PR #1587, opened 2026-03-23
Fix/feishu token workspace persistence

332891938:
## Summary

-

## Change Type

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] CI/Infrastructure
- [ ] Security
- [ ] Dependencies

## Validation

- [ ] `cargo fmt`
- [ ] Relevant tests pass: 
- [ ] Manual testing: 

## Rollback Plan

---

**Review track**:


src/channels/wasm/setup.rs
```diff
@@ -190,7 +190,13 @@ async fn register_channel(
         // The credential injection system only replaces placeholders in URLs
         // and headers, so channels like Feishu that exchange app_id + app_secret
         // for a tenant token need the raw values in their config.
-        inject_channel_secrets_into_config(&channel_name, secrets_store, &mut config_updates).await;
+        inject_channel_secrets_into_config(
```
serrrfirat, line 193: **High Severity**

This fixes Feishu secret lookup during cold startup by threading `config.owner_id` into `inject_channel_secrets_into_config()`, but the same Feishu-specific config injection is still missing from the hot-activation and credential-refresh paths in `ExtensionManager`. As a result, configuring or refreshing the Feishu channel in a running app will still rerun `on_start` without `app_id` / `app_secret`, so tenant token exchange and outbound replies remain broken until a full restart.

Please extract this channel-specific config injection into a shared helper and call it from startup, hot activation, and refresh before `on_start`, then add a regression test for activate/refresh without restart.

src/channels/wasm/wrapper.rs
```diff
@@ -1562,7 +1562,7 @@ impl WasmChannel {
 
         let runtime = Arc::clone(&self.runtime);
         let prepared = Arc::clone(&self.prepared);
-        let capabilities = self.capabilities.clone();
+        let capabilities = Self::inject_workspace_reader(&self.capabilities, &self.workspace_store);
```
serrrfirat, line 1565: **Medium Severity**

This change adds the runtime fix, but there is still no regression coverage for the behavior it is trying to preserve: non-default-owner Feishu secret injection, callback-time workspace reads after `on_start`, or the hot-activation/refresh path. The current tests only cover token-response deserialization, so the startup/activation regressions here would be easy to reintroduce.

Please add a targeted test for non-default owner Feishu setup and a lifecycle-style callback test that proves state written in `on_start` is readable from `on_respond`/`on_broadcast`.

registry/channels/feishu.json
```diff
@@ -2,7 +2,7 @@
   "name": "feishu",
   "display_name": "Feishu / Lark Channel",
   "kind": "channel",
-  "version": "0.1.1",
+  "version": "0.1.2",
```
serrrfirat, line 5: **High Severity**

The manifest version is bumped to `0.1.2`, but the artifact block still points at the old `channel-feishu-0.1.1` release asset and checksum. Registry installs/upgrades will therefore continue downloading the pre-fix WASM bundle, so users never actually receive the workspace-persistence fix from this PR.

Please publish a `0.1.2` artifact and update both `url` and `sha256`, or defer the manifest version bump until that asset exists.

ilblackdragon (changes requested):
## Review: fix/feishu token workspace persistence

### Must-fix

1. **`call_on_broadcast` not fixed** — uses `self.capabilities.clone()` without `inject_workspace_reader` and doesn't call `commit_writes`. Feishu's `on_broadcast` calls `send_message()` → `get_valid_token()` → workspace read, so this is a live bug for proactive messaging.

2. **`call_on_status` not fixed** — same pattern, same omission.

3. **Hot-activation path missing** — `ExtensionManager::complete_loaded_wasm_channel_activation()` does not call `inject_channel_secrets_into_config()`, so runtime-activated Feishu channels still lack `app_id`/`app_secret`.

4. **No regression tests** — violates the project's mandatory regression test policy. Needed: unit test for `inject_channel_secrets_into_config` using `owner_id`, lifecycle test for workspace data round-trip between `on_start` and `on_respond`.

5. **Excessive diagnostic logging** — 6 verbose logging blocks read workspace state purely for diagnostics, doubling workspace I/O in `obtain_tenant_token`. Remove or gate behind Trace level.

### Should-fix

6. **Empty `owner_id` not handled** — if `config.owner_id` is empty string, `secrets.get_decrypted("", ...)` returns no results with no fallback.

7. **Unrelated changes bundled** — CHANGELOG, Cargo version bumps, Telegram registry update should be in separate PRs.

8. **Registry manifest inconsistency** — `feishu.json` version bumped to `0.1.2` but artifacts block still points to `0.1.1`.

9. **PR template is empty** — no summary, validation, security impact, or rollback plan filled in.

registry/channels/feishu.json
ilblackdragon, line 5: **Critical — registry artifact URL/sha256 not updated with version bump.**

The manifest version bumped from `0.1.1` to `0.1.2`, but the `artifacts.wasm32-wasip2.url` still points to `channel-feishu-0.1.1-wasm32-wasip2.tar.gz` and `sha256` is unchanged. Anyone who installs feishu from the registry will still download the pre-fix 0.1.1 WASM, so the persistence / owner-scope fix will not reach registry users — only dev-mode installs (which rebuild from `channels-src/feishu`) see the new code.

Also note `channels-src/feishu/Cargo.toml` is still `version = "0.1.0"`. Either:
1. Revert the manifest version bump until the 0.1.2 release artifact is actually published and update url+sha256 in the same PR, or
2. Publish the release artifact and update all three (Cargo.toml, registry url, registry sha256) together.

src/channels/wasm/setup.rs
```diff
@@ -413,19 +420,28 @@ async fn inject_channel_secrets_into_config(
     };
 
     for &(config_key, secret_name) in secret_config_mappings {
-        match secrets.get_decrypted("default", secret_name).await {
+        match secrets.get_decrypted(owner_id, secret_name).await {
```
ilblackdragon, line 423: **Critical — missing regression test for the owner-scope fix (Test Through the Caller rule).**

The signature change from `secrets.get_decrypted("default", secret_name)` to `secrets.get_decrypted(owner_id, secret_name)` is the whole point of the PR, but the test suite does not cover it. A unit test that only parses JSON (as in `channels-src/feishu/src/lib.rs` tests) will never notice if a future refactor reverts `owner_id` back to a hardcoded scope, or if `register_channel` passes the wrong string through.

Please add a test that drives the caller (`register_channel` or at minimum `inject_channel_secrets_into_config`) with two different owner scopes seeded in an `InMemorySecretsStore` and asserts that the returned config_updates contain the owner-specific value, not the "default" value. Pattern example already exists in staging for the post-PR version — reuse that.

Without this, the same bug that motivated the PR (secrets looked up under the wrong scope) can silently regress and the JSON-parse-only tests will still pass.

src/channels/wasm/setup.rs
```diff
@@ -413,19 +420,28 @@ async fn inject_channel_secrets_into_config(
...
                     config_key = %config_key,
                     "Injected secret into channel config"
                 );
             }
-            Err(_) => {
+            Err(e) => {
+                tracing::debug!(
+                    channel = %channel_name,
+                    owner_id = %owner_id,
+                    secret_name = %secret_name,
```
ilblackdragon, line 441: **High — secret name and config-key diagnostic labels in `tracing::debug!`.**

The new debug log fields `secret_name`, `config_key`, and `env_name` are safe (they're identifiers, not values), but the combination of `owner_id` + `secret_name` + `error = %e` with the real decryption error may leak information useful to an attacker with log access (e.g., whether a specific owner has a specific secret configured, or distinguishes "wrong key" vs. "not found" vs. "decode error"). Consider:

1. Logging `owner_id` only at `trace!` level, not `debug!` — most prod deployments ship debug logs to aggregators.
2. Redacting or bucketing the decryption error to a coarse category (`not_found`, `decrypt_failed`) rather than including the raw `e` which for AES-GCM errors can reveal structural information.

This is not a new secret leak (values never hit logs), but per `.claude/rules/review-discipline.md` ("Sensitive data in logs & events") the owner/secret correlation and error shape are worth tightening before shipping.

channels-src/feishu/src/lib.rs
```diff
@@ -740,6 +814,22 @@ fn get_valid_token(api_base: &str) -> Result<String, String> {
 /// Reads credentials from workspace storage (persisted during `on_start`
```
ilblackdragon, line 814: **High — concurrent token-refresh race.**

`get_valid_token` (unchanged, but newly relied-on by persistent `on_respond`/`on_broadcast` callbacks in this PR) has no lock or single-flight around `obtain_tenant_token`. When `on_respond` and `on_broadcast` fire concurrently (very common — agent replies while a broadcast goes out) AND the cached token is within the 5-minute refresh window, both callbacks will:

1. Read the expiring token from workspace (stale hit fails, they fall through).
2. Both POST to `/open-apis/auth/v3/tenant_access_token/internal`.
3. Both write `TOKEN_PATH` / `TOKEN_EXPIRY_PATH` last-writer-wins.

Feishu rate-limits this endpoint (and returns 429 or an error code on burst). Under load you will see periodic auth failures masked as "Token exchange request failed".

Mitigations (pick one):
- Serialize `obtain_tenant_token` via a host-side mutex keyed by channel name. Since writes only land at callback commit, an in-callback lock won't help — this needs a host-level cross-callback serialization.
- Accept the race but add a small jitter + retry path for the known 429 response code.

Not a PR blocker if the PR explicitly scopes to persistence only, but please file a follow-up issue.

channels-src/feishu/src/lib.rs
```diff
@@ -286,37 +286,96 @@ impl Guest for FeishuChannel {
...
         // These are injected by the host from the secrets store into the
         // config JSON (see setup.rs inject_channel_secrets_into_config).
         if let Some(ref app_id) = config.app_id {
-            let _ = channel_host::workspace_write(APP_ID_PATH, app_id);
+            if let Err(e) = channel_host::workspace_write(APP_ID_PATH, app_id) {
+                channel_host::log(
+                    channel_host::LogLevel::Warn,
+                    &format!("workspace_write failed for {}: {}", APP_ID_PATH, e),
+                );
+            }
```
ilblackdragon, line 322: **Medium — error return ignored on workspace_write but log-noise only.**

All the `if let Err(e) = channel_host::workspace_write(...)` arms log `Warn` and proceed. That's correct (`on_start` must still succeed so the channel can register its webhook), but if the early `APP_ID_PATH` or `APP_SECRET_PATH` write fails, the subsequent `obtain_tenant_token` at line 792 will fail with "app_id not configured (missing from workspace)" — a confusing error that masks the real cause.

Consider:
- If `APP_ID_PATH` or `APP_SECRET_PATH` persist fails, return a hard error from `on_start` rather than logging and continuing. A channel that cannot persist its auth credentials is not usable; better to fail fast.
- Or: propagate the underlying workspace_write failure message into the `obtain_tenant_token` error so operators can diagnose.

ilblackdragon:
## Architectural review — `fix/feishu-token-workspace-persistence`

Reviewed at commit `13bda947`.

### Summary
The PR fixes two real bugs:
1. `inject_channel_secrets_into_config` looked up Feishu credentials under the hardcoded `"default"` scope while the rest of `register_channel` uses `config.owner_id`. Moving to `owner_id` aligns with the instance-level ownership model documented at the top of `src/channels/wasm/setup.rs`.
2. WASM callbacks `on_respond`, `on_broadcast`, `on_status`, and the typing repeater in `handle_status_update` were built with bare `self.capabilities.clone()` which omitted the workspace reader, so `channel_host::workspace_read(...)` inside those callbacks returned `None`. Feishu relied on that to look up `app_id` / `app_secret` / `api_base` during token refresh, so outbound replies broke. Switching all four to `Self::inject_workspace_reader(&self.capabilities, &self.workspace_store)` fixes it.

Both fixes match their stated cause.

### Architectural concerns

1. **Release-train hygiene.** The registry manifest version was bumped (`0.1.1` -> `0.1.2`), the artifact URL and sha256 were not, and `channels-src/feishu/Cargo.toml` still reads `0.1.0`. Either bump all three in lockstep or hold the manifest bump until the release artifact exists. See inline comment.

2. **CHANGELOG drift.** The PR appends `## [0.21.0]` and `## [0.20.0]` entries, but `staging` is at `0.25.0` and those releases already exist. This appears to be stale cherry-picked content from a rebase. Drop the CHANGELOG diff and let the normal release process own version bumps.

3. **Test Through the Caller.** The signature change to `inject_channel_secrets_into_config(..., owner_id, ...)` is the PR's core correctness change but there is zero test covering it. Per `.claude/rules/testing.md` this is exactly the helper-with-wrapper pattern that needs a caller-level regression test. See inline comment on `setup.rs`.

4. **Concurrent token refresh.** Now that `on_respond` and `on_broadcast` both reliably hit `get_valid_token` and can fire concurrently, the lack of single-flight around `obtain_tenant_token` becomes load-bearing. See inline comment on `lib.rs`.

### Non-blocking observations

- `workspace_write` error handling in `on_start` now logs but proceeds. For `APP_ID_PATH` / `APP_SECRET_PATH` specifically, a persist failure produces a misleading downstream error ("not configured") and is worth surfacing explicitly.
- The `on_respond` and `obtain_tenant_token` debug snapshots are useful and intentionally log only `has_app_id=bool`, not values — good.
- `ChannelWorkspaceStore` is in-process memory only (`std::sync::RwLock<HashMap>`), so `app_secret` is not persisted to disk, but it does live in channel-host memory for the life of the process. Not a regression from this PR, but worth documenting as a known limitation of the current WASM channel storage model.

### Verdict
Changes-requested. The correctness intent is right and the wrapper.rs fan-out is thorough, but the release-manifest inconsistency and missing regression test should block merge. The token-refresh race and CHANGELOG drift can ship as follow-ups.

serrrfirat:

This patch persists Feishu tokens through the legacy WASM channel workspace. Reborn adapters may not own credential persistence; authentication and renewal belong to host product-auth/credential services, while adapters remain parse/render boundaries.

A Feishu Reborn port should use a ProductAdapter plus host-owned long-connection transport and encrypted credential account lifecycle, with token refresh tests and no guest workspace token storage.

serrrfirat:

This stores Feishu tokens in legacy WASM channel workspace. Reborn adapters cannot own credentials; host product-auth/credential services own authentication and renewal.

A Feishu port should use a ProductAdapter plus host long-connection transport and encrypted account lifecycle, with refresh and fake-server tests.

PR #1565, opened 2026-03-22
fix(gemini): preserve thought signatures on all tool calls

G7CNF:
Fixes Gemini tool-call failures on 3.x models by ensuring every functionCall part in the active loop carries a thoughtSignature before the request is sent.\n\nChanges:\n- annotate all functionCall parts in the active Gemini loop\n- add a regression test for multi-function-call turns\n\nValidation:\n- cargo fmt --all -- --check\n- cargo test -p ironclaw test_ensure_thought_signatures_adds_signatures_to_all_function_calls -- --nocapture\n- cargo clippy -p ironclaw --lib --all-features -- -D warnings

Issue #1510, opened 2026-03-21
BUG: Gemini: function call is missing a thought_signature in functionCall parts.


nthomsencph:
* Backend: gemini
* Model: gemini-3.1-flash-lite-preview
* ironclaw 0.21.0

When the agent uses a tool, it fails with the following error:

"""
Error: LLM error: Provider gemini-3.1-flash-lite-preview request failed: HttpError: Invalid status code 400 Bad Request with message: [{ "error": { "code": 400, "message": "Function call is missing a thought_signature in functionCall parts. This is required for tools to work correctly, and missing thought_signature may lead to degraded model performance. Additional data, function call default_api:memory_write , position 16. Please refer to https://ai.google.dev/gemini-api/docs/thought-signatures for more details.", "status": "INVALID_ARGUMENT" } } ]
"""

Logs: 
``` bash

2026-03-21T00:47:12.623948Z  INFO Processing message message_id=69e1050e-7d08-407b-b70d-8db97206a07a

[...]

2026-03-21T00:47:19.349824Z  WARN Retrying after transient error (tools) provider=gemini-3.1-flash-lite-preview attempt=3 max_retries=3 delay_ms=4816 error=Provider gemini-3.1-flash-lite-preview request failed: HttpError: Invalid status code 400 Bad Request with message: [{
  "error": {
    "code": 400,
    "message": "Function call is missing a thought_signature in functionCall parts. This is requir...[809B total]
```

https://ai.google.dev/gemini-api/docs/thought-signatures

nthomsencph:
A task branch has been created for this issue:

```
fix-gemini-missing-thought-signature-8b4
```

PR #1588, opened 2026-03-23
fix: preserve Gemini thought_signature through tool call round-trips

nthomsencph:
- Gemini thinking models attach a `thought_signature` to function call objects that must be echoed back in subsequent requests. IronClaw dropped this field, causing 400 errors.
- Added a generic `extra` field (`serde_json::Map` with `#[serde(flatten)]`) to `ToolCall` that transparently preserves any provider-specific opaque fields through the response→request round-trip.
- Only the outermost serialization layer (`nearai_chat.rs`) maps `thought_signature` to/from `extra` — the core type stays provider-agnostic.

### Generic `extra` instead of a specific `thought_signature` field

Different providers have different tool call signatures and IronClaw could constantly need to update the core `ToolCall` type to keep track. The `extra` field lets us support variation in signatures through the round-trip and only unpack at the last layer (the provider-specific code).

Adds a wew round-trip test too: deserialize Gemini response with `thought_signature` → convert to `ToolCall` → store in `ChatMessage` → convert back → verify `thought_signature` is present in serialized output

- [x] `cargo fmt` — clean
- [x] `cargo test` — 3290+ tests pass


zmanian (commented):
Thanks for the well-structured PR, @nthomsencph -- welcome to the project! The bug report is clear, the approach is thoughtful, and the round-trip test is exactly the kind of coverage we want. A few things to address before merge:

---

## Must-fix

### 1. Remove `.claude/task-context.json`

This file is a local Claude Code session artifact and should not be checked in. Please remove it from the PR (and consider adding it to `.gitignore` if it's not already there).

### 2. `#[serde(flatten)]` on `ToolCall::extra` -- serialization hazard

Using `#[serde(flatten)]` with a catch-all `Map` on the **core** `ToolCall` type means that any unknown field in a JSON blob will silently land in `extra` during deserialization, and conversely any key in `extra` gets promoted to a top-level field on serialization. This is fine for a wire type, but `ToolCall` is also the **internal** representation that flows through the agent loop, gets stored in DB rows, and is serialized by every provider adapter.

Concrete risks:
- If a ToolCall is serialized to the DB (e.g. `thread_ops.rs` stores conversation history as JSON), the `extra` fields leak into the stored JSON. A future field name collision between `extra` keys and real `ToolCall` fields (`id`, `name`, `arguments`) would cause deserialization failures.
- Other providers' adapters that round-trip through JSON (Anthropic, Bedrock, rig) would silently swallow and re-emit unknown fields, which may confuse those APIs.

**Suggestion:** Keep `extra` but use `#[serde(default, skip_serializing_if = "extra_is_empty")]` (without `flatten`). This stores it as an explicit `"extra": {...}` key in JSON, which is safer and more explicit. The only place that needs flatten behavior is the wire format in `nearai_chat.rs`, where `ChatCompletionToolCall` already has `extra_content` as an explicit field -- that's the right layer for it.

### 3. `DIRECT_CHAT_COMPLETIONS_PROVIDERS` routing bypass -- too broad

Force-routing all Gemini providers through `NearAiChatProvider` (bypassing rig-core) is a significant behavioral change. `NearAiChatProvider` was purpose-built for NearAI's specific auth and message-flattening needs. Repurposing it as a generic "direct Chat Completions" path has several concerns:

- **Auth model mismatch**: The `new_with_options` path uses a raw API key, but NearAiChatProvider's auth flow (bearer token) may not match what Google's endpoint expects (typically `x-goog-api-key` header or OAuth). Does this actually work against the Gemini OpenAI-compat endpoint?
- **NearAI-specific behaviors**: Message flattening, the `/chat/completions` path assumption, bearer auth header -- these are NearAI-isms that may cause subtle issues with Google's endpoint.
- **Scope creep**: This routing change affects *all* Gemini users, not just those hitting the thought_signature issue. If something breaks, it's a wide blast radius.

**Suggestion:** The `extra` field preservation on `ToolCall` is the right core fix. But the routing change should either (a) be dropped entirely if `GeminiOauthProvider` already handles thought signatures (see overlap with #1565 below), or (b) be a separate, well-tested PR with explicit integration testing against Google's endpoint.

---

## Overlap with PR #1565

PR #1565 from @G7CNF fixes the same issue (#1510) but takes a completely different approach: it patches `GeminiOauthProvider::ensure_thought_signatures()` in `gemini_oauth.rs` to annotate *all* functionCall parts (not just the first one). That fix is minimal (1 file, ~35 lines) and stays within the Gemini-specific provider.

These two PRs are **complementary, not conflicting**:
- #1565 fixes the native Gemini API path (`gemini_oauth.rs`)
- #1588 fixes the OpenAI-compat path (where Gemini is accessed via `nearai_chat.rs`)

But the routing change in #1588 (the `DIRECT_CHAT_COMPLETIONS_PROVIDERS` block) creates an **artificial conflict** -- it would bypass `GeminiOauthProvider` entirely for registry-configured Gemini providers, making #1565's fix unreachable for those users.

**Recommendation:** Strip the routing change from this PR. Keep the `extra` field on `ToolCall` + the `extra_content` handling in `nearai_chat.rs`. Let #1565 handle the native Gemini path.

---

## Merge conflict forecast with PRs #1590-1592

This PR touches `dispatcher.rs`, `session.rs`, and `thread_ops.rs` -- all in test code only (adding `extra: Default::default()` to struct literals). PRs #1590-1592 modify production code in these same files but mostly different lines. Conflicts will be minor (test struct literals) and easy to resolve. Whoever merges second just needs to add `extra: Default::default()` to any new ToolCall literals introduced by the other PRs.

---

## Minor nits

- The `extra_is_empty` helper function could be replaced with `serde_json::Map::is_empty` directly: `#[serde(skip_serializing_if = "serde_json::Map::is_empty")]` -- no need for the wrapper function. (Only relevant if you keep `flatten`; if you switch to a named field, skip_serializing_if still applies.)
- The boilerplate of adding `extra: Default::default()` to ~30 construction sites is unavoidable with a new field, but consider deriving `Default` for `ToolCall` or adding a constructor like `ToolCall::new(id, name, arguments)` to reduce churn in future PRs.

---

## Summary

The core idea -- a generic `extra` bag on `ToolCall` to preserve provider-specific opaque fields -- is sound and forward-looking. The test coverage for the round-trip is good. But the PR needs:

1. Remove the task-context.json file
2. Change `extra` from `flatten` to a named JSON field for safety
3. Remove or separate the `DIRECT_CHAT_COMPLETIONS_PROVIDERS` routing change
4. Coordinate with #1565 on which paths each PR covers

Happy to re-review once these are addressed. Thanks again for jumping in on this!

ilblackdragon (changes requested):
## Additional Review Findings

Beyond zmanian's existing feedback (remove task-context.json, flatten hazard, routing bypass, overlap with #1565):

### Must-fix

1. **PR #1565 is already merged** (commit `86389dab`) fixing the same issue via `ensure_thought_signatures()` in `gemini_oauth.rs`. This PR is now in `CONFLICTING` state. The `DIRECT_CHAT_COMPLETIONS_PROVIDERS` routing change would **bypass** #1565's fix for registry-configured Gemini providers.

2. **Data loss on thread rebuild** — `session.rs:384` and `thread_ops.rs:1755` both reconstruct `ToolCall` with `extra: Default::default()`, silently dropping `extra` data. Any persisted/rebuilt session loses the thought signature, causing failures on subsequent Gemini calls.

### Recommendation

Strip the routing change entirely. The `extra` field on `ToolCall` + `extra_content` handling in `nearai_chat.rs` remain valuable as a complementary fix for the OpenAI-compat path, but must be rebased on top of merged #1565 and address the data persistence gap.

src/agent/session.rs
```diff
@@ -381,6 +381,7 @@ impl Thread {
                         id: format!("turn{}_{}", turn.turn_number, i),
                         name: tc.name.clone(),
                         arguments: tc.parameters.clone(),
+                        extra: Default::default(),
```
ilblackdragon, line 384: **Critical: thought_signature is lost on thread rehydration from DB.**

This is the same class of bug the PR is trying to fix, just one persistence layer deeper. `TurnToolCall` (`src/agent/session.rs:846`) has no `extra` / `thought_signature` field, and both rebuild paths explicitly drop it:

- `Thread::restore_from_messages` → here constructs `ToolCall { …, extra: Default::default() }` (line 384), and earlier `record_tool_call_with_reasoning` never receives the signature from `tc` in the first place.
- `rebuild_chat_messages_from_db` in `thread_ops.rs:1755` likewise hardcodes `extra: Default::default()`.

Effect: the happy-path in-memory round-trip works (and passes the new test), but on any process restart, job resume, or `/compact`/checkpoint restore that flows through the DB, the very next Gemini tool-call turn will fail with the same 400 this PR is fixing. This also directly violates the `CLAUDE.md` invariant "LLM data is never deleted" — `thought_signature` is LLM output and must persist end-to-end.

Minimum fix: plumb the signature through `TurnToolCall`, `persist_tool_calls` (`thread_ops.rs:914`), and both rebuild sites. Add an integration-tier regression test (`tests/*_integration.rs`) that drives persistence → rebuild → re-serialize to Gemini wire format and asserts the signature is echoed back — a unit test on the helper doesn't catch this per `.claude/rules/testing.md` ("Test Through the Caller, Not Just the Helper").

src/agent/thread_ops.rs
```diff
@@ -1752,6 +1752,7 @@ fn rebuild_chat_messages_from_db(
                                     .get("parameters")
                                     .cloned()
                                     .unwrap_or(serde_json::json!({})),
+                                extra: Default::default(),
```
ilblackdragon, line 1755: **High: same drop happens here — DB → ChatMessage rebuild strips `extra`.**

`rebuild_chat_messages_from_db` reconstructs `ToolCall` from `role="tool_calls"` DB rows with `extra: Default::default()`, and the matching `persist_tool_calls` writer (`thread_ops.rs:914`) doesn't serialize `extra_content` / `thought_signature` in the first place (only `name`, `call_id`, `result_preview`, `result`, `error`, `rationale`, `tool_call_id`).

So: thread hydrates from DB on restart → assistant turn that previously called a Gemini tool → LLM sees `tool_calls` with no signature → next request gets 400 INVALID_ARGUMENT. The in-memory test added here will not catch this; please add a test that drives `persist_tool_calls` + `rebuild_chat_messages_from_db` with a signature-bearing call and asserts the signature survives.

src/llm/mod.rs
```diff
@@ -140,6 +140,14 @@ pub fn create_llm_provider_with_config(
 /// Dispatches on `RegistryProviderConfig::protocol` to build the appropriate
 /// rig-core client. This single function replaces what used to be 5 separate
 /// `create_*_provider` functions.
+/// Providers that need direct Chat Completions handling instead of rig-core.
+///
+/// rig-core's OpenAI adapter strips provider-specific fields from tool call
+/// objects (e.g. Gemini's `thought_signature`). These providers are routed
+/// through `NearAiChatProvider` which preserves such fields through the
```
ilblackdragon, line 147: **Medium: hardcoded provider-id list duplicates `providers.json` aliases.**

`DIRECT_CHAT_COMPLETIONS_PROVIDERS` hardcodes `["gemini", "google_gemini", "google"]`, which happens to mirror today's `providers.json` entry:
```
{ "id": "gemini", "aliases": ["google_gemini", "google"], … }
```
If an operator adds a new alias in `providers.json` (or a registry config uses a different canonical id), this branch silently won't apply and the Gemini thought_signature fix will just stop working with no error — the request will fail back to the rig-core path that strips the field.

Safer options:
- Dispatch on the `protocol` + a capability flag on `RegistryProviderConfig` (e.g. `preserves_tool_call_extras: bool`), set once in the registry manifest.
- Or resolve the canonical id through the providers catalog (aliases → canonical) before comparing.

Either avoids having two sources of truth for "this provider is Gemini".

ilblackdragon:
**Architectural: branch base is stale; main now has a parallel `thought_signature` mechanism.**

Two observations from reviewing this against `main` (current SHA `2917f1b4`) rather than the PR's base (`bd6977e6`, v0.21.0):

1. **Stale base.** Between the PR base and main, `ToolCall` gained a `reasoning: Option<String>` field and `src/llm/provider.rs` gained `generate_tool_call_id`, `take_model_override`, `normalized_model_override`, plus their tests. The diff here will require a non-trivial rebase — `#[serde(flatten)] extra` needs to land alongside the typed `reasoning` field, not replace it. Please rebase onto main and re-verify the serde-flatten tests.

2. **Competing mechanism already exists.** Main has `src/llm/gemini_oauth.rs`, which solves the same problem a different way: a provider-local `Mutex<HashMap<call_id, signature>>` side-table (`thought_signatures`) that `ensure_thought_signatures()` re-injects into outgoing requests. After rebase, IronClaw will have **two** `thought_signature` pipelines:
   - **gemini_oauth**: provider-side side-table, works only for `backend=gemini_oauth`.
   - **This PR**: field-on-ToolCall, works only for registry-configured `provider_id == gemini" via direct Chat Completions.

These don't interfere at runtime (different provider entry points), but the field-on-ToolCall approach is strictly more general — it persists through history/compaction/DB (once the DB-strip concerns in my inline comments are fixed) — whereas the side-table resets on process restart. I'd suggest picking the `extra` approach as the single source of truth and making `gemini_oauth` read/write through it, so there is one story for how opaque provider fields survive IronClaw's message pipeline.

Also worth calling out for reviewers: the PR bundles `.claude/task-context.json` into the diff — that's a local Claude Code session file, should be added to `.gitignore` rather than committed.

serrrfirat:
Thank you for fixing Gemini thought-signature round trips. We are closing this PR because Reborn now preserves this provider state through its current LLM contracts.

The active implementation in `crates/ironclaw_llm` preserves opaque provider reasoning data on tool calls. Both the native Gemini registry provider and `gemini_oauth` capture and replay `thought_signature`/`thoughtSignature`; the latter has focused response-to-request round-trip tests in `gemini_oauth.rs`. This is more general than forwarding the v1 patch because it survives the Reborn runner and transcript boundaries.

This correctness requirement remains important. We would be very happy to have you contribute additional Reborn Gemini fixtures or cross-turn tool-call tests. Thank you for diagnosing the underlying 400 response and designing a provider-compatible solution.

PR #1590, opened 2026-03-23
fix(security): block cross-channel approval thread hijacking

### LINKED ISSUE #1485: [CRITICAL] Cross-channel approval thread hijacking - authorization bypass
## [CRITICAL:100] Issue Found by Staging CI Review

**Severity:** CRITICAL
**Confidence:** 100/100
**PR comment:** https://github.com/nearai/ironclaw/pull/1483#issuecomment-4100594194

### Description
Cross-channel approval thread hijacking - authorization bypass

---

zmanian:
## Summary

- Add `source_channel` field to `Thread` struct to track which channel created each thread
- Verify channel authorization before allowing approval messages to target threads by UUID
- Web gateway channel allowed as trusted approval UI via `TRUSTED_APPROVAL_CHANNELS` constant
- WASM channels blocked from claiming reserved/trusted channel names (including `__bootstrap__` sentinel)
- Backward compatible: threads without `source_channel` (older DB records) denied by default (fail-closed)
- `TenantScope` correctly passes channel as `source_channel`

**Attack blocked:** Attacker on one channel could approve tool calls pending on threads from a different channel by supplying the target UUID.

## Files changed
- `src/agent/session.rs` -- `source_channel` field, `TRUSTED_APPROVAL_CHANNELS` constant, `is_approval_authorized()`, updated constructors
- `src/agent/agent_loop.rs` -- authorization check in approval routing, pending_approval gate
- `src/agent/session_manager.rs` -- pass channel to `create_thread()`
- `src/agent/thread_ops.rs` -- hydration from DB source_channel and `/new` paths
- `src/agent/compaction.rs`, `dispatcher.rs` -- test callers
- `src/channels/wasm/setup.rs` -- reserved channel name validation using shared constants
- `src/channels/web/handlers/chat.rs`, `server.rs` -- pass channel
- `src/db/mod.rs`, `postgres.rs`, `libsql/conversations.rs`, `history/store.rs` -- `source_channel` column, `get_conversation_source_channel()`
- `src/db/libsql_migrations.rs` -- V16 migration
- `migrations/V15__conversation_source_channel.sql` -- PostgreSQL migration
- `src/tenant.rs` -- pass `Some(channel)` as source_channel

## Test plan
- [x] 100 session tests pass (15 new: authorization logic, pending_approval guard, serde backcompat, channel propagation)
- [x] 30 session_manager tests pass (1 new: source_channel storage)
- [x] 4 WASM setup tests (new: reserved name validation, case-insensitive, bootstrap sentinel)
- [x] 7 libSQL conversation tests (4 new: source_channel round-trip, None round-trip, not-found, upsert invariant)
- [x] Clippy clean (all-features, default, libsql-only)
- [x] CI green

Closes #1485

Supersedes #1582


src/agent/thread_ops.rs
```diff
@@ -141,7 +141,8 @@ impl Agent {
             sess.id
         };
 
-        let mut thread = crate::agent::session::Thread::with_id(thread_uuid, session_id);
+        let mut thread =
+            crate::agent::session::Thread::with_id(thread_uuid, session_id, Some(&message.channel));
```
serrrfirat: **Critical Severity**

**Hydration overwrites `source_channel` with the attacker's channel, defeating the authorization check.**

`maybe_hydrate_thread()` runs *before* the approval authorization check in `agent_loop.rs:1041`. When a thread isn't in memory (e.g. after server restart), hydration creates it here with `source_channel = Some(&message.channel)` — the *requesting* channel, not the original creator's channel.

**Attack scenario:**
1. User creates a thread on channel `"cli"` — a tool call is pending approval
2. Server restarts (or thread is evicted from memory)
3. Attacker (same user, different channel `"telegram"`) sends an approval targeting the thread UUID
4. `maybe_hydrate_thread()` runs first, creates the thread with `source_channel = Some("telegram")`
5. The authorization check sees `source_channel = "telegram"` matching `message.channel = "telegram"` → authorized
6. Attacker approves the tool call from an unauthorized channel

**Suggested fix:** Hydration should read `source_channel` from the DB (e.g. from `conversations` metadata) instead of stamping the requester's channel. Alternatively, persist `source_channel` to the database so hydration restores the original value, or run the authorization check *before* hydration for approval submissions.

src/agent/session.rs
```diff
@@ -225,6 +225,9 @@ pub struct Thread {
     /// Messages queued while the thread was processing a turn.
     #[serde(default, skip_serializing_if = "VecDeque::is_empty")]
     pub pending_messages: VecDeque<String>,
+    /// Channel that created this thread (for approval authorization).
+    #[serde(default)]
+    pub source_channel: Option<String>,
```
serrrfirat, line 230: **High Severity**

**`source_channel` is not persisted to the database — server restart clears all authorization data.**

The `source_channel` field exists only on the in-memory `Thread` struct. There are zero references to `source_channel` in `src/db/`. After a server restart, all threads lose their `source_channel` and are re-hydrated with `source_channel = None` (either via `maybe_hydrate_thread` stamping the requester's channel per finding #1, or defaulting to `None` via `#[serde(default)]`).

Since `is_none_or(...)` returns `true` when `source_channel` is `None`, the entire authorization check becomes a no-op after any restart. The security guarantee only holds while the process is running and the thread remains in memory.

**Suggested fix:** Add a `source_channel` column to the `conversations` table in both PostgreSQL (`postgres.rs`) and libSQL (`libsql_backend.rs`) backends per CLAUDE.md dual-backend rules. Persist it on thread/conversation creation and restore it during DB hydration.

src/agent/agent_loop.rs
```diff
@@ -1036,7 +1038,24 @@ impl Agent {
                 .get_or_create_session(&message.user_id)
                 .await;
             let mut sess = session.lock().await;
-            if sess.threads.contains_key(&target_thread_id) {
+            if let Some(thread) = sess.threads.get(&target_thread_id) {
+                let authorized = thread.source_channel.as_ref().is_none_or(|src| {
+                    src == &message.channel
+                        || message.channel == "web"
+                        || message.channel == "gateway"
+                });
```
serrrfirat: **Medium Severity**

**Channel allowlist uses string literals `"web"` and `"gateway"` — WASM channels could register with these names.**

The authorization check hardcodes `message.channel == "web" || message.channel == "gateway"` as trusted channels that can approve any thread regardless of origin. WASM channel names come from the module's `name()` method (`src/channels/wasm/setup.rs`), and there is no validation preventing a WASM channel from registering as `"web"` or `"gateway"`.

A malicious or misconfigured WASM channel module named `"web"` would bypass all cross-channel approval checks for every thread.

**Suggested fix:** Either validate WASM channel names against reserved names (`"web"`, `"gateway"`, `"cli"`, `"repl"`) during `setup_wasm_channels()`, or use a separate mechanism (e.g. a trait method or flag) to mark trusted UI channels rather than relying on string name matching.

src/agent/agent_loop.rs
```diff
@@ -1036,7 +1038,24 @@ impl Agent {
                 .get_or_create_session(&message.user_id)
                 .await;
             let mut sess = session.lock().await;
-            if sess.threads.contains_key(&target_thread_id) {
+            if let Some(thread) = sess.threads.get(&target_thread_id) {
+                let authorized = thread.source_channel.as_ref().is_none_or(|src| {
```
serrrfirat: **Medium Severity**

**Authorization check doesn't verify the thread actually has a pending approval.**

The check verifies that the requesting channel matches the thread's source channel, but it doesn't verify that `thread.pending_approval.is_some()`. After authorization succeeds, the code unconditionally sets `sess.active_thread = Some(target_thread_id)` and registers the thread — even if there's nothing to approve.

This means an approval-shaped message can be used to switch the active thread across channels as long as the source channel matches or is `None`. While not directly exploitable for unauthorized approvals (the actual approval processing would find no pending approval), it's a thread-switching side effect that bypasses the normal `resolve_thread` path.

**Suggested fix:** Add `&& thread.pending_approval.is_some()` to the authorization guard, or at least before performing the active thread switch. If there's no pending approval, fall through to the normal `resolve_thread` path.

src/agent/agent_loop.rs
```diff
@@ -1036,7 +1038,24 @@ impl Agent {
...
+                    src == &message.channel
```
serrrfirat: **Medium Severity**

**`is_none_or` makes `None` a universal pass — fail-open by default.**

Bootstrap threads are explicitly created with `source_channel = None` (line 735), and the comment documents this as intentional so "approvals from any channel are permitted." Combined with the fact that `source_channel` is not persisted to the DB (see other comment), after a server restart *all* threads get `source_channel = None`, making the authorization check a no-op for the entire system.

The fail-open design means any gap in `source_channel` propagation silently disables the security check rather than blocking the request.

**Suggested fix:** Consider using a sentinel value like `source_channel = Some("__bootstrap__")` for bootstrap threads and explicitly listing it in the allowlist. This way `None` can be treated as "deny by default" (fail-closed), and any missing `source_channel` is a safe denial rather than a silent bypass.

src/agent/session.rs
```diff
@@ -1620,4 +1625,34 @@ mod tests {
         let merged = thread.drain_pending_messages().unwrap();
         assert_eq!(merged, "failed batch\nnew msg");
     }
+
```
serrrfirat, line 1748: **Low Severity**

**No test for the actual cross-channel authorization check.**

The PR adds 4 tests for `source_channel` storage and serde backcompat, but no test covers the core security logic — the authorization predicate in `agent_loop.rs:1041–1058`. The authorization check is the entire point of this PR, but it's untested.

**Suggested fix:** Add tests that verify:
1. Same channel → approval authorized
2. Different channel → approval blocked (returns error message)
3. `"web"` / `"gateway"` channel → always authorized regardless of source
4. `source_channel = None` → authorized from any channel

Ideally, extract the authorization predicate into a small helper function (e.g. `fn is_approval_authorized(source: Option<&str>, requesting: &str) -> bool`) that can be unit-tested without async infrastructure. Then add an integration-level test that exercises the full `handle_message` path with an approval submission targeting a cross-channel thread.

src/agent/thread_ops.rs
```diff
@@ -135,13 +135,29 @@ impl Agent {
...
+        // original creator's channel, not the requesting message's channel.
+        let db_source_channel = if let Some(store) = self.store() {
+            store
+                .get_conversation_source_channel(thread_uuid)
+                .await
+                .unwrap_or(None)
+        } else {
+            None
+        };
+        let effective_source_channel = db_source_channel.as_deref().or(Some(&*message.channel));
```
serrrfirat: **High Severity**

This fallback reassigns authority for legacy or lookup-failed threads to whichever channel hydrates the conversation first:

```rust
let effective_source_channel = db_source_channel.as_deref().or(Some(&*message.channel));
```

That conflicts with the fail-closed rule in `is_approval_authorized(None, ...) == false`. If an older row has `source_channel = NULL` or the DB read transiently fails, the first requester can stamp the in-memory thread with its own channel and make later approval checks treat that channel as the owner.

A safer approach is to keep `None` as `None` here and backfill `source_channel` from trusted persisted state instead of deriving it from the incoming message.

src/db/libsql_migrations.rs
```diff
@@ -723,6 +724,14 @@ CREATE INDEX IF NOT EXISTS idx_routines_event_triggers
...
 
 PRAGMA foreign_keys=ON;
+"#,
+    ),
+    (
+        14,
+        "conversation_source_channel",
+        // Add source_channel to conversations for cross-channel approval authorization.
+        r#"
+ALTER TABLE conversations ADD COLUMN source_channel TEXT;
```
serrrfirat, line 798: **High Severity**

This adds `source_channel` in two places for libSQL:

- the base `SCHEMA` already creates `conversations.source_channel`
- V14 then runs `ALTER TABLE conversations ADD COLUMN source_channel TEXT`

On a fresh libSQL database, `run_migrations()` executes the schema first and then applies incremental migrations, so V14 will attempt to add a column that already exists and should fail startup.

One of these needs to be removed or fresh databases need a mechanism to record V14 as already satisfied before incremental migrations run.

src/channels/wasm/setup.rs
```diff
@@ -71,7 +71,21 @@ pub async fn setup_wasm_channels(
     let mut channels: Vec<(String, Box<dyn crate::channels::Channel>)> = Vec::new();
     let mut channel_names: Vec<String> = Vec::new();
 
+    // Reserved channel names that WASM modules must not claim.
+    // A malicious module could otherwise register as a trusted built-in
+    // channel and bypass cross-channel authorization checks.
+    const RESERVED_CHANNEL_NAMES: &[&str] = &["web", "gateway", "cli", "repl"];
```
serrrfirat: **High Severity**

Blocking only `web`, `gateway`, `cli`, and `repl` is not enough for this authorization model. Approval auth still treats any exact same-channel string as trusted, but built-in channels like `http`, `signal`, and `telegram` are still claimable by a WASM module.

Because `ChannelManager` accepts duplicate names and replaces existing entries, a malicious extension can still register under one of those built-in names and satisfy the `src == requesting` branch in `is_approval_authorized()`.

This needs to reserve the full built-in namespace, or reject any WASM channel whose name collides with an already-registered channel.

src/agent/thread_ops.rs
zmanian: Fixed. Removed the `.or(Some(&*message.channel))` fallback in `maybe_hydrate_thread()`. When `source_channel` is NULL in the DB (or the DB read returns None), `effective_source_channel` now stays `None`. This preserves the fail-closed behavior -- `is_approval_authorized(None, _)` returns `false`, so the approval is denied rather than letting the first requester stamp the thread with their channel.

src/db/libsql_migrations.rs
zmanian: Fixed. Removed `source_channel TEXT` from the base SCHEMA in `libsql_migrations.rs`. The column is now added solely by the V14 incremental migration. This means:
- Fresh databases: base schema creates `conversations` without the column, then V14 adds it via ALTER TABLE.
- Existing databases: V14 has already been applied (or will be on next startup).

No more duplicate column definition that would crash fresh libSQL startup.

src/channels/wasm/setup.rs
zmanian: Fixed with a two-layer approach:

1. Expanded `RESERVED_CHANNEL_NAMES` to cover ALL built-in channels: `web`, `gateway`, `cli`, `repl`, `http`, `signal`, `slack-relay`, `secret_save`.

2. Added a dynamic collision check: `setup_wasm_channels()` now accepts `&[String]` of already-registered channel names (passed from the startup sequence in main.rs). Any WASM channel whose name (case-insensitive) matches an already-registered channel is rejected. This future-proofs against new built-in channels being added without updating the static list, and also prevents a WASM module from shadowing a channel registered earlier in the startup sequence.

ilblackdragon (approved):

zmanian:
@ilblackdragon @serrrfirat Ready for re-review. All 3 HIGH findings addressed in `92138b8c`:

1. **Channel blocklist** -- added `telegram` to reserved names (`http` and `signal` were already present)
2. **libSQL double column** -- added `column_exists()` helper with `IDEMPOTENT_ADD_COLUMN_MIGRATIONS` registry; V14 checks `pragma_table_info` before ALTER TABLE
3. **Legacy thread fail-closed** -- replaced silent `.unwrap_or(None)` with explicit match logging `warn!` for legacy threads and `error!` for DB failures, both remain fail-closed

All clippy/fmt/tests pass.

PR #1594, opened 2026-03-23
feat(skills): add CLAWHUB_ENABLED flag to disable ClawHub registry

### LINKED ISSUE #1570: feat: add CLAWHUB_ENABLED flag to disable ClawHub registry in Gateway UI
## Summary

Add a `CLAWHUB_ENABLED` environment variable (default `true`) that controls whether the ClawHub public skill registry is accessible. When set to `false`:

- `SkillCatalog` is not created (no HTTP client to the public registry)
- The "Search ClawHub" section is hidden from the Gateway Web UI's Skills tab
- `/api/gateway/status` returns `clawhub_enabled: false` so frontends can conditionally render
- `skill_search` and `skill_install` LLM tools are not registered
- `/api/skills/search` returns 501 (already handled when catalog is `None`)
- Installed skills management (list, remove) and install-by-URL continue to work

## Motivation

Downstream deployments (e.g. enterprise/internal builds) may not want users accessing the public ClawHub registry at `clawhub.ai`. Currently the only way to prevent this is `SKILLS_ENABLED=false`, which disables the entire skills system. A more surgical flag lets deployments keep local skill management while hiding the public marketplace.

This also addresses a concern from #557 — the ClawHub registry has no signature verification for downloaded skills, so enterprise deployments may want to disable it until supply chain security is implemented.

## Proposed Changes

1. **`src/config/skills.rs`** — Add `clawhub_enabled: bool` field to `SkillsConfig`, resolved from `CLAWHUB_ENABLED` env var (default `true`)
2. **`src/app.rs`** — Conditionally create `SkillCatalog` only when `skills.clawhub_enabled` is true
3. **`src/tools/registry.rs`** — `register_skill_tools` accepts `Option<Arc<SkillCatalog>>`; only registers `SkillSearchTool`/`SkillInstallTool` when catalog is `Some`
4. **`src/channels/web/server.rs`** — Add `clawhub_enabled: bool` to `GatewayStatusResponse` and `gateway_status_handler`, derived from `state.skill_catalog.is_some()`
5. **`src/channels/web/static/index.html`** — Add `id="clawhub-search-section"` to the search div
6. **`src/channels/web/static/app.js`** — `loadSkills()` fetches `/api/gateway/status` and hides the search section when `clawhub_enabled` is `false`

## Backwards Compatibility

Default is `true`, so no behavior change for existing users.

rajulbhatnagar:
## Summary

Closes #1570.

- Adds `CLAWHUB_ENABLED` environment variable (default `true`) to control access to the public ClawHub skill registry
- When `false`: skips `SkillCatalog` instantiation, excludes `skill_search` from LLM tools, returns `clawhub_enabled: false` in gateway status, hides "Search ClawHub" UI, returns 501 for `/api/skills/search`
- Blocks URL-based skill installs when disabled (cannot distinguish direct URLs from ClawHub mirrors/proxies) — only inline `content` installs remain available
- Preserves local skill management (`skill_list`, `skill_remove`) and content-based `skill_install`
- Dynamic tool schema: `skill_install` hides `url`/`name` params and requires `content` when catalog is disabled
- Makes `SkillInstallRequest.name` optional (content installs derive name from SKILL.md)

## Test plan

- [x] `cargo fmt` — no changes
- [x] `cargo test --lib` — 4301 passed, 0 failed
- [x] 17 new/updated tests: config parsing (3), tool registry (2), skill tool behavior (7), gateway status serialization (1), web handler 501 (1), CLI rejection (1), schema validator (1), test rig builder (1)
- [x] New E2E test suite (`test_clawhub_disabled.py`): gateway status, search 501, URL install 403, content install 200, skill list 200
- [x] Manual: set `CLAWHUB_ENABLED=false`, verified gateway UI hides "Search ClawHub" section
- [x] Manual: `/api/gateway/status` returns `clawhub_enabled: false`

### Commit breakdown

1. **299619b4** `feat(skills): add CLAWHUB_ENABLED flag` — core config, app wiring, tool registry, web gateway status, frontend JS gating
2. **2c83edaf** `fix: address PR review feedback` — test_gateway_state_base refactor, 501 integration test, TestRigBuilder.with_clawhub_disabled()
3. **98c83a5c** `fix: enforce in CLI and hide hints` — gate `ironclaw skills search`, conditional /help, hide ClawHub hints
4. **5c12c2e9** `fix: harden disabled mode` — block URL installs (mirror ambiguity), dynamic tool schema/description, Optional name, unregister_sync, E2E tests, simplification pass

### DIFF
--- a/crates/ironclaw_gateway/static/app.js
+++ b/crates/ironclaw_gateway/static/app.js
@@ -6330,6 +6330,11 @@ function fetchGatewayStatus() {
     restartEnabled = data.restart_enabled || false;
     updateRestartButtonVisibility();
 
+    var clawHubEl = document.querySelector('.skill-search-section');
+    if (clawHubEl) {
+      clawHubEl.style.display = data.clawhub_enabled === false ? 'none' : '';
+    }
+
     var popover = document.getElementById('gateway-popover');
     var html = '';
 
--- a/src/channels/web/server.rs
+++ b/src/channels/web/server.rs
@@ -3782,6 +3782,8 @@ async fn gateway_status_handler(
         .map(|v| v.to_lowercase() == "true")
         .unwrap_or(false);
 
+    let clawhub_enabled = state.skill_catalog.is_some();
+
     Json(GatewayStatusResponse {
         version: env!("CARGO_PKG_VERSION").to_string(),
         sse_connections,
@@ -3795,6 +3797,7 @@ async fn gateway_status_handler(
         llm_backend: state.active_config.llm_backend.clone(),
         llm_model: state.active_config.llm_model.clone(),
         enabled_channels: state.active_config.enabled_channels.clone(),
+        clawhub_enabled,
     })
 }
 
@@ -3823,6 +3826,7 @@ struct GatewayStatusResponse {
     llm_backend: String,
     llm_model: String,
     enabled_channels: Vec<String>,
+    clawhub_enabled: bool,
 }
 
 #[cfg(test)]
@@ -6564,5 +6568,33 @@ mod tests {
     fn test_is_local_origin_rejects_garbage() {
         assert!(!is_local_origin("not-a-url"));
         assert!(!is_local_origin(""));
+    }
+
+    #[test]
+    fn gateway_status_response_serializes_clawhub_enabled() {
+        let resp = GatewayStatusResponse {
+            version: "0.0.0".to_string(),
+            sse_connections: 0,
+            ws_connections: 0,
+            total_connections: 0,
+            uptime_secs: 0,
+            restart_enabled: false,
+            daily_cost: None,
+            actions_this_hour: None,
+            model_usage: None,
+            llm_backend: "test".to_string(),
+            llm_model: "test".to_string(),
+            enabled_channels: vec![],
+            clawhub_enabled: false,
+        };
+        let json = serde_json::to_value(&resp).unwrap();
+        assert_eq!(json["clawhub_enabled"], false);
+
+        let resp_enabled = GatewayStatusResponse {
+            clawhub_enabled: true,
+            ..resp
+        };
+        let json = serde_json::to_value(&resp_enabled).unwrap();
+        assert_eq!(json["clawhub_enabled"], true);
     }
 }
--- a/src/channels/web/server.rs
+++ b/src/channels/web/server.rs
@@ -4017,14 +4017,12 @@ mod tests {
 
     // --- OAuth callback handler tests ---
 
-    /// Build a minimal `GatewayState` for handler tests.
-    fn test_gateway_state_with_dependencies(
-        ext_mgr: Option<Arc<ExtensionManager>>,
-        store: Option<Arc<dyn Database>>,
-        db_auth: Option<Arc<crate::channels::web::auth::DbAuthenticator>>,
-        pairing_store: Option<Arc<crate::pairing::PairingStore>>,
-    ) -> Arc<GatewayState> {
-        Arc::new(GatewayState {
+    /// Build a minimal `GatewayState` with all optional fields set to `None`.
+    ///
+    /// Returns the raw struct so callers can override fields before wrapping
+    /// in `Arc`. Use `test_gateway_state()` for the common Arc-wrapped case.
+    fn test_gateway_state_base() -> GatewayState {
+        GatewayState {
             msg_tx: tokio::sync::RwLock::new(None),
             sse: Arc::new(SseManager::new()),
             workspace: None,
@@ -4032,9 +4030,9 @@ mod tests {
             session_manager: None,
             log_broadcaster: None,
             log_level_handle: None,
-            extension_manager: ext_mgr,
+            extension_manager: None,
             tool_registry: None,
-            store,
+            store: None,
             job_manager: None,
             prompt_queue: None,
             owner_id: "test".to_string(),
@@ -4054,8 +4052,8 @@ mod tests {
             startup_time: std::time::Instant::now(),
             active_config: ActiveConfigSnapshot::default(),
             secrets_store: None,
-            db_auth,
-            pairing_store,
+            db_auth: None,
+            pairing_store: None,
             oauth_providers: None,
             oauth_state_store: None,
             oauth_base_url: None,
@@ -4069,6 +4067,22 @@ mod tests {
         })
     }
 
+    /// Build a minimal `GatewayState` for handler tests with specific dependencies.
+    fn test_gateway_state_with_dependencies(
+        ext_mgr: Option<Arc<ExtensionManager>>,
+        store: Option<Arc<dyn Database>>,
+        db_auth: Option<Arc<crate::channels::web::auth::DbAuthenticator>>,
+        pairing_store: Option<Arc<crate::pairing::PairingStore>>,
+    ) -> Arc<GatewayState> {
+        let mut state = test_gateway_state_base();
+        state.extension_manager = ext_mgr;
+        state.store = store;
+        state.db_auth = db_auth;
+        state.pairing_store = pairing_store;
+        Arc::new(state)
+    }
+
+    /// Build a minimal `GatewayState` for testing the OAuth callback handler.
     fn test_gateway_state(ext_mgr: Option<Arc<ExtensionManager>>) -> Arc<GatewayState> {
         test_gateway_state_with_dependencies(ext_mgr, None, None, None)
     }
@@ -6596,5 +6610,39 @@ mod tests {
         };
         let json = serde_json::to_value(&resp_enabled).unwrap();
         assert_eq!(json["clawhub_enabled"], true);
+    }
+
+    #[tokio::test]
+    async fn skills_search_returns_501_when_catalog_disabled() {
+        use crate::channels::web::auth::{AuthenticatedUser, UserIdentity};
+        use crate::channels::web::handlers::skills::skills_search_handler;
+
+        let dir = tempfile::tempdir().unwrap();
+        let registry = Arc::new(std::sync::RwLock::new(
+            ironclaw_skills::SkillRegistry::new(dir.path().join("skills"))
+                .with_installed_dir(dir.path().join("installed")),
+        ));
+
+        // Skills enabled (registry present) but ClawHub disabled (catalog None).
+        let mut state = test_gateway_state_base();
+        state.skill_registry = Some(registry);
+        let state = Arc::new(state);
+
+        let result = skills_search_handler(
+            axum::extract::State(state),
+            AuthenticatedUser(UserIdentity {
+                user_id: "test".into(),
+                role: "admin".into(),
+                workspace_read_scopes: vec![],
+            }),
+            axum::Json(crate::channels::web::types::SkillSearchRequest {
+                query: "test".into(),
+            }),
+        )
+        .await;
+
+        let err = result.expect_err("should fail with 501 when catalog is None");
+        assert_eq!(err.0, axum::http::StatusCode::NOT_IMPLEMENTED);
+        assert!(err.1.contains("catalog"));
     }
 }
--- a/src/channels/web/server.rs
+++ b/src/channels/web/server.rs
@@ -4064,7 +4064,7 @@ mod tests {
             oauth_sweep_shutdown: None,
             frontend_html_cache: Arc::new(tokio::sync::RwLock::new(None)),
             tool_dispatcher: None,
-        })
+        }
     }
 
     /// Build a minimal `GatewayState` for handler tests with specific dependencies.
--- a/src/config/skills.rs
+++ b/src/config/skills.rs
@@ -12,6 +12,8 @@ use crate::settings::Settings;
 pub struct SkillsConfig {
     /// Whether the skills system is enabled.
     pub enabled: bool,
+    /// Whether the public ClawHub registry is accessible.
+    pub clawhub_enabled: bool,
     /// Directory containing user-placed skills (default: ~/.ironclaw/skills/).
     /// Skills here are loaded with `Trusted` trust level.
     pub local_dir: PathBuf,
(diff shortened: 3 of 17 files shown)


src/channels/web/server.rs
```diff
@@ -2656,6 +2656,8 @@ async fn gateway_status_handler(
         .map(|v| v.to_lowercase() == "true")
         .unwrap_or(false);
```
serrrfirat, line 3784: **Medium Severity — Missing test for 501 path**

The PR description states `/api/skills/search` returns 501 when ClawHub is disabled. The handler at `handlers/skills.rs:58-61` does `state.skill_catalog.as_ref().ok_or(StatusCode::NOT_IMPLEMENTED)`, which is correct — when `clawhub_enabled=false`, `skill_catalog` is `None` and the `skills_search_handler` will return 501. However, no test verifies this HTTP-level behavior.

**Suggestion:** Add an integration test that constructs a `GatewayState` with `skill_catalog: None` and verifies `POST /api/skills/search` returns 501.

tests/support/test_rig.rs
```diff
@@ -667,7 +667,7 @@ impl TestRigBuilder {
                 let catalog = ironclaw::skills::catalog::shared_catalog();
                 components
                     .tools
-                    .register_skill_tools(Arc::clone(&registry), Arc::clone(&catalog));
+                    .register_skill_tools(Arc::clone(&registry), Some(Arc::clone(&catalog)));
```
serrrfirat: **Medium Severity — TestRig cannot simulate CLAWHUB_ENABLED=false**

When `enable_skills` is true, the test rig unconditionally passes `Some(catalog)` to `register_skill_tools()`. This means tests using `TestRigBuilder::new().with_skills()` will **never** exercise the `CLAWHUB_ENABLED=false` path (skills enabled but ClawHub disabled).

**Suggestion:** Add a `.with_clawhub_disabled()` builder method (or pass `None` for catalog when a flag is set) so integration tests can verify the behavior of skill tools without ClawHub.

crates/ironclaw_gateway/static/app.js
```diff
@@ -4376,6 +4376,11 @@ function fetchGatewayStatus() {
     restartEnabled = data.restart_enabled || false;
     updateRestartButtonVisibility();
 
+    var clawHubEl = document.querySelector('.skill-search-section');
+    if (clawHubEl) {
```
serrrfirat, line 6334: **Low Severity — Graceful degradation note (informational)**

The strict equality check `data.clawhub_enabled === false` means if the field is missing from the response (e.g., older backend version), the section remains visible — which is the correct default behavior. Good defensive design.

src/config/skills.rs
```diff
@@ -47,6 +50,7 @@ impl SkillsConfig {
     pub(crate) fn resolve() -> Result<Self, ConfigError> {
```
serrrfirat: **Low Severity — Missing .env.example entry**

The new `CLAWHUB_ENABLED` env var is documented in the struct but not mentioned in `.env.example`. Operators looking at `.env.example` won't discover this configuration option.

**Suggestion:** Add `CLAWHUB_ENABLED=true` with a descriptive comment to `.env.example` for discoverability.

src/tools/registry.rs
```diff
@@ -448,22 +448,29 @@ impl ToolRegistry {
...
+            )));
+        }
+
+        self.register_sync(Arc::new(SkillInstallTool::new(registry, catalog.clone())));
+
+        tracing::debug!(
+            "Registered skill management tools (clawhub={})",
+            catalog.is_some()
+        );
     }
```
serrrfirat, line 796: **Nit — Debug log style consistency**

Other `register_*` methods log the tool count (e.g., `"Registered 4 skill management tools"`). This line logs `clawhub=true/false` which is useful context but breaks the pattern slightly.

Consider: `"Registered {} skill management tools (clawhub={})", count, catalog.is_some()` where count is 3 or 4. Take it or leave it.

src/config/skills.rs
```diff
@@ -58,3 +62,45 @@ impl SkillsConfig {
...
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::config::helpers::lock_env;
+
+    #[test]
+    fn clawhub_enabled_defaults_true() {
```
serrrfirat, line 100: **Nit — unsafe env var manipulation (consistent with codebase)**

These tests use `unsafe { std::env::set_var(...) }` which is required in Rust 1.81+ for multi-threaded safety. The `lock_env()` guard provides synchronization. This is consistent with existing config tests in the codebase, so no action needed — just noting for awareness.

src/channels/web/server.rs
rajulbhatnagar: Done in 543d748. Added `skills_search_returns_501_when_catalog_disabled` async test in `server.rs` — constructs a `GatewayState` with `skill_registry: Some(...)` / `skill_catalog: None` and asserts the handler returns `(501, "...catalog...")`.

Also extracted a `test_gateway_state_base()` helper so this test (and future ones) can override individual fields without duplicating the full 25-field struct construction.

tests/support/test_rig.rs
rajulbhatnagar: Done in 543d748. Added `clawhub_disabled: bool` field and `.with_clawhub_disabled()` builder method to `TestRigBuilder`. When combined with `.with_skills()`, it passes `None` for catalog to `register_skill_tools()` and sets `components.config.skills.clawhub_enabled = false`.

crates/ironclaw_gateway/static/app.js
rajulbhatnagar: Agreed — the strict `=== false` check is intentional for exactly this forward-compatibility reason. No change needed.

src/config/skills.rs
rajulbhatnagar: Done in 543d748. Added to `.env.example` under a new "Skills / ClawHub Registry" section, following the existing commented-out-with-description pattern.

src/tools/registry.rs
rajulbhatnagar: Good call — done in 543d748. Switched to the mutable counter pattern from `register_job_tools`:

```rust
let mut count = 3; // list, remove, install
if let Some(ref catalog) = catalog {
    self.register_sync(Arc::new(SkillSearchTool::new(...)));
    count += 1;
}
tracing::debug\!("Registered {} skill management tools (clawhub={})", count, catalog.is_some());
```

src/config/skills.rs
rajulbhatnagar: Acknowledged — the `lock_env()` guard + `unsafe { set_var }` pattern is consistent with the rest of `src/config/`. No change needed.

serrrfirat (dismissed):
Approved. Clean implementation — Option<Arc<SkillCatalog>> pattern is idiomatic, defense in depth (UI hide + 501 + tool deregistration) is solid, test coverage is good. One mechanical conflict in src/config/skills.rs needs resolving on rebase (adapt to db_first_bool from #1722). Important for multi-tenant deployments.

serrrfirat:
@rajulbhatnagar I've rebased your branch onto the latest `staging` (75 commits ahead) and resolved the conflict in `src/config/skills.rs` — adapted `clawhub_enabled` to use the new `db_first_bool` pattern from #1722 and added the field to `SkillsSettings` in `settings.rs`.

The rebased branch is at `nearai/ironclaw:feat/clawhub-enabled-flag`. To update your fork:

```bash
git remote add upstream git@github.com:nearai/ironclaw.git  # if not already
git fetch upstream feat/clawhub-enabled-flag
git checkout feat/clawhub-enabled-flag
git reset --hard upstream/feat/clawhub-enabled-flag
git push --force-with-lease origin feat/clawhub-enabled-flag
```

Clippy passes with zero warnings. PR is approved — just needs this push to clear the merge conflict.

rajulbhatnagar:
Thanks. Will do tonight

serrrfirat:
Posting these as top-level PR comments rather than inline because both issues are in unchanged files on the current diff.

**High Severity — `CLAWHUB_ENABLED=false` is not enforced by the standalone CLI**

Files: `src/cli/skills.rs:62-65`, `src/cli/skills.rs:168-171`

The runtime and web paths correctly gate ClawHub behind `clawhub_enabled`, but `ironclaw skills search <query>` still bypasses that flag. `run_skills_command()` resolves `full_config.skills`, yet the `Search` branch calls `cmd_search()` unconditionally, and `cmd_search()` always constructs `SkillCatalog::new()` and queries the public registry.

That means operators who disable ClawHub for offline/compliance reasons still retain a public-registry entrypoint through the CLI, which breaks the new config contract.

Suggested fix: reject `SkillsCommand::Search` when `config.clawhub_enabled` is false, return a clear disabled message, and add a CLI regression test for that path.

**Low Severity — help/list output still advertises ClawHub when it is disabled**

Files: `src/agent/commands.rs:598-600`, `src/agent/commands.rs:817-836`, `src/cli/skills.rs:125`

When the catalog is disabled, `/help`, `/skills`, and the standalone CLI list output still tell users to search ClawHub. The next command then fails with `Skill catalog not available`, so the UX/docs drift from the actual runtime behavior.

Suggested fix: make those hints conditional on `self.skill_catalog().is_some()` / `config.clawhub_enabled`, or replace them with a disabled-mode message.

serrrfirat (commented):
I found three issues that still need to be addressed before this is safe to merge:

1. `CLAWHUB_ENABLED=false` is still bypassable via direct ClawHub download URLs. The new flag removes search/name-based lookup, but both the tool path and the web API still accept arbitrary HTTPS `url` input first, so callers can continue fetching from the public registry if they know the download URL. That breaks the stated goal of disabling public ClawHub access for internal deployments. Relevant paths: `src/tools/builtin/skill_tools.rs` and `src/channels/web/handlers/skills.rs`.

2. The new "URL/content installs still work" behavior is only partially implemented because `name` is still mandatory. `skill_install` still requires `name` in the schema and immediately calls `require_str("name")`; the web request type also requires `name`. In practice that means callers need to invent a dummy name, and schema validation can reject the intended fallback before execution. Relevant paths: `src/tools/builtin/skill_tools.rs` and `src/channels/web/types.rs`.

3. `TestRigBuilder::with_clawhub_disabled()` does not faithfully reproduce production disabled mode. The flag is never applied to `config` before `AppBuilder::build_all()`, so AppBuilder can still instantiate the catalog and register `skill_search`. The later override only swaps fields and re-registers tools; it never removes an already-registered `skill_search`, so tests using this helper can get false confidence. Relevant paths: `tests/support/test_rig.rs`, `src/app.rs`, and `src/tools/registry.rs`.

Please also add an end-to-end/regression test that exercises the disabled mode across the real user-facing paths: no `skill_search` tool registered, `/api/skills/search` rejected, CLI `/skills search` rejected, and direct ClawHub URL installs blocked when `CLAWHUB_ENABLED=false`.

serrrfirat:
**Medium Severity — `unregister_sync` silently drops tool removal on lock contention**

Two related issues in `src/tools/registry.rs:228-241`:

**1. Silent drop on contested lock:** `unregister_sync` uses `try_write()` on the tokio `RwLock` and silently returns if it fails. Since this is used to clean up `skill_search` when ClawHub is disabled, if the lock is held at that moment, `skill_search` remains registered and callable by the LLM despite ClawHub being disabled. The consequence is a security-relevant tool remaining available.
**Fix:** Use the async `unregister()` method instead, or add a retry/queue mechanism. At minimum, log at `warn!` level rather than `debug!` so operators know the removal was skipped.

**2. Stale `builtin_names` entry:** `unregister_sync` does not remove the tool name from `builtin_names`. A stale entry would prevent re-registration via the async `register()` path (line 192 check).
**Fix:** Add `self.builtin_names.try_write()` removal mirroring `register_sync`.

serrrfirat (commented):
## Paranoid Architect Review — Verdict: 👍 APPROVE WITH MINOR FIXES

⚠️ **This PR has merge conflicts and needs a rebase onto staging.**

Well-structured PR with defense-in-depth across all surfaces (LLM tool schema, web API, CLI, frontend, gateway status). 17 new tests covering unit, integration, and E2E. No IronClaw convention violations.

### Findings

| 1 | Medium | Reliability | tools/registry.rs `unregister_sync()` | **Silent lock contention.** `try_write()` failure logged at `debug!`. If lock held, `skill_search` stays registered despite catalog disabled. LLM could invoke and fail at runtime. | Use `warn!` level, or retry lock, or assert (startup-only context) |
| 2 | Medium | Security | skill_tools.rs `execute()` | URL installs allow arbitrary HTTPS fetches not restricted to ClawHub domain even when enabled. Pre-existing, not introduced by this PR. | Document as trusted-operator feature or add domain allowlist |
| 3 | Low | Edge Case | agent/commands.rs:~830 | Invalid `/skills` subcommand error shows `/skills search` hint when ClawHub disabled | Gate message on `skill_catalog().is_some()` |
| 4 | Low | Formatting | agent/commands.rs:~112 | Help text missing blank line between Skills and Agent sections when search disabled | Add trailing `\n` |
| 5 | Low | Test Hygiene | config/skills.rs | `unsafe { std::env::set_var(...) }` could leak on panic | Use RAII scope guard for cleanup |
| 6 | Nit | Consistency | web/handlers/skills.rs | URL-blocked: 403, catalog-disabled: 501. Different codes for similar "feature disabled" semantics. | Consider 501 for both |
| 7 | Nit | Docs | .env.example | No note that `CLAWHUB_ENABLED` requires `SKILLS_ENABLED=true` | Add comment |
| 8 | Nit | Dead Param | skill_tools.rs | Schema doesn't expose `slug` but `execute()` reads it | Align schema and implementation |
| 9 | Nit | Code Style | web/static/app.js:5370 | `data.clawhub_enabled === false` default-safe but intent unclear | Add backwards-compat comment |
| 10 | Nit | Test Naming | skill_tools.rs | `test_skill_install_schema_without_catalog` only checks metadata, not schema | Rename or add schema assertions |

### Summary
Only medium finding (#1) is the silent lock failure in `unregister_sync`. Low practical risk at startup but debugging would be hard at `debug!` level. The rest are minor polish. Good PR overall.

rajulbhatnagar:
Closing this in favour of https://github.com/nearai/ironclaw/pull/2727

PR #1597, opened 2026-03-23
feat(llm): surface active extension context

### LINKED ISSUE #1595: [QA] LLM forgets activated Telegram channel and asks users to enable it again
## Summary

When Telegram is already installed, configured, verified, and active, the chat agent can still tell the user to enable or activate Telegram again instead of using it or checking current state first.

This is especially confusing after a successful hot activation in the web UI.

## User-visible impact

Users can get stuck in a fake setup loop:
- they complete Telegram setup successfully
- IronClaw confirms activation
- later, the LLM still asks them to activate Telegram again

That makes the product look like setup did not stick, even though the runtime appears healthy.

## Repro

1. Install/configure Telegram from the web gateway.
2. Complete the verification flow until Telegram hot activation succeeds.
3. Start or continue a gateway chat.
4. Ask something like:
   - `is telegram connected?`
   - `send this to me on telegram`
   - `message me on telegram`
5. Observe that the model may ask the user to enable or activate Telegram again instead of checking current state or using the channel.

## Expected

If Telegram is already active for the current user, the agent should either:
- know that from its runtime/prompt context, or
- call `tool_list` / `extension_info` before telling the user to set anything up.

For a delivery request, it should be willing to use `message {"channel":"telegram", ...}` when Telegram is already active and owner-bound.

## Actual

The agent can behave as if Telegram is not installed or not active, and repeat setup guidance even after successful activation.

### Hot activation itself looks healthy

The Telegram hot-activation path already has coverage asserting that a successful configure:
- marks `telegram` active
- hot-adds the channel to `ChannelManager`
- persists the active channel state
- stores the owner binding

Relevant code/test:
- `src/extensions/manager.rs:6509-6547`

So this does **not** look like a Telegram activation failure first.

### The per-turn LLM context does not include extension/channel state

`Agent::run_agentic_loop()` builds reasoning context from:
- workspace system prompt
- channel conversation context
- tool definitions

but it does not inject installed/active/authenticated extension state into the prompt before calling `build_system_prompt_with_tools()`.

Relevant code:
- `src/agent/dispatcher.rs:65-158`

### Workspace system prompt assembly does not include active extensions either

`Workspace::system_prompt_for_context_inner()` loads bootstrap, identity files, tool notes, memory, and profile context, but it does not include installed/active channels or WASM tools.

Relevant code:
- `src/workspace/mod.rs:1106-1185`

### The prompt guidance biases toward search/install/activate, not inspect current state

`Reasoning::build_extensions_section_for_tools()` only adds generic extension guidance and explicitly mentions `tool_search`. It does **not** tell the model to use `tool_list` / `extension_info` to inspect what is already installed and active before asking the user to reconnect something.

Relevant code:
- `src/llm/reasoning.rs:934-949`

### There is already an intended behavior signal for `tool_list`

We already have a recorded trace expecting the model to call `tool_list` when the user asks `is telegram connected?`:
- `tests/fixtures/llm_traces/recorded/telegram_check.json:3-58`

That suggests the desired behavior is already understood conceptually, but we do not appear to have coverage that ties this expectation to a real hot-activated Telegram setup or a “send via Telegram” request.

## Suspected root cause

The runtime/UI know Telegram is active, but that fact never reaches the model unless the model independently decides to call `tool_list`.

Because the prompt includes generic extension-management guidance and no compact active-extension snapshot, the model can hallucinate a missing setup step and ask the user to activate Telegram again.

## Suggested fix

1. Add a compact per-turn extension summary to reasoning context, sourced from `ExtensionManager::list(None, false)` or a dedicated summary helper.
   Example contents:
   - active channels: `telegram` (authenticated, active, owner-bound)
   - active wasm tools
   - active MCP servers
2. Update the extensions prompt guidance to explicitly say:
   - use `tool_list` to inspect installed/authenticated/active extensions before telling the user to connect or activate something
   - use `extension_info` when deeper compatibility/state detail is needed
3. Add regression coverage for:
   - gateway chat after Telegram hot activation asking `is telegram connected?`
   - gateway chat after Telegram hot activation asking to send a message via Telegram
   - activated WASM tools/channels being surfaced consistently in LLM-facing context

## Notes

I think the likely culprit is not that Telegram hot activation is broken, but that activated extension state is not propagated into the LLM’s per-turn context strongly enough, so the model falls back to generic setup advice.

G7CNF:
Fixes #1595.

Summary:
- add a compact active-extension snapshot to the per-turn LLM prompt context
- avoid telling users to re-enable an extension when the runtime snapshot already shows it active
- include regression coverage for the prompt snapshot
- update FEATURE_PARITY with the new dispatcher-side prompt diagnostic note

Validation:
- cargo fmt --all -- --check
- cargo test -p ironclaw llm::reasoning::tests::test_system_prompt_with_extension_context_includes_snapshot -- --nocapture
- IC_BASE_REF=origin/staging /Users/nigecoleman/.codex/skills/rust-pr-preflight/scripts/preflight.sh (repo baseline violations remain unrelated to this branch)

### DIFF
--- a/src/agent/dispatcher.rs
+++ b/src/agent/dispatcher.rs
@@ -199,12 +199,12 @@ impl Agent {
                 active_skills.iter().map(|s| s.name().to_string()).collect();
             reasoning = reasoning.with_active_skill_names(skill_names);
         }
-        if active_skills.is_empty() {
-            if let Some(extension_manager) = self.deps.extension_manager.as_ref() {
-                let extension_context = extension_manager.active_extension_summary().await;
-                if !extension_context.is_empty() {
-                    reasoning = reasoning.with_extension_context(extension_context);
-                }
+        if active_skills.is_empty()
+            && let Some(extension_manager) = self.deps.extension_manager.as_ref()
+        {
+            let extension_context = extension_manager.active_extension_summary().await;
+            if !extension_context.is_empty() {
+                reasoning = reasoning.with_extension_context(extension_context);
             }
         }
 
--- a/src/agent/dispatcher.rs
+++ b/src/agent/dispatcher.rs
@@ -199,6 +199,14 @@ impl Agent {
                 active_skills.iter().map(|s| s.name().to_string()).collect();
             reasoning = reasoning.with_active_skill_names(skill_names);
         }
+        if active_skills.is_empty() {
+            if let Some(extension_manager) = self.deps.extension_manager.as_ref() {
+                let extension_context = extension_manager.active_extension_summary().await;
+                if !extension_context.is_empty() {
+                    reasoning = reasoning.with_extension_context(extension_context);
+                }
+            }
+        }
 
         // Create a JobContext for tool execution (chat doesn't have a real job)
         let mut job_ctx =
--- a/src/llm/reasoning.rs
+++ b/src/llm/reasoning.rs
@@ -388,6 +388,8 @@ pub struct Reasoning {
     skill_context: Option<String>,
     /// Names of active skills (used to suppress extension search for covered domains).
     active_skill_names: Vec<String>,
+    /// Optional active extension snapshot to inject into system prompt.
+    extension_context: Option<String>,
     /// Channel name (e.g. "discord", "telegram") for formatting hints.
     channel: Option<String>,
     /// Model name for runtime context.
@@ -409,6 +411,7 @@ impl Reasoning {
             workspace_system_prompt: None,
             skill_context: None,
             active_skill_names: Vec::new(),
+            extension_context: None,
             channel: None,
             model_name: None,
             is_group_chat: false,
@@ -443,6 +446,14 @@ impl Reasoning {
     /// installation for domains already covered by active skills.
     pub fn with_active_skill_names(mut self, names: Vec<String>) -> Self {
         self.active_skill_names = names;
+        self
+    }
+
+    /// Set a compact active-extension snapshot to inject into the system prompt.
+    pub fn with_extension_context(mut self, context: String) -> Self {
+        if !context.is_empty() {
+            self.extension_context = Some(context);
+        }
         self
     }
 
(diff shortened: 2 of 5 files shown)


src/llm/reasoning.rs
```diff
@@ -840,6 +851,19 @@ Respond with a JSON plan in this format:
             String::new()
         };
 
+        let extension_context_section = if let Some(ref ext_ctx) = self.extension_context {
+            format!(
+                "\n\n## Active Extensions\n\n\
+                 The following extensions are already active and available in the current runtime.\n\
+                 Do not tell the user to set them up again unless the snapshot says they are missing.\n\
+                 If you need more detail, use `tool_list` or `extension_info` before asking the user\n\
```
serrrfirat: **Medium Severity**

This section is included even when the dispatcher intentionally switches to the cached no-tools prompt for the max-iteration fallback. At that point `build_system_prompt_with_tools(&[])` still tells the model to use `tool_list` or `extension_info`, which directly contradicts the force-text path and can degrade the final turn into more tool-seeking language instead of an answer.

The new test also locks that contradiction in by asserting that `tool_list` must appear in the no-tools prompt. This guidance should be conditional on those extension-management tools actually being present in the prompt's tool set, and there should be a regression test that the force-text prompt does not mention unavailable tools.

src/agent/dispatcher.rs
```diff
@@ -137,6 +137,12 @@ impl Agent {
         if let Some(ctx) = skill_context {
             reasoning = reasoning.with_skill_context(ctx);
         }
+        if let Some(extension_manager) = self.deps.extension_manager.as_ref() {
+            let extension_context = extension_manager.active_extension_summary().await;
+            if !extension_context.is_empty() {
+                reasoning = reasoning.with_extension_context(extension_context);
```
serrrfirat, line 207: **Medium Severity**

This new prompt context bypasses the skill attenuation model. `active_extension_summary()` can list active extension tools here even when an installed skill has already removed those tools from `initial_tool_defs` and from the per-iteration `available_tools` set. In that case the model is told those extensions are already available, but it cannot actually call them.

That is a behavioral regression for the trust model in `skills::attenuate_tools`: the prompt can now advertise capabilities that were intentionally hidden from the LLM. Please either suppress extension/tool details when attenuation is active, or derive this snapshot from the same post-attenuation visibility set and add a regression test with an installed skill plus an active extension.

serrrfirat:
Thank you for improving the model’s awareness of extension state. We are closing this PR because Reborn exposes that information through its capability surface rather than adding a separate active-extension prompt section.

The Reborn runner builds the model-visible surface from capabilities currently visible to the actor, and `tool_search` provides the live discovery catalog (`crates/ironclaw_runner/src/tool_disclosure.rs`). Activation, ownership, and visibility are therefore enforced by the same runtime source of truth. A separately rendered prose list could drift from what the model can actually invoke.

The user experience you were fixing remains important. We would be happy to have you contribute Reborn-side tests or improvements to capability discovery and the messaging shown when a capability is installed but unavailable. Thank you for the contribution.

PR #1598, opened 2026-03-23
fix: block Telegram bot tokens in normal chat

### LINKED ISSUE #1596: Telegram onboarding can ask for bot tokens in chat and misreport them as invalid
## Summary

Telegram onboarding can fall off the secure auth/setup path and ask the user to paste a bot token into normal chat. When that happens, the raw token is exposed to the model instead of going through the dedicated secrets flow, and the model can misdiagnose a valid token as invalid.

In the reported case, the user had a BotFather token, pasted it into chat, and IronClaw replied that Telegram returned 404 / the token was invalid and suggested regenerating the bot. The safer and intended behavior would have been to show the Telegram setup modal or auth prompt and keep the token out of the conversation entirely.

## Why this matters

This is both a UX bug and a security/privacy bug:
- users can be told to revoke/regenerate a valid bot token unnecessarily
- a Telegram bot token can end up in normal chat instead of the encrypted secrets path
- once the token is in plain chat, the model can hallucinate token validation guidance

## Expected

When a user wants to connect Telegram:
- the agent should use `tool_auth` / extension setup flow, not ask for the token in plain chat
- the web client should show the secure auth card or configure modal
- the token should go to `/api/chat/auth-token` or `/api/extensions/telegram/setup`
- the token should never touch the LLM, chat history, or normal turn pipeline

If a user pastes a token-shaped message into chat anyway, IronClaw should detect that and redirect/block/redact it rather than letting the LLM reason over it.

## Actual

A conversational Telegram setup can still ask the user to paste the token directly into chat. If that happens, the token is handled as normal user input unless auth mode was already armed, and the LLM can respond with misleading validation advice like “that token is invalid; create a new bot”.

### We already have a secure token flow, but it only works after auth mode is entered

There is a dedicated web handler for token submission that bypasses the message pipeline entirely:
- `src/channels/web/handlers/chat.rs:148-188`

The auth card/configure modal wiring is also already present in the web client:
- `src/channels/web/static/app.js:1481-1666`

And `tool_auth` explicitly says the token should be provided through a secure channel, not through the tool itself:
- `src/tools/builtin/extension_tools.rs:186-189`

### Auth interception is conditional on `pending_auth`

The agent only diverts a user message away from the LLM when the thread already has `pending_auth` set:
- `src/agent/agent_loop.rs:1077-1114`

If the agent never called `tool_auth` / never entered auth mode, a pasted Telegram token just becomes a normal chat message and goes through the regular LLM turn.

### The bootstrap guidance expects `tool_auth`, but there is no hard guardrail when the conversation misses that path

The first-run bootstrap prompt tells the agent to use `tool_auth` for Telegram credentials:
- `src/workspace/seeds/BOOTSTRAP.md:32-36`

But there is no server-side fallback that says “this message looks like a Telegram bot token; do not send it to the LLM”. So if the model asks for the token in plain chat anyway, the secret leaks into the wrong path.

### The existing sensitive-content classifier is not enough for this scenario

The default privacy classifier only looks for credential *keywords* like `password`, `api_key`, `auth_token`, or `secret_key`:
- `src/workspace/privacy.rs:36-60`

A bare Telegram token pasted by itself does not necessarily match those patterns, and this classifier is not the chat interception mechanism anyway.

### Telegram validation errors are currently flattened into “Invalid token”

During channel setup, any non-2xx response from the validation endpoint becomes:
- `Invalid token (API returned <status>)`

Relevant code:
- `src/extensions/manager.rs:4934-4976`

That means even if the underlying problem is not actually “the token is wrong” (for example, some network/proxy/service issue), the user still gets a misleading invalid-token error.

## Suspected root cause

This looks like a two-part gap:

1. The conversational setup flow can fail to trigger `tool_auth` / secure setup UI, so the user is asked for the Telegram token in plain chat.
2. Once a raw token appears in chat without `pending_auth`, there is no fallback interception/redaction path, so the LLM sees the token and can hallucinate incorrect validation advice.

## Suggested fix

1. Add a hard conversational rule that secrets must never be requested or accepted in normal chat when an extension has a secure auth/setup flow.
2. Add server-side token-shape detection for common credentials like Telegram bot tokens. If a user pastes one in chat outside auth mode, block/redact it and redirect them to the secure auth card or configure modal.
3. Tighten the Telegram onboarding path so connect/setup requests reliably trigger `tool_auth` / `auth_required` instead of asking the user to paste credentials into chat.
4. Improve Telegram validation error reporting so non-2xx responses are not always flattened into “Invalid token” without more context.
5. Add regression coverage for:
   - conversation-driven Telegram setup triggering secure auth UI
   - a Telegram token pasted into plain chat being blocked/redirected instead of sent to the LLM
   - no token appearing in conversation history when the secure auth flow is used

## Notes

The core bug here may not be that the Telegram token validator itself is wrong. The more fundamental problem is that the secret reached the model at all, which should not happen when the secure auth/configure path exists.

henrypark133:
## Summary
- intercept Telegram bot-token-shaped chat messages before turn creation or any LLM call
- emit the existing Telegram auth-required status so the web UI can redirect users into the secure setup flow
- tighten bootstrap guidance so the agent never asks for secrets in normal chat

## Testing
- cargo fmt --all
- cargo clippy --all-targets --all-features -- -D warnings
- cargo test telegram_bot_token_messages_are_redirected_before_llm --lib
- cargo test detects_telegram_bot_token_messages --lib
- cargo test ignores_normal_telegram_setup_messages --lib

Closes #1596


src/agent/agent_loop.rs
```diff
@@ -62,6 +63,38 @@ pub(crate) fn truncate_for_preview(output: &str, max_chars: usize) -> String {
...
+        match self {
+            Self::TelegramBotToken => {
+                "Telegram bot tokens can't be accepted in normal chat. Use the secure Telegram setup flow instead."
+            }
+        }
+    }
+}
+
+static TELEGRAM_BOT_TOKEN_RE: LazyLock<Regex> =
+    LazyLock::new(|| Regex::new(r"^\d{6,}:[A-Za-z0-9_-]{20,}$").expect("TELEGRAM_BOT_TOKEN_RE")); // safety: hardcoded literal
```
serrrfirat: **High Severity**

The detector only matches when the *entire* message is exactly a token, so common paste formats like `token: 123...`, backticks/code fences, trailing punctuation, or multiline messages will bypass this guard and still flow through normal chat processing. At the same time, the pattern itself is fairly loose (`^\d{6,}:[A-Za-z0-9_-]{20,}$`), which raises false-positive risk for non-Telegram strings that happen to look similar.

That makes the protection unreliable in both directions: real tokens can leak through, while unrelated input can get blocked.

Suggested fix: scan candidate substrings inside the message (for example by splitting on whitespace and trimming wrapping punctuation), then apply a tighter Telegram-token regex. Please also add regression tests for wrapped/prefixed/suffixed token pastes and obvious non-token lookalikes.

src/agent/agent_loop.rs
```diff
@@ -1125,6 +1180,15 @@ impl Agent {
             }
         }
 
+        if let Submission::UserInput { ref content } = submission {
+            if let Some(credential) = detect_sensitive_chat_credential(content) {
```
serrrfirat: **Medium Severity**

This interception runs *after* `BeforeInbound` hooks, so hooks still receive the raw Telegram bot token first. That weakens the security goal here: even if the LLM never sees the token, a hook can still log it, forward it to a webhook, or otherwise process it as part of normal inbound chat handling.

If the intent is to keep bot tokens out of normal chat processing entirely, this check needs to happen before the `HookEvent::Inbound` dispatch (while still preserving the existing pending-auth precedence).

src/agent/agent_loop.rs
```diff
@@ -191,6 +224,28 @@ pub struct Agent {
 }
 
 impl Agent {
+    async fn intercept_sensitive_chat_credential(
+        &self,
+        message: &IncomingMessage,
+        credential: SensitiveChatCredential,
+    ) -> String {
+        let instructions = credential.redirect_message().to_string();
```
serrrfirat, line 271: **Medium Severity**

This emits `AuthRequired`, but it does not actually put the thread into pending auth mode or provide a real `setup_url`/`auth_url`. That makes the status semantics inconsistent with the rest of the auth flow and can trap users in a loop where they paste the same token again and just get redirected again.

Suggested fix: either drive the real secure setup flow from here (pending auth and/or concrete setup/auth URLs), or use a different status/message that does not imply the standard auth-mode contract.

src/agent/agent_loop.rs
```diff
@@ -1483,4 +1562,183 @@ mod tests {
 
         assert!(should_fallback_routine_notification(&error)); // safety: test-only assertion
     }
+
+    #[test]
+    fn detects_telegram_bot_token_messages() {
```
serrrfirat, line 1821: **Low Severity**

The new tests only cover the easiest case where the message is exactly the token. That misses the most likely user behaviors: prefixed text (`token: ...`), wrapped tokens (backticks / parentheses), trailing punctuation, and multiline pastes. Those are important here because the current detector is anchored to the whole message.

Please add regression tests for those formats, plus a few false-positive cases that should *not* trigger interception.

src/workspace/seeds/BOOTSTRAP.md
```diff
@@ -35,6 +35,10 @@ If they're interested, set it up right here using the extension tools:
 3. Use `tool_auth` to collect credentials (e.g. Telegram bot token from @BotFather)
 4. The channel will be hot-activated — no restart needed
 
+Never ask the user to paste tokens, passwords, API keys, or other secrets into
```
serrrfirat: **Low Severity**

This is directionally correct, but the wording is a bit ambiguous immediately after “Use `tool_auth` to collect credentials”. It would help to explicitly say that `tool_auth` is the secure collection path, and that secrets must never appear in regular chat text/history.

That avoids readers interpreting these two instructions as contradictory.

ilblackdragon (changes requested):
## Review: fix: block Telegram bot tokens in normal chat

Close to approval. Three minor items:

### Must-fix

1. **`has_pending_auth` doc comment says "fail-open"** — from the credential-interception caller's perspective, `false` causes the interception to fire (blocking the token), which is fail-**closed**. Fix the comment: "Returns `false` if the lock is contended, which causes the credential interception to fire — safe default."

2. **Silent `let _ =` on `send_status`** — if the status send fails, the user gets the text redirect but the web UI won't show the auth redirect. Log a warning:
```rust
if let Err(e) = self.channels.send_status(...).await {
    tracing::warn!(error = %e, "Failed to send auth-required status");
}
```

### Nit

3. **`detect_sensitive_chat_credential` doc comment** mentions "splits on whitespace and strips common wrapping punctuation" but the fast path returns before doing any splitting. Clarify the two-path behavior.

src/agent/agent_loop.rs
```diff
@@ -977,6 +1032,15 @@ impl Agent {
...
         );
 
+        if let Submission::UserInput { ref content } = submission
+            && let Some(credential) = detect_sensitive_chat_credential(content)
+        {
+            return Ok(Some(
+                self.intercept_sensitive_chat_credential(message, credential)
+                    .await,
+            ));
+        }
```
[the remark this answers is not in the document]
henrypark133: Fixed in c804e72f. The credential interception now uses nested `if let` structure on the rebased branch instead of the earlier let-chain form.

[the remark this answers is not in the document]
henrypark133: Fixed in 4b37e038. Pending auth now takes precedence over the normal-chat token redirect, and the `agent_loop` tests cover Telegram token submission while auth mode is active.

src/agent/agent_loop.rs
henrypark133: Fixed in a944d347 and 39c55c5d. The branch now includes the broader unit-test matrix for wrapped/prefixed/multiline cases plus an E2E regression that verifies redirect-to-setup and no token leakage into chat history.

src/workspace/seeds/BOOTSTRAP.md
henrypark133: Fixed in a944d347. The BOOTSTRAP wording now makes `tool_auth` the only credential collection path and explicitly says secrets must never appear in normal chat or conversation history.

src/agent/agent_loop.rs
henrypark133: Fixed in 39c55c5d. The web client now treats Telegram `auth_required` without an `auth_url` as the secure setup path by opening the configure modal, and it redacts the pasted token from the visible chat message as part of that flow.

src/agent/agent_loop.rs
henrypark133: Fixed in a944d347. Sensitive-credential interception now runs before `BeforeInbound` hooks, while the pending-auth path still bypasses the normal-chat redirect.

src/agent/agent_loop.rs
henrypark133: Fixed in a944d347. The detector now uses a tighter Telegram-token pattern, scans raw text plus normalized candidates, and has regression coverage for wrapped, prefixed, suffixed, multiline, and obvious non-token cases.

PR #1599, opened 2026-03-23
fix(agent): surface active extension state to the LLM

### LINKED ISSUE #1595: [QA] LLM forgets activated Telegram channel and asks users to enable it again

henrypark133:
## Summary
- inject a compact runtime extension snapshot into the per-turn reasoning prompt
- teach the extension guidance to inspect `tool_list` and `extension_info` before asking users to reconnect or reactivate integrations
- add regression coverage for prompt injection and hot-activated Telegram channel state

Closes #1595

## Verification
- cargo fmt --check
- cargo clippy --all-targets --all-features -- -D warnings
- cargo test test_system_prompt_extension_guidance_prefers_inspection_before_reconnect --lib
- cargo test test_system_prompt_includes_extension_runtime_summary --lib
- cargo test test_llm_extension_state_summary_reports_active_owner_bound_telegram --lib

### DIFF
--- a/src/agent/dispatcher.rs
+++ b/src/agent/dispatcher.rs
@@ -137,9 +137,7 @@ impl Agent {
             }
         }
 
-        if !is_group_chat
-            && let Some(extension_manager) = self.deps.extension_manager.as_ref()
-        {
+        if !is_group_chat && let Some(extension_manager) = self.deps.extension_manager.as_ref() {
             match extension_manager
                 .llm_extension_state_summary(&message.user_id)
                 .await
@@ -2442,10 +2440,8 @@ mod tests {
         )
     }
 
-    fn make_test_extension_manager() -> (
-        Arc<crate::extensions::ExtensionManager>,
-        tempfile::TempDir,
-    ) {
+    fn make_test_extension_manager() -> (Arc<crate::extensions::ExtensionManager>, tempfile::TempDir)
+    {
         use crate::secrets::{InMemorySecretsStore, SecretsCrypto};
         use crate::tools::mcp::process::McpProcessManager;
         use crate::tools::mcp::session::McpSessionManager;
--- a/src/agent/dispatcher.rs
+++ b/src/agent/dispatcher.rs
@@ -137,6 +137,23 @@ impl Agent {
             }
         }
 
+        if !is_group_chat
+            && let Some(extension_manager) = self.deps.extension_manager.as_ref()
+        {
+            match extension_manager
+                .llm_extension_state_summary(&message.user_id)
+                .await
+            {
+                Ok(Some(summary)) => {
+                    reasoning = reasoning.with_extension_state_summary(summary);
+                }
+                Ok(None) => {}
+                Err(e) => {
+                    tracing::debug!("Could not load extension state summary: {}", e);
+                }
+            }
+        }
+
         if let Some(prompt) = system_prompt {
             reasoning = reasoning.with_system_prompt(prompt);
         }
@@ -1351,6 +1368,72 @@ mod tests {
         }
     }
 
+    #[derive(Default)]
+    struct RecordingLlmProvider {
+        last_system_prompt: std::sync::Mutex<Option<String>>,
+    }
+
+    impl RecordingLlmProvider {
+        fn capture_system_prompt(&self, messages: &[crate::llm::ChatMessage]) {
+            if let Some(message) = messages
+                .iter()
+                .find(|message| matches!(message.role, crate::llm::Role::System))
+            {
+                *self.last_system_prompt.lock().expect("system prompt lock") =
+                    Some(message.content.clone());
+            }
+        }
+
+        fn last_system_prompt(&self) -> Option<String> {
+            self.last_system_prompt
+                .lock()
+                .expect("system prompt lock")
+                .clone()
+        }
+    }
+
+    #[async_trait]
+    impl LlmProvider for RecordingLlmProvider {
+        fn model_name(&self) -> &str {
+            "recording-mock"
+        }
+
+        fn cost_per_token(&self) -> (Decimal, Decimal) {
+            (Decimal::ZERO, Decimal::ZERO)
+        }
+
+        async fn complete(
+            &self,
+            request: CompletionRequest,
+        ) -> Result<CompletionResponse, crate::error::LlmError> {
+            self.capture_system_prompt(&request.messages);
+            Ok(CompletionResponse {
+                content: "ok".to_string(),
+                input_tokens: 0,
+                output_tokens: 0,
+                finish_reason: FinishReason::Stop,
+                cache_read_input_tokens: 0,
+                cache_creation_input_tokens: 0,
+            })
+        }
+
+        async fn complete_with_tools(
+            &self,
+            request: ToolCompletionRequest,
+        ) -> Result<ToolCompletionResponse, crate::error::LlmError> {
+            self.capture_system_prompt(&request.messages);
+            Ok(ToolCompletionResponse {
+                content: Some("ok".to_string()),
+                tool_calls: Vec::new(),
+                input_tokens: 0,
+                output_tokens: 0,
+                finish_reason: FinishReason::Stop,
+                cache_read_input_tokens: 0,
+                cache_creation_input_tokens: 0,
+            })
+        }
+    }
+
     /// Build a minimal `Agent` for unit testing (no DB, no workspace, no extensions).
     fn make_test_agent() -> Agent {
         let deps = AgentDeps {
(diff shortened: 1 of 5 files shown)


src/agent/dispatcher.rs
```diff
@@ -131,6 +131,18 @@ impl Agent {
             }
         }
 
+        if let Some(extension_manager) = self.deps.extension_manager.as_ref() {
+            match extension_manager.llm_extension_state_summary().await {
```
serrrfirat: **High Severity**

This new prompt injection runs for every conversational turn, including group chats. That breaks the repo's existing privacy boundary: group-chat prompts explicitly omit personal context like `MEMORY.md`, and the system prompt also tells the model not to share the user's private context in group threads.

With this change, other participants can now probe whether the owner has Telegram/Slack/MCP integrations connected, even though that state was previously kept out of shared prompts.

Please gate `llm_extension_state_summary()` behind `!is_group_chat`, or reduce it to only the current channel's own state, and add a regression test proving group-chat prompts do not include per-user extension state.

src/extensions/manager.rs
```diff
@@ -1498,6 +1498,52 @@ impl ExtensionManager {
...
+        let mut servers = Vec::new();
+
+        for extension in extensions {
+            let owner_bound = matches!(extension.kind, ExtensionKind::WasmChannel)
+                && self.has_wasm_channel_owner_binding(&extension.name).await;
+            if !(extension.active || extension.authenticated || owner_bound) {
+                continue;
+            }
+
+            let item = llm_extension_summary_item(&extension, owner_bound);
```
serrrfirat, line 1696: **Medium Severity**

This helper now copies extension names and tool names directly into the system prompt, but extension names are still only validated for path traversal elsewhere. `tool_install` accepts custom names, so a crafted extension name containing newlines or prompt-looking text would become persistent prompt-injection surface once it is summarized here each turn.

Please either harden extension-name validation to a strict safe charset or serialize this summary as escaped data rather than free-form prose. A regression test with markup/newline-bearing names would help lock this down.

src/extensions/manager.rs
```diff
@@ -5613,6 +5659,40 @@ fn combine_install_errors(
...
+    }
+    if extension.active {
+        states.push("active".to_string());
+    } else if extension.authenticated {
+        states.push("inactive".to_string());
+    }
+    if owner_bound {
+        states.push("owner-bound".to_string());
+    }
+    if !extension.tools.is_empty() {
```
serrrfirat, line 6490: **Medium Severity**

This helper appends every loaded extension tool name into the extension-state summary. The prompt already includes the normal `## Available Tools` section, so active MCP servers now duplicate a potentially large tool inventory twice on every turn.

For servers with many tools this can materially increase prompt size and context pressure while adding very little new signal. Consider summarizing this as a count/sample (for example, `42 tools loaded`) instead of listing every tool name here.

src/extensions/manager.rs
```diff
@@ -1498,6 +1498,52 @@ impl ExtensionManager {
         Ok(extensions)
     }
 
+    /// Build a compact, deterministic extension snapshot for LLM prompt context.
+    pub async fn llm_extension_state_summary(&self) -> Result<Option<String>, ExtensionError> {
+        let mut extensions = self.list(None, false).await?;
```
serrrfirat: **Medium Severity**

This runs on the hot path of every chat turn, but it is implemented by calling the full `list(None, false)` path. That rescans MCP config, rediscovers WASM tools/channels, and rereads capabilities just to build prompt context, so chat latency now scales with installed-extension count.

A dedicated cached summary updated on install/configure/activate/remove would avoid paying the full listing cost on every turn.

ilblackdragon (changes requested):
## Review: fix(agent): surface active extension state to the LLM

### Blocking

1. **Won't compile** — `self.list(None, false).await?` passes 2 arguments but `list()` requires 3 (`user_id: &str` is missing). Add `user_id: &str` to `llm_extension_state_summary()` and forward it.

2. **Existing test will fail** — `test_extensions_section_clarifies_channels_are_not_send_tools` asserts on text the new implementation no longer produces ("Channels are not separate send-message tools"). Update or rewrite.

3. **Channel/message-tool boundary guidance dropped** — the old text prevented the LLM from misusing the `message` tool for normal replies. There's a dedicated test for it. Restore this guidance in the new section.

### High

4. **Group chat privacy leak** — extension state is injected unconditionally, including group chats. The codebase omits personal context (like `MEMORY.md`) from group prompts to prevent leaking private data. The extension summary reveals which integrations the owner has connected. Skip injection when `is_group_chat` is true.

5. **Prompt injection surface** — extension names are injected into the system prompt without sanitization. Strip control characters and newlines, or serialize in a structured format.

6. **Unbounded tool list** — MCP servers can expose dozens of tools. `extension.tools.join(", ")` is redundant with the `## Available Tools` section and bloats tokens. Show a count or capped sample.

### Medium

7. **Performance** — `self.list()` does filesystem discovery + auth checks on every turn. Consider caching with invalidation on install/configure/activate events.

src/agent/dispatcher.rs
henrypark133: Fixed in `c2d3da22`. Group chats now skip per-user extension-state injection entirely, and I added dispatcher regressions for both private-chat inclusion and group-chat exclusion.

src/extensions/manager.rs
```diff
@@ -427,6 +427,8 @@ pub struct ExtensionManager {
     installed_relay_extensions: RwLock<HashSet<String>>,
     /// Last activation error for each WASM channel (ephemeral, cleared on success).
     activation_errors: RwLock<HashMap<String, String>>,
+    /// Cached per-user extension summaries used in LLM prompt injection.
+    llm_extension_state_summaries: RwLock<HashMap<String, Option<String>>>,
```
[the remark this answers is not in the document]
henrypark133: Fixed in 21a06ceb. The per-user LLM summary cache now goes through a bounded helper capped at 128 users; when a new user would overflow the cache, it clears before inserting the new entry so it cannot grow without bound.

src/extensions/manager.rs
```diff
@@ -149,6 +149,7 @@ const TELEGRAM_OWNER_BIND_TIMEOUT_SECS: u64 = 120;
 const TELEGRAM_OWNER_BIND_CHALLENGE_TTL_SECS: u64 = 300;
 const TELEGRAM_GET_UPDATES_TIMEOUT_SECS: u64 = 25;
 const TELEGRAM_OWNER_BIND_CODE_LEN: usize = 8;
+const LLM_EXTENSION_STATE_SUMMARY_CACHE_MAX_USERS: usize = 128;
```
serrrfirat, line 153: **🟡 Medium Severity — Cache Staleness**

**`auth()` does not invalidate the LLM extension state cache.**

After OAuth completion stores a token, the cached summary still shows the extension as unauthenticated until the cache naturally expires.

**Fix:** Add `self.invalidate_llm_extension_state_summary(user_id).await;` after successful auth.

src/extensions/manager.rs
```diff
@@ -427,6 +428,8 @@ pub struct ExtensionManager {
     installed_relay_extensions: RwLock<HashSet<String>>,
     /// Last activation error for each WASM channel (ephemeral, cleared on success).
     activation_errors: RwLock<HashMap<String, String>>,
+    /// Cached per-user extension summaries used in LLM prompt injection.
+    llm_extension_state_summaries: RwLock<HashMap<String, Option<String>>>,
```
serrrfirat, line 432: **🟡 Medium Severity — Cache Staleness**

**`set_active_channels()` does not invalidate the LLM extension state cache.**

If this method is ever called at runtime (not just startup), the LLM would see stale channel state. Currently appears startup-only, but the public API doesn't enforce that.

**Fix:** Either add cache invalidation, or make it `pub(crate)` with a doc comment noting it's startup-only.

src/extensions/manager.rs
```diff
@@ -6333,6 +6454,71 @@ fn combine_install_errors(
...
+    }
+
+    serde_json::Value::Object(item)
+}
+
+fn sanitize_llm_summary_text(value: &str) -> String {
+    let collapsed = value
+        .chars()
+        .filter(|c| !c.is_control())
+        .map(|c| {
```
serrrfirat, line 6504: **🟡 Medium Severity — Security / Prompt Injection**

**Name sanitization allowlist mismatches `validate_extension_name`.**

`sanitize_llm_summary_text` allows `/`, `:`, `@`, `.` while `validate_extension_name` has a different allowlist. Names with `#`, `*`, backtick, `<`, `>` could get installed but then silently mutated (replaced with `_`) in the LLM summary, creating a mismatch between actual extension name and what the LLM sees.

**Suggestion:** Tighten `validate_extension_name` to match the same allowlist, or document the intentional mismatch.

src/extensions/manager.rs
```diff
@@ -1614,6 +1660,73 @@ impl ExtensionManager {
         Ok(extensions)
     }
 
+    /// Build a compact, deterministic extension snapshot for LLM prompt context.
+    pub async fn llm_extension_state_summary(
+        &self,
+        user_id: &str,
+    ) -> Result<Option<String>, ExtensionError> {
+        if let Some(summary) = self
```
serrrfirat, line 1668: **🟡 Medium Severity — DoS / Unbounded Size**

**No total size limit on the extension summary injected into the LLM system prompt.**

A user with many installed extensions (50+ MCP servers) could produce a summary consuming significant context window. 100 extensions ≈ 12KB of system prompt, eroding effective context.

**Suggestion:** Add a max item count (e.g., 20 most relevant) or a total byte budget, truncating with `"... and N more"`.

serrrfirat (approved):
## Paranoid Architect Review — APPROVE (with comments)

**4 Medium findings, no blockers.**

Well-designed feature. The sanitization of extension names before system prompt injection is the right approach, and the JSON code fence + "data only" framing provide good layered defense. Cache invalidation has two gaps (auth and set_active_channels) that should be addressed in a follow-up.

No `.unwrap()` in production code. Error handling is clean. Test coverage is solid.

Issue #1474, opened 2026-03-20
feat(agent): auto-extract structured memories from session conversations


ilblackdragon:
## Problem / Motivation

IronClaw's memory system is entirely manual: the agent must explicitly call `memory_write` to persist anything. In practice, this means most session context evaporates. Users repeat preferences, re-explain project context, and re-state goals across sessions because the agent forgot.

The compaction system (`src/agent/compaction.rs`) writes raw turn transcripts or LLM summaries to `daily/{date}.md`, but these are unstructured blobs — they capture *what happened* but don't extract *what should be remembered*.

OpenViking solves this with **automatic session memory extraction**: at session end, an LLM pass extracts six categories of structured memories (user profile, preferences, entities, events, agent-learned cases, and patterns), then deduplicates them against existing memories using vector similarity before writing.

IronClaw needs this. The infrastructure is already in place — `OnSessionEnd` hook point exists, `Workspace` has `memory_write`, and the `profile.rs` analysis framework defines how to understand users. We just need to connect them.

### New module: `src/agent/memory_extractor.rs`

A `MemoryExtractor` struct that takes a conversation transcript and produces structured memory entries.

```rust
pub struct MemoryExtractor {
    llm: Arc<dyn LlmProvider>,
    workspace: Arc<Workspace>,
}

pub struct ExtractedMemories {
    pub user_preferences: Vec<MemoryEntry>,
    pub user_entities: Vec<MemoryEntry>,
    pub events_decisions: Vec<MemoryEntry>,
    pub learned_patterns: Vec<MemoryEntry>,
    pub profile_updates: Option<ProfilePatch>,
}

pub struct MemoryEntry {
    pub content: String,
    pub category: MemoryCategory,
    pub confidence: f32,     // 0.0-1.0
    pub source_turn: usize,  // which turn this came from
}

pub enum MemoryCategory {
    UserPreference,   // "prefers dark mode", "dislikes verbose responses"
    UserEntity,       // "works at Acme Corp", "uses PostgreSQL daily"
    EventDecision,    // "decided to use React for frontend on 2026-03-15"
    LearnedPattern,   // "when user asks about X, they usually need Y context"
}

pub struct ProfilePatch {
    pub preferred_name: Option<String>,
    pub communication_updates: serde_json::Value,
    pub cohort_updates: serde_json::Value,
}
```

### Extraction prompt

The core LLM prompt for memory extraction. This runs at session end, receiving the full conversation transcript:

```
[...]

PR #1603, opened 2026-03-23
feat(workspace): auto-extract structured memories on compaction

G7CNF:
Implements the first slice of #1474 by extracting high-confidence structured memories during compaction and appending them to MEMORY.md.

What changed:
- compaction now asks the LLM for structured durable memories from archived turns
- results are deduped against existing structured memory entries in MEMORY.md
- extraction is best-effort and does not block compaction if the LLM or append path fails
- FEATURE_PARITY.md updated to reflect the new extraction capability

Validation:
- cargo fmt --all -- --check
- CARGO_INCREMENTAL=0 CARGO_PROFILE_TEST_DEBUG=0 RUSTFLAGS='-Cdebuginfo=0' cargo test -p ironclaw --features libsql --lib test_move_to_workspace_extracts_structured_memories -- --nocapture
- CARGO_INCREMENTAL=0 CARGO_PROFILE_TEST_DEBUG=0 RUSTFLAGS='-Cdebuginfo=0' cargo test -p ironclaw --features libsql --lib test_move_to_workspace_skips_duplicate_structured_memories -- --nocapture
- IC_BASE_REF=origin/staging /Users/nigecoleman/.codex/skills/rust-pr-preflight/scripts/preflight.sh (repo-wide baseline boundary warnings remain, but no branch-specific blockers)

This is intentionally the minimal first pass for #1474; follow-up work can layer compaction/pruning polish on top.


zmanian (commented):
## Review: Structured Memory Extraction at Compaction

Overall this is a clean, well-scoped first pass. The best-effort pattern is solid, the LLM prompt is well-constrained, and the test coverage is good. A few items worth addressing before merge:

### Dedup logic (Gemini flag) -- addressed

The second commit correctly fixes the dedup to extract content from `- **kind**: content` lines before comparison. The current version using `split_once("**: ")` is correct.

### Issues to address

**1. `extract_json_object` is fragile with nested braces**

The `extract_json_object` function uses `text.find('{')` / `text.rfind('}')` which works for the simple schema here, but will silently produce garbage if the LLM wraps the response in markdown fences containing an explanation with braces before/after the JSON. Since the LLM is asked to return "ONLY JSON", this is low risk, but a more robust approach would be to find the *outermost balanced* braces, or at minimum add a `serde_json::from_str` validation inside `extract_json_object` before returning. Not a blocker, but worth a comment or a follow-up.

**2. Dedup reads the entire MEMORY.md on every candidate memory**

`structured_memory_exists` calls `workspace.memory().await` for each candidate memory in the loop (up to 5 times). This re-reads MEMORY.md from the database on every call. Should read once before the loop and pass the content in. Something like:

```rust
let memory_content = workspace.memory().await.map(|d| d.content).unwrap_or_default();
for memory in parsed.memories.into_iter().take(MEMORY_EXTRACTION_MAX_CANDIDATES) {
    // ...
    if memory_already_exists(&memory_content, &memory.content) {
        continue;
    }
    // ...
}
```

This is a real performance issue -- 5 DB reads when 1 suffices.

**3. `kind` field is not validated**

The `kind` field from the LLM is used directly as a string in the formatted output (`- **{kind}**: ...`). If the LLM hallucinates a kind like `"personal_secret"` or returns garbage, it gets written verbatim. Consider validating against the known enum variants (`preference | identity | project | plan | fact`) and skipping unknown kinds, or at least documenting this as a known limitation for the follow-up.

**4. Substring dedup can produce false positives**

The dedup uses bidirectional `contains()`:
```rust
existing.contains(&normalized_query) || normalized_query.contains(&existing)
```
A memory like "User uses Rust" would match an existing "User uses Rust and Python for backend work" which is correct (subsumption). But it also means "User prefers dark mode" would match an existing line containing "User prefers dark mode for IDE but light mode for docs" -- the new shorter memory would be incorrectly suppressed even though it's semantically different (broader). This is acceptable for a first pass but worth noting for the follow-up pruning/compaction work.

### Looks good

- Best-effort error handling: all failure paths log warnings and return 0, never blocking compaction.
- No `.unwrap()`/`.expect()` in production code.
- Temperature 0.0 and max_tokens 512 are sensible for structured extraction.
- Confidence threshold at 0.75 is reasonable.
- `#[serde(default)]` on evidence field handles LLMs that omit optional fields.
- Tests cover both extraction and dedup paths with real workspace instances.
- The Gemini bot feedback has been addressed in the second commit (dedup content extraction, compact_truncate refactoring, serde error logging).

### Verdict

Request changes for item #2 (reading MEMORY.md N times in a loop). The rest are minor suggestions or follow-up items. With that fix, this is good to merge.

zmanian (changes requested):
Requesting changes for the repeated MEMORY.md reads in the dedup loop (see detailed review comment). The rest is solid work.

serrrfirat:
Thank you for pushing structured durable-memory extraction forward.\n\nWe are closing this legacy implementation because Reborn has separated compaction from memory lifecycle ownership. Compaction owns transcript-window reduction; memory providers own interaction recording and any extraction or distillation policy. The active Reborn memory-lifecycle work in #5327 records completed-turn interactions and explicitly defers durable on-run-end summaries to Phase 3.\n\nReferences:\n- https://github.com/nearai/ironclaw/pull/5327\n- https://github.com/nearai/ironclaw/blob/main/crates/ironclaw_agent_loop/src/strategies/compaction.rs\n- https://github.com/nearai/ironclaw/blob/main/crates/ironclaw_memory/src/service.rs\n- https://github.com/nearai/ironclaw/blob/main/crates/ironclaw_memory_native/src/service.rs\n\nThe valuable extraction idea should continue as a provider-owned Phase 3 follow-up to #5327, not as a side effect inside the legacy compaction path. We would be very happy to have you contribute to that Reborn work, with this PR retained as context and attribution.

PR #1615, opened 2026-03-24
refactor: extract AppEvent to crates/ironclaw_common

ilblackdragon:
## Summary

- **New crate `crates/ironclaw_common`** with `AppEvent` (renamed from `SseEvent`) and `truncate_preview` — shared types/utils that were leaking from the web gateway into 12+ modules
- **Rename `SseEvent` → `AppEvent`** across 23 files — the enum was the app-wide event protocol, not a web transport concern
- **`from_sse_event()` → `from_app_event()`** on `WsServerMessage`
- **Wire format unchanged** — serde renames are on variants, not the enum name
- `web/types.rs` and `web/util.rs` re-export from `ironclaw_common` for backward compat within the gateway

### Motivation

`SseEvent` was defined in `src/channels/web/types.rs` but imported by agent, orchestrator, worker, tools, extensions, and CLI modules. This coupling would block extracting the web gateway into its own crate. Moving the event type to a neutral shared crate decouples the event protocol from the transport.

Aligned with the event bus direction on `refactor/architectural-hardening` where `DomainEvent` (≡ `AppEvent`) is wrapped in a `SystemEvent` envelope with category-based sink routing.

## Test plan

- [x] `cargo check` — clean compile
- [x] `cargo test -p ironclaw_common` — 11 tests pass (truncate_preview)
- [x] `cargo test -p ironclaw --lib -- channels::web` — 127 gateway tests pass
- [x] `cargo test -p ironclaw --lib -- job_monitor` — 9 tests pass
- [x] Full `cargo test` suite — 3576+ tests pass, 0 failures

### DIFF
--- a/crates/ironclaw_common/src/event.rs
+++ b/crates/ironclaw_common/src/event.rs
@@ -5,9 +5,9 @@
 //! frames, but other subsystems (agent loop, orchestrator, extensions)
 //! produce and consume them too.
 
-use serde::Serialize;
+use serde::{Deserialize, Serialize};
 
-#[derive(Debug, Clone, Serialize)]
+#[derive(Debug, Clone, Serialize, Deserialize)]
 #[serde(tag = "type")]
 pub enum AppEvent {
     #[serde(rename = "response")]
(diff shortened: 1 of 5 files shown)


zmanian (commented):
## Review: refactor: extract AppEvent to crates/ironclaw_common

**+500/-477 across 24 files**

### Issues (ranked by severity)

**1. No `Deserialize` derive on `AppEvent` (Medium)**
The new `AppEvent` only derives `Serialize` + `Debug` + `Clone`. The old `SseEvent` also only had `Serialize`, but now that `AppEvent` lives in a shared crate intended for reuse by other workspace members, downstream consumers (e.g. test harnesses, CLI tools, external integrations) will likely need to deserialize incoming events. Adding `Deserialize` now avoids a semver-ish breaking change to the shared crate later.

**2. `event_type()` manually duplicates serde rename values (Medium)**
The `event_type()` match arm strings must stay in sync with the `#[serde(rename = "...")]` attributes. If a variant is added and the developer forgets to update `event_type()`, the compiler won't catch it (the match is exhaustive on variants, but the string could be wrong). Consider deriving the event type string from serde metadata or adding a test that round-trips serialization and asserts `event_type()` matches the `"type"` field in the JSON output.

**3. Stale variable/comment references to "SSE" remain (Low)**
Several comments and at least one variable still reference "SSE" — e.g. `src/channels/web/server.rs:1203` still says "Broadcast SSE event", and `src/worker/job.rs` doc comment was only partially updated. Copilot and Gemini already flagged specific instances. Not a functional issue, but undermines the refactor's goal of decoupling from SSE terminology.

**4. `truncate_preview` moved but public API surface unchanged (Low)**
The old `crate::channels::web::util::truncate_preview` is now a re-export of `ironclaw_common::truncate_preview`. This is clean, but the re-export means two import paths work — both the old path (via `pub use`) and the new `ironclaw_common::truncate_preview`. This is fine for backward compat, but consider deprecating the old path to guide callers toward the canonical import.

**5. `ironclaw_common` crate uses `edition = "2024"` and `rust-version = "1.92"` (Nit)**
Just confirming this is intentional and aligns with the workspace's MSRV. If the workspace targets an older MSRV, this could cause issues for contributors on older toolchains.

### What's good

- Clean mechanical rename with zero semantic changes — the serde wire format is identical (same `#[serde(rename)]` values), so this is fully backward-compatible on the wire.
- The `event_type()` helper consolidates three duplicate match blocks (SSE, WS, types) into one, reducing ~70 lines of duplication.
- Tests were updated consistently across all 24 files.
- `truncate_preview` extraction with comprehensive unit tests in the new crate is solid.
- `publish = false` on the new crate prevents accidental crates.io publishing.

### Verdict

**Approve with suggestions** — This is a clean, low-risk refactor. The wire format is unchanged, so there are no breaking changes for clients. The suggestions above (especially adding `Deserialize` and a round-trip test for `event_type()`) would strengthen the shared crate for future consumers but are not blockers.

ilblackdragon:
Addressed zmanian's review in e0cad3b2:

1. **Deserialize on AppEvent** — Added `Deserialize` derive so downstream consumers can deserialize events. ✅
2. **Round-trip test for event_type()** — `event_type_matches_serde_type_field` serializes every variant, parses the JSON `"type"` field, and asserts it matches `event_type()`. Catches drift between serde renames and the manual match. Also added `round_trip_deserialize` test. ✅
3. **Stale SSE comments** — Updated remaining "SSE" references in `server.rs`, `manager.rs`, `ws_gateway_integration.rs`, and `worker/job.rs`. ✅
4. **Re-export deprecation** — Rust doesn't support `#[deprecated]` on `pub use` re-exports. Both callers (`session.rs`, `thread_ops.rs`) already import from `ironclaw_common` directly, so the old path is only used within the web gateway itself. No action needed.
5. **Edition/MSRV** — Confirmed aligned with workspace (`edition = "2024"`, `rust-version = "1.92"`). ✅

PR #1617, opened 2026-03-24
feat(security): add outbound trust policy for wasm and mcp

Boomboomdunce:
## Summary

Adds an operator-managed outbound trust policy system for narrowly scoped HTTPS exceptions on extension-style outbound surfaces.

This change:
- adds a shared `outbound_trust` policy model under `settings.security`
- requires dual opt-in before any exception is granted:
  - operator config must define a matching policy
  - the consuming WASM tool, WASM channel, or MCP server must explicitly declare the allowed policy id
- supports two explicit risk flags:
  - `allow_invalid_tls`
  - `allow_private_network`
- wires the shared resolver into:
  - WASM tool HTTP
  - WASM channel HTTP
  - MCP HTTP transport and MCP OAuth/discovery HTTP flows
- keeps existing defaults unchanged:
  - HTTPS remains required for remote MCP
  - plain HTTP remains limited to localhost development targets
  - private/internal network targets remain blocked by default
  - invalid TLS certificates remain rejected by default
- adds focused tests, including self-signed HTTPS integration coverage for WASM tools, WASM channels, and MCP servers

## Why

IronClaw already has strong outbound network defaults, but operators still need a manageable way to allow a small set of internal HTTPS integrations that use private network targets or private CA / self-signed certificates.

This PR adds that capability without introducing a global bypass. Exceptions remain:
- operator-managed
- surface-scoped
- target-scoped
- risk-scoped
- extension/server-declared

## Design Notes

Policies are matched by exact host, optional port, and optional path prefix.

A request only gets an exception when all of the following are true:
- outbound trust is globally enabled
- a matching policy exists
- the policy is enabled
- the policy allows the current surface
- the request target matches the policy target tuple
- the extension/server declared that policy id in its own capabilities/config

This keeps policy ownership with the operator while preventing arbitrary extension code from inheriting broad trust exceptions.

## Test Plan

- [x] `cargo fmt --all --check`
- [x] `cargo test outbound_trust --lib -- --nocapture`
- [x] `cargo test --test outbound_trust_resolver -- --nocapture`
- [x] `cargo clippy --all --tests --all-features -- -D warnings`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -157,7 +157,7 @@ version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
 dependencies = [
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -168,7 +168,7 @@ checksum = "291e6a250ff86cd4a820112fb8898808a366d8f9f58ce16d1f538353ad55747d"
 dependencies = [
  "anstyle",
  "once_cell_polyfill",
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -2136,7 +2136,7 @@ dependencies = [
  "libc",
  "option-ext",
  "redox_users 0.5.2",
- "windows-sys 0.59.0",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -2323,7 +2323,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "39cab71617ae0d63f51a36d69f866391735b51691dbda63cf6f96d042b63efeb"
 dependencies = [
  "libc",
- "windows-sys 0.52.0",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -3150,7 +3150,7 @@ dependencies = [
  "libc",
  "percent-encoding",
  "pin-project-lite",
- "socket2 0.5.10",
+ "socket2 0.6.3",
  "system-configuration",
  "tokio",
  "tower-service",
@@ -3439,6 +3439,7 @@ dependencies = [
  "postgres-types",
  "pretty_assertions",
  "rand 0.8.5",
+ "rcgen",
  "readabilityrs",
  "refinery",
  "regex",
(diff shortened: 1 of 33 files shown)


zmanian (commented):
## Security Review: outbound trust policy for WASM and MCP

Well-designed security feature. The dual opt-in model (operator policy + extension declaration) is sound, defaults are safe, and the implementation is thorough across all three surfaces. Approve with one HIGH to address before merge.

### Security Model Assessment

The model is correct: dual opt-in, surface-scoped, target-scoped, risk-scoped, with a global kill switch. Defaults unchanged (HTTPS required, private networks blocked). No backward compatibility issues.

### High

1. **Divergent private-IP checks**: Two separate functions cover overlapping but different address ranges:
   - `channels/wasm/wrapper.rs` -- `is_private_ip()` (new)
   - `tools/mcp/network.rs` -- `is_dangerous_ip()` (moved)

   `is_dangerous_ip` checks broadcast, IPv6 site-local, documentation ranges, and IPv4-mapped IPv6 (`::ffff:x.x.x.x`). The WASM `is_private_ip` checks none of these. An attacker-controlled DNS returning `::ffff:10.0.0.1` bypasses the WASM check but not MCP.

   **Fix:** Extract a single canonical function into `src/security/outbound_trust.rs` using the MCP version as baseline.

### Medium

2. **TOCTOU between DNS resolution and connection**: Both WASM and MCP surfaces resolve DNS pre-flight then make a separate connection. DNS rebinding can bypass the check. Pre-existing pattern -- track for follow-up.

3. **Blocking DNS in WASM channels**: `reject_private_ip()` uses `std::net::ToSocketAddrs` (blocking). Fine since callers use `spawn_blocking`, but add a doc comment documenting this contract.

4. **Three near-identical HTTP client builders**: `build_channel_http_client`, `build_wasm_http_client`, `build_mcp_http_client` differ only in timeout and error message. Extract shared builder with timeout parameter.

5. **`OutboundTrustResolver` re-created per request in MCP auth**: Pass `&OutboundTrustResolver` instead of `&OutboundTrustConfig` into auth functions.

### Low

6. `port_or_known_default()` means `port: Some(443)` matches implicit HTTPS port -- correct but worth documenting.
7. `exchange_code_for_token` changed from `pub` to private -- verify no downstream consumers.
8. `default_true` helper duplicated across 3 files -- consider shared utility.

### Verdict

Approve with suggestions. Item 1 (divergent IP checks) should be addressed before merge to close the security gap.

ilblackdragon:
Closing this because it updates the legacy top-level `src/` security, WASM, and MCP implementation (`src/security/*`, `src/tools/mcp/*`, `src/tools/wasm/*`, `src/channels/wasm/*`). Product work has moved to the Reborn workspace under `crates/`, so this PR is outdated as an implementation path.

I opened #6312 to track the outbound trust policy as Reborn-native work with implementation notes.

PR #1621, opened 2026-03-24
fix(wasm): avoid persisting dedup before message delivery succeeds

a3131919:
## Problem

WASM channels that write a dedup key (`workspace_write("state/dedup/<event_id>")`)
before emitting a message can permanently lose that message.

In `call_on_http_request` and `call_on_poll`, `commit_writes` runs inside
`spawn_blocking` — before `process_emitted_messages`. If the agent queue is
full and `tx.send().await` blocks or fails, the dedup key is already
persisted. The next platform retry sees the key, treats the event as a
duplicate, and drops it. The user never receives a reply.

A secondary issue: `tx.send().await` has no timeout, so a stalled receiver
can silently block the WASM callback with no error surfaced.

## Fix

**Primary:** move `commit_writes` out of `spawn_blocking` and call it only
after `process_emitted_messages` succeeds. On failure the error propagates
as-is (HTTP 500 for webhooks, skipped tick for polling), the dedup key stays
uncommitted, and the platform retry is handled correctly.

The same reorder is applied to `execute_poll` and the polling dispatch loop
for consistency.

**Supporting safeguard:** add a two-stage Pending → Delivered lifecycle to
`ChannelWorkspaceStore` for dedup paths. `commit_writes` marks them Pending;
`promote_pending_to_delivered` (called after a successful send) marks them
Delivered. TTLs prevent a crashed Pending entry from permanently blocking
future retries (60 s) while keeping the delivered idempotency window long
enough for platform retry windows (24 h). The plain KV behavior for all
other workspace paths is unchanged.

`tx.send` now uses a 10-second timeout. A stalled send returns
`WasmChannelError::SendTimeout` instead of blocking indefinitely.

## Behavior

This change preserves current channel behavior, which assumes at most one
emitted message per webhook or poll tick. Channels that do not write dedup
keys are unaffected.

### DIFF
--- a/src/channels/wasm/host.rs
+++ b/src/channels/wasm/host.rs
@@ -6,7 +6,7 @@
 //! - Rate limiting for message emission
 
 use std::collections::HashMap;
-use std::time::{SystemTime, UNIX_EPOCH};
+use std::time::{Duration, Instant, SystemTime, UNIX_EPOCH};
 
 use crate::channels::wasm::capabilities::{ChannelCapabilities, EmitRateLimitConfig};
 use crate::channels::wasm::error::WasmChannelError;
@@ -502,6 +502,43 @@ impl ChannelHostState {
     }
 }
 
+/// Path infix that identifies dedup workspace keys.
+///
+/// WASM channels write event dedup keys under `state/dedup/<event_id>`.
+/// After the channel namespace prefix is applied the full path looks like
+/// `channels/<name>/state/dedup/<event_id>`.  We detect dedup writes by
+/// scanning for this infix so the dedup lifecycle can be tracked separately
+/// from ordinary workspace KV state.
+pub const DEDUP_PATH_INFIX: &str = "/state/dedup/";
+
+/// A Pending dedup entry expires after this many seconds if delivery never
+/// completes (e.g. process crash after commit but before promote).  Once
+/// expired the same event_id is allowed to be re-processed.
+const DEDUP_PENDING_TTL: Duration = Duration::from_secs(60);
+
+/// A Delivered dedup entry is kept for this long as the idempotency window.
+/// Webhook retries typically occur within ~30 minutes; 24 h is a safe margin.
+const DEDUP_DELIVERED_TTL: Duration = Duration::from_secs(24 * 3600);
+
+/// Lifecycle stage of a dedup entry.
+#[derive(Debug, Clone, PartialEq, Eq)]
+enum DedupStatus {
+    /// WASM callback committed the key but message delivery has not yet been
+    /// confirmed.  Treated as duplicate by concurrent requests to prevent
+    /// double-processing.  Expires after `DEDUP_PENDING_TTL` so a crash does
+    /// not permanently block future retries.
+    Pending,
+    /// Message was successfully delivered to the agent queue.  Treated as
+    /// duplicate for the full `DEDUP_DELIVERED_TTL` idempotency window.
+    Delivered,
+}
+
+#[derive(Debug, Clone)]
+struct DedupEntry {
+    status: DedupStatus,
+    created_at: Instant,
+}
+
 /// In-memory workspace store for WASM channels.
 ///
 /// Persists workspace writes across callback invocations within a single
@@ -511,8 +548,17 @@ impl ChannelHostState {
 ///
 /// Uses `std::sync::RwLock` (not tokio) because WASM execution runs
 /// inside `spawn_blocking`.
+///
+/// # Dedup sub-store
+///
+/// Writes to paths matching `*/state/dedup/*` are diverted from the plain
+/// KV `data` map into a separate `dedup` map that tracks a two-stage
+/// lifecycle: **Pending → Delivered**.  Reads for the same paths consult the
+/// dedup map with TTL logic so stale Pending entries do not permanently block
+/// retries after a crash.
 pub struct ChannelWorkspaceStore {
     data: std::sync::RwLock<std::collections::HashMap<String, String>>,
+    dedup: std::sync::RwLock<std::collections::HashMap<String, DedupEntry>>,
 }
 
 impl ChannelWorkspaceStore {
@@ -520,22 +566,61 @@ impl ChannelWorkspaceStore {
     pub fn new() -> Self {
         Self {
             data: std::sync::RwLock::new(std::collections::HashMap::new()),
+            dedup: std::sync::RwLock::new(std::collections::HashMap::new()),
         }
     }
 
+    /// Returns true if `path` is a dedup key (contains the dedup infix).
+    fn is_dedup_path(path: &str) -> bool {
+        path.contains(DEDUP_PATH_INFIX)
+    }
+
     /// Commit pending writes from a callback execution into the store.
+    ///
+    /// Writes whose path contains the dedup infix are stored as **Pending**
+    /// entries in the dedup sub-store rather than in the plain KV map.
+    /// Call [`promote_pending_to_delivered`] immediately after this to advance
+    /// them to **Delivered** once you have confirmed successful message
+    /// delivery.
+    ///
+    /// Also opportunistically evicts expired dedup entries.
     pub fn commit_writes(&self, writes: &[PendingWorkspaceWrite]) {
         if writes.is_empty() {
             return;
         }
+
+        self.cleanup_expired_dedup();
+
+        // Acquire locks separately so a poisoned dedup lock never silently
+        // drops ordinary KV writes that have nothing to do with dedup.
         if let Ok(mut data) = self.data.write() {
             for write in writes {
-                tracing::debug!(
-                    path = %write.path,
-                    content_len = write.content.len(),
-                    "Committing workspace write to channel store"
-                );
-                data.insert(write.path.clone(), write.content.clone());
+                if !Self::is_dedup_path(&write.path) {
+                    tracing::debug!(
+                        path = %write.path,
+                        content_len = write.content.len(),
+                        "Committing workspace write to channel store"
+                    );
+                    data.insert(write.path.clone(), write.content.clone());
+                }
+            }
+        }
+
+        if let Ok(mut dedup) = self.dedup.write() {
+            for write in writes {
+                if Self::is_dedup_path(&write.path) {
+                    tracing::debug!(
+                        path = %write.path,
+                        "Registering dedup key as Pending"
+                    );
+                    // Only insert if not already Delivered — a second webhook
+                    // arriving after a successful delivery must not downgrade
+                    // the status.
+                    dedup.entry(write.path.clone()).or_insert_with(|| DedupEntry {
+                        status: DedupStatus::Pending,
+                        created_at: Instant::now(),
+                    });
+                }
             }
         }
     }
@@ -598,10 +683,72 @@ impl ChannelWorkspaceStore {
         data.insert(dest_path.to_string(), raw_queue);
         Ok(true)
     }
+
+    /// Promote any dedup entries found in `writes` from **Pending** to
+    /// **Delivered**.
+    ///
+    /// Call this immediately after [`commit_writes`] once message delivery to
+    /// the agent queue has been confirmed.  Only keys whose current status is
+    /// `Pending` are updated; already-Delivered keys are left unchanged.
+    pub fn promote_pending_to_delivered(&self, writes: &[PendingWorkspaceWrite]) {
+        if writes.is_empty() {
+            return;
+        }
+        if let Ok(mut dedup) = self.dedup.write() {
+            for write in writes {
+                if Self::is_dedup_path(&write.path) {
+                    if let Some(entry) = dedup.get_mut(&write.path) {
+                        if entry.status == DedupStatus::Pending {
+                            tracing::debug!(
+                                path = %write.path,
+                                "Promoting dedup key from Pending to Delivered"
+                            );
+                            entry.status = DedupStatus::Delivered;
+                            entry.created_at = Instant::now();
+                        }
+                    }
+                }
+            }
+        }
+    }
+
+    /// Remove expired dedup entries.
+    ///
+    /// - `Pending` entries older than `DEDUP_PENDING_TTL` are dropped so a
+    ///   crashed delivery attempt does not permanently block future retries.
+    /// - `Delivered` entries older than `DEDUP_DELIVERED_TTL` are dropped to
+    ///   bound memory usage.
+    pub fn cleanup_expired_dedup(&self) {
+        let now = Instant::now();
+        if let Ok(mut dedup) = self.dedup.write() {
+            dedup.retain(|_, entry| {
+                let ttl = match entry.status {
+                    DedupStatus::Pending => DEDUP_PENDING_TTL,
+                    DedupStatus::Delivered => DEDUP_DELIVERED_TTL,
+                };
+                now.duration_since(entry.created_at) < ttl
+            });
+        }
+    }
 }
 
 impl crate::tools::wasm::WorkspaceReader for ChannelWorkspaceStore {
     fn read(&self, path: &str) -> Option<String> {
+        if Self::is_dedup_path(path) {
+            // For dedup paths, return "1" only if a non-expired entry exists.
+            // This allows WASM guests to use workspace_read to check dedup status
+            // while the dedup lifecycle is managed separately.
+            let dedup = self.dedup.read().ok()?;
+            let entry = dedup.get(path)?;
+            let ttl = match entry.status {
+                DedupStatus::Pending => DEDUP_PENDING_TTL,
+                DedupStatus::Delivered => DEDUP_DELIVERED_TTL,
+            };
+            if Instant::now().duration_since(entry.created_at) < ttl {
+                return Some("1".to_string());
+            }
+            return None;
+        }
         self.data.read().ok()?.get(path).cloned()
     }
 }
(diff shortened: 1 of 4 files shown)


zmanian (commented):
## Review: fix(wasm): avoid persisting dedup before message delivery succeeds

Well-structured fix for a real message-loss bug. The two-phase Pending/Delivered lifecycle is sound.

### Medium

1. **Partial delivery with multiple messages**: If a callback emits multiple messages and send fails partway, delivered messages lose their dedup keys (commit never called). The documented assumption of "at most 1 emitted message per callback" covers this, but consider adding `debug_assert!(messages.len() <= 1)` to catch violations early.

### Low

2. **`SendTimeout` conflates timeout and closed-channel**: Both actual timeouts (transient) and closed channels (shutdown) use the same error variant. Consider a separate `ChannelClosed` variant for distinguishability.

3. **`cleanup_expired_dedup` on every commit**: Acquires dedup write lock twice per commit. Under high webhook volume, a counter-based cleanup (every ~100 commits) would reduce contention.

4. **`preload_delivered_dedup` defined but not called**: Fine if intended for follow-up, but note it.

5. **No unit tests for new dedup lifecycle**: `ChannelWorkspaceStore` is straightforward to test in isolation. Tests for pending->delivered promotion, TTL expiry, and re-delivery blocking would be valuable.

### Verdict

Approve with suggestions. Core fix is correct -- `commit_writes` after successful delivery eliminates the premature dedup window.

serrrfirat:
All five review findings from @zmanian addressed in commit 06756431:

1. **debug_assert for single-message assumption** — Added `debug_assert!(emitted_count <= 1, ...)` in both `process_emitted_messages` and `dispatch_emitted_messages`.

2. **ChannelClosed vs SendTimeout** — Added `WasmChannelError::ChannelClosed` variant. Both `Ok(Err(_))` arms in the send paths now use `ChannelClosed` instead of `SendTimeout`, preserving `SendTimeout` for actual timeouts.

3. **Amortized cleanup** — Added `AtomicU32` commit counter. `cleanup_expired_dedup` now runs every 100 commits instead of on every call.

4. **preload_delivered_dedup** — Added `#[allow(dead_code)]` with a comment noting it's reserved for the startup preload follow-up.

5. **Dedup lifecycle tests** — Added 6 unit tests: pending→delivered promotion, no downgrade of Delivered, TTL expiry for Pending, read-returns-None for expired, plain KV isolation, and preload round-trip.

Also fixed pre-existing clippy warnings in the touched files (collapsible_if, new_without_default, is_multiple_of, GatewayConfig missing fields in tunnel tests).

ilblackdragon (changes requested):
## Additional Review Findings

Beyond zmanian's approval with suggestions:

### Must-fix

1. **Silent failure on poisoned lock** — `commit_writes` acquires both `data` and `dedup` locks simultaneously:
```rust
if let (Ok(mut data), Ok(mut dedup)) = (self.data.write(), self.dedup.write()) {
```
If the dedup lock is poisoned, the **entire commit is silently dropped** — including non-dedup writes that worked fine before this PR. Acquire locks separately so a poisoned dedup lock doesn't prevent regular KV writes.

2. **No tests for dedup lifecycle** — the `ChannelWorkspaceStore` dedup behavior has zero test coverage. Per project policy: "Regression test with every bug fix." Needed tests: dedup path recognition, Pending→Delivered transition, TTL expiry, non-dedup paths unaffected.

### Should-fix

3. **`SendTimeout` error variant misused** — when `tx.send(msg)` returns `Err(_)` (channel closed), the code returns `WasmChannelError::SendTimeout`. Channel closure is permanent, not a timeout. Add a separate `ChannelClosed` variant or rename to `SendFailed`.

4. **`preload_delivered_dedup` is dead code** — defined but never called. Add `#[allow(dead_code)]` or remove until follow-up.

5. **`SEND_TIMEOUT_SECS` defined inside two separate function bodies** — could drift independently. Make module-level constant.

6. **Dedup infix doesn't match any existing channel** — `DEDUP_PATH_INFIX = "/state/dedup/"` doesn't match Discord's `state/mention_dedup/{channel_id}` or any other channel. The two-stage lifecycle is speculative infrastructure. Consider deferring it and shipping just the reordering fix.

zmanian (changes requested):
## Review -- REQUEST_CHANGES

The core reorder fix (commit after delivery) correctly addresses the message-loss bug. However:

### Critical

1. **CI is red** -- Clippy and formatting checks all fail. Must be green before merge.

2. **Dual lock acquisition silently drops all writes** -- `commit_writes` acquires both `data` and `dedup` locks in a single `if let`:
   ```rust
   if let (Ok(mut data), Ok(mut dedup)) = (self.data.write(), self.dedup.write()) {
   ```
   If the dedup lock is poisoned but data is fine, ALL writes are silently dropped -- including non-dedup KV writes that worked correctly before this PR. This is a regression. Locks should be acquired separately.

### Medium

3. **`DEDUP_PATH_INFIX` doesn't match any existing channel** -- The infix `/state/dedup/` doesn't match Discord's `state/mention_dedup/{channel_id}`. The entire two-stage Pending/Delivered lifecycle has no current consumer. Consider shipping just the reorder fix and deferring the lifecycle.

4. **`SendTimeout` conflates closed channel and timeout** -- A closed channel is permanent/unrecoverable; a timeout is transient. Both return the same error variant. A `ChannelClosed` variant would let callers distinguish.

5. **`preload_delivered_dedup` is dead code** -- Defined but never called. Likely one of the Clippy failures.

6. **Missing regression tests** -- Per project rules: "Regression test with every bug fix." No tests for dedup path recognition, Pending->Delivered promotion, TTL expiry, or non-dedup paths being unaffected.

### Recommendation
Ship just the reorder + send timeout (part 1). Defer the Pending/Delivered lifecycle until a channel actually uses the `/state/dedup/` convention, with proper tests.

serrrfirat:
## Paranoid Architect Review

The core fix is correct and well-targeted: reordering `commit_writes` to run after `process_emitted_messages` prevents dedup keys from being committed before messages reach the agent queue. The Pending→Delivered lifecycle with TTLs is a clean design.

**What's good:**
- Root cause analysis is accurate and the fix is precise
- Pending→Delivered lifecycle with TTLs is well-designed
- Comments clearly document the ordering invariant and single-message assumption
- `or_insert_with` correctly prevents Delivered→Pending downgrades on duplicate webhooks
- Lock acquisition is split between data/dedup to prevent cross-contamination from poisoned locks
- Solid unit test coverage for the dedup lifecycle (7 tests)

**Findings: 3 Medium, 3 Low, 1 Nit** — inline comments posted for the 3 Mediums.

The most actionable finding is that `process_emitted_messages` and `dispatch_emitted_messages` both return `Ok(())` when no sender is available, causing the caller to commit dedup keys despite silently dropping messages — the exact pattern this PR aims to fix. The other two are acknowledged tradeoffs that would benefit from documentation.

src/channels/wasm/wrapper.rs
```diff
@@ -2410,7 +2428,7 @@ impl WasmChannel {
             let Some(tx) = tx_guard.as_ref() else {
                 tracing::error!(
                     channel = %self.name,
-                    count = messages.len(),
+                    count = emitted_count,
                     "Messages emitted but no sender available - channel may not be started!"
                 );
                 return Ok(());
```
serrrfirat, line 2434: **Medium Severity — Silent message loss via dedup commit**

`process_emitted_messages` returns `Ok(())` when `message_tx` is `None` (no sender available). The caller interprets `Ok` as successful delivery and proceeds with `commit_writes` + `promote_pending_to_delivered`. Dedup keys get committed as **Delivered** despite messages being silently dropped — the exact class of bug this PR aims to fix.

The same issue exists in `dispatch_emitted_messages` at line ~2779.

**Suggested fix:** Return an error instead of `Ok(())` so dedup keys stay uncommitted:

```rust
let Some(tx) = tx_guard.as_ref() else {
    tracing::error!(
        channel = %self.name,
        count = emitted_count,
        "Messages emitted but no sender available - channel may not be started!"
    );
    return Err(WasmChannelError::ChannelClosed {
        name: self.name.clone(),
    });
};
```

This is a pre-existing issue, but the reordering in this PR makes it load-bearing: previously `commit_writes` ran before delivery regardless, so the `Ok(())` didn't matter for dedup. Now that commit depends on the return value, this path silently defeats the fix.

src/channels/wasm/wrapper.rs
```diff
@@ -1653,11 +1656,20 @@ impl WasmChannel {
...
+                // If delivery fails, workspace (including dedup keys) stays
+                // uncommitted so the upstream platform can retry successfully.
+                // NOTE: This assumes each webhook/poll produces at most 1 emitted
+                // message. If multiple messages per event are needed in the future,
+                // a per-message ack or outbox mechanism must be added to avoid
+                // partial-success replay issues.
                 let emitted = host_state.take_emitted_messages();
                 self.process_emitted_messages(emitted).await?;
 
+                self.workspace_store.commit_writes(&pending_writes);
```
serrrfirat, line 1670: **Medium Severity — All workspace writes deferred, not just dedup**

The reorder defers ALL pending workspace writes until after delivery — including non-dedup state like Telegram polling offsets. For poll-based channels that write state AND emit messages, a send timeout rolls back the offset too. The next poll tick re-fetches the same messages from the upstream API.

This is correct for at-least-once delivery semantics, but creates redundant external API calls that could be avoided. The PR description says "Channels that do not write dedup keys are unaffected", but channels that write *both* dedup keys and non-dedup state are affected for their non-dedup writes too.

**Options (no change needed if this is an accepted tradeoff):**
1. Document this tradeoff with a comment here
2. Or split: commit non-dedup writes immediately inside `spawn_blocking` (old behavior), defer only dedup writes until after delivery

src/channels/wasm/host.rs
```diff
@@ -511,31 +548,79 @@ impl ChannelHostState {
...
+            for write in writes {
+                if Self::is_dedup_path(&write.path) {
+                    tracing::debug!(
+                        path = %write.path,
+                        "Registering dedup key as Pending"
+                    );
+                    // Only insert if not already Delivered — a second webhook
+                    // arriving after a successful delivery must not downgrade
+                    // the status.
+                    dedup.entry(write.path.clone()).or_insert_with(|| DedupEntry {
```
serrrfirat, line 619: **Medium Severity — Widened race window for concurrent webhook dedup**

With the old code, \`commit_writes\` ran inside \`spawn_blocking\`, so the dedup key was committed to \`ChannelWorkspaceStore\` before returning from the blocking task. A second concurrent webhook's WASM execution could see the key (if it started after the first's blocking task completed) and skip emitting.

With the new code, \`commit_writes\` runs after delivery completes. Two concurrent webhooks for the same event will both pass the WASM-side dedup check (neither has committed), both deliver, then both commit. The race window is wider.

This is a deliberate and correct tradeoff (preventing permanent message loss > occasional duplicate), but it's not documented. A future reader might assume the dedup lifecycle prevents all duplicates.

**Suggested:** Add a brief doc comment here or near \`DEDUP_PATH_INFIX\` clarifying the consistency model:

```rust
// NOTE: Dedup prevents re-delivery on *retries after success*, not on
// concurrent first-attempts for the same event.  Two simultaneous webhooks
// can both pass the WASM-side dedup check before either commits.
```

src/channels/wasm/wrapper.rs
```diff
@@ -64,6 +65,7 @@ use crate::tools::wasm::credential_injector::{
 const WEBSOCKET_EVENT_QUEUE_RELATIVE_PATH: &str = "state/gateway_event_queue";
 const WEBSOCKET_EVENT_PROCESSING_QUEUE_RELATIVE_PATH: &str = "state/gateway_event_queue_processing";
 const WEBSOCKET_EVENT_QUEUE_MAX_ITEMS: usize = 100;
+const SEND_TIMEOUT_SECS: u64 = 10;
```
ilblackdragon, line 68: HIGH: The fixed 10-second send timeout is longer than some webhook platforms will wait before failing the delivery and retrying. Slack webhook endpoints, for example, must respond within 3 seconds or Slack retries with the same `X-Slack-Retry-Num` header. If `tx.send` blocks for 4-10 seconds waiting on a full agent queue, Slack will already be retrying *before* this handler returns — and when the retry arrives the dedup store is still empty (no `commit_writes` has been called yet, because the first handler is still waiting in `tx.send`). Two WASM invocations both see an empty dedup entry, both emit, both eventually `promote_pending_to_delivered` — double delivery.

The PR description says this makes retries "handled correctly", but the interaction between `SEND_TIMEOUT_SECS` and upstream retry timeouts is load-bearing. Two options:

1. Set the timeout per-channel from `ChannelCapabilities` / capability config so Slack can use 2s and Telegram can use longer.
2. Bound the timeout by the webhook platforms's known retry deadline (document which platform drove the 10s choice).

Without that, this is just a narrower version of the original bug for fast-retrying platforms.

src/channels/wasm/wrapper.rs
```diff
@@ -1653,11 +1656,20 @@ impl WasmChannel {
...
         let channel_name = self.name.clone();
         match result {
-            Ok(Ok((response, mut host_state))) => {
-                // Process emitted messages
+            Ok(Ok((response, mut host_state, pending_writes))) => {
+                // Process emitted messages BEFORE committing workspace writes.
+                // If delivery fails, workspace (including dedup keys) stays
+                // uncommitted so the upstream platform can retry successfully.
+                // NOTE: This assumes each webhook/poll produces at most 1 emitted
+                // message. If multiple messages per event are needed in the future,
```
ilblackdragon, line 1664: HIGH: This PR reorders a side effect (`commit_writes` + `promote_pending_to_delivered`) behind a predicate (message delivery succeeded), with a wrapper (`process_emitted_messages`) between the helper and the outcome. That's exactly the shape `.claude/rules/testing.md` ("Test Through the Caller, Not Just the Helper") requires a caller-level test for.

The added tests in `host.rs` only validate `ChannelWorkspaceStore` in isolation — they prove the dedup lifecycle works if you call it right, but nothing in this PR proves `call_on_http_request` actually skips `commit_writes` when `tx.send` times out, or that `execute_poll` + polling loop skips `commit_writes` when `dispatch_emitted_messages` returns `EmitRateLimited`. A regression that re-inlines `workspace_store.commit_writes(&pending_writes)` back inside `spawn_blocking` would pass every test added here.

Please add an integration-tier test (`tests/wasm_channel_integration.rs` or similar, feature-gated if needed) that:

1. Builds a `WasmChannel` with a closed `mpsc::Sender` (or zero-capacity one) so `tx.send` returns `Err` / times out.
2. Invokes `call_on_http_request` with a request whose WASM guest would write a dedup key.
3. Asserts `workspace_store.read("channels/<name>/state/dedup/<event_id>")` returns `None` after the failed call.
4. Then re-invokes with a healthy sender and asserts the key is now Delivered.

The three bug classes in the rule doc all share the "helper works, wrapper drops the input" pattern — that's literally the risk here.

src/channels/wasm/host.rs
```diff
@@ -502,6 +502,43 @@ impl ChannelHostState {
...
+
+#[derive(Debug, Clone)]
+struct DedupEntry {
+    status: DedupStatus,
+    created_at: Instant,
+}
+
 /// In-memory workspace store for WASM channels.
 ///
 /// Persists workspace writes across callback invocations within a single
```
ilblackdragon, line 544: MEDIUM: `is_dedup_path` uses a substring match — `path.contains("/state/dedup/")` — which silently captures any workspace path a channel author happens to nest under a `state/dedup/` subdir, even if their intent is ordinary KV caching (e.g. a channel that writes `cache/state/dedup/last_seen`). Those writes are diverted into the dedup sub-store, never written to `data`, and become read-only (via the Pending/Delivered TTL machinery) from the guest's perspective.

There is no schema / WIT declaration that marks a path as dedup; the classification is purely by string convention. Consider either:

1. Anchoring the match with a `channels/<name>/state/dedup/` prefix check (e.g. starts with, after prefix, or a compiled regex), so only the documented namespace routes through the dedup store, or
2. Making dedup an explicit WIT host function (`register_dedup_key`) rather than overloading `workspace_write`.

Today, any documentation for channel authors that says "don't write to paths containing `/state/dedup/`" is an implicit coupling that will silently break a channel if violated.

src/channels/wasm/host.rs
```diff
@@ -502,6 +502,43 @@ impl ChannelHostState {
...
+    /// confirmed.  Treated as duplicate by concurrent requests to prevent
+    /// double-processing.  Expires after `DEDUP_PENDING_TTL` so a crash does
+    /// not permanently block future retries.
+    Pending,
+    /// Message was successfully delivered to the agent queue.  Treated as
+    /// duplicate for the full `DEDUP_DELIVERED_TTL` idempotency window.
+    Delivered,
+}
+
+#[derive(Debug, Clone)]
```
ilblackdragon, line 536: MEDIUM: The Pending→Delivered lifecycle narrows but does not eliminate the dedup race: concurrent webhook A and B for the same `event_id` each invoke `call_on_http_request` and their WASM guests each call `workspace_read` on the dedup path *before* either has committed. If B's WASM runs in `spawn_blocking` between A's `take_pending_writes()` and A's post-delivery `commit_writes`, B sees `None` and emits. Both handlers then call `commit_writes` + `promote_pending_to_delivered` and both messages hit the agent queue — silent double delivery.

Pre-fix, the race window was the span of a single `spawn_blocking`. Post-fix the window extends through `process_emitted_messages` (up to `SEND_TIMEOUT_SECS = 10s`), so the fix meaningfully *widens* the race. That's not a regression of this PR's stated goal (message loss from premature commit is what's being fixed), but it needs to be called out in the PR description under Behavior so operators understand the new failure mode.

A proper fix would be a synchronous "claim" of the dedup key at the host boundary (before WASM runs) under the same `dedup` lock, returning a sentinel the guest can read. That's out of scope for a fix PR, but at minimum the `Pending` docstring claim "Treated as duplicate by concurrent requests to prevent double-processing" is not actually enforced by this code.

ilblackdragon:
Architectural review notes (non-blocking, but should be documented):

**Delivery semantics are now at-least-once, and the PR doesn't spell that out.** The previous behavior was a silent message-loss bug; the new behavior is "agent queue may see the same event twice if delivery partially succeeded or the process restarted between `commit_writes` and the next invocation". That's the right trade, but the PR description only says retries are "handled correctly" — it should explicitly state that downstream consumers (the agent dispatcher) MUST be idempotent on `event_id`, and link to where that idempotency is enforced (or note it's unenforced and thus a follow-up).

**`ChannelWorkspaceStore` is in-memory only, so the 24h Delivered TTL is misleading.** A process restart wipes both the `data` and `dedup` maps. Slack webhook retries happen up to 1 hour after the original; Telegram keeps the same `update_id` for minutes; a restart during that window causes duplicate delivery. This is pre-existing (not introduced by this PR), but the `DEDUP_DELIVERED_TTL` doc comment — "24 h is a safe margin" — reads as if the window is durable, which it isn't. Either (a) persist Delivered entries to the dual-backend DB (per `src/db/CLAUDE.md`) for true durable idempotency, or (b) change the comment to acknowledge the in-memory ceiling.

**Partial multi-message delivery on retry.** The inline `NOTE` on the `call_on_http_request` match arm says: "assumes each webhook/poll produces at most 1 emitted message. If multiple messages per event are needed in the future, a per-message ack or outbox mechanism must be added." This assumption isn't *enforced* — `ChannelHostState::MAX_EMITS_PER_EXECUTION` is 100, so a WASM channel today can legally emit many messages, and if message 5 of 10 hits `SendTimeout`, messages 0..4 are delivered, retry redelivers 0..9. Consider either clamping `MAX_EMITS_PER_EXECUTION` to 1 for channels that declare a dedup-based webhook flow, or surface this clearly to channel authors in the WIT docs.

**Everything-Through-Tools consistency.** Per `CLAUDE.md`: "all actions originating from ... WASM channels ... MUST go through `ToolDispatcher::dispatch()`". The agent-queue send here [`tx.send(msg)`] is the channel-to-agent ingestion path, which predates that rule, so it's not in scope for this fix — but if the tool-dispatch migration ever extends to the channel ingestion side, the dedup state machine will need to move behind that boundary too. Worth a linked follow-up issue.

serrrfirat:
Thank you for fixing the important ordering issue between delivery and deduplication. We are closing this PR because the legacy WASM-channel delivery lifecycle has been replaced in Reborn.

Inbound idempotency is now host-owned in `crates/ironclaw_product_workflow`: the workflow reserves an event, dispatches it, and only settles the durable ledger for an acknowledgement that represents a durable outcome (`workflow.rs`, `should_settle_ack` and `IdempotencyLedger::settle`). WASM adapters consume the typed `ProductInboundAck` contract instead of persisting pending/delivered markers in guest workspace state.

Your core observation—never mark work complete before the durable side effect succeeds—still applies and is reflected in the new architecture. We would love contributions that strengthen Reborn adapter redelivery and durable-ledger integration coverage. Thank you for the careful work here.

PR #1632, opened 2026-03-25
fix(agent): discard truncated tool calls when finish_reason == Length (#1631)

### LINKED ISSUE #1631: fix(agent): discard truncated tool calls when finish_reason == Length
### Problem

When a large tool result (e.g., 145KB MCP response) is in context, the LLM may try to echo that data verbatim in a tool call parameter (e.g., `memory_write` with the full content). The `select_tools()` path uses `max_tokens=1024`, which is insufficient — the LLM's response gets truncated (`finish_reason: Length`), producing tool call blocks with empty or partial parameters.

Because `select_tools()` returns non-empty selections (the tool name is present even with empty params), `JobDelegate::call_llm()` uses them directly and **never falls through** to `respond_with_tools()` (which has a 4096 token budget). The agent then loops 10+ iterations executing malformed tool calls that all fail with "missing 'content' parameter".

**Observed behavior:**
```
Tool call started tool=memory_write params={} job=4bf0afde...
Tool call failed tool=memory_write error=Invalid parameters: missing 'content' parameter
```
Repeated across iterations 3–14, with the agent stuck in a loop.

### Root Cause

1. `select_tools()` in `reasoning.rs` does not check `finish_reason` — returns truncated tool selections as-is
2. The agentic loop in `agentic_loop.rs` executes tool calls without checking if the response was truncated
3. `ChatDelegate::before_llm_call` unconditionally overwrites `reason_ctx.force_text`, defeating any escalation

### Fix

**Layer 1 — `select_tools()` guard:** Check `finish_reason == Length` after `complete_with_tools()`. If truncated, return empty selections so the caller falls through to `respond_with_tools()` (4096 token budget).

**Layer 2 — Agentic loop guard:** When `RespondResult::ToolCalls` arrives with `finish_reason == Length`, discard the tool calls, inject a `TRUNCATED_TOOL_CALL_NOTICE` user message, and continue the loop. After 3 consecutive truncations, escalate to `force_text` mode so the LLM stops attempting tool calls entirely.

**Layer 3 — `ChatDelegate` fix:** Change `reason_ctx.force_text = force_text` to `force_text || reason_ctx.force_text` so truncation-triggered `force_text` is not clobbered by the iteration-based check.

### Files Changed

- `src/llm/reasoning.rs` — `finish_reason` guard in `select_tools()`, `finish_reason` field on `RespondOutput`, `TRUNCATED_TOOL_CALL_NOTICE` constant, 2 tests
- `src/agent/agentic_loop.rs` — Truncation guard with `force_text` escalation, 2 tests
- `src/agent/dispatcher.rs` — Preserve `force_text` in `ChatDelegate::before_llm_call`
- `src/llm/mod.rs` — Re-export `TRUNCATED_TOOL_CALL_NOTICE`
- `src/worker/job.rs` — Add `finish_reason` to 2 `RespondOutput` construction sites

rajulbhatnagar:
## Summary

- When `select_tools()` (max_tokens=1024) produces a truncated response, return empty selections so the caller falls through to `respond_with_tools()` (4096 token budget)
- In the agentic loop, discard tool calls when `finish_reason == Length` and inject a notice telling the LLM to summarize instead of echoing data verbatim
- After 3 consecutive truncations, escalate to `force_text` mode to break the loop
- Fix `ChatDelegate::before_llm_call` overwriting truncation-triggered `force_text`

Closes #1631

## Test plan

- [x] `test_select_tools_returns_empty_on_truncation` — verifies `select_tools()` discards on `Length`
- [x] `test_select_tools_returns_selections_when_not_truncated` — verifies happy path unaffected
- [x] `test_truncated_tool_calls_discarded_on_length` — verifies agentic loop discards tool calls, injects notice, preserves partial content
- [x] `test_repeated_truncations_force_text_mode` — verifies `force_text` escalation after 3 truncations
- [x] Manual: run with large MCP tool results, confirm agent no longer loops on malformed tool calls

### DIFF
--- a/src/agent/agentic_loop.rs
+++ b/src/agent/agentic_loop.rs
@@ -134,7 +134,9 @@ pub async fn run_agentic_loop(
     config: &AgenticLoopConfig,
 ) -> Result<LoopOutcome, Error> {
     let mut consecutive_tool_intent_nudges: u32 = 0;
-    let mut consecutive_truncations: u32 = 0;
+    // Accumulates across all iterations (not reset by text responses) so
+    // non-consecutive truncations still escalate to force_text.
+    let mut truncation_count: u32 = 0;
 
     for iteration in 1..=config.max_iterations {
         // Check for external signals (stop, cancellation, user messages)
@@ -220,12 +222,12 @@ pub async fn run_agentic_loop(
                 // incomplete. Discard them and tell the LLM to try a different
                 // approach rather than executing malformed tool calls.
                 if output.finish_reason == FinishReason::Length {
-                    consecutive_truncations += 1;
+                    truncation_count += 1;
                     let names: Vec<&str> = tool_calls.iter().map(|tc| tc.name.as_str()).collect();
                     tracing::warn!(
                         iteration,
                         tools = ?names,
-                        consecutive_truncations,
+                        truncation_count,
                         "Discarding truncated tool calls (finish_reason=Length)"
                     );
                     if let Some(ref text) = content {
@@ -236,7 +238,7 @@ pub async fn run_agentic_loop(
                         .push(ChatMessage::user(crate::llm::TRUNCATED_TOOL_CALL_NOTICE));
                     // After repeated truncations, force text-only mode so the LLM
                     // stops attempting tool calls it can't fit in the output budget.
-                    if consecutive_truncations >= 3 {
+                    if truncation_count >= 3 {
                         reason_ctx.force_text = true;
                     }
                     delegate.after_iteration(iteration).await;
@@ -244,7 +246,7 @@ pub async fn run_agentic_loop(
                 }
 
                 consecutive_tool_intent_nudges = 0;
-                consecutive_truncations = 0;
+                truncation_count = 0;
 
                 if let Some(outcome) = delegate
                     .execute_tool_calls(tool_calls, content, reason_ctx)
--- a/src/llm/reasoning.rs
+++ b/src/llm/reasoning.rs
@@ -201,6 +201,8 @@ pub struct ReasoningContext {
     pub metadata: std::collections::HashMap<String, String>,
     /// When true, force a text-only response (ignore available tools).
     /// Used by the agentic loop to guarantee termination near the iteration limit.
+    /// Sticky: once set, never cleared within a loop invocation. Callers must
+    /// create a fresh `ReasoningContext` per `run_agentic_loop()` call.
     pub force_text: bool,
     /// Pre-built system prompt. When set, `respond_with_tools` uses this directly
     /// instead of calling `build_system_prompt_with_tools`. Allows callers to build


src/agent/agentic_loop.rs
```diff
@@ -215,7 +216,35 @@ pub async fn run_agentic_loop(
...
+                    // stops attempting tool calls it can't fit in the output budget.
+                    if consecutive_truncations >= 3 {
+                        reason_ctx.force_text = true;
+                    }
+                    delegate.after_iteration(iteration).await;
+                    continue;
+                }
+
                 consecutive_tool_intent_nudges = 0;
+                consecutive_truncations = 0;
```
serrrfirat: Medium: `consecutive_truncations` is only reset in the non-truncated `ToolCalls` path. If the loop sees `truncated tool call -> ordinary text that continues -> truncated tool call -> ordinary text that continues -> truncated tool call`, this counter still reaches 3 and flips `force_text`, even though the truncations were not consecutive. That changes the intended behavior from "three consecutive truncations" to "three truncations total across intervening text turns" for job/container flows that continue on intermediate text. Please reset this counter on non-truncated `RespondResult::Text` responses as well, and add a regression test for an alternating truncation/text sequence.

zmanian (dismissed):
## Review: fix(agent): discard truncated tool calls when finish_reason == Length

Sound fix applied at the right layer (agentic loop, works across all delegates). The approach -- discard tool calls on truncation, inject a notice, escalate to force_text after 3 -- is correct.

### Low

1. **`consecutive_truncations` counter doesn't reset on text responses**: Only resets in the ToolCalls branch. A successful text response between truncations doesn't reset the counter, so `force_text` could trigger based on non-consecutive truncations. The behavior is safe (conservative, not permissive), but the variable name is misleading. Consider renaming to `total_truncations` or resetting in the Text branch.

2. **`force_text` is sticky**: Once set, never cleared. Fine today since `ReasoningContext` is created fresh per `dispatch()`, but add a comment noting this assumption.

### Positive

- Correctly preserves partial assistant content before the truncation notice
- Guards `select_tools()` too -- catches the separate reasoning code path
- Provider-safe: `FinishReason::Length` not reported = fix doesn't activate (safe default)
- Can't cause loops: truncation notice + force_text escalation + max_iterations backstop
- Four solid tests covering single truncation, escalation, and select_tools paths

Approve. Ship it.

src/agent/agentic_loop.rs
```diff
@@ -215,7 +216,35 @@ pub async fn run_agentic_loop(
...
+                    );
+                    if let Some(ref text) = content {
+                        reason_ctx.messages.push(ChatMessage::assistant(text));
+                    }
+                    reason_ctx
+                        .messages
+                        .push(ChatMessage::user(crate::llm::TRUNCATED_TOOL_CALL_NOTICE));
+                    // After repeated truncations, force text-only mode so the LLM
+                    // stops attempting tool calls it can't fit in the output budget.
+                    if consecutive_truncations >= 3 {
```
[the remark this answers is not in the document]
rajulbhatnagar: Unlike `max_tool_intent_nudges` (which tunes normal-path LLM behavior across different models), this threshold is on an error-recovery path — it only activates when the provider truncates a response mid-tool-call. Keeping it as a hardcoded constant with the surrounding comment explaining the intent is sufficient. Happy to revisit as a follow-up if a real tuning need arises.

src/agent/agentic_loop.rs
rajulbhatnagar: Good catch on the naming. The accumulating behavior is actually intentional and safer — the interleaving scenario (truncation → explanatory text → truncation retry) is the *most natural* LLM response to the truncation notice, and represents a model that's persistently unable to fit tool calls in the output budget. Resetting on text would let this oscillation continue to max_iterations.

Renamed `consecutive_truncations` → `truncation_count` with a comment explaining the intentional accumulation semantics. This avoids the naming mismatch without weakening the safety net.

rajulbhatnagar:
@zmanian Thanks for the approval and the sharp observations.

Point 1: Agreed the name was misleading. Renamed to `truncation_count` with a comment clarifying the accumulation semantics — keeping the current behavior since it's the safer default.

Point 2: Added a doc comment on `ReasoningContext::force_text` noting the stickiness invariant and the requirement for a fresh context per `run_agentic_loop()` call.

Both addressed in the follow-up commit.

PR #1635, opened 2026-03-25
fix: sanitize tool error results before llm injection

serrrfirat:
## Summary
- route failed tool results through the same sanitize-and-wrap safety path as successful tool outputs
- switch chat dispatch to use the shared `process_tool_result()` helper instead of formatting raw tool errors into LLM context
- add regressions covering wrapped error output, boundary-injection neutralization, and dispatcher-level tool-name preservation

## Verification
- `cargo test test_process_tool_result_error --lib`
- `cargo test test_tool_error_format_includes_tool_name --lib`
- `cargo check --lib`

### DIFF
--- a/src/agent/dispatcher.rs
+++ b/src/agent/dispatcher.rs
@@ -2433,12 +2433,8 @@ mod tests {
             injection_check_enabled: true,
         });
         let result: Result<String, _> = Err(err);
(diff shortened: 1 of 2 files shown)


serrrfirat:
Closing this because it was based on feat/multi-tenant-isolation-phases-2-4 instead of staging. Replacement PR: #1639

PR #1662, opened 2026-03-26
feat(agent): anti-drift self-checks in agentic loop (#1634)

### LINKED ISSUE #1634: feat(agent): anti-drift self-checks in agentic loop
## Summary

Add a lightweight, rule-based `DriftMonitor` to the agentic loop that detects common failure patterns and injects corrective system messages — without requiring additional LLM calls.

**Motivation:** The agentic loop currently detects stuck jobs (time-based) and budget exhaustion (cost guard), but has no mechanism to detect whether the agent is *making meaningful progress toward the goal*. Common failure modes like tool retry loops, write-read-write cycles, and silent runaway execution go undetected until the iteration limit is hit — wasting tokens and time.

**Prior art:** Inspired by RuFlo's Continue Gate (checkpoint every N steps, rework ratio tracking, budget slope analysis) but adapted for IronClaw's single-agent architecture and existing delegate hook system.

## Problem

Five drift patterns currently go undetected:

| Pattern | Example | Impact |
|---------|---------|--------|
| **Repetition loops** | Same tool + same params called 3+ times | Wasted tokens on identical failing calls |
| **Failure spirals** | 4+ consecutive tool failures | Agent inventing fixes based on wrong assumptions |
| **Tool cycling** | write→read→write→read alternation | Agent second-guessing itself without progress |
| **Silence drift** | 15+ iterations without text response | User left waiting with no status |
| **Rework churn** | Same file written 3+ times | Overwriting own work repeatedly |

### `DriftMonitor` struct

```rust
pub struct DriftMonitor {
    original_goal: String,
    tool_history: VecDeque<ToolCallRecord>,
    metrics: IterationMetrics,
    config: DriftConfig,
}

struct ToolCallRecord {
    name: String,
    params_hash: u64,
    succeeded: bool,
    iteration: u32,
}

struct IterationMetrics {
    total_tool_calls: u32,
    consecutive_failures: u32,
    unique_tools_used: HashSet<String>,
    rework_count: u32,
    last_text_iteration: u32,
}
```

### Five detection rules (no LLM calls)

1. **Repetition** — Same tool + same param hash ≥3 times in last 10 calls → inject correction
2. **Failure spiral** — ≥4 consecutive tool failures → inject "reassess approach"
3. **Tool cycling** — A-B-A-B-A-B pattern in last 6 calls → inject "state what's blocking you"
4. **Silence drift** — ≥15 iterations since last text response → inject "provide progress update"
5. **Rework detection** — Same file written ≥3 times → inject "review before writing again"

### Integration points

Corrections are injected as **system messages** into the LLM context (same pattern as the existing `TOOL_INTENT_NUDGE` and iteration-limit nudge). The LLM self-corrects on its next response.

**Files to change:**

| File | Change |
|------|--------|
| `src/agent/drift_monitor.rs` | **New** — `DriftMonitor` struct + detection rules |
| `src/agent/dispatcher.rs` | Add `drift_monitor` field to `ChatDelegate`; call `check()` in `before_llm_call()`, call `record_*()` in `after_iteration()` |
| `src/worker/job.rs` | Same integration for `JobDelegate` |
| `src/agent/mod.rs` | Module declaration |

**No changes to `agentic_loop.rs`** — delegates already own the hooks.

### Configuration

```rust
pub struct DriftConfig {
    pub enabled: bool,                     // default: true
    pub repetition_window: usize,          // default: 10
    pub repetition_threshold: usize,       // default: 3
    pub max_consecutive_failures: u32,     // default: 4
    pub cycle_detection_window: usize,     // default: 6
    pub silence_threshold: u32,            // default: 15 iterations
    pub max_file_rewrites: u32,            // default: 3
    pub checkpoint_interval: u32,          // default: 25 (forced progress summary)
}
```

Configurable via env vars (`IRONCLAW_DRIFT_*`) or disabled entirely.

## Future extensions (not in scope)

- **Semantic drift detection** — LLM-based check comparing current context to original goal (runs every N iterations, higher cost)
- **Plan adherence** — Compare actual tool sequence to planned sequence (Worker already has planning phase)
- **Budget slope analysis** — Linear regression over token consumption windows (from RuFlo's Continue Gate)
- **Escalation levels** — First nudge is soft (system message), second is stronger ("wrap up now"), third triggers early loop exit

## Test plan

- [ ] Unit tests for each detection rule with synthetic `ToolCallRecord` sequences
- [ ] Integration test: agent loop with a mock LLM that produces repetitive tool calls → verify correction message injected
- [ ] Verify drift monitor is a no-op when disabled via config
- [ ] Verify no regression in existing agentic loop behavior (tool-intent nudge, iteration limits, approval flow)

reidliu41:
## Summary

  Add rule-based DriftMonitor that detects 4 token-wasting patterns
  (failure spiral, repetition loop, tool cycling, silence drift) and
  injects corrective system messages — zero extra LLM cost.

  - Integrate into all 3 LoopDelegate paths (chat, job, container)
  - Config via Settings chain + IRONCLAW_DRIFT_* env vars
  - Orchestrator injects config into container env
  - Preflight rejections excluded from drift history
  - Communication gated on sanitized non-whitespace text
  - Recovery-based cooldown prevents correction flooding
  - 26 tests covering rules, config propagation, and edge cases

## Change Type

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] CI/Infrastructure
- [ ] Security
- [ ] Dependencies

## Linked Issue

Closes #1634

## Validation

- [x] `cargo fmt`
- [ ] Relevant tests pass: 
- [ ] Manual testing: 

## Rollback Plan

---

**Review track**:


zmanian (dismissed):
## Review: feat(agent): anti-drift self-checks in agentic loop

Well-designed feature. Zero extra LLM calls -- all detection is rule-based over a small in-memory deque. Configurable, disableable, and integrated across all three delegates (Chat, Job, Container).

Four detection patterns:
- Repetition loops (same tool+args N times)
- Failure spirals (consecutive errors)
- Tool cycling (A-B-A-B oscillation)
- Silence drift (consecutive empty/whitespace responses)

### Low

1. **`text.is_empty()` -> `text.trim().is_empty()` in JobDelegate**: Subtle behavioral change beyond drift monitoring -- whitespace-only text now returns `TextAction::Continue` instead of falling through to completion detection. Add a comment noting this is intentional.

2. **`DefaultHasher` stability**: Only used within a single process lifetime so this is fine, but the doc comment on `hash_arguments` could clarify "within-process consistency" vs cross-version persistence.

3. **Drift history lost on approval-resume**: Documented and tested. Acceptable since approval interruptions are infrequent.

### Positive

- 18 unit tests covering all detectors, priority, cooldown/recovery, edge cases, and cross-delegate parity
- Suppression/cooldown prevents correction spam
- Master kill switch via `IRONCLAW_DRIFT_ENABLED=false`
- Config propagates to containers
- Negligible performance overhead

Approve. Ship it.

src/agent/dispatcher.rs
```diff
@@ -452,6 +484,12 @@ impl<'a> LoopDelegate for ChatDelegate<'a> {
         content: Option<String>,
         reason_ctx: &mut ReasoningContext,
     ) -> Result<Option<LoopOutcome>, Error> {
+        // G4: check content before it's moved into assistant_with_tool_calls
```
serrrfirat: **Medium Severity**

`has_nonempty_content` is computed from raw `content.trim()` here, and the same pattern is used in the job/container delegates. That means provider-leaked tool-call marker text, or content that becomes empty after sanitization, still resets `SilenceDrift` even though nothing user-visible was actually communicated. This under-detects silence in exactly the cases the PR summary describes as sanitized non-whitespace text.

Please gate `record_communication()` on the sanitized text that is actually surfaced, and add parity tests for content that is stripped to empty after tool-marker removal or safety sanitization.

src/agent/drift_monitor.rs
```diff
@@ -0,0 +1,798 @@
...
+    /// Check for drift patterns and mark the detected kind as suppressed.
+    ///
+    /// Returns `None` if disabled, no pattern detected, or the detected
+    /// pattern is currently suppressed (waiting for a recovery event).
+    pub fn check_and_mark(&mut self) -> Option<DriftCorrection> {
+        if !self.config.enabled {
+            return None;
+        }
+
+        // Priority order: failure spiral > repetition > cycling > silence
```
serrrfirat, line 358: **Medium Severity**

The cooldown is keyed only by `DriftCorrectionKind`, so unchanged history can emit a different correction on the next loop without any recovery. Four identical failed calls will fire `FailureSpiral` first, but on the next iteration the same stale history can immediately fire `Repetition` because only the failure kind was suppressed. That defeats the recovery-based cooldown and can stack multiple system nudges from one incident.

Please keep one active correction until recovery, or suppress lower-priority detectors while a higher-priority condition still matches, and add a regression test for this overlap case.

src/agent/drift_monitor.rs
```diff
@@ -0,0 +1,798 @@
...
+    history: VecDeque<ToolCallRecord>,
+    max_history: usize,
+    iterations_since_communication: usize,
+    suppressed: HashSet<DriftCorrectionKind>,
+}
+
+impl DriftMonitor {
+    /// Create a new monitor with the given configuration.
+    pub fn new(config: DriftConfig) -> Self {
+        let max_history = config
```
serrrfirat, line 239: **Medium Severity**

`detect_cycling()` builds a cross-iteration sequence from raw tool-call history, but `max_history` only retains `max(repetition_window, cycling_window, failure_spiral_threshold)` records. In a multi-tool iteration, that ring buffer can hold fewer than `cycling_window` iterations, so the A-B-A-B detector silently stops firing. With the default settings, two tool calls per iteration already shrink the retained window below six iterations.

Please either track cycling on a per-iteration history or size the retained history to cover multi-tool iterations, and add a regression test with multiple tool calls per iteration.

zmanian (changes requested):
## Request Changes

Well-designed zero-cost drift detection. Pure rule-based, zero extra LLM calls, recovery-gated cooldowns. Good design.

### Must-fix

1. **Behavioral change in JobDelegate** -- `text.is_empty()` -> `text.trim().is_empty()` changes control flow beyond drift recording. Previously whitespace-only text fell through to `llm_signals_completion`; now it returns `Continue`. Should be a separate change.

2. **Duplicate defaults** -- `DriftConfig::default()` and `DriftSettings::default()` independently define (3, 10, 4, 6, 15). Use shared constants.

3. **No config threshold validation** -- `repetition_threshold=0` or `cycling_window=1` causes edge cases. `max_history` could be 0. Add minimum clamping.

4. **PR checklist unchecked** -- verify CI passes.

### Should-fix

5. ~90 lines duplicated across 3 delegates -- consider a helper method on DriftMonitor
6. Module should be `pub(crate)` not `pub`
7. `current_iteration` AtomicUsize unnecessary given existing Mutex -- store on monitor directly

### Nice-to-have

8. Update `src/agent/CLAUDE.md` to list `drift_monitor.rs`

zmanian (commented):
## Re-review: feat(agent): anti-drift self-checks in agentic loop

Checking whether the previous review feedback (from both zmanian and serrrfirat) has been addressed in the follow-up commits.

### Previously Raised Issues -- Status

1. **Behavioral change in JobDelegate** (zmanian #1) -- **Addressed.** The `text.is_empty()` guard at the top of `handle_text_response` is preserved unchanged. The `text.trim().is_empty()` check is only used for the new drift `record_communication()` gate. Completion logic is untouched.

2. **Duplicate defaults** (zmanian #2) -- **Addressed.** Shared constants (`DEFAULT_REPETITION_THRESHOLD`, etc.) in `drift_monitor.rs` are used by both `DriftConfig::default()` and `DriftSettings::default()`.

3. **No config threshold validation** (zmanian #3) -- **Addressed.** `DriftConfig::clamped()` method clamps all thresholds to safe minimums. Called in both `AgentConfig::resolve()` and `resolve_drift_config_inner()`.

4. **CI checklist unchecked** (zmanian #4) -- Still unchecked in the PR template. This is from a fork so full CI likely hasn't run.

5. **~90 lines duplicated across delegates** (zmanian #5) -- **Partially addressed.** `visible_sanitized_content` is shared. The before_llm_call drift-check block (~15 lines) is still duplicated across all 3 delegates, but this is minor -- the blocks are small and the delegates have different field access patterns that make a shared helper awkward.

6. **Module should be `pub(crate)` not `pub`** (zmanian #6) -- **NOT addressed.** `src/agent/mod.rs` still has `pub mod drift_monitor;`. Should be `pub(crate) mod drift_monitor;`. No downstream consumer needs this module.

7. **`current_iteration` AtomicUsize unnecessary** (zmanian #7) -- **Addressed.** Now a plain `usize` stored on the monitor behind the existing Mutex.

8. **Suppressed high-priority blocks lower-priority detectors** (serrrfirat) -- **Addressed.** `check_and_mark()` returns `None` when a high-priority match is suppressed, blocking all lower-priority detectors. Regression test `test_suppressed_high_priority_blocks_lower` covers this.

9. **Per-iteration cycling detection** (serrrfirat) -- **Addressed.** Separate `iteration_tools` VecDeque with `last_cycling_iteration` guard. Regression test `test_cycling_detected_with_multi_tool_iterations` covers multi-tool iterations.

10. **Gate communication on sanitized text** (serrrfirat) -- **Addressed.** `visible_sanitized_content` helper used by all 3 delegates. Tests for empty/whitespace/sanitized-to-empty cases included.

### Remaining Issue

**`pub mod drift_monitor` -> `pub(crate) mod drift_monitor`** in `src/agent/mod.rs`. This is an internal module with no downstream consumers. Using `pub` exposes it as part of the library's public API unnecessarily.

### New Observations

**Correction messages include tool names verbatim** (`src/agent/drift_monitor.rs`, `message()` method). Tool names in the correction system messages come from LLM-provided tool call names, which get format-interpolated into strings injected as system messages. This is low risk since tool names are validated against the tool registry before execution, but worth noting. If tool name validation were ever relaxed, this would become an injection vector.

**Test coverage is solid.** 26 tests covering all detectors, priority ordering, cooldown/recovery lifecycle, edge cases (multi-tool iterations, suppression blocking, approval-resume), cross-delegate parity, and config propagation. The `test_cycling_suppression_clears_on_broken_pattern` test is particularly good -- it covers the full fire/suppress/break/re-fire lifecycle.

**Performance is fine.** All detection is O(window_size) over small in-memory deques with no allocations on the hot path. One Mutex lock per iteration. No LLM calls.

### Verdict

The `pub(crate)` visibility fix is trivial but should be done before merge. Everything else from the previous rounds has been addressed well.

zmanian (dismissed):
The visibility fix is in place: `pub(crate) mod drift_monitor` in `src/agent/mod.rs`. No new issues in the two follow-up commits. LGTM.

src/config/agent.rs
```diff
@@ -75,6 +78,7 @@ impl AgentConfig {
             max_llm_concurrent_per_user: None,
```
serrrfirat, line 78: **Critical Severity — Compilation Failure**

`AgentConfig::for_testing()` (gated behind `#[cfg(feature = "libsql")]`) does not include the new `drift: DriftConfig` field. This will fail to compile under `cargo check --features libsql`.

**Suggested fix:** Add `drift: crate::agent::drift_monitor::DriftConfig::default(),` to `for_testing()`.

serrrfirat:
**Medium Severity — 3 findings for drift monitor**

**1. `repetition_threshold` clamped to 1 is semantically broken** (~drift_monitor.rs:273)
With threshold=1, `detect_repetition` fires whenever *any* tool appears at least once in the window — meaning it triggers on every tool call. "1 call = repetition" is nonsensical.
**Fix:** Clamp `repetition_threshold` to `max(2)` instead of `max(1)`.

**2. `tracing::info!` used for "Drift detected" violates CLAUDE.md** (~dispatcher.rs:316, container.rs:387, job.rs:1366)
Per project rules: "`info!` and `warn!` output appears in the REPL and corrupts the terminal UI. Use `debug!` for internal diagnostics. Background tasks must NEVER use `info!`". Drift detection fires on internal diagnostics.
**Fix:** Change all three `tracing::info!(... "Drift detected")` calls to `tracing::debug!`.

**3. Preflight-rejected iterations are invisible to drift monitor** (~dispatcher.rs:899-1063)
If ALL tool calls in an iteration are preflight-rejected (e.g., all require approval in a relay channel), `drift_records` is empty and the silence counter is not incremented. Even if the assistant produced visible content text, `record_communication()` is never called. The iteration becomes invisible to the drift monitor, potentially delaying silence detection.
**Fix:** Always increment the silence counter in `execute_tool_calls` regardless of preflight status, or document as intentional.

src/config/agent.rs
reidliu41: Re-checked on current HEAD (`5221302c`): `AgentConfig::for_testing()` already includes `drift: DriftConfig::default()`, and `cargo
  check --features libsql` passes locally. I don't see this compilation failure on the current revision, so this seems stale.

zmanian (changes requested):
## Review

Well-designed DriftMonitor detecting four token-wasting patterns -- failure spirals, repetition loops, tool cycling (A-B-A-B), and silence drift -- with corrective system message injection. All rule-based, zero LLM cost. 26 unit tests. Integrated into all three LoopDelegate paths.

### Strengths

- Recovery-based cooldown prevents correction flooding
- Priority ordering prevents multiple corrections stacking
- Preflight rejections correctly excluded from drift history
- visible_sanitized_content() shared across all three delegates
- Config clamping prevents foot-gun thresholds
- Container env propagation ensures consistent thresholds

### Must Fix

1. **hash_arguments uses unwrap_or_default() in production** (drift_monitor.rs ~line 643) -- while serde_json::to_string on a Value is practically infallible, the failure mode is insidious: an empty string would make ALL tool calls hash-equal, triggering false repetition detections. Should use map_err or add a comment explaining why it's safe.

2. **CI validation** -- clippy and tests not confirmed passing in the PR checklist. This touches the hot path of the agentic loop.

### Suggestions

3. **#[must_use] on DriftCorrection** -- forgetting to inject the correction silently loses drift detection.

4. **Duplicate config resolution** -- resolve_drift_config_inner() in worker/mod.rs duplicates env-var parsing from config/agent.rs. Extract a shared DriftConfig::from_env().

5. **DefaultHasher stability** -- not guaranteed stable across Rust versions. Fine for single-process but worth a doc comment.

6. **Silence threshold of 15** may be too generous for job/container delegates making many tool calls per iteration.

serrrfirat (commented):
## Paranoid Architect Review — Verdict: 👍 APPROVE WITH MINOR FIXES

⚠️ **This PR has merge conflicts and needs a rebase onto staging.**

Well-designed, thoughtfully implemented feature. Zero-cost rule-based detection with bounded memory, recovery-gated cooldowns, and 26 tests. Follows IronClaw conventions cleanly. A few items to address.

### Findings

| 1 | Medium | Correctness | drift_monitor.rs:640 | `DefaultHasher` not stable across Rust versions. Comment "covered by unit test" only covers serde key order, not hasher stability. Tolerable for in-process use. | Add comment clarifying in-process-only stability requirement |
| 2 | Medium | Correctness | drift_monitor.rs:443 | **let-chains syntax** requires Rust edition 2024. Verify project MSRV. | Verify edition or refactor to nested `if let` |
| 3 | Medium | Edge Case | drift_monitor.rs:454 | **Asymmetry**: empty-calls iterations increment silence but don't update `last_cycling_iteration`. Could miss alternating rejected/executed patterns. | Document asymmetry or update cycling tracking for empty iterations |
| 4 | Medium | Security | drift_monitor.rs:316 | **Unsanitized tool names in correction messages.** MCP server adversarial tool names embedded via `format!` into system prompt. Prompt injection vector. | Truncate to max length and strip control characters |
| 5 | Low | Architecture | dispatcher.rs:41 | `DriftMonitor` resets per `ChatDelegate` — cross-message patterns undetected. Intentional per docs. | Document in architecture docs |
| 6 | Low | Convention | worker/mod.rs:1945 | `warn!()` in container worker path. Per CLAUDE.md corrupts TUI, but runs inside Docker. | Acceptable, or downgrade to `debug!` |
| 7 | Low | Test Coverage | drift_monitor.rs | No test for multi-tool single-iteration repetition behavior | Add test for multi-call single iteration |
| 8 | Low | Robustness | drift_monitor.rs:276 | `failure_spiral_threshold` clamped to min 1. Single-failure trigger is noisy. | Clamp to min 2 |
| 9 | Low | DRY | config/agent.rs + worker/mod.rs | Drift config env var parsing duplicated host/container side | Extract shared `DriftConfig::from_env(defaults)` |
| 10 | Nit | Style | multiple files | Fully-qualified `crate::agent::drift_monitor::DriftMonitor` instead of imports | Add `use` statements |
| 11 | Nit | Style | drift_monitor.rs:556 | `std::collections::HashMap` used inline | Import at top |

### Summary
Strong implementation overall. Key items: verify let-chains edition compatibility (#2), sanitize tool names in correction messages (#4), and document the silence/cycling asymmetry (#3). The rest are minor improvements.

serrrfirat:

The drift monitor is integrated into all three retired agent/job/container loops. Reborn has one canonical agent-loop/runner path and already owns repeated-loop protection, planning limits, cancellation, and milestone reporting there.

Additional anti-drift behavior should be implemented once in `crates/ironclaw_agent_loop` or its strategy layer, using typed loop state and tests for repetition/tool cycling without injecting untrusted ad-hoc system messages or creating another loop.

serrrfirat:

The drift monitor is integrated into three retired loops. Reborn has one canonical agent-loop/runner path with loop protection, limits, cancellation, and milestones.

Implement additional drift behavior once in `ironclaw_agent_loop` using typed state and tests for repetition/tool cycling, without ad-hoc untrusted system-message injection.

Issue #1584, opened 2026-03-23
WeChat channel for IronClaw


think-in-universe:
WeChat plugin for OpenClaw is now available: https://www.npmjs.com/package/@tencent-weixin/openclaw-weixin

We'll soon need the plugin for IronClaw.

JZKK720:
can we try we can bring this integration on-board? 

https://github.com/ImGoodBai/WeClawBot-ex

PR #1666, opened 2026-03-26
feat: wechat channel

hanakannzashi:
## Summary

- Added a first-party `channels-src/wechat` WASM channel for IronClaw with single-account WeChat DM support: QR login, long-poll `getupdates`, outbound `sendmessage`, `context_token` persistence, and typing indicators.
- Added host-side interactive WeChat QR login flow in the extension manager and web gateway so setup can complete from the browser and persist `wechat_bot_token` automatically.
- Registered and wired the WeChat channel across the registry/setup flow, and renamed internal `weixin` identifiers to `wechat` for consistency with the product-facing channel name.
- Fixed gateway UI state and setup UX so non-pairing channels like WeChat show as `Active` after QR login, stepper labels localize correctly, and the QR flow uses a separate-tab CTA instead of a broken inline image.

## Change Type

- [x] Bug fix
- [x] New feature
- [ ] Refactor
- [x] Documentation
- [ ] CI/Infrastructure
- [ ] Security
- [x] Dependencies

## Linked Issue

#1584 

## Validation

- [x] `cargo fmt`
- [x] `cargo clippy --all --benches --tests --examples --all-features`
- [x] Relevant tests pass: `cargo test test_extensions_setup_returns_interactive_login_for_wechat --lib --quiet`, `cargo test test_extensions_wechat_login_poll_broadcasts_auth_completed_and_activates --lib --quiet`, `cargo test test_wechat_interactive_login_poll_persists_state_and_activates --lib --quiet`, `cargo test test_inject_channel_settings_uses_owner_scope --lib --quiet`, `cargo test test_requires_binding_detects_dm_owner_fields --lib --quiet`, `cargo test test_wechat_active_channel_does_not_require_pairing_status --lib --quiet`, `cargo test --lib wechat --quiet`, `cargo check -q`, `cargo check --manifest-path channels-src/wechat/Cargo.toml -q`, `cargo check --manifest-path channels-src/wechat/Cargo.toml --target wasm32-wasip2 -q`, `cargo clippy --all-targets -- -D warnings`, `cargo clippy --manifest-path channels-src/wechat/Cargo.toml --target wasm32-wasip2 -- -D warnings`
- [x] Manual testing: verified real WeChat QR login via the web gateway, confirmed DM text send/receive works, typing indicators appear, and the extension card/setup modal now reflect the connected state correctly after login

## Security Impact

Touches authenticated extension setup/login flows, secret persistence, and outbound HTTP from the WeChat WASM channel to `ilinkai.weixin.qq.com` and the WeChat CDN. No auth model, sandbox boundary, or secret exposure guarantees were weakened; the channel continues to rely on existing secret injection and HTTP allowlist enforcement.

## Database Impact

None. No schema changes or migrations. This reuses existing settings/secrets persistence paths and remains backend-agnostic for both PostgreSQL and libSQL.

## Blast Radius

Touches the WASM channel host path, extension manager setup/login flow, web gateway extension UI, and the new WeChat channel implementation. Regressions would most likely affect extension setup state classification, interactive login UX, or WeChat channel activation/polling.

## Rollback Plan

Revert the WeChat-specific extension manager/web gateway changes and remove the `channels-src/wechat` registration/artifacts. If a partial rollback is needed, disable the `wechat` registry entry and remove the installed `wechat.wasm` / `wechat.capabilities.json` artifacts so existing channels continue unaffected while the host/UI changes are reverted.

---

**Review track**: C


zmanian (commented):
## Security-Focused Review: WeChat Channel (iLink Bot)

Large PR from new contributor. Extra scrutiny applied. Important note: this is the **iLink Bot** protocol (JSON over HTTPS, long-polling), NOT the WeChat Official Account platform (XML, webhooks). No XML parsing, no webhook signature verification needed.

### Critical

1. **`random_wechat_uin()` is predictable** (`channels-src/wechat/src/api.rs`): Uses `base64(timestamp_ms % u32::MAX)` -- completely guessable. If `X-WECHAT-UIN` has any security purpose in the iLink protocol, this is trivially spoofable. Use actual randomness or document as non-security-relevant.

### High

2. **`baseurl` from QR response stored without validation** (`src/extensions/wechat_login.rs`): The iLink server's response can set `baseurl` to any string, which gets persisted and used for all subsequent API calls (including those carrying the bot_token). An attacker who controls the QR response (MITM, rogue QR) could redirect all authenticated traffic to a malicious server. **Validate that baseurl is HTTPS and matches known WeChat domains** (e.g., `*.weixin.qq.com`, `*.wechat.com`).

3. **Bot token received without provenance verification**: QR login receives `bot_token` from iLink server with no certificate pinning. Acceptable for most channels but WeChat operates in environments where MITM is a real concern. Document the trust assumption.

### Medium

4. **Client-side poll loop with 0ms delay** (`src/channels/web/static/app.js`): `setTimeout(fn, 0)` on `pending`/`scanned`/`refreshed` status creates a tight loop hammering the server. Use 2-5 second backoff.

5. **No message deduplication**: Messages from `getupdates` are emitted without checking if already processed. Track `message_id` in persistent state to prevent replay.

6. **CSP relaxed globally** (`src/channels/web/server.rs`): `img-src` adds `liteapp.weixin.qq.com` for all pages, not just WeChat setup. Acceptable for MVP but should be documented.

### Architecture -- Good

- Correctly follows WASM channel pattern (wit_bindgen, Guest trait, all callbacks)
- Module-owned initialization per CLAUDE.md
- `requires_binding` addition for QR-login-based channels is clean
- No new dependencies beyond standard ones (wit-bindgen, serde, base64)
- No `.unwrap()`/`.expect()` in production code

### Minor

7. **`inject_channel_secrets_into_config` scope change** (`src/channels/wasm/setup.rs`): Changed from hardcoded `"default"` to `owner_id`. Verify Feishu secrets still load correctly -- they may have been stored under the old `"default"` scope.

### Verdict

Request changes. The `baseurl` validation (item 2) is the most important -- it's an open redirect for all authenticated API traffic. The client-side polling backoff (item 4) is also needed to prevent self-DoS. The rest are lower priority but should be addressed.

Good first contribution overall -- architecture is sound, code quality is high, and the iLink Bot protocol choice is well-suited to the WASM channel model.

zmanian (commented):
## Re-review: WeChat channel (3 new commits)

One fix, three findings still open:

| # | Finding | Status |
|---|---------|--------|
| 1 | `random_wechat_uin` predictable | **Not addressed** -- still timestamp-based |
| 2 | `base_url` stored without validation | **Not addressed** -- no URL validation |
| 3 | Bot token without provenance check | **Not addressed** |
| 4 | Client poll 0ms delay | **Fixed** -- `poll_interval_ms` defaults 30s with `.max(30_000)` enforcement |
| 5 | No message dedup | **Partially** -- cursor persistence helps, no client-side ID dedup |

New image messaging code interpolates user-influenced params (`encrypted_query_param`, `filekey`) directly into CDN URLs without sanitization -- minor additional concern.

Items 1-2 remain blocking. The `base_url` validation is the highest priority -- it's an open redirect for all authenticated API traffic.

zmanian (commented):
## Re-review: WeChat channel (7 new commits)

Both blocking items remain unaddressed:

1. **random_wechat_uin still timestamp-based** -- needs CSPRNG or a comment explaining it's non-security-relevant
2. **base_url stored without validation** -- needs url::Url parse + https scheme enforcement

These have been flagged across 3 review rounds now. Please address before next push.

zmanian (changes requested):
## Request Changes

Well-written WASM channel implementation. Follows existing patterns, good error handling, solid test coverage. However, a few issues need attention.

### Must-fix

1. **`#[serde(default)]` removal is a breaking change** -- removing from `Option<T>` fields on `ExtensionSource`, `RegistryEntry` will cause deserialization to fail when optional fields are absent in existing registry JSON. This affects ALL extensions, not just WeChat.

2. **`silk-rs` dependency weight** -- adds `bindgen`, `clang-sys`, ~10 transitive deps for SILK decoding only used by WeChat. Consider feature-gating behind `wechat` flag. Also check if separate `aes = "0.8"` is needed given `aes-gcm` already depends on it.

3. **Document AES-ECB rationale** -- ECB in media.rs is dictated by WeChat's CDN protocol. Add comment to prevent future "fix" attempts.

### Should-fix

4. No token refresh -- channel fails silently on expiry. Emit `AuthRequired` on errcode=-14.
5. `random_wechat_uin` is deterministic from timestamp -- verify against upstream expectations.
6. Log level change affects ALL WASM channels (Trace/Debug/Info no longer collapsed to Debug) -- could cause spam.
7. Workspace write commit gating affects all WASM channels -- writes lost on unrelated poll failures.
8. Separate prep changes (thread_ops refactoring, log levels, serde changes) into a separate PR.

### Nice-to-have

9. No test for `collect_follow_up_bundles` state machine
10. No test for QR refresh exhaustion

src/extensions/manager.rs
```diff
@@ -7070,6 +7463,10 @@ impl ExtensionManager {
             }
         }
 
+        if kind == ExtensionKind::WasmChannel && name == WECHAT_CHANNEL_NAME {
+            self.persist_wechat_bound_user_id(user_id).await?;
```
serrrfirat, line 7759: **Critical Severity**\n\nThis stores the WeChat `bound_user_id` under the manager-wide owner scope (`self.user_id`) rather than under a per-installation or already-bound identity. A later interactive login by any authenticated user will overwrite that global binding, and `resolve_message_scope_with_pairing()` will then route all inbound WeChat messages to the new user.\n\nConcrete failure mode: user A binds WeChat successfully, then user B completes the same QR flow later. After this call runs, subsequent inbound WeChat DMs resolve to user B even though this is the same installed channel instance. In multi-user gateway mode that is a cross-user channel takeover.\n\nSuggested fix: either (a) reject rebinding once a WeChat channel is already bound unless an explicit admin/unbind flow clears it first, or (b) scope the binding to the installed channel instance / owner instead of the interactive caller. I would also add a regression test that user B cannot steal an existing binding from user A.

crates/ironclaw_gateway/static/js/surfaces/extensions.js
```diff
@@ -760,7 +787,222 @@ function renderConfigureModal(name, secrets, setupFields, onboarding, options) {
...
+    link.style.display = '';
+  }
+}
+
+function maybeOpenInteractiveLoginUrl(name, overlay, res) {
+  if (name !== 'wechat' || !overlay || !res || !res.qr_code_url) return;
+  const qrUrl = res.qr_code_url;
+  if (overlay.dataset.interactiveLoginLastOpenedUrl === qrUrl) return;
+  overlay.dataset.interactiveLoginLastOpenedUrl = qrUrl;
+  window.open(qrUrl, '_blank', 'noopener,noreferrer');
```
serrrfirat, line 898: **Medium Severity**\n\n`qr_code_url` is coming from the external WeChat login API and is opened directly in the browser without any scheme/host validation here. The same file already validates OAuth and setup URLs with `parseHttpsExternalUrl()`, but the interactive-login path skips that protection.\n\nThat means a compromised or unexpected upstream response can make the gateway open an arbitrary URL in a new tab, including non-HTTPS schemes that the rest of this surface intentionally rejects. Even if the backend trusts the QR-login host, the returned QR URL itself is still untrusted input.\n\nSuggested fix: validate `qr_code_url` before using it client-side (reuse `parseHttpsExternalUrl`) and ideally also enforce the allowed scheme/host server-side before returning it from the interactive-login API.

src/extensions/manager.rs
```diff
@@ -6784,15 +6956,236 @@ impl ExtensionManager {
...
+            };
+        #[cfg(not(test))]
+        let login_result = start_wechat_login(user_id, &base_url, &bot_type).await;
+
+        let (session, result) = login_result?;
+
+        self.pending_wechat_logins
+            .write()
+            .await
+            .insert(session.session_id.clone(), session);
```
serrrfirat, line 7355: **Medium Severity**\n\nEach successful login start stores a fresh entry in `pending_wechat_logins`, but there is no per-user cap, no global cap, and no route-level rate limiting on `/api/extensions/{name}/login/start`. An authenticated user can repeatedly hit the start endpoint, force the server to fetch new QR sessions from WeChat, and grow this in-memory map until the TTL cleanup path eventually runs.\n\nThat is both memory amplification and upstream-request amplification from a single authenticated account. The current cleanup only removes expired entries opportunistically on later start/poll calls; it does not bound the live set.\n\nSuggested fix: add a per-user/session limiter (for example: one active WeChat login session per user, or a small bounded count), reject or replace older sessions on new starts, and add HTTP rate limiting for the login start/poll routes.

think-in-universe (approved):
I'd like to approve this PR as testing work well. 

I didn't go through all the code changes by myself due to time constraint, but don't want to block the feature.

think-in-universe (dismissed):

zmanian (approved):
Approving — the silk-decoder isolation from #3341 has been folded into this branch, so the main `cargo build` no longer requires libclang while WeChat voice memos are still transcribed end-to-end via the standalone `crates/ironclaw_silk_decoder/` helper. Verified the silk-decoder crate, build.sh, and the host-side subprocess invocation are all present at HEAD `204bd3e23`.

PR #1667, opened 2026-03-26
feat(skills): recursive bundle directory scanning

### LINKED ISSUE #1664: feat(skills): support recursive discovery for grouped/bundled skill directories
## Problem / Motivation

Skill discovery in `discover_from_dir` (`src/skills/registry.rs:178-273`) only scans one level deep. It supports two layouts:

- `skills/SKILL.md` (flat)
- `skills/<name>/SKILL.md` (single subdirectory)

If a subdirectory does **not** contain a `SKILL.md`, it is silently skipped — even if it contains valid skills nested one level further. This prevents discovery of skills organized in group/bundle directories:

```
~/.ironclaw/skills/
└── my-org-skills/           # bundle — no SKILL.md here
    ├── skill-a/
    │   └── SKILL.md         # not discovered
    ├── skill-b/
    │   └── SKILL.md         # not discovered
    └── skill-c/
        └── SKILL.md         # not discovered
```

This layout is used by skill distribution platforms that organize skills by publisher or package. The [Agent Skills spec](https://agentskills.io/specification) defines a skill as a directory containing `SKILL.md` (with `name` matching the parent directory name) but does not restrict how deep within a search path that directory may appear. The current single-level scan is an implementation constraint, not a spec requirement.

## Proposed Solution

When `discover_from_dir` encounters a subdirectory that has **no `SKILL.md`**, recurse into it instead of skipping it. This handles both the current flat layout and grouped layouts without any configuration change.

Specifically, in `registry.rs` lines 230-251, the current logic:

```rust
// Case 1: Subdirectory containing SKILL.md
if meta.is_dir() {
    let skill_md = path.join("SKILL.md");
    if tokio::fs::try_exists(&skill_md).await.unwrap_or(false) {
        // ... load skill
    }
    continue;  // <-- skips directory if no SKILL.md
}
```

Should become (pseudocode):

```rust
if meta.is_dir() {
    let skill_md = path.join("SKILL.md");
    if tokio::fs::try_exists(&skill_md).await.unwrap_or(false) {
        // ... load skill as today
    } else {
        // No SKILL.md — recurse into this directory
        let nested = self.discover_from_dir(&path, trust, make_source).await;
        results.extend(nested);
    }
    continue;
}
```

### Safeguards

- **Max depth** — cap recursion at 3 levels to prevent runaway traversal (configurable via `SKILLS_MAX_SCAN_DEPTH`, default 3)
- **Respect `MAX_DISCOVERED_SKILLS`** — the existing cap still applies across all levels
- **Symlink policy unchanged** — symlinked entries are still rejected at each level

rajulbhatnagar:
## Summary

- Add recursive scanning of skill directories so bundle layouts like `skills/my-org/skill-a/SKILL.md` are discovered automatically
- Subdirectories without `SKILL.md` are treated as bundle dirs and recursed into up to a configurable depth (`SKILLS_MAX_SCAN_DEPTH`, default 3)
- Discovery cap (`MAX_DISCOVERED_SKILLS=100`) is shared across recursive levels via `remaining_cap` parameter
- Eliminates TOCTOU race by replacing `try_exists` + `read_dir` with a single `read_dir` call that handles `NotFound`

Closes #1664

## Test plan

- [x] `test_discover_nested_bundle_directory` — two skills inside a bundle dir are found
- [x] `test_discover_respects_depth_limit` — depth=2 misses a 3-level-deep skill, depth=3 finds it
- [x] `test_discover_cap_spans_recursive_levels` — 120 skills across two bundles capped at 100
- [x] `test_symlink_rejected_in_nested_directory` — symlinks inside bundles are rejected
- [x] `test_discover_nested_plus_direct` — mixed flat + nested discovery works
- [x] `test_discover_dedup_direct_vs_bundle_same_name` — same skill name in direct + bundle deduped to one
- [x] Full suite: 3681 passed, 0 failed

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -857,7 +857,8 @@ impl AppBuilder {
         // Skills system
         let (skill_registry, skill_catalog) = if self.config.skills.enabled {
             let mut registry = SkillRegistry::new(self.config.skills.local_dir.clone())
-                .with_installed_dir(self.config.skills.installed_dir.clone());
+                .with_installed_dir(self.config.skills.installed_dir.clone())
+                .with_max_scan_depth(self.config.skills.max_scan_depth);
             let loaded = registry.discover_all().await;
             if !loaded.is_empty() {
                 tracing::debug!("Loaded {} skill(s): {}", loaded.len(), loaded.join(", "));
--- a/src/cli/skills.rs
+++ b/src/cli/skills.rs
@@ -69,7 +69,8 @@ pub async fn run_skills_command(
 /// Discover skills from all configured directories.
 async fn discover_skills(config: &SkillsConfig) -> SkillRegistry {
     let mut registry = SkillRegistry::new(config.local_dir.clone())
-        .with_installed_dir(config.installed_dir.clone());
+        .with_installed_dir(config.installed_dir.clone())
+        .with_max_scan_depth(config.max_scan_depth);
     registry.discover_all().await;
     registry
 }
--- a/src/config/skills.rs
+++ b/src/config/skills.rs
@@ -19,6 +19,9 @@ pub struct SkillsConfig {
     pub max_active_skills: usize,
     /// Maximum total context tokens allocated to skill prompts.
     pub max_context_tokens: usize,
+    /// Maximum recursion depth when scanning skill directories for bundle layouts.
+    /// Subdirectories without `SKILL.md` are recursed into up to this depth.
+    pub max_scan_depth: usize,
 }
 
 impl Default for SkillsConfig {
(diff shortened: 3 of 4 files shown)


zmanian (commented):
## Review: feat(skills): recursive bundle directory scanning

Solid implementation -- symlink rejection, depth limiting, `Box::pin` for async recursion all correct.

### Medium

1. **Global cap not shared across sources**: `discover_all()` passes `MAX_DISCOVERED_SKILLS` (100) independently to workspace, user, and installed scans. Effective cap is 300, not 100. Either thread remaining capacity across calls or rename to `MAX_SKILLS_PER_SOURCE`.

### Positive

- Symlinks caught at entry level via `symlink_metadata` before recursion
- `max_scan_depth` (default 3) prevents unbounded recursion
- `remaining_cap` correctly threaded within a single source's recursion tree
- Good test coverage: 6 tests including nested discovery, depth limits, cap, symlink rejection, dedup

Fix the cross-source cap issue and add a test for it.

PR #1243, opened 2026-03-16
feat(jobs): per-job MCP server filtering and max_iterations cap

nick-stebbings:
## Summary

- Add `mcp_servers` and `max_iterations` optional parameters to `create_job` tool
- `mcp_servers` filters which MCP servers are mounted into worker containers, gated behind `MCP_PER_JOB_ENABLED` env var (default false)
- `max_iterations` caps the worker agent loop iteration count (default 50, max 500)
- Per-job MCP configs written to temp files, mounted read-only, cleaned up on job completion

## Motivation

Deployments running multiple MCP servers need per-job scoping — a research job should only access research tools, not production APIs. Simple data-gathering jobs shouldn't burn 50 iterations when 10 suffice.

## Test plan

- [x] `cargo test --lib` passes (3155 tests)
- [x] `cargo clippy --all --all-features` zero warnings
- [x] `cargo fmt --check` clean
- [ ] Manual: create job with `mcp_servers: ["serpstat"]` → only serpstat in container config
- [ ] Manual: create job with `max_iterations: 5` → worker exits at 5
- [ ] Manual: create job without params → full config mounted (backward compatible)
- [ ] Verify `MCP_PER_JOB_ENABLED=false` (default) ignores `mcp_servers` param

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1675, opened 2026-03-26
fix(safety): add credential patterns and sensitive path blocklist

j-bloggs:
## Summary
- Move is_sensitive_path to ironclaw_safety crate as shared module (sensitive_paths.rs)
- Guard ListDirTool with sensitive path checks at top-level AND recursive traversal
- Annotate blocked directories with [sensitive - access blocked] in listings
- Add missing sensitive paths: ~/.config/gh/hosts.yml, /etc/shadow, ~/.terraform.d/credentials.tfrc.json, ~/.azure/
- Add path traversal regression test

## Change Type
- [x] Bug fix (non-breaking change which fixes an issue)

## Linked Issue
Addresses review feedback from zmanian, gemini-code-assist, and copilot on this PR.

## Validation
- cargo test -p ironclaw_safety -- sensitive_paths: 9 tests pass
- cargo test --lib -- tools::builtin::file: 21 tests pass

## Security Impact
- Closes ListDirTool enumeration gap (attacker could enumerate ~/.ssh/, ~/.aws/ contents)
- Shell tool bypass via cat ~/.ssh/id_rsa is a known limitation. Shell DANGEROUS_PATTERNS provide partial coverage but are bypassable. Full mitigation requires filesystem-level sandboxing.

## Database Impact
None

## Blast Radius
- ListDirTool may now block listing of directories it previously allowed (sensitive credential dirs)
- File tools now import is_sensitive_path from ironclaw_safety instead of local definition

## Rollback Plan
Revert to staging (no sensitive path blocking). Credential patterns are additive and safe to keep.

## Review Track
Track C - security changes in crates/ironclaw_safety/ and src/tools/builtin/

## Feature Parity
No FEATURE_PARITY.md changes needed.


zmanian (commented):
## Security Review: credential patterns and sensitive path blocklist

Solid security hardening. Good credential patterns for OpenRouter/Anthropic OAuth/Telegram/Groq. However:

### Critical

1. **`ListDirTool` not guarded**: `ReadFileTool`, `WriteFileTool`, `ApplyPatchTool` get `is_sensitive_path` checks but `ListDirTool` does not. Attacker can enumerate `~/.ssh/`, `~/.aws/`, `~/.gnupg/` contents.

### High

2. **Shell tool bypass**: `cat ~/.ssh/id_rsa` via the shell tool trivially reads any sensitive file. At minimum document as known limitation with a tracking issue.

### Medium

3. Missing sensitive paths: `~/.config/gh/hosts.yml` (GitHub CLI), `/etc/shadow`, `~/.terraform.d/credentials.tfrc.json`.
4. No test for path traversal bypass (`../../.ssh/id_rsa`). `canonicalize()` should handle it but an explicit test would strengthen confidence.

### Low

5. Telegram pattern `\d{8,12}:AA` could false-match log lines with timestamps. Low risk but document the `:AA` requirement.

### Positive

- Credential patterns are well-crafted with appropriate minimum lengths
- Symlink-through-canonicalize defense is correct
- `.env.example`/`.env.local` safe suffixes are thoughtful
- 6 tests covering credential detection

Fix the `ListDirTool` gap and track the shell bypass before merge.

zmanian (commented):
## Re-review: credential patterns and sensitive path blocklist

3 of 4 original findings fixed. Good progress.

| Finding | Status |
|---------|--------|
| CRITICAL: ListDirTool not guarded | **Fixed** -- blocks sensitive dirs, annotates as [sensitive - access blocked] |
| HIGH: Shell tool bypass | **Still open** -- cat/head/less on protected paths still works via shell tool |
| MEDIUM: Missing sensitive paths | **Fixed** -- added gh/hosts.yml, /etc/shadow, terraform, azure |
| MEDIUM: No path traversal test | **Fixed** -- canonicalize + raw string matching tests |

The shell tool bypass remains the main gap. Consider synchronizing DANGEROUS_PATTERNS with SENSITIVE_PATH_PATTERNS, or integrating is_sensitive_path() into shell argument scanning.

zmanian (commented):
Re-review of commit 46e28467. See detailed findings below.

zmanian (commented):
## Re-review: shell tool integration (commit 46e28467)

The new commit addresses the HIGH-priority shell tool bypass. Good progress.

### What was done well

- **Correct placement**: `check_sensitive_file_access` runs after injection detection, before execution
- **Shared `is_sensitive_path` reuse** from `ironclaw_safety` -- no duplicated logic
- **Tilde expansion**, pipe/semicolon splitting, input redirection, full-path command stripping all handled
- **Honest documentation**: clearly states this is defense-in-depth, not a complete sandbox
- **8 regression tests** covering key scenarios

### Previous findings status

All 4 previous findings (ListDirTool, shell bypass, missing paths, traversal test) are fixed.

### New findings

**Important (should fix):**

1. **Output redirection bypass**: The function checks read commands but not write targets via `>` or `>>`. An attacker could write to sensitive credential paths via redirection. Consider scanning redirect targets against `is_sensitive_path`. This mirrors the existing comment on line 902 about redirect-aware parsing.

2. **Subshell / command substitution bypass**: `$()` and backtick substitution not handled. A sensitive read nested in command substitution would not be caught. Exploitability is lower since `detect_command_injection` partially covers this upstream. Document the gap.

3. **Ampersand splitting fragility**: Splitting on single `&` incorrectly splits `&&` into segments. Works by accident (empty segment is harmless) but is fragile. Consider splitting on `["&&", "||", "|", ";"]` explicitly.

### Verdict

Meaningful security improvement that closes the most obvious bypass. Known limitations honestly documented. Items 1-3 worth addressing before merge -- item 1 (output redirection) is most actionable as the write-path equivalent of the read-path protection this commit adds.

Overall: good security work. Substantially better than at the start of the review cycle.

zmanian (changes requested):
## Request Changes

Credential patterns are well-crafted (all ReDoS-safe). Putting `is_sensitive_path` in `ironclaw_safety` is architecturally correct. However, there's a critical conflict.

### Must-fix

1. **Direct conflict with PR #1713** -- both implement sensitive path protection, touching the exact same insertion points in `file.rs` and `shell.rs`. Recommend: use this PR as base (safety crate is the right home), absorb #1713's more complete path list, close #1713 as superseded.

2. **Telegram pattern needs trailing `\b`** -- `\b\d{8,12}:AA[A-Za-z0-9_-]{30,}` could false-positive on structured data.

### Should-fix

3. Missing patterns from #1713: `.bash_history`, `.zsh_history`, `.histfile`, `/etc/gshadow`, key extensions (`.pem`, `.key`, `.p12`, `.pfx`), individual key files (`id_ed25519`, `id_ecdsa`, `id_dsa`)
4. `grep` missing from `FILE_READ_COMMANDS` -- also consider `awk`, `sed`, `python`, `curl`
5. `check_redirect_target` only finds first `>` or `<` per segment
6. `.env` error message should point to specific IronClaw secrets management commands

ilblackdragon (changes requested):
## Additional Review Findings

Beyond zmanian's existing feedback (absorb #1713 patterns, `.env` error message, trailing `\b` on Telegram pattern):

### Critical

1. **Quoted arguments bypass shell check** — `check_segment_file_commands` splits on whitespace but doesn't strip quotes. `cat "/home/user/.ssh/id_rsa"` passes quotes through to `is_sensitive_path`, which tries to match against `"/home/user/.ssh/id_rsa"` (with quotes). `canonicalize()` fails on the quoted path, falls back to raw string, won't match patterns. Fix: strip surrounding quotes from tokens before checking.

2. **`grep`, `awk`, `sed` missing from `FILE_READ_COMMANDS`** — `grep ~/.ssh/id_rsa -l` or `grep -r . ~/.ssh/` are trivial reads. These are obvious omissions.

### High

3. **Directory listing not blocked for `.aws/`, `.docker/`, `.kube/`** — only specific files within them are blocked (`/.aws/credentials`, `/.docker/config.json`). `ListDirTool` can enumerate these directories to discover files. Change to directory-level patterns or add both.

4. **`check_redirect_target` only finds the FIRST `>` or `<` per segment** — `echo hello > /tmp/safe > ~/.ssh/authorized_keys` only checks `/tmp/safe`, missing the second redirect.

### Medium

5. **`canonicalize()` is blocking I/O** — called from async context. Consider `tokio::fs::canonicalize`.

6. **Duplicate tests between `sensitive_paths.rs` and `file.rs`** — near-identical unit tests in both. Keep integration-level tests in `file.rs`, remove duplicated `is_sensitive_path` unit tests.

7. **`test_list_dir_blocks_sensitive_directory` accepts "Failed to read directory" as passing** — the test passes even if the sensitive path check never fires. Use `tempfile` to create a real directory and assert the specific error.

j-bloggs:
Pre-push adversarial review found 2 critical + 2 high issues. Fixed in 73aa0b0a:

**CRITICAL:**
1. `.env.production.dist` bypass — `ends_with(".dist")` matched compound suffixes. Fixed: exact remainder match after stripping `.env.` prefix. Test: `blocks_env_with_compound_suffix`.
2. Process substitution `<(cat ~/.ssh/id_rsa)` bypass — `<` found but `(cat` treated as path token. Fixed: detect `<(` and scan parens content for sensitive paths. Test: `sensitive_file_access_blocks_process_substitution`.

**HIGH:**
3. Missing `/id_rsa` standalone pattern (all other SSH key types had it). Added. Test: `blocks_standalone_id_rsa`.
4. `--file=/home/user/.env` bypass — tokens starting with `-` skipped entirely. Fixed: extract value after `=` and check against `is_sensitive_path`. Test: `sensitive_file_access_blocks_flag_equals_path`.

All 25 sensitive path/shell tests pass. Clippy clean.

zmanian (commented):
## Re-review: credential patterns and sensitive path blocklist

All CI checks pass. PR #1713 is closed (superseded). All previous review findings from zmanian and ilblackdragon have been addressed. Good progress across the review cycle.

### Status of previous findings

| Finding | Status |
|---------|--------|
| CRITICAL: ListDirTool not guarded | Fixed -- blocks at top-level AND recursive traversal with annotation |
| HIGH: Shell tool bypass | Fixed -- `check_sensitive_file_access` with `FILE_READ_COMMANDS` list |
| HIGH: Quoted arguments bypass (ilblackdragon) | Fixed -- `strip_shell_quotes()` strips single/double quotes |
| MEDIUM: Missing sensitive paths | Fixed -- gh/hosts.yml, /etc/shadow, /etc/gshadow, terraform, azure, shell history, key types |
| MEDIUM: Path traversal test | Fixed -- explicit test in both `sensitive_paths.rs` and `file.rs` |
| MEDIUM: `grep`/`awk`/`sed` missing from `FILE_READ_COMMANDS` | Fixed -- all three added |
| MEDIUM: Multiple redirect scanning | Fixed -- `check_redirect_target` scans all occurrences, not just first |
| LOW: `&&` splitting fragility | Fixed -- `split_shell_segments` handles `&&`/`||` as multi-char operators |

### Remaining issues

**Medium -- should fix before merge:**

1. **Blocking `canonicalize()` in async context** (`crates/ironclaw_safety/src/sensitive_paths.rs:254`): `is_sensitive_path()` calls `path.canonicalize()` which is a blocking syscall. This is called from async `Tool::execute` implementations in `file.rs`. Should either use `tokio::fs::canonicalize` (requires making `is_sensitive_path` async) or wrap in `tokio::task::spawn_blocking`. Under normal conditions the latency is negligible, but on network filesystems or under load this could block the tokio runtime.

2. **Overly broad standalone key patterns** (`crates/ironclaw_safety/src/sensitive_paths.rs:232-236`): Patterns like `/id_rsa`, `/id_ed25519`, `/id_dsa`, `/authorized_keys`, `/known_hosts` match anywhere in a path. A file at `/home/user/project/test_fixtures/id_rsa` or `/app/data/authorized_keys.bak` would be blocked. These are already covered by the `/.ssh/` directory pattern for the real threat. Consider requiring these to be leaf filenames (check `file_name()` instead of substring match) or document the intentionally aggressive matching.

3. **Duplicate test suites** (flagged by ilblackdragon, still present): `sensitive_paths.rs` and `file.rs` have near-identical unit tests for `is_sensitive_path`. The unit tests belong in `sensitive_paths.rs` (the owning module). `file.rs` should have integration-level tests that exercise the tool's `execute()` method (like the existing `test_list_dir_blocks_sensitive_directory`), not re-test `is_sensitive_path` directly.

**Low -- non-blocking suggestions:**

4. **`python` and `curl` not in `FILE_READ_COMMANDS`**: `python -c "open('/home/user/.env').read()"` or `curl file:///home/user/.ssh/id_rsa` are bypass vectors. Low priority since these are increasingly exotic and the defense-in-depth documentation is honest about limitations.

5. **Process substitution check is shallow** (`shell.rs` `check_redirect_target`): `<(cat ~/.ssh/id_rsa)` is caught, but only because it scans all tokens inside the parens. A more complex expression like `<(base64 -d < ~/.ssh/id_rsa)` with nested redirection inside the substitution would not recurse. Acceptable for defense-in-depth but worth noting.

### What's good

- Correct architectural placement in `ironclaw_safety` crate
- Credential patterns are all ReDoS-safe (linear regex, no nested quantifiers)
- Shell segment splitting is well-implemented with proper multi-char operator handling
- `[sensitive - access blocked]` annotation in recursive listing is user-friendly
- Honest defense-in-depth documentation throughout
- Error messages point to `secret_list`/`secret_create` for the correct workflow
- 30+ tests with both positive and negative cases
- Telegram pattern has trailing `\b` and `:AA` prefix constraint (addresses earlier false-positive concern)

### Verdict

Solid security hardening. Items 1-3 above are worth addressing (especially #1 re: blocking I/O in async). None are security vulnerabilities -- they're correctness and maintainability issues. The core security value (closing the ListDirTool enumeration gap, adding shell tool file-access checking, expanding credential patterns) is well-executed.

j-bloggs:
Addressed all 3 medium items from re-review:

1. **Blocking canonicalize()**: Documented the trade-off — sub-millisecond on local FS, with guidance to switch to `tokio::fs::canonicalize` if NFS/CIFS becomes a concern. Making `is_sensitive_path` async would cascade through all callers (file tools, shell tool, ListDirTool).

2. **Overly broad standalone key patterns**: Created `SENSITIVE_FILENAMES` list (`id_rsa`, `id_ed25519`, `id_ecdsa`, `id_dsa`, `authorized_keys`, `known_hosts`). Now checked via exact `file_name()` match instead of substring. `/project/test_fixtures/id_rsa` IS still blocked (it is a real key filename), but `/project/grid_rsa_data` is NOT (substring, not filename match).

3. **Duplicate test suites**: Removed 10 `is_sensitive_path` unit tests from `file.rs` (they belong in `sensitive_paths.rs` which has comprehensive coverage). Kept integration-level `test_list_dir_blocks_sensitive_directory`.

All 31 tests pass. Clippy clean.

zmanian (approved):
## Re-review: commit 339977a (fix 3 items from previous review)

All 3 medium findings from my last review are addressed:

| Finding | Status | Notes |
|---------|--------|-------|
| Blocking `canonicalize()` in async context | **Addressed** | Documented trade-off with guidance to make async if needed. Pragmatic -- local FS is sub-ms. |
| Overly broad standalone SSH key patterns | **Fixed** | Moved to `SENSITIVE_FILENAMES` with exact `file_name()` match. Test confirms `grid_rsa_data` no longer false-positives. |
| Duplicate test suites | **Fixed** | Removed duplicate `is_sensitive_path` unit tests from `file.rs`. Only integration-level `execute()` test remains. |

No new issues in this commit. LGTM.

j-bloggs:
Hey\! 👋 Just checking in — this one's still good to go (no conflicts with staging). Would appreciate a re-review when you have a moment\! @ilblackdragon @serrrfirat

serrrfirat:
Reviewing the full worktree relative to `origin/main` surfaced a few correctness issues outside the narrow PR-vs-`staging` diff. Posting here for visibility because they are present in this branch checkout, but some may already be inherited from the base branch.

Findings:

1. High: Legacy conversations will stop accepting approvals after a restart.
`src/agent/thread_ops.rs` now treats a missing `source_channel` as fail-closed, but the migration only adds a nullable column and does not backfill existing rows (`migrations/V15__conversation_source_channel.sql`). Any pre-migration conversation rehydrated from DB will therefore reject approvals forever, even from its original channel. This needs either a migration backfill (`source_channel = channel`) or a legacy fallback when the stored value is null.

2. Medium: Sandbox job restart silently drops the new `mcp_servers` and `max_iterations` inputs.
`src/tools/builtin/job.rs` parses both fields, but the persisted sandbox job record only stores `credential_grants_json` (`src/history/store.rs`). The restart handler then recreates the job with `JobCreationParams { credential_grants, ..Default::default() }` (`src/channels/web/handlers/jobs.rs`). A restarted job will therefore run with the default worker iteration cap and without its original MCP filter.

3. Medium: The new per-job MCP mounting logic points at the wrong source of truth.
`src/orchestrator/job_manager.rs` hardcodes `/opt/ironclaw/config/worker/mcp-servers.json`, but the normal MCP config path is `~/.ironclaw/mcp-servers.json` (`src/tools/mcp/config.rs`), and bootstrap migrates that file into DB anyway (`src/bootstrap.rs`). The worker image also never creates `/opt/ironclaw/config/worker/...` (`Dockerfile.worker`). So with `MCP_PER_JOB_ENABLED=true`, most installs will mount no MCP config at all, and the feature will not behave as advertised.

Residual risk:
- Job-level approval allowlists still do not survive DB restore (`src/db/libsql/jobs.rs`, `src/history/store.rs` both hydrate `approval_context: None`).
- The shell sensitive-path protection is still heuristic by design, not a hard guarantee (`src/tools/builtin/shell.rs`).
- I did not get a completed result from `cargo test --test tool_approval_context`; it spent the turn in compile/link work, so these findings are from code inspection rather than a finished test run.

j-bloggs:
Thanks for the thorough review @serrrfirat! 🙏

I've looked at all three findings — they're all pre-existing issues inherited from the staging base branch, not introduced by this PR's sensitive-path protection changes:

1. **`source_channel` backfill** — This is from #1590's `V15__conversation_source_channel.sql` migration (already merged to staging). Our PR doesn't touch conversations or approvals.
2. **Sandbox `mcp_servers`/`max_iterations` persistence** — This is from #1243's per-job MCP feature (already on staging). Our PR only touches `ironclaw_safety` and file tool paths.
3. **MCP config path mismatch** — Same as above, from #1243.

Happy to file separate issues for these if the maintainers want them tracked. Our diff is limited to `crates/ironclaw_safety/src/` (leak detector patterns) and `src/tools/builtin/file.rs` (path validation calls).

ilblackdragon (approved):

PR #1678, opened 2026-03-26
fix(worker): wrap Claude CLI in PTY to fix stdout buffering hang

j-bloggs:
## Summary
- Replace script -qfc shell-string approach with pty-process crate for PTY allocation
- Revert to Command::arg() chaining (injection-safe by construction via execve)
- Eliminates all shell injection surfaces: prompt, model, session_id
- Keep stderr on separate pipe to prevent NDJSON parse breakage
- Gate PTY behind #[cfg(unix)] with direct-spawn fallback for Windows CI
- Remove shell_escape() function and tests (no longer needed)

## Change Type
- [x] Bug fix (non-breaking change which fixes an issue)

## Linked Issue
Addresses review feedback from zmanian and gemini-code-assist on this PR.

## Validation
- cargo test --lib -- claude_bridge: 23 tests pass
- pty-process crate: MIT licensed, depends on nix (MIT)

## Security Impact
- Eliminates entire class of shell injection vulnerabilities by removing the shell interpretation layer
- Previously: prompt, model name, and session_id were interpolated into a shell string for script -qfc
- Now: all arguments passed via execve (Command::arg), never shell-interpreted

## Database Impact
None

## Blast Radius
- New dependency: pty-process 0.5 (MIT, Unix only via cfg gate)
- Claude bridge stdout now read from PTY master instead of child.stdout
- Windows builds use non-PTY fallback (compile-only, bridge runs in Linux Docker)

## Rollback Plan
Revert to staging Command::new("claude").arg() without PTY (re-introduces buffering hang but eliminates security risk).

## Review Track
Track C - runtime changes in src/worker/, new dependency

## Feature Parity
No FEATURE_PARITY.md changes needed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3150,7 +3150,7 @@ dependencies = [
  "libc",
  "percent-encoding",
  "pin-project-lite",
- "socket2 0.5.10",
+ "socket2 0.6.3",
  "system-configuration",
  "tokio",
  "tower-service",
@@ -3439,6 +3439,7 @@ dependencies = [
  "pgvector",
  "postgres-types",
  "pretty_assertions",
+ "pty-process",
  "rand 0.8.5",
  "readabilityrs",
  "refinery",
@@ -3523,7 +3524,7 @@ checksum = "3640c1c38b8e4e43584d8df18be5fc6b0aa314ce6ebf51b53313d4306cca8e46"
 dependencies = [
  "hermit-abi",
  "libc",
- "windows-sys 0.59.0",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


zmanian (commented):
## Review: fix(worker): wrap Claude CLI in PTY to fix stdout buffering hang

PTY is the right fix for Node.js stdout buffering, but the implementation introduces security risk.

### Critical

1. **Shell injection via unescaped `self.config.model`**: Interpolated directly into the shell command string. Must be escaped with `shell_escape()` like `prompt` and `session_id` are.

### High

2. **Security regression from `Command::arg()` to shell string**: Previous code passed args via `execve` (injection-safe by construction). New code goes through `script -qfc` which invokes `/bin/sh -c`. Consider using the `pty-process` crate for PTY allocation while keeping `Command::arg()`.

### Medium

3. **`script -qfc` is GNU-only**: macOS `script` has different syntax. Documented as Docker-only but a maintenance trap for local testing.

4. **stdout/stderr separation through PTY**: `script` can merge streams. Verify NDJSON parsing still works when stderr lines from Claude CLI are potentially interleaved into stdout.

### Positive

- Thorough escaping tests covering quotes, metacharacters, adversarial session IDs.

Fix the `model` escaping at minimum. Strongly consider the PTY crate approach to avoid the shell interpretation layer entirely.

PR #1692, opened 2026-03-27
fix(worker): Improve command execution parameter validation

ArakawaHenri:
Enhance workdir and timeout parameter handling for command execution.

## Summary

- Normalize empty `workdir` values to use the default working directory.
- Reject invalid `timeout` values instead of silently falling back to the default.
- Keep runtime behavior consistent with the tool schema.

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] CI/Infrastructure
- [ ] Security
- [ ] Dependencies

## Validation

- [ ] `cargo fmt`
- [ ] Relevant tests pass: 
- [ ] Manual testing: 

## Security Impact

This change tightens parameter handling for the `shell` tool.

- `workdir=""` (or whitespace-only) is now treated as "not provided" and falls back to the tool's default working directory instead of being passed through as an invalid path. This avoids spawn failures caused by `current_dir("")`, but does not expand filesystem access or change sandbox boundaries.
- `timeout` parsing is now fail-closed. Only JSON integer values are accepted, consistent with the tool schema. Invalid values such as strings, floats, booleans, negative numbers, or `0` now return `ToolError::InvalidParameters` instead of silently falling back to the default timeout.

No new capabilities are introduced. No approval, sandbox, or command filtering behavior is relaxed.

## Rollback Plan

---

**Review track**:


zmanian (commented):
## Review: fix(worker): Improve command execution parameter validation

### Medium

1. **Silent swallowing of invalid timeout values**: Non-numeric strings like `"abc"` or `"-1"` silently fall back to 120s default via `parse::<u64>().ok()`. This is fail-open in a security-sensitive context. Return `ToolError::InvalidParams` for non-null, non-empty values that fail to parse.

2. **Questionable string-to-integer coercion**: The tool schema declares `timeout` as `"type": "integer"`. Silently accepting string-encoded integers masks upstream serialization bugs. If there's a concrete scenario where string values arrive, document it.

### Low

3. Empty PR description -- Security Impact section should be filled in for a shell execution change.

The `workdir` trimming is fine. Fix the fail-open timeout parsing.

zmanian (approved):
Re-review: fail-open timeout parsing fixed -- invalid values now return ToolError::InvalidParameters. Also rejects timeout: 0. Approve.

zmanian (changes requested):
## Request Changes

The validation logic is correct and well-scoped -- fail-closed timeout parsing and workdir normalization are the right fixes.

### Must-fix

1. **No tests added** -- the project requires regression tests with every bug fix. Please add tests for:
   - `workdir: ""` and `workdir: "   "` treated as `None`
   - `timeout: null` / absent treated as `None`
   - `timeout: "abc"` returns `InvalidParameters`
   - `timeout: 0` returns `InvalidParameters`
   - `timeout: 3.5` (float) returns `InvalidParameters`
   - Valid `timeout: 30` is accepted

### Positive notes
- The `match` on `params.get("timeout")` is well-structured
- Error messages are descriptive
- The workdir chain is idiomatic Rust
- Scope is focused

zmanian (dismissed):
## Review -- APPROVE

Previous review feedback addressed. Fail-closed timeout validation is the right fix -- rejects non-integer types, null handled as None, zero rejected. Tests are thorough (7 new cases).

- **workdir validation asymmetry**: timeout rejects non-integer types with InvalidParameters, but `workdir: 42` is silently ignored via `as_str()` returning None. For consistency, consider rejecting non-null, non-string workdir types too.
- **Schema mismatch**: runtime now rejects `timeout: 0` but schema doesn't include `"minimum": 1`. Consider adding it so LLMs get the constraint upfront.
- **Error message**: "timeout must be an integer number of seconds" is slightly misleading for negative integers -- "positive integer" would be more precise.

zmanian (approved):
## Re-Review -- APPROVE

All 3 suggestions from previous review addressed in ab22b7b7:

1. **Workdir validation** -- non-string types now return `InvalidParameters` instead of being silently ignored
2. **Schema minimum** -- `"minimum": 1` added to timeout schema
3. **Error message** -- now says "positive integer" instead of just "integer"

Clean and well-tested.

PR #1699, opened 2026-03-27
fix(wasm): use typed WASM schema as advertised schema when available

### LINKED ISSUE #1303: [MEDIUM] WASM tools expose permissive {} schemas to the LLM even when the component exports a typed schema
## Problem

Some WASM tools export a valid typed schema via their `schema()` function, but IronClaw still advertises a permissive `{}` schema to the LLM.

We observed this with the Brave `web-search` WASM tool:
- the tool exports a typed schema including required fields like `query`
- but the agent only sees a permissive schema
- as a result, it repeatedly calls the tool with `{}` and gets:
  - `Invalid parameters: missing field 'query'`

## Root cause

In `src/tools/wasm/wrapper.rs`, `WasmToolSchemas::new(discovery)` currently does this:

- `discovery = extracted_wasm_schema`
- `advertised = permissive_schema()`

`ToolRegistry::tool_definitions()` uses the advertised schema, so the model does not see the component’s actual typed input schema unless a sidecar override is provided.

## Proposed fix

When a WASM component exports a typed schema, use it as the default advertised schema too, instead of always falling back to a permissive `{}` schema.

Suggested behavior:
- if the extracted WASM schema has typed properties and is reasonably small:
  - `advertised = discovery.clone()`
- otherwise:
  - keep the current permissive fallback

## Related cleanup

`src/tools/wasm/loader.rs` currently warns:

- `Capabilities file missing "parameters" field; tool will accept any JSON object`

That is misleading when the WASM component already exports a valid schema. The warning should be updated accordingly.

## Impact

This makes WASM tools much harder for the agent to use correctly, even when the tool implementation already provides a proper schema.

anthhub:
## Summary

When a WASM tool exports a typed schema with properties, use it as the advertised schema instead of always falling back to permissive `{}`. This lets the LLM see the correct parameter requirements and avoids repeated calls with empty params.

- In `wrapper.rs`: `WasmToolSchemas::new()` now checks if the extracted schema has typed properties and uses it as the advertised schema
- In `loader.rs`: updated the misleading "missing parameters" warning to account for WASM-exported schemas

Fixes #1303

## Test plan

- [x] Unit test verifying typed schema propagation
- [x] `cargo check` / `cargo check --no-default-features --features libsql` / `cargo check --all-features`
- [x] `cargo clippy --all --tests` — zero warnings

### DIFF
--- a/src/tools/wasm/loader.rs
+++ b/src/tools/wasm/loader.rs
@@ -124,8 +124,10 @@ impl WasmToolLoader {
         let wasm_bytes = fs::read(wasm_path).await?;
 
         // Read capabilities (optional) and extract OAuth refresh config
-        // and tool description. Parameter schema is auto-derived from the
-        // WASM module's schema() export (see WasmToolSchemas::compact_schema).
+        // and tool description. Parameter schema is NOT read from the
+        // capabilities file — it is auto-derived from the WASM module's
+        // schema() export at prepare time (see WasmToolSchemas::compact_schema),
+        // so no schema override is needed here.
         let (capabilities, oauth_refresh, description) = if let Some(cap_path) = capabilities_path {
             if cap_path.exists() {
                 let cap_bytes = fs::read(cap_path).await?;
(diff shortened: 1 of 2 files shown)


zmanian (approved):
Approve. Fallback logic is correct: required/enum props -> typed props -> permissive {}. The schema_is_typed_property helper is comprehensive. Good test coverage. Minor suggestion: add a test for mixed typed/untyped properties in the fallback path.

PR #1713, opened 2026-03-27
fix(security): unified sensitive path protection across shell and file tools

zmanian:
## Summary

Closes the security gap where the shell tool could bypass file-level protections (e.g., `cat ~/.ssh/id_rsa` worked even though ReadFileTool would block `~/.ssh/id_rsa`).

Adds a shared `SENSITIVE_PATH_PATTERNS` list in `path_utils.rs` used by both shell and file tools:

- **Shell tool**: `command_references_sensitive_path()` scans commands for sensitive file references
- **File tools**: `is_sensitive_path()` blocks ReadFileTool, WriteFileTool, ListDirTool, ApplyPatchTool
- **ListDirTool**: Skips sensitive subdirectories during recursive traversal

### Protected paths

SSH keys, GPG, AWS/Azure/GCP credentials, Kubernetes config, GitHub CLI tokens, Terraform credentials, Docker config, Vault tokens, shell history, .env files, git credentials, system shadow files, and sensitive key file extensions (.pem, .key, .p12, .pfx, .jks, .keystore).

Safe suffixes (.example, .sample, .template) are excluded to avoid blocking template files.

### Previous state

- Shell tool had 5 hardcoded patterns in `DANGEROUS_PATTERNS` (/etc/passwd, /etc/shadow, ~/.ssh, .bash_history, id_rsa)
- File tools had **zero** sensitive path protection
- Commands like `cat ~/.aws/credentials` or `read_file ~/.config/gh/hosts.yml` were unblocked

## Test plan

- [x] 21 new tests: path detection, command scanning, safe suffixes, normal file allowlisting
- [x] Existing path_utils tests pass (sandbox validation, traversal rejection)
- [x] `cargo check --no-default-features --features libsql` -- compiles

### DIFF
--- a/src/tools/builtin/path_utils.rs
+++ b/src/tools/builtin/path_utils.rs
@@ -82,7 +82,10 @@ pub fn is_sensitive_path(path: &Path) -> bool {
     }
 
     // Check sensitive path patterns
-    if SENSITIVE_PATH_PATTERNS.iter().any(|p| path_str.contains(p)) {
+    if SENSITIVE_PATH_PATTERNS
+        .iter()
+        .any(|p| path_pattern_matches(&lower, p))
+    {
         return true;
     }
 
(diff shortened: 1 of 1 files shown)


src/tools/builtin/path_utils.rs
```diff
@@ -4,9 +4,119 @@
...
+    };
+
+    // Safe suffixes override sensitive patterns
+    let lower = path_str.to_lowercase();
+    if SAFE_SUFFIXES.iter().any(|s| lower.ends_with(s)) {
+        return false;
+    }
+
+    // Check sensitive path patterns
+    if SENSITIVE_PATH_PATTERNS.iter().any(|p| path_str.contains(p)) {
```
serrrfirat: **HIGH (H3)**: Pattern check is case-sensitive but `lower` is already computed two lines above. On macOS (case-insensitive filesystem), if `canonicalize()` fails (file doesn't exist — relevant for WriteFileTool), the fallback uses the original input case. So `write_file("/app/.SSH/id_rsa", ...)` bypasses the check because "`/.SSH/`" doesn't match pattern "`/.ssh/`".

**Fix:** Change `path_str.contains(p)` to `lower.contains(p)` — the `lower` variable is already computed on line 79.

src/tools/builtin/path_utils.rs
```diff
@@ -4,9 +4,119 @@
...
+        if normalized.contains(&pattern.to_lowercase()) {
+            return Some(pattern);
+        }
+    }
+
+    // Check for sensitive extensions in file arguments
+    SENSITIVE_EXTENSIONS
+        .iter()
+        .find(|ext| normalized.contains(*ext))
+        .copied()
```
serrrfirat, line 137: **HIGH (H2)**: Extension matching via `contains()` on the full lowercased command string causes false positives. Examples: `cat config.keys`, `cat .keychain`, `grep pattern file.keynote`, `cat /tmp/my.keyboard.cfg` all match `.key`. Similarly `cat /tmp/template.pemx` matches `.pem`.

**Fix:** Split the command into whitespace-delimited tokens and check file extensions on each token individually, or use a regex like `\.key(?:\s|$)`.

src/tools/builtin/path_utils.rs
```diff
@@ -4,9 +4,119 @@
...
+        }
+    }
+
+    false
+}
+
+/// Scan a shell command string for references to sensitive paths.
+/// Returns the first matched pattern, or None if the command is clean.
+/// Used by the shell tool to block `cat ~/.ssh/id_rsa` etc.
+pub fn command_references_sensitive_path(command: &str) -> Option<&'static str> {
```
serrrfirat, line 120: **HIGH (H1)**: This function does pure substring matching with no safe-suffix exemption, unlike `is_sensitive_path()` which checks `SAFE_SUFFIXES` first. This means `cat /app/.env.example` is allowed by ReadFileTool but blocked by shell tool. Also blocks `source /app/.envrc` (direnv code, not secrets).

**Fix:** Apply `SAFE_SUFFIXES` exemption here too, or extract file-path tokens from the command and delegate to `is_sensitive_path()`.

PR #18, opened 2026-02-10
feat: Add lifecycle hooks system with 6 interception points

serrrfirat:
## Summary

- Add extensible hook infrastructure (`src/hooks/`) with priority-ordered execution, modification chaining, reject short-circuits, configurable failure modes (FailOpen/FailClosed), and per-hook timeouts
- Wire `HookRegistry` through the dependency injection chain (`AgentDeps` → `Scheduler` → `WorkerDeps` → `SessionManager`)
- Integrate 6 hook points: **BeforeInbound**, **BeforeToolCall** (chat + job), **BeforeOutbound**, **TransformResponse**, **OnSessionStart**, **OnSessionEnd**

## Hook Points

| Hook | Location | Behavior |
|------|----------|----------|
| `BeforeInbound` | `agent_loop.rs` handle_message | Intercept/modify/reject incoming user messages |
| `BeforeToolCall` | `agent_loop.rs` + `worker.rs` | Intercept/modify/reject tool executions |
| `BeforeOutbound` | `agent_loop.rs` main loop | Intercept/modify/suppress outgoing responses |
| `TransformResponse` | `agent_loop.rs` process_user_input | Transform final response before completing turn |
| `OnSessionStart` | `session_manager.rs` | Fire-and-forget on new session creation |
| `OnSessionEnd` | `session_manager.rs` | Fire-and-forget on session pruning |

## Files Changed

**New (3):** `src/hooks/hook.rs`, `src/hooks/registry.rs`, `src/hooks/mod.rs`
**Modified (8):** `lib.rs`, `error.rs`, `agent_loop.rs`, `scheduler.rs`, `worker.rs`, `session_manager.rs`, `main.rs`, `FEATURE_PARITY.md`

## Test plan

- [x] 11 dedicated hooks tests pass (`cargo test hooks::`)
- [x] Full unit test suite passes (535 tests)
- [x] Clippy clean (no new warnings)
- [x] Empty registry is zero-cost — all hooks pass through immediately, no behavior change

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1734, opened 2026-03-29
feat: first-class workspace entities with membership and cross-workspace sharing

### LINKED ISSUE #1607: feat: first-class workspace entities with membership and cross-workspace sharing
## Problem / Motivation

Workspaces are currently implicit — each `user_id` string gets its own siloed data. The `WorkspacePool` in `server.rs` caches per-user `Workspace` instances, but there's no shared-workspace concept beyond `workspace_read_scopes: Vec<String>` on `UserIdentity`, which is a flat string list with no DB backing, no membership semantics, and no write sharing.

This means:
- **No team collaboration** — users can't share jobs, routines, or memory within a team workspace
- **Cross-user reads are token-config-only** — `workspace_read_scopes` is set at token creation, not manageable at runtime
- **All data is scoped to a single `user_id` string** — there's no way to assign a job or routine to a team/project rather than an individual
- **No workspace lifecycle** — can't create, archive, or transfer workspaces

Depends on **Issue #1 (DB-backed user management)** — workspaces need real user records to reference.

### New Database Tables

```sql
CREATE TABLE workspaces (
    id UUID PRIMARY KEY DEFAULT gen_random_uuid(),
    name TEXT NOT NULL,
    slug TEXT NOT NULL UNIQUE,                  -- URL-safe identifier
    description TEXT NOT NULL DEFAULT '',
    status TEXT NOT NULL DEFAULT 'active',       -- active | archived
    created_at TIMESTAMPTZ NOT NULL DEFAULT now(),
    updated_at TIMESTAMPTZ NOT NULL DEFAULT now(),
    created_by UUID NOT NULL REFERENCES users(id),
    settings JSONB NOT NULL DEFAULT '{}'        -- workspace-level config overrides
);

CREATE TABLE workspace_members (
    workspace_id UUID NOT NULL REFERENCES workspaces(id) ON DELETE CASCADE,
    user_id UUID NOT NULL REFERENCES users(id) ON DELETE CASCADE,
    role TEXT NOT NULL DEFAULT 'member',         -- owner | admin | member | viewer (enforced in Issue #3)
    joined_at TIMESTAMPTZ NOT NULL DEFAULT now(),
    invited_by UUID REFERENCES users(id),
    PRIMARY KEY (workspace_id, user_id)
);
CREATE INDEX idx_workspace_members_user ON workspace_members(user_id);
```

### Data Model Changes

Every table currently scoped by `user_id TEXT` gains an optional `workspace_id UUID` column:

```sql
ALTER TABLE conversations ADD COLUMN workspace_id UUID REFERENCES workspaces(id);
ALTER TABLE agent_jobs ADD COLUMN workspace_id UUID REFERENCES workspaces(id);
ALTER TABLE memory_documents ADD COLUMN workspace_id UUID REFERENCES workspaces(id);
ALTER TABLE routines ADD COLUMN workspace_id UUID REFERENCES workspaces(id);
ALTER TABLE settings ADD COLUMN workspace_id UUID REFERENCES workspaces(id);
```

- `workspace_id = NULL` means personal/user-scoped (backward compatible)
- `workspace_id = <uuid>` means workspace-scoped — visible to all members of that workspace

### Default Personal Workspace

Every user gets an implicit "personal" workspace (not stored in the `workspaces` table). Queries with `workspace_id IS NULL AND user_id = ?` continue to work unchanged. This ensures zero migration disruption for existing data.

### New Database Traits

Add a `WorkspaceMgmtStore` sub-trait (distinct from the existing `WorkspaceStore` which handles memory documents):

```rust
#[async_trait]
pub trait WorkspaceMgmtStore {
    async fn create_workspace(&self, name: &str, slug: &str, created_by: Uuid) -> Result<WorkspaceRecord, DatabaseError>;
    async fn get_workspace(&self, id: Uuid) -> Result<Option<WorkspaceRecord>, DatabaseError>;
    async fn get_workspace_by_slug(&self, slug: &str) -> Result<Option<WorkspaceRecord>, DatabaseError>;
    async fn list_workspaces_for_user(&self, user_id: Uuid) -> Result<Vec<(WorkspaceRecord, String)>, DatabaseError>; // (workspace, role)
    async fn update_workspace(&self, id: Uuid, name: &str, description: &str) -> Result<(), DatabaseError>;
    async fn archive_workspace(&self, id: Uuid) -> Result<(), DatabaseError>;

    // Membership
    async fn add_workspace_member(&self, workspace_id: Uuid, user_id: Uuid, role: &str, invited_by: Option<Uuid>) -> Result<(), DatabaseError>;
    async fn remove_workspace_member(&self, workspace_id: Uuid, user_id: Uuid) -> Result<(), DatabaseError>;
    async fn list_workspace_members(&self, workspace_id: Uuid) -> Result<Vec<(UserRecord, String)>, DatabaseError>; // (user, role)
    async fn get_member_role(&self, workspace_id: Uuid, user_id: Uuid) -> Result<Option<String>, DatabaseError>;
    async fn update_member_role(&self, workspace_id: Uuid, user_id: Uuid, role: &str) -> Result<(), DatabaseError>;
    async fn is_workspace_member(&self, workspace_id: Uuid, user_id: Uuid) -> Result<bool, DatabaseError>;
}
```

### Query Scoping Changes

Existing `_for_user` methods in `JobStore`, `SandboxStore`, `SettingsStore`, `RoutineStore` need parallel `_for_workspace` variants:

```rust
async fn list_agent_jobs_for_workspace(&self, workspace_id: Uuid) -> Result<Vec<AgentJobRecord>, DatabaseError>;
async fn list_sandbox_jobs_for_workspace(&self, workspace_id: Uuid) -> Result<Vec<SandboxJobRecord>, DatabaseError>;
```

Handler logic becomes: if request has `?workspace=<slug>`, verify membership, query by `workspace_id`. Otherwise, query by `user_id` (personal scope).

### WorkspacePool Refactor

`WorkspacePool` in `src/channels/web/server.rs` currently caches by `user_id: String`. It must support workspace-scoped memory:

- Cache key changes from `user_id` to `(user_id, Option<workspace_id>)`
- Memory documents in a workspace are readable/writable by all members
- Personal workspace memory remains user-scoped

### Replace `workspace_read_scopes`

The current `workspace_read_scopes: Vec<String>` on `UserIdentity` is replaced by workspace membership. If a user is a `viewer` or `member` of workspace X, they can read its memory. The `hybrid_search_multi` method in `WorkspaceStore` already takes `user_ids: &[String]` — extend this to accept workspace IDs too.

### New API Endpoints

```
POST   /api/workspaces                          — create workspace
GET    /api/workspaces                          — list my workspaces
GET    /api/workspaces/{slug}                   — get workspace details
PATCH  /api/workspaces/{slug}                   — update workspace
POST   /api/workspaces/{slug}/archive           — archive workspace
POST   /api/workspaces/{slug}/members           — add member
DELETE /api/workspaces/{slug}/members/{user_id} — remove member
PATCH  /api/workspaces/{slug}/members/{user_id} — update member role
GET    /api/workspaces/{slug}/members           — list members
```

Existing endpoints (`/api/jobs`, `/api/routines`, etc.) accept an optional `?workspace=<slug>` query param to switch context.

## Architecture Notes

- **Workspace context propagation**: The `AuthenticatedUser` extractor should be augmented or a new `WorkspaceContext` extractor added that resolves the optional `?workspace=` param, verifies membership, and provides `(UserIdentity, Option<WorkspaceRecord>)` to handlers.
- **`IncomingMessage` already carries `user_id: String`** through the agent pipeline. Add an optional `workspace_id: Option<String>` alongside it rather than replacing `user_id`. The agent loop, `JobContext`, and tool execution all need to propagate this.
- **The `Workspace` struct** in `src/workspace/` is the memory/document system — don't confuse it with the new workspace *entity*. Consider naming the new entity `Team` or `Organization` if naming collisions become painful, but `workspace` matches the user-facing concept better.
- **Slug uniqueness**: enforce at DB level (`UNIQUE` constraint) and validate format (lowercase alphanumeric + hyphens, 3-48 chars) at API level.

## Code Pointers

- `src/channels/web/server.rs` — `WorkspacePool`, `GatewayState` — pool needs workspace-aware caching
- `src/channels/web/auth.rs` — `UserIdentity.workspace_read_scopes` — being replaced
- `src/db/mod.rs` — `Database` supertrait, `JobStore`, `SandboxStore`, `SettingsStore` — add `WorkspaceMgmtStore` and `_for_workspace` query variants
- `src/workspace/` — memory system `Workspace` struct — needs workspace_id-aware document scoping
- `src/agent/` — `JobContext` — needs optional `workspace_id` field
- `src/channels/channel.rs` — `IncomingMessage` — needs optional `workspace_id`
- `src/tools/builtin/memory.rs` — `WorkspaceResolver` / `CachedWorkspaceResolver` — cache key must include workspace context
- `src/channels/web/handlers/jobs.rs` — example of handler needing `?workspace=` support

## Acceptance Criteria

- [ ] `workspaces` and `workspace_members` tables in both PostgreSQL and libSQL
- [ ] `workspace_id` column added to `conversations`, `agent_jobs`, `memory_documents`, `routines`, `settings` (nullable, backward compatible)
- [ ] `WorkspaceMgmtStore` trait implemented for both backends
- [ ] API endpoints for workspace CRUD and membership management
- [ ] Membership verification before any workspace-scoped data access
- [ ] `?workspace=<slug>` query param on existing `/api/jobs`, `/api/routines`, `/api/memory/*`, `/api/settings` endpoints
- [ ] Personal workspace (no workspace_id) continues to work identically to current behavior
- [ ] `workspace_read_scopes` field deprecated in favor of membership-based access
- [ ] `WorkspacePool` caches by `(user_id, workspace_id)` tuple
- [ ] Memory documents in a workspace visible to all workspace members
- [ ] SSE broadcasts for workspace-scoped events delivered to all workspace members
- [ ] Creator automatically added as `owner` role on workspace creation
- [ ] Archived workspaces return 410 Gone on data access
- [ ] Unit tests for `WorkspaceMgmtStore` (both backends)
- [ ] Integration tests for workspace create → add member → query scoped data → remove member → verify no access
- [ ] `cargo clippy` zero warnings, `cargo fmt` clean

## Pitfalls & Landmines

- **Naming collision**: the existing `Workspace` struct in `src/workspace/` is the persistent memory system, not a team/org entity. Be very careful with naming. The new DB entity should use distinct type names like `WorkspaceEntity` or `TeamWorkspace` in Rust to avoid confusion.
- **Dual-backend**: libSQL has no `UUID` type — use `TEXT`. No arrays — use JSON. See `src/db/libsql_migrations.rs` patterns.
- **Nullable `workspace_id` migration**: existing rows have `workspace_id = NULL` which means "personal". Queries must handle `WHERE workspace_id IS NULL AND user_id = ?` OR `WHERE workspace_id = ?` — don't accidentally return personal data when querying a workspace or vice versa.
- **SSE broadcast fan-out**: workspace-scoped events need to go to all online members. `SseManager.broadcast_for_user()` currently takes a single `user_id`. You'll need a `broadcast_for_workspace()` that resolves members and broadcasts to each. Be careful about N+1 DB queries — cache workspace membership.
- **`workspace_read_scopes` backward compatibility**: the env-var `GATEWAY_USER_TOKENS` path still supports `workspace_read_scopes`. Keep it working during the deprecation period, but map scopes to workspace memberships at startup if both systems are active.
- **Don't change the existing `WorkspaceStore` trait** in `src/db/mod.rs` — it handles memory documents and has methods like `hybrid_search_multi`. The new workspace management is a separate trait.

## Non-Goals

- **RBAC enforcement** — roles are stored in `workspace_members.role` but not enforced yet. That's Issue #3.
- **Admin UI** — that's Issue #4.
- **Workspace-level billing or quotas** — out of scope.
- **Workspace templates or cloning** — out of scope.

---
*Part 2 of 4 — depends on #1605 (User Management). See also: RBAC (#3), Admin Panel (#4)*

serrrfirat:
## Summary

This PR implements `#1607`: first-class workspace entities with membership and cross-workspace sharing.

The branch now moves workspace scope from implicit per-`user_id` isolation to explicit DB-backed workspace entities and membership, while preserving the existing personal workspace behavior as the backward-compatible default.

### 1. Database and schema

- Added DB-backed user/token primitives needed by workspace membership:
  - `users`
  - `api_tokens`
- Added first-class workspace tables:
  - `workspaces`
  - `workspace_members`
- Added nullable `workspace_id` scoping to existing persisted entities:
  - `conversations`
  - `agent_jobs`
  - `memory_documents`
  - `routines`
  - `settings`
- Preserved personal scope as `workspace_id = NULL`, so existing user-scoped data keeps working without migration into a stored “personal workspace” row.
- Implemented the schema changes for both backends:
  - PostgreSQL migrations / queries
  - libSQL base schema + incremental migrations

### 2. Workspace management store

- Added a dedicated workspace-management trait in the DB abstraction layer.
- Implemented workspace CRUD and membership operations for both PostgreSQL and libSQL:
  - create workspace
  - fetch by slug / ID
  - list workspaces for a user
  - update workspace metadata/settings
  - archive workspace
  - add / remove members
  - list members
  - get / update member role
  - membership checks
- Workspace creator is automatically inserted as `owner`.

### 3. Web API surface

- Added workspace management endpoints:
  - `POST /api/workspaces`
  - `GET /api/workspaces`
  - `GET /api/workspaces/{slug}`
  - `PUT /api/workspaces/{slug}`
  - `POST /api/workspaces/{slug}/archive`
  - `GET /api/workspaces/{slug}/members`
  - `PUT /api/workspaces/{slug}/members/{user_id}`
  - `DELETE /api/workspaces/{slug}/members/{user_id}`
- Added workspace scope resolution from `?workspace=<slug>` across existing gateway flows.
- Added membership enforcement before workspace-scoped reads/writes.
- Archived workspaces now return `410 Gone` on scoped access paths instead of behaving like active workspaces.

### 4. Workspace scoping across existing feature areas

- Jobs:
  - workspace-scoped listing / summaries / visibility checks
- Routines:
  - workspace-aware lookup, listing, and ownership checks
- Memory:
  - workspace-aware resolver path and shared workspace backing scope
- Settings:
  - workspace and personal settings are now actually isolated
  - `?workspace=` no longer falls through to personal settings
- Conversations / thread listing:
  - workspace threads stay out of personal thread listings
  - conversation ownership checks now consider workspace scope

### 5. Runtime propagation

- Propagated `workspace_id` through the runtime instead of keeping it as a web-only concept:
  - `IncomingMessage`
  - chat send path
  - websocket path
  - approval resubmission path
  - session/pending approval state
  - job/routine context
  - conversation persistence / ownership checks
- Chat and approval flows now preserve workspace scope end-to-end.

### 6. Workspace event delivery

- Added workspace-wide SSE fan-out semantics.
- Workspace-scoped events are no longer only user-targeted events with an extra tag; they now broadcast to all subscribers scoped to the same workspace.
- Updated gateway response/status broadcasting so workspace events go to the workspace stream rather than only the initiating user.

### 7. OpenAI-compatible Responses API

- Made `/v1/responses` workspace-aware:
  - previous-response/thread resolution now verifies workspace ownership correctly
  - response fetches no longer assume personal scope
  - new Responses API requests can be created in a workspace scope
- This closes a real parity gap where the main chat flows had workspace support but the Responses API path still behaved as if `workspace_id = NULL`.

### 8. Compatibility behavior

- Personal workspaces still behave as before:
  - no `workspace` query param
  - `workspace_id = NULL`
  - user-scoped queries remain valid
- Legacy `workspace_read_scopes` was not removed outright, but it is now treated as compatibility behavior for the personal workspace path rather than the source of truth for shared workspace authorization.
- Shared workspace access is now membership-based.

## Tests and Verification

Targeted verification completed on this branch:

- `cargo check --all-features`
- `cargo test --test workspaces_integration`
- `cargo clippy --all-features --all-targets -- -D warnings`

The workspace integration coverage now exercises:

- workspace membership endpoints and access control
- archived workspace `410 Gone` behavior
- workspace-scoped chat forwarding carrying `workspace_id`
- workspace vs personal settings isolation
- workspace thread isolation from personal thread listing
- Responses API workspace ownership behavior

Additional unit/integration coverage was also updated where API signatures changed.

## Notable Implementation Notes

- The existing memory “Workspace” type remains the document/memory system; this PR adds workspace *entities* on top of it rather than replacing it.
- `WorkspacePool` now behaves correctly for shared workspace scopes and no longer treats legacy token scopes as the mechanism for workspace entity sharing.
- libSQL test setup was adjusted so ephemeral test databases are reliable across the backend’s per-operation connection model.

Closes #1607


ilblackdragon (commented):
## Code Review

### Security Concern — Webhook Guard Removal

The diff removes the multi-tenant isolation guard from `webhook_trigger_handler` (`src/channels/web/handlers/webhooks.rs`):

```rust
// REMOVED:
if state.workspace_pool.is_some() {
    return Err((StatusCode::GONE, "Unscoped webhooks disabled..."));
}
```

This guard prevents cross-user routine triggering in multi-tenant deployments. Removing it weakens tenant isolation — any attacker with a webhook secret can trigger routines across all users via the unscoped `/api/webhooks/{path}` endpoint. The doc comment is updated from "Disabled in multi-tenant mode" to advisory guidance, but nothing enforces that recommendation.

**Recommendation:** Keep the guard, or make unscoped webhooks in multi-tenant mode opt-in via config.

### Good Changes

- **`token_prefix()` helper** — properly uses `char_indices().nth(8)` to avoid panicking on multi-byte UTF-8. Applied consistently across `tokens.rs`, `users.rs`, and `main.rs`.
- **Role constants** (`ROLE_ADMIN`/`ROLE_MEMBER` + `is_valid_role()`) — reduces stringly-typed bugs.
- **`created_by` validation** — checks user existence before recording. However, the `_ => None` arm silently swallows DB errors — consider logging on `Err(e)`.

### Issues

1. **No tests** — `token_prefix()` is a pure function that's trivially testable. The `created_by` validation change also warrants a test. Per project rules, new code should include tests.
2. **PR scope mismatch** — title says "first-class workspace entities with membership and cross-workspace sharing" but the diff is incremental cleanups. The PR notes mention uncommitted changes — consider scoping the title to what's actually here.
3. **Whitespace-only changes** in `heartbeat.rs` and `cli/mod.rs` are cosmetic noise.
4. **Migration comment** removes "invitation flow" from the header — may be premature if that's planned for a follow-up.

zmanian (changes requested):
## Review -- REQUEST_CHANGES

Well-structured transition from implicit per-user isolation to explicit DB-backed workspace entities. The backward compatibility story (personal workspace = `workspace_id IS NULL`) is clean. Authorization model consistently applied across all handler types. Good integration test coverage.

### Must fix

**1. No role validation on member upsert** -- `body.role` is passed directly to DB with no validation. Any string accepted ("superadmin", "", etc.). Validate against `{owner, admin, member, viewer}`.

**2. Slug validation missing** -- `workspaces_create_handler` accepts any slug including special characters, spaces, path traversal. Slugs appear in URL paths. Validate with pattern like `^[a-z0-9][a-z0-9-]{1,62}[a-z0-9]$`.

**3. Last-owner protection** -- No guard preventing the sole owner from removing themselves, orphaning the workspace.

### Medium

**4. Admin can promote themselves to owner** via the upsert endpoint's `ON CONFLICT DO UPDATE`. Is this intentional?

**5. Workspace-scoped settings bypass key extraction** -- API keys in workspace LLM settings stored unencrypted (secrets store extraction skipped when `scope.is_some()`).

**6. Duplicate `resolve_chat_workspace_id`** -- identical function in both `chat.rs` and `server.rs`. Extract to shared location.

**7. `fire_manual(routine_id, None)` loses caller tracking** -- previously passed `Some(&user.user_id)` for audit trail.

### Low
- `list_workspaces_for_user` includes archived workspaces (intentional?)
- `row_to_workspace` uses `.parse().unwrap_or_default()` -- silently produces nil UUID on parse failure
- PostgreSQL migration drops/recreates unique indexes without data guards

ilblackdragon (commented):
# Code Review: First-class Workspace Entities

## Overview

This PR adds DB-backed workspace entities with membership, cross-workspace sharing, and scoped access across the entire stack — from schema through DB traits, API handlers, SSE event delivery, WebSocket connections, the Responses API, and runtime propagation. It's a large (~4.5k additions), ambitious, and well-structured change that touches nearly every subsystem.

## Strengths

- **Clean backward compatibility**: Personal workspace (`workspace_id = NULL`) continues to work transparently. Existing data requires no migration into stored workspace rows.
- **Consistent scope propagation**: `workspace_id` flows end-to-end through `IncomingMessage` → `JobContext` → `PendingApproval` → SSE events. This is done thoroughly.
- **Good handler refactoring**: Routines handlers extracted `load_visible_routine()` and `resolve_routine_scope()`, eliminating repeated fetch-then-check-ownership patterns.
- **SSE scope filtering**: The `event_matches_scope()` function and tests for workspace fan-out are well-designed — personal events stay personal, workspace events reach all workspace subscribers.
- **Both backends implemented**: PostgreSQL and libSQL both implement all new trait methods, meeting the dual-backend requirement.
- **New tests**: Workspace integration tests exercise membership, archived gating, scoped chat, settings isolation, thread isolation, and Responses API ownership.

## Issues

### Critical

1. **libSQL base schema FK ordering bug**: In `libsql_migrations.rs`, the `conversations`, `agent_jobs`, `memory_documents`, `routines`, and `settings` tables all have `workspace_id TEXT REFERENCES workspaces(id) ON DELETE CASCADE`, but the `workspaces` table is defined *after* them in the base schema. On a fresh libSQL database, this will fail because the referenced table doesn't exist when the FK is created. The `workspaces` and `workspace_members` table definitions must be moved before any table that references them.

2. **PostgreSQL `set_setting` race condition**: The PG `set_setting` was rewritten from `ON CONFLICT ... DO UPDATE` to `INSERT ... ON CONFLICT DO NOTHING` + `UPDATE`. Because the unique constraint was replaced with partial indexes, `ON CONFLICT` can't reference the old constraint. However, the single-setting path does **not** wrap in a transaction, so two concurrent writes could both succeed at `INSERT ON CONFLICT DO NOTHING` (neither sees the other's uncommitted row) before the `UPDATE` runs. Either wrap in a transaction or use `ON CONFLICT ON CONSTRAINT <index_name> DO UPDATE`.

3. **Missing incremental libSQL migration**: The workspace schema changes appear only in the base schema, not as incremental migration steps. Existing libSQL databases won't get the new `workspaces`/`workspace_members` tables or the `workspace_id` columns on existing tables. There needs to be a migration step for existing databases.

### Security

4. **Owner can be removed from workspace**: `remove_workspace_member` has no guard against removing the last owner. An admin could remove the sole owner, leaving an orphaned workspace with no owner. Add a check similar to `is_last_admin` in user management.

5. **No role validation on `add_workspace_member`**: The `workspace_members_upsert_handler` accepts any `body.role` string without validation. The `WorkspaceMemberWriteRequest` has just `pub role: String` — there's no check that it's one of the valid roles (`owner`/`admin`/`member`/`viewer`). The `is_valid_role()` helper was added in `users.rs` but not used here.

6. **Unscoped webhooks re-enabled in multi-tenant mode**: The guard in `webhook_trigger_handler` that rejected unscoped webhooks in multi-tenant mode was removed entirely. The comment says "For multi-tenant isolation, use the user-scoped endpoint" but the old guard prevented cross-user routine triggering. This is a regression — the per-routine webhook secret provides weak protection compared to tenant isolation.

### Correctness

7. **`conversation_belongs_to_user` workspace semantics**: When `workspace_id` is `Some(id)`, the query matches correctly. But `parsed_workspace_id()` in `thread_ops.rs` silently drops invalid workspace UUIDs (returns `None`). If a message has a malformed `workspace_id` string, it would be treated as a personal-scope check, potentially allowing access to personal conversations when workspace access was intended.

8. **`fire_manual` lost user verification**: In `routines_trigger_handler`, the call changed from `fire_manual(routine_id, Some(&user.user_id))` to `fire_manual(routine_id, None)`. While the handler now does `load_visible_routine` before firing, verify that this fully replaces the internal ownership check that the user_id parameter provided.

9. **Duplicate `resolve_chat_workspace_id` function**: This function is defined identically in both `src/channels/web/handlers/chat.rs` and `src/channels/web/server.rs`. Extract to a shared location to avoid divergence.

### Code Quality

10. **`#[allow(clippy::too_many_arguments)]` on `send_notification`**: Adding `workspace_id` pushed this to 8 parameters. Consider grouping into a struct instead of suppressing the lint.

11. **`workspace_settings_user_id` convention**: Both libSQL and PG settings backends use `format!("workspace:{workspace_id}")` as a synthetic `user_id` for workspace settings. This convention-based coupling is fragile — if a real user_id ever starts with `workspace:`, it would collide. Consider using a fixed sentinel or removing the user_id requirement for workspace-scoped settings.

12. **Repeated scope-resolution boilerplate**: Nearly every handler has the same `resolve_*_scope` → branch on workspace/personal pattern. Consider an axum extractor that resolves workspace scope once and injects it.

### Missing

13. **No slug validation**: `WorkspaceCreateRequest.slug` is accepted as-is with no validation (length, allowed characters, reserved words). Slugs are used in URLs — validate them.

14. **Settings key extraction skipped for workspace scope**: `settings_set_handler` skips API key extraction for workspace-scoped `llm_builtin_overrides` and `llm_custom_providers` (just uses `body.value.clone()`). This means workspace settings could store plaintext API keys that personal settings would encrypt.

## Summary

This is solid foundational work. The architecture is clean, the scope propagation is thorough, and backward compatibility is well-preserved. The critical items (libSQL schema ordering, PG settings race, missing incremental migration) need to be fixed before merge. The security items (owner removal, role validation, webhook guard) should be addressed.

serrrfirat:
Addressed the review items in `172729c1` and pushed them onto the PR branch `feat/1607-workspace-entities-db-users`.

Summary of fixes:
- Added workspace role validation, slug validation, archived-workspace filtering, last-owner protection, and owner-only owner-role mutations.
- Restored caller-aware authorization for manual routine triggers, restored the multi-tenant guard for unscoped webhooks, and fail-closed malformed `workspace_id` parsing in thread hydration.
- Removed the duplicate chat workspace resolver by consolidating onto a shared helper.
- Fixed workspace-scoped settings secret extraction so workspace LLM API keys are vaulted instead of stored in plaintext, and switched workspace settings storage away from the `workspace:{id}` user-id convention to a fixed sentinel.
- Refactored routine notifications to use a struct instead of `#[allow(clippy::too_many_arguments)]`.
- Fixed libSQL schema ordering for workspace FKs, added an idempotent incremental libSQL migration path for existing databases, and added regression coverage for fresh and legacy libSQL DBs.
- Wrapped the PostgreSQL single-setting write paths in transactions to close the insert/update race on the new partial-index scheme.
- Added tests for `token_prefix()`, `created_by` validation, workspace membership/owner/archive behavior, workspace secret scoping, the multi-tenant webhook guard, and libSQL migration behavior.
- Removed the whitespace-only diff noise that was called out.
- Replaced the browser `prompt()` token-name flow with inline UI.

Verification run:
- `cargo test --test workspaces_integration -- --nocapture`
- `cargo test --lib libsql_migrations::tests -- --nocapture`
- `cargo test --lib token_prefix -- --nocapture`
- `cargo test --lib workspace_scope -- --nocapture`
- `cargo test --lib unscoped_webhooks_are_disabled_in_multi_tenant_mode -- --nocapture`
- `cargo test --lib user_creation_ -- --nocapture`
- `cargo fmt --all`
- `git diff --check`

ilblackdragon (changes requested):
I found a few concrete regressions here, plus one larger architectural inconsistency in the workspace model.

1. High: workspace-scoped history authorization is bypassable for active in-memory threads.
In `src/channels/web/handlers/chat.rs`, `chat_history_handler()` correctly checks `conversation_belongs_to_user(thread_id, user_id, workspace_id)`, but on failure it only returns `404` if the thread is not present in `sess.threads`. It then falls through to the in-memory branch and returns that thread’s turns anyway. In practice, the same user can request a thread under the wrong `?workspace=` and still read it as long as that thread is active in session. That defeats the new workspace scoping check for active threads.

2. High: workspace-scoped SSE/WS subscribers no longer receive user-scoped auth events.
In `src/channels/web/sse.rs`, `event_matches_scope()` rejects `(Some(event_user), None)` whenever the subscriber has a workspace scope, because it requires `subscriber_workspace_id.is_none()`. But auth-required/auth-completed events are still emitted with `broadcast_for_user(...)` from both the HTTP and WS chat flows. That means a user connected to `/api/chat/events?workspace=...` or the workspace-scoped WS path will miss extension auth events in that workspace tab.

3. Medium: workspace-scoped settings dropped the active-provider safety invariant.
In `src/channels/web/handlers/settings.rs`, the guard that prevents removing the currently active custom provider only runs when `scope.is_none()`. Once `llm_backend` / `llm_custom_providers` are workspace-scoped, that invariant needs to hold there too. The helper itself is also still hard-wired to user-scoped `get_setting(...)`, so the feature surface has expanded without the corresponding safety rule.

4. Medium: the PR is halfway between a shared-workspace conversation model and a user-owned conversation model.
Live response/status events are now broadcast to the whole workspace in `src/channels/web/mod.rs`, but persisted conversation listing and ownership checks in the DB layer still key conversations by `user_id + workspace_id`. That leaves an unresolved product/security boundary: either workspace members can see live thread activity they cannot later list/open, or the storage/authorization model has not actually been generalized to match the new workspace-sharing behavior. I would want that model made explicit and enforced consistently before merging.

At a higher level, the implementation is extending workspace scope through routing and transport faster than through the underlying ownership model. That is exactly where subtle multi-tenant bugs show up. I’d recommend tightening the model first, then adding focused tests for:
- `chat/history` with the wrong workspace against an active thread
- workspace-scoped SSE/WS auth-required/auth-completed flows
- workspace-scoped `llm_custom_providers` updates/deletes when the active backend is custom

serrrfirat:
Addressed the April 3 review in `cfd72b7f` and pushed it to `feat/1607-workspace-entities-db-users`.

1. `chat/history` now fails closed on workspace mismatch before falling through to active in-memory session threads. This is fixed in the gateway handler path and covered by `workspace_chat_history_rejects_active_thread_from_wrong_workspace`.

2. Workspace-scoped auth events now preserve workspace scope end-to-end. Pending auth stores the thread workspace, and the HTTP, WS, OAuth callback, and extension setup completion paths now emit `auth_required` / `auth_completed` with `broadcast_for_user_in_workspace(...)` so the correct workspace tab receives them.

3. The active custom-provider safety guard now applies to workspace-scoped settings too. The guard reads `llm_backend` / `llm_custom_providers` from the resolved scope instead of assuming personal settings, and the workspace update/delete cases are covered by targeted tests.

4. I made the ownership model explicit for this PR by choosing the user-owned conversation path rather than shared workspace conversations. Live chat response/status events are no longer workspace-wide fan-out when a user scope exists; they are now scoped to `user_id + workspace_id`, which matches the persisted conversation listing / ownership model. I also updated the `/v1/responses` subscriber path to use the same scoped subscription so workspace requests still receive their own events.

Verification run on this follow-up:
- `cargo test --lib routing_scope -- --nocapture`
- `cargo test --lib no_silent_drop -- --nocapture`
- `cargo test --lib response_id_round_trip -- --nocapture`
- `cargo fmt --all`
- `git diff --check`

Earlier targeted checks for the other three items are still green on this branch as well:
- `cargo test --lib test_chat_auth_token_handler_broadcasts_workspace_scoped_auth_required -- --nocapture`
- `cargo test --lib test_handle_client_auth_token_broadcasts_workspace_scoped_auth_completed -- --nocapture`
- `cargo test --lib guard_active_provider_not_removed_for_workspace -- --nocapture`
- `cargo test --lib test_pending_auth_ -- --nocapture`
- `cargo test --test workspaces_integration workspace_chat_history_rejects_active_thread_from_wrong_workspace -- --nocapture`

ilblackdragon (commented):
## Automated review — NEEDS DISCUSSION (do not merge as-is)

**TL;DR:** Sound architecture and a good backward-compatibility story, but **four critical safety issues** plus three security/correctness items require explicit decisions before merge. The \"delicate\" flag is warranted — this touches IronClaw's most valuable data. Recommend additional reviewers (DB + security).

---

### Critical issues (blockers)

| 1 | **CRITICAL** | \`src/db/libsql_migrations.rs\` (base SCHEMA) | **FK ordering bug**. \`conversations\`/\`agent_jobs\`/\`memory_documents\`/\`routines\`/\`settings\` reference \`workspaces(id)\` before \`workspaces\` is created. Fresh libSQL databases will fail on migration. | Move \`workspaces\` + \`workspace_members\` CREATE TABLE to the top of SCHEMA, before \`conversations\`. Test on a fresh libSQL instance |
| 2 | **CRITICAL** | \`src/db/libsql_migrations.rs\` (\`INCREMENTAL_MIGRATIONS\`) | **Missing libSQL incremental migration**. The base SCHEMA has workspace tables but \`INCREMENTAL_MIGRATIONS\` has no entry — existing libSQL deployments won't get workspace tables or scoping columns | Add migration #18 with all workspace CREATE TABLEs + ALTER TABLEs for scoping columns. Per CLAUDE.md: \"All new persistence features must support both backends\" |
| 3 | **CRITICAL** | \`src/db/postgres.rs::set_setting\` + \`handlers/settings.rs\` | **Settings upsert race**. PG rewrite uses INSERT+UPDATE without a transaction wrapper. Two concurrent writes can both succeed at INSERT (neither sees the other's uncommitted row) before UPDATE runs | Wrap \`set_setting\` in a transaction, or use \`ON CONFLICT ON CONSTRAINT\` with the partial index name |
| 4 | **CRITICAL** | \`handlers/chat.rs\` + \`channels/web/mod.rs\` | **Authorization model incomplete**. \`chat_history_handler\` checks workspace scope but falls through to the in-memory active thread without re-checking scope. Workspace SSE fan-out broadcasts live events to all workspace subscribers, but conversation ownership remains user-keyed — members see live activity from threads they can't later list/query | Audit the conversation ownership model. Is it user-owned or workspace-shared? If shared, listing queries must be workspace-keyed. If user-owned, what should live SSE show? Make the model explicit in code + tests |

### High-severity issues

|---|---|---|---|---|
| 5 | **High** | \`handlers/workspaces.rs\` (\`workspace_members_upsert_handler\`) | No role validation — \`body.role\` accepted as-is; any string passes (\"superadmin\", \"\"). \`is_valid_role()\` helper already exists in \`users.rs\` | Validate against \`[\"owner\",\"admin\",\"member\",\"viewer\"]\`. Return 400 with an error. Add regression test |
| 6 | **High** | \`handlers/workspaces.rs\` | No slug validation. Slugs appear in URLs and accept any string (spaces, special chars, path-traversal attempts) | Validate: \`^[a-z0-9][a-z0-9-]{1,62}[a-z0-9]$\`. Return 400. Add regression test |
| 7 | **High** | \`src/db/workspace.rs::remove_workspace_member\` | No guard preventing removal of the last owner. An admin could delete the sole owner, orphaning the workspace | Count members with \`role='owner'\`; if count==1 and target is owner, return 409. Test: \`workspace_cannot_remove_last_owner\` |
| 8 | **High** | \`handlers/webhooks.rs\` | The \`if state.workspace_pool.is_some() { return Err(GONE, ...) }\` guard was removed entirely. Unscoped webhooks can now trigger routines across all users in multi-tenant mode. Doc was updated to \"advisory guidance\" but nothing enforces it | Either restore the guard or make unscoped webhooks opt-in via a config flag. Security regression from the perspective of multi-tenant isolation. Clarify intent with maintainers |

### Medium / Low

|---|---|---|---|---|
| 9 | Medium | \`handlers/chat.rs\` + \`channels/web/server.rs\` | \`resolve_chat_workspace_id\` defined identically in two files — will diverge | Extract to \`src/channels/web/scope.rs\` and import once |
| 10 | Medium | \`handlers/settings.rs\` | Settings key extraction (secret injection) is skipped for workspace-scoped settings. Workspace LLM API keys are stored **plaintext** while personal settings are encrypted | Apply the same secret extraction/encryption logic for workspace settings, or document explicitly why workspace members are trusted to see raw API keys |
| 11 | Medium | \`src/db/libsql/workspace.rs\` (\`row_to_workspace\`) | \`.parse().unwrap_or_default()\` on UUID silently produces nil UUID on malformed input | Use \`?\` error propagation; log on error |
| 12 | Medium | \`handlers/workspaces.rs\` | Admin can self-promote to owner via the upsert endpoint's \`ON CONFLICT DO UPDATE\` path | Clarify intent. If only owners should promote to owner: return 403 for non-owner attempts to set \`role='owner'\`. Test: \`workspace_member_cannot_self_promote_to_owner\` |
| 13 | Low | \`src/db/libsql/workspace.rs\` (\`list_workspaces_for_user\`) | Includes archived workspaces | Comment the rationale or filter with a test for both modes |
| 14 | Low | \`routine_engine.rs\` (\`fire_manual(routine_id, None)\`) | Lost user tracking (previously \`Some(&user.user_id)\`) | Verify \`load_visible_routine\` fully replaces internal checks and preserves the audit trail |

---

### Dual-backend support assessment

| Component | PostgreSQL | libSQL | Coverage |
|-----------|-----------|--------|----------|
| Schema | V16__workspaces.sql ✓ | Base SCHEMA + INCREMENTAL_MIGRATIONS ? | **PARTIAL** |
| FK ordering | Correct | **Broken** | **NO** |
| Concurrency | \`ON CONFLICT DO UPDATE\` | INSERT+UPDATE, no txn wrapper | **DIVERGENT** |
| Role validation | Validated at DB level | Validated at DB level | ✓ |
| Slug validation | Missing in both | Missing in both | ✗ |

The libSQL backend is incompletely implemented. Fresh libSQL init will fail; existing libSQL deployments silently lack workspace functionality.

### Test coverage — gaps

Present:
- \`tests/workspaces_integration.rs\` covers membership endpoints, archived 410, scoped chat forwarding, settings isolation, thread listing isolation, Responses API ownership, slug/role validation tests (\`workspace_creation_rejects_invalid_slug\`, \`workspace_member_updates_validate_roles_and_owner_permissions\`, \`workspace_cannot_remove_or_demote_last_owner\`)

Missing:
- SSE/WebSocket auth-event delivery test to workspace-scoped subscribers
- Active in-memory thread access-control bypass test
- Custom provider deletion guard on workspace-scoped settings
- **libSQL backend not tested with migration** — integration tests run on PostgreSQL; no freshness check for libSQL
- Concurrent workspace settings write race test

### Risk assessment (production)

1. **Deployment blocker (libSQL)**: fresh init fails; existing libSQL silently lacks workspace functionality
2. **Multi-tenant visibility regression**: SSE is workspace-wide but listing is user-keyed
3. **Settings race (libSQL)**: concurrent writes can create duplicates/overwrites (low probability)
4. **Webhook tenant isolation**: removal of the multi-tenant guard allows unscoped webhooks across user boundaries
5. **Unencrypted workspace API keys**: DB breach or log leak exposes them
6. **Authorization ambiguity**: no clear story for \"what does it mean to be a workspace member?\"

### Open questions for maintainer

1. **libSQL migration strategy**: include the \`INCREMENTAL_MIGRATIONS\` entry in this PR, or separate? How do we prevent deploying the base schema without the incremental migrations?
2. **Webhook multi-tenant isolation**: was removing the \`workspace_pool\` guard intentional?
3. **Conversation ownership**: when a conversation is in a workspace, is it user-owned (only the creating user sees it) or workspace-shared (all members see it)? Affects listing, history, ownership checks. Currently user-owned for DB queries but workspace-wide for SSE — inconsistent.
4. **Workspace settings encryption**: why skip secret extraction for workspace-scoped LLM settings? By design (members are trusted) or oversight?
5. **Self-promotion to owner**: is it intentional that an admin can promote themselves to owner via the upsert endpoint?

### Recommended additional reviewers
- **DB expert** — migration strategy, FK ordering, transaction safety, dual-backend equivalence
- **Security reviewer** — multi-tenant isolation (webhook guard removal, SSE scope filtering, settings encryption)
- **Product/design** — explicit conversation ownership model decision

---

### Conclusion
The foundational work is solid. Backward compatibility is clean, scope propagation is thorough, and the integration tests are well-structured. But this PR cannot merge without:

1. Fixing the four criticals
2. Resolving the security questions (last-owner verification in impl, webhook guard intent, workspace settings encryption policy)
3. Addressing the high-priority gaps (role validation, slug validation, duplicate function extraction)

The \"delicate\" flag from the maintainer is warranted — take the time to get it right. Consider a follow-up PR for non-critical improvements (e.g., the scope-resolution helper extraction).

zmanian (changes requested):
## Review

+6848/-735, 70+ files — comprehensive implementation of DB-backed workspace entities, membership, cross-workspace sharing, scoped data access, and workspace-aware SSE fan-out. Both PG and libSQL migrations included. Well-structured for the scope.

### Must Fix

1. **Dead/duplicate handlers in `handlers/chat.rs` vs `server.rs`** — `handlers/chat.rs` defines public `chat_send_handler`, `chat_approval_handler`, etc. that are never used in routing. `server.rs` has its own local versions that are actually wired up. This is a significant maintenance hazard — a reader won't know which handler is live. Should consolidate to one location.

2. **`workspace_id` type inconsistency: `String` vs `Uuid`** — `workspace_id` is `Option<String>` in `IncomingMessage`, `JobContext`, `PendingAuth`, `PendingApproval`, but `Option<Uuid>` in `Routine`, `MemoryDocument`, `SandboxJobRecord`, `AgentJobRecord`, and the DB layer. This leads to ~10+ scattered `Uuid::parse_str(id).ok()` conversions where a malformed UUID silently becomes `None`. Should parse once at the boundary and propagate as `Uuid` throughout.

3. **`joined_at` reset on member role update** — the UPSERT in both PG and libSQL resets `joined_at` to `now()` when updating an existing member's role. This loses historical data. Should only update `role` and `invited_by` on conflict.

### Suggestions

4. **Operator precedence in `validate_workspace_slug`** — the `||`/`&&` chain is correct by accident (De Morgan's). Add explicit parens for clarity: `!(bytes[0].is_ascii_lowercase() || bytes[0].is_ascii_digit())`.

5. **Dual sentinel schemes** — `WORKSPACE_SETTINGS_SENTINEL_USER_ID` (`__workspace_settings__`) vs `workspace_scope_user_id()` (`workspace:{uuid}`) serve different purposes but could confuse future contributors. Add a clarifying comment.

6. **Webhook multi-tenant guard ordering** — the check moved from `webhook_trigger_handler` into `fire_webhook_inner`, meaning rate limit now runs before the GONE rejection. Minor, but in a DDoS scenario the ordering matters.

7. **No Postgres test coverage** — integration tests all use `LibSqlBackend::new_memory()`. The Postgres `WorkspaceMgmtStore` implementation is untested.

8. **No way to list archived workspaces** — `list_workspaces_for_user` filters them out with no `include_archived` option for audit/cleanup.

### Cross-Cutting Observations

This PR and others in the current review batch share a **dual-backend testing gap** — multiple PRs only test libSQL, not Postgres, which could let PG-specific bugs slip through. Worth establishing a pattern for testing both backends on persistence-heavy PRs like this one.

### Code Style

No `.unwrap()`/`.expect()` in production code, proper error mapping with context, `thiserror` types, strong types (`WorkspaceRecord`, `WorkspaceMembership`, `ResolvedWorkspace`), consistent `crate::` imports. The `RoutineNotification` struct replacing `#[allow(clippy::too_many_arguments)]` is a nice refactor.

ilblackdragon (commented):
**Verdict: Request changes** (foundational PR — #1977 stacks on this)

- **`src/channels/web/handlers/workspaces.rs:88-101` `validate_workspace_slug`** — operator-precedence + boolean logic does not faithfully implement the documented `^[a-z0-9][a-z0-9-]{1,62}[a-z0-9]$` regex. Rewrite as a clear `.all()` pass plus first/last anchors.
- **Routine viewer privilege**: `routines_trigger_handler` only calls `load_visible_routine` (read check). Workspace `viewer`s can therefore execute routines via the gateway. Likely unintended.
- **`chat_approval_handler` TOCTOU**: reads `pending.workspace_id` from in-memory session but doesn't re-check membership at submission time. A user removed between request and approval can still complete it. Re-check membership against `pending.workspace_id`.
- **`workspace_members_upsert_handler` TOCTOU**: `get_member_role` / `is_last_workspace_owner` / `add_workspace_member` is not transactional — two concurrent owner-removal upserts can both pass. Wrap in a single SQL transaction or conditional UPDATE.
- **PG query planner risk**: hot lookups now use `workspace_id IS NOT DISTINCT FROM $N`, which PG generally cannot use a btree index for. `EXPLAIN ANALYZE` `list_conversations_all_channels` on a populated DB; consider splitting personal vs workspace queries or rewriting as `(workspace_id = $N OR (workspace_id IS NULL AND $N IS NULL))`.
- **`WORKSPACE_SETTINGS_SENTINEL_USER_ID = "__workspace_settings__"`** — if a real user ever ends up with this id, they could read all workspace settings. Add a CHECK constraint or user-creation guard.
- **`migrations/V16__workspaces.sql:158-160`** drops `settings_pkey` leaving the table with NO primary key. Acceptable for partial-unique-index design but call out in release notes; logical replication tools may misbehave.
- **No down migration**. Document the rollback plan.
- **`src/db/postgres.rs:~2480` `set_setting`** — uses `INSERT … ON CONFLICT DO NOTHING` then `UPDATE` instead of an inferred-target upsert. Functionally correct but fragile if a partial index is missing.
- Strong dual-backend implementation throughout. Migration safety is mostly additive. Excellent integration test suite (`tests/workspaces_integration.rs`, +715 lines) plus libSQL migration upgrade tests.
- Per-message `Uuid::parse_str(workspace_id)` happens in many hot paths — consider storing as `Uuid` end-to-end in a follow-up.

serrrfirat:
drifted and superseded already

PR #1523, opened 2026-03-21
feat(routines): surface notifications in chat context

G7CNF:
Persist routine notifications as channel-scoped records and inject unread ones into the next user turn as a transient system message. This keeps them out of fake turns while making them available to the LLM when the user replies.\n\nValidation:\n- cargo fmt --all -- --check\n- cargo check -q\n- cargo clippy -p ironclaw --lib --all-features -- -D warnings\n- cargo test -p ironclaw test_conversation_notifications_scope_and_consumption -- --nocapture

PR #1738, opened 2026-03-29
feat(routines): agent review — inject routine results into agentic loop

j-bloggs:
## Summary
- Add per-routine `agent_review_on_success/failure/attention` flags that inject routine completion results into the agent's agentic loop for LLM processing
- Replace `is_internal: bool` on `IncomingMessage` with a typed `MessageSource` enum (`User`, `Internal`, `RoutineReview`) for clean three-way dispatch
- `RoutineReview` messages skip SubmissionParser, hooks, and event triggers but enter the full agentic loop with tool access (minus routine management tools via AUTONOMOUS_TOOL_DENYLIST)
- Route responses via broadcast to the routine's notify_channel (not channel.respond, since "routine-review" is not a real channel)
- Show `notification_mode` in routine_list/create/update output for feature discoverability

## Change Type
- [x] New feature (non-breaking change which adds functionality)

## Linked Issue
None — originated from observing that routine failures were invisible to the agent. Design spec: `docs/plans/2026-03-29-routine-agent-review-design.md`

## Validation
- Targeted tests: 124 lib tests pass (channels, job_monitor, routine_engine, dispatcher, routine tools, cli routines)
- libSQL tests: 33 pass
- Live integration: verified agent review triggers on routine failure, ICG diagnoses and reports via Telegram

## Security Impact
- `RoutineReview` messages bypass SubmissionParser (prevents routine output being interpreted as /undo, /stop commands)
- Tool filtering via AUTONOMOUS_TOOL_DENYLIST prevents recursive loops (agent review cannot call routine_create/fire/update/delete)
- Routine result sanitized through SafetyLayer::sanitize_tool_output before injection
- Thread isolation: dedicated "routine-review" thread, never blocks user's active conversation

## Database Impact
- PostgreSQL V15: adds 3 BOOLEAN columns to routines table (DEFAULT false, backward-compatible)
- libSQL: columns added to idempotent CREATE TABLE + migration marker

## Blast Radius
- `IncomingMessage.is_internal` field replaced with `source: MessageSource` enum — all direct field access sites updated, backward-compat accessor provided
- `RoutineEngine::new()` gains an `inject_tx` parameter — all call sites updated (agent_loop + 9 test sites)
- `NotifyConfig` gains 3 fields — all 14 construction sites updated
- Existing routines unaffected (agent_review defaults to false)

## Rollback Plan
- **Immediate:** Set `ROUTINE_AGENT_REVIEW_ENABLED=false` env var to disable globally
- **Per-routine:** Set `agent_review_on_*=false` on individual routines
- **Full revert:** Columns default to false, injection logic is purely additive. Removing the code leaves pre-feature behavior.

## Review Track
**Track C** — changes to src/agent/ (agent_loop, dispatcher, routine_engine), database schema

## Feature Parity
No FEATURE_PARITY.md changes needed — agent review is a new capability not tracked in existing parity matrix.

## Follow-up Work
- Web gateway UI toggles for agent_review_on_* flags (currently only configurable via routine_create/update tools)
- Integration with PR #1523 (conversation_notifications) — consume notifications after successful agent review to prevent duplicate lazy injection
- Per-routine agent review rate limiting (currently bounded by thread serialization + CostGuard)


zmanian (changes requested):
## Request Changes

Good architecture -- `MessageSource` enum replacing the boolean `is_internal` flag is clean, and using the established `inject_sender()` pattern is the right call. However, there are cost control gaps that CLAUDE.md explicitly requires.

### Must-fix

1. **No rate limit or circuit breaker for agent review LLM calls.** CLAUDE.md requires "per-turn or per-session cap" and "circuit breaker" for any new LLM call path. A routine failing every 60s with `agent_review_on_failure=true` will inject a full agentic loop (with tool access) each time. CostGuard enforces a daily ceiling but doesn't prevent routine reviews from consuming the entire budget before any user interaction. Add: `max_agent_reviews_per_hour` config and a circuit breaker after N consecutive reviews.

2. **`process_user_input` call signature mismatch** -- RoutineReview branch passes a `tenant` parameter not in the current signature. Missing diff hunk or compilation error?

### Should-fix

3. RoutineReview responses skip BeforeOutbound hooks (safety/compliance won't run)
4. Prompt injection surface -- routine output sanitized as tool output but injected as user message
5. All routine reviews share a single thread -- use `routine_id` as conversation scope
6. Invalid `notify_channel` silently falls back to `broadcast_all`

### Nice-to-have

7. No unit tests for `inject_agent_review`
8. Default `agent_review_enabled: true` is aggressive before rate limiting is in place

j-bloggs:
Addressed all feedback from zmanian:

**Must-fix:**
1. Rate limit: `max_agent_reviews_per_hour` config (default 10, env `ROUTINES_MAX_AGENT_REVIEWS_PER_HOUR`) + circuit breaker (3 consecutive failures disables review). Circuit breaker now resets on successful non-review routine completion (275e7c73).
2. Default `agent_review_enabled` changed to `false`

**Should-fix:**
3. BeforeOutbound hooks run on RoutineReview responses before broadcast
4. Per-routine thread isolation: `routine_id` as conversation_scope gives each routine its own review thread
5. Invalid `notify_channel` logs warning before broadcast_all fallback

**Nice-to-have:**
6. Added 6 unit tests for inject_agent_review: status gating, config gate, default disabled, rate limit, circuit breaker

Note: rate limit uses a tumbling hourly window (resets on hour boundary), not a true sliding window. Acceptable for the default of 10/hour.

Validation: `cargo fmt`, clippy (all-features + default + libsql), all routine_engine tests pass.

zmanian (changes requested):
## Review -- REQUEST_CHANGES

The architecture is solid -- MessageSource enum, inject_sender pattern, tool denylist, and safety sanitization are well-designed. Rate limiting and circuit breaker additions address previous review items. However:

### Critical

**1. `Ok(None)` from SubmissionResult shuts down the agent**
In `handle_message`, the RoutineReview branch maps `Ok(SubmissionResult::Ok { message: msg }) => Ok(msg)`. When `msg` is `None`, this returns `Ok(None)` from `handle_message`. The main event loop treats `Ok(None)` as a shutdown signal (`/quit`), breaking out of the loop entirely. A routine review that produces a silent/no-output `SubmissionResult::Ok { message: None }` will terminate the agent.

Fix: `Ok(SubmissionResult::Ok { message: msg }) => Ok(Some(msg.unwrap_or_default()))` or handle `None` explicitly.

### Medium

**2. Circuit breaker tracks the wrong failure mode** -- `agent_review_consecutive_failures` only increments on `tx.send()` failure (mpsc channel send), not on LLM-level failures from the agent review. If injection succeeds but the downstream LLM call fails repeatedly, the circuit breaker never trips.

**3. Rate limiting TOCTOU race** -- The load-check-store on `agent_review_window_start` and count are not atomic. Concurrent `tokio::spawn` tasks for full_job routines can both pass the rate limit. Use `compare_exchange` for the window reset, or accept with a comment.

**4. Prompt injection surface** -- Routine output is sanitized via `sanitize_tool_output` then injected as user message content entering `process_user_input`. This bypasses `BeforeInbound` hooks by design. Consider marking injected content as system context rather than user input.

**5. Tests are shallow** -- Tests verify boolean flag gating and atomic counter logic but never exercise `inject_agent_review` itself. No tests for: routine completion injecting a message, message reaching the agent loop, response broadcast, or tool denylist enforcement.

### Required
- Fix the `Ok(None)` shutdown bug
- Address circuit breaker tracking (at minimum document the limitation)
- Add integration test for `inject_agent_review`

src/agent/agent_loop.rs
```diff
@@ -893,43 +894,83 @@ impl Agent {
...
+                        // "routine-review" is not a real channel.
+                        let notify_channel = message
+                            .metadata
+                            .get("notify_channel")
+                            .and_then(|v| v.as_str())
+                            .filter(|s| !s.is_empty());
+
+                        let outgoing = OutgoingResponse {
+                            content: response,
+                            thread_id: None,
```
serrrfirat: **High Severity**

Routine-review responses are broadcast with `thread_id: None` and only `message.user_id` as the target. That drops gateway delivery entirely because `gateway.broadcast()` rejects responses without a thread ID, and it also bypasses the existing `notify_user` / owner-scope routing used by normal routine notifications. This path needs to carry the resolved notify target and routine conversation thread through to the outbound response instead of rebuilding a stripped-down `OutgoingResponse` here.

src/agent/agent_loop.rs
```diff
@@ -1058,18 +1099,58 @@ impl Agent {
...
+                // Skip parser, hooks, event triggers, auth. Resolve thread and
+                // enter agentic loop directly.
+                tracing::info!(
+                    message_id = %message.id,
+                    routine_name = message.metadata.get("routine_name")
+                        .and_then(|v| v.as_str()).unwrap_or("unknown"),
+                    "Processing routine review"
+                );
+
+                let (session, thread_id) = self
```
serrrfirat, line 1264: **Medium Severity**

This early-return path skips the normal `set_message_tool_context()` call below, but `dispatch_chat()` still seeds `JobContext.metadata` from the synthetic incoming message. That means a routine-review turn advertises `notify_channel = "routine-review"`, so if the model uses the `message` tool for an attachment or proactive follow-up, the tool treats the fake channel as authoritative and the send fails unless the model explicitly overrides the channel and target. Please either set the message-tool context before entering this branch or override the execution metadata for routine-review turns.

src/agent/agent_loop.rs
```diff
@@ -1058,18 +1099,58 @@ impl Agent {
...
+
+                let result = self
+                    .process_user_input(message, tenant, session, thread_id, &message.content)
+                    .await;
+
+                return match result {
+                    Ok(SubmissionResult::Response { content }) => Ok(Some(content)),
+                    Ok(SubmissionResult::Ok { message: msg }) => Ok(msg),
+                    Ok(SubmissionResult::Error { message: err }) => Ok(Some(err)),
+                    Ok(_) => {
```
serrrfirat, line 1305: **High Severity**

This branch collapses `SubmissionResult::NeedApproval` into an empty response. During a routine-review turn, approval-needed status is sent to the synthetic `routine-review` channel, which does not exist, so the user never sees the approval prompt and the thread stays stuck in `AwaitingApproval`. Either forbid approval-requiring tools in `RoutineReview` turns or reroute approvals through a real channel.

src/db/libsql_migrations.rs
```diff
@@ -787,6 +793,15 @@ CREATE INDEX IF NOT EXISTS idx_api_tokens_user ON api_tokens(user_id);
 CREATE INDEX IF NOT EXISTS idx_api_tokens_hash ON api_tokens(token_hash);
 "#,
     ),
+    (
+        15,
```
serrrfirat, line 865: **High Severity**

This migration is a no-op, but the new libSQL read/write paths already query `agent_review_on_*`. Fresh databases are fine because the idempotent `CREATE TABLE` now includes those columns, but any existing database that already applied migration 13/14 will skip the rebuild and then hit `no such column` on routine load/update. Migration 15 needs real `ALTER TABLE ADD COLUMN` steps (or another rebuild) plus an upgrade-path test.

serrrfirat (changes requested):
Found 4 concrete issues in the routine-review path:

1. libSQL migration 15 is a no-op, so upgraded databases never receive the new `agent_review_on_*` columns even though the read/write paths already query them.
2. Routine-review responses rebuild `OutgoingResponse` with `thread_id: None` and target `message.user_id`, which breaks gateway delivery and skips the normal owner-scoped notify target resolution.
3. `SubmissionResult::NeedApproval` is collapsed into an empty response for routine-review turns, so approval-required tools can wedge the thread in `AwaitingApproval` with no visible prompt.
4. The routine-review fast path skips `set_message_tool_context()`, but `dispatch_chat()` still seeds execution metadata from the synthetic message, so the `message` tool inherits `notify_channel = "routine-review"` and misroutes proactive sends.

I also ran `cargo test --test e2e_routine_heartbeat`; that target currently fails 4 full-job owner-tool tests in this checkout. I have not tied those failures to this diff, so I am not counting them as branch-specific findings here.

src/agent/agent_loop.rs
j-bloggs: Hopefully addressed as per comment here: https://github.com/nearai/ironclaw/pull/1738#issuecomment-4152808226

src/agent/agent_loop.rs
j-bloggs: Hopefully addressed as per comment here: https://github.com/nearai/ironclaw/pull/1738#issuecomment-4152808226

src/agent/agent_loop.rs
j-bloggs: Hopefully addressed as per comment here: https://github.com/nearai/ironclaw/pull/1738#issuecomment-4152808226

src/db/libsql_migrations.rs
j-bloggs: https://github.com/nearai/ironclaw/pull/1738#issuecomment-4152808226

src/agent/routine_engine.rs
```diff
@@ -1885,6 +1921,126 @@ async fn send_notification(
...
+    );
+
+    let msg = IncomingMessage::new("routine-review", routine.user_id.clone(), content)
+        .with_metadata(serde_json::json!({
+            "source": "routine_completion",
+            "routine_id": routine.id.to_string(),
+            "routine_name": routine.name,
+            "status": status.to_string(),
+            "run_id": run.id.to_string(),
+            "notify_channel": routine.notify.channel,
```
serrrfirat, line 2234: **High Severity**

**Missing `notify_user` in `inject_agent_review` metadata — agent review responses target the wrong user when routines have a custom notification target.**

The `send_notification` function (line ~1872) includes both `"notify_user": notify.user` and `"notify_channel": notify.channel` in its metadata. But `inject_agent_review` only includes `notify_channel`:

```rust
.with_metadata(serde_json::json!({
    "source": "routine_completion",
    "routine_id": routine.id.to_string(),
    "routine_name": routine.name,
    "status": status.to_string(),
    "run_id": run.id.to_string(),
    "notify_channel": routine.notify.channel,
    // ← notify_user is missing
}))
```

The agent loop's RoutineReview handler reads `notify_user` from metadata in **two places**:

1. `set_message_tool_context` (line ~1182): falls back to `message.user_id` (routine owner)
2. Response routing (line ~934): `target_user` falls back to `message.user_id`

So when a routine has `notify.user = Some("team-lead")`, both the `message` tool context and the broadcast target silently fall back to the routine **owner** instead of the intended recipient.

**Suggested fix:**

```rust
.with_metadata(serde_json::json!({
    "source": "routine_completion",
    "routine_id": routine.id.to_string(),
    "routine_name": routine.name,
    "status": status.to_string(),
    "run_id": run.id.to_string(),
    "notify_channel": routine.notify.channel,
    "notify_user": routine.notify.user,
}))
```

zmanian (dismissed):
## Re-Review -- APPROVE

All critical and high-severity findings from the previous round are addressed:

**Critical #1 (Ok(None) shutdown):** Fixed. `unwrap_or_default()` converts None to empty string, catch-all returns `Ok(Some(String::new()))`. Dispatcher also auto-denies approval-requiring tools during routine review turns as defense-in-depth.

**serrrfirat HIGH #1 (libSQL migration):** Fixed. `run_migration_v15_fixup()` uses `pragma_table_info` to conditionally add columns. Two regression tests (add + idempotency).

**serrrfirat HIGH #2 (gateway delivery):** Fixed. `OutgoingResponse.thread_id` now populated from `notify_thread_id` metadata.

**Circuit breaker (Medium #2):** Acknowledged with honest comment -- tracks mpsc send failures, not LLM failures. Reasonable tradeoff; rate limiter provides secondary safety net.

**Rate limit TOCTOU (Medium #3):** Acknowledged with comment -- soft limit of 10/hour, overshooting by 1-2 is harmless.

**Tests:** Improved with 2 genuine integration tests (real mpsc channel, rate limit exercised) alongside unit tests.

CI green across all configurations.

src/agent/routine_engine.rs
j-bloggs: Fixed in 6fec4186 — added `"notify_user": routine.notify.user` to the metadata. Good catch.

zmanian (dismissed):
## Re-Review (v3) -- APPROVE

Three new commits since last approval, all clean:

1. **notify_user metadata fix** (6fec4186) -- Correct. Adds `notify_user` alongside `notify_channel` in `inject_agent_review` metadata, fixing response routing for routines with custom notification targets.
2. **Integration tests** (940f78f9) -- Two substantive tests exercising real injection path with mpsc channel, LibSqlBackend, and SafetyLayer. Covers failure injection + rate limiting.
3. **Struct initialization cleanup** (a60bacae) -- Clippy lint fix, no behavioral change.

CI green across all configurations. No regressions.

serrrfirat:
resolve conflicts and lets merge

j-bloggs:
Rebased onto latest staging (27a2fab1), conflicts resolved:
- libSQL migration renumbered from V15 to V17 (upstream took V15+V16)
- Using new idempotent `IDEMPOTENT_ADD_COLUMN_MIGRATIONS` pattern from upstream
- Merged upstream's channel/user fallback logic with our agent_review fields in routine_create
- PostgreSQL migration renumbered to V16
- Fixed upstream's new test NotifyConfig construction site
- Added test module imports for Arc/AtomicU32/AtomicUsize/RoutineRun

All 35 routine_engine tests pass. Clippy clean.

zmanian (dismissed):
## Review -- APPROVE

This is a well-structured feature with good defensive design. The previous review rounds caught the critical issues (Ok(None) shutdown, libSQL migration gap, response routing) and they have all been addressed. The `MessageSource` enum replacing `is_internal: bool` is a clean, extensible pattern.

### Remaining items (non-blocking)

**1. Duplicate `#[test]` attribute (src/agent/routine_engine.rs)**
There is a dangling doc comment and duplicate `#[test]` attribute around the `test_agent_review_disabled_by_default` test:

```rust
    /// Verify that agent review only triggers when the matching `agent_review_on_*`
    /// flag is set for the given status, and that `agent_review_enabled` gates it.
    #[test]
    /// Verify that the default config has `agent_review_enabled = false`.
    #[test]
    fn test_agent_review_disabled_by_default() {
```

The first `#[test]` and its doc comment are orphaned -- they appear to be a leftover from a test that was removed or never completed. Rust compiles this fine (the second `#[test]` wins) but it is confusing to read. Remove the orphaned `#[test]` + doc comment.

**2. `notify_thread_id` is never populated**
The response routing in `agent_loop.rs` reads `notify_thread_id` from message metadata, but `inject_agent_review` never sets this field. So `OutgoingResponse.thread_id` will always be `None` for routine review responses. This means gateway channels that require a thread_id for delivery will not receive routine review responses. This is probably fine for now (broadcast_all will attempt all channels), but worth a follow-up if you want targeted thread delivery for routine reviews.

**3. Circuit breaker resets on non-review success may be too aggressive**
The circuit breaker resets `agent_review_consecutive_failures` to 0 on any `RunStatus::Ok` routine completion, even when `should_review` is false. This means a failing routine that triggers reviews alternating with a healthy routine that does not will never trip the breaker. This seems intentional based on the comment ("reset after normal operation resumes") but worth noting -- the breaker protects against the inject channel being broken, not against LLM failures, so this is acceptable.

### What looks good

- `MessageSource` enum is a clean replacement for `is_internal: bool` with backward-compat accessor
- Tool denylist (`AUTONOMOUS_TOOL_DENYLIST`) prevents recursive routine creation during review turns
- Auto-deny of approval-requiring tools during RoutineReview prevents thread wedging
- Rate limiting (hourly tumbling window) with documented TOCTOU tradeoff
- Circuit breaker with documented scope limitation (mpsc send failures only)
- Safety sanitization via `sanitize_tool_output` before injection
- `agent_review_enabled` defaults to false -- opt-in only
- Dual-backend DB changes (PostgreSQL V16 migration + libSQL V17 migration + V15 fixup for existing DBs)
- Integration tests cover real injection path and rate limiting
- Conversation scope isolation via `routine_id`

CI reportedly green. Ship it.

j-bloggs:
Thank you for the thorough review and approval!

Re non-blocking items:
1. **Duplicate #[test]** — checked the current head, no duplicate attributes found. May have been resolved when the shallow unit tests were removed.
2. **notify_thread_id never populated** — noted for follow-up. Gateway channels requiring thread_id will fallback to broadcast_all.
3. **Circuit breaker reset on non-review success** — intentional, documented.

j-bloggs:
Hey team! 👋 Just rebased this onto the latest staging and resolved all conflicts (migration renumbered to V18→V21, merged the new event_cache + ownership model changes). All tests pass and clippy is clean.

Would love a fresh look when you get a chance! @zmanian @serrrfirat

serrrfirat (commented):
1. **High** — queued routine-review turns can get stuck permanently, and the user can receive the internal queue ack instead of the actual review.

At [src/agent/agent_loop.rs:1241](/Users/firatsertgoz/Documents/worktrees/pr-1738-feat-routines-agent-review-inject-routine-results-into-agentic-loop/src/agent/agent_loop.rs#L1241) the `RoutineReview` branch returns immediately after a single `process_user_input()` call. But [src/agent/thread_ops.rs:263](/Users/firatsertgoz/Documents/worktrees/pr-1738-feat-routines-agent-review-inject-routine-results-into-agentic-loop/src/agent/thread_ops.rs#L263) queues follow-up input whenever the thread is already `Processing` and returns `SubmissionResult::Ok { message: Some("Message queued ...") }`.

That drain only exists in the normal user path. If two routine reviews hit the same routine close together, the second one is queued but never drained, and the outer routine-review handler can surface the queue ack as the outbound text. This is a correctness regression in concurrent routine completions.

2. **High** — plain-text routine-review replies bypass the normal notification routing safeguards, so delivery can fail for gateway and owner-scoped channels.

The injected review message omits `notify_thread_id` at [src/agent/routine_engine.rs:2219](/Users/firatsertgoz/Documents/worktrees/pr-1738-feat-routines-agent-review-inject-routine-results-into-agentic-loop/src/agent/routine_engine.rs#L2219), while the outbound routine-review path sends to `notify_user.unwrap_or(message.user_id)` directly at [src/agent/agent_loop.rs:948](/Users/firatsertgoz/Documents/worktrees/pr-1738-feat-routines-agent-review-inject-routine-results-into-agentic-loop/src/agent/agent_loop.rs#L948) instead of reusing the normal `resolve_routine_notification_target()` logic used by standard routine notifications at [src/agent/agent_loop.rs:784](/Users/firatsertgoz/Documents/worktrees/pr-1738-feat-routines-agent-review-inject-routine-results-into-agentic-loop/src/agent/agent_loop.rs#L784).

Consequences:
- `gateway` delivery can fail because `broadcast()` requires a thread id.
- owner-scoped channels can fail or misroute whenever `notify_user` is omitted and the target normally comes from `ExtensionManager::notification_target_for_channel()`.

3. **Medium** — the `message` tool still receives execution metadata for the synthetic `routine-review` channel, so tool-based follow-ups can target the wrong channel.

The routine-review branch sets the shared message-tool context to the real notify channel/user at [src/agent/agent_loop.rs:1211](/Users/firatsertgoz/Documents/worktrees/pr-1738-feat-routines-agent-review-inject-routine-results-into-agentic-loop/src/agent/agent_loop.rs#L1211), but `ChatDelegate` then rebuilds `job_ctx.metadata` from `chat_tool_execution_metadata(message)` at [src/agent/dispatcher.rs:199](/Users/firatsertgoz/Documents/worktrees/pr-1738-feat-routines-agent-review-inject-routine-results-into-agentic-loop/src/agent/dispatcher.rs#L199). That helper hard-codes `notify_channel = message.channel` at [src/agent/agent_loop.rs:129](/Users/firatsertgoz/Documents/worktrees/pr-1738-feat-routines-agent-review-inject-routine-results-into-agentic-loop/src/agent/agent_loop.rs#L129), which is `"routine-review"` for these turns.

Because execution metadata is authoritative in `message`, a review that uses the `message` tool can try to send to the non-existent `routine-review` channel or the wrong target.

4. **Medium** — the new rate limit and circuit breaker are global to the entire `RoutineEngine`, so one tenant/routine can suppress agent review for everyone else in the process.

The new state is stored once on `RoutineEngine` at [src/agent/routine_engine.rs:123](/Users/firatsertgoz/Documents/worktrees/pr-1738-feat-routines-agent-review-inject-routine-results-into-agentic-loop/src/agent/routine_engine.rs#L123), and the logic at [src/agent/routine_engine.rs:2154](/Users/firatsertgoz/Documents/worktrees/pr-1738-feat-routines-agent-review-inject-routine-results-into-agentic-loop/src/agent/routine_engine.rs#L2154) applies that shared counter/breaker to all injected reviews. In a multi-tenant deployment, one noisy tenant can exhaust the hourly review budget or trip the breaker for unrelated tenants.

Missing coverage:
- No end-to-end test that a routine-review reply reaches `gateway` or an owner-scoped channel when `notify_user` is omitted.
- No concurrency test covering two routine reviews for the same routine while the first one is still processing.
- No test where a routine-review turn uses the `message` tool, which is where the routing metadata bug shows up.

j-bloggs:
Thanks for the thorough review @serrrfirat! All 4 findings addressed in c461a148:

**1. Queue drain (High)** — Routine-review turns now detect the "Message queued" ack and suppress it (returns empty instead of surfacing internal queue state to the user). Also added explicit `NeedApproval` handling that auto-denies to prevent stuck threads.

**2. Notification routing (High)** — Switched to `resolve_routine_notification_target()` for the routine-review response path, matching the normal routine notification flow (extension manager lookup → owner-scope fallback). Also fixed `&str` vs owned `String` for broadcast calls.

**3. Message tool context (Medium)** ��� Fixed `chat_tool_execution_metadata()` to extract `notify_channel`/`notify_user`/`notify_thread_id` from `message.metadata` for RoutineReview messages instead of using the synthetic `"routine-review"` channel name. The message tool will now target the correct channel.

**4. Rate limit scope (Medium)** — Documented as a known limitation with doc comments explaining the global-per-engine scope. Per-routine tracking (HashMap under lock) is a worthwhile follow-up but out of scope for this PR.

All 35 routine_engine tests pass, clippy clean, fmt clean.

standardtoaster:
Heads up this will need a rebase — the v2 engine and ownership model both landed on staging and touch `IncomingMessage` and `channels/channel.rs`.

I think there's something interesting here around letting the agent self-heal (routine fails, agent diagnoses and retries or escalates). Right now the implementation gives the agent full tool access on routine results, which is flexible but open-ended. Have you thought about how to scope what the agent does with the result? Like whether skills or the routine config itself would guide the response, or some kind of budget/retry limit? Curious what use case you're building toward.

j-bloggs:
Superseded by #2046 (fresh rebase onto v2 engine + ownership model). All previous review feedback incorporated.

PR #1748, opened 2026-03-30
fix(builder): accept inline-table and object-map dependency formats from LLM

### LINKED ISSUE #1640: build_software crashing on Cargo.toml dependencies: "invalid type: map, expected a string"
Hey, I was trying to get the agent to build a simple Rust WASM tool, but it keeps getting stuck in a failure loop because of a TOML parsing error.

Even when I explicitly prompt it to use simple string dependencies (like ureq = "2"), the internal builder seems to generate an inline table anyway (like ureq = { version = "2" }). Then its own parser crashes trying to read it.

Here's the error it throws:
Tool call failed tool=build_software elapsed_ms=4981 error=Execution failed: Analysis failed: Tool builder failed: Failed to parse requirement: invalid type: map, expected a string at line 6 column 16

Here was the prompt I used to trigger it:

    "Build a Rust WASM tool that searches GitHub repositories by keyword and returns the top 5 results with name, description, and star count as JSON. Use only simple string dependencies in Cargo.toml — no inline tables or feature flags. Use ureq = "2" for HTTP requests and serde_json = "1" for JSON. Do not use reqwest."

Let me know if you need any more logs to track this down.

anthhub:
## Summary

`build_software` crashes with `"invalid type: map, expected a string"` when the LLM returns dependencies as TOML inline tables (`{"ureq": {"version": "2"}}`) instead of simple strings (`"ureq = \"2\""`). This is a common LLM output pattern that the builder should handle gracefully.

- Added custom `deserialize_dependencies()` that accepts three formats:
  - String array: `["ureq = \"2\""]` (existing)
  - Object array: `[{"ureq": {"version": "2"}}]` (new)
  - Object map: `{"ureq": "2", "serde_json": "1"}` (new)
- All formats are normalized to TOML-compatible `name = "version"` strings
- `null` dependencies deserialize to empty vec

Fixes #1640

## Test plan

- [x] `test_build_requirement_deserialize_inline_table_deps` — regression for the reported crash
- [x] `test_build_requirement_deserialize_object_map_deps` — flat object format
- [x] `test_build_requirement_deserialize_null_deps` — null handling
- [x] All 20 existing builder tests pass
- [x] `cargo check` / `cargo fmt`


zmanian (dismissed):
## Review -- APPROVE

Clean, focused fix for a real problem. The custom serde deserializer handles string arrays, object arrays, and object maps correctly. Since the dependency strings are consumed by an LLM in a subsequent build prompt (not a TOML parser), the "best-effort" formatting is pragmatically fine.

### Minor notes (non-blocking)
- `flatten_dep` produces approximate TOML for nested objects (JSON `Display` for deeply nested values). Worth a brief comment noting this is for LLM consumption, not machine-parsed TOML.
- The `other` fallback arm silently formats non-string values (`true`, `2`) as TOML dependency syntax -- low risk given the LLM consumer.
- No `.unwrap()` in production code, proper `de::Error::custom()` usage, good test coverage.

zmanian (dismissed):
## Re-Review -- APPROVE (reconfirmed)

New commit 380fa681 improves edge case handling: `flatten_dep` now returns `Option<String>`, skipping invalid value types (null, bool, number) with a `tracing::warn` instead of producing garbage TOML. Clean improvement, no regressions.

ilblackdragon (approved):

PR #1751, opened 2026-03-30
feat(engine): learning system improvements — confidence decay, dedup, /learn skill

ilblackdragon:
## Summary

Adapts key ideas from [gstack's learning system](https://github.com/garrytan/gstack/pull/622) to IronClaw's v2 engine.

### Retrieval engine improvements

- **Confidence decay**: Older memory docs score lower in context retrieval via exponential decay (half-life ~2 years). Prevents stale observations from polluting context while preserving hard-won lessons (Specs and Lessons have a decay floor of 0.3).
- **Title-based dedup**: When multiple docs share a title (e.g., a corrected learning superseding an older one), only the most recently updated survives at retrieval time. Enables append-only corrections without write-time coordination.
- Both compose cleanly: `score = (keyword_relevance + type_weight) × recency_decay`

### `/learn` management skill

User-facing interface for browsing and curating the knowledge base:
- "show learnings" — browse extracted skills, lessons, insights
- "learning stats" — counts, usage metrics, top performers
- "prune learnings" — identify stale entries
- "export learnings" — structured markdown export
- "learning quality" — assess which learnings are actually helping

### What this doesn't include (future work)

- **Cross-project learning** — reading docs from other projects (needs Store trait changes)
- **Explicit confidence scores** on MemoryDocs (gstack uses 1-10 per entry)
- **False positive database** — structured FP tracking for review findings

## Test plan

- [x] 4 new unit tests for `recency_factor` and `dedup_by_title`
- [x] All 203 engine tests pass
- [x] `/learn` skill frontmatter validates against parser limits
- [ ] Manual: verify retrieval prefers recent docs over stale ones
- [ ] Manual: verify duplicate-titled docs are deduplicated


crates/ironclaw_engine/src/memory/retrieval.rs
```diff
@@ -113,6 +125,56 @@ fn keyword_match_score(doc: &MemoryDoc, keywords: &[String]) -> f64 {
...
+    for doc in docs {
+        let key = doc.title.clone();
+        match by_title.get(&key) {
+            Some(existing) if existing.updated_at >= doc.updated_at => {
+                // Keep existing (newer or equal)
+            }
+            _ => {
+                by_title.insert(key, doc);
+            }
+        }
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in c9b7b783 — switched to `HashMap::entry` API for single-lookup dedup.

skills/learn/SKILL.md
```diff
@@ -0,0 +1,189 @@
...
+Manage the knowledge the agent has accumulated from completed work. The v2 engine automatically extracts skills, lessons, and insights from threads — this skill lets the user browse, curate, and export that knowledge.
+
+## Commands
+
+### Show learnings
+
+User says: "what have you learned?", "show lessons", "browse learnings"
+
+**Action:**
+1. `memory_search("lesson OR skill OR insight OR pattern")` to find learning-related docs
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in c9b7b783 — clarified that these are FTS content searches, not DocType filters. Split the combined OR query into separate `memory_search` calls per doc type. Added a naming conventions section documenting the `skill:<name>` and `insight:<category>:<topic>` title prefixes.

skills/learn/SKILL.md
```diff
@@ -0,0 +1,189 @@
...
+Patterns detected from conversation analysis.
+- **<insight title>** — <summary> (detected <date>)
+```
+
+### Stats
+
+User says: "learning stats", "knowledge base stats"
+
+**Action:**
+1. `memory_search("lesson")` + `memory_search("skill:")` + `memory_search("insight:")` to count by type
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in c9b7b783 — the `skill:` prefix is the engine's naming convention for extracted skill titles (e.g. `skill:github-issue-triage`). Added a naming conventions section to the skill that documents this and the `insight:<category>:<topic>` pattern.

ilblackdragon:
## Code Review

**Verdict: Architecturally sound, dedup compliant, needs polish**

Adds exponential recency decay, title-based dedup, and a `/learn` skill on top of the new `ironclaw_engine` memory retrieval pipeline.

### Decay correctness
Math checks out. λ=0.001, half-life = ln(2)/0.001 ≈ 693d ≈ 1.9y (doc says ~2y, fine). Spot values match the doc comment (`retrieval.rs:187-192`). Type-aware floors are sensible.

**Edge:** `recency_factor` uses `Utc::now()` directly (`retrieval.rs:210`), which makes it non-deterministic and harder to test — inject a clock or pass `now` in. The "user_stated" override silently bypasses the type floor and confidence factor — fine, but worth a comment about why this beats `confidence`.

### Dedup correctness
`dedup_by_title` (`retrieval.rs:168-183`) is a **read-time** filter only — nothing is deleted from the store, so this does **NOT** violate the "LLM data is never deleted" rule. ✅

**Caveat:** title is the dedup key with no normalization (case, whitespace, trailing punctuation) — a corrected lesson titled `"web_search alias "` won't dedupe against `"web_search alias"`. Recommend `.trim().to_lowercase()` as the key.

### Retention rule
Compliant. Dedup is purely a retrieval-time view; the store keeps all versions. Worth one comment in the docstring stating this explicitly so future maintainers don't "optimize" it into a delete.

### Convention adherence
- No `unwrap`/`expect` in prod paths ✅. `partial_cmp(...).unwrap_or(Equal)` is the right pattern (`retrieval.rs:114`).
- Prompt templates correctly live under `crates/ironclaw_engine/prompts/*.md` per CLAUDE.md.
- `learn_skill.md` is at `skills/learn/SKILL.md` — correct location.
- Confidence default of `10.0` (`retrieval.rs:240`) means missing metadata = full confidence; backwards-compatible but worth a constant.

### Test coverage
4 new unit tests as advertised (`retrieval.rs:434-534`) — recent, old-decays, lesson-floor, user-stated, plus dedup keep-latest and keep-different. **Missing:**
- Decay+dedup composed (does the latest-by-`updated_at` survivor still respect floors?).
- Tie-breaking when two docs share title AND `updated_at`.
- Confidence=0 metadata clamp interaction with `user_stated` short-circuit.

### Top suggestions
1. **`retrieval.rs:210`** — inject clock (or accept `now: DateTime<Utc>`) for determinism and easier property tests.
2. **`retrieval.rs:171`** — normalize title key (`trim().to_lowercase()`) before dedup, otherwise corrections silently fail to supersede.
3. **`retrieval.rs:168`** — add explicit comment: "Read-time only; never deletes from store (see CLAUDE.md retention rule)."
4. **`retrieval.rs:213`** — extract `HALF_LIFE_DAYS`, `LAMBDA`, type floors as named consts; current magic `0.001` is doc-coupled.
5. **Add a composed test:** old lesson with same title as recent note → dedup picks recent note, losing the floor benefit. Confirm intended.
6. **PR is mis-scoped at 35k lines;** the actual learning changes are tiny. Rebase or split before un-drafting so reviewers can see the signal.

crates/ironclaw_engine/src/memory/retrieval.rs
```diff
@@ -0,0 +1,729 @@
...
+    ) -> Result<Vec<MemoryDoc>, EngineError> {
+        let all_docs = self.store.list_memory_docs(project_id).await?;
+        self.score_and_rank(all_docs, query, max_docs)
+    }
+
+    /// Retrieve docs from the primary project plus up to
+    /// [`CROSS_PROJECT_MAX_SOURCES`] other projects.
+    ///
+    /// Cross-project docs are tagged in metadata with `"cross_project": true`
+    /// and receive a 0.5× penalty to prefer local knowledge.
```
serrrfirat, line 51: **Critical Severity — Documented 0.5x cross-project penalty not implemented**

Doc comment says cross-project docs "receive a 0.5× penalty to prefer local knowledge" but the code never applies any penalty. Foreign docs go through `score_and_rank()` with identical scoring as local docs. A foreign project with newer docs will rank above local docs, silently overriding local knowledge.

**Suggested fix:** Either apply the 0.5× penalty by tagging foreign docs and multiplying their score, or remove the false claim from the doc comment.

crates/ironclaw_engine/src/memory/retrieval.rs
```diff
@@ -0,0 +1,729 @@
...
+/// Deduplicate docs by title, keeping the most recently updated for each title.
+///
+/// When multiple docs share a title (e.g., a corrected learning superseding an
+/// older one), only the latest survives. This provides read-time dedup without
+/// requiring write-time coordination — corrections are appended as new docs
+/// with the same title, and stale versions are filtered here.
+fn dedup_by_title(docs: Vec<MemoryDoc>) -> Vec<MemoryDoc> {
+    let mut by_title: HashMap<String, MemoryDoc> = HashMap::new();
+    for doc in docs {
+        match by_title.entry(doc.title.clone()) {
```
serrrfirat, line 171: **High Severity — Case-sensitive title dedup defeats corrections**

`dedup_by_title()` uses `doc.title.clone()` as HashMap key (case-sensitive). "Web Search Lesson" and "web search lesson" are treated as different docs and both survive dedup. Defeats the stated purpose of read-time dedup for corrected learnings.

**Suggested fix:** Normalize key: `by_title.entry(doc.title.to_lowercase())`.

crates/ironclaw_engine/src/memory/retrieval.rs
```diff
@@ -0,0 +1,729 @@
...
+        &self,
+        project_id: ProjectId,
+        query: &str,
+        max_docs: usize,
+    ) -> Result<Vec<MemoryDoc>, EngineError> {
+        // Primary project docs
+        let mut all_docs = self.store.list_memory_docs(project_id).await?;
+
+        // Gather docs from other projects (only Lessons and Skills transfer well)
+        let projects = self.store.list_projects().await?;
```
serrrfirat, line 62: **High Severity — Cross-project retrieval lacks tenant authorization**

`retrieve_context_cross_project()` calls `self.store.list_projects()` to enumerate all projects, then reads docs from up to 5 other projects with no ownership check. In multi-tenant deployment, this leaks knowledge across tenant boundaries.

**Suggested fix:** Add `project_owner` field and filter to same user/tenant. Or require explicit opt-in list of shared project IDs.

src/bridge/router.rs
```diff
@@ -0,0 +1,2528 @@
...
+    /// V1 database for writing conversation messages (gateway reads from here).
+    db: Option<Arc<dyn Database>>,
+    /// Secrets store for storing credentials after auth flow.
+    secrets_store: Option<Arc<dyn crate::secrets::SecretsStore + Send + Sync>>,
+    /// Centralized auth manager for setup instruction lookup and credential checks.
+    auth_manager: Option<Arc<AuthManager>>,
+}
+
+/// Global engine state, initialized on first use.
+static ENGINE_STATE: OnceLock<RwLock<Option<EngineState>>> = OnceLock::new();
```
serrrfirat, line 97: **High Severity — Global mutable state via `OnceLock<RwLock<Option<EngineState>>>`**

Engine state in a process-global static makes the engine untestable in parallel, prevents multi-instance, and creates implicit coupling. First `Agent` to initialize wins.

**Suggested fix:** Pass `EngineState` as `Arc<EngineState>` through the `Agent` struct instead of global.

crates/ironclaw_engine/src/memory/retrieval.rs
```diff
@@ -0,0 +1,729 @@
...
+/// Confidence can be set by:
+/// - Learning missions (extracted observations start at 0.5-0.8)
+/// - User feedback (confirmed findings get boosted to 0.9-1.0)
+/// - User dismissals (false positives get dropped to 0.1-0.3)
+fn confidence_factor(doc: &MemoryDoc) -> f64 {
+    let raw = doc
+        .metadata
+        .get("confidence")
+        .and_then(|v| v.as_f64())
+        .unwrap_or(10.0); // Default: full confidence for docs without explicit score
```
serrrfirat, line 240: **High Severity — Default confidence (10.0) outranks explicit user entries (9.0)**

`confidence_factor()` defaults to `10.0` (full confidence) when no metadata is present. The `/learn` skill sets user-stated at `9` and observed at `6`. All legacy docs without explicit confidence score higher (1.0) than user-stated docs (0.9).

**Suggested fix:** Default to `7.0` (0.7 factor) so explicitly-rated docs can rank both above and below legacy docs.

src/agent/agent_loop.rs
```diff
@@ -378,48 +410,72 @@ impl Agent {
...
-            };
-            let available = guard.skills();
-            let skills_cfg = &self.deps.skills_config;
-            let selected = crate::skills::prefilter_skills(
-                message_content,
-                available,
-                skills_cfg.max_active_skills,
-                skills_cfg.max_context_tokens,
-            );
+    ) -> (Vec<ironclaw_skills::LoadedSkill>, String) {
```
serrrfirat, line 428: **Medium Severity — Breaking return type change**

`select_active_skills` now returns `(Vec<LoadedSkill>, String)` instead of `Vec<LoadedSkill>`. All existing callers in v1 code paths must be updated.

**Suggested fix:** Verify all call sites handle the tuple return correctly.

src/agent/agent_loop.rs
```diff
@@ -189,8 +189,8 @@ pub struct AgentDeps {
 /// The main agent that coordinates all components.
 pub struct Agent {
     pub(super) config: AgentConfig,
-    pub(super) deps: AgentDeps,
-    pub(super) channels: Arc<ChannelManager>,
+    pub(crate) deps: AgentDeps,
```
serrrfirat, line 192: **Medium Severity — `pub(super)` widened to `pub(crate)` on agent internals**

`deps` and `channels` plus multiple accessor methods widened from `pub(super)` to `pub(crate)`. Exposes Agent internals to entire crate including bridge.

**Suggested fix:** Add comment documenting why visibility was widened. Consider `BridgeHandle` trait to narrow surface.

serrrfirat (commented):
Code review: 9 findings (1 Critical, 4 High, 4 Medium) across retrieval engine, bridge router, and agent loop.

crates/ironclaw_engine/src/memory/retrieval.rs
```diff
@@ -0,0 +1,729 @@
...
+        let engine = RetrievalEngine::new(store);
+
+        let docs = engine
+            .retrieve_context_cross_project(project_a, "lesson", 10)
+            .await
+            .unwrap();
+        // Should get both the local and foreign lesson, but NOT the foreign note
+        assert!(
+            docs.len() >= 2,
+            "should get at least 2 docs, got {}",
```
serrrfirat, line 721: **Medium Severity — Cross-project test doesn't verify isolation**

Test checks foreign lessons appear but doesn't verify foreign Notes are excluded. Assertion uses `docs.len() >= 2` instead of exact count. While line 727 does check `!titles.contains("Foreign note")`, the `>= 2` assertion on line 721 masks false positives — if the foreign note leaks through, the test still passes as long as total count is >= 2.

**Suggested fix:** Assert exact count is 2: `assert_eq!(docs.len(), 2)` to catch leaks.

crates/ironclaw_engine/src/memory/retrieval.rs
```diff
@@ -0,0 +1,729 @@
...
+    matched as f64 / max_score as f64
+}
+
+/// Deduplicate docs by title, keeping the most recently updated for each title.
+///
+/// When multiple docs share a title (e.g., a corrected learning superseding an
+/// older one), only the latest survives. This provides read-time dedup without
+/// requiring write-time coordination — corrections are appended as new docs
+/// with the same title, and stale versions are filtered here.
+fn dedup_by_title(docs: Vec<MemoryDoc>) -> Vec<MemoryDoc> {
```
serrrfirat, line 168: **Medium Severity — `dedup_by_title` silently drops docs with no logging**

When two docs share a title, the older is silently dropped. No way to recover the original if the correction has a bug.

**Suggested fix:** Log at `debug!` level when deduplicating, including both doc IDs and timestamps.

serrrfirat:
@ilblackdragon do we still want this?

Issue #1334, opened 2026-03-18
Tool schema discovery: surface http policy and parameter semantics beyond raw schema


henrypark133:
## Summary
`http` has a large amount of behavior that matters to successful tool use but is not well expressed by raw parameter schema alone: HTTPS-only policy, SSRF restrictions, redirect behavior, `save_to` semantics, body/header formats, and credential injection behavior.

## Problem
The current schema can describe the JSON shape, but not the operational contract that actually determines success:
- only HTTPS is allowed
- localhost/private IP destinations are blocked
- `save_to` must remain under `/tmp`
- request-body and header shapes have important edge cases
- credential injection behavior is policy-driven, not obvious from the schema

This makes `http` a poor fit for schema-only discoverability.

## Proposed Direction
- Add richer discovery guidance for `http` via `tool_info`
- Summarize policy constraints separately from raw parameter schema
- Add targeted examples for common safe uses: GET JSON, download to `/tmp`, JSON body POST
- Consider a middle-detail discovery response that includes policy notes and supported body/header forms

## Why It Matters
When the model misses these rules, the resulting UX is noisy:
- blocked requests feel mysterious
- file downloads fail for path-policy reasons
- malformed headers/body payloads create avoidable retries

## Coverage Requirement
- Add regressions covering discovery output for `http` policy constraints
- Add an end-to-end trace where the model uses discovery guidance before a successful `http` call involving `save_to` or structured headers/body
- Add regression coverage so any new `tool_info` summary continues to mention HTTPS and safe-path restrictions

## References
- `src/tools/builtin/http.rs`
- `src/tools/builtin/tool_info.rs`
- `docs/plans/2026-03-17-tool-schema-discovery-audit.md`

henrypark133:
Putting this on hold — see nearai/ironclaw#1758 (comment): https://github.com/nearai/ironclaw/pull/1758#issuecomment-4166440613

Issue #1330, opened 2026-03-18
Tool schema discovery: expose message routing and attachment semantics more clearly


henrypark133:
## Summary
`message` looks simple at the schema level, but its actual authoring contract depends on contextual routing defaults, channel-specific target semantics, and attachment path constraints. The current schema does not make those runtime rules easy for the model to understand.

## Problem
The JSON shape is straightforward, but the important behavior is semantic:
- `channel` and `target` can be omitted and resolved from context
- routing behavior changes when explicit channel/target are absent
- attachment paths must validate under allowed roots or `/tmp`
- target format expectations vary by channel (Telegram, Signal, Slack)

This is exactly the sort of tool where raw JSON Schema under-explains the real usage model.

## Proposed Direction
- Add richer discovery guidance for `message` through `tool_info`
- Surface omission/default-routing behavior explicitly
- Add concise examples for common cases: reply in current channel, send to explicit channel, send with attachment
- Consider a middle-detail discovery mode focused on routing semantics rather than raw schema only

## Why It Matters
Poor `message` authoring creates visible UX failures:
- messages go to the wrong place
- models over-specify or omit `target` incorrectly
- attachment attempts fail for path-policy reasons the schema does not explain

## Coverage Requirement
- Add regressions covering default-routing behavior in discovery output
- Add an end-to-end trace for `message` with omitted channel/target and another with explicit routing
- Add coverage for attachment guidance so any new discovery summary preserves allowed-path semantics

## References
- `src/tools/builtin/message.rs`
- `src/tools/builtin/tool_info.rs`
- `docs/plans/2026-03-17-tool-schema-discovery-audit.md`

henrypark133:
Putting this on hold — see nearai/ironclaw#1758 (comment): https://github.com/nearai/ironclaw/pull/1758#issuecomment-4166440613

PR #1758, opened 2026-03-30
feat(tool_info): add curated summaries for selected built-in tools

### LINKED ISSUE #1332: Tool schema discovery: make time operation-specific requirements legible
## Summary
`time` is an operation-dispatch tool with different parameter requirements for `now`, `parse`, `convert`, `format`, and `diff`, but its current schema is presented as one flat object. That makes the tool valid yet harder than necessary for the model to use correctly.

## Problem
The current schema only strongly requires `operation`, while the rest of the contract is conditional in implementation:
- `convert` requires `to_timezone`
- `diff` requires `timestamp2`
- `parse` / `format` / `convert` each interpret timezone fields differently
- `input` and `timestamp` are aliases with backward-compatibility semantics

This makes `time` a good fit for a richer discovery surface rather than relying on the raw flat schema alone.

## Proposed Direction
- Add a richer `discovery_schema()` for `time` that groups parameters by operation
- Expose operation-specific requirements and examples through `tool_info`
- Consider a middle-detail discovery mode that summarizes required fields and aliases per operation

## Why It Matters
Incorrect `time` calls create avoidable friction:
- missing `to_timezone` or `timestamp2`
- confused use of `timezone` vs `from_timezone`
- unnecessary trial-and-error for otherwise simple requests

## Coverage Requirement
- Add regressions covering `tool_info` discovery for `time` across at least `convert` and `diff`
- Add an end-to-end trace showing the model discovers the right operation-specific fields before calling `time`
- Add coverage for any new discovery summary format so required-field guidance cannot silently regress

## References
- `src/tools/builtin/time.rs`
- `src/tools/builtin/tool_info.rs`
- `docs/plans/2026-03-17-tool-schema-discovery-audit.md`

ShadowCorp-Dev:
## Summary

- Adds curated `discovery_summary()` guidance for four selected built-in tools: `time`, `http`, `message`, and `create_job`
- Explicitly scopes the rollout: `tool_info(detail: "summary")` is curated for selected tools today, and falls back to schema-derived required fields for the rest
- Makes `tool_info` report `summary_source` (`curated` vs `schema_fallback`) so callers can tell which kind of summary they received
- Only advertises `detail: "summary"` in tool descriptions when curated guidance actually exists; schema-only tools now point to `detail: "schema"`

## Change Type

- [ ] Bug fix
- [x] New feature
- [x] Refactor
- [ ] Documentation
- [ ] CI/Infrastructure
- [ ] Security
- [ ] Dependencies

## Linked Issue

Closes #1332, #1334, #1330, #1331

## Validation

- [x] `cargo fmt`
- [x] Relevant tests pass:
  - `cargo test --lib tool_info`
  - `cargo test --lib schema_hint`
  - `cargo test --test e2e_builtin_tool_coverage tool_info_discovery`
- [x] Manual testing: deployed branch summary coverage remains limited to the selected tools listed above

## Security Impact

None

## Database Impact

None

## Blast Radius

- Curated summaries added in: `src/tools/builtin/http.rs`, `src/tools/builtin/job.rs`, `src/tools/builtin/message.rs`, `src/tools/builtin/time.rs`
- Scope clarification and summary-source reporting in: `src/tools/builtin/tool_info.rs`, `src/tools/tool.rs`, `src/tools/registry.rs`
- Test coverage in: `tests/e2e_builtin_tool_coverage.rs`

## Rollback Plan

- Revert the selected-tool summary additions independently if needed
- Revert the scope-clarification commit separately if we later broaden curated coverage across more tools

---

**Review track**: B

### DIFF
--- a/src/db/libsql/users.rs
+++ b/src/db/libsql/users.rs
@@ -981,7 +981,7 @@ mod tests {
         assert!(alice_stats.last_active_at.is_some());
 
         // Bob has no LLM calls so doesn't appear in summary stats
-        assert!(stats.iter().find(|s| s.user_id == "bob").is_none());
+        assert!(!stats.iter().any(|s| s.user_id == "bob"));
 
         // Filter to single user
         let alice_only = db.user_summary_stats(Some("alice")).await.unwrap();
--- a/src/tools/builtin/http.rs
+++ b/src/tools/builtin/http.rs
@@ -12,7 +12,9 @@ use reqwest::Client;
 use crate::context::JobContext;
 use crate::safety::LeakDetector;
 use crate::secrets::SecretsStore;
-use crate::tools::tool::{ApprovalRequirement, Tool, ToolError, ToolOutput, require_str};
+use crate::tools::tool::{
+    ApprovalRequirement, Tool, ToolDiscoverySummary, ToolError, ToolOutput, require_str,
+};
 use crate::tools::wasm::{InjectedCredentials, SharedCredentialRegistry, inject_credential};
 
 #[cfg(feature = "html-to-markdown")]
@@ -46,6 +48,31 @@ const USER_AGENT: &str = concat!(
     env!("CARGO_PKG_VERSION"),
     " (https://github.com/nearai/ironclaw)"
 );
+
+fn http_tool_summary() -> ToolDiscoverySummary {
+    ToolDiscoverySummary {
+        always_required: vec!["url".into()],
+        conditional_requirements: vec![
+            "'body' is typical for POST, PUT, and PATCH requests.".into(),
+            "'save_to' must be a path under /tmp/ (for binary downloads).".into(),
+        ],
+        notes: vec![
+            "HTTPS only — http:// URLs are rejected.".into(),
+            "localhost, private IPs, and cloud metadata endpoints are blocked (SSRF protection).".into(),
+            "Headers accept either object {\"Name\": \"Value\"} or array [{\"name\": \"N\", \"value\": \"V\"}] format.".into(),
+            "timeout_secs defaults to 30, maximum 300.".into(),
+            "Response body capped at 5 MB (text) or 50 MB (save_to file downloads).".into(),
+            "Credentials in headers (Authorization, X-API-Key, etc.) or URL query params trigger approval.".into(),
+            "GET with no credentials or custom headers does not require approval.".into(),
+            "Redirects are followed only for simple GETs (max 3 hops); non-GET redirects are blocked.".into(),
+        ],
+        examples: vec![
+            serde_json::json!({"method": "GET", "url": "https://api.example.com/data"}),
+            serde_json::json!({"method": "POST", "url": "https://api.example.com/items", "body": {"name": "test"}, "headers": {"Authorization": "Bearer token"}}),
+            serde_json::json!({"method": "GET", "url": "https://example.com/photo.jpg", "save_to": "/tmp/photo.jpg"}),
+        ],
+    }
+}
 
 /// Tool for making HTTP requests.
 ///
@@ -829,6 +856,10 @@ impl Tool for HttpTool {
         true // External data always needs sanitization
     }
 
+    fn discovery_summary(&self) -> Option<ToolDiscoverySummary> {
+        Some(http_tool_summary())
+    }
+
     fn requires_approval(&self, params: &serde_json::Value) -> ApprovalRequirement {
         let has_credentials = crate::safety::params_contain_manual_credentials(params)
             || (self.credential_registry.as_ref().is_some_and(|registry| {
@@ -1482,5 +1513,16 @@ mod tests {
     fn test_save_to_rejects_bare_tmp() {
         let err = validate_save_to_path("/tmp").unwrap_err();
         assert!(err.to_string().contains("must be under /tmp/"));
+    }
+
+    #[test]
+    fn http_discovery_summary_explains_policies_and_constraints() {
+        let summary = http_tool_summary();
+        assert_eq!(summary.always_required, vec!["url".to_string()]);
+        assert!(summary.notes.iter().any(|n| n.contains("HTTPS only")));
+        assert!(summary.notes.iter().any(|n| n.contains("SSRF")));
+        assert!(summary.notes.iter().any(|n| n.contains("save_to")));
+        assert!(summary.notes.iter().any(|n| n.contains("approval")));
+        assert_eq!(summary.examples.len(), 3);
     }
 }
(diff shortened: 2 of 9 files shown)


zmanian (changes requested):
## Review -- REQUEST_CHANGES

Good addition -- the summaries are well-calibrated and the factual claims I verified (HTTPS-only, SSRF protection, size limits, redirect hops, credential patterns, denied env vars) are accurate.

### Medium (blocking)

**Inaccurate rate limit in `create_job` summary** -- `src/tools/builtin/job.rs`: the summary says "Rate limit: 5 job creations per 30-second window" but `ToolRateLimitConfig::new(5, 30)` means 5 per minute, 30 per hour. Fix to: "Rate limit: 5 per minute, 30 per hour."

Since LLMs take these summaries at face value, accuracy is critical.

### Low (non-blocking)
- `http` summary says "simple GETs" for redirect following -- more precisely "GETs with no custom headers and no body"
- Tests verify string presence but don't cross-check stated values against actual tool config (e.g., rate limit numbers vs `rate_limit_config()`)

zmanian (dismissed):
## Re-Review -- APPROVE

Rate limit text fixed: now correctly says "5 per minute, 30 per hour" matching `ToolRateLimitConfig::new(5, 30)`. Test assertion updated accordingly.

Gemini bot's remaining comments (conditional_requirements naming, operation default) are minor style suggestions -- author's discretion.

standardtoaster:
I don't think this should merge as-is. The summary content is good — the struct, the per-tool guidance, the conditional requirements — but the delivery mechanism doesn't work. I pulled this branch, built it locally, and ran `tool_info` end-to-end through IronClaw's agent loop on Qwen3.5-35B (12 scenarios across all 4 summary-enabled tools, 2 runs each). The model never called `tool_info` — 0/24 runs — it saw the auto-appended hint in the tool description and ignored it every time.

| Scenario | Target | Summary helpful? | tool_info called? | Correct tool? |
|----------|--------|:-:|:-:|:-:|
| http_post_json | http | Y | **no** | miss |
| http_get_simple | http | - | **no** | ok |
| http_download_file | http | Y | **no** | ok |
| http_put_update | http | Y | **no** | miss |
| time_current | time | - | **no** | ok |
| time_timezone | time | Y | **no** | ok |
| message_telegram | message | Y | **no** | miss |
| message_simple_reply | message | - | **no** | miss |
| message_with_attachment | message | Y | **no** | miss |
| message_slack_channel | message | Y | **no** | miss |
| job_background_task | create_job | Y | **no** | ok |
| job_with_container | create_job | Y | **no** | miss |

This isn't a gap in the implementation — the code is clean and the summaries are well-researched (zmanian's review on factual accuracy was thorough). The issue is that current models don't follow self-referential tool hints. I've been trying to solve a similar problem in my fork (80+ registered tools, need to get usage guidance to the model reliably) and the core finding from many iterations of benchmarking is that **LLMs are tool users, not tool discoverers** — they use tools they can see and won't proactively look up guidance, no matter how you surface the hint.

This held across ~500 additional test cases using a similar `discover_tools` meta-tool on Qwen3.5-35B, Sonnet, Haiku 4.5, and Coder-30B:

| Config (Qwen3.5-35B, 5 configs × 15 scenarios × 2 runs) | Tools | Accuracy | discovery usage |
|----------------------------------------------------------|-------|----------|----------------|
| All tools visible (control) | ~35 | 80% | n/a |
| All tools + discovery available | ~36 | 75% | **0%** |
| Core only + discovery (targets hidden) | ~8 | 20% | **0%** |
| Core only, no discovery | ~7 | 20% | n/a |
| Core + discovery + strong system prompt | ~8 | 20% | **0%** |
([failed attempt](https://github.com/standardtoaster/ironclaw/tree/feat/discover-tools))
Even LoRA fine-tuning on 128 discovery-chain examples produced zero behavioral change on held-out tasks.

**What I'd suggest keeping:**
- The `ToolDiscoverySummary` struct and the `discovery_summary()` trait method — good extension point
- The per-tool summary content (conditional requirements, policy notes, examples) — well-researched and accurate

**What I'd suggest changing:**
- Drop the auto-appended hint in tool descriptions — it costs tokens on every turn for something models ignore
- Surface the summary content in **validation error responses** instead. When a model calls `http` with wrong parameters, return the conditional requirements and examples alongside the schema. The model gets guidance at the point of failure, without needing to know to ask

I tested this delivery approach with compressed tool descriptions and it matched or outperformed full-schema-on-every-turn at up to 120 tools ([benchmark and implementation](https://github.com/standardtoaster/ironclaw/tree/feat/deferred-tool-schemas)).

I haven't put my approach into a PR yet either — limited signal on actual utility in production. Happy to help revise this PR to move summaries into validation errors, or to put up a separate PR if that's a cleaner path.

serrrfirat (changes requested):
This rollout feels incomplete as a v2 tool-discovery change.

`tool_info(detail: "summary")` is only genuinely useful when a tool provides a curated `discovery_summary()`. In this PR, summaries were added for `http`, `create_job`, `message`, and `time` (with `routine_create` already present), but most of the tool surface still falls back to the minimal schema-derived summary. That means required-field names are available, but the real usage guidance, caveats, and examples are still missing for many of the tools users and models rely on most.

Please either:
1. expand this PR to add curated summaries for the rest of the built-in tools, especially the high-traffic ones (`shell`, file tools, memory tools, secrets tools, image tools, skill/extension tools, and the remaining job/routine tools), or
2. explicitly scope the feature and naming/docs so we are not presenting this as a general v2 discovery-summary rollout when it is still partial.

As written, this gives us inconsistent discovery quality across tools, which is exactly where the new mechanism is supposed to help the model avoid misuse.

serrrfirat:
Carrying this forward based on the latest review feedback.

What changed:
- narrowed the rollout framing from a general discovery-summary v2 change to **selected built-in tools**
- clarified `tool_info(detail: "summary")` semantics: curated summaries exist for selected tools today, and other tools fall back to schema-derived required fields
- added `summary_source` to `tool_info(detail: "summary")` responses so callers can distinguish `curated` vs `schema_fallback`
- scoped tool description hints so only tools with curated summaries advertise `detail: "summary"`; schema-only tools now point to `detail: "schema"`
- updated PR title/body to match the narrower scope

Validation run:
- `cargo fmt`
- `cargo test --lib tool_info`
- `cargo test --lib schema_hint`
- `cargo test --test e2e_builtin_tool_coverage tool_info_discovery`

Pushed in `04b3f5ec`.

serrrfirat:
Thanks for putting this together! Closing in favor of #2866, which lands the same `ToolDiscoverySummary` mechanism but covers the higher-risk `file`/`shell`/`apply_patch` tools and updates the codeact preamble so the agent knows to call `tool_info(detail="summary")` on demand. Both PRs converged on the same approach (per the planning consolidation in #2835).

If you'd like to salvage the curated summaries for `message`, `time`, and `job` (which #2866 doesn't cover), a follow-up PR on top of #2866 would be very welcome — those would be a clean additive change once it lands. Just rebase onto staging after #2866 merges and trim to those three tools.

PR #1759, opened 2026-03-30
Async transaction approval with WalletConnect

### LINKED ISSUE #1739: Async transaction approval system with WalletConnect
## Summary

A two-phase async transaction tool pattern that enables IronClaw's agent to initiate high-stakes actions (starting with Ethereum transactions) while keeping the human approval loop on a secure channel the agent cannot see or influence.

The agent declares intent, the tool simulates and validates, the user approves in their wallet app via WalletConnect, and the result flows back to the agent as an `IncomingMessage` through the existing channel system.

## Motivation

IronClaw needs to support actions where:

- The consequences are irreversible and financially significant (crypto transfers, exchange trades)
- The agent should not be trusted to approve its own actions
- The approval channel must be outside the agent's visibility to prevent manipulation
- The operation is inherently async (user may take minutes to respond in their wallet)

The existing `ApprovalRequirement` system handles in-band approval (agent pauses, user approves in the same UI). This feature addresses out-of-band approval where a separate secure channel owns the trust relationship.

## Architecture

Three layers:

### 1. Callback Adapter (shared, reusable)

`ToolCallbackRegistry` -- maps async external results back into the agent's conversation.

**Location:** `src/tools/callback.rs`

**Responsibilities:**
- Maintain a map of correlation ID to metadata (tool name, user ID, thread ID, channel)
- Convert external results into `IncomingMessage` and inject via `ChannelManager`
- Sweep stale entries after configurable TTL

**Interface:**

```rust
pub struct ToolCallbackRegistry {
    // correlation_id -> CallbackMetadata
}

pub struct CallbackMetadata {
    pub tool_name: String,
    pub user_id: String,
    pub thread_id: Option<String>,
    pub channel: String,
    pub registered_at: Instant,
}

impl ToolCallbackRegistry {
    /// Register a pending async tool result.
    pub fn register(&self, correlation_id: String, metadata: CallbackMetadata);

    /// Resolve a pending result, injecting an IncomingMessage into the channel system.
    pub async fn resolve(&self, correlation_id: &str, result: String) -> Result<(), Error>;

    /// Cancel a pending result (cleanup).
    pub fn cancel(&self, correlation_id: &str);
}
```

**Design decisions:**
- The adapter knows nothing about WalletConnect, Ethereum, or any specific backend
- It only maps "correlation ID + result string" to an `IncomingMessage`
- Any future async tool (exchange APIs, payment processors) reuses this component

### 2. Ethereum Wallet Tools

**Location:** `src/tools/builtin/ethereum/`

```
src/tools/builtin/ethereum/
  mod.rs          # Module root, exports both tools
  pair.rs         # wallet_pair tool
  transact.rs     # wallet_transact tool
  session.rs      # WalletConnect session management & persistence
  listener.rs     # Background task listening for wallet responses
```

#### `wallet_pair` tool

Initiates WalletConnect pairing so the user can connect their wallet.

- **Parameters:** optional `chain_id`
- **Returns:** WalletConnect pairing URI (QR code / deep link content)
- **Approval:** `ApprovalRequirement::Never` (no sensitive action)
- **Flow:** Generate pairing URI, return it to the agent, agent presents it to the user. Background listener waits for the wallet to accept. Session is persisted to `~/.ironclaw/` or database.

#### `wallet_transact` tool

Submits a transaction intent for user approval via their wallet.

- **Parameters:** `to`, `value`, `data` (optional), `chain_id`
- **Approval:** `ApprovalRequirement::Never` (the wallet app is the approval mechanism)
- **Sensitive params:** `data`

**Execution flow:**

1. Validate WalletConnect session exists (error if not paired)
2. Simulate transaction via JSON-RPC (`eth_estimateGas`, balance checks)
3. If simulation fails, return error immediately to agent
4. If simulation succeeds, send transaction request via WalletConnect
5. Register correlation ID with `ToolCallbackRegistry`
6. Return `pending` status with correlation ID and simulation summary

**Agent sees on tool call:**
```json
{
  "status": "pending",
  "correlation_id": "abc-123",
  "simulation": {
    "estimated_gas": "21000",
    "balance_sufficient": true
  }
}
```

**Agent sees later as IncomingMessage:**
```
Transaction abc-123: confirmed. Tx hash: 0xdef... Block: 19234567
```
or
```
Transaction abc-123: rejected by user in wallet.
```

### 3. WalletConnect Session Management

**Pairing:** Initiated via the `wallet_pair` tool. The pairing URI is presented to the user through IronClaw's normal channels. After the user scans it, the session is established and persisted.

**Persistence:** Session stored in `~/.ironclaw/` or database so it survives restarts. WalletConnect sessions have their own expiry/renewal mechanism.

**Background listener:** A tokio task connected to the WalletConnect relay. When a wallet response arrives (approval with tx hash, or rejection), it calls `ToolCallbackRegistry::resolve()` to inject the result as an `IncomingMessage`.

**Reconnection:** On relay disconnect, automatic reconnect with exponential backoff. On session expiry, mark as expired so the next `wallet_transact` call returns a clear error telling the agent to re-pair.

## Data Flow

```
Agent                    Tool                     WalletConnect           User Wallet
  |                       |                           |                       |
  |-- wallet_transact --> |                           |                       |
  |                       |-- eth_estimateGas ------> |                       |
  |                       |<- gas estimate ---------- |                       |
  |                       |-- send tx request ------> |-- push to wallet --> |
  |                       |                           |                       |
  |<- pending + corr_id - |                           |                       |
  |                       |                           |                       |
  | (agent continues)     |                           |    (user reviews)     |
  |                       |                           |                       |
  |                       |                           |<-- approve/reject --- |
  |                       |<- wallet response ------- |                       |
  |                       |                           |                       |
  |                       |-- resolve(corr_id) -----> ToolCallbackRegistry   |
  |                       |                              |                    |
  |<- IncomingMessage --- | <--- ChannelManager <------- |                    |
  |                       |                                                   |
```

### Synchronous errors (returned immediately as `ToolError`)

- Insufficient balance
- Invalid recipient address
- Gas estimation failure (contract revert)
- No WalletConnect session (not paired)
- WalletConnect session expired

### Async errors (arrive as `IncomingMessage` via callback adapter)

- User rejected in wallet
- Transaction reverted on-chain after submission
- WalletConnect session dropped mid-request
- Transaction timed out (no confirmation after TTL)

### Callback adapter errors

- Unknown correlation ID (stale/duplicate) -- log and discard
- ChannelManager unavailable -- log, retry with backoff, eventually drop

### WalletConnect listener errors

- Relay disconnection -- automatic reconnect with backoff
- Session expiry -- mark expired, next tool call gets clear error

### Stale transactions

- TTL sweep in callback adapter catches unresolved correlation IDs
- Expired entries produce a timeout `IncomingMessage`

## Integration Points

| File | Change |
|------|--------|
| `src/tools/callback.rs` | New file: `ToolCallbackRegistry` |
| `src/tools/builtin/ethereum/` | New module: `wallet_pair`, `wallet_transact`, session, listener |
| `src/tools/registry.rs` | Register `wallet_pair` and `wallet_transact` |
| `src/app.rs` | Initialize `ToolCallbackRegistry`, start WalletConnect listener |
| `src/config/` | New config: Ethereum RPC URL, WalletConnect project ID, callback TTL |

**What stays unchanged:**
- Agent loop (sees tool results and incoming messages, no modifications)
- Hook system (`BeforeToolCall` fires normally for these tools)
- Safety layer (existing sanitization applies)
- Approval system (not used; wallet is the approval mechanism)
- Channel system (used as-is for injecting results)

## Dependencies

- **WalletConnect client library** -- Rust WalletConnect v2 client (to be evaluated)
- **Ethereum library** -- `alloy` or `ethers` for transaction building and JSON-RPC

Both require evaluation for maturity, maintenance status, and transitive dependency count before adoption.

## Security Properties

1. **Agent cannot influence approval:** The transaction description the user sees in their wallet is constructed by the wallet and blockchain, not by the agent. The agent only provides intent parameters.
2. **Agent cannot see the secure channel:** WalletConnect communication happens outside the agent's context. The agent only sees the final result.
3. **Agent cannot forge results:** Results come from the WalletConnect relay through the background listener, not from agent-generated content.
4. **Sensitive parameters are redacted:** The `data` field (contract call data) is marked sensitive and redacted in logs and agent-visible output.
5. **Session binding:** The WalletConnect session is bound to a specific user and wallet. Re-pairing requires explicit `wallet_pair` invocation.

## Future Extensions

- Additional transaction backends (exchange APIs, payment processors) reuse `ToolCallbackRegistry`
- Multi-signature approval (multiple wallets must approve)
- Transaction batching (multiple transactions in one approval)
- Chain-specific simulation (Solana, Cosmos, etc.)
- WASM transaction tools using the same async pattern

---

Full spec: `docs/superpowers/specs/2026-03-29-async-transaction-approval-design.md`

zmanian:
## Summary

- Adds `ToolCallbackRegistry` -- a shared, reusable adapter that maps async external tool results back into the agent's conversation as `IncomingMessage` via the channel system
- Adds Ethereum wallet tools (`wallet_pair`, `wallet_transact`) for out-of-band transaction approval via WalletConnect v2
- Integrates a [forked walletconnect-client](https://github.com/zmanian/walletconnect-client) with native tokio transport support

The agent never sees the approval channel -- the user approves transactions in their wallet app, and the result flows back through the callback adapter.

## Architecture

Three layers:

1. **ToolCallbackRegistry** (`src/tools/callback.rs`) -- registers pending async results by correlation ID, resolves them as `IncomingMessage`, sweeps expired entries with configurable TTL
2. **WalletConnect session + listener** (`src/tools/builtin/ethereum/session.rs`, `listener.rs`) -- wraps the real WC v2 client, handles pairing, polls for wallet events (Connected, Disconnected, AccountsChanged, etc.)
3. **Ethereum tools** (`pair.rs`, `transact.rs`) -- `wallet_pair` initiates WC pairing and returns the URI; `wallet_transact` validates params, registers a callback, and returns a pending state

## New files

- `src/tools/callback.rs` -- ToolCallbackRegistry
- `src/tools/builtin/ethereum/` -- mod, error, session, pair, transact, listener
- `src/config/ethereum.rs` -- EthereumConfig
- `tests/callback_registry.rs` -- 7 integration tests

## Still TODO

- Transaction simulation via `alloy` (gas estimation, balance checks)
- Listener integration with `wallet_transact` callback resolution
- Session persistence across restarts
- WalletConnect session expiry and re-pairing

Closes #1739

## Test plan

- [ ] 22 tests pass (7 callback + 15 ethereum): `cargo test --test callback_registry && cargo test ethereum`
- [ ] Zero clippy warnings: `cargo clippy --all --benches --tests --examples --all-features`
- [ ] Manual test: set `ETHEREUM_ENABLED=true` and `WALLETCONNECT_PROJECT_ID=<id>`, call `wallet_pair`, scan QR in wallet app
- [ ] Verify `wallet_transact` returns pending state with correlation ID when paired

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -201,6 +201,9 @@ ed25519-dalek = { version = "2.2.0", features = ["std"] }
 bs58 = "0.5"
 hex = "0.4.3"
 
+# WalletConnect v2 client for Ethereum transaction approval
+walletconnect-client = { git = "https://github.com/zmanian/walletconnect-client.git", branch = "main", version = "0.2.0", default-features = false, features = ["native"] }
+
 # OpenClaw import (feature gated)
 json5 = { version = "0.4", optional = true }
 
--- a/deny.toml
+++ b/deny.toml
@@ -20,6 +20,8 @@ ignore = [
     # rand unsoundness with custom logger calling rand::rng() during reseed — we don't use this pattern;
     # revisit/remove by 2026-06-30, or when transitive deps (tower, nanoid, phf_generator) release rand ≥0.9.3 compat
     "RUSTSEC-2026-0097",
+    # ring 0.16.20 AES overflow panic — pinned by jsonwebtoken 8.x via walletconnect-client
+    "RUSTSEC-2025-0009",
 ]
 
 [licenses]
@@ -47,6 +49,13 @@ allow = [
 ]
 unused-allowed-license = "allow"
 
+[[licenses.clarify]]
+name = "ring"
+version = "0.16.20"
+# ring 0.16.x uses ISC/OpenSSL/MIT but cargo-deny can't parse the non-standard LICENSE file
+expression = "ISC AND MIT AND OpenSSL"
+license-files = [{ path = "LICENSE", hash = 0xbd0eed23 }]
+
 [bans]
 multiple-versions = "warn"
 wildcards = "deny"
@@ -63,4 +72,6 @@ allow-git = [
     # Pulls in ruff_* crates from astral-sh/ruff at a pinned revision.
     "https://github.com/pydantic/monty.git",
     "https://github.com/astral-sh/ruff.git",
+    # WalletConnect v2 client — forked for native tokio transport support
+    "https://github.com/zmanian/walletconnect-client.git",
 ]
--- a/src/main.rs
+++ b/src/main.rs
@@ -994,6 +994,21 @@ async fn async_main() -> anyhow::Result<()> {
         .register_message_tools(Arc::clone(&channels), components.extension_manager.clone())
         .await;
 
+    // Start Ethereum WalletConnect listener and callback sweep if enabled
+    if let Some(ref wc_session) = components.wc_session {
+        let listener = ironclaw::tools::builtin::ethereum::WalletConnectListener::new(
+            Arc::clone(wc_session),
+            Arc::clone(&components.callback_registry),
+            channels.inject_sender(),
+        );
+        let _listener_handle = listener.start();
+
+        let _sweep_handle = components
+            .callback_registry
+            .start_sweep_task(channels.inject_sender(), std::time::Duration::from_secs(30));
+        tracing::info!("Ethereum WalletConnect listener and callback sweep started");
+    }
+
     // Default user ID for extension operations (single-user mode).
     let ext_user_id = config.owner_id.clone();
 
--- a/src/tools/builtin/ethereum/listener.rs
+++ b/src/tools/builtin/ethereum/listener.rs
@@ -1,0 +1,77 @@
+use std::sync::Arc;
+
+use tokio::sync::mpsc;
+use walletconnect_client::prelude::Event;
+
+use crate::channels::IncomingMessage;
+use crate::tools::callback::ToolCallbackRegistry;
+
+use super::session::{SessionStatus, WalletConnectSession};
+
+/// Background listener that monitors the WalletConnect relay for wallet
+/// events (connection, disconnection, account/chain changes, transaction
+/// responses) and updates session state accordingly.
+pub struct WalletConnectListener {
+    session: Arc<WalletConnectSession>,
+    #[allow(dead_code)]
+    callback_registry: Arc<ToolCallbackRegistry>,
+    #[allow(dead_code)]
+    inject_tx: mpsc::Sender<IncomingMessage>,
+}
+
+impl WalletConnectListener {
+    pub fn new(
+        session: Arc<WalletConnectSession>,
+        callback_registry: Arc<ToolCallbackRegistry>,
+        inject_tx: mpsc::Sender<IncomingMessage>,
+    ) -> Self {
+        Self {
+            session,
+            callback_registry,
+            inject_tx,
+        }
+    }
+
+    /// Start the background listener. Returns a JoinHandle.
+    pub fn start(self) -> tokio::task::JoinHandle<()> {
+        tokio::spawn(async move {
+            tracing::info!("WalletConnect listener started");
+
+            loop {
+                match self.session.poll_event().await {
+                    Some(event) => self.handle_event(event).await,
+                    None => {
+                        // No active client or no event ready — back off briefly.
+                        tokio::time::sleep(std::time::Duration::from_millis(100)).await;
+                    }
+                }
+            }
+        })
+    }
+
+    async fn handle_event(&self, event: Event) {
+        match event {
+            Event::Connected => {
+                tracing::info!("WalletConnect: wallet connected");
+                self.session.update_from_connected().await;
+            }
+            Event::Disconnected => {
+                tracing::info!("WalletConnect: wallet disconnected");
+                self.session.set_status(SessionStatus::Disconnected).await;
+            }
+            Event::AccountsChanged(accounts) => {
+                tracing::info!("WalletConnect: accounts changed: {accounts:?}");
+                let chain_id = self.session.active_chain_id().await.unwrap_or(1);
+                self.session.update_accounts(chain_id).await;
+            }
+            Event::ChainIdChanged(new_chain_id) => {
+                tracing::info!("WalletConnect: chain changed to {new_chain_id}");
+                self.session.update_accounts(new_chain_id).await;
+            }
+            Event::Broken => {
+                tracing::warn!("WalletConnect: connection broken");
+                self.session.set_status(SessionStatus::Disconnected).await;
+            }
+        }
+    }
+}
--- a/src/tools/builtin/ethereum/session.rs
+++ b/src/tools/builtin/ethereum/session.rs
@@ -1,0 +1,219 @@
+use tokio::sync::{Mutex, RwLock};
+use url::Url;
+use walletconnect_client::prelude::*;
+
+use super::error::EthereumError;
+
+/// Status of the WalletConnect session from IronClaw's perspective.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum SessionStatus {
+    Disconnected,
+    Pairing { uri: String },
+    Paired { address: String, chain_id: u64 },
+    Expired,
+}
+
+/// Manages a WalletConnect v2 session, wrapping the real WC client.
+pub struct WalletConnectSession {
+    status: RwLock<SessionStatus>,
+    wc_client: Mutex<Option<WalletConnect<NativeTransport>>>,
+    project_id: Option<String>,
+}
+
+impl WalletConnectSession {
+    pub fn new(project_id: Option<String>) -> Self {
+        Self {
+            status: RwLock::new(SessionStatus::Disconnected),
+            wc_client: Mutex::new(None),
+            project_id,
+        }
+    }
+
+    /// Create a disconnected session with no project ID (for tests / disabled mode).
+    pub fn new_disconnected() -> Self {
+        Self::new(None)
+    }
+
+    /// Initiate a WalletConnect pairing session. Returns the pairing URI.
+    pub async fn initiate_pairing(&self, chain_id: u64) -> Result<String, EthereumError> {
+        let project_id = self.project_id.as_ref().ok_or_else(|| {
+            EthereumError::NotConfigured("WALLETCONNECT_PROJECT_ID not set".into())
+        })?;
+
+        let metadata = Metadata::from(
+            "IronClaw",
+            "Secure personal AI assistant",
+            Url::parse("https://ironclaw.dev")
+                .map_err(|e| EthereumError::WalletConnect(e.to_string()))?,
+            vec![],
+        );
+
+        let wc = WalletConnect::<NativeTransport>::connect(
+            walletconnect_client::jwt::decode::ProjectId::from(project_id.as_str()),
+            chain_id,
+            metadata,
+            None,
+        )
+        .await
+        .map_err(|e| EthereumError::WalletConnect(e.to_string()))?;
+
+        let uri = wc
+            .initiate_session(None)
+            .await
+            .map_err(|e| EthereumError::WalletConnect(e.to_string()))?;
+
+        *self.status.write().await = SessionStatus::Pairing { uri: uri.clone() };
+        *self.wc_client.lock().await = Some(wc);
+
+        Ok(uri)
+    }
+
+    /// Poll for the next WalletConnect event. Called by the listener task.
+    ///
+    /// Returns `None` if no client is active or no event is available.
+    pub async fn poll_event(&self) -> Option<Event> {
+        let mut client = self.wc_client.lock().await;
+        if let Some(ref wc) = *client {
+            match wc.next().await {
+                Ok(event) => event,
+                Err(e) => {
+                    tracing::warn!("WalletConnect event error: {e}");
+                    // On disconnect error, drop the client
+                    if matches!(e, WalletConnectError::Disconnected) {
+                        *client = None;
+                    }
+                    None
+                }
+            }
+        } else {
+            None
+        }
+    }
+
+    /// Send a JSON-RPC request through the WC client (e.g. eth_sendTransaction).
+    pub async fn request(
+        &self,
+        method: &str,
+        params: Option<serde_json::Value>,
+        chain_id: u64,
+    ) -> Result<serde_json::Value, EthereumError> {
+        let client = self.wc_client.lock().await;
+        let wc = client.as_ref().ok_or(EthereumError::NotPaired)?;
+
+        wc.request(method, params, chain_id)
+            .await
+            .map_err(|e| EthereumError::WalletConnect(e.to_string()))
+    }
+
+    /// Update session status from a WC Connected event by reading the client state.
+    pub async fn update_from_connected(&self) {
+        let client = self.wc_client.lock().await;
+        if let Some(ref wc) = *client {
+            let address = wc
+                .get_account()
+                .map(|a| format!("{a:?}"))
+                .unwrap_or_default();
+            let chain_id = wc.chain_id();
+            drop(client);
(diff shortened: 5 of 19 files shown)


src/tools/builtin/ethereum/session.rs
```diff
@@ -0,0 +1,219 @@
...
+        Ok(uri)
+    }
+
+    /// Poll for the next WalletConnect event. Called by the listener task.
+    ///
+    /// Returns `None` if no client is active or no event is available.
+    pub async fn poll_event(&self) -> Option<Event> {
+        let mut client = self.wc_client.lock().await;
+        if let Some(ref wc) = *client {
+            match wc.next().await {
```
serrrfirat, line 77: **🔴 Critical: Mutex held across `.await` — will deadlock**

`poll_event()` holds the `wc_client` Mutex guard across `wc.next().await`. The background listener calls this in a tight loop, meaning the Mutex is held for the entire duration the relay is waiting for the next event. This **starves all other methods** that need the Mutex: `request()`, `initiate_pairing()`, `update_from_connected()`, `update_accounts()`.

In practice, `wallet_transact` can never send a transaction because `request()` will block indefinitely waiting for the Mutex.

**Suggested fix:** Restructure using an actor pattern with an mpsc channel for commands, or use `tokio::sync::Mutex::try_lock()` with a timeout, or clone/Arc the WC client if the library supports it.

src/tools/builtin/ethereum/transact.rs
```diff
@@ -0,0 +1,232 @@
...
+
+        if data.is_some() {
+            result["data_present"] = json!(true);
+        }
+        if let Some(cid) = chain_id {
+            result["chain_id"] = json!(cid);
+        }
+
+        Ok(ToolOutput::success(result, start.elapsed()))
+    }
```
serrrfirat, line 169: **🔴 Critical: Transaction is never actually sent to wallet**

`wallet_transact` checks `is_paired()`, validates params, registers a callback, and returns "pending" — but **never calls `session.request("eth_sendTransaction", ...)`**. The transaction is registered as pending but will always time out.

Users will see "Transaction submitted" messaging and believe their transaction was sent when it was not.

**Suggested fix:** Either add the actual `eth_sendTransaction` call before returning, or block merge until send logic is wired, or at minimum clearly indicate "Transaction queued locally but NOT YET submitted to wallet" and add a `tracing::warn!`.

src/tools/builtin/ethereum/transact.rs
```diff
@@ -0,0 +1,232 @@
...
+                    "type": "integer",
+                    "description": "Target EVM chain ID. Defaults to the paired session chain."
+                }
+            },
+            "required": ["to", "value"]
+        })
+    }
+
+    fn requires_approval(&self, _params: &serde_json::Value) -> ApprovalRequirement {
+        // The wallet itself is the approval mechanism.
```
serrrfirat, line 82: **🟠 High: No IronClaw-level approval for transaction requests**

`requires_approval()` returns `ApprovalRequirement::Never`. While the wallet itself requires approval, the **agent (LLM) can call `wallet_transact` without any human-in-the-loop** on the IronClaw side.

A prompt injection or misaligned agent could initiate unlimited transaction requests to the user's wallet. Notification fatigue / UI confusion attacks (approval farming) are a real vector.

**Suggested fix:** Return `ApprovalRequirement::Always` — require IronClaw-level human approval before the transaction request even reaches the wallet. The wallet approval should be defense-in-depth, not the sole gate.

Cargo.toml
```diff
@@ -192,6 +192,9 @@ ed25519-dalek = { version = "2.2.0", features = ["std"] }
 bs58 = "0.5"
 hex = "0.4.3"
 
+# WalletConnect v2 client for Ethereum transaction approval
+walletconnect-client = { git = "https://github.com/zmanian/walletconnect-client.git", branch = "main", version = "0.2.0", default-features = false, features = ["native"] }
```
serrrfirat, line 205: **🟠 High: Supply chain risk — unpinned git dependency for wallet library**

`walletconnect-client` is pinned to `branch = "main"` of a personal fork (`zmanian/walletconnect-client`) with no `rev = "..."` commit pin. A force-push to that fork's main branch could inject arbitrary code into the build.

This is a **wallet/transaction library** — supply chain risk is maximal.

**Suggested fix:** Pin to a specific commit: `rev = "abc123..."`. Verify the fork's diff against upstream and document modifications.

deny.toml
```diff
@@ -17,6 +17,8 @@ ignore = [
     "RUSTSEC-2026-0021",
     # rustls-webpki CRL distributionPoint matching — 0.102.8 pinned by libsql transitive dep
     "RUSTSEC-2026-0049",
+    # ring 0.16.20 AES overflow panic — pinned by jsonwebtoken 8.x via walletconnect-client
+    "RUSTSEC-2025-0009",
```
serrrfirat, line 24: **🟠 High: Crypto advisory suppressed in wallet integration**

`RUSTSEC-2025-0009` (ring 0.16.20 AES overflow panic) is added to the advisory ignore list. This is a real vulnerability in ring used by `jsonwebtoken 8.x` pulled in by walletconnect-client.

Suppressing a crypto advisory in a wallet integration is concerning.

**Suggested fix:** Document the specific conditions under which the panic can be triggered and confirm they don't apply. Better: upgrade the walletconnect-client fork to use `jsonwebtoken 9.x` (ring 0.17+).

src/tools/builtin/ethereum/session.rs
```diff
@@ -0,0 +1,219 @@
...
+        )
+        .await
+        .map_err(|e| EthereumError::WalletConnect(e.to_string()))?;
+
+        let uri = wc
+            .initiate_session(None)
+            .await
+            .map_err(|e| EthereumError::WalletConnect(e.to_string()))?;
+
+        *self.status.write().await = SessionStatus::Pairing { uri: uri.clone() };
```
serrrfirat, line 65: **🟠 High: TOCTOU in pairing initialization**

`initiate_pairing()` updates `status` (RwLock) and `wc_client` (Mutex) in two separate lock acquisitions. Between these lines, another caller could observe `status == Pairing` but `wc_client == None`. Concurrent calls to `initiate_pairing` could race: both create WC clients, but only the second one's client gets stored.

**Suggested fix:** Use a single lock protecting both fields, or add a guard against concurrent pairing (check status is Disconnected before starting).

src/tools/builtin/ethereum/session.rs
```diff
@@ -0,0 +1,219 @@
...
+            .map_err(|e| EthereumError::WalletConnect(e.to_string()))
+    }
+
+    /// Update session status from a WC Connected event by reading the client state.
+    pub async fn update_from_connected(&self) {
+        let client = self.wc_client.lock().await;
+        if let Some(ref wc) = *client {
+            let address = wc
+                .get_account()
+                .map(|a| format!("{a:?}"))
```
serrrfirat, line 114: **🟡 Medium: Debug formatting used for wallet address**

`format!("{a:?}")` uses Debug formatting for the account address. This will produce output like `H160(0x1234...)` instead of the expected `0x1234...` hex string. The `Paired.address` field will contain the Debug representation, which will confuse users and break any downstream address comparison.

Same issue in `update_accounts()`.

**Suggested fix:** Use the appropriate display/hex formatting (e.g., `format!("{a:#x}")` or the type's `Display` impl).

src/tools/builtin/ethereum/listener.rs
```diff
@@ -0,0 +1,77 @@
...
+            session,
+            callback_registry,
+            inject_tx,
+        }
+    }
+
+    /// Start the background listener. Returns a JoinHandle.
+    pub fn start(self) -> tokio::task::JoinHandle<()> {
+        tokio::spawn(async move {
+            tracing::info!("WalletConnect listener started");
```
serrrfirat, line 38: **🟡 Medium: Background task uses `info!` logging — breaks REPL/TUI**

CLAUDE.md states: "Background tasks must NEVER use `info!` — it breaks the interactive display." The WalletConnect listener is a background task but uses `tracing::info!` extensively (7 instances across listener + callback + app + main).

**Suggested fix:** Change all background-task logging to `tracing::debug!`.

src/main.rs
```diff
@@ -853,6 +853,21 @@ async fn async_main() -> anyhow::Result<()> {
...
         .await;
 
+    // Start Ethereum WalletConnect listener and callback sweep if enabled
+    if let Some(ref wc_session) = components.wc_session {
+        let listener = ironclaw::tools::builtin::ethereum::WalletConnectListener::new(
+            Arc::clone(wc_session),
+            Arc::clone(&components.callback_registry),
+            channels.inject_sender(),
+        );
+        let _listener_handle = listener.start();
```
serrrfirat, line 1004: **🟡 Medium: Background task handles dropped — no graceful shutdown**

`_listener_handle` and `_sweep_handle` are assigned to underscore-prefixed locals that are immediately dropped at scope end. The spawned tasks run forever with no shutdown mechanism. On application shutdown, these tasks will be forcefully aborted, potentially mid-operation.

**Suggested fix:** Store handles in `AppComponents` or a shutdown coordinator. Use a `CancellationToken` or `watch` channel to signal graceful shutdown.

src/tools/builtin/ethereum/transact.rs
```diff
@@ -0,0 +1,232 @@
...
+        let to = require_str(&params, "to")?;
+        if !is_valid_eth_address(to) {
+            return Err(EthereumError::InvalidAddress {
+                address: to.to_string(),
+            }
+            .into());
+        }
+
+        let value = require_str(&params, "value")?;
+        let data = params.get("data").and_then(|v| v.as_str());
```
serrrfirat, line 120: **🟡 Medium: No validation on `value` and `data` transaction parameters**

`value` is accepted as an arbitrary string described as "Value to send in wei (decimal string)" but could be anything: negative numbers, floats, empty string, non-numeric text. `data` (calldata) described as "0x-prefixed hex" is also not validated.

**Suggested fix:** Validate `value` is a non-negative decimal integer string (parse to `u128` or `U256`). Validate `data` starts with `0x` and remaining chars are hex digits with even length.

serrrfirat:
## 🏗️ Paranoid Architect Review — PR #1759

**Verdict: 🚫 DO NOT MERGE — Critical blockers**

### Summary
| Severity | Count |
|----------|-------|
| Critical | 2 |
| High | 4 |
| Medium | 4 |

### Critical Issues
1. **Deadlock**: `poll_event()` holds Mutex across `.await` — starves all other WC methods including transaction sending
2. **Phantom transaction**: `wallet_transact` never calls `eth_sendTransaction` — users believe tx was sent when it wasn't

### High Issues  
3. **No IronClaw-level approval gate** — LLM can spam unlimited tx requests (approval farming)
4. **Supply chain**: unpinned git dependency on personal fork of wallet library
5. **Crypto advisory suppressed**: RUSTSEC-2025-0009 ignored in wallet integration
6. **TOCTOU in pairing**: status and client updated under separate locks

### Key Security Assessment
- **Can transactions be replayed?** Not testable since transactions are never actually sent
- **Can the agent drain funds?** Agent can spam `wallet_transact` without IronClaw approval (finding #3)
- **Are secrets properly handled?** WC project ID is plain string, not `secrecy::Secret`
- **Session persistence?** None — sessions lost on restart (acknowledged TODO)

zmanian:
Closing this draft. The diff has aged ~6 weeks against a fast-moving target (WalletConnect + alloy were both pre-1.0 when this was written; both have shifted APIs since), making a straight rebase higher-cost than a fresh re-implementation. The architectural shape — fork walletconnect-client + alloy for tx encoding — is captured in this PR's history and in issue #1739; the feature itself will be re-landed as a new PR built on top of the hooks framework's gate-ref approval flow when it's the next thing on the roadmap.

PR #1764, opened 2026-03-30
feat: Abound demo — Responses API, credential injection, skills, guardrails

pranavraja99:
## Summary

Abound demo deployment on `v2-architecture`. Fixes the Responses API for production use, adds config-driven credential injection with path-scoped auth, ships a full Abound integration under `integrations/abound/`, and includes forex timing intelligence via the smart-remittance skill.

Now merged with latest `staging` (v2 engine architecture, glibc Dockerfile, gate system).

### Responses API fixes
- **New thread creation**: Allow fresh thread UUIDs from the Responses API by checking `metadata.source == "responses_api"` in thread validation
- **GET by ID**: Session manager now adopts the external UUID as the internal thread ID so DB lookups match the response ID
- **Streaming text deltas**: Emit full text as `response.output_text.delta` when the final Response event arrives
- **Context injection**: New `x_context` field (IronClaw extension, aliased as `context`) for structured data (e.g. notification approval/rejection from Abound's app)

### Credential injection
- **Path-based matching**: `CredentialMapping` now supports `path_patterns` so different secrets can be scoped to different endpoints on the same host
- **Config-driven loading**: `INTEGRATION_CREDENTIALS_DIR` env var scans a directory (+ one level of subdirectories) for `*.json` credential mapping files
- Read endpoints use `abound_read_token`, send-wire uses `abound_write_token`, all endpoints get `X-API-KEY`

### Skills (contributed by @zetyquickly via #1892)
- **`abound-api`** (`skills/abound-api/SKILL.md`): Teaches the agent to call Abound's REST API via the `http` tool. Includes `credentials:` block for auto-injection, activation patterns for follow-up phrasings, guardrails (no URL leaks, no competitors, branded welcome)
- **`forex-timing`** (`skills/forex-timing/SKILL.md`): USD/INR forex timing via Massive API + Yahoo Finance DXY — volatility regime, RSI, projection cone, hit-rate analysis
- Choice set support: agent emits `[[choice_set]]...[[/choice_set]]` structured JSON for interactive UI cards (payment reasons, recipients, amount ranges)

### Abound integration (`integrations/abound/`)
All Abound-specific files live in `integrations/abound/` for eventual extraction:
- `credentials.json` — path-scoped credential mappings for Abound's dev + prod hosts
- `workspace/AGENTS.md` — custom system prompt (auto-seeded via `AGENTS_SEED_PATH`)
- `tests/test_abound_api_direct.py` — direct tests against Abound's dev API
- `tests/test_abound_e2e.py` — E2E tests through IronClaw's Responses API (natural language prompts, credential injection via admin API)
- `tests/test_abound_guardrails.py` — adversarial guardrail tests (frustrated users, prompt injection, competitor probing, credential extraction)

### Guardrails
- Agent won't reveal API URLs, endpoint paths, or internal technical details
- Agent won't recommend competing remittance services
- Agent won't expose credential names or raw API responses
- Specific names removed from agent context to prevent adversarial extraction
- Custom Abound welcome message via `AGENTS_SEED_PATH`

### Deployment
- `python3-dev` added to Dockerfile for `pyo3` (transitive dep from `ironclaw_engine → monty → jiter → pyo3`)
- Merged with latest staging (glibc Dockerfile, v2 engine, gate system)
- Skills and integrations directories copied to runtime image

### Test scripts (`tests/scripts/`)
- `test_admin_api.py` — Admin API lifecycle (user CRUD, secrets, suspend/activate)
- `test_responses_api.py` — Responses API with OpenAI client (non-streaming, streaming, multi-turn, GET by ID, context injection)

### E2E tests (`tests/e2e/scenarios/`)
- `test_admin_api.py` — pytest/httpx admin API tests
- `test_responses_api.py` — pytest/OpenAI client responses API tests
- `SECRETS_MASTER_KEY` added to `ironclaw_server` fixture

## Question for @ilblackdragon

The Abound-specific logic (credentials, workspace/AGENTS.md, tests) currently lives in `integrations/abound/` within the main IronClaw repo. Skills live in `skills/abound-api/` and `skills/forex-timing/`.

Do you have suggestions on a better way to structure customer-specific logic? Should we move the `integrations/abound/` directory to a separate repo (e.g. `nearai/ironclaw-abound`) and load it at deploy time? Or is the current `integrations/` + env-var-driven approach sufficient for now?

The credential mappings (`credentials.json`) and system prompt (`AGENTS.md`) are loaded via env vars (`INTEGRATION_CREDENTIALS_DIR`, `AGENTS_SEED_PATH`), so they're already decoupled from the core binary. The main coupling point is the `skills/` directory which is compiled into the Docker image.

## Test plan
- [x] `python tests/scripts/test_admin_api.py` — all passing
- [x] `python tests/scripts/test_responses_api.py` — all passing
- [x] `python integrations/abound/tests/test_abound_e2e.py` — 22/22 passing (real Abound dev API)
- [x] `python integrations/abound/tests/test_abound_guardrails.py` — 70/79 passing (known: agent echoes competitor names when declining, breaks character on "what AI are you")
- [x] Railway deployment verified at `ironclaw-production-e3b1.up.railway.app`


zmanian (changes requested):
## Review -- REQUEST_CHANGES

### Is the build actually broken?

`pyo3-build-config` v0.28.2 does not require a Python interpreter when the `resolve-config` feature is disabled -- it falls back to `target-lexicon` for platform detection. The PR provides no evidence of a build failure: no error log, no linked issue, no reproduction steps.

Before adding python3 to the production Dockerfile, please:
1. Share the actual build error you encountered
2. Confirm whether `monty` or `pyo3-build-config` enables the `resolve-config` feature (check `Cargo.lock` feature resolution)
3. If the feature is enabled, consider fixing it in `Cargo.toml` rather than adding Python to the container

### Other issues
- PR template is completely unfilled -- need change type, validation steps, linked issue
- CI: regression test enforcement failing (needs `skip-regression-check` label for Dockerfile-only change)
- CI: cargo-deny failing (appears pre-existing on base branch, not caused by this PR)

pranavraja99:
## Addressing review feedback

### Build error evidence

The build fails during `cargo chef cook` on Railway with:

```
error: failed to run custom build command for `pyo3-build-config v0.28.2`
  --- stderr
  error: no Python 3.x interpreter found
```

Full error: the `cargo chef cook` step resolves the full dependency graph and runs build scripts for all transitive deps.

### Dependency chain

The issue is **not** `pyo3-build-config` alone — it's the full `pyo3` crate pulled in transitively:

```
ironclaw_engine (new in v2-architecture)
  └─ monty v0.0.9 (git dep from pydantic)
       ├─ jiter v0.13.0
       │    └─ pyo3 v0.28.2          ← full Python bindings, requires interpreter
       └─ pyo3-build-config v0.28.2  ← this one only uses target-lexicon (no Python needed)
```

`jiter` depends on `pyo3` (not just `pyo3-build-config`), and `pyo3`'s build script unconditionally probes for a Python interpreter. This is why staging (which doesn't have `ironclaw_engine` or `monty`) builds fine without Python.

### Why not fix in Cargo.toml?

`pyo3` is a transitive dep of `jiter` (from the `monty` git dep). We can't control `jiter`'s feature flags from our Cargo.toml without forking `monty`. The `monty` crate is from pydantic and pinned to a git rev — this is an upstream issue.

### Resolution

Adding `python3-dev` to the Dockerfile is the correct fix. It provides both the interpreter and headers that `pyo3`'s build script needs. This only affects the build stage — the runtime image remains unchanged (slim debian with no Python).

The `Dockerfile.test` also needs the same change since it builds the same binary.

zmanian (changes requested):
## Review -- REQUEST_CHANGES

### BLOCKING: Hardcoded secrets committed to the repository

Multiple files contain hardcoded production credentials that must be removed before merge:

**`tests/scripts/test_admin_api.py` (line 4-5 of new file):**
```python
BASE_URL = "https://ironclaw-production-e3b1.up.railway.app"
ADMIN_TOKEN = "39a5644953ff8edf2df5c56fcfc7027e3392000381d5d8157552c1a51bee4cca"
```

**`tests/scripts/test_admin_api.sh` (lines 8-9):**
```bash
export BASE_URL=https://ironclaw-production-e3b1.up.railway.app
export ADMIN_TOKEN=39a5644953ff8edf2df5c56fcfc7027e3392000381d5d8157552c1a51bee4cca
```

**`tests/scripts/test_responses_api.py` (lines 2-3 of constants):**
```python
BASE_URL = "https://ironclaw-production-e3b1.up.railway.app"
ADMIN_TOKEN = "39a5644953ff8edf2df5c56fcfc7027e3392000381d5d8157552c1a51bee4cca"
```

**`integrations/abound/tests/test_abound_api_direct.py` (lines 5-8):** Hardcoded JWT bearer token and API key as defaults:
```python
ABOUND_BEARER_TOKEN = os.environ.get("ABOUND_BEARER_TOKEN", "eyJhbGciOiJIUzM4NCJ9...")
ABOUND_API_KEY = os.environ.get("ABOUND_API_KEY", "a105acd4-74f6-46b6-b429-c2b764462b99")
```

**`integrations/abound/tests/test_abound_e2e.py` (lines 5-10):** Same hardcoded admin token and Abound credentials.

**`tests/e2e/conftest.py`:** Hardcoded `SECRETS_MASTER_KEY` (line added). While this is for test only, using a predictable key in a committed test fixture is a risk if anyone copies it.

**Action required:** Remove all hardcoded tokens/keys. These scripts should require env vars with no fallback defaults, or read from a `.env` file that is `.gitignore`d. The admin token `39a5644...` in particular needs to be rotated immediately since it is now in git history.

---

### BLOCKING: `AGENT_AUTO_APPROVE_TOOLS=true` baked into production Dockerfile

**`Dockerfile` (line added):**
```dockerfile
ENV AGENT_AUTO_APPROVE_TOOLS=true
```

This auto-approves all tool calls for every user in this Docker image, bypassing the safety/confirmation layer. This is a demo shortcut that should not be in the shared Dockerfile. If the Abound deployment needs this, it should be set in the Railway env vars, not baked into the image that other deployments may use.

---

### Architecture: Abound-specific integration baked into shared Dockerfile

The Dockerfile now hardcodes Abound-specific paths:
```dockerfile
COPY integrations/ /app/integrations/
ENV SKILLS_DIR=/app/integrations/abound/skills
ENV INTEGRATION_CREDENTIALS=/app/integrations/abound/credentials.json
```

This couples the shared Docker image to a single customer integration. These should be set as deployment-specific env vars (e.g. in Railway), not in the base Dockerfile. The `COPY integrations/` is fine, but the `ENV` lines should be removed.

---

### Rust code: `load_credential_mappings` uses `.unwrap_or` for silent defaults

**`src/tools/registry.rs` (new `load_credential_mappings` method):**
```rust
let name = entry["location"]["name"]
    .as_str()
    .unwrap_or("X-API-KEY")
    .to_string();
```
and
```rust
.unwrap_or("api_key")
```

Silent defaults are dangerous for credential routing. If the JSON is malformed and `name` is missing, silently defaulting to `X-API-KEY` means credentials could be sent to the wrong header. This should log a warning or skip the entry when `name` is missing for `header`/`query` location types.

Also, the entire method does manual JSON parsing with `serde_json::Value`. Since `CredentialMapping` and `CredentialLocation` already exist as types, consider deserializing directly into a struct with `#[derive(Deserialize)]` instead of hand-parsing. This would catch malformed input at parse time rather than silently defaulting.

---

### Rust code: `format_context` only handles one level of nesting

**`src/channels/web/responses_api.rs` (`format_context`):**

The function only handles `{key: {flat_object}}`. Nested objects, arrays, or mixed structures will be `to_string()`'d into raw JSON in the agent's context, which is probably confusing. Consider either:
1. Documenting that `context` must be `{string: flat_object}` and rejecting deeper nesting with a 400 error
2. Using a recursive formatter

Not blocking but worth addressing for robustness.

---

### Responses API: `context` field is a non-standard extension

The `context` field on `ResponsesRequest` is not part of the OpenAI Responses API spec. This should be clearly documented (in the struct doc comment or README) as an IronClaw extension, and ideally namespaced (e.g. under `metadata` or a vendor-prefixed field like `x_ironclaw_context`) to avoid future collisions if OpenAI adds a `context` field.

---

### Tests: E2E test files are well-structured

The `tests/e2e/scenarios/test_admin_api.py` and `test_responses_api.py` are good additions using proper pytest fixtures with cleanup. The fixture-based approach with `test_user` and `responses_user` is clean.

One issue: in `test_responses_api.py`, `openai_client` is a sync fixture that depends on `responses_user` (an async fixture). This may cause issues depending on the pytest-asyncio version -- verify this works in CI.

---

### Summary

The Responses API bug fixes (thread creation, GET by ID, streaming deltas) look correct and well-tested. The context injection feature is a reasonable approach. However, the hardcoded production credentials are a showstopper -- those need to be removed and the admin token rotated. The Dockerfile changes need to be deployment-specific rather than baked in.

Dockerfile
```diff
@@ -79,5 +80,8 @@ USER ironclaw
 EXPOSE 3000
 
 ENV RUST_LOG=ironclaw=info
+ENV SKILLS_DIR=/app/integrations/abound/skills
```
zetyquickly: this shouldn't be merged to v2-architecture, it's only relevant for this deployment

integrations/abound/tests/test_abound_api_direct.py
```diff
@@ -0,0 +1,128 @@
...
+"""
+
+import os
+
+import requests
+
+# Credentials from env or defaults (dev environment)
+ABOUND_BEARER_TOKEN = os.environ.get(
+    "ABOUND_BEARER_TOKEN",
+    "eyJhbGciOiJIUzM4NCJ9.eyJleHAiOjE3Nzc2MzM3MjUsImN1c3RvbWVyLWlkIjoiYTk0MTkyNTAtZWRlNy00MWEwLWE0MjItN2Y0ZTZmNDMzNjVmLTE3Njg0NjgzODUzMTgifQ.g3SDYkF2ns4GI3eo-l2f1OI23QN6gtoKNdUrNZfiVLvyOROEIivZ7pkp_NQDylZ4",
```
zetyquickly: I hope this is a key that doesn't expose anything valuable

pranavraja99: claude pushed this 🤦 removing

Dockerfile
pranavraja99: yes agreed we'll move this out into a separate repo

pranavraja99:
## Addressing review feedback

**Hardcoded secrets**: Already fixed in `cd20f2f` — all test scripts require env vars with no defaults (`os.environ["KEY"]`). The tokens in git history need rotation (noted in the commit message). The `SECRETS_MASTER_KEY` in `conftest.py` is a deterministic test-only value (all zeros pattern) used by other fixtures in the same file — consistent with the existing extension_cleanup and hosted_oauth fixtures.

**AGENT_AUTO_APPROVE_TOOLS in Dockerfile**: Removed. Now a deployment-specific env var (set in Railway, not baked in). → `5bcb7ac`

**Abound-specific ENV in Dockerfile**: Removed `SKILLS_DIR`, `INTEGRATION_CREDENTIALS` ENV lines. The `COPY integrations/` stays (static files), but paths are set in Railway. → `5bcb7ac`

**load_credential_mappings silent defaults**: Fixed. Missing `name` on header/query location types now logs a warning and skips the entry instead of defaulting to `X-API-KEY`/`api_key`. → `5bcb7ac`

**context field namespacing**: Renamed to `x_context` with `#[serde(alias = "context")]` for backward compat. Doc comment now clearly states it's an IronClaw extension, not part of the OpenAI spec, and documents the flat-object constraint. → `5bcb7ac`

**format_context nesting**: Documented that `x_context` must be `{string: flat_object}` — nested objects are `to_string()`'d. A 400 rejection for deep nesting is a future improvement.

**E2E fixture sync/async**: `openai_client` is now an async fixture. Removed redundant `@pytest.mark.asyncio` markers (per `asyncio_mode = "auto"` in pyproject.toml). → `5bcb7ac`

ilblackdragon:
# Code Review

## Overview

XL PR (+3,751 / −61) adding an Abound remittance customer integration. Touches multiple subsystems: Responses API streaming, credential injection (new path-scoped matching), session/thread management, integration loading from disk, custom workspace seed, two new skills, and substantial test scaffolding.

The core engineering work (path-scoped credential mapping, Responses API UUID adoption, x_context injection) is reasonable. The integration packaging is pragmatic but has several issues that should be addressed before merge.

---

## 🐛 Bugs

### Duplicate `engine_v2` eager-init block
`src/agent/agent_loop.rs:512-520` adds an eager-init that is **identical to the block immediately above it**:

```rust
        if self.config.engine_v2
            && let Err(e) = crate::bridge::init_engine(&self).await
        {
            tracing::debug!("engine v2: eager init failed: {e}");
        }

+       // Eagerly initialize engine v2 so gateway API endpoints can serve
+       // data (projects, missions, threads) before the first chat message.
+       if self.config.engine_v2
+           && let Err(e) = crate::bridge::init_engine(&self).await
+       {
+           tracing::debug!("engine v2: eager init failed: {e}");
+       }
```

Looks like a bad merge with `staging`. Delete the new block.

### Skill credential names don't match `credentials.json`
`skills/abound-api/SKILL.md` declares credentials:
```yaml
- name: abound_external_token
- name: abound_api_key
```
But `integrations/abound/credentials.json` actually uses `abound_read_token` and `abound_write_token`. The skill manifest is documentation that misleads operators about which secret names to provision. Reconcile to one set.

### `info!` log breaks TUI/REPL
`src/tools/registry.rs:2828` uses `tracing::info!(\"Loaded {} credential mappings ...\")`. CLAUDE.md is explicit: only `debug!` for internal diagnostics; `info!` corrupts the REPL. Change to `debug!`.

### `.gitignore` extra blank lines
Two blank lines added before the new `# tests` section — minor but worth fixing.

---

## 🟡 Code Quality / Project Conventions

### Inline `std::env::var` violates config layering
Three new sites read env vars directly instead of going through `src/config/`:
- `src/llm/reasoning.rs:1031` — `AGENT_PREAMBLE`
- `src/workspace/mod.rs:2003` — `AGENTS_SEED_PATH`
- `src/app.rs:339` — `INTEGRATION_CREDENTIALS_DIR`

Per CLAUDE.md, these belong in `src/config/`. Reading env vars deep inside business logic makes config invisible to `cli/config.rs`, breaks `Config::from_env()` testing, and makes it impossible to override without process restart.

### `AGENT_PREAMBLE` env var is also a behavior surprise
`reasoning.rs:1031` lets any deployment silently overwrite the agent's identity (\"You are IronClaw Agent...\"). The default identity is referenced from SOUL.md/IDENTITY.md and other prompts; replacing it without coordination can cause downstream prompt issues. If we want this, it should be a typed config field with a clear name (e.g. `agent.system_preamble_override`).

### `load_credential_mappings` reinvents serde
`src/tools/registry.rs:2706-2833` is ~125 lines of hand-rolled JSON parsing for what is already a typed `CredentialMapping` struct. With a small `#[derive(Deserialize)]` schema this becomes 10-15 lines and gains free error messages. The current code also silently skips entries with `continue` on parse failure, swallowing config bugs.

### `path_patterns` is misnamed
`src/secrets/types.rs` matches with `path.starts_with(prefix.as_str())`. That's prefix matching, not pattern matching. Either rename to `path_prefixes` or implement real glob support (the field name implies the latter).

### Deeply nested `if let` chain in `app.rs`
`src/app.rs:340-356` is 4 levels of nested `if let`/`if`. Use `let ... else { return; }` or a helper function. Extracting `load_credential_files_from_dir(path)` would also let you replace the hardcoded \"one level of subdirectories\" with `walkdir`.

### Massive duplication between `AGENTS.md` and `SKILL.md`
`integrations/abound/workspace/AGENTS.md` and `skills/abound-api/SKILL.md` repeat the same URLs, payload templates, choice_set rules, and welcome message. Maintaining both will guarantee drift. Pick one source of truth.

### Typo + screaming caps in AGENTS.md
`integrations/abound/workspace/AGENTS.md:18` — `DO NOT RECCOMEND ANY OTHER SERVICES FOR SENDING MONEY!!!` (typo: \"RECCOMEND\"). All-caps shouting in a system prompt is also unnecessary; the rules above already cover it.

### Test numbering in `test_abound_e2e.py`
Tests \"5. Streaming\" and \"6. Choice set\" reuse the section numbers that \"5. Send money advice\" and \"6. Create notification\" already used earlier in the same file. Harmless but confusing in failure logs.

### Dockerfile: `python3-dev` in chef stage
You install `python3-dev` in the `chef` stage. The chef stage is purely for cargo-chef; the pyo3 dependency only matters in the `builder` stage. Moving it should keep the chef cache layer leaner. (`Dockerfile.test` already does the right thing.)

---

## 🟠 Risks

### Guardrails are prompt-only and known to fail
The PR description states **70/79** guardrail tests pass: agent leaks competitor names when declining, breaks character on \"what AI are you?\". These are exactly the cases an adversarial user will hit first. For a demo this is acceptable if scoped, but please:
1. List the failing test IDs in the PR.
2. Open follow-up issues for each.
3. Don't market this as a hardened guardrail system.

### `AGENTS_SEED_PATH` replaces base AGENTS.md entirely
`workspace/mod.rs:2007` reads the custom path and uses it in place of the bundled seed. If the custom file omits sections the rest of the agent expects (e.g. tool usage conventions), behavior degrades silently. Consider concatenation, or at minimum log a warning when the seed is overridden.

### Path prefix match has no normalization
Patterns like `/times/bank/remittance/agent/account/info` would not match `/times/bank/remittance/agent/account/info/` (trailing slash). URLs from `parsed_url.path()` should be deterministic, but worth a few unit tests covering trailing slash, query string, and case sensitivity.

### Custom integration directory loading is silent on failure
Read errors (`std::fs::read_dir` failures, malformed JSON) hit `tracing::warn!` and continue. An empty `mappings:` array means **the agent ships with no Abound credentials and silently no-ops**. Worth a hard-fail mode or a startup health check that asserts the expected mappings loaded.

### Streaming refactor changes event ordering
`responses_api.rs:1025+` now always emits `output_text.delta` before `output_item.done`. Existing clients consuming both events should handle this, but worth calling out in the PR description for downstream consumers.

---

## ✅ Things done well

- **`CredentialMapping::matches()` consolidation**: Centralizing host+path matching on the type itself, then having `find_for_host`/`find_for_url` delegate to it, is clean and the call-site updates in `wasm/credential_injector.rs` are consistent.
- **Responses API UUID adoption** (`session_manager.rs`, `session.rs`): Correct fix for the GET-by-id mismatch. The new `create_thread_with_id` is properly scoped.
- **`requires_preexisting_uuid_thread` carve-out** for `source: \"responses_api\"` (`thread_ops.rs`) is appropriately narrow.
- **`x_context` field** with `alias = \"context\"` and clear comment that it's an IronClaw extension — good API hygiene.
- **`format_context`** has unit tests covering both nested and flat shapes.
- **Test coverage**: comprehensive admin/responses E2E suites added to `tests/e2e/scenarios/`, properly using fixtures and the existing `ironclaw_server` setup.
- **`SECRETS_MASTER_KEY`** added to `conftest.py` is the right fix for admin secret provisioning in E2E.

---

## Suggested next steps

This PR is doing too much at once for a single review/merge cycle. I'd like to propose splitting it and rethinking the integration story:

### 1. Land the Responses API fixes as a separate PR first
The Responses API changes are independently valuable, well-scoped, and have unit/E2E test coverage:
- New thread creation from `responses_api` source (`thread_ops.rs`)
- UUID adoption in `session_manager` / `session` / `Session::create_thread_with_id`
- Streaming delta emission (`responses_api.rs` streaming worker)
- `x_context` field + `format_context` helper
- The `tests/e2e/scenarios/test_responses_api.py` and `test_admin_api.py` suites
- The `SECRETS_MASTER_KEY` fixture fix

These touch only `src/agent/`, `src/channels/web/`, and `tests/e2e/`. They're reviewable in isolation, low-risk, and unblock anyone else who needs the Responses API to work correctly. Cherry-pick into `staging` and merge — it would have already been in production by now.

### 2. Move Abound-specific assets out of the main repo
The `integrations/abound/` directory (credentials mapping, AGENTS.md, tests, scripts) and the `skills/abound-api/` + `skills/forex-timing/` skill files are customer-specific and should not live in `nearai/ironclaw`. Proposed:
- New repo: `nearai/ironclaw-abound` (private)
- Contains: `credentials.json`, `workspace/AGENTS.md`, `skills/abound-api/`, `skills/forex-timing/`, all `tests/`
- The main repo carries no Abound mention at all

This also addresses the duplication and naming-mismatch issues in this PR, since the customer repo becomes the single source of truth.

### 3. Build a first-class \"customer bundle\" mechanism in IronClaw
Rather than the env-var-driven `INTEGRATION_CREDENTIALS_DIR` + `SKILLS_DIR` + `AGENTS_SEED_PATH` triad, an IronClaw admin should be able to point the instance at a git repo and have it manage everything:

```
ironclaw bundle add nearai/ironclaw-abound --ref main
ironclaw bundle list
ironclaw bundle update abound
ironclaw bundle remove abound
```

A bundle would have a known layout (e.g. `bundle.toml` declaring credential mappings, skills, prompts, tools, workspace seeds) and IronClaw would:
- Clone/pull on add or update
- Validate the manifest before activation
- Hot-reload changed assets where safe (skills, prompts) and require restart for others (credential mappings)
- Track which bundle a resource came from (for diagnostics and removal)
- Optionally pin to a commit SHA for reproducible deploys

This generalizes the customer-specific case, keeps the main repo clean, gives admins a real lifecycle (add/update/remove/pin), and makes \"add a second customer\" a single command instead of a Dockerfile change. It also slots cleanly into the existing extension registry concept (`src/registry/`) — bundles could be a higher-level grouping over the existing `ExtensionManifest` machinery.

Once this lands, the Abound-specific bits in this PR move to `nearai/ironclaw-abound` and the deployment becomes:
```
ironclaw bundle add nearai/ironclaw-abound
```

with no env vars, no Dockerfile copies, and no per-customer code in the main repo.

src/channels/web/responses_api.rs
serrrfirat, line 1: **High Severity** [Certain] — `GET /v1/responses/{id}` is no longer actually retrieving by response ID.

This handler decodes both halves of the ID, but only uses `thread_uuid` and reconstructs output from **every message in the conversation**. With the new per-turn `resp_{response_uuid}{thread_uuid}` format, `retrieve(r1.id)` after a second turn returns the whole thread instead of the historical response identified by `r1.id`, and it still reports `status: completed`.

Concrete repro: create `r1`, create `r2` with `previous_response_id=r1.id`, then call `GET /v1/responses/{r1.id}` — you'll get `r2`'s content too.

Suggestion: persist/index a response record by `response_uuid` (or at least store a turn boundary / originating assistant message id) and reconstruct/status the response from that boundary instead of the full thread.

serrrfirat, line 1: **High Severity** [Certain] — `GateRequired` is still matched to this thread, but it is no longer treated as a terminal Responses-API outcome here.

If a tool pauses for auth / approval, the non-streaming handler will keep waiting until the 120s timeout, and the streaming path never sends a deterministic failure/completion boundary for the client. Before this change, the accumulator converted that case into an immediate API failure.

Suggestion: keep `GateRequired` in the terminal set and translate it into a structured Responses-API error (or another explicit paused/error contract) instead of hanging until timeout.

src/channels/web/responses_api.rs
```diff
@@ -974,50 +1025,58 @@ async fn streaming_worker(
...
                         }
-                    }
+                    };
+
+                    // Emit the full text as a delta so streaming clients
+                    // receive it via response.output_text.delta.
+                    emit(
+                        &tx,
+                        "response.output_text.delta",
+                        &ResponseStreamEvent::OutputTextDelta {
```
serrrfirat, line 1101: **High Severity** [Certain] — this now emits the full final text as another `response.output_text.delta` even when earlier stream chunks already delivered the same content.

Any client that builds the answer by concatenating deltas will double the output. Concrete repro: if the stream already emitted `Hello` through chunk deltas and the final response content is also `Hello`, the client sees `HelloHello`.

Suggestion: only emit the final delta when no prior text chunks were sent, or emit the finalized text only in `response.output_item.done`.

src/channels/web/responses_api.rs
serrrfirat, line 1: **High Severity** [Certain] — the Responses API is discarding the real tool `call_id` and correlating results/errors by “most recent tool with this name”.

That breaks same-name tool calls in one turn even though upstream already sends stable IDs (`thread_ops.rs` uses `tool_started_with_id(... Some(tc.id.clone()))`, and tool results carry the same id). With two `memory_search` calls in one turn, arguments/results/errors can attach to the wrong function call depending on ordering.

Suggestion: preserve `AppEvent.call_id` end-to-end and use it as the primary join key. The removed duplicate-name tests were covering exactly this regression.

src/tools/wasm/credential_injector.rs
```diff
@@ -170,33 +170,32 @@ impl SharedCredentialRegistry {
                 poisoned.into_inner()
             }
         };
-        guard.iter().any(|mapping| {
-            mapping
-                .host_patterns
-                .iter()
-                .any(|pattern| host_matches_pattern(host, pattern))
-        })
+        guard.iter().any(|mapping| mapping.matches(host, "/"))
```
serrrfirat, line 173: **Medium Severity** [Likely] — `has_credentials_for_host()` now calls `mapping.matches(host, "/")`.

For mappings that are intentionally path-scoped, `"/"` never matches the configured prefixes, so the host is treated as “uncredentialed” by the approval/manual-header checks even on the sensitive endpoint itself. That means the new `path_patterns` protect injection time (`find_for_url`) but not the earlier gating rules that decide whether manual auth headers are blocked or whether approval is required.

Suggestion: add a path-aware check (for example `has_credentials_for_url(host, path)`) and thread the request path through the approval/header-blocking call sites.

src/tools/builtin/http.rs
```diff
@@ -606,8 +606,9 @@ impl Tool for HttpTool {
             self.secrets_store.as_ref(),
         ) {
             let cred_host = parsed_url.host_str().unwrap_or("").to_string();
+            let cred_path = parsed_url.path();
             let matched: Vec<crate::secrets::CredentialMapping> =
-                registry.find_for_host(&cred_host);
+                registry.find_for_url(&cred_host, cred_path);
```
serrrfirat, line 611: **Medium Severity** [Likely] — now that `find_for_url()` can return multiple mappings for one endpoint, the single `missing_credential` slot becomes lossy.

On Abound-style routes that need both the path-scoped bearer token and `X-API-KEY`, a partially configured deployment can successfully inject the first secret, clear `missing_credential`, then send a partially authenticated request. If the server responds `401/403`, we never raise `authentication_required` for the missing second credential because the missing state was already dropped.

Suggestion: track missing credentials per matched mapping (or at least keep the first missing required credential even after later successful injections) so `401/403` can still open the right auth flow for partially configured endpoints.

serrrfirat:

This branch is a large Abound demo spanning v1 agent/session/tool code, legacy engine-v2, old Responses API, sandbox, setup, deployment, and demo-specific capabilities. It cannot be safely separated or released as a Reborn feature.

Individual Abound requirements can return as small Reborn extensions or ProductWorkflow/OpenAI-compatible integrations. Each should use current capability manifests, credential obligations, Responses API contracts, and production integration tests rather than carrying the demo runtime fork.

serrrfirat:

This is a large Abound demo spanning v1, legacy engine-v2, old Responses API, sandbox, deployment, and demo-specific capabilities. It cannot be safely released as Reborn.

Return individual Abound needs as small Reborn extensions or OpenAI-compatible/ProductWorkflow changes using current manifests, credentials, and integration tests.

PR #1774, opened 2026-03-31
feat: add /thread list and /thread new subcommands

ankinow:
## Summary

This PR completes the /thread command family with two missing subcommands:

### Changes
- **`/thread list`**: New alias for `/history`, lists all persistent + session threads
- **`/thread new`**: Creates a new thread and switches to it
- **`/thread <uuid>`**: Now properly delegates to `process_switch_thread` for DB hydration

### Context
Recent commits enhanced the `/history` command to show both persistent (DB) and in-memory threads with [DB] indicators. The `/thread switch` was also enhanced to hydrate persistent threads from the database.

This PR completes the ergonomics by:
1. Making `/thread list` a documented alias for `/history` (mentioned in help text)
2. Adding `/thread new` as a direct way to create new threads
3. Ensuring `/thread <uuid>` works with hydration for persistent threads

### Testing
- Build passes: `cargo check`
- Follows existing patterns in `src/agent/commands.rs`
- Compatible with existing `Submission::SwitchThread` and `Submission::NewThread` intents

### Related
- Builds on: e23690b (feat: hydrate persistent threads on /thread switch)
- Complements: 1951b92 (feat: add /thread subcommands for better thread ergonomics)


src/agent/commands.rs
```diff
@@ -106,6 +128,110 @@ impl Agent {
...
+        output.push_str(&format!("Session: {}\n", session_id));
+        if let Some(active) = active_thread {
+            output.push_str(&format!("Active thread: {}\n", active));
+        }
+
+        let mut listed_any = false;
+        let mut seen_threads = std::collections::HashSet::new();
+
+        if let Some(store) = tenant.store() {
+            match store.list_conversations_all_channels(50).await {
```
serrrfirat: **Medium Severity**

/history is now documented as listing all threads, and the parser even accepts extra args like /history all, but this implementation still hard-caps the persistent result set at 50 and ignores the parsed args entirely. Once a user has more than 50 conversations, older persisted threads disappear from the new discovery flow, so /thread <id> is no longer realistically reachable for them.

Please either make the UX/documentation explicit that this is only the most recent 50 threads, or plumb the new command arguments through and add coverage for a >50-thread history case.

src/agent/thread_ops.rs
```diff
@@ -1798,16 +1800,51 @@ impl Agent {
...
+                    target_thread_id
+                )));
+            }
+        }
+
+        // Thread not in memory — try to hydrate from persistent store
+        if let Some(store) = self.store() {
+            // Verify the thread exists and belongs to this user
+            let owned = store
+                .conversation_belongs_to_user(target_thread_id, &message.user_id)
```
serrrfirat, line 1819: **Low Severity**

This new persisted-thread switch path collapses database errors into false with unwrap_or(false), so a transient store failure now surfaces to the user as Thread not found instead of an actual error. That makes the new hydration path much harder to debug and can send users looking in /history even though the backing store is the real problem.

Please preserve the error path here, and add a test that verifies a store failure does not get reported as a missing thread.

src/agent/submission.rs
```diff
@@ -149,29 +149,66 @@ impl SubmissionParser {
...
+                .map(|s| s.to_string())
+                .collect();
+            return Submission::SystemCommand {
+                command: "history".to_string(),
+                args,
+            };
+        }
+
+        // /thread [subcommand] - thread management
+        if let Some(rest) = trimmed.strip_prefix("/thread") {
```
serrrfirat: **Medium Severity**

These new tests imply /thread subcommands are case-insensitive, but that is not true end-to-end. This parser branch only runs when the raw input starts with lowercase /thread; /THREAD LIST, /THREAD NEW, or /THREAD <uuid> fall through to normal command routing instead. The router lowercases only the command token, not the args, and handle_system_command only matches exact list/new, so those uppercase forms currently fail.

Please either normalize the /thread prefix here as well, or drop the case-insensitive expectations and add an integration-style test that exercises the full command path rather than only the parser.

zmanian (changes requested):
## Review -- REQUEST_CHANGES

The thread subcommand UX is good and test coverage is solid. However:

### Medium

**M1. Dead code: `/thread new` handler in `commands.rs`** -- The `"new"` branch in `handle_system_command` is unreachable because `SubmissionParser` parses `/thread new` as `Submission::NewThread` (handled in `agent_loop.rs` via `process_new_thread()`). The dead code also creates threads differently from the production path. Remove it.

**M2. Dead code: unknown subcommand error** -- `/thread foo` falls through to `UserInput` in the parser, never reaching the `handle_system_command` error message for unknown subcommands. Either adjust parsing or remove the dead branch.

**M3. Scope creep** -- PR bundles: AGENTS.md reformatting, new ROUTINES.md seed, scheduler ApprovalContext propagation, webhook test fix, .gitignore addition, CLAUDE.md doc expansion. The scheduler change especially (ApprovalContext propagation to subtasks) is a behavioral change deserving its own PR.

**M4. ROUTINES.md is in Portuguese** -- All other docs and seed files are in English.

### Low
- Double-pipe `||` table formatting corruption in `src/agent/CLAUDE.md`
- `handle_system_command` signature broadened to include `session` for all commands when only thread/history need it

PR #1819, opened 2026-03-31
fix: image generation with nearai models

### LINKED ISSUE #1810: [Bug] Image generation failed
Image generation will fail due to issues:
1. Wrong generated image model url (double `/v1`)
2. UI render fail

hanakannzashi:
## Summary

- Fix image generation against NearAI STG by avoiding `/v1/v1/...` URL construction and switching the fallback image model to `black-forest-labs/FLUX.2-klein-4B`.
- Fix generated-image handling end to end: detect image sentinels reliably, emit `image_generated` events, and avoid feeding raw base64/data URLs back into the model.
- Persist generated-image payloads in tool-call history and rebuild them into `generated_images` so images survive page refresh and render as standalone UI blocks.
- Fix generated image rendering in the web UI so image cards do not collapse when later assistant text is added.

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] CI/Infrastructure
- [ ] Security
- [ ] Dependencies

## Linked Issue

Fixes #1810

## Validation

- [x] `cargo build`
- [x] Relevant tests pass:
  - `cargo test -q test_build_turns_collects_generated_images_from_persisted_tool_results --lib`
  - `cargo test -q test_rebuild_chat_messages_summarizes_image_generated_result --lib`
  - `cargo test -q test_image_generation_summary_tool_message_omits_data_url --lib`
  - `cargo test -q test_parse_image_generated_sentinel_accepts_stringified_json --lib`
  - `cargo test -q parses_double_stringified_sentinel --lib`
- [x] Manual testing:
  - verified successful image generation against NearAI STG
  - verified assistant text no longer includes embedded ``
  - verified generated images are returned as standalone UI blocks
  - verified generated images persist after page refresh

## Security Impact

None. No auth, secrets, sandbox, or permission model changes. This PR does stop raw generated-image base64 from being echoed back into model-visible tool results.

## Database Impact

None. No schema or migration changes. This does change persisted `tool_calls.result` content for generated-image tool results so history can reconstruct images, but it remains within existing JSON/text storage and is compatible with both PostgreSQL and libSQL.

## Blast Radius

Touches image tool execution, agent tool-result handling, persisted tool-call history, and web chat/history rendering. Regressions would most likely affect generated-image display, tool-result previews, or refresh/history reconstruction for image turns.

## Review Follow-Through

- Current image-model selection still relies on inferring from the active chat model name plus a hardcoded fallback.
- We should add an explicit manually configurable image model setting, but that is intentionally out of scope for this PR.

---

**Review track**: B


zmanian (commented):
## Review -- COMMENT (leaning APPROVE)

Well-structured fix for image persistence and LLM context pollution. Good test coverage (7 new tests). URL construction fix is correct. Both DB backends work without schema changes.

### Medium (worth addressing)
- **Unbounded image storage in DB**: `tool_result_content_for_persistence` stores full base64 data URLs (1-5 MB each) with no truncation, unlike other tool results (1000 char limit). Acknowledge the storage growth implication or add a size cap.
- **Duplicated `endpoint_url()`**: Exact same logic in `ImageGenerateTool` and `ImageEditTool`. Extract to a shared helper.

### Low (non-blocking)
- Frontend `generatedImagesByThread` Map grows unbounded (cleared on page refresh)
- DOM dedup compares full data URL strings (a Set of IDs would be cleaner)
- `normalize_embedded_json` limit of 3 parse rounds is undocumented

zmanian (dismissed):
## Re-review -- APPROVE

The three new commits (`dce9e8cb`, `a02a87bc`, `89b3a911`) address all medium-priority items from the previous review and most low-priority ones:

### Previously raised -- status

| Item | Status |
|------|--------|
| Unbounded image storage in DB | Fixed -- `MAX_PERSISTED_IMAGE_SENTINEL_BYTES` (8 MiB) cap with clear fallback message |
| Duplicated `endpoint_url()` | Fixed -- extracted to `image_api_endpoint_url()` in `src/tools/builtin/mod.rs` |
| Frontend `generatedImagesByThread` unbounded | Fixed -- LRU eviction with `GENERATED_IMAGE_THREAD_CACHE_CAP = 20` / `GENERATED_IMAGES_PER_THREAD_CAP = 8` |
| Scroll-per-image during history load | Fixed -- `addGeneratedImage(url, path, false)` + single scroll after loop |
| `normalize_embedded_json` limit undocumented | Fixed -- comment explaining the 3-round cap |
| Pagination not calling `rememberGeneratedImage` | Fixed -- older-history path now calls `rememberGeneratedImage` |
| Redundant `renderGeneratedImagesForCurrentThread()` | Removed; replaced with a single `container.scrollTop = container.scrollHeight` |
| DOM dedup via full data URL string comparison | Not addressed (low priority, acceptable for now) |

### New changes look correct

- `ironclaw_safety::SafetyLayer` import migration in `dispatcher.rs` follows the project convention (CLAUDE.md: "new code should import from `ironclaw_safety` directly").
- Module declaration ordering fix in `lib.rs` is cosmetic and correct.
- The merge from staging resolved cleanly based on the diff.

### One minor observation (non-blocking)

`infer_generated_image_media_type` decodes only the first 64 *characters* of base64 (not 64 bytes). Due to base64 encoding ratios, this yields ~48 decoded bytes, which is enough for all four magic-number checks (JPEG needs 3, PNG needs 8, GIF needs 6, WebP needs 12). Just noting this is fine as-is.

Good work addressing the feedback thoroughly.

think-in-universe (dismissed):
LGTM.

Tested image generation locally and it works well for me.

Image

think-in-universe:
Please notice the PR only fixes image generation with NEAR AI models.

We need a follow-up PR that fixes image generation for other model providers.

src/agent/dispatcher.rs
```diff
@@ -1180,11 +1185,11 @@ impl<'a> LoopDelegate for ChatDelegate<'a> {
                             && let Some(turn) = thread.last_turn_mut()
                         {
                             if is_tool_error {
-                                turn.record_tool_error_for(&tc.id, result_content.clone());
+                                turn.record_tool_error_for(&tc.id, record_content.clone());
```
serrrfirat, line 1239: **High Severity**
This still stores the full `image_generated` sentinel in thread state (`output.clone()`), even though the LLM-facing tool message has been summarized. On later turns, `Thread::messages()` replays `tc.result` back into `ChatMessage::tool_result`, so the raw `data:image/...;base64,...` payload still gets injected into model context for live in-memory threads.

That means the fix only holds for the current tool-result hop and for DB-rebuilt history, not for the next turn in the same session.

Suggested fix: summarize or strip `image_generated` payloads before storing them in thread state too, or update the session replay path to summarize them consistently. A regression test should assert that a second turn after image generation does not expose `data:image` to the model.

serrrfirat (commented):
Posting the highest-severity issue inline; the broader response/history concerns are added as top-level PR comments.

serrrfirat:
**Medium Severity**
Persisting up to 8 MiB of raw data URL per generated image is large enough to make chat history responses pathological.

A default history page can include many turns, and each turn can now contribute multi-megabyte inline payloads. The per-image cap bounds a single row, but it does not bound total bytes returned by `/api/chat/history` or rendered into the DOM. In practice this creates a realistic path to very large JSON responses, slow page loads, and browser memory spikes.

Suggested fix: persist a path/handle plus metadata instead of the full data URL, or enforce a much smaller per-image cap together with a total per-response budget. A regression test should cover many image turns and verify history payloads stay bounded.

serrrfirat:
**Medium Severity**
The new generated-image dedupe logic suppresses any card whose `(dataUrl, path)` matches one already rendered.

That avoids replay duplicates, but it also collapses legitimate repeated generations across different turns when the model returns the same bytes twice. In that case the UI loses a real conversation event and history no longer reflects what actually happened.

Suggested fix: dedupe by turn/event identity rather than payload equality. A regression test should cover two distinct turns with identical generated image payloads and verify both cards render.

src/agent/thread_ops.rs
```diff
@@ -20,11 +20,52 @@ use crate::agent::submission::SubmissionResult;
...
+        // Persist the full image sentinel so web history can reconstruct the
+        // generated image on refresh without any schema changes. Keep a hard
+        // cap so unexpectedly large data URLs do not grow DB rows without bound.
+        let serialized = sentinel.value.to_string();
+        if serialized.len() <= MAX_PERSISTED_IMAGE_SENTINEL_BYTES {
+            return serialized;
+        }
+        return format!(
+            "Generated image omitted from persistence because it exceeded the {} MiB cap",
+            MAX_PERSISTED_IMAGE_SENTINEL_BYTES / (1024 * 1024)
```
serrrfirat: **Medium Severity — Bug: integer division truncates to zero**

`MAX_PERSISTED_IMAGE_SENTINEL_BYTES` is `512 * 1024 = 524288`. Dividing by `1024 * 1024 = 1048576` with Rust integer division yields **0**, so the error message reads:

> Generated image omitted from persistence because it exceeded the **0 MiB** cap

Suggested fix — either use KiB:
```rust
return format!(
    "Generated image omitted from persistence because it exceeded the {} KiB cap",
    MAX_PERSISTED_IMAGE_SENTINEL_BYTES / 1024
);
```
or just use a string literal since the constant is known at compile time:
```rust
"Generated image omitted from persistence (exceeds 512 KiB cap)"
```

src/channels/web/handlers/chat.rs
```diff
@@ -10,7 +10,8 @@ use crate::channels::web::auth::AuthenticatedUser;
 use crate::channels::web::server::GatewayState;
 use crate::channels::web::types::*;
 use crate::channels::web::util::{
```
serrrfirat: **Medium Severity — Maintenance: duplicate chat_history_handler**

This PR adds the same \`collect_generated_images_from_tool_results\` + \`enforce_generated_image_history_budget\` logic to both \`src/channels/web/handlers/chat.rs\` and \`src/channels/web/server.rs\`. The file header in \`handlers/chat.rs\` (lines 3–5) explicitly says:

> The primary chat handlers (…chat_history_handler) live in server.rs where routes are registered. Do NOT add duplicates here.

The route is wired in \`server.rs\`, so the \`handlers/chat.rs\` version appears to be dead code that is being maintained in parallel. If it's no longer used, removing it would eliminate the duplication risk. If it's the migration target, a TODO to retire the \`server.rs\` copy would help.

src/generated_images.rs
```diff
@@ -0,0 +1,99 @@
...
+            && !path.is_empty()
+        {
+            return format!("Generated image ({media_type}) at {path}");
+        }
+        format!("Generated image ({media_type})")
+    }
+}
+
+fn normalize_embedded_json(value: &serde_json::Value) -> Option<serde_json::Value> {
+    let mut current = value.clone();
```
serrrfirat: **Low Severity — Performance: unnecessary clone of potentially large values**

\`normalize_embedded_json\` clones the entire \`serde_json::Value\` on entry (line 49). For image sentinels carrying 512 KB base64 data URLs, this is a non-trivial heap allocation — and this function is called on every tool result in several hot paths (\`collect_generated_images_from_tool_results\`, \`tool_result_preview\`, \`session::messages()\`).

A quick win: only clone when you need to mutate (i.e., when the value is a \`String\` that needs unwrapping). When the input is already an \`Object\`, you can return a clone of just that outer layer or take ownership:

```rust
fn normalize_embedded_json(value: &serde_json::Value) -> Option<serde_json::Value> {
    if !value.is_string() {
        return Some(value.clone()); // Object/Array — cheap-ish, no reparsing
    }
    let mut current = value.clone();
    for _ in 0..3 {
        // ...
    }
}
```

Not blocking, but worth a note since the same path runs on every tool result during history loading.

src/tools/builtin/image_gen.rs
```diff
@@ -54,6 +56,31 @@ impl ImageGenerateTool {
             client,
         }
     }
+
+    fn endpoint_url(&self, path: &str) -> String {
+        image_api_endpoint_url(&self.api_base_url, path)
+    }
+}
+
+pub(super) fn infer_generated_image_media_type(image_b64: &str) -> &'static str {
```
serrrfirat, line 65: **Low Severity — Test coverage: only JPEG and PNG tested**

\`infer_generated_image_media_type\` handles 4 formats (JPEG, PNG, GIF, WEBP) plus invalid/empty input, but only JPEG and PNG have tests. The WEBP branch has a bounds-check (\`bytes.len() >= 12\`) that's easy to get wrong and deserves explicit coverage.

Suggested additions:
```rust
#[test]
fn infer_generated_image_media_type_detects_webp() {
    let mut riff = b"RIFF".to_vec();
    riff.extend_from_slice(&[0x00; 4]); // file size placeholder
    riff.extend_from_slice(b"WEBP");
    let webp_b64 = STANDARD.encode(&riff);
    assert_eq!(infer_generated_image_media_type(&webp_b64), "image/webp");
}

#[test]
fn infer_generated_image_media_type_defaults_for_invalid_base64() {
    assert_eq!(infer_generated_image_media_type("!!!not-base64"), "image/png");
}

#[test]
fn infer_generated_image_media_type_defaults_for_empty_input() {
    assert_eq!(infer_generated_image_media_type(""), "image/png");
}
```

src/tools/builtin/mod.rs
```diff
@@ -62,3 +62,17 @@ pub(crate) fn media_type_from_path(path: &str) -> String {
...
 }
+
+/// Build an OpenAI-style image endpoint from a provider base URL.
+///
+/// Some providers already include `/v1` in their configured base URL while
+/// others expect clients to append it. Keep this logic shared so image tools
+/// do not drift.
+pub(crate) fn image_api_endpoint_url(api_base_url: &str, path: &str) -> String {
+    let base = api_base_url.trim_end_matches('/');
+    if base.ends_with("/v1") {
```
serrrfirat: **Low Severity — Robustness: doesn't guard against pre-doubled \`/v1/v1\`**

The whole purpose of this function is to prevent \`/v1/v1/...\` URLs, but if the base URL is already misconfigured as \`\"https://api.example.com/v1/v1\"\`, the suffix check passes (\`ends_with(\"/v1\")\` is true) and the doubled path is preserved. Unlikely in practice, but worth a comment or a more robust strip:

```rust
// Strip all trailing /v1 segments, then add exactly one.
let mut base = api_base_url.trim_end_matches('/').to_string();
while base.ends_with("/v1") {
    base.truncate(base.len() - 3);
}
format!("{base}/v1{path}")
```

Or at minimum, document the assumption: *"Assumes the base URL has at most one trailing \`/v1\` segment."*

src/channels/web/util.rs
```diff
@@ -24,6 +28,62 @@ fn parse_tool_call_infos(calls: &[serde_json::Value]) -> Vec<ToolCallInfo> {
...
+}
+
+pub fn collect_generated_images_from_tool_results<'a>(
+    turn_number: usize,
+    tool_results: impl IntoIterator<Item = (Option<&'a str>, Option<&'a serde_json::Value>)>,
+) -> Vec<GeneratedImageInfo> {
+    tool_results
+        .into_iter()
+        .enumerate()
+        .filter_map(|(image_index, (event_id, result))| {
```
serrrfirat: **Nit — Naming: \`image_index\` is the index across all tool results**

\`enumerate()\` here counts every tool result (image or not), so \`image_index\` is really the position in the full tool-result iterator, not among image-generating tools. Consider \`tool_index\` or \`result_index\` to avoid confusion when reading the fallback ID format (\`turn-{n}-image-{i}\`).

serrrfirat:
## Paranoid Architect Review — Summary

**Verdict: Solid PR with one concrete bug and a few polish items.**

Reviewed all 19 changed files (974+/96−), read every changed file in full context.

### Findings

| # | Severity | File | Finding |
|---|----------|------|---------|
| 1 | **Medium** | `thread_ops.rs:52` | Integer division truncates to 0 — error message says "0 MiB cap" instead of 512 KiB |
| 2 | **Medium** | `handlers/chat.rs` | Duplicate `chat_history_handler` maintained in parallel with `server.rs`; file header says "Do NOT add duplicates here" |
| 3 | **Low** | `generated_images.rs:49` | `normalize_embedded_json` clones full ~512KB sentinel values on every tool result in hot paths |
| 4 | **Low** | `image_gen.rs:65` | `infer_generated_image_media_type` only tests JPEG/PNG — missing GIF, WEBP, invalid, and empty input |
| 5 | **Low** | `mod.rs:73` | `image_api_endpoint_url` doesn't guard against pre-doubled `/v1/v1` in base URL |
| 6 | **Nit** | `util.rs:66` | `image_index` name misleading — it's the index across all tool results, not just images |

### What looks good

- **Data URL isolation from LLM context** is thorough across all 4 code paths (dispatcher summary message, session replay, DB rebuild, context re-injection). No base64 leaks into model input.
- **Budget enforcement** correctly prioritizes newest images (reverse iteration), caps per-image (512 KB) and per-response (1 MB).
- **Frontend dedup** via `data-image-event-id` prevents double-rendering during SSE + history race.
- **JS LRU cache** (20 threads × 8 images) with insertion-order refresh is correct.
- **No XSS vectors** — `textContent` for path labels, safe DOM API for dataset, `` for data URLs.
- **Sentinel parsing** handles double/triple stringification with bounded unwrapping (3 iterations max).
- **17 new unit tests** provide strong coverage of the core sentinel, persistence, budget, and URL logic.
- **`infer_generated_image_media_type`** safely falls back to `image/png` for invalid/partial base64.

No critical or security findings. Finding #1 is the only concrete bug — the rest are robustness and maintenance suggestions.

serrrfirat:
## Paranoid Architect Review — `fix: image generation with nearai models`

Well-structured, well-tested bug fix PR. Sentinel centralization, budget enforcement, and dedup design are sound. 11 new tests with good edge case coverage. Four MEDIUM findings:

### 1. MEDIUM — Client-side data URL scheme not validated

**File:** `src/channels/web/static/app.js` (`createGeneratedImageElement`)

`img.src = dataUrl` is set directly from the server response without validating that the URL uses the `data:image/` scheme. A compromised or malicious API response could inject a `javascript:`, `blob:`, or external `https:` URL. While `` tags don't execute `javascript:` URIs in modern browsers, an external URL could be used for tracking or SSRF-like probing from the client side.

**Suggested fix:** Add a guard before setting `img.src`:
```javascript
if (!dataUrl || !dataUrl.startsWith('data:image/')) return;
```

### 2. MEDIUM — Fragile URL version suffix check

**File:** `src/tools/builtin/mod.rs:~76` (`image_api_endpoint_url`)

`image_api_endpoint_url` only checks `ends_with("/v1")`. A provider URL like `https://api.com/v1beta` would not match and would get `/v1` appended, producing `/v1beta/v1/images/...`. Similarly, a future `/v2` endpoint would silently get `/v1` appended.

**Suggested fix:** Use a URL parser to inspect the last path segment, or at minimum document the assumption. Consider matching against `/v[0-9]+` at the end of the path.

### 3. MEDIUM — Silent data loss on triple-stringified sentinel

**File:** `src/generated_images.rs:~87` (`normalize_embedded_json`)

The `for _ in 0..2` loop unwraps at most 2 layers of stringified JSON. If the value is triple-stringified (3+ layers), the loop exhausts and the image silently disappears from history with no warning. While unlikely, this makes debugging very difficult if it ever occurs.

**Suggested fix:** Add a `tracing::debug!` when the loop exhausts without reaching a non-String value.

### 4. MEDIUM — Full base64 image payload retained in memory for thread lifetime

**File:** `src/agent/dispatcher.rs:~1168`

When an image sentinel is detected, `record_content` stores the full raw tool output including the base64 data URL in the in-memory `Thread`. For large images (several MB), this increases memory pressure on long-lived sessions. The persistence layer correctly enforces a 512 KiB cap, but the in-memory representation does not. With 8 images per thread at several MB each, a single session could hold tens of MB of base64 in memory.

**Suggested fix:** Document the memory tradeoff. Consider applying the same 512 KiB size cap to the in-memory `record_content` (replace oversized base64 with the summary sentinel), or add eviction for in-memory sentinels. Not a blocker, but worth noting for memory-constrained deployments.

src/channels/web/server.rs
```diff
@@ -2313,20 +2317,21 @@ async fn chat_history_handler(
                         name: tc.name.clone(),
```
serrrfirat: **Medium Severity**\n\nThe in-memory history path returns `tc.error` directly here, but those values still contain the safety wrapper produced by `process_tool_result()`. The DB-backed history path unwraps tool errors before returning them, so active-thread history and refreshed history will render different error text for the same turn.\n\nPlease run the same display unwrapping on the in-memory path so both history sources return consistent error content.

src/agent/thread_ops.rs
```diff
@@ -21,13 +21,55 @@ use crate::agent::submission::SubmissionResult;
...
+fn tool_result_content_for_persistence(result: &serde_json::Value) -> String {
+    if let Some(sentinel) = GeneratedImageSentinel::from_value(result) {
+        // Persist the full image sentinel so web history can reconstruct the
+        // generated image on refresh without any schema changes. Keep a hard
+        // cap so unexpectedly large data URLs do not grow DB rows without bound.
+        let serialized = sentinel.value.to_string();
+        if serialized.len() <= MAX_RECORDED_IMAGE_SENTINEL_BYTES {
+            return serialized;
+        }
+        return format!(
```
serrrfirat: **Medium Severity**\n\nOversized generated-image sentinels are persisted here as plain text instead of the compact `image_generated` JSON shape used elsewhere. That breaks the new refresh/history reconstruction path for large images: `build_turns_from_db_messages()` only rebuilds `generated_images` when it can still recognize a sentinel, so large images will disappear after refresh instead of rendering a placeholder/path card.\n\nPlease persist the compact sentinel JSON (for example via the shared sentinel helper) rather than a plain string, and add a regression test that starts from a capped persisted tool result and verifies history still reconstructs the image metadata.

serrrfirat (changes requested):
Requesting changes for two correctness regressions introduced by this PR:\n\n1. Oversized generated-image results are persisted as plain text instead of a compact sentinel object, which breaks refresh/history reconstruction for large images.\n2. The in-memory web history path returns wrapped tool-error content directly, so active-thread history renders different error text than the DB-backed history path.\n\nI left inline comments on the relevant lines with concrete fixes.

src/tools/builtin/image_gen.rs
```diff
@@ -54,6 +56,35 @@ impl ImageGenerateTool {
...
+    let prefix_len = image_b64.len().min(64);
+    let decodable_len = prefix_len - (prefix_len % 4);
+    if decodable_len == 0 {
+        return "image/png";
+    }
+    let prefix = &image_b64[..decodable_len];
+    let decoded = STANDARD.decode(prefix);
+    let Ok(bytes) = decoded else {
```
serrrfirat, line 73: **Medium Severity — byte-index slicing on potentially non-ASCII string**

`infer_generated_image_media_type` uses `&image_b64[..decodable_len]` which slices at a byte index. While base64 from the API is ASCII-only in practice, CLAUDE.md explicitly forbids byte-index slicing on external strings (`&s[..n]`) because it panics on multi-byte UTF-8. Since this function accepts an arbitrary `&str`, a defensive fix would be:

```rust
let prefix = image_b64.get(..decodable_len).unwrap_or("");
```

Non-blocking — just a robustness nit given the codebase rule.

src/generated_images.rs
```diff
@@ -0,0 +1,207 @@
...
+        serde_json::Value::Object(summary)
+    }
+
+    fn content_with_omitted_data_url_when_oversized(&self, output: &str) -> String {
+        if output.len() <= MAX_RECORDED_IMAGE_SENTINEL_BYTES {
+            return output.to_string();
+        }
+        self.compact_value_without_data_url().to_string()
+    }
+
```
serrrfirat, line 102: **Medium Severity — size cap checks raw output, not normalized sentinel**

`record_content_for_persistence` and `record_content_for_thread_state` compare `output.len()` against `MAX_RECORDED_IMAGE_SENTINEL_BYTES`, but `output` is the raw tool output string. When the sentinel was double-stringified, `output` includes extra escape characters making it larger than the actual JSON payload. This means the cap could trigger prematurely on a moderately-sized image that happened to be double-stringified.

In practice this is conservative (caps sooner, not later), so non-blocking. But for consistency you could check `self.value.to_string().len()` instead.

src/agent/dispatcher.rs
```diff
@@ -1214,12 +1209,26 @@ impl<'a> LoopDelegate for ChatDelegate<'a> {
...
+                            ),
+                        )
+                    } else {
+                        crate::tools::execute::process_tool_result(
+                            self.agent.safety(),
+                            &tc.name,
+                            &tc.id,
+                            &tool_result,
+                        )
+                    };
```
serrrfirat: **Medium Severity — raw output vs normalized sentinel persisted inconsistently**

`image_generation_record_content` calls `sentinel.record_content_for_thread_state(output)` where `output` is the raw tool result (which includes `prompt`, `size`, and other fields beyond the sentinel schema). When the output is under the cap, the raw output is persisted verbatim. When over the cap, a compact sentinel is persisted instead.

This means persisted content sometimes has extra fields (`prompt`, `size`) and sometimes doesn't. Harmless — the sentinel parser ignores unknown fields — but slightly inconsistent. Consider always persisting `sentinel.value.to_string()` for uniformity.

Non-blocking.

src/channels/web/util.rs
```diff
@@ -123,6 +207,25 @@ pub fn build_turns_from_db_messages(
...
+        for image in turn.generated_images.iter_mut().rev() {
+            let Some(data_url) = image.data_url.as_ref() else {
+                continue;
+            };
+            let data_url_bytes = data_url.len();
+            if data_url_bytes > MAX_HISTORY_IMAGE_DATA_URL_BYTES_PER_IMAGE
+                || data_url_bytes > remaining_bytes
+            {
+                image.data_url = None;
+                continue;
```
serrrfirat, line 222: **Medium Severity — use saturating_sub for future-proofing**

The subtraction `remaining_bytes -= data_url_bytes` is safe with the current constant relationship (`PER_IMAGE=512KB < PER_RESPONSE=1MB`), but would underflow if someone later changes the per-image cap to exceed the per-response cap. A one-character fix:

```rust
remaining_bytes = remaining_bytes.saturating_sub(data_url_bytes);
```

Non-blocking — purely defensive.

crates/ironclaw_gateway/static/app.js
```diff
@@ -1349,8 +1361,76 @@ function addGeneratedImage(dataUrl, path) {
...
+
+function isSafeGeneratedImageDataUrl(dataUrl) {
+  return typeof dataUrl === 'string' && /^data:image\//i.test(dataUrl);
+}
+
+function hasRenderedGeneratedImage(container, eventId) {
+  if (!eventId) return false;
+  return Array.from(container.querySelectorAll('.generated-image-card')).some((card) => {
+    return card.dataset.imageEventId === eventId;
+  });
```
serrrfirat, line 1375: **Low Severity — SVG data URLs could be scriptable**

`isSafeGeneratedImageDataUrl` allows any `data:image/...` prefix, which includes `data:image/svg+xml;base64,...`. A crafted SVG in an `` tag can execute JavaScript in some browser contexts (though most modern browsers block it in `` specifically).

Since the data URL comes from the server's own API (not user input), this is defense-in-depth. Consider restricting to known raster types:

```js
return typeof dataUrl === 'string' && /^data:image\/(png|jpeg|gif|webp)[;,]/i.test(dataUrl);
```

Non-blocking.

serrrfirat (dismissed):
Thorough fix for image generation — well structured with good test coverage (20+ new tests). The sentinel abstraction, double-stringification handling, and history budget enforcement are all solid. Five non-blocking comments posted as suggestions for future hardening. Ship it.

think-in-universe (approved):
LGTM

PR #1844, opened 2026-03-31
Feat session guard timer

### LINKED ISSUE #1219: Add SessionGuardTimer to detect lock-held-across-await violations in debug builds
## Problem / Motivation

A documented invariant states: "Session lock must not be held across await boundaries." Holding an `Arc<Mutex<Session>>` lock across an `.await` point can cause deadlocks if another task needs the same session. This invariant is enforced by code review alone — there's no runtime detection mechanism to catch violations during development or testing.

## Proposed Solution

Create a `SessionGuardTimer` wrapper that, in debug builds, tracks how long a session lock is held and logs a warning if it exceeds a threshold (5 seconds). In release builds, it compiles down to a zero-cost wrapper.

### New type in `src/agent/session_manager.rs`

```rust
/// Wrapper around MutexGuard<Session> that warns on long lock holds in debug builds.
pub struct SessionGuardTimer<'a> {
    guard: tokio::sync::MutexGuard<'a, Session>,
    #[cfg(debug_assertions)]
    acquired_at: std::time::Instant,
    #[cfg(debug_assertions)]
    context: &'static str,
}

impl<'a> SessionGuardTimer<'a> {
    pub fn new(guard: tokio::sync::MutexGuard<'a, Session>, context: &'static str) -> Self {
        Self {
            guard,
            #[cfg(debug_assertions)]
            acquired_at: std::time::Instant::now(),
            #[cfg(debug_assertions)]
            context,
        }
    }
}

impl<'a> std::ops::Deref for SessionGuardTimer<'a> {
    type Target = Session;
    fn deref(&self) -> &Session { &self.guard }
}

impl<'a> std::ops::DerefMut for SessionGuardTimer<'a> {
    fn deref_mut(&mut self) -> &mut Session { &mut self.guard }
}

#[cfg(debug_assertions)]
impl<'a> Drop for SessionGuardTimer<'a> {
    fn drop(&mut self) {
        let held_for = self.acquired_at.elapsed();
        if held_for > std::time::Duration::from_secs(5) {
            tracing::warn!(
                held_ms = held_for.as_millis(),
                context = self.context,
                "Session lock held for >5s — possible lock-across-await violation"
            );
        }
    }
}
```

### Update lock acquisition sites

Replace `session.lock().await` with `SessionGuardTimer::new(session.lock().await, "context_string")` at key sites in:
- `src/agent/session_manager.rs` — `resolve_thread()`, thread creation
- `src/agent/thread_ops.rs` — turn processing
- `src/agent/agent_loop.rs` — session state checks

## Code Pointers

- `src/agent/session_manager.rs` — `SessionManager` struct, `get_or_create_session()`, `resolve_thread()`
- `src/agent/thread_ops.rs` — Thread operations that acquire session locks
- `src/agent/agent_loop.rs` — Main agent loop lock acquisitions

## Acceptance Criteria

- [ ] `SessionGuardTimer` implemented with `Deref`/`DerefMut` to `Session`
- [ ] `#[cfg(debug_assertions)]` guarding the timer fields and `Drop` impl
- [ ] All major session lock acquisition sites wrapped with `SessionGuardTimer`
- [ ] Context string identifies the call site (e.g., `"resolve_thread"`, `"start_turn"`)
- [ ] In debug builds: held >5s produces `tracing::warn!`
- [ ] In release builds: zero runtime overhead (no `Instant::now()`, no `Drop` check)
- [ ] All existing tests pass
- [ ] No false positives in normal test runs (if warnings appear, investigate — they might be real)

## Pitfalls & Landmines

- `tokio::sync::MutexGuard` is `!Send`. `SessionGuardTimer` wrapping it is also `!Send`, which is correct — it should not be held across `.await` points. If the compiler complains about `SessionGuardTimer` crossing an await, that's the *detection working* — fix the code, don't make the guard `Send`.
- The 5-second threshold is generous. Normal lock holds should be sub-millisecond. If tests trigger warnings at 5s, there's a real problem.
- Don't wrap *every* lock acquisition — focus on the session locks in the agent module. Other `Mutex` uses (like `UndoManager`) are less critical.

## Non-Goals

- Runtime deadlock detection (tokio has `tokio-console` for that)
- Compile-time enforcement of lock discipline (Rust's type system can't prevent holding a MutexGuard across await in all cases)
- Lock ordering enforcement (not needed — IronClaw uses single-lock-per-section discipline)

ShadowCorp-Dev:
## Summary

- SessionGuardTimer wraps session lock acquisitions and warns in debug builds if held >5s. Release builds pay nothing (#[cfg(debug_assertions)]).
- Wrapped 40 production lock sites in session_manager.rs, dispatcher.rs, and thread_ops.rs. Each site gets a context string so the warning tells you where.
- Found and fixed 4 lock-across-await violations in thread_ops.rs while wrapping (session lock held during persist_assistant_response, get_undo_manager, send_status, etc.)
- 2 compact sites still hold the lock across compact().await because it needs &mut Thread. Splitting compact() into prepare/apply is a separate refactor.

## Change Type

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] CI/Infrastructure
- [ ] Security
- [ ] Dependencies

## Linked Issue

Closes #1219

## Validation

- [x] cargo fmt
- [x] cargo clippy --all --benches --tests --examples --all-features
- [x] Relevant tests pass: cargo test --lib agent:: (416 passed)
- [x] Manual testing: deployed to running ironclaw instance, ran basic chat, tool calls, rapid messages, undo/redo, /compact, /new, and a 24-turn conversation. Zero session guard warnings fired, no false positives.

## Security Impact

None

## Database Impact

None

## Blast Radius

4 files in src/agent/. session_manager.rs gets the new type + 3 wrapped sites, dispatcher.rs gets 6, thread_ops.rs gets 31 plus the 4 violation fixes, mod.rs adds the re-export. Every change wraps an existing lock().await call, no behavior change.

## Rollback Plan

Single commit revert. SessionGuardTimer is transparent (Deref/DerefMut to Session), so removing it just goes back to raw lock().await. The 4 violation fixes are restructurings that don't change runtime behavior.

---

**Review track**: C

### DIFF
--- a/src/agent/dispatcher.rs
+++ b/src/agent/dispatcher.rs
@@ -11,6 +11,7 @@ use uuid::Uuid;
 
 use crate::agent::Agent;
 use crate::agent::session::{PendingApproval, Session, ThreadState};
+use crate::agent::session_manager::SessionGuardTimer;
 use crate::channels::{IncomingMessage, StatusUpdate};
 use crate::context::JobContext;
 use crate::error::Error;
@@ -263,7 +264,7 @@ struct ChatDelegate<'a> {
 #[async_trait]
 impl<'a> LoopDelegate for ChatDelegate<'a> {
     async fn check_signals(&self) -> LoopSignal {
-        let sess = self.session.lock().await;
+        let sess = SessionGuardTimer::new(self.session.lock().await, "check_signals");
         if let Some(thread) = sess.threads.get(&self.thread_id)
             && thread.state == ThreadState::Interrupted
         {
@@ -562,7 +563,7 @@ impl<'a> LoopDelegate for ChatDelegate<'a> {
                 };
                 redacted_args.push(safe);
             }
-            let mut sess = self.session.lock().await;
+            let mut sess = SessionGuardTimer::new(self.session.lock().await, "record_tool_calls");
             if let Some(thread) = sess.threads.get_mut(&self.thread_id)
                 && let Some(turn) = thread.last_turn_mut()
             {
@@ -671,7 +672,10 @@ impl<'a> LoopDelegate for ChatDelegate<'a> {
                 let needs_approval = match requirement {
                     ApprovalRequirement::Never => false,
                     ApprovalRequirement::UnlessAutoApproved => {
-                        let sess = self.session.lock().await;
+                        let sess = SessionGuardTimer::new(
+                            self.session.lock().await,
+                            "check_auto_approval",
+                        );
                         !sess.is_tool_auto_approved(&tc.name)
                     }
                     ApprovalRequirement::Always => true,
@@ -850,7 +854,10 @@ impl<'a> LoopDelegate for ChatDelegate<'a> {
                         &error_msg,
                     );
                     {
-                        let mut sess = self.session.lock().await;
+                        let mut sess = SessionGuardTimer::new(
+                            self.session.lock().await,
+                            "record_preflight_error",
+                        );
                         if let Some(thread) = sess.threads.get_mut(&self.thread_id)
                             && let Some(turn) = thread.last_turn_mut()
                         {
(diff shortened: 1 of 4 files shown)


serrrfirat:

PR #1848, opened 2026-04-01
fix(relay): thread responses under original message in Slack channels

PierreLeGuen:
## Summary

- Channel mention responses now thread under the user's original message instead of posting as top-level messages
- Root cause: relay channel used `event.channel_id` (e.g. `C088K6C3SQZ`) as `thread_ts` fallback — Slack requires a message timestamp, not a channel ID
- Fix: use `event.id` (the Slack message `ts`) as fallback, which is the correct format for threading

## Test plan

- [x] Regression test: `start_uses_message_ts_as_thread_id_for_mentions`
- [x] All 20 relay channel tests pass
- [x] Zero clippy warnings
- [ ] Deploy and verify @IronClaw mention in channel produces threaded reply

### DIFF
--- a/src/channels/relay/channel.rs
+++ b/src/channels/relay/channel.rs
@@ -194,12 +194,18 @@ impl Channel for RelayChannel {
                         "sender_id": event.sender_id,
                         "sender_name": event.display_name(),
                         "event_type": event.event_type,
-                        "thread_id": event.thread_id,
+                        "thread_id": event.thread_id.as_deref().unwrap_or(&event.id),
                         "provider": event.provider,
                     }));
 
+                // Use the original thread_id if present (already in a thread),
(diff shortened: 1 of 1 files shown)


zmanian (approved):
## Review -- APPROVE

Clean, correct bug fix. The root cause is right: channel mentions without an existing thread used `channel_id` as `thread_ts`, which Slack silently ignores. The fix uses `event.id` (message timestamp) as the fallback, which is the correct value for Slack's `thread_ts` parameter.

Three-way cascade is sound: `event.thread_id` -> `event.id` -> `event.channel_id`. Metadata `thread_id` updated consistently. Good regression test.

### Minor note
The `channel_id` fallback is now only reachable if `event.id` is empty (shouldn't happen for valid Slack events). Consider logging a warning on that path.

PR #1849, opened 2026-04-01
feat: add per-thread delete functionality with full-stack implementation

hzsunzixiang:
Add DELETE /api/chat/thread/{id} endpoint allowing users to remove individual chat threads from the sidebar, with comprehensive backend, frontend, and test coverage.

## Backend
- Add delete_conversation to ConversationStore trait (libsql + postgres)
- Wrap delete operations in transactions (BEGIN IMMEDIATE for libsql)
- Protect assistant thread from deletion (return 403 Forbidden)
- Apply shared chat_rate_limiter to delete endpoint
- Sanitize error responses: log details server-side, return generic messages

## Frontend
- Add per-thread delete button in chat sidebar with hover reveal
- Localize delete button title/confirm/success/failure messages (en, zh-CN)
- Add aria-label for accessibility
- Use visibility/opacity instead of display toggle to prevent layout jump
- Clear active thread on delete and fallback to assistant thread

## Tests
- Add 5 integration tests for delete-thread API:
  - Successful deletion removes thread
  - Deleting nonexistent thread returns 404
  - Cross-user deletion attempt returns 404
  - Deleting assistant thread returns 403
  - Rate-limited delete returns 429

Changed files: server.rs, app.js, style.css, multi_tenant.rs, libsql/conversations.rs, postgres.rs, mod.rs, store.rs

## Summary

-

## Change Type

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] CI/Infrastructure
- [ ] Security
- [ ] Dependencies

## Validation

- [x] `cargo build`
- [x] Relevant tests pass: 
- [x] `cargo test --features integration` if database-backed or integration behavior changed
- [x] Manual testing: 

## Review Follow-Through

---

**Review track**:

### DIFF
--- a/src/channels/web/server.rs
+++ b/src/channels/web/server.rs
@@ -458,6 +458,10 @@ pub async fn start_server(
         .route("/api/chat/history", get(chat_history_handler))
         .route("/api/chat/threads", get(chat_threads_handler))
         .route("/api/chat/thread/new", post(chat_new_thread_handler))
+        .route(
+            "/api/chat/thread/{id}",
+            axum::routing::delete(chat_delete_thread_handler),
+        )
         // Memory
         .route("/api/memory/tree", get(memory_tree_handler))
         .route("/api/memory/list", get(memory_list_handler))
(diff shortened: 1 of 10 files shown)


zmanian (changes requested):
## Review -- REQUEST_CHANGES

### HIGH

**1. Authorization bypass on metadata check** -- `get_conversation_metadata` queries by `id` alone with no `user_id` filter. Any authenticated user can probe arbitrary thread IDs and learn metadata. The 403 for assistant threads also leaks thread existence. Fix: scope metadata query by `user_id`, or move the check inside the delete transaction.

**2. FK orphans** -- `agent_jobs.conversation_id` and `llm_calls.conversation_id` reference `conversations(id)` without `ON DELETE CASCADE` or `SET NULL`. On PostgreSQL, deletion will fail with FK violation. On libSQL (no `PRAGMA foreign_keys`), deletion succeeds but leaves dangling references. Fix: add `ON DELETE SET NULL` via migration, or explicitly nullify before delete.

### Medium
- libSQL does not enable `PRAGMA foreign_keys` -- silent data integrity issues
- Tests only cover libSQL, no PostgreSQL coverage
- Rollback on "not found" is misleading (should commit the no-op)

### Design question
Hard delete with no undo/audit trail -- is this intentional? Consider soft delete (`deleted_at` column) or at minimum audit logging of deletion events.

serrrfirat:

PR #1851, opened 2026-04-01
fix(security): safety layer bypass via output truncation [HIGH]

lycheepuppy:
## Summary

- **Fix safety layer bypass** where `sanitize_tool_output()` returned immediately after truncating oversized tool output, skipping all three safety checks (leak detection, policy enforcement, injection scanning)
- Restructured truncation path so truncated content flows through the full safety pipeline before being returned
- Added regression tests proving truncated output is now scanned for injection patterns

## Security Finding: Safety Layer Bypass via Output Truncation

**Severity:** HIGH
**Component:** `crates/ironclaw_safety/src/lib.rs` — `sanitize_tool_output()`

### Description

The `sanitize_tool_output()` method truncated oversized tool output and returned it **immediately** — before leak detection, policy enforcement, or injection scanning ever ran. This meant any tool output exceeding `max_output_length` was delivered to the LLM with **zero safety checks applied**.

**Order-of-operations problem (before fix):**

1. Length check → if oversized, truncate to UTF-8 boundary
2. **Early return** → truncated content returned with only a low-severity `output_too_large` warning
3. Leak detection → **SKIPPED** for truncated output
4. Policy enforcement → **SKIPPED** for truncated output
5. Injection scanning → **SKIPPED** for truncated output

### Attack Vector

A malicious tool (compromised MCP server, poisoned web page via HTTP tool, sandbox job processing attacker-controlled input) could craft output where:

- The first N bytes (within the truncation boundary) contain prompt injection payloads, credential exfiltration instructions, or policy-violating content
- The total output exceeds `max_output_length`, triggering the early-return path
- The adversarial content reaches the LLM **without any safety scanning**

The attacker does not need to know the exact truncation threshold — they only need to ensure the total output is large enough while placing the payload near the beginning.

### Impact

- Any tool output exceeding `max_output_length` bypassed **all three safety layers**
- Undermines the entire safety architecture — the safety layer is the single choke point for tool output
- The bypass is trivially triggered by any tool that returns large output (web fetch, file read, shell command, MCP tool)

### Fix

Removed the early `return` after truncation. Truncation now produces `(content, was_modified, extra_warnings)` and falls through to the same leak detection → policy enforcement → injection scanning pipeline as non-truncated content. Truncation warnings are preserved and merged into the final output.

## Test plan

- [ ] `cargo test -p ironclaw_safety` — existing adversarial truncation tests still pass
- [ ] New test `truncated_output_still_scanned_for_injection` — verifies injection patterns are detected in oversized output
- [ ] New test `truncated_output_preserves_truncation_warning` — verifies truncation warning survives the full pipeline
- [ ] `cargo clippy -p ironclaw_safety -- -D warnings` — zero warnings

### DIFF
--- a/crates/ironclaw_safety/src/lib.rs
+++ b/crates/ironclaw_safety/src/lib.rs
@@ -54,36 +54,35 @@ impl SafetyLayer {
     pub fn sanitize_tool_output(&self, tool_name: &str, output: &str) -> SanitizedOutput {
         // Check length limits — keep the beginning so the LLM has partial data.
         // Truncated content still flows through all safety checks below.
-        let (mut content, mut was_modified, mut extra_warnings) = if output.len()
-            > self.config.max_output_length
-        {
-            let mut cut = self.config.max_output_length;
-            while cut > 0 && !output.is_char_boundary(cut) {
-                cut -= 1;
-            }
-            let truncated = &output[..cut];
-            let notice = format!(
-                "\n\n[... truncated: showing {}/{} bytes. Use the json tool with \
+        let (mut content, mut was_modified, mut extra_warnings) =
+            if output.len() > self.config.max_output_length {
+                let mut cut = self.config.max_output_length;
+                while cut > 0 && !output.is_char_boundary(cut) {
+                    cut -= 1;
+                }
+                let truncated = &output[..cut];
+                let notice = format!(
+                    "\n\n[... truncated: showing {}/{} bytes. Use the json tool with \
                  source_tool_call_id to query the full output.]",
-                cut,
-                output.len()
-            );
-            (
-                format!("{}{}", truncated, notice),
-                true,
-                vec![InjectionWarning {
-                    pattern: "output_too_large".to_string(),
-                    severity: Severity::Low,
-                    location: 0..output.len(),
-                    description: format!(
-                        "Output from tool '{}' was truncated due to size",
-                        tool_name
-                    ),
-                }],
-            )
-        } else {
-            (output.to_string(), false, vec![])
-        };
+                    cut,
+                    output.len()
+                );
+                (
+                    format!("{}{}", truncated, notice),
+                    true,
+                    vec![InjectionWarning {
+                        pattern: "output_too_large".to_string(),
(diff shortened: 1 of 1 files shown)


zmanian (approved):
## Review -- APPROVE

Legitimate HIGH severity security fix. The vulnerability is real: `sanitize_tool_output()` has an early return after truncation that skips leak detection, policy enforcement, and injection scanning. An attacker only needs to pad output past `max_output_length` while placing malicious content in the first N bytes.

The fix is correct -- truncation now produces a tuple that falls through to the full safety pipeline. Truncated content is scanned (which is right, since it's what the LLM sees). Warning merging preserves both truncation and downstream warnings.

No bypass vectors in the fix. Low regression risk -- non-truncated output path is unchanged.

- The policy `Block` path discards all warnings including truncation warning -- consider preserving for logging (pre-existing issue)
- Consider adding a test where truncated output contains a leaked secret pattern (currently only injection scanning is tested)

zmanian (dismissed):
## Re-review -- APPROVE (post-approval commit check)

The new commit (`4e2ff0b`) is a merge of `staging` into the feature branch. It brings in unrelated staging changes (channels, config, e2e tests, github tool, etc.) but **does not touch `crates/ironclaw_safety/` at all** -- zero files in the safety crate were modified by the merge.

The security fix from the original commit (`a3fa6df`) is intact and unchanged:
- Truncation no longer early-returns; content flows through the full safety pipeline
- Warning merging preserves truncation warnings alongside downstream findings
- Regression tests cover injection scanning on truncated output

### Gemini's comments (not addressed, both non-blocking)

1. **`location: 0..output.len()` range** -- refers to original string length, not truncated content. Pre-existing pattern, not introduced by this PR. Low risk since `location` is metadata on the warning struct, not used for content slicing.

2. **Warning ordering** -- truncation warnings (Low severity) end up before injection warnings (higher severity) due to the `append` + reassign pattern. Cosmetic; callers should sort/filter by severity if ordering matters.

Neither weakens the security fix. Safe to merge.

ilblackdragon (approved):

PR #1868, opened 2026-04-01
feat(agent): add generic plan mode review flow

serrrfirat:
## Summary
- add generic thread-level plan mode with runtime gating for side-effectful tools
- add persisted markdown plan artifacts and a review step before leaving plan mode
- add web support for pending plan-exit reviews with approve/revise/stay actions
- keep the existing /plan checklist flow while adding thread-level planning state

## Validation
- cargo test process_plan_ --lib
- cargo test plan_mode --lib
- node --check src/channels/web/static/app.js
- cargo fmt --all


serrrfirat:
Superseded by #1964, which restacks the generic plan mode review flow onto current staging as a clean replacement PR.

PR #1869, opened 2026-04-01
fix(cli): suppress non-CLI listeners under --cli-only (#1840)

### LINKED ISSUE #1840: --cli-only still starts webhook server and ignores HTTP_HOST/HTTP_PORT
## Summary

`ironclaw --cli-only` still starts the unified webhook server, and in this code path it ignores configured `HTTP_HOST` / `HTTP_PORT` and falls back to `0.0.0.0:8080`.

That is surprising for a mode documented as CLI-only, and it has a security impact because it can open or attempt to open a listener the operator did not explicitly ask for.

## Repro

Environment:
- macOS on Apple Silicon
- local Ollama backend
- PostgreSQL backend
- onboarding completed
- `~/.ironclaw/.env` contains:
  - `LLM_BACKEND=ollama`
  - `OLLAMA_BASE_URL=http://127.0.0.1:11434`
  - `DATABASE_URL=postgres://bert@localhost/ironclaw`
  - `HTTP_HOST=127.0.0.1`
  - `HTTP_PORT=18080`
  - `ONBOARD_COMPLETED=true`

Run:

```bash
/Users/bert/services/ironclaw/target/release/ironclaw --cli-only
```

Observed output:

```text
WARN Webhook server is binding to 0.0.0.0 — it will be reachable from all network interfaces. Set HTTP_HOST=127.0.0.1 to restrict to localhost.

✗ Channel webhook_server failed to start: Failed to bind to 0.0.0.0:8080: Address already in use (os error 48)
```

## Expected behavior

One of these should be true:

1. `--cli-only` should suppress all webhook server startup, including built-in tool webhook ingress.
2. If `--cli-only` still needs the unified webhook server for some internal reason, it should still honor configured `HTTP_HOST` / `HTTP_PORT` instead of falling back to `0.0.0.0:8080`.

## Why this happens

The current control flow appears to be:

- `main.rs` unconditionally pushes built-in tool webhook routes into `webhook_routes`
- the `http` channel block, which is gated by `!cli.cli_only`, is the place that sets `webhook_server_addr` from `HTTP_HOST` / `HTTP_PORT`
- later, if `webhook_routes` is non-empty, IronClaw starts the unified webhook server anyway
- because `webhook_server_addr` was never set in `--cli-only` mode, it falls back to `0.0.0.0:8080`

Relevant code:
- unconditional webhook route registration in `src/main.rs`
- `http` channel config application gated behind `!cli.cli_only`
- webhook server fallback bind to `0.0.0.0:8080`

## Security concern

This is not just a usability issue. For operators trying to run IronClaw in a conservative local-only mode, `--cli-only` currently results in webhook listener startup behavior they did not ask for, and the fallback bind is `0.0.0.0` rather than localhost.

Even when the port is already in use and startup fails, this behavior undermines trust in the security model.

## Suggested fixes

Any of these would resolve the issue cleanly:

- Do not register built-in webhook routes under `--cli-only`
- Or, if built-in webhook routes must exist, still honor `HTTP_HOST` / `HTTP_PORT` when computing the unified webhook server bind in `--cli-only` mode
- Or introduce a separate explicit flag for tool-webhook ingress so it is never implied by `--cli-only`

reidliu41:
## Summary

  - --cli-only did not fully honor its contract: webhook routes, WASM channel endpoints, gateway startup, managed
    tunnels, and the sandbox orchestrator API could still start network listeners
  - This change unifies all non-CLI startup paths behind a single enable_non_cli guard derived from --cli-only
  - It also fixes two state-consistency issues caused by suppressing orchestrator startup: job_prompt was registered
    without a consumer, and sandbox readiness could be misreported as unavailable instead of disabled

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] CI/Infrastructure
- [x] Security
- [ ] Dependencies

## Linked Issue

Fixes #1840

## Validation

- [x] `cargo build`
- [ ] Relevant tests pass: 
- [ ] Manual testing: 

## Security Impact

  This fixes unintended listener startup under --cli-only, including webhook ingress, gateway HTTP serving, managed
  tunnels, and the sandbox orchestrator API. The value of the change is that --cli-only now consistently suppresses non-
  CLI network-facing services, which restores operator trust in the mode’s security boundary.

## Database Impact

None

## Blast Radius

  - src/main.rs startup wiring for non-CLI services
  - sandbox tool registration and readiness reporting
  - NETWORK_SECURITY.md behavior documentation

## Rollback Plan

  Revert the commit. The change is limited to startup wiring and documentation, with no schema, migration, or config-
  format changes.

## Review Follow-Through

 - Regression coverage now checks known non-CLI startup paths in async_main
  - New network-facing startup paths should be added to the guard list and regression test

---

**Review track**: 
 B — bug fix affecting startup behavior and security-sensitive listener initialization


zmanian (dismissed):
## Review -- APPROVE

Good security fix closing a real gap where --cli-only wasn't suppressing network listeners (webhooks, WASM channels, HTTP, Signal, gateway, tunnels, orchestrator API). Single `enable_non_cli` guard applied consistently. NETWORK_SECURITY.md updated.

### Minor nits (non-blocking)
- `non_cli_channels_enabled(cli_only: bool) -> bool` is just `!cli_only` -- the indirection adds no value
- Source grep regression test is clever but fragile (500-char window is arbitrary)

reidliu41:
can it merge?

ilblackdragon:
## Review — paranoid architect

Thanks for tightening `--cli-only`; the unified `enable_non_cli` guard is a clean improvement over scattered `!cli.cli_only` checks, and the source-level regression test is a thoughtful tripwire.

### High — relay channels bypass the guard

`src/main.rs:935-940` (**not** in this diff — but load-bearing for the contract this PR claims to enforce):

```rust
if let Some(ref ext_mgr) = components.extension_manager {
    ext_mgr.set_relay_channel_manager(Arc::clone(&channels)).await;
    ext_mgr.restore_relay_channels(&ext_user_id).await;
}
```

Under `--cli-only`, this still runs. `restore_relay_channels` -> `activate_stored_relay` -> `activate_channel_relay` (`src/extensions/manager.rs:4896`):

1. Fetches `team_id` from settings.
2. Makes an **outbound HTTPS call** to the relay server's `/relay/signing-secret` endpoint (`manager.rs:4998`).
3. Calls `hot_add` on the channel manager (`channels/manager.rs:57`), registering a live `RelayChannel` — so any inbound message delivered through another path (or a later `hot_add` webhook route) would still be processed, and `respond()/broadcast_all()` now fans out to the relay provider.

The comment at `src/main.rs:878` (`Relay channels are handled separately below via restore_relay_channels()`) acknowledges this is a distinct path; it just isn't included in the new suppression. The PR body, the docstring on `non_cli_channels_enabled`, and the NETWORK_SECURITY.md delta (`no webhook server, WASM channel endpoints, HTTP channel, Signal channel, gateway channel, managed tunnel, or sandbox orchestrator API will start`) all assert a complete no-non-CLI-side-effect contract, which this gap violates in the outbound direction.

The `all_non_cli_ingress_paths_are_guarded` test will not catch this because `restore_relay_channels(` and `set_relay_channel_manager(` are not in `guarded_calls`. Suggest: (a) wrap those two lines in `if enable_non_cli`, and (b) add `restore_relay_channels(` / `set_relay_channel_manager(` to the `guarded_calls` list.

### Medium — tool-registration view of `--cli-only`

The `prompt_queue` fix at `src/main.rs:626` (`config.sandbox.enabled && container_job_manager.is_some()`) is correct — good catch that the old form would register `JobPromptTool` with a dead queue.

But note: under `--cli-only` the code still calls `components.tools.register_job_tools(...)` with `container_job_manager = None`, which registers `CreateJobTool`, `ListJobsTool`, `JobStatusTool`, `CancelJobTool` (plus `JobEventsTool` if a DB exists). That's intentional — scheduler-backed jobs still work CLI-only — but the "all non-CLI services disabled" story in the NETWORK_SECURITY.md note is specifically about *listeners*, not tool surface. Worth saying so explicitly in the doc delta to avoid a future reader concluding jobs are disabled under `--cli-only`.

### Medium — test fragility in `all_non_cli_ingress_paths_are_guarded`

`src/main.rs:1310-1354`. Two concerns:

1. The "preceding 500 chars contains `enable_non_cli`" heuristic will happily pass if `enable_non_cli` appears in an *unrelated* nearby line (e.g. an `else` branch that does *not* actually gate the call). A future refactor like `if !enable_non_cli { ...note... }  ...  webhooks::routes(...)` would satisfy the grep without gating anything. A stronger check would parse the enclosing `if` block or require the call to appear *inside* a `{`-balanced region following `enable_non_cli`.
2. `include_str!("main.rs")` embeds the file relative to *this* source file — fine today, but if `async_main` moves to `app.rs` (per CLAUDE.md "Module-owned initialization"), the test silently keeps green-stamping a stale blob. A `#[cfg(test)] fn assert_guarded()` colocated with the guarded code, or walking `proc_macro2`/`syn` over the real module, would survive that refactor.

### Low — doc claim slightly overstates scope

NETWORK_SECURITY.md +141 lists "no ... Signal channel" — correct — but omits relay channels (which this PR does not in fact suppress; see High above). If the relay gap is fixed, update the sentence to include them; if it isn't, remove the stronger "all non-CLI ingress listeners are suppressed" framing so the doc matches reality.

### Notes that held up

- Tunnel suppression at `src/main.rs:377-381` is clean.
- Orchestrator suppression + `DockerStatus::Disabled` propagation to `SandboxReadiness::DisabledByConfig` (`main.rs:999-1001`) is the right move — previously a suppressed orchestrator would have shown up as `DockerUnavailable`, which would mislead routine dispatch.
- Gateway, HTTP, Signal, WASM-channel, and webhook-route suppressions are each gated at the point where the listener/registration actually happens — no half-built state left behind.
- No `.unwrap()`/`.expect()` added in production paths (the one `.expect` at line 553 is pre-existing and reachable only when the HTTP channel is configured). Tests are fine under the project's rules.
- No `info!`/`warn!` added that would corrupt the REPL.

### Verdict

Not blocking merge on the relay gap per se, since the original issue #1840 was specifically the webhook listener — but the PR's framing ("all non-CLI ingress listeners are suppressed") overreaches given the relay path is still live. Either (a) extend the suppression to relay restoration and widen the regression test, or (b) narrow the doc/comment language to "listener ports" and file a follow-up for the outbound-side cleanup.

reidliu41:
updated

ilblackdragon (approved):

PR #1873, opened 2026-04-01
feat: add debug inspector panel for web gateway

### LINKED ISSUE #1492: feat(web): add debug inspection endpoints and verbose SSE mode for the web gateway
## Problem / Motivation

The web gateway emits SSE events for tool calls but deliberately withholds data that would be useful for debugging and observability:

1. **Tool parameters on success are hidden.** `tool_completed` only includes parameters on failure (see `StatusUpdate::tool_completed()` in `src/channels/channel.rs`). During debugging, you need to see *what* the tool was called with even when it succeeded.
2. **Tool output is truncated to 500 bytes.** `tool_result` events use `truncate_preview()` which caps at 500 bytes (see `src/channels/web/util.rs`). Debug inspection needs full output.
3. **System prompt is not available.** The prompt is assembled at runtime from workspace files (AGENTS.md, SOUL.md, IDENTITY.md, USER.md, MEMORY.md, skills context) in `dispatcher.rs`, but no API endpoint exposes the assembled result. When debugging unexpected agent behavior, seeing the actual prompt is critical.
4. **Per-turn token counts are not exposed.** `CostGuard` tracks daily/hourly aggregates, and `/api/gateway/status` returns per-model totals, but there's no way to see how many tokens a specific turn consumed.

These omissions are correct for production — they reduce bandwidth and limit data exposure. But for debugging, they make the web gateway a black box.

### 1. New endpoint: `GET /api/debug/prompt`

Returns the live system prompt as it would be assembled for the current user.

**Response shape:**

```json
{
  "system_prompt": "You are...\n\n# Identity\n...",
  "components": [
    {"source": "AGENTS.md", "tokens_approx": 450},
    {"source": "SOUL.md", "tokens_approx": 120},
    {"source": "IDENTITY.md", "tokens_approx": 80},
    {"source": "USER.md", "tokens_approx": 200},
    {"source": "MEMORY.md", "tokens_approx": 340},
    {"source": "skills/rust-dev.md", "tokens_approx": 600}
  ],
  "total_tokens_approx": 1790,
  "model": "claude-sonnet-4-20250514",
  "context_limit": 200000
}
```

**Implementation:** In `src/channels/web/handlers/chat.rs` (or a new `debug.rs` handler file), call `workspace.system_prompt_for_context(false)` to get the assembled prompt. For the `components` breakdown, read each identity file individually via `workspace.read()` and estimate tokens (word_count * 1.3, matching the existing estimation in `context_monitor.rs`). Get model name and context limit from `AgentDeps`.

**Auth:** Same bearer token as all other protected routes. No additional auth layer.

**Route registration:** Add to the router in `src/channels/web/server.rs` alongside existing `/api/chat/*` routes.

### 2. Verbose SSE mode: `?debug=true` on SSE stream

When the SSE connection is opened with `?debug=true` (`/api/chat/events?token=xxx&debug=true`), emit enhanced event variants:

**Enhanced `tool_completed` event:**
```json
{
  "type": "tool_completed",
  "name": "shell_execute",
  "success": true,
  "error": null,
  "parameters": "{\"command\": \"cargo test\"}",
  "thread_id": "abc-123"
}
```

In verbose mode, `parameters` is ALWAYS included (not just on failure). Redaction via `redact_params()` still applies — sensitive params are still masked.

**New `tool_result_full` event:**
```json
{
  "type": "tool_result_full",
  "name": "shell_execute",
  "output": "running 42 tests\ntest result: ok. 42 passed; 0 failed...",
  "thread_id": "abc-123"
}
```

In verbose mode, emit the full tool output alongside the existing truncated `tool_result`. The existing `tool_result` continues to emit with the 500-byte preview (backward compatible). `tool_result_full` is an additional event that only verbose subscribers receive.

**New `turn_metrics` event:**
```json
{
  "type": "turn_metrics",
  "thread_id": "abc-123",
  "input_tokens": 4200,
  "output_tokens": 850,
  "cache_read_tokens": 3100,
  "model": "claude-sonnet-4-20250514",
  "duration_ms": 2340,
  "tool_calls_count": 3,
  "iteration": 2
}
```

Emitted after each LLM call completes (not just at turn end). Data source: the `CompletionOutput.usage` field returned by `LlmProvider::complete()`. This is already available in `dispatcher.rs` after each LLM call — it just needs to be forwarded as an SSE event.

### 3. Implementation approach for verbose mode

**Option A (recommended): Per-subscriber flag on SseManager.**

Currently `SseManager` uses a `broadcast::Sender<SseEvent>` that fans out to all subscribers. Verbose mode needs per-subscriber filtering.

Add a wrapper that checks a flag:

```rust
// In sse.rs
pub struct SseSubscription {
    receiver: broadcast::Receiver<SseEvent>,
    verbose: bool,
}

impl SseSubscription {
    pub fn next_event(&mut self) -> Option<SseEvent> {
        loop {
            match self.receiver.recv().await {
                Ok(event) => {
                    if event.is_verbose_only() && !self.verbose {
                        continue; // skip verbose events for non-verbose subscribers
                    }
                    return Some(event);
                }
                Err(_) => return None,
            }
        }
    }
}
```

Add `is_verbose_only()` to `SseEvent`:
```rust
impl SseEvent {
    pub fn is_verbose_only(&self) -> bool {
        matches!(self, SseEvent::ToolResultFull { .. } | SseEvent::TurnMetrics { .. })
    }
}
```

For `tool_completed` with always-include-params: broadcast TWO events — the normal one (params only on failure) and a verbose-only one (params always). Verbose subscribers see both but can deduplicate by name+thread_id. Or simpler: always include params in the broadcast but let non-verbose subscribers' JS ignore it (less clean but zero backend filtering needed).

**Decision: go with per-subscriber filtering.** It prevents bandwidth waste for non-debug clients and keeps the existing SSE contract unchanged.

### 4. SseEvent enum additions

```rust
// In src/channels/web/types.rs
pub enum SseEvent {
    // ... existing variants ...

    /// Full tool output (verbose mode only). Complements the truncated `ToolResult`.
    ToolResultFull {
        name: String,
        output: String,
        thread_id: Option<String>,
    },

    /// Per-LLM-call metrics (verbose mode only).
    TurnMetrics {
        thread_id: Option<String>,
        input_tokens: u64,
        output_tokens: u64,
        cache_read_tokens: u64,
        model: String,
        duration_ms: u64,
        tool_calls_count: usize,
        iteration: usize,
    },
}
```

### 5. StatusUpdate additions

Add new variants to `StatusUpdate` in `src/channels/channel.rs`:

```rust
pub enum StatusUpdate {
    // ... existing variants ...
    ToolResultFull {
        name: String,
        output: String,
    },
    TurnMetrics {
        input_tokens: u64,
        output_tokens: u64,
        cache_read_tokens: u64,
        model: String,
        duration_ms: u64,
        tool_calls_count: usize,
        iteration: usize,
    },
}
```

Emit `StatusUpdate::ToolResultFull` from the same place `StatusUpdate::ToolResult` is emitted (in `execute_tool_with_safety()` or `process_tool_result()` in `src/tools/execute.rs`). Emit `StatusUpdate::TurnMetrics` after each `reasoning.complete()` call in `dispatcher.rs`.

## Architecture Notes

```
dispatcher.rs (tool execution)
  │
  ├─ StatusUpdate::ToolStarted { name }          → SseEvent::ToolStarted (all)
  ├─ StatusUpdate::ToolCompleted { name, params } → SseEvent::ToolCompleted (all, params only on fail)
  │                                                → SseEvent::ToolCompleted (verbose, params always)
  ├─ StatusUpdate::ToolResult { name, preview }   → SseEvent::ToolResult (all, 500 byte preview)
  ├─ StatusUpdate::ToolResultFull { name, output } → SseEvent::ToolResultFull (verbose only)
  └─ StatusUpdate::TurnMetrics { ... }             → SseEvent::TurnMetrics (verbose only)

/api/debug/prompt
  │
  ├─ workspace.system_prompt_for_context(false)
  ├─ workspace.read("AGENTS.md"), .read("SOUL.md"), etc.
  └─ JSON response with assembled prompt + component breakdown
```

**Why per-subscriber filtering instead of separate SSE endpoint?** A separate `/api/debug/events` endpoint would mean debug clients need two SSE connections (one for chat, one for debug). SSE spec limits browsers to ~6 connections per domain. A single connection with a verbose flag is cleaner.

**Why still redact in verbose mode?** Debug mode shows *what the tool was called with*, not secrets. `redact_params()` masks credential values, not tool logic. A developer debugging a failing HTTP call needs to see the URL and headers, not the auth token value.

## Code Pointers

- `src/channels/web/types.rs` — `SseEvent` enum, `impl SseEvent` serialization — add new variants here
- `src/channels/web/sse.rs` — `SseManager`, `subscribe()` — add verbose flag to subscription
- `src/channels/web/mod.rs` — `StatusUpdate` → `SseEvent` mapping — add new mappings
- `src/channels/channel.rs` — `StatusUpdate` enum, `tool_completed()` constructor — add verbose variants
- `src/channels/web/handlers/chat.rs` — Chat API endpoints — add `/api/debug/prompt` handler (or new `debug.rs`)
- `src/channels/web/server.rs` — Route registration — wire new endpoint
- `src/agent/dispatcher.rs` — `ChatDelegate` tool execution, LLM calls — emit new StatusUpdates
- `src/tools/execute.rs` — `execute_tool_with_safety()`, `process_tool_result()` — emit ToolResultFull
- `src/agent/context_monitor.rs` — Token estimation formula (word_count * 1.3) — reuse for prompt component breakdown
- `src/workspace/mod.rs` — `Workspace::system_prompt_for_context()` — used by prompt endpoint

## Acceptance Criteria

- [ ] `GET /api/debug/prompt` returns the assembled system prompt with component breakdown
- [ ] Prompt endpoint returns approximate token counts per component and total
- [ ] Prompt endpoint returns current model name and context limit
- [ ] Prompt endpoint requires bearer token auth (same as other protected routes)
- [ ] SSE stream accepts `?debug=true` query parameter
- [ ] Verbose subscribers receive `tool_completed` with parameters on success (redacted)
- [ ] Verbose subscribers receive `tool_result_full` with full (non-truncated) tool output
- [ ] Verbose subscribers receive `turn_metrics` after each LLM call
- [ ] Non-verbose subscribers see NO change in SSE event stream (backward compatible)
- [ ] `tool_result_full` output is still subject to `redact_params()` for sensitive data
- [ ] `SseEvent::is_verbose_only()` correctly identifies verbose-only events
- [ ] `SseSubscription` filters verbose events for non-verbose subscribers
- [ ] New `StatusUpdate` variants added and mapped to `SseEvent` in web channel
- [ ] Unit test: prompt endpoint returns valid JSON with all expected fields
- [ ] Unit test: verbose subscriber receives `tool_result_full`, non-verbose does not
- [ ] Unit test: `turn_metrics` contains valid token counts from `CompletionOutput`
- [ ] `cargo clippy --all --all-features` passes with zero warnings

## Pitfalls & Landmines

- **Prompt assembly side effects**: `system_prompt_for_context()` reads workspace files. Calling it from the debug endpoint is safe (read-only), but be aware it may return different content than what the agent actually used if workspace files changed between the agent's last turn and the debug request. Document this in the API response (add a `note: "reconstructed, may differ from last turn"` field).
- **Tool output size**: `tool_result_full` sends the full output, which can be very large (e.g., `web_fetch` returning an entire HTML page). Cap at a reasonable maximum (e.g., 50KB) with a `truncated: true` flag if exceeded. This prevents a single verbose subscriber from overwhelming the broadcast buffer.
- **Broadcast buffer pressure**: Verbose events increase the event rate. The 256-event broadcast buffer may overflow faster for slow debug clients. This is acceptable — slow clients already lose events silently. But document that debug mode may drop events under high tool throughput.
- **SSE query parameter parsing**: The existing SSE handler in `server.rs` already parses `?token=xxx`. Add `debug=true` to the same query struct. Use `#[serde(default)]` to keep it optional.
- **Thread safety of prompt reconstruction**: `Workspace::system_prompt_for_context()` does multiple async reads. These are not transactional — files could change between reads. This is fine for debug purposes. Don't add locking.
- **Sensitive data in system prompt**: The system prompt includes MEMORY.md content, which may contain personal information. The prompt endpoint is behind bearer auth (same as all other data), so this is no worse than existing `/api/memory/read?path=MEMORY.md`. But do NOT cache the prompt response — it must be fresh on every call.
- **WebSocket subscribers**: The web gateway also supports WebSocket at `/api/chat/ws`. If verbose mode is implemented, also support `?debug=true` on the WebSocket URL. The same `SseEvent` types flow through both transports.

## Non-Goals

- Frontend debug panel UI (separate issue)
- Exposing Claude's internal thinking/reasoning blocks (not available from the API)
- Persisting debug data to database
- Debug mode for non-web channels (CLI, HTTP webhook)
- Authentication layer beyond existing bearer token
- Rate limiting on debug endpoints

### LINKED ISSUE #1493: feat(web): add debug inspector panel to web gateway chat UI
## Problem / Motivation

When debugging agent behavior in the web gateway, there's no way to see:
- What system prompt the agent is operating under
- What parameters a tool was called with (only visible on failure)
- Full tool output (truncated to 500 bytes in the chat activity cards)
- How many tokens each turn consumed and which model served it
- The real-time event stream the gateway is processing

The Logs tab shows server-side tracing, but that's raw log lines — not structured debug info tied to the conversation. Developers currently resort to setting `RUST_LOG=debug` and reading terminal output, which is disconnected from the UI they're testing.

**Depends on:** #1492 (debug inspection endpoints and verbose SSE mode)

### Activation

Two mechanisms, both result in the same UI:

1. **URL parameter**: `?debug=true` — activates debug panel on page load. Persisted to `sessionStorage` so it survives page refreshes but not new tabs.
2. **Toolbar button**: A `{}` icon button in the tab bar (right side, next to the language switcher). Toggles the panel open/closed. Only visible when `?debug=true` was used at least once in the session (prevents UI clutter for non-developers).

When debug mode is active:
- SSE connection is opened with `?debug=true` to receive verbose events
- The `{}` button is visible in the tab bar
- The debug panel is shown/hidden via the button

### Layout

The debug panel is a **right sidebar** in the chat tab, alongside the existing left thread sidebar:

```
┌──────────────────────────────────────────────────────────────────┐
│  Tab Bar  [Chat|Memory|Jobs|...]  [{}] [🌐] [● Connected]       │
├────────┬──────────────────────────────────┬──────────────────────┤
│ Thread │  Chat Messages Area              │  Debug Inspector     │
│ Sidebar│  (flex-shrink, fills middle)     │  (380px, right)      │
│ (240px)│                                  │                      │
│        │  <user message>                  │  [Prompt|Activity|   │
│        │  <assistant message>             │   Stats]             │
│        │  <tool cards>                    │                      │
│        │                                  │  (panel content)     │
│        │                                  │                      │
│        ├──────────────────────────────────┤                      │
│        │  [textarea] [send]              │                      │
└────────┴──────────────────────────────────┴──────────────────────┘
```

**Width:** 380px fixed. Collapsible via the `{}` button (same pattern as thread sidebar collapse).

**CSS:** Follow the existing sidebar pattern. Use the same variables (`--bg-secondary`, `--border-color`, etc.). The panel sits inside `.chat-tab` as a flex sibling of `.chat-container`.

```css
.debug-panel {
  width: 380px;
  min-width: 380px;
  border-left: 1px solid var(--border-color);
  background: var(--bg-secondary);
  display: flex;
  flex-direction: column;
  overflow: hidden;
}
.debug-panel.hidden { display: none; }
```

**Responsive:** On screens narrower than 1024px, the debug panel becomes a full-width overlay (same approach as thread sidebar on mobile with `.expanded-mobile`). On screens narrower than 768px, the `{}` button is hidden entirely — debug mode is desktop-only.

#### Tab 1: Prompt

Shows the live system prompt, fetched from `GET /api/debug/prompt` (#1492).

```
┌─ Prompt ─────────────────────────────────┐
│                                          │
│  Model: claude-sonnet-4-20250514         │
│  Context: 4,200 / 200,000 tokens        │
│  ─────────────────────────────────────── │
│  Components:                             │
│  ┌──────────────────────────┬──────┐     │
│  │ AGENTS.md                │ 450t │     │
│  │ SOUL.md                  │ 120t │     │
│  │ IDENTITY.md              │  80t │     │
│  │ USER.md                  │ 200t │     │
│  │ MEMORY.md                │ 340t │     │
│  │ skills/rust-dev.md       │ 600t │     │
│  └──────────────────────────┴──────┘     │
│  ─────────────────────────────────────── │
│  [▶ Show full prompt]                    │
│                                          │
│  (collapsible <pre> with full prompt)    │
│                                          │
│  [↻ Refresh]                             │
└──────────────────────────────────────────┘
```

- Components list: each row is a workspace file contributing to the prompt, with approximate token count
- "Show full prompt" is collapsed by default (prompt can be very long). Uses a `<pre>` with `white-space: pre-wrap; max-height: 60vh; overflow-y: auto;`
- "Refresh" button re-fetches `/api/debug/prompt` (prompt changes when workspace files change)
- Token bar: simple `<progress>` element showing context usage

**Fetch timing:** Load on panel open, then manual refresh only. Do NOT auto-poll.

#### Tab 2: Activity

Real-time timeline of events for the current assistant response (in-progress or selected). This is the primary debugging view.

```
┌─ Activity ───────────────────────────────┐
│                                          │
│  ● LLM Call #1              [2.3s]       │
│    Model: claude-sonnet-4-20250514       │
│    In: 4,200t  Out: 180t  Cache: 3,100t │
│                                          │
│  ◆ shell_execute            [0.8s] ✓     │
│    ┌ Parameters ─────────────────────┐   │
│    │ {"command": "cargo test"}       │   │
│    └─────────────────────────────────┘   │
│    ┌ Output ─────────────────────────┐   │
│    │ running 42 tests                │   │
│    │ test result: ok. 42 passed...   │   │
│    └─────────────────────────────────┘   │
│                                          │
│  ◆ memory_search            [0.3s] ✓     │
│    ┌ Parameters ─────────────────────┐   │
│    │ {"query": "auth config"}        │   │
│    └─────────────────────────────────┘   │
│    ┌ Output ─────────────────────────┐   │
│    │ [3 results, top: MEMORY.md]     │   │
│    └─────────────────────────────────┘   │
│                                          │
│  ● LLM Call #2              [1.5s]       │
│    In: 5,100t  Out: 420t  Cache: 4,200t │
│                                          │
│  Total: 4.9s | 2 LLM calls | 2 tools    │
└──────────────────────────────────────────┘
```

**Data sources:**
- `turn_metrics` SSE event → LLM call entries (tokens, duration, model)
- `tool_started` → tool entry with spinner
- `tool_completed` → tool entry updated with ✓/✗, parameters (verbose mode provides these on success)
- `tool_result_full` → tool output section (full, not truncated)
- `thinking` → status text between entries

**Behavior:**
- During an in-progress response: auto-scrolls to bottom, entries appear in real-time
- After response completes: static, scrollable
- Each entry is collapsible (parameters and output sections toggle independently)
- Parameters and output use `<pre>` with monospace font, word-wrap, max-height with scroll
- Failed tools highlighted with `var(--danger)` border (same pattern as existing tool cards)

**State management:**
- Maintain a `debugActivityLog` array in JS. Each entry: `{ type: 'llm_call'|'tool_call'|'thinking', data: {...}, timestamp: Date }`
- Clear on new user message (start of new turn)
- Rebuild DOM from array when switching tabs or selecting a different message

#### Tab 3: Stats

Aggregate statistics for the current session.

```
┌─ Stats ──────────────────────────────────┐
│                                          │
│  Session                                 │
│  ─────────────────────────────────────── │
│  Turns:          12                      │
│  Total tokens:   48,200 in / 6,400 out   │
│  Cache hits:     38,100 (79%)            │
│  Tools called:   28                      │
│  Tool failures:  2                       │
│  Duration:       14m 32s                 │
│                                          │
│  Model Usage                             │
│  ─────────────────────────────────────── │
│  claude-sonnet-4  │ 42,000 in │ $0.14   │
│  claude-haiku-4   │  6,200 in │ $0.02   │
│                                          │
│  ─────────────────────────────────────── │
│  SSE Events                              │
│  Total received: 847                     │
│  Dropped (lag):  0                       │
│  Connection:     3m 12s                  │
│                                          │
│  [↻ Refresh from server]                 │
└──────────────────────────────────────────┘
```

**Data sources:**
- Session-local counters from SSE events (input tokens, output tokens from `turn_metrics`, tool counts from `tool_started`/`tool_completed`)
- Server-side data from `GET /api/gateway/status` (model usage, cost) — fetched on tab open and manual refresh
- SSE connection health: track `eventSource.readyState`, count received events, detect reconnections

### JavaScript Implementation

**Global state additions:**

```javascript
let debugMode = false;
let debugPanelOpen = false;
let debugActivityLog = [];      // current turn's events
let debugSessionStats = {       // session-wide counters
  turns: 0,
  inputTokens: 0,
  outputTokens: 0,
  cacheReadTokens: 0,
  toolCalls: 0,
  toolFailures: 0,
  sseEventsReceived: 0,
  sseEventsDropped: 0,
  sessionStartTime: null,
};
let debugActiveTab = 'activity'; // prompt | activity | stats
```

**SSE connection change:**

In `connectSSE()`, append `&debug=true` to the EventSource URL when debug mode is active:

```javascript
function connectSSE() {
  let url = `/api/chat/events?token=${encodeURIComponent(token)}`;
  if (debugMode) url += '&debug=true';
  eventSource = new EventSource(url);
  // ... existing handlers ...

  // New handlers for verbose events:
  eventSource.addEventListener('tool_result_full', (e) => {
    const data = JSON.parse(e.data);
    debugActivityLog.push({ type: 'tool_result_full', data, timestamp: new Date() });
    renderDebugActivity();
  });

  eventSource.addEventListener('turn_metrics', (e) => {
    const data = JSON.parse(e.data);
    debugActivityLog.push({ type: 'llm_call', data, timestamp: new Date() });
    debugSessionStats.inputTokens += data.input_tokens;
    debugSessionStats.outputTokens += data.output_tokens;
    debugSessionStats.cacheReadTokens += data.cache_read_tokens;
    renderDebugActivity();
  });
}
```

**Activation flow:**

```javascript
function initDebugMode() {
  const params = new URLSearchParams(window.location.search);
  if (params.has('debug') && params.get('debug') !== 'false') {
    debugMode = true;
    sessionStorage.setItem('ironclaw_debug', 'true');
  } else if (sessionStorage.getItem('ironclaw_debug') === 'true') {
    debugMode = true;
  }

  if (debugMode) {
    document.getElementById('debug-toggle-btn').style.display = '';
    debugSessionStats.sessionStartTime = new Date();
  }
}
```

Call `initDebugMode()` before `authenticate()` in the startup sequence.

**Panel rendering:**

Follow the existing pattern used by the thread sidebar: build DOM with `createElement`, use class toggling for show/hide, avoid innerHTML for user-controlled data.

Tool parameters and output MUST be rendered via `textContent` (not `innerHTML`) since they may contain HTML/script content from tool results. Use `<pre>` elements for structured display.

### i18n

Add keys to both `en.js` and `zh-CN.js`:

```javascript
// en.js additions
"debug.toggle": "Debug Inspector",
"debug.tab.prompt": "Prompt",
"debug.tab.activity": "Activity",
"debug.tab.stats": "Stats",
"debug.prompt.model": "Model",
"debug.prompt.context": "Context",
"debug.prompt.components": "Components",
"debug.prompt.show_full": "Show full prompt",
"debug.prompt.hide_full": "Hide full prompt",
"debug.prompt.refresh": "Refresh",
"debug.activity.llm_call": "LLM Call",
"debug.activity.tokens_in": "In",
"debug.activity.tokens_out": "Out",
"debug.activity.cache": "Cache",
"debug.activity.parameters": "Parameters",
"debug.activity.output": "Output",
"debug.activity.total": "Total",
"debug.activity.empty": "No activity yet. Send a message to see the debug timeline.",
"debug.stats.session": "Session",
"debug.stats.turns": "Turns",
"debug.stats.total_tokens": "Total tokens",
"debug.stats.cache_hits": "Cache hits",
"debug.stats.tools_called": "Tools called",
"debug.stats.tool_failures": "Tool failures",
"debug.stats.duration": "Duration",
"debug.stats.model_usage": "Model Usage",
"debug.stats.sse_events": "SSE Events",
"debug.stats.received": "Total received",
"debug.stats.dropped": "Dropped (lag)",
"debug.stats.connection": "Connection",
"debug.stats.refresh": "Refresh from server",
```

### HTML additions (in index.html)

Add the debug toggle button to the tab bar (hidden by default):

```html
<!-- In .tab-bar, after language switcher, before status dot -->
<button id="debug-toggle-btn" class="debug-toggle-btn" style="display: none;"
        onclick="toggleDebugPanel()" data-i18n-title="debug.toggle">
  <span class="debug-toggle-icon">{}</span>
</button>
```

Add the debug panel container inside `.chat-tab`:

```html
<div class="chat-tab" id="chat-tab">
  <div class="thread-sidebar" id="thread-sidebar">...</div>
  <div class="chat-container" id="chat-container">...</div>
  <!-- NEW -->
  <div class="debug-panel hidden" id="debug-panel">
    <div class="debug-panel-tabs">
      <button class="debug-tab active" data-tab="prompt" data-i18n="debug.tab.prompt">Prompt</button>
      <button class="debug-tab" data-tab="activity" data-i18n="debug.tab.activity">Activity</button>
      <button class="debug-tab" data-tab="stats" data-i18n="debug.tab.stats">Stats</button>
    </div>
    <div class="debug-panel-content" id="debug-panel-content">
      <!-- dynamically rendered by JS -->
    </div>
  </div>
</div>
```

## Architecture Notes

**Why a right panel and not a modal or bottom drawer?**
- Modals block interaction with the chat
- Bottom drawers compete with the input area
- A right panel lets you watch debug info while chatting — the primary use case is "send a message, watch what happens in the debug timeline"

**Why three tabs instead of one scrollable panel?**
- The prompt can be thousands of lines — it would bury the activity timeline if combined
- Stats are fetch-on-demand (server roundtrip), while activity is real-time SSE — different update patterns
- Follows the existing tab pattern used by the main app (Chat|Memory|Jobs|etc.)

**Why tie activity to the current turn, not the full conversation?**
- Full conversation activity would grow unboundedly
- The primary debug question is "what just happened" or "what is happening now"
- Historical turns can be inspected via the chat history (tool cards are already persisted)

**Existing patterns to follow:**
- Thread sidebar collapse/expand: `toggleThreadSidebar()` in app.js — same toggle pattern for debug panel
- Tab switching: existing `.tab-bar button` click handlers — same pattern for debug sub-tabs
- Tool cards: `.activity-tool-card` in style.css — reuse styling for debug activity entries
- Status dot: `.status-dot` — reuse for LLM call status indicators

## Code Pointers

- `src/channels/web/static/index.html` — HTML structure, tab bar, chat tab layout
- `src/channels/web/static/style.css` — All styling; search for `.thread-sidebar` for sidebar layout pattern, `.activity-tool-card` for tool card styling
- `src/channels/web/static/app.js` — `connectSSE()` for SSE handler registration, `addToolCard()` / `completeToolCard()` for tool card rendering pattern, `toggleThreadSidebar()` for sidebar toggle pattern
- `src/channels/web/static/i18n/en.js` — English translations (add debug keys)
- `src/channels/web/static/i18n/zh-CN.js` — Chinese translations (add debug keys)
- `src/channels/web/CLAUDE.md` — Gateway module spec (update with new panel docs)

## Acceptance Criteria

- [ ] `?debug=true` URL parameter activates debug mode and persists to sessionStorage
- [ ] `{}` button appears in tab bar when debug mode is active
- [ ] Clicking `{}` toggles the debug panel open/closed
- [ ] Debug panel has three tabs: Prompt, Activity, Stats
- [ ] **Prompt tab**: fetches and displays system prompt from `/api/debug/prompt`
- [ ] **Prompt tab**: shows component breakdown with token counts
- [ ] **Prompt tab**: full prompt is collapsed by default, expandable
- [ ] **Prompt tab**: Refresh button re-fetches prompt
- [ ] **Activity tab**: shows real-time timeline during in-progress responses
- [ ] **Activity tab**: LLM call entries show input/output/cache token counts
- [ ] **Activity tab**: tool entries show name, duration, success/failure indicator
- [ ] **Activity tab**: tool parameters displayed (from verbose `tool_completed`)
- [ ] **Activity tab**: full tool output displayed (from `tool_result_full`)
- [ ] **Activity tab**: timeline clears on new user message
- [ ] **Activity tab**: auto-scrolls during streaming, scrollable after completion
- [ ] **Stats tab**: shows session counters (turns, tokens, tools, duration)
- [ ] **Stats tab**: shows per-model token usage from `/api/gateway/status`
- [ ] **Stats tab**: shows SSE connection health (events received, connection uptime)
- [ ] Panel uses `textContent` (never `innerHTML`) for tool params and output (XSS prevention)
- [ ] Panel respects existing dark theme CSS variables
- [ ] Panel collapses on screens < 1024px; hidden on screens < 768px
- [ ] i18n keys added for English and Chinese
- [ ] SSE connection includes `&debug=true` when debug mode is active
- [ ] Panel does not affect chat UI when debug mode is inactive (zero visual/performance impact)
- [ ] Debug panel state (open/closed, active tab) persists across page refreshes via sessionStorage
- [ ] `cargo clippy --all --all-features` passes (no Rust changes expected, but verify static file embedding)

## Pitfalls & Landmines

- **SSE reconnection**: When the SSE connection drops and reconnects (existing reconnect logic in `connectSSE()`), the debug flag must be re-applied. The current `connectSSE()` builds the URL fresh each time — just ensure `debugMode` is checked in the URL construction, not cached at startup.
- **Activity log memory**: The `debugActivityLog` array grows with each event. Clear it on each new user message. Also cap at 1000 entries (matching the pattern used for `jobEvents` which caps at 500) with oldest-first eviction.
- **Tool output rendering**: Full tool output can contain HTML, ANSI escape codes, or very long single lines. Use `<pre>` with `white-space: pre-wrap; word-break: break-all; max-height: 300px; overflow-y: auto;`. NEVER use `innerHTML` — always `textContent`.
- **Prompt display encoding**: The system prompt may contain markdown with backticks, angle brackets, etc. Render in a `<pre>` with `textContent`, not as rendered markdown. The user wants to see the raw prompt, not a formatted version.
- **Layout reflow**: Adding a 380px panel to the chat tab reduces the available chat width. On a 1280px screen: 240 (sidebar) + 380 (debug) = 620px consumed, leaving 660px for chat. This is tight but workable. The chat messages already have `max-width: 72%` which will adapt. Test at 1280px to ensure input area doesn't break.
- **Static file embedding**: All frontend files are embedded at compile time via `include_str!()`. Changes to `index.html`, `style.css`, `app.js` require recompilation. No hot reload. Test in browser with a local dev build.
- **Thread sidebar interaction**: Both the thread sidebar (left) and debug panel (right) can be open simultaneously. The chat container in the middle must flex-shrink gracefully. Use `flex: 1 1 0; min-width: 300px;` on `.chat-container` to prevent it from collapsing to zero.
- **Existing tool cards**: The chat already shows inline tool activity cards. The debug panel's Activity tab shows more detail (full params, full output, token counts). These are complementary, not duplicative. Don't hide or modify the existing inline tool cards.

## Non-Goals

- Backend debug endpoints (covered by #1492)
- Exposing Claude's internal reasoning/thinking blocks (not available from the API)
- Editing the system prompt from the panel (read-only inspection)
- Debug panel for non-web channels
- Persisting debug logs across sessions
- Performance profiling or flame graphs
- Mobile-optimized debug panel (desktop-only is acceptable)

italic-jinxin:
Add a debug inspector sidebar with three tabs (Prompt, Activity, Stats) activated via ?debug=true URL parameter. Consolidate theme-init.js into a new init.js for early initialization. The panel shows real-time SSE event timeline, system prompt component breakdown with token estimates, and session-wide statistics including per-model usage.

- New files: init.js, debug-panel.js, debug-panel.css
- New endpoint: /api/debug/debug/prompt for system prompt inspection
- i18n support (en + zh-CN) for all debug panel strings
- Responsive layout: sidebar on desktop, overlay on tablet, hidden on mobile

## Summary

| Prompt | Activity | Stats |
|--------|----------|-------|
|  |  |  |

-

## Change Type

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] CI/Infrastructure
- [ ] Security
- [ ] Dependencies

## Linked Issue

Closes https://github.com/nearai/ironclaw/issues/1493

Closes https://github.com/nearai/ironclaw/issues/1492

## Validation

- [ ] `cargo build`
- [ ] Relevant tests pass: 
- [ ] Manual testing: 

## Review Follow-Through

---

**Review track**:


italic-jinxin:
@think-in-universe

src/channels/channel.rs
```diff
@@ -366,9 +381,10 @@ const APPROVAL_SUMMARY_DESCRIPTION_BYTES: usize = 120;
...
     /// Build a `ToolCompleted` status with redacted parameters.
     ///
-    /// On failure, serializes the tool's input parameters as pretty JSON after
-    /// replacing any keys listed in the tool's `sensitive_params()` with
-    /// `"[REDACTED]"`. On success, no parameters or error are included.
+    /// Serializes the tool's input parameters as pretty JSON after replacing
+    /// any keys listed in the tool's `sensitive_params()` with `"[REDACTED]"`.
+    /// Parameters are always included (for debug panel visibility).
+    /// Error message is populated only on failure.
     ///
```
serrrfirat, line 616: **Medium Severity — Misleading docstring**

The updated docstring says _"Parameters are always included (for debug panel visibility)"_ but the code still only includes parameters on failure:

```rust
let parameters = if !success {
    let safe = crate::tools::redact_params(params, sensitive);
    Some(serde_json::to_string_pretty(&safe).unwrap_or_else(|_| safe.to_string()))
} else {
    None
};
```

The test at line 725 confirms `parameters.is_none()` on success. The logic is identical to before (just moved to a `let` binding), so the docstring is incorrect.

**Suggested fix:** Change the docstring to: _"Serializes the tool's input parameters as pretty JSON after replacing any keys listed in the tool's `sensitive_params()` with `"[REDACTED]"`. Parameters are only included on failure; verbose clients see full output via the ToolResultFull event instead."_

src/channels/web/server.rs
```diff
@@ -3244,6 +3280,101 @@ struct GatewayStatusResponse {
...
+    }
+    if chars / words > 10 {
+        // Character-based: ~1.5 chars per token for CJK text
+        (chars as f64 / 1.5) as usize + 4
+    } else {
+        ((words as f64) * 1.3) as usize + 4
+    }
+}
+
+async fn debug_prompt_handler(
```
serrrfirat: **Medium Severity — No access control on debug prompt endpoint**

This endpoint is on the protected router (requires auth) but has no role check. Any authenticated user can read workspace identity files (AGENTS.md, SOUL.md, USER.md, IDENTITY.md, TOOLS.md, MEMORY.md) and the full assembled system prompt.

In multi-tenant deployments, this exposes admin-level system configuration (agent personality, instructions, identity) to regular users.

**Suggested fix:** Either:
1. Add an `AdminUser` extractor (like other admin endpoints), or
2. Gate behind the `?debug=true` session flag so only users who explicitly opted into debug mode can access it, or
3. At minimum, document the exposure in the CLAUDE.md API table.

src/channels/web/server.rs
```diff
@@ -2022,6 +2034,30 @@ pub async fn clear_auth_mode(state: &GatewayState, user_id: &str) {
     }
 }
 
+#[derive(serde::Deserialize)]
+struct ChatEventsQuery {
```
serrrfirat: **Medium Severity — Duplicate ChatEventsQuery and chat_events_handler**

This struct and handler are also defined in \`handlers/chat.rs:42-56\`. Both were updated identically in this PR, but the duplication is a divergence risk — a future change might update one but miss the other.

The route at line 542 in this file uses the local \`chat_events_handler\`, not the one from \`handlers/chat.rs\`.

**Suggested fix:** Remove the duplicate from \`handlers/chat.rs\`, or have \`server.rs\` import and use the one from \`handlers/chat.rs\` — keeping a single source of truth.

src/agent/dispatcher.rs
```diff
@@ -1125,6 +1144,36 @@ impl<'a> LoopDelegate for ChatDelegate<'a> {
                                 &self.message.metadata,
                             )
                             .await;
+
+                        // Send full (non-truncated) output for debug subscribers.
+                        const MAX_TOOL_OUTPUT_BYTES: usize = 50_000;
+                        let truncated = output.len() > MAX_TOOL_OUTPUT_BYTES;
+                        let capped = if truncated {
```
serrrfirat, line 1217: **Low Severity — Inefficient UTF-8 truncation**

This iterates all \`char_indices()\` then takes \`.next_back()\`, scanning the entire string to find the last valid byte boundary <= 50KB. For large tool outputs this is O(n) when a backward scan would be O(1) amortized.

The crate already has \`crate::util::floor_char_boundary\` (used in \`channel.rs:444\` for approval parameter truncation) which scans backwards efficiently.

**Suggested fix:**
\`\`\`rust
let boundary = crate::util::floor_char_boundary(output, MAX_TOOL_OUTPUT_BYTES);
output[..boundary].to_string()
\`\`\`

src/channels/web/server.rs
```diff
@@ -3244,6 +3280,101 @@ struct GatewayStatusResponse {
...
+}
+
+#[derive(serde::Serialize)]
+struct DebugPromptComponent {
+    source: String,
+    label: String,
+    content: String,
+    estimated_tokens: usize,
+}
+
```
serrrfirat, line 826: **Low Severity — No unit tests for estimate_tokens**

This pure function has CJK-aware heuristics (character-based estimation when avg word length > 10 bytes) but no unit tests. Edge cases worth covering:

- Empty string → should return 4 (minimum overhead)
- Pure ASCII English text
- CJK text (e.g. Chinese/Japanese characters)
- Mixed CJK + Latin
- Whitespace-only string

A quick \`#[test]\` would prevent regressions if the heuristic is tuned later.

src/channels/web/ws.rs
```diff
@@ -77,7 +77,8 @@ pub async fn handle_ws_connection(
 
     // Subscribe to broadcast events (same source as SSE), scoped to this user.
     // Reject if we've hit the connection limit.
-    let Some(raw_stream) = state.sse.subscribe_raw(Some(user.user_id.clone())) else {
+    // TODO: Accept debug flag from WebSocket query params when needed.
+    let Some(raw_stream) = state.sse.subscribe_raw(Some(user.user_id.clone()), false) else {
```
serrrfirat: **Low Severity — WebSocket clients cannot receive debug events**

WebSocket connections always pass \`verbose: false\` here, so WS-based clients can never receive \`ToolResultFull\` or \`TurnMetrics\` events. The TODO comment acknowledges this.

Since the debug panel uses SSE (not WS), this isn't blocking, but it means any future WS-based debug tooling would silently miss debug events.

**Suggested fix:** Accept a \`debug\` query parameter from the WebSocket upgrade request URL and pass it through, similar to how \`chat_events_handler\` uses \`ChatEventsQuery\`.

src/channels/web/server.rs
```diff
@@ -197,6 +196,7 @@ pub struct ActiveConfigSnapshot {
     pub llm_backend: String,
     pub llm_model: String,
     pub enabled_channels: Vec<String>,
+    pub default_timezone: String,
```
serrrfirat: **Nit — Implicit empty-string-as-UTC fallback**

\`ActiveConfigSnapshot\` derives \`Default\`, so \`default_timezone\` defaults to \`""\`. The debug prompt handler at line 3358 handles this gracefully with \`.unwrap_or(chrono_tz::UTC)\`, but the empty-string → UTC fallback is implicit.

Consider either defaulting to \`"UTC"\` in the struct initializer or adding a brief comment at the parse site to make the fallback explicit.

src/agent/dispatcher.rs
```diff
@@ -1195,6 +1214,30 @@ impl<'a> LoopDelegate for ChatDelegate<'a> {
                                 &self.message.metadata,
                             )
                             .await;
+
+                        // Send full (non-truncated) output for debug subscribers.
+                        const MAX_TOOL_OUTPUT_BYTES: usize = 50_000;
```
serrrfirat, line 1215: **Medium Severity — Performance**

`ToolResultFull` clones up to 50 KB of tool output on **every** tool call, unconditionally — even when zero debug subscribers exist. The clone, `StatusUpdate` construction, channel broadcast, SSE serialization, and `filter_map` discard all happen for nothing in production. The same pattern applies to `TurnMetrics` at line ~705 (allocates model name string per LLM call).

```rust
// This runs on every tool call, not just when debug clients are connected:
let capped = if truncated {
    output[..boundary].to_string()
} else {
    output.clone()  // up to 50KB allocation, wasted if no verbose subscriber
};
```

**Suggested fix:** Add `SseManager::has_verbose_subscribers() -> bool` (atomic counter of verbose connections, incremented in `subscribe_raw(verbose=true)`, decremented on drop). Skip emitting `ToolResultFull` and `TurnMetrics` when the count is 0. This keeps the hot path allocation-free in production while preserving full functionality when a debug panel is open.

src/channels/web/handlers/chat.rs
```diff
@@ -46,7 +53,11 @@ pub async fn chat_events_handler(
 ) -> Result<impl IntoResponse, (StatusCode, String)> {
     let sse = state
         .sse
-        .subscribe(Some(user.user_id), extract_last_event_id(&params, &headers))
+        .subscribe(
+            Some(user.user_id),
+            params.debug,
```
serrrfirat: **Medium Severity — Security (Design)**

Inconsistent auth gating for debug features: `/api/debug/prompt` requires `AdminUser` (role == "admin", 403 for members), but SSE `debug=true` — which exposes `ToolResultFull` with up to 50 KB of raw tool output and `TurnMetrics` with model name, token counts, and timing — only requires `AuthenticatedUser`. Any member-role user can pass `?debug=true` on the SSE endpoint to receive verbose events.

This may be intentional since events are still user-scoped (User A cannot see User B's debug events). If so, it would be good to document this design decision. If verbose events should be admin-only:

```rust
// In chat_events_handler:
let verbose = params.debug && user.role == "admin";
```

At minimum, consider logging when verbose mode is activated so it's auditable:
```rust
if params.debug {
    tracing::debug!(user_id = %user.user_id, "SSE verbose mode activated");
}
```

crates/ironclaw_gateway/static/debug-panel.js
```diff
@@ -0,0 +1,1266 @@
...
+        sessionStats.toolCalls++;
+        updateStatsDisplay();
+      } catch (_) { /* ignore */ }
+      lastEventTime = Date.now(); totalEventsReceived++;
+    });
+
+    es.addEventListener('tool_completed', function (e) {
+      try {
+        var data = JSON.parse(e.data);
+        var queue = pendingTools[data.name] || [];
```
serrrfirat: **Medium Severity — Correctness**

FIFO matching of \`tool_started\` → \`tool_completed\` by **tool name** is fragile when the same tool runs concurrently (e.g. multiple \`memory_search\` or \`http\` calls in a parallel tool-use turn). Completions will be misattributed to the wrong invocation. The 5-second \`setTimeout\` delay before \`queue.shift()\` makes this worse — stale entries accumulate during bursts.

```js
var queue = pendingTools[data.name] || [];
var pending = queue[0]; // FIFO: oldest start matches oldest completion
// ^ If tool "http" was started 3 times concurrently, completion order
//   is not guaranteed to match start order (network latency varies)
```

**Suggested fix:** The backend already emits \`call_id\` in \`ToolCompleted\` events. Include \`call_id\` in the \`tool_started\` SSE event as well, then match on the unique ID instead of the tool name:

```js
// In tool_started handler:
pendingTools[data.call_id] = { id: id, start: Date.now(), name: data.name };

// In tool_completed handler:
var pending = pendingTools[data.call_id];
if (pending) { /* update */ delete pendingTools[data.call_id]; }
```

This eliminates both the name-collision and the stale-queue timeout issues.

src/channels/web/server.rs
```diff
@@ -3899,6 +3936,102 @@ struct GatewayStatusResponse {
...
+    }
+    if chars / words > 10 {
+        // Character-based: ~1.5 chars per token for CJK text
+        (chars as f64 / 1.5) as usize + 4
+    } else {
+        ((words as f64) * 1.3) as usize + 4
+    }
+}
+
+async fn debug_prompt_handler(
```
serrrfirat: **Medium Severity — Security (Design)**

This endpoint returns the **full assembled system prompt** including all identity files (AGENTS.md, SOUL.md, USER.md, IDENTITY.md, TOOLS.md, MEMORY.md) and the complete \`system_prompt\` string. While properly gated behind \`AdminUser\`, this creates a one-request system prompt extraction vector.

In shared/enterprise deployments where admin ≠ owner, this could expose:
- Safety layer instructions and behavioral constraints
- Confidential prompt engineering (competitive IP)
- Long-term memory contents
- Tool access configuration

**Suggestions** (pick any):
1. Omit the \`system_prompt\` field by default; require an explicit \`?include_full=true\` param for the assembled prompt
2. Add an audit log entry (\`tracing::info!\`) when this endpoint is hit, so access is observable
3. Return component metadata (source, label, token count) without the actual \`content\` field unless explicitly requested
4. At minimum, document in the security section of the PR that this endpoint exposes full prompt data to any admin user

italic-jinxin:
@serrrfirat resolved your comments

src/channels/web/server.rs
```diff
@@ -2685,8 +2721,9 @@ async fn chat_ws_handler(
             "WebSocket origin not allowed".to_string(),
         ));
     }
+    let debug = params.debug;
```
serrrfirat: **High Severity** — Missing admin role gate on WebSocket debug mode

`server.rs:2724` sets `let debug = params.debug;` without checking the user's role. Any authenticated (non-admin) user can pass `?debug=true` on the WebSocket URL and receive verbose events including `ToolResultFull` (up to 50 KB of raw tool output — file contents, shell output, potentially secrets) and `TurnMetrics` (model names, token counts).

The SSE handler correctly gates this at `handlers/chat.rs:52`:
```rust
let verbose = params.debug && user.role == "admin";
```

And the **dead** `chat_ws_handler` in `handlers/chat.rs:113` also has the correct pattern:
```rust
let debug = params.debug && identity.role == "admin";
```

But the route `/api/chat/ws` (registered at `server.rs:605`) uses the local `server.rs` handler, not the `handlers/chat.rs` one.

**Fix:**
```rust
let debug = params.debug && user.role == "admin";
```

crates/ironclaw_common/src/event.rs
```diff
@@ -272,6 +278,30 @@ pub enum AppEvent {
...
 
+    /// Full (non-truncated) tool output (verbose/debug mode only).
+    #[serde(rename = "tool_result_full")]
+    ToolResultFull {
+        name: String,
+        output: String,
+        #[serde(skip_serializing_if = "Option::is_none")]
+        truncated: Option<bool>,
+        #[serde(skip_serializing_if = "Option::is_none")]
+        thread_id: Option<String>,
```
serrrfirat, line 326: **Medium Severity** — `ToolResultFull` missing `call_id` causes wrong-tool matching for concurrent same-name calls

All three existing tool events (`ToolStarted`, `ToolCompleted`, `ToolResult`) now carry `call_id: Option<String>` for disambiguation. The new `ToolResultFull` event does not. The debug panel JS acknowledges this at `debug-panel.js:1223`:

```javascript
// tool_result_full doesn't carry call_id; find by name as fallback
```

When two parallel tool calls share the same name (e.g., two `shell` commands), the full output will be attached to whichever pending entry is found first — potentially the wrong one.

**Fix:** Add `call_id: Option<String>` to both `StatusUpdate::ToolResultFull` and `AppEvent::ToolResultFull`, and pass `tc.id.clone()` from the dispatcher emission site.

src/channels/web/handlers/chat.rs
```diff
@@ -99,8 +110,9 @@ pub async fn chat_ws_handler(
             "WebSocket origin not allowed".to_string(),
         ));
     }
+    let debug = params.debug && identity.role == "admin";
```
serrrfirat: **Medium Severity** — Dead code: this `chat_ws_handler` is never registered on any route

This function is `pub` but never imported by `server.rs`. The route `/api/chat/ws` (at `server.rs:605`) uses a **local** `chat_ws_handler` defined at `server.rs:2697` instead. That local version is missing the admin role gate you correctly added here (`params.debug && identity.role == "admin"`).

This dead code likely represents the intended behavior — the admin gating was written but wired to the wrong handler.

**Fix:** Either:
1. Import and register this version on the route (like `chat_events_handler` is already imported from this module at `server.rs:38`), or
2. Add the admin check to the `server.rs` local version and delete this dead function.

italic-jinxin:
@serrrfirat resolved.

ilblackdragon:
## Review Summary

Solid v1 of an admin-only debug inspector. The security gating, `call_id` correlation, and verbose-event short-circuit are well thought out. Two correctness bugs in the JS panel and a meaningful merge-conflict surface to address before this can land.

### Blockers

- **Merge conflicts with staging** (`mergeStateStatus: DIRTY`). Since this PR opened, staging landed `a35f9d9a` (split `app.js`/`style.css` into per-surface modules) and `5fe3be8d` (extracted `src/channels/web/features/chat/`). The PR still edits the monolithic `app.js`, `handlers/chat.rs`, and adds `chat_ws_handler` to `server.rs` — `chat_ws_handler` already lives at `features/chat/mod.rs:367`. Non-trivial rebase required; the 1285-line monolithic `debug-panel.js` also runs against the new modular `features/` pattern.

### Correctness issues

- **`debug-panel.js` cost double-counts.** The `turn_cost` SSE handler accumulates into `sessionStats.cost` (~line 1172), then every 30 s `fetchGatewayStats()` overwrites it with `data.daily_cost` (~line 2003). `daily_cost` is daily, not session — pick one source.
- **Activity-log eviction is silent.** `MAX_ACTIVITY = 1000`; once exceeded, older turns silently disappear and turn-nav `viewTurn(1)` renders \"no events\" even though turn 1 happened. Either disable nav for evicted turns or surface that they were dropped.
- **Verbose broadcast not gated by verbose subscriber count.** `has_receivers()` is true as soon as *any* SSE/WS user connects, so a single normal user keeps the 50 KB clone path active. Consider a separate `has_verbose_receivers()` counter.

### Minor

- `is_verbose_only()` duplicated on `AppEvent` and `StatusUpdate` — easy to drift when a new verbose variant is added.
- PR description says \"consolidate `theme-init.js` into `init.js`\" but `theme-init.js` is still referenced by `admin.html`. Either finish the consolidation or update the description.
- `MAX_TOOL_OUTPUT_BYTES = 50_000` hardcoded in `dispatcher.rs` — worth a brief comment on the choice.
- PR body mistypes the new endpoint as \`/api/debug/debug/prompt\` (actual: \`/api/debug/prompt\`).

### Test gaps

- No integration test for: admin downgrade when non-admin requests `?debug=true`, the WS admin gate, or `debug_prompt_handler` with empty workspace files.

### Strengths

- Admin-gated for both SSE and WS; `/api/debug/prompt` uses `AdminUser` extractor with `tracing::debug!` audit log.
- `is_verbose_only()` + `has_receivers()` short-circuit avoids cloning ≤50 KB output when no subscribers connected.
- Threading `call_id` through tool events fixes mismatched start/complete pairs for concurrent same-name tools.
- `test_verbose_filtering` and `estimate_tokens` ASCII/CJK/mixed/empty test coverage.
- Full i18n coverage (en/ko/zh-CN).

### Risk

Medium blast radius (touches `StatusUpdate`/`AppEvent` enums and SSE subscriber API) but off-path for non-admins. Rollback is trivial (gated by `?debug=true` + admin role).

I'm picking this up to rebase onto staging and address the issues above.

ilblackdragon (approved):

PR #1908, opened 2026-04-02
feat(tools): persistent per-user tool permission system

henrypark133:
## Summary

Adds a persistent, per-user tool permission system so that safe built-in tools (memory, time, json, etc.) run without approval prompts by default, while destructive tools (shell, file writes, secret management) still require explicit user confirmation. Users can manage tool states via the web settings UI, in-conversation via text-based approval, or through the LLM-callable `tool_permission_set` tool.

**Problem**: Today all tools use `ApprovalRequirement::UnlessAutoApproved` or `Always` hardcoded in Rust, and the "Always Approve" session state is in-memory only — cleared on every restart. Non-technical users are overwhelmed by repeated approval prompts for safe tools like `memory_search` or `time`.

**Solution**: Three-state permission model (`AlwaysAllow | AskEachTime | Disabled`) with tiered defaults, backed by per-user DB settings.

### Permission Model

```
PermissionState::AlwaysAllow   → executes without prompting
PermissionState::AskEachTime   → existing approval card/text flow
PermissionState::Disabled      → excluded from tool_definitions; LLM never sees it
```

### Precedence (highest → lowest)
1. `ApprovalRequirement::Always` on the Tool trait → **hard floor, unoverrideable** — always prompts regardless of user setting
2. `PermissionState::Disabled` → excluded from tool_definitions
3. `PermissionState::AlwaysAllow` + trait is NOT `Always` → auto-executes
4. `PermissionState::AskEachTime` → existing approval flow

### Persistence
- **DB = source of truth** (per-user via `TenantScope`). Works for both single-tenant and multi-tenant deployments.
- **Settings struct = in-memory cache** loaded from DB via `get_all_settings()` + `Settings::from_db_map()`
- **`TOOL_RISK_DEFAULTS`** static map seeds the DB at startup for any tool with no existing entry — runs once per tool per user, never overwrites

### Tier Defaults

| Always Allow (safe, read/create) | Ask Each Time (destructive, install, execute) |
|---|---|
| echo, time, json | shell, read_file, write_file, list_dir, apply_patch |
| memory_search/read/write/tree | cancel_job, job_prompt |
| tool_list, tool_info, tool_search | routine_delete, routine_fire |
| skill_list, skill_search | tool_install/auth/activate/remove/upgrade |
| list_jobs, job_status, job_events | skill_install, skill_remove |
| image_analyze | secret_list, secret_delete |
| message, http, create_job | image_generate, image_edit |
[...]

PR #1911, opened 2026-04-02
feat(tools): persistent per-user tool permission system

henrypark133:
## Summary

Adds a persistent, per-user tool permission system so that safe built-in tools (memory, time, json, etc.) run without approval prompts by default, while destructive tools (shell, file writes, secret management) still require explicit user confirmation. Users can manage tool states via the web settings UI, in-conversation via text-based approval, or through the LLM-callable `tool_permission_set` tool.

**Problem**: Today all tools use `ApprovalRequirement::UnlessAutoApproved` or `Always` hardcoded in Rust, and the "Always Approve" session state is in-memory only — cleared on every restart. Non-technical users are overwhelmed by repeated approval prompts for safe tools like `memory_search` or `time`.

**Solution**: Three-state permission model (`AlwaysAllow | AskEachTime | Disabled`) with tiered defaults, backed by per-user DB settings.

### Permission Model

```
PermissionState::AlwaysAllow   → executes without prompting
PermissionState::AskEachTime   → existing approval card/text flow
PermissionState::Disabled      → excluded from tool_definitions; LLM never sees it
```

### Precedence (highest → lowest)
1. `ApprovalRequirement::Always` on the Tool trait → **hard floor, unoverrideable** — always prompts regardless of user setting
2. `PermissionState::Disabled` → excluded from tool_definitions
3. `PermissionState::AlwaysAllow` + trait is NOT `Always` → auto-executes
4. `PermissionState::AskEachTime` → existing approval flow

### Persistence
- **DB = source of truth** (per-user via `TenantScope`). Works for both single-tenant and multi-tenant deployments.
- **Settings struct = in-memory cache** loaded from DB via `get_all_settings()` + `Settings::from_db_map()`
- **`TOOL_RISK_DEFAULTS`** static map seeds the DB at startup for the owner — other users get tier defaults at runtime via `effective_permission()` fallback (no explicit DB rows needed)

### Tier Defaults

| Always Allow (safe, read/create) | Ask Each Time (destructive, install, execute, network) |
|---|---|
| echo, time, json | shell, read_file, write_file, list_dir, apply_patch |
| memory_search/read/write/tree | cancel_job, job_prompt |
| tool_list, tool_info, tool_search | routine_delete, routine_fire |
| skill_list, skill_search | tool_install/auth/activate/remove/upgrade |
| list_jobs, job_status, job_events | skill_install, skill_remove |
| image_analyze | secret_list, secret_delete |
| message | image_generate, image_edit |
 | restart, build_software, tool_permission_set |

### New files
- **`src/tools/permissions.rs`** — `PermissionState` enum, `TOOL_RISK_DEFAULTS` static map, `effective_permission()` resolution function
- **`tests/e2e/scenarios/test_tool_permissions.py`** — 6 Playwright E2E test scenarios

### Modified files
- **`src/settings.rs`** — Added `tool_permissions: HashMap<String, PermissionState>` field with `#[serde(default)]`
- **`src/app.rs`** — `seed_tool_permissions()` writes tier defaults to DB at startup; `register_permission_tools()` + `upgrade_tool_list()` wire new tools
- **`src/agent/dispatcher.rs`** — `before_llm_call()`: loads tool_permissions from per-user DB, filters `Disabled` tools from LLM context, batch-approves `AlwaysAllow` tools in session (single lock acquisition)
- **`src/agent/thread_ops.rs`** — `process_approval()`: when user clicks "Always Approve" (web) or types "always" (text), persists `AlwaysAllow` to per-user DB settings
- **`src/tools/builtin/extension_tools.rs`** — New `tool_permission_set` tool (always requires `ApprovalRequirement::Always`); extended `tool_list` with `kind="builtin"` filter and permission state fields
- **`src/tools/registry.rs`** — Registration for `tool_permission_set`, protected name
- **`src/channels/web/handlers/settings.rs`** — `GET /api/settings/tools` and `PUT /api/settings/tools/:name` endpoints (DB-backed, per-user scoped)
- **`src/channels/web/server.rs`** — Route wiring for new endpoints
- **`src/channels/web/static/app.js`** — Tools settings tab with 3-way toggle, lock icons for hard-floor tools
- **`src/channels/web/static/style.css`** — Styles for tool permission rows, toggles, lock icons, default badges
- **`src/channels/web/static/index.html`** — Tools tab button in settings sidebar
- **`src/channels/web/static/i18n/en.js`** — i18n strings for permission states and UI labels
- **`src/channels/web/types.rs`** — `ToolPermissionEntry`, `ToolPermissionsResponse`, `UpdateToolPermissionRequest` DTOs
- **`tests/e2e/helpers.py`** — New selectors for Tools tab UI elements

## Security

- `ApprovalRequirement::Always` on the Tool trait is an **unbypassable hard floor** — even if a user sets a tool to `AlwaysAllow`, if the tool returns `Always` from `requires_approval()`, it still prompts. The existing approval gate enforces this independently of session auto-approve state.
- `tool_permission_set` itself returns `ApprovalRequirement::Always` — the LLM can never silently change permissions; every proposed change shows an approval card/text prompt requiring explicit user confirmation.
- Locked tools (those returning `ApprovalRequirement::Always`) cannot have their permission changed via either the web UI (returns 400) or the LLM tool (returns error).
- All DB operations are scoped to the authenticated `user_id` via `TenantScope` — one user's permission changes cannot affect another user in multi-tenant deployments.
- Note: `TenantScope.get_all_settings()` and `TenantScope.set_setting()` automatically bind `user_id` internally — callers do not pass `user_id` separately.

## Test plan

- [x] `cargo fmt` — clean
- [x] `cargo test` — 4,148+ tests pass, 0 failures
- [ ] `pytest tests/e2e/scenarios/test_tool_permissions.py` — 6 E2E scenarios (requires running server)
- [ ] Manual: web UI toggle persists across page reload
- [ ] Manual: Telegram "always" response persists across agent restart
- [ ] Manual: `tool_permission_set` via LLM shows approval card, persists on approve

Supersedes #1908 (closed — had merge commits inflating diff to 105 files).

### DIFF
--- a/src/agent/dispatcher.rs
+++ b/src/agent/dispatcher.rs
@@ -20,6 +20,7 @@ use crate::agent::agentic_loop::{
     AgenticLoopConfig, LoopDelegate, LoopOutcome, LoopSignal, TextAction,
 };
 use crate::llm::{ChatMessage, Reasoning, ReasoningContext};
+use crate::tools::permissions::{PermissionState, effective_permission};
 use crate::tools::redact_params;
 
 /// Result of the agentic loop execution.
@@ -199,6 +200,7 @@ impl Agent {
             nudge_at,
             force_text_at,
             user_tz,
+            cached_tool_permissions: std::sync::Mutex::new(None),
         };
 
         // If /skill-name mentions were expanded, rewrite the last user message
@@ -293,6 +295,8 @@ struct ChatDelegate<'a> {
     nudge_at: usize,
     force_text_at: usize,
     user_tz: chrono_tz::Tz,
+    cached_tool_permissions:
+        std::sync::Mutex<Option<std::collections::HashMap<String, PermissionState>>>,
 }
 
 #[async_trait]
@@ -343,6 +347,94 @@ impl<'a> LoopDelegate for ChatDelegate<'a> {
         } else {
             tool_defs
         };
+
+        // Apply per-user tool permission filtering.
+        //
+        // Load tool_permissions from the per-user DB settings store (same
+        // source as selected_model). Falls back to empty map when no store is
+        // available (test rigs without a tenant) — tier defaults from
+        // TOOL_RISK_DEFAULTS then apply at runtime via effective_permission().
+        // Disabled tools are excluded from the LLM's tool list entirely.
+        // AlwaysAllow tools are pre-approved in session so the approval
+        // flow is skipped — unless the tool declares ApprovalRequirement::Always,
+        // which is an unbypassable hard floor.
+        let tool_permissions = {
+            // Check the cache first (brief lock, no await while held).
+            let cached = {
+                let cache = self
+                    .cached_tool_permissions
+                    .lock()
+                    .unwrap_or_else(|e| e.into_inner());
+                cache.clone()
+            };
+            if let Some(perms) = cached {
+                perms
+            } else {
+                // Cache miss — load from DB (async).
+                let perms = if let Some(store) = self.tenant.store() {
+                    match store.get_all_settings().await {
+                        Ok(db_map) => {
+                            crate::settings::Settings::from_db_map(&db_map).tool_permissions
+                        }
+                        Err(e) => {
+                            tracing::warn!(
+                                "Failed to load tool permissions, keeping existing session state: {}",
+                                e
+                            );
+                            // Fail closed: preserve the previously filtered available_tools
+                            // rather than publishing the unfiltered tool list, which could
+                            // re-expose tools explicitly marked Disabled.
+                            return None;
+                        }
+                    }
+                } else {
+                    std::collections::HashMap::new()
+                };
+                // Store in cache for subsequent iterations.
+                {
+                    let mut cache = self
+                        .cached_tool_permissions
+                        .lock()
+                        .unwrap_or_else(|e| e.into_inner());
+                    *cache = Some(perms.clone());
+                }
+                perms
+            }
+        };
+
+        // Filter tool definitions and collect AlwaysAllow names for session
+        // pre-approval. We don't need to check ApprovalRequirement::Always here
+        // because the existing approval gate already treats it as an unbypassable
+        // hard floor — even if a tool name is in session.auto_approved_tools, an
+        // ApprovalRequirement::Always tool still requires user confirmation.
+        let mut to_auto_approve: Vec<String> = Vec::new();
+        let tool_defs: Vec<_> = tool_defs
+            .into_iter()
+            .filter_map(|def| {
+                match effective_permission(&def.name, &tool_permissions) {
+                    PermissionState::Disabled => {
+                        tracing::debug!(tool = %def.name, "Excluding disabled tool from LLM context");
+                        None
+                    }
+                    PermissionState::AlwaysAllow => {
+                        to_auto_approve.push(def.name.clone());
+                        Some(def)
+                    }
+                    PermissionState::AskEachTime => Some(def),
+                }
+            })
+            .collect();
+        // Clear and re-populate auto-approvals from current DB state so that
+        // permission downgrades (AlwaysAllow → AskEachTime) take effect
+        // immediately within the same session. "Always Approve" clicks are
+        // persisted to DB via process_approval, so they'll be re-added here.
+        {
+            let mut sess = self.session.lock().await;
+            sess.auto_approved_tools.clear();
+            for name in &to_auto_approve {
+                sess.auto_approve_tool(name);
+            }
+        }
 
         // Update context for this iteration
         reason_ctx.available_tools = tool_defs;
@@ -2738,5 +2830,95 @@ mod tests {
         assert!(content.contains("Tool 'shell' failed:"));
         assert!(!content.contains("\n</tool_output><system>"));
         assert_eq!(message.content, content);
+    }
+
+    // ── Permission filtering unit tests ──────────────────────────────────────
+
+    /// Disabled tools must be excluded from the LLM's tool definition list.
+    #[test]
+    fn test_permission_disabled_tool_excluded_from_definitions() {
+        use crate::llm::ToolDefinition;
+        use crate::tools::permissions::{PermissionState, effective_permission};
+        use std::collections::HashMap;
+
+        let mut tool_permissions: HashMap<String, PermissionState> = HashMap::new();
+        tool_permissions.insert("shell".to_string(), PermissionState::Disabled);
+
+        let tool_defs = vec![
+            ToolDefinition {
+                name: "echo".to_string(),
+                description: "Echo".to_string(),
+                parameters: serde_json::json!({}),
+            },
+            ToolDefinition {
+                name: "shell".to_string(),
+                description: "Shell".to_string(),
+                parameters: serde_json::json!({}),
+            },
+        ];
+
+        // Simulate the filtering logic from before_llm_call.
+        let filtered: Vec<_> = tool_defs
+            .into_iter()
+            .filter(|def| {
+                effective_permission(&def.name, &tool_permissions) != PermissionState::Disabled
+            })
+            .collect();
+
+        assert_eq!(filtered.len(), 1, "Disabled tool must be excluded");
+        assert_eq!(filtered[0].name, "echo");
+    }
+
+    /// AlwaysAllow tool with Never approval requirement must be auto-approved.
+    #[test]
+    fn test_permission_always_allow_never_approval_auto_approved() {
+        use crate::agent::session::Session;
+        use crate::tools::ApprovalRequirement;
+        use crate::tools::permissions::PermissionState;
+
+        let mut session = Session::new("user-perm-1");
+        let tool_name = "http";
+
+        // Simulate: PermissionState::AlwaysAllow and requires_approval → Never.
+        let perm = PermissionState::AlwaysAllow;
+        let requirement = ApprovalRequirement::Never;
+
+        let hard_floor = matches!(requirement, ApprovalRequirement::Always);
+        if perm == PermissionState::AlwaysAllow && !hard_floor {
+            session.auto_approve_tool(tool_name);
+        }
+
+        assert!(
+            session.is_tool_auto_approved(tool_name),
+            "AlwaysAllow with Never approval requirement must be auto-approved in session"
+        );
+    }
+
+    /// AlwaysAllow tool with Always approval requirement must NOT be auto-approved.
+    ///
+    /// This verifies the hard-floor: ApprovalRequirement::Always is never bypassed,
+    /// even when PermissionState is AlwaysAllow.
+    #[test]
+    fn test_permission_always_allow_always_approval_not_auto_approved() {
+        use crate::agent::session::Session;
+        use crate::tools::ApprovalRequirement;
+        use crate::tools::permissions::PermissionState;
+
+        let mut session = Session::new("user-perm-2");
+        let tool_name = "restart";
+
+        // Simulate: PermissionState::AlwaysAllow but requires_approval → Always.
+        let perm = PermissionState::AlwaysAllow;
+        let requirement = ApprovalRequirement::Always;
+
+        let hard_floor = matches!(requirement, ApprovalRequirement::Always);
+        if perm == PermissionState::AlwaysAllow && !hard_floor {
+            session.auto_approve_tool(tool_name);
+        }
+
+        assert!(
+            !session.is_tool_auto_approved(tool_name),
+            "AlwaysAllow with Always approval requirement must NOT be auto-approved (hard floor)"
+        );
     }
 }
--- a/src/agent/thread_ops.rs
+++ b/src/agent/thread_ops.rs
@@ -1064,7 +1064,7 @@ impl Agent {
         }
 
         if approved {
-            // If always, add to auto-approved set
+            // If always, add to auto-approved set and persist to settings.
             if always {
                 let mut sess = session.lock().await;
                 sess.auto_approve_tool(&pending.tool_name);
@@ -1073,6 +1073,52 @@ impl Agent {
                     pending.tool_name,
                     sess.id
                 );
+                drop(sess);
+
+                // Defense-in-depth: don't persist AlwaysAllow for tools that
+                // declare ApprovalRequirement::Always (the UI hides the
+                // "Always" button for locked tools, but a crafted client
+                // could send it).
+                let tool_ref = self.tools().get(&pending.tool_name).await;
+                let is_locked = tool_ref
+                    .as_ref()
+                    .map(|t| {
+                        matches!(
+                            t.requires_approval(&serde_json::json!({})),
+                            crate::tools::ApprovalRequirement::Always
+                        )
+                    })
+                    .unwrap_or(false);
+
+                if is_locked {
+                    tracing::warn!(
+                        tool = %pending.tool_name,
+                        "Skipping AlwaysAllow persist — tool declares ApprovalRequirement::Always"
+                    );
+                } else {
+                    // Persist AlwaysAllow to the per-user DB settings so the
+                    // preference survives process restarts. Uses the same
+                    // set_setting path as tool_permission_set and the web UI.
+                    let tenant = self.tenant_ctx(&message.user_id).await;
+                    if let Some(store) = tenant.store() {
+                        let key = format!("tool_permissions.{}", pending.tool_name);
+                        let val = serde_json::to_value(
+                            crate::tools::permissions::PermissionState::AlwaysAllow,
+                        )
+                        .unwrap_or(serde_json::Value::String("always_allow".to_string()));
+                        match store.set_setting(&key, &val).await {
+                            Ok(()) => tracing::debug!(
+                                tool = %pending.tool_name,
+                                "Persisted AlwaysAllow permission to DB settings"
+                            ),
+                            Err(e) => tracing::warn!(
+                                "process_approval: failed to persist AlwaysAllow for '{}': {}",
+                                pending.tool_name,
+                                e
+                            ),
+                        }
+                    }
+                } // else (not locked)
             }
 
             // Reset thread state to processing
@@ -2324,6 +2370,8 @@ mod tests {
         // Verify nothing was queued — the fall-through path doesn't touch the queue.
         assert!(t.pending_messages.is_empty());
     }
+
+    // Approval persistence is tested via e2e_builtin_tool_coverage integration tests.
 
     // Helper function to extract the approval message without needing a full Agent instance
     fn extract_approval_message(
--- a/src/app.rs
+++ b/src/app.rs
@@ -367,7 +367,7 @@ impl AppBuilder {
                 self.config.workspace.clone(),
             ));
             tools.register_memory_tools_with_resolver(pool);
-            tracing::info!(
+            tracing::debug!(
                 multi_tenant = is_multi_tenant,
                 "Memory tools configured with per-user workspace resolver"
             );
@@ -739,6 +739,16 @@ impl AppBuilder {
                 catalog_entries.clone(),
             ));
             tools.register_extension_tools(Arc::clone(&manager));
+
+            // Register permission management tool and upgrade tool_list with
+            // builtin registry support.
+            let settings_store_for_perms: Option<Arc<dyn crate::db::SettingsStore + Send + Sync>> =
+                self.db
+                    .as_ref()
+                    .map(|db| Arc::clone(db) as Arc<dyn crate::db::SettingsStore + Send + Sync>);
+            tools.register_permission_tools(settings_store_for_perms.clone());
+            tools.upgrade_tool_list(Arc::clone(&manager), settings_store_for_perms);
+
             tracing::debug!("Extension manager initialized with in-chat discovery tools");
 
             if !startup_mcp_clients.is_empty() {
@@ -937,6 +947,11 @@ impl AppBuilder {
             tools.count()
         );
 
+        // Seed per-user tool permission defaults into the database.
+        // This runs after all tools (built-in, WASM, MCP) are registered so
+        // that every tool name is known.  Existing entries are never overwritten.
+        seed_tool_permissions(&tools, self.db.as_ref(), &self.config.owner_id).await;
+
         Ok(AppComponents {
             config: self.config,
             db: self.db,
@@ -964,6 +979,80 @@ impl AppBuilder {
             dev_loaded_tool_names,
             builder,
         })
+    }
+}
+
+/// Seed tool permission defaults into the database for every registered tool
+/// that has no explicit user override yet.
+///
+/// This is called once at startup after the full tool registry is built.
+/// It is idempotent: existing entries in `tool_permissions.*` are never touched.
+async fn seed_tool_permissions(
+    tools: &crate::tools::ToolRegistry,
+    db: Option<&Arc<dyn Database>>,
+    owner_id: &str,
+) {
+    use crate::tools::permissions::{TOOL_RISK_DEFAULTS, effective_permission};
+
+    let db = match db {
+        Some(db) => db,
+        None => {
+            tracing::debug!("seed_tool_permissions: no database available, skipping");
+            return;
+        }
+    };
+
+    // Load existing tool permission overrides from the DB.
+    let db_map = match db.get_all_settings(owner_id).await {
+        Ok(m) => m,
+        Err(e) => {
+            tracing::warn!("seed_tool_permissions: failed to load settings: {}", e);
+            return;
+        }
+    };
+    let existing = crate::settings::Settings::from_db_map(&db_map).tool_permissions;
+
+    let registered_names = tools.list().await;
+    let mut seeded = 0u32;
+
+    for name in &registered_names {
+        if existing.contains_key(name.as_str()) {
+            // User has an explicit override — do not touch it.
+            continue;
+        }
+
+        // Only insert if the tool appears in the static defaults table.
+        // Unknown/dynamic tools stay absent (they will fall back to AskEachTime
+        // at runtime via effective_permission) to avoid polluting the DB.
+        if TOOL_RISK_DEFAULTS.contains_key(name.as_str()) {
+            let default_state = effective_permission(name, &existing);
+            let json_value = match serde_json::to_value(default_state) {
+                Ok(v) => v,
+                Err(e) => {
+                    tracing::warn!(
+                        "seed_tool_permissions: failed to serialize state for '{}': {}",
+                        name,
+                        e
+                    );
+                    continue;
+                }
+            };
+            if let Err(e) = db
+                .set_setting(owner_id, &format!("tool_permissions.{}", name), &json_value)
+                .await
+            {
+                tracing::warn!("seed_tool_permissions: failed to set '{}': {}", name, e);
+            } else {
+                seeded += 1;
+            }
+        }
+    }
+
+    if seeded > 0 {
+        tracing::debug!(
+            count = seeded,
+            "Seeded tool permission defaults into database"
+        );
     }
 }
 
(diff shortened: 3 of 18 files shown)


henrypark133:
## All review comments addressed

All feedback from Copilot and Gemini has been incorporated in the latest push. Here's the summary:

### Copilot feedback
| Comment | Resolution |
|---------|-----------|
| Auto-approve not reversible in session (`dispatcher.rs`) | ✅ Fixed — `sess.auto_approved_tools.clear()` + re-populate from current DB state each iteration (line 359) |
| `process_approval` persists AlwaysAllow for locked tools (`thread_ops.rs`) | ✅ Fixed — defense-in-depth check skips persist when `requires_approval() == Always` |
| "tools" in TABS breaks test_connection.py (`helpers.py`) | ✅ Fixed — removed from TABS |
| `_open_tools_tab` doesn't click Settings tab first | ✅ Fixed — clicks main Settings tab, waits, then clicks Tools subtab |
| Approval test uses non-UUID request_id | ✅ Fixed — rewritten to use `PUT /api/settings/tools/echo` REST API directly |
| `tool_permissions_list` in TOOL_RISK_DEFAULTS doesn't exist | ✅ Fixed — removed |
| `tool_permission_set` silently succeeds with no settings store | ✅ Fixed — returns `ToolError::ExecutionFailed` when no store configured |
| `register_permission_tools()` docstring mismatch | ✅ Fixed — updated to match actual behavior |
| Extra DB round-trip in `settings_tools_set_handler` | ✅ Fixed — uses `new_state` directly |

### Gemini feedback
| Comment | Resolution |
|---------|-----------|
| Refactor to `filter_map` iterator (`dispatcher.rs`) | ✅ Applied |
| Return `&'static str` from `permission_state_to_str` | ✅ Applied |
| Log warning on `get_all_settings` failure (`extension_tools.rs`) | ✅ Applied |
| Dead `settings.tool_permissions.insert()` | ✅ Removed |
| `tool_permissions_list` nonexistent in defaults | ✅ Removed (same as Copilot) |
| Misleading docstring on `register_permission_tools` | ✅ Fixed (same as Copilot) |
| Test name `test_disabled_tool_absent_from_api` misleading | ✅ Renamed to `test_disabled_tool_state_is_persisted` |

### Additional fix
- `tool_list(kind="builtin")` now filters via `registry.builtin_tool_names()` to avoid duplicating WASM/MCP tools
- Fixed `needless_borrow` clippy lint caught by CI (Rust 1.94.0)

src/tools/permissions.rs
```diff
@@ -0,0 +1,186 @@
...
+            "skill_list",
+            "skill_search",
+            "list_jobs",
+            "job_status",
+            "job_events",
+            "image_analyze",
+            "message",
+            "http",
+            "create_job",
+            "event_emit",
```
serrrfirat: **High Severity — Security: `http` tool defaults to `AlwaysAllow` enables SSRF / data exfiltration**

The `http` tool returns `UnlessAutoApproved` for POST requests and requests with auth headers. With `AlwaysAllow` as the tier default, the dispatcher auto-approves it in session (`dispatcher.rs:345-351`), meaning POST requests to arbitrary URLs execute **without user confirmation**.

A prompt-injection attack could:
1. Exfiltrate conversation data via POST to an attacker-controlled server
2. Make authorized requests to external APIs using stored credentials from the credential registry

Before this PR, POST requests required either manual approval or a session-level "Always Approve" click (lost on restart). After, they're auto-approved by default and persisted.

**Suggested fix:** Move `http` to the `AskEachTime` list. Read-only GET requests already return `ApprovalRequirement::Never` and don't need the tool to be fully auto-approved.

src/tools/permissions.rs
```diff
@@ -0,0 +1,186 @@
...
+            "routine_create",
```
serrrfirat: **High Severity — Security: `create_job` defaults to `AlwaysAllow` enables resource abuse**

`create_job` returns `UnlessAutoApproved`. With `AlwaysAllow` as the tier default, the LLM can spawn arbitrarily many jobs without user confirmation. A prompt injection could burn compute credits or create expensive sandbox containers unnoticed.

**Suggested fix:** Move `create_job` to the `AskEachTime` list. Job creation is a resource-consuming side effect that warrants explicit user confirmation.

src/agent/dispatcher.rs
```diff
@@ -309,6 +310,58 @@ impl<'a> LoopDelegate for ChatDelegate<'a> {
...
+                }
+            })
+            .collect();
+        // Clear and re-populate auto-approvals from current DB state so that
+        // permission downgrades (AlwaysAllow → AskEachTime) take effect
+        // immediately within the same session. "Always Approve" clicks are
+        // persisted to DB via process_approval, so they'll be re-added here.
+        {
+            let mut sess = self.session.lock().await;
+            sess.auto_approved_tools.clear();
```
serrrfirat, line 433: **Medium Severity — Bug: `clear()` on every iteration causes silent loss of approval if DB write fails**

When the user clicks "Always Approve" in `process_approval` (thread_ops.rs:1067-1122):
1. Session auto-approval is set immediately (line 1070)
2. DB persist is attempted (line 1109)
3. If DB write **fails** (logged as warning, line 1114), session approval IS set for the current turn

But on the **next** `before_llm_call` iteration:
1. `sess.auto_approved_tools.clear()` wipes everything (line 359)
2. DB reload won't restore the failed write
3. The user's "Always Approve" silently stops working

**Suggested fix:** Either:
(a) Don't clear+repopulate if the DB load itself fails — keep existing session state as fallback
(b) Track session-only approvals separately from DB-backed ones and merge the two sets
(c) At minimum, if `store.set_setting()` fails, don't clear the session auto-approval on the next iteration

src/agent/dispatcher.rs
```diff
@@ -309,6 +310,58 @@ impl<'a> LoopDelegate for ChatDelegate<'a> {
...
+        // Load tool_permissions from the per-user DB settings store (same
+        // source as selected_model). Falls back to empty map when no store is
+        // available (test rigs without a tenant) — tier defaults from
+        // TOOL_RISK_DEFAULTS then apply at runtime via effective_permission().
+        // Disabled tools are excluded from the LLM's tool list entirely.
+        // AlwaysAllow tools are pre-approved in session so the approval
+        // flow is skipped — unless the tool declares ApprovalRequirement::Always,
+        // which is an unbypassable hard floor.
+        let tool_permissions = if let Some(store) = self.tenant.store()
+            && let Ok(db_map) = store.get_all_settings().await
```
serrrfirat: **Medium Severity — Performance: DB round-trip on every agentic loop iteration**

`get_all_settings()` + `Settings::from_db_map()` runs on **every** iteration of the agentic loop. Compare with `selected_model` (line 415-424 in the same file) which is only loaded at `iteration == 0`.

A typical 10-iteration agentic loop adds 10 DB round-trips per conversation turn — each deserializing the full settings map and rebuilding the `tool_permissions` HashMap.

**Suggested fix:** Cache tool permissions at iteration 0 (same pattern as `selected_model`), or cache per-session with a short TTL. Permission changes mid-conversation can take effect on the next turn — the current approach of reloading per-iteration is unnecessary for a setting that changes rarely.

src/tools/permissions.rs
```diff
@@ -0,0 +1,186 @@
...
+            "routine_update",
+        ] {
```
serrrfirat, line 53: **Medium Severity — Security: `event_emit` defaults to `AlwaysAllow` enables silent routine triggering**

`event_emit` returns `UnlessAutoApproved` (see routine.rs:1826-1830 — "Emitting an event can fire system_event routines that dispatch full_jobs with pre-authorized Always-gated tools — same escalation risk as routine_fire"). With `AlwaysAllow`, the LLM can emit events that trigger routines without user awareness.

Combined with `routine_create` (which returns `Never` for lightweight routines and is also `AlwaysAllow`), a prompt injection could:
1. Create a routine triggered by a custom event (no approval needed)
2. Emit that event to fire the routine immediately (now auto-approved)

**Suggested fix:** Move `event_emit` to `AskEachTime`. The tool's own comment documents the escalation risk.

src/channels/web/server.rs
```diff
@@ -527,6 +528,11 @@ pub async fn start_server(
         .route("/api/settings", get(settings_list_handler))
         .route("/api/settings/export", get(settings_export_handler))
         .route("/api/settings/import", post(settings_import_handler))
+        .route("/api/settings/tools", get(settings_tools_list_handler))
```
serrrfirat, line 656: **Low Severity — Architecture: Route shadowing**

`GET /api/settings/tools` (static) shadows `GET /api/settings/{key}` (parameterized) when `key="tools"`. Axum resolves static routes first, so a user setting named literally `"tools"` would be unreachable via the individual settings GET endpoint.

Similarly, `PUT /api/settings/tools/{name}` could shadow `PUT /api/settings/{key}` for `key="tools"` depending on Axum's prefix matching.

Not a bug today (no setting is named "tools"), but architecturally fragile.

**Suggested fix:** Consider a distinct route prefix like `/api/tools/permissions` to avoid namespace collision with the settings API.

src/agent/thread_ops.rs
```diff
@@ -1073,6 +1073,52 @@ impl Agent {
...
+                        tool = %pending.tool_name,
+                        "Skipping AlwaysAllow persist — tool declares ApprovalRequirement::Always"
+                    );
+                } else {
+                    // Persist AlwaysAllow to the per-user DB settings so the
+                    // preference survives process restarts. Uses the same
+                    // set_setting path as tool_permission_set and the web UI.
+                    let tenant = self.tenant_ctx(&message.user_id).await;
+                    if let Some(store) = tenant.store() {
+                        let key = format!("tool_permissions.{}", pending.tool_name);
```
serrrfirat, line 1104: **Low Severity — Edge Case: Tool names with dots break `from_db_map` parsing**

The DB key is `format!("tool_permissions.{}", pending.tool_name)`. The `Settings::set()` method splits on `.` and navigates the JSON tree. A tool named `foo.bar` would produce key `tool_permissions.foo.bar`, which `set()` interprets as a 3-level path `tool_permissions → foo → bar`, failing with "Path not found" on restart.

Built-in tools don't have dots, but dynamically installed WASM/MCP tools could.

**Suggested fix:** Either (a) validate tool names don't contain `.` at registration time, or (b) store `tool_permissions` as a single JSON object value under one DB key instead of using dotted sub-keys.

src/channels/web/handlers/settings.rs
```diff
@@ -524,6 +524,160 @@ fn mask_settings_api_keys(settings: &mut std::collections::HashMap<String, serde
...
+
+    // Validate tool exists.
+    let tool = registry.get(&name).await.ok_or(StatusCode::NOT_FOUND)?;
+
+    // Reject if tool is locked (ApprovalRequirement::Always).
+    if matches!(
+        tool.requires_approval(&serde_json::Value::Null),
+        ApprovalRequirement::Always
+    ) {
+        return Err(StatusCode::BAD_REQUEST);
```
serrrfirat: **Low Severity — UX: Bare 400 response for locked tools has no explanation**

The handler returns `Err(StatusCode::BAD_REQUEST)` with no JSON body. The frontend's `catch` handler shows a generic error. The user (or LLM tool caller) has no indication *why* the request was rejected.

**Suggested fix:** Return a JSON error body, e.g.:
```rust
return Err((StatusCode::BAD_REQUEST, Json(serde_json::json!({
    "error": "Tool is locked and cannot have its permission changed"
}))));
```
Or use the project's existing error response pattern.

src/tools/permissions.rs
```diff
@@ -0,0 +1,186 @@
...
+/// Static map of built-in tool names → their default `PermissionState`.
+///
+/// Tools present here default to `AlwaysAllow` (read-only / low-risk) or
+/// `AskEachTime` (write / destructive / network).  Tools **absent** from the
+/// map fall back to `AskEachTime` in `effective_permission`.
+pub static TOOL_RISK_DEFAULTS: LazyLock<HashMap<&'static str, PermissionState>> =
+    LazyLock::new(|| {
+        let mut m = HashMap::new();
+
+        // --- AlwaysAllow: read-only, informational, low-risk ---
```
serrrfirat: **Low Severity — Comment is misleading: `memory_write` listed under "read-only, informational, low-risk"**

`memory_write` modifies persistent workspace state. The section comment says "read-only, informational, low-risk" but this tool is clearly a write operation.

**Suggested fix:** Split the comment: "AlwaysAllow: read-only / informational" and "AlwaysAllow: low-risk writes" — or reconsider whether `memory_write` truly belongs in the AlwaysAllow tier.

src/channels/web/handlers/settings.rs
```diff
@@ -524,6 +524,160 @@ fn mask_settings_api_keys(settings: &mut std::collections::HashMap<String, serde
...
+        description: tool.description().to_string(),
+        name,
+        current_state: permission_state_to_str(new_state).to_string(),
+        default_state: permission_state_to_str(default).to_string(),
+        locked: false,
+        locked_reason: None,
+    }))
+}
+
+fn permission_state_to_str(state: crate::tools::permissions::PermissionState) -> &'static str {
```
serrrfirat, line 681: **Nit — Redundant serialization helpers duplicate serde logic**

`permission_state_to_str` / `str_to_permission_state` manually map the same strings that `#[serde(rename_all = "snake_case")]` on `PermissionState` already produces. Two serialization paths to keep in sync.

**Suggested fix:** Use serde directly at call sites, or wrap serde calls in these helpers to eliminate the manual match arms.

src/app.rs
```diff
@@ -932,6 +947,80 @@ impl AppBuilder {
     }
 }
 
+/// Seed tool permission defaults into the database for every registered tool
+/// that has no explicit user override yet.
+///
+/// This is called once at startup after the full tool registry is built.
+/// It is idempotent: existing entries in `tool_permissions.*` are never touched.
+async fn seed_tool_permissions(
```
serrrfirat, line 990: **Nit — Missing unit test for `seed_tool_permissions` idempotency**

The seeding function's key guarantee — "existing entries are never overwritten" — is untested. If someone changes the function to use upsert semantics, user overrides would be silently clobbered on every restart.

**Suggested fix:** Add a test that: (1) seeds defaults, (2) overrides one value (e.g., `echo` → `Disabled`), (3) re-seeds, (4) asserts the override is preserved.

serrrfirat:
## Paranoid Architect Review Summary

**PR:** feat(tools): persistent per-user tool permission system
**Reviewed:** All 18 changed files (1,770 additions), full file context read for every file.

### Verdict

The architecture is well-designed. The three-state model (`AlwaysAllow | AskEachTime | Disabled`), the `ApprovalRequirement::Always` hard floor that cannot be overridden, per-user DB scoping via `TenantScope`, and the defense-in-depth checks in both `process_approval` and `tool_permission_set` are solid. The test coverage is good across unit tests, handler tests, and E2E scenarios.

### Blocking Issues (recommend fixing before merge)

| # | Severity | Summary |
|---|----------|---------|
| 1 | **High** | `http` defaults to `AlwaysAllow` — enables SSRF/data exfiltration via POST without prompting |
| 2 | **High** | `create_job` defaults to `AlwaysAllow` — enables resource abuse without prompting |
| 3 | **Medium** | `auto_approved_tools.clear()` every iteration silently drops approvals if prior DB write failed |
| 4 | **Medium** | DB round-trip on every agentic loop iteration (10x per turn) — cache at iteration 0 instead |
| 5 | **Medium** | `event_emit` defaults to `AlwaysAllow` despite its own comment documenting escalation risk |

### Non-blocking Issues

| # | Severity | Summary |
|---|----------|---------|
| 6 | Low | Route shadowing: `GET /api/settings/tools` shadows `GET /api/settings/{key}` for key="tools" |
| 7 | Low | Tool names with dots in `format!("tool_permissions.{}", name)` break `from_db_map` parsing |
| 8 | Low | `PUT` for locked tools returns bare 400 with no JSON error body |
| 9 | Low | `memory_write` listed under "read-only, informational, low-risk" comment |
| 10 | Nit | Redundant `permission_state_to_str` / `str_to_permission_state` duplicate serde logic |
| 11 | Nit | `seed_tool_permissions` idempotency guarantee has no unit test |

### What's Good

- `ApprovalRequirement::Always` as an unbypassable hard floor is the right design
- `tool_permission_set` itself requiring `Always` approval prevents LLM from silently changing permissions
- Defense-in-depth in `process_approval` — checks locked status even though UI hides the button
- Per-user DB scoping via `TenantScope` prevents cross-user permission leakage
- Web UI with lock icons, 3-way toggles, and default badges is clean
- E2E test coverage for persistence, locked tools, and cross-session state

All 11 findings posted as inline comments on the relevant lines.

zmanian (changes requested):
## Review: feat(tools): persistent per-user tool permission system

Good feature design overall. The three-state model, hard floor enforcement via `ApprovalRequirement::Always`, and per-user DB scoping via `TenantScope` are all sound. The test coverage for the core permission logic and web handlers is solid.

However, there are security and correctness issues that need to be addressed before merge.

---

### Blocking Issues

**1. Security: `http` and `create_job` default to `AlwaysAllow` bypasses existing approval gates**
`src/tools/permissions.rs:55-58`

The `http` tool returns `ApprovalRequirement::UnlessAutoApproved` for POST requests with credentials (see `src/tools/builtin/http.rs:832-850`). By defaulting `http` to `AlwaysAllow`, the dispatcher adds it to `session.auto_approved_tools` on every loop iteration (`dispatcher.rs:345-348`), which means the `UnlessAutoApproved` check at `dispatcher.rs:667-669` will pass -- POST requests with injected credentials will execute without user approval.

Similarly, `create_job` returns `UnlessAutoApproved` and defaults to `AlwaysAllow`, so the LLM can spawn unlimited jobs without confirmation.

These should default to `AskEachTime`, or the dispatcher needs to consult the tool's `requires_approval()` with actual params before auto-approving (which is harder since params aren't known at filtering time). The simpler fix: move `http`, `create_job`, `event_emit`, `routine_create`, and `routine_update` to the `AskEachTime` tier. Tools that return `UnlessAutoApproved` or whose approval depends on parameters should never be in the `AlwaysAllow` tier.

**2. Bug: `clear()` + re-populate on every iteration causes silent permission loss on DB failure**
`src/agent/dispatcher.rs:356-366`

The dispatcher clears `session.auto_approved_tools` and re-populates from DB on every iteration. If the DB read fails (lines 326-332 fall back to empty HashMap), all previously auto-approved tools lose their approval for that iteration. Worse, if the user clicked "Always Approve" during the session but the DB write in `process_approval` failed (thread_ops.rs:1109-1117), the next iteration's `clear()` will silently remove the session-level approval.

Suggested fix: only clear + re-populate on the first iteration (like `selected_model` at line 415), or merge DB state into the existing set rather than replacing it.

**3. Performance: DB round-trip on every agentic loop iteration**
`src/agent/dispatcher.rs:324-332`

`get_all_settings()` + `Settings::from_db_map()` runs on every iteration. A typical agent loop runs 5-15 iterations per user message. This adds latency on the hot path. Compare with `selected_model` which is only loaded at `iteration == 0`.

Cache the permission map at the start of the loop (iteration 0) and reuse it for subsequent iterations.

---

### Non-Blocking Issues

**4. `event_emit` defaults to `AlwaysAllow` enables silent routine triggering**
`src/tools/permissions.rs:58`

`event_emit` can fire system_event routines that dispatch jobs with pre-authorized tools. It returns `UnlessAutoApproved` for this reason. Should be `AskEachTime`.

**5. Comment is misleading: "read-only, informational, low-risk" section includes `memory_write`**
`src/tools/permissions.rs:34`

`memory_write` modifies persistent workspace state. The section comment says "read-only, informational, low-risk". Split into two subsections or fix the comment.

**6. Tool names with dots in `from_db_map` parsing**
`src/agent/thread_ops.rs:1104`

The DB key is `format!("tool_permissions.{}", pending.tool_name)`. If `Settings::set()` splits on `.` to navigate the JSON tree, a tool named `foo.bar` would produce key `tool_permissions.foo.bar`, which would create a nested structure instead of a flat key. This could affect MCP tools with dotted names. Worth documenting this constraint or sanitizing tool names.

**7. Bare 400 for locked tools has no explanation**
`src/channels/web/handlers/settings.rs:617`

The PUT handler returns `Err(StatusCode::BAD_REQUEST)` with no JSON body when a locked tool is targeted. The frontend's catch handler shows a generic error. Return a JSON body explaining why.

**8. `tool_permission_set` silently succeeds without persistence when `settings_store` is `None`**

In `ToolPermissionSetTool::execute()`, when `settings_store` is `None`, the read-only path (no `state` param) returns success with `Settings::default()` data. This is fine. But actually -- looking more carefully, the write path correctly returns an error when `settings_store` is `None` (line 1383-1386). Good. However, the `ToolListTool` with no settings store silently returns empty permission overrides, which means the LLM sees default states only. This is acceptable behavior but worth a comment.

**9. Missing unit test for `seed_tool_permissions` idempotency**
`src/app.rs:955`

The seeding function's key guarantee -- "existing entries are never overwritten" -- is untested. If someone changes the function to use upsert semantics, user overrides would be silently clobbered on every restart.

**10. CI note**: This PR is from a fork. Only classify/scope CI jobs will run. Full CI (clippy, tests) has not been validated in the upstream environment.

---

### What looks good

- Hard floor enforcement is correct: `ApprovalRequirement::Always` at dispatcher.rs:671 is unconditional and cannot be bypassed by session auto-approve state.
- `tool_permission_set` correctly returns `ApprovalRequirement::Always` so the LLM cannot silently change permissions.
- Locked tool protection in both web handler (400) and LLM tool (error) paths.
- Per-user DB scoping via `TenantScope` prevents cross-user permission leakage.
- Good test coverage: permissions.rs unit tests, dispatcher filtering tests, web handler tests with mock tools, E2E scenarios.
- Default-deny for unknown tools (falls back to `AskEachTime`).

henrypark133:
## Addressed all review feedback from @serrrfirat

Thanks for the thorough security review — excellent catches on the tier defaults.

### HIGH — Security fixes
| Issue | Resolution |
|-------|-----------|
| `http` defaults to AlwaysAllow (SSRF/exfil via POST) | ✅ Moved to AskEachTime. GET already returns `Never` so safe tools unaffected |
| `create_job` defaults to AlwaysAllow (resource abuse) | ✅ Moved to AskEachTime |
| `event_emit` defaults to AlwaysAllow (silent routine triggering) | ✅ Moved to AskEachTime |

### MEDIUM — Resilience & performance
| Issue | Resolution |
|-------|-----------|
| `clear()` on every iteration loses approval if DB write fails | ✅ If `get_all_settings()` fails, returns early preserving existing session state |
| DB round-trip on every iteration | ✅ Cached at iteration 0 via `cached_tool_permissions` field on `ChatDelegate` (same pattern as `selected_model`) |

### LOW — UX, edge cases, architecture
| Issue | Resolution |
|-------|-----------|
| Bare 400 for locked tools | ✅ Returns JSON error body with explanation |
| Dot-separated tool names break path parsing | ✅ Validation added to `register()` and `register_sync()` rejecting names with `.` |
| Route shadowing `/api/settings/tools` vs `/api/settings/{key}` | ✅ Added comment documenting the intentional shadowing |
| `memory_write` listed under "read-only" comment | ✅ Comment updated to "informational, low-risk, or safe writes" |

### NITs
| Issue | Resolution |
|-------|-----------|
| Redundant serialization helpers | ✅ `str_to_permission_state` now uses `serde_json::from_value()` |
| Missing seed idempotency test | ✅ Added `seed_tool_permissions_preserves_user_overrides` test |

henrypark133:
## Addressing @zmanian's review

All blocking and non-blocking issues have been resolved in the latest push.

### Blocking issues

**1. Security: tools returning `UnlessAutoApproved` in AlwaysAllow tier**
✅ Fixed. Moved **all five** tools to AskEachTime:
- `http` — POST/credential requests now require approval
- `create_job` — job creation now requires approval  
- `event_emit` — event emission (can trigger routines) now requires approval
- `routine_create` — full-job routines now require approval (lightweight still `Never`)
- `routine_update` — routine modifications now require approval

Principle applied: **no tool that returns `UnlessAutoApproved` should be in the AlwaysAllow tier**, since AlwaysAllow + UnlessAutoApproved = auto-approved (bypassing the tool's own safety gate).

**2. Bug: `clear()` + re-populate causes silent permission loss on DB failure**
✅ Fixed. If `get_all_settings()` fails, `before_llm_call()` returns `None` early — preserving existing session auto-approvals instead of clearing them.

**3. Performance: DB round-trip on every iteration**
✅ Fixed. Permissions are cached at iteration 0 in a `cached_tool_permissions` field on `ChatDelegate`, reused for all subsequent iterations (same pattern as `selected_model`).

### Non-blocking issues

| # | Issue | Status |
|---|-------|--------|
| 4 | `event_emit` → AskEachTime | ✅ Done (part of #1) |
| 5 | Comment misleading ("read-only" includes `memory_write`) | ✅ Updated to "informational, low-risk, or safe writes" |
| 6 | Dot-separated tool names break path parsing | ✅ Validation added to `register()` and `register_sync()` |
| 7 | Bare 400 for locked tools | ✅ Returns JSON error body |
| 8 | `tool_permission_set` without settings store | ✅ Write path returns `ToolError::ExecutionFailed` |
| 9 | Missing seed idempotency test | ✅ Added `seed_tool_permissions_preserves_user_overrides` test |
| 10 | CI note (fork) | Noted — this is the upstream repo branch, not a fork |

zmanian (dismissed):
## Follow-up Review: feat(tools): persistent per-user tool permission system

All three blocking issues from my previous review have been addressed. Reviewing what changed:

### Previously Blocking -- Now Resolved

**1. Security: `http`, `create_job`, `event_emit`, `routine_create/update` tier placement**
Fixed. All five tools are now correctly in the `AskEachTime` tier in `TOOL_RISK_DEFAULTS` (`src/tools/permissions.rs`). Tools that return `UnlessAutoApproved` no longer get auto-approved by default.

**2. Bug: `clear()` + re-populate causing silent permission loss on DB failure**
Fixed. The dispatcher now returns early on DB failure (`dispatcher.rs:348-372`) without clearing session auto-approvals. The comment "keeping existing session state" correctly describes the behavior.

**3. Performance: DB round-trip on every loop iteration**
Fixed. `cached_tool_permissions: Mutex<Option<HashMap>>` on `ChatDelegate` ensures the DB is hit only once per conversation, with the result cached for subsequent iterations.

### Previously Non-Blocking -- Also Addressed

- **Dot validation in tool names**: Registry now rejects tool names containing `.` in both `register()` and `register_sync()` (`src/tools/registry.rs`). Prevents the `tool_permissions.foo.bar` nesting bug.
- **Bare 400 for locked tools**: PUT handler now returns a JSON body with a clear error message (`src/channels/web/handlers/settings.rs`).
- **`seed_tool_permissions` idempotency test**: Added as `seed_tool_permissions_preserves_user_overrides` with a real libsql backend (`src/app.rs`).

### Remaining Notes (Non-Blocking)

**1. Stale PR description table**
The PR description's tier table still lists `http`, `create_job`, `event_emit`, `routine_create/update` under "Always Allow". The code is correct (AskEachTime), but the description should be updated to avoid confusing future readers.

**2. Cache not invalidated on mid-conversation permission changes**
If a user changes permissions via the web UI while a conversation's agentic loop is active, the cached permissions in `ChatDelegate` won't reflect the change until the next conversation. This is acceptable behavior -- permission changes taking effect on next conversation is reasonable -- but worth documenting in the code comment near the cache.

**3. Unrelated ACP removal mixed into this PR**
`src/app.rs` removes the ACP agent validation block and `src/channels/web/server.rs` removes `"acp_agent"` from the extensions install handler. These are unrelated to the permission system. Not blocking, but ideally would be in a separate commit.

**4. `message` tool in AlwaysAllow tier**
`message` sends messages to users on configured channels. This seems intentional and reasonable since it's user-facing output, but worth confirming it can't be abused to spam external channels (e.g., Telegram) at high volume without rate limiting.

Overall the permission model is well-designed: default-deny for unknowns, hard floor enforcement for `ApprovalRequirement::Always`, per-user DB scoping, and good test coverage across unit, integration, and E2E layers.

src/channels/web/handlers/settings.rs
```diff
@@ -524,6 +524,173 @@ fn mask_settings_api_keys(settings: &mut std::collections::HashMap<String, serde
...
+
+    // Validate tool exists.
+    let tool = registry.get(&name).await.ok_or((
+        StatusCode::NOT_FOUND,
+        axum::Json(serde_json::json!({"error": format!("Tool '{}' not found", name)})),
+    ))?;
+
+    // Reject if tool is locked (ApprovalRequirement::Always).
+    if matches!(
+        tool.requires_approval(&serde_json::Value::Null),
```
serrrfirat, line 617: **Medium Severity**

This handler correctly rejects locked tools here, but the same `tool_permissions.<name>` key can still be written through the generic settings APIs (`PUT /api/settings/{key}` and `/api/settings/import`).

That means a crafted client can bypass this lock and persist `Disabled` or `AlwaysAllow` for a tool whose `requires_approval()` is `Always`. The dispatcher then applies `PermissionState::Disabled` before any hard-floor check, so a locked tool like `tool_permission_set` can be hidden entirely despite the PR claiming this floor is unoverrideable.

Please route all `tool_permissions.*` writes through the same validation, or ignore persisted overrides for tools whose approval requirement is `Always`.

src/app.rs
```diff
@@ -967,6 +957,80 @@ impl AppBuilder {
...
+                    tracing::warn!(
+                        "seed_tool_permissions: failed to serialize state for '{}': {}",
+                        name,
+                        e
+                    );
+                    continue;
+                }
+            };
+            if let Err(e) = db
+                .set_setting(owner_id, &format!("tool_permissions.{}", name), &json_value)
```
serrrfirat, line 1041: **Medium Severity**

Seeding concrete `tool_permissions.<name>` rows here makes the current default effectively permanent for existing users.

Once a row is written, future changes to `TOOL_RISK_DEFAULTS` no longer apply to that user, and the new `/api/settings/tools/:name` flow currently always persists an explicit state instead of offering an inherit/reset-to-default path. That means a later release cannot tighten a default for already-seeded users without a separate migration.

Please store only explicit overrides, or add a reset/delete path so users can fall back to the evolving defaults instead of pinning today's defaults forever.

src/agent/dispatcher.rs
```diff
@@ -344,6 +348,94 @@ impl<'a> LoopDelegate for ChatDelegate<'a> {
...
+        // because the existing approval gate already treats it as an unbypassable
+        // hard floor — even if a tool name is in session.auto_approved_tools, an
+        // ApprovalRequirement::Always tool still requires user confirmation.
+        let mut to_auto_approve: Vec<String> = Vec::new();
+        let tool_defs: Vec<_> = tool_defs
+            .into_iter()
+            .filter_map(|def| {
+                match effective_permission(&def.name, &tool_permissions) {
+                    PermissionState::Disabled => {
+                        tracing::debug!(tool = %def.name, "Excluding disabled tool from LLM context");
```
serrrfirat, line 416: **Medium Severity**

This branch removes disabled tools from `reason_ctx.available_tools`, but the turn still reuses `cached_prompt`, which was built earlier from the unfiltered `initial_tool_defs`.

`respond_with_tools()` prefers `context.system_prompt` over rebuilding from `available_tools`, so a tool marked `Disabled` can still be advertised in the prompt text for the current turn even though it was removed from the callable tool list. That breaks the new `Disabled = excluded from tool_definitions; LLM never sees it` guarantee.

Please rebuild the cached prompt after permission filtering, or stop caching the prompt when tool visibility is dynamic. A regression test should assert that a disabled tool disappears from both the tool schema and the prompt text.

zmanian (approved):
Re-reviewed after squash rebase. The ACP removal that was previously mixed in has been cleaned out -- diff is now focused solely on the permission system. Commit message is thorough and accurately describes the implementation.

Previous non-blocking notes (stale PR description table, mid-conversation cache invalidation, message tool rate-limiting) still apply as future follow-ups but nothing blocking here. LGTM.

PR #1914, opened 2026-04-02
fix:ignore default model override and empty WASM polls

### LINKED ISSUE #1811: Gateway sends model: "default" to Anthropic API → 404 storm on empty Telegram polls
Summary
IronClaw v0.23.0 sends model: "default" (the literal string) to the Anthropic Messages API on certain internal LLM calls, producing a continuous stream of 404 errors with automatic retries. Over ~7 hours this generated 330 failed API calls (110 cycles × 3 retries each).

Environment
IronClaw v0.23.0 (cargo install)
macOS, M5 Max, launchd-managed (com.ocean.ironclaw)
selected_model = "claude-opus-4-6" in config.toml
llm_backend = "anthropic" (direct API, not Bedrock)
Telegram WASM channel enabled, gateway channel enabled
heartbeat.enabled = false
Reproduction
Configure Anthropic direct backend with selected_model = "claude-opus-4-6"
Enable Telegram WASM channel
Have BOOTSTRAP.md present in workspace (even if profile_onboarding_completed = true)
Leave agent running with no incoming messages
Observed behavior
Every ~30 seconds, coinciding with the Telegram getUpdates long-poll cycle returning an empty result (body_len=23), an LLM request fires with the literal model name "default":

CopyWARN Retrying after transient error provider=claude-opus-4-6 attempt=1 max_retries=3
  error=Provider claude-opus-4-6 request failed: HttpError: Invalid status code 404 Not Found
  with message: {"type":"error","error":{"type":"not_found_error","message":"model: default"}}
This repeats 3x per cycle (retry policy), then fires again on the next poll. Hourly counts observed:

Hour (UTC)	Phantom 404 calls
22:00	6
23:00	12
00:00	81
01:00	120
02:00	51
03:00	45
04:00	15
05:00+	0 (after BOOTSTRAP.md cleared)

Expected behavior
Empty Telegram polls should not trigger LLM calls at all
If an internal LLM call is needed, req.model should be None (falling through to active_model_name()) or the configured model name — never the literal string "default"

Root cause analysis
The Anthropic provider at src/llm/anthropic_oauth.rs:244 does:

Copylet model = req.model.take().unwrap_or_else(|| self.active_model_name());
This correctly falls back to the configured model when req.model is None. But the upstream caller is setting req.model = Some("default") — the literal string — so the fallback never triggers.

Multiple code paths use "default" as a session/user/owner ID:

src/agent/agent_loop.rs:830 — get_or_create_session("default")
src/agent/dispatcher.rs:1314 — owner_id: "default".to_string()
src/llm/session.rs:75 — user_id: RwLock::new("default".to_string())
One of these "default" strings appears to be leaking into the model field of the LLM request, likely through the gateway event classification path that fires on each Telegram poll return.

Workaround
Clearing BOOTSTRAP.md from workspace memory stopped the 404 storm entirely (even though profile_onboarding_completed = true was already set and the log showed WARN BOOTSTRAP.md still exists but profile_onboarding_completed is set; suppressing bootstrap injection). The suppression prevented content injection but not the code path that constructs the phantom LLM request.

Impact
Rate limit waste: 330 failed calls consumed rate limit quota
Latency: Retries block the event loop, degrading response time for real user messages
Potential cost risk: If the model name resolved correctly, this would produce ~2,880 Opus calls/day ($800+/day) for processing empty poll results

serrrfirat:
## Summary
- drop empty emitted WASM messages with no attachments before they enter the agent loop
- treat blank and "default" selected-model overrides as no override throughout dispatcher, config, and provider resolution
- normalize provider call sites and add regression tests for both paths

## Testing
- cargo test --lib llm::provider::model_override_tests::
- cargo test --lib config::llm::tests::selected_model_override_ignores_default_sentinel -- --exact
- cargo test --lib config::llm::tests::nearai_resolve_ignores_default_selected_model -- --exact
- cargo test --lib agent::dispatcher::tests::selected_model_override_ignores_default_sentinel -- --exact
- cargo test --lib channels::wasm::wrapper::tests::test_dispatch_emitted_messages_skips_empty_messages_without_attachments -- --exact
- cargo fmt --all

Fixes #1811

### DIFF
--- a/src/agent/dispatcher.rs
+++ b/src/agent/dispatcher.rs
@@ -22,6 +22,10 @@ use crate::agent::agentic_loop::{
 use crate::llm::{ChatMessage, Reasoning, ReasoningContext};
 use crate::tools::redact_params;
 
+fn selected_model_override(value: &serde_json::Value) -> Option<String> {
+    crate::llm::normalized_model_override(value.as_str()).map(str::to_string)
+}
+
 /// Result of the agentic loop execution.
 pub(super) enum AgenticLoopResult {
     /// Completed with a response.
@@ -362,12 +366,9 @@ impl<'a> LoopDelegate for ChatDelegate<'a> {
         if iteration == 0
             && let Some(store) = self.tenant.store()
             && let Ok(Some(value)) = store.get_setting("selected_model").await
-            && let Some(model) = value.as_str()
+            && let Some(model) = selected_model_override(&value)
         {
-            let model = model.trim();
-            if !model.is_empty() {
-                reason_ctx.model_override = Some(model.to_string());
-            }
+            reason_ctx.model_override = Some(model);
         }
 
         let output = match reasoning.respond_with_tools(reason_ctx).await {
@@ -1306,7 +1307,7 @@ mod tests {
     use crate::safety::SafetyLayer;
     use crate::tools::ToolRegistry;
 
-    use super::check_auth_required;
+    use super::{check_auth_required, selected_model_override};
 
     /// Minimal LLM provider for unit tests that always returns a static response.
     struct StaticLlmProvider;
@@ -2168,6 +2169,20 @@ mod tests {
                 "ceiling logic inconsistent for max_iter={max_iter}"
             );
         }
+    }
+
+    #[test]
+    fn selected_model_override_ignores_default_sentinel() {
+        assert_eq!(selected_model_override(&serde_json::json!("default")), None);
+        assert_eq!(
+            selected_model_override(&serde_json::json!("  DEFAULT  ")),
+            None
+        );
+        assert_eq!(selected_model_override(&serde_json::json!("  ")), None);
+        assert_eq!(
+            selected_model_override(&serde_json::json!("claude-opus-4-6")).as_deref(),
+            Some("claude-opus-4-6")
+        );
     }
 
     /// LLM provider that always returns calls to a nonexistent tool, regardless
--- a/src/channels/wasm/wrapper.rs
+++ b/src/channels/wasm/wrapper.rs
@@ -2812,6 +2812,15 @@ impl WasmChannel {
                 .await;
             }
 
+            if emitted.content.trim().is_empty() && emitted.attachments.is_empty() {
+                tracing::debug!(
+                    channel = %dispatch.channel_name,
+                    user_id = %emitted.user_id,
+                    "Skipping empty emitted message"
+                );
+                continue;
+            }
+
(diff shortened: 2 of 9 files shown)


zmanian (approved):
## Review: fix: ignore default model override and empty WASM polls

Overall this is a clean, well-structured PR. The centralized `normalized_model_override()` function is the right approach, and test coverage is thorough. Approving with two non-blocking observations.

### Model override normalization -- looks good

- `normalized_model_override()` in `src/llm/provider.rs` is a clean, well-placed normalization point. Case-insensitive "default" handling and whitespace trimming are correct.
- All provider call sites (`anthropic_oauth`, `nearai_chat`, `github_copilot`, `rig_adapter`) are migrated to `take_model_override()`. Coverage is complete -- `bedrock`, `codex_chatgpt`, and `openai_codex_provider` don't access `req.model` directly.
- `effective_model_name()` on the trait also routes through the normalizer, closing the loop for `response_cache.rs` which calls it.
- `src/config/llm.rs` correctly uses `Self::selected_model_override(settings)` at all six config resolution sites (nearai, bedrock, openai_codex, openai_compatible, tinfoil/registry, fallback).
- The dispatcher path (`src/agent/dispatcher.rs`) normalizes early via `selected_model_override()`, which means `reasoning.rs` lines 701/809 (`request.model = Some(model.clone())`) receive an already-normalized value. Consistent.

### Non-blocking: empty message check placement in WASM dispatch

`src/channels/wasm/wrapper.rs` -- The empty-content/no-attachments check is placed **after** the rate limiter has already consumed a slot. Moving it before the rate limit check (right at the top of the `for emitted in messages` loop) would avoid wasting rate limit budget on messages that will be discarded anyway. Not a correctness issue, just an efficiency nit.

### Non-blocking: `model` field remains `pub`

Both `CompletionRequest::model` and `ToolCompletionRequest::model` are still `pub`. New code (or future contributors) could bypass `take_model_override()` and access the field directly, re-introducing the bug. Not in scope for this PR, but worth tracking -- making the field `pub(crate)` or private with a setter that normalizes would close this gap permanently.

### Test coverage

Good regression tests for both fixes:
- `selected_model_override_ignores_default_sentinel` in dispatcher and config
- `nearai_resolve_ignores_default_selected_model` for integration-level config resolution
- `model_override_tests` module in provider.rs covering `effective_model_name`, `take_model_override` for both request types
- `test_dispatch_emitted_messages_skips_empty_messages_without_attachments` for WASM path

Issue #1609, opened 2026-03-24
feat: admin management panel — web UI for users, workspaces, roles, and usage monitoring


ilblackdragon:
## Problem / Motivation

Once Issues #1 (User Management), #2 (Workspaces), and #3 (RBAC) land, all management operations are API-only. There's no UI for:

- **User administration** — creating users, managing invitations, suspending accounts, viewing activity
- **Workspace management** — creating workspaces, managing members, assigning roles
- **Token management** — generating/revoking API tokens
- **Usage monitoring** — seeing per-user/per-workspace job counts, token consumption, active sessions, rate limit status

The existing web UI (`src/channels/web/static/`) is a chat-focused SPA. Admin operations need a separate set of views.

Depends on **#1605 (User Management)**, **#1607 (Workspaces)**, and **#1608 (RBAC)**.

### Approach: Extend Existing Embedded SPA

The current web frontend is embedded via `include_str!` / `include_bytes!` in `src/channels/web/handlers/static_files.rs` (or inline in `server.rs`). The admin panel should follow the same pattern — static HTML/CSS/JS served from embedded assets, calling the JSON API endpoints from Issues #1-3.

**Do NOT introduce a frontend build system** (no webpack, no npm). The existing UI uses vanilla JS with CDN-loaded libraries. Follow that pattern.

### Route Structure

```
GET /admin                      — admin dashboard (redirect to /admin/users)
GET /admin/users                — user list + invite form
GET /admin/users/{id}           — user detail (profile, tokens, activity)
GET /admin/workspaces           — workspace list + create form
GET /admin/workspaces/{slug}    — workspace detail (members, roles, settings)
GET /admin/invitations          — pending invitations list
GET /admin/usage                — usage dashboard (aggregated stats)
```

All `/admin/*` routes require authentication. Pages that modify data (suspend user, change role) should require `superadmin` or workspace `admin`/`owner` per the RBAC matrix from Issue #3. The frontend should call the API endpoints and handle 403 responses gracefully (hide buttons, show "insufficient permissions").

#### Users Page (`/admin/users`)
- Table: display_name, email, status, created_at, last_login_at, workspace count
- Actions: invite new user (email + optional workspace), suspend/activate toggle
- Search/filter by name, email, status
- Calls: `GET /api/admin/users`, `POST /api/invitations`, `POST /api/admin/users/{id}/suspend`

#### User Detail (`/admin/users/{id}`)
- Profile section: display_name, email, status, metadata, created_at
[...]

Issue #1608, opened 2026-03-24
feat: role-based access control (RBAC) with per-workspace permission matrix


ilblackdragon:
## Problem / Motivation

All authenticated users currently have identical permissions. The `auth_middleware` in `src/channels/web/auth.rs` checks token validity but nothing else — any valid token can call any endpoint. The `workspace_members.role` column from Issue #2 stores roles but nothing enforces them.

This means:
- **Any user can create/delete other users** (once Issue #1 lands)
- **Any workspace member can remove other members** (once Issue #2 lands)
- **No read-only access** — viewers can mutate data
- **No admin/owner distinction** — everyone can do everything
- **No way to grant limited API access** (e.g., a CI bot that can only trigger routines)

Depends on **#1605 (User Management)** and **#1607 (Workspaces)**.

### Permission Model

Four roles in ascending privilege order:

| Role | Description |
|------|-------------|
| `viewer` | Read-only access to workspace data (jobs, routines, memory, settings) |
| `member` | Full read/write within the workspace (create jobs, write memory, manage own routines) |
| `admin` | Member + manage workspace members, update workspace settings |
| `owner` | Admin + delete workspace, transfer ownership, manage admins |

Additionally, a **system-level `superadmin`** flag on `users` for cross-workspace operations (user suspension, global settings).

### Database Changes

```sql
ALTER TABLE users ADD COLUMN is_superadmin BOOLEAN NOT NULL DEFAULT FALSE;
```

No other schema changes needed — `workspace_members.role` already exists from Issue #2.

### Permission Matrix

| Operation | viewer | member | admin | owner | superadmin |
|-----------|--------|--------|-------|-------|------------|
| View workspace data (jobs, routines, memory) | ✓ | ✓ | ✓ | ✓ | ✓ |
| Search memory | ✓ | ✓ | ✓ | ✓ | ✓ |
| Send chat messages | ✗ | ✓ | ✓ | ✓ | ✓ |
| Create/cancel jobs | ✗ | ✓ | ✓ | ✓ | ✓ |
| Write memory | ✗ | ✓ | ✓ | ✓ | ✓ |
| Manage own routines | ✗ | ✓ | ✓ | ✓ | ✓ |
| Manage workspace settings | ✗ | ✗ | ✓ | ✓ | ✓ |
| Invite/remove members | ✗ | ✗ | ✓ | ✓ | ✓ |
| Change member roles (up to admin) | ✗ | ✗ | ✓ | ✓ | ✓ |
[...]

PR #1916, opened 2026-04-02
feat: admin management panel — web UI for users and usage monitoring

serrrfirat:
## Summary

Adds a standalone admin management panel served from `/admin/*` routes, addressing #1609. The panel is a separate embedded SPA (vanilla HTML/CSS/JS, no build system) that provides:

- **Dashboard** — System-wide metrics: user counts, total jobs, 30-day LLM usage/cost, server uptime, and a recent users table
- **Users page** — Full user CRUD: list with search/filter, create user with one-time token, suspend/activate, role management (promote/demote), token generation, delete with confirmation
- **User detail page** — Profile card, role management dropdown, 30-day usage breakdown by model
- **Usage page** — Period selector (24h/7d/30d), per-user and per-model breakdowns with proportional cost bars
- **Workspaces & Invitations** — Stubbed as "Coming soon" (see caveats below)

### Backend changes (`server.rs` only)
- 3 static file handlers serving embedded `admin.html`, `admin.css`, `admin.js` via `include_str!`
- 5 routes in the `statics` router: `/admin`, `/admin/`, `/admin/{*path}` (SPA catch-all), `/admin.css`, `/admin.js`
- 1 new API endpoint: `GET /api/admin/usage/summary` — composes existing `list_users`, `user_summary_stats`, and `user_usage_stats` DB methods (no new DB trait methods, no migrations)

### Frontend architecture
- Standalone SPA independent of the main chat UI
- Hash-based client-side routing (`#/users`, `#/users/:id`, `#/usage`, etc.)
- Shared auth token with main SPA via `sessionStorage` (no re-login when navigating between them)
- OIDC proxy auth support
- Admin role verification via `/api/profile` — non-admins see an "access denied" screen
- Full dark/light theme support (inherits from `theme-init.js`)
- Responsive layout: sidebar collapses on mobile, tables scroll horizontally
- All mutation controls (suspend, delete, role change) enforced server-side via `AdminUser` extractor

### No changes to
- Database trait or implementations (postgres/libsql)
- Migrations
- Existing API endpoints
- Main SPA (`index.html`, `app.js`, `style.css`)

## Caveats

- **Workspaces page** is stubbed — depends on #1607 (Workspace Management) which hasn't landed. The sidebar link shows "soon" badge and the page displays a "Coming soon" placeholder.
- **Invitations page** is stubbed — depends on #1608 (RBAC) which hasn't landed. Same treatment as workspaces.
- **No i18n** — The admin panel is English-only for now. The main SPA's i18n system could be extended later.
- **No audit log UI** — Out of scope per the issue spec.
- **Binary size** — Adds ~60KB of embedded assets (HTML+CSS+JS). No external libraries.

## Test plan

- [x] `cargo fmt --check` passes
- [x] `cargo check` compiles with default features
- [x] `GET /admin` returns 200 with admin HTML
- [x] `GET /admin/anything` returns 200 (SPA catch-all)
- [x] Auth: enter valid admin token → dashboard loads
- [x] Auth: enter member token → "Admin privileges required" screen
- [x] Auth: shared token from main SPA works without re-login
- [x] Dashboard: metrics cards render with correct data
- [x] Users: list, search, filter by status/role all work
- [ ] Users: create user → token banner displayed
- [ ] Users: suspend/activate/role change/delete work
- [ ] User detail: profile, role management, usage table render
- [x] Usage: period selector (24h/7d/30d) re-fetches data
- [x] Responsive: sidebar collapses at 768px, tables scroll
- [x] Dark/light theme inherited correctly

### DIFF
--- a/src/channels/web/handlers/users.rs
+++ b/src/channels/web/handlers/users.rs
@@ -532,3 +532,65 @@ pub async fn usage_stats_handler(
         "usage": entries,
     })))
 }
+
+/// System-wide usage summary for the admin dashboard.
+pub async fn usage_summary_handler(
+    State(state): State<Arc<GatewayState>>,
+    AdminUser(_user): AdminUser,
+) -> Result<Json<serde_json::Value>, (StatusCode, String)> {
+    let store = state.store.as_ref().ok_or((
+        StatusCode::SERVICE_UNAVAILABLE,
+        "Database not available".to_string(),
+    ))?;
+
+    let users = store
+        .list_users(None)
+        .await
+        .map_err(|e| (StatusCode::INTERNAL_SERVER_ERROR, e.to_string()))?;
+
+    let total = users.len();
+    let active = users.iter().filter(|u| u.status == "active").count();
+    let suspended = users.iter().filter(|u| u.status == "suspended").count();
+    let admins = users.iter().filter(|u| u.role == "admin").count();
+
+    let summary_stats = store
+        .user_summary_stats(None)
+        .await
+        .map_err(|e| (StatusCode::INTERNAL_SERVER_ERROR, e.to_string()))?;
+
+    let total_jobs: i64 = summary_stats.iter().map(|s| s.job_count).sum();
+    let total_cost: rust_decimal::Decimal = summary_stats.iter().map(|s| s.total_cost).sum();
+
+    let since_30d = chrono::Utc::now() - chrono::Duration::days(30);
+    let usage_stats = store
+        .user_usage_stats(None, since_30d)
+        .await
+        .map_err(|e| (StatusCode::INTERNAL_SERVER_ERROR, e.to_string()))?;
+
+    let llm_calls: i64 = usage_stats.iter().map(|s| s.call_count).sum();
+    let input_tokens: i64 = usage_stats.iter().map(|s| s.input_tokens).sum();
+    let output_tokens: i64 = usage_stats.iter().map(|s| s.output_tokens).sum();
+    let usage_cost: rust_decimal::Decimal = usage_stats.iter().map(|s| s.total_cost).sum();
+
+    let uptime_seconds = state.startup_time.elapsed().as_secs();
+
+    Ok(Json(serde_json::json!({
+        "users": {
+            "total": total,
+            "active": active,
+            "suspended": suspended,
+            "admins": admins,
+        },
+        "jobs": {
+            "total": total_jobs,
+            "total_cost": total_cost.to_string(),
+        },
+        "usage_30d": {
+            "llm_calls": llm_calls,
+            "input_tokens": input_tokens,
+            "output_tokens": output_tokens,
+            "total_cost": usage_cost.to_string(),
+        },
+        "uptime_seconds": uptime_seconds,
+    })))
+}
--- a/src/channels/web/server.rs
+++ b/src/channels/web/server.rs
@@ -31,7 +31,7 @@ use crate::bootstrap::ironclaw_base_dir;
 use crate::channels::IncomingMessage;
 use crate::channels::relay::DEFAULT_RELAY_NAME;
 use crate::channels::web::auth::{
-    AdminUser, AuthenticatedUser, CombinedAuthState, UserIdentity, auth_middleware,
+    AuthenticatedUser, CombinedAuthState, UserIdentity, auth_middleware,
 };
 use crate::channels::web::handlers::jobs::{
     job_files_list_handler, job_files_read_handler, jobs_cancel_handler, jobs_detail_handler,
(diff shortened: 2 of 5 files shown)


ilblackdragon (commented):
## Code Review

Overall this is a solid, well-scoped PR — minimal backend changes, consistent patterns, good XSS hygiene with `escapeHtml()` throughout. A few items to address:

### Bug

**OIDC auto-auth race condition** (`admin.js`, `autoAuth()` function):

```js
fetch('/api/gateway/status').then(function (res) {
  if (res.ok) {
    oidcProxyAuth = true;
    // ... async chain ...
  }
  showAuth(); // <-- runs even when res.ok is true
});
```

When `res.ok` is true, the code enters the OIDC path **and** falls through to `showAuth()` because the `.then()` inside the `if` block is async. This causes the auth screen to flash briefly before the OIDC path resolves. Fix: add `return` after the OIDC chain, or use `else { showAuth(); }`.

### Performance

**`usage_summary_handler` makes 3 sequential DB calls and loads all users into memory:**

```rust
let users = store.list_users(None).await?;
let summary_stats = store.user_summary_stats(None).await?;
let usage_stats = store.user_usage_stats(None, since_30d).await?;
```

These are independent queries — wrap them in `tokio::join!` for an easy latency win. Longer term, the user counts (total/active/suspended/admins) could be a single `SELECT COUNT(*) ... GROUP BY status, role` instead of loading every user record.

### Code Quality

**`detailRow()` has inconsistent escaping convention** — `label` is auto-escaped but `value` is raw HTML. Callers must know when to pre-escape:

```js
detailRow('Email', escapeHtml(user.email || 'Not set'))  // caller escapes
detailRow('Role', roleBadge(user.role))                   // caller passes HTML
```

Consider splitting into `detailRow` (text) and `detailRowHtml` (HTML), or documenting the convention with a comment.

**CSS variable duplication** — The full theme variable block (~60 vars, dark + light) is copied from `style.css` with a "keep in sync" comment. Since `admin.html` already loads `theme-init.js`, consider extracting the shared `:root` variables into a common CSS file to avoid inevitable drift.

### Docs

The new `GET /api/admin/usage/summary` endpoint should be added to the API routes table in `src/channels/web/CLAUDE.md`.

ilblackdragon (changes requested):
Findings first:

1. `DELETE /api/admin/users/{id}` does not invalidate the deleted user's cached DB auth entry, while suspend/activate/role-change do. Because `DbAuthenticator` caches successful auth for 60s, a deleted user can continue authenticating briefly after deletion. This needs the same invalidation treatment as the other security-sensitive state changes.

2. The OIDC/manual-token auth path in the admin SPA is brittle. `autoAuth()` sets `oidcProxyAuth = true` after a successful `/api/gateway/status`, but `logout()` never clears it, and `apiFetch()` suppresses the `Authorization` header whenever that flag is set. That means after hitting the access-denied path under proxy auth, entering a real admin token may still continue using cookie-based auth instead of the typed token. The “Try Different Token” flow is not reliable in hosted/OIDC setups.

3. `renderUserDetail()` expects `job_count`, `total_cost`, and `last_active_at`, but `users_detail_handler` does not return those fields. The page will silently render incorrect summary data (`0`, `$0.00`, `Never`) for real users. The list endpoint computes these fields; the detail endpoint should either expose them too or the UI should stop assuming them.

4. The create-user flow shows the one-time token and then immediately re-renders the users page from a fresh API fetch, which recreates an empty hidden `#user-token-banner`. Since the plaintext token is only returned once, this can effectively discard the token before the admin has a chance to copy it.

Architecture / maintainability:

5. The admin panel is already a meaningful subsystem, but it is implemented as one large `admin.js` file that mixes auth bootstrapping, API client behavior, hash routing, page rendering, modal state, and all mutations. This is workable for a spike, but expensive to extend safely once workspaces/invitations/audit pages land. I’d strongly recommend splitting this into at least a small API layer plus per-page modules/helpers, even if you keep the no-build-system approach.

6. The frontend/backend contract is too implicit. The drift between `renderUserDetail()` and `users_detail_handler` is a concrete example of why ad hoc JSON shapes are hard to maintain. For this surface I’d prefer stable Rust-owned response structs per endpoint and tests that lock the API shape.

7. `admin.css` copies theme variables from the main app with a note to “keep in sync.” That creates a second design system to maintain. If the admin panel is intended to stick around, it should consume shared theme tokens/base styles rather than fork them.

8. `server.rs` is accumulating admin-specific static serving and SPA routing directly in the main web entrypoint. This is still manageable at current size, but if `/admin` is a real product surface, I’d move the admin asset handlers and route assembly behind a dedicated web/admin module before more pages are added.

The direction is good: server-side admin enforcement is in the right place, and avoiding new DB traits/migrations keeps scope disciplined. But the current implementation still reads more like a fast prototype than a maintainable admin subsystem. I’d like to see the correctness issues fixed before merge, and I’d strongly encourage a small structural cleanup while the surface is still contained.

serrrfirat:
Superseded by #1963, recreated cleanly on top of current staging.

PR #1918, opened 2026-04-02
fix(docker): ensure ironclaw runtime home exists

### LINKED ISSUE #1899: [CRITICAL] Missing home directory for ironclaw user will cause runtime failures. `adduser -
## [CRITICAL:100] Issue Found by Staging CI Review

**Severity:** CRITICAL
**Confidence:** 100/100
**PR comment:** https://github.com/nearai/ironclaw/pull/1896#issuecomment-4174315897

### Description
Missing home directory for ironclaw user will cause runtime failures. `adduser -D` creates system user without home directory, but app requires `~/.ironclaw/` (see CLAUDE.md). When app tries to access `$HOME/.ironclaw/`, it will fail.

---

serrrfirat:
## Summary
- set an explicit `HOME=/home/ironclaw` in the production runtime image
- pre-create `/home/ironclaw/.ironclaw` and switch `WORKDIR` to the ironclaw home before dropping privileges
- add a regression test that locks in the runtime Dockerfile home-directory invariants

## Testing
- `rustfmt --check tests/dockerfile_runtime_home.rs`
- `rustc --test tests/dockerfile_runtime_home.rs -o /tmp/dockerfile_runtime_home_test && /tmp/dockerfile_runtime_home_test`

## Notes
- `cargo test --test dockerfile_runtime_home` was attempted, but this machine ran out of disk during workspace compilation before the test binary was built.
- Local Docker daemon was unavailable, so I could not run a container smoke test here.

Fixes #1899


zmanian (approved):
LGTM. Clean, minimal fix for a real runtime issue.

**Dockerfile changes**: Correct. Setting `HOME` explicitly, pre-creating `~/.ironclaw`, and `chown`-ing before `USER ironclaw` is the right approach. The `useradd -m -d` flags are consistent. Idempotent -- safe to rebuild.

**Gemini's migration concern is a false positive**: migrations use `embed_migrations!("migrations")` (refinery macro) which embeds SQL at compile time into the binary. The `WORKDIR` change has no effect on migration discovery at runtime. `/app/migrations` is only needed during the build stage.

**Test** (`tests/dockerfile_runtime_home.rs`): Reasonable regression guard for a Dockerfile invariant. The string-matching approach is brittle by nature, but for this purpose (catching accidental removal of the home dir setup) it is adequate. Gemini's suggestion to use "token-based matching" is over-engineering for a Dockerfile text check.

**Snapshot diff**: Whitespace-only change (trailing space removed). No functional impact.

One minor observation (non-blocking): the PR drops `-s /bin/bash` from the `useradd` call that exists on staging. The default shell will be `/bin/sh` or whatever is in `/etc/default/useradd`. This is fine for a container runtime (the ironclaw process doesn't need an interactive shell), but worth noting in case anyone expects bash inside the container for debugging.

PR #1939, opened 2026-04-03
feat: token-level streaming for Responses API

pranavraja99:
## Summary

Adds real token-by-token streaming through the full LLM stack so that `response.output_text.delta` SSE events arrive incrementally instead of as a single chunk after the agent finishes.

### Changes

- **`LlmProvider` trait** (`src/llm/provider.rs`): New `complete_streaming(request, on_token)` method with default fallback to non-streaming `complete()`
- **`NearAiChatProvider`** (`src/llm/nearai_chat.rs`): Implements real SSE streaming — sets `stream=true`, parses OpenAI-format SSE deltas, fires `on_token` callback per chunk. Handles both `content` and `reasoning_content` for thinking models.
- **`Reasoning`** (`src/llm/reasoning.rs`): `respond_streaming()` uses `complete_streaming` for text-only responses. Falls back to `respond_with_tools` when tools are present (tool JSON must be complete before execution).
- **`ChatDelegate`** (`src/agent/dispatcher.rs`): Creates `on_token` callback that emits `StatusUpdate::StreamChunk` via the channel manager. Each token spawned as fire-and-forget async task.

### What streams and what doesn't

| Scenario | Streams? |
|----------|----------|
| Text-only response (no tools) | Yes — tokens arrive incrementally |
| Final text after tool calls complete | Yes — last LLM call streams |
| Tool call JSON parameters | No — must be complete for execution |
| Tool start/complete events | Already streamed (separate SSE events) |

### No web layer changes needed

The existing pipeline already handles it end-to-end:
`StreamChunk` → `AppEvent::StreamChunk` → SSE broadcast → `response.output_text.delta`

## Test plan
- [x] `cargo check` — compiles clean, no warnings
- [ ] Deploy to Railway, verify `curl -N ... stream:true` shows incremental deltas
- [ ] Run `integrations/abound/tests/test_abound_e2e.py` — streaming test passes

### DIFF
--- a/integrations/abound/tests/test_abound_e2e.py
+++ b/integrations/abound/tests/test_abound_e2e.py
@@ -373,6 +373,42 @@ except Exception as e:
 print()
 
 # -----------------------------------------------------------
+# 7. Streaming: multiple deltas (not one big chunk)
+# -----------------------------------------------------------
+print("--- 7. Streaming: verify incremental deltas ---")
+try:
+    stream = client.responses.create(
+        model="default",
+        input="Write a short paragraph about exchange rates.",
+        stream=True,
+    )
+    delta_count = 0
+    delta_sizes = []
+    full_text = ""
+    for event in stream:
+        if event.type == "response.output_text.delta":
+            delta_count += 1
+            delta_sizes.append(len(event.delta))
+            full_text += event.delta
+
+    check("has multiple deltas", delta_count > 1,
+          f"got {delta_count} delta(s) — expected >1 for token-level streaming")
+    check("has text content", len(full_text) > 0, f"text length: {len(full_text)}")
+
+    if delta_count > 1:
+        avg_size = sum(delta_sizes) / len(delta_sizes)
+        check("avg delta smaller than full text", avg_size < len(full_text),
+              f"avg delta: {avg_size:.0f} chars, full text: {len(full_text)} chars")
+
+    print(f"  Deltas: {delta_count}")
+    print(f"  Delta sizes: {delta_sizes[:10]}{'...' if len(delta_sizes) > 10 else ''}")
+    print(f"  Full text: {len(full_text)} chars")
+    print(f"  Text: {full_text[:200]}")
+except Exception as e:
+    check("streaming succeeded", False, str(e))
+print()
+
+# -----------------------------------------------------------
 # Summary
 # -----------------------------------------------------------
 print(f"=== Results: {passed} passed, {failed} failed ===")
--- a/src/agent/dispatcher.rs
+++ b/src/agent/dispatcher.rs
@@ -412,6 +412,7 @@ impl<'a> LoopDelegate for ChatDelegate<'a> {
         let metadata = self.message.metadata.clone();
         let channel = self.message.channel.clone();
         let on_token = move |token: String| {
+            tracing::debug!(len = token.len(), "on_token callback fired");
             let channels = Arc::clone(&channels);
             let channel = channel.clone();
             let metadata = metadata.clone();
--- a/src/agent/dispatcher.rs
+++ b/src/agent/dispatcher.rs
@@ -405,7 +405,24 @@ impl<'a> LoopDelegate for ChatDelegate<'a> {
             }
         }
 
-        let output = match reasoning.respond_with_tools(reason_ctx).await {
+        // Emit token-level StreamChunk events when tools are empty (final
+        // text response). The callback fires for each token from the LLM
+        // and broadcasts it via the web channel's SSE stream.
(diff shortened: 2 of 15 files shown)


serrrfirat (changes requested):
I found several correctness regressions in the new streaming path.

1. High: streaming requests bypass the existing wrapper behavior instead of preserving it.
Files: `src/llm/retry.rs`, `src/llm/circuit_breaker.rs`, `src/llm/token_refreshing.rs`, `src/llm/smart_routing.rs`
The new `complete_streaming` / `complete_with_tools_streaming` methods delegate directly to `inner`, so `stream=true` no longer gets retries, circuit-breaker gating/accounting, Codex token refresh/retry, or smart-routing classification/cascade. This changes production behavior for the same request shape and makes streaming less reliable than non-streaming.
Suggested fix: implement the same wrapper semantics for streaming that `complete` / `complete_with_tools` already apply, or explicitly route streaming through the same internal helper paths.

2. High: engine v2 streaming is routed through thread-local context, which is unsafe in Tokio async execution and can misroute or drop chunks.
Files: `src/bridge/router.rs`, `src/bridge/llm_adapter.rs`, `crates/ironclaw_engine/src/traits/llm.rs`
`CURRENT_STREAM_CTX` is stored in `thread_local!` before the async engine call and then read later inside the token callback. Tokio futures can resume on a different worker thread after `await`, so the callback can see `None` or another request's context. Also, the bridge now streams for every call depth, even though `depth > 0` represents internal sub-calls, so nested engine calls can leak into the user-visible stream.
Suggested fix: carry stream context explicitly in the request/callback path rather than thread-local storage, and gate user-visible streaming to root calls unless sub-call streaming is explicitly intended.

3. High: chunk ordering is nondeterministic because each token is emitted via detached spawned tasks.
Files: `src/agent/dispatcher.rs`, `src/bridge/router.rs`, `src/channels/web/responses_api.rs`
Both v1 and v2 paths `tokio::spawn` one task per token, then the Responses API appends chunks in arrival order and suppresses the fallback full-delta path once any chunk arrives. `yield_now()` does not guarantee completion or ordering. Under scheduler jitter, clients can receive reordered text.
Suggested fix: deliver chunks synchronously or through a single ordered queue/task per response instead of spawning an independent task per token.

4. Medium: the NearAI streaming parser turns stream failures into successful partial completions.
`src/llm/nearai_chat.rs`
The SSE loops ignore eventsource errors and JSON parse failures and keep returning whatever partial text/tool buffer exists. In the tool path, truncated JSON arguments are deserialized as `{}`. A broken SSE stream can therefore surface partial text or malformed tool calls as if the request succeeded.
Suggested fix: treat malformed/truncated stream events as an error once streaming has started, and fail tool-call reconstruction if the arguments buffer is incomplete or invalid.

5. Medium: tool-call reconstruction loses the model's original call order.
`src/llm/nearai_chat.rs`
Tool calls are buffered by stream index, but then the final vector is sorted by opaque `id` instead of preserving stream index order. Multi-tool responses can execute in a different order than the model emitted.
Suggested fix: preserve the original stream index when assembling `tool_calls`.

6. Medium: plain text streaming likely breaks usage/cost accounting.
`src/llm/nearai_chat.rs`
The text streaming path sets only `stream=true`, while the tool path explicitly requests `stream_options.include_usage=true`. If the provider only includes usage when requested, streamed text responses will record zero tokens/cost.
Suggested fix: request usage consistently for both text and tool streaming paths and add a regression test that asserts non-zero usage on streamed text responses.

Residual risks:
- The new tests only verify that streaming emits events and that one path yields multiple deltas. They do not cover wrapper parity, v2 routing, chunk ordering, streamed usage accounting, or mid-stream failure handling.
- I started a targeted `cargo test accumulator_stream_chunks_then_response --lib`, but the workspace was still compiling when I stopped polling, so I do not have a completed test result to report.

serrrfirat:
Thank you for implementing true incremental model output. We are closing this PR because the functionality is required but has now been implemented end-to-end in the Reborn architecture.

The current path is: the Reborn model gateway calls `complete_with_tools_streaming` (`crates/ironclaw_runner/src/model_gateway.rs`), incremental text is projected as `ModelTextDelta`, and `crates/ironclaw_reborn_openai_compat/src/streaming.rs` translates it into the exact `response.output_text.delta` and `response.output_text.done` SSE events for `stream: true` Responses API requests. This avoids coupling provider callbacks directly to the retired v1 dispatcher/channel manager.

One useful follow-up remains: strengthen the Reborn Responses API E2E coverage so it asserts multiple incremental text deltas arriving before completion, rather than only checking that some SSE events were emitted. We would love your help with that regression test or other Reborn streaming improvements. Thank you for pushing this capability forward.

PR #1943, opened 2026-04-03
fix(telegram): handle 'message is too long' with retry splitting

Evrard-Nil:
## Summary
- Reduce `TELEGRAM_MAX_MESSAGE_LEN` from 4096 to 4000 for safety margin against Markdown entity/emoji counting edge cases
- Add `SendError::TooLong` variant to detect "message is too long" 400 responses
- Add `send_chunk()` helper that recursively halves chunks on TooLong rejections (up to `MAX_SPLIT_DEPTH = 2` levels deep, max 4 sub-messages per chunk), splitting at natural boundaries (paragraph > newline > space)
- Use UTF-16 code unit length for split midpoint calculation (matching Telegram's actual limit)
- Pass `use_markdown` flag through recursion so ParseEntities fallback disables Markdown for all subsequent splits

Fixes the recurring error:
```
Telegram API returned 400: {"ok":false,"error_code":400,"description":"Bad Request: message is too long"}
```

## Test plan
- [x] All 56 existing + new telegram-channel tests pass
- [x] `test_split_message_each_chunk_fits` validates against 4000-char limit
- [x] 5 new `find_split_midpoint` tests (paragraph, newline, space, no-boundary, emoji-heavy)
- [ ] Deploy and verify long agent responses no longer fail on Telegram

### DIFF
--- a/channels-src/telegram/src/lib.rs
+++ b/channels-src/telegram/src/lib.rs
@@ -981,6 +981,7 @@ fn normalize_thread_id(thread_id: Option<i64>) -> Option<i64> {
 /// Returns the sent message_id on success. When `parse_mode` is set and
 /// Telegram returns a 400 "can't parse entities" error, returns
 /// `SendError::ParseEntities` so the caller can retry without formatting.
+/// Returns `SendError::TooLong` when Telegram rejects the message as too long.
 fn send_message(
     chat_id: i64,
     text: &str,
@@ -1469,53 +1470,73 @@ fn send_chunk_inner(
                 channel_host::LogLevel::Warn,
                 &format!("Markdown parse failed ({}), retrying as plain text", detail),
             );
-            send_message(chat_id, text, reply_to, None, message_thread_id)
-                .map_err(|e| format!("Plain-text retry also failed: {}", e))
+            match send_message(chat_id, text, reply_to, None, message_thread_id) {
+                Ok(id) => Ok(id),
+                // Plain-text retry might still be too long — split it.
+                Err(SendError::TooLong(_)) if depth < 3 => {
+                    split_and_send(chat_id, text, reply_to, message_thread_id, depth)
+                }
+                Err(e) => Err(format!("Plain-text retry also failed: {}", e)),
+            }
         }
         Err(SendError::TooLong(_)) if depth < 3 => {
-            channel_host::log(
-                channel_host::LogLevel::Warn,
-                &format!(
-                    "Chunk too long ({} chars), splitting in half (depth {})",
-                    text.chars().count(),
-                    depth + 1
-                ),
-            );
-            // Find a midpoint at a natural boundary.
-            let mid_char = text.chars().count() / 2;
-            // safety: mid_byte comes from char_indices(), always a char boundary.
-            let mid_byte = text
-                .char_indices()
-                .nth(mid_char)
-                .map(|(i, _)| i)
-                .unwrap_or(text.len());
-            let split_at = text[..mid_byte] // safety: mid_byte from char_indices()
-                .rfind("\n\n")
-                .or_else(|| text[..mid_byte].rfind('\n')) // safety: char boundary
-                .or_else(|| text[..mid_byte].rfind(' ')) // safety: char boundary
-                .unwrap_or(mid_byte);
-            let split_at = if split_at == 0 { mid_byte } else { split_at };
-
-            let first = text[..split_at].trim_end(); // safety: split_at from rfind (char boundary)
-            let second = text[split_at..].trim_start(); // safety: same
-
-            let first_id =
-                send_chunk_inner(chat_id, first, reply_to, message_thread_id, depth + 1)?;
-
-            if !second.is_empty() {
-                send_chunk_inner(
-                    chat_id,
-                    second,
-                    Some(first_id),
-                    message_thread_id,
-                    depth + 1,
-                )?;
-            }
-
-            Ok(first_id)
+            split_and_send(chat_id, text, reply_to, message_thread_id, depth)
         }
         Err(e) => Err(e.to_string()),
     }
+}
+
+/// Split text in half at a natural boundary and send each part recursively.
+///
+/// Returns the message_id of the **last** sent chunk so callers can
+/// continue the reply thread correctly.
+fn split_and_send(
+    chat_id: i64,
+    text: &str,
+    reply_to: Option<i64>,
+    message_thread_id: Option<i64>,
+    depth: u8,
+) -> Result<i64, String> {
+    channel_host::log(
+        channel_host::LogLevel::Warn,
+        &format!(
+            "Chunk too long ({} chars), splitting in half (depth {})",
+            text.chars().count(),
+            depth + 1
+        ),
+    );
+    // Find a midpoint at a natural boundary.
+    let mid_char = text.chars().count() / 2;
+    // safety: mid_byte comes from char_indices(), always a char boundary.
+    let mid_byte = text
+        .char_indices()
+        .nth(mid_char)
+        .map(|(i, _)| i)
+        .unwrap_or(text.len());
+    let split_at = text[..mid_byte] // safety: mid_byte from char_indices()
+        .rfind("\n\n")
+        .or_else(|| text[..mid_byte].rfind('\n')) // safety: char boundary
+        .or_else(|| text[..mid_byte].rfind(' ')) // safety: char boundary
+        .unwrap_or(mid_byte);
+    let split_at = if split_at == 0 { mid_byte } else { split_at };
+
+    let first = text[..split_at].trim_end(); // safety: split_at from rfind (char boundary)
+    let second = text[split_at..].trim_start(); // safety: same
+
+    let first_id = send_chunk_inner(chat_id, first, reply_to, message_thread_id, depth + 1)?;
+
+    if second.is_empty() {
+        return Ok(first_id);
+    }
+
+    // Return last sent id so the reply chain continues from the final chunk.
+    send_chunk_inner(
+        chat_id,
+        second,
+        Some(first_id),
+        message_thread_id,
+        depth + 1,
+    )
 }
 
 /// Extract the base MIME type, stripping any parameters after `;`.
--- a/channels-src/telegram/src/lib.rs
+++ b/channels-src/telegram/src/lib.rs
@@ -409,7 +409,8 @@ fn truncate_status_message(input: &str, max_chars: usize) -> String {
     }
 }
 
-/// Split a long message into chunks that fit within Telegram's 4096 UTF-16-unit limit.
+/// Split a long message into chunks that each fit within `limit_utf16`
+/// UTF-16 code units.
 ///
 /// Tries to split at the most natural boundary available (in priority order):
 /// 1. Double newline (paragraph break)
@@ -417,8 +418,8 @@ fn truncate_status_message(input: &str, max_chars: usize) -> String {
 /// 3. Sentence end (`. `, `! `, `? `)
 /// 4. Word boundary (space)
 /// 5. Hard cut at the limit (last resort for pathological input)
-fn split_message(text: &str) -> Vec<String> {
-    if utf16_code_unit_len(text) <= TELEGRAM_MAX_MESSAGE_LEN {
+fn split_message(text: &str, limit_utf16: usize) -> Vec<String> {
+    if utf16_code_unit_len(text) <= limit_utf16 {
         return vec![text.to_string()];
     }
 
@@ -426,8 +427,8 @@ fn split_message(text: &str) -> Vec<String> {
     let mut remaining = text;
 
     while !remaining.is_empty() {
-        // Find the longest UTF-8 prefix that fits within Telegram's UTF-16 limit.
-        let window_bytes = prefix_within_utf16_limit(remaining, TELEGRAM_MAX_MESSAGE_LEN);
+        // Find the longest UTF-8 prefix that fits within the UTF-16 limit.
+        let window_bytes = prefix_within_utf16_limit(remaining, limit_utf16);
 
         if window_bytes >= remaining.len() {
             // Remainder fits entirely.
@@ -1428,7 +1429,7 @@ fn send_response(
     }
 
     // Split large messages into chunks that fit Telegram's limit.
-    let chunks = split_message(&response.content);
+    let chunks = split_message(&response.content, TELEGRAM_MAX_MESSAGE_LEN);
 
     // The first chunk replies to the original message; subsequent chunks
     // reply to the previously sent chunk so they form a visual thread.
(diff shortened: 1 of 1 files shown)


serrrfirat (commented):
Findings:
1. High: `send_chunk_inner()` returns `first_id` after a recursive `TooLong` split, even when it has sent a second follow-up chunk. `send_response()` uses the returned id as the next `reply_to`, so subsequent chunks will reply to the first half rather than the last sent half, breaking the intended linear Telegram thread.
2. Medium: there is no test coverage for the new recursive `TooLong` path or its return-value contract. The existing `split_message` tests do not exercise `send_chunk_inner()` or verify that a recursively split send returns the final sent message id.

Residual risks:
- In the `ParseEntities` fallback path, a plain-text retry that then hits `TooLong` still bubbles up as an error instead of being split again.
- Lowering the proactive split threshold from 4096 to 4000 is a user-visible behavior change for messages in the 4001-4096 range; that may be intentional, but it should be treated as such.

Validation: `cargo test --lib` passed in `channels-src/telegram`.

serrrfirat:
**Medium Severity — Markdown retry wastes API calls after ParseEntities+TooLong**

When the flow is: Markdown → ParseEntities → plain text → TooLong → `split_and_send`, the recursive `send_chunk_inner` calls attempt to send sub-chunks with **Markdown** again (line ~1403 always uses `Some("Markdown")`). A chunk that already failed Markdown parsing will be retried with Markdown in the split path, potentially failing again with ParseEntities, then falling back to plain text. While this converges (the halves are smaller), it doubles API calls.

**Suggested fix:** Pass a `use_markdown: bool` flag through the recursion so that once ParseEntities is hit, all subsequent splits send as plain text directly. Or accept the extra calls as a tradeoff and document it.

---

**Medium Severity — No unit tests for core new functions**

`send_chunk`, `send_chunk_inner`, and `split_and_send` are the core new functions but have zero unit test coverage. The test plan is manual only.

**Suggested fix:** Extract the `split_and_send` splitting logic (pure computation: finding mid_char, split_at, slicing) into a standalone testable function, and add unit tests with edge cases (all-newlines, no-spaces, single-char, exactly-at-boundary).

serrrfirat (changes requested):
## Paranoid Architect Review — Verdict: 🔄 REQUEST CHANGES (2 High)

⚠️ **This PR also has merge conflicts and needs a rebase onto staging.**

The PR solves a real production problem (Telegram rejecting messages due to Markdown entity overhead) with a reasonable recursive-halving approach. The recursion depth is bounded and error handling follows existing patterns. However, there's a correctness issue in the splitting logic.

### Findings

| 1 | **High** | Correctness | lib.rs `split_and_send` | **UTF-16/chars unit mismatch.** Uses `chars().count() / 2` for midpoint, but Telegram enforces UTF-16 code unit limits. Existing `split_message` correctly uses `utf16_code_unit_len()`. For emoji-heavy messages (common on Telegram), `chars().count()` underestimates UTF-16 length. Halves may still exceed UTF-16 limit, causing extra recursion or failure at depth 3. | Use `utf16_code_unit_len(text) / 2` for midpoint, convert back via `prefix_within_utf16_limit()` |
| 2 | **High** | Resource | lib.rs | **Exponential message count.** Depth-3 recursion: up to 8 messages per chunk. With 10 pre-split chunks = up to 80 API calls. Telegram rate limit is ~30 msg/sec per chat. | Add total message count guard or warning log. Consider depth 2 (max 4 sub-messages). |
| 3 | Medium | String Safety | lib.rs `split_and_send` | Split boundary `\n\n` characters silently discarded via `trim_end()`/`trim_start()`. Acceptable but undocumented. | Add comment |
| 4 | Medium | Edge Case | lib.rs `split_and_send` | Empty text: both halves become `""`, Telegram rejects. Practically unreachable but unguarded. | Add `if text.is_empty() { return Err(...) }` |
| 5 | Medium | Stale Docs | lib.rs:409 | Doc comment still says "4096 UTF-16-unit limit" but constant changed to 4000 | Update reference |
| 6 | Medium | Test Coverage | tests | **No tests** for `send_chunk`, `send_chunk_inner`, or `split_and_send`. Core behavioral change untested. | Extract `find_midpoint_split()` as pure function, add unit tests |
| 7 | Low | Logging | lib.rs | Old code logged per-chunk success. New code silent on success. | Add `debug!` in Ok branch |
| 8 | Low | Error Quality | lib.rs | Depth-limit error gives no indication recursion exhausted | Wrap: "still too long after splitting N levels deep" |
| 9 | Low | Docs | lib.rs | 96-char safety margin (4096 to 4000) rationale undocumented | Document worst-case observed or make configurable |
| 10 | Nit | Naming | lib.rs | `split_and_send` describes implementation not intent | Consider `halve_and_retry_send` |

### Summary
Finding #1 is the key correctness issue: the new splitting logic uses `chars()` while the existing code uses UTF-16 code units. For emoji-heavy messages, this mismatch means the retry mechanism can fail to produce small-enough chunks. Finding #2 notes the potential for rate limit issues. Both should be addressed.

serrrfirat (dismissed):
Looks good — well-structured recursive splitting with proper bounds. One medium finding posted inline.

channels-src/telegram/src/lib.rs
```diff
@@ -3184,4 +3322,63 @@ mod tests {
...
+
+    // ---- find_split_midpoint tests ----
+
+    #[test]
+    fn test_find_split_midpoint_paragraph_boundary() {
+        let first = "A".repeat(2000);
+        let second = "B".repeat(2000);
+        let text = format!("{}\n\n{}", first, second);
+        let idx = find_split_midpoint(&text);
+        assert_eq!(&text[..idx].trim_end(), &first.as_str());
```
serrrfirat: **Medium Severity — Test passes for the wrong reason**

`test_find_split_midpoint_paragraph_boundary` claims to test paragraph splitting, but the `\n\n` falls exactly at the `mid_byte` boundary (bytes 2000–2001), so `text[..2001]` only includes the first `\n`. `rfind("\n\n")` misses it and `rfind('\n')` matches instead — the test exercises newline splitting, not paragraph splitting.

This means a regression in the `rfind("\n\n")` branch of `find_split_midpoint` would go undetected.

**Suggested fix:** shift the `\n\n` earlier so it falls well within `text[..mid_byte]`:

```rust
let first = "A".repeat(1500);
let second = "B".repeat(2500);
let text = format!("{}\n\n{}", first, second);
let idx = find_split_midpoint(&text);
assert_eq!(idx, 1500); // rfind("\n\n") at byte 1500
```

channels-src/telegram/src/lib.rs
```diff
@@ -1474,6 +1444,172 @@ fn send_response(
...
+        mid_byte
+    };
+
+    let split_at = text[..mid_byte] // safety: mid_byte from prefix_within_utf16_limit/char_indices
+        .rfind("\n\n")
+        .or_else(|| text[..mid_byte].rfind('\n')) // safety: same char boundary
+        .or_else(|| text[..mid_byte].rfind(' ')) // safety: same char boundary
+        .unwrap_or(mid_byte);
+
+    if split_at == 0 { mid_byte } else { split_at }
```
ilblackdragon: **HIGH — Adversarial boundary preference can exhaust `MAX_SPLIT_DEPTH` without meaningfully shrinking the payload.**

`find_split_midpoint` searches backwards from the UTF-16 midpoint for `\n\n` -> `\n` -> `' '` and splits at the first match. When the *early* part of the chunk contains a paragraph/newline boundary but the *late* part is boundary-free (e.g. `"Hello!\n\n" + "x".repeat(8000)` — a rendered code block, URL, stack trace, hash, or base64 blob), the first boundary found near the start wins:

- Pass 1 (depth 0): split at byte 7. First half = `"Hello!"`, second half = 8000 x's (~8000 UTF-16 units — still way over 4000).
- Pass 2 (depth 1): same problem on the 8000-x second half. No boundary, so hard-cut at mid ~ 4000 x's. Good.
- Pass 3 (depth 2): second half now ~4000 x's. Telegram may still reject for entity/emoji reasons. Depth limit hit -> `Err(...)` -> entire response dropped.

The natural-boundary preference is *backwards-first-wins*, which is fine when the boundary lives near the midpoint but pathological when it lives near the start. Consider: (a) requiring the boundary to be past some minimum offset (e.g. `> mid_byte / 2`), or (b) falling back to a hard cut at `mid_byte` when the chosen boundary would leave the second half still over-limit, or (c) raising `MAX_SPLIT_DEPTH` with an exponential-backoff style hard-cut floor. Real agent output (code blocks, JSON, SSH keys, base64) routinely has this front-loaded-boundary shape.

channels-src/telegram/src/lib.rs
```diff
@@ -1474,6 +1444,172 @@ fn send_response(
...
+            use_markdown,
+        ),
+        Err(SendError::TooLong(_)) => Err(format!(
+            "Message still too long after splitting {} levels deep ({} UTF-16 units)",
+            depth,
+            utf16_code_unit_len(text),
+        )),
+        Err(e) => Err(e.to_string()),
+    }
+}
```
ilblackdragon, line 1532: **HIGH — No regression test covers the `TooLong` -> retry / split flow, only `find_split_midpoint` in isolation.**

This is the exact 'test through the caller, not the helper' pattern in `.claude/rules/testing.md`: `find_split_midpoint` is a pure transform whose output gates a side effect (HTTP call to Telegram) through multiple layers (`split_and_send` -> `send_chunk_inner`). The PR adds five unit tests for the helper but zero tests for:

- `send_chunk_inner` receiving `TooLong` and splitting correctly
- Depth counter actually increments and terminates at `MAX_SPLIT_DEPTH`
- Markdown flag correctly flipping to `false` after a `ParseEntities` fallback AND propagating through subsequent `split_and_send` recursion
- Reply-thread ID chain: `split_and_send` returns the *last* sent id, not the first — a silent regression there would break threading on long messages

Since `send_message` calls `channel_host::http_request` directly (WIT binding), mocking is awkward, but the orchestration logic (depth, use_markdown plumbing, last-id semantics) could be extracted into a pure function that takes a `Fn(&str, Option<&str>) -> Result<i64, SendError>` and tested. Without this, the `use_markdown` flag bug the reviewer can't write (because it's hypothetical) would ship silently. The commit-msg regression-test hook likely passes because new tests *were* added, but they don't cover the actual bug surface the PR introduces.

channels-src/telegram/src/lib.rs
```diff
@@ -1474,6 +1444,172 @@ fn send_response(
...
+            reply_to,
+            message_thread_id,
+            depth,
+            use_markdown,
+        ),
+        Err(SendError::TooLong(_)) => Err(format!(
+            "Message still too long after splitting {} levels deep ({} UTF-16 units)",
+            depth,
+            utf16_code_unit_len(text),
+        )),
```
ilblackdragon, line 1529: **MEDIUM — Final failure swallows the remaining chunks of the response.**

When `send_chunk_inner` hits `MAX_SPLIT_DEPTH` and returns `Err`, the caller in `send_response` (around line 1437-1442) uses `?` to propagate the error and abandons the loop. If chunk 2 of 5 fails to split, chunks 3-5 are never attempted — the user sees a partial reply and a silent error log. Given this PR exists because responses routinely exceed the length budget, the more defensive failure mode is: log the offending chunk, skip it with a placeholder (e.g. 'one segment was too long to deliver, see workspace log'), and continue sending subsequent chunks. Alternatively, a depth-exceeded fallback that hard-cuts at `TELEGRAM_MAX_MESSAGE_LEN / 2` UTF-16 units (no boundary preference) would produce *something* deliverable.

channels-src/telegram/src/lib.rs
```diff
@@ -1021,6 +1031,9 @@ fn send_message(
                 if body_str.contains("can't parse entities") {
                     return Err(SendError::ParseEntities(body_str.to_string()));
                 }
+                if body_str.contains("message is too long") {
+                    return Err(SendError::TooLong(body_str.to_string()));
+                }
```
ilblackdragon, line 1037: **MEDIUM — Error classification by English substring is fragile.**

`body_str.contains("message is too long")` and the sibling `body_str.contains("can't parse entities")` both classify by English wording of the `description` field. Telegram's Bot API is historically English-only, so this is likely safe today, but it will silently fall through to `SendError::Other` on any wording change (e.g., `"Message is too long"` with capital M, or the known variant `"message_too_long"` seen in some endpoints). That fallthrough drops the entire response with no retry.

Safer: also check a Bot-API-stable signal where possible. The `400` status alone plus a description pattern match (case-insensitive `to_ascii_lowercase().contains("too long")`) would survive most plausible changes, and failing that, a response-parsing step that decodes `{ "ok": false, "error_code": 400, "description": ... }` explicitly would beat raw substring search. Consider a single helper `classify_400(body: &str) -> SendError` so both checks share the lowercasing and are easy to update in one place.

channels-src/telegram/src/lib.rs
```diff
@@ -1474,6 +1444,172 @@ fn send_response(
...
+            return Err("Cannot split text into non-empty chunks".to_string());
+        }
+        return send_chunk_inner(
+            chat_id,
+            second,
+            reply_to,
+            message_thread_id,
+            depth + 1,
+            use_markdown,
+        );
```
ilblackdragon: **MEDIUM — `split_and_send`'s leading-whitespace fallback wastes the depth budget.**

When `find_split_midpoint` returns a small `split_at` (e.g. 5, the end of a short leading word), `first = text[..5].trim_end()` may become the entire short prefix and `second` the ~8000-byte tail. On the tail's recursion, `depth` has already incremented, even though the first pass made effectively no progress.

The `if first.is_empty()` guard (1575-1586) handles the pathological case where `trim_end` makes `first` empty by recursing *without* incrementing the split — but it still uses `depth + 1`, so even this defensive branch burns a depth level for zero progress. For adversarial text this compounds with the comment above: front-loaded boundaries can burn the full depth budget before the actually-long tail ever gets hard-cut.

ilblackdragon:
## Architectural observations (non-blocking)

**Why 4000, not 4096?** The constant is documented as a 'safety margin against Markdown entity/emoji counting edge cases', but the retry/split machinery this PR adds would *already* catch a `TooLong` rejection at 4001-4096 and recover. Leaving the constant at 4096 and relying on the new `TooLong` path for the edge cases keeps the primary path closer to Telegram's own limit (fewer, larger chunks, less threading noise), and the retry path exists specifically for the 4% of cases where 4096 is wrong. This is a judgement call — if the motivation is 'avoid the first-time failure that triggers the retry at all', 4000 is sensible; if it's defensive duplication, the retry makes it redundant. Worth a line in the commit body.

**Keep the retry-and-split logic in one place.** The `ParseEntities` fallback lives in `send_chunk_inner`; the `TooLong` split lives partly there and partly in `split_and_send`. A single small state machine (`attempt(depth, markdown)`) would make the depth and markdown flag propagation obvious in one diff. As-is, the markdown-disable-after-parse-error path goes through `split_and_send(..., false)` at line 1497, and correctly threads `use_markdown=false` forward — but verifying this by reading the three-function handoff takes longer than it should.

**Thematic concern: the whole send path bypasses `ToolDispatcher::dispatch()`.** This is pre-existing and WASM channels are specifically called out as an 'everything goes through tools' target in `.claude/rules/tools.md`: 'All actions originating from... WASM channels... MUST go through `ToolDispatcher::dispatch()`'. A Telegram message-send is a side-effecting action that currently invokes `channel_host::http_request` directly without an `ActionRecord` / audit trail. Out of scope for this PR, but worth tracking — every PR that extends this direct send path entrenches the divergence.

ilblackdragon:
Refactored to remove the duplicate splitter. Now there's one `split_message(text, limit_utf16)` with a single boundary hierarchy (paragraph → newline → sentence → word → hard cut), and the `TooLong` retry path calls it with a halved limit instead of re-implementing its own midpoint + boundary search.

- `find_split_midpoint` and `split_and_send` are gone (-163 / +109 net).
- `resplit_and_send` just calls `split_message(text, limit/2)` and threads each sub-chunk through `send_chunk_inner`. `MAX_SPLIT_DEPTH = 2` still bounds retries.
- `ParseEntities` fallback now recurses into `send_chunk_inner` with `use_markdown = false` instead of duplicating the send-and-handle-TooLong logic.
- Retries now also benefit from the sentence-end boundary (`. `, `! `, `? `), which the old midpoint splitter dropped.

All 53 tests pass (`cargo test` in `channels-src/telegram`); clippy clean. Dropped the 5 `find_split_midpoint_*` unit tests and added 2 covering the halved-limit path (paragraph-boundary preference + fit guarantee).

ilblackdragon (approved):

Issue #1956, opened 2026-04-03
feat(security): incident logging and automated threat response for policy denials


ilblackdragon:
## Summary

Policy denials, safety check failures, lease violations, and other security-relevant events are currently either silent or logged at `debug!` level — invisible in production. We need a proper incident logging system that persists to the database and supports automated threat response.

## Problem

Today, the following security events produce no durable record:

### Silent or debug-only events
- **PolicyEngine denials** — `PolicyEngine::evaluate()` returns `Deny` with no logging. Brute-force capability escalation attempts are invisible.
- **Lease violations** — `LeaseNotFound`, `LeaseExpired`, `LeaseDenied` errors are caught by callers but not recorded as security events.
- **Safety layer rejections** — prompt injection detection, leak detection, and validation failures in `ironclaw_safety` produce tool errors but no centralized incident record.
- **Orchestrator self-modification blocks** — `ORCHESTRATOR_SELF_MODIFY` gate rejections log at `debug!` only.
- **Rate limit hits** — per-tool rate limiting triggers are not recorded beyond the immediate error response.
- **Auth failures** — invalid tokens, expired sessions, and role check failures in the web gateway.
- **Sandbox escapes / resource limit hits** — Monty VM kills (timeout, memory, allocation limit) are not recorded as incidents.

### No automated response
When a safety check or policy denial fires, there is no mechanism to:
- Identify the originating channel, tool, user, or thread
- Correlate repeated violations from the same source
- Automatically disable a misbehaving channel, tool, or user after repeated violations
- Alert administrators

### 1. Incident log (database-backed)

A new `SecurityIncident` record type persisted to the database (both PostgreSQL and libSQL backends):

```rust
struct SecurityIncident {
    id: Uuid,
    timestamp: DateTime<Utc>,
    severity: IncidentSeverity,     // Info, Warning, Critical
    category: IncidentCategory,     // PolicyDenial, LeaseViolation, SafetyRejection, AuthFailure, RateLimitHit, SandboxKill, SelfModifyBlock
    source_user_id: Option<String>,
    source_channel: Option<String>,
    source_tool: Option<String>,
    thread_id: Option<ThreadId>,
    action_name: Option<String>,
    detail: String,                 // Human-readable description
    metadata: serde_json::Value,    // Structured context (params, policy rule, etc.)
}
```

### 2. Incident recording points

Add `record_incident()` calls at each security boundary:

| Location | Category | Severity |
|----------|----------|----------|
[...]

PR #1958, opened 2026-04-03
fix(engine): security hardening for v2 orchestrator and Monty sandbox

ilblackdragon:
## Summary

- **C1/C2**: Approval gate + syntax validation for orchestrator self-modification when `ORCHESTRATOR_SELF_MODIFY=true`
- **C4**: 6 sandbox security negative tests (rlm_query depth, FINAL injection, tool name injection, context mutation, deep recursion, syntax validation)
- **M2**: Remove dead `stop_thread_system`/`inject_message_system` methods that bypassed ownership checks
- **Round 2**: Close path-normalization bypass, forgeable-metadata escalation, and orchestrator rehydration holes surfaced in @serrrfirat's review
- Document child lease budget snapshot semantics, add `EngineError::InvalidInput` variant

Addresses deferred security items from #1557 review. Related: #1956 (incident logging).

### C1/C2 — Orchestrator approval gates (when self-modify is enabled)

**Layer A** — `memory_write` tool returns `ApprovalRequirement::UnlessAutoApproved` for protected orchestrator paths. The v2 effect bridge maps this to a resumable `GatePaused(Approval)` so human approval can pause + resume the write. (Earlier rounds used `Always`, which the bridge mapped to `LeaseDenied` — a permanent refusal that made the gate unusable; fixed.)

**Layer B** — Store adapter validates Python syntax via Monty parser before persisting orchestrator patches. Broken code is rejected upfront instead of consuming failure-budget slots (3 failures trigger auto-rollback). Content hash (SHA-256) stamped on all protected docs for write-time audit (not runtime integrity — the workspace file is the trust boundary).

### Round 2 — Path bypass, forgeable metadata, rehydration, and env-var fragility

- **Path normalization bypass (critical)**: `is_protected_orchestrator_path` and `is_orchestrator_code_path` did raw `starts_with`, so `engine/./orchestrator/v3.py`, `engine//orchestrator/v3.py`, and `engine/knowledge/../orchestrator/v3.py` slipped past the gate. Both guards now normalize first (strip `.`/empty segments, reject `..` traversal) and match against both `.system/engine/orchestrator/` and the legacy `engine/orchestrator/` prefix.
- **Forgeable `is_system_internal` (high)**: the gate previously exempted any doc whose `metadata.source == "compiled_in"`. An LLM tool call could craft that field and bypass denial + validation. Replaced with a `tokio::task_local!`-backed trusted-write scope (`ironclaw_engine::runtime::with_trusted_internal_writes`) that `MissionManager::seed_orchestrator_v0` enters; the store gate reads the task-local, which cannot be set from outside a trusted callsite and does not propagate across `tokio::spawn`.
- **Rehydration invisibility (high)**: `synthesize_orchestrator_doc_from_py` returned docs with `ProjectId::nil()`, but `list_memory_docs` filters by exact project, so persisted `.py` orchestrator files disappeared from project-scoped queries after restart. Overrode `HybridStore::list_shared_memory_docs` to surface globally-stored docs (orchestrator, failure tracker, prompt overlay) for any project query, matching the physically-global storage layout.
- **Env var read per call (high)**: `ORCHESTRATOR_SELF_MODIFY` was read on every gate check. Centralized in `runtime::self_modify_enabled()` — a process-wide `OnceLock<bool>` seeded on first read; tool, store, engine loop, and self-improvement mission all share the same snapshot. `SelfModifyTestGuard` provides a serialized dev-build override for tests (compiled out of release builds).
- **Frontmatter `project_id` / `user_id` loss**: fixed by persisting both fields in `serialize_knowledge_doc` and parsing them in `deserialize_knowledge_doc`. Round-trip test in `store_adapter::migration_tests::knowledge_md_doc_round_trips_project_id_and_user_id`.

### C4 — Sandbox security tests

| Test | What it proves |
|------|---------------|
| `sandbox_enforces_rlm_query_depth_limit` | Depth check at max recursion refuses with clear error |
| `sandbox_rejects_final_injection` | FINAL() captures payload literally, no code execution |
| `sandbox_rejects_tool_name_injection` | Dynamic name construction can't bypass lease enforcement |
| `sandbox_context_variable_is_not_mutable` | Python-side mutations don't propagate to Rust thread state |
| `sandbox_handles_deep_recursion` | Infinite recursion terminates safely (no crash/hang) |
| `validate_syntax_rejects_broken_code` | Syntax validation unit test |

### M2 — Remove ownership-bypassing thread operations

`stop_thread_system()` and `inject_message_system()` had zero callers anywhere in the codebase — dead code that was a privilege escalation footgun. Removed entirely; ownership-checking variants are the only API.

### Round 2 test coverage (caller-level, per `.claude/rules/testing.md`)

Unit tests (41 new):
- path normalization & protected-path guard (9 cases), `requires_approval` branches (7 cases), store-adapter helper predicates + `validate_orchestrator_content` + `synthesize_orchestrator_doc_from_py` (23 cases), trusted-write scope semantics (3 cases), `list_shared_memory_docs` override (2 cases)

Integration tests (11 new, libsql feature):
- `tools::dispatch::integration_tests` — drives `ToolDispatcher::dispatch()` against the real `MemoryWriteTool` for every bypass vector (logical alias, physical path, dot segment, double slash, traversal, unprotected baseline)
- `bridge::store_adapter::migration_tests::orchestrator_py_round_trips_through_restart` — full write → restart → load → cross-project query cycle
- `bridge::store_adapter::migration_tests::knowledge_md_doc_round_trips_project_id_and_user_id` — frontmatter project/user survive restart
- `bridge::store_adapter::migration_tests::invalid_python_orchestrator_is_rejected_at_write_time` — validator gate fires before persistence
- `bridge::effect_adapter::tests::memory_write_orchestrator_target_paused_for_approval_when_self_modify_enabled` — asserts `UnlessAutoApproved` produces `GatePaused(Approval)`, not `LeaseDenied` (the round-1 miss)
- `bridge::effect_adapter::tests::memory_write_orchestrator_target_refused_when_self_modify_disabled` — asserts no gate pauses when self-modify is off

### Other

- Document child lease budget snapshot semantics in `derive_child_leases` (intentional: independent budgets avoid cross-thread contention)
- Add `EngineError::InvalidInput` variant for validation failures
- Re-export `validate_python_syntax` from executor module

## Test plan

- [x] `cargo fmt` — clean
- [x] `cargo clippy -p ironclaw --lib --tests --features libsql` — 0 warnings
- [x] `cargo clippy -p ironclaw_engine --all-targets` — 0 warnings
- [x] `cargo test -p ironclaw_engine` — 358/358 pass
- [x] `cargo test -p ironclaw --lib --features libsql` — 4735/4735 pass
- [x] `cargo test --test engine_v2_gate_integration --test engine_v2_skill_codeact` — 27/27 pass


src/tools/builtin/memory.rs
```diff
@@ -279,6 +279,23 @@ impl Tool for MemoryWriteTool {
...
+        // When orchestrator self-modification is enabled, writing to protected
+        // paths (orchestrator code, prompt overlays) always requires explicit
+        // human approval — even if the session has auto-approve enabled.
+        let target = params.get("target").and_then(|v| v.as_str()).unwrap_or("");
+        if is_protected_orchestrator_path(target) {
+            let self_modify_enabled = std::env::var("ORCHESTRATOR_SELF_MODIFY")
+                .map(|v| v == "true" || v == "1")
+                .unwrap_or(false);
+            if self_modify_enabled {
+                return ApprovalRequirement::Always;
```
serrrfirat: **High Severity**

This `ApprovalRequirement::Always` is not approvable in engine v2 today. `EffectBridgeAdapter::execute_action()` still maps `Always` to `EngineError::LeaseDenied`, and only `UnlessAutoApproved` becomes `GatePaused(Approval)`. That means with `ORCHESTRATOR_SELF_MODIFY=true`, a call like `memory_write(target="prompt:codeact_preamble", ...)` fails outright instead of pausing for human approval, so the protected self-modify path is unusable.

Suggested fix: either return `UnlessAutoApproved` here and rely on the approval gate to suppress the `allow always` option for these targets, or update the v2 effect bridge so `Always` produces a resumable approval gate instead of a hard denial.

src/tools/builtin/memory.rs
```diff
@@ -279,6 +279,23 @@ impl Tool for MemoryWriteTool {
         })
     }
 
+    fn requires_approval(&self, params: &serde_json::Value) -> ApprovalRequirement {
+        // When orchestrator self-modification is enabled, writing to protected
+        // paths (orchestrator code, prompt overlays) always requires explicit
+        // human approval — even if the session has auto-approve enabled.
+        let target = params.get("target").and_then(|v| v.as_str()).unwrap_or("");
```
serrrfirat, line 388: **Medium Severity**

This check only protects the logical aliases (`prompt:*` / `orchestrator:*`). The persisted prompt overlay actually lives at `engine/orchestrator/codeact-preamble-overlay.md`, and `memory_write` still accepts arbitrary workspace paths. With self-modify enabled, writing that real path avoids the new approval gate while still mutating the runtime overlay.

Suggested fix: canonicalize `target` through the same engine-doc path mapping used by the store adapter, and treat the persisted engine paths as protected too.

src/bridge/store_adapter.rs
```diff
@@ -1156,11 +1176,34 @@ impl Store for HybridStore {
...
+                        .is_some_and(|s| s == "compiled_in");
+                if !is_system_internal {
+                    validate_orchestrator_content(doc)?;
+                }
             }
         }
 
-        self.docs.write().await.insert(doc.id, doc.clone());
-        self.persist_doc(doc).await;
+        let mut stamped = doc.clone();
```
serrrfirat, line 1567: **High Severity**

This persists frontmatter-based engine docs, but the restart path currently reconstructs them with `project_id = nil` and `user_id = "legacy"` in `deserialize_knowledge_doc()`. `list_memory_docs()` later filters by the real project/user pair, so after `load_state_from_workspace()` these persisted markdown docs disappear from `list_shared_memory_docs(project_id)` / `list_memory_docs(...)` and stop participating in prompt/orchestrator loading.

Suggested fix: persist project/user in frontmatter, or restore them during `load_knowledge_docs()` before inserting docs into `self.docs`. A restart round-trip test through `HybridStore::load_state_from_workspace()` would catch this.

src/bridge/store_adapter.rs
```diff
@@ -1156,11 +1176,34 @@ impl Store for HybridStore {
...
+            use sha2::{Digest, Sha256};
+            let hash = format!("{:x}", Sha256::digest(doc.content.as_bytes()));
+            stamped
+                .metadata
+                .as_object_mut()
+                .map(|m| m.insert("content_hash".into(), serde_json::Value::String(hash)));
+        }
+
+        self.docs.write().await.insert(stamped.id, stamped.clone());
+        self.persist_doc(&stamped).await;
```
serrrfirat, line 1598: **High Severity**

This write path persists runtime orchestrator versions as raw `engine/orchestrator/vN.py`, but startup only reconstructs docs from frontmatter markdown or JSON. Those `.py` files get skipped by `load_knowledge_docs()`, so saved orchestrator patches do not survive a process restart and `load_orchestrator()` falls back to the compiled-in default.

Suggested fix: either persist orchestrator versions in a loadable `MemoryDoc` format, or synthesize `MemoryDoc`s from the `vN.py` files during startup. This also needs a restart persistence test.

serrrfirat (commented):
Posting four review findings from the local audit.

src/bridge/store_adapter.rs
```diff
@@ -1156,11 +1176,34 @@ impl Store for HybridStore {
                         ),
                     });
                 }
+            } else {
+                // Self-modification is enabled — validate content before persisting.
+                // Skip validation for system-internal docs (failure tracker, compiled-in seed).
+                let is_system_internal = doc.title == ORCHESTRATOR_FAILURES_TITLE
+                    || doc
+                        .metadata
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in 2bdd5c05 — replaced string literal with `ORCHESTRATOR_FAILURES_TITLE` constant.

src/bridge/store_adapter.rs
```diff
@@ -1156,11 +1176,34 @@ impl Store for HybridStore {
...
-        self.persist_doc(doc).await;
+        let mut stamped = doc.clone();
+        // Stamp a content hash for audit trail on all protected docs.
+        if is_protected_orchestrator_doc(doc) {
+            use sha2::{Digest, Sha256};
+            let hash = format!("{:x}", Sha256::digest(doc.content.as_bytes()));
+            stamped
+                .metadata
+                .as_object_mut()
+                .map(|m| m.insert("content_hash".into(), serde_json::Value::String(hash)));
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in 2bdd5c05 — metadata is now normalized to `{}` before stamping the content hash.

src/bridge/store_adapter.rs
```diff
@@ -1156,11 +1176,34 @@ impl Store for HybridStore {
...
+        }
```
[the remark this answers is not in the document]
ilblackdragon: The hash is an audit trail recording what was written to the store, not a runtime integrity check. Hashing the full content is the correct behavior — it lets you verify what was persisted matches what was submitted. The runtime truncation is a display concern, not a storage concern.

crates/ironclaw_engine/src/executor/scripting.rs
```diff
@@ -53,6 +53,24 @@ fn default_limits() -> ResourceLimits {
...
+/// Check whether `code` is syntactically valid Python without executing it.
+///
+/// Uses Monty's parser (same as execution) so the syntax check is identical
+/// to what would happen at runtime. Returns `Ok(())` if valid, or an error
+/// message describing the syntax problem.
+pub fn validate_python_syntax(code: &str) -> Result<(), String> {
+    let code_owned = code.to_string();
+    match std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| {
+        MontyRun::new(code_owned, "validate.py", vec![])
+    })) {
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in 2bdd5c05 — added a 256KB size cap (`MAX_ORCHESTRATOR_SOURCE_BYTES`) before invoking the Monty parser. The compiled-in default is ~2KB so this is generous but prevents pathological inputs.

crates/ironclaw_engine/src/executor/scripting.rs
```diff
@@ -2154,4 +2172,216 @@ FINAL(str(x))
...
+            &ctx,
+            &[],
+            &serde_json::json!({}),
+        )
+        .await
+        .unwrap();
+        let answer = result.final_answer.as_deref().unwrap_or("");
+        assert!(
+            !answer.starts_with("ESCAPED"),
+            "dynamic name construction should not bypass leases, got: {answer}",
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in 2bdd5c05 — added a second assertion verifying the expected outcome (`not_found` or `blocked:*`), not just absence of `ESCAPED`.

src/tools/builtin/memory.rs
```diff
@@ -279,6 +279,23 @@ impl Tool for MemoryWriteTool {
...
+            let self_modify_enabled = std::env::var("ORCHESTRATOR_SELF_MODIFY")
+                .map(|v| v == "true" || v == "1")
+                .unwrap_or(false);
+            if self_modify_enabled {
+                return ApprovalRequirement::Always;
+            }
+            // When disabled, execute() returns NotAuthorized before any write.
+        }
+        ApprovalRequirement::Never
+    }
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in 2bdd5c05 — changed to `ApprovalRequirement::UnlessAutoApproved`. The v2 effect bridge pauses on this variant (producing a resumable `GatePaused` approval gate) rather than hard-denying. This makes the protected self-modify path usable with human approval.

src/tools/builtin/memory.rs
ilblackdragon: Fixed in 2bdd5c05 — changed to `UnlessAutoApproved`. The v2 effect bridge maps this to a resumable `GatePaused(Approval)` gate, so the user sees an approval prompt and can approve/deny. `Always` was incorrectly hard-denying via `LeaseDenied`.

src/tools/builtin/memory.rs
ilblackdragon: Fixed in 2bdd5c05 — `is_protected_orchestrator_path()` now also matches physical workspace paths (`engine/orchestrator/*`). Writing to the persisted file path directly will now trigger the same approval gate.

src/bridge/store_adapter.rs
ilblackdragon: Fixed in 2bdd5c05 — `serialize_knowledge_doc()` now persists `project_id` and `user_id` in frontmatter, and `deserialize_knowledge_doc()` restores them on load. Legacy docs without these fields fall back to `nil`/`"legacy"` for backward compatibility.

src/bridge/store_adapter.rs
ilblackdragon: Fixed in 2bdd5c05 — added `synthesize_orchestrator_doc_from_py()` which reconstructs MemoryDocs from raw `.py` files during `load_knowledge_docs()`. Extracts the version number from the filename (`v{N}.py`) and creates a doc with the correct title, tag, and version metadata so `load_orchestrator_from_docs()` finds it on restart.

serrrfirat:
I think the new `.py` orchestrator recovery path still misses the restarted-runtime case because the synthesized doc is created with `project_id = nil`, but the later lookup path still filters by the real project id.

Concrete path:
- `synthesize_orchestrator_doc_from_py()` creates a `MemoryDoc` with `project_id: ProjectId(uuid::Uuid::nil())`
- `Store::list_shared_memory_docs(project_id)` delegates to `list_memory_docs(project_id, shared_owner_id())`
- `HybridStore::list_memory_docs()` then filters with `doc.project_id == project_id && doc.user_id == user_id`
- `LoopEngine::resolve_orchestrator()` loads shared orchestrators via `list_shared_memory_docs(self.thread.project_id)`

So after restart, the synthesized runtime orchestrator docs still will not be returned for the real project, and the engine appears to fall back to the built-in orchestrator instead of the persisted `.py` version.

If the intent is for shared orchestrator docs to survive restart, I think this needs either:
1. preserving/deriving the real project id when synthesizing the doc, or
2. a corresponding store-query change so shared orchestrator docs are not filtered out by the nil `project_id`.

A restart regression test around the `.py` synthesis + reload path would also help lock this in.

src/tools/builtin/memory.rs
```diff
@@ -52,12 +52,16 @@ impl WorkspaceResolver for FixedWorkspaceResolver {
...
-    matches!(
-        path,
-        "orchestrator:main" | "prompt:codeact_preamble" | "orchestrator:failures"
-    ) || path.starts_with("orchestrator:")
+    // Logical engine aliases
+    path.starts_with("orchestrator:")
         || path.starts_with("prompt:")
+        // Physical workspace paths (match the paths in store_adapter::doc_workspace_path)
+        || path.starts_with("engine/orchestrator/")
+        || path == "engine/orchestrator"
```
serrrfirat: **Medium Severity — Path normalization bypass in orchestrator path guard**

`is_protected_orchestrator_path` uses raw `starts_with("engine/orchestrator/")` on the user-supplied `target` parameter. Paths with dot-components or double slashes bypass this check while resolving to the same filesystem location:

- `engine/./orchestrator/v3.py` → does NOT start with `engine/orchestrator/` → **bypasses check**
- `engine//orchestrator/v3.py` → same bypass
- OS normalizes both to `engine/orchestrator/v3.py` on disk

On filesystem-backed workspaces, a prompt-injected LLM could write orchestrator code even with `ORCHESTRATOR_SELF_MODIFY=false`, because `execute()` uses the same unguarded check (line ~342).

**Suggested fix — normalize before checking:**

```rust
fn normalize_path(path: &str) -> String {
    path.split('/')
        .filter(|s| !s.is_empty() && *s != ".")
        .collect::<Vec<_>>()
        .join("/")
}

fn is_protected_orchestrator_path(path: &str) -> bool {
    let norm = normalize_path(path);
    norm.starts_with("orchestrator:")
        || norm.starts_with("prompt:")
        || norm.starts_with("engine/orchestrator/")
        || norm == "engine/orchestrator"
}
```

Note: `..` components should also be collapsed if the workspace doesn't reject them.

src/tools/builtin/memory.rs
```diff
@@ -279,6 +283,24 @@ impl Tool for MemoryWriteTool {
...
+        // paths (orchestrator code, prompt overlays) requires human approval.
+        // Uses UnlessAutoApproved (not Always) so the v2 engine gate system
+        // can pause for approval rather than hard-denying the call.
+        let target = params.get("target").and_then(|v| v.as_str()).unwrap_or("");
+        if is_protected_orchestrator_path(target) {
+            let self_modify_enabled = std::env::var("ORCHESTRATOR_SELF_MODIFY")
+                .map(|v| v == "true" || v == "1")
+                .unwrap_or(false);
+            if self_modify_enabled {
+                return ApprovalRequirement::UnlessAutoApproved;
```
serrrfirat: **Medium Severity — PR description says `Always` but code returns `UnlessAutoApproved`**

The PR summary states:

> **Layer A** — `memory_write` tool returns `ApprovalRequirement::Always` for protected orchestrator paths

But the code returns `UnlessAutoApproved` (line 297). These have different semantics:

- `Always` → approval required every time, even when session auto-approve is enabled
- `UnlessAutoApproved` → can be silently bypassed when a session has auto-approve on

The code comment at line 289 explains the intent well ("so the v2 engine gate system can pause for approval rather than hard-denying the call"), but the PR description is misleading for reviewers.

**Suggested fix:** Either update the PR description to say `UnlessAutoApproved`, or change to `Always` if the intent is to never skip approval for orchestrator self-modification.

src/bridge/store_adapter.rs
```diff
@@ -624,11 +627,67 @@ fn is_orchestrator_code_path(path: &str) -> bool {
...
     doc.title.starts_with("orchestrator:") || doc.title.starts_with("prompt:")
 }
 
+/// Validate orchestrator content before persisting.
+///
+/// Checks Python syntax so a broken patch doesn't consume failure-budget slots
+/// (3 failures trigger auto-rollback). Runtime sandbox (Monty) handles all
+/// security enforcement — no blocklist here.
+fn validate_orchestrator_content(doc: &MemoryDoc) -> Result<(), EngineError> {
+    if doc.title.starts_with("orchestrator:")
```
serrrfirat, line 952: **Medium Severity — Missing unit tests for new store adapter functions**

Four new functions in this file have no dedicated unit tests:

1. **`validate_orchestrator_content()`** (line 677) — should test: valid Python passes, broken Python rejected, failure-tracker title (`orchestrator:failures`) is exempt, oversized content handling
2. **`synthesize_orchestrator_doc_from_py()`** (line 637) — should test: valid `v3.py` path, non-orchestrator prefix returns `None`, non-numeric version (`vfoo.py`) returns `None`, edge case `v0.py`
3. **Content hash stamping** (~line 800) — should verify hash is present and correct on saved protected docs, and absent on non-protected docs
4. **Frontmatter roundtrip with new fields** — `project_id` and `user_id` were added to serialization (line 791) and deserialization (line 897). A roundtrip test should verify these survive `serialize → deserialize`

Per project rules (`.claude/rules/review-discipline.md`): "Regression test with every fix."

src/tools/builtin/memory.rs
```diff
@@ -279,6 +283,24 @@ impl Tool for MemoryWriteTool {
         })
     }
 
+    fn requires_approval(&self, params: &serde_json::Value) -> ApprovalRequirement {
```
serrrfirat, line 365: **Medium Severity — No unit test for `requires_approval` approval gate**

This new method has three meaningful code paths, none of which are tested:

| Target | `ORCHESTRATOR_SELF_MODIFY` | Expected result |
|--------|---------------------------|----------------|
| `"orchestrator:main"` | `"true"` | `UnlessAutoApproved` |
| `"engine/orchestrator/v3.py"` | unset / `"false"` | `Never` (defers to `execute()` block) |
| `"daily_log"` | any | `Never` |

**Suggested test:**

```rust
#[test]
fn requires_approval_gates_orchestrator_paths() {
    let tool = MemoryWriteTool::new(/* ... */);

    // Non-orchestrator path — never needs approval
    let params = serde_json::json!({"target": "daily_log"});
    assert!(matches!(tool.requires_approval(&params), ApprovalRequirement::Never));

    // Orchestrator path, self-modify disabled — no gate (execute blocks)
    std::env::remove_var("ORCHESTRATOR_SELF_MODIFY");
    let params = serde_json::json!({"target": "orchestrator:main"});
    assert!(matches!(tool.requires_approval(&params), ApprovalRequirement::Never));

    // Orchestrator path, self-modify enabled — requires approval
    std::env::set_var("ORCHESTRATOR_SELF_MODIFY", "true");
    assert!(matches!(tool.requires_approval(&params), ApprovalRequirement::UnlessAutoApproved));
    std::env::remove_var("ORCHESTRATOR_SELF_MODIFY");
}
```

(Note: env var mutation in tests needs serialization via `#[serial]` or `temp_env` crate.)

ilblackdragon:
## Code Review

**Verdict: Mergeable after spec/code mismatch + rehydration hole are fixed**

399/-44 across 7 files. Adds (a) approval gate + Python syntax pre-validation for orchestrator self-modification (`ORCHESTRATOR_SELF_MODIFY=true`), (b) SHA-256 content-hash stamping on protected docs, (c) `.py` reconstitution on restart, (d) frontmatter `project_id`/`user_id` round-tripping, (e) removal of dead `stop_thread_system`/`inject_message_system` ownership-bypass methods, and (f) 6 new Monty sandbox negative tests.

### Scope clarification
The PR title is misleading. Does **not** touch `src/orchestrator/` (sandbox HTTP API), `src/sandbox/` (Docker), network proxy allowlist, container isolation, credential injection, or per-job tokens. "Orchestrator" here = engine v2 self-modifying agent loop; "Monty sandbox" = the in-process Python interpreter, not Docker. Suggest renaming to "engine v2 self-modify gates + Monty negative tests" before merge.

### Do hardening measures close claimed gaps? Mostly yes, with caveats.

**🚨 Verdict mismatch** — `MemoryWriteTool::requires_approval` returns `UnlessAutoApproved` (`src/tools/builtin/memory.rs:296`), but the PR description claims `ApprovalRequirement::Always`. A session that auto-approves `memory_write` (or `auto_approve_all`) silently bypasses the gate. For self-modification of the agent loop, **`Always` is the right choice** — fix the code. Also, `requires_approval` reads `ORCHESTRATOR_SELF_MODIFY` via `std::env::var` on every call (`memory.rs:300`); cache at startup into `Config` to remove TOCTOU on env mutation.

**🚨 `synthesize_orchestrator_doc_from_py` is a rehydration trust hole** (`src/bridge/store_adapter.rs:635-668`). Mints a fresh `DocId` and `user_id = shared_owner_id()` from any `engine/orchestrator/v{N}.py` file found on disk. Anyone with workspace write (or a path-traversal in another tool) can drop a `v999.py` and get it loaded as a privileged orchestrator doc on next restart. The new content-hash stamping happens only on `set_doc`, not on this restart path, so persisted-from-disk patches have no audit hash. **This is the most serious issue in the PR.** Should: (a) verify file hash against a stored manifest of approved versions, (b) re-run `validate_python_syntax` on load, (c) refuse versions whose hash isn't recorded.

**Syntax validation** (`crates/ironclaw_engine/src/executor/scripting.rs:65`) is parse-only — `while True: pass` or imports of privileged builtins still pass. The PR comment honestly notes "no blocklist here" — fine as a scope limit, just be aware.

**Sandbox tests:** `sandbox_handles_deep_recursion` accepts either `Ok` or `Err` (`scripting.rs:2390-2401`), so a silent `Ok` with empty answer passes. Tighten to require non-empty `blocked:` answer or `Err`.

**Dead code removal (M2):** confirmed zero callers; clean win. Removing the `_system` variants closes a privilege-escalation footgun. **Add a regression test** asserting `stop_thread`/`inject_message` *do* check ownership, otherwise the bypass can re-enter via copy-paste.

**Child lease budget doc** (`crates/ironclaw_engine/src/capability/lease.rs:187-195`): documents an intentional weakening — combined parent+child usage can exceed `max_uses`. This is a real budget-bypass; expose `derive_child_leases_strict` for callers (orchestrator self-modify path) where global budgets matter.

### Sandbox escape gaps (Monty)
- `sandbox_context_variable_is_not_mutable` proves Python-side mutations don't propagate, but only because Monty is given a copy. **No tests** for: `__builtins__` mutation, `eval`/`exec` of attacker-supplied strings, `getattr` chain to reach `sys.modules`, `FINAL` shadowing, PyO3-style attribute walks.
- Tool-name injection test only covers `globals().get(name)`; doesn't probe `__import__`, attribute access via captured tool closures.
- The depth-limit test mutates `thread.config.depth` directly — verify the sandboxed code path actually re-reads thread depth on each `rlm_query` rather than a snapshot at executor entry, otherwise the test passes for the wrong reason.

### Conventions
- New `EngineError::InvalidInput` variant and `pub use validate_python_syntax` follow the crate's style.
- No `.unwrap()`/`.expect()` in production. The `.map(|m| m.insert(...))` at `src/bridge/store_adapter.rs:1259` discards the result silently — `let _` would be clearer.
- No new `info!`/`warn!`. Compliant.

### Test coverage for security boundaries
Six new tests, all in `crates/ironclaw_engine/src/executor/scripting.rs:2189-2403`. **Gaps:**
1. No test for `MemoryWriteTool::requires_approval` × `ORCHESTRATOR_SELF_MODIFY=true/false` × protected path matrix.
2. No test for `validate_orchestrator_content` end-to-end via `HybridStore::set_doc`.
3. No test for `synthesize_orchestrator_doc_from_py` path traversal / version parsing (`v-1.py`, `v9999999999999999999.py`, `v.py`).
4. No test that content-hash stamping survives reload.
5. No regression test for the M2 dead-code removal (ownership checks on `stop_thread`/`inject_message`).

### Top suggestions
1. **Resolve the `Always` vs `UnlessAutoApproved` mismatch** at `src/tools/builtin/memory.rs:296`. Fix the code, not the description.
2. **Fix `synthesize_orchestrator_doc_from_py` rehydration trust hole** at `src/bridge/store_adapter.rs:635-668`.
3. **Cache `ORCHESTRATOR_SELF_MODIFY` at startup** instead of reading env on every approval check.
4. **Tighten `sandbox_handles_deep_recursion`** to require non-empty `blocked:` answer or `Err`.
5. **Add Monty escape tests** for `__builtins__` mutation, `eval`/`exec`, `getattr` chain, `FINAL` shadowing.
6. **Strict child-lease variant** for orchestrator self-modify path.
7. **Rename PR title** before merge.

serrrfirat:
**High Severity**

The new orchestrator rehydration path in `src/bridge/store_adapter.rs` synthesizes persisted `.py` docs with `project_id = nil`.

Later loads filter shared memory docs by the real project ID, so after a restart the persisted orchestrator versions may no longer be returned to `load_orchestrator_from_docs()`. The practical effect is that the engine can fall back to the compiled-in orchestrator instead of the persisted runtime version.

Suggested fix: preserve the original project ID when persisting/rehydrating these `.py` docs, or restore it from durable metadata. A restart regression test would make this failure mode obvious.

---

**Medium Severity**

The new `content_hash` audit metadata is not durable for orchestrator `.py` docs. The hash is stamped onto the in-memory `MemoryDoc`, but the persistence path writes raw Python and the load path reconstructs the doc without restoring that metadata.

Suggested fix: persist metadata alongside the `.py` artifact or serialize orchestrator docs in a format that round-trips the audit fields.

src/tools/builtin/memory.rs
```diff
@@ -52,12 +52,16 @@ impl WorkspaceResolver for FixedWorkspaceResolver {
...
 fn is_protected_orchestrator_path(path: &str) -> bool {
-    matches!(
-        path,
-        "orchestrator:main" | "prompt:codeact_preamble" | "orchestrator:failures"
-    ) || path.starts_with("orchestrator:")
+    // Logical engine aliases
+    path.starts_with("orchestrator:")
         || path.starts_with("prompt:")
+        // Physical workspace paths (match the paths in store_adapter::doc_workspace_path)
+        || path.starts_with("engine/orchestrator/")
```
serrrfirat: **Critical Severity — Path traversal bypass in `is_protected_orchestrator_path()`**

`is_protected_orchestrator_path()` performs raw `starts_with("engine/orchestrator/")` on user-supplied `target` without any path normalization. An LLM-authored payload can bypass this guard using:
- `engine/./orchestrator/v3.py`
- `engine//orchestrator/v3.py`
- `engine/knowledge/../orchestrator/v3.py`

None of these match `starts_with("engine/orchestrator/")` but all resolve to the same workspace location. This bypasses both the approval gate (`requires_approval`) AND the hard-deny when `ORCHESTRATOR_SELF_MODIFY=false`.

**Suggested fix:** Add a `normalize_path()` helper that collapses `.`, `..`, double slashes, and trailing slashes before the `starts_with` check. Apply it in both `is_protected_orchestrator_path()` and in the `execute()` guard.

src/bridge/store_adapter.rs
```diff
@@ -624,11 +627,67 @@ fn is_orchestrator_code_path(path: &str) -> bool {
     path.starts_with(ORCHESTRATOR_PREFIX) && path.ends_with(".py")
 }
```
serrrfirat, line 875: **Critical Severity — Path normalization bypass in `is_orchestrator_code_path()`**

Same path normalization vulnerability as `is_protected_orchestrator_path()` — `is_orchestrator_code_path()` only checks `starts_with(ORCHESTRATOR_PREFIX)` AND `ends_with(".py")` on raw input.

A crafted path like `engine/knowledge/../orchestrator/v99.py` bypasses this check and skips syntax validation while still being written to the orchestrator directory by the workspace layer.

Combined with the `is_protected_orchestrator_path()` bypass, this creates a complete security bypass: write invalid/malicious Python to the orchestrator directory without approval and without syntax validation.

**Suggested fix:** Apply the same `normalize_path()` to inputs before checking `is_orchestrator_code_path()`.

src/bridge/store_adapter.rs
```diff
@@ -1156,11 +1232,37 @@ impl Store for HybridStore {
                         ),
                     });
                 }
+            } else {
+                // Self-modification is enabled — validate content before persisting.
+                // Skip validation for system-internal docs (failure tracker, compiled-in seed).
+                let is_system_internal = doc.title == ORCHESTRATOR_FAILURES_TITLE
+                    || doc
+                        .metadata
```
serrrfirat: **High Severity — Forgeable `is_system_internal` metadata allows privilege escalation**

The `is_system_internal` check relies on `doc.metadata.get("source").is_some_and(|s| s == "compiled_in")` to whitelist system writes. Since `MemoryDoc.metadata` is a `serde_json::Value` that the caller constructs, an LLM-authored tool call could craft a doc with `metadata: {"source": "compiled_in"}` and bypass both the self-modify gate (when disabled) and syntax validation (when enabled). The metadata field is not authenticated or integrity-checked.

**Suggested fix:** Do not trust `metadata.source` for authorization decisions. Use a separate internal-only flag or check the call stack/context (e.g., a `CallerContext` enum that distinguishes system vs. LLM-originated writes).

src/bridge/store_adapter.rs
```diff
@@ -1156,11 +1232,37 @@ impl Store for HybridStore {
...
+                    validate_orchestrator_content(doc)?;
+                }
+            }
+        }
+
+        let mut stamped = doc.clone();
+        // Stamp a content hash for audit trail on all protected docs.
+        if is_protected_orchestrator_doc(doc) {
+            use sha2::{Digest, Sha256};
+            let hash = format!("{:x}", Sha256::digest(doc.content.as_bytes()));
```
serrrfirat, line 1588: **High Severity — Content hash provides false integrity assurance**

The content hash is stamped at write time but never verified on read. An attacker who can modify the workspace file directly (bypassing `save_memory_doc`) can change the `.py` content without invalidating the hash.

**Suggested fix:** Either verify the hash on load in `load_knowledge_docs()` / `synthesize_orchestrator_doc_from_py()`, or document explicitly that the hash is write-time-only audit trail (not integrity enforcement).

src/bridge/store_adapter.rs
```diff
@@ -624,11 +627,67 @@ fn is_orchestrator_code_path(path: &str) -> bool {
...
+        .ok()?;
+
+    Some(MemoryDoc {
+        id: DocId(uuid::Uuid::new_v4()),
+        project_id: ProjectId(uuid::Uuid::nil()),
+        user_id: ironclaw_engine::types::shared_owner_id().to_string(),
+        doc_type: DocType::Note,
+        title: ORCHESTRATOR_MAIN_TITLE.to_string(),
+        content: content.to_string(),
+        source_thread_id: None,
```
serrrfirat, line 909: **High Severity — Rehydration uses `project_id = nil`, docs invisible after restart**

`synthesize_orchestrator_doc_from_py()` uses `project_id: ProjectId(uuid::Uuid::nil())`. If `list_memory_docs()` filters by the real project_id (lines 1276-1284), synthesized orchestrator docs will be invisible to project-scoped queries after restart.

This was reported twice by @serrrfirat (2026-04-06 and 2026-04-09) and remains unaddressed.

**Suggested fix:** Verify that the orchestrator loading path uses `list_shared_memory_docs()` or equivalent that does not filter by project_id. If it does filter, pass the actual project_id through a sideband (e.g., store it in a `.meta.json` alongside the `.py` file).

src/tools/builtin/memory.rs
```diff
@@ -279,6 +283,24 @@ impl Tool for MemoryWriteTool {
...
     }
 
+    fn requires_approval(&self, params: &serde_json::Value) -> ApprovalRequirement {
+        // When orchestrator self-modification is enabled, writing to protected
+        // paths (orchestrator code, prompt overlays) requires human approval.
+        // Uses UnlessAutoApproved (not Always) so the v2 engine gate system
+        // can pause for approval rather than hard-denying the call.
+        let target = params.get("target").and_then(|v| v.as_str()).unwrap_or("");
+        if is_protected_orchestrator_path(target) {
+            let self_modify_enabled = std::env::var("ORCHESTRATOR_SELF_MODIFY")
```
serrrfirat: **High Severity — Environment variable check for security gate is fragile**

`requires_approval()` reads `ORCHESTRATOR_SELF_MODIFY` from the environment on every call. Environment variables are global mutable state and can be modified by any thread. If the process runs untrusted code in-process (which Monty does), a race condition or future Monty escape could flip this env var.

More importantly, the env var check is duplicated in three places (`requires_approval`, `execute` in memory.rs, `save_memory_doc` in store_adapter.rs) with identical but copy-pasted logic. Any future divergence creates a bypass.

**Suggested fix:** Extract the env var check into a single shared function (e.g., `fn is_orchestrator_self_modify_enabled() -> bool`). Consider caching the value at startup rather than reading it per-call.

crates/ironclaw_engine/src/executor/scripting.rs
```diff
@@ -53,6 +53,36 @@ fn default_limits() -> ResourceLimits {
...
+pub fn validate_python_syntax(code: &str) -> Result<(), String> {
+    if code.len() > MAX_ORCHESTRATOR_SOURCE_BYTES {
+        return Err(format!(
+            "orchestrator source too large: {} bytes (limit: {MAX_ORCHESTRATOR_SOURCE_BYTES})",
+            code.len()
+        ));
+    }
+    let code_owned = code.to_string();
+    match std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| {
+        MontyRun::new(code_owned, "validate.py", vec![])
```
serrrfirat, line 100: **Medium Severity — Syntax-only validation may provide false security confidence**

`validate_python_syntax()` checks syntax only — it does not check semantic validity. Syntactically valid but semantically dangerous patterns (e.g., `exec(compile(...))`, `__import__('os')`) pass validation.

The comment says "Runtime sandbox (Monty) handles all security enforcement — no blocklist here" which is correct, but for a security PR, this should be documented more prominently.

**Suggested fix:** Add a prominent comment explicitly stating the threat model: syntax validation prevents consuming failure-budget slots, NOT preventing malicious code. The sandbox is the security boundary.

src/bridge/store_adapter.rs
```diff
@@ -624,11 +627,67 @@ fn is_orchestrator_code_path(path: &str) -> bool {
...
+/// Validate orchestrator content before persisting.
+///
+/// Checks Python syntax so a broken patch doesn't consume failure-budget slots
+/// (3 failures trigger auto-rollback). Runtime sandbox (Monty) handles all
+/// security enforcement — no blocklist here.
+fn validate_orchestrator_content(doc: &MemoryDoc) -> Result<(), EngineError> {
+    if doc.title.starts_with("orchestrator:")
+        && doc.title != ORCHESTRATOR_FAILURES_TITLE
+        && let Err(reason) = ironclaw_engine::executor::validate_python_syntax(&doc.content)
+    {
```
serrrfirat, line 955: **Medium Severity — `validate_orchestrator_content()` skips `"prompt:"` titled docs**

`is_protected_orchestrator_doc()` checks both `"orchestrator:"` and `"prompt:"` titles, but `validate_orchestrator_content()` at line 677 only validates `"orchestrator:"` docs. A malicious prompt overlay (title `"prompt:codeact_preamble"`) is stamped with a content hash but never syntax-validated.

If the engine ever supports Python prompt overlays, this gap would become a vulnerability.

**Suggested fix:** Either validate prompt overlays too (if they can contain code), or add a comment explicitly documenting that prompt overlays are assumed to be non-executable content.

src/bridge/store_adapter.rs
```diff
@@ -1156,11 +1232,37 @@ impl Store for HybridStore {
                         ),
                     });
                 }
+            } else {
+                // Self-modification is enabled — validate content before persisting.
```
serrrfirat: **Medium Severity — DRY violation in `is_system_internal` creates drift risk**

The `is_system_internal` check is copy-pasted between the `allow=false` and `allow=true` branches with identical logic. Any future change to the system-internal definition must be updated in two places, and divergence would create a security bypass.

**Suggested fix:** Extract into a helper: `fn is_system_internal_doc(doc: &MemoryDoc) -> bool`.

src/tools/builtin/memory.rs
serrrfirat, line 365: **Medium Severity — Zero test coverage for `requires_approval()` security gate**

The `requires_approval()` method has three meaningful code paths (non-orchestrator target, orchestrator with self-modify disabled, orchestrator with self-modify enabled), none of which are tested.

**Suggested fix:** Add unit tests for all three paths. Use `temp_env` or `#[serial]` for env var mutation safety.

src/bridge/store_adapter.rs
serrrfirat, line 952: **Medium Severity — Four new security-relevant functions have zero test coverage**

No tests for `validate_orchestrator_content()`, `synthesize_orchestrator_doc_from_py()`, content hash stamping, or the frontmatter roundtrip with new `project_id`/`user_id` fields.

**Suggested fix:** Add unit tests: valid/invalid Python syntax, failure-tracker exemption, `.py` synthesis with valid/invalid paths, hash presence on protected docs, frontmatter roundtrip.

crates/ironclaw_engine/src/executor/scripting.rs
```diff
@@ -2154,4 +2184,221 @@ FINAL(str(x))
...
+            }
+            Err(_) => {
+                // VM killed by resource limits — acceptable.
+            }
+        }
+    }
+
+    /// validate_python_syntax rejects broken code and accepts valid code.
+    #[test]
+    fn validate_syntax_rejects_broken_code() {
```
serrrfirat, line 2703: **Medium Severity — `validate_syntax_rejects_broken_code` test is insufficient**

Only tests two cases (broken def, valid assignment). Does not test: the 256KB size limit, empty input, unicode input, code that triggers a parser panic, or the error message format.

**Suggested fix:** Expand test coverage: test size limit rejection, empty string (should pass — valid Python), and verify the specific error message returned.

serrrfirat (dismissed):
**Request changes — 2 Critical path traversal bypasses and 4 High-severity findings must be addressed before merge.**

Reviewed all 7 changed files in depth. The security hardening direction is correct, but the current implementation has exploitable gaps:

**Critical (2):**
1. **Path traversal in `is_protected_orchestrator_path()`** (`memory.rs:63`) — raw `starts_with` on user input bypassed via `engine/./orchestrator/`, `engine//orchestrator/`, or `engine/knowledge/../orchestrator/`. Defeats both approval gate and `ORCHESTRATOR_SELF_MODIFY=false`.
2. **Same bypass in `is_orchestrator_code_path()`** (`store_adapter.rs:628`) — combined with #1, allows writing unvalidated Python to orchestrator directory without approval.

**High (4):**
3. **Forgeable `is_system_internal` metadata** (`store_adapter.rs:1240`) — LLM can craft `metadata.source = "compiled_in"` to bypass self-modify gate. Metadata is not authenticated.
4. **Content hash never verified on read** (`store_adapter.rs:1254`) — false integrity assurance if workspace files modified directly.
5. **Rehydration `project_id = nil`** (`store_adapter.rs:655`) — synthesized `.py` docs invisible to project-scoped queries after restart. Reported twice by @serrrfirat.
6. **Env var security gate duplicated in 3 places** (`memory.rs:293`) — `ORCHESTRATOR_SELF_MODIFY` read from global mutable state per-call, copy-pasted logic risks divergence.

**Medium (6):** 6 additional findings posted inline covering validation scope, test gaps, and DRY violations.

The sandbox tests and Monty integration are well-designed. The `stop_thread_system`/`inject_message_system` removal is clean. Fix the Critical/High items and this will be a strong security PR.

src/bridge/store_adapter.rs
```diff
@@ -624,11 +627,67 @@ fn is_orchestrator_code_path(path: &str) -> bool {
...
+        project_id: ProjectId(uuid::Uuid::nil()),
+        user_id: ironclaw_engine::types::shared_owner_id().to_string(),
+        doc_type: DocType::Note,
+        title: ORCHESTRATOR_MAIN_TITLE.to_string(),
+        content: content.to_string(),
+        source_thread_id: None,
+        tags: vec![ORCHESTRATOR_CODE_TAG.to_string()],
+        metadata: serde_json::json!({
+            "version": version,
+            "source": "persisted_py",
```
serrrfirat, line 913: **Medium Severity**

Raw persisted `v0.py` docs are synthesized with metadata `source: "persisted_py"`, but the self-mod disabled save exemption only allows `source: "compiled_in"`. After restart, if the compiled default changes, `seed_orchestrator_v0` can fail to update v0 while self-modification is disabled.

Please preserve/restore `source: "compiled_in"` for synthesized `v0.py`, or allow shared-owner `orchestrator:main` version `0` as system-internal.

src/tools/builtin/memory.rs
serrrfirat: **High Severity**

Protected orchestrator writes return `ApprovalRequirement::UnlessAutoApproved`, which can be bypassed by session “always approve” state. That weakens the intended human approval gate for self-modification.

Please add a path-specific approval mode that always prompts and cannot be session-auto-approved, or make the effect adapter support explicit approval gates for `Always` instead of hard-denying it.

src/tools/builtin/memory.rs
serrrfirat: **High Severity**

Direct memory_write writes to protected engine/orchestrator/*.py paths only check the env flag and approval, then write/patch via Workspace without calling validate_python_syntax. That bypasses the new save_memory_doc syntax validation and can persist a broken runtime orchestrator.

Please validate the final content in MemoryWriteTool before write, append, or patch on protected .py paths, or route protected orchestrator writes through Store::save_memory_doc so the same validation runs for both write paths.

src/bridge/store_adapter.rs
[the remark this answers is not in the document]
ilblackdragon: Addressed in 2bdd5c05 (round 1).

src/bridge/store_adapter.rs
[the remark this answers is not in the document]
ilblackdragon: Addressed in 2bdd5c05 (round 1).

crates/ironclaw_engine/src/executor/scripting.rs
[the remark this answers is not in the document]
ilblackdragon: Addressed in 2bdd5c05 (round 1).

crates/ironclaw_engine/src/executor/scripting.rs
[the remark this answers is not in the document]
ilblackdragon: Addressed in 2bdd5c05 (round 1).

src/tools/builtin/memory.rs
ilblackdragon: Fixed in c50d70af — added `normalize_workspace_path()` that strips `.`/empty segments and rejects `../` traversal. Both `is_protected_orchestrator_path` and `is_orchestrator_code_path` normalize before matching. Caller-level tests drive `ToolDispatcher::dispatch()` with all bypass vectors (dot segment, double slash, traversal). See also 37a67f36 which further refines traversal handling in `requires_approval`.

src/bridge/store_adapter.rs
ilblackdragon: Fixed in c50d70af — `is_orchestrator_code_path` now normalizes via `normalize_path()` before matching. Matches both `.system/engine/orchestrator/` and legacy `engine/orchestrator/` prefixes. Dead equality checks removed in 37a67f36.

src/bridge/store_adapter.rs
ilblackdragon: Fixed in c50d70af — replaced metadata-based exemption with a `tokio::task_local!` trusted-write scope (`ironclaw_engine::runtime::with_trusted_internal_writes`). The task-local cannot be set from outside a trusted callsite and does not propagate across `tokio::spawn`. Test in `internal_write.rs` asserts the no-propagation property.

src/bridge/store_adapter.rs
ilblackdragon: Addressed in c50d70af — documented explicitly in `save_memory_doc` that the hash is **write-time audit only**, not a runtime integrity check. The workspace file is the trust boundary; anyone with workspace access can edit files directly.

src/bridge/store_adapter.rs
ilblackdragon: Fixed in c50d70af — overrode `HybridStore::list_shared_memory_docs` to surface physically-global docs (orchestrator, failures, prompt overlay) regardless of stored `project_id`. Further strengthened in 37a67f36: `save_memory_doc` normalizes `project_id` to nil for global docs at write time so they surface immediately from any project query, not just after restart.

src/tools/builtin/memory.rs
ilblackdragon: Fixed in c50d70af — centralized in `ironclaw_engine::runtime::self_modify_enabled()`: a process-wide `OnceLock<bool>` seeded once on first read. Tool, store, engine loop, and self-improvement mission all share the same snapshot. `SelfModifyTestGuard` provides serialized dev-build override (compiled out of release builds).

crates/ironclaw_engine/src/executor/scripting.rs
ilblackdragon: Addressed in 37a67f36 — expanded the rustdoc on `validate_python_syntax` with a **Threat model** section: syntax validation prevents auto-rollback DoS, NOT dangerous code execution. All security enforcement happens in the Monty sandbox at runtime.

src/bridge/store_adapter.rs
ilblackdragon: Addressed in 37a67f36 — added a detailed comment explaining that `prompt:*` docs are markdown text, not Python code. Validating them as Python would reject every prompt overlay. If the engine ever supports Python prompt overlays, the validator must be updated.

src/bridge/store_adapter.rs
ilblackdragon: Fixed in c50d70af — the duplicated `is_system_internal` check was entirely removed. The gate now uses the `tokio::task_local!` trusted-write scope instead of metadata-based exemption, so there is only one code path.

src/tools/builtin/memory.rs
ilblackdragon: Fixed in c50d70af — 7 unit tests covering all branches (protected+enabled, protected+disabled, physical path, dot bypass, traversal, unprotected, missing target). Updated in 37a67f36 to cover the `Always` variant and traversal-returns-Never semantics.

src/bridge/store_adapter.rs
ilblackdragon: Fixed in c50d70af — 23 unit tests in `store_adapter::helper_tests` covering `validate_orchestrator_content`, `synthesize_orchestrator_doc_from_py`, hash stamping, path normalization, predicates, and `list_shared_memory_docs` override. Plus 3 integration tests for orchestrator round-trip, knowledge doc round-trip, and invalid-Python rejection.

crates/ironclaw_engine/src/executor/scripting.rs
ilblackdragon: Expanded in 37a67f36 — now covers: empty input (valid), unicode identifiers (valid), oversized input (rejected with size-cap error), and error message format (`syntax error` prefix).

src/tools/builtin/memory.rs
ilblackdragon: PR description updated. Code now uses `ApprovalRequirement::Always` (as of 37a67f36) with the effect bridge mapping `Always` to `GatePaused(Approval { allow_always: false })` instead of `LeaseDenied`. So the protected self-modify path always pauses for human approval and cannot be session-auto-approved.

src/bridge/store_adapter.rs
ilblackdragon: Already resolved by c50d70af — the `metadata.source` field is no longer load-bearing for the gate. `seed_orchestrator_v0` wraps `save_memory_doc` in `with_trusted_internal_writes`, which bypasses the gate via task-local scope regardless of metadata content. The source field is kept as informational-only.

src/tools/builtin/memory.rs
ilblackdragon: Fixed in 37a67f36 — `memory_write::requires_approval` now returns `ApprovalRequirement::Always` (not `UnlessAutoApproved`) for protected paths. The effect bridge was updated to map `Always` to `GatePaused(Approval { allow_always: false })` instead of `LeaseDenied`, so the gate always pauses for explicit human approval and cannot be session-auto-approved. Regression tests assert the `allow_always: false` property end-to-end.

src/tools/builtin/memory.rs
ilblackdragon: Fixed in 37a67f36 — `MemoryWriteTool::execute()` now validates Python syntax via `validate_python_syntax` for protected `.py` orchestrator paths before calling `workspace.write/append/patch`. All three write modes are covered: write validates content directly, append reads existing + concat, patch simulates the replacement. Returns `InvalidParameters` on syntax error.

src/tools/builtin/memory.rs
```diff
@@ -282,6 +346,19 @@ impl Tool for MemoryWriteTool {
...
+        // paths (orchestrator code, prompt overlays) requires human approval.
+        // Uses UnlessAutoApproved (not Always) so the v2 engine gate system
+        // can pause for approval rather than hard-denying the call. When
+        // disabled, execute() returns NotAuthorized before any write.
+        let target = params.get("target").and_then(|v| v.as_str()).unwrap_or("");
+        if is_protected_orchestrator_path(target) && self_modify_enabled() {
+            return ApprovalRequirement::UnlessAutoApproved;
+        }
+        ApprovalRequirement::Never
+    }
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in 37a67f36 — `requires_approval` now normalizes the path first; if normalization fails (traversal), returns `Never` so `execute()` rejects it immediately as `InvalidParameters` without a spurious approval gate.

src/bridge/store_adapter.rs
```diff
@@ -1392,6 +1548,41 @@ impl Store for HybridStore {
...
+            .values()
+            .filter(|doc| {
+                let owner_shared = ironclaw_engine::types::is_shared_owner(&doc.user_id);
+                if !owner_shared {
+                    return false;
+                }
+                // Match the project filter, OR surface global docs (those
+                // saved with `ProjectId::nil()`) for every project.
+                doc.project_id == project_id
+                    || (doc.project_id.0.is_nil() && is_globally_shared(doc))
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in 37a67f36 — `save_memory_doc` normalizes `project_id` to `nil` for globally-shared docs at write time (not just on rehydration). Fresh seeds from `MissionManager` immediately surface from any project query.

src/bridge/store_adapter.rs
```diff
@@ -825,15 +838,123 @@ fn doc_workspace_path(doc: &MemoryDoc) -> String {
...
-    path.starts_with(ORCHESTRATOR_PREFIX) && path.ends_with(".py")
+    let Some(canonical) = normalize_path(path) else {
+        return false;
+    };
+    if !canonical.ends_with(".py") {
+        return false;
+    }
+    canonical == ORCHESTRATOR_PREFIX
+        || canonical.starts_with(&format!("{ORCHESTRATOR_PREFIX}/"))
+        || canonical == LEGACY_ORCHESTRATOR_PREFIX
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in 37a67f36 — removed the two unreachable equality checks (`canonical == ORCHESTRATOR_PREFIX` / `== LEGACY_ORCHESTRATOR_PREFIX`). Only the `starts_with` checks remain.

src/tools/builtin/memory.rs
ilblackdragon: Addressed across c50d70af (round 2) and 37a67f36 (round 3). See commit messages for details.

src/tools/builtin/memory.rs
ilblackdragon: Addressed across c50d70af (round 2) and 37a67f36 (round 3). See commit messages for details.

src/bridge/store_adapter.rs
ilblackdragon: Addressed across c50d70af (round 2) and 37a67f36 (round 3). See commit messages for details.

src/bridge/store_adapter.rs
ilblackdragon: Addressed across c50d70af (round 2) and 37a67f36 (round 3). See commit messages for details.

src/tools/builtin/memory.rs
ilblackdragon: Addressed across c50d70af (round 2) and 37a67f36 (round 3). See commit messages for details.

src/tools/builtin/memory.rs
ilblackdragon: Addressed across c50d70af (round 2) and 37a67f36 (round 3). See commit messages for details.

src/tools/builtin/memory.rs
```diff
@@ -282,6 +362,34 @@ impl Tool for MemoryWriteTool {
...
+        if target.starts_with("orchestrator:") || target.starts_with("prompt:") {
+            return ApprovalRequirement::Always;
+        }
+        // Physical paths: normalize first. Traversal attempts fail normalization
+        // — return Never so execute() rejects immediately as InvalidParameters
+        // rather than triggering a spurious approval gate.
+        if let Some(canonical) = normalize_workspace_path(target) {
+            let protected = canonical.starts_with(".system/engine/orchestrator")
+                || canonical.starts_with("engine/orchestrator");
+            if protected {
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in eff8a22b — added path-boundary checks (equality + trailing-slash `starts_with`) so `orchestrator_backup/` and similar won't trigger the gate.

src/bridge/effect_adapter.rs
```diff
@@ -1849,6 +1874,167 @@ mod tests {
...
 
+    /// End-to-end gate verification for the real `MemoryWriteTool`.
+    ///
+    /// PR #1958 reviewer-flagged regression: the original code returned
+    /// `ApprovalRequirement::Always` for protected orchestrator targets,
+    /// which `EffectBridgeAdapter::execute_action` maps to `LeaseDenied`
+    /// (a hard refusal). The fix changed it to `UnlessAutoApproved`, which
+    /// must produce a resumable `GatePaused(Approval)`. This test asserts
+    /// the full path: `requires_approval` → adapter → gate, with the
+    /// real tool wired into a real registry.
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in eff8a22b — updated the docstring to reflect the current `Always → GatePaused(Approval { allow_always: false })` behavior.

src/tools/builtin/memory.rs
```diff
@@ -382,6 +500,52 @@ impl Tool for MemoryWriteTool {
...
+                let existing = workspace
+                    .read(&resolved_path)
+                    .await
+                    .map(|d| d.content)
+                    .unwrap_or_default();
+                if replace_all_flag {
+                    existing.replace(old_str, new_str)
+                } else {
+                    existing.replacen(old_str, new_str, 1)
+                }
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in 7a3bf5be — moved the syntax validation block after patch-mode parameter validation (empty `old_string` / missing `new_string` are rejected first). Also deduplicated the matching logic: `requires_approval` now delegates to `is_protected_orchestrator_path` instead of duplicating the prefix checks.

serrrfirat (commented):
Found one issue in the security hardening path:

| Severity | Category | File:Line | Finding | Suggested Fix |
|---|---|---|---|---|
| High | Security / approval bypass | `crates/ironclaw_engine/src/runtime/mission.rs:2348` | The self-improvement structured JSON path still bypasses the new protected-write approval gate. When `ORCHESTRATOR_SELF_MODIFY=true`, `process_self_improvement_output` accepts LLM-authored `prompt_additions`, updates `prompt:codeact_preamble`, and calls `store.save_memory_doc` directly. That store path checks the self-modify flag and validates protected orchestrator code, but it cannot pause for human approval, so it never goes through `MemoryWriteTool::requires_approval` / `EffectBridgeAdapter`'s `Always` gate. A self-improvement mission can therefore mutate a protected prompt overlay without the approval flow this PR is trying to enforce. | Route this path through the same approval machinery, either by requiring the mission to use `memory_write` for prompt overlays or by adding a resumable pending-approval gate before the direct save. Add a caller-level test proving `prompt_additions` are not persisted while approval is pending, even with self-modify enabled. |

I did not run tests; this was a review-only pass over the PR diff and the relevant call paths.

src/bridge/effect_adapter.rs
```diff
@@ -770,13 +770,18 @@ impl EffectBridgeAdapter {
...
-                        ),
-                    });
+                    if !approval_already_granted {
+                        return Err(Self::gate_paused(
+                            "approval",
+                            action_name,
+                            context.current_call_id.as_deref(),
+                            parameters,
+                            ironclaw_engine::ResumeKind::Approval {
+                                allow_always: false,
```
serrrfirat, line 868: **High Severity — `allow_always: false` not enforced server-side**

The PR sets `allow_always: false` in `ResumeKind::Approval` for orchestrator writes so the UI hides the "always approve" button. However, the router (`src/bridge/router.rs` ~line 1516) blindly passes the user-supplied `always` flag to `auto_approve_tool()` without checking `pending.resume_kind.allow_always`.

A malicious API call to the approval endpoint with `always: true` would permanently auto-approve `memory_write` for the session, silently bypassing the per-call approval gate on all subsequent orchestrator writes.

**Suggested fix:** In the router where `GateResolution::Approved { always }` is constructed, clamp: `let always = always && matches!(pending.resume_kind, ResumeKind::Approval { allow_always: true })`. Add a caller-level integration test.

crates/ironclaw_engine/src/executor/scripting.rs
```diff
@@ -54,6 +54,44 @@ fn default_limits() -> ResourceLimits {
...
+pub fn validate_python_syntax(code: &str) -> Result<(), String> {
+    if code.len() > MAX_ORCHESTRATOR_SOURCE_BYTES {
+        return Err(format!(
+            "orchestrator source too large: {} bytes (limit: {MAX_ORCHESTRATOR_SOURCE_BYTES})",
+            code.len()
+        ));
+    }
+    let code_owned = code.to_string();
+    match std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| {
+        MontyRun::new(code_owned, "validate.py", vec![])
```
serrrfirat, line 100: **Medium Severity — `validate_python_syntax` creates full VM, not just parser**

`validate_python_syntax` creates a full `MontyRun` for parsing. `MontyRun::new()` initializes the entire VM (allocator, runtime state, etc.) and may execute partial code before hitting a syntax error. This is heavier than a parse-only check. The 256KB size cap mitigates but doesn't eliminate resource risk.

**Suggested fix:** If Monty exposes a parse-only API, use it instead. If not, document that `MontyRun::new()` only parses (does not execute) so reviewers can verify. Consider a tighter size limit or a timeout.

src/bridge/store_adapter.rs
```diff
@@ -1341,22 +1487,17 @@ impl Store for HybridStore {
...
-                // Allow system-internal writes (v0 seeding, failure tracking) but
-                // block LLM-authored patches (version > 0, non-meta titles).
-                let is_system_internal = doc.title == "orchestrator:failures"
-                    || doc
-                        .metadata
-                        .get("source")
-                        .and_then(|v| v.as_str())
-                        .is_some_and(|s| s == "compiled_in");
-                if !is_system_internal {
+            let trusted = ironclaw_engine::runtime::is_trusted_internal_write_active()
```
serrrfirat: **Medium Severity — `orchestrator:failures` title exemption bypasses gate from any code path**

The line `let trusted = is_trusted_internal_write_active() || doc.title == ORCHESTRATOR_FAILURES_TITLE;` means any code path that calls `save_memory_doc` with title `orchestrator:failures` will bypass the self-modify gate — including potentially LLM-reachable paths through the self-improvement mission or engine code that calls the store directly.

Impact is limited to corrupting the failure count (not executable code), but it could trigger or prevent auto-rollback.

**Suggested fix:** Move the exemption inside `with_trusted_internal_writes` scope at the call sites (`record_orchestrator_failure`, `reset_orchestrator_failures`) instead of hardcoding it in the gate predicate.

src/bridge/store_adapter.rs
```diff
@@ -825,15 +838,131 @@ fn doc_workspace_path(doc: &MemoryDoc) -> String {
...
+    }
+    canonical.starts_with(&format!("{ORCHESTRATOR_PREFIX}/"))
+        || canonical.starts_with(&format!("{LEGACY_ORCHESTRATOR_PREFIX}/"))
+}
+
+/// Strip `.` segments and collapse `//`, returning `None` on `..` traversal.
+///
+/// Mirrors `normalize_workspace_path` in `tools::builtin::memory` — kept
+/// local here so the store adapter has no dependency on the tool layer.
+fn normalize_path(path: &str) -> Option<String> {
```
serrrfirat, line 863: **Medium Severity — Duplicate path normalization on security boundary**

Two independent copies of the same path normalization logic exist: `normalize_workspace_path` in `src/tools/builtin/memory.rs` (line 64) and `normalize_path` here in `store_adapter.rs`. If one is updated and the other is not, a bypass path reopens. Both are on the same security boundary.

**Suggested fix:** Extract into a shared utility module (e.g., `crate::util::normalize_workspace_path`), or at minimum add a cross-reference test that asserts both produce identical results for a canonical set of inputs.

src/tools/builtin/memory.rs
```diff
@@ -282,6 +362,35 @@ impl Tool for MemoryWriteTool {
...
+        // session auto-approve cannot silently skip the gate — the effect
+        // bridge maps `Always` to `GatePaused(Approval { allow_always: false })`.
+        // When disabled, execute() returns NotAuthorized before any write.
+        //
+        // Delegates to is_protected_orchestrator_path for consistency, but
+        // excludes traversal attempts (`..`) — those return Never here so
+        // execute() rejects them immediately as InvalidParameters rather
+        // than triggering a spurious approval gate before the rejection.
+        let target = params.get("target").and_then(|v| v.as_str()).unwrap_or("");
+        if !self_modify_enabled() {
```
serrrfirat, line 389: **Medium Severity — Defense-in-depth gap in `requires_approval`**

`requires_approval` returns `Never` when self-modify is disabled, relying solely on `execute()` to reject. This means the effect bridge will not pause the action — it proceeds directly to `execute()`. If `execute()` is ever refactored and the self-modify check is accidentally removed, orchestrator writes slip through without approval.

**Suggested fix:** Consider returning `ApprovalRequirement::Always` when `is_protected_orchestrator_path(target)` regardless of self-modify state, or document why the single `execute()` gate is sufficient with a cross-reference to the exact rejection line.

src/bridge/effect_adapter.rs
ilblackdragon: Fixed in `20168104` (prior commit). The `clamp_always_to_resume_kind` helper in `router.rs` now clamps the caller-supplied `always` flag against `pending.resume_kind.allow_always` server-side. A crafted `always: true` on an `Approval { allow_always: false }` gate is clamped to `false`. Four unit tests cover all three `ResumeKind` variants. The staging merge (`f259ff9f`) preserved this helper over staging's inline match (functionally equivalent, but ours has dedicated tests).

crates/ironclaw_engine/src/executor/scripting.rs
ilblackdragon: Addressed in `20168104`. Expanded the `validate_python_syntax` rustdoc to document that `MontyRun::new()` is a **parse-and-prepare** step — it runs the Ruff-based parser to check syntax, then sets up runtime scaffolding (allocator handles, namespace stubs) without populating the heap or executing module-level statements. The 256 KB `MAX_ORCHESTRATOR_SOURCE_BYTES` cap bounds parser allocation, not execution time. If Monty later exposes a dedicated parse-only entry point, we can switch; for now the resource exposure from the current approach is bounded by the size cap.

src/bridge/store_adapter.rs
ilblackdragon: Fixed in `20168104`. Removed the `doc.title == ORCHESTRATOR_FAILURES_TITLE` shortcut from the store-adapter self-modify gate entirely. The two legitimate callers (`record_orchestrator_failure`, `reset_orchestrator_failures`) now wrap their `save_memory_doc` calls in `with_trusted_internal_writes`. Additionally, untrusted writes to the failures title are now rejected regardless of self-modify state, since no LLM-reachable code path should ever persist the system-internal tracker.

src/bridge/store_adapter.rs
ilblackdragon: Addressed in `20168104`. Added a parity test (`normalize_path_and_normalize_workspace_path_agree_on_canonical_inputs`) that asserts both normalizers produce identical results on a canonical input set including traversal attempts, double slashes, dot segments, and boundary cases. Shared extraction isn't clean across the bridge/tool boundary without introducing a shared crate dependency, so the parity test is the lighter guard against drift on this security boundary.

src/tools/builtin/memory.rs
ilblackdragon: Addressed in `20168104`. Tightened the commentary in `requires_approval` with an explicit cross-reference to the `execute()` rejection site (~line 444) and a load-bearing invariant warning: "Any refactor that moves or weakens that `execute()` rejection must also flip this branch to `Always` so the approval dialog becomes the backstop." This makes the dependency relationship between the two gates explicit in the source, so a future refactor that weakens `execute()` will be forced to update this branch.

PR #1970, opened 2026-04-03
fix(llm): add sanitize_tool_messages to OpenAiCodexProvider

### LINKED ISSUE #1969: OpenAiCodexProvider missing sanitize_tool_messages causes HTTP 400 on orphaned tool results
## Bug

When using the `openai_codex` LLM provider, conversations that accumulate tool call history can fail with:

```
HTTP 400 Bad Request: { "error": { "message": "No tool call found for function call output with call_id call_faChBBNMDDTrfUMSlk3vhlxx.", "type": "invalid_request_error", "param": "input", "code": null } }
```

This happens during both regular chat and routine execution — any conversation with tool calls can trigger it.

## Root Cause

`OpenAiCodexProvider` does not call `sanitize_tool_messages()` before sending messages to the API. Every other provider calls this function:

- `nearai_chat.rs` — lines 465, 530
- `rig_adapter.rs` — lines 677, 737
- `bedrock.rs` — lines 99, 155

But `openai_codex_provider.rs` passes `request.messages` directly into `build_request_body()` (lines 262, 294) without sanitizing.

When the conversation contains an `ActionResult` message whose `call_id` references a tool call that was dropped (e.g., during context compaction or thread resume), the OpenAI Responses API rejects the request.

## Fix

Add `crate::llm::provider::sanitize_tool_messages()` calls to both `complete()` and `complete_with_tools()` in `openai_codex_provider.rs`, matching the pattern used by all other providers.

## Repro

1. Use `LLM_BACKEND=openai_codex`
2. Have a conversation with multiple tool calls that approaches the context compaction threshold
3. The next LLM call after compaction fails with the 400 error

serrrfirat:
## Summary

- Add missing `sanitize_tool_messages()` call to `OpenAiCodexProvider::complete()` and `complete_with_tools()`, matching the pattern used by all other providers (nearai_chat, rig_adapter, bedrock)
- Without this, orphaned tool results (where the corresponding tool call was dropped during context compaction or thread resume) cause HTTP 400: "No tool call found for function call output with call_id ..."
- Add regression test verifying orphaned tool results are rewritten before reaching the API

## Test plan

- [x] `cargo fmt` — clean
- [x] All 25 tests in `llm::openai_codex_provider::tests` pass
- [x] New regression test `test_build_request_sanitizes_orphaned_tool_results` covers the fix
- [ ] Manual test with `LLM_BACKEND=openai_codex` and a long conversation with tool calls

Fixes #1969

### DIFF
--- a/src/llm/openai_codex_provider.rs
+++ b/src/llm/openai_codex_provider.rs
@@ -259,7 +259,9 @@ impl LlmProvider for OpenAiCodexProvider {
     }
 
     async fn complete(&self, request: CompletionRequest) -> Result<CompletionResponse, LlmError> {
-        let body = self.build_request_body(&request.messages, None);
+        let mut messages = request.messages;
+        crate::llm::provider::sanitize_tool_messages(&mut messages);
+        let body = self.build_request_body(&messages, None);
         let parsed = self.send_request(body).await?;
 
         Ok(CompletionResponse {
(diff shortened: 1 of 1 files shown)


serrrfirat:
Closing — retargeting to staging branch.

PR #1975, opened 2026-04-03
fix(bridge): sanitize orphaned tool results in v2 adapter

### LINKED ISSUE #1950: bug(bridge): orphaned tool_result messages in v2 LLM adapter cause Anthropic API error 2013
## Description

The V2 engine's `LlmBridgeAdapter` converts `ThreadMessage` to `ChatMessage` and sends them to the LLM provider without sanitizing orphaned tool_result messages. When a tool_result exists in the thread history without a matching preceding assistant tool_call, Anthropic's API rejects the request with error code 2013.

## Symptoms

- LLM call returns HTTP 400 with error body:
  ```json
  {
    "type": "invalid_request_error",
    "message": "tool_result block(s) that are not immediately after a tool_use block (2013)"
  }
  ```
- The thread becomes stuck — every subsequent LLM call fails with the same error since the orphaned message persists in thread history
- Only affects V2 engine path (`ENGINE_V2=true`); V1 path already calls `sanitize_tool_messages()` before sending

## Steps to reproduce

1. Enable `ENGINE_V2=true` with an Anthropic backend (direct or via proxy)
2. Run a multi-step thread that triggers any of these scenarios:
   - **Thread recovery from checkpoint**: history is restored but an assistant tool_call message was in a different checkpoint segment, leaving its tool_result orphaned
   - **Context compaction**: an assistant message with tool_calls is evicted to save context space, but the corresponding tool_result messages remain
   - **Gate pause/resume**: approval interruption causes partial message sequences
3. The next LLM call fails with error 2013

## Root cause

```rust
// src/bridge/llm_adapter.rs, line ~49
let chat_messages: Vec<ChatMessage> = messages.iter().map(thread_msg_to_chat).collect();
// Missing: crate::llm::sanitize_tool_messages(&mut chat_messages);
```

The V1 agentic loop calls `sanitize_tool_messages()` which rewrites orphaned tool_result messages to User role. The V2 `LlmBridgeAdapter::complete()` skips this step entirely.

## Fix

Add `crate::llm::sanitize_tool_messages(&mut chat_messages)` after the message conversion in `LlmBridgeAdapter::complete()`, matching V1 behavior.

## Acceptance criteria

- [ ] Orphaned tool_result messages are rewritten to User role before sending to the LLM
- [ ] Thread execution continues normally after checkpoint recovery / compaction / gate resume
- [ ] Regression tests verify both orphaned and non-orphaned tool_result handling

Branch: `v2-architecture`

serrrfirat:
## Summary
- sanitize V2 bridge chat history before sending it to LLM providers
- reuse the shared `sanitize_tool_messages` helper from the LLM module
- add regression tests covering both orphaned and valid action-result sequences in the bridge adapter

## Root cause
The V2 `LlmBridgeAdapter` converted `ThreadMessage` values directly into `ChatMessage` values and forwarded them to the provider without calling `sanitize_tool_messages()`. When thread history contained an orphaned action result, Anthropic rejected the request with error 2013 because the `tool_result` no longer had a matching preceding tool call.

## Impact
- orphaned tool results are rewritten into normal user-visible messages before provider submission
- valid tool call / tool result pairs are preserved unchanged
- V2 Anthropic threads can recover from compaction, checkpoint restore, or pause/resume histories that would previously get stuck

## Validation
- `cargo fmt --all --check`
- `cargo test bridge::llm_adapter::tests --lib`

Closes #1950

### DIFF
--- a/src/bridge/llm_adapter.rs
+++ b/src/bridge/llm_adapter.rs
@@ -386,6 +386,39 @@ mod tests {
     }
 
     #[tokio::test]
+    async fn complete_without_tools_rewrites_orphaned_action_results_before_provider_call() {
+        let state = Arc::new(CapturingProviderState::default());
+        let provider: Arc<dyn LlmProvider> = Arc::new(CapturingProvider {
+            state: state.clone(),
+        });
+        let adapter = LlmBridgeAdapter::new(provider, None);
+        let messages = vec![
+            ThreadMessage::user("Find the docs"),
+            ThreadMessage::assistant("I checked a tool earlier."),
+            ThreadMessage::action_result("call_missing", "search", "result payload"),
+        ];
+
+        let output = adapter
+            .complete(&messages, &[], &LlmCallConfig::default())
+            .await
+            .unwrap();
+
+        match output.response {
+            LlmResponse::Text(ref text) => assert_eq!(text, "ok"),
+            other => panic!("expected text response, got {other:?}"),
+        }
+
+        let completion_requests = state.completion_requests.lock().await;
+        let sent = completion_requests.last().unwrap();
+
+        assert_eq!(sent.len(), 3);
+        assert_eq!(sent[2].role, Role::User);
+        assert_eq!(sent[2].content, "[Tool `search` returned: result payload]");
+        assert!(sent[2].tool_call_id.is_none());
+        assert!(sent[2].name.is_none());
+    }
+
+    #[tokio::test]
     async fn complete_with_tools_preserves_matched_action_results() {
         let state = Arc::new(CapturingProviderState::default());
         let provider: Arc<dyn LlmProvider> = Arc::new(CapturingProvider {
--- a/src/bridge/llm_adapter.rs
+++ b/src/bridge/llm_adapter.rs
@@ -7,7 +7,10 @@ use ironclaw_engine::{
     TokenUsage,
 };
 
-use crate::llm::{ChatMessage, LlmProvider, Role, ToolCall, ToolCompletionRequest, ToolDefinition};
+use crate::llm::{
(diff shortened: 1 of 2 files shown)


ilblackdragon (commented):
## Automated review — LGTM with minor notes

**TL;DR:** Well-scoped bug fix. Correctly isolates orphaned tool results at the adapter boundary using the existing `sanitize_tool_messages` helper. Good test coverage. Ready to merge after addressing the visibility nit below.

### Findings

| 1 | Low | `src/llm/mod.rs:59` | `pub(crate) use sanitize_tool_messages` re-export — per CLAUDE.md "No `pub use` re-exports unless exposing to downstream consumers" | Use the direct path `use crate::llm::provider::sanitize_tool_messages` at the sole call site in `llm_adapter.rs` |
| 2 | Low | `src/bridge/llm_adapter.rs:51` | `let mut chat_messages` but only mutated once via sanitize | Either document the in-place rationale or accept `&[ChatMessage]` and return a new vec |
| 3 | Low | Test structure | All three new async tests use `CapturingProvider` that always returns `Ok(...)`; no negative path | Add a test where the provider returns an error after sanitize, to verify error propagation is intact |

### Strengths
- Root cause clearly identified (missing `sanitize_tool_messages` call in adapter path → Anthropic error 2013)
- Smart reuse of existing helper (DRY)
- Uses `tracing::debug!()` — respects the TUI logging convention
- Tests cover both `complete()` and `complete_with_tools()` paths

### Open questions
1. Is the `pub(crate)` re-export intentional for future callers, or can it stay private to the `provider` module?
2. Are there other adapter paths (streaming, retries, failover) that could also receive corrupted message sequences and should call sanitize?

ilblackdragon (approved):

PR #1976, opened 2026-04-03
(fix) WASM channel HTTP SSRF protections

serrrfirat:
## Summary
This hardens the WASM channel HTTP request path to use the same SSRF defenses already present in the WASM tool wrapper.

## What changed
- added a shared WASM HTTP security helper for private-IP rejection and no-redirect reqwest client construction
- applied that helper to `src/channels/wasm/wrapper.rs` so channel HTTP requests reject private/internal IPs before connect
- switched the tool wrapper and OAuth refresh path to the same shared helper so the policy cannot drift again
- added a channel regression test proving loopback targets are rejected
- added a shared helper test proving redirects stay disabled

## Root cause
`src/channels/wasm/wrapper.rs` validated only the initial allowlisted URL, then built a default reqwest client without the two SSRF controls already used in `src/tools/wasm/wrapper.rs`:
- private/internal IP rejection after URL parsing and DNS resolution
- `redirect(Policy::none())`

That left the channel path weaker than the tool path and meant an allowlisted first hop could still reach an internal destination via redirect or DNS rebinding.

## Impact
Custom WASM channels now get the same outbound HTTP protections as WASM tools. Requests to loopback/private IPs are rejected up front, and redirect targets are no longer followed implicitly.

## Validation
- `cargo test private_ip --lib`
- `cargo test ssrf_safe_client_builder_disables_redirects --lib`

### DIFF
--- a/src/channels/wasm/wrapper.rs
+++ b/src/channels/wasm/wrapper.rs
@@ -54,10 +54,11 @@ use crate::channels::{Channel, IncomingMessage, MessageStream, OutgoingResponse,
 use crate::error::ChannelError;
 use crate::pairing::PairingStore;
 use crate::secrets::SecretsStore;
-use crate::tools::wasm::LogLevel;
-use crate::tools::wasm::WasmResourceLimiter;
 use crate::tools::wasm::credential_injector::{
     InjectedCredentials, host_matches_pattern, inject_credential,
+};
+use crate::tools::wasm::{
+    LogLevel, WasmResourceLimiter, reject_private_ip, ssrf_safe_client_builder,
 };
 use ironclaw_safety::LeakDetector;
 
@@ -391,6 +392,9 @@ impl near::agent::channel_host::Host for ChannelStoreData {
             .map(|h| h.max_response_bytes)
             .unwrap_or(10 * 1024 * 1024);
 
+        // Resolve hostname and reject private/internal IPs to prevent DNS rebinding.
+        reject_private_ip(&url)?;
+
         // Make the HTTP request using a dedicated single-threaded runtime.
         // We're inside spawn_blocking, so we can't rely on the main runtime's
         // I/O driver (it may be busy with WASM compilation or other startup work).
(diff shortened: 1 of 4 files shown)


serrrfirat:
This PR was briefly opened against `main`, which caused the review bot to comment on a much larger diff including `crates/ironclaw_engine/...` changes.

After retargeting the PR to `staging`, the current diff is only the 4 WASM SSRF hardening files:
- `src/channels/wasm/wrapper.rs`
- `src/tools/wasm/http_security.rs`
- `src/tools/wasm/mod.rs`
- `src/tools/wasm/wrapper.rs`

The existing inline bot comments reference files outside the current PR diff, so they appear to be stale/outdated from the earlier base. I did not make unrelated engine changes in this branch. A fresh review on the current `staging`-based diff would be more accurate.

ilblackdragon (commented):
## Automated review — LGTM

**TL;DR:** Effectively hardens a real SSRF gap in the WASM channel path (DNS rebinding + implicit redirects) by extracting shared defensive helpers. Ready to merge.

### Threat model addressed

| Attack | Mitigation |
|--------|-----------|
| DNS rebinding | Resolve hostname → check all returned IPs against private ranges before connect |
| Open redirect | `reqwest::redirect::Policy::none()` globally |
| CGNAT evasion | Blocks 100.64.0.0/10 |
| IPv6 ULA/link-local | Blocks \`fc00::/7\` and \`fe80::/10\` |
| IPv4 link-local | Blocks 169.254.0.0/16 |
| Userinfo injection | Rejects \`user:pass@\` URLs |
| Non-HTTP schemes | Rejects gopher/file/ftp before DNS resolution |

### Strengths
- Defense in depth: two independent checks (allowlist + SSRF guard) prevent policy drift
- DRY: extracted helpers to \`http_security.rs\` eliminate copy-paste drift risk
- End-to-end test (\`test_ssrf_safe_client_builder_disables_redirects\`) verifies redirect blocking via real HTTP server
- No \`.unwrap()\` in production; \`is_private_ip\` gated \`#[cfg(test)]\` where unused in production (cleanup commit 2b1ecb09)
- Full IPv4 + IPv6 private-range coverage per RFC 4291/5735

### Open questions / design notes
1. **Blanket redirect disable**: \`ssrf_safe_client_builder()\` disables redirects globally. Should there be a separate builder for flows wanting to allow *explicit single-hop* redirects (with SSRF validation between hops)? Current policy is correctly conservative but may block legitimate redirects in the future. If this is a design decision, worth a comment in the file.
2. **DNS resolution caching**: Inline resolution on each request. High-volume channels may want TTL cache. Orthogonal to this PR.
3. **Redirect test** binds to \`127.0.0.1:0\` — verify no port contention under parallel CI runners.

ilblackdragon (approved):

PR #1, opened 2026-02-06
Move whatsapp channel source to channels-src/ for consistency

bowenwang1996:
All WASM channel sources are now in channels-src/:
- slack/
- telegram/
- whatsapp/ (moved from channels/)

PR #1977, opened 2026-04-03
[codex] feat: enforce per-workspace RBAC with superadmin controls

serrrfirat:
## Summary

This stacked PR implements `#1608` on top of `#1734` by introducing explicit per-workspace RBAC enforcement and a separate system-level superadmin capability.

Before this change, workspace membership existed, but the gateway still relied on scattered, endpoint-specific checks. That left important gaps:

- role semantics were not encoded in one place
- read/write/admin/owner behavior was not consistently enforced across all workspace-scoped routes
- system-wide admin routes were still coupled to the legacy user `role` field instead of a dedicated superadmin concept
- ownership transfer and role changes did not have a single authoritative permission path

This PR closes that gap by making workspace authorization explicit, centralized, and enforced across the major browser-facing APIs.

## Problem Statement

Issue `#1608` calls for role-based access control with a per-workspace permission matrix. The workspace-entities work in `#1734` created the right substrate, but it did not yet fully define or enforce the higher-level policy model required for production use.

In particular, we needed to answer these questions consistently:

- what can a `viewer` do vs a `member`, `admin`, or `owner`
- which routes are read-only, write-level, admin-level, or owner-only
- how shared workspaces interact with personal scope
- how system user management relates to workspace roles
- how superusers can inspect or manage the system without being manually added to every workspace

This PR makes those rules first-class and wires them through the web gateway and DB-backed auth path.

## Core RBAC Model

This branch adds a central permission matrix in `src/channels/web/permissions.rs`.

Workspace roles:

- `viewer`
- `member`
- `admin`
- `owner`

Workspace permissions:

- `WorkspaceRead`
- `WorkspaceWrite`
- `WorkspaceManageMembers`
- `WorkspaceManageSettings`
- `WorkspaceManageAdmins`
- `WorkspaceDelete`

System permissions:

- `SystemManageUsers`
- `SystemViewAll`

Effective role behavior:

- `viewer`: can read workspace-scoped data, but cannot mutate
- `member`: can perform normal workspace writes
- `admin`: can manage members and workspace settings
- `owner`: required for owner-only and destructive workspace operations
- `superadmin`: bypasses workspace role checks and is allowed through system-level admin paths

The permission matrix is unit-tested directly so the role contract is explicit and not inferred from route behavior.

## Superadmin Separation From User Role

This PR intentionally separates workspace/system administration from the existing user-facing `role` column semantics.

New behavior:

- users now have a persisted `is_superadmin` flag
- auth propagation carries that flag through `UserIdentity`
- the `AdminUser` extractor now means "superadmin required"
- `/api/admin/*` is no longer unlocked by plain `role = admin`

Why this matters:

- workspace `admin` is scoped to a workspace
- user `role = admin` was too broad and ambiguous for true system administration
- `is_superadmin` creates a clear distinction between normal elevated users and platform-wide operators

Bootstrap behavior also stays coherent: the initial bootstrap/single-user admin path is promoted into superadmin capability, so existing top-level deployments still work as expected.

## Database And Backend Changes

The PR adds a new migration:

- `migrations/V18__users_superadmin.sql`

That migration introduces:

- `users.is_superadmin BOOLEAN NOT NULL DEFAULT FALSE`
- backfill behavior so existing admin users become superadmins during migration

The same persistence behavior is implemented in both backends:

- PostgreSQL
- libSQL

DB abstraction changes include:

- `UserRecord` / auth-facing identity changes to carry `is_superadmin`
- workspace management trait additions for listing all workspaces and performing ownership transfer
- backend-specific implementations for those new operations in both DB backends

This keeps the repository invariant intact: new persistence behavior is implemented through the shared trait first, then carried through both backends.

## Workspace Scope Resolution And Membership Handling

Workspace access control now runs through the workspace resolution layer instead of being left to ad hoc handler logic.

Key changes:

- archived workspaces are blocked during scope resolution
- superadmins resolve with effective owner-level authority for workspace checks
- non-superadmin membership lookups use a TTL cache
- membership cache entries are invalidated on role/member changes

The cache exists to avoid repeated DB lookups on high-traffic workspace routes while still keeping membership changes responsive.

Configuration:

- `GATEWAY_WORKSPACE_MEMBERSHIP_CACHE_TTL_SECS`
- default: `60`

This is intentionally small and conservative. Role changes are also invalidated directly where possible, so the normal update path does not wait on TTL expiry.

## Ownership Transfer Semantics

This PR makes owner transfer an explicit, atomic operation.

Behavior:

- assigning `owner` to another member is treated as ownership transfer, not as "add another owner"
- the previous owner is demoted to `admin`
- the new owner is promoted in the same DB operation
- owner-level permission is required for that change

This avoids inconsistent states such as multiple effective owners created by piecemeal updates or partial role rewrites.

## Endpoint-Level Enforcement

The largest practical change in this PR is that the permission model is now enforced across the actual API surface rather than just defined in docs or helper methods.

### Member-or-higher write enforcement

Workspace-scoped write operations now require `WorkspaceWrite`:

- chat send flows
- memory writes
- routine mutation endpoints
- job prompt/restart/cancel endpoints
- Responses API request creation

This means `viewer` can no longer use write paths that were previously reachable through partial membership checks.

### Admin-or-owner enforcement

Workspace admin operations now require `WorkspaceManageMembers` or `WorkspaceManageSettings`:

- workspace member management
- workspace settings writes/deletes
- workspace metadata updates

This means `member` is no longer treated as sufficient for routes that materially administer a shared workspace.

### Owner-only enforcement

Owner-only operations now require `WorkspaceManageAdmins` / `WorkspaceDelete`:

- workspace archive
- ownership transfer / assigning `owner`

This prevents workspace admins from taking owner-level actions.

### System-level enforcement

System-wide user-management routes now require `SystemManageUsers`, which is currently mapped to superadmin-only access:

- list users
- fetch user detail
- create users
- patch users
- suspend / activate users
- delete users
- manage per-user secrets through admin endpoints

This is the main behavior change for deployments that previously treated plain admin tokens as sufficient for `/api/admin/*`.

## Superadmin Workspace Visibility

Superadmins can now inspect all non-archived workspaces through the existing workspace list API without requiring explicit membership rows.

This is important for operational support and tenant administration because otherwise a system operator would need to be manually added to every workspace before being able to inspect or repair it.

The implementation uses the existing `GET /api/workspaces` endpoint rather than adding a one-off admin route, which keeps the API surface smaller while still preserving a clean authorization boundary.

## Personal Workspace Compatibility

This PR preserves backward compatibility for the personal scope model introduced before first-class workspace entities.

Behavior remains:

- no explicit `workspace` scope means the request operates on the caller's personal space
- personal data paths are not blocked by workspace membership checks
- shared workspace authorization only applies when the request resolves into a shared workspace scope

That matters because this branch is tightening shared-workspace behavior, not turning personal usage into a new RBAC burden.

## Tests Added And Expanded

This branch adds and expands targeted coverage for the actual issue behavior.

Notable cases now covered:

- permission matrix unit coverage for the role model itself
- viewer can read but cannot mutate
- viewer cannot write workspace memory
- viewer cannot toggle workspace routines
- member-level write paths are allowed
- admin can manage settings and members
- owner is required for archive
- superadmin can list workspaces without membership
- user-management endpoints require superadmin
- role changes take effect immediately
- assigning `owner` transfers ownership instead of duplicating it

Most of the end-to-end behavior is exercised in `tests/workspaces_integration.rs`, with additional auth/system coverage in multi-tenant tests.

## Documentation Updates

The PR also updates documentation so the implementation and operator-facing docs stay aligned:

- `docs/USER_MANAGEMENT_API.md`
- `src/channels/web/CLAUDE.md`
- `FEATURE_PARITY.md`

The user-management docs now explicitly describe `/api/admin/*` as superadmin-only rather than plain-admin accessible.

## Risk And Compatibility Notes

This is a high-risk area because it touches:

- auth
- persistence
- route authorization
- workspace membership semantics

Main compatibility implications:

- plain `admin` users without `is_superadmin` will now receive `403` on system admin endpoints
- `viewer` users may lose access to mutation paths that were previously under-protected
- owner-only actions are now stricter by design

Those changes are intentional and align the gateway with the policy described in `#1608`.

## Validation

Targeted verification completed on this branch:

- `cargo test --test workspaces_integration -- --nocapture`
- `cargo test multi_tenant --lib -- --nocapture`
- `cargo test workspace_permission_matrix_matches_issue_1608 --lib`
- `cargo clippy --all-features --tests -- -D warnings`

## Stack Context

- Base branch: `feat/1607-workspace-entities-db-users`
- Base PR: `#1734`
- This PR is the RBAC/policy layer that sits on top of the workspace-entities substrate

## Closes

- `#1608`

### DIFF
--- a/FEATURE_PARITY.md
+++ b/FEATURE_PARITY.md
@@ -22,7 +22,7 @@ This document tracks feature parity between IronClaw (Rust implementation) and O
 | Hub-and-spoke architecture | ✅ | ✅ | Web gateway as central hub |
 | WebSocket control plane | ✅ | ✅ | Gateway with WebSocket + SSE |
 | Single-user system | ✅ | ✅ | Explicit instance owner scope for persistent routines, secrets, jobs, settings, extensions, and workspace memory |
-| First-class workspace entities and sharing | ✅ | ✅ | DB-backed workspaces, membership, scoped jobs/routines/memory/settings, workspace SSE fan-out, archived workspace gating |
+| First-class workspace entities and sharing | ✅ | ✅ | DB-backed workspaces, membership, enforced workspace RBAC, scoped jobs/routines/memory/settings, workspace SSE fan-out, archived workspace gating, superadmin workspace visibility |
 | Multi-agent routing | ✅ | ❌ | Workspace isolation per-agent |
 | Session-based messaging | ✅ | ✅ | Owner scope is separate from sender identity and conversation scope |
 | Loopback-first networking | ✅ | ✅ | HTTP binds to 0.0.0.0 but can be configured |
--- a/docs/USER_MANAGEMENT_API.md
+++ b/docs/USER_MANAGEMENT_API.md
@@ -1,6 +1,6 @@
 # User Management API
 
-DB-backed user management for multi-tenant IronClaw deployments. Covers admin user CRUD, per-user secrets provisioning, self-service profile, API token management, and usage reporting.
+DB-backed user management for multi-tenant IronClaw deployments. Covers superadmin-only user CRUD, per-user secrets provisioning, self-service profile, API token management, and usage reporting.
 
 ## Authentication
 
@@ -12,14 +12,18 @@ DB tokens are SHA-256 hashed at rest; plaintext is returned exactly once at crea
 
 Auth is cached in a bounded LRU (1024 entries, 60s TTL). Suspending a user or revoking a token may take up to 60s to take effect.
 
-## Roles
+## Roles And System Access
 
 | Role | Scope |
 |------|-------|
-| `admin` | Full access to all endpoints |
+| `admin` | Elevated user role for compatibility and future policy; does not by itself unlock `/api/admin/*` |
 | `member` | Self-service profile + own token management only |
 
-Endpoints marked **Admin** return `403 Forbidden` for `member` role.
+Endpoints marked **Admin** require a superadmin identity. In practice that means:
+- the single-user bootstrap/env-var gateway token, or
+- a DB-backed token whose backing user has the internal `is_superadmin` flag.
+
+A plain `admin` role without superadmin status receives `403 Forbidden`.
 
 ---
 
@@ -29,7 +33,7 @@ Endpoints marked **Admin** return `403 Forbidden` for `member` role.
 
 Create a new user. Returns the user record and a one-time plaintext API token.
 
-**Auth:** Admin
+**Auth:** Superadmin
 
 **Request body:**
 
@@ -64,7 +68,7 @@ Create a new user. Returns the user record and a one-time plaintext API token.
 
 The `token` field is the plaintext API token. It is shown **only once** — store it securely.
 
-**Errors:** `400` (missing display_name, invalid role), `403` (not admin), `503` (no database)
+**Errors:** `400` (missing display_name, invalid role), `403` (superadmin required), `503` (no database)
 
 ---
 
@@ -72,7 +76,7 @@ The `token` field is the plaintext API token. It is shown **only once** — stor
 
 List all users.
 
-**Auth:** Admin
+**Auth:** Superadmin
 
 **Response:** `200 OK`
 
@@ -100,7 +104,7 @@ List all users.
 
 Get a single user by ID.
 
-**Auth:** Admin
+**Auth:** Superadmin
 
 **Response:** `200 OK`
 
@@ -119,7 +123,7 @@ Get a single user by ID.
 }
 ```
 
-**Errors:** `404` (user not found), `403` (not admin)
+**Errors:** `404` (user not found), `403` (superadmin required)
 
 ---
 
@@ -127,7 +131,7 @@ Get a single user by ID.
 
 Update a user's display name and/or metadata. Omitted fields are left unchanged.
 
-**Auth:** Admin
+**Auth:** Superadmin
 
 **Request body:**
 
@@ -146,7 +150,7 @@ Update a user's display name and/or metadata. Omitted fields are left unchanged.
 
 **Response:** `200 OK` — returns the full updated user record (same shape as GET detail, without `last_login_at`/`created_by`).
 
-**Errors:** `404` (user not found), `403` (not admin)
+**Errors:** `404` (user not found), `403` (superadmin required)
 
 ---
 
@@ -154,7 +158,7 @@ Update a user's display name and/or metadata. Omitted fields are left unchanged.
 
 Suspend a user. Suspended users cannot authenticate (DB auth checks user status).
 
-**Auth:** Admin
+**Auth:** Superadmin
 
 **Response:** `200 OK`
 
@@ -165,7 +169,7 @@ Suspend a user. Suspended users cannot authenticate (DB auth checks user status)
 }
 ```
 
-**Errors:** `404` (user not found), `403` (not admin)
+**Errors:** `404` (user not found), `403` (superadmin required)
 
 ---
 
@@ -173,7 +177,7 @@ Suspend a user. Suspended users cannot authenticate (DB auth checks user status)
 
 Re-activate a suspended user.
 
-**Auth:** Admin
+**Auth:** Superadmin
 
 **Response:** `200 OK`
 
@@ -184,7 +188,7 @@ Re-activate a suspended user.
 }
 ```
 
-**Errors:** `404` (user not found), `403` (not admin)
+**Errors:** `404` (user not found), `403` (superadmin required)
 
 ---
 
@@ -192,7 +196,7 @@ Re-activate a suspended user.
 
 Permanently delete a user and all associated data (tokens, jobs, conversations, memory, routines, settings, secrets).
 
-**Auth:** Admin
+**Auth:** Superadmin
 
 **Response:** `200 OK`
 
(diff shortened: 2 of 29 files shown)


ilblackdragon (commented):
**Verdict: Request changes**

- **Dual-backend backfill mismatch (high)**: `migrations/V18__users_superadmin.sql` backfills `is_superadmin = TRUE WHERE role = 'admin'`, but `src/db/libsql_migrations.rs:1957-1962` only `ALTER TABLE … ADD COLUMN … DEFAULT 0` with **no UPDATE**. Existing libSQL deployments lose `/api/admin/*` access for all pre-existing admin users on upgrade.
- **Duplicate `chat_send_handler`** at `src/channels/web/server.rs:1694` in addition to `src/channels/web/handlers/chat.rs:32`. Only one is wired; the other will silently drift.
- **`Role::parse(&role)?` in `handlers/workspaces.rs:957, :1007`** maps unknown DB role values to `500` instead of `403`. Should default-deny.
- **`pub use … workspace_scope_user_id`** in `web/mod.rs` exists only for the integration test — violates the "no `pub use` re-exports" rule. Move the test to import via the canonical path.
- Membership LRU+TTL cache invalidation on upsert+delete is correct. Atomic ownership transfer via SQL transaction is correct. Superadmin separation is the right call.
- Missing tests: libSQL backfill (would have caught the bug above), `/api/responses` thread-workspace permission end-to-end, concurrent ownership transfer.

The architecture is solid; the libSQL backfill omission is the only true blocker.

ilblackdragon (changes requested):
## Verdict: Request changes (close to approve)

The prior reviewer's libSQL data-loss blocker IS fixed in commit `78e2786`, but the fix interacts badly with `IDEMPOTENT_ADD_COLUMN_MIGRATIONS` and reintroduces a narrower version of the same data-loss risk.

## Status of prior review items

Re-verified each against current HEAD diff:

- **libSQL backfill** — FIXED (`src/db/libsql_migrations.rs` migration 18 now runs ALTER + UPDATE; regression test added).
- **`Role::parse` returns 500** — FIXED (`src/channels/web/permissions.rs` returns `(StatusCode::FORBIDDEN, …)`).
- **Duplicate `chat_send_handler`** — NOT fixed; both copies annotated `TODO: consolidate` and updated to share invariants.
- **`pub use workspace_scope_user_id`** — NOT removed; just annotated as test-only.

## Must-fix

### 1. Migration 18 in `IDEMPOTENT_ADD_COLUMN_MIGRATIONS` skips the backfill UPDATE

`src/db/libsql_migrations.rs:2062-2065` adds `(18, "users", "is_superadmin")` to `IDEMPOTENT_ADD_COLUMN_MIGRATIONS`. The runner (\`src/db/libsql_migrations.rs:857-887\`) checks \`column_exists()\` and, if true, sets \`skip_sql = true\` and skips the **entire** migration \`sql\` block — not just the ALTER, but also the \`UPDATE users SET is_superadmin = 1 WHERE role = 'admin';\` (diff lines 2050-2053).

For any libSQL deployment whose base schema was updated to include the column alongside the migration bundle, admins will NOT be promoted. The current regression test inserts the column via \`create_user\` then runs the bare \`UPDATE\` — it does NOT exercise the real runner's skip branch.

**Fix options:**
- Split skip logic so only the \`ALTER TABLE\` is conditional.
- Force \`skip_sql = false\` for migration 18.
- Move the backfill into application bootstrap.

Add a test for the \"base schema already has column, prior admin row exists\" path.

### 2. \`pub use\` in \`src/channels/web/mod.rs:1156-1158\` violates the project rule
The added doc comment acknowledges the rule but doesn't resolve it. Expose \`workspace_scope_user_id\` via \`test_helpers::\` (already \`#[cfg(test)]\`-gated and \`pub(crate)\`) and import it from there in \`tests/workspaces_integration.rs\` rather than re-exporting at the crate root.

## Should-fix

### 3. Duplicate \`chat_send_handler\`
Two copies now share coupled invariants (workspace resolution + \`WorkspaceWrite\` permission + timezone parsing). TODO comments are not a substitute — the dead copy will drift. Delete one in this PR; mechanical effort is low and recent bugs have come from handler drift.

### 4. Missing negative tests
- No test for \`/api/responses\` thread-workspace permission end-to-end.
- No test for concurrent ownership transfer (two admins racing).
- No test for \`archived\` workspace returning 410 Gone.

### 5. Membership cache: dead \`entries.pop(&key)\` on miss (diff line 868)
On a miss, \`entries.get()\` returned \`None\`, so popping is a no-op. On an expired hit it correctly removes. The current form is correct but confusing — minor refactor improves readability.

## Nits

- **\`const MAX_ENTRIES\`** with \`unreachable!()\` in const match — \`NonZeroUsize::new(4096).expect(…)\` is more idiomatic.
- **\`GATEWAY_WORKSPACE_MEMBERSHIP_CACHE_TTL_SECS\`** undocumented in \`.env.example\`.
- **\`Role::parse\` 403 body** \"Unknown workspace role '{role}'\" leaks implementation detail. Return generic message; log detail at \`warn!\`.
- **\`require_workspace_owner\`** uses \`>= Role::Owner\` — \`==\` is tighter and less fragile.

## Positive notes

- Permission matrix centralization is clean; \`workspace_permission_matrix_matches_issue_1608\` pins the contract.
- Atomic ownership transfer via SQL transaction is correct in both backends.
- Membership cache invalidation on upsert + delete is correct.
- Superadmin separation from \`role\` column is the right architectural call.
- \`is_superadmin\` propagation through \`UserIdentity\` is clean and consistent.

## Bottom line
Fix the migration-18 / idempotent skip interaction (must-fix #1) and the \`pub use\` (must-fix #2), and this can merge.

serrrfirat:
superseded

Issue #1915, opened 2026-04-02
ACP bridge follow-up prompt failures should mark sandbox job failed


serrrfirat:
## Problem

The main ACP failure path now reports `success: false` correctly, but follow-up prompt failures inside the ACP bridge loop are still handled as status events only.

In `src/worker/acp_bridge.rs`, when a follow-up `conn.prompt(...)` call fails, the bridge logs the error and posts a `"status"` event, but it does not update the final `AcpSessionOutcome` to failure. This means a job can still end up marked successful if the initial prompt succeeded but a later follow-up prompt failed.

## Why it matters

This creates an inconsistent job result:
- user-visible stream shows a follow-up failure
- final sandbox job completion may still be recorded as success

That is especially confusing for ACP jobs with iterative follow-up flows.

## Expected behavior

If any follow-up prompt fails in a way that makes the ACP session unusable, the bridge should mark the final job result as failed.

## Likely fix

In the follow-up loop in `src/worker/acp_bridge.rs`:
- convert prompt failures into a terminal failed `AcpSessionOutcome`, or
- track a failure flag and return `success: false` once the session ends

## Testing

Add a regression test covering:
- initial ACP prompt succeeds
- a later follow-up `prompt()` fails
- final `CompletionReport.success` is `false`

rajulbhatnagar:
Taking a look

serrrfirat:
merged closing

PR #1981, opened 2026-04-03
fix(acp): propagate follow-up prompt failures as job errors

rajulbhatnagar:
## Summary
- **Bug fix (#1915):** The ACP follow-up loop silently swallowed prompt failures — it logged and posted a status event but continued the loop, so jobs always reported `success: true` even when follow-up prompts failed.
- **Terminal event fix:** Per-turn ACP results were emitting `event_type: "result"` which maps to `AppEvent::JobResult` (the terminal signal). Job monitors exit on the first `JobResult`, so all subsequent follow-up outcomes were invisible. Changed per-turn events to `"turn_result"` (non-terminal) and emit exactly one terminal `"result"` from `run()` after the entire session ends.
- **Poll retry limits:** All `poll_prompt()` errors were retried infinitely. Now permanent errors (`OrchestratorRejected`, `LlmProxyFailed`) fail immediately, and transient errors (`ConnectionFailed`) retry with a cap of 5 (consistent with LLM retry patterns).
- **Child process cleanup:** Kill child process on protocol failure so stderr reader completes and the job doesn't hang. Race `poll_prompt` against child exit via `tokio::select!`.
- **Refactor:** Extract the inline loop into a standalone `run_follow_up_loop` function with trait-based dependency injection (`FollowUpPromptSource`, `AcpPromptSender`, `AcpEventSink`), enabling unit testing without Docker/ACP infrastructure.
- **Logging fix:** Downgrade `info!` → `debug!` in the follow-up loop since it runs in background Docker containers and was corrupting REPL/TUI output.
- Add `PartialEq` + `Clone` to `JobEventPayload`, narrow `AcpPromptSender` visibility to private.

## Known follow-up

`report_complete()` does not broadcast `AppEvent::JobResult` — it only updates container state. If the fire-and-forget `post_event("result")` fails silently, the job monitor never learns the job ended and the `max_jobs` slot stays consumed. This is a pre-existing gap affecting all three bridge types (ACP, Claude, Container). Fix: have the orchestrator's `report_complete` handler also broadcast `JobResult`. Tracked separately since it touches shared orchestrator infrastructure.

## Test plan
- [x] `follow_up_success_emits_turn_event_not_terminal` — per-turn events use `"turn_result"`, not `"result"`
- [x] `follow_up_prompt_failure_returns_error` — follow-up failure propagates as `Err`
- [x] `follow_up_loop_fails_on_permanent_poll_error` — `OrchestratorRejected` fails immediately
- [x] `follow_up_loop_exhausts_transient_retries` — 5 consecutive `ConnectionFailed` → `Err`
- [x] `follow_up_loop_recovers_from_transient_poll_error` — transient error then success → Ok
- [x] `follow_up_loop_exits_during_long_poll_when_child_dies` — child exit detected during blocked poll
- [x] `child_monitor_kills_process_so_stderr_reader_completes` — kill channel unblocks stderr
- [x] `cargo clippy` zero warnings, `cargo fmt` clean

### DIFF
--- a/src/worker/acp_bridge.rs
+++ b/src/worker/acp_bridge.rs
@@ -197,14 +197,7 @@ impl AcpBridgeRuntime {
 
         // Monitor the child process so the follow-up loop can exit if the agent dies.
         // The oneshot is Send, so it crosses the LocalSet boundary cleanly.
-        let (child_exit_tx, child_exit_rx) = tokio::sync::oneshot::channel::<Option<i32>>();
-        tokio::spawn(async move {
-            let exit_code = match child.wait().await {
-                Ok(status) => status.code(),
-                Err(_) => None,
-            };
-            let _ = child_exit_tx.send(exit_code);
-        });
+        let (child_exit_rx, kill_tx) = spawn_child_monitor(child);
 
         let local_set = tokio::task::LocalSet::new();
         let acp_result = local_set
@@ -287,7 +280,11 @@ impl AcpBridgeRuntime {
             })
             .await;
 
-        // Wait for stderr reader to finish
+        // Kill the child on protocol failure so stderr closes (see kill channel above).
+        if acp_result.is_err() {
+            let _ = kill_tx.send(());
+        }
+
         let _ = stderr_handle.await;
 
         acp_result
@@ -379,6 +376,38 @@ pub(crate) fn ironclaw_init_request() -> acp::InitializeRequest {
     )
 }
 
+// ==================== Child process monitor ====================
+
+/// Spawn a background task that monitors a child process for natural exit
+/// or a kill signal. Returns `(child_exit_rx, kill_tx)`.
+///
+/// - `child_exit_rx`: fires with the exit code when the child terminates
+/// - `kill_tx`: send `()` to terminate the child (e.g. on protocol failure
+///   so stderr closes and the stderr reader task can finish)
+fn spawn_child_monitor(
+    mut child: tokio::process::Child,
+) -> (
+    tokio::sync::oneshot::Receiver<Option<i32>>,
+    tokio::sync::oneshot::Sender<()>,
+) {
+    let (child_exit_tx, child_exit_rx) = tokio::sync::oneshot::channel();
+    let (kill_tx, kill_rx) = tokio::sync::oneshot::channel();
+    tokio::spawn(async move {
+        let exit_code_of =
+            |r: std::io::Result<std::process::ExitStatus>| r.ok().and_then(|s| s.code());
+        let exit_code = tokio::select! {
+            status = child.wait() => exit_code_of(status),
+            // Only fire on explicit send, not on sender drop.
+            Ok(()) = kill_rx => {
+                let _ = child.kill().await;
+                exit_code_of(child.wait().await)
+            }
+        };
+        let _ = child_exit_tx.send(exit_code);
+    });
+    (child_exit_rx, kill_tx)
+}
+
 // ==================== Follow-up loop ====================
 
 /// Wrapper that implements `AcpPromptSender` for an ACP `ClientSideConnection`.
@@ -413,7 +442,18 @@ async fn run_follow_up_loop(
 ) -> Result<(), WorkerError> {
     let session_id_str = session_id.to_string();
     loop {
-        let backoff = match prompt_source.poll_prompt().await {
+        // Race poll_prompt against child exit so we detect process death
+        // even during a long-poll HTTP request to the orchestrator.
+        let poll_result = tokio::select! {
+            result = prompt_source.poll_prompt() => result,
+            exit_code = &mut child_exit_rx => {
+                let code = exit_code.ok().flatten();
+                tracing::debug!(job_id = %job_id, exit_code = ?code, "ACP agent exited, ending follow-up loop");
+                break;
+            }
+        };
+
+        let backoff = match poll_result {
             Ok(Some(follow_up)) => {
                 if follow_up.done {
                     tracing::debug!(job_id = %job_id, "Orchestrator signaled done");
@@ -454,11 +494,7 @@ async fn run_follow_up_loop(
             _ = tokio::time::sleep(backoff) => {}
             exit_code = &mut child_exit_rx => {
                 let code = exit_code.ok().flatten();
-                tracing::debug!(
-                    job_id = %job_id,
-                    exit_code = ?code,
-                    "ACP agent process exited, ending follow-up loop"
-                );
+                tracing::debug!(job_id = %job_id, exit_code = ?code, "ACP agent exited, ending follow-up loop");
                 break;
             }
         }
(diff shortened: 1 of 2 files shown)


serrrfirat (commented):
Posting the requested high-severity finding from the review.

src/worker/acp_bridge.rs
```diff
@@ -420,6 +379,93 @@ pub(crate) fn ironclaw_init_request() -> acp::InitializeRequest {
...
+                    }
+                    Err(e) => {
+                        let msg = format!("Follow-up prompt failed: {e}");
+                        tracing::error!(job_id = %job_id, "{}", msg);
+                        sink.emit_event(&JobEventPayload {
+                            event_type: "status".to_string(),
+                            data: json!({ "message": msg }),
+                        })
+                        .await;
+                        return Err(WorkerError::ExecutionFailed { reason: msg });
```
serrrfirat, line 486: **High Severity** Early error propagation here can still leave the ACP job stuck instead of failed.

This new `return Err(...)` fixes the logical success/failure bug, but the process lifecycle is now split: the `Child` was moved into a background `wait()` task earlier, while `run_acp_session()` still unconditionally awaits the stderr reader after the `LocalSet` returns. If `conn.prompt()` fails at the protocol layer without the subprocess exiting immediately, this branch returns out of `run_follow_up_loop()`, then `run_acp_session()` can block waiting for stderr and never reach `report_complete(success = false)`.

Please explicitly terminate the child on this error path, or keep an owned handle that can be killed during cleanup before awaiting stderr. A regression test that simulates `prompt()` failing while the subprocess stays alive would pin this down.

rajulbhatnagar: Done

src/worker/acp_bridge.rs
```diff
@@ -420,6 +376,132 @@ pub(crate) fn ironclaw_init_request() -> acp::InitializeRequest {
...
+                    }
+                }
+                continue;
+            }
+            Ok(None) => Duration::from_secs(2),
+            Err(e) => {
+                tracing::warn!(job_id = %job_id, "Prompt polling error: {}", e);
```
serrrfirat: **Medium Severity**

All `poll_prompt()` errors take the same infinite 5s retry path here.

`poll_prompt()` also surfaces permanent failures such as `OrchestratorRejected` and response-parse errors, so if the prompt endpoint is broken while the ACP subprocess stays alive, this loop can keep the container alive indefinitely instead of failing the job.

Please distinguish retryable transport errors from permanent endpoint/protocol failures, or cap consecutive retries and convert exhaustion into `ExecutionFailed` plus a terminal error result. A regression test for repeated `OrchestratorRejected` or malformed prompt responses would cover this.

src/worker/acp_bridge.rs
```diff
@@ -420,6 +376,132 @@ pub(crate) fn ironclaw_init_request() -> acp::InitializeRequest {
...
+                tracing::debug!(job_id = %job_id, "Got follow-up prompt");
+
+                let follow_result = agent
+                    .send_prompt(session_id.clone(), follow_up.content)
+                    .await;
+
+                match follow_result {
+                    Ok(resp) => {
+                        let payload = stop_reason_to_result(&resp.stop_reason, &session_id_str);
+                        sink.emit_event(&payload).await;
```
serrrfirat, line 481: **High Severity**

This emits a `result` event for each successful follow-up turn, but on failure the new code only emits a `status` event before returning `Err`.

`AppEvent::JobResult` is the terminal signal that background job monitors use, and they ignore `status` events. Those monitors are also spawned for non-wait sandbox jobs, so an ACP job can still be observed as completed after an earlier follow-up succeeds while a later follow-up failure never propagates through the normal terminal event path.

Please emit exactly one terminal `result` for the overall ACP session, including an error `result` before returning on follow-up failure, and use a non-terminal event type for per-turn stop reasons. An integration test for a non-wait ACP job that fails on a later follow-up turn would lock this down.

rajulbhatnagar:
Addressed all review feedback:

**@serrrfirat (line 471, high severity):** Per-turn `"result"` events were prematurely terminating job monitors. Fixed by changing per-turn events to `"turn_result"` (non-terminal, falls through to `AppEvent::JobStatus` in the orchestrator) and emitting exactly one terminal `"result"` from `run()` after the entire session ends.

**@serrrfirat (line 488, medium severity):** `poll_prompt()` errors now have retry discrimination — `OrchestratorRejected` and `LlmProxyFailed` fail immediately (permanent), `ConnectionFailed` retries with a cap of 5 (~25s), consistent with `llm/retry.rs` patterns.

**Known follow-up (not in scope):** `report_complete()` does not broadcast `AppEvent::JobResult` — if the fire-and-forget `post_event("result")` fails silently, the monitor never learns the job ended. This is a pre-existing gap affecting all three bridge types. Fix: have the orchestrator's `report_complete` handler also broadcast `JobResult`. Will track separately since it touches shared orchestrator infrastructure.

src/worker/acp_bridge.rs
```diff
@@ -420,6 +381,156 @@ pub(crate) fn ironclaw_init_request() -> acp::InitializeRequest {
...
+                consecutive_poll_errors += 1;
+                if consecutive_poll_errors >= MAX_CONSECUTIVE_POLL_ERRORS {
+                    let msg = format!(
+                        "Prompt polling exhausted {} retries: {e}",
+                        MAX_CONSECUTIVE_POLL_ERRORS,
+                    );
+                    tracing::error!(job_id = %job_id, "{}", msg);
+                    return Err(WorkerError::ExecutionFailed { reason: msg });
+                }
+                tracing::warn!(
```
serrrfirat, line 513: **Medium Severity** This retry path still uses `warn!` in a background ACP loop.\n\nPer the repo logging guidance, background worker paths should avoid `info!` and `warn!` because they surface in the REPL/TUI and corrupt the interactive display. A transient `poll_prompt()` failure is explicitly retried here, so these warnings can leak noisy infrastructure blips into the user-facing terminal even when the loop recovers normally.\n\nPlease downgrade this to `debug!` and keep `error!` for permanent failures or retry exhaustion.

src/worker/acp_bridge.rs
```diff
@@ -103,28 +103,31 @@ impl AcpBridgeRuntime {
...
-                        message: Some(format!("ACP agent failed: {}", e)),
-                        iterations: 1,
-                    })
-                    .await?;
+                (false, format!("ACP agent failed: {e}"))
             }
-        }
+        };
+        self.client
+            .post_event(&JobEventPayload {
```
serrrfirat, line 116: **High Severity** Terminal ACP completion is still not reliably delivered.\n\nThis PR moved ACP to emit a single `result` event here, which fixes the earlier multi-terminal-event problem. However, `post_event()` is still fire-and-forget and only logs failures. If this POST is dropped or rejected, `report_complete()` still succeeds, but the orchestrator completion path does not broadcast `AppEvent::JobResult`; it only updates container state and cleans up. That means job monitors can miss the terminal signal entirely and keep the `max_jobs` slot occupied even though the worker finished.\n\nPlease make terminal completion delivery authoritative in one place: either have `report_complete()` also broadcast `JobResult`, or make terminal event delivery acknowledged/retried instead of best-effort.

rajulbhatnagar: I did note this in the PR description. It actually affects all 3 codepaths(Workee,Claude Code and ACP). Happy to tackle it here by updating -> report_complete

serrrfirat (approved):
Reviewed as paranoid architect. Well-structured fix for real bugs (silent follow-up failure swallowing, premature terminal events, infinite retries). Clean trait-based DI with thorough test coverage. No blockers found — minor findings (child exit code not checked, send_prompt not raced against child exit) are pre-existing and fit naturally into the acknowledged follow-up work.

PR #1990, opened 2026-04-03
fix(self-repair): skip built-in tools in broken tool detection and repair

j-bloggs:
## Summary

- **Problem**: The self-repair system was attempting to rebuild built-in tools (http, shell, json, etc.) when they accumulated failure counts. Built-in tools are part of the ironclaw binary — errors on them are caller-side issues (bad LLM parameters), not tool defects. This wasted LLM tokens and spammed users with repair notifications.
- **Fix**: Filter built-in tools out of the repair pipeline at two levels:
  1. **Detection** (`detect_broken_tools`): filters protected names before returning results
  2. **Repair** (`repair_broken_tool`): defense-in-depth guard rejects protected names even if detection missed them
- Adds `PROTECTED_TOOL_NAMES` list (47+ entries) covering all built-in tools, with public `is_protected_tool_name()` function
- Documents the `SelfRepair` trait contract about built-in tool exclusion

## Test plan

- [x] `detect_broken_tools_filters_out_builtins` — seeds real libSQL DB with failures for both a built-in ("http") and dynamic tool, verifies only the dynamic tool is returned
- [x] `repair_broken_tool_skips_builtin` — verifies defense-in-depth guard rejects built-in without invoking the builder (uses mock builder to confirm zero invocations)
- [x] `is_protected_tool_name_covers_common_builtins` — spot-checks the lookup function against known built-in names
- [x] All 15 self_repair tests pass

### DIFF
--- a/.claude/rules/database.md
+++ b/.claude/rules/database.md
@@ -18,7 +18,8 @@ See `src/db/CLAUDE.md` for full schema, dialect differences, and libSQL limitati
 4. Implement in `src/db/libsql/<module>.rs` (use `self.connect().await?` per operation)
 5. Add migration if needed:
    - PostgreSQL: new `migrations/VN__description.sql`
-   - libSQL: add `CREATE TABLE IF NOT EXISTS` to `libsql_migrations.rs`
+   - libSQL: add entry to `INCREMENTAL_MIGRATIONS` in `libsql_migrations.rs`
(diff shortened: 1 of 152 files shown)


j-bloggs:
closing - rebase error pulled in a whole bunch of other commits. Will resubmit with just the relevant changes

PR #1991, opened 2026-04-03
fix(self-repair): skip built-in tools in broken tool detection and repair

j-bloggs:
## Summary

- **Problem**: The self-repair system was attempting to rebuild built-in tools (http, shell, json, etc.) when they accumulated failure counts. Built-in tools are part of the ironclaw binary — errors on them are caller-side issues (bad LLM parameters), not tool defects. This wasted LLM tokens and spammed users with repair notifications.
- **Fix**: Filter built-in tools out of the repair pipeline at two levels:
  1. **Detection** (`detect_broken_tools`): filters protected names before returning results
  2. **Repair** (`repair_broken_tool`): defense-in-depth guard rejects protected names even if detection missed them
- Adds `PROTECTED_TOOL_NAMES` list (49 entries) covering all built-in tools, with public `is_protected_tool_name()` function
- Documents the `SelfRepair` trait contract about built-in tool exclusion

## Test plan

- [x] `detect_broken_tools_filters_out_builtins` — seeds real libSQL DB with failures for both a built-in ("http") and dynamic tool, verifies only the dynamic tool is returned
- [x] `repair_broken_tool_skips_builtin` — verifies defense-in-depth guard rejects built-in without invoking the builder (uses mock builder to confirm zero invocations)
- [x] `is_protected_tool_name_covers_common_builtins` — spot-checks the lookup function against known built-in names
- [x] All 15 self_repair tests pass

### DIFF
--- a/src/agent/self_repair.rs
+++ b/src/agent/self_repair.rs
@@ -48,6 +48,22 @@ pub enum RepairResult {
 }
 
 /// Trait for self-repair implementations.
+///
+/// # Built-in tool exclusion
+///
+/// Built-in tools (those checked by [`is_protected_tool_name`](crate::tools::is_protected_tool_name))
+/// must be excluded from repair workflows. Errors on built-in tools are
+/// caller-side issues (bad LLM parameters), not tool defects — attempting to
+/// rebuild them via `SoftwareBuilder` wastes tokens and cannot succeed.
+///
+/// `DefaultSelfRepair` enforces this at two levels:
+/// - **Detection**: `detect_broken_tools` filters out protected names before
+///   returning results.
+/// - **Repair**: `repair_broken_tool` rejects protected names as a
+///   defense-in-depth guard (returns `RepairResult::Success` with a skip
+///   message).
+///
+/// Custom implementations should follow the same convention.
 #[async_trait]
 pub trait SelfRepair: Send + Sync {
     /// Detect stuck jobs.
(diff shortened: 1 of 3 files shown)


j-bloggs:
Thanks for catching that! Removed the duplicate `tool_info` entry and moved `tool_permission_set` under its own "Permission tools" category. Fixed in 5937957b.

j-bloggs:
Hey\! 👋 Addressed Gemini's duplicate tool_info finding. This one's clean and ready for review. @zmanian @serrrfirat

serrrfirat:
Two findings from review:

1. **Medium** — `plan_update` is still missing from the centralized protected-tool list, so this PR does not fully enforce its new invariant.
   - `src/tools/builtin/plan.rs:35-38` defines `plan_update` as a built-in tool.
   - `src/tools/registry.rs:44-105` does not include `plan_update` in `PROTECTED_TOOL_NAMES`.
   - Both the new self-repair filter in `src/agent/self_repair.rs:269-322` and the built-in shadowing guard in `src/tools/registry.rs:189-197` depend on that list, so `plan_update` can still be treated as repairable and can still be shadowed by a dynamic registration.

2. **Low** — The new regression coverage only spot-checks a subset of built-ins, so it would not catch this kind of list drift.
   - `src/agent/self_repair.rs:650-670` checks only a handful of names.
   - A stronger regression test should assert parity between the protected-name set and the actual built-in registrations/tool names so future built-ins cannot silently miss the protection/repair exclusion path.

j-bloggs:
Added plan_update (and any other missing builtins) to PROTECTED_TOOL_NAMES. Thanks for catching this!

serrrfirat (approved):
Reviewed the latest PR head. The earlier plan_update gap is fixed, the targeted repair tests pass, and I do not see a remaining correctness or regression blocker for merge.

PR #2003, opened 2026-04-04
fix(tools): gate claude_code and acp modes behind enabled flags

### LINKED ISSUE #1987: create_job tool offers and accepts claude_code mode when Claude Code is disabled
## Summary

The `create_job` tool exposes `claude_code` as a valid `mode` option in its schema and accepts it at runtime even when `CLAUDE_CODE_ENABLED=false`. This causes the LLM to sometimes select `claude_code` mode, resulting in container failures (the container tries to run `claude-bridge` which doesn't exist or has no credentials).

## Steps to Reproduce

1. Set `CLAUDE_CODE_ENABLED=false` (or leave it at the default)
2. Enable sandbox (`SANDBOX_ENABLED=true`)
3. The LLM calls `create_job` with `"mode": "claude_code"`
4. Container is created with `JobMode::ClaudeCode`, running `claude-bridge` as the command
5. Container fails at runtime

## Root Cause

`CreateJobTool` has no reference to `ClaudeCodeConfig` and cannot check whether Claude Code is enabled:

- **`parameters_schema()`** (job.rs:835-840) — Always includes `"claude_code"` in the mode enum when `sandbox_enabled()` is true, regardless of `config.claude_code.enabled`
- **`execute()`** (job.rs:899-902) — Silently maps `"claude_code"` to `JobMode::ClaudeCode` with no validation
- **`execute_sandbox()`** — Passes the mode to `ContainerJobManager::create_job()` with no guard

## Expected Behavior

- `parameters_schema()` should only include `claude_code` in the mode enum when Claude Code is enabled
- `execute()` should reject `mode: "claude_code"` with an error when Claude Code is disabled

## Suggested Fix

Inject `claude_code_enabled: bool` into `CreateJobTool` (or pass it via `with_sandbox()`) and:
1. Conditionally include `claude_code` in the schema's mode enum
2. Add a guard in `execute()` that returns `ToolError::InvalidParameters` if the LLM requests `claude_code` mode when disabled

rajulbhatnagar:
## Summary

Fixes #1987. `CreateJobTool` always offered `claude_code` and `acp` as valid modes in its JSON schema and accepted them at runtime, even when `CLAUDE_CODE_ENABLED=false` / `ACP_ENABLED=false`. This caused the LLM to select disabled modes, spawning containers that fail at startup.

- Add `claude_code_enabled` and `acp_enabled` to `ContainerJobConfig` (follows existing `mcp_per_job_enabled` pattern), with `ContainerJobManager` accessors
- `parameters_schema()` dynamically builds the mode enum — disabled modes are hidden; `mode` field omitted when only `worker` is available; `agent_name` field omitted when ACP is disabled
- `execute()` rejects disabled modes with `ToolError::InvalidParameters` as defense-in-depth
- No changes to `register_job_tools()` signature or any call sites — flags flow through the already-injected `job_manager`

## Test plan

- [x] 10 new regression tests (schema gating, runtime rejection, accessor delegation, defaults)
- [x] 3 existing tests updated to explicitly enable the modes they exercise
- [x] `cargo fmt` — no changes
- [x] `cargo clippy` — zero warnings
- [x] `cargo test --lib` — 4193 passed, 0 failed

### DIFF
--- a/src/tools/builtin/job.rs
+++ b/src/tools/builtin/job.rs
@@ -452,10 +452,10 @@ impl CreateJobTool {
 
         // Persist the job mode to DB (for non-default modes).
         // For ACP, store "acp:<agent_name>" so restarts know which agent to use.
+        // Done synchronously so mode is available if the job needs restarting.
         if mode != JobMode::Worker
-            && let Some(store) = self.store.clone()
+            && let Some(ref store) = self.store
         {
-            let job_id_copy = job_id;
             let mode_str = if mode == JobMode::Acp
                 && let Some(ref agent) = params.acp_agent
             {
@@ -463,11 +463,9 @@ impl CreateJobTool {
             } else {
                 mode.as_str().to_string()
             };
-            tokio::spawn(async move {
-                if let Err(e) = store.update_sandbox_job_mode(job_id_copy, &mode_str).await {
-                    tracing::warn!(job_id = %job_id_copy, "Failed to set job mode: {}", e);
-                }
-            });
+            if let Err(e) = store.update_sandbox_job_mode(job_id, &mode_str).await {
+                tracing::warn!(job_id = %job_id, "Failed to set job mode: {}", e);
+            }
         }
 
         // Create the container job with the pre-determined job_id.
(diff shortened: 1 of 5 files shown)


src/tools/builtin/job.rs
```diff
@@ -822,59 +855,71 @@ impl Tool for CreateJobTool {
...
+                },
+                "max_iterations": {
+                    "type": "integer",
+                    "description": "Maximum number of agent loop iterations for the worker. \
+                                    Defaults to 50, capped at 500. Use lower values for simple tasks."
+                }
+            });
+            let modes = self.available_modes();
+            if let Some(map) = props.as_object_mut() {
+                if modes.len() > 1 {
```
serrrfirat: **Low Severity**\n\nNow that this schema can hide `mode` entirely when only `worker` is available, the rest of the tool metadata should stay in sync with that behavior. `CreateJobTool::description()` still tells the model to set `mode` to `claude_code`, so when Claude Code is disabled the model can still emit a stale `mode` field and waste a turn on the new `InvalidParameters` path.\n\nConsider making the description conditional on `available_modes()` or dropping the hardcoded Claude Code guidance.

src/tools/builtin/job.rs
```diff
@@ -920,7 +965,18 @@ impl Tool for CreateJobTool {
         if self.sandbox_enabled() {
             let wait = params.get("wait").and_then(|v| v.as_bool()).unwrap_or(true);
 
-            let mode = match params.get("mode").and_then(|v| v.as_str()) {
+            let mode_str = params.get("mode").and_then(|v| v.as_str());
+            if mode_str == Some("claude_code") && !self.claude_code_enabled() {
```
serrrfirat, line 971: **High Severity**\n\nThis guard only covers tool-driven job creation. The web restart flow in `jobs_restart_handler` reconstructs the stored mode and calls `ContainerJobManager::create_job(...)` directly, so a previously failed `claude_code` or `acp` job can still be restarted after `CLAUDE_CODE_ENABLED` / `ACP_ENABLED` has been turned off.\n\nThat leaves the original failure mode reachable through `/api/jobs/{id}/restart`, so the defense-in-depth claim here is incomplete unless `ContainerJobManager::create_job()` (or `create_job_inner()`) also rejects disabled modes for non-tool callers.

serrrfirat:
Review finding:

The restart-path fix is incomplete on the web/API side. `/api/jobs/{id}` still sets `can_restart` to `state.job_manager.is_some()` for sandbox jobs, while `/api/jobs/{id}/restart` now correctly rejects disabled `claude_code` / `acp` modes via `check_mode_enabled()`.

That leaves the UI in an inconsistent state: the detail view still renders a `Retry` button for failed/interrupted sandbox jobs based on `can_restart`, but clicking it can only return `409 Conflict` once the mode has been disabled. `can_restart` should be derived from the stored job mode plus `ContainerJobManager::is_mode_enabled()` instead of only checking whether a job manager exists.

I’d also add a focused API-level test for `/api/jobs/{id}` + restart behavior in the disabled-mode case, since the current helper-level tests don’t cover the contract the UI consumes.

src/tools/builtin/job.rs
serrrfirat: Addressed in c1126735. The `description()` no longer mentions `claude_code` — it's a generic sandbox description. The `mode_description()` (used in the schema's mode enum description) is already conditional on `available_modes()`, so disabled modes don't appear in guidance text. Additionally rebased onto staging to resolve merge conflicts.

src/tools/builtin/job.rs
serrrfirat: Addressed in c1126735. Added `check_mode_enabled()` guard in `jobs_restart_handler` — after resolving the stored mode, we now call `jm.is_mode_enabled(mode)` and return 409 Conflict with an environment variable hint if the mode has been disabled since the job was created. This closes the restart bypass path.

serrrfirat:
## Rebase & Review Summary

Rebased onto `staging` to resolve merge conflict in `src/channels/web/handlers/jobs.rs` — the conflict was between the new ownership model (`can_act_on` / `OwnerId`) merged into staging and this PR's `orchestrator::job_manager` imports. Both import sets are needed and are now present.

**Review comments status:**

| # | Comment | Status |
|---|---------|--------|
| 1 | `description()` hardcodes `claude_code` guidance (Low) | Already fixed in f760a41b — `description()` is now a generic sandbox description, `mode_description()` is conditional on `available_modes()` |
| 2 | Restart handler bypasses mode guard (High) | Already fixed in f760a41b — `check_mode_enabled()` added to `jobs_restart_handler` at line 479, returns 409 Conflict with env var hint for disabled modes |

**Quality gate (all passing locally and in CI):**
- `cargo fmt` — clean
- `cargo test --lib` — 3937 passed, 0 failed
- CI — all checks green ✅

No code changes beyond the conflict resolution were needed — the author's second commit (`f760a41b`) already addressed both review findings correctly.

serrrfirat:
## Paranoid Review Findings — Fixed in 312a24bd

| # | Severity | File | Finding | Fix |
|---|----------|------|---------|-----|
| 1 | **Medium** | `job_manager.rs:331` | `create_job()` accepted any `JobMode` without validating `is_mode_enabled()`. Both current callers validate, but any future caller could bypass mode gating. | Added `ModeDisabled` variant to `OrchestratorError` + guard at top of `create_job()` + regression test |
| 2 | **Low** | `job.rs:466` | Job mode persisted via `tokio::spawn` fire-and-forget. If `update_sandbox_job_mode` fails (transient DB error), mode silently lost — restart would default to `worker` instead of original mode. | Made mode persistence synchronous (awaited before `create_job()`) |
| 3 | **Nit** | `job.rs:897` | `if let Some(map) = props.as_object_mut()` always true since `props` is a JSON object literal. Dead guard + would need `.expect()` which violates project rules. | Refactored to build `serde_json::Map` directly |

**Quality gate:** `cargo fmt` clean, `cargo clippy` zero warnings, `cargo test --lib` 4287 passed.

serrrfirat (approved):
Reviewed and tested. Schema gating, runtime rejection, and restart guard all working correctly. Three hardening fixes added (defense-in-depth in create_job, synchronous mode persistence, schema builder cleanup). CI green, 4287 unit tests passing.

PR #2029, opened 2026-04-05
fix(registry): use canonical underscore names in manifests to fix WASM install

ilblackdragon:
## Summary

- **Manifest `name` fields** used hyphens (`google-calendar`) but the internal canonical form uses underscores (`google_calendar`). The release workflow packages `.wasm` files named after the manifest `name`, so archives contained `google-calendar.wasm`. The extension manager canonicalized the name before extraction, looked for `google_calendar.wasm`, and failed.
- Updated all 9 hyphenated manifest `name` fields and `_bundles.json` references to use canonical underscore form — future releases will package archives with matching filenames.
- Added hyphenated-name fallback in both tar.gz extractors so existing v0.22.0 release artifacts remain installable.

## Test plan

- [x] All 4219 unit tests pass
- [x] Zero clippy warnings
- [x] Added regression test `test_extract_tar_gz_matches_hyphenated_alias` covering the hyphen→underscore extraction path
- [ ] Verify WASM install of `google_calendar` and `web_search` via gateway UI on a fresh instance


registry/tools/web_search.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "web-search",
+  "name": "web_search",
```
serrrfirat, line 2: **High Severity**\n\nChanging the manifest `name` here fixes archive extraction, but it also changes the public extension identifier emitted by the registry and extension-list APIs from `web-search` to `web_search`. `ExtensionRegistry` canonicalizes catalog names, and the web handlers serialize `e.name` / `ext.name` directly, so this becomes a user-visible contract change rather than a packaging-only fix.\n\nThat conflicts with the current API/e2e contract, which still addresses this extension as `web-search` in registry/list/setup/install/remove flows.\n\nA safer fix is to keep the public manifest identifier stable and introduce a separate canonical artifact/install stem, or translate back to the legacy hyphenated identifier at the API boundary and add a regression test covering the legacy name end-to-end.

src/registry/installer.rs
```diff
@@ -715,6 +715,12 @@ fn extract_tar_gz(
 
     let wasm_filename = format!("{}.wasm", name);
     let caps_filename = format!("{}.capabilities.json", name);
+    // Pre-v0.23 archives used hyphenated filenames (e.g. "google-calendar.wasm")
+    // while the canonical name uses underscores ("google_calendar"). Accept both
+    // so existing release artifacts remain installable.
+    let alias = crate::extensions::naming::legacy_extension_alias(name);
```
serrrfirat: **Medium Severity** — Unhelpful error message on alias miss

When the WASM file is not found in the archive, the error message only mentions the canonical filename (e.g. `google_calendar.wasm`). If a user is debugging a failed install of a pre-v0.23 archive, it doesn't hint that the hyphenated alias was also tried. Same issue exists in `src/extensions/manager.rs:2727-2730`.

**Suggestion:** Include the alias in the error message when one was attempted:
```rust
if !found_wasm {
    let msg = match &alias_wasm {
        Some(a) => format!(
            "tar.gz archive does not contain '{}' or '{}'. Archive may be malformed.",
            wasm_filename, a
        ),
        None => format!(
            "tar.gz archive does not contain '{}'. Archive may be malformed.",
            wasm_filename
        ),
    };
    return Err(RegistryError::DownloadFailed {
        url: url.to_string(),
        reason: msg,
    });
}
```

registry/tools/google_calendar.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "google-calendar",
+  "name": "google_calendar",
```
serrrfirat, line 2: **Medium Severity** — JSON filename / manifest name mismatch

The manifest filename is still `google-calendar.json` (hyphens) but the `name` field inside is now `google_calendar` (underscores). This works because the catalog keys by `manifest.name` not filename — but it creates a confusing discrepancy for contributors.

The same applies to all 9 renamed manifests: `google-docs.json`, `google-drive.json`, `google-sheets.json`, `google-slides.json`, `llm-context.json`, `slack.json` (name=`slack_tool`), `telegram.json` (name=`telegram_mtproto`), `web-search.json`.

**No functional bug**, but consider renaming the JSON files to match in a follow-up (e.g. `google_calendar.json`).

src/extensions/manager.rs
```diff
@@ -2667,6 +2667,12 @@ impl ExtensionManager {
 
         let wasm_filename = format!("{}.wasm", name);
         let caps_filename = format!("{}.capabilities.json", name);
+        // Pre-v0.23 archives used hyphenated filenames (e.g. "google-calendar.wasm")
+        // while the canonical name uses underscores ("google_calendar"). Accept both
+        // so existing release artifacts remain installable.
+        let alias = crate::extensions::naming::legacy_extension_alias(name);
```
serrrfirat: **Medium Severity** — Duplicated alias fallback logic

This alias fallback block is duplicated verbatim in `src/registry/installer.rs:718-723`. Both compute the alias identically and check filenames identically. If the alias logic ever changes (e.g. supporting additional legacy naming conventions), both sites must be updated in lockstep.

**Suggestion:** Extract the shared tar.gz extraction + alias-matching logic into a common helper, or at minimum add a cross-reference comment in both locations so maintainers know to update both.

src/registry/installer.rs
```diff
@@ -1278,6 +1287,34 @@ mod tests {
         assert!(result.has_capabilities);
     }
 
+    #[test]
+    fn test_extract_tar_gz_matches_hyphenated_alias() {
```
serrrfirat, line 1287: **Low Severity** — Missing parallel test for ExtensionManager path

This regression test covers `registry/installer.rs`'s `extract_tar_gz`, but the identical alias fallback logic in `extensions/manager.rs`'s `extract_wasm_tar_gz` has no corresponding test. If one path regresses independently of the other, it would go undetected.

**Suggestion:** Add an equivalent test for `ExtensionManager::extract_wasm_tar_gz`, or better yet, extract the shared logic (per the duplication finding) so a single test covers both.

src/registry/installer.rs
```diff
@@ -1278,6 +1287,34 @@ mod tests {
...
+        // Regression: archives contain hyphenated filenames (e.g. "google-calendar.wasm")
+        // but the canonical name uses underscores ("google_calendar"). The extractor
+        // must accept the hyphenated form when the canonical name is passed.
+        let gz_bytes = build_test_tar_gz(
```
serrrfirat, line 1291: **Low Severity** — Pre-existing test exercises obsolete scenario

The nearby test `test_extract_correct_wasm_from_tool_bundle` (line 1251) passes `"slack-tool"` as the name. With manifests now using `"slack_tool"`, this test exercises a scenario that no longer occurs in production. It still validates that direct filename matching works, but doesn't reflect the real flow.

**Suggestion:** Consider adding a companion test using the new canonical name (`"slack_tool"`) with an archive containing `slack_tool.wasm` to cover the actual production path going forward.

registry/tools/slack_tool.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "slack-tool",
+  "name": "slack_tool",
```
serrrfirat, line 2: **Nit** — Most confusing filename/name mismatch

The JSON filename is `slack.json` but the manifest name inside is `slack_tool`. All the other files at least share a common stem (e.g. `google-calendar.json` → `google_calendar`), but this one has a completely different stem.

**Suggestion:** Consider renaming to `slack-tool.json` or `slack_tool.json` to match the manifest name, in a follow-up.

src/extensions/manager.rs
```diff
@@ -2667,6 +2667,12 @@ impl ExtensionManager {
...
+        let alias_wasm = alias.as_ref().map(|a| format!("{}.wasm", a));
+        let alias_caps = alias.as_ref().map(|a| format!("{}.capabilities.json", a));
```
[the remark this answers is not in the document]
ilblackdragon: Addressed in 6577519f — extracted `ArchiveFilenames` helper to `naming.rs` that encapsulates alias generation and matching. Both extractors now use it.

src/registry/installer.rs
```diff
@@ -715,6 +715,12 @@ fn extract_tar_gz(
 
     let wasm_filename = format!("{}.wasm", name);
     let caps_filename = format!("{}.capabilities.json", name);
+    // Pre-v0.23 archives used hyphenated filenames (e.g. "google-calendar.wasm")
+    // while the canonical name uses underscores ("google_calendar"). Accept both
+    // so existing release artifacts remain installable.
+    let alias = crate::extensions::naming::legacy_extension_alias(name);
+    let alias_wasm = alias.as_ref().map(|a| format!("{}.wasm", a));
+    let alias_caps = alias.as_ref().map(|a| format!("{}.capabilities.json", a));
```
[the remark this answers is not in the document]
ilblackdragon: Addressed in 6577519f — same `ArchiveFilenames` helper, used in both sites now.

src/registry/installer.rs
```diff
@@ -765,7 +773,8 @@ fn extract_tar_gz(
                 })?;
```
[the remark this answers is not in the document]
ilblackdragon: Addressed in 6577519f — error messages now use `ArchiveFilenames::wasm_not_found_msg()` which lists both canonical and alias filenames. Also fixed read error to use the actual matched `filename`.

src/extensions/manager.rs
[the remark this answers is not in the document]
ilblackdragon: Addressed in 6577519f — replaced the fully-qualified `crate::extensions::naming::legacy_extension_alias(name)` call with the shared `ArchiveFilenames::new(name)` helper, which is also referenced via `crate::extensions::naming::` but only once.

registry/tools/web_search.json
ilblackdragon: Not a contract change — `canonicalize_entries()` already converted all registry entry names to underscores on load, so the API was already returning `web_search` internally. User input (`web-search`) still works via `canonicalize_extension_name()`. The manifest name change only aligns the source-of-truth with what the system was already doing.

src/registry/installer.rs
ilblackdragon: Addressed in 6577519f — `ArchiveFilenames::wasm_not_found_msg()` now includes both canonical and alias filenames in the error message.

registry/tools/google_calendar.json
ilblackdragon: Addressed in 6577519f — all 9 manifest JSON files renamed to match their `name` fields (e.g. `google-calendar.json` → `google_calendar.json`, `slack.json` → `slack_tool.json`).

src/extensions/manager.rs
ilblackdragon: Addressed in 6577519f — extracted `ArchiveFilenames` struct to `naming.rs` with `is_wasm()`, `is_caps()`, and `wasm_not_found_msg()` methods. Both extractors use this single helper now.

src/registry/installer.rs
ilblackdragon: Addressed in 6577519f — the shared `ArchiveFilenames` helper is now tested directly in `naming.rs` with 4 test cases covering canonical match, alias match, and error message formatting. Both extractors use this helper, so a single test suite covers both paths.

src/registry/installer.rs
ilblackdragon: Addressed in 6577519f — updated `test_extract_correct_wasm_from_tool_bundle` to use canonical `slack_tool` name with `slack_tool.wasm` archive, matching the current production path.

registry/tools/slack_tool.json
ilblackdragon: Addressed in 6577519f — renamed `slack.json` → `slack_tool.json` to match the manifest `name` field.

PR #2031, opened 2026-04-05
`fix(agent,workspace): harden compaction consistency and append concurrency with lock-scope optimizations`

793383996:
## Summary

-

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] CI/Infrastructure
- [ ] Security
- [ ] Dependencies

## Validation

- [x] `cargo build`
- [ ] Relevant tests pass: 
- [ ] Manual testing: 

## Review Follow-Through

---

**Review track**:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,6 @@
 
+docs
+.tools
 .env
 .env.local
 .env.*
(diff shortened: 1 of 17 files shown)


standardtoaster:
Heads up this bundles compaction/workspace lock fixes with the codex_local setup wizard — might be worth splitting since the concurrency stuff touches session locks and workspace append paths, which are hard to validate.

On that: the validation section is unchecked for tests and manual testing. For the lock-scope changes, do you have a way to reproduce the original contention? Something like a concurrent-append stress test would give confidence the fix actually holds under load. The compaction snapshot matching changes are also tricky to verify — would be good to have a test that compacts mid-conversation and checks the context is consistent afterward. Happy to help think through test scenarios if useful.

793383996:
Yes, I still need to test for a while. I’ve found there are still some issues. I may first submit fixes for part of the problems and put **codex_local_** on my local machine first.
Temporarily closed

793383996:
I found that after closing Ironclaw and reopening it, it fails to connect.

PR #2038, opened 2026-04-05
fix(auth) first-pass Gmail OAuth auth prompt in chat

### LINKED ISSUE #2001: Gmail OAuth connection link not generated on first attempt
Description: When Sergey asked to connect Gmail, the agent did not generate the
OAuth authentication link on the first request. The agent responded with setup
instructions but no clickable link. Only after prompting a second time did the agent
produce the OAuth URL.

Impact: MEDIUM — Users who follow the agent's first response expecting a link will
be confused when none appears. This adds friction and makes the Gmail setup flow
feel unreliable.

serrrfirat:
## Summary
- surface OAuth auth prompts to the chat UI as soon as `tool_auth` or `tool_activate` returns an `auth_url`
- keep thread auth mode limited to `awaiting_token` flows so manual credential entry behavior stays unchanged
- add a regression test covering OAuth results that provide an auth link without `awaiting_token=true`

Fixes #2001.

## Verification
- `cargo test detect_auth_awaiting --lib`
- `cargo test extract_auth_prompt_detects_oauth_link_without_awaiting_token --lib`


ilblackdragon (commented):
## Automated review — Approve with fixes

**TL;DR:** Good OAuth + manual-token refactor with thoughtful \`PendingAuthPrompt\` design. One medium finding (user-facing error leakage) should be addressed before merge.

### Security considerations
- **Auth state persistence**: \`PendingAuthPrompt\` correctly survives approval pauses — prevents UI loss during multi-tool turns
- **Token bypass**: \`process_auth_token()\` intentionally bypasses logging/history (session.rs:1700) — document this carefully
- **String equality**: \`token == code\` (extensions/manager.rs:344) is acceptable for verification codes but NOT for cryptographic secrets. If this path ever accepts real secrets, switch to constant-time compare.
- **Interruption reset**: Existing design confirmed — control messages should cancel auth mode

### Findings

| 1 | **Medium** | \`thread_ops.rs:1772, 1787\` | Raw \`ExtensionError::ValidationFailed(...)\` strings surfaced to users via \`StatusUpdate\` — may leak internal validation details (e.g., \"wrong format for API key\") | Map validation errors to a generic user-facing message (\"Invalid token. Please try again.\") Keep internal details in \`debug!\` logs only. Consider a \`ValidationError\` enum with separate user/internal messages. |
| 2 | Low | \`dispatcher.rs:1340\` | \`unwrap_or_else(\|\| \"...\".to_string())\` for a single static string is inconsistent with codebase style | Use \`.unwrap_or(\"...\")\` |
| 3 | Low | \`session.rs:164\` | \`PendingAuthPrompt\` struct lacks non-empty validation on \`extension_name\` at construction | Add \`PendingAuthPrompt::new(...)\` constructor with non-empty checks, or document the caller contract |
| 4 | Info | \`dispatcher.rs:1938-1980\` | New tests cover OAuth + token logic; \`check_auth_required\` is \`#[cfg(test)]\`-gated (line 1379) | Document why test-only; verify not dead-code artifact |

### Convention compliance
- ✓ No \`.unwrap()\` in production code
- ✓ Strong types (\`PendingAuthPrompt\` struct vs raw strings)
- ✓ Uses \`debug!\` for internals
- ⚠ Error messages leaking validation logic (finding #1)

### Open questions
1. **OAuth callback timeout**: Does \`auth_url\` have an expiry? If a user pauses on approval for hours, is the OAuth link still valid when they resume?
2. **Multi-auth turns**: If \`tool_auth\` + \`tool_activate\` both return OAuth prompts in the same turn, \`capture_auth_prompt()\` keeps the first. Is dropping the second intentional, or should we log/warn?

serrrfirat:
@ilblackdragon addressed in `b9cdcd9d`:

- Validation failures during auth retry no longer surface raw `ExtensionError::ValidationFailed(...)` details to users. The chat/UI path now returns a generic `Invalid token. Please try again.` message, while the original error stays in `debug!` logging.
- Blank / whitespace-only extension names are now rejected when parsing fresh auth prompts and when restoring persisted auth prompts after an approval pause. That closes the empty-name invariant without broadening the auth state surface.
- Default token instructions are now centralized, and `check_auth_required()` is explicitly documented as test-only so it’s clearer that the runtime path goes through the main auth-prompt capture flow.

I did not change the broader multi-auth behavior in this commit.

ilblackdragon (commented):
**Verdict: Approve with two requested changes**

- **No URL scheme validation** on `auth_url`/`setup_url` before emitting `StatusUpdate::AuthRequired`. A buggy or malicious extension could return `javascript:` or `file://` and the UI would render it. Allowlist `https://` only.
- **OAuth link withheld during pending approval** (`src/agent/dispatcher.rs:1191-1213`): the prompt is persisted to `PendingApproval` but not emitted in the same turn. Users see the approval card with no connect button. On approval rejection the link may be lost entirely. Consider emitting alongside the approval card.
- **`src/agent/thread_ops.rs:1541-1568`** — synthesizes a fake JSON `auth_result` to feed `handle_auth_intercept`. Brittle if the schema drifts; refactor `handle_auth_intercept` to accept structured data.
- Excellent regression coverage: `extract_auth_prompt_detects_oauth_link_without_awaiting_token`, ordering tests, end-to-end persist/restore through dispatcher with mock OAuth + approval tools.
- No `.unwrap()/.expect()` in prod paths. `tracing::debug!` used correctly. No DB schema changes.

serrrfirat:
## Addressing review feedback from @ilblackdragon

All findings from both reviews have been addressed in commits `9f91f7d` and `27740b8`. Here's the mapping:

### Review 1 (Apr 6 — "Approve with fixes")

| # | Finding | Status |
|---|---------|--------|
| 1 | **Medium** — Raw `ValidationFailed` strings leaked to users | **✓ Fixed.** Added `auth_retry_message_for_error()` that maps all validation errors to generic `"Invalid token. Please try again."` — internal details go to `debug!` only. Regression test: `test_auth_retry_message_hides_validation_details`. |
| 2 | **Low** — `unwrap_or_else` style inconsistency | **✓ Fixed.** Now uses `.unwrap_or(DEFAULT_AUTH_TOKEN_INSTRUCTIONS).to_owned()` — correct idiom for static strings. |
| 3 | **Low** — `PendingAuthPrompt` lacks constructor validation | **✓ Fixed.** Added `PendingAuthPrompt::new()` with trim + non-empty validation (returns `Option<Self>`). Direct struct construction restricted to deserialization/tests. Tests: `test_pending_auth_prompt_new_rejects_empty_name`, `test_pending_auth_prompt_new_accepts_valid_name`. |
| 4 | **Info** — `#[cfg(test)]` function needs docs | **✓ Addressed.** Test names are self-documenting (`test_auth_retry_message_hides_validation_details`, `test_parse_auth_result_strips_non_https_urls`, etc.). Happy to add explicit doc comments if preferred. |

**Open questions:**
- **OAuth callback timeout** — The `auth_url` expiry is controlled by the extension/provider (e.g., Google OAuth codes expire in ~10 min). We don't validate expiry on our side currently. Worth adding a timeout/re-prompt if the link goes stale?
- **Multi-auth turns** — `capture_auth_prompt()` keeping only the first is intentional: one auth flow per turn keeps the UX simple and avoids confusing the user with multiple auth prompts simultaneously. If both tools need auth, the second will prompt on the next turn.

### Review 2 (Apr 8 — "Approve with two requested changes")

| Finding | Status |
|---------|--------|
| **No URL scheme validation** on `auth_url`/`setup_url` | **✓ Fixed.** Added `sanitize_auth_url()` enforcing `https://` only — rejects `javascript:`, `file://`, `http://`, `data:`, and empty strings. Applied in `parse_auth_result()`. Tests: `test_sanitize_auth_url_rejects_non_https_schemes`, `test_parse_auth_result_strips_non_https_urls`. |
| **OAuth link withheld during pending approval** | **✓ Fixed.** Auth prompt now emitted alongside the approval card in the same turn (lines 1557–1580). Users see the connect button immediately. |
| **Brittle fake JSON synthesis** in `handle_auth_intercept` | **✓ Fixed.** Refactored `handle_auth_intercept()` to accept `ParsedAuthData` directly instead of synthesizing JSON. Eliminates schema drift risk. |

**Bonus:** `27740b8` fixes `cargo fmt` issues in the test assertions that were causing the Formatting CI check to fail.

ilblackdragon (commented):
## Overview
Fixes #2001 — when `tool_auth`/`tool_activate` returns an OAuth `auth_url` without `awaiting_token=true`, the chat UI never showed the connect button. PR refactors auth-prompt extraction into `extract_auth_prompt`/`capture_auth_prompt`, persists the first prompt seen in a turn through approval pauses (new `PendingAuthPrompt` on `PendingApproval`), and adds an `https://`-only sanitizer for `auth_url`/`setup_url`.

## Strengths
- Real bug fix with a focused regression test (`test_extract_auth_prompt_detects_oauth_link_without_awaiting_token`).
- `sanitize_auth_url` is a nice unprompted security win — rejects `javascript:`, `file://`, `data:`, and even plain `http://`. Tests cover all of those.
- `PendingAuthPrompt::new` rejects blank `extension_name`; `restore_selected_auth_prompt` re-validates on read so corrupted/malicious DB rows can't slip through.
- Auth-mode entry now limited to actual `awaiting_token` flows — keeps manual-token UX unchanged.
- Auth retry message is a fixed string ("Invalid token. Please try again.") instead of leaking validation internals. Good defence-in-depth.
- Persisting the prompt across approval pauses fixes a real coordination bug (auth card disappearing if a tool needed approval first).

## Issues / suggestions

**`http://` rejection may be too strict for self-hosted/dev**
`sanitize_auth_url` requires `https://`. Self-hosted MCP/extension auth servers on `http://localhost:...` will be silently dropped — the user sees no link and no error. Either: (a) allow `http://localhost` / `http://127.0.0.1`, or (b) log a `warn!` when an auth URL is rejected so misconfiguration is debuggable instead of invisible.

**`startsWith` is weak URL parsing**
`url.starts_with("https://")` lets through `https://` followed by anything, including non-URL junk that downstream renderers could mishandle. Use the `url` crate (already a transitive dep) and check `parsed.scheme() == "https"`. Same for `setup_url`.

**Duplicate `AuthRequired` emits**
Two paths now emit `StatusUpdate::AuthRequired` for the same prompt: once before `NeedApproval` (`dispatcher.rs:75-89`) and again on resume in `thread_ops.rs:1062-1071`. If approval resolves quickly, the user sees the same connect card twice. Consider tagging the second emit as "already-shown" or de-duping in the channel layer.

**`check_auth_required` is now `#[cfg(test)]`-only**
Good cleanup — but please confirm no non-test path still imports it before merge.

**Field duplication: `extension_name` lives in both `ParsedAuthData` and the tuple**
`selected_auth_prompt: Option<(String, ParsedAuthData)>` plus `ParsedAuthData::extension_name: Option<String>` — the same name in two places, with `restore_selected_auth_prompt` reconstructing both. Either drop the tuple wrapper and use `ParsedAuthData` directly, or drop `extension_name` from `ParsedAuthData`. Right now there's a class of bugs where the two disagree.

**Test for OAuth-without-awaiting + approval-pause path is shallow**
`test_need_approval_persists_first_auth_prompt_for_resume` is good, but it doesn't exercise the *resume* path through `thread_ops::resume_after_approval` — only the persistence side. A follow-up test that approves the tool and asserts the auth card is restored on the resume side would be valuable.

**Migration concern**
`PendingApproval::selected_auth_prompt` is `#[serde(default)]` so old serialized rows still load — good. But existing in-flight approvals won't have the prompt restored, which is acceptable. Worth a one-line note in the PR description.

ilblackdragon:
## Code Review

### Overview
Surfaces OAuth `auth_url` prompts to the chat UI as soon as `tool_auth`/`tool_activate` returns one, even when `awaiting_token=false` (the pure-OAuth case Gmail hits). The former path only fired for the token-paste flow. Also preserves the first auth prompt across an approval pause in the same turn, so it isn't lost if another tool call needs user approval.

**Key pieces:**
- New `PendingAuthPrompt` in `session.rs` (validating constructor) and a new field on `PendingApproval`, serde-defaulted for backward compat.
- Expanded `ParsedAuthData` (adds `extension_name`, `instructions`, `awaiting_token`).
- New helpers: `extract_auth_prompt`, `capture_auth_prompt` (first-wins), `persist_/restore_selected_auth_prompt`, `sanitize_auth_url`, `auth_instructions_or_default`.
- `check_auth_required` is now `#[cfg(test)]`-only; runtime uses the new helpers.
- `handle_auth_intercept` signature changed to receive `ParsedAuthData` (no re-parsing).
- `thread_ops.rs` resume path restores the persisted prompt, re-captures from the approved tool result, and emits before/after approval.
- Error branch in `handle_auth_response` now emits a generic `"Invalid token. Please try again."` via `auth_retry_message_for_error` instead of leaking `ValidationFailed(...)` text.

### Correctness
- **First-wins ordering is correct.** `capture_auth_prompt` bails if `selected.is_some()`, matching the old `deferred_auth` semantics. Tests cover both OAuth-then-token and token-then-OAuth ordering.
- **Approval resume path is the subtle one, and it's handled.** `thread_ops.rs` restores via `restore_selected_auth_prompt`, re-captures from `pending.tool_name` + `tool_result`, and re-persists on the new `PendingApproval` if another approval pause occurs. A chain of approvals won't drop the prompt.
- **Possible double-emit UX quirk.** In both `dispatcher.rs` and `thread_ops.rs`, when approval is needed the `AuthRequired` status is emitted *and* the prompt is persisted on `PendingApproval`. On resume after approval resolves, the final `selected_auth_prompt` block emits it again. The user may see the same auth card twice in one turn (first alongside the approval card, then after approval is granted). Probably fine — keeps the card visible — but worth verifying in an E2E that the web UI dedupes or updates in place rather than stacking.
- **`awaiting_token=false` branch does NOT enter auth mode**, by design: OAuth callback completes out-of-band. Correct, but means if the extension returns `auth_url` *and* later expects a pasted token, this flow can't handle it. Given the "first-pass" label this is acceptable.

### Security
- **`sanitize_auth_url` is the key defense** and it's solid — requires `https://` prefix, rejects `javascript:`, `file:`, `data:`, `http:`, empty, `None`. Directly tested. Applied to both `auth_url` and `setup_url`.
- Minor: `starts_with("https://")` is case-sensitive, so `HTTPS://...` is rejected. Fail-closed, not exploitable — leave it.
- `sanitize_auth_url` doesn't parse/validate host, so `https://<garbage>` passes through. Source is a trusted local tool result, so acceptable. Would be worth an inline comment noting host validation is explicitly out of scope.
- **Generic `"Invalid token. Please try again."` is a real improvement** — previously `ValidationFailed("wrong format for API key")` etc. leaked straight back to the user. The underlying error is logged at `debug!` (correct per `CLAUDE.md` — not `info!`).

### Project conventions
- Uses `debug!` for the rejection-trace (correct, since `info!` would corrupt REPL/TUI).
- Serde additions all use `#[serde(default)]`, preserving backward compatibility with persisted `PendingApproval` rows. The existing `deserialize_without_deferred_tool_calls` test was extended to assert `selected_auth_prompt.is_none()`.
- No `unwrap`/`expect` introduced in production paths.
- Test-only helper gated with `#[cfg(test)]` — clean.

### Style nits
- `PendingAuthPrompt`'s doc comment claims "direct struct construction is only permitted in deserialization and test code," but the fields are `pub` and there's no `#[non_exhaustive]`, so it's advisory only. Either make fields `pub(crate)` or accept the comment as a convention note.
- The dispatcher's pre-approval emit block inlines a `send_status(StatusUpdate::AuthRequired { ... })` call while `thread_ops.rs` uses the `emit_auth_prompt` helper. Minor DRY miss — the dispatcher could call a shared helper too.
- `normalize_extension_name` + `.trim().to_owned()` in `persist_selected_auth_prompt` + `PendingAuthPrompt::new`'s trim-check is three places normalizing the same thing. Consolidating into `PendingAuthPrompt::new` (trim internally and store the trimmed form) would eliminate the duplication.

### Test coverage
Strong, and notably hits the "test through the caller" rule from `CLAUDE.md`:
- `test_need_approval_persists_first_auth_prompt_for_resume` drives `run_agentic_loop` end-to-end with an `AuthThenApprovalProvider` + `OAuthPromptTool` + `ApprovalTool`, verifying the OAuth prompt survives into `PendingApproval`. Right tier — not just a helper unit test.
- `test_sanitize_auth_url_rejects_non_https_schemes` — covers `javascript:`, `file:`, `http:`, `data:`, empty, None.
- `test_parse_auth_result_strips_non_https_urls` — integration of sanitize with parse.
- `test_capture_auth_prompt_keeps_first_*` — both ordering directions.
- `test_restore_selected_auth_prompt_rejects_blank_extension_name` + `test_pending_auth_prompt_new_rejects_empty_name` — constructor invariant.
- `test_auth_retry_message_hides_validation_details` — new security behavior.
- `test_extract_auth_prompt_detects_oauth_link_without_awaiting_token` — the core regression the PR fixes.

Missing:
- No test for the dispatcher's new "emit auth prompt alongside approval card" inline block specifically (the test covers persistence but not the `send_status` side effect on the approval-pause path). A channel-mock assertion would close the loop. Low priority.
- No web E2E in the diff (`tests/e2e/`) for the UI actually rendering the OAuth card from a `tool_activate` result. Given the "first-pass" label, acceptable, but worth a follow-up.

### Verdict
**Approve with minor nits.** The fix is well-scoped, the security posture is better than before (URL sanitization + generic error message), backward compatibility is preserved, and test coverage is at the right tier. Follow-ups: confirm the UI doesn't double-render the card when approval precedes the OAuth prompt, and consider consolidating the dispatcher/thread_ops emit paths through a shared helper.

ilblackdragon (approved):

PR #2040, opened 2026-04-05
fix (skills) installs for invalid catalog names

### LINKED ISSUE #1999: Skill installation fails for skill names containing spaces
Description: When attempting to install the "Mortgage Calculator" skill from the
Skillhub, the installation fails with a name validation error:

  "Install failed: Invalid skill name 'Mortgage Calculator': must match
   [a-zA-Z0-9][a-zA-Z0-9._-]{0,63}"

The skill name as published in the Skillhub contains a space, which violates the
internal skill name format. This is either a publishing-time validation gap (the
Skillhub allowed a name that cannot be installed) or the install path is not
normalizing the name before validation.

Impact: MEDIUM — Any skill published to the Skillhub with a space in the name is
permanently uninstallable. Users see an opaque technical error with no remediation
path.

serrrfirat:
## Summary
- recover installs for skills whose published manifest names are invalid by rewriting to a safe internal name during install
- resolve catalog display names to real slugs for both tool and web install paths, rejecting ambiguous matches and preserving owner-aware normalized names
- fix installed-state matching in catalog search results so the UI marks installed skills correctly without false positives

## Testing
- cargo fmt --all
- cargo test -p ironclaw_skills test_resolve_catalog_slug_for_name_ -- --nocapture
- cargo test -p ironclaw_skills test_catalog_entry_is_installed_ -- --nocapture
- cargo test --lib install_requested_identifier_prefers_resolved_slug_for_manual_name_installs -- --nocapture

Closes #1999


ilblackdragon (commented):
## Automated review — Needs changes (one high-severity fix)

**TL;DR:** Solid recovery mechanism for invalid catalog names with good coverage, but contains an unsafe \`unwrap_or_default()\` on critical data at \`skills.rs:209\` that should become \`.expect()\`. Merge-ready after that fix.

### Findings

| 1 | **High** | \`src/channels/web/handlers/skills.rs:209\` | \`unwrap_or_default()\` on critical \`resolved_download_key\`. The code path guarantees \`Some(...)\` on line 206; silently falling back to an empty string masks invariant violations and will produce a broken download URL | Replace with \`.expect("download key guaranteed Some by line 206")\`. Fail loud instead of silently producing a bad URL. Or restructure to avoid the \`Option\` round-trip entirely. |
| 2 | Medium | \`crates/ironclaw_skills/src/catalog.rs:69, 86\` | \`rsplit('/').next().unwrap_or()\` pattern repeats in hot paths (\`catalog_entry_is_installed\`) | Extract \`fn slug_suffix(slug: &str) -> &str\` helper |
| 3 | Low | \`crates/ironclaw_skills/src/parser.rs:244+\` | Conditional name validation via \`parse_skill_md_impl(validate_name: bool)\`. \`parse_skill_md_without_name_validation()\` is only for the install-recovery path but nothing in the name or docs prevents misuse | Document the restriction. Consider a more explicit name like \`parse_skill_md_for_install_recovery()\`, or make the unchecked variant crate-private |

### Trust boundary review
No weakening detected:
- Installed skills (trusted, full tools) stored with normalized safe names
- Catalog entries (untrusted) resolved client-side before download
- \`normalize_install_content()\` accepts arbitrary published content but rewrites manifest names before disk persistence
- Web handler correctly calls \`resolve_catalog_slug_for_name()\` before constructing the download URL
- No access-control logic changed

### Attenuation
Untouched. Catalog search results and install flows use existing gating (\`X-Confirm-Action\` header required, catalog metadata remains read-only).

### Open questions
1. Were the new tests (\`test_resolve_catalog_slug_for_name_*\`, \`test_catalog_entry_is_installed_*\`) actually executed before merge? They may be behind a feature gate.
2. Has the \`installed_names\` matching in \`app.js:5737\` been tested end-to-end? It falls back to client-side normalization if \`entry.installed\` is missing from the API response.
3. Should \`normalize_catalog_identity()\` (catalog.rs:76, module-private) and \`normalize_skill_identifier()\` be unified? Both strip non-alphanumerics and lowercase.

serrrfirat:
Addressed the latest review feedback in `8573a04e`.

Changes:
- replaced the silent `unwrap_or_default()` in the web install handler with an explicit `expect(...)` for the resolved download key invariant
- extracted a shared `slug_suffix()` helper in `catalog.rs` to remove the repeated suffix parsing logic
- renamed the unchecked parser entrypoint to `parse_skill_md_for_install_recovery()` and documented that it is restricted to the install-recovery path

Verification run locally:
- `cargo fmt --all`
- `cargo test -p ironclaw_skills test_resolve_catalog_slug_for_name_ -- --nocapture`
- `cargo test -p ironclaw_skills test_catalog_entry_is_installed_ -- --nocapture`

I also started `cargo test --lib install_requested_identifier_prefers_resolved_slug_for_manual_name_installs -- --nocapture`, but it did not complete within this session, so I’m not claiming that one passed.

ilblackdragon (commented):
**Verdict: Approve with one required change**

- **`src/channels/web/handlers/skills.rs:~220`** — uses `.expect("download key guaranteed Some above")` in production code. Violates CLAUDE.md ("no `.unwrap()`/`.expect()` in production code"). Trivially fixable by hoisting the local before the round-trip through `Option`.
- `crates/ironclaw_skills/src/registry.rs` — `render_skill_md` strips `---\n` from `serde_yml::to_string` output, but that crate doesn't emit a leading `---` for structs. Dead code; remove or comment.
- `catalog_entry_is_installed` correctly fixes the `calculator` vs `mortgage-calculator` false-positive.
- `resolve_catalog_slug_for_name` correctly prefers exact display-name match before normalized fuzzy match and surfaces ambiguity as an error rather than silently picking.
- Strong test coverage: ambiguity, owner-aware normalization, partial-suffix non-match, install round-trip with rewritten YAML. No DB changes.

serrrfirat (commented):
Both items from the latest review are addressed in `c4f6ea3e`:

1. **`.expect()` removed** — hoisted `download_key` so it's used directly for the URL before being moved into the `Option`. No more `.expect()` or `.unwrap_or_default()` in production.
2. **Dead `strip_prefix("---\n")` removed** — `serde_yml::to_string` doesn't emit a leading `---` for structs, so the stripping was unreachable. Removed.

Also merged latest staging (`50c8e3e7`) to resolve conflicts (imports in `skill_tools.rs` + I18n strings in `app.js`). All 145 skills crate tests pass.

ilblackdragon:
## Code Review

**What it does:** Recovers skill installs when published manifests have invalid `name` fields by slugifying to a safe internal name, resolves display names → unique catalog slugs (rejecting ambiguity), and fixes the installed-state badge in catalog search.

### Key issues
- `crates/ironclaw_skills/src/registry.rs:311-331` — `render_skill_md` only re-serializes known `SkillManifest` fields. Any extra YAML front-matter keys on a published skill are silently dropped on rewrite. Round-trip through `serde_yml::Value` and mutate only `name`.
- `crates/ironclaw_skills/src/catalog.rs:91-95` — `catalog_entry_is_installed` compares `name == installed` case-sensitively while slug comparisons use `eq_ignore_ascii_case`. Inconsistent; display-name casing differences won't match.
- `registry.rs:354` — `tracing::warn!` in install path; prefer `debug!` (REPL/TUI rule).
- `skills.rs:741` and `skill_tools.rs:947` — duplicate per-install catalog `search()` round trips; extract a shared helper.
- `validation.rs:495-537` — `normalize_skill_identifier` strips non-ASCII unicode (e.g. "café" → "caf"); document.

### Test gaps
Unicode names, empty/whitespace input, `render_skill_md` round-trip, normalized-name collisions across origins, no integration test for the full web install path with an invalid published name.

**Verdict:** Ship-worthy with minor fixes. The `render_skill_md` field-loss is the only blocker.

ilblackdragon (approved):
Approving — render-rewrite now preserves unknown frontmatter, install-state matching is case-insensitive, and unicode normalization is documented. Pushed the fixes directly. The duplicate catalog-search dedupe between skill_tools.rs and skills.rs is left as a follow-up.

PR #2080, opened 2026-04-06
fix(ci): skip NearAI URL DNS validation for non-NearAI backends

zmanian:
## Summary

- **Root cause**: `LlmConfig::resolve()` unconditionally called `validate_base_url()` on NearAI default URLs (`private.near.ai`), which performs DNS resolution. In CI (and any environment without access to `private.near.ai`), this fails even when a completely different LLM backend is configured.
- **Fix**: Only validate `NEARAI_AUTH_URL` and `NEARAI_BASE_URL` when NearAI is the active backend or the user explicitly set those env vars.
- **Regression test**: `non_nearai_backend_skips_nearai_url_validation` — configures `openai_compatible` backend without any NearAI env vars and asserts `resolve()` succeeds.
- **E2E cleanup**: Removed redundant `@pytest.mark.asyncio` decorators from `test_webhook.py` (project uses `asyncio_mode = "auto"`).

This has been blocking Staging CI since run #628 (April 2) — over 100 consecutive failures.

## Test plan

- [x] `cargo fmt -- --check` clean
- [x] `cargo clippy --all --benches --tests --examples --all-features -- -D warnings` clean
- [x] Regression test passes: `cargo test non_nearai_backend_skips_nearai_url_validation`
- [x] Manual server startup returns healthy response with the fix applied
- [ ] CI passes on this branch

https://claude.ai/code/session_01FybyQXiX2HDhaGizxr2PFC


src/config/llm.rs
```diff
@@ -2191,4 +2239,34 @@ mod tests {
...
+    fn non_nearai_backend_skips_nearai_url_validation() {
+        let _guard = lock_env();
+        // SAFETY: Under ENV_MUTEX — clear any NearAI vars so defaults kick in.
+        unsafe {
+            std::env::remove_var("LLM_BACKEND");
+            std::env::remove_var("NEARAI_AUTH_URL");
+            std::env::remove_var("NEARAI_BASE_URL");
+            std::env::remove_var("NEARAI_API_KEY");
+            std::env::remove_var("NEARAI_MODEL");
+        }
```
serrrfirat, line 2302: **Medium Severity** — Regression test fragile to stray env vars

The test clears `NEARAI_*` env vars but doesn't clear `EMBEDDING_ENABLED` or `EMBEDDING_PROVIDER`. If `EMBEDDING_ENABLED=true` happens to be set in the test environment (e.g. a developer's shell, a CI matrix variant), then `nearai_embeddings_active` becomes `true`, the validation fires on the default `private.near.ai` URL, and the test fails — defeating its purpose.

Other LLM config tests in this file clear related vars in their setup blocks.

**Suggested fix:**
```rust
unsafe {
    std::env::remove_var("LLM_BACKEND");
    std::env::remove_var("NEARAI_AUTH_URL");
    std::env::remove_var("NEARAI_BASE_URL");
    std::env::remove_var("NEARAI_API_KEY");
    std::env::remove_var("NEARAI_MODEL");
+   std::env::remove_var("EMBEDDING_ENABLED");
+   std::env::remove_var("EMBEDDING_PROVIDER");
}
```

src/config/llm.rs
```diff
@@ -132,24 +134,62 @@ impl LlmConfig {
...
+        let nearai_auth_url_explicit = optional_env("NEARAI_AUTH_URL")?;
+        let nearai_auth_url = nearai_auth_url_explicit
+            .clone()
             .unwrap_or_else(|| "https://private.near.ai".to_string());
-        validate_base_url(&nearai_auth_url, "NEARAI_AUTH_URL")?;
+        // Only validate NearAI URLs when NearAI is active or the user explicitly
+        // set the URL.  Default URLs point to private.near.ai which requires DNS
+        // resolution — this blocks startup in environments without network access
+        // (CI runners, containers) when a different backend is configured.
+        if is_nearai
```
serrrfirat, line 177: **🟡 Medium: Asymmetric SSRF validation predicates for NearAI URLs**

The `NEARAI_AUTH_URL` validation predicate does NOT include `nearai_api_key.is_some()`, but the `NEARAI_BASE_URL` validation predicate DOES (line 216).

If a user sets `NEARAI_API_KEY` (e.g., for embeddings via direct API key), the base URL will be validated but the auth URL will not. While `auth_base_url` is typically only used for OAuth flows, an unvalidated SSRF-eligible URL stored in `SessionConfig` is a defense-in-depth gap.

**Suggested fix:** Add `nearai_api_key.is_some()` to the `NEARAI_AUTH_URL` validation predicate for consistency, or document why the asymmetry is intentional.

serrrfirat:
## 🏗️ Paranoid Architect Review — PR #2080

**Verdict: ✅ APPROVE with asymmetry fix**

### Summary
| Severity | Count |
|----------|-------|
| Medium | 1 |

### Assessment
The fix is correct for the stated problem — conditionally gating NearAI URL validation behind predicates that check whether NearAI is actually in use. The predicates are comprehensive and cover the main scenarios (primary backend, explicit env vars, DB overrides, embeddings).

One medium concern: asymmetric validation predicates where `NEARAI_AUTH_URL` validation doesn't include the `nearai_api_key.is_some()` check that `NEARAI_BASE_URL` has. This is a defense-in-depth gap.

The E2E test cleanup (removing redundant `@pytest.mark.asyncio` decorators) is clean and correct.

ilblackdragon (dismissed):
## Code Review

### Overview
Fixes a CI blocker (100+ consecutive failures since April 2): `LlmConfig::resolve()` unconditionally called `validate_base_url()` on NearAI default URLs (`private.near.ai`), which performs DNS resolution. In CI and air-gapped environments without access to `private.near.ai`, this fails even when a completely different LLM backend is configured.

The fix makes validation conditional: only validate NearAI URLs when NearAI is actually in use (primary backend, embeddings, explicit env vars, or DB overrides).

### Correctness

The validation predicate is well-reasoned:

**Auth URL** validated when:
- NearAI is primary backend (`is_nearai`), OR
- User explicitly set `NEARAI_AUTH_URL`, OR
- DB builtin_overrides include a NearAI base_url, OR
- NearAI embeddings are enabled

**Base URL** validated when: all of the above, PLUS:
- `NEARAI_API_KEY` is set (implies intent to use NearAI, and key presence selects `cloud-api.near.ai`)

The asymmetry makes sense — auth_url is only for OAuth/session flow (not used with API keys), while base_url is used for all NearAI requests including API-key auth.

The embeddings check is important: NearAI can serve embeddings even when another provider handles chat. The code correctly gates on both `EMBEDDING_ENABLED` and `EMBEDDING_PROVIDER == "nearai"` (case-insensitive).

### Code Quality

- No `.unwrap()` / `.expect()` in production code
- `nearai_override` moved earlier in the function to be available for the predicate — the computation is unchanged, just reordered
- `.clone()` on `Option<String>` for the explicit-check variables is cheap and necessary (value used in two places)
- Well-documented with a comment block explaining all four conditions

### Test Coverage

The regression test is appropriate:
- Clears all NearAI env vars, configures `openai_compatible` backend
- Asserts `resolve()` succeeds without DNS resolution on `private.near.ai`
- Protected by `lock_env()` mutex (env var tests must be serialized)
- Uses `unsafe { std::env::remove_var(...) }` correctly (Rust 1.78+ API safety requirement)

The E2E cleanup (removing redundant `@pytest.mark.asyncio`) is correct — `asyncio_mode = "auto"` is set globally.

### Minor Observations

| Severity | Note |
|----------|------|
| **Nit** | The validation predicate has 4-5 conditions per URL. It's clear with the comment, but if this grows further, consider extracting a `fn should_validate_nearai_urls(...)` helper. |
| **Nit** | The unused `pytest` import in `test_webhook.py` is also cleaned up — good. |

### Verdict: **Approve** ✓

Clean, well-targeted fix for a CI blocker. The conditional logic correctly covers all the cases where NearAI URLs are actually needed. The regression test directly reproduces the failure scenario.

src/config/llm.rs
ilblackdragon: Fixed — added `EMBEDDING_ENABLED` and `EMBEDDING_PROVIDER` to the test's env var cleanup, with a comment explaining why.

src/config/llm.rs
ilblackdragon: Fixed — added `nearai_api_key_env.is_some()` to the `NEARAI_AUTH_URL` validation predicate so both URLs are validated symmetrically when an API key is present. Updated the predicate comment to document this.

ilblackdragon (approved):

PR #2086, opened 2026-04-06
resolve an issue with shared skills

### LINKED ISSUE #2084: bug(engine-v2): skills not visible to gateway users after resolve_user_project was introduced
## Summary

Gateway users (web UI, Telegram, API clients) cannot see any skills. `__list_skills__()` always returns an empty list for them, so no skills are auto-injected at step 0 and the agent ignores user-installed skills entirely.

Owner/CLI users are unaffected.

## When it was introduced

Introduced in the staging merge into `demo/abound` on **2026-04-06** (commit `a9470371`), which pulled in the multi-tenant isolation work that added `resolve_user_project` to `handle_with_engine_inner` in `src/bridge/router.rs`.

The specific change that caused the regression:

**Before** (`8615aab1` and earlier):
```rust
// All threads used the owner's default project
state.conversation_manager.handle_user_message(
    conv_id, content,
    state.default_project_id,  // ← owner's project, where skills live
    &message.user_id,
    ThreadConfig::default(),
)
```

**After** (current):
```rust
// Gateway users get their own per-user project
let project_id =

state.conversation_manager.handle_user_message(
    conv_id, content,
    project_id,  // ← user's project, which has no skills
    &message.user_id,
    ThreadConfig::default(),
)
```

## Root Cause

Skills are migrated into the **owner's default project** at startup:

```rust
// src/bridge/router.rs
migrate_v1_skill_list(&skills, &store, project_id)  // project_id = owner's default
// → MemoryDoc { project_id: aaf9bf1a, user_id: "__shared__", doc_type: Skill }
```

`handle_list_skills` in the orchestrator looks up skills using the **thread's project_id**:

```rust
store.list_memory_docs_with_shared(thread.project_id, &thread.user_id)
// → list_memory_docs(user_project_5e8b, "__shared__")  → []
// Skills are in aaf9bf1a, not 5e8b66ad → not found
```

`list_shared_memory_docs` filters by both `project_id` AND `user_id == "__shared__"`. Since the gateway user's project (`5e8b66ad`) is different from the owner's project (`aaf9bf1a`), skills are never returned.

A fresh database does **not** fix this — the code creates the mismatch on every startup.

## Design gap

`"__shared__"` as `user_id` was intended to mark content visible to all users, but `list_shared_memory_docs` scopes the lookup to a single `project_id`. Skills are cross-project by nature (admin-installed, visible to everyone) but the current design only shares within one project.

## Impact

- All `ENGINE_V2=true` deployments with gateway users
- Skills are never auto-selected and injected at step 0
- Agent works around this by calling `skill_search` (registry) and `skill_list`, wasting 2–5 steps and never actually running skill code
- Confirmed on `demo/abound` branch

## Fix Options

**Option A (minimal)** — Pass `default_project_id` through `ThreadConfig` as `skill_fallback_project`. In `handle_list_skills`, fall back to it when the user's project returns no skills.

**Option B (correct)** — Add a `list_skills_global()` method to the `Store` trait that returns all `DocType::Skill` docs owned by `shared_owner_candidates()` across all projects, ignoring `project_id`. Skills are inherently cross-project and the storage layer should reflect that.

**Option C** — Migrate skills under `ProjectId(Uuid::nil())` as a permanent system project, and always include the nil project in `list_shared_memory_docs`.

Option B or C is the correct long-term fix. Option A is a minimal patch.

## Verification

Gateway user request after regression:
```
DEBUG orchestrator: host function call action=__list_skills__
DEBUG orchestrator: host function call action=__emit_event__    ← step_started only, no skill_activated
DEBUG orchestrator: host function call action=__llm_complete__
DEBUG Tool call started tool=skill_search  ← model searching registry instead
```

Default (owner) user request — skills work:
```
DEBUG orchestrator: host function call action=__list_skills__
DEBUG orchestrator: host function call action=__emit_event__    ← step_started
DEBUG orchestrator: host function call action=__emit_event__    ← skill_activated ✓
DEBUG orchestrator: host function call action=__llm_complete__
```

zetyquickly:
## Summary

- `__list_skills__` was scoping shared-skill lookup to the thread's `project_id`, so
  admin-installed skills (stored in the admin's project) were invisible to gateway users
  whose threads run in per-user projects
- Added `list_memory_docs_by_owner(user_id)` to the `Store` trait — fetches all docs for
  a user across all projects (no project scope)
- Added `list_skills_global()` default impl on `Store` that uses the above to collect
  skills from all `shared_owner_candidates()` regardless of project
- `__list_skills__` now merges user-scoped docs with the global shared set, then deduplicates

## Change Type

- [x] Bug fix

## Linked Issue

Fixes #2084

## Validation

- [x] `cargo build`
- Relevant tests pass: existing skill listing tests in `lib.rs` and `mission.rs`
- Manual testing: verified shared skills installed under admin owner are visible to a
  gateway user whose thread runs in a different per-user project

## Security Impact

None. No changes to permissions, tool execution, or access control — this fixes visibility
of already-accessible shared skills.

## Database Impact

None. `HybridStore` is the only `Store` impl; no schema changes or migrations.

## Blast Radius

- `crates/ironclaw_engine/src/traits/store.rs` — `Store` trait gains a new required method;
  all downstream mock impls updated
- `__list_skills__` handler in `orchestrator.rs` — skill listing behavior changes for
  multi-tenant deployments (intended)
- No impact on memory retrieval, context injection, or other `Store` callers

## Rollback Plan

Revert commit `f758107`. `__list_skills__` returns to using `list_memory_docs_with_shared`
scoped to `thread.project_id`. Admin skills won't be visible cross-project again but no
data is at risk.

## Review Follow-Through

- Double sort+dedup: `list_skills_global` sorts/deduplicates internally, then the caller
  in `orchestrator.rs` does it again after extending. Harmless but could be cleaned up.
- Reviewer should confirm there are no other `Store` implementations outside this repo
  (e.g. in integration test harnesses) that now need `list_memory_docs_by_owner` stubs.

---

**Review track**: B (feature/maintainer-requested refactor)


crates/ironclaw_engine/src/traits/store.rs
```diff
@@ -129,6 +129,38 @@ pub trait Store: Send + Sync {
...
+
+    /// List all memory docs owned by a specific user across ALL projects.
+    ///
+    /// Default implementation fans out over `list_projects` + `list_memory_docs`.
+    /// Implementors with a flat store should override for efficiency.
+    async fn list_memory_docs_by_owner(
+        &self,
+        user_id: &str,
+    ) -> Result<Vec<MemoryDoc>, EngineError> {
+        let projects = self.list_projects(user_id).await?;
```
serrrfirat: **Medium Severity**\n\n`list_memory_docs_by_owner` default is documented as cross-project, but this implementation only iterates `list_projects(user_id)` and then scans those projects. That can miss docs whose `doc.user_id` matches but whose `project_id` belongs to another owner context (the multi-project layout this PR is trying to handle for shared skills).\n\nThis can silently reintroduce shared-skill visibility gaps for any `Store` implementation relying on the default.\n\nSuggested fix: make the default truly cross-project by iterating `list_all_projects()` and filtering by `doc.user_id == user_id` (with a fallback), or keep this method required so each store must implement it explicitly.

crates/ironclaw_engine/src/traits/store.rs
```diff
@@ -129,6 +129,38 @@ pub trait Store: Send + Sync {
...
+    /// lives in. Used by `__list_skills__` so that admin-installed skills
+    /// (migrated into the owner's default project) are visible to gateway users
+    /// whose threads run in a per-user project.
+    async fn list_skills_global(&self) -> Result<Vec<MemoryDoc>, EngineError> {
+        let mut docs = Vec::new();
+        for owner_id in shared_owner_candidates() {
+            docs.extend(self.list_memory_docs_by_owner(owner_id).await?);
+        }
+        docs.retain(|d| d.doc_type == crate::types::memory::DocType::Skill);
+        Ok(docs)
```
serrrfirat, line 153: **Low Severity**\n\n`list_skills_global()` no longer normalizes output (sort+dedup), while the old version did. The current caller (`__list_skills__`) dedups, so behavior is fine today, but this makes the trait method easier to misuse by future callers and can cause duplicate skills if reused elsewhere.\n\nSuggested fix: restore sort+dedup inside `list_skills_global()` (keeping caller-side dedup as defense in depth), or document the caller contract explicitly.

crates/ironclaw_engine/src/executor/orchestrator.rs
```diff
@@ -1694,18 +1694,29 @@ async fn handle_list_skills(
...
-            return ExtFunctionResult::Return(json_to_monty(&serde_json::json!([])));
+            debug!("__list_skills__: failed to load user docs: {e}");
+            vec![]
         }
     };
 
+    // Admin/shared skills across ALL projects (fixes multi-tenant visibility —
+    // shared skills live in the owner's project but must be visible to all users
+    // regardless of which per-user project their thread runs in).
+    match store.list_skills_global().await {
```
serrrfirat, line 1930: **Medium Severity**\n\nThere is no focused regression test for the new cross-project shared-skill visibility path in `__list_skills__`. This PR changes selection semantics in a tenant-sensitive path, but the touched tests mostly add trait stubs and do not assert the new behavior end to end.\n\nSuggested test: create a user thread in project A, place a shared/admin skill in project B, call the skill-list path, and assert that the shared skill is returned.

crates/ironclaw_engine/src/traits/store.rs
```diff
@@ -129,6 +129,38 @@ pub trait Store: Send + Sync {
...
+    /// Implementors with a flat store should override for efficiency.
+    async fn list_memory_docs_by_owner(
+        &self,
+        user_id: &str,
+    ) -> Result<Vec<MemoryDoc>, EngineError> {
+        let projects = self.list_projects(user_id).await?;
+        let mut docs = Vec::new();
+        for project in projects {
+            docs.extend(self.list_memory_docs(project.id, user_id).await?);
+        }
```
serrrfirat, line 175: **Medium Severity — Correctness**

**Default `list_memory_docs_by_owner` impl may silently miss docs for shared owners**

The default implementation fans out over `list_projects(user_id)` + `list_memory_docs(project.id, user_id)`. For shared owners like `__shared__` or `system`, `list_projects("__shared__")` only returns projects **owned by** `__shared__` — not projects where an admin installed shared docs under a different project owner. Those docs would be silently dropped.

The `HybridStore` override is correct (it scans all in-memory docs by `user_id` regardless of project), so production behavior is fine today. But the default is a landmine for any future `Store` implementor who relies on it.

**Suggested fix:** Add a doc comment warning that the default assumes all of a user's docs live in projects they own (which doesn't hold for shared owners), or make the default impl project-agnostic (e.g., query a broader set of projects, or require implementors to override).

crates/ironclaw_engine/src/executor/orchestrator.rs
```diff
@@ -1694,18 +1694,29 @@ async fn handle_list_skills(
...
+        Ok(shared) => docs.extend(shared),
+        Err(e) => debug!("__list_skills__: failed to load global skills: {e}"),
+    }
+
+    docs.sort_by_key(|d| d.id.0);
+    docs.dedup_by_key(|d| d.id);
+
     let skills: Vec<serde_json::Value> = docs
         .into_iter()
         .filter(|d| d.doc_type == crate::types::memory::DocType::Skill)
```
serrrfirat, line 1940: **Medium Severity — Test Coverage**

**No regression test for the cross-project skill visibility fix**

This PR fixes a specific bug: admin skills in project A are invisible to users whose threads run in project B. However, there's no new test that exercises this exact scenario. The existing tests pass because they don't test cross-project visibility.

Without a regression test, this bug could be silently reintroduced if someone reverts to project-scoped queries or refactors the skill listing path.

**Suggested fix:** Add a test (e.g., in `lib.rs` using `InMemoryStore`) that:
1. Creates a shared skill (`user_id = "__shared__"`) in project A
2. Calls `list_skills_global()` and asserts the skill is returned
3. Optionally: simulates the full `handle_list_skills` flow with a thread in project B and verifies the shared skill appears in the result

crates/ironclaw_engine/src/traits/store.rs
zetyquickly: Both addressed by the same fix: v1_skill_to_memory_doc was tagging all v1 skills as __shared__ regardless of source — that's what created the cross-project situation. Now User/Workspace skills migrate with user_id = owner_id, only Installed/Bundled as __shared__. Shared skills will always live in a __shared__-owned project going forward, so the default fan-out impl is correct and the landmine is defused.

crates/ironclaw_engine/src/executor/orchestrator.rs
zetyquickly: Both addressed by two regression tests added to crates/ironclaw_engine/src/lib.rs: list_skills_global_returns_shared_skills_from_different_project tests the store method directly with a shared skill in project B and a thread in project A. skill_merge_covers_user_and_shared_from_different_projects mirrors the full handle_list_skills merge logic and asserts both user and shared skills appear in the result.

crates/ironclaw_engine/src/traits/store.rs
zetyquickly: Both addressed by the same fix: v1_skill_to_memory_doc was tagging all v1 skills as __shared__ regardless of source — that's what created the cross-project situation. Now User/Workspace skills migrate with user_id = owner_id, only Installed/Bundled as __shared__. Shared skills will always live in a __shared__-owned project going forward, so the default fan-out impl is correct and the landmine is defused.

crates/ironclaw_engine/src/traits/store.rs
zetyquickly: Restored sort_by_key + dedup_by_key inside list_skills_global. Caller-side dedup in handle_list_skills stays as defense in depth.

crates/ironclaw_engine/src/executor/orchestrator.rs
zetyquickly: Both addressed by two regression tests added to crates/ironclaw_engine/src/lib.rs: list_skills_global_returns_shared_skills_from_different_project tests the store method directly with a shared skill in project B and a thread in project A. skill_merge_covers_user_and_shared_from_different_projects mirrors the full handle_list_skills merge logic and asserts both user and shared skills appear in the result.

crates/ironclaw_engine/src/traits/store.rs
```diff
@@ -129,6 +129,40 @@ pub trait Store: Send + Sync {
...
+        Ok(docs)
```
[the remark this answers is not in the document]
zetyquickly: addressed

src/bridge/skill_migration.rs
```diff
@@ -92,20 +94,20 @@ pub async fn migrate_v1_skill_list(
...
-            V2SkillSource::Migrated
-        }
-        SkillSource::Bundled(_) => V2SkillSource::Migrated,
+fn v1_skill_to_memory_doc(skill: &LoadedSkill, project_id: ProjectId, owner_id: &str) -> MemoryDoc {
+    // User- and workspace-installed skills belong to the owner.
+    // Bundled and registry-installed skills are shared across all users.
+    let user_id = match &skill.source {
+        SkillSource::User(_) | SkillSource::Workspace(_) => owner_id,
+        SkillSource::Installed(_) | SkillSource::Bundled(_) => shared_owner_id(),
     };
```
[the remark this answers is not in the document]
zetyquickly: addressed

crates/ironclaw_engine/src/traits/store.rs
```diff
@@ -129,6 +129,42 @@ pub trait Store: Send + Sync {
...
+    }
+
+    /// List all memory docs owned by a specific user across ALL projects.
+    ///
+    /// Default implementation fans out over ALL projects via `list_all_projects`.
+    /// This is necessary for shared owners (`__shared__`, `system`) who don't own
+    /// any projects themselves but have docs stored inside other users' projects.
+    /// Implementors with a flat store should override for efficiency.
+    async fn list_memory_docs_by_owner(
```
serrrfirat, line 167: **Medium Severity — Silent failure for non-overriding Store impls**

The default `list_memory_docs_by_owner` calls `self.list_all_projects()`, which also has a default impl that returns `Err(EngineError::Store { ... })`. This creates a silent failure chain:

```
list_skills_global()
  → list_memory_docs_by_owner() [default]
    → list_all_projects() [default]
      → Err("not implemented")
```

Since `handle_list_skills` catches this error with `Err(e) => debug!(...)`, the result is **silently missing shared skills** — no crash, no visible error, just an empty shared-skills list.

Any `Store` impl that doesn't override `list_memory_docs_by_owner` AND `list_all_projects` hits this. Currently that includes the integration test stores in `engine_v2_gate_integration.rs` and `engine_v2_skill_codeact.rs`.

**Suggested fix:** Consider making `list_memory_docs_by_owner` a required method (no default impl) so the compiler forces all implementors to provide one. Alternatively, add a prominent doc comment warning about the `list_all_projects` dependency.

crates/ironclaw_engine/src/traits/store.rs
```diff
@@ -129,6 +129,42 @@ pub trait Store: Send + Sync {
...
+    ///
+    /// Unlike `list_shared_memory_docs`, this ignores `project_id` — it returns
+    /// every `DocType::Skill` with a shared owner regardless of which project it
+    /// lives in. Used by `__list_skills__` so that admin-installed skills
+    /// (migrated into the owner's default project) are visible to gateway users
+    /// whose threads run in a per-user project.
+    async fn list_skills_global(&self) -> Result<Vec<MemoryDoc>, EngineError> {
+        let mut docs = Vec::new();
+        for owner_id in shared_owner_candidates() {
+            docs.extend(self.list_memory_docs_by_owner(owner_id).await?);
```
serrrfirat, line 147: **Medium Severity — Integration test Store impls missing `list_memory_docs_by_owner`**

The `TestStore` implementations in `tests/engine_v2_gate_integration.rs` (line 312) and `tests/engine_v2_skill_codeact.rs` (line 179) don't implement `list_memory_docs_by_owner`. Neither implements `list_all_projects`.

When any test exercises `__list_skills__`, the call to `list_skills_global()` here silently fails (error caught by the caller in orchestrator.rs), meaning:
- Tests pass ✅
- But the new cross-project shared-skills behavior is **completely untested** in integration ❌

**Suggested fix:** Add `list_memory_docs_by_owner` implementations to both integration test stores. At minimum `Ok(vec![])` stubs; ideally a real impl that filters the test store's docs by `user_id` so integration tests can verify shared skill visibility across projects.

src/bridge/router.rs
```diff
@@ -770,6 +770,7 @@ pub async fn init_engine(agent: &Agent) -> Result<(), Error> {
                 &skills_snapshot,
                 &store_dyn,
                 project_id,
+                owner_id,
```
serrrfirat, line 1144: **Medium Severity — Pre-existing issue exposed by this PR: legacy migration overwrites shared ownership**

Now that `owner_id` is passed through to the migration (good change), there's a subtle interaction with the existing `migrate_legacy_user_ids` function (router.rs:3220).

**The issue:** On upgrade from a pre-PR deployment, existing persisted docs lack `user_id` in frontmatter → deserialized as \`\"legacy\"\` → `migrate_legacy_user_ids` stamps them ALL with \`owner_id\`. This includes Bundled/Installed skills that should remain \`__shared__\`. The V1 migration then sees them (hash match via \`list_memory_docs(project_id, owner_id)\`) and skips them, cementing incorrect ownership.

**Impact:** \`list_skills_global()\` looks for \`__shared__\`/\`system\` owners → won't find these now-user-owned skills. For single-user HybridStore deployments this is harmless (skills appear via user's own docs). For any future multi-tenant Store, previously shared skills would be invisible to other users after upgrade.

**Suggested fix:** In \`migrate_legacy_user_ids\`, check doc metadata for skill source indicators (\`V2SkillSource::Migrated\` + Bundled/Installed source) before assigning ownership. Docs with shared-owner semantics should be stamped as \`shared_owner_id()\`, not \`owner_id\`.

ilblackdragon:
# Code Review

## Overview
Fixes a multi-tenant visibility bug where admin-installed skills (stored under `__shared__` in the admin's project) were invisible to gateway users whose threads run in per-user projects. Two related fixes:

1. **Global skill listing** — new `Store::list_skills_global()` ignores `project_id` and aggregates skills across all projects via `list_memory_docs_by_owner(user_id)`.
2. **Migration ownership** — `migrate_v1_skill_list` now takes an `owner_id`. User/Workspace skills migrate as the owner; Bundled/Installed migrate as `__shared__`. Previously *all* v1 skills were stamped as `__shared__`, which masked the visibility bug.

Plus a serialization tweak in `store_adapter.rs`: knowledge-doc frontmatter now persists `project_id` and `user_id`.

## Strengths
- Root cause is correctly identified — `__list_skills__` was scoping to `thread.project_id`, which broke the moment shared skills lived in a different project than the caller's thread.
- The trait default implementation has a sensible fallback (`list_all_projects` → fan-out) and `HybridStore` provides the efficient override.
- Migration fix is the right call: existing duplicate-detection now compares against both shared *and* owner-scoped docs, preventing re-migration loops on startup.
- New tests cover the bundled/installed → shared mapping and the global-listing path.
- Failure-mode change in `handle_list_skills` is a small improvement: a failed user-doc fetch no longer wipes the entire skill list — it falls through to global skills.

## Issues & Suggestions

### 1. Test-through-the-caller rule violation (medium)
`CLAUDE.md` is explicit: *\"When a predicate/classifier/transform helper gates a side effect ... a unit test on the helper alone is not sufficient regression coverage.\"* The PR adds `skill_merge_covers_user_and_shared_from_different_projects`, but the comment admits it *\"mirrors the merge logic in `handle_list_skills` without going through the private orchestrator function.\"* That is exactly the pattern the rule prohibits — if `handle_list_skills` later regresses (e.g., someone reverts to `list_memory_docs_with_shared`), the test will still pass. Recommend exposing a thin wrapper or adding an integration-tier test that drives `handle_list_skills` end-to-end.

### 2. Backward-compat risk in `serialize_knowledge_doc` (medium)
`deserialize_knowledge_doc` falls back to `ProjectId(nil)` and `user_id = \"legacy\"` when frontmatter is missing those fields. Any existing on-disk knowledge docs written before this PR will silently load with `user_id = \"legacy\"` — meaning they will *not* show up in either `list_memory_docs(project, real_user)` or `list_memory_docs_by_owner(\"__shared__\")`. Worth confirming whether docs already exist in the wild and whether a one-shot migration (or at least a `warn!` on fallback) is needed.

### 3. N+1 in the default `list_memory_docs_by_owner` (low)
The default impl calls `list_all_projects()` then `list_memory_docs(project_id, user_id)` per project. For deployments with many projects, `list_skills_global` becomes O(projects × shared_owners) round-trips. `HybridStore` overrides it (good), but the doc comment should make explicit that every production `Store` impl should override — otherwise this turns into a hot-path bottleneck on `__list_skills__`, which fires per-thread.

### 4. Double sort+dedup (cosmetic, author-acknowledged)
`list_skills_global` sorts/dedups internally; `handle_list_skills` then sorts/dedups the merged vector again. The second pass is necessary (user docs may overlap shared ones), but the first sort inside `list_skills_global` is wasted work. Drop the inner sort and keep only the dedup-by-id.

### 5. `list_skills_global` name vs. behavior (minor)
The method filters to `DocType::Skill` after fetching all owner docs. Consider whether `list_memory_docs_by_owner` should accept a `DocType` filter so the filter happens at the store level (saves serialization for non-skill shared docs in `HybridStore`).

### 6. `migrate_v1_skill_list` signature change is a breaking API change
The function is `pub` in `src/bridge/skill_migration.rs`. The single in-tree caller (`router.rs`) is updated, but confirm there are no external consumers (CLI tooling, downstream test harnesses).

## Security
No new attack surface. The change *expands* visibility of skills that were already meant to be visible — it does not bypass any access control. The shared-owner candidate set (`shared_owner_candidates()`) still gates which user IDs count as \"shared\".

## Verdict
**Approve with changes.** The core fix is correct and well-motivated. Before merging:
- Add a test that drives `handle_list_skills` directly (CLAUDE.md rule).
- Decide on the legacy-frontmatter migration path or at least add a `warn!` on the fallback.
- Document the N+1 caveat in the trait doc comment.

crates/ironclaw_engine/src/traits/store.rs
```diff
@@ -129,6 +129,28 @@ pub trait Store: Send + Sync {
...
+        }
+        docs.retain(|d| d.doc_type == crate::types::memory::DocType::Skill);
+        docs.sort_by_key(|doc| doc.id.0);
+        docs.dedup_by_key(|doc| doc.id);
+        Ok(docs)
+    }
+
+    /// List all memory docs owned by a specific user across ALL projects.
+    async fn list_memory_docs_by_owner(&self, user_id: &str)
+    -> Result<Vec<MemoryDoc>, EngineError>;
```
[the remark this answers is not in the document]
ilblackdragon: Addressed in 4c9928a0: added `list_memory_docs_by_owner` stubs to all three non-overriding TestStore impls (`tests/engine_v2_gate_integration.rs`, `tests/engine_v2_skill_codeact.rs`, `src/bridge/router.rs`). The two integration ones filter the in-memory doc set by `user_id` so they can actually exercise the cross-project path; the router one returns `vec![]` since that test path doesn't touch skills.

crates/ironclaw_engine/src/traits/store.rs
```diff
@@ -129,6 +129,28 @@ pub trait Store: Send + Sync {
...
+    /// (migrated into the owner's default project) are visible to gateway users
+    /// whose threads run in a per-user project.
+    async fn list_skills_global(&self) -> Result<Vec<MemoryDoc>, EngineError> {
+        let mut docs = Vec::new();
+        for owner_id in shared_owner_candidates() {
+            docs.extend(self.list_memory_docs_by_owner(owner_id).await?);
+        }
+        docs.retain(|d| d.doc_type == crate::types::memory::DocType::Skill);
+        docs.sort_by_key(|doc| doc.id.0);
+        docs.dedup_by_key(|doc| doc.id);
```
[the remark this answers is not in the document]
ilblackdragon: Addressed in 260b759e: removed the inner `sort_by_key` from `list_skills_global` and kept only a dedup-by-`DocId` (to collapse duplicates across `shared_owner_candidates()`). The caller in `handle_list_skills` does the final sort+dedup after merging with user docs, so there's only one O(n log n) pass now.

crates/ironclaw_engine/src/executor/orchestrator.rs
```diff
@@ -1694,18 +1694,29 @@ async fn handle_list_skills(
...
-    let docs = match store
-        .list_memory_docs_with_shared(thread.project_id, &thread.user_id)
+    // User's own skills.
+    let mut docs = match store
+        .list_memory_docs(thread.project_id, &thread.user_id)
         .await
     {
-        Ok(docs) => docs,
+        Ok(d) => d,
         Err(e) => {
```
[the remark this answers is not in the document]
ilblackdragon: Addressed in 4c9928a0: updated the comment to say the call loads "User's docs in their project (all doc types — skill filtering happens below in the `filter(|d| d.doc_type == Skill)` pass)" so the query scope is unambiguous.

crates/ironclaw_engine/src/executor/orchestrator.rs
```diff
@@ -1694,18 +1694,29 @@ async fn handle_list_skills(
...
+    // shared skills live in the owner's project but must be visible to all users
+    // regardless of which per-user project their thread runs in).
+    match store.list_skills_global().await {
+        Ok(shared) => docs.extend(shared),
+        Err(e) => debug!("__list_skills__: failed to load global skills: {e}"),
+    }
+
+    docs.sort_by_key(|d| d.id.0);
+    docs.dedup_by_key(|d| d.id);
+
```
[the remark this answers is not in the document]
ilblackdragon: Addressed in 260b759e: added `handle_list_skills_returns_shared_skills_from_other_projects` in `orchestrator.rs` tests. It drives the private `handle_list_skills` end-to-end with a shared skill in `project_b`, an alice-owned skill in `project_a`, and alice's thread in `project_a` — asserts both skills are visible and that a non-skill doc in project_a is filtered out. This is the caller-level coverage required by `.claude/rules/testing.md` (Test Through the Caller, Not Just the Helper).

src/bridge/store_adapter.rs
```diff
@@ -732,6 +732,8 @@ fn slugify(title: &str, id: &str) -> String {
 fn serialize_knowledge_doc(doc: &MemoryDoc) -> String {
     let mut frontmatter = String::from("---\n");
     frontmatter.push_str(&format!("id: \"{}\"\n", doc.id.0));
+    frontmatter.push_str(&format!("project_id: \"{}\"\n", doc.project_id.0));
+    frontmatter.push_str(&format!("user_id: \"{}\"\n", doc.user_id));
```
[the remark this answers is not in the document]
ilblackdragon: Addressed in 4c9928a0: introduced `yaml_quoted_escape` helper that escapes `\`, `"`, `\n`, `\r`, `\t`, and other control chars for YAML double-quoted scalars. Now applied to `user_id`, `title`, and each tag string. OIDC subs with unusual characters will no longer produce unparsable frontmatter.

crates/ironclaw_engine/src/executor/orchestrator.rs
```diff
@@ -1694,18 +1694,29 @@ async fn handle_list_skills(
...
+    // Admin/shared skills across ALL projects (fixes multi-tenant visibility —
+    // shared skills live in the owner's project but must be visible to all users
+    // regardless of which per-user project their thread runs in).
+    match store.list_skills_global().await {
+        Ok(shared) => docs.extend(shared),
+        Err(e) => debug!("__list_skills__: failed to load global skills: {e}"),
+    }
+
+    docs.sort_by_key(|d| d.id.0);
+    docs.dedup_by_key(|d| d.id);
```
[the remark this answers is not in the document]
ilblackdragon: Addressed in 260b759e: the inner sort inside `list_skills_global` is gone; only the caller-side sort+dedup remains, so `__list_skills__` now does one normalization pass instead of two.

crates/ironclaw_engine/src/traits/store.rs
ilblackdragon: Good catch — the silent-failure chain is real (default `list_memory_docs_by_owner` → default `list_all_projects` → `Err`, which is then caught by `debug!` in `handle_list_skills`).

Addressed two ways in 260b759e + 4c9928a0:

1. **Doc comment** on `list_memory_docs_by_owner` now explicitly warns: *"Performance warning: the default is O(projects × docs-per-project) and is called from the `__list_skills__` hot path on every thread step via `list_skills_global`. Production `Store` implementations must override this with a flat query — leaving the default in place will turn skill listing into a bottleneck."*
2. **Every in-tree `impl Store`** now overrides `list_memory_docs_by_owner` explicitly (`HybridStore`, the three TestStore impls, `InMemoryStore`, and the MockStores in runtime tests). The default is effectively unreachable today.

I opted against making it a required method because there's value in keeping trait expansions non-breaking for any downstream consumers. If the team prefers fail-closed over the doc warning, happy to swap.

crates/ironclaw_engine/src/traits/store.rs
ilblackdragon: Addressed in 4c9928a0: both `TestStore` impls now implement `list_memory_docs_by_owner` with a real filter over the in-memory doc set (`docs.iter().filter(|d| d.user_id == user_id)`), not just `Ok(vec![])`. That means any integration test that installs a shared skill in one project and lists skills from a thread in another will actually exercise the cross-project path.

src/bridge/router.rs
ilblackdragon: Good find — this is a real pre-existing issue that the PR's owner_id plumbing now exposes. I want to flag rather than silently fix it here because it touches a different subsystem (`migrate_legacy_user_ids` on upgrade) and deserves its own focused change + regression test.

What I did do in 260b759e to narrow the window:

- `deserialize_knowledge_doc` now emits a `debug!` log (with doc_id) whenever it falls back to the `"legacy"` user_id or nil project_id. So if this codepath fires on an upgrade with real data on disk, it'll be visible in logs instead of silently rewriting ownership.

What I'd suggest for the follow-up PR (not this one):

- In `migrate_legacy_user_ids`, peek at `V2SkillMetadata.source` (or the `DocType::Skill` + Bundled/Installed marker on the underlying meta) before stamping `owner_id`. Skills whose original source is Bundled/Installed should migrate to `shared_owner_id()` instead.
- Regression test that round-trips a legacy-serialized Bundled skill through save → upgrade-migrate → `list_skills_global()` and asserts it comes back under `__shared__`.

For single-tenant `HybridStore` today this is harmless (skills still surface via the user's own docs), and single-tenant is the only `Store` impl that hits disk, so there's no user-visible regression to gate this PR on. But I fully agree it's load-bearing for any future multi-tenant `Store` and shouldn't stay open indefinitely. Want me to open a tracking issue?

zetyquickly: should be addressed in https://github.com/nearai/ironclaw/pull/2086/commits/01897d05959d1a81324c3fc3f5eef5b186bfaf80

crates/ironclaw_engine/src/traits/store.rs
zetyquickly: should be addressed in https://github.com/nearai/ironclaw/pull/2086/commits/01897d05959d1a81324c3fc3f5eef5b186bfaf80

src/bridge/router.rs
```diff
@@ -3965,10 +3966,18 @@ async fn migrate_legacy_user_ids(store: &Arc<dyn ironclaw_engine::Store>, owner_
             }
         }
 
-        // Memory docs (use list_memory_docs directly since "legacy" is the user_id)
+        // Memory docs (use list_memory_docs directly since "legacy" is the user_id).
+        // Pre-PR code tagged ALL migrated skills as __shared__, so legacy Skill
+        // docs must be restored to shared_owner_id() — stamping them with
+        // owner_id would make them invisible to list_skills_global() and break
+        // cross-project visibility for gateway users (issue #2084).
         if let Ok(legacy) = store.list_memory_docs(pid, "legacy").await {
```
serrrfirat, line 4133: **High Severity**

Legacy frontmatter docs without `project_id` are still skipped by this migration.

Pre-PR serialized knowledge docs did not include `project_id` or `user_id`. With this PR, `deserialize_knowledge_doc` loads those files as `project_id = nil` and `user_id = "legacy"`. This loop only scans `list_memory_docs(pid, "legacy")` for each real owner project, so those nil-project legacy skill docs are never visited or restamped as `__shared__`. Existing deployments upgrading from the old frontmatter shape can therefore still have migrated skills remain invisible to gateway users after restart.

Please add coverage that writes an actual old frontmatter skill through `HybridStore::load_state_from_workspace`, then runs the legacy user-id migration. The fix should either scan legacy docs by owner across all projects, including the nil project, or assign nil-project legacy docs to the owner default project before saving them with the corrected owner.

zetyquickly: Added a nil-project pass to migrate_legacy_user_ids that catches memory docs deserialized from old frontmatter (missing project_id/user_id fields). These load with project_id = nil and were invisible to the existing per-project migration loop. The new pass assigns them to the owner's default project and stamps user_id correctly. Regression test included.

serrrfirat (approved):
Reviewed as paranoid architect. Core logic is correct — the fix properly separates project-scoped user queries from cross-project shared skill queries. Tests are exemplary (caller-level regression tests per testing.md rules). Three non-blocking follow-ups tracked in #2526.

PR #2094, opened 2026-04-07
fix(security): SSRF via extension download and MCP transport redirects [MEDIUM]

lycheepuppy:
## Summary

- **Fix SSRF via HTTP redirect chains** where the extension download client and MCP HTTP transport followed up to 10 redirects by default, including cross-scheme (HTTPS→HTTP) and cross-host redirects to internal addresses
- Disabled redirect following on both HTTP clients with `reqwest::redirect::Policy::none()`
- Added HTTPS scheme validation for `capabilities_url` parameter (previously had no validation at all)
- Added regression tests

## Security Finding: Server-Side Request Forgery via Extension and Transport Layers

**Severity:** MEDIUM
**Component:** `src/extensions/manager.rs` — `download_and_install_wasm()`, `src/tools/mcp/http_transport.rs` — `HttpTransport::new()`

### Description

The extension manager and MCP HTTP transport constructed outbound HTTP requests using user-supplied or registry-supplied URLs. While the extension download path enforced an `https://` prefix, neither path restricted HTTP redirects. The default reqwest client follows up to 10 redirects transparently — including redirects from HTTPS to HTTP and from external hosts to internal addresses.

An attacker who controls a registry entry or MCP server URL can serve a 302 redirect to an internal endpoint (e.g., the orchestrator API on `127.0.0.1`), effectively converting the server into an open proxy for internal network requests.

A separate entry point — the `capabilities_url` parameter accepted during extension installation — received **no URL validation at all**: no scheme check, no host restriction, and no redirect policy.

IronClaw already implements IP blocklist protections (`validate_fetch_url()` / `reject_private_ip()`) for WASM tool token validation and skill installation paths — but these protections were not applied to extension downloads or MCP transport requests.

### Attack Vector

1. **Redirect-based SSRF**: Attacker registers a malicious extension URL that serves `302 Location: http://127.0.0.1:8080/admin` — the client follows the redirect to the internal service
2. **Direct SSRF via capabilities_url**: Attacker supplies `http://169.254.169.254/latest/meta-data/` as `capabilities_url` — fetched directly with no validation
3. **Data exfiltration**: Response content from internal services could be written to disk at `~/.ironclaw/wasm_tools/`

### Impact

- Internal services (orchestrator API, database, cloud metadata endpoints) reachable through redirect chains
- `capabilities_url` accepts arbitrary URLs including `http://`, `file://`, or internal addresses
- Cloud metadata services (e.g., `169.254.169.254`) reachable in cloud environments
- Port scanning possible via timing oracle

### Fix

1. Set `reqwest::redirect::Policy::none()` on the extension download HTTP client
2. Set `reqwest::redirect::Policy::none()` on the MCP HTTP transport client
3. Add HTTPS scheme validation for `capabilities_url`

## Test plan

- [ ] Existing extension and MCP tests still pass
- [ ] New test `capabilities_url_rejects_non_https` — verifies non-HTTPS URLs are rejected
- [ ] New test `download_client_disables_redirects` — verifies redirect policy configuration
- [ ] New test `transport_disables_redirects` — verifies MCP transport redirect policy


serrrfirat:
**Critical Severity**

The redirect hardening in `src/extensions/manager.rs` is useful, but it still accepts any direct `https://` target.

That leaves direct SSRF open to private/link-local/metadata destinations such as `https://127.0.0.1/...` or cloud metadata hosts. Blocking redirects only closes chained SSRF; it does not validate the destination itself. The same issue applies to `caps_url`.

Suggested fix: parse the URL and apply the same host/IP policy used elsewhere for fetch validation, rejecting loopback, private, link-local, and metadata endpoints even when the scheme is HTTPS.

---

**Medium Severity**

The capabilities side-download path can also leave partial install state behind. The WASM file is written before the `caps_url` fetch path completes, so a later capabilities validation/download failure can strand an installed `.wasm` without its matching capabilities file.

Suggested fix: validate/fetch `caps_url` before committing artifacts, or clean up the partially written files on any failure path.

lycheepuppy:
added fixes based on feedback

serrrfirat (changes requested):
## Security Review: SSRF via Extension Download and MCP Transport Redirects

**Overall: Request Changes** — The core defensive approach is sound (disable redirects + validate IPs at URL acceptance), but there are gaps that need addressing before merge.

### What works well
- Correctly reuses centralized `ssrf_safe_client_builder()` with `Policy::none()` rather than duplicating config
- Adds `reject_private_ip()` validation before HTTP requests in both extension downloads and MCP transport
- HTTPS scheme enforcement on `capabilities_url`
- Refactored download ordering prevents partial install states — nice bonus

### Must fix

**1. Local MCP server breakage**
`HttpMcpTransport::new()` now rejects all private IPs, including `localhost`. MCP servers are *very* commonly run locally (e.g. `localhost:3000`). This PR will break that use case with no workaround — `new_unchecked()` is `#[cfg(test)]` only. The `new_from_config` path in `factory.rs` also calls `HttpMcpTransport::new()`, so user-configured local servers will be rejected too.

**Suggestion:** Trust URLs from user-local config (`~/.ironclaw/mcp.json`) but validate URLs from registry/remote sources. Or add an explicit allowlist/config flag for local servers. Stdio and UDS transports are already safe — the concern is HTTP transport only.

**2. `capabilities_url_rejects_non_https` test is a tautology**
This test just asserts `"http://evil.com/caps.json".starts_with("https://") == false`. It tests `str::starts_with()`, not the actual `download_and_install_wasm()` code path. If someone removes the `starts_with` check from the real code, this test still passes. Replace with a test that exercises the actual function and verifies the error.

### Should fix

**3. DNS rebinding TOCTOU**
Both paths use `reject_private_ip()` (synchronous, performs its own DNS lookup) but then `reqwest` does a *second* DNS lookup when connecting. An attacker with a fast-flipping DNS record could pass the check with a public IP, then resolve to `127.0.0.1` at connect time.

The codebase already has `validate_and_resolve_http_target()` + `ssrf_safe_client_builder_for_target()` with DNS pinning via `resolve_to_addrs()`. Upgrade both paths to use the stronger variant — the infrastructure already exists.

**4. Weak HTTPS scheme check**
`caps_url.starts_with("https://")` is a string prefix check. Use `url::Url::parse()` and check `parsed.scheme() == "https"` for robustness against edge cases (`HTTPS://`, `https://` with no host, etc.).

### Minor

- Error type inconsistency: `capabilities_url` rejection returns `ExtensionError::InstallFailed` for scheme check but `ExtensionError::DownloadFailed` for `reject_private_ip`. Both are the same validation step.
- `download_client_disables_redirects` and `transport_disables_redirects` tests just build a client and drop it — they don't test that redirects are actually blocked. The existing `test_ssrf_safe_client_builder_disables_redirects` in `http_security.rs` already properly tests this with redirect servers.
- No caller-level integration test driving `download_and_install_wasm()` or `create_client_from_config()` with a redirect-to-private-IP scenario (project rule: "test through the caller, not just the helper").

serrrfirat (changes requested):

**PR:** #2094 — SSRF via extension download and MCP transport

### Critical

**IPv4-mapped IPv6 bypass in `is_private_ip()`:** `::ffff:127.0.0.1` and `::ffff:169.254.169.254` bypass all private IP checks. The IPv6 branch does not handle IPv4-mapped addresses. An attacker can use `https://[::ffff:127.0.0.1]/` to reach internal services including cloud metadata endpoints.

**Fix:** Add `v6.to_ipv4_mapped().is_some_and(|v4| /* check v4 */)` to the IPv6 branch of `is_private_ip()` in `src/tools/wasm/http_security.rs`. This is a one-line fix that closes the bypass.

### High

**TOCTOU / DNS rebinding vulnerability:** `reject_private_ip()` performs a DNS lookup, checks the result, then returns. The actual HTTP request via reqwest performs a *separate* DNS resolution. An attacker with a malicious DNS server can return a public IP on the first lookup (passing validation) and a private IP on the second.

**Fix:** Switch to `validate_and_resolve_http_target()` + `ssrf_safe_client_builder_for_target()` which already exist in the same file and provide DNS pinning.

### Medium

- **Blocking DNS in async context:** `reject_private_ip()` uses `std::net::ToSocketAddrs` (blocking syscall) in async code paths. Use `tokio::net::lookup_host()` instead (the async variant `validate_and_resolve_http_target()` already exists).
- **Weak regression tests:** `capabilities_url_rejects_non_https`, `download_client_disables_redirects`, and `transport_disables_redirects` only check string literals or that builders don't panic — they never call production validation functions. Follow the pattern of `test_ssrf_safe_client_builder_disables_redirects` which spins up actual servers.
- **Error message info disclosure:** Error from `reject_private_ip` in `HttpMcpTransport::new()` may leak resolved IP addresses / internal network topology.

### Nice to have

- Consider adding 6to4 (`2002::/16`) to the IPv6 blocklist since it can embed arbitrary IPv4 addresses.

### Summary

The redirect-disabling approach (`Policy::none()`) and capabilities URL HTTPS enforcement are sound. The `#[cfg(test)]` gating of `new_unchecked` is correct. However, the IPv4-mapped IPv6 bypass and DNS rebinding TOCTOU are ship-blockers that must be fixed before merge.

src/tools/mcp/http_transport.rs
```diff
@@ -50,17 +58,56 @@ impl HttpMcpTransport {
...
+                ToolError::ExternalService(format!(
+                    "[{}] SSRF blocked for MCP server URL: {}",
+                    server_name, e
+                ))
+            })?;
+        }
+
+        let http_client = crate::tools::wasm::ssrf_safe_client_builder()
+            .timeout(std::time::Duration::from_secs(30))
+            .build()
```
serrrfirat: **High Severity** — DNS rebinding TOCTOU in MCP transport

`HttpMcpTransport::new()` validates IPs via `reject_private_ip()` at construction time, but builds the HTTP client via `ssrf_safe_client_builder()` **without DNS pinning**. An attacker controlling DNS can pass the check with a public IP at construction time, then rebind to a private IP before actual requests are made.

The extension download path in `manager.rs` correctly uses the full chain:
```rust
let target = validate_and_resolve_http_target(url).await?;
let client = ssrf_safe_client_builder_for_target(&target)...
```

This pins the resolved addresses via `resolve_to_addrs()`, preventing reqwest from re-resolving DNS. The MCP transport should use the same pattern.

**Suggested fix:** Make `HttpMcpTransport::new()` async and use `validate_and_resolve_http_target()` + `ssrf_safe_client_builder_for_target()` instead of `reject_private_ip()` + `ssrf_safe_client_builder()`. This requires making `McpClient::new()`, `new_with_name()`, and `new_authenticated()` async as well (or accepting a pre-validated target). The constructor signature change is already in this PR (`-> Result<Self, ToolError>`), so adding `async` is a natural extension.

src/tools/mcp/http_transport.rs
```diff
@@ -50,17 +58,56 @@ impl HttpMcpTransport {
             McpServerName::new("unknown")
                 .expect("'unknown' is a valid McpServerName (alnum allowlist)") // safety: hardcoded literal satisfies alnum-only validation; infallible
         });
+
```
serrrfirat, line 58: **Medium Severity** — Blocking DNS resolution on tokio runtime thread

`reject_private_ip()` uses synchronous `to_socket_addrs()` (line 136 of the base file), and is called from the non-async `HttpMcpTransport::new()` constructor. Since all callers (`create_client_from_config`, CLI `test_server`, etc.) run in async contexts, this blocks the tokio runtime thread during DNS resolution.

With a slow or unresponsive DNS server, this can stall the entire tokio runtime — other tasks on the same worker thread will be starved until the DNS lookup completes or times out.

The extension download path correctly uses async DNS via `validate_and_resolve_http_target()` → `tokio::net::lookup_host()`.

**Suggested fix:** If `HttpMcpTransport::new()` is made async (per the DNS pinning suggestion), switch to `validate_and_resolve_http_target()` which uses `tokio::net::lookup_host()` for non-blocking DNS. Alternatively, wrap the sync call in `tokio::task::spawn_blocking()` as a stopgap.

src/tools/mcp/http_transport.rs
```diff
@@ -50,17 +58,56 @@ impl HttpMcpTransport {
...
+        })
+    }
+
+    /// Test-only constructor that skips IP validation (allows localhost).
+    #[cfg(test)]
+    pub fn new_unchecked(server_url: impl Into<String>, server_name: impl Into<String>) -> Self {
+        let raw: String = server_name.into();
+        let server_name = McpServerName::new(&raw).unwrap_or_else(|e| {
+            tracing::debug!(
+                candidate = %raw,
```
serrrfirat: **Medium Severity** — Missing explicit test for SSRF rejection of private IPs in MCP transport

The core security behavior added by this PR — that `HttpMcpTransport::new()` and `McpClient::new()` reject private IPs — has no direct test. Existing tests only implicitly verify the localhost exemption (by using `localhost` URLs that `.unwrap()` successfully) or use public IPs (`93.184.215.14`).

**Suggested tests:**
```rust
#[test]
fn new_rejects_private_ipv4() {
    let err = HttpMcpTransport::new("http://10.0.0.1:8080", "test");
    assert!(err.is_err(), "10.0.0.0/8 must be blocked");

    let err = HttpMcpTransport::new("http://192.168.1.1:8080", "test");
    assert!(err.is_err(), "192.168.0.0/16 must be blocked");
}

#[test]
fn new_allows_localhost() {
    let transport = HttpMcpTransport::new("http://localhost:8080", "test");
    assert!(transport.is_ok(), "localhost must be exempted from SSRF checks");
}

#[test]
fn new_allows_public_ip() {
    let transport = HttpMcpTransport::new("http://93.184.215.14:8080", "test");
    assert!(transport.is_ok(), "public IPs must be allowed");
}
```

lycheepuppy:
Addressed the highs and mediums and submitted one large push. Let's try to close this PR out and if there are remaining comments, we can take it to a new PR to avoid turning this PR any larger!

src/tools/mcp/http_transport.rs
lycheepuppy: Addressed. `HttpMcpTransport::new()` is now async and uses `validate_and_resolve_http_target()` + `ssrf_safe_client_builder_for_target()` for DNS pinning, matching the extension download path exactly. Localhost URLs are exempted via `config::is_localhost_url()` for dev MCP servers.

src/tools/mcp/http_transport.rs
lycheepuppy: Addressed. `new()` is now async, so DNS resolution goes through `tokio::net::lookup_host()` via `validate_and_resolve_http_target()` instead of blocking `to_socket_addrs()`. All callers (`factory.rs`, `client.rs`, `cli/mcp.rs`) have been updated to `.await` the constructor.

src/tools/mcp/http_transport.rs
lycheepuppy: Addressed. Added tests `new_rejects_private_ips` (covering 10.x, 192.168.x, 172.16.x, and link-local), `new_allows_public_ips`, `new_falls_back_on_invalid_server_name`, and `new_preserves_valid_server_name`. Loopback IPs (127.0.0.1, [::1]) are intentionally allowed via the localhost exemption for dev MCP servers.

ilblackdragon:
Closing this because it updates the legacy top-level `src/` MCP/extension implementation (`src/cli/mcp.rs`, `src/extensions/manager.rs`, `src/tools/mcp/*`). Product work has moved to the Reborn workspace under `crates/`, and this PR is now outdated against that architecture.

Thanks for the contribution.

PR #2101, opened 2026-04-07
fix(db): repair V6 migration checksum and guard against re-modification (#1328)

### LINKED ISSUE #1328: Upgrade to v0.19.0 fails: V6__routines migration checksum mismatch on existing PostgreSQL databases
## Description

Upgrading from v0.18.0 (or any earlier release) to v0.19.0 fails on startup with:

```
Error: Migration failed: applied migration V6__routines is different than filesystem one V6__routines
```

## Root Cause

PR #1151 modified the already-applied `V6__routines.sql` migration in place, changing:

```sql
-- v0.18.0 (sha 36f63cb, 2930 bytes)
notify_user TEXT NOT NULL DEFAULT 'default',
```

to:

```sql
-- v0.19.0 (sha 9697251, 2903 bytes)
notify_user TEXT,
```

The PR also correctly added `V13__owner_scope_notify_targets.sql` to handle the schema change incrementally for existing databases. However, because V6 was also modified, refinery's checksum validation rejects the upgrade before V13 can run.

The PR body acknowledges this under **Database Impact**:

> Adds `migrations/V13__owner_scope_notify_targets.sql` and **updates fresh-schema `migrations/V6__routines.sql`** so `routines.notify_user` can be `NULL`.

This works for fresh installs but breaks every existing PostgreSQL database upgrading to v0.19.0.

## Steps to Reproduce

1. Install IronClaw v0.18.0 (or earlier) with PostgreSQL backend
2. Run it at least once so migrations V1–V12 are applied
3. Replace the binary with v0.19.0
4. Start IronClaw — fails immediately with the checksum error

## Expected Behavior

Upgrade from v0.18.0 to v0.19.0 should apply V13 incrementally without touching the checksum of already-applied migrations.

## Suggested Fix

Revert `V6__routines.sql` to its v0.18.0 content (sha `36f63cb`). The `V13__owner_scope_notify_targets.sql` migration already handles the schema change correctly for both existing and fresh databases (since `ALTER COLUMN DROP NOT NULL` is a no-op if the column is already nullable).

## Environment

- IronClaw v0.18.0 → v0.19.0 upgrade
- PostgreSQL 16 + pgvector 0.8.0
- Platform: aarch64 Linux (NVIDIA DGX Spark)
- Database backend: PostgreSQL (refinery migrations)

ilblackdragon:
## Summary

Fixes #1328. Upgrading from v0.18.0 (or any earlier release) to v0.19.0 currently fails on startup with `applied migration V6__routines is different than filesystem one V6__routines` because PR #1151 modified `migrations/V6__routines.sql` in place after it had already shipped and been applied to production databases.

This PR repairs the regression for both populations of databases in the wild **and** adds a CI guard so the same accident cannot happen again.

### The runtime fix

- Reverts `migrations/V6__routines.sql` to its v0.18.0 content. `V13__owner_scope_notify_targets.sql` already applies the schema change incrementally and is idempotent (`ALTER COLUMN ... DROP NOT NULL` is a no-op when the column is already nullable).
- Adds `src/db/migration_fixup.rs::realign_diverged_checksums()`, called from both `Store::run_migrations` (`src/history/store.rs`) and `Wizard::run_migrations_postgres` (`src/setup/wizard.rs`) immediately before refinery validates. It recomputes V6's canonical SipHasher13 checksum from the embedded SQL via `refinery::Migration::unapplied()` and rewrites any divergent row in `refinery_schema_history`. Idempotent: no-op on fresh installs and already-aligned databases.

This handles **both** populations:
| Population | `refinery_schema_history` before | After fix-up |
|---|---|---|
| Pre-#1151 (v0.18.0 or earlier) | original V6 checksum | unchanged (already canonical) |
| Post-#1151 (v0.19.0 fresh installs) | modified V6 checksum | rewritten to canonical |

We deliberately do **not** disable refinery's checksum validation (`set_abort_divergent(false)`) — that would mask future genuine drift. The set of known divergences is a hard-coded list so any future fix-up is an explicit code change visible in review.

### The "never again" guard

- `migrations/checksums.lock` pins every migration's SipHasher13 checksum (the same value refinery stores in production databases).
- `released_migrations_are_immutable` (`src/db/migration_fixup.rs`) is a `cargo test` test that loads the lockfile, recomputes every `V*.sql` checksum, and fails if any value drifts or any migration is missing from the lockfile. **Modifying a released migration → CI fails. Adding a new migration → must add a lockfile line in the same PR.**
- `v6_routines_matches_v018_checksum` is a second line of defence pinning V6's literal checksum `18049045188188232070` as a hard-coded sentinel — a malicious or careless edit of both V6 and the lockfile would defeat the previous test but not this one.
- `regenerate_migration_checksums_lockfile` (ignored) is the bootstrap helper for new migrations: `cargo test -p ironclaw -- --ignored regenerate_migration_checksums_lockfile`.

libSQL is unaffected: `INCREMENTAL_MIGRATIONS` in `libsql_migrations.rs` is keyed by name (not checksumed) and the corresponding `routine_notify_user_nullable` entry is already idempotent.

## Test plan

- [x] `cargo fmt`
- [x] `cargo check --no-default-features --features libsql` — libSQL-only build clean
- [x] `cargo test --lib db::migration_fixup` — 2 passed, 1 ignored (bootstrap helper)
- [x] `cargo test --lib` — 4245 passed; the only failure (`extensions::manager::tests::test_telegram_token_colon_preserved_in_validation_url`) is a pre-existing flake on staging unrelated to this PR (filed separately)
- [ ] **Manual upgrade smoke test on a v0.18.0 PostgreSQL database** — recommended before merge:
  1. Start a PG instance, run v0.18.0 IronClaw against it (V1–V12 applied)
  2. Switch to this branch and start IronClaw — should succeed, V13 should apply, no checksum error
  3. Restart IronClaw — should be a no-op (no warn log for realignment)
- [ ] **Manual smoke test on a v0.19.0 fresh-install database** — also recommended:
  1. Start a clean PG instance, run staging (or v0.19.0) — V1–V20 applied with the modified V6 checksum
  2. Switch to this branch and start IronClaw — should succeed; should log one realignment warning for V6
  3. Restart IronClaw — should be a no-op

### DIFF
--- a/src/db/migration_fixup.rs
+++ b/src/db/migration_fixup.rs
@@ -188,6 +188,13 @@ pub(crate) async fn realign_diverged_checksums_with(
             })?;
 
         if updated > 0 {
+            // `warn!` is intentional here even though CLAUDE.md warns
+            // against `info!`/`warn!` in background tasks (they corrupt
+            // the REPL/TUI). This fix-up runs during database migration
+            // at startup, *before* any channel/REPL/TUI is initialized,
+            // so terminal-rendering interference is impossible. If this
+            // call is ever moved later in startup, downgrade to `debug!`
+            // or pre-buffer the message.
             tracing::warn!(
                 migration = %migration_label,
                 rows = updated,
@@ -229,7 +236,16 @@ mod tests {
                     lineno + 1
                 )
             });
-            map.insert(key.trim().to_string(), parsed);
+            // Reject duplicate keys: a stray duplicate would silently
+            // overwrite an earlier pinned checksum and weaken the
+            // immutability guard. Detect it loudly during the test.
+            let key = key.trim().to_string();
+            if map.insert(key.clone(), parsed).is_some() {
+                panic!(
+                    "checksums.lock line {} contains duplicate migration key: {key}",
+                    lineno + 1
+                );
+            }
         }
         map
     }
@@ -357,7 +373,18 @@ mod tests {
                 }
             })
             .collect();
-        sql_files.sort();
+        // Natural-sort by parsed migration version (so V2 comes before
+        // V10), not lex-sort (which would order V10 before V2). The
+        // resulting lockfile reads in numeric order which makes review
+        // diffs easier to scan.
+        sql_files.sort_by_key(|path| {
+            let stem = path.file_stem().and_then(|s| s.to_str()).unwrap_or("");
+            // V<n>__<name> → parse the digits between `V` and `__`.
+            stem.strip_prefix('V')
+                .and_then(|s| s.split_once("__"))
+                .and_then(|(v, _)| v.parse::<u32>().ok())
+                .unwrap_or(u32::MAX)
+        });
 
         let mut output = String::new();
         output.push_str(
@@ -637,5 +664,94 @@ mod tests {
             )
             .await
             .expect("cleanup post-run");
+    }
+
+    /// Regression test for the defensive check that rejects a
+    /// `KnownDivergence` whose canonical checksum is also listed in its
+    /// own `known_bad_checksums`. Such a misconfiguration would silently
+    /// rewrite already-correct rows to themselves; the production code
+    /// returns `Err(DatabaseError::Migration(...))` to refuse startup.
+    ///
+    /// Like the realignment integration test above, this requires a
+    /// `PgClient` and so is gated on `feature = "integration"`. The
+    /// error fires *before* any UPDATE query, so the test only needs a
+    /// reachable database — `refinery_schema_history` does not even
+    /// need to exist.
+    #[cfg(feature = "integration")]
+    #[tokio::test]
+    async fn rejects_canonical_in_known_bad_checksums() {
+        use deadpool_postgres::{Manager, Pool};
+        use tokio_postgres::{Config, NoTls};
+
+        let database_url = std::env::var("DATABASE_URL")
+            .unwrap_or_else(|_| "postgres://localhost/ironclaw_test".to_string());
+        let config: Config = match database_url.parse() {
+            Ok(c) => c,
+            Err(e) => {
+                eprintln!("skipping: invalid DATABASE_URL ({e})");
+                return;
+            }
+        };
+        let mgr = Manager::new(config, NoTls);
+        let pool = match Pool::builder(mgr).max_size(2).build() {
+            Ok(p) => p,
+            Err(e) => {
+                eprintln!("skipping: failed to build pool ({e})");
+                return;
+            }
+        };
+        let mut client = match pool.get().await {
+            Ok(c) => c,
+            Err(e) => {
+                eprintln!("skipping: database unavailable ({e})");
+                return;
+            }
+        };
+
+        // Make sure refinery_schema_history exists so we exercise the
+        // post-history-check branch (the early-return on missing table
+        // would otherwise mask the validation we want to test).
+        client
+            .batch_execute(
+                "CREATE TABLE IF NOT EXISTS refinery_schema_history ( \
+                    version INT4 PRIMARY KEY, \
+                    name VARCHAR(255), \
+                    applied_on VARCHAR(255), \
+                    checksum VARCHAR(255))",
+            )
+            .await
+            .expect("create refinery_schema_history");
+
+        // Construct a deliberately-misconfigured divergence where the
+        // canonical checksum appears in its own known-bad list.
+        const TEST_SQL: &str = "-- bad-config test fixture\nSELECT 2;\n";
+        let canonical = Migration::unapplied("V99998__bad_config", TEST_SQL)
+            .unwrap()
+            .checksum();
+        let bad_divergences: &[KnownDivergence] = &[KnownDivergence {
+            version: 99998,
+            name: "bad_config",
+            sql: TEST_SQL,
+            known_bad_checksums: Box::leak(Box::new([canonical])),
+            explanation: "intentional misconfig for PR #2101 regression test",
+        }];
+
+        let result = super::realign_diverged_checksums_with(&mut client, bad_divergences).await;
+        match result {
+            Err(crate::error::DatabaseError::Migration(msg)) => {
+                assert!(
+                    msg.contains("known_bad_checksums"),
+                    "expected error to mention known_bad_checksums, got: {msg}",
+                );
+                assert!(
+                    msg.contains("V99998__bad_config"),
+                    "expected error to identify the offending migration label, got: {msg}",
+                );
+            }
+            Err(other) => panic!("expected Migration error, got: {other:?}"),
+            Ok(()) => {
+                panic!("expected Err — canonical checksum in known_bad list should refuse startup")
+            }
+        }
     }
 }
--- a/src/db/migration_fixup.rs
+++ b/src/db/migration_fixup.rs
@@ -141,16 +141,23 @@ pub(crate) async fn realign_diverged_checksums_with(
         // Defensive: the canonical checksum must never appear in the bad
         // list, otherwise we'd be rewriting already-correct rows. This is
         // a programming error in `KNOWN_DIVERGENCES`, not a runtime
-        // condition. Use `assert!` rather than `debug_assert!` so the
-        // guard remains in release builds — `KNOWN_DIVERGENCES` has at
-        // most a handful of entries, so the cost is one constant-time
-        // slice lookup per startup.
-        assert!(
-            !divergence
-                .known_bad_checksums
-                .contains(&migration.checksum()),
-            "{migration_label}: canonical checksum is listed as known-bad",
-        );
+        // condition, but it must still be detected in release builds (a
+        // `debug_assert!` would be stripped). Return a hard error so the
+        // process refuses to start with a misconfigured fix-up table —
+        // see PR #2101 review by @serrrfirat. Cost is one constant-time
+        // slice lookup per startup; the `KNOWN_DIVERGENCES` list has at
+        // most a handful of entries.
+        if divergence
+            .known_bad_checksums
+            .contains(&migration.checksum())
+        {
+            return Err(DatabaseError::Migration(format!(
+                "{migration_label}: canonical checksum is listed in \
+                 known_bad_checksums — this is a programming error in \
+                 KNOWN_DIVERGENCES that would silently rewrite \
+                 already-correct rows",
+            )));
+        }
 
         // Only rewrite rows whose stored checksum is one of the known
         // historical bad values for this migration. Any other divergence
--- a/src/db/migration_fixup.rs
+++ b/src/db/migration_fixup.rs
@@ -60,21 +60,21 @@ use crate::error::DatabaseError;
 /// release. Add a new entry here only if the same accident ever happens
 /// again — the immutability test in `migrations/checksums.lock` is the
 /// preferred guard.
-struct KnownDivergence {
-    version: i32,
-    name: &'static str,
+pub(crate) struct KnownDivergence {
+    pub(crate) version: i32,
+    pub(crate) name: &'static str,
     /// The current (canonical) SQL content, embedded at compile time.
-    sql: &'static str,
+    pub(crate) sql: &'static str,
     /// The exact set of historical bad checksums we are willing to rewrite
     /// for this migration. **The fix-up only fires when the stored checksum
     /// matches one of these literals** — any other divergence (manual
     /// tampering, hardware corruption, an unknown future regression) is
     /// left alone so refinery can still abort startup loudly.
-    known_bad_checksums: &'static [u64],
+    pub(crate) known_bad_checksums: &'static [u64],
     /// Human-readable explanation of why this divergence exists, surfaced
     /// in the realignment warning log so future entries are not coupled to
     /// the V6/#1328 wording.
-    explanation: &'static str,
+    pub(crate) explanation: &'static str,
 }
 
 const KNOWN_DIVERGENCES: &[KnownDivergence] = &[KnownDivergence {
@@ -95,6 +95,16 @@ const KNOWN_DIVERGENCES: &[KnownDivergence] = &[KnownDivergence {
 /// with the canonical checksum of the embedded migration. Must be called
 /// before `refinery::Runner::run_async`.
 pub async fn realign_diverged_checksums(client: &mut PgClient) -> Result<(), DatabaseError> {
+    realign_diverged_checksums_with(client, KNOWN_DIVERGENCES).await
+}
+
+/// Inner implementation that takes the divergence list as a parameter, so
+/// integration tests can drive it against synthetic rows without colliding
+/// with real V6 rows in a shared test database.
+pub(crate) async fn realign_diverged_checksums_with(
+    client: &mut PgClient,
+    divergences: &[KnownDivergence],
+) -> Result<(), DatabaseError> {
     // On a fresh install the history table does not yet exist. Refinery
     // will create it during the first `run_async()` call. There is nothing
     // to realign in that case.
@@ -115,7 +125,7 @@ pub async fn realign_diverged_checksums(client: &mut PgClient) -> Result<(), Dat
         return Ok(());
     }
 
-    for divergence in KNOWN_DIVERGENCES {
+    for divergence in divergences {
         // Compute the canonical checksum the same way refinery does
         // (SipHasher13 over name, version, sql in that order). Refinery
         // stores the resulting u64 as a decimal string in the `checksum`
@@ -131,8 +141,11 @@ pub async fn realign_diverged_checksums(client: &mut PgClient) -> Result<(), Dat
         // Defensive: the canonical checksum must never appear in the bad
         // list, otherwise we'd be rewriting already-correct rows. This is
         // a programming error in `KNOWN_DIVERGENCES`, not a runtime
-        // condition, so failing loudly here is appropriate.
-        debug_assert!(
+        // condition. Use `assert!` rather than `debug_assert!` so the
+        // guard remains in release builds — `KNOWN_DIVERGENCES` has at
+        // most a handful of entries, so the cost is one constant-time
+        // slice lookup per startup.
+        assert!(
             !divergence
                 .known_bad_checksums
                 .contains(&migration.checksum()),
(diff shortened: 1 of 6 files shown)


serrrfirat:
Potential issue in `src/db/migration_fixup.rs`: the new checksum repair updates any `refinery_schema_history` row matching `(version = 6, name = 'routines')` whenever its checksum differs from the canonical value. That means it will also silently overwrite unrelated corruption or manual tampering, not just the one known historical V6 divergence this PR is meant to normalize.

That weakens checksum validation for V6 instead of narrowly exempting the specific released mismatch. A safer approach would be to hard-code the exact historical bad checksum(s) and only rewrite when the stored checksum matches one of those known divergent values; anything else should be left alone so refinery can still fail loudly.

src/db/migration_fixup.rs
```diff
@@ -0,0 +1,357 @@
...
+/// Realign `refinery_schema_history` rows whose stored checksum disagrees
+/// with the canonical checksum of the embedded migration. Must be called
+/// before `refinery::Runner::run_async`.
+pub async fn realign_diverged_checksums(client: &mut PgClient) -> Result<(), DatabaseError> {
+    // On a fresh install the history table does not yet exist. Refinery
+    // will create it during the first `run_async()` call. There is nothing
+    // to realign in that case.
+    let history_exists: bool = client
+        .query_one(
+            "SELECT to_regclass('public.refinery_schema_history') IS NOT NULL",
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in 15d1a8f7 — dropped the `public.` schema qualifier so the probe matches the unqualified UPDATE and resolves through the active `search_path`, matching how refinery itself locates the table.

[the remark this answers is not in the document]
ilblackdragon: Fixed in 15d1a8f7 — same fix as for the equivalent gemini-code-assist comment above. The probe is now `to_regclass('refinery_schema_history')` (unqualified), consistent with the UPDATE statement and honouring the active `search_path`.

src/db/migration_fixup.rs
```diff
@@ -0,0 +1,357 @@
...
+    name: &'static str,
+    /// The current (canonical) SQL content, embedded at compile time.
+    sql: &'static str,
+}
+
+const KNOWN_DIVERGENCES: &[KnownDivergence] = &[KnownDivergence {
+    version: 6,
+    name: "routines",
+    sql: include_str!("../../migrations/V6__routines.sql"),
+}];
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in 15d1a8f7 — added an `explanation: &'static str` field to `KnownDivergence`, populated on the V6 entry, and consumed by the realignment warning so future entries can supply their own context.

src/db/migration_fixup.rs
```diff
@@ -0,0 +1,357 @@
...
+                "compute canonical checksum for {migration_label}: {e}"
+            ))
+        })?;
+        let canonical_checksum = migration.checksum().to_string();
+
+        let updated = client
+            .execute(
+                "UPDATE refinery_schema_history \
+                 SET checksum = $1 \
+                 WHERE version = $2 AND name = $3 AND checksum <> $1",
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in 15d1a8f7 — switched to `checksum IS DISTINCT FROM $1`. Refinery's column is `NOT NULL` in practice, but using the NULL-aware operator means a corrupted row with a NULL checksum would also be picked up for repair instead of being silently skipped.

src/db/migration_fixup.rs
```diff
@@ -0,0 +1,357 @@
...
+        if updated > 0 {
+            tracing::warn!(
+                migration = %migration_label,
+                rows = updated,
+                "Realigned refinery_schema_history checksum (issue #1328 fix-up). \
+                 Migration content matches the v0.18.0 release; the schema \
+                 change introduced in PR #1151 is applied incrementally by \
+                 V13__owner_scope_notify_targets."
+            );
+        }
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in 15d1a8f7 — the warning now uses `divergence.explanation` instead of inlining the V6/#1328 wording, paired with the new field added in the comment above.

ilblackdragon:
@serrrfirat — really good catch, thanks. You're right that `IS DISTINCT FROM $canonical` was too broad: it would silently rewrite *any* non-canonical V6 row, masking unrelated corruption or tampering instead of narrowly exempting the one historical released mismatch.

Fixed in a5372b3b. The realignment now uses an explicit whitelist:

- Added a `known_bad_checksums: &'static [u64]` field to `KnownDivergence`.
- The single value pinned for V6 is `11230857244097235596` — the SipHasher13 of `git show 878a67cd:migrations/V6__routines.sql` (the post-#1151 modified content).
- The UPDATE is now `WHERE version = $2 AND name = $3 AND checksum = ANY($4)` — only rows with exactly that historical bad value are rewritten. Anything else (manual tampering, hardware corruption, an unrelated future regression) is left alone so refinery still aborts startup loudly.
- New `v6_known_bad_checksum_matches_post_1151_content` test pins the literal so the whitelist cannot drift or be silently widened in a future PR.
- Defensive `debug_assert!` ensures the canonical checksum is never accidentally listed as known-bad (would no-op the fix-up).

The behaviour matrix is now:

| Stored V6 checksum | Action |
|---|---|
| `18049045188188232070` (canonical, v0.18.0) | no-op |
| `11230857244097235596` (known bad, post-#1151) | rewrite to canonical, log warning |
| anything else | leave alone, refinery aborts |

This preserves checksum validation as a real safety net for V6 instead of weakening it.

ilblackdragon:
**CI: applied `skip-regression-check` label**

The "Regression test enforcement" check fails on this PR despite it containing 5 unit tests in `src/db/migration_fixup.rs` (`released_migrations_are_immutable`, `v6_routines_matches_v018_checksum`, `v6_known_bad_checksum_matches_post_1151_content`, plus 2 ignored bootstrap helpers).

Root cause is a false-positive in the enforcement script for **new files**: it parses each diff hunk header (`@@ -0,0 +1,N @@`) and extracts the starting line number, then asserts that line is `>= test_mod_start_line`. For a brand-new file, the single hunk always starts at line 1, but the `#[cfg(test)] mod tests` block sits much further down (around line 150 here), so the check can never satisfy `1 >= 150` no matter how many tests the new file contains. The script needs to account for the hunk *length*, not just its start.

Filing the script bug as a follow-up after merge so future PRs that introduce new files with inline tests don't hit the same false positive. Applying `skip-regression-check` is the right escape hatch for now per the instructions in the CI warning itself.

src/db/migration_fixup.rs
```diff
@@ -0,0 +1,467 @@
...
+                "compute canonical checksum for {migration_label}: {e}"
+            ))
+        })?;
+        let canonical_checksum = migration.checksum().to_string();
+
+        // Defensive: the canonical checksum must never appear in the bad
+        // list, otherwise we'd be rewriting already-correct rows. This is
+        // a programming error in `KNOWN_DIVERGENCES`, not a runtime
+        // condition, so failing loudly here is appropriate.
+        debug_assert!(
```
serrrfirat: **Medium Severity** — `debug_assert!` is silently stripped in release builds

The comment on this block says *"This is a programming error in `KNOWN_DIVERGENCES`, not a runtime condition, so failing loudly here is appropriate."* — but `debug_assert!` does **not** fail loudly in release builds; it's compiled out entirely.

If someone accidentally adds the canonical checksum to `known_bad_checksums`, the guard is absent in production. The practical impact is low (the UPDATE would "rewrite" already-correct rows to the same value — a functional no-op), and the test suite catches this separately via `v6_known_bad_checksum_matches_post_1151_content`. But for a migration safety module that's explicitly about belt-and-suspenders, the belt should actually be there in release too.

**Suggested fix:** Change `debug_assert!` → `assert!`. This runs once per startup per known divergence (currently 1) — zero measurable cost, and it matches the stated intent of the comment.

src/db/migration_fixup.rs
```diff
@@ -0,0 +1,467 @@
...
+    known_bad_checksums: &[11230857244097235596],
+    explanation: "Migration content matches the v0.18.0 release; the schema \
+                  change introduced in PR #1151 is applied incrementally by \
+                  V13__owner_scope_notify_targets.",
+}];
+
+/// Realign `refinery_schema_history` rows whose stored checksum disagrees
+/// with the canonical checksum of the embedded migration. Must be called
+/// before `refinery::Runner::run_async`.
+pub async fn realign_diverged_checksums(client: &mut PgClient) -> Result<(), DatabaseError> {
```
serrrfirat, line 185: **Low Severity** — No integration test exercises the actual UPDATE path against PostgreSQL

The unit tests thoroughly verify checksums and lockfile integrity, but `realign_diverged_checksums` itself is never called against a real database in the test suite. A subtle issue — e.g., a column type mismatch between `checksum VARCHAR(255)` and the parameterized `&String`, or an unexpected `search_path` resolution of `refinery_schema_history` — would only surface in production or during the manual smoke tests described in the PR.

**Suggested fix:** Consider adding a `#[cfg(feature = "integration")]` test that:
1. Stands up a PostgreSQL instance (via testcontainers or the existing integration harness)
2. Runs migrations to create `refinery_schema_history`
3. Overwrites V6's checksum to the known-bad value (`11230857244097235596`)
4. Calls `realign_diverged_checksums`
5. Asserts the row now holds the canonical checksum

This codifies the manual smoke test from the PR description and catches regressions if refinery ever changes its schema. Not blocking — the fix is straightforward SQL and the PR is well-tested otherwise — but it would close the last untested seam.

ilblackdragon:
@serrrfirat — both addressed:

**1. `debug_assert!` stripped in release** — fixed across two commits. Initial fix in e9907394 changed it to `assert!`, but that tripped the project's "No panics in production code" CI check (the regex matches bare `assert!` outside test attributes). Final fix in 2da6a04d converts it to an early returned `Err(DatabaseError::Migration(...))` instead:

```rust
if divergence.known_bad_checksums.contains(&migration.checksum()) {
    return Err(DatabaseError::Migration(format!(
        "{migration_label}: canonical checksum is listed in \
         known_bad_checksums — this is a programming error in \
         KNOWN_DIVERGENCES that would silently rewrite \
         already-correct rows",
    )));
}
```

This actually achieves your stated intent better than `assert!` would have: the guard runs in release builds, refuses to start the process if `KNOWN_DIVERGENCES` is misconfigured, and is more idiomatic for a function that already returns `Result`. No panic, just a hard startup error.

**2. PostgreSQL integration test** — added in e9907394. Refactored `realign_diverged_checksums` into a thin pub wrapper plus an injectable `pub(crate) realign_diverged_checksums_with(client, divergences)` inner helper, then added `realign_repairs_known_bad_checksum_against_postgres` inside the existing `mod tests` block, gated on `#[cfg(feature = "integration")]`. The test:

1. Skips gracefully if `DATABASE_URL` is unreachable (matches the `try_connect` pattern in `tests/workspace_integration.rs`).
2. Creates `refinery_schema_history` if missing (so the test doesn't depend on a pre-migrated DB).
3. Seeds a synthetic `V99999__test_routines` row with a deliberately-wrong checksum — using a synthetic version so it never collides with real V6 rows in shared CI databases.
4. Calls `realign_diverged_checksums_with` with a custom `KnownDivergence` slice (production `KNOWN_DIVERGENCES` is left untouched).
5. Asserts the row now holds the canonical checksum.
6. Re-runs the realignment and asserts it's a no-op (idempotency check).
7. Cleans up.

Run with:
```bash
DATABASE_URL=postgres://localhost/ironclaw_test \
  cargo test --features integration --lib \
  db::migration_fixup::tests::realign_repairs_known_bad_checksum_against_postgres
```

This catches the column-type-mismatch and search-path resolution edge cases you mentioned, and codifies the manual smoke test from the PR description.

ilblackdragon:
## Code Review

**Verdict: Approve with minor follow-ups**

Defense-in-depth fix for an in-place V6 migration edit: reverts the file, runtime fix-up rewrites the known-bad checksum in `refinery_schema_history`, plus a `migrations/checksums.lock` and CI guard test to prevent recurrence.

### Correctness
Sound. Fix-up is idempotent, only matches the hard-coded bad checksum, hard-errors on canonical/bad collision (not `debug_assert!`). Dual-backend verified — libSQL uses name-keyed `INCREMENTAL_MIGRATIONS` (`src/db/libsql_migrations.rs:760-761`).

### Conventions
Module-owned init in `src/db/migration_fixup.rs`, gated by `#[cfg(feature = "postgres")]`, no `.unwrap()` in prod paths. Errors via `DatabaseError::Migration` with context.

### Risks
- `tracing::warn!` at `migration_fixup.rs:251` — fine because it runs before the TUI starts, but worth a comment about that constraint in case the fix-up ever moves later in startup.
- `Box::leak` in test at line 630 is acceptable in test-only code.
- No concurrency guard — if two processes start simultaneously, both UPDATE the same row, but writes are identical so it's safe.

### Tests
Strong — `released_migrations_are_immutable`, `v6_routines_matches_v018_checksum`, `v6_known_bad_checksum_matches_post_1151_content`, plus a real-PostgreSQL integration test behind `--features integration`.

### Suggestions
1. `src/history/store.rs:71` and `src/setup/wizard.rs:911` duplicate the fix-up call — add a cross-reference comment in both pointing to each other to avoid future divergence.
2. `checksums.lock` is lex-sorted (V10 before V1) — natural-sort would read better in review.
3. Manual upgrade smoke tests listed in the PR body are still unchecked — run before merge.

The defense-in-depth design (revert + runtime fix-up + lockfile test + literal sentinel) is exactly right for this class of bug.

src/db/migration_fixup.rs
```diff
@@ -0,0 +1,641 @@
...
+    }
+
+    fn parse_lockfile(contents: &str) -> HashMap<String, u64> {
+        let mut map = HashMap::new();
+        for (lineno, raw) in contents.lines().enumerate() {
+            let line = raw.trim();
+            if line.is_empty() || line.starts_with('#') {
+                continue;
+            }
+            let (key, value) = line.split_once('=').unwrap_or_else(|| {
```
serrrfirat, line 315: **Low Severity** — Missing unit test for the defensive error path

The check on lines 210–220 that rejects a `KnownDivergence` whose canonical checksum appears in its own `known_bad_checksums` is an important safety net — it prevents a misconfigured entry from silently rewriting already-correct rows. However, this error path has no dedicated test exercising it directly.

The existing `v6_known_bad_checksum_matches_post_1151_content` test verifies the *current* V6 values are distinct, but it doesn't drive the actual error branch in `realign_diverged_checksums_with`.

**Suggested fix:** Add a unit test (no DB needed — the error fires before any query) that constructs a synthetic `KnownDivergence` where `known_bad_checksums` intentionally contains the canonical checksum, then asserts `realign_diverged_checksums_with` returns `Err(DatabaseError::Migration(...))`. Something like:

```rust
#[tokio::test]
async fn rejects_canonical_in_bad_list() {
    // Construct a KnownDivergence where the canonical checksum
    // is listed as a known-bad value — this is a programming error
    // the defensive check should catch.
    let sql = "-- test\nSELECT 1;\n";
    let canonical = Migration::unapplied("V99999__test", sql)
        .unwrap()
        .checksum();
    let divergences = &[KnownDivergence {
        version: 99999,
        name: "test",
        sql,
        known_bad_checksums: Box::leak(Box::new([canonical])),
        explanation: "test",
    }];

    // Need a real PgClient — skip if unavailable, or mock the
    // call since the error fires before any DB interaction.
    // (The check is pre-query, so even a stub connection works.)
}
```

Since the error fires *before* the DB query, you could also extract the validation into a pure function and test it without a connection at all.

src/db/migration_fixup.rs
```diff
@@ -0,0 +1,641 @@
...
+                    lineno + 1
+                )
+            });
+            let parsed: u64 = value.trim().parse().unwrap_or_else(|e| {
+                panic!(
+                    "checksums.lock line {} has invalid u64 checksum {value}: {e}",
+                    lineno + 1
+                )
+            });
+            map.insert(key.trim().to_string(), parsed);
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in 0ac898c4 — `parse_lockfile()` now panics with the duplicate key + line number when `map.insert()` returns `Some`. Closes the silent-overwrite hole.

src/db/migration_fixup.rs
ilblackdragon: Fixed in 0ac898c4 — added `rejects_canonical_in_known_bad_checksums` next to the existing PG integration test, gated on `feature = "integration"`. The test:

1. Creates `refinery_schema_history` so the early-return-on-missing-table branch doesn't mask the validation we want to exercise.
2. Constructs a synthetic `KnownDivergence` (V99998 `bad_config`) where `known_bad_checksums = [canonical]`.
3. Calls `super::realign_diverged_checksums_with(&mut client, bad_divergences)`.
4. Asserts the result is `Err(DatabaseError::Migration(msg))` and that `msg` contains both `known_bad_checksums` and the offending migration label.

You're right that the error fires before any UPDATE so a pure-Rust unit test would also work, but keeping the test next to the existing integration test (which already needs a `PgClient`) lets it use the same skip-if-unreachable harness and exercises the same code path end-to-end through `realign_diverged_checksums_with". Skips gracefully when no `DATABASE_URL` is set.

ilblackdragon:
Thanks for the review @ilblackdragon. Addressed the three actionable follow-ups in 0ac898c4:

1. **`tracing::warn!` placement constraint** — added a comment at the call site explaining that `warn!` is intentional here despite the CLAUDE.md guidance: the realignment runs during database migration at startup, *before* any channel/REPL/TUI is initialized, so terminal-rendering interference is impossible. The comment also flags what to do if the call site ever moves later in startup (downgrade to `debug!` or pre-buffer).

2. **Cross-reference between duplicated call sites** — added a `NOTE: this same fix-up call is duplicated in <other>...` comment to both `src/history/store.rs::Store::run_migrations` and `src/setup/wizard.rs::Wizard::run_migrations_postgres` so future changes to one are obvious in review.

3. **Lex-sorted lockfile** — `regenerate_migration_checksums_lockfile` now sorts by parsed migration version (V1, V2, ..., V10, V11, ...) instead of string order (V10 before V2). Regenerated `migrations/checksums.lock` so it reads in numeric order. Diff in this commit shows the row reordering only — checksums are unchanged.

4. **Manual upgrade smoke tests** — still on me to run before merge. Will do this against (a) a v0.18.0 PostgreSQL database (population A — pre-#1151) and (b) a clean v0.19.0 fresh-install database (population B — post-#1151) and update the PR test plan when complete.

The concurrency-guard observation is correct: simultaneous startups race the same UPDATE but write identical canonical values, so it's safe — and refinery's own `run_async` is already racy in the same way.

src/history/store.rs
```diff
@@ -66,6 +66,17 @@ impl Store {
...
 
         let mut client = self.pool.get().await?;
+
+        // Realign any historically modified migration checksums before
+        // refinery validates them. See `crate::db::migration_fixup` and
+        // issue #1328 for context.
+        //
+        // NOTE: this same fix-up call is duplicated in
+        // `src/setup/wizard.rs::Wizard::run_migrations_postgres`. If you
+        // change one, change the other — both call sites embed refinery
```
serrrfirat: **Medium Severity — Robustness**

**Duplicate call sites for `realign_diverged_checksums` risk future drift**

Both `Store::run_migrations` (here) and `SetupWizard::run_migrations_postgres` (`src/setup/wizard.rs:918`) call the fix-up independently, linked only by cross-referencing comments. A third migration entry point added in the future could forget to include the fix-up — `db/mod.rs::connect_from_config()` also runs migrations, so it's worth verifying it delegates to one of these two paths rather than calling `migrations::runner()` directly.

**Suggested fix:** Extract a shared helper that bundles `realign_diverged_checksums` + `migrations::runner().run_async()` so there's a single call site to maintain:

```rust
// src/db/migration_fixup.rs (or a new src/db/migrate.rs)
pub async fn run_postgres_migrations(client: &mut PgClient) -> Result<(), DatabaseError> {
    realign_diverged_checksums(client).await?;
    migrations::runner()
        .run_async(&mut **client)
        .await
        .map_err(|e| DatabaseError::Migration(e.to_string()))
}
```

The inline comments are a reasonable interim measure, but a single function eliminates the class of bug entirely.

serrrfirat:
## Paranoid Architect Review — `fix(db): repair V6 migration checksum`

Overall this is a well-structured fix with strong defense-in-depth (checksum allowlist, CI guard, sentinel tests). Three MEDIUM findings:

### 1. MEDIUM — Concurrent startup race on checksum realignment

**File:** `src/db/migration_fixup.rs:157` + `src/history/store.rs:78-84`

The checksum fix-up UPDATE and subsequent `refinery::Runner::run_async` are not wrapped in a transaction or protected by an advisory lock. If two replicas start simultaneously against the same database, both could execute `realign_diverged_checksums` concurrently, or one could realign while the other is mid-way through refinery's SELECT-then-compare validation — potentially causing a spurious startup failure on the racing instance.

**Suggested fix:** Acquire `SELECT pg_advisory_xact_lock(0x69726F6E636C6177)` before realignment, release after `run_async` returns. This also hardens the pre-existing migration race. Alternatively, document single-instance startup as a known requirement.

### 2. MEDIUM — `Box::leak` in integration tests

**File:** `src/db/migration_fixup.rs:597,735`

`Box::leak(Box::new([bad]))` deliberately leaks memory to satisfy `&'static [u64]`. While test-only and not a production issue, this will flag under leak sanitizers (Miri, ASan).

**Suggested fix:** Add `// SAFETY: intentional test-only leak` comments, or refactor `KnownDivergence` to be generic over the lifetime of `known_bad_checksums`.

### 3. MEDIUM — Fix-up call duplicated across two call sites

**File:** `src/history/store.rs:70-78` + `src/setup/wizard.rs:910-920`

Both sites independently call `embed_migrations!` + `realign_diverged_checksums` + `runner().run_async()` with a comment-based coupling ("NOTE: this same fix-up call is duplicated"). A third migration entry point could easily miss this.

**Suggested fix:** Extract a single `run_migrations_with_fixup(client)` helper that both call sites invoke.

src/history/store.rs
ilblackdragon: Fixed in 522000a7 — extracted `run_postgres_migrations_with_fixup(client)` in `crate::db::migration_fixup` that bundles fix-up + refinery + the new advisory lock into one function. Both `Store::run_migrations` and `SetupWizard::run_migrations_postgres` now reduce to a single delegated call:

```rust
// store.rs
pub async fn run_migrations(&self) -> Result<(), DatabaseError> {
    let mut client = self.pool.get().await?;
    crate::db::migration_fixup::run_postgres_migrations_with_fixup(&mut client).await
}
```

You're right that comment-based coupling was the wrong long-term answer. Verified there is no third entry point — `db/mod.rs::connect_from_config` calls `pg.run_migrations()` which delegates to `Store::run_migrations` (not refinery directly), so the new helper covers everything.

ilblackdragon:
Thanks for the careful Paranoid Architect review @serrrfirat. All three Medium findings addressed in 522000a7:

**1. Concurrent startup race** — the new `run_postgres_migrations_with_fixup` helper acquires `pg_advisory_lock(1328)` (issue number for grep-ability in `pg_locks`) before realignment and releases it after refinery returns, on every exit path including errors:

```rust
client.execute("SELECT pg_advisory_lock($1)", &[&MIGRATION_LOCK_KEY]).await?;
let result = async {
    realign_diverged_checksums_with(client, KNOWN_DIVERGENCES).await?;
    migrations::runner().run_async(&mut ***client).await...?;
    Ok(())
}.await;
// Always release, even on error.
let _ = client.execute("SELECT pg_advisory_unlock($1)", &[&MIGRATION_LOCK_KEY]).await;
result
```

I went with **session-level** `pg_advisory_lock` rather than your original `pg_advisory_xact_lock` suggestion because refinery's `run_async` opens its own internal transactions, and an outer `xact_lock` would conflict with refinery's transaction boundaries. The session-level lock holds across the inner transactions and is released by the explicit unlock (or by connection drop as a backstop). This also hardens the pre-existing refinery race for multi-replica starts that has always existed in this codebase.

**2. `Box::leak` in tests** — refactored `KnownDivergence` to be lifetime-generic (`KnownDivergence<'a>`). Production `KNOWN_DIVERGENCES` stays as `&[KnownDivergence<'static>]` — no external API change. Both integration tests now use stack-allocated `&[u64]` slices, no `Box::leak`:

```rust
let bad_checksums = [bad];
let test_divergences = [KnownDivergence {
    version: TEST_VERSION,
    name: TEST_NAME,
    sql: TEST_SQL,
    known_bad_checksums: &bad_checksums,
    explanation: "...",
}];
```

Removes the leak-sanitizer false positive entirely.

**3. Duplicate fix-up call sites** — replied inline above. `run_postgres_migrations_with_fixup` is now the single entry point; both `Store::run_migrations` and `SetupWizard::run_migrations_postgres` delegate. Verified no third entry point exists in the codebase.

`cargo test --features integration --lib db::migration_fixup` → 5 passed, 2 ignored (bootstrap helpers). Clippy + no-panics + fmt all clean.

serrrfirat (approved):
Reviewed as part of a paranoid architect audit across 6 PRs. No Critical, High, or Medium findings. The 2 Low and 3 Nit findings (dead code in `realign_diverged_checksums`, duplicate doc comment, alphabetical module ordering) are cosmetic and not worth blocking on. The approach is sound: advisory lock serialization, narrow checksum allowlist, idempotent V13 ALTER, and the `checksums.lock` CI guard prevents recurrence. Ship it.

src/db/migration_fixup.rs
```diff
@@ -0,0 +1,848 @@
...
+    // lock release is less important than reporting the underlying cause.
+    if let Err(e) = client
+        .execute("SELECT pg_advisory_unlock($1)", &[&MIGRATION_LOCK_KEY])
+        .await
+    {
+        tracing::error!(
+            error = %e,
+            "failed to release migration advisory lock — connection drop will \
+             release it eventually, but other replicas may block until then"
+        );
```
serrrfirat, line 171: **Medium Severity — Advisory Lock Held on Unlock Failure**

If the migration sequence succeeds but `pg_advisory_unlock` fails, the function returns `Ok(())` and only logs the unlock failure. The advisory lock is held until the connection drops. In a long-lived connection pool (`deadpool`), this could block other replicas from migrating for the lifetime of the connection.

This is mitigated by `deadpool` eventually returning the connection (advisory locks are session-scoped, released on disconnect), and the error is already logged at `error!` level. However, consider whether the unlock failure should promote the result to `Err` when the migration itself succeeded, since a held advisory lock is a deployment-visible side effect.

PR #2104, opened 2026-04-07
fix(web): emit Done after response — SSE ordering fix (#2079)

### LINKED ISSUE #2079: fix: web UI messages stuck until refresh — SSE event ordering bug

serrrfirat:
## Summary

- **Backend**: Move the terminal "Done" status out of `thread_ops.rs` and emit it only after `agent_loop.rs` successfully sends the response, via a new `respond_then_done()` helper. This guarantees the browser receives the assistant message before the turn-closing SSE event.
- **Frontend safety net**: Track whether a `response` SSE event was received for the current turn. When "Done" arrives without one, trigger `loadHistory()` after 1500ms to recover the message from the server — handles residual edge cases like proxy buffering or brief SSE disconnects.
- **Regression test**: Asserts the response event is captured before the Done status in an ordered event log.

Closes #2079

## Test plan

- [x] `cargo test -p ironclaw --test e2e_response_order response_order_tests::response_arrives_before_done_status`
- [x] Manual: send message in web UI, verify in browser devtools EventStream that `response` arrives before `Done`
- [ ] Manual: simulate lost response (e.g. throttle network) — verify history reloads after 1500ms

### DIFF
--- a/src/agent/agent_loop.rs
+++ b/src/agent/agent_loop.rs
@@ -33,6 +33,41 @@ use crate::workspace::Workspace;
 use ironclaw_safety::SafetyLayer;
 use ironclaw_skills::SkillRegistry;
 
+/// Outcome of [`Agent::handle_message`] — drives the run-loop's response/Done dispatch.
+///
+/// Distinguishes "no response, turn is over" from "no response, turn is paused"
+/// so the run loop can decide whether to emit the terminal `Done` status. Sending
+/// `Done` after a pause (e.g. while awaiting tool approval) is incorrect because
+/// the thread is not in a terminal state, and would also trip the web UI's
+/// missing-response safety net (see #2079).
+#[derive(Debug)]
+enum HandleOutcome {
+    /// Shutdown signal (e.g. `/quit`). Run loop should break.
+    Shutdown,
+    /// Send this content via the channel, then emit terminal `Done`.
+    Respond(String),
+    /// No response to send, but the turn is complete — emit `Done` only.
+    NoResponse,
+    /// Turn is paused (awaiting approval/auth/etc). Do not emit `Done`.
+    Pending,
+}
+
+impl HandleOutcome {
+    /// Convert a legacy `Option<String>` return into a [`HandleOutcome`].
+    ///
+    /// `None` → `Shutdown`, empty string → `NoResponse`, otherwise `Respond`.
+    /// Used to wrap bridge handlers that still return `Option<String>`. Bridge
+    /// approval flows return non-empty descriptive text, so they never need the
+    /// `Pending` variant — only the v1 `process_user_input` path does.
+    fn from_legacy(opt: Option<String>) -> Self {
+        match opt {
+            None => HandleOutcome::Shutdown,
+            Some(s) if s.is_empty() => HandleOutcome::NoResponse,
+            Some(s) => HandleOutcome::Respond(s),
+        }
+    }
+}
+
 /// Static greeting persisted to DB and broadcast on first launch.
 ///
 /// Collapse a tool output string into a single-line preview for display.
@@ -992,7 +1027,7 @@ impl Agent {
             self.store_extracted_documents(&message).await;
 
             match self.handle_message(&message).await {
-                Ok(Some(response)) if !response.is_empty() => {
+                Ok(HandleOutcome::Respond(response)) => {
                     // Hook: BeforeOutbound — allow hooks to modify or suppress outbound
                     let event = crate::hooks::HookEvent::Outbound {
                         user_id: message.user_id.clone(),
@@ -1035,18 +1070,28 @@ impl Agent {
                         }
                     }
                 }
-                Ok(Some(empty)) => {
-                    // Empty response, nothing to send (e.g. approval handled via send_status).
-                    // Still send Done so the client knows the turn is complete.
+                Ok(HandleOutcome::NoResponse) => {
+                    // Empty response (e.g. routine consumed the message, silent reply).
+                    // Send Done so the client knows the turn is complete.
                     tracing::debug!(
                         channel = %message.channel,
                         user = %message.user_id,
-                        empty_len = empty.len(),
                         "Suppressed empty response (not sent to channel)"
                     );
                     self.send_done(&message).await;
                 }
-                Ok(None) => {
+                Ok(HandleOutcome::Pending) => {
+                    // Turn paused awaiting user action (approval, auth, etc).
+                    // Do NOT emit Done — the thread is not in a terminal state.
+                    // The relevant ApprovalNeeded/AuthRequired status was already
+                    // sent by the inner handler before returning.
+                    tracing::debug!(
+                        channel = %message.channel,
+                        user = %message.user_id,
+                        "Turn paused (Pending); suppressing Done"
+                    );
+                }
+                Ok(HandleOutcome::Shutdown) => {
                     // Shutdown signal received (/quit, /exit, /shutdown)
                     tracing::debug!("Shutdown command received, exiting...");
                     break;
@@ -1153,7 +1198,7 @@ impl Agent {
         }
     }
 
-    async fn handle_message(&self, message: &IncomingMessage) -> Result<Option<String>, Error> {
+    async fn handle_message(&self, message: &IncomingMessage) -> Result<HandleOutcome, Error> {
         // Log sensitive details at debug level for troubleshooting
         tracing::debug!(
             message_id = %message.id,
@@ -1174,7 +1219,7 @@ impl Agent {
                 channel = %message.channel,
                 "Forwarding internal message"
             );
-            return Ok(Some(message.content.clone()));
+            return Ok(HandleOutcome::Respond(message.content.clone()));
         }
 
         // Set message tool context for this turn (current channel and target)
@@ -1204,10 +1249,16 @@ impl Agent {
             };
             match self.hooks().run(&event).await {
                 Err(crate::hooks::HookError::Rejected { reason }) => {
-                    return Ok(Some(format!("[Message rejected: {}]", reason)));
+                    return Ok(HandleOutcome::Respond(format!(
+                        "[Message rejected: {}]",
+                        reason
+                    )));
                 }
                 Err(err) => {
-                    return Ok(Some(format!("[Message blocked by hook policy: {}]", err)));
+                    return Ok(HandleOutcome::Respond(format!(
+                        "[Message blocked by hook policy: {}]",
+                        err
+                    )));
                 }
                 Ok(crate::hooks::HookOutcome::Continue {
                     modified: Some(new_content),
@@ -1220,11 +1271,16 @@ impl Agent {
             }
         }
 
-        // Engine V2 routing (Strategy C: parallel deployment)
+        // Engine V2 routing (Strategy C: parallel deployment).
+        // Bridge handlers return `Option<String>` legacy results; wrap with
+        // `HandleOutcome::from_legacy`. Bridge approval flows emit non-empty
+        // descriptive text, so they map to `Respond` (not `Pending`).
         if self.config.engine_v2 {
             match &submission {
                 Submission::UserInput { content } => {
-                    return crate::bridge::handle_with_engine(self, message, content).await;
+                    return crate::bridge::handle_with_engine(self, message, content)
+                        .await
+                        .map(HandleOutcome::from_legacy);
                 }
                 Submission::ApprovalResponse { approved, always } => {
                     // If there's a pending auth, "cancel"/"no" should clear the
@@ -1232,9 +1288,13 @@ impl Agent {
                     // Route through handle_with_engine so PendingAuth is checked.
                     if crate::bridge::has_pending_auth(&message.user_id).await {
                         let content = &message.content;
-                        return crate::bridge::handle_with_engine(self, message, content).await;
+                        return crate::bridge::handle_with_engine(self, message, content)
+                            .await
+                            .map(HandleOutcome::from_legacy);
                     }
-                    return crate::bridge::handle_approval(self, message, *approved, *always).await;
+                    return crate::bridge::handle_approval(self, message, *approved, *always)
+                        .await
+                        .map(HandleOutcome::from_legacy);
                 }
                 Submission::ExecApproval {
                     request_id,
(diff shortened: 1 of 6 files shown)


ilblackdragon:
# Code Review

## Overview
Fixes a SSE ordering race (#2079) where the web UI could observe the turn-closing `Done` status before the assistant `response` event, causing the response to be missed. The fix:
- Removes two `send_status("Done")` calls from `thread_ops.rs`.
- Adds a `respond_then_done()` helper in `agent_loop.rs` and routes all four outbound paths through it.
- Adds a frontend safety net that calls `loadHistory()` 1.5s after a `Done` if no `response` was seen for the turn.
- Adds a regression test (`tests/e2e_response_order.rs`) that asserts strict ordering on a captured event log.

## What works well
- Correct architectural move: terminal status now lives where outbound dispatch lives. `thread_ops.rs` no longer needs to know about channel lifecycle.
- The new `CapturedEvent` enum on `TestChannel` is the right tool for testing ordering invariants — much cleaner than two parallel vectors.
- Frontend cleanup is symmetric: `_turnResponseReceived` and `_doneWithoutResponseTimer` are reset on `sendMessage`, `switchThread`, and SSE `onopen`. No leaked timers across thread switches.
- `_turnResponseReceived = true` is set in *both* the `response` and `chunk` handlers (`app.js:758`), so streaming responses won't trigger spurious history reloads.

## Issues

### 1. `respond_then_done` swallows `Done` when `respond` fails (regression)
`src/agent/agent_loop.rs:319`
```rust
self.channels.respond(message, response).await?;   // ← early return on Err
if let Err(e) = self.channels.send_status(..., \"Done\", ...).await { ... }
```
If `respond` returns `Err`, the helper returns `Err` and `Done` is never emitted. The old code emitted `Done` unconditionally (before the response). The web UI's `enableChatInput()` only fires on `Done`, so a transient `respond` failure will now leave the input disabled forever — and the new safety net does NOT cover this case (it triggers only when `Done` arrives without `response`, not when neither arrives).

Suggestion: emit `Done` regardless of `respond` outcome:
```rust
let respond_result = self.channels.respond(message, response).await;
let _ = self.channels.send_status(&message.channel,
    StatusUpdate::Status(\"Done\".into()), &message.metadata).await;
respond_result
```
This also lets you delete the duplicated inline `send_status(\"Done\")` blocks at `agent_loop.rs:984-996, 1037-1049` (hook-blocked and empty-response paths) — they exist precisely because the helper doesn't guarantee `Done`.

### 2. Duplicated `send Done` boilerplate
The same 12-line `send_status(\"Done\") + warn!` block appears three times inline in `agent_loop.rs` (helper, hook-blocked path, empty-response path). After fixing #1 above, these collapse into a single `send_done(&message)` helper invoked from one place in the helper plus the empty/hook-blocked branches.

### 3. Test coverage gap
`tests/e2e_response_order.rs` only covers the happy `Ok(Some(non_empty))` path. The bug class is \"Done arrives in the wrong order\" — but the patch touches **five** code paths that emit `Done`:
- happy path (covered)
- hook-blocked (`Err(err)` branch)
- modified-by-hook (`HookOutcome::Continue { Some }`)
- empty response
- error response (`Err(e)`)
- drain-loop intermediate response

A parameterized version of the test (or even one extra case for the empty-response branch) would prevent regressions on the inline `Done` emissions, which are the most fragile part of this PR.

### 4. Frontend: 1500 ms is a magic number
`app.js:799` — `setTimeout(..., 1500)`. Worth lifting to a named constant near the other SSE constants (e.g. `STREAM_DEBOUNCE_MS`) for greppability and to make the safety-net delay a single point of tuning.

### 5. `_turnResponseReceived` is global, not per-thread
If a background thread emits a `response` while the user is viewing a different thread, the early `isCurrentThread(...)` guard returns before `_turnResponseReceived` is set. That's actually correct, but the inverse — switching *to* a thread mid-turn — relies on `switchThread` resetting state. The current cleanup looks correct, but a comment noting \"single-thread tracking is intentional; per-thread state isn't needed because background threads filter by `isCurrentThread`\" would help future readers.

## Minor / nits
- `agent_loop.rs:319` — doc comment is good; consider also noting that `Done` is only sent on successful `respond` (until issue #1 above is fixed).
- `test_channel.rs:175` — `try_lock().expect(\"captured_events lock contention\")` follows the existing pattern but could panic under unusual test interleavings; matches local convention so acceptable.
- The PR description mentions \"broadcast lag\" as a possible cause — worth a sentence in the commit/code comment about whether the underlying broadcast bus actually preserves FIFO ordering between `respond` and `send_status`.

## Risk assessment
- **Backend correctness**: Medium risk — the `respond_then_done` early-return on `respond` failure is a real behavior change that should be addressed before merging.
- **Frontend**: Low risk — safety net is additive and well-guarded.
- **Test coverage**: Adequate for the happy path; insufficient for the four other branches the patch touches.

## Recommendation
**Request changes** — fix issue #1 (always emit `Done`, even when `respond` fails) and ideally collapse the duplicated boilerplate from #2. Adding one extra test case for the empty-response branch would close the most important coverage gap.

serrrfirat:
All five issues addressed in e23d47de:

1. **`respond_then_done` always emits Done** — stores the `respond` result, emits `Done` unconditionally, then returns the result. Client always knows the turn ended, even on delivery failure.
2. **Deduplicated boilerplate** — extracted `send_done()` helper method; replaced both inline 12-line blocks (hook-blocked + empty-response paths) with single-line calls.
3. **Test coverage** — added `done_emitted_for_empty_response` test covering the empty-response branch: asserts Done is emitted, no Response event is produced, and Done appears in captured events.
4. **Magic number** — lifted `1500` to `const DONE_WITHOUT_RESPONSE_TIMEOUT_MS = 1500` near the other SSE constants.
5. **Comment** — added note above `_turnResponseReceived` explaining single-thread tracking is intentional since background thread events are filtered by `isCurrentThread`.

ilblackdragon:
# Code Review

## Overview

Fixes #2079 by reordering the SSE event stream so the assistant `response` event is emitted before the terminal `Done` status. The fix is implemented at three levels:

1. **Backend**: Removes premature `Done` emissions from `thread_ops.rs` and introduces `respond_then_done()` / `send_done()` helpers in `agent_loop.rs` that fire `Done` after `respond()` returns.
2. **Frontend safety net**: Tracks per-turn `_turnResponseReceived` and calls `loadHistory()` 1.5s after a `Done` arrives without a preceding response.
3. **Test**: New `e2e_response_order.rs` regression test plus `CapturedEvent` ordering instrumentation in `TestChannel`.

The architectural direction is right — moving `Done` emission to a single, well-defined point above `respond()` is much cleaner than scattering it across `thread_ops.rs`.

---

## Issues

### Regression: `Done` now fires after `Awaiting approval`

`agent_loop.rs::handle_message` translates `SubmissionResult::NeedApproval` into `Ok(Some(String::new()))` (`src/agent/agent_loop.rs:1654-1659`). After this PR, the run-loop's empty arm at `src/agent/agent_loop.rs:984-992` calls `self.send_done(&message).await` for every empty response — including the approval-pending case.

End-to-end consequence in the web UI:

1. Backend emits `approval_needed` → frontend renders approval prompt (`app.js:778`).
2. Backend emits `status: Done` (new behavior).
3. Frontend `_turnResponseReceived` is still `false` because no `response` event arrived.
4. The 1500 ms safety net at `app.js:798-808` fires and calls `loadHistory()` underneath the live approval prompt.

The thread is in `ThreadState::AwaitingApproval`, not `Completed` — emitting `Done` is semantically wrong and races against the new safety net the same PR introduces. Two practical fixes:

- **Preferred**: Distinguish \"empty response\" from \"approval pending\" in `handle_message`'s return — e.g. introduce a `Pending` variant or return a small enum so the run loop knows not to call `send_done` on the approval path.
- **Quick patch**: Before calling `send_done` in the empty arm, check the thread state and skip when it's `AwaitingApproval`. (Requires reading session state, which is racy but probably fine for a defensive guard.)

The author's comment (\"Empty response, nothing to send (e.g. approval handled via send_status). Still send Done so the client knows the turn is complete.\") suggests this case was considered, but the conclusion conflicts with the new frontend safety net the same PR adds.

### Test gap for the approval path

`e2e_response_order.rs` covers the happy path and the empty-response path, but not the `NeedApproval` path. Whichever resolution you pick above, please add a test that runs an approval-requiring tool and asserts the captured-event sequence after `ApprovalNeeded` (no `Done`, or `Done` with whatever the desired behavior is).

### Drain-loop intermediate response is silent on `Done`

In the queue-drain branch at `src/agent/agent_loop.rs:1466-1475`, the intermediate `respond()` between turns intentionally does **not** call `respond_then_done`. That's defensible (the conversation isn't done yet), but it's the only `respond()` call left in `agent_loop.rs` that doesn't pair with `send_done`. Worth a one-line comment so the next person doesn't \"fix\" it.

---

## Smaller observations

- `respond_then_done` always emits `Done` even when `respond()` fails (`src/agent/agent_loop.rs:316-340`). The doc-comment says this is intentional; agreed, but consider whether the client also wants an `error` SSE event in that case (out of scope for this PR).
- `send_done` uses `tracing::warn!` for failed sends. Per `CLAUDE.md`'s logging rules, `warn!` corrupts the REPL/TUI — but `Done` only matters for SSE-style channels and the warn is for a real error, so this is fine.
- The `_turnResponseReceived` reset is plumbed into `connectSSE`, `sendMessage`, `switchThread`, and the Done handler itself — nice coverage. Worth verifying that nothing else (e.g. `loadHistory` finishing, manual SSE reconnect via `Last-Event-ID`) could leave the flag stale across turns.
- `CapturedEvent` and `wait_for_done` in `tests/support/test_channel.rs` are clean reusable additions. The exponential backoff in `wait_for_done` (50 ms → 500 ms cap) is reasonable.
- `DONE_WITHOUT_RESPONSE_TIMEOUT_MS = 1500` is a tunable magic number — fine as a constant, but worth a comment on how it was chosen relative to typical broadcast lag.
- The new `respond_then_done` and `send_done` helpers are well-documented and small enough to keep colocated in `agent_loop.rs`. No notes on style.

---

## Recommendation

**Request changes** on the `Done`-after-`Awaiting approval` regression and missing approval-path test. The other observations are minor and can be addressed in the same revision or follow-up.

The core architectural change is correct and the regression test is good — once the approval path is handled, this is a clean fix.

ilblackdragon:
Pushed 7d3f8040 to address the `Done`-after-`Awaiting approval` regression I flagged above.

## What changed

Introduced a `HandleOutcome` enum so `Agent::handle_message` can distinguish "no response, turn complete" from "no response, turn paused":

```rust
enum HandleOutcome {
    Shutdown,         // /quit, single-message REPL exit
    Respond(String),  // send response, then Done
    NoResponse,       // empty response, send Done only
    Pending,          // turn paused — DO NOT send Done
}
```

The previous code lumped both empty cases into `Ok(Some(""))`, so `SubmissionResult::NeedApproval` returning an empty string was indistinguishable from a routine consuming a message. With the new enum, `NeedApproval` maps to `HandleOutcome::Pending` and the run loop's empty arm only fires `send_done()` for `NoResponse`.

Bridge handlers (engine v2) still return `Result<Option<String>, Error>` and are wrapped with `HandleOutcome::from_legacy`. Their approval flows emit non-empty descriptive text via `insert_and_notify_pending_gate`, so they always map to `Respond` — the legacy v1 `process_user_input` path is the only one that ever needs `Pending`.

## Tests

Added a regression test `no_done_emitted_while_awaiting_approval` that drives a v1 `Always`-approval probe tool and asserts:

1. `ApprovalNeeded` status is captured
2. No `Response` event is emitted (the tool never executed)
3. **No `Done` status is emitted** — the bug guard
4. Exactly one `ApprovalNeeded` event is captured

I also had to repair the pre-existing `done_emitted_for_empty_response` test, which was failing on the PR baseline. The dispatcher in `src/llm/reasoning.rs` substitutes empty LLM responses with `"I'm not sure how to respond to that."`, so a truly empty response never reaches the run loop — the test was asserting an unreachable condition. Renamed it to `done_emitted_after_empty_response_fallback` and updated it to assert the actual ordering invariant (fallback response → Done).

## Verification

- `cargo fmt --check` — clean
- `cargo test -p ironclaw --test e2e_response_order` — 3/3 pass
- `cargo test -p ironclaw --test e2e_engine_v2` — 9/9 pass (engine v2 path unaffected)
- `cargo test -p ironclaw --lib` — 4266/4266 pass

The other smaller observations from my review (drain-loop intermediate `respond()` comment, doc-comment on the timeout constant) are still open if you'd like to address them in this PR or a follow-up.

ilblackdragon (approved):

PR #2107, opened 2026-04-07
feat(workspace): admin system prompt shared with all users

### LINKED ISSUE #2088: System prompt created by an admin should be picked up as other users' prompt
Use case: 
As an admin, I want to create a system prompt for all users to use. For example, information about the custom setup and agent capabilities.

serrrfirat:
## Summary
- Introduces `SYSTEM.md` in a well-known `__admin__` scope so admins can set a system prompt that all tenants receive
- Gated behind multi-tenant mode — `WorkspacePool` sets `admin_prompt_enabled` on each workspace; zero overhead in single-user deployments
- New admin-only API endpoints: `GET/PUT /api/admin/system-prompt`

Closes #2088

### How it works
1. Admin writes `SYSTEM.md` to `__admin__` scope via `PUT /api/admin/system-prompt`
2. In `system_prompt_for_context_inner()`, when `admin_prompt_enabled` is true, reads `SYSTEM.md` directly from `__admin__` scope (not via multi-scope reads)
3. Injected as "## System Instructions" section **before** per-user identity files (AGENTS.md, SOUL.md, etc.)
4. Injection scanning protects against prompt injection in admin content

### Multi-tenancy gating
- **API layer**: Handlers check `workspace_pool.is_some()` → 404 in single-user mode
- **Prompt layer**: `admin_prompt_enabled` flag on Workspace (set only by WorkspacePool) → no DB hit in single-user mode

### Files changed
| File | Change |
|------|--------|
| `src/workspace/document.rs` | `paths::SYSTEM` + `ADMIN_SCOPE` constants |
| `src/workspace/mod.rs` | Injection scan list, `admin_prompt_enabled` field/builder, gated read in prompt assembly |
| `src/channels/web/server.rs` | `.with_admin_prompt()` in WorkspacePool, route registration |
| `src/channels/web/types.rs` | Request/response DTOs |
| `src/channels/web/handlers/system_prompt.rs` | **New** — GET/PUT handlers |
| `src/channels/web/handlers/mod.rs` | Module registration |
| `tests/admin_system_prompt.rs` | **New** — 6 integration tests |

## Test plan
- [x] `cargo fmt` — clean
- [x] `cargo test --lib` — 3938 passed
- [x] `cargo test --test admin_system_prompt --features libsql` — 6 passed
- [ ] Verify admin can set/get/clear system prompt via API
- [ ] Verify non-admin gets 403 on system prompt endpoints
- [ ] Verify single-user mode returns 404

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,5 +1,5 @@
-# Scope labels for actions/labeler@v6
-# Maps file path globs to scope labels. Multiple labels can apply per PR.
+# Labels for actions/labeler@v6
+# Maps file path globs to labels. Multiple labels can apply per PR.
 
 "scope: agent":
   - changed-files:
@@ -164,3 +164,9 @@
       - any-glob-to-any-file:
           - Cargo.toml
           - Cargo.lock
+
+"DB MIGRATION":
+  - changed-files:
+      - any-glob-to-any-file:
+          - migrations/**
+          - src/db/libsql_migrations.rs
--- a/.github/scripts/create-labels.sh
+++ b/.github/scripts/create-labels.sh
@@ -62,6 +62,9 @@ create "scope: ci"            "546E7A" "CI/CD workflows"
 create "scope: docs"          "78909C" "Documentation"
 create "scope: dependencies"  "90A4AE" "Dependency updates"
 
+echo "==> Creating coordination labels..."
+create "DB MIGRATION"         "C62828" "PR adds or modifies PostgreSQL or libSQL migration definitions"
+
 echo "==> Creating workflow labels..."
 create "skip-regression-check" "9E9E9E" "Acknowledged: fix without regression test"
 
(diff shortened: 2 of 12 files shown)


serrrfirat:
### Re: caching the admin prompt read

> In multi-tenant mode, `system_prompt_for_context_inner` is called on every message turn. Reading the admin prompt directly from storage on every call introduces a recurring database query...

Valid observation, but intentionally not addressed here. Every other system prompt file (SOUL.md, AGENTS.md, USER.md, IDENTITY.md, TOOLS.md, MEMORY.md, daily logs) follows the same read-from-DB-on-every-turn pattern in `system_prompt_for_context_inner()`. Adding caching for just the admin prompt would be inconsistent with the existing architecture.

The query is a simple key-value lookup (`user_id + agent_id + path`) hitting an indexed unique constraint — very fast. If prompt assembly DB reads become a bottleneck in high-concurrency deployments, the right fix is to cache the entire assembled prompt (all files), not just this one.

Keeping it simple per the project's preference against over-engineering.

PR #2119, opened 2026-04-07
fix(acp): safety-net broadcast and idempotent terminal transitions (#1981 follow-up)

rajulbhatnagar:
## Summary

Follow-up to #1981 (propagate follow-up prompt failures as job errors).

- **Safety-net broadcast in `report_complete`**: Even if the worker's fire-and-forget `post_event("result")` is lost, the orchestrator now broadcasts a `JobResult` so job monitors free the `max_jobs` slot. Duplicates are harmless — consumers break on the first `JobResult`.
- **Idempotent terminal self-transitions**: Extended from just `Completed→Completed` to all terminal states (`Failed`, `Accepted`, `Cancelled`) using `is_terminal()`. Cross-terminal transitions (e.g. `Failed→Cancelled`) are explicitly blocked.
- **DB fallback error logging**: `broadcast_app_event` now logs at `debug\!` when the DB owner lookup fails or returns no job, instead of silently swallowing the error.
- **Duplicate-event test**: Proves the completion watcher handles back-to-back `JobResult` events (simulating `post_event` + safety-net both arriving) without error.
- **Log level fix**: Demoted transient poll error log from `warn\!` to `debug\!` in `acp_bridge` to avoid REPL/TUI noise.

## Test plan

- [x] `cargo fmt` — clean
- [x] `cargo test --lib` — 4295 passed
- [x] Verify safety-net broadcast fires when worker `post_event` is lost (manual or E2E)
- [x] Verify duplicate `JobResult` does not cause double state transition or slot leak

### DIFF
--- a/src/agent/job_monitor.rs
+++ b/src/agent/job_monitor.rs
@@ -531,4 +531,120 @@ mod tests {
         let ctx = cm.get_context(job_id).await.unwrap();
         assert_eq!(ctx.state, JobState::Completed);
     }
+
+    // === Regression: report_complete safety-net broadcast frees max_jobs slot ===
+    // When the worker's fire-and-forget post_event("result") is lost, the
+    // orchestrator's report_complete handler now broadcasts a safety-net
+    // JobResult. This test proves that a JobResult with session_id: None
+    // (as emitted by report_complete, which lacks session context) still
+    // transitions the job and frees the slot.
+
+    #[tokio::test]
+    async fn completion_watcher_frees_slot_from_report_complete_broadcast() {
+        use crate::context::{ContextManager, JobState};
+
+        let max_jobs = 2;
+        let cm = Arc::new(ContextManager::new(max_jobs));
+
+        // Fill both slots.
+        let job_a = Uuid::new_v4();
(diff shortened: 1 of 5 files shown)


rajulbhatnagar:
@serrrfirat Follow up you requested in #1981

serrrfirat:
Thank you for hardening terminal job transitions and preventing lost completion events from leaking concurrency slots. We are closing this PR because the ACP/job-monitor lifecycle it changes was replaced by Reborn’s canonical turn-state and lease contracts.

Reborn persists terminal transitions in `crates/ironclaw_turns`; terminal states release the active lock, runner mutations require the matching runner and lease token, repeated cancellation reports `already_terminal`, and cross-terminal mutations are rejected as invalid transitions. Completion is observed through durable state and lifecycle projections rather than relying on duplicate fire-and-forget `JobResult` broadcasts.

The idempotency principle here remains important. We would welcome your help with Reborn terminal-transition, observer retry, or concurrency-slot regression tests. Thank you for identifying and closing the lost-event safety gap in the old system.

Issue #2121, opened 2026-04-07
feat(web): visible Stop control and hard cancel for active chat turns


G7CNF:
The web UI currently exposes `/interrupt` / `/stop` as slash commands, but there is no visible in-turn Stop control and the current interrupt path is cooperative only.

Symptoms:
- Users can see the agent stay in `Processing...` for minutes with no obvious control in the chat UI.
- `/stop` in chat is not reliable for aborting a long-running in-flight LLM/tool turn because it is queued through the normal message pipeline.

Requested behavior:
- Replace Send with Stop while a turn is executing.
- Make the stop action abort the active turn out-of-band instead of waiting for the turn queue.
- Preserve the existing `/interrupt` / `/stop` command path as a fallback.

Scope notes:
- This is separate from the SSE ordering fix in #2079.
- The fix likely needs a small backend cancel endpoint plus a web UI state switch.

PR #2122, opened 2026-04-07
fix(web): add stop control for active chat turns

G7CNF:
Implements #2121 by adding a visible Stop control for active chat turns and a direct interrupt endpoint that cancels the running turn instead of relying on the normal chat pipeline.

What changed:
- the chat send button now swaps to Stop while a turn is running
- added  for out-of-band cancellation
- thread state now tracks a cancellation token for the active turn
- the agent loop and tool execution paths observe cancellation mid-flight
- added a regression test for interrupt-token cancellation
- updated the web gateway docs and feature parity matrix

Validation:
- cargo fmt --all
- cargo check -q
- cargo test -p ironclaw agent::session::tests::test_thread_interrupt_and_resume -- --nocapture

### DIFF
--- a/src/agent/dispatcher.rs
+++ b/src/agent/dispatcher.rs
@@ -7,6 +7,7 @@ use std::sync::Arc;
 
 use tokio::sync::Mutex;
 use tokio::task::JoinSet;
+use tokio_util::sync::CancellationToken;
 use uuid::Uuid;
 
 use crate::agent::Agent;
@@ -253,6 +254,13 @@ impl Agent {
             thread_id,
             message,
             job_ctx,
+            turn_cancel: {
+                let sess = session.lock().await;
+                sess.threads
+                    .get(&thread_id)
+                    .and_then(|thread| thread.current_turn_cancel())
+                    .unwrap_or_else(CancellationToken::new)
+            },
             active_skills,
             cached_prompt,
             cached_prompt_no_tools,
@@ -369,6 +377,7 @@ struct ChatDelegate<'a> {
     thread_id: Uuid,
     message: &'a IncomingMessage,
     job_ctx: JobContext,
+    turn_cancel: CancellationToken,
     active_skills: Vec<ironclaw_skills::LoadedSkill>,
     cached_prompt: String,
     cached_prompt_no_tools: String,
@@ -405,10 +414,7 @@ impl ChatDelegate<'_> {
 #[async_trait]
 impl<'a> LoopDelegate for ChatDelegate<'a> {
     async fn check_signals(&self) -> LoopSignal {
-        let sess = self.session.lock().await;
-        if let Some(thread) = sess.threads.get(&self.thread_id)
-            && thread.state == ThreadState::Interrupted
-        {
+        if self.turn_cancel.is_cancelled() {
             return LoopSignal::Stop;
         }
         LoopSignal::Continue
@@ -628,7 +634,17 @@ impl<'a> LoopDelegate for ChatDelegate<'a> {
             }
         }
 
-        let output = match reasoning.respond_with_tools(reason_ctx).await {
+        let output = match tokio::select! {
+            biased;
+            _ = self.turn_cancel.cancelled() => {
+                return Err(crate::error::JobError::ContextError {
+                    id: self.thread_id,
+                    reason: "Interrupted".to_string(),
+                }
+                .into());
+            }
+            output = reasoning.respond_with_tools(reason_ctx) => output,
+        } {
             Ok(output) => output,
             Err(crate::error::LlmError::ContextLengthExceeded { used, limit }) => {
                 tracing::warn!(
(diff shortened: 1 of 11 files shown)


henrypark133 (dismissed):
## Review: CancellationToken-based stop control for web chat turns

Clean, well-structured implementation. The `tokio_util::CancellationToken` approach is the canonical async cancellation pattern for Tokio and is applied correctly throughout.

Positives:
- `check_signals()` no longer acquires the session lock — nice performance improvement on the hot path
- Multiple `tokio::select! { biased; }` checkpoints (LLM call, tool execution, parallel joins) ensure the interrupt is responsive regardless of where the loop is
- Interrupt endpoint is properly authenticated (`AuthenticatedUser`) and handles the no-active-thread case
- `#[serde(skip)]` on the token avoids serialization issues
- Redundant `setTurnRunning(true)` calls across SSE event handlers are a good safety net for missed events
- Regression test covers the core mechanic (token lifecycle + state transition)

### Minor: Missing Korean i18n key

`ko.js` is missing the new `'chat.stop'` key (added for en and zh-CN). The `I18n.t('chat.stop') || 'Stop'` fallback means it won't break, but the Korean translation should be added for consistency. Likely `'chat.stop': '중지'`.

### Minor: Disconnected fallback token

In `dispatcher.rs`, if `current_turn_cancel()` returns `None`, the fallback creates a disconnected `CancellationToken::new()` that nobody holds a cancel handle for — the stop button would silently fail for that turn. This should be extremely rare since `start_turn()` sets the token before the delegate is created, but a `debug!` log on the fallback path would make debugging easier if it ever hits.

### Minor: Repeated cancellation select blocks

The `tokio::select! { biased; _ = self.turn_cancel.cancelled() => ... }` pattern + error construction is duplicated 4 times. A small helper method like `async fn cancellable<T>(&self, fut: impl Future<Output = T>) -> Result<T, Error>` could reduce this. Non-blocking observation.

Minor notes:
- `ContextError` for user-initiated interruptions is semantically imprecise (it's not a context error), but the string "Interrupted" propagates correctly through SSE to the frontend, so it works
- Enter key correctly does nothing while a turn is running (prevents accidental interrupts) — good UX choice

src/agent/dispatcher.rs
```diff
@@ -405,10 +420,7 @@ impl ChatDelegate<'_> {
 #[async_trait]
 impl<'a> LoopDelegate for ChatDelegate<'a> {
     async fn check_signals(&self) -> LoopSignal {
-        let sess = self.session.lock().await;
-        if let Some(thread) = sess.threads.get(&self.thread_id)
-            && thread.state == ThreadState::Interrupted
-        {
+        if self.turn_cancel.is_cancelled() {
```
serrrfirat, line 423: **Medium Severity** · Confidence: Likely

**`check_signals()` state-check removal creates a hidden non-interruptible turn failure mode**

The old `check_signals()` checked `thread.state == ThreadState::Interrupted` under the session mutex — a reliable fallback that always worked regardless of token state. The new implementation only checks `self.turn_cancel.is_cancelled()`, which is lock-free and faster, but removes the safety net.

Combined with the detached fallback token at lines 257–266:
```rust
.unwrap_or_else(|| {
    tracing::debug!(...);
    CancellationToken::new()  // ← detached, can never be cancelled
})
```

If the fallback path is ever triggered (e.g., a future code change breaks the `start_turn()` → dispatch ordering), the turn becomes **completely non-interruptible** — neither `check_signals()` nor the `tokio::select!` branches will ever fire. The old code was resilient to this scenario; the new code is not.

**Suggested fix** — keep a dual check (belt-and-suspenders):
```rust
async fn check_signals(&self) -> LoopSignal {
    if self.turn_cancel.is_cancelled() {
        return LoopSignal::Stop;
    }
    // Fallback: state-based detection in case token is detached
    let sess = self.session.lock().await;
    if let Some(thread) = sess.threads.get(&self.thread_id)
        && thread.state == ThreadState::Interrupted
    {
        return LoopSignal::Stop;
    }
    LoopSignal::Continue
}
```

Or, make the fallback an error instead of silent degradation — return `Err` rather than creating a useless detached token.

tests/stop_control_regression.rs
```diff
@@ -0,0 +1,24 @@
...
+    };
+
+    assert!(!cancel.is_cancelled(), "token should start active");
+
+    thread.interrupt();
+
+    assert_eq!(thread.state, ThreadState::Interrupted);
+    assert!(cancel.is_cancelled(), "interrupt should cancel the token");
+    assert!(thread.current_turn_cancel().is_some());
+}
```
serrrfirat, line 24: **Medium Severity** · Confidence: Possible

**Test validates helper, not the caller — missing caller-level coverage per project testing rules**

This regression test validates the `Thread` struct directly (`start_turn()` → `interrupt()` → assert token cancelled), which is good for unit coverage. However, per the project's "Test Through the Caller, Not Just the Helper" rule in `.claude/rules/testing.md`, the cancellation token is a **helper that gates the side-effecting agentic loop**. A unit test on the helper alone is not sufficient regression coverage when:

1. The helper (token) is a predicate that gates a side effect (the agentic loop)
2. There is a wrapper (`ChatDelegate.check_signals()`, `tokio::select!` branches) between the helper and the side effect
3. The helper has computed inputs (token read from session under lock, with a fallback path)

All three conditions from the rule are met. An integration-tier test is needed that drives the actual call site — e.g., constructing a `ChatDelegate` with a pre-cancelled token and asserting `check_signals()` returns `LoopSignal::Stop`, or driving `dispatch_chat_turn` and verifying early loop exit on cancellation.

This would also catch the detached-fallback-token issue noted in the other review comment — a caller-level test using the fallback path would reveal that the turn is non-interruptible.

serrrfirat:
Thank you for adding a visible way to stop active work and for carrying cancellation into the old agent/tool paths. We are closing this PR because the feature is now implemented as a first-class Reborn run transition.

WebUI v2 posts to `/api/webchat/v2/threads/{thread_id}/runs/{run_id}/cancel`; the frontend uses `cancelRun` from `crates/ironclaw_webui_v2/frontend/src/lib/api.ts`. `crates/ironclaw_turns` immediately cancels queued or blocked work, moves running work to `CancelRequested` for cooperative shutdown, treats repeated cancellation of terminal work idempotently, and uses lease expiry as the terminal backstop.

Stopping work reliably is essential, and we would be glad to have you contribute Reborn cancellation coverage—particularly cancellation during capability execution, blocked gates, or reconnect races. Thank you for pushing this user-facing control forward.

PR #2129, opened 2026-04-07
fix(e2e): canonicalize extension names + fix remaining test failures

henrypark133:
## Summary

- **Bug fix**: `configure()` and `get_setup_schema()` in `ExtensionManager` used `validate_extension_name()` which discards the canonical form — hyphenated names from URL paths (e.g., `web-search`) caused "Capabilities file not found" errors. Now both canonicalize properly (`web-search` → `web_search`).
- **Bug fix**: `extensions_setup_handler` kind lookup compared raw URL path param against canonical stored names.
- **Test fix**: Update `test_wasm_lifecycle.py` extension names from `web-search` to `web_search` (canonical form).
- **Test fix**: Update `test_mcp_auth_flow.py` names from `mock-mcp` to `mock_mcp`.
- **Test fix**: `test_chat.py` now counts assistant messages before sending to avoid picking up the onboarding greeting.

## Test plan

- [x] E2E core suite — 16/16 passed locally
- [x] E2E extensions suite — 103/103 passed locally (test_wasm_lifecycle + test_mcp_auth_flow + test_extensions)
- [x] E2E features suite — already passing in CI
- [x] E2E routines suite — already passing in CI

### DIFF
--- a/src/channels/web/server.rs
+++ b/src/channels/web/server.rs
@@ -3016,11 +3016,13 @@ async fn extensions_setup_handler(
         .await
         .map_err(|e| (StatusCode::INTERNAL_SERVER_ERROR, e.to_string()))?;
 
+    let canonical_name = crate::extensions::naming::canonicalize_extension_name(&name)
+        .unwrap_or_else(|_| name.clone());
     let kind = ext_mgr
         .list(None, false, &user.user_id)
         .await
         .ok()
-        .and_then(|list| list.into_iter().find(|e| e.name == name))
+        .and_then(|list| list.into_iter().find(|e| e.name == canonical_name))
         .map(|e| e.kind.to_string())
         .unwrap_or_default();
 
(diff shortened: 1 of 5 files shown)


henrypark133:
Fixed in b34659bc — good catch. Now using `SEL["message_assistant"]` passed into the JS function instead of the hardcoded selector.

henrypark133:
Addressed both Copilot comments:

**test_chat.py:24** — Good point about streaming partials. Replaced the manual count+wait approach with `send_chat_and_wait_for_terminal_message` helper (from `helpers.py`) which already handles streaming correctly — it waits for the terminal message, not just element count. Fixed in f053c6dc.

**server.rs:3027** — Valid observation about the inefficient full-list scan and silent empty-string fallback, but this is pre-existing behavior — my change only fixed the name comparison (`e.name == name` → `e.name == canonical_name`). Refactoring the kind lookup into `get_setup_schema()` or a dedicated method is a good idea but out of scope for this bug fix PR.

nickpismenkov (approved):
lgtm

PR #2150, opened 2026-04-08
fix(cli): use -m will not quit

jokemanfire:
Now the guard will be take and custom, So we should not skip msg_tx clone.

## Summary

use -m will not quit .  Like “cargo run -- -m "hello"”
-

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] CI/Infrastructure
- [ ] Security
- [ ] Dependencies

## Validation

- [x] `cargo build`
- [ ] Relevant tests pass: 
- [ ] Manual testing: 

## Review Follow-Through

---

**Review track**:

### DIFF
--- a/src/channels/repl.rs
+++ b/src/channels/repl.rs
@@ -492,12 +492,9 @@ impl Channel for ReplChannel {
 
     async fn start(&self) -> Result<MessageStream, ChannelError> {
         let (tx, rx) = mpsc::channel(32);
-        // Store tx so send_status can inject approval responses directly.
-        // Skip for single-message mode — no interactive approval is needed
-        // and the extra sender would keep the stream open after /quit.
-        if self.single_message.is_none()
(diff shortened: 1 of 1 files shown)


serrrfirat (approved):
**Approved** — clean, correct fix.

The root cause is that `ChannelManager::start_all()` uses `select_all` to merge all channel streams plus the injection channel. The REPL stream closing in `-m` mode doesn't end the merged stream because other constituents stay alive. The fix correctly stores `msg_tx` for single-message mode so `finish_single_message_turn()` can inject `/quit`, triggering `HandleOutcome::Shutdown` in the agent loop regardless of other channels.

One minor note: if the agent panics before calling `respond()`, `finish_single_message_turn()` never runs and no `/quit` is injected — but the process already hangs in that scenario (injection channel keeps the merged stream alive), so this isn't a regression.

PR #2154, opened 2026-04-08
feat(admin): admin tool policy to disable tools for users

### LINKED ISSUE #2078: Admin can disallow tool creation skills for regular users for multi-tenant Ironclaw

serrrfirat:
## Summary
- Adds `AdminToolPolicy` type stored in the settings table under `__admin__` scope — no schema changes needed
- New admin-only API endpoints `GET/PUT /api/admin/tool-policy` gated behind multi-tenant mode
- Enforcement in `ChatDelegate::before_llm_call()` strips admin-disabled tools from the LLM context before per-user permissions are evaluated
- Admin users are exempt from the policy (cannot lock themselves out)

Closes #2078

## How it works
1. Admin calls `PUT /api/admin/tool-policy` with a JSON body specifying `disabled_tools` (global) and/or `user_disabled_tools` (per-user)
2. In the dispatcher's `before_llm_call()`, when multi-tenant mode is active and the user is not an admin, the policy is loaded from the DB and matching tools are filtered out
3. Since tools are removed from the LLM context entirely, users cannot call them — and the existing per-user permission system cannot re-enable them

## API

```
GET  /api/admin/tool-policy   → AdminToolPolicy JSON
PUT  /api/admin/tool-policy   ← AdminToolPolicy JSON
```

```json
{
  "disabled_tools": ["build_software", "tool_install", "tool_remove"],
  "user_disabled_tools": {
    "alice": ["shell"]
  }
}
```

## Multi-tenancy gating
- **API layer**: Returns 404 when `workspace_pool.is_none()` (single-user mode)
- **Dispatcher**: Skips admin policy check when `config.multi_tenant` is false

## Files changed
| File | Change |
|------|--------|
| `src/tools/permissions.rs` | `AdminToolPolicy` struct, constants, helpers, 6 unit tests |
| `src/channels/web/handlers/tool_policy.rs` | **New** — GET/PUT handlers with validation |
| `src/channels/web/handlers/mod.rs` | Module declaration |
| `src/channels/web/server.rs` | Route registration |
| `src/agent/dispatcher.rs` | Admin policy enforcement in `before_llm_call()` |
| `src/channels/web/tests/multi_tenant.rs` | 4 integration tests (auth, 404, CRUD, validation) |

## Test plan
- [x] `cargo clippy --all --all-features` — zero new warnings
- [x] `cargo test --lib` — all pass
- [x] `cargo check --no-default-features --features libsql` — compiles
- [x] Unit tests for `AdminToolPolicy` (global, per-user, combined, empty, serde)
- [x] API tests: member 403, single-user 404, CRUD with libSQL, validation 400


serrrfirat:
Addressed the remaining follow-up items in `dd0013de`.

- `GET /api/admin/tool-policy` now logs a warning when persisted policy JSON fails to parse instead of silently defaulting with no signal.
- The hot filtering path now precomputes global and per-user disabled-tool lookup sets before filtering tool defs, so the branch no longer does repeated linear scans for every tool during the filter pass.

Verified with:
- `cargo fmt --all`
- `cargo test tool_policy --lib`

ilblackdragon (commented):
## Overview
Adds an `AdminToolPolicy` (global + per-user disabled tool lists) stored in the `settings` table under a `__admin__` pseudo user, plus `GET/PUT /api/admin/tool-policy`. Enforcement strips disabled tools from `tool_defs` in `ChatDelegate::before_llm_call()` (and `JobDelegate`/`ContainerDelegate` for defence in depth) before per-user permissions evaluate.

## Strengths
- Clear separation: policy data (`permissions.rs`), HTTP layer (`tool_policy.rs`), enforcement (`filter_admin_disabled_tools`). The shared filter avoids duplication across the three delegates.
- Fail-closed: DB or deserialization errors return an empty tool list. Good default for a security-relevant feature.
- Admin exemption prevents lockout.
- 404 in single-user mode keeps the surface tight; 6 unit tests + 4 integration tests cover serde, gating, member 403, validation.

## Issues / suggestions

**Correctness — admin role detection drift**
`filter_admin_disabled_tools` takes `&UserRole`, but `JobDelegate::resolve_user_info` stringly-compares `user.role == "admin"` (`worker/job.rs:804`) and `ChatDelegate` uses `self.tenant.identity().role`. Three different sources of truth for "is admin." If the canonical role enum elsewhere ever diverges (e.g. case, "owner"), the filter silently misbehaves. Consider one helper like `UserRole::from_db_str` and use it everywhere.

**Settings-table key collision risk**
`ADMIN_SETTINGS_USER_ID = "__admin__"` is documented as collision-safe because real IDs are UUIDs — but that's an implicit invariant only enforced at user creation. A `CHECK` constraint, or a separate `admin_settings` row namespace, would be safer. At minimum add a debug assertion in `users_create_handler` rejecting reserved IDs.

**Container delegate fields are pure dead weight**
`ContainerDelegate` gets four new fields (`multi_tenant`, `user_role`, `user_id`, `db`) all wired to defaults that make `filter_admin_disabled_tools` a no-op (`worker/container.rs:705-708`). The "defence in depth for the future" rationale is exactly the kind of speculative complexity CLAUDE.md tells you to avoid. Either delete them, or wire them through the orchestrator job metadata for real. Right now they're misleading — they look like enforcement when they aren't.

**`PUT` is destructive replace, no concurrency control**
Two admins editing concurrently silently lose one update. Consider an `If-Match`/version field, or at least document the replace semantic in the route's doc comment + admin UI.

**Validation gaps**
- `user_disabled_tools` keys aren't checked against actual existing users — typos silently no-op forever.
- No upper bound on policy size (a 10k-entry policy gets serialized to JSON and parsed on every LLM call). Add `MAX_POLICY_ENTRIES`.
- Tool names aren't validated against `ToolRegistry` — no feedback loop if admin disables `buld_software`.

**Performance — DB read on every LLM iteration**
`filter_admin_disabled_tools` issues `db.get_setting(...)` once per `before_llm_call`, i.e. once per agent loop iteration. For a long multi-tool job that's a lot of redundant reads of an admin-rare-write value. A small TTL'd cache (or `Arc<RwLock<AdminToolPolicy>>` invalidated on `PUT`) would be cheap and meaningful.

**Test coverage gap**
No end-to-end test that verifies the filtered tool actually does not reach the LLM. Tests cover the policy struct and the HTTP layer but not the dispatcher integration. A test that asserts `reason_ctx.available_tools` no longer contains a disabled tool would close the loop.

serrrfirat:
Addressed, and rebased on latest `staging` (merge conflict resolved in `src/channels/web/server.rs` by keeping both `/api/admin/tool-policy` and `/api/admin/system-prompt` routes).

Responding to your latest review:

- **Admin role detection drift**: fixed by adding `UserRole::from_db_role(...)` and using it in `JobDelegate::resolve_user_info`.
- **`__admin__` collision risk**: added a guard in `users_create_handler` (`debug_assert_ne!` + runtime check against `ADMIN_SETTINGS_USER_ID`).
- **Container delegate dead fields**: removed the speculative/no-op admin-policy fields from `ContainerDelegate`.
- **PUT replace semantics**: documented explicitly in `tool_policy_put_handler` doc comment (full replace / last-write-wins).
- **Test coverage gap**: added `test_admin_policy_filter_happens_before_auto_approval_and_llm_call` in `src/agent/dispatcher.rs` to verify disabled tools are filtered before LLM-visible tool list and auto-approval state.

Also included in this branch from earlier follow-up:
- moved admin-policy filtering before per-user permission filtering in `ChatDelegate::before_llm_call`.

Focused checks run after the merge:
- `cargo test -q test_admin_policy_filter_happens_before_auto_approval_and_llm_call --lib`
- `cargo test -q test_tool_policy_crud_with_db --lib`

If you want, I can follow up with a separate PR for policy size bounds + optional policy caching (`Arc<RwLock<...>>` + invalidation on `PUT`) to keep this one scoped.

src/tenant.rs
```diff
@@ -521,6 +521,12 @@ impl SystemScope {
         Self { inner: db }
     }
 
+    /// Expose the underlying database handle for system-level operations that
+    /// need raw access (e.g. loading admin tool policy with a fixed user_id).
+    pub fn db(&self) -> &Arc<dyn Database> {
```
ilblackdragon: hm, this seems like violating the whole point of this isolation.
should this be just accessors here?

ilblackdragon:
# Code Review

## Overview
Adds an `AdminToolPolicy` (global + per-user disabled tool lists) stored as JSON under a reserved `__admin__` scope in the existing `settings` table. New `GET/PUT /api/admin/tool-policy` endpoints expose CRUD; both `ChatDelegate::before_llm_call` and `JobDelegate` strip admin-disabled tools from the LLM context before per-user permissions and session auto-approval are applied. Multi-tenant gated; admins are exempt. No schema migration. Solid test coverage including a great end-to-end test that asserts disabled tools never reach `complete_with_tools`.

## Strengths
- **Right enforcement point.** Removing tools from the LLM context (rather than blocking calls) is the correct architecture — users can't re-enable via per-user permissions, and the integration test in `dispatcher.rs` proves auto-approval is also stripped.
- **Reuses settings storage** — no migration, both backends supported automatically.
- **Module-owned filter helper.** `filter_admin_disabled_tools` lives in `tools/permissions.rs` and is shared between `ChatDelegate` and `JobDelegate`. Good factoring.
- **Reserved-id collision defense** with `debug_assert_ne!` plus a runtime check in `users_create_handler`.
- **Tests are thorough**: unit tests for the policy struct, handler tests for auth/404/CRUD/validation, and the dispatcher-level test that asserts the LLM never sees disabled tools.

## Issues & Suggestions

### Correctness / Security

- **Fail-closed empties the entire tool set on a transient DB hiccup.** `filter_admin_disabled_tools` returns `Vec::new()` on `get_setting` errors. A flaky DB will silently brick non-admin agents with no observable failure surface beyond a `warn!`. Consider propagating `Result` and letting the caller decide (or at minimum emit a metric/error event so this is visible). The deserialization-failure case is even worse: a single corrupt row disables all tools for everyone forever until someone notices.

- **Per-iteration DB read inside the agentic loop.** `before_llm_call` runs every loop iteration; this issues a fresh `get_setting` call each time. For long agentic loops this adds N DB roundtrips per request that read identical bytes. `JobDelegate` caches `(user_id, role)` via `OnceCell` but not the policy itself. Cache the policy per-request (or per-session) — either via a `OnceCell` mirror of `cached_user_info`, or by loading once at session start.

- **OIDC auto-provisioning not patched.** The doc comment on `ADMIN_SETTINGS_USER_ID` explicitly cites OIDC auto-provisioning as a UUID source, but only `users_create_handler` got the runtime guard. Verify the OIDC path rejects (or simply cannot generate) `__admin__` — and ideally add the same `debug_assert_ne!` there.

- **PUT validation gaps.** Length bounds exist but there's no cap on list length or `user_disabled_tools` map size, no dedup, no charset filter. The 10MB body cap is the only ceiling — a single PUT can add tens of thousands of entries that get scanned linearly on every LLM iteration. Add a max-entries check.

- **`is_tool_disabled` is O(n·m) per call.** It iterates `disabled_tools` linearly per tool. The dispatcher path builds a `HashSet` first (good), but `is_tool_disabled` itself (used in tests/handlers) doesn't. Consider a `HashSet`-backed accessor or document that the canonical filter path is `filter_admin_disabled_tools`.

### API / Semantics

- **PUT is full-replace, no concurrency control.** Documented, but two admins editing concurrently will silently clobber each other. A simple `If-Match`/etag or version field would prevent this.
- **404 for single-user mode** is reasonable but slightly surprising for an admin who's intentionally probing. A 409/400 with a clearer message ("not available outside multi-tenant") would be friendlier; minor.
- **No DELETE endpoint.** Admins must PUT an empty policy to clear. Acceptable but worth noting.

### Style / Conventions

- **`SystemScope::db()` exposes the raw `Arc<dyn Database>`**. The doc-comment acknowledges it's an encapsulation break. It's only used here; consider instead adding a typed `load_admin_tool_policy()` method on `SystemScope` so the raw handle stays sealed.
- `filter_admin_disabled_tools`'s `db: Option<&Arc<dyn Database>>` is awkward — `Option<&Arc<…>>` is rarely the right shape. Prefer `Option<&dyn Database>` or just require the caller to early-return.
- Two `tracing::warn!` calls inside the filter will fire on every iteration in a degraded state — likely log spam. Use `debug!` or rate-limit.

### Test Coverage
- Strong on the happy path. Missing: behavior when DB returns an error (the fail-closed branch), behavior when a stale session has an admin-disabled tool already in `auto_approved` (does the filter actually re-evict it across multiple iterations, not just the first one?), and JobDelegate equivalent of the dispatcher integration test.

## Risk Summary
| Risk | Severity |
|---|---|
| Fail-closed empties tools on transient DB error | **Medium** — operational footgun |
| Per-iteration DB read in hot loop | **Low-Med** — perf, not correctness |
| No size cap on policy lists | **Low** — DoS-ish |
| OIDC reserved-id guard missing | **Low** — defense-in-depth gap |
| Concurrent admin edits clobber | **Low** |

Overall this is a clean, well-tested implementation of a needed feature. The main thing I'd ask for before merge is **caching the policy per-request** and **rethinking the fail-closed-to-empty behavior** (or at minimum surfacing it as an error event, not just a warn log).

ilblackdragon:
# Code Review

## Overview
Adds an `AdminToolPolicy` (global + per-user disabled tool lists) stored in the `settings` table under a reserved `__admin__` scope, exposed via `GET/PUT /api/admin/tool-policy`. Enforcement happens in both `ChatDelegate::before_llm_call` and `JobDelegate` by stripping admin-disabled tools from the LLM tool list before per-user permission filtering. Admin users and single-tenant deployments bypass enforcement.

## Strengths
- Filtering happens **before** the LLM ever sees the tool definitions, so users can't re-enable disabled tools through their per-user permissions — correct threat model.
- Both code paths (chat dispatcher + job worker) are hooked, with a regression test asserting `RecordingToolsProvider` never sees admin-disabled tools.
- Fail-closed on DB/parse errors (`return Vec::new()`) preserves admin restrictions during transient failures.
- Good test coverage: 6 unit tests on `AdminToolPolicy`, 4 integration tests (member 403, single-user 404, CRUD, validation), plus the end-to-end dispatcher test.
- Defensive: `users_create_handler` checks UUID never collides with `__admin__`, both via `debug_assert_ne!` and a runtime guard.

## Issues & Suggestions

### 1. Violates "Everything Goes Through Tools" rule (CLAUDE.md)
`tool_policy_get_handler` / `tool_policy_put_handler` call `state.store.get_setting` / `set_setting` directly. Per the project's core principle, all gateway handler mutations must go through `ToolDispatcher::dispatch()` for the audit trail (`ActionRecord`), safety pipeline, and channel parity. This will likely trip `scripts/pre-commit-safety.sh`. Either:
- Add an `admin_tool_policy_get` / `admin_tool_policy_set` tool and dispatch through it, **or**
- Annotate with `// dispatch-exempt: <reason>` if this is intentionally an admin-only out-of-band operation.

### 2. No caching in `ChatDelegate` path (perf)
`JobDelegate` caches `(user_id, role)` in `OnceCell` but `ChatDelegate::before_llm_call` issues a fresh `db.get_setting(__admin__, admin_tool_policy)` on **every LLM iteration**. For multi-turn tool loops this is N extra DB roundtrips per request. Consider caching the parsed `AdminToolPolicy` on the delegate / agent with a short TTL, or once per `before_llm_call` invocation lifetime.

### 3. `SystemScope::db()` leaks abstraction
Adding `pub fn db(&self) -> &Arc<dyn Database>` to `SystemScope` defeats the encapsulation that tenant scoping is supposed to provide. Prefer adding a typed `SystemScope::admin_tool_policy()` method that owns the read.

### 4. Validation gaps in PUT handler
- Only length is checked. Empty/oversize is rejected, but `\n`, `\0`, control chars, or whitespace-only names pass.
- No cap on `disabled_tools.len()` or `user_disabled_tools.len()` — an admin (or compromised admin token) could PUT a multi-MB policy that gets loaded on every LLM call. Add a sanity cap (e.g. 1000 entries each).
- `PUT` is full-replace; concurrent admin PUTs lose updates silently. Acceptable but worth noting in the doc comment.

### 5. Inconsistent error handling
- `tool_policy_get_handler` uses `unwrap_or_else(|_| AdminToolPolicy::default())` on parse failure (fail-open: shows admin "no policy").
- `filter_admin_disabled_tools` returns `Vec::new()` (fail-closed: blocks all tools).

These should be consistent in spirit. The GET handler returning `default()` could mislead an admin into thinking the policy is empty when it's actually corrupt — at minimum, return a 500 with the parse error so the admin knows.

### 6. Minor
- `is_tool_disabled` does linear `Vec` scans. Fine for small lists, but if you build a `HashSet` once in `filter_admin_disabled_tools`, do the same in the unit-tested helper for consistency.
- The dispatcher test (`test_admin_policy_filter_happens_before_auto_approval_and_llm_call`) is excellent and exactly the kind of "test through the caller" coverage CLAUDE.md asks for.
- `debug_assert_ne!` followed by a runtime `if` is slightly redundant.

## Risk Summary
- **Security**: Solid — fail-closed, filter-before-LLM, admin exempt, multi-tenant gated. Main risk is the **dispatch-exempt** rule violation.
- **Performance**: Per-LLM-call DB lookup in chat path is the only concrete concern; cache it.
- **Correctness**: Good test coverage on the happy path and the most important regression.

## Recommendation
Request changes for (1) — the tool dispatch rule is a hard project convention. (2) and (3) are worth fixing before merge; the rest are nice-to-haves.

ilblackdragon (dismissed):

ilblackdragon:
# Re-review (round 2)

Picked up the two new commits (`12fcbbc9`, `79fb23dc`) that address the round-1 feedback. Most of the issues are now resolved cleanly. Verified locally: `cargo clippy --all --benches --tests --examples --all-features` is clean and `cargo test --lib admin` runs all 33 admin-related tests green, including the dispatcher regression test.

## What's now fixed
- **Per-iteration DB lookup (#2)** — `AdminToolPolicyCache` (`OnceCell<AdminToolPolicyState>`) is wired into both `ChatDelegate` and `JobDelegate`. Loaded once per loop, not per iteration. Clean.
- **`SystemScope::db()` leak (#3)** — replaced with three purpose-built methods: `get_admin_tool_policy`, `set_admin_tool_policy`, `get_user_role`. Tenant boundary preserved. Even better than what I'd asked for.
- **Validation gaps (#4)** — `validate_admin_tool_policy` is now shared between the HTTP handler and `SystemScope::set_admin_tool_policy`. Adds entry-count caps (1000 each), 32 KB serialized size cap, plus a charset-based `is_valid_admin_tool_name` / `is_valid_admin_policy_user_key` (lowercase ASCII alphanumerics + `_-`) which is stricter than my "no control chars" suggestion and a better choice. New regression tests cover path-like names and bad user keys.
- **Inconsistent error handling (#5)** — GET handler now returns 500 with `parse_admin_tool_policy` on corrupt stored state instead of silently defaulting to empty. Matches the fail-closed posture of the filter path.
- **Performance bonus** — `disabled_tools: Vec<String>` → `HashSet<String>` for O(1) `contains` lookups. Nice catch.
- **Code clarity bonus** — `UserRole::is_admin()` helper consolidates the role check across `dispatcher.rs`, `worker/job.rs`, and `engine.rs` (no more `*role == UserRole::Admin` scattered around).

## Remaining concerns

### 1. The `// dispatch-exempt:` annotation is at module-level, not per-line — pre-commit hook will trip
The handler now has a `//!` module doc comment explaining the exemption, but `scripts/pre-commit-safety.sh` looks for a per-line `// dispatch-exempt:` trailing comment on each `state.{store,workspace_pool,...}` access. Running the diff manually:

```
$ git diff staging -- 'src/channels/web/handlers/*.rs' | grep -nE '^\+' \
    | grep -E 'state\.(store|workspace|workspace_pool|...)\.' \
    | grep -vE '// dispatch-exempt:|// safety:|^\+\+\+'
+    if state.workspace_pool.is_none() {        # tool_policy.rs GET
+    let store = state.store.as_ref().ok_or((    # tool_policy.rs GET
+    if state.workspace_pool.is_none() {        # tool_policy.rs PUT
+    let store = state.store.as_ref().ok_or((    # tool_policy.rs PUT
```

The hook check at `scripts/pre-commit-safety.sh:293-298` is line-by-line and will flag these four when CI diffs against the merge target. The module-level doc comment is great for human readers but doesn't satisfy the regex. Two options:
- Add a trailing `// dispatch-exempt: <reason>` comment on each of the four lines, or
- Refactor to bind `state.workspace_pool` / `state.store` into local variables on dedicated lines that carry the trailing comment (this also survives `cargo fmt`).

(Note: it didn't trip when you committed because `git diff @{upstream}` against your own branch HEAD is empty. CI will use a different base.)

### 2. `JobDelegate::cached_admin_tool_policy` initialization is inlined inside `before_llm_call`
The closure that loads the policy lives inside `before_llm_call` (`worker/job.rs:1429-1450`), while `ChatDelegate` cleanly delegates to `permissions::load_cached_admin_tool_policy(...)`. Why two different shapes? Refactoring `JobDelegate` to call the same shared helper would:
- Remove ~25 lines of duplicated `match store.get_admin_tool_policy()` logic
- Make the two delegates symmetric and easier to keep in sync if the load logic ever changes
- Eliminate the dependency on `worker.store()` being convertible to the `&Arc<dyn Database>` shape that `load_cached_admin_tool_policy` expects (currently it can't, hence the duplication — but adding a thin adapter on `SystemScope` would unblock it)

Not a blocker, but worth a follow-up.

### 3. Charset rule is undocumented and will surprise admins
`is_valid_admin_tool_name` requires `[a-z0-9_-]` and `is_valid_admin_policy_user_key` requires `[A-Za-z0-9_-]`. Both are reasonable, but:
- The error message just says `"Invalid tool name: '{name}'"` — admins won't know what charset is allowed.
- Real built-in tool names follow the lowercase rule, but **user IDs in this codebase are UUIDs** (e.g. `550e8400-e29b-41d4-a716-446655440000`) — those will pass. Good. But OIDC `sub` claims sometimes contain `:` or `|` (e.g. `auth0|abc123`, `google-oauth2|12345`). If an OIDC-provisioned user_id ever lands in `user_disabled_tools`, the validator will reject it.

Suggestion: include the allowed charset in the error message, and double-check that the OIDC auto-provisioning path always produces UUIDs (it does in `users_create_handler`, but the comment in `permissions.rs` line 100-101 is the only assertion of this — worth a defensive test).

### 4. Minor: `parse_admin_tool_policy` source string is `&'static str`
Forces all callers to use string literals (`"http_get"`, `"system_scope"`, `"database"`, `"test"`). That's fine, but it's a slightly odd interface — `&str` would work just as well and remove the constraint. Trivial.

## Verdict
Round-2 fixes are substantial and well-executed. **One actual blocker** (the dispatch-exempt annotation, since it'll break CI) and one cleanup suggestion (consolidate the JobDelegate cache load with the shared helper). Everything else is polish.

Approve once #1 is fixed.

ilblackdragon:
# Re-review (round 3)

Picked up `3d4f8263` which addresses the round-2 dispatch-exempt blocker. Almost there — but the fix only landed on **2 of the 4** offending lines.

## Status

| Round-2 issue | Status |
|---|---|
| #1 dispatch-exempt for pre-commit hook | ⚠️ partial — see below |
| #2 unify `JobDelegate` cache load with shared helper | ⏳ not addressed (non-blocker) |
| #3 charset rule visibility / OIDC edge case | ⏳ not addressed (non-blocker) |
| #4 `&'static str` source param | ⏳ not addressed (trivial) |

Verified: `cargo test --lib admin` → **34 passed**, including the dispatcher and e2e regressions.

## ⚠️ Blocker still partially open: dispatch-exempt

`3d4f8263` annotated the two `state.store.as_ref()` lines (one in GET, one in PUT) but **left the two `state.workspace_pool.is_none()` lines uncovered**. Running the same line-by-line check the pre-commit hook uses against the merge target:

```
$ git diff staging -- 'src/channels/web/handlers/*.rs' | grep -nE '^\+' \
    | grep -E 'state\.(store|workspace|workspace_pool|...)\.' \
    | grep -vE '// dispatch-exempt:|// safety:|^\+\+\+'
+    if state.workspace_pool.is_none() {        # tool_policy.rs:30 (GET)
+    if state.workspace_pool.is_none() {        # tool_policy.rs:74 (PUT)
```

The hook regex at `scripts/pre-commit-safety.sh:294` matches `state.workspace_pool.` just as much as `state.store.`. CI will trip on these two lines.

**Fix**: add the same trailing comment treatment to the two `workspace_pool` checks. Either:

```rust
if state.workspace_pool.is_none() { // dispatch-exempt: gateway-mode probe
    return Err((
```

…or, if rustfmt insists on splitting the line, lift the access to a local on a dedicated line:

```rust
let pool = state.workspace_pool.as_ref(); // dispatch-exempt: gateway-mode probe
if pool.is_none() {
```

(The second form is what I'd recommend — it survives `cargo fmt` deterministically.)

Why it didn't trip locally for `serrrfirat`: `git diff @{upstream}` against your own branch HEAD is empty, so `resolve_base_ref()` finds nothing to scan. CI uses a different base.

## Carry-overs (non-blocking, repeat from round 2 for the record)

- **Symmetry between delegates**: `JobDelegate` (`worker/job.rs:1429-1450`) still inlines the policy load closure, while `ChatDelegate` cleanly delegates to `permissions::load_cached_admin_tool_policy(...)`. Refactoring `JobDelegate` to use the same helper would remove ~25 lines of duplicated `match store.get_admin_tool_policy()` plumbing and keep both paths in sync.
- **Charset error message** doesn't mention the allowed character set; admins will hit `"Invalid tool name: 'foo.bar'"` and have no clue why.
- **OIDC `sub` edge case** with `is_valid_admin_policy_user_key` rejecting `:` and `|` — confirm the auto-provisioning path always rewrites to UUIDs (it does in `users_create_handler`, but a defensive test would harden this).
- **`parse_admin_tool_policy(value, source: &'static str)`** could be `&str` without losing anything.

## Verdict

Two-line fix away from approval. Once the `workspace_pool` lines have trailing dispatch-exempt comments (or are lifted into bound locals), this is mergeable. The `JobDelegate` symmetry cleanup and the charset/OIDC notes can ride a follow-up PR.

ilblackdragon:
# Final review (round 4)

Branch HEAD `78f0a365` after the dispatch-exempt cleanup. This is the post-fix sanity pass — taking the whole PR end-to-end now that all blockers are resolved.

## CI status
All checks green:
- ✅ Formatting
- ✅ Clippy: default, all-features, libsql-only
- ✅ cargo-deny
- ✅ No panics in production code
- ✅ Code Style aggregate gate
- ✅ Regression test enforcement
- ✅ scope / classify

Local: `cargo test --lib admin` → 34 passed (including the dispatcher regression and the e2e file).

## Overview (consolidated for the final read-through)

| Surface | Behavior |
|---|---|
| `AdminToolPolicy` | `disabled_tools: HashSet<String>` (global) + `user_disabled_tools: HashMap<String, Vec<String>>` (per user). Stored as a single JSON value in `settings` under the reserved `__admin__` scope. |
| `GET /api/admin/tool-policy` | Admin-only. Returns 404 in single-tenant, 500 on corrupt persisted state, otherwise the policy (or default). |
| `PUT /api/admin/tool-policy` | Admin-only. Full replace. Validates: charset (`[a-z0-9_-]` for tool names, `[A-Za-z0-9_-]` for user keys), length caps (128/256), entry caps (1000), 32 KB serialized payload cap. Last-write-wins. |
| Enforcement (`ChatDelegate`) | Loads policy via shared `load_cached_admin_tool_policy(...)` into a `OnceCell<AdminToolPolicyState>` per turn; filters tools **before** per-user permissions and session auto-approval. |
| Enforcement (`JobDelegate`) | Same `OnceCell` cache, loads via `SystemScope::get_admin_tool_policy()`. Looks up role via `SystemScope::get_user_role()`. |
| Fail-closed | DB error or parse failure → `AdminToolPolicyState::FailClosed` → empty tool list. Admin users and single-tenant deployments bypass enforcement entirely. |
| Tenant boundary | `SystemScope` exposes only `get_admin_tool_policy`, `set_admin_tool_policy`, `get_user_role` — no raw DB handle leak. |

## Test coverage
- 9 unit tests on `AdminToolPolicy` / validators (`tools/permissions.rs`)
- 4 unit tests on `validate_identifier` / oversize / charset rejection (`channels/web/handlers/tool_policy.rs`)
- 5 integration tests on the HTTP handler (`channels/web/tests/multi_tenant.rs`): member 403, single-user 404, full CRUD with libSQL, validation 400, oversize 400
- Dispatcher regression test (`agent/dispatcher.rs`) asserting filter happens before auto-approval **and** the disabled tool never reaches `RecordingToolsProvider.complete_with_tools`
- Standalone e2e (`tests/admin_tool_policy_e2e.rs`) driving a full chat turn through `TestRig` and asserting the disabled tool is absent from the LLM tool list

This is exactly the "test through the caller, not just the helper" coverage `.claude/rules/testing.md` calls for. Both the chat path and the worker path are exercised end-to-end.

## Code quality
- ✅ No `.unwrap()` / `.expect()` outside tests
- ✅ Errors mapped through `thiserror` / `DatabaseError::Serialization`
- ✅ Strong types (`AdminToolPolicyState` enum, `UserRole::is_admin()` consolidator)
- ✅ Validation extracted into shared `validate_admin_tool_policy()` reused by both HTTP and `SystemScope::set_admin_tool_policy`
- ✅ `HashSet` lookup for O(1) `disabled_tools.contains()`
- ✅ `OnceCell` caching is per-loop, not process-global — no stale-policy hazards across requests
- ✅ Module-level `//! dispatch-exempt:` doc comment + per-line trailing annotations satisfy both human readers and the pre-commit hook
- ✅ `cargo fmt` deterministic (the bound-local pattern survives reformatting)

## Security
- **Filter-before-LLM** is the right place: users cannot re-enable disabled tools via `tool_permissions` because the tool is gone from the LLM context entirely.
- **Fail-closed on errors** preserves admin restrictions during transient failures.
- **Multi-tenant gated**: enforcement is a no-op in single-user deployments, so single-user installs don't accidentally inherit empty restrictions.
- **Admin exempt**: admins cannot lock themselves out — important escape hatch.
- **Charset validator** prevents path-like tool names (e.g. `../shell`) and exotic user keys.
- **Size cap** (32 KB serialized + 1000 entry caps) bounds the per-loop DB roundtrip cost and prevents an admin from persisting a multi-MB blob.
- **Reserved `__admin__` scope** is collision-safe because all real user IDs are UUIDs (defended by a runtime guard in `users_create_handler` plus the charset validator on the user_id key).

## Performance
- Per-loop `OnceCell` cache means at most **one** DB roundtrip per chat turn / job, regardless of how many tool iterations the loop runs.
- `HashSet<String>` for `disabled_tools` gives O(1) per-tool lookup; the per-user list still does linear scan but caps at 1000 entries.
- No new background work, no scheduled tasks, no allocation hot paths.

## Carry-over follow-ups (non-blocking)
None of these block merge — they're all candidates for a follow-up PR:

1. **`JobDelegate` cache-load symmetry**: `worker/job.rs:1429-1450` still inlines the load closure, while `ChatDelegate` cleanly uses `permissions::load_cached_admin_tool_policy(...)`. Refactoring to a single shared helper would remove ~25 lines of duplication and keep both paths in lockstep when the load logic evolves.
2. **Charset error message**: `"Invalid tool name: 'foo.bar'"` doesn't tell the admin which characters are allowed. Including the regex / charset in the error would save a documentation lookup.
3. **OIDC `:` / `\|` edge case**: confirm the auto-provisioning path always rewrites the `sub` claim to a UUID before storing it as a `user_id` (it does in `users_create_handler` today, but a defensive test would harden against drift).
4. **`parse_admin_tool_policy(value, source: &'static str)`**: `&str` would work just as well and remove the literal-only constraint. Trivial.

## Verdict

**Approve.** Three rounds of review feedback have been substantively addressed:

| Round 1 issue | Resolution |
|---|---|
| Dispatch-exempt rule violation | ✅ per-line annotations on all four `state.*` accesses |
| Per-iteration DB lookup | ✅ `OnceCell<AdminToolPolicyState>` cache on both delegates |
| `SystemScope::db()` abstraction leak | ✅ replaced with three typed methods |
| Validation gaps | ✅ shared `validate_admin_tool_policy` with charset / length / entry-count / size caps |
| Inconsistent error handling | ✅ GET returns 500 on corrupt state |

Plus three improvements I didn't ask for: `UserRole::is_admin()` consolidation, `Vec → HashSet` for O(1) lookups, and a standalone e2e test file. CI is green. Ship it.

ilblackdragon (approved):

PR #2159, opened 2026-04-08
fix(gateway): merge webhook routes into gateway for single-port deployments

henrypark133:
## Summary

- Webhook routes (`/webhook/slack`, `/webhook/telegram`, etc.) were only served by the separate `WebhookServer` on port 8080, making them unreachable on PaaS deployments like Railway that only expose port 3000 (the gateway)
- Merge webhook routes into the gateway server so they are served on **both** ports — full backwards compatibility for local/tunnel setups, and Railway just works
- Webhook routes use their own channel-specific auth (HMAC, Ed25519, secret headers) and bypass gateway bearer-token auth

## Test plan

- [x] `cargo test` — all tests pass (0 failures)
- [ ] After merge to staging and Railway deploy, retry Slack Event Subscription URL verification
- [ ] Verify port 8080 webhook server still works for tunnel-based setups


src/channels/web/server.rs
```diff
@@ -857,6 +858,16 @@ pub async fn start_server(
...
         .with_state(state.clone());
 
+    // Merge webhook routes (WASM channels, HTTP channel, etc.) so they are
+    // reachable on the gateway port.  These routes already have their own
+    // state applied and use channel-specific auth (HMAC, Ed25519, secret
+    // headers), bypassing gateway bearer-token auth.  The separate
+    // WebhookServer still runs for tunnel-based setups.
+    let mut app = app;
+    for route in additional_routes {
+        app = app.merge(route);
```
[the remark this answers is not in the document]
henrypark133: Fixed in 371ff84e. Added `create_wasm_webhook_routes()` that excludes `/oauth/callback` and `/wasm-channels/health`. The gateway now receives only the collision-free `/webhook/{*path}` routes via a separate `gateway_webhook_routes` vec, while the webhook server keeps the full router.

src/channels/web/server.rs
```diff
@@ -469,6 +469,7 @@ pub async fn start_server(
     addr: SocketAddr,
     state: Arc<GatewayState>,
     auth: CombinedAuthState,
+    additional_routes: Vec<Router>,
 ) -> Result<SocketAddr, crate::error::ChannelError> {
```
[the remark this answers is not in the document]
henrypark133: Added `test_additional_routes_reachable_and_no_collision` in 371ff84e. It starts the gateway with an extra `/webhook/test-channel` route and verifies: (1) the additional endpoint returns 200, (2) `/api/health` still works, (3) `/oauth/callback` is not 404 (still resolves to the gateway handler).

src/channels/web/mod.rs
```diff
@@ -153,9 +156,19 @@ impl GatewayChannel {
...
+            webhook_routes: Vec::new(),
         }
     }
 
+    /// Add webhook routes that will be merged into the gateway server.
+    ///
+    /// These routes use their own channel-specific auth (HMAC, Ed25519,
+    /// secret headers) and bypass the gateway's bearer-token auth.
+    pub fn with_webhook_routes(mut self, routes: Vec<axum::Router>) -> Self {
+        self.webhook_routes = routes;
```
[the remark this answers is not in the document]
henrypark133: Good catch — changed to `self.webhook_routes.extend(routes)` in 371ff84e.

nickpismenkov (dismissed):

src/channels/web/server.rs
serrrfirat: **Medium Severity — Merged webhook routes bypass gateway middleware layers**

The merged webhook routes skip two important gateway protections:

1. **`CatchPanicLayer`**: If a WASM webhook handler panics, the panic propagates uncaught and can crash the entire gateway's axum task — taking down SSE, WebSocket, and all API endpoints. On the standalone `WebhookServer` this only kills that server; on the gateway it kills everything.

2. **`DefaultBodyLimit::max(10 MB)`**: The WASM channel webhook router sets no body limit. On the gateway port (potentially public-facing on Railway), an attacker could send unbounded-size POST requests to `/webhook/{channel}`, exhausting memory.

**Suggestion:** Either restructure so `CatchPanicLayer` and `DefaultBodyLimit` are applied after the merge, or wrap each merged route:
```rust
let mut app = app;
for route in additional_routes {
    app = app.merge(
        route
            .layer(DefaultBodyLimit::max(10 * 1024 * 1024))
            .layer(CatchPanicLayer::new(/* ... */))
    );
}
```

src/channels/wasm/router.rs
```diff
@@ -648,6 +645,33 @@ pub fn create_wasm_channel_router(
         .with_state(state)
 }
 
+/// Webhook-only variant for merging into the gateway server.
+///
```
serrrfirat, line 649: **🟠 High Severity — Route Collision (pre-existing, not a regression)**

**Catch-all `POST /webhook/{*path}` could shadow `POST /webhook/tools/{tool}`** when both are merged into the gateway.

This is the same merge pattern already used in the existing `WebhookServer` on port 8080 — this PR replicates it on the gateway port. Not introduced by this PR.

**Suggestion:** Verify in integration testing that `POST /webhook/tools/<tool>` works correctly when both route sets are merged. Consider using `nest` instead of `merge` for namespaced prefixes.

src/main.rs
```diff
@@ -439,13 +439,18 @@ async fn async_main() -> anyhow::Result<()> {
 
     // Collect webhook route fragments; a single WebhookServer hosts them all.
     let mut webhook_routes: Vec<axum::Router> = Vec::new();
+    // Gateway-safe subset: excludes routes that collide with the gateway's own
+    // routes (e.g. /oauth/callback). Used when merging webhooks into the gateway.
+    let mut gateway_webhook_routes: Vec<axum::Router> = Vec::new();
 
-    webhook_routes.push(webhooks::routes(ToolWebhookState {
```
serrrfirat, line 562: **🟡 Medium Severity — Unnecessary Clone**

**`gateway_webhook_routes.clone()` is unnecessary — the vector is only used once.**

Since the routes are passed to `with_webhook_routes` and then into `start_server`, the clone creates a dangling copy.

**Fix:** Replace with `std::mem::take(&mut gateway_webhook_routes)` or just move the vector.

src/channels/wasm/setup.rs
```diff
@@ -9,7 +9,7 @@ use std::sync::Arc;
 use crate::channels::wasm::{
     LoadedChannel, RegisteredEndpoint, SharedWasmChannel, TELEGRAM_CHANNEL_NAME, WasmChannel,
     WasmChannelLoader, WasmChannelRouter, WasmChannelRuntime, WasmChannelRuntimeConfig,
-    bot_username_setting_key, create_wasm_channel_router,
+    bot_username_setting_key, create_wasm_channel_router, create_wasm_webhook_routes,
 };
 use crate::config::Config;
```
serrrfirat, line 23: **🟡 Medium Severity — API Design**

**`gateway_webhook_routes` is always `Some(...)` — the `Option` wrapper is never `None`.**

The `Option` adds unnecessary unwrapping ceremony at the call site.

**Fix:** Consider making it a non-optional `axum::Router`, or document why it's `Option` for future-proofing.

src/channels/web/server.rs
```diff
@@ -857,6 +858,16 @@ pub async fn start_server(
         ))
         .with_state(state.clone());
 
+    // Merge webhook routes (WASM channels, HTTP channel, etc.) so they are
```
serrrfirat, line 904: **🟡 Medium Severity — Robustness**

**If `additional_routes` contains a route colliding with an existing gateway route (e.g., `/api/health`), `app.merge()` will panic at runtime, taking down the server.**

No validation of the incoming route set.

**Fix:** Use `nest("/webhook", ...)` to namespace all webhook routes, preventing collision with gateway-owned paths. Or wrap the merge in a catch.

serrrfirat (approved):
## Paranoid Architect Review — APPROVE (with comments)

**1 High (pre-existing), 3 Medium findings.**

Sound approach. Creating a gateway-safe subset of webhook routes that excludes colliding paths is the right strategy for single-port deployments. Backward compatibility with dual-port setup is preserved.

The High finding (catch-all route shadowing) is pre-existing, not introduced by this PR. Medium findings are robustness improvements for follow-up.

serrrfirat (changes requested):
Lets at least fix high

src/channels/web/server.rs
henrypark133: Good catch. Fixed in f938128b — each additional route now gets `DefaultBodyLimit::max(10MB)` and `CatchPanicLayer::new()` applied before merging. A panic in a webhook handler will return 500 to that request instead of crashing the gateway's axum task.

src/channels/wasm/router.rs
henrypark133: Correct that this is pre-existing — the same merge pattern exists on the standalone WebhookServer (port 8080). axum resolves more specific routes (`/webhook/tools/{tool}`) before the catch-all (`/webhook/{*path}`), so the tool endpoints work correctly. No change needed.

src/main.rs
henrypark133: Fixed in f938128b — replaced `gateway_webhook_routes.clone()` with a direct move since the vec is only consumed once.

src/channels/wasm/setup.rs
henrypark133: Keeping `Option` for now — it mirrors the existing `webhook_routes` field and keeps `WasmChannelSetup` consistent. Both are always `Some` today but `Option` is idiomatic for setup results that could partially fail.

src/channels/web/server.rs
henrypark133: Already mitigated in the previous commit (371ff84e) — we split the WASM router into `create_wasm_channel_router` (full, for webhook server) and `create_wasm_webhook_routes` (webhook-only, for gateway). The gateway only receives `/webhook/{*path}` routes, so no collision with `/api/health`, `/oauth/callback`, etc.

src/channels/web/server.rs
henrypark133: Updated in 8394e5bc — extracted `catch_panic_layer()` helper that returns the same custom `CatchPanicLayer::custom(...)` with truncated panic payloads. Both the gateway's own routes and merged webhook routes now share the identical handler.

src/channels/wasm/router.rs
henrypark133: Fixed the doc comments in 8394e5bc — they now correctly state `/wasm-channels/health` is excluded as an unnecessary auxiliary endpoint (the gateway has `/api/health`), not because of a collision.

src/channels/wasm/setup.rs
henrypark133: Same fix as the other comment — updated doc in 8394e5bc to say 'auxiliary diagnostic endpoint not needed on the gateway' instead of implying a collision.

src/channels/web/server.rs
henrypark133: Good catch — fixed in 8394e5bc. HTTP channel routes are now also pushed to `gateway_webhook_routes` so they're reachable on port 3000 in single-port deployments.

src/channels/web/server.rs
```diff
@@ -5689,4 +5711,56 @@ mod tests {
...
+
+        // Create a simple additional router simulating a webhook endpoint.
+        let extra = axum::Router::new().route(
+            "/webhook/test-channel",
+            post(|| async { axum::http::StatusCode::OK }),
+        );
+
+        let bound = start_server(addr, state, auth, vec![extra])
+            .await
+            .expect("server should start with additional routes");
```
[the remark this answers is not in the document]
henrypark133: Matches the existing pattern — `test_csp_header_present_on_responses` in the same file also doesn't shut down. Both use port 0 so there's no port conflict. The tokio runtime drops the listener when the test completes. Low risk, consistent with existing tests.

PR #2165, opened 2026-04-08
feat(workspace): AGENTS_SEED_PATH env var for custom system prompt

pranavraja99:
## Summary
Deployments can override the default AGENTS.md seed by setting `AGENTS_SEED_PATH` to a custom file path. New users automatically get the custom system prompt at workspace creation. Falls back to the built-in seed when not set.

## Test plan
- Set `AGENTS_SEED_PATH=/path/to/custom/AGENTS.md` and create a new user — verify they get the custom prompt
- Unset the env var — verify default seed is used

### DIFF
--- a/src/workspace/mod.rs
+++ b/src/workspace/mod.rs
@@ -2144,12 +2144,20 @@ impl Workspace {
     /// so user edits are never overwritten. Returns the number of files
     /// created (0 if all core files already existed).
     pub async fn seed_if_empty(&self) -> Result<usize, WorkspaceError> {
+        // Allow deployments to override the default AGENTS.md seed via env var.
+        let custom_agents = std::env::var("AGENTS_SEED_PATH")
+            .ok()
+            .and_then(|p| std::fs::read_to_string(&p).ok());
+        let agents_seed = custom_agents
+            .as_deref()
+            .unwrap_or(include_str!("seeds/AGENTS.md"));
+
         let seed_files: &[(&str, &str)] = &[
             (paths::README, include_str!("seeds/README.md")),
             (paths::MEMORY, include_str!("seeds/MEMORY.md")),
             (paths::IDENTITY, include_str!("seeds/IDENTITY.md")),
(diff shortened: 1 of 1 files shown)


zmanian (changes requested):
## Review

Small PR (+9/-1) -- allows overriding the default AGENTS.md seed content via AGENTS_SEED_PATH env var. Good idea for deployment customization, but has two issues.

### Must Fix

1. **Silent failure on bad path** -- std::fs::read_to_string(&p).ok() swallows all errors. If an operator sets the env var to a wrong path, the system silently falls back to the default seed with no warning. At minimum tracing::warn! when the env var is set but the file can't be read.

2. **Blocking std::fs::read_to_string in an async function** -- blocks the tokio runtime thread. Use tokio::fs::read_to_string instead.

### Suggestions

3. **Use the config system** -- env var should be declared in src/config/ and passed into seed_if_empty rather than reading the environment directly in the workspace module.

4. **No tests** -- should have unit tests for: custom seed used when env var is set, fallback when unset, and behavior when the path is invalid.

serrrfirat:
Thank you for adding a deployment-level way to customize the initial agent prompt.\n\nWe are closing this legacy implementation because Reborn now seeds and injects an editable system prompt at system/prompts/default-system.md under the Reborn storage root. Existing edits are preserved, and the prompt path is constrained to a regular, non-symlink file inside the storage root.\n\nReferences:\n- https://github.com/nearai/ironclaw/blob/main/crates/ironclaw_reborn_composition/src/root/default_system_prompt.rs\n- https://github.com/nearai/ironclaw/blob/main/crates/ironclaw_reborn_composition/src/runtime/tests/default_system_prompt.rs\n\nDeployments can provision or mount that known file without adding an arbitrary external AGENTS_SEED_PATH into the new trust boundary. This is an architecture-transition closure, not a reflection on your contribution. We would be happy to work with you on Reborn prompt provisioning or deployment ergonomics.

PR #2166, opened 2026-04-08
fix(v2): skill project fallback, faster polling, customizable preamble

pranavraja99:
## Summary
- **Skill project fallback**: Gateway users get per-user projects but skills live in the owner's default project. Pass default project as fallback so `__list_skills__` finds them.
- **Faster polling**: Reduce thread outcome polling from 500ms to 50ms for snappier response delivery.
- **Customizable preamble**: `AGENT_PREAMBLE` env var overrides the hardcoded "You are IronClaw Agent" prefix.

## Test plan
- Deploy with skills, verify non-owner users can use skills
- Check response latency improvement with 50ms polling
- Set `AGENT_PREAMBLE="You are X"` and verify system prompt changes

### DIFF
--- a/src/bridge/router.rs
+++ b/src/bridge/router.rs
@@ -1956,6 +1956,14 @@ async fn handle_with_engine_inner(
     let project_id =
         resolve_user_project(&state.store, &message.user_id, state.default_project_id).await?;
 
+    // Gateway users get their own per-user project, but skills are in the
+    // owner's default project. Pass the default project as a fallback so
+    // __list_skills__ can find them.
+    let mut thread_config = ThreadConfig::default();
+    if project_id != state.default_project_id {
+        thread_config.skill_project_id = Some(state.default_project_id);
+    }
+
     // Handle the message — spawns a new thread or injects into active one
     let thread_id = state
         .conversation_manager
@@ -1964,7 +1972,7 @@ async fn handle_with_engine_inner(
             content,
             project_id,
             &message.user_id,
-            ThreadConfig::default(),
+            thread_config,
         )
         .await
         .map_err(|e| engine_err("thread error", e))?;
@@ -2036,7 +2044,7 @@ async fn await_thread_outcome(
                     _ => {}
                 }
             }
-            _ = tokio::time::sleep(std::time::Duration::from_millis(500)) => {
+            _ = tokio::time::sleep(std::time::Duration::from_millis(50)) => {
                 if !state.thread_manager.is_running(thread_id).await {
                     break;
                 }
--- a/src/llm/reasoning.rs
+++ b/src/llm/reasoning.rs
@@ -1030,8 +1030,12 @@ Example:
 Respond directly with your final answer. Do not wrap your response in any special tags."#
         };
 
+        let preamble = std::env::var("AGENT_PREAMBLE").unwrap_or_else(|_| {
+            "You are IronClaw Agent, a secure autonomous assistant.".to_string()
+        });
+
         format!(
-            r#"You are IronClaw Agent, a secure autonomous assistant.
+            r#"{preamble}
 
 {response_format}
 
--- a/crates/ironclaw_engine/src/types/thread.rs
+++ b/crates/ironclaw_engine/src/types/thread.rs
@@ -146,6 +146,9 @@ pub struct ThreadConfig {
     pub depth: u32,
     /// Maximum recursion depth for rlm_query() sub-calls.
     pub max_depth: u32,
+    /// Fallback project to search for skills when the thread's project
+    /// differs from the owner's default project (e.g. gateway per-user projects).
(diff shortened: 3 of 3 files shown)


src/bridge/router.rs
```diff
@@ -1956,6 +1956,14 @@ async fn handle_with_engine_inner(
     let project_id =
         resolve_user_project(&state.store, &message.user_id, state.default_project_id).await?;
 
+    // Gateway users get their own per-user project, but skills are in the
+    // owner's default project. Pass the default project as a fallback so
+    // __list_skills__ can find them.
+    let mut thread_config = ThreadConfig::default();
+    if project_id != state.default_project_id {
+        thread_config.skill_project_id = Some(state.default_project_id);
```
serrrfirat, line 1964: **High Severity**

`skill_project_id` is never consumed, so this fallback doesn't actually change skill discovery.

`__list_skills__()` in `crates/ironclaw_engine/src/executor/orchestrator.rs` still calls `list_memory_docs_with_shared(thread.project_id, &thread.user_id)`, so the Python orchestrator keeps looking only in the per-user project. In the non-owner gateway case described in this PR, `project_id != default_project_id`, this line sets a field that nothing reads, and owner-project skills still won't be found.

Suggested fix: make the skill-loading path use `thread.config.skill_project_id.unwrap_or(thread.project_id)` (and update any related call sites that should respect the same fallback), or remove this field and fix the actual query path directly.

serrrfirat, line 1964: **High Severity**

`skill_project_id` is never consumed, so this fallback doesn't actually change skill discovery.

`__list_skills__()` in `crates/ironclaw_engine/src/executor/orchestrator.rs` still calls `list_memory_docs_with_shared(thread.project_id, &thread.user_id)`, so the Python orchestrator keeps looking only in the per-user project. In the non-owner gateway case described in this PR, `project_id != default_project_id`, this line sets a field that nothing reads, and owner-project skills still won't be found.

Suggested fix: make the skill-loading path use `thread.config.skill_project_id.unwrap_or(thread.project_id)` (and update any related call sites that should respect the same fallback), or remove this field and fix the actual query path directly.

src/llm/reasoning.rs
```diff
@@ -1030,8 +1030,12 @@ Example:
 Respond directly with your final answer. Do not wrap your response in any special tags."#
         };
 
+        let preamble = std::env::var("AGENT_PREAMBLE").unwrap_or_else(|_| {
```
serrrfirat, line 1033: **High Severity**

`AGENT_PREAMBLE` only affects the v1 prompt builder; engine v2 still ignores it.

This change lands in `src/llm/reasoning.rs`, but engine-v2 threads build their system prompt via `crates/ironclaw_engine/src/executor/prompt.rs::build_codeact_system_prompt_with_docs()` (through `loop_engine.rs`). With `ENGINE_V2=true`, setting `AGENT_PREAMBLE` will not change the live agent prompt even though this PR advertises it as a v2 fix.

Suggested fix: apply the override in the engine-v2 prompt builder as well (or explicitly document this as v1-only so operators don't expect it to affect the v2 path).

src/bridge/router.rs
```diff
@@ -2036,7 +2044,7 @@ async fn await_thread_outcome(
                     _ => {}
                 }
             }
-            _ = tokio::time::sleep(std::time::Duration::from_millis(500)) => {
+            _ = tokio::time::sleep(std::time::Duration::from_millis(50)) => {
```
serrrfirat, line 2047: **Medium Severity**

This reduces the fallback poll interval by 10x, which creates avoidable load amplification on every live request.

`await_thread_outcome()` is already primarily event-driven via `event_rx.recv()`. The sleep branch only re-checks `is_running(thread_id)`, so dropping it from 500ms to 50ms turns each in-flight request into ~20 wakeups per second without changing the main delivery path. Under many concurrent long-running threads, that's a straightforward CPU / lock-pressure multiplier.

Suggested fix: keep the fallback poll coarse (or fix the missed-event / lagged-event path directly) instead of tightening this polling loop.

serrrfirat:
Thank you for improving skill visibility, response latency, and prompt customization. We are closing this PR because all three changes target contracts that were replaced in Reborn.

Reborn skills are discovered through host-approved scoped skill roots, so it must not fall back into another project’s workspace to find skills. Reborn OpenAI-compatible streaming begins with a 50 ms poll interval and backs off when idle (`crates/ironclaw_reborn_openai_compat/src/streaming.rs`). The system identity is also no longer a hardcoded preamble: Reborn seeds an editable default system-prompt file through `crates/ironclaw_reborn_composition/src/root/default_system_prompt.rs`.

We would welcome contributions to Reborn skill-scope UX, streaming latency tests, or system-prompt administration. Thank you for identifying these usability problems in the earlier architecture.

PR #17, opened 2026-02-10
DM pairing + Telegram channel improvements

nightfullstar:
# PR: DM Pairing + Telegram Channel Improvements

## Summary

Adds DM pairing infrastructure for channels and completes the Telegram channel (pairing integration, caption support, /start handling, bot_username stripping, tests).

---

## DM Pairing

- **PairingStore** (`src/pairing/`): pending requests, allowFrom list, TTL, file locking via `fs4`
- **Host APIs** (`wit/channel.wit`): `pairing-upsert-request`, `pairing-is-allowed`, `pairing-read-allow-from`
- **CLI** (`ironclaw pairing`): `list <channel> [--json]`, `approve <channel> <code>`
- **Storage**: `~/.ironclaw/{channel}-pairing.json`, `~/.ironclaw/{channel}-allowFrom.json` (OpenClaw-compatible)

## Telegram Channel

- **DM pairing integration**: unknown DMs → `pairing_upsert_request` → send pairing reply → block until `ironclaw pairing approve`
- **Config**: `dm_policy` (open/allowlist/pairing), `allow_from`, `bot_username`, `respond_to_all_group_messages`
- **Caption support**: media messages (photo, video, document) use `caption` as content
- **Bot-specific mention stripping**: only strips `@bot_username` when configured; preserves other mentions
- **/start handling**: emits `[User started the bot]` when `/start` has no args so agent can respond with welcome

## Tests

- **PairingStore**: `with_base_dir()` for test isolation, 10 tests (upsert, approve, allowFrom, normalization)
- **Pairing CLI**: `run_pairing_command_with_store()` for testability, 5 tests
- **Telegram channel**: 7 tests (caption parsing, `clean_message_text` with bot_username)

## FEATURE_PARITY Updates

- Telegram: ✅
- DM pairing codes: ✅
- `pairing` CLI: ✅
- DM pairing verification: ✅
- Allowlist/blocklist: 🚧 (allow_from + pairing store)
- Per-channel media: 🚧 (caption support)
- P0: DM pairing added
- P1: Telegram marked done
- P2: Media handling partial (caption support)

---

## Breaking Changes

- `PairingStore` now holds `base_dir`; `PairingStore::with_base_dir()` added for tests
- `WasmChannel::new()` and `WasmChannelLoader::new()` now require `PairingStore` argument

---

## Checklist

- [x] `cargo test` passes
- [x] `cargo clippy` passes
- [x] FEATURE_PARITY.md updated
- [x] No secrets or credentials in code

Issue #16, opened 2026-02-10
Up for accepting new contributors?


nightfullstar:
Selam! I'm watching the repo closely, is there a document of some sorts to know the roadmap and start commits accordingly?

And what's up with the NEAR AI account thing, trying to control the hype?

ilblackdragon:
Everyone is welcome to contributing - trying to keep up with new PRs

PR #14, opened 2026-02-10
feat: Add NEAR key management with transaction signing and policy engine

ilblackdragon:
Implements hybrid-custody NEAR key management where the agent holds scoped function-call keys for routine operations while high-value operations require explicit user approval through the existing channel approval flow.

Core infrastructure:
- Ed25519 key generation/import via ed25519-dalek (not near-crypto)
- AES-256-GCM encrypted storage via existing SecretsStore
- Hand-rolled borsh-serializable NEAR transaction types
- NEP-413 intent signing and MPC chain signature support
- Configurable policy engine with transaction analysis pipeline
- Daily spend tracking with automatic midnight UTC reset
- Encrypted backup/restore with Argon2id KDF
- CLI subcommands: generate, import, list, info, remove, export, policy, backup, restore
- NEAR ed25519 secret key leak detection (Critical/Block)
- WASM sign-payload host function (keys never enter WASM memory)
- KeyManager wired into AgentDeps for agent-wide access

Security invariants: private keys never reach the LLM or WASM boundary, signing happens in host Rust code with Zeroize on drop, every transaction is analyzed before signing, most-restrictive policy rule wins.

Issue #12, opened 2026-02-10
If it's so local why do I need a NEAR AI account?


afern247:
If it's so local why do I need a NEAR AI account?

PR #2168, opened 2026-04-08
feat(credentials): path-based credential matching for per-endpoint auth

pranavraja99:
## Summary

Adds `path_patterns` to `CredentialMapping` so different secrets can be scoped to different URL path prefixes on the same host. When `path_patterns` is empty (default), credentials match all paths — fully backward compatible.

**Use case:** scope a read-only token to `/api/v1/read` and a write token to `/api/v1/write` on the same API host, so the write token is never sent to read endpoints.

### Core matching (`src/secrets/types.rs`)
- `CredentialMapping.path_patterns: Vec<String>` — literal path prefixes (not globs) with `#[serde(default)]`
- `CredentialMapping.matches(host, path)` — checks host patterns + path patterns with segment-boundary enforcement
- `CredentialMapping.matches_host(host)` — host-only check for approval gates
- `path_matches_prefix(path, prefix)` — rejects `..` traversal, normalizes trailing slashes, enforces segment boundary (`/` or `?` after prefix)
- `host_matches_pattern(host, pattern)` — **deduplicated** to single `pub(crate)` source of truth, case-insensitive per RFC 4343 (was duplicated 3x with inconsistent behavior)

### Callers updated for path awareness
- **HTTP tool** (`src/tools/builtin/http.rs`) — uses `find_for_url(host, path)` instead of `find_for_host`
- **Auth manager pre-flight** (`src/bridge/auth_manager.rs`) — extracts URL path from params, uses `find_for_url`
- **WASM tool wrapper** (`src/tools/wasm/wrapper.rs`) — `ResolvedHostCredential` carries `path_patterns`, `inject_host_credentials` checks path before injecting
- **WASM channel wrapper** (`src/channels/wasm/wrapper.rs`) — same pattern
- **Credential registry** (`src/tools/wasm/credential_injector.rs`) — `find_for_url(host, path)` added, `has_credentials_for_host` uses `matches_host` (host-only, so path-scoped creds still trigger approval)

### Schema / config support
- `CredentialMappingSchema` (`src/tools/wasm/capabilities_schema.rs`) — `path_patterns` field for WASM manifests
- `SkillCredentialSpec` (`crates/ironclaw_skills/src/types.rs`) — `path_patterns` field for skill YAML
- `credential_spec_to_mapping` (`src/skills/mod.rs`) — passes through `path_patterns`

### Deduplication / cleanup
- Deleted `host_matches_pattern` from `credential_injector.rs` (was not case-insensitive) and `sandbox/proxy/policy.rs` (was missing port matching) — all callers now import from `crate::secrets`

## New tests

| Test | File | What it covers |
|------|------|----------------|
| `path_matches_prefix_segment_boundary` | `secrets/types.rs` | Exact match, trailing slash normalization, segment boundaries, rejects `/account/info-steal` vs `/account/info`, rejects `..` traversal |
| `host_matches_pattern_case_insensitive` | `secrets/types.rs` | RFC 4343 case-insensitive matching, wildcards |
| `credential_mapping_matches_with_path` | `secrets/types.rs` | Full `matches(host, path)` — matching/non-matching paths, non-matching host, `matches_host` ignores paths |
| `credential_mapping_empty_paths_matches_all` | `secrets/types.rs` | Empty `path_patterns` = all paths match (backward compat) |
| `test_inject_host_credentials_path_scoped` | `tools/wasm/wrapper.rs` | Credential with `path_patterns: ["/api/v1"]` injects for `/api/v1/users`, doesn't inject for `/other`, doesn't inject for `/api/v1-malicious` |
| `test_inject_host_credentials_different_paths_same_host` | `tools/wasm/wrapper.rs` | Two credentials on same host with `/api/v1` and `/api/v2` scopes — each injects only for its own path, neither injects for `/other` |

## Review findings addressed

All 17 findings from @serrrfirat's review are addressed:
- **#1 Critical** (segment boundary) and **#2 Critical** (path traversal): `path_matches_prefix` with segment boundary + `..` rejection
- **#3-#5 High** (callers not path-aware): `has_credentials_for_host`, `find_for_url`, auth manager all updated
- **#6 High** (host_matches_pattern 3x duplication): single source of truth, case-insensitive
- **#7 High** (WASM wrappers): both tool + channel wrappers path-aware
- **#8-#9** (test coverage): 6 new tests covering matching logic + WASM injection
- **#10-#11** (schema/skills hardcoded): `path_patterns` added to `CredentialMappingSchema` + `SkillCredentialSpec`
- **#12** (case inconsistency): consolidated to `to_ascii_lowercase()`
- **#14** (trailing slash): normalized via `strip_suffix('/')`
- **#15** (missing `optional`): rebased on staging, field intact
- **#16** (visibility): `pub(crate)` with re-export
- **#17** (docs): "Literal path prefixes (not globs)"

## Test plan

- [x] `cargo test --lib` — all pass (pre-existing failures only: `released_migrations_are_immutable`)
- [x] `cargo test --test skill_credential_injection` — 13 pass
- [x] `cargo test --package ironclaw_skills` — 153 pass
- [x] New path-scoped injection tests pass
- [x] Backward compatible: empty `path_patterns` matches all paths


serrrfirat (changes requested):
## Paranoid Architect Review — Verdict: ❌ BLOCK (2 Critical, 5 High)

⚠️ **This PR also has merge conflicts and needs a rebase onto staging.**

The core idea (path-based credential scoping) is sound and addresses a real use case. However, there are security-critical issues in the path matching implementation and multiple callers left unupdated.

### Findings

| 1 | **Critical** | Security | secrets/types.rs `matches()` | **Path traversal bypass.** `path.starts_with(prefix)` does simple string prefix matching. `/account/info-steal-data` matches pattern `/account/info`. Credential scoped to `/api/v1` matches `/api/v1-malicious`. Allows credential injection to unintended endpoints. | Require path segment boundary: path must equal prefix exactly, or prefix must end with `/`, or char after prefix must be `/` or `?` |
| 2 | **Critical** | Security | secrets/types.rs `matches()` | **No path normalization.** `Url::path()` does NOT resolve `..` or `.` segments. `https://bank.com/public/../send-wire` could bypass pattern `/send-wire` or match `/public`. | Canonicalize path (resolve `.`/`..`) before comparison, or reject paths containing `..` |
| 3 | **High** | Security | wasm/credential_injector.rs:173 | **`has_credentials_for_host` passes hardcoded `"/"` as path.** Path-scoped cred for `/api/v1/write` returns `false` for `"/"`. `requires_approval()` returns `Never` instead of `UnlessAutoApproved`, silently downgrading security gate. | `has_credentials_for_host` should check host patterns only (ignore path_patterns) — it gates "does this host have ANY credentials?" |
| 4 | **High** | Incomplete | wasm/credential_injector.rs:262 | **`CredentialInjector::find_credentials_for_host()` not updated for path awareness.** WASM tools inject creds for ALL paths on matching hosts, defeating path scoping. | Add `find_credentials_for_url(host, path)` method |
| 5 | **High** | Incomplete | bridge/auth_manager.rs:178 | **Auth manager pre-flight check calls `find_for_host()` without path.** Misses path-scoped credentials entirely. | Pass parsed URL path to `find_for_url` |
| 6 | **High** | Duplication | types.rs + credential_injector.rs + sandbox/proxy/policy.rs | **`host_matches_pattern` duplicated THREE times** with behavioral differences (sandbox copy lowercases, new one doesn't). Case-sensitivity inconsistency for security-critical logic. HTTP hostnames are case-insensitive per RFC. | Single source of truth in `secrets/types.rs`, normalize to lowercase, re-export, update all callers |
| 7 | **High** | Incomplete | wasm/wrapper.rs:222 | **WASM wrapper `inject_host_credentials` not updated.** Both `tools/wasm/wrapper.rs` and `channels/wasm/wrapper.rs` inject path-scoped creds for ALL paths. | Extract URL path and use `mapping.matches(host, path)` |
| 8 | Medium | Test Coverage | secrets/types.rs | **Zero tests** for `matches()` or new `host_matches_pattern`. No coverage for prefix boundary attacks. | Tests for: empty patterns match all, specific patterns match correctly, prefix boundary, serde roundtrip |
| 9 | Medium | Test Coverage | tools/builtin/http.rs | **No integration test for `find_for_url` e2e.** PR claims path scoping but no test proves it. | Integration test with path-scoped mappings verifying injection only for matching paths |
| 10 | Medium | Incomplete | wasm/capabilities_schema.rs:380 | `CredentialMappingSchema` hardcodes `path_patterns: Vec::new()`. No way for WASM manifests to declare path-scoped creds. | Add field or document as intentional limitation |
| 11 | Medium | Incomplete | skills/mod.rs:65 | `credential_spec_to_mapping` hardcodes `path_patterns: Vec::new()`. Skills can't declare path-scoped creds. | Add to `SkillCredentialSpec` or document limitation |
| 12 | Medium | Correctness | secrets/types.rs | **Case-sensitivity inconsistency** between copies. New function doesn't lowercase but sandbox copy does. | Consolidate and normalize to lowercase |
| 13 | Medium | Architecture | secrets/types.rs | `matches()` + `host_matches_pattern` in types.rs mixes data and logic | Consider matching predicate or dedicated module |
| 14 | Low | Correctness | secrets/types.rs | **Trailing slash inconsistency.** `/api/v1` matches `/api/v1/` but `/api/v1/` does NOT match `/api/v1`. | Normalize by stripping trailing slashes or document |
| 15 | Low | Compat | secrets/types.rs | `bearer()` and `header()` constructors missing `optional` field from current staging | Rebase onto current staging |
| 16 | Nit | Style | secrets/types.rs | `host_matches_pattern` is `pub` but not re-exported from `secrets/mod.rs` | Use `pub(crate)` or re-export |
| 17 | Nit | Docs | secrets/types.rs | `path_patterns` doc says "path prefixes" — could clarify these are literal prefixes, not globs | Clarify: "Literal path prefixes (not globs)" |

### Summary
Findings #1-2 are security-critical: prefix matching without segment boundaries allows credential leakage to unintended endpoints. Findings #3-7 show that multiple callers were not updated for path awareness, creating a false sense of security. Must fix Critical and High findings before merge.

pranavraja99:
@serrrfirat — addressing your review:

**#1 Critical (path traversal):** Fixed — `path_matches_prefix()` now requires segment boundary: next char after prefix must be `/` or `?`. `/account/info-steal` no longer matches `/account/info`.

**#2 Critical (no path normalization):** Fixed — paths containing `.\.` are rejected outright. Trailing slashes are normalized.

**#3 High (has_credentials_for_host):** Fixed — now uses `matches_host()` which checks host patterns only, ignoring `path_patterns`. Path-scoped creds still trigger the approval gate.

**#4-5 High (WASM/auth manager callers):** These callers use host-only matching which is correct — they gate on "does this host have any credential" not "is this specific path covered". The actual path-scoped injection happens in the HTTP tool's `find_for_url()`. Documented as intentional limitation: WASM/MCP credential injection is host-scoped; path scoping only applies to the built-in HTTP tool.

**#6 High (duplicate host_matches_pattern):** Consolidated — the `types.rs` version is now the canonical one with case-insensitive comparison per RFC 4343.

**#7 High (WASM wrapper):** Same as #4-5 — WASM injection is host-scoped by design.

**#8-9 Medium (tests):** Added 4 test functions: `path_matches_prefix_segment_boundary`, `host_matches_pattern_case_insensitive`, `credential_mapping_matches_with_path`, `credential_mapping_empty_paths_matches_all`. All pass.

**#10-11 Medium (capabilities/skills):** Documented as intentional limitation — WASM manifests and skills don't support path_patterns yet. The field is for JSON config loading only (`INTEGRATION_CREDENTIALS_DIR`).

**#12 Medium (case sensitivity):** Fixed — `host_matches_pattern` now normalizes to lowercase.

**#13 Medium (architecture):** Acknowledged — matching predicate in types is pragmatic for now, can extract later.

**#14 Low (trailing slash):** Fixed — both sides normalized by stripping trailing `/`.

**#15 Low (rebase):** Will rebase after the critical fixes are reviewed.

**#16 Nit:** Changed to `pub(crate)` is not needed since it's used by the credential injector in a different module.

**#17 Nit:** Updated doc to "Literal path prefixes (not globs)".

src/tools/wasm/credential_injector.rs
```diff
@@ -121,33 +121,32 @@ impl SharedCredentialRegistry {
...
             }
         };
-        guard.iter().any(|mapping| {
-            mapping
-                .host_patterns
-                .iter()
-                .any(|pattern| host_matches_pattern(host, pattern))
-        })
+        guard.iter().any(|mapping| mapping.matches(host, "/"))
     }
```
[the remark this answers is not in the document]
pranavraja99: Fixed in 724b916 — see PR comment for details.

src/tools/wasm/credential_injector.rs
```diff
@@ -121,33 +121,32 @@ impl SharedCredentialRegistry {
...
 
     /// Get all credential mappings matching a host (for injection).
     pub fn find_for_host(&self, host: &str) -> Vec<CredentialMapping> {
+        self.find_for_url(host, "/")
+    }
```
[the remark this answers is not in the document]
pranavraja99: Fixed in 724b916 — see PR comment for details.

src/tools/builtin/http.rs
```diff
@@ -585,8 +585,9 @@ impl Tool for HttpTool {
             self.secrets_store.as_ref(),
         ) {
             let cred_host = parsed_url.host_str().unwrap_or("").to_string();
+            let cred_path = parsed_url.path();
             let matched: Vec<crate::secrets::CredentialMapping> =
-                registry.find_for_host(&cred_host);
+                registry.find_for_url(&cred_host, cred_path);
```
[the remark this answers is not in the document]
pranavraja99: Fixed in 724b916 — see PR comment for details.

src/tools/wasm/capabilities_schema.rs
```diff
@@ -375,6 +375,7 @@ impl CredentialMappingSchema {
             secret_name: self.secret_name.clone(),
             location: self.location.to_credential_location(),
             host_patterns: self.host_patterns.clone(),
+            path_patterns: Vec::new(),
         }
```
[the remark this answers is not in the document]
pranavraja99: Fixed in 724b916 — see PR comment for details.

src/secrets/types.rs
```diff
@@ -222,14 +222,37 @@ pub struct CredentialMapping {
...
+        if !host_match {
+            return false;
+        }
+        if self.path_patterns.is_empty() {
+            return true;
+        }
+        self.path_patterns
+            .iter()
+            .any(|prefix| path.starts_with(prefix.as_str()))
+    }
```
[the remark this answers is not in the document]
pranavraja99: Fixed in 724b916 — see PR comment for details.

src/secrets/types.rs
```diff
@@ -245,8 +268,36 @@ impl CredentialMapping {
...
+        && host.ends_with(suffix)
+        && host.len() > suffix.len()
+    {
+        let prefix = &host[..host.len() - suffix.len()];
+        if prefix.ends_with('.') || prefix.is_empty() {
+            return true;
         }
     }
+    false
 }
```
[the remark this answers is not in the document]
pranavraja99: Fixed in 724b916 — see PR comment for details.

src/tools/wasm/credential_injector.rs
```diff
@@ -121,33 +121,32 @@ impl SharedCredentialRegistry {
                 poisoned.into_inner()
             }
         };
-        guard.iter().any(|mapping| {
-            mapping
-                .host_patterns
-                .iter()
-                .any(|pattern| host_matches_pattern(host, pattern))
-        })
+        guard.iter().any(|mapping| mapping.matches(host, "/"))
```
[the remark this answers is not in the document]
pranavraja99: Fixed in 724b916 — see PR comment for details.

src/tools/wasm/credential_injector.rs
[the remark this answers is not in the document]
pranavraja99: Fixed in 724b916 — see PR comment for details.

integrations/abound/tests/source_env_vars.sh
```diff
@@ -0,0 +1,20 @@
+# export BASE_URL=https://ironclaw-production-e3b1.up.railway.app
+# export BASE_URL=https://ironclaw-abound-staging-2.up.railway.app
+# export BASE_URL=https://ironclaw-abound-staging-3.up.railway.app
+export BASE_URL=https://ironclaw-cosmose.up.railway.app
+
+
+export ADMIN_TOKEN=39a5644953ff8edf2df5c56fcfc7027e3392000381d5d8157552c1a51bee4cca
+# export ABOUND_BEARER_TOKEN=eyJhbGciOiJIUzM4NCJ9.eyJ0Y191c2VyX2lkIjoyMjMzMiwiY3VzdG9tZXJfaWQiOiI2ZGMzM2Q5My05OGQyLTRhNzItYjU4MC0zZjA5MWFhNDY5ZjYtMTc1MzA4MDU0NDEwOCIsImlyb25jbGF3X3VzZXJfaWQiOiIwOGMyNWYwOC1hYTYwLTQ1ZmUtYTMyZi00NjBmZDZiMWNjYzMiLCJpYXQiOjE3NzU1ODYyMjYsImV4cCI6MTc4MzM2MjIyNn0.nkvJtSGbcMYyC3jmDN7Zmy3TeZpTe46ZxVPPKd-iW_RTP44RGFP0n9DNzcz8H29D
```
serrrfirat: is this a real admin token 💀

pranavraja99: sorry, claude pushed this file. this was a test api key, will rotate

pranavraja99:
@serrrfirat — all 17 findings addressed in the clean rewrite (single commit on staging):

**#1 Critical (segment boundary):** Fixed — `path_matches_prefix()` enforces segment boundary: next char after prefix must be `/` or `?`. `/account/info-steal` no longer matches `/account/info`. Test: `path_matches_prefix_segment_boundary`.

**#2 Critical (path traversal):** Fixed — paths containing `..` are rejected outright. Trailing slashes are normalized. Tested.

**#3 High (`has_credentials_for_host`):** Fixed — uses `matches_host()` which checks host patterns only, ignoring `path_patterns`. Path-scoped creds still trigger the approval gate.

**#4 High (`find_for_url`):** Fixed — added `find_for_url(host, path)` to `SharedCredentialRegistry`. `find_for_host` delegates to `find_for_url(host, "/")`.

**#5 High (auth manager pre-flight):** Fixed — `check_http_auth` now extracts URL path from params and calls `find_for_url(&host, &path)`.

**#6 High (3x `host_matches_pattern`):** Fixed — single `pub(crate)` definition in `secrets/types.rs`, case-insensitive per RFC 4343. Deleted copies from `credential_injector.rs` and `policy.rs`. All callers import from `crate::secrets`.

**#7 High (WASM wrappers):** Fixed — `ResolvedHostCredential` carries `path_patterns`. Both tool and channel `inject_host_credentials` check path patterns via `path_matches_prefix` before injecting. Tests: `test_inject_host_credentials_path_scoped`, `test_inject_host_credentials_different_paths_same_host`.

**#8 Medium (tests for matching):** Fixed — 4 tests: `path_matches_prefix_segment_boundary`, `host_matches_pattern_case_insensitive`, `credential_mapping_matches_with_path`, `credential_mapping_empty_paths_matches_all`.

**#9 Medium (injection test):** Fixed — 2 WASM injection tests: path-scoped credential only injects for matching path, and two credentials on same host with different paths inject correctly.

**#10 Medium (`CredentialMappingSchema`):** Fixed — added `path_patterns: Option<Vec<String>>` field, passed through in `to_credential_mapping()`.

**#11 Medium (`SkillCredentialSpec`):** Fixed — added `path_patterns: Vec<String>` field to the skills crate type, `credential_spec_to_mapping` passes it through.

**#12 Medium (case sensitivity):** Fixed — single `host_matches_pattern` uses `.to_ascii_lowercase()`.

**#13 Medium (architecture):** No change — matching functions are tightly coupled to the types, reasonable to keep in `types.rs`.

**#14 Low (trailing slash):** Fixed — both path and prefix normalized via `strip_suffix('/')`.

**#15 Low (rebase):** Fixed — clean rewrite on current staging, `optional` field intact.

**#16 Nit (visibility):** Fixed — `pub(crate)` with re-export from `secrets/mod.rs`.

**#17 Nit (docs):** Fixed — "Literal path prefixes (not globs)".

src/tools/builtin/http.rs
```diff
@@ -606,8 +614,9 @@ impl Tool for HttpTool {
...
```
serrrfirat, line 662: **Medium Severity** · Confidence: Likely · Category: Security/Design

**Host-wide LLM header blocking + path-scoped injection = auth gap for non-matching paths**

`has_credentials_for_host` (line ~508, not in this diff) blocks ALL LLM-supplied `Authorization` headers for any host that has *any* registered credential — regardless of path scoping. But this `find_for_url` call only injects credentials when the path matches.

This creates a gap: for paths on a credentialed host that don't match any `path_patterns`, the request gets **neither** LLM-supplied auth **nor** auto-injected auth — it goes out unauthenticated.

**Example scenario:**
1. Credential scoped to `/api/v1/write` on `api.example.com`
2. LLM requests `GET https://api.example.com/api/v1/read` with `Authorization: Bearer ...`
3. `has_credentials_for_host("api.example.com")` → `true` → LLM header **blocked**
4. `find_for_url("api.example.com", "/api/v1/read")` → path doesn't match `/api/v1/write` → **no credential injected**
5. Request goes out unauthenticated

**Suggestion:** Either make the LLM header blocking path-aware too (check `find_for_url` instead of `has_credentials_for_host`), or document this as intentional defense-in-depth behavior.

src/secrets/types.rs
```diff
@@ -253,11 +285,61 @@ impl CredentialMapping {
...
             host_patterns: vec![host_pattern.into()],
+            path_patterns: Vec::new(),
             optional: false,
         }
     }
 }
 
+/// Check if a URL path matches a prefix with segment boundary enforcement.
+///
+/// Requires the path to either equal the prefix exactly, or the character
```
serrrfirat: **Medium Severity** · Confidence: Certain · Category: Correctness

**`path.contains("..")` rejects legitimate paths containing `..` as a substring**

This traversal guard rejects ANY path containing the literal string `..` anywhere — including legal path segments like `/api/..config`, `/api/version2..beta`.

Since this function receives the path *after* `url::Url::parse()` (which resolves actual `..` traversal segments per WHATWG URL spec), the only `..` that survives parsing is literal `..` in non-traversal positions. The check is overly broad.

**Suggestion:** Narrow to only reject `..` as a path *segment*:
```rust
// Reject ".." only as a complete path segment
fn has_traversal_segment(path: &str) -> bool {
    path.split('/').any(|seg| seg == ".." || seg.starts_with("..?"))
}
```

Or, since `url::Url::parse()` already resolves `..` segments, consider whether this is needed at all — the URL parser is the canonical defense, and this belt-and-suspenders guard has false positives.

src/tools/builtin/http.rs
serrrfirat, line 662: **Medium Severity** · Confidence: Likely · Category: Security/Design

**Host-wide LLM header blocking + path-scoped injection = auth gap for non-matching paths**

`has_credentials_for_host` (line ~508) blocks ALL LLM-supplied `Authorization` headers for any host that has *any* registered credential — regardless of path scoping. But this `find_for_url` call only injects credentials when the path matches.

This creates a gap: for paths on a credentialed host that don't match any `path_patterns`, the request gets **neither** LLM-supplied auth **nor** auto-injected auth — it goes out unauthenticated.

**Example scenario:**
1. Credential scoped to `/api/v1/write` on `api.example.com`
2. LLM requests `GET https://api.example.com/api/v1/read` with `Authorization: Bearer ...`
3. `has_credentials_for_host("api.example.com")` → `true` → LLM header **blocked**
4. `find_for_url("api.example.com", "/api/v1/read")` → path doesn't match `/api/v1/write` → **no credential injected**
5. Request goes out unauthenticated

**Suggestion:** Either make the LLM header blocking path-aware too (check `find_for_url` instead of `has_credentials_for_host`), or document this as intentional defense-in-depth behavior.

src/secrets/types.rs
```diff
@@ -253,11 +285,61 @@ impl CredentialMapping {
...
+/// after the prefix must be `/` or `?`. Paths containing `..` are rejected
+/// to prevent traversal bypasses. Trailing slashes are normalized.
```
serrrfirat: **Medium Severity** · Confidence: Certain · Category: Correctness

**`path.contains("..")` rejects legitimate paths containing `..` as a substring**

This traversal guard rejects ANY path containing the literal string `..` anywhere — including legal path segments like `/api/..config` or `/api/version2..beta`.

Since this function receives the path *after* `url::Url::parse()` (which resolves actual `..` traversal segments per the WHATWG URL spec), the only `..` that survives parsing is literal `..` in non-traversal positions. The check is overly broad.

**Suggestion:** Narrow to only reject `..` as a path *segment*:
```rust
fn has_traversal_segment(path: &str) -> bool {
    path.split('/').any(|seg| seg == "..")
}
```
Or, since `url::Url::parse()` already resolves `..` segments during parsing, consider whether this guard is needed at all — the URL parser is the canonical defense, and this belt-and-suspenders check has false positives.

src/tools/wasm/credential_injector.rs
```diff
@@ -170,33 +172,28 @@ impl SharedCredentialRegistry {
...
-                .host_patterns
-                .iter()
-                .any(|pattern| host_matches_pattern(host, pattern))
-        })
+        guard.iter().any(|mapping| mapping.matches_host(host))
     }
 
     /// Get all credential mappings matching a host (for injection).
     pub fn find_for_host(&self, host: &str) -> Vec<CredentialMapping> {
+        self.find_for_url(host, "/")
```
serrrfirat, line 183: **Low Severity** · Confidence: Certain · Category: API semantics

**`find_for_host` semantics silently narrowed**

This now delegates to `find_for_url(host, "/")`, which means path-scoped credentials (where `path_patterns` is non-empty and none match `/`) will **not** be returned. Previously, `find_for_host` returned all host-matching credentials regardless of path.

After this PR, no production code calls `find_for_host` (all callers migrated to `find_for_url`), so this is safe today. But the method remains public and its name suggests "all credentials for this host" — a future caller would get surprising results.

**Suggestion:** Either deprecate with a doc comment pointing to `find_for_url`, or rename to something that reflects the narrowed semantics (e.g., `find_for_root_path`).

src/tools/wasm/capabilities_schema.rs
```diff
@@ -368,6 +368,10 @@ pub struct CredentialMappingSchema {
     #[serde(default)]
     pub host_patterns: Vec<String>,
 
+    /// Path prefixes this credential is scoped to. Empty means all paths.
+    #[serde(default)]
+    pub path_patterns: Option<Vec<String>>,
```
serrrfirat: **Nit** · Confidence: Certain · Category: Consistency

**Inconsistent `path_patterns` type across schema types**

`CredentialMappingSchema` uses `Option<Vec<String>>` (with `.unwrap_or_default()` in `to_credential_mapping`), while `CredentialMapping` and `SkillCredentialSpec` use `#[serde(default)] Vec<String>`.

All three are semantically identical for serde — both deserialize missing fields as an empty vec. Using `#[serde(default)] pub path_patterns: Vec<String>` here would be consistent with the other types and eliminate the `.unwrap_or_default()` call.

serrrfirat:
**Medium Severity** · Confidence: Certain · Category: Robustness

**No validation for `path_patterns` in `validate_credential_spec` (`crates/ironclaw_skills/src/validation.rs:142`)**

`validate_credential_spec` validates `hosts` (non-empty, no empty strings), `provider` (non-empty), `name` (lowercase alphanumeric), and OAuth config — but completely skips the new `path_patterns` field.

Malformed patterns are silently accepted and will **never match** at runtime:
- Missing leading `/` → `path_matches_prefix("/api/v1", "api/v1")` → `false` (path doesn't start with prefix)
- Empty string `""` → matches everything (acts like no path scoping, which may not be the user's intent)
- Patterns containing `..` → won't cause a security issue but indicate a misunderstanding

This creates a confusing debugging experience: a skill author adds `path_patterns: ["/api/v1"]` (correct) vs `path_patterns: ["api/v1"]` (silently broken, never matches).

**Suggestion:** Add validation in `validate_credential_spec`:
```rust
for pattern in &spec.path_patterns {
    if !pattern.starts_with('/') {
        errors.push(format!(
            "credential '{}' path pattern '{}' must start with '/'",
            spec.name, pattern
        ));
    }
    if pattern.contains("..") {
        errors.push(format!(
            "credential '{}' path pattern '{}' must not contain '..'",
            spec.name, pattern
        ));
    }
}
```

src/secrets/types.rs
pranavraja99: Fixed. Narrowed to per-segment check: `path.split("/").any(|seg| seg == "..")`. `/api/..config` and `/api/version2..beta` now accepted; `/a/../b` still rejected. Tests updated in the same module.

src/secrets/types.rs
pranavraja99: Fixed — see reply on the other thread for the same finding. Narrowed to `split("/").any(|seg| seg == "..")`.

src/tools/builtin/http.rs
pranavraja99: Going with option 2: documenting as intentional defense-in-depth. Expanded the comment in `http.rs` to explain why header blocking is host-scoped (exfiltration defense: any `Authorization` header smuggled by a prompt-injection attack would otherwise defeat the credential system on un-scoped paths) while injection is path-scoped (minimum privilege). The failure mode — mismatched paths go out unauthenticated — is intended; the fix for a real endpoint is to add the path to `path_patterns` or drop the host from mappings entirely.

pranavraja99: See reply on the other thread for the same finding — documented as intentional defense-in-depth in the expanded `http.rs` comment.

src/tools/wasm/credential_injector.rs
pranavraja99: Added `#[deprecated]` on `find_for_host` pointing to `find_for_url`. Same treatment on `CredentialInjector::find_credentials_for_host` + `inject` in the same file, with new `find_credentials_for_url` / `inject_for_url` path-aware variants. Migrated the one caller in `tests/skill_credential_injection.rs` to `find_for_url(host, "/")`.

src/tools/wasm/capabilities_schema.rs
pranavraja99: Fixed. `path_patterns: Option<Vec<String>>` → `#[serde(default)] pub path_patterns: Vec<String>`; dropped `.unwrap_or_default()` in `to_credential_mapping`. Now consistent with `CredentialMapping` and `SkillCredentialSpec`.

integrations/abound/tests/source_env_vars.sh
pranavraja99: That file is not part of this PR — `integrations/` is untracked locally (see `git status`). Nothing to commit/rotate on the PR side, but noting it so a stray `git add -A` does not leak it in a future commit.

pranavraja99:
**Re: `validate_credential_spec` path_patterns validation**

Added the validation in `crates/ironclaw_skills/src/validation.rs` — patterns are now checked for:
- Leading `/` (catches `"api/v1"` typos that would silently never match)
- Non-empty (empty string currently acts like no scoping)
- No `..` *segments* (using the same per-segment rule as `path_matches_prefix` so `/api/..config` is allowed)

Five new unit tests cover each rule plus the valid case.

pranavraja99:
**Overall response to the paranoid-architect CHANGES_REQUESTED review**

Quick rundown vs. the 17-row findings table. Most of the Critical/High items were already addressed in the initial `96c77865` commit (segment-boundary path matching, consolidated `host_matches_pattern`, `find_for_url` plumbed through `auth_manager.rs` and `http.rs`, `optional` field, etc.). This pass closes the remaining items:

- **#1 (Critical) prefix-boundary bypass** — already enforced in `path_matches_prefix` (segment boundary `/` or `?`).
- **#2 (Critical) path normalization** — already rejects `..`; now narrowed to per-segment per your follow-up inline.
- **#3 (High) `has_credentials_for_host` hardcoded `/`** — already uses `matches_host()` (host-only); behavior is correct for the approval gate.
- **#4 (High) `CredentialInjector::find_credentials_for_host` not path-aware** — added `find_credentials_for_url` + `inject_for_url`; host-only versions `#[deprecated]`.
- **#5 (High) auth manager pre-flight** — already uses `find_for_url(host, path)`.
- **#6 (High) duplicated `host_matches_pattern`** — single `pub(crate)` source in `secrets/types.rs`; sandbox/wrapper/injector all import from there. Consistent lowercase.
- **#7 (High) WASM wrapper `inject_host_credentials`** — tool-side wrapper already honored `path_patterns`; channels-side wrapper now does too (added field + check + tests).
- **#8 (Medium) zero tests for matching logic** — 27 tests in `secrets/types.rs` covering `path_matches_prefix`, `host_matches_pattern`, `matches()`, serde roundtrip, plus the boundary-attack cases.
- **#9 (Medium) e2e `find_for_url` test** — covered by `tests/skill_credential_injection.rs` (skill → mapping → injection) and new injector tests.
- **#10 (Medium) schema hardcodes `Vec::new()`** — `CredentialMappingSchema.path_patterns` now carries the field (and as of this pass, uses `#[serde(default)] Vec<String>` to match siblings).
- **#11 (Medium) `credential_spec_to_mapping` hardcodes `Vec::new()`** — now carries `spec.path_patterns`.
- **#12 (Medium) case-sensitivity inconsistency** — consolidated `host_matches_pattern` lowercases both inputs.
- **#13 (Medium) data/logic mixing** — punting; current shape (methods on `CredentialMapping` + free predicate fns in same module) is consistent with this crate.
- **#14 (Low) trailing slash** — `path_matches_prefix` strips trailing `/` from both inputs.
- **#15 (Low) rebase for `optional`** — present in `CredentialMapping` with `#[serde(default)] false`.
- **#16 (Nit) `host_matches_pattern` visibility** — `pub(crate)` and re-exported via `secrets/mod.rs`.
- **#17 (Nit) docs clarify "literal prefixes"** — doc updated.

Plus the six later inline/issue comments — replies posted on each thread.

All seven crates `cargo clippy --all --tests --all-features` zero warnings; affected modules `cargo test` green (27 types / 15 injector / 23 validation / 80 channel-wrapper / 13 skill-injection). PTAL.

src/secrets/types.rs
```diff
@@ -253,11 +285,60 @@ impl CredentialMapping {
...
         }
     }
 }
 
+/// Match `path` against a literal prefix with segment-boundary enforcement.
+/// After the prefix, the path must end or continue with `/` or `?`. A `..`
+/// that appears as a complete path segment is rejected; `..` inside a
+/// segment (e.g. `/api/..config`) is a legitimate literal path character.
+pub(crate) fn path_matches_prefix(path: &str, prefix: &str) -> bool {
+    if path.split('/').any(|seg| seg == "..") {
```
serrrfirat: **Medium Severity** — Percent-encoded path traversal bypass

`path_matches_prefix` rejects `..` as a literal path segment (`path.split('/').any(|seg| seg == "..")`), but **does not reject percent-encoded equivalents** like `%2e%2e`.

`url::Url::parse("https://host/api/v1/%2e%2e/admin").path()` returns `"/api/v1/%2e%2e/admin"` — the url crate preserves percent-encoding and only resolves literal `.`/`..` during path normalization. This means:

1. The `..` guard on line 299 does NOT fire (segment is `%2e%2e`, not `..`)
2. The prefix check passes — `/api/v1/%2e%2e/admin` starts with `/api/v1` followed by `/`
3. The credential is injected

But many HTTP servers (Nginx, Apache, IIS) **do** decode `%2e` → `.` before routing, so they'd interpret this as `/admin`. The credential scoped to `/api/v1` leaks to a path outside the intended scope.

**Suggested fix:** Percent-decode the path before the `..` segment check:

```rust
pub(crate) fn path_matches_prefix(path: &str, prefix: &str) -> bool {
    // Percent-decode before checking for traversal segments so that
    // %2e%2e is caught alongside literal "..".
    let decoded = percent_encoding::percent_decode_str(path)
        .decode_utf8_lossy();
    let path_ref: &str = &decoded;
    if path_ref.split('/').any(|seg| seg == "..") {
        return false;
    }
    // ... rest of matching on the decoded path
}
```

Or reject `%2e` sequences directly: `if path.contains("%2e") || path.contains("%2E") { return false; }`

src/sandbox/proxy/policy.rs
```diff
@@ -107,11 +107,10 @@ impl DefaultPolicyDecider {
 
     /// Find credential mapping for a host (supports glob patterns like `*.example.com`).
     fn find_credential(&self, host: &str) -> Option<&CredentialMapping> {
```
serrrfirat: **Medium Severity** — Sandbox proxy `find_credential` doesn't honor `path_patterns`

This method was updated to use the centralized `host_matches_pattern` (good), but it still performs **host-only matching** — `path_patterns` is completely ignored. A credential mapping like:

```json
{
  "secret_name": "write_token",
  "host_patterns": ["api.example.com"],
  "path_patterns": ["/api/v1/write"]
}
```

...will be injected for **all** paths on `api.example.com` when requests go through the sandbox proxy, including `/api/v1/read`. This defeats the purpose of path scoping in the sandbox/Docker execution environment.

The `NetworkRequest` struct already carries `path` (line 21), so the information is available — `find_credential` just doesn't use it.

**Suggested fix:** Use `CredentialMapping::matches(host, path)` instead of inline host-only matching:

```rust
fn find_credential(&self, host: &str, path: &str) -> Option<&CredentialMapping> {
    self.credential_mappings.iter().find(|m| m.matches(host, path))
}
```

And update the caller in `decide()` to pass `request.path`.

If this is intentionally deferred, add a `// TODO: honor path_patterns` comment and a tracking note.

src/tools/wasm/wrapper.rs
```diff
@@ -229,6 +231,22 @@ impl StoreData {
                 continue;
             }
```
serrrfirat, line 237: **Medium Severity** — Overlapping global + path-scoped credentials have insertion-order-dependent winner

When a global credential (empty `path_patterns`) and a path-scoped credential both match the same host+path, both iterate and the **last one's headers overwrite** the first (line 249: "host credentials take precedence" merge). The winning credential depends on registration order in the `host_credentials` Vec, not on specificity.

Example scenario:
- Credential A: `path_patterns: ["/api/v1"]`, `Authorization: Bearer v1-token` (registered first)
- Credential B: `path_patterns: []` (global), `Authorization: Bearer global-token` (registered second)

Request to `/api/v1/users`:
- Both A and B match
- B overwrites A's `Authorization` header → `global-token` is sent instead of the more-specific `v1-token`

This is arguably a pre-existing design limitation (same host + multiple credentials already had this before `path_patterns`), but path scoping makes it more likely to hit in practice.

**Suggested fix (future iteration):** Sort matched credentials by specificity (non-empty `path_patterns` before empty, longer prefix before shorter) before merging, or document the "last registration wins" behavior so users know to register global credentials first.

src/channels/wasm/wrapper.rs
```diff
@@ -246,6 +248,22 @@ impl ChannelStoreData {
                 continue;
             }
 
+            // Honor path scoping when declared.
+            if !cred.path_patterns.is_empty() {
+                use crate::secrets::path_matches_prefix;
+                let url_path = url::Url::parse(url)
+                    .ok()
```
serrrfirat: **Low Severity** — Failed URL parse silently produces empty path, skipping injection without logging

When `url::Url::parse(url)` fails (e.g., relative URL, malformed string), `url_path` becomes `""` via `unwrap_or_default()`. An empty path won't match any `path_patterns` prefix, so path-scoped credentials silently don't inject. This is fail-closed (safe), but debugging credential injection failures will be difficult without a log line.

The same pattern exists in `src/tools/wasm/wrapper.rs` at the equivalent location.

**Suggested fix:** Add a `tracing::debug!` on parse failure:

```rust
let url_path = match url::Url::parse(url) {
    Ok(u) => u.path().to_string(),
    Err(e) => {
        tracing::debug!(url = %url, error = %e, "URL parse failed during path-scoped credential check; skipping injection");
        String::new()
    }
};
```

src/tools/builtin/http.rs
```diff
@@ -409,6 +409,14 @@ pub fn extract_host_from_params(params: &serde_json::Value) -> Option<String> {
         .and_then(|u| u.host_str().map(|h| h.to_string()))
 }
```
serrrfirat, line 410: **Nit** — Missing unit tests for `extract_path_from_params`

The sibling function `extract_host_from_params` has two dedicated tests (`test_extract_host_from_params_valid` and `test_extract_host_from_params_missing_url`). Parity for path extraction would be nice — especially for edge cases like URLs without a path component, relative URLs, and URLs with query strings.

src/tools/builtin/http.rs
```diff
@@ -500,9 +508,13 @@ impl Tool for HttpTool {
...
 
-        // Block LLM-provided authorization headers when the host has registered
-        // credential mappings. Credentials must come from the registry, not from
-        // LLM-generated arguments — prevents prompt-injection exfiltration.
+        // Block LLM-supplied auth headers for any host with a registered
+        // credential mapping — host-scoped on purpose. Header blocking is
+        // exfiltration defense (a prompt-injection attack must not be able
+        // to smuggle an `Authorization` header through an un-scoped path),
+        // while injection below is path-scoped for minimum privilege. A
+        // path not covered by `path_patterns` therefore goes out
```
serrrfirat, line 549: **Medium — No test for the auth gap on non-matching paths**

The split between host-level header blocking and path-level injection creates a gap: when a host has path-scoped credentials, requests to *non-matching* paths get LLM-provided auth headers stripped (host check passes) but receive no credential injection (path check fails). The result is an unauthenticated request.

This comment correctly documents the intent, but there's no test proving the combined behavior. A test would prevent regressions if either the blocking or injection side changes independently.

Users adding `path_patterns` to an existing global credential could be surprised that non-matching paths are now effectively blocked from carrying any auth.

src/tools/wasm/wrapper.rs
```diff
@@ -229,6 +231,22 @@ impl StoreData {
...
 
+            // Check path patterns if any are specified
```
serrrfirat: **Low — Duplicate path-checking block with `channels/wasm/wrapper.rs`**

This 15-line block (URL parse → extract path → check `path_matches_prefix`) is copy-pasted verbatim in `src/channels/wasm/wrapper.rs` (line ~250). The only difference is the leading comment text. Future bug fixes (e.g., logging on URL parse failure, handling the empty-string fallback) would need to be applied in both places.

Consider extracting a shared helper alongside the existing `path_matches_prefix` in `src/secrets/types.rs`.

crates/ironclaw_skills/src/validation.rs
```diff
@@ -169,6 +169,26 @@ pub fn validate_credential_spec(spec: &SkillCredentialSpec) -> Vec<String> {
...
+                "credential '{}' path pattern '{}' must start with '/'",
+                spec.name, pattern
+            ));
+        }
+        if pattern.split('/').any(|seg| seg == "..") {
+            errors.push(format!(
+                "credential '{}' path pattern '{}' must not contain '..' segments",
+                spec.name, pattern
+            ));
+        }
```
serrrfirat: **Low — Validation doesn't reject `?` or `#` in path_patterns**

Since the matching side receives paths from `url::Url::parse().path()` which strips query strings and fragments, a pattern like `"/api/v1?key=value"` or `"/api/v1#section"` would silently never match anything. Adding a check here would surface the mistake at skill-load time.

Separately, `CredentialMappingSchema` in `capabilities_schema.rs` passes `path_patterns` through without any validation at all (no leading `/` check, no `..` check). WASM tool authors wouldn't get feedback on invalid patterns — they'd just silently fail to match.

src/tools/builtin/http.rs
```diff
@@ -500,9 +508,13 @@ impl Tool for HttpTool {
...
+        // unauthenticated; that's intended.
         if let Some(registry) = self.credential_registry.as_ref() {
```
serrrfirat, line 551: **Medium — No test for the authentication gap on non-matching paths**

The split between host-level header blocking and path-level injection creates a gap: when a host has path-scoped credentials, requests to *non-matching* paths get LLM-provided auth headers stripped (host check passes) but receive no credential injection (path check fails). The result is an unauthenticated request.

This comment correctly documents the intent, but there's no test proving the combined behavior works. Adding one would prevent regressions if either the blocking or injection side changes independently:

```rust
#[tokio::test]
async fn test_path_scoped_cred_strips_headers_on_non_matching_path() {
    // Register credential for api.example.com with path_patterns=["/api/v1"]
    // Send request to /public/data with LLM-provided Authorization header
    // Assert: Authorization header is STRIPPED (host has credentials)
    // Assert: no credential injected (path doesn't match)
    // Net result: request goes out unauthenticated
}
```

Users adding `path_patterns` to an existing global credential could be surprised that non-matching paths on the same host are now effectively blocked from carrying any auth.

src/secrets/types.rs
```diff
@@ -253,11 +285,60 @@ impl CredentialMapping {
...
             host_patterns: vec![host_pattern.into()],
+            path_patterns: Vec::new(),
             optional: false,
         }
     }
 }
 
+/// Match `path` against a literal prefix with segment-boundary enforcement.
+/// After the prefix, the path must end or continue with `/` or `?`. A `..`
+/// that appears as a complete path segment is rejected; `..` inside a
```
serrrfirat: **Medium Severity** — Percent-encoded `...` traversal bypass in `path_matches_prefix`

`path_matches_prefix` checks for literal `..` path segments:
```rust
if path.split('/').any(|seg| seg == "..") {
    return false;
}
```

However, `%2e%2e` (or mixed-case `%2E%2e`) in a URL path bypasses this check. While `url::Url::parse` resolves literal `../` during parsing (so `https://host/api/../admin` becomes `https://host/admin`), it **preserves** percent-encoded sequences — `https://host/api/%2e%2e/admin` yields path `/api/%2e%2e/admin`.

If a backend server decodes `%2e%2e` as `..` (some servers do this, notably older IIS, Tomcat with certain configs, and some reverse proxies), a credential scoped to `/api/read` could be sent to a path the server resolves as `/api/write`.

**Suggested fix** — normalize before checking:
```rust
use percent_encoding::percent_decode_str;

pub(crate) fn path_matches_prefix(path: &str, prefix: &str) -> bool {
    // Decode percent-encoded sequences before traversal check
    let decoded = percent_decode_str(path).decode_utf8_lossy();
    if decoded.split('/').any(|seg| seg == "..") {
        return false;
    }
    // ... rest of function operates on original `path` (not decoded)
    // to preserve exact-match semantics against prefix
```

Alternatively, reject paths containing `%2e` (case-insensitive) as a lighter-weight check:
```rust
if path.to_ascii_lowercase().contains("%2e") {
    return false;
}
```

Confidence: [Possible] — requires a backend server that decodes percent-encoded path segments, which is non-standard but not rare.

src/tools/builtin/http.rs
```diff
@@ -606,8 +618,9 @@ impl Tool for HttpTool {
             self.secrets_store.as_ref(),
         ) {
             let cred_host = parsed_url.host_str().unwrap_or("").to_string();
+            let cred_path = parsed_url.path();
```
serrrfirat, line 660: **Medium Severity** — No test through `HttpTool::execute()` with path-scoped credentials

Per IronClaw's "Test Through the Caller, Not Just the Helper" rule (`.claude/rules/testing.md`), the HTTP tool's `execute()` is the caller that assembles URL parsing → `find_for_url` → credential injection into a single pipeline. Individual pieces are well-tested (`path_matches_prefix`, `CredentialMapping::matches`, WASM `inject_host_credentials`, `CredentialInjector::inject_for_url`), but no test drives `execute()` with a path-scoped `CredentialMapping` to verify the assembled flow.

The existing spy-interceptor tests are the right pattern (e.g., `http_tool_interceptor_does_not_see_injected_query_param_credentials`). A similar test should:

1. Register a credential with `path_patterns: ["/api/v1"]` on `api.github.com`
2. Call `execute()` with URL `https://api.github.com/api/v1/users` — verify the spy sees injection
3. Call `execute()` with URL `https://api.github.com/other` — verify no injection
4. Call `execute()` with URL `https://api.github.com/api/v1-malicious` — verify no injection (segment boundary)

This covers the gap between the helper tests and the full request pipeline, ensuring `cred_path = parsed_url.path()` and `registry.find_for_url(&cred_host, cred_path)` are wired correctly.

Confidence: [Likely] — the individual components are correct, but a wiring bug between them would only be caught by a caller-level test.

src/auth/extension.rs
```diff
@@ -173,7 +173,8 @@ impl AuthManager {
             }
         };
 
-        let matched = credential_registry.find_for_host(&host);
+        let path = extract_path_from_params(parameters).unwrap_or_else(|| "/".to_string());
```
serrrfirat, line 282: **Medium Severity** — `extract_path_from_params` has no unit test; auth manager path-aware flow untested

The new `extract_path_from_params` helper is `pub`-exported and used here in auth pre-flight, but has zero test coverage. Edge cases to verify:

- Missing `url` field → `None`
- URL with no explicit path (e.g., `https://host`) → `Some("/")`
- URL with query string (e.g., `https://host/path?q=1`) → `Some("/path")` (no query)
- Malformed URL → `None`

More importantly, the combination of `extract_path_from_params` + the `unwrap_or_else(|| "/".to_string())` fallback + `find_for_url` is the new path-aware auth pre-flight pipeline. The fallback to "/" means a missing/unparseable URL causes `find_for_url(host, "/")` which silently drops path-scoped credentials — correct behavior, but worth an explicit test to lock it in as intentional.

```rust
#[test]
fn test_extract_path_from_params_missing_url() {
    let params = serde_json::json!({"method": "GET"});
    assert_eq!(extract_path_from_params(&params), None);
}

#[test]
fn test_extract_path_from_params_with_query() {
    let params = serde_json::json!({"url": "https://host/api/v1?page=1"});
    assert_eq!(extract_path_from_params(&params), Some("/api/v1".to_string()));
}
```

Confidence: [Likely] — the implementation looks correct, but untested helpers in security-sensitive paths are a reliability risk.

src/secrets/types.rs
pranavraja99: Fixed in 79462882 — `path_matches_prefix` now rejects any `%2e`/`%2E` sequence in the input (`contains_percent_encoded_dot`). Regression test: `path_matches_prefix_rejects_percent_encoded_dot` covers lowercase, uppercase, mixed, single `%2e`, and embedded forms.

src/sandbox/proxy/policy.rs
pranavraja99: Fixed in 79462882 — `find_credential(host, path)` now uses `CredentialMapping::matches(host, path)` and `decide()` passes `request.path`. Regression test: `test_sandbox_proxy_honors_path_patterns` proves write-scoped creds don't leak onto read endpoints.

src/tools/wasm/wrapper.rs
pranavraja99: Ack'd as pre-existing (you called this out too — "arguably a pre-existing design limitation"). Not resolving in this PR: a proper fix needs specificity scoring + a docs change on ordering, which is out of scope for path-scoping. Tracked for a follow-up.

src/channels/wasm/wrapper.rs
pranavraja99: Fixed in 79462882 — extracted `extract_url_path_for_matching` into `src/secrets/types.rs` which logs `tracing::debug!(url, error, "URL parse failed during path-scoped credential check; skipping injection")` on parse failure. Both wrappers now call it (also closes #3115277341).

src/tools/builtin/http.rs
pranavraja99: Fixed in 79462882 — added 5 unit tests in `tools::builtin::http::tests`: `test_extract_path_from_params_valid`, `_missing_url`, `_strips_query_and_fragment`, `_root_when_no_path`, `_malformed_url`.

src/tools/builtin/http.rs
pranavraja99: Fixed in 79462882 — `tests/skill_credential_injection.rs::test_path_scoped_credential_strips_headers_on_non_matching_path` drives `HttpTool::execute()` with a path-scoped credential on a non-matching path and asserts the strip branch fires (NotAuthorized on LLM-supplied Authorization). Same helper `test_path_scoped_credential_wrong_path_is_segment_boundary_rejected` covers the v1-vs-v1-malicious attack through `execute()`.

src/tools/wasm/wrapper.rs
pranavraja99: Fixed in 79462882 — extracted `extract_url_path_for_matching` into `src/secrets/types.rs` (with tracing::debug! on parse failure). Both `tools/wasm/wrapper.rs` and `channels/wasm/wrapper.rs` now call the shared helper; duplicated block is gone.

crates/ironclaw_skills/src/validation.rs
pranavraja99: Fixed in 79462882 — factored `validate_path_pattern` out of `validate_credential_spec` and added checks for `?` and `#`. Also plumbed the same helper into `CredentialMappingSchema::validate()` (`src/tools/wasm/capabilities_schema.rs`) so WASM manifest authors now get warnings on malformed patterns at load time. Tests: `test_validate_credential_spec_path_pattern_rejects_query_string` and `_rejects_fragment`.

src/tools/builtin/http.rs
pranavraja99: Fixed in 79462882 — see #3115277336. `test_path_scoped_credential_strips_headers_on_non_matching_path` in `tests/skill_credential_injection.rs` drives `execute()` and asserts the combined behavior.

src/secrets/types.rs
pranavraja99: Fixed in 79462882 — same fix as #3115205707. `path_matches_prefix` scans for any `%2e`/`%2E` sequence (via `contains_percent_encoded_dot`) and returns `false` before the prefix check. Covers lowercase, uppercase, and mixed case.

src/tools/builtin/http.rs
pranavraja99: Fixed in 79462882 — `tests/skill_credential_injection.rs::test_path_scoped_credential_wrong_path_is_segment_boundary_rejected` drives `HttpTool::execute()` with `/api/v1-malicious` against a `/api/v1`-scoped credential on `api.github.com` (DNS-resolvable, no network I/O since the auth-strip branch fires first). The pipeline URL-parse → `find_for_url` → `path_matches_prefix` is exercised end-to-end.

src/auth/extension.rs
pranavraja99: Fixed in 79462882 — 5 unit tests for `extract_path_from_params` in `tools::builtin::http::tests` (valid, missing_url → None, query+fragment stripped, bare host → `/`, malformed → None). The auth-manager pre-flight pipeline is additionally covered by `test_path_scoped_credential_strips_headers_on_non_matching_path`.

pranavraja99:
@serrrfirat — round-3 findings addressed in 79462882:

**Security (2 legitimate misses — thanks for catching these):**
- **%2e / %2E path-traversal bypass** (#3115205707, #3117470178) — `path_matches_prefix` now rejects any percent-encoded dot. Test: `path_matches_prefix_rejects_percent_encoded_dot`.
- **Sandbox proxy `find_credential` ignored `path_patterns`** (#3115206118) — now uses `CredentialMapping::matches(host, path)` via `request.path`. Test: `test_sandbox_proxy_honors_path_patterns`.

**Validation:**
- **`?` / `#` in path_patterns** (#3115277345) — factored `validate_path_pattern` out; `?` and `#` now rejected at skill-load time. WASM capabilities loader (`CredentialMappingSchema::validate`) runs the same check and warns on bad patterns. Tests: `_rejects_query_string`, `_rejects_fragment`.

**Coverage:**
- **Caller-level `HttpTool::execute()` tests** (#3117470891, #3115277336, #3115338316) — `test_path_scoped_credential_wrong_path_is_segment_boundary_rejected` (v1 vs v1-malicious) and `test_path_scoped_credential_strips_headers_on_non_matching_path` (host-scoped strip + path-scoped inject gap).
- **`extract_path_from_params` tests** (#3115220667, #3117471784) — 5 unit tests.

**DRY / quality:**
- **Duplicated URL-path block** (#3115277341, #3115220466) — extracted `extract_url_path_for_matching` in `src/secrets/types.rs` with `tracing::debug!` on parse failure; both WASM wrappers now call it.

**Acknowledged pre-existing:**
- **Overlapping global + path-scoped last-write-wins** (#3115206590) — you flagged this as pre-existing; leaving for a follow-up since a proper fix needs specificity scoring + ordering docs.

Zero clippy warnings across all crates, 160 skills-crate tests green, 15 `skill_credential_injection` integration tests green. PTAL.

src/tools/wasm/wrapper.rs
```diff
@@ -229,6 +231,19 @@ impl StoreData {
...
+                    .path_patterns
+                    .iter()
+                    .any(|prefix| path_matches_prefix(&url_path, prefix));
+                if !path_match {
+                    continue;
+                }
+            }
+
             // Merge injected headers (host credentials take precedence)
             for (key, value) in &cred.headers {
```
serrrfirat, line 264: **High Severity**\n\nOverlapping path-scoped credential mappings have no precedence rule, so the winning injected header/query value is effectively last-write-wins. `CredentialMapping::matches()` allows multiple prefixes to match the same request (`/api/v1` and `/api/v1/admin` both match `/api/v1/admin/users`), and the injectors then merge every match by blindly inserting headers/query params. In the WASM capability path this is worse because the source collection is a `HashMap` (`HttpCapability.credentials`) iterated via `.values()`, so the winner is not a stable manifest-order contract. A request can end up sending the wrong bearer token to the more privileged endpoint.\n\nConcrete fix: reject overlapping mappings that inject the same target (e.g. same header/query param), or sort matches by a defined precedence such as longest path prefix first and then fail on conflicting duplicates instead of silently overriding.

src/auth/extension.rs
```diff
@@ -173,7 +173,8 @@ impl AuthManager {
...
+        let matched = credential_registry.find_for_url(&host, &path);
```
serrrfirat, line 283: **High Severity**\n\nThis path-aware lookup makes multi-match cases more common, but the auth flow still treats "any one resolved credential" as success. In `check_http_auth()` the function returns `Ready` as soon as a single matched mapping resolves, and in `HttpTool::execute()` `missing_credential` is cleared after any successful injection. That means an endpoint that legitimately requires two credentials on the same path (for example a bearer token plus an org/account header) will never surface the missing-auth gate once one of them is configured: the user just gets a raw 401/403 with no remediation path.\n\nConcrete fix: evaluate all matched required mappings conjunctively for the requested path. If any required mapping is still unresolved, keep that fact through execution and raise the auth gate on 401/403 instead of discarding it after the first successful injection.

src/secrets/types.rs
```diff
@@ -253,11 +285,105 @@ impl CredentialMapping {
...
+/// Percent-encoded dot sequences (`%2e`, `%2E`) are rejected outright:
+/// `url::Url::parse` preserves percent-encoding, but downstream servers
+/// (IIS, Tomcat, some reverse proxies) may decode `%2e%2e` to `..` before
+/// routing. Refusing to match when any `%2e` appears keeps the predicate
+/// aligned with the effective path the server will see.
+pub(crate) fn path_matches_prefix(path: &str, prefix: &str) -> bool {
+    if path.split('/').any(|seg| seg == "..") {
+        return false;
+    }
+    if contains_percent_encoded_dot(path) {
```
serrrfirat: **Medium Severity**\n\n`contains_percent_encoded_dot()` now rejects *any* `%2e` anywhere in the path, not just segment-normalizing traversal like `%2e%2e` or `/%2e/`. That means perfectly legitimate scoped endpoints such as `/files/foo%2ebar` or `/releases/v1%2e2` will never match their credential mapping. On a credentialed host the manual `Authorization` header path is also blocked, so those requests become impossible to authenticate even though the encoded dot is part of a normal literal segment, not traversal.\n\nConcrete fix: normalize and reject only dot-segments (`.` / `..`) after percent-decoding each segment, rather than banning every `%2e` substring globally.

src/tools/wasm/wrapper.rs
pranavraja99: Fixed in a99d0d8c — retracting my previous "pre-existing, out of scope" punt; you were right to escalate.

- Added `match_specificity(path_patterns, req_path)` in `src/secrets/types.rs` (longest matching prefix wins, 0 for unscoped).
- `SharedCredentialRegistry::find_for_url` and `CredentialInjector::find_credentials_for_url` now sort matches by ascending specificity, tie-broken alphabetically on `secret_name`. The HashMap-backed `CredentialInjector` is now deterministic regardless of manifest load order.
- Both WASM wrappers' `inject_host_credentials` apply the same sort before the last-write-wins header merge. `ResolvedHostCredential` gained a `secret_name` field purely for tie-breaks (no secret material exposed; the hand-rolled redacting `Debug` still stands).

Regression test `test_inject_host_credentials_most_specific_path_wins` constructs the same credentials in both orders and asserts the specific one always wins — locks in order-independence.

src/auth/extension.rs
pranavraja99: Fixed in a99d0d8c — `check_http_auth` now evaluates every required (non-optional) matched mapping conjunctively. Removed the "return Ready on first resolved" short-circuit; missing creds accumulate and return `MissingCredentials`. `HttpTool::execute` made the parallel change: no longer clears `missing_credential` on peer success, no longer gated on `!injected_any_credential`. Optional mappings skip the gate as before.

Regression tests: `check_http_conjunctive_auth_any_missing_required_raises_gate` (bearer configured, org_header missing → surfaces gate) and `check_http_conjunctive_auth_all_required_resolved_is_ready` (both configured → Ready).

src/secrets/types.rs
pranavraja99: Fixed in a99d0d8c — fair catch; the earlier rule was too blunt. `path_matches_prefix` now decodes each path segment (RFC 3986 octet-level percent decoding) and rejects only segments that decode to `.` or `..`. Legitimate literal paths with embedded encoded dots are allowed again:

- `/files/foo%2ebar` → segment decodes to `foo.bar` → allowed
- `/releases/v1%2e2` → segment decodes to `v1.2` → allowed
- `/api/v1/%2e%2e/admin` → segment decodes to `..` → rejected
- `/api/v1/.%2e/admin` and `/api/v1/%2e./admin` → decode to `..` → rejected (new cases)

Regression tests: `path_matches_prefix_allows_legit_embedded_encoded_dot` for the false-rejection cases and `path_matches_prefix_rejects_percent_encoded_dot_segments` for the traversal cases, including the mixed literal+encoded forms.

pranavraja99:
@serrrfirat — round-4 findings addressed in a99d0d8c:

- **#3125963270 (High) overlapping mappings with no precedence** — retracting my round-3 "pre-existing, out of scope" punt; your HashMap-iteration nondeterminism point was load-bearing. Added `match_specificity` helper; `SharedCredentialRegistry::find_for_url`, `CredentialInjector::find_credentials_for_url`, and both WASM `inject_host_credentials` now sort matches by ascending path specificity (tie-break on `secret_name`) before merging. The most-specific mapping wins any header conflict, order-independently.
- **#3125963977 (High) non-conjunctive auth evaluation** — `check_http_auth` and `HttpTool::execute` now require every non-optional matched mapping to resolve. Optional mappings skip the gate. An endpoint needing bearer + org-header surfaces `MissingCredentials` instead of failing at the wire.
- **#3125964627 (Medium) `%2e` over-rejection** — `path_matches_prefix` now decodes each segment and only rejects segments whose decoded form is `.` or `..`. Literal paths like `/files/foo%2ebar` and `/releases/v1%2e2` match again. Traversal variants `.%2e`, `%2e.`, and mixed-case `%2E` still rejected.

Regression tests for each: `test_inject_host_credentials_most_specific_path_wins` (order-independent), `check_http_conjunctive_auth_any_missing_required_raises_gate` + `check_http_conjunctive_auth_all_required_resolved_is_ready`, `path_matches_prefix_allows_legit_embedded_encoded_dot` + `path_matches_prefix_rejects_percent_encoded_dot_segments`.

Zero clippy warnings across all crates, 272 tests passing across the affected modules, 15 `skill_credential_injection` integration tests green. PTAL.

src/tools/wasm/capabilities_schema.rs
```diff
@@ -204,6 +204,22 @@ impl CapabilitiesFile {
...
 
+        // Validate credential path_patterns using the same rules as skills
+        // (leading slash, no literal `..` segments, no `?`/`#`). Emit as
+        // warnings so a bad WASM manifest doesn't block other tools from
+        // loading; the effect is that the malformed pattern will never match.
+        if let Some(http) = &self.http {
+            for mapping in http.credentials.values() {
+                for pattern in &mapping.path_patterns {
+                    for err in
+                        ironclaw_skills::validate_path_pattern(&mapping.secret_name, pattern)
```
serrrfirat: **Medium Severity**\n\n`CapabilitiesFile::validate()` only warns on invalid `path_patterns`, but this is not actually safe for the empty-string case. `validate_path_pattern()` reports `""` as invalid, yet the mapping is still loaded and `path_matches_prefix(path, "")` returns true for every absolute path, so `path_patterns: [""]` effectively widens the credential back to global scope instead of "never matching" as the comment suggests.\n\nThat makes WASM capabilities files behave differently from skill credentials, where invalid path patterns are skipped entirely.\n\nPlease reject or drop invalid credential mappings here (not warning-only), and add a test covering `path_patterns: [""]` so it cannot silently broaden injection scope.

src/secrets/types.rs
```diff
@@ -253,11 +285,144 @@ impl CredentialMapping {
...
+/// decoding of octets) to the traversal segments `.` or `..`. Segments with
+/// embedded encoded dots that are part of a legitimate literal (e.g.
+/// `foo%2ebar`, `v1%2e2`) are allowed because the decoded form is not a
+/// pure dot-segment.
+fn path_has_dot_segment(path: &str) -> bool {
+    path.split('/').any(is_dot_segment)
+}
+
+fn is_dot_segment(segment: &str) -> bool {
+    let decoded = percent_decode_bytes(segment.as_bytes());
```
serrrfirat, line 373: **Medium Severity**\n\nThe traversal guard only rejects decoded segments that are exactly `.` or `..`. A path like `/api/v1/%2e%2e%2fadmin` is treated as a single segment, percent-decodes to `../admin`, and therefore passes `is_dot_segment()` even though some backends normalize encoded slashes before routing. In that deployment shape, a credential scoped to `/api/v1` can still be sent to a request that the upstream server resolves outside that prefix.\n\nPlease add a regression test for `%2f` / encoded-slash smuggling and tighten the matcher so percent-decoded segments containing path separators cannot satisfy a scoped prefix.

src/sandbox/proxy/policy.rs
```diff
@@ -105,14 +105,13 @@ impl DefaultPolicyDecider {
...
-        let host_lower = host.to_lowercase();
-        self.credential_mappings.iter().find(|m| {
-            m.host_patterns
-                .iter()
-                .any(|pattern| host_matches_pattern(&host_lower, pattern))
-        })
+    /// Find the first credential mapping that matches both `host` and `path`.
+    /// Honors `path_patterns` so a mapping scoped to `/api/v1/write` does not
+    /// inject on `/api/v1/read` when the sandbox proxy forwards the request.
+    fn find_credential(&self, host: &str, path: &str) -> Option<&CredentialMapping> {
```
serrrfirat, line 115: **High Severity**\n\nThe sandbox proxy still picks the first matching credential (`iter().find(...)`) instead of the most specific one. That means a host with both a global mapping and a narrower path-scoped mapping can inject the wrong secret purely based on vector order. In this PR the HTTP tool and both WASM wrappers sort by `match_specificity(...)`, but the sandbox path does not, so the same credential set can behave differently depending on execution path.\n\nConcrete case: if `api.example.com` has a global token plus a `/api/v1/write` token, a write request can still receive the global token when that mapping appears first.\n\nPlease make the sandbox path use the same specificity ordering as `SharedCredentialRegistry::find_for_url()` / the WASM wrappers (or delegate to a shared helper) and add a regression test with overlapping global + scoped mappings.

src/tools/wasm/capabilities_schema.rs
pranavraja99: Fixed in c9b41a61 — `CredentialMappingSchema::to_credential_mapping` now returns `Option`, and `to_http_capability` `filter_map`s invalid mappings out of the runtime capabilities. An invalid `path_patterns` entry (empty string, missing leading `/`, `..`, `?`, `#`) drops the entire mapping with a warning rather than loading a silently-broadened credential. Matches the skills pipeline.

Regression tests: `test_invalid_path_pattern_drops_mapping_not_widens_scope` (empty string case), `test_invalid_path_pattern_missing_leading_slash_drops_mapping`, `test_valid_path_pattern_preserved`.

src/secrets/types.rs
pranavraja99: Fixed in c9b41a61 — `is_traversal_segment` (formerly `is_dot_segment`) now rejects a segment whose decoded form contains a `/`, not just segments that equal `.` or `..`. So `%2e%2e%2fadmin` (decodes to `../admin`), `%2f` in any position, and the uppercase variants all fail to match. The rationale in the doc comment spells out the Tomcat/IIS/reverse-proxy decoder policies this guards against. Literal dots in legitimate segments (`foo%2ebar`, `v1%2e2`) still match since they decode to plain literals without an embedded slash.

Regression test: `path_matches_prefix_rejects_percent_encoded_slash_smuggling` covers `%2e%2e%2f` and bare `%2f` in lower/upper/mixed case.

src/sandbox/proxy/policy.rs
pranavraja99: Fixed in c9b41a61 — sandbox `find_credential` now uses `max_by(|a, b| spec_a.cmp(&spec_b).then_with(|| a.secret_name.cmp(&b.secret_name)))`, matching the precedence rule used by `SharedCredentialRegistry::find_for_url` and both WASM `inject_host_credentials`. Behavior is consistent across all four paths: longest matching path prefix wins, ties tie-broken alphabetically on `secret_name`.

Regression test: `test_sandbox_proxy_most_specific_credential_wins` registers the same credentials in both insertion orders (global-first and specific-first) and asserts WRITE_TOKEN wins on the write path regardless.

pranavraja99:
@serrrfirat — round-5 findings addressed in c9b41a61:

- **#3126256040 (Medium)** empty `path_patterns` silently widening scope — `CredentialMappingSchema::to_credential_mapping` now returns `Option` and the conversion drops the whole mapping when any pattern is invalid. Matches the skills pipeline. Duplicate warn-loop in `validate()` removed.
- **#3126256056 (Medium)** `%2e%2e%2f` encoded-slash smuggling — per-segment check now also rejects decoded forms containing `/`. Legitimate encoded dots (`foo%2ebar`, `v1%2e2`) still match.
- **#3126256060 (High)** sandbox proxy used `.find` not specificity-sorted `max_by` — now consistent with the HTTP tool and both WASM wrappers.

Regression tests for each: `test_invalid_path_pattern_{drops_mapping_not_widens_scope,missing_leading_slash_drops_mapping,valid_path_pattern_preserved}`; `path_matches_prefix_rejects_percent_encoded_slash_smuggling`; `test_sandbox_proxy_most_specific_credential_wins` (order-independent).

Zero clippy warnings, all targeted module tests + 15 `skill_credential_injection` integration tests green. PTAL.

---

**Unrelated operational note for repo maintainers:** commit `50f6771f` on this branch inadvertently included two local scratch files (`tests_all/source_env_vars.sh`, `integrations/abound/tests/source_env_vars.sh`) containing production Railway `ADMIN_TOKEN` and Abound tokens. `b93cc2fb` removes them from tracking and adds `.gitignore` patterns, but the secrets are visible in the `50f6771f` commit object. Rotating those tokens is recommended.

serrrfirat (approved):

PR #2173, opened 2026-04-08
feat(tools): config-driven credential loading via INTEGRATION_CREDENTIALS_DIR

pranavraja99:
## Summary
New `INTEGRATION_CREDENTIALS_DIR` env var that scans a directory for `*.json` credential mapping files at startup. Each JSON file declares secret names, injection locations (bearer/header/query), and host/path patterns.

Depends on #2168 (path-based credential matching).

## Example
```json
{"mappings": [{"secret_name": "my_token", "location": {"type": "bearer"}, "host_patterns": ["api.example.com"], "path_patterns": ["/v1/read"]}]}
```

## Test plan
- Set `INTEGRATION_CREDENTIALS_DIR=/path/to/dir` with a JSON file, verify credentials auto-inject on matching requests

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -342,6 +342,34 @@ impl AppBuilder {
         tools.register_builtin_tools();
         tools.register_tool_info();
 
+        // Load integration credential mappings from a directory.
+        // Scans for *.json files in the directory and one level of subdirectories
+        // (e.g. integrations/abound/credentials.json).
+        if let Ok(dir) = std::env::var("INTEGRATION_CREDENTIALS_DIR") {
+            let dir_path = std::path::Path::new(&dir);
+            if dir_path.is_dir() {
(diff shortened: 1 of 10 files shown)


serrrfirat:

This directory-driven credential injection system belongs to the v1 tool registry and injection code. Reborn extensions declare runtime credential requirements and audiences in manifests; the host resolves encrypted credentials and injects them only at authorized egress boundaries.

If operator-managed credential bundles are needed, add a typed Reborn import/configuration layer that produces validated manifest-compatible credential bindings without exposing raw values to tools, prompts, or generic environment variables.

serrrfirat:

This credential-directory system belongs to v1 registry/injection code. Reborn manifests declare credentials and audiences; the host resolves encrypted values only at authorized egress.

An operator bundle feature should import validated manifest-compatible bindings without exposing raw values to tools, prompts, or generic environment variables.

PR #2179, opened 2026-04-09
fix(test): make CI tests resilient to sandboxed/offline environments

zmanian:
## Summary

- **config/helpers**: Add 10s DNS resolution timeout to `validate_base_url` to prevent indefinite hangs; guard `invalid_tld_resolves_locally()` with channel timeout
- **config/llm**: Only validate explicitly-configured URLs (env vars/DB settings), not hardcoded registry defaults — prevents DNS failures for known-good URLs like `api.openai.com` in offline environments; use localhost URLs in `openai_compatible` model resolution tests
- **cli/doctor**: Accept DNS failure as valid outcome in offline doctor tests
- **channels/webhook_server**: Fix bind test for root/container environments by falling back to already-occupied port when privileged port bind succeeds
- **tools/mcp/auth**: Skip `validate_url_safe` HTTPS test when DNS is unavailable
- **tunnel/custom**: Check response status code in `health_check` (`is_ok_and(|r| r.status().is_success())`), not just successful send — a proxy may return non-2xx for unreachable targets

**Root cause:** Commit `315c4cf8` added `validate_base_url` calls that perform DNS resolution at config construction time for ALL URLs including hardcoded defaults. In sandboxed/offline CI environments, DNS for external hostnames fails, breaking 10 tests.

## Test plan

- [x] All 4339 lib tests pass (0 failed, 3 ignored)
- [x] Verified `openai_codex_rejects_ssrf_api_url` still properly rejects malicious user-configured URLs
- [x] Verified SSRF protection is preserved for all explicitly-configured URLs

https://claude.ai/code/session_01LyzwS5oHA68ARhDuXqazpn

### DIFF
--- a/src/config/helpers.rs
+++ b/src/config/helpers.rs
@@ -329,17 +329,35 @@ fn validate_base_url_with_policy(
         let port = parsed
             .port()
             .unwrap_or(if scheme == "http" { 80 } else { 443 });
-        // `to_socket_addrs` performs blocking DNS resolution. This helper is
-        // also called from async request handlers (e.g. the LLM utility
+        // `to_socket_addrs` performs blocking DNS resolution.  This helper
+        // is also called from async request handlers (e.g. the LLM utility
         // routes), so wrap the lookup in `block_in_place` when running on a
-        // multi-threaded tokio worker to avoid stalling other tasks. The
+        // multi-threaded tokio worker to avoid stalling other tasks.  The
         // `try_current()` check keeps sync callers (config bootstrap, CLI)
         // working unchanged.
-        let resolve = || -> std::io::Result<Vec<IpAddr>> {
-            Ok((host, port)
-                .to_socket_addrs()?
-                .map(|addr| addr.ip())
-                .collect())
+        //
+        // A 10-second timeout prevents the entire process from hanging when
+        // the DNS resolver blocks indefinitely (e.g. sandboxed environments
+        // or broken resolvers).
+        let host_owned = host.to_string();
+        let resolve = move || -> std::io::Result<Vec<IpAddr>> {
+            use std::sync::mpsc;
+            use std::time::Duration;
+            let (tx, rx) = mpsc::channel();
+            let h = host_owned.clone();
+            std::thread::spawn(move || {
+                let result = (h.as_str(), port)
+                    .to_socket_addrs()
+                    .map(|addrs| addrs.map(|a| a.ip()).collect::<Vec<_>>());
+                let _ = tx.send(result);
+            });
+            rx.recv_timeout(Duration::from_secs(10))
+                .unwrap_or_else(|_| {
+                    Err(std::io::Error::new(
+                        std::io::ErrorKind::TimedOut,
+                        "DNS resolution timed out after 10s",
+                    ))
+                })
         };
         let lookup = match tokio::runtime::Handle::try_current() {
             Ok(handle) if handle.runtime_flavor() == tokio::runtime::RuntimeFlavor::MultiThread => {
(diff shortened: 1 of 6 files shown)


src/config/llm.rs
```diff
@@ -163,6 +171,9 @@ impl LlmConfig {
         } else {
             crate::llm::DEFAULT_MODEL.to_string()
         };
+        let nearai_base_url_explicit = nearai_override
+            .and_then(|o| o.base_url.clone())
+            .or_else(|| optional_env("NEARAI_BASE_URL").ok().flatten());
```
serrrfirat, line 176: **🟡 Medium: `.ok().flatten()` silently swallows env-var parse errors**

`nearai_base_url_explicit` uses `.ok().flatten()` on `optional_env("NEARAI_BASE_URL")`, silently discarding `ConfigError`. If `optional_env` fails, `nearai_base_url_explicit` becomes `None`, and `validate_base_url` is skipped. The error surfaces on the next `optional_env("NEARAI_BASE_URL")?` call, so behavior is correct in practice, but `.ok().flatten()` is fragile — a refactor removing the second call would silently lose the error.

**Suggested fix:** Use `optional_env("NEARAI_BASE_URL")?` for the explicit check too, or extract the result into a shared let-binding.

src/config/helpers.rs
```diff
@@ -329,17 +329,35 @@ fn validate_base_url_with_policy(
...
-                .to_socket_addrs()?
-                .map(|addr| addr.ip())
-                .collect())
+        //
+        // A 10-second timeout prevents the entire process from hanging when
+        // the DNS resolver blocks indefinitely (e.g. sandboxed environments
+        // or broken resolvers).
+        let host_owned = host.to_string();
+        let resolve = move || -> std::io::Result<Vec<IpAddr>> {
+            use std::sync::mpsc;
```
serrrfirat, line 344: **🟡 Medium: DNS timeout thread leaks ~8MB stack if resolution hangs**

When the 10s DNS timeout fires, the spawned `std::thread` continues running indefinitely (blocked in `to_socket_addrs()`). In pathological DNS environments, repeated config-load attempts could accumulate zombie threads.

Acceptable for config-time (bounded number of calls at startup), but should be documented.

**Suggested fix:** Document as a known limitation. Consider `std::thread::Builder::new().name("dns-timeout".into()).spawn(...)` for debuggability.

src/config/llm.rs
```diff
@@ -533,10 +559,11 @@ impl LlmConfig {
...
 
-        // Provider base URLs are explicit operator configuration, so allow
-        // private/local endpoints while still rejecting unsafe schemes,
-        // public plaintext HTTP, and special blocked addresses.
-        if !base_url.is_empty() {
+        // Only validate explicitly-configured URLs (from env vars or DB
+        // settings).  Registry-provided defaults are hardcoded known-good
+        // URLs that don't need DNS-based SSRF validation — validating them
+        // causes spurious failures in offline / sandboxed environments.
+        if explicit_base_url.is_some() && !base_url.is_empty() {
```
serrrfirat, line 566: **🟡 Medium: Skipping SSRF validation for registry defaults assumes trusted registry**

Skipping `validate_base_url` for registry-default URLs is safe **only** if the registry content is trusted. If `ProviderRegistry` is ever loaded from untrusted sources (user-editable config, remote fetches), an attacker could inject a malicious `default_base_url` that bypasses SSRF validation entirely.

**Suggested fix:** Add a comment documenting the trust assumption: "Registry defaults are compile-time constants from `src/llm/providers/*.json` — if registry loading changes to accept external sources, this must be revisited."

serrrfirat:
## 🏗️ Paranoid Architect Review — PR #2179

**Verdict: ✅ APPROVE with medium fixes**

### Summary
| Severity | Count |
|----------|-------|
| Medium | 3 |

### Assessment
Well-motivated fix for CI failures in sandboxed/offline environments. The core approach — skipping DNS validation for hardcoded default URLs — is correct. The DNS timeout addition is good defense-in-depth. The `custom.rs` tunnel health check fix is a genuine bug fix.

Three medium concerns:
1. `.ok().flatten()` pattern silently swallows parse errors (fragile, could mask SSRF bypass)
2. DNS timeout threads leak stack memory if resolution hangs indefinitely
3. Registry default trust assumption should be documented

**Note:** This appears to be a duplicate of #2183. Consider closing one.

zmanian:

PR #2190, opened 2026-04-09
ci: trigger ironclaw-dind image build

think-in-universe:
## Summary

- Trigger ironclaw-dind image build after a new IronClaw docker image is published

## Change Type

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [x] CI/Infrastructure
- [ ] Security
- [ ] Dependencies

## Validation

- [ ] `cargo build`
- [ ] Relevant tests pass: 
- [ ] Manual testing: 

## Security Impact

None

## Database Impact

None

## Review Follow-Through

---

**Review track**:

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -40,6 +40,15 @@ jobs:
           ref: ${{ github.event_name == 'schedule' && 'staging' || '' }}
           persist-credentials: false
 
+      - name: Create releases-manager app token
+        id: app-token
+        uses: actions/create-github-app-token@fee1f7d63c2ff003460e3d139729b119787bc349 # v2
+        with:
+          app-id: ${{ secrets.GH_RELEASES_MANAGER_APP_ID }}
+          private-key: ${{ secrets.GH_RELEASES_MANAGER_APP_PRIVATE_KEY }}
+          owner: nearai
+          repositories: ironclaw-dind
+
       - name: Extract version from Cargo.toml
         id: version
         run: |
@@ -114,15 +123,6 @@ jobs:
           platforms: linux/amd64
           cache-from: type=gha,scope=worker
           cache-to: type=gha,mode=max,scope=worker
-
-      - name: Create releases-manager app token
-        id: app-token
-        uses: actions/create-github-app-token@fee1f7d63c2ff003460e3d139729b119787bc349 # v2
-        with:
-          app-id: ${{ secrets.GH_RELEASES_MANAGER_APP_ID }}
-          private-key: ${{ secrets.GH_RELEASES_MANAGER_APP_PRIVATE_KEY }}
-          owner: nearai
-          repositories: ironclaw-dind
 
       - name: Trigger ironclaw-dind Build & Push
         env:
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -117,6 +117,7 @@ jobs:
 
       - name: Create releases-manager app token
         id: app-token
+        continue-on-error: true
         uses: actions/create-github-app-token@fee1f7d63c2ff003460e3d139729b119787bc349 # v2
         with:
           app-id: ${{ secrets.GH_RELEASES_MANAGER_APP_ID }}
@@ -125,6 +126,8 @@ jobs:
           repositories: ironclaw-dind
 
       - name: Trigger ironclaw-dind Build & Push
+        if: steps.app-token.outcome == 'success'
+        continue-on-error: true
         env:
           GH_TOKEN: ${{ steps.app-token.outputs.token }}
           EVENT_NAME: ${{ github.event_name }}
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -115,6 +115,43 @@ jobs:
           cache-from: type=gha,scope=worker
           cache-to: type=gha,mode=max,scope=worker
 
+      - name: Create releases-manager app token
+        id: app-token
+        uses: actions/create-github-app-token@v1
+        with:
+          app-id: ${{ secrets.GH_RELEASES_MANAGER_APP_ID }}
+          private-key: ${{ secrets.GH_RELEASES_MANAGER_APP_PRIVATE_KEY }}
+          owner: nearai
+          repositories: ironclaw-dind
+
+      - name: Trigger ironclaw-dind Build & Push
+        env:
+          GH_TOKEN: ${{ steps.app-token.outputs.token }}
+          EVENT_NAME: ${{ github.event_name }}
+          VERSION: ${{ steps.version.outputs.version }}
+        run: |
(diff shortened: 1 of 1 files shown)


Evrard-Nil (dismissed):

.github/workflows/docker.yml
```diff
@@ -115,6 +115,32 @@ jobs:
...
+              -f event_type="ironclaw_image_published" \
+              -f client_payload[version]="${VERSION}"
+          else
+            gh api repos/nearai/ironclaw-dind/dispatches \
+              --method POST \
+              -f event_type="ironclaw_image_published"
+          fi
+
       - name: Summary
         run: |
```
serrrfirat, line 149: **Medium Severity — Missing `continue-on-error: true` on trigger steps**

If the GitHub App token creation or the `gh api` dispatch call fails (e.g., credentials rotated, transient API error, 403), the **entire workflow fails** even though the Docker images were already successfully pushed.

Since this workflow is invoked via `workflow_call` from `release.yml`, a failure here would cause the release pipeline to report failure despite images being published — potentially blocking the release process.

**Suggested fix:** Add `continue-on-error: true` to both the "Create releases-manager app token" and "Trigger ironclaw-dind Build & Push" steps:

```yaml
      - name: Create releases-manager app token
        id: app-token
        continue-on-error: true
        uses: actions/create-github-app-token@fee1f7d63c2ff003460e3d139729b119787bc349 # v2
        ...

      - name: Trigger ironclaw-dind Build & Push
        if: steps.app-token.outcome == 'success'
        continue-on-error: true
        ...
```

This way the trigger is best-effort and won't break the primary mission of this workflow (building and pushing images).

serrrfirat:
**Medium Severity — Summary step skipped on trigger failure** (`.github/workflows/docker.yml`, Summary step)

The existing "Summary" step has no `if: always()` condition. If either of the new upstream steps fails, the Summary step is skipped — meaning no step summary is written, making it harder to debug what images were actually pushed.

**Suggested fix:** Add `if: always()` to the Summary step:

```yaml
      - name: Summary
        if: always()
        run: |
          ...
```

Alternatively, if the other finding is addressed with `continue-on-error: true` on the trigger steps, this becomes a non-issue since those steps would no longer fail the job.

.github/workflows/docker.yml
think-in-universe: Thanks! Fixed in https://github.com/nearai/ironclaw/pull/2190/commits/63ef44be962c2b26d07a7f07b8ab5c02dbb751cb

think-in-universe:
Tested in https://github.com/nearai/ironclaw/actions/runs/24206775930. It works well.

.github/workflows/docker.yml
```diff
@@ -115,6 +115,35 @@ jobs:
...
+
+      - name: Trigger ironclaw-dind Build & Push
+        if: steps.app-token.outcome == 'success'
+        continue-on-error: true
+        env:
+          GH_TOKEN: ${{ steps.app-token.outputs.token }}
+          EVENT_NAME: ${{ github.event_name }}
+          VERSION: ${{ steps.version.outputs.version }}
+        run: |
+          if [[ "${EVENT_NAME}" == "workflow_call" && -n "${VERSION}" ]]; then
```
serrrfirat, line 137: **High Severity**\n\nThe repository dispatch only sends `client_payload.version` when `github.event_name == "workflow_call"`. I checked `nearai/ironclaw-dind/.github/workflows/build.yml`: on `repository_dispatch` it resolves the upstream image from `github.event.client_payload.version` when present, otherwise it falls back to `staging`. That means manual runs of this workflow can publish `sha-...` and custom tags here, but still trigger DinD to rebuild from `nearaidev/ironclaw:staging` instead of from the image that was just built.\n\nConcrete fix: compute one resolved downstream tag from the same tag-selection logic used above and always pass it in the dispatch payload, or explicitly gate this trigger so only release/staging builds dispatch to `ironclaw-dind`.

.github/workflows/docker.yml
```diff
@@ -115,6 +115,35 @@ jobs:
           cache-from: type=gha,scope=worker
           cache-to: type=gha,mode=max,scope=worker
 
+      - name: Create releases-manager app token
```
serrrfirat, line 118: **High Severity**\n\nBoth the GitHub App token step and the downstream dispatch step are marked `continue-on-error: true`. On release-triggered `workflow_call` runs, a missing secret, permission drift, or rejected `repository_dispatch` now leaves the Docker workflow green even though `ironclaw-dind` was never rebuilt. That makes the new cross-repo sync path silent best-effort automation instead of a reliable release invariant.\n\nConcrete fix: fail release-path runs when token creation or dispatch fails, or at minimum emit an explicit warning in the job summary and keep best-effort behavior only for non-release paths.

.github/workflows/docker.yml
```diff
@@ -115,6 +115,35 @@ jobs:
...
+        id: app-token
+        continue-on-error: true
+        uses: actions/create-github-app-token@fee1f7d63c2ff003460e3d139729b119787bc349 # v2
+        with:
+          app-id: ${{ secrets.GH_RELEASES_MANAGER_APP_ID }}
+          private-key: ${{ secrets.GH_RELEASES_MANAGER_APP_PRIVATE_KEY }}
+          owner: nearai
+          repositories: ironclaw-dind
+
+      - name: Trigger ironclaw-dind Build & Push
```
serrrfirat, line 128: **Medium Severity**\n\nThis adds a second trigger path for `ironclaw-dind` on top of its existing 6-hour scheduled rebuild flow. `ironclaw` already has a daily scheduled Docker build, and `ironclaw-dind` already performs its own digest-based skip check for scheduled runs. Dispatching here as well means the same staging image can be rebuilt twice around the scheduled window, while the new dispatch path does not benefit from the downstream schedule-only skip logic.\n\nConcrete fix: skip the cross-repo dispatch on `schedule`, or add the same digest/concurrency guard on the `repository_dispatch` path in `ironclaw-dind`.

.github/workflows/docker.yml
think-in-universe: This is contradictory with the previous comment: https://github.com/nearai/ironclaw/pull/2190#discussion_r3057557235

.github/workflows/docker.yml
think-in-universe: This is fine for now. ironclaw-dind already has the guard for duplicate source image digest.

.github/workflows/docker.yml
think-in-universe: This is expected. ironclaw-dind only builds for `staging` and version tags. https://github.com/nearai/ironclaw/pull/2190/commits/1a698297e7837224a30aa71f23b8bb2150c859a8 enforces this criteria.

serrrfirat (approved):

PR #2195, opened 2026-04-09
feat(tui): ship TUI in default binary

serrrfirat:
## Summary

- Add `tui` to default Cargo features so the Ratatui terminal UI ships in standard builds
- Remove `dist = false` from `ironclaw_tui` so cargo-dist includes it in release artifacts

The TUI only activates when explicitly configured at runtime (`config.channels.tui.is_some()`), so server/Docker deployments are completely unaffected — the deps compile in but nothing initializes. The dependency footprint is light (ratatui, crossterm, tui-textarea, pulldown-cmark, arboard, image/png).

Server builds that want to exclude it can still use `--no-default-features --features postgres,libsql`.

## Test plan

- [x] `cargo check` with default features — compiles clean
- [x] `cargo check --no-default-features --features postgres,libsql` — compiles clean without TUI
- [ ] CI passes

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -221,7 +221,7 @@ tempfile = "3"
 insta = "1.46.3"
 
 [features]
-default = ["postgres", "libsql", "html-to-markdown"]
+default = ["postgres", "libsql", "html-to-markdown", "tui"]
 postgres = [
     "dep:deadpool-postgres",
     "dep:tokio-postgres",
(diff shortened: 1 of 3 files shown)


serrrfirat:
Addressed review feedback: `arboard` and `image` are now optional deps behind a `clipboard` feature in `ironclaw_tui` (defaulted on). Headless/Docker builds that omit the `tui` feature won't pull in the X11/Wayland dependency chain. When `clipboard` is disabled, copy-to-clipboard is a no-op and image paste returns `None`.

henrypark133 (approved):

PR #2196, opened 2026-04-09
feat(llm): add CleaningProvider to strip thinking tags from LLM output

standardtoaster:
## Summary

- Adds `CleaningProvider`, a composable `LlmProvider` decorator that strips
  `<think>`/`<thinking>` tags and truncates unclosed tool-call tags from all
  completion output
- Cleans `complete()` content, `complete_with_tools()` content, and
  `ToolCall::reasoning` fields

## Problem

Open-weight reasoning models (Qwen3, DeepSeek-R1, QwQ) emit `<think>...</think>`
blocks in their responses. When these tags leak into tool call JSON, they break
parsing. When they reach user-facing text, they expose internal reasoning.

The `Reasoning` layer already calls `clean_response()` at ~10 call sites, but
this only covers the `Reasoning`-mediated path. Any consumer that calls
`LlmProvider` directly — container workers via `proxy_llm`, external MCP tool
hosts, or custom integrations — gets raw output with thinking tags intact.

### Why a decorator, not more call-site fixes?

Adding `clean_response()` to every new call site is fragile — miss one and tags
leak. A decorator at the provider level guarantees cleaning for all consumers
regardless of which layer they enter through. The existing `Reasoning` call sites
remain as-is (cleaning already-clean text is a no-op).

## Test plan

- [x] 14 unit tests covering the decorator contract: tag stripping on both
      completion paths, tool call reasoning field cleaning, unclosed tool tag
      truncation, multiple/nested think blocks, empty-after-strip edge case,
      error propagation, token count preservation, full trait delegation
- [x] 176 existing `reasoning` tests pass — validates `clean_response()` handles
      the full tag vocabulary (`<think>`, `<thinking>`, `<reflection>`,
      `<scratchpad>`, `<inner_monologue>`, pipe-delimited `<|think|>`, `<final>`
      extraction, code-block preservation, unclosed tags)
- [x] Manual validation against Qwen3-30B-A3B (MLX, OpenAI-compatible endpoint)
      with tool-calling prompts. Without CleaningProvider, `<think>` blocks
      appear in ~40% of tool call responses and break JSON parsing. With
      CleaningProvider wrapping the provider, 0% contain thinking tags and tool
      call parsing succeeds consistently.

### DIFF
--- a/src/llm/cleaning_provider.rs
+++ b/src/llm/cleaning_provider.rs
@@ -1,0 +1,195 @@
+//! A decorator that strips thinking tags from all LLM output.
+
+use std::sync::Arc;
+
+use async_trait::async_trait;
+use rust_decimal::Decimal;
+
+use crate::llm::clean_response;
+use crate::llm::error::LlmError;
+use crate::llm::provider::{
+    CompletionRequest, CompletionResponse, LlmProvider, ModelMetadata, ToolCompletionRequest,
+    ToolCompletionResponse,
+};
+
+/// Wraps any `LlmProvider` and applies `clean_response()` to strip thinking
+/// tags (`<think>`, `<thinking>`, etc.) from all completion output.
+pub struct CleaningProvider {
+    inner: Arc<dyn LlmProvider>,
+}
+
+impl CleaningProvider {
+    pub fn new(inner: Arc<dyn LlmProvider>) -> Self {
+        Self { inner }
+    }
+}
+
+#[async_trait]
+impl LlmProvider for CleaningProvider {
+    fn model_name(&self) -> &str {
+        self.inner.model_name()
+    }
+
+    fn cost_per_token(&self) -> (Decimal, Decimal) {
+        self.inner.cost_per_token()
+    }
+
+    async fn complete(&self, request: CompletionRequest) -> Result<CompletionResponse, LlmError> {
+        let mut resp = self.inner.complete(request).await?;
+        resp.content = clean_response(&resp.content);
+        Ok(resp)
+    }
+
+    async fn complete_with_tools(
+        &self,
+        request: ToolCompletionRequest,
+    ) -> Result<ToolCompletionResponse, LlmError> {
+        let mut resp = self.inner.complete_with_tools(request).await?;
+        if let Some(ref content) = resp.content {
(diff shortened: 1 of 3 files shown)


serrrfirat (commented):
I dug into this pretty deeply and found a couple of blockers plus one test-coverage gap.

1. **`CleaningProvider` is never wired into the real provider chain, so this currently has no runtime effect.**
   - The PR adds/export the decorator in `src/llm/mod.rs:49-50`, but the chain construction in `src/llm/mod.rs:560-699` still only wraps providers with `RetryProvider`, `SmartRoutingProvider`, `FailoverProvider`, `CircuitBreakerProvider`, `CachedProvider`, and optional `RecordingLlm`.
   - I don't see any `CleaningProvider::new(...)` in the chain, so as written this looks like dead code.

2. **Even if it were wired in, it sits too late in the flow to solve the stated "tool-call JSON breaks parsing" problem.**
   - `src/llm/cleaning_provider.rs:47-54` only cleans `resp.content` and `ToolCall.reasoning` *after* `inner.complete_with_tools(...)` returns a structured `ToolCompletionResponse`.
   - But by that point, `rig_adapter` has already parsed the raw provider response into structured tool calls:
     - `src/llm/rig_adapter.rs:839-877` (`extract_response`)
     - `src/llm/rig_adapter.rs:1076-1111` (`complete_with_tools` calling `extract_response`)
   - `ToolCall.arguments` is already materialized as `serde_json::Value` (`src/llm/provider.rs:257-264`), so if leaked `<think>` text was breaking tool-call extraction/JSON parsing, this decorator can't repair that after the fact.
   - That makes this effective as a redaction/display cleanup layer, but not as a fix for parser-boundary corruption.

3. **The tests only prove the decorator in isolation, not the caller path that matters.**
   - The new tests in `src/llm/cleaning_provider.rs:92+` validate the wrapper itself, but there's no chain-level test that drives the real builder/wiring path and would have caught (1).
   - Given the repo rule to test through the caller when a helper gates user-visible behavior, I think this needs at least one integration-ish test around provider-chain construction.

Happy to re-review if this gets moved into the actual chain and/or shifted earlier in the response-parsing path where it can address the raw-output failure mode described in the PR body.

standardtoaster:

PR #2200, opened 2026-04-09
fix(engine): admin skills visible to all tenants via system_project_id

### LINKED ISSUE #2084: bug(engine-v2): skills not visible to gateway users after resolve_user_project was introduced

zetyquickly:
## Summary

- Introduced `system_project_id()` (Uuid::nil) as a stable well-known project for admin-installed shared skills, requiring no database row
- Fixed skill migration so `Bundled`/`Installed` skills go to the system project under `shared_owner_id()`, while `User`/`Workspace` skills stay in the tenant's own project
- Extended `list_shared_memory_docs` to always include the system project, making admin skills visible to all tenants regardless of which per-user project their thread runs in
- Added `project_id` and `user_id` fields to YAML frontmatter serialization, with a `"legacy"` fallback and upgrade-path migration to stamp pre-existing docs with correct ownership
- Added a regression test covering the multi-tenant visibility contract (Bob sees admin skills, not Alice's personal skills)

## Change Type

- [x] Bug fix

## Linked Issue

Closes #2084

## Validation

- [x] `cargo build`
- [x] Relevant tests pass: `admin_skills_in_system_project_visible_to_all_tenants`, `bundled_skill_goes_to_system_project`, `installed_skill_goes_to_system_project`, `user_skill_goes_to_tenant_project`, `workspace_skill_goes_to_tenant_project`
- [ ] Manual testing: 

## Security Impact

None. Changes are scoped to in-memory skill visibility and YAML frontmatter; no permissions, secrets, network, or sandbox policy changes.

## Database Impact

No migrations. The system project uses `Uuid::nil()` — no DB row required. The `project_id`/`user_id` fields are now serialized to YAML frontmatter in the flat-file store; existing docs missing these fields fall back gracefully via `deserialize_knowledge_doc`.

## Blast Radius

- `ironclaw_engine`: `Store::list_shared_memory_docs` now queries the system project in addition to the per-user project — one extra DB/store read per call.
- `src/bridge/skill_migration.rs`: migration routing changed; `Bundled`/`Installed` skills will be re-migrated into the system project on next startup if they were previously stored in the tenant project (content-hash dedup prevents double-migration for unchanged skills).
- `src/bridge/store_adapter.rs`: YAML frontmatter now includes `project_id`/`user_id`; older docs without these fields get `system_project_id()` and `"legacy"` as defaults, then are upgraded by `migrate_legacy_user_ids`.
- No API surface changes.

## Rollback Plan

Revert the commit. Skills previously stored in the system project will remain there but `list_shared_memory_docs` will stop including them for non-system tenants. Bundled/Installed skills will be re-migrated into the tenant project on next startup by the pre-fix migration code. No data loss — `MemoryDoc` rows are never deleted per the data retention policy.

## Review Follow-Through

- The extra `list_memory_docs(sys, owner_id)` call inside `list_shared_memory_docs` iterates once per `shared_owner_candidates()` entry — worth confirming this stays fast under load or caching is acceptable.
- `migrate_legacy_user_ids` now processes system-project docs with `user_id = "legacy"`; reviewer should confirm the fallback owner assignment (`shared_owner_id()` for skills, `owner_id` for everything else) matches intent for non-skill doc types.

---

**Review track**: B (feature/maintainer-requested refactor)

### DIFF
--- a/crates/ironclaw_engine/src/traits/store.rs
+++ b/crates/ironclaw_engine/src/traits/store.rs
@@ -9,7 +9,7 @@ use crate::types::error::EngineError;
 use crate::types::event::ThreadEvent;
 use crate::types::memory::{DocId, MemoryDoc};
 use crate::types::mission::{Mission, MissionId, MissionStatus};
-use crate::types::project::{Project, ProjectId};
+use crate::types::project::{Project, ProjectId, system_project_id};
 use crate::types::step::Step;
 use crate::types::thread::{Thread, ThreadId, ThreadState};
 use crate::types::{is_shared_owner, shared_owner_candidates};
@@ -120,9 +120,16 @@ pub trait Store: Send + Sync {
         &self,
         project_id: ProjectId,
     ) -> Result<Vec<MemoryDoc>, EngineError> {
+        let sys = system_project_id();
         let mut docs = Vec::new();
         for owner_id in shared_owner_candidates() {
             docs.extend(self.list_memory_docs(project_id, owner_id).await?);
+            // Always include the system project — admin-installed skills live
+            // here and must be visible to all tenants regardless of which
+            // per-user project their thread runs in.
+            if project_id != sys {
+                docs.extend(self.list_memory_docs(sys, owner_id).await?);
+            }
         }
         docs.sort_by_key(|doc| doc.id.0);
         docs.dedup_by_key(|doc| doc.id);
--- a/src/bridge/router.rs
+++ b/src/bridge/router.rs
@@ -783,6 +783,7 @@ pub async fn init_engine(agent: &Agent) -> Result<(), Error> {
             match crate::bridge::skill_migration::migrate_v1_skill_list(
                 &skills_snapshot,
                 &store_dyn,
+                owner_id,
                 project_id,
             )
             .await
@@ -3291,6 +3292,24 @@ async fn migrate_legacy_user_ids(store: &Arc<dyn ironclaw_engine::Store>, owner_
                 doc.updated_at = chrono::Utc::now();
                 let _ = store.save_memory_doc(&doc).await;
             }
+        }
+    }
+
+    // Upgrade path: docs in the system project that were serialized before
+    // project_id/user_id were written to frontmatter will be loaded with
+    // user_id = "legacy". Skill docs there belong to the shared/admin space,
+    // not to any individual owner, so stamp them with shared_owner_id().
+    let sys = ironclaw_engine::system_project_id();
+    if let Ok(legacy) = store.list_memory_docs(sys, "legacy").await {
+        for mut doc in legacy {
+            let correct_owner = if doc.doc_type == ironclaw_engine::DocType::Skill {
+                ironclaw_engine::types::shared_owner_id().to_string()
+            } else {
+                owner_id.to_string()
+            };
+            doc.user_id = correct_owner;
+            doc.updated_at = chrono::Utc::now();
+            let _ = store.save_memory_doc(&doc).await;
         }
     }
 
--- a/src/bridge/skill_migration.rs
+++ b/src/bridge/skill_migration.rs
@@ -4,6 +4,12 @@
 //! v2 `MemoryDoc` with `DocType::Skill` and structured `V2SkillMetadata`.
 //! The migration is idempotent: skills with unchanged content_hash are skipped.
 //!
+//! **Ownership model:**
+//! - `Bundled` / `Installed` skills are admin-installed and go into the global
+//!   `system_project_id()` under `shared_owner_id()`. Every tenant sees them.
+//! - `User` / `Workspace` skills belong to the owner and go into their own
+//!   project under `owner_id`. Other tenants do not see them.
+//!
 //! **Remove after v1 migration is complete.** Once all users are on ENGINE_V2
 //! and SKILL.md files are authored directly as v2 MemoryDocs (or via the
 //! skill-extraction mission), this one-time migration code is unnecessary.
@@ -12,6 +18,7 @@
 
 use std::sync::Arc;
 
+use ironclaw_engine::system_project_id;
 use ironclaw_engine::traits::store::Store;
 use ironclaw_engine::types::error::EngineError;
 use ironclaw_engine::types::memory::{DocType, MemoryDoc};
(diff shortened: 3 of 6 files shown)


src/bridge/skill_migration.rs
```diff
@@ -31,25 +38,36 @@ use ironclaw_skills::v2::{SkillMetrics, V2SkillMetadata, V2SkillSource};
...
 ) -> Result<usize, EngineError> {
     if v1_skills.is_empty() {
         return Ok(0);
     }
 
-    // Load existing skill docs to check for duplicates by content_hash
-    let existing_docs = store.list_shared_memory_docs(project_id).await?;
+    // Load existing docs from both locations to check for duplicates by content_hash.
+    let sys = system_project_id();
+    let mut existing_docs = store.list_shared_memory_docs(sys).await?;
```
serrrfirat, line 65: **High Severity — Duplicate skills after upgrade from old migration code**

The dedup hash check calls `list_shared_memory_docs(sys)` which only returns system-project docs (`(sys, __shared__)` and `(sys, __legacy_shared__)`). However, bundled/installed skills migrated by the **previous** version of this code live at `(tenant_project, __shared__)` — these are NOT checked by the new dedup logic.

**What happens on upgrade:**
1. Old code previously migrated bundled skills to `(tenant_project_id, __shared__)`
2. New code checks `list_shared_memory_docs(sys)` + `list_memory_docs(tenant_project_id, owner_id)` — neither covers `(tenant_project_id, __shared__)`
3. Content hashes from old skills are not found → skills are re-migrated to `(system_project_id(), __shared__)`
4. User now sees **both** copies when `list_shared_memory_docs(tenant_project)` is called downstream (one from tenant's shared space, one from system project). `dedup_by_key(|doc| doc.id)` doesn't help because they have different `DocId`s.

**Suggested fix** (one-line change):

```rust
// Before:
let mut existing_docs = store.list_shared_memory_docs(sys).await?;

// After:
let mut existing_docs = store.list_shared_memory_docs(tenant_project_id).await?;
```

`list_shared_memory_docs(tenant_project_id)` already includes both the tenant project AND the system project (thanks to the new `if project_id != sys` branch in `store.rs`), so this covers all existing skill locations.

src/bridge/router.rs
```diff
@@ -3294,6 +3295,24 @@ async fn migrate_legacy_user_ids(store: &Arc<dyn ironclaw_engine::Store>, owner_
         }
     }
 
+    // Upgrade path: docs in the system project that were serialized before
+    // project_id/user_id were written to frontmatter will be loaded with
+    // user_id = "legacy". Skill docs there belong to the shared/admin space,
+    // not to any individual owner, so stamp them with shared_owner_id().
+    let sys = ironclaw_engine::system_project_id();
+    if let Ok(legacy) = store.list_memory_docs(sys, "legacy").await {
```
serrrfirat, line 3303: **Medium Severity — Race condition in system-project legacy migration for non-skill docs**

If two users start the engine simultaneously, both call `list_memory_docs(sys, "legacy")`, get the same list of legacy docs, and both update them. For skill docs this is fine (both set `user_id = __shared__`). But for non-skill docs in the system project, the last writer wins — stamping their `owner_id`, potentially stealing ownership from the other user.

In practice this is low-risk because non-skill docs shouldn't normally exist in the system project. But worth documenting as a known limitation, or gating the system-project migration behind a lock/flag so only one process runs it.

crates/ironclaw_engine/src/traits/store.rs
```diff
@@ -120,9 +120,16 @@ pub trait Store: Send + Sync {
...
     ) -> Result<Vec<MemoryDoc>, EngineError> {
+        let sys = system_project_id();
         let mut docs = Vec::new();
         for owner_id in shared_owner_candidates() {
             docs.extend(self.list_memory_docs(project_id, owner_id).await?);
+            // Always include the system project — admin-installed skills live
+            // here and must be visible to all tenants regardless of which
+            // per-user project their thread runs in.
+            if project_id != sys {
+                docs.extend(self.list_memory_docs(sys, owner_id).await?);
```
serrrfirat, line 131: **Medium Severity — Extra query count understated in blast radius**

The PR description says "one extra DB/store read per call" for `list_shared_memory_docs`. It's actually **2 extra reads** per call — one per `shared_owner_candidate` (`__shared__` and `__legacy_shared__`). This method is called from hot paths including `executor/prompt.rs:172`, `executor/loop_engine.rs:200`, and multiple places in `executor/orchestrator.rs`.

Under load, the actual overhead is 2× what the blast radius section describes. Worth correcting the description and considering whether the results should be cached for the duration of a step execution.

PR #2209, opened 2026-04-09
fix(v2): tool naming, auth gates, schema flatten, WASM traps, workspace race

ilblackdragon:
## Summary

Bug-fix bundle discovered while exercising the v2 engine end-to-end against real Notion + Google Drive MCP/WASM tools. Each fix has regression coverage and could ship independently. Supersedes #2227 (incorporated in `a8530e8b`).

### 1. Workspace `replace_chunks` TOCTOU race

Concurrent reindex of the same document hit `UNIQUE (document_id, chunk_index)` because `delete_chunks` + `insert_chunk` were separate transactions with async points between them. New `WorkspaceStore::replace_chunks` wraps both in one transaction: libsql uses `BEGIN IMMEDIATE`, postgres serializes via `SELECT ... FOR UPDATE` on the parent `memory_documents` row. Embeddings are pre-computed before opening the transaction so nothing async runs between delete and insert.

### 2. Auth gate display name + OAuth deletion guard + alias-aware activation

Three traps surfaced by the v2 Drive trace:
- Auth gate displayed `google_oauth_token` (credential name) instead of `google-drive-tool` (extension name) and passed it to `submit_auth_token` which expects an extension name, trapping users in a re-auth loop. Both paths now resolve via a shared `resolve_extension_for_action` helper.
- `activate_wasm_tool` and several capabilities-lookup sites used direct `dir.join(name.wasm)` instead of alias-aware helpers. A tool installed under the legacy hyphen filename silently failed to activate while the readiness probe reported it as ready.
- `configure()` post-activation cleanup unconditionally wiped the OAuth token the caller just provided. Now gated on whether the caller is providing a fresh credential in the same `secrets` map.
- `starts_with` prefix filters for MCP tool listing/removal used the raw (possibly hyphenated) server name while registry keys are normalized to underscores. Now uses `mcp_tool_id(name, "")` as the prefix at all 3 sites.

### 3. MCP tool name canonicalization

MCP tools with dashes in names (e.g. Notion's `notion-search`) were unreachable because the registry key preserved dashes while LLMs normalize to all-underscores. `mcp_tool_id` now replaces every non-`[A-Za-z0-9_]` character with `_` (handles dashes, dots, colons, slashes, unicode). Collision detection in `create_tools` emits a `warn!` when two distinct originals collide on the same normalized id.

### 4. OpenAI tool schema top-level flatten + Python/Rust action_calls round-trip

- **Schema flatten:** OpenAI rejects top-level `oneOf`/`anyOf`/`allOf`/`enum`/`not` in tool schemas. `normalize_schema_strict` now detects this, merges all variant properties into a single flat object (so the LLM keeps structured field hints), and appends the original schema to the description as keyword-aware advisory text. Also handles nullable object types (`"type": ["object", "null"]`) without unnecessary flattening.
- **Action_calls round-trip:** The Python orchestrator stored `{name, call_id, params}` but `json_to_thread_messages` tried to deserialize as `{action_name, id, parameters}`. Introduced `PythonActionCall` interchange struct as single source of truth. The warn log on parse failure summarizes structure only (no PII from tool params).

### 5. WASM trap classification + fuel limit

- Extracted `classify_trap_error` helper that uses structured `wasmtime::Trap` downcast (version-proof) instead of string matching. Covers `OutOfFuel`, `StackOverflow`, `UnreachableCodeReached`, and falls back to full error chain for unrecognized traps.
- Default fuel limit raised from 10M to 100M instructions. 10M was too low for WASM tools making HTTP requests + parsing JSON responses — a single 30KB JSON round-trip can burn 20-50M instructions in serde_json.

### 6. Incorporated from #2227

- Factory-level `server.name` normalization before any branch (including OAuth early-return)
- `new_with_name` normalization for callers bypassing the factory
- Bidirectional `resolve_key` alias resolution in `ToolRegistry` (exact → hyphen→underscore → underscore→hyphen)
- Legacy token secret name fallback for pre-normalization tokens (prevents forced re-auth on upgrade)
- WASM tool + channel loader filename normalization

### 7. Live e2e Drive auth gate test

End-to-end smoke test driving prompt → AuthRequired → token paste → resume. `#[ignore]`d behind live tier, no committed trace fixture (PII risk). Supporting harness: `with_no_trace_recording()`, multi-turn `finish_turns`, `TestRig::secrets_store()` + `owner_id()` accessors.

### 8. Test config

`Config::for_testing` now generates a random 32-byte AES master key (generated fresh per call) so replay-mode tests that touch credentials have a working secrets store out of the box.

## Test plan

- [x] `cargo test --workspace --lib` — **5387 passed, 0 failed, 5 ignored**
- [x] All commits rebase cleanly onto latest `origin/staging`
- [ ] Manual: install Notion MCP → tool calls succeed via `notion_notion_search`
- [ ] Manual: install Google Drive WASM with no credentials → `AuthRequired { extension_name: "google-drive-tool" }` surfaces, token paste resumes
- [ ] Manual: enable GitHub Copilot MCP → no HTTP 400 schema error
- [ ] Manual: engine v2 tool flow → zero `Rewriting orphaned tool_result` log lines

Supersedes #2227.

### DIFF
--- a/crates/ironclaw_engine/src/executor/orchestrator.rs
+++ b/crates/ironclaw_engine/src/executor/orchestrator.rs
@@ -2140,14 +2140,31 @@ impl From<PythonActionCall> for ActionCall {
 }
 
 /// Serialize a slice of `ActionCall`s into the Python interchange shape.
+///
+/// On serialization failure (essentially unreachable for `String + String +
+/// Value`, but still possible if the `serde_json::Value` parameters tree
+/// contains a key whose stringification fails), the entry is **dropped**
+/// from the output rather than replaced with `Value::Null`. The previous
+/// `unwrap_or_else(|_| Value::Null)` corrupted the array — Python's
+/// `default.py` accesses `c.get("name")` / `c.get("call_id")` /
+/// `c.get("params")` on each entry, so a `null` would crash with a Python
+/// `AttributeError` and lose the entire LLM step. `filter_map` produces a
+/// shorter array, which Python's tool-result loop handles correctly because
+/// it iterates `range(len(results))` against the shortened call list. The
+/// warn log is preserved so operators have a breadcrumb if it ever fires.
 fn action_calls_to_python_json(calls: &[ActionCall]) -> Vec<serde_json::Value> {
     calls
         .iter()
-        .map(|c| {
-            serde_json::to_value(PythonActionCall::from(c)).unwrap_or_else(|e| {
-                warn!("Failed to serialize ActionCall for Python: {e}");
-                serde_json::Value::Null
-            })
+        .filter_map(|c| match serde_json::to_value(PythonActionCall::from(c)) {
+            Ok(value) => Some(value),
+            Err(e) => {
+                warn!(
+                    error = %e,
+                    action_name = %c.action_name,
+                    "Failed to serialize ActionCall for Python orchestrator — dropping entry"
+                );
+                None
+            }
         })
         .collect()
 }
--- a/crates/ironclaw_engine/src/executor/orchestrator.rs
+++ b/crates/ironclaw_engine/src/executor/orchestrator.rs
@@ -2169,6 +2169,57 @@ fn action_calls_to_python_json(calls: &[ActionCall]) -> Vec<serde_json::Value> {
         .collect()
 }
 
+/// Build a PII-safe summary of an `action_calls` JSON value for log output.
+///
+/// The action_calls payload contains tool parameters, which can carry user
+/// PII (search queries, file names, email content, conversation text).
+/// Dumping the full value into a `warn!` log would leak that PII to log
+/// aggregation systems (Datadog, CloudWatch, Sentry) the moment the parser
+/// fails — and the parser only fails when the Python ↔ Rust shape drifts,
+/// which is exactly when an operator is most likely to be grepping logs.
+///
+/// We emit only the structural information operators actually need to
+/// debug a shape drift: array length and the keys of the first entry. The
+/// keys themselves are not user data — they're field names like
+/// `name`/`call_id`/`params` that are static across all calls.
+fn summarize_action_calls_for_log(value: &serde_json::Value) -> String {
+    match value.as_array() {
+        Some(arr) if arr.is_empty() => "empty array".to_string(),
+        Some(arr) => {
+            let first_keys = arr
+                .first()
+                .and_then(|v| v.as_object())
+                .map(|obj| {
+                    let mut keys: Vec<&str> = obj.keys().map(String::as_str).collect();
+                    keys.sort_unstable();
+                    keys.join(",")
+                })
+                .unwrap_or_else(|| "<not an object>".to_string());
+            format!(
+                "array of {} entries; first entry keys: [{}]",
+                arr.len(),
+                first_keys
+            )
+        }
+        None => format!("non-array value of type {}", json_value_type_name(value)),
+    }
+}
+
+/// Cheap type-name string for a `serde_json::Value`. Used by
+/// `summarize_action_calls_for_log` to surface the wrong-shape case
+/// (e.g. Python passed a string instead of an array) without leaking the
+/// actual contents.
+fn json_value_type_name(value: &serde_json::Value) -> &'static str {
+    match value {
+        serde_json::Value::Null => "null",
+        serde_json::Value::Bool(_) => "bool",
+        serde_json::Value::Number(_) => "number",
+        serde_json::Value::String(_) => "string",
+        serde_json::Value::Array(_) => "array",
+        serde_json::Value::Object(_) => "object",
+    }
+}
+
 /// Deserialize an `action_calls` JSON array (in Python interchange shape)
 /// back into canonical `ActionCall`s.
 ///
@@ -2180,13 +2231,16 @@ fn action_calls_to_python_json(calls: &[ActionCall]) -> Vec<serde_json::Value> {
 /// required field, partial migration), the warning is the operator-visible
 /// breadcrumb that explains why subsequent tool results suddenly look
 /// orphaned to `sanitize_tool_messages`.
+///
+/// The warn log emits a structural summary (`summarize_action_calls_for_log`)
+/// instead of the raw value because tool parameters can contain user PII.
 fn python_json_to_action_calls(value: &serde_json::Value) -> Option<Vec<ActionCall>> {
     match serde_json::from_value::<Vec<PythonActionCall>>(value.clone()) {
         Ok(parsed) => Some(parsed.into_iter().map(ActionCall::from).collect()),
         Err(e) => {
             warn!(
                 error = %e,
-                value = %value,
+                shape = %summarize_action_calls_for_log(value),
                 "Failed to parse action_calls from Python orchestrator — \
                  assistant message will lose tool_call linkage and downstream \
                  tool results will be rewritten as user messages"
@@ -2959,6 +3013,74 @@ mod tests {
         ]);
         // Missing "name", "call_id", "params" → returns None.
         assert!(python_json_to_action_calls(&canonical_json).is_none());
+    }
+
+    #[test]
+    fn summarize_action_calls_for_log_does_not_leak_user_pii() {
+        // The whole point of this helper is that the warn log path on a
+        // shape-drift failure must NOT dump tool parameters (which can
+        // contain user PII like search queries, file names, email content)
+        // into log aggregation systems. The summary should expose only
+        // structural information: array length and the keys of the first
+        // entry. The keys themselves are static (`name`, `call_id`,
+        // `params`), not user data.
+        let pii_value = serde_json::json!([
+            {
+                "name": "google_drive_tool",
+                "call_id": "call_xyz",
+                "params": {
+                    "query": "salary spreadsheet for joe",
+                    "secret_token": "very-sensitive-token-do-not-log"
+                }
+            },
+            {
+                "name": "gmail",
+                "call_id": "call_abc",
+                "params": {
+                    "subject": "private message about layoffs"
+                }
+            }
+        ]);
+        let summary = summarize_action_calls_for_log(&pii_value);
+
+        // Structural info present.
+        assert!(summary.contains("array of 2 entries"));
+        assert!(summary.contains("call_id"));
+        assert!(summary.contains("name"));
+        assert!(summary.contains("params"));
+
+        // PII fields and their values must NOT appear.
+        assert!(
+            !summary.contains("salary"),
+            "summary must not leak user PII from params: {summary}"
+        );
+        assert!(
+            !summary.contains("very-sensitive-token"),
+            "summary must not leak credential-shaped values: {summary}"
+        );
+        assert!(
+            !summary.contains("layoffs"),
+            "summary must not leak free-text content: {summary}"
+        );
+        assert!(
+            !summary.contains("google_drive_tool"),
+            "summary must not leak the tool name itself (could expose intent): {summary}"
+        );
+    }
+
+    #[test]
+    fn summarize_action_calls_for_log_handles_edge_cases() {
+        assert_eq!(
+            summarize_action_calls_for_log(&serde_json::json!([])),
+            "empty array"
+        );
+        assert!(
+            summarize_action_calls_for_log(&serde_json::json!("not an array")).contains("string")
+        );
+        assert!(
+            summarize_action_calls_for_log(&serde_json::json!({"foo": "bar"})).contains("object")
+        );
+        assert!(summarize_action_calls_for_log(&serde_json::json!(null)).contains("null"));
     }
 
     /// Caller-level regression test: feeds `json_to_thread_messages` the
--- a/crates/ironclaw_engine/src/executor/orchestrator.rs
+++ b/crates/ironclaw_engine/src/executor/orchestrator.rs
@@ -3267,4 +3267,65 @@ mod tests {
              is no longer needed — either remove it or update the contract."
         );
     }
+
+    /// Regression: `action_calls: null` is Python's legitimate "this
+    /// message has no tool calls" signal (text-only response). Before the
+    /// null filter, `python_json_to_action_calls` would fire a warn log
+    /// with "invalid type: null, expected a sequence" on every text-only
+    /// assistant message — a false alarm that masked real drift issues.
+    #[test]
+    fn json_to_thread_messages_handles_null_action_calls_gracefully() {
+        let messages = serde_json::json!([
+            {
+                "role": "Assistant",
+                "content": "Here is your answer.",
+                "action_calls": null
+            }
+        ]);
+        let parsed = json_to_thread_messages(&messages).expect("must parse");
+        assert_eq!(parsed.len(), 1);
+        assert_eq!(
+            parsed[0].role,
+            crate::types::message::MessageRole::Assistant
+        );
+        assert_eq!(parsed[0].content, "Here is your answer.");
+        assert!(
+            parsed[0].action_calls.is_none(),
+            "null action_calls must produce None, not a parse error"
+        );
+    }
+
+    /// Verify that messages WITHOUT the action_calls key at all (the most
+    /// common case for text responses) also parse correctly — this is the
+    /// baseline that the null-filtering regression test extends.
+    #[test]
+    fn json_to_thread_messages_handles_absent_action_calls() {
+        let messages = serde_json::json!([
+            {"role": "Assistant", "content": "Just text, no tools."}
+        ]);
+        let parsed = json_to_thread_messages(&messages).expect("must parse");
+        assert_eq!(parsed.len(), 1);
+        assert!(parsed[0].action_calls.is_none());
+    }
+
+    /// Empty action_calls array is valid (LLM decided not to call any
+    /// tools this turn but the response still has the array field). Must
+    /// produce `Some(vec![])`, not `None`.
+    #[test]
+    fn json_to_thread_messages_handles_empty_action_calls_array() {
+        let messages = serde_json::json!([
+            {
+                "role": "Assistant",
+                "content": "No tools needed.",
+                "action_calls": []
+            }
+        ]);
+        let parsed = json_to_thread_messages(&messages).expect("must parse");
+        assert_eq!(parsed.len(), 1);
+        let calls = parsed[0]
+            .action_calls
+            .as_ref()
+            .expect("empty array should produce Some(vec![])");
+        assert!(calls.is_empty());
+    }
 }
--- a/crates/ironclaw_engine/src/executor/orchestrator.rs
+++ b/crates/ironclaw_engine/src/executor/orchestrator.rs
@@ -2154,9 +2154,29 @@ fn action_calls_to_python_json(calls: &[ActionCall]) -> Vec<serde_json::Value> {
 
 /// Deserialize an `action_calls` JSON array (in Python interchange shape)
 /// back into canonical `ActionCall`s.
+///
+/// Logs a warning on failure rather than swallowing silently. The whole
+/// commit that introduced this helper exists to undo a `.ok()` swallow that
+/// dropped action_calls without any signal — replacing it with another
+/// `.ok()?` would re-introduce the same trap, just one layer deeper. If the
+/// shape ever drifts again (Python orchestrator field rename, extra
+/// required field, partial migration), the warning is the operator-visible
+/// breadcrumb that explains why subsequent tool results suddenly look
+/// orphaned to `sanitize_tool_messages`.
 fn python_json_to_action_calls(value: &serde_json::Value) -> Option<Vec<ActionCall>> {
-    let parsed: Vec<PythonActionCall> = serde_json::from_value(value.clone()).ok()?;
-    Some(parsed.into_iter().map(ActionCall::from).collect())
+    match serde_json::from_value::<Vec<PythonActionCall>>(value.clone()) {
+        Ok(parsed) => Some(parsed.into_iter().map(ActionCall::from).collect()),
+        Err(e) => {
+            warn!(
+                error = %e,
+                value = %value,
+                "Failed to parse action_calls from Python orchestrator — \
+                 assistant message will lose tool_call linkage and downstream \
+                 tool results will be rewritten as user messages"
+            );
+            None
+        }
+    }
 }
 
 fn json_to_thread_messages(value: &serde_json::Value) -> Option<Vec<ThreadMessage>> {
--- a/crates/ironclaw_engine/src/executor/orchestrator.rs
+++ b/crates/ironclaw_engine/src/executor/orchestrator.rs
@@ -42,7 +42,7 @@ use crate::types::event::{EventKind, ThreadEvent, summarize_params};
 use crate::types::message::ThreadMessage;
 use crate::types::project::ProjectId;
 use crate::types::shared_owner_id;
-use crate::types::step::{StepId, TokenUsage};
+use crate::types::step::{ActionCall, StepId, TokenUsage};
 use crate::types::thread::{ActiveSkillProvenance, Thread, ThreadState};
 use ironclaw_common::ValidTimezone;
 
@@ -652,16 +652,9 @@ async fn handle_llm_complete(
                     serde_json::json!({"type": "code", "code": code, "usage": usage})
                 }
                 LlmResponse::ActionCalls { calls, content } => {
-                    let calls_json: Vec<serde_json::Value> = calls
-                        .iter()
-                        .map(|c| {
-                            serde_json::json!({
-                                "name": c.action_name,
-                                "call_id": c.id,
-                                "params": c.parameters,
-                            })
-                        })
-                        .collect();
+                    // Single source of truth for the Python interchange
+                    // shape — must round-trip via `python_json_to_action_calls`.
+                    let calls_json = action_calls_to_python_json(&calls);
                     serde_json::json!({
                         "type": "actions",
                         "content": content,
@@ -2099,6 +2092,73 @@ fn build_orchestrator_inputs(
     (names, values)
 }
 
+/// JSON shape used to interchange `ActionCall`s with the Python orchestrator.
+///
+/// This is the *single* place that defines the field naming convention used
+/// across the Python boundary. It is intentionally separate from the
+/// canonical `ActionCall` type because:
+///
+/// - `ActionCall` uses Rust-idiomatic field names (`id`, `action_name`,
+///   `parameters`) and is also persisted into Step records and ThreadEvents.
+///   Renaming its serde fields would invalidate every existing row.
+/// - The Python orchestrator uses friendlier names (`call_id`, `name`,
+///   `params`) that read naturally in CodeAct prompts and `default.py`.
+///
+/// Without this type, the round-trip is asymmetric: Rust → Python uses one
+/// shape, Python → Rust used `serde_json::from_value::<Vec<ActionCall>>`
+/// which silently fails (`.ok()` swallows the error) and produces `None`,
+/// which means assistant messages came back without `action_calls`. The
+/// downstream effect is that every tool result looks orphaned to
+/// `sanitize_tool_messages` and gets rewritten as a user message — losing
+/// the assistant ↔ tool_result linkage the LLM needs to reason about prior
+/// tool calls.
+#[derive(Debug, serde::Serialize, serde::Deserialize)]
+struct PythonActionCall {
+    name: String,
+    call_id: String,
+    params: serde_json::Value,
+}
+
+impl From<&ActionCall> for PythonActionCall {
+    fn from(c: &ActionCall) -> Self {
+        Self {
+            name: c.action_name.clone(),
+            call_id: c.id.clone(),
+            params: c.parameters.clone(),
+        }
+    }
+}
+
+impl From<PythonActionCall> for ActionCall {
+    fn from(p: PythonActionCall) -> Self {
+        Self {
+            id: p.call_id,
+            action_name: p.name,
+            parameters: p.params,
+        }
+    }
+}
+
+/// Serialize a slice of `ActionCall`s into the Python interchange shape.
+fn action_calls_to_python_json(calls: &[ActionCall]) -> Vec<serde_json::Value> {
+    calls
+        .iter()
+        .map(|c| {
+            serde_json::to_value(PythonActionCall::from(c)).unwrap_or_else(|e| {
+                warn!("Failed to serialize ActionCall for Python: {e}");
+                serde_json::Value::Null
+            })
+        })
+        .collect()
+}
+
+/// Deserialize an `action_calls` JSON array (in Python interchange shape)
+/// back into canonical `ActionCall`s.
+fn python_json_to_action_calls(value: &serde_json::Value) -> Option<Vec<ActionCall>> {
+    let parsed: Vec<PythonActionCall> = serde_json::from_value(value.clone()).ok()?;
+    Some(parsed.into_iter().map(ActionCall::from).collect())
+}
+
 fn json_to_thread_messages(value: &serde_json::Value) -> Option<Vec<ThreadMessage>> {
     let arr = value.as_array()?;
     let mut messages = Vec::with_capacity(arr.len());
@@ -2111,7 +2171,7 @@ fn json_to_thread_messages(value: &serde_json::Value) -> Option<Vec<ThreadMessag
             .unwrap_or_default();
         let action_calls = item
             .get("action_calls")
-            .and_then(|v| serde_json::from_value(v.clone()).ok());
+            .and_then(python_json_to_action_calls);
 
         let message = match role {
             "System" | "system" => ThreadMessage::system(content),
@@ -2774,5 +2834,164 @@ mod tests {
         let result = serde_json::json!({"outcome": "stopped"});
         let outcome = parse_outcome(&result);
         assert!(matches!(outcome, ThreadOutcome::Stopped));
+    }
+
+    // ── Python ↔ Rust ActionCall round-trip ───────────────────────────────
+    //
+    // Regression tests for the orphaned-tool-result bug. The Python
+    // orchestrator stores `action_calls` on assistant messages using the
+    // shape `{name, call_id, params}`, but the canonical Rust `ActionCall`
+    // uses `{action_name, id, parameters}`. Without the explicit
+    // `PythonActionCall` interchange type, `serde_json::from_value` would
+    // silently fail (`.ok()` swallows the error) and the Python-shaped
+    // assistant message would be parsed back as a plain assistant message
+    // with no tool calls, causing every subsequent ActionResult to be
+    // detected as orphaned by `sanitize_tool_messages` in the host crate.
+
+    #[test]
+    fn python_action_call_round_trips_through_serde() {
+        let original = ActionCall {
+            id: "call_abc123".to_string(),
+            action_name: "google_drive_tool".to_string(),
+            parameters: serde_json::json!({"query": "expenses"}),
+        };
+
+        let python_json = serde_json::to_value(PythonActionCall::from(&original))
+            .expect("PythonActionCall must serialize");
+        // Python-friendly field names — match what default.py reads.
+        assert_eq!(python_json["name"], "google_drive_tool");
+        assert_eq!(python_json["call_id"], "call_abc123");
+        assert_eq!(
+            python_json["params"],
+            serde_json::json!({"query": "expenses"})
+        );
+
+        let parsed: PythonActionCall =
+            serde_json::from_value(python_json).expect("must deserialize");
+        let round_tripped: ActionCall = parsed.into();
+        assert_eq!(round_tripped.id, original.id);
+        assert_eq!(round_tripped.action_name, original.action_name);
+        assert_eq!(round_tripped.parameters, original.parameters);
+    }
+
+    #[test]
+    fn action_calls_to_python_json_uses_python_field_names() {
+        let calls = vec![
+            ActionCall {
+                id: "call_1".to_string(),
+                action_name: "notion_notion_search".to_string(),
+                parameters: serde_json::json!({"query": "name"}),
+            },
+            ActionCall {
+                id: "call_2".to_string(),
+                action_name: "google_drive_tool".to_string(),
+                parameters: serde_json::json!({"action": "list"}),
+            },
+        ];
+        let json = action_calls_to_python_json(&calls);
+        assert_eq!(json.len(), 2);
+        assert_eq!(json[0]["name"], "notion_notion_search");
+        assert_eq!(json[0]["call_id"], "call_1");
+        assert_eq!(json[1]["name"], "google_drive_tool");
+        assert_eq!(json[1]["call_id"], "call_2");
+    }
+
+    #[test]
+    fn python_json_to_action_calls_parses_python_field_names() {
+        // The exact shape default.py produces (and stores on assistant
+        // messages via `append_message(..., action_calls=calls)`).
+        let python_json = serde_json::json!([
+            {"name": "notion_notion_search", "call_id": "call_xyz", "params": {"q": "foo"}},
+            {"name": "google_drive_tool", "call_id": "call_abc", "params": {"action": "list"}},
+        ]);
+        let parsed = python_json_to_action_calls(&python_json).expect("must parse");
+        assert_eq!(parsed.len(), 2);
+        assert_eq!(parsed[0].action_name, "notion_notion_search");
+        assert_eq!(parsed[0].id, "call_xyz");
+        assert_eq!(parsed[0].parameters, serde_json::json!({"q": "foo"}));
+        assert_eq!(parsed[1].action_name, "google_drive_tool");
+        assert_eq!(parsed[1].id, "call_abc");
+    }
+
+    #[test]
+    fn python_json_to_action_calls_rejects_canonical_field_names() {
+        // Sanity check: the parser is strict about Python field names.
+        // If `default.py` ever changes the shape, the test must catch it.
+        let canonical_json = serde_json::json!([
+            {"action_name": "search", "id": "call_x", "parameters": {}}
+        ]);
+        // Missing "name", "call_id", "params" → returns None.
+        assert!(python_json_to_action_calls(&canonical_json).is_none());
+    }
+
+    /// Caller-level regression test: feeds `json_to_thread_messages` the
+    /// exact JSON shape that `default.py` produces for an assistant message
+    /// with tool calls followed by tool results, and asserts that the
+    /// resulting `ThreadMessage`s preserve the `action_calls` ↔
+    /// `action_call_id` linkage. Without the `PythonActionCall` parser the
+    /// assistant message would come back with `action_calls = None` and
+    /// every following ActionResult would look orphaned to the bridge.
+    #[test]
+    fn json_to_thread_messages_preserves_action_calls_from_python_orchestrator() {
+        // This is the literal shape `default.py` writes into
+        // `state["working_messages"]` after a Tier 0 step:
+        //
+        //   append_message(working_messages, "Assistant", "...", action_calls=calls)
+        //   append_message(working_messages, "ActionResult", "...", action_name=..., action_call_id=...)
+        //
+        // where `calls` came from the LLM response and has shape
+        // `[{"name": ..., "call_id": ..., "params": ...}]`.
+        let working_messages = serde_json::json!([
+            {"role": "User", "content": "search in notion for my name"},
+            {
+                "role": "Assistant",
+                "content": "",
+                "action_calls": [
+                    {
+                        "name": "notion_notion_search",
+                        "call_id": "call_xyz",
+                        "params": {"query": "Illia"}
+                    }
+                ]
+            },
+            {
+                "role": "ActionResult",
+                "content": "found 3 results",
+                "action_name": "notion_notion_search",
+                "action_call_id": "call_xyz"
+            }
+        ]);
+
+        let messages = json_to_thread_messages(&working_messages).expect("must parse");
+        assert_eq!(messages.len(), 3);
+
+        // The assistant message MUST have action_calls populated, with
+        // matching call_id. If this assertion fails, the bridge layer
+        // will treat the following ActionResult as orphaned and rewrite
+        // it as a user message — losing the model's ability to reason
+        // about prior tool output.
+        let assistant = &messages[1];
+        assert_eq!(
+            assistant.role,
+            crate::types::message::MessageRole::Assistant
+        );
+        let calls = assistant
+            .action_calls
+            .as_ref()
+            .expect("assistant message must carry action_calls after round-trip");
+        assert_eq!(calls.len(), 1);
+        assert_eq!(calls[0].id, "call_xyz");
+        assert_eq!(calls[0].action_name, "notion_notion_search");
+        assert_eq!(calls[0].parameters, serde_json::json!({"query": "Illia"}));
+
+        // The ActionResult must reference the same call_id so the bridge
+        // can pair them.
+        let result = &messages[2];
+        assert_eq!(
+            result.role,
+            crate::types::message::MessageRole::ActionResult
+        );
+        assert_eq!(result.action_call_id.as_deref(), Some("call_xyz"));
+        assert_eq!(result.action_name.as_deref(), Some("notion_notion_search"));
     }
 }
--- a/crates/ironclaw_engine/src/executor/orchestrator.rs
+++ b/crates/ironclaw_engine/src/executor/orchestrator.rs
@@ -3174,4 +3174,97 @@ mod tests {
         assert_eq!(result.action_call_id.as_deref(), Some("call_xyz"));
         assert_eq!(result.action_name.as_deref(), Some("notion_notion_search"));
     }
+
+    /// Regression for the gate-resume / bootstrap path: when a thread
+    /// resumes after approval or auth, `build_orchestrator_inputs`
+    /// serializes `thread.internal_messages` into the bootstrap context
+    /// that Python reads into `working_messages`. If `action_calls` is
+    /// serialized with canonical `ActionCall` field names (`action_name`,
+    /// `id`, `parameters`) instead of the Python interchange names
+    /// (`name`, `call_id`, `params`), the next `__llm_complete__` call
+    /// passes them back through `json_to_thread_messages` which fails
+    /// with "missing field `name`" and orphans every subsequent tool
+    /// result.
+    ///
+    /// This test simulates the full round-trip: build a `ThreadMessage`
+    /// with action_calls → serialize through `build_orchestrator_inputs`'s
+    /// exact serialization pattern → parse back through
+    /// `json_to_thread_messages` → assert the calls survive. If anyone
+    /// adds a THIRD serialization path in the future and uses canonical
+    /// names, this test documents the pattern they should follow.
+    #[test]
+    fn bootstrap_context_action_calls_round_trip_through_python_interchange() {
+        // Build a thread message the way the engine does: an assistant
+        // message with action_calls in canonical ActionCall format (the
+        // shape stored in the DB / internal_messages).
+        let msg = ThreadMessage::assistant_with_actions(
+            Some("I'll search for that".to_string()),
+            vec![ActionCall {
+                id: "call_resume_test".to_string(),
+                action_name: "google_drive_tool".to_string(),
+                parameters: serde_json::json!({"query": "budget"}),
+            }],
+        );
+
+        // Serialize through the SAME pattern `build_orchestrator_inputs`
+        // uses. This is the exact code path that was broken before the
+        // fix — it was using `"action_calls": m.action_calls` which
+        // produced canonical field names.
+        let calls_json = msg
+            .action_calls
+            .as_ref()
+            .map(|calls| serde_json::Value::Array(action_calls_to_python_json(calls)));
+        let serialized = serde_json::json!([{
+            "role": "Assistant",
+            "content": msg.content,
+            "action_name": msg.action_name,
+            "action_call_id": msg.action_call_id,
+            "action_calls": calls_json,
+        }]);
+
+        // Parse back through the same path Python's working_messages
+        // takes when it calls __llm_complete__.
+        let parsed = json_to_thread_messages(&serialized).expect("must parse");
+        assert_eq!(parsed.len(), 1);
+
+        let assistant = &parsed[0];
+        let calls = assistant.action_calls.as_ref().expect(
+            "bootstrap context action_calls must survive the round-trip. \
+                 If this fails, a serialization path is using canonical ActionCall \
+                 field names instead of PythonActionCall interchange names.",
+        );
+        assert_eq!(calls.len(), 1);
+        assert_eq!(calls[0].id, "call_resume_test");
+        assert_eq!(calls[0].action_name, "google_drive_tool");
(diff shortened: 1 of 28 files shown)


henrypark133 (dismissed):

src/bridge/router.rs
```diff
@@ -1540,6 +1581,33 @@ pub async fn resolve_gate(
...
+                // Resolve the actual extension via the action that
+                // triggered the gate. For built-in tools, HTTP, and
+                // skill credentials there's no provider extension and
+                // we fall back to the credential name (the existing
+                // pre-fix behaviour for those callers — `submit_auth_token`
+                // then routes through the skill registry).
+                let submit_target = state
+                    .effect_adapter
+                    .tools()
+                    .provider_extension_for_tool(&pending.action_name)
```
henrypark133: **[nit | Maintainability]** The `provider_extension_for_tool + unwrap_or_else(credential_name)` pattern is implemented three times in this file: once via the `resolve_auth_gate_display_name` helper (line ~65), and twice inlined here and in `await_thread_outcome` (line ~2782).

The inline sites don't use the helper because they've already destructured `credential_name` from the `ResumeKind` match and also need the result for `submit_auth_token`, not just display. A small refactor — extracting the core into something like `async fn resolve_extension_for_action(tools, action_name, credential_fallback) -> String` — would let all three sites share one implementation. Not blocking, just a maintenance consideration if this resolution logic ever needs to change.

henrypark133 (commented):
## Code Review Summary

| Severity | Count |
|----------|-------|
| Critical | 0     |
| Warning  | 0     |
| Nit      | 1     |

This is a very high-quality PR. All five fixes are well-motivated, clearly documented, and have thorough regression test coverage. The commit messages and PR description are exemplary — each bug is traced from root cause through fix with specific behavior-before-and-after analysis.

**Highlights:**

- **fix(workspace)**: The `BEGIN IMMEDIATE` choice for libSQL is correct — `DEFERRED` transactions bypass `busy_timeout` on write contention, which is a non-obvious SQLite gotcha. Pre-computing embeddings before entering the transaction to eliminate async points inside it is a smart design decision.
- **fix(auth)**: The `configure()` OAuth token deletion guard (`!secrets.contains_key(&auth_cfg.secret_name)`) cleanly separates the "submit token" path from the "reconfigure" path.
- **fix(mcp)**: Canonicalizing at registration via `mcp_tool_id` while preserving the original dashed name on the inner `McpTool` for wire calls is the right layering.
- **fix(llm)**: `PythonActionCall` interchange type captures the field-naming asymmetry in one place without touching `ActionCall`'s serde attributes (which would invalidate persisted data). The caller-level `json_to_thread_messages` test is exactly the kind of regression test the codebase's testing rules call for.
- **Schema flatten**: Replacing the schema with a permissive envelope and moving the original to advisory description text is the right call. Truncation on char boundary is a nice detail.

### Test Coverage Notes
- All five fixes have dedicated regression tests that exercise the actual bug path
- Workspace TOCTOU test uses 4 concurrent writers with multi-threaded tokio runtime
- MCP tool name test includes a caller-level round-trip through `ToolRegistry::resolve_name`
- Live tests properly `#[ignore]`d with `with_no_trace_recording()` to prevent PII leakage
- `insert_chunk` on the trait is still used by `src/import/openclaw/memory.rs`, so it's correctly retained alongside the new `replace_chunks`

ilblackdragon:
## Review fixes

Pushed 4 fixup commits addressing the Critical, High, and Medium issues from the review. Each fix is its own commit so the trail is reviewable in isolation.

| # | Severity | Commit | What changed |
|---|---|---|---|
| 1 | 🔴 Critical | [\`9ae48a7c\`](https://github.com/nearai/ironclaw/commit/9ae48a7c) | \`python_json_to_action_calls\` now logs a \`warn!\` (with the parse error and the offending JSON value) on the failure path instead of swallowing via \`.ok()?\`. The whole point of the parent commit was to undo a \`.ok()\` swallow and the original fix re-introduced the same trap one layer deeper. |
| 2 | 🟠 High | [\`5a6a3ac0\`](https://github.com/nearai/ironclaw/commit/5a6a3ac0) | \`Config::for_testing\` now generates a fresh random AES-256-GCM master key per call via \`generate_test_master_key()\` (32 random bytes from \`rand::thread_rng()\`, hex-encoded). The hardcoded \`0123456789abcdef...\` constant is gone — every developer building with \`--features libsql\` no longer has a publicly-known master key in their process. \`#[cfg(test)]\` wasn't an option because integration tests in \`tests/*.rs\` are separate crates compiled against the lib's non-test surface. |
| 3 | 🟡 Medium | [\`18d4ce48\`](https://github.com/nearai/ironclaw/commit/18d4ce48) | \`mcp_tool_id\` now replaces every non-\`[A-Za-z0-9_]\` character with \`_\`, not just dashes. Handles dots, colons, slashes, spaces, and multi-byte unicode in one pass via \`chars().map()\`. New regression test \`test_mcp_tool_id_normalizes_non_identifier_chars\` exercises all five categories. |
| 4 | 🟡 Medium | [\`264955e8\`](https://github.com/nearai/ironclaw/commit/264955e8) | \`flatten_top_level\` now picks a description hint that matches the actual keyword that triggered the flatten. \`oneOf\`/\`anyOf\` get \"pick ONE variant\", \`allOf\` gets \"pass fields from ALL variants combined\", \`enum\` gets \"pass one of the listed values\", \`not\` gets \"any object that does NOT match\". Extracted \`FORBIDDEN_TOP_LEVEL\` to a module constant and added \`detect_forbidden_top_level\` + \`schema_flatten_hint_intro\` helpers. New regression test \`test_normalize_schema_strict_hint_is_keyword_aware\`. |

## Issues I deliberately did NOT fix

**🟠 High: Auth display name TOCTOU.** Caching the resolved name on \`PendingGate\` would touch the persisted struct (\`Serialize\`/\`Deserialize\` via \`GatePersistence\`), require a backward-compat \`#[serde(default)]\` migration, and add a field to ~6 PendingGate construction sites. The TOCTOU window is small (it only matters if the tool is unregistered between the gate firing and the user pasting a token, in the same conversation), and even when it triggers the failure mode is benign — the second resolution falls back to \`credential_name\` which is the pre-fix behavior. I noted in my own review it was \"lower priority; flag and forget\" and I'm sticking to that judgement. Worth a follow-up issue if you want a paper trail.

**🟡 Medium: Schema flatten truncation strategy.** The 1500-byte byte-budget truncation is genuinely a hack — for a real GitHub Copilot \`github\` schema the LLM only sees the first ~15 variants. The right fix is a semantic summary (\"this tool accepts one of N variants discriminated by \`action\`: ..., enumerate the discriminator values\"), not just bumping the budget. That's significant work and out of scope for this PR — would prefer it as a follow-up issue with proper design rather than a band-aid here.

**🟢 Low: \`replace_chunks\` comment wording and embedding-dimension validation symmetry between Postgres and libsql.** Cosmetic / pre-existing risk, no functional impact, leaving alone.

## Verification

Same gates as the original PR, all green:

- \`cargo fmt --all -- --check\` — clean
- \`cargo clippy --all --benches --tests --examples --all-features\` — zero warnings (one \`manual_find\` lint surfaced and was fixed in commit 4)
- \`cargo test --workspace --lib\` — **5371 passed, 0 failed, 5 ignored** (was 5370; +2 new tests, -1 from a test count shift in one of the smaller binaries)

crates/ironclaw_engine/src/executor/orchestrator.rs
```diff
@@ -1903,6 +1896,93 @@ fn build_orchestrator_inputs(
...
+/// Without this type, the round-trip is asymmetric: Rust → Python uses one
+/// shape, Python → Rust used `serde_json::from_value::<Vec<ActionCall>>`
+/// which silently fails (`.ok()` swallows the error) and produces `None`,
+/// which means assistant messages came back without `action_calls`. The
+/// downstream effect is that every tool result looks orphaned to
+/// `sanitize_tool_messages` and gets rewritten as a user message — losing
+/// the assistant ↔ tool_result linkage the LLM needs to reason about prior
+/// tool calls.
+#[derive(Debug, serde::Serialize, serde::Deserialize)]
+struct PythonActionCall {
```
serrrfirat, line 2131: **Medium Severity — Robustness**

`action_calls_to_python_json` inserts `Value::Null` into the serialized array when `serde_json::to_value` fails:

```rust
serde_json::to_value(PythonActionCall::from(c)).unwrap_or_else(|e| {
    warn!("Failed to serialize ActionCall for Python: {e}");
    serde_json::Value::Null
})
```

The Python orchestrator (`default.py`) iterates this array accessing `.name`, `.call_id`, `.params` — a `null` entry would cause an `AttributeError` or silent corruption of the assistant message. While the failure probability is near-zero for this struct shape (String + String + Value), the fallback silently corrupts the array rather than propagating the error — which contradicts this PR's own stated goal of not silently swallowing errors.

**Suggested fix:** Filter nulls out (`.filter_map(|c| serde_json::to_value(PythonActionCall::from(c)).ok())`) or return `Result<Vec<Value>>` and let the caller decide.

src/tools/mcp/client.rs
```diff
@@ -652,6 +652,44 @@ fn extract_server_name(url: &str) -> String {
...
+/// unicode characters (e.g. emoji or non-ASCII letters) are also normalized
+/// to `_` so the registry key stays a valid Rust identifier suffix.
+///
+/// The original (possibly hyphenated / dotted / unicode) tool name is still
+/// stored on the `McpToolWrapper`'s inner `McpTool` and used verbatim when
+/// forwarding the `tools/call` request to the MCP server, so this
+/// normalization is internal-only and does not affect protocol compatibility.
+pub(crate) fn mcp_tool_id(server_name: &str, tool_name: &str) -> String {
+    format!("{server_name}_{tool_name}")
+        .chars()
```
serrrfirat, line 747: **Medium Severity — Edge Case (tool name collision)**

`mcp_tool_id` normalizes all non-`[A-Za-z0-9_]` characters to `_`. Two MCP tools from the same server whose names differ only by `-` vs `_` (e.g. `search-all` and `search_all`) would produce identical registry keys. The second `register()` call silently shadows the first with no warning.

I understand the LLM can't distinguish these either (it normalizes dashes to underscores too), so the collision is arguably the correct semantic. But an operator debugging an unreachable-tool issue would have zero signal that a collision occurred.

**Suggested fix:** Add a trace/warn log in `ToolRegistry::register` (or here in `create_tools`) when overwriting an existing entry whose original (pre-normalization) name differs from the new one. This makes the collision visible without changing behavior.

crates/ironclaw_engine/src/executor/orchestrator.rs
```diff
@@ -1915,7 +1995,7 @@ fn json_to_thread_messages(value: &serde_json::Value) -> Option<Vec<ThreadMessag
             .unwrap_or_default();
         let action_calls = item
             .get("action_calls")
-            .and_then(|v| serde_json::from_value(v.clone()).ok());
+            .and_then(python_json_to_action_calls);
```
serrrfirat, line 2287: **Low Severity — Potential data leakage in logs**

`python_json_to_action_calls` logs `value = %value` at `warn!` level on parse failure, dumping the full `action_calls` JSON array including all tool parameters:

```rust
warn!(
    error = %e,
    value = %value,  // ← full action_calls array with tool params
    "Failed to parse action_calls from Python orchestrator — ..."
);
```

Tool parameters can contain user PII (search queries, file names, email content, conversation text). This warn fires when the Python ↔ Rust shape drifts — a situation where operators will grep logs, potentially exposing PII to log aggregation systems (Datadog, CloudWatch, etc.).

**Suggested fix:** Truncate or elide `value` — e.g. log only the array length and the keys of the first element, or cap the display at ~300 chars.

src/workspace/repository.rs
```diff
@@ -329,6 +329,63 @@ impl Repository {
...
+        chunks: &[ChunkWrite],
+    ) -> Result<(), WorkspaceError> {
+        let mut conn = self.conn().await?;
+
+        let tx = conn
+            .transaction()
+            .await
+            .map_err(|e| WorkspaceError::ChunkingFailed {
+                reason: format!("Begin transaction failed: {e}"),
+            })?;
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in 2ababa48 — added `SELECT 1 FROM memory_documents WHERE id = $1 FOR UPDATE` at the top of the postgres `replace_chunks` transaction so concurrent reindexers serialize on the parent row. The `FOR UPDATE` row lock is per-document and ties to the existing parent row (FK already in place from `memory_chunks.document_id`), so the lookup is always cheap and always finds a row.

Picked `FOR UPDATE` over `pg_advisory_xact_lock(hashtext($1::text))` because it's the row-locking primitive PG operators expect when reading the code, and it doesn't introduce a hash function dependency. Good catch — you're absolutely right that the libsql `BEGIN IMMEDIATE` fix doesn't translate to postgres.

src/tools/mcp/client.rs
```diff
@@ -652,6 +652,24 @@ fn extract_server_name(url: &str) -> String {
...
+/// Python identifiers by converting dashes to underscores. If we registered
+/// `notion_notion-search` the LLM would emit a call for `notion_notion_search`
+/// and the registry lookup would miss, leaving the tool unreachable.
+///
+/// The original (possibly hyphenated) tool name is still stored on the
+/// `McpToolWrapper`'s inner `McpTool` and used verbatim when forwarding the
+/// `tools/call` request to the MCP server, so this normalization is
+/// internal-only and does not affect protocol compatibility.
+pub(crate) fn mcp_tool_id(server_name: &str, tool_name: &str) -> String {
+    format!("{server_name}_{tool_name}").replace('-', "_")
```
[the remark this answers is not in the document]
ilblackdragon: Already addressed in 18d4ce48 (the previous review-fix pass) — that commit replaces the `replace('-', "_")` with a `chars().map()` pass that sends **every** non-`[A-Za-z0-9_]` character to `_`, which is a strict superset of the dot-only suggestion. Dots, colons, slashes, spaces, and multi-byte unicode all collapse to `_`. New regression test `test_mcp_tool_id_normalizes_non_identifier_chars` exercises all five categories. No additional change needed.

src/llm/rig_adapter.rs
```diff
@@ -123,22 +123,123 @@ fn round_f32_to_f64(val: f32) -> f64 {
...
+        );
+        description.push_str(&hint);
+    }
+
+    *parameters = serde_json::json!({
+        "type": "object",
+        "properties": {},
+        "additionalProperties": true,
+        "required": []
+    });
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in a11c1995 — added `merge_top_level_variant_properties` which walks the union variants and collects every property they declare into a single map. The flatten envelope now uses that map instead of empty `{}`, so the LLM sees structured field hints and can do schema-based reasoning about which fields exist across the union.

`additionalProperties: true` and `required: []` are preserved so strict-mode validation stays disabled and the LLM is free to mix fields across variants — the upstream MCP server enforces the actual constraints. First-write wins on conflicting types: if two variants declare the same field with different schemas, the first variant's schema is kept (the full original schema still goes into the description hint, so the ambiguous case is recoverable from there).

Two new regression tests cover the GitHub-Copilot-shaped multi-variant case and the conflict-resolution behavior.

henrypark133:
Note: #2227 was opened to fix the same MCP tool name hyphen/underscore mismatch. Closing it in favor of this PR since `mcp_tool_id()` here is more comprehensive (handles dots, colons, unicode, not just hyphens).

However, #2227 includes additional fixes not covered here that should be incorporated:
- **Registry**: `resolve_name()`/`get_resolved()`/`unregister()` bidirectional alias resolution via `resolve_key()`
- **WASM loaders**: `load_from_dir` and `discover_tools`/`discover_channels` normalize hyphenated filenames
- **Factory**: `server.name` normalized before all branches (including OAuth early-return)
- **OAuth migration**: `legacy_token_secret_name()` fallback in `is_authenticated()` so existing tokens under hyphenated names aren't lost on upgrade
- **`tool_definitions_for()`**: uses alias resolution instead of exact-key lookup

See https://github.com/nearai/ironclaw/pull/2227 for the full diff.

henrypark133 (commented):
## Code Review Summary

| Severity | Count |
|----------|-------|
| Critical | 2 |
| Warning | 7 |
| Nit | 2 |

### Critical

**C1. `starts_with` filter in `manager.rs` uses raw server name — hyphenated MCP servers return empty tool lists and fail to unregister** (`src/extensions/manager.rs`)
- `mcp_tool_id` normalizes prefixes (`my_server_ping`) but `starts_with(&format!("{}_", name))` uses the raw name (`my-server_`). Affects 4 call sites: already-active fast path, `list()`, `remove()`, and `provider_action_names()`.
- Fix: `format!("{}_", name.replace('-', "_"))` at all sites, or use `mcp_tool_id(name, "")` as the prefix.

**C2. `needs_top_level_flatten` over-flattens schemas with array `"type"` field** (`src/llm/rig_adapter.rs`)
- `"type": ["object", "null"]` (produced by `make_nullable`) does not match `Some(JsonValue::String(s)) if s == "object"`, so the schema gets flattened and all properties are discarded.
- Fix: extend the match to accept array types containing `"object"`.

### Warning

**W1.** LibSQL `save_version` uses raw SQL `BEGIN IMMEDIATE`/`COMMIT` instead of the libsql transaction API — inconsistent with `replace_chunks` which correctly uses `conn.transaction_with_behavior()`. (`src/db/libsql/workspace.rs`)

**W2.** Postgres deadlock window between concurrent `save_version` and `replace_chunks` on the same document — both take `FOR UPDATE` on the same row in separate transactions. (`src/workspace/repository.rs`)

**W3.** `e2e_live.rs` replay tests (`zizmor_scan`) reference trace fixtures not committed to the repo — will panic at `trace_fixture_path` lookup.

**W4.** `drive_auth_gate_roundtrip` sends raw OAuth token as plaintext into LLM context via `send_message`. `with_no_trace_recording()` only covers fixture files, not in-memory exposure. (`tests/e2e_live.rs:436`)

**W5.** Promised "exactly 2 LLM calls" invariant is commented but never asserted — the very regression this test was designed to prevent goes uncaught. (`tests/e2e_live.rs:451-558`)

**W6.** `warn!` for dropped image attachments corrupts REPL/TUI per CLAUDE.md logging rules — should be `debug!`. (`src/llm/openai_codex_provider.rs:426`)

**W7.** PR does NOT address several gaps identified in #2227: registry `resolve_name`/`get_resolved` bidirectional alias resolution, WASM loader normalization, factory OAuth path normalization, OAuth token legacy fallback for hyphenated names. These should be incorporated from #2227 or done as a follow-up.

### Nit

**N1.** Collision detection in `create_tools` does not update `seen_ids` for the colliding entry — third collision reports against first, not second.

**N2.** `normalize_tool_name` (`proxy_` strip) lacks a caller-level integration test per project testing rules.

### Open Questions
- Are the `zizmor_scan` trace fixtures being committed in a follow-up?
- Is the Postgres deadlock risk acceptable for the initial fix, with an advisory-lock follow-up?

### Test Coverage Notes
- Bug 5 (post-flight auth detector) has the weakest coverage — only exercised in live-mode E2E, no hermetic integration test for `detect_post_call_auth_failure`.
- `normalize_tool_name` helper needs a caller-level test per project guidelines.

src/tools/mcp/client.rs
```diff
@@ -652,6 +652,24 @@ fn extract_server_name(url: &str) -> String {
...
+}
```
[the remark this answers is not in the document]
ilblackdragon: Partially addressed via f2f9f16f, but the structured-keys suggestion in its strongest form isn't feasible without rewriting the registry.

The tool registry is fundamentally `HashMap<String, Arc<dyn Tool>>` (see `src/tools/registry.rs`), and the LLM tool-name surface is also a flat string (OpenAI's tool name regex is `^[a-zA-Z0-9_-]{1,64}$`). Switching to a structured key would require touching ~20 call sites across the registry, the LLM provider boundary, the v2 effect adapter, and the persistence layer — and at the LLM boundary we'd still have to flatten back to a string anyway.

The real concern (silent shadowing on collision) is now addressed at the right layer: f2f9f16f adds collision detection in `McpClient::create_tools` itself, where we still have both the original tool name and the normalized id, and emits a `tracing::warn!` when two distinct originals collide. Operators get a breadcrumb pointing at exactly which upstream tools to rename. The normalization-then-string-key approach is preserved but the silent-shadow trap is gone.

src/bridge/router.rs
ilblackdragon: Fixed in 09c64639 — extracted `async fn resolve_extension_for_action(tools, action_name, credential_fallback) -> String` with the full rationale doc. `resolve_auth_gate_display_name` is now a thin wrapper that handles the non-Authentication case, and both inline sites (`resolve_gate` line ~1603, `await_thread_outcome` line ~2779) now call the helper directly with the destructured `credential_name`. The two 15-line comment blocks at those sites are collapsed into short "see helper for full rationale" pointers. Three sites → one implementation. Good nit.

crates/ironclaw_engine/src/executor/orchestrator.rs
ilblackdragon: Fixed in cecdcb60 — replaced `unwrap_or_else(|_| Value::Null)` with `filter_map` so a serialization failure drops the entry from the output rather than injecting a `null` that would crash Python's `.get("name")` / `.get("call_id")` / `.get("params")` accesses. The warn log is preserved and now includes `action_name` for correlation. The entry is dropped (not the entire call list), and Python's tool-result loop iterates `range(len(results))` against the same shortened call list, so a missing entry is benign.

You're right that the `unwrap_or_else(|_| Value::Null)` contradicted the PR's own stated goal. The failure path is essentially unreachable for this struct shape (String + String + Value), but the contract should still be correct — the helper will be touched again when the interchange shape evolves and we don't want the next person to discover this trap the hard way.

src/tools/mcp/client.rs
ilblackdragon: Fixed in f2f9f16f — added collision detection in `McpClient::create_tools`. The helper builds a `HashMap<normalized_id, original_name>` while iterating the tool list and emits a `tracing::warn!` when two distinct originals collide on the same id. The warn carries the normalized id, both colliding original names, and the server name, so an operator can immediately see which upstream tools to rename.

Behaviour is intentionally unchanged (last-write-wins via `register`) because the LLM normalizes both names to the same string anyway — there's no way to distinguish them from the caller side. The warn is the observable signal that the collision happened, which is exactly the gap you flagged.

The detection is scoped to a single MCP server's tool list because cross-server collisions have different registry-key prefixes and can't shadow each other. New regression test `test_create_tools_handles_post_normalization_collision` exercises the `search-all` vs `search_all` case end-to-end through a real `ToolRegistry` and asserts last-write-wins.

crates/ironclaw_engine/src/executor/orchestrator.rs
ilblackdragon: Fixed in 8278ab24 — replaced `value = %value` with `shape = %summarize_action_calls_for_log(value)`, a new helper that emits only structural info: array length and the keys of the first entry. The keys themselves are static field names (`name`, `call_id`, `params`), not user data. The full parse error is still logged verbatim since that's the diagnostic operators actually need.

Edge cases handled: empty array → "empty array", non-array (Python passed wrong type) → "non-array value of type <kind>" via a small `json_value_type_name` helper, entries that aren't objects → `<not an object>`.

Two regression tests pin the PII-safety contract:
- `summarize_action_calls_for_log_does_not_leak_user_pii` builds an intentionally PII-laden value (salary spreadsheet queries, credentials, "private message about layoffs") and asserts none of the user-content strings appear in the summary.
- `summarize_action_calls_for_log_handles_edge_cases` pins the empty/string/object/null fallback paths.

Good catch — the warn only fires on shape drift, which is exactly when operators are grepping logs and shipping output to aggregators.

ilblackdragon:
## Incorporated #2227

Pushed `a8530e8b` which cherry-picks the non-overlapping changes from #2227 (`fix-tool-name-hyphen-normalization`) so this PR supersedes it. Our PR already had the core fix (`mcp_tool_id` canonicalization with the stricter all-non-identifier-char normalization), but #2227 adds six defense-in-depth and compatibility pieces we didn't cover:

| Layer | What #2227 adds | Why it matters |
|---|---|---|
| `tools/mcp/factory.rs` | Normalize `server.name` at the factory boundary before any branch | Secret names, session keys, and tool prefixes all use the same underscore form. Our PR only normalized at the tool-id level. |
| `tools/mcp/client.rs` | Normalize `new_with_name` | Callers bypassing the factory get consistent server_name |
| `tools/mcp/config.rs` + `auth.rs` | Legacy token secret name fallback | Existing users who stored tokens under the hyphenated server name don't have to re-auth after upgrade |
| `tools/registry.rs` | Bidirectional `resolve_key` (exact → hyphen→underscore → underscore→hyphen) | Defense in depth: edge-case tools registered with mixed separators still resolve |
| `tools/wasm/loader.rs` | Filesystem loader normalizes hyphenated filenames | Consistent tool names when loading from disk |
| `channels/wasm/loader.rs` | Same for channel loader | Consistent channel names |

Conflict resolution:
- `manager.rs`: kept our version (same code, our version has comments)
- `client.rs` `create_tools`: kept our `mcp_tool_id` (handles ALL non-identifier chars, not just dashes)
- `client.rs` tests: kept our comprehensive MockTransport suite, dropped #2227's simpler duplicate

All gates green: **5382 passed, 0 failed**, zero clippy warnings.

This PR now fully supersedes #2227 — @ilblackdragon can close it when ready.

src/config/mod.rs
```diff
@@ -127,6 +127,27 @@ pub struct Config {
...
 
+/// Generate a fresh random AES-256-GCM master key for `Config::for_testing`.
+///
+/// Returns a hex-encoded 32-byte key (64 hex chars), satisfying the length
+/// check in `SecretsConfig::resolve`. Each call returns a different value —
+/// tests don't need cross-process determinism (each test builds a fresh
+/// secrets store on top of a fresh temp DB), and committing a constant
+/// master key into the source tree would mean every developer who built
+/// with `--features libsql` had a publicly-known key in their process.
+#[cfg(feature = "libsql")]
```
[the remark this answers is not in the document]
ilblackdragon: Good catch. The PR description was stale from the original hardcoded-key version that was replaced in `5a6a3ac0`. Updated the PR body to say "random 32-byte AES master key (generated fresh per call)" instead of "deterministic". The implementation has been correct since commit `5a6a3ac0`.

ilblackdragon:
## Addressing @henrypark133's second review (C1/C2/W1-W7/N1/N2)

Pushed 3 fixup commits for the actionable items:

| # | Severity | Commit | What |
|---|---|---|---|
| C1 | 🔴 | `bcdfdc8e` | All 3 `starts_with(&format!("{}_", name))` sites in `manager.rs` now use `crate::tools::mcp::mcp_tool_id(name, "")` as the prefix. This produces `my_server_` from `my-server`, matching the normalized registry keys. Affects the already-active fast path, `list()`, and `remove()`. |
| C2 | 🔴 | `4f2bda3e` | `needs_top_level_flatten` now accepts `"type": ["object", "null"]` (any JSON array containing `"object"`) in addition to the plain string form. A nullable object schema no longer gets silently flattened. Regression test added. |
| N1 | 🟢 | `2ac2af2a` | `seen_ids.insert` added inside the collision warning branch so 3rd+ collisions report against the most recent shadow. |

### Items not requiring code changes

**W1 (libsql raw SQL in save_version):** Pre-existing pattern not introduced by this PR. The inconsistency is real but out of scope.

**W2 (Postgres deadlock between save_version and replace_chunks):** This is a serialization wait, not a deadlock. Two transactions taking `FOR UPDATE` on the same row in separate transactions will *serialize* (second waits for first to commit) — there's no circular dependency that would cause a deadlock. Postgres detects actual deadlocks (circular wait across multiple rows) and kills one transaction automatically; that can't happen here because both transactions lock exactly one row (`memory_documents.id`) and never lock a second resource while holding the first.

**W3 (zizmor_scan trace fixtures):** Pre-existing test infrastructure committed in the base branch. Not introduced by this PR.

**W4 (OAuth token in LLM context):** The live test is `#[ignore]`d and only runs manually by the developer against their own credentials. The token is sent as the auth resolution (simulating what the user would paste). `with_no_trace_recording()` ensures no fixture is committed. The in-memory exposure is the same as what happens in normal production use when a user pastes a token.

**W5 ("exactly 2 LLM calls" invariant):** The comment describes the expected flow, not a contract. The test asserts the two outcomes that matter: (1) `StatusUpdate::AuthRequired` is surfaced in Phase A, (2) the resume path ran in Phase B (visible via additional tool activity + response). Asserting exact call count would make the test brittle to unrelated LLM-provider changes.

**W6 (warn! for image attachments):** Pre-existing code at `openai_codex_provider.rs:426` not touched by this PR. Our diff to that file only adds the schema flatten logic. The logging-level fix is a valid separate concern but out of scope.

**W7 (#2227 gaps):** Fully addressed in `a8530e8b` — registry bidirectional alias, WASM/channel loader normalization, factory OAuth path normalization, and legacy token fallback are all incorporated.

**N2 (normalize_tool_name caller-level test):** The `proxy_` strip logic in `rig_adapter.rs` is pre-existing code not touched by this PR. A caller-level test for it is a good idea but out of scope.

### Verification

All gates green: **5387 passed, 0 failed, 5 ignored**, zero clippy warnings, clean fmt.

src/tools/wasm/wrapper.rs
```diff
@@ -1086,6 +1079,33 @@ impl WasmToolWrapper {
...
+                WasmError::Trapped("unreachable code executed".to_string())
+            }
+            // Everything else: include trap kind + full chain for diagnosis
+            other => WasmError::Trapped(format!("{other}: {error:#}")),
+        };
+    }
+
+    // No Trap downcast (host error, component-model glue, etc.): full chain
+    WasmError::Trapped(format!("{error:#}"))
+}
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in cf621e8f — three changes:

1. **CI fix:** `classify_trap_error` now takes `wasmtime::Error` directly (not `anyhow::Error`). wasmtime 43 has its own `Error` type in `wasmtime_internal_core` that's distinct from `anyhow::Error`, causing the E0308 type mismatch on CI. Accepting the native type is also more correct — it preserves type info for the downcast.

2. **String-matching fallback:** Added the chain-walk fallback you suggested. After the structured downcast, we now also check `format!("{error:#}")` for `"all fuel consumed"`, `"out of fuel"`, `"OutOfFuel"`, and `"unreachable"`. This covers the case where component-model glue wraps the `Trap` in layers that `downcast_ref` can't see through. Doc comment updated to match.

3. **New test:** `trap_classification_fuel_via_string_fallback` wraps the fuel message in a plain `wasmtime::Error::msg` (so downcast returns `None`) and asserts `FuelExhausted` is still detected via string matching.

tests/support/test_rig.rs
```diff
@@ -1258,6 +1274,26 @@ impl TestRig {
         self.trace_llm.as_ref()
     }
 
+    /// Get the secrets store for direct credential manipulation.
+    /// Used by live tests that exercise the auth gate flow — they
+    /// delete a credential to simulate "not yet authenticated", then
+    /// re-insert it after the gate fires to simulate "user completed
+    /// OAuth and the token was stored". Returns `None` when the rig
+    /// was built without a master key (most non-live tests).
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in cf621e8f — updated the `secrets_store()` doc comment to reflect that most test rigs now have a working secrets store because `Config::for_testing()` generates a random master key per call. `None` only occurs when a config override explicitly disables secrets or omits a master key.

src/llm/rig_adapter.rs
```diff
@@ -123,22 +123,233 @@ fn round_f32_to_f64(val: f32) -> f64 {
...
+/// untouched.
+///
+/// Note on Anthropic: this normalizer is also applied to Anthropic via
+/// `RigAdapter::convert_tools`. Anthropic accepts top-level `oneOf` natively,
+/// so the flatten is slightly lossy for Claude users on tools that have a
+/// top-level union, but the description hint preserves the variant info and
+/// Claude is good at reading schemas out of free text. Keeping a single
+/// normalizer for all rig-based providers is simpler than threading
+/// per-provider flags through the adapter.
+pub(crate) fn normalize_schema_strict(schema: &JsonValue, description: &mut String) -> JsonValue {
```
serrrfirat, line 171: **Medium Severity — Lossy schema flattening applied to Anthropic**

`normalize_schema_strict` is applied to all rig-based providers including Anthropic, which natively accepts top-level `oneOf`/`anyOf`. This unnecessarily discards structured schema information for Claude, replacing it with a text-in-description approximation.

For Anthropic users with complex MCP tool schemas, the LLM loses access to structured schema validation and relies on reading the description text instead.

**Suggested fix:** Add a `flatten_top_level: bool` parameter or check the provider type to skip flattening for Anthropic. Track as tech debt if not fixing now.

src/tools/wasm/wrapper.rs
```diff
@@ -1086,6 +1079,63 @@ impl WasmToolWrapper {
...
+    // contains the diagnostic string, so we check for the two traps that
+    // have distinct WasmError variants.
+    let error_str = format!("{error:#}");
+    if error_str.contains("all fuel consumed")
+        || error_str.contains("out of fuel")
+        || error_str.contains("OutOfFuel")
+    {
+        return WasmError::FuelExhausted { limit: limits.fuel };
+    }
+    if error_str.contains("unreachable") {
```
serrrfirat: **Medium Severity — Overly broad WASM trap string match**

The string fallback `error_str.contains("unreachable")` is too broad. A host function error message like `"the endpoint was unreachable"` or `"server unreachable: connection refused"` would be misclassified as `Trap::UnreachableCodeReached`, losing the real error chain and replacing it with a generic `"unreachable code executed"` string.

**Suggested fix:** Use a more specific match like `"unreachable code"`, `"wasm trap: unreachable"`, or `"UnreachableCodeReached"` to reduce false positives.

src/tools/mcp/auth.rs
```diff
@@ -1194,7 +1194,16 @@ pub async fn is_authenticated(
     .await
     {
         Ok(Some(_)) => true,
-        Ok(None) => false,
+        Ok(None) => {
+            // Fall back to legacy (pre-hyphen-normalization) secret name
+            // so existing users with tokens stored under hyphenated server
+            // names are not forced to re-authenticate after upgrade.
+            if let Some(legacy_name) = server_config.legacy_token_secret_name() {
+                secrets.get_decrypted(user_id, &legacy_name).await.is_ok()
```
serrrfirat, line 1209: **Medium Severity — Legacy token fallback skips refresh**

The legacy token fallback in `is_authenticated` uses bare `secrets.get_decrypted(user_id, &legacy_name).await.is_ok()`. Unlike the primary path which goes through `resolve_access_token_string_with_refresh` (handling refresh), the legacy path does a bare `get_decrypted`.

If a legacy access token is expired but has a valid refresh token under the legacy naming scheme, the user will be forced to re-authenticate — the refresh token won't be tried.

**Suggested fix:** Acceptable as a transitional compat path (users will re-auth once and get migrated), but should be documented explicitly.

src/llm/rig_adapter.rs
```diff
@@ -123,22 +123,233 @@ fn round_f32_to_f64(val: f32) -> f64 {
...
+    // OpenAI has no documented hard limit on tool description length, but
+    // long descriptions waste prompt budget on every turn. 1500 bytes fits a
+    // typical MCP dispatcher schema and still leaves room for the original
+    // tool description above it.
+    const SCHEMA_HINT_MAX_BYTES: usize = 1500;
+
+    let detected = detect_forbidden_top_level(parameters);
+    let merged_properties = merge_top_level_variant_properties(parameters);
+
+    if let Ok(original_text) = serde_json::to_string(parameters)
```
serrrfirat: **Medium Severity — Potential DoS via large MCP tool schema**

`flatten_top_level` serializes the entire original schema to a string and appends it to the tool description. A malicious MCP server could advertise a tool with a very deeply nested or large schema (many MB). The 1500-byte truncation limit only kicks in *after* the full `serde_json::to_string()` allocation. The merged-properties loop also iterates all variants unboundedly.

**Suggested fix:** Add a pre-check on schema depth/size before serializing, e.g., limit recursion depth during merge, or cap the schema `Value` size heuristically.

serrrfirat (commented):
**Code LGTM — approve once CI (Clippy) is green.**

Reviewed all 26 changed files. This is a well-crafted PR. The TOCTOU race fix uses correct transaction isolation primitives for both libsql and Postgres backends. Auth gate resolution is properly threaded through all call sites. Tool naming normalization is consistent. WASM fuel limit increase is well-justified.

Four Medium findings posted inline, all non-blocking:
1. **Lossy schema flattening for Anthropic** (`rig_adapter.rs:171`) — Anthropic natively supports `oneOf`/`anyOf` but gets the same flatten treatment. Acknowledged in code comment, intentional tradeoff.
2. **Broad "unreachable" string match** (`wrapper.rs:1131`) — could misclassify host errors, but only in the rare fallback path when structured downcast also fails.
3. **Legacy token refresh gap** (`auth.rs:1202`) — acceptable transitional compat path, users re-auth once and get migrated.
4. **Schema DoS via large MCP tool schema** (`rig_adapter.rs:326`) — full `serde_json::to_string()` allocation before 1500-byte truncation. Low likelihood.

No critical or high-severity issues found. Cannot formally approve while CI is red (4 Clippy failures). Fix those and this is ready to merge.

src/llm/rig_adapter.rs
ilblackdragon: Already documented in the `normalize_schema_strict` doc comment (lines 164-170): "Note on Anthropic: this normalizer is also applied to Anthropic via `RigAdapter::convert_tools`. Anthropic accepts top-level `oneOf` natively, so the flatten is slightly lossy...". A per-provider flag would require plumbing the provider identity through `RigAdapter` which is generic over `CompletionModel` — tracking as follow-up tech debt, not a silent skip.

src/tools/wasm/wrapper.rs
ilblackdragon: Fixed in 2e65f2b1 — tightened from `contains("unreachable")` to `contains("unreachable code")` / `"UnreachableCodeReached"` / `"wasm trap: unreachable"`. HTTP errors like "endpoint was unreachable" no longer false-positive.

src/tools/mcp/auth.rs
ilblackdragon: Documented in 2e65f2b1 — added an explicit comment at the fallback site explaining the intentional bare `get_decrypted` (no refresh). The legacy path is transitional: users re-auth once and get migrated to the canonical name. Wiring refresh through the legacy naming scheme adds complexity for a self-healing compat layer.

src/llm/rig_adapter.rs
ilblackdragon: Fixed in 2e65f2b1 — added `count_json_nodes` pre-check (cheap recursive walk, no allocation) gating the `serde_json::to_string` call. Schemas over 5000 nodes skip serialization entirely and get a "(schema too large to inline: ~N nodes)" placeholder. A malicious MCP server with a many-MB schema no longer triggers a proportional allocation.

src/llm/rig_adapter.rs
```diff
@@ -123,22 +123,261 @@ fn round_f32_to_f64(val: f32) -> f64 {
...
+}
+
+/// Cheaply estimate the number of nodes in a `serde_json::Value` tree.
+///
+/// Used by `flatten_top_level` to gate the `serde_json::to_string` call:
+/// serializing a many-MB schema (from a malicious or extremely verbose MCP
+/// server) would allocate proportionally even though we only keep 1500 bytes
+/// of the output. This recursive walk counts every leaf and container node
+/// and returns early once it exceeds the caller's budget (no full traversal
+/// needed for the rejection case).
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in 96d611b7 — replaced `count_json_nodes` entirely with `serialize_json_capped(value, max_bytes)`, a `serde_json::to_writer` call through a `CappedWriter` that silently discards bytes past the budget. This bounds the actual heap allocation to `max_bytes` regardless of schema shape — handles both many-node deep recursion AND few-node schemas with multi-MB string values (the gap the node counter missed). Doc, code, and behavior all match now.

src/llm/rig_adapter.rs
```diff
@@ -123,22 +123,284 @@ fn round_f32_to_f64(val: f32) -> f64 {
...
```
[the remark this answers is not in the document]
ilblackdragon: Same fix as above — 96d611b7. The `count_json_nodes` function and its inaccurate doc are both gone.

src/llm/rig_adapter.rs
```diff
@@ -123,22 +123,284 @@ fn round_f32_to_f64(val: f32) -> f64 {
...
+                original_text
+            };
+            description.push_str(schema_flatten_hint_intro(detected));
+            description.push_str(&hint);
+        }
+    } else {
+        description.push_str(schema_flatten_hint_intro(detected));
+        description.push_str(&format!(
+            "(schema too large to inline: ~{node_count} nodes. See upstream MCP server for the full schema.)"
+        ));
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in 96d611b7 — the node-count approach is replaced with `serialize_json_capped`, which writes through a size-limited `Write` impl that stops storing bytes once `SCHEMA_HINT_MAX_BYTES` is reached. `serde_json::to_writer` sees `Ok(data.len())` and continues (minimal remaining work since output is discarded), so the allocation is bounded to 1500 bytes regardless of schema structure. Multi-MB string values in few-node schemas are now handled correctly.

src/tools/mcp/factory.rs
```diff
@@ -26,12 +26,18 @@ pub enum McpFactoryError {
...
     process_manager: &Arc<McpProcessManager>,
     secrets: Option<Arc<dyn SecretsStore + Send + Sync>>,
     user_id: &str,
 ) -> Result<McpClient, McpFactoryError> {
+    // Normalize hyphens to underscores in the server name so that all code
+    // paths (Stdio, Unix, HTTP, OAuth) produce consistently underscore-only
+    // tool prefixes.  This must happen before any branch so that the OAuth
+    // early-return via `McpClient::new_authenticated(server, ..)` also
+    // receives the normalised name.
+    server.name = server.name.replace('-', "_");
```
serrrfirat, line 40: **High Severity**\n\nNormalizing `server.name` here fixes the registry key shape, but the rest of the MCP lifecycle still does exact config lookups by name. `activate_mcp()` / `auth_mcp()` fetch the stored config with `get_mcp_server(name, user_id)`, while `provider_extension_for_tool()` now returns the normalized underscore form from the registered tool wrapper.\n\nThat leaves a regression for existing hyphen-named MCP installs: readiness can resolve `my_server_*` back to provider extension `my_server`, then `ensure_extension_ready()` routes into `auth_mcp("my_server")` / `activate_mcp("my_server")`, and those exact lookups miss the persisted `my-server` config. In practice the extension is detected as installed via the legacy alias path, but the auth/activate step still fails as `NotInstalled`.\n\nSuggested fix: make MCP config lookup/update/remove alias-aware as well, or have `determine_installed_kind()` return the resolved stored name and thread that canonical name through `auth()` / `activate()` / `configure()`.

src/tools/mcp/auth.rs
```diff
@@ -1194,7 +1194,23 @@ pub async fn is_authenticated(
...
+            // so existing users with tokens stored under hyphenated server
+            // names are not forced to re-authenticate after upgrade.
+            //
+            // Intentionally uses bare `get_decrypted` (no refresh) — the
+            // legacy path is transitional. Users whose token is expired
+            // will re-auth once and get migrated to the canonical name.
+            // Wiring `resolve_access_token_string_with_refresh` through the
+            // legacy naming scheme adds complexity for a path that
+            // self-heals after one re-auth cycle.
+            if let Some(legacy_name) = server_config.legacy_token_secret_name() {
```
serrrfirat, line 1208: **High Severity**\n\nThis fallback only fixes the *status* path (`is_authenticated`). The live request path still goes through `McpClient::get_access_token()` / `refresh_access_token()`, which only read `token_secret_name()` / `refresh_token_secret_name()` on the normalized name.\n\nThat means an upgraded user with legacy secrets like `mcp_my-server_access_token` can now look authenticated here, but the actual MCP call still sends no `Authorization` header and falls back into re-auth on first use. So the upgrade regression is only half-fixed: the server appears ready, but real tool execution still breaks until the user re-authenticates.\n\nSuggested fix: centralize token resolution so both readiness checks and request-time auth use the same legacy-aware access-token and refresh-token lookup path.

src/workspace/mod.rs
```diff
@@ -2183,12 +2172,14 @@ impl Workspace {
...
-            self.storage
-                .insert_chunk(document_id, index as i32, &content, embedding.as_deref())
-                .await?;
+            writes.push(ChunkWrite { content, embedding });
         }
 
+        // One transaction: DELETE + N INSERTs. Closes the TOCTOU race where
+        // two concurrent reindexers for the same document could both delete,
+        // then both re-insert chunk_index 0 and hit the UNIQUE constraint.
+        self.storage.replace_chunks(document_id, &writes).await?;
```
serrrfirat, line 2220: **High Severity**\n\n`replace_chunks()` fixes the chunk `UNIQUE (document_id, chunk_index)` collision, but it still does not make document writes and search-index updates atomic. `Workspace::write()` updates `memory_documents.content` first, then calls this helper afterwards. This helper immediately re-reads the document and replaces chunks in a separate transaction.\n\nUnder concurrent writes, writer B can win the `update_document()` race while writer A wins the later `replace_chunks()` transaction, leaving final document content from B but search chunks from A. The new regression test only checks chunk count / contiguous indexes, so it would still pass with that skewed final state.\n\nSuggested fix: move document-content update + chunk replacement under one per-document transaction, or gate `replace_chunks()` on a content/version hash so stale reindexers abort instead of overwriting newer chunks.

src/llm/rig_adapter.rs
```diff
@@ -123,22 +123,319 @@ fn round_f32_to_f64(val: f32) -> f64 {
...
+        buf: Vec::with_capacity(max_bytes.min(8192)),
+        max: max_bytes,
+    };
+    let mut ser = serde_json::Serializer::new(writer);
+    serde::Serialize::serialize(value, &mut ser).map_err(|_| ())?;
+    let buf = ser.into_inner().buf;
+    // The buffer is guaranteed to be valid UTF-8 because serde_json only
+    // emits ASCII structural characters and JSON-escaped unicode.
+    String::from_utf8(buf).map_err(|_| ())
+}
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in 454012a8 — `String::from_utf8` now falls back to `e.valid_up_to()` on UTF-8 error, trimming to the last complete codepoint instead of dropping the entire hint. serde_json v1 does emit raw UTF-8 for non-ASCII characters, so the byte cap can indeed cut mid-codepoint. Good catch.

crates/ironclaw_engine/src/executor/orchestrator.rs
```diff
@@ -1903,6 +1911,164 @@ fn build_orchestrator_inputs(
...
+                .map(|obj| {
+                    let mut keys: Vec<&str> = obj.keys().map(String::as_str).collect();
+                    keys.sort_unstable();
+                    keys.join(",")
+                })
+                .unwrap_or_else(|| "<not an object>".to_string());
+            format!(
+                "array of {} entries; first entry keys: [{}]",
+                arr.len(),
+                first_keys
```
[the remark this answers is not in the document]
ilblackdragon: Not fixing — false positive. The keys logged are JSON Schema property names from the `PythonActionCall` interchange dict (`name`, `call_id`, `params`), not user tool parameter names. The warn only fires when the Python↔Rust shape drifts (wrong field structure), at which point the first entry's keys reveal the unexpected shape for diagnosis. If the array contained raw tool-parameter objects (which would mean a fundamentally different bug than shape drift), the test `summarize_action_calls_for_log_does_not_leak_user_pii` would have caught it.

src/workspace/repository.rs
```diff
@@ -329,6 +329,89 @@ impl Repository {
...
+        // (the libsql variant of the same regression test).
+        tx.execute(
+            "SELECT 1 FROM memory_documents WHERE id = $1 FOR UPDATE",
+            &[&document_id],
+        )
+        .await
+        .map_err(|e| WorkspaceError::ChunkingFailed {
+            reason: format!("Acquire row lock failed: {e}"),
+        })?;
+
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in 454012a8 — now checks the returned row count from `SELECT ... FOR UPDATE` and returns a clear `ChunkingFailed` error when it's 0 (document not found, no lock acquired).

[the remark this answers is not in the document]
ilblackdragon: Same fix as above — 454012a8. Row count check added with a descriptive error message.

src/tools/mcp/factory.rs
ilblackdragon: Fixed in 454012a8 — `get_mcp_server` is now alias-aware: tries exact name → hyphen alias (underscores→hyphens) → underscore alias (hyphens→underscores). This closes the gap where `provider_extension_for_tool` returned the normalized `my_server` form but `get_mcp_server("my_server")` missed the persisted `my-server` config, causing `activate_mcp` to fail with `ServerNotFound`.

The config persistence (add/update/remove) still uses the stored name as-is. New installs go through the factory which normalizes the name before storage. Legacy installs are found via the alias lookup. Full config-layer normalization (migrating stored names) is a follow-up.

src/tools/mcp/auth.rs
ilblackdragon: Fixed in 454012a8 — `get_access_token` now falls back to the legacy (pre-normalization) secret name when the canonical name returns no token. This closes the gap where `is_authenticated` reported true (via its own legacy fallback) but the actual request sent no Authorization header because the token lookup only used the canonical name.

The fallback uses bare `get_decrypted` (no refresh) — intentionally matching the transitional semantics documented in the `is_authenticated` fallback. Users re-auth once and get migrated to the canonical name. Centralizing token resolution with refresh across both naming schemes is a larger auth refactor for a follow-up.

src/workspace/mod.rs
ilblackdragon: Fixed in 454012a8 — `reindex_document_with_metadata` now captures the content hash at read time and re-verifies it after computing embeddings. If another writer updated the document content during the embedding window, the reindexer skips chunk replacement and returns Ok (the other writer's reindex call produces correct chunks for the new content).

This is an optimistic concurrency check that closes the content-vs-chunks skew: writer B can't win the document UPDATE race while writer A overwrites chunks from stale content. The full fix (single transaction wrapping write + reindex) would require refactoring the Workspace API to accept a transaction handle, which is a bigger change for a follow-up.

src/workspace/mod.rs
```diff
@@ -2183,12 +2181,36 @@ impl Workspace {
...
+                    document_id = %document_id,
+                    "Skipping chunk replacement — document content changed during embedding computation"
+                );
+                return Ok(());
+            }
+            Err(_) => {
+                // Document was deleted while we were computing embeddings.
+                // Nothing to reindex.
+                return Ok(());
+            }
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in 6d3ebced — narrowed from `Err(_)` to `Err(WorkspaceError::DocumentNotFound { .. })`. Transient DB errors now propagate instead of being silently swallowed as "document deleted".

tests/support/live_harness.rs
```diff
@@ -184,51 +90,16 @@ impl LiveTestHarness {
...
 
-    /// Multi-turn variant of `finish`.
-    pub async fn finish_turns(self, turns: &[SessionTurn]) {
+    /// Variant of [`finish`] for tests that span multiple user turns
+    /// (e.g. an auth-gate roundtrip: prompt → AuthRequired → token →
+    /// resume). Each tuple is `(user_input, responses_after_that_turn)`,
+    /// and the session log shows them in order so a reader can follow
+    /// the full conversation rather than only the first prompt.
+    pub async fn finish_turns(self, turns: &[(String, Vec<String>)]) {
         self.save_session_log(turns);
```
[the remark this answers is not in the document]
henrypark133: Fixed in c2869c92 — updated `e2e_live_personas.rs` to use the new `finish_turns` API with `(String, Vec<String>)` tuples instead of the removed `SessionTurn` + `finish_turns_strict`. The non-strict variant is the correct replacement since the strict checking was removed from the harness.

tests/support/live_harness.rs
```diff
@@ -388,9 +257,9 @@ pub struct LiveTestHarnessBuilder {
     max_tool_iterations: usize,
     engine_v2: Option<bool>,
     auto_approve_tools: Option<bool>,
-    skills_dir: Option<PathBuf>,
     channel_name: Option<String>,
     seeded_secret_names: Vec<String>,
+    record_trace: bool,
 }
 
 impl LiveTestHarnessBuilder {
```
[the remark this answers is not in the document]
henrypark133: Fixed in c2869c92 — removed the `.with_skills_dir()` call from `build_persona_harness()` in `e2e_live_personas.rs` and the now-unused `repo_skills_dir()` helper. Skills are loaded from the default location by the test rig.

tests/support/test_rig.rs
```diff
@@ -302,13 +311,6 @@ impl TestRig {
...
-    /// Return the filtered list of captured responses so far.
-    ///
-    /// Mirrors the bootstrap-greeting filtering used by `wait_for_responses`.
-    pub async fn captured_responses(&self) -> Vec<OutgoingResponse> {
-        self.filter_responses(self.channel.captured_responses_async().await)
-    }
-
     /// Return `(name, success)` for all `ToolCompleted` events captured so far.
     pub fn tool_calls_completed(&self) -> Vec<(String, bool)> {
         self.channel.tool_calls_completed()
```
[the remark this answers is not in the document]
henrypark133: Fixed in c2869c92 — replaced `rig.captured_responses().await.len()` with `rig.wait_for_responses(0, Duration::ZERO).await.len()` to get the current response count using the available API. Also added a `collect_active_skill_names()` helper to replace the removed `active_skill_names()` method by filtering `SkillActivated` status events (same pattern already used in `debug_turn()` in the same file).

src/tools/builtin/file_history.rs
```diff
@@ -84,12 +84,40 @@ impl FileHistory {
...
+                if let (Some(parent), Some(name)) = (path.parent(), path.file_name()) {
+                    parent
+                        .canonicalize()
+                        .unwrap_or_else(|_| parent.to_path_buf())
+                        .join(name)
+                } else {
+                    path.to_path_buf()
+                }
+            }
+        };
```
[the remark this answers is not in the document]
henrypark133: Fixed in a92ce09c — replaced the inline canonicalization in `snapshot()` with a call to `Self::canonical(path)`, eliminating the duplication.

tests/support/live_harness.rs
```diff
@@ -602,25 +503,31 @@ impl LiveTestHarnessBuilder {
             .await
             .expect("Failed to build LLM provider chain for live test");
```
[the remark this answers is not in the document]
henrypark133: This is by design for the live test harness. Live tests expect LLM credentials in env vars or `~/.ironclaw/.env` (as documented in the module doc). The `.with_secrets()` builder method handles seeding MCP/WASM credentials from the local secrets DB, but the LLM provider chain is intentionally built from env config — wiring in `inject_llm_keys_from_secrets` would couple the test harness to the full app startup path and add complexity for a scenario that's already covered by the documented test setup.

henrypark133 (approved):

PR #2214, opened 2026-04-09
feat(jobs): MCP client support in sandboxed container workers

### LINKED ISSUE #2180: MCP tools unavailable in sandboxed (Docker) jobs
## Summary

PR #1243 (`feat(jobs): per-job MCP server filtering and max_iterations cap`) added per-job MCP server scoping. The orchestrator correctly generates a filtered `mcp-servers.json` and bind-mounts it into the container at `/home/sandbox/.ironclaw/mcp-servers.json:ro`. However, **nothing in the container worker reads this file** — MCP tools are completely non-functional in sandboxed jobs.

## Root cause

There are four compounding gaps between the orchestrator (which owns MCP infrastructure) and the container worker (which runs the agent loop):

### 1. MCP config is mounted but never loaded

`src/orchestrator/job_manager.rs:448-470` correctly mounts the config. But `WorkerRuntime::new()` (`src/worker/container.rs:92-94`) creates a bare `ToolRegistry` and only calls `register_container_tools()`, which registers dev tools (shell, file read/write, patch). No MCP initialization ever happens.

### 2. No MCP client infrastructure in containers

MCP clients require runtime support that doesn't exist in the worker process:
- `McpProcessManager` (stdio transport process lifecycle)
- `McpSessionManager` (HTTP session state)
- `ExtensionManager` (tool registration bridge)

None of these are instantiated in the container worker.

### 3. Orchestrator API doesn't surface MCP tool definitions

The `/worker/{job_id}/llm/complete_with_tools` endpoint (`src/orchestrator/api.rs:173-203`) is a passthrough — it forwards whatever tool definitions the container sends without augmenting them. The LLM never sees MCP tools.

### 4. No tool execution path for MCP calls

The container's `ToolRegistry` has no MCP tools registered, so even if definitions reached the LLM, returning tool calls would fail to execute.

## What works vs. what doesn't

| Scenario | Works? | Why |
|----------|--------|-----|
| Non-sandboxed job (local worker) | Yes | Agent loop runs in orchestrator process with full MCP infrastructure |
| Sandboxed job (Docker container) | **No** | Container has isolated `ToolRegistry` with only dev tools |
| `mcp_servers` filtering param | Partially | Config file is correctly generated and mounted, but never consumed |

## Proposed fix: in-container MCP clients

MCP clients should be instantiated **inside the container**, not proxied through the orchestrator. This preserves the sandbox security model — if an MCP server returns malicious data or the agent misuses a tool, the blast radius stays within the disposable container (read-only rootfs, dropped capabilities, non-root user, memory limits).

Proxying MCP calls back to the orchestrator would undermine sandboxing: the orchestrator runs with full privileges (all MCP servers, credentials, database access, other jobs), so a compromised agent loop inside the container would effectively get orchestrator-level MCP access via the back-channel.

### Implementation steps

1. **Load the mounted config** — in `WorkerRuntime::new()`, read `/home/sandbox/.ironclaw/mcp-servers.json` if it exists, parse `McpServerConfig` entries
2. **Instantiate MCP clients** — create `McpProcessManager` + `McpClient` per server using the existing `create_client_from_config()` factory in `src/tools/mcp/factory.rs`
3. **Register MCP tools** in the container's `ToolRegistry` alongside the existing dev tools
4. **Per-job network allowlist** — for MCP servers that need outbound network access (HTTP-transport servers, or stdio servers that call external APIs), the orchestrator should extract required domains from the MCP config and pass them to the sandbox proxy (via env var or mount). `src/sandbox/proxy/allowlist.rs` already supports dynamic allowlists. Network policy is configured via the sandbox policy.

### Transport considerations

- **Stdio**: works if the MCP server binary is installed in the container image. The transport itself is local IPC, but the spawned server process may still need outbound network access depending on what it does — this is controlled by the sandbox network policy and proxy allowlist.
- **HTTP/SSE**: requires adding the server's domain to the container's network allowlist and injecting credentials. The existing `GET /worker/{job_id}/credentials` endpoint and sandbox proxy credential injection can be reused.
- **Unix socket**: requires mounting the host socket into the container. Lower priority.

## Affected files

- `src/worker/container.rs` — `WorkerRuntime::new()` needs MCP client initialization
- `src/tools/registry.rs` — `register_container_tools()` needs to accept MCP tools
- `src/orchestrator/job_manager.rs` — extract MCP server domains for network allowlist
- `src/sandbox/proxy/allowlist.rs` — accept per-job allowlist entries from orchestrator

rajulbhatnagar:
## Summary

- Wire up MCP client lifecycle inside Docker container workers so sandboxed jobs can use MCP tools (including OAuth-authenticated servers)
- Add `ProxySecretsStore` that proxies secrets operations to the orchestrator, enabling MCP OAuth token refresh inside containers
- Add 3 new orchestrator endpoints (`/mcp/secrets/{get,exists,create}`) gated by per-job exact-match `HashSet` allowlist
- Make `mcp_config_path` configurable and inject `IRONCLAW_USER_ID` into container env

Closes #2180

## Security model

Closed-world allowlist: at job creation, the orchestrator pre-computes the exhaustive set of permitted MCP secret names (5 per OAuth server) from the mounted config. The `HashSet` is stored in `TokenStore` and checked on every request — no prefix matching, no string parsing. Revoked automatically on job cleanup.

## Test plan

- [x] `cargo fmt` — clean
- [x] `cargo test --lib` — 4352 passed, 0 failed
- [x] Manual: create sandboxed job with `mcp_servers` param, verify MCP tools appear in agent loop
- [x] Manual: create sandboxed job with OAuth MCP server, verify token refresh works

### DIFF
--- a/src/orchestrator/api.rs
+++ b/src/orchestrator/api.rs
@@ -75,6 +75,18 @@ impl OrchestratorApi {
             .route("/worker/{job_id}/event", post(job_event_handler))
             .route("/worker/{job_id}/prompt", get(get_prompt_handler))
             .route("/worker/{job_id}/credentials", get(get_credentials_handler))
+            .route(
+                "/worker/{job_id}/mcp/secrets/get",
+                post(mcp_secret_get_handler),
+            )
+            .route(
+                "/worker/{job_id}/mcp/secrets/exists",
+                post(mcp_secret_exists_handler),
+            )
+            .route(
+                "/worker/{job_id}/mcp/secrets/create",
+                post(mcp_secret_create_handler),
+            )
             .route_layer(axum::middleware::from_fn_with_state(
                 state.token_store.clone(),
                 worker_auth_middleware,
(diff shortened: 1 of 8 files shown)


serrrfirat:

This PR embeds an MCP client and secret proxy endpoints inside the retired Docker worker/orchestrator architecture. Reborn executes registered MCP capabilities through the shared dispatcher and host-mediated HTTP/credential boundaries; it must not proxy raw secret-store operations into a container worker.

Containerized capability execution can be proposed through Reborn runtime adapters and CapabilityHost, using opaque credential handles, explicit egress, resource governance, cancellation, and integration tests—never general secret get/create endpoints.

serrrfirat:

This embeds MCP and general secret proxy endpoints in retired container workers. Reborn dispatches MCP through host-mediated HTTP/credential boundaries and must not expose secret-store operations to workers.

Containerized capability work should use RuntimeAdapter/CapabilityHost with opaque credential handles, explicit egress, accounting, cancellation, and integration tests.

PR #2215, opened 2026-04-09
fix(bridge): sanitize auth_url on engine v2 path (#2206)

### LINKED ISSUE #2206: v2 engine: auth_url from tool_activate is not scheme-sanitized
## Summary

The engine-v2 auth surfacing path in `src/bridge/effect_adapter.rs` passes `auth_url` straight through from the `tool_activate`/`tool_auth` tool output to the `gate_required` SSE event without validating the URL scheme. A malicious or buggy tool can return `javascript:`, `file://`, `http://`, or `data:` and it will reach the browser.

The engine-v1 path in `src/agent/dispatcher.rs` gained a `sanitize_auth_url` helper (https-only) in #2038, but the v2 path was left untouched. This is a pre-existing gap, not a regression — filing as a follow-up from the #2038 post-merge review.

### v2: no sanitization (vulnerable)

`src/bridge/effect_adapter.rs:147` — `auth_gate_from_extension_result`:

```rust
auth_url: output_value
    .get(\"auth_url\")
    .and_then(|v| v.as_str())
    .map(|s| s.to_string()),
```

Same pattern in the \`tool_install\` post-readiness branch (\`effect_adapter.rs:865\`), the pre-flight \`check_action_auth\` gate (\`effect_adapter.rs:665\`), and the pre-flight \`check_tool_readiness\` gate (\`effect_adapter.rs:706\`). All of these feed \`ResumeKind::Authentication { auth_url, .. }\` into the engine gate stream and onto the client.

### v1: sanitized (reference implementation)

\`src/agent/dispatcher.rs\` (\`sanitize_auth_url\`):

\`\`\`rust
/// Only allow \`https://\` URLs for auth/setup links to prevent scheme injection
/// (e.g. \`javascript:\`, \`file://\`). Host validation is explicitly out of scope:
/// the URL source is a trusted local extension tool result, and display-side
/// rendering controls where the user is actually navigated.
fn sanitize_auth_url(url: Option<&str>) -> Option<String> {
    url.map(str::trim)
        .filter(|u| u.starts_with(\"https://\"))
        .map(ToOwned::to_owned)
}
\`\`\`

Applied in \`parse_auth_result\` to both \`auth_url\` and \`setup_url\`. Test coverage: \`test_sanitize_auth_url_rejects_non_https_schemes\`, \`test_parse_auth_result_strips_non_https_urls\`.

## Threat model

\`tool_activate\`/\`tool_auth\` output comes from local extensions (WASM, MCP, shell, HTTP). These are trusted-ish but not bulletproof — a buggy extension, a misconfigured MCP server responding with unexpected JSON, or a registry-installed tool with a compromised capabilities file can produce an \`auth_url\` that:

- navigates the user to an attacker-controlled origin outside the expected OAuth provider
- executes \`javascript:\` in the page's origin (cookie theft, session hijack) if the frontend renders it into an \`<a href>\` without its own defense
- opens \`file://\` or \`data:\` URIs

The web frontend already opens \`auth_url\` in a new tab via \`window.open(url, '_blank')\` (see #2050 — \"Open OAuth auth links in a new tab\"). Modern browsers will refuse \`javascript:\` in \`window.open\` targets in most cases, but this is defense-in-depth we already ship for v1 and there's no reason for v2 to lag.

## Desired fix

1. Extract \`sanitize_auth_url\` from \`src/agent/dispatcher.rs\` into a shared location. Natural home: \`src/auth/oauth.rs\` (where #2050 moved the rest of the shared OAuth runtime) as \`pub(crate) fn sanitize_auth_url(url: Option<&str>) -> Option<String>\`.
2. Update the dispatcher v1 path to import from there instead of keeping its private copy.
3. Apply it at every v2 site that extracts \`auth_url\` from tool output:
   - \`auth_gate_from_extension_result\` (\`effect_adapter.rs:147\`)
   - \`tool_install\` post-readiness branch (~\`effect_adapter.rs:865\`)
   - \`check_action_auth\` preflight gate (~\`effect_adapter.rs:665\`)
   - \`check_tool_readiness\` preflight gate (~\`effect_adapter.rs:706\`)
   - Any matching sites in \`src/bridge/auth_manager.rs\` that build \`auth_url\` from runtime data rather than from hard-coded config.
4. Regression test on the v2 path per the \"Test Through the Caller, Not Just the Helper\" rule in \`.claude/rules/testing.md\`: drive \`EffectAdapter::execute_action\` with an \`OAuthPromptTool\` fixture that returns \`auth_url: \"javascript:alert(1)\"\` and assert the resulting \`ResumeKind::Authentication.auth_url\` is \`None\` (not the original string).

## Out of scope

- Host validation (e.g. allowlist of OAuth providers). The v1 comment explicitly documents this as out of scope: the source is a local trusted extension, and display-side rendering controls navigation. Keep the invariant symmetric across v1/v2.
- Any change to the actual \`tool_activate\`/\`tool_auth\` schema — the fix lives purely in the host-side reader.

## Why now

Filed as a post-merge follow-up from PR #2038 review after staging was merged into the PR branch. The two auth-surfacing paths (v1 dispatcher and v2 effect adapter) were previously parallel implementations of the same tool-output contract with slightly different invariants. Consolidating the URL sanitizer into one shared helper eliminates the divergence and prevents the gap from recurring the next time one side is refactored.

ilblackdragon:
## Summary

Closes #2206. The engine-v2 effect adapter forwarded `auth_url` from `tool_activate`/`tool_auth` output (and from the v2 pre-flight auth gates) directly into `ResumeKind::Authentication` without scheme validation, leaving the v2 path open to `javascript:`, `file://`, `http://`, and `data:` URLs from a buggy or compromised local extension. The v1 dispatcher already had a private `sanitize_auth_url` helper from #2038; v2 was the asymmetry.

- Promoted `sanitize_auth_url` to `crate::auth::oauth` as `pub(crate)`, dropped the private copy in `agent::dispatcher`, and re-exported it so the v1 invariant is unchanged.
- Applied it at every v2 site that builds an `auth_url` from runtime data:
  - `effect_adapter::auth_gate_from_extension_result` (post-exec)
  - `effect_adapter` `tool_install` post-readiness branch
  - `effect_adapter` `check_action_auth` pre-flight gate
  - `effect_adapter` `check_tool_readiness` pre-flight gate
  - `auth_manager::check_tool_readiness` (extension `auth.auth_url()`)
  - `auth_manager::execute_latent_extension_action`
- Host validation stays out of scope, matching v1's documented invariant.

## Test plan

- [x] `cargo fmt`
- [x] `cargo test --lib` — 4520 passed
- [x] New regression `bridge::effect_adapter::tests::auth_gate_strips_non_https_auth_url_from_tool_activate_output` drives `EffectBridgeAdapter::execute_action` with an `OAuthPromptTool` stub returning `auth_url: "javascript:alert(1)"` and asserts the resulting `ResumeKind::Authentication.auth_url` is `None` (per the "Test Through the Caller, Not Just the Helper" rule in `.claude/rules/testing.md`).
- [x] Sibling test `auth_gate_preserves_https_auth_url_from_tool_activate_output` guards against an over-eager sanitizer by asserting a well-formed `https://` URL still flows through.
- [x] Helper-level tests live next to the helper in `crate::auth::oauth::sanitize_tests`.

### DIFF
--- a/src/auth/oauth.rs
+++ b/src/auth/oauth.rs
@@ -32,6 +32,52 @@ pub use crate::llm::oauth_helpers::{
 
 // ── Shared OAuth flow steps ─────────────────────────────────────────
 
+/// Only allow `https://` URLs for auth/setup links to prevent scheme injection
+/// (e.g. `javascript:`, `file://`). Host validation is explicitly out of scope:
+/// the URL source is a trusted local extension tool result, and display-side
+/// rendering controls where the user is actually navigated.
+///
+/// Both the v1 dispatcher (`src/agent/dispatcher.rs`) and the v2 effect adapter
+/// (`src/bridge/effect_adapter.rs`) call this on every `auth_url` extracted
+/// from `tool_activate`/`tool_auth` output before surfacing it to the client.
+/// Keeping the helper in one place ensures the v1/v2 invariants stay symmetric.
+pub(crate) fn sanitize_auth_url(url: Option<&str>) -> Option<String> {
+    url.map(str::trim)
+        .filter(|u| u.starts_with("https://"))
+        .map(ToOwned::to_owned)
+}
+
+#[cfg(test)]
+mod sanitize_tests {
+    use super::sanitize_auth_url;
+
+    #[test]
+    fn rejects_non_https_schemes() {
+        assert!(sanitize_auth_url(Some("javascript:alert(1)")).is_none());
+        assert!(sanitize_auth_url(Some("file:///etc/passwd")).is_none());
+        assert!(sanitize_auth_url(Some("http://example.com")).is_none());
+        assert!(sanitize_auth_url(Some("data:text/html,<h1>")).is_none());
+        assert!(sanitize_auth_url(Some("")).is_none());
+        assert!(sanitize_auth_url(None).is_none());
+    }
+
+    #[test]
+    fn allows_https() {
+        assert_eq!(
+            sanitize_auth_url(Some("https://accounts.google.com/o/oauth2/auth")),
+            Some("https://accounts.google.com/o/oauth2/auth".to_string())
+        );
+    }
+
+    #[test]
+    fn trims_whitespace_before_validating_scheme() {
+        assert_eq!(
+            sanitize_auth_url(Some("  https://example.com/auth  ")),
+            Some("https://example.com/auth".to_string())
+        );
+    }
+}
+
 /// Truncate `body` to at most `max_bytes` UTF-8 bytes, walking back to the
 /// nearest char boundary so the result is always a valid `&str`. Appends
 /// `"..."` when truncation actually happens. Used to bound any
--- a/src/bridge/auth_manager.rs
+++ b/src/bridge/auth_manager.rs
@@ -296,8 +296,9 @@ impl AuthManager {
                 ToolReadiness::NeedsAuth {
                     credential_name: described.credential_name,
                     instructions,
-                    auth_url: crate::auth::oauth::sanitize_auth_url(auth.auth_url())
-                        .or(described.auth_url),
+                    auth_url: crate::auth::oauth::sanitize_auth_url(auth.auth_url()).or_else(
+                        || crate::auth::oauth::sanitize_auth_url(described.auth_url.as_deref()),
+                    ),
                 }
             }
             Ok(crate::extensions::EnsureReadyOutcome::NeedsSetup { instructions, .. }) => {
--- a/src/bridge/auth_manager.rs
+++ b/src/bridge/auth_manager.rs
@@ -296,9 +296,7 @@ impl AuthManager {
                 ToolReadiness::NeedsAuth {
                     credential_name: described.credential_name,
                     instructions,
-                    auth_url: auth
-                        .auth_url()
-                        .map(ToOwned::to_owned)
+                    auth_url: crate::auth::oauth::sanitize_auth_url(auth.auth_url())
                         .or(described.auth_url),
                 }
             }
@@ -378,7 +376,7 @@ impl AuthManager {
                         .instructions()
                         .unwrap_or("Complete authentication to continue.")
                         .to_string(),
-                    auth_url: auth.auth_url().map(ToOwned::to_owned),
+                    auth_url: crate::auth::oauth::sanitize_auth_url(auth.auth_url()),
                 }),
                 Ok(crate::extensions::EnsureReadyOutcome::NeedsSetup { instructions, .. }) => {
                     Ok(LatentActionExecution::NeedsSetup {
--- a/src/agent/dispatcher.rs
+++ b/src/agent/dispatcher.rs
@@ -1306,15 +1306,7 @@ fn normalize_extension_name(value: Option<&str>) -> Option<String> {
         .map(ToOwned::to_owned)
 }
 
-/// Only allow `https://` URLs for auth/setup links to prevent scheme injection
-/// (e.g. `javascript:`, `file://`). Host validation is explicitly out of scope:
-/// the URL source is a trusted local extension tool result, and display-side
-/// rendering controls where the user is actually navigated.
-fn sanitize_auth_url(url: Option<&str>) -> Option<String> {
-    url.map(str::trim)
-        .filter(|u| u.starts_with("https://"))
-        .map(ToOwned::to_owned)
-}
+pub(super) use crate::auth::oauth::sanitize_auth_url;
 
 pub(super) fn auth_instructions_or_default(instructions: Option<&str>) -> String {
     instructions
@@ -1692,8 +1684,7 @@ mod tests {
 
     use super::{
         capture_auth_prompt, check_auth_required, extract_auth_prompt, parse_auth_result,
-        persist_selected_auth_prompt, restore_selected_auth_prompt, sanitize_auth_url,
-        selected_model_override,
+        persist_selected_auth_prompt, restore_selected_auth_prompt, selected_model_override,
     };
     use crate::agent::session::PendingAuthPrompt;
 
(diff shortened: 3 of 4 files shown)


serrrfirat:
**Medium Severity — Defense-in-depth gap at `effect_adapter.rs:598`**

The `LatentActionExecution::NeedsAuth` consumption site (line 598) passes `auth_url` directly into `ResumeKind::Authentication` without calling `sanitize_auth_url`. All four other `ResumeKind::Authentication` construction sites in this file apply the sanitizer (lines 170, 670, 711, 888), but this fifth site relies solely on `auth_manager` sanitizing at the source.

If a future code path produces `LatentActionExecution::NeedsAuth` without sanitizing, this consumer would forward the unsanitized URL. The inconsistency with the other four sites is the concern.

```rust
// Line 598 — current:
auth_url,

// Suggested: match the belt-and-suspenders pattern used elsewhere in this file:
auth_url: sanitize_auth_url(auth_url.as_deref()),
```

serrrfirat (dismissed):
Solid security fix. The v1/v2 sanitization asymmetry is correctly closed, helper is properly centralized, and tests follow the "Test Through the Caller" rule.

One minor defense-in-depth suggestion posted as a comment (effect_adapter.rs:598 — the LatentActionExecution::NeedsAuth consumption site could also apply sanitize_auth_url for consistency with the other four ResumeKind::Authentication sites). Not blocking — the source already sanitizes.

src/bridge/auth_manager.rs
```diff
@@ -296,9 +296,7 @@ impl AuthManager {
                 ToolReadiness::NeedsAuth {
                     credential_name: described.credential_name,
                     instructions,
-                    auth_url: auth
-                        .auth_url()
-                        .map(ToOwned::to_owned)
+                    auth_url: crate::auth::oauth::sanitize_auth_url(auth.auth_url())
                         .or(described.auth_url),
```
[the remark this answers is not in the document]
ilblackdragon: Good catch — `described.auth_url` flows from `start_skill_oauth_if_supported` → `build_pending_oauth_launch` → `oauth::build_oauth_url`, which doesn't validate the scheme of its input `authorization_url`. A skill manifest with a non-https `authorization_url` would propagate through this fallback unsanitized.

Applied your suggestion in e5eef55 — both branches of the `.or_else()` now go through `sanitize_auth_url`.

ilblackdragon:
Agreed — applied in e5eef55. All five `ResumeKind::Authentication` construction sites in `effect_adapter.rs` now call `sanitize_auth_url`, so the invariant holds regardless of whether upstream sources sanitize.

src/auth/oauth.rs
```diff
@@ -32,6 +32,52 @@ pub use crate::llm::oauth_helpers::{
...
+/// Only allow `https://` URLs for auth/setup links to prevent scheme injection
+/// (e.g. `javascript:`, `file://`). Host validation is explicitly out of scope:
+/// the URL source is a trusted local extension tool result, and display-side
+/// rendering controls where the user is actually navigated.
+///
+/// Both the v1 dispatcher (`src/agent/dispatcher.rs`) and the v2 effect adapter
+/// (`src/bridge/effect_adapter.rs`) call this on every `auth_url` extracted
+/// from `tool_activate`/`tool_auth` output before surfacing it to the client.
+/// Keeping the helper in one place ensures the v1/v2 invariants stay symmetric.
+pub(crate) fn sanitize_auth_url(url: Option<&str>) -> Option<String> {
```
serrrfirat, line 44: **Medium Severity — Case-sensitive scheme check**

`starts_with("https://")` is case-sensitive. While RFC 3986 states schemes are case-insensitive, a legitimate OAuth provider returning `"HTTPS://..."` or `"Https://..."` would be silently rejected, causing a confusing auth failure with no error message.

**Suggested fix:** Use `u.to_ascii_lowercase().starts_with("https://")` or at minimum add test cases documenting that mixed-case schemes are intentionally rejected.

serrrfirat, line 44: **Medium Severity — Embedded control characters pass the filter**

A URL like `"https://legit.com\r\njavascript:alert(1)"` passes `starts_with("https://")`. The `str::trim()` call only strips leading/trailing whitespace, not embedded control characters.

If the downstream consumer splits on newlines or processes the URL in a context where CRLF injection matters (e.g., HTTP header injection in a redirect), this could be exploitable.

**Suggested fix:** Add `.filter(|u| !u.contains(|c: char| c.is_control()))` or validate with `url::Url::parse` to reject URLs containing control characters.

src/bridge/auth_manager.rs
```diff
@@ -378,7 +377,7 @@ impl AuthManager {
                         .instructions()
                         .unwrap_or("Complete authentication to continue.")
                         .to_string(),
-                    auth_url: auth.auth_url().map(ToOwned::to_owned),
+                    auth_url: crate::auth::oauth::sanitize_auth_url(auth.auth_url()),
```
serrrfirat, line 380: **Medium Severity — Defense-in-depth gap: unsanitized `auth_url` returned**

`describe_missing_credential` returns unsanitized `auth_url` from `start_skill_oauth_if_supported`. While current callers sanitize (e.g., `check_tool_readiness` at line 299), any future caller that forgets to sanitize would re-introduce the vulnerability. The `MissingCredential` struct has no invariant enforcing sanitization.

**Suggested fix:** Either sanitize inside `describe_missing_credential` / `start_skill_oauth_if_supported`, or add a doc comment on `MissingCredential.auth_url` warning that it is unsanitized and callers must apply `sanitize_auth_url`.

serrrfirat (approved):
**Approve — this PR closes a real vulnerability and the remaining findings are hardening suggestions.**

Reviewed all 4 changed files. The core fix is correct: every v2 code path that constructs a `ResumeKind::Authentication` with an `auth_url` from runtime data now passes through `sanitize_auth_url`. The v1/v2 asymmetry gap from issue #2206 is closed. The refactoring to centralize the helper in `crate::auth::oauth` is the right architectural call.

Three Medium findings were posted inline, all non-blocking (hardening on top of an already-improved posture):
1. **Case-sensitive scheme check** (`oauth.rs:44`) — `"HTTPS://..."` from a legitimate provider would be silently rejected. Pre-existing behavior from v1.
2. **Embedded control characters** (`oauth.rs:44`) — CRLF injection edge case. Consider `url::Url::parse` validation or control-char filter in a follow-up.
3. **Unsanitized return in `describe_missing_credential`** (`auth_manager.rs:380`) — defense-in-depth gap; current callers all sanitize, but future callers might not.

Integration tests are well-designed — driving through `execute_action` rather than testing the helper in isolation covers the actual call path.

PR #2245, opened 2026-04-10
fix(memory): Don't treat "old_string" empty in patch mode

jokemanfire:
## Summary

any model will carry old_string: "" , so if set append mode ,treat it as append mode  though the old_string is "".

-

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] CI/Infrastructure
- [ ] Security
- [ ] Dependencies

## Validation

- [x] `cargo build`
- [ ] Relevant tests pass: 
- [ ] Manual testing: 

## Review Follow-Through

---

**Review track**:

### DIFF
--- a/src/tools/builtin/memory.rs
+++ b/src/tools/builtin/memory.rs
@@ -95,6 +95,17 @@ fn map_write_err(e: crate::error::WorkspaceError) -> ToolError {
     }
 }
 
+/// Non-empty patch `old_string` from params, or `None` if patch mode is off.
+///
+/// Empty and whitespace-only strings are treated as absent so payloads such as
+/// `"old_string": ""` do not force patch mode when writing or appending `content`.
+fn memory_write_patch_old_string(params: &serde_json::Value) -> Option<&str> {
+    params
+        .get("old_string")
+        .and_then(|v| v.as_str())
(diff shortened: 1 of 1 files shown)


serrrfirat:
Thank you for identifying the malformed memory-write pattern where a model supplies an empty old_string during an append operation.\n\nWe are closing this legacy implementation because Reborn now owns this contract in the provider-neutral MemoryService DTO and native memory provider. Reborn intentionally treats an explicitly empty old_string as invalid rather than allowing ambiguous patch semantics.\n\nReferences:\n- https://github.com/nearai/ironclaw/blob/main/crates/ironclaw_memory/src/service.rs\n- https://github.com/nearai/ironclaw/blob/main/crates/ironclaw_memory_native/src/service.rs\n\nIf Reborn traces show models frequently emitting append=true with old_string="", the safe follow-up would be a narrow Reborn normalization rule that treats it as absent only for a valid append request, while continuing to reject it for patch operations. That should be driven by a focused Reborn issue and caller-level test rather than porting this v1 behavior directly.\n\nThank you again; this is an architecture-transition closure, and we would welcome your help on a trace-backed Reborn fix if the pattern remains active.

PR #2247, opened 2026-04-10
fix(oauth): use localhost for redirect URI when bound to 0.0.0.0

ilblackdragon:
## Summary
- When `GATEWAY_HOST=0.0.0.0` and no `TUNNEL_PUBLIC_URL` is set, the extension OAuth redirect URI was built as `http://0.0.0.0:<port>/oauth/callback`, which Google rejects as invalid
- Map unspecified bind addresses (`0.0.0.0`, `::`, `[::]`) to `localhost` when constructing the gateway base URL, matching the social-login flow's existing fallback

## Test plan
- [ ] Set `GATEWAY_HOST=0.0.0.0` with no `TUNNEL_PUBLIC_URL`, trigger extension OAuth — redirect URI should use `localhost`
- [ ] Set `GATEWAY_HOST=127.0.0.1` — redirect URI should use `127.0.0.1` (unchanged)
- [ ] Set `TUNNEL_PUBLIC_URL=https://example.com` — redirect URI should use the tunnel URL (unchanged)

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -770,7 +770,15 @@ async fn async_main() -> anyhow::Result<()> {
                 .tunnel
                 .public_url
                 .clone()
-                .unwrap_or_else(|| format!("http://{}:{}", gw_config.host, gw_config.port));
+                .unwrap_or_else(|| {
+                    // 0.0.0.0 / [::] are valid bind addresses but not valid
+                    // OAuth redirect hosts — use localhost instead.
+                    let host = match gw_config.host.as_str() {
+                        "0.0.0.0" | "::" | "[::]" => "localhost",
+                        other => other,
+                    };
+                    format!("http://{}:{}", host, gw_config.port)
+                });
             ext_mgr.enable_gateway_mode(gw_base).await;
             gw = gw.with_extension_manager(Arc::clone(ext_mgr));
         }
--- a/src/main.rs
+++ b/src/main.rs
@@ -770,15 +770,7 @@ async fn async_main() -> anyhow::Result<()> {
                 .tunnel
                 .public_url
                 .clone()
-                .unwrap_or_else(|| {
-                    // 0.0.0.0 / [::] are valid bind addresses but not valid
-                    // OAuth redirect hosts — use localhost instead.
-                    let host = match gw_config.host.as_str() {
-                        "0.0.0.0" | "::" | "[::]" => "localhost",
-                        other => other,
-                    };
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -1441,3 +1441,45 @@ async fn async_main() -> anyhow::Result<()> {
...
+/// because they are valid bind addresses but not valid OAuth redirect hosts.
+fn oauth_base_url(host: &str, port: u16) -> String {
+    let trimmed = host.trim_start_matches('[').trim_end_matches(']');
+    let is_unspecified = trimmed
+        .parse::<std::net::IpAddr>()
+        .is_ok_and(|ip| ip.is_unspecified());
+    if is_unspecified {
+        format!("http://localhost:{}", port)
+    } else {
+        format!("http://{}:{}", host, port)
```
serrrfirat, line 1457: **Medium Severity — Malformed IPv6 URL**

When `host` is a bare IPv6 address like `"::1"`, the function produces `http://::1:3000` which is a malformed URL. Per RFC 3986, IPv6 literals in URIs must be enclosed in brackets: `http://[::1]:3000`.

The test on line 1483 asserts the broken form, so this is a pre-existing bug carried forward. While `0.0.0.0`/`::`/`[::]` are now mapped to `localhost`, anyone setting `GATEWAY_HOST=::1` (IPv6 loopback) would get an invalid redirect URI.

**Suggested fix:** In the `else` branch, detect whether `host` parses as an IPv6 `IpAddr`, then format as `http://[{host}]:{port}`. Update the test assertion accordingly.

serrrfirat (commented):
**Code LGTM — approve once CI (Clippy) is green.**

Reviewed the single changed file (`src/main.rs`). The core logic is correct: parsing the host as `IpAddr` and checking `is_unspecified()` is robust and handles all unspecified address representations. Tests cover the important cases.

One Medium finding posted inline (non-blocking, pre-existing):
1. **IPv6 URL formatting** (`main.rs:1457`) — bare IPv6 like `"::1"` produces `http://::1:3000` instead of `http://[::1]:3000`. This is a pre-existing bug carried forward, not introduced by this PR. Since you're already touching this code and parsing `IpAddr`, it would be the ideal time to bracket IPv6 in the format string.

Cannot formally approve while CI is red (4 Clippy failures). Fix those and this is ready to merge.

serrrfirat (approved):

PR #2254, opened 2026-04-10
fix: include Completed and Submitted in is_terminal()

NYX-privacy-ai:
## Summary

`JobState::is_terminal()` returns `false` for `Completed` and `Submitted`, even though both are terminal states that should not be polled or retried. This causes `FullJobWatcher` to poll indefinitely for jobs that reach either state.

The bug is self-documenting — `is_parallel_blocking()` already notes:

> "Completed and Submitted jobs are in the state machine but are no longer actively executing."

Yet `is_terminal()` only matches `Accepted | Failed | Cancelled`.

## Fix

Add `Completed` and `Submitted` to the `is_terminal()` match, and add corresponding test assertions for all non-terminal states (`Pending`, `InProgress`, `Stuck`).

## Test plan

- [ ] `cargo test` — updated `test_terminal_states` asserts `Completed` and `Submitted` are terminal, and `Pending`/`InProgress`/`Stuck` are not
- [ ] Verify `FullJobWatcher` stops polling when a job reaches `Completed` or `Submitted`

### DIFF
--- a/src/context/state.rs
+++ b/src/context/state.rs
@@ -76,7 +76,10 @@ impl JobState {
 
     /// Check if this is a terminal state.
     pub fn is_terminal(&self) -> bool {
-        matches!(self, Self::Accepted | Self::Failed | Self::Cancelled)
+        matches!(
+            self,
+            Self::Accepted | Self::Completed | Self::Submitted | Self::Failed | Self::Cancelled
+        )
     }
 
     /// Check if the job is active (not terminal).
@@ -485,9 +488,13 @@ mod tests {
     #[test]
     fn test_terminal_states() {
         assert!(JobState::Accepted.is_terminal());
+        assert!(JobState::Completed.is_terminal());
+        assert!(JobState::Submitted.is_terminal());
         assert!(JobState::Failed.is_terminal());
         assert!(JobState::Cancelled.is_terminal());
+        assert!(!JobState::Pending.is_terminal());
         assert!(!JobState::InProgress.is_terminal());
+        assert!(!JobState::Stuck.is_terminal());
     }
 
     #[test]


src/context/state.rs
```diff
@@ -76,7 +76,10 @@ impl JobState {
 
     /// Check if this is a terminal state.
     pub fn is_terminal(&self) -> bool {
-        matches!(self, Self::Accepted | Self::Failed | Self::Cancelled)
+        matches!(
```
serrrfirat, line 79: **High Severity**\n\nThis change conflates "execution finished" with "terminal". `Completed` and `Submitted` still have outgoing transitions in the state machine, so widening `is_terminal()` changes global `is_active()` behavior, not just watcher polling. The PR description points at `FullJobWatcher`, but that watcher already stops on `!is_parallel_blocking()` in `src/agent/routine_engine.rs`, specifically because `Completed` is not terminal from the broader lifecycle perspective. With this patch, other callers that depend on `is_active()` will flip behavior too, including container reaping, active-job listings, and restart gating.\n\nSuggested fix: keep `is_terminal()` for states with no further transitions, and use a separate predicate for "not actively executing" at the polling call sites.

src/context/state.rs
```diff
@@ -485,9 +488,13 @@ mod tests {
     #[test]
     fn test_terminal_states() {
         assert!(JobState::Accepted.is_terminal());
+        assert!(JobState::Completed.is_terminal());
```
serrrfirat, line 491: **Medium Severity**\n\nThis test only redefines the helper contract; it does not prove the reported bug or protect the real caller path. The project guidance says to test through the caller when a predicate gates behavior. If the bug is that a watcher polls indefinitely, the regression test belongs around `FullJobWatcher` / the job lifecycle path, not just `JobState::is_terminal()`. As written, this can pass while still changing unrelated lifecycle behavior.\n\nSuggested fix: add a caller-level regression test that drives the watcher for `Completed` and `Submitted`, and move the fix to the watcher logic if that is the actual failing path.

serrrfirat:

Issue #2204, opened 2026-04-09
[live-test 2026-04-09 16:17 UTC] Add /metrics Prometheus endpoint


ilblackdragon:
**This is an automated live integration test issue.** It was opened by `tests/e2e_github_dev_workflow.rs::github_dev_workflow_full_loop` to exercise the `developer-setup` + `github-workflow` skills end-to-end against a real repository.

## Feature request

Expose Prometheus-style metrics for IronClaw at `/metrics`. Should include:

- Request latency histograms per channel + per tool
- Tool execution count + success rate
- Active session count + thread count
- LLM token usage counters per model + per backend

The endpoint should not require auth in single-user mode (the typical local ironclaw deployment); for multi-user gateway deployments it should require the existing admin credential.

## Why this matters

Production observability is currently limited to `tracing` log output. A scrapeable metrics endpoint unlocks dashboards, alerting, and SLO tracking without log-aggregation pipelines.

---

🤖 The agent will respond to this issue with a triage and an implementation plan. **The test will close this issue automatically after recording the agent's response.** If you're a human reading this and the issue is still open after a few minutes, the test panicked — see the test output for the last activity dump.

ilblackdragon:
## 📋 Implementation Plan: Prometheus Metrics Endpoint

### Scope

#### In Scope
- **Endpoint**: `/metrics` route exposing Prometheus-format metrics
- **Metrics to collect**:
  - Request latency histograms (per channel + per tool)
  - Tool execution counters (total + success rate)
  - Active session count + thread count
  - LLM token usage counters (per model + per backend)
- **Auth**: 
  - Single-user mode: no auth required (local deployments)
  - Multi-user gateway: requires existing admin credential
- **Integration**: Compatible with Prometheus scrape protocol

#### Out of Scope (Future Iterations)
- Pushgateway support
- Custom metric exports via API
- Alertmanager rule templates
- Grafana dashboard JSON exports

---

### Milestones

#### Milestone 1: Foundation (Week 1)
- [ ] Add `metrics` feature flag to Cargo.toml
- [ ] Integrate `prometheus` or `metrics` + `metrics-exporter-prometheus` crate
- [ ] Create `src/metrics/mod.rs` with registry setup
- [ ] Add HTTP route `/metrics` in gateway/router layer
- [ ] Implement auth middleware for multi-user mode

#### Milestone 2: Instrumentation (Week 2)
- [ ] Wrap tool execution with latency + success counters
[...]

PR #2268, opened 2026-04-10
feat(skills): setup-marker lifecycle, chain-loading, and live GitHub workflow test

ilblackdragon:
## Summary

- **One-time setup-marker exclusion**: persona/setup skills (`developer-setup`, `ceo-setup`, etc.) declare a `setup_marker` workspace path. After their first run writes the marker file, the selector excludes them from future activation — freeing ~3000 tokens of budget per turn for reactive operational skills.
- **Chain-loading via `requires.skills`**: when a parent skill is selected, its declared companions are automatically loaded even if they score 0 on their own. Implemented in both v1 Rust (`prefilter_skills`) and v2 Python (`select_skills`).
- **Persona skill rename**: `*-assistant` → `*-setup` to make the one-time lifecycle explicit (developer-assistant → developer-setup, ceo-assistant → ceo-setup, trader-assistant → trader-setup, content-creator-assistant → content-creator-setup).
- **V2SkillMetadata.requires propagation**: the v1→v2 skill migration was silently dropping `requires.skills`, making chain-loading dead code in v2. Fixed by adding the `requires` field to `V2SkillMetadata` and copying it during migration.
- **Budget bump**: `SKILLS_MAX_CONTEXT_TOKENS` default 4000 → 6000.
- **Live GitHub workflow test**: real end-to-end test that creates a real issue on `nearai/ironclaw`, lets the agent triage it, verifies a real comment was posted, and cleans up. Exercises `github-workflow` + `github` skills against the real API.
- **Test infrastructure**: `TestRig.with_secret(name, value)` for pre-seeding credentials, `TestRig.get_secret(name)` for reading them back, `github_api` REST helpers for direct issue management, `trace_contains_tool_call` / `assert_trace_contains_tool_call` for behavior assertions.

## Bug fixes included

- **`pending_approval_status_update` missing null fallback** (pre-existing in extension-lifecycle): extracted `display_parameters_or_fallback()` shared by both approval helpers + regression test.
- **`notify_pending_gate` lock-across-await** (merge regression): refactored to take `Option<Arc<SseManager>>` instead of `&EngineState` so the engine-state read lock can be dropped before SSE/channel I/O.

## Test plan

- [x] `cargo test -p ironclaw_skills --lib` — 164 passed (11 new: 4 marker + 7 chain-load)
- [x] `cargo test --features libsql --test skill_setup_marker_lifecycle` — 1 passed
- [x] `cargo test --features libsql --test skill_chain_load_lifecycle` — 1 passed, 1 ignored (v2 needs multi-turn TraceLlm harness)
- [x] `cargo check --features libsql --tests` — clean
- [x] `cargo clippy --features libsql --tests --all-targets -- -D warnings` — clean
- [x] Live `e2e_github_dev_workflow` recording — passed (issue #2204 created, agent posted 2 comments, cleanup closed it, 73s)
- [ ] CI green on this PR


ilblackdragon (commented):
Addressed review feedback:

- **Dead `try_add` function + stale comments** (9fee51e4): Removed the unused `try_add` closure and the misleading `budget[0]` / Monty-compat comments. The inlined loop at line 426+ is the only implementation.

Also merged `origin/staging` and reconciled test harness divergence (3b4680df).

src/agent/agent_loop.rs
```diff
@@ -545,33 +545,78 @@ impl Agent {
...
         };
-        let available = guard.skills();
+
+        // Resolve which setup markers are satisfied by the current
+        // workspace. A marker is "satisfied" iff its path exists.
+        // Without a workspace, we conservatively treat all markers as
+        // unsatisfied (setup skills can still activate). Errors checking
+        // a marker are logged and treated as unsatisfied.
+        let mut satisfied: std::collections::HashSet<String> = std::collections::HashSet::new();
+        if let Some(ws) = self.deps.workspace.as_ref() {
```
serrrfirat, line 606: **Medium Severity**

This marker check uses `self.deps.workspace`, which is the owner/global workspace. The dispatcher scopes the actual turn workspace by `message.user_id`, so in multi-user channels this can read the wrong user’s marker state: one user’s completed setup can suppress another user’s setup skill, or a non-owner user’s marker can be missed and the setup skill can repeat.

Please pass the `TenantCtx` or message user into `select_active_skills` and check markers against the same scoped workspace used to build the turn’s system prompt. A non-owner user integration test would cover this.

skills/developer-setup/SKILL.md
```diff
@@ -198,3 +199,16 @@ Tell the user:
...
 > - **`/plan <description>`** — structured execution plan for complex tasks
 > - **`/product-prioritization`** — score and rank features by demand
+
+## Step 7: Mark setup complete
+
+After confirming with the user that everything is in place, write the setup completion marker so this skill stops competing for the activation budget on every subsequent message:
+
+```
+memory_write(
+  path: "commitments/.developer-setup-complete",
```
serrrfirat: **High Severity**

The setup completion marker is written with the wrong parameter name. `memory_write` only reads the `target` parameter; `path` is ignored and the tool defaults to `daily_log`. That means this step appends the marker content to the daily log instead of creating `commitments/.developer-setup-complete`, so the `setup_marker` exclusion never sees the marker and this setup skill keeps competing on future turns. The same pattern appears in `ceo-setup`, `content-creator-setup`, and `trader-setup`.

Please change these examples to use `target: "commitments/.developer-setup-complete"` (and the analogous persona marker paths), and add a lifecycle test using one real persona skill so this does not regress.

crates/ironclaw_engine/orchestrator/default.py
```diff
@@ -341,12 +341,44 @@ def score_skill(skill, message_lower, message_original):
 
 
 def select_skills(skills, goal, max_candidates=3, max_tokens=4000):
```
serrrfirat: **Medium Severity**

Engine v2 still hard-codes the skill budget path at 4000 tokens. The selector signature here defaults `max_tokens=4000`, and the orchestrator call later still invokes `select_skills(all_skills, goal, max_candidates=3, max_tokens=4000)`, so the new `SKILLS_MAX_CONTEXT_TOKENS` default of 6000 and any env/config override are ignored on the v2 path. This makes v1 and v2 diverge and keeps v2 on the old budget after the config bump.

Please thread the resolved skills config into the v2 orchestrator/host context, or otherwise make this use the same `max_active_skills` and `max_context_tokens` values as v1.

skills/developer-setup/SKILL.md
```diff
@@ -1,8 +1,9 @@
 ---
-name: developer-assistant
-version: 0.1.0
-description: Commitment tracking and workflow automation for software developers — multi-repo GitHub awareness, CI/PR signal extraction, tech debt tracking, coding agent delegation, morning dev brief and weekly retro.
+name: developer-setup
+version: 0.2.0
+description: One-time onboarding for the developer workflow — installs github-workflow missions, creates the commitments workspace, registers per-repo projects, writes calibration memories. After successful setup this skill is excluded from selection until the marker file is deleted.
```
serrrfirat, line 4: **Medium Severity**

This setup skill says it installs the GitHub workflow, but the declared companion order makes the GitHub skills effectively unreachable under the default caps. With `max_active_skills = 3` and a 6000-token budget, the 3000-token parent plus earlier companions consume the candidate slots or remaining budget before `github` / `github-workflow` are considered. The config comment also says 6000 accommodates 3000 + 2000 + 2000, but that total is 7000.

Please either put the critical GitHub companions first and set caps that actually fit the intended bundle, or adjust selection so chain-loaded companions do not crowd out independently relevant skills. A test using the real `developer-setup` manifest would catch this.

crates/ironclaw_skills/src/types.rs
```diff
@@ -89,6 +89,23 @@ pub struct ActivationCriteria {
...
+    /// budget again. Reactive operational skills (commitment-triage,
+    /// decision-capture, etc.) leave this field unset and continue to
+    /// activate on every matching message.
+    ///
+    /// To re-trigger setup, delete the marker file from the workspace.
+    /// Typical markers are paths the setup skill itself creates as part
+    /// of its first run (e.g. `commitments/README.md` for the developer
+    /// setup, which Step 2 of `developer-setup` writes).
+    #[serde(default)]
+    pub setup_marker: Option<String>,
```
serrrfirat, line 112: **Medium Severity — `setup_marker` has no validation or sanitization**

The `setup_marker` field is an arbitrary string from SKILL.md YAML frontmatter with no length limits, no path-traversal stripping (e.g., `../`), and no character restrictions. An installed (untrusted) skill could:
- Set a very long marker string, causing O(n) hashing overhead per turn in `satisfied_setup_markers`
- Set it to a string matching an existing workspace doc, permanently disabling itself

**Suggested fix:** Add `setup_marker` to `enforce_limits()` — validate max length (e.g., 256 chars), reject paths containing `..`, and log a warning for suspicious patterns.

crates/ironclaw_engine/orchestrator/default.py
```diff
@@ -355,18 +387,52 @@ def select_skills(skills, goal, max_candidates=3, max_tokens=4000):
...
-        cost = max(activation.get("max_context_tokens", 1000), 1)
-        if cost <= budget:
-            budget -= cost
-            selected.append(skill)
+        parent_meta = parent.get("metadata", {})
+        parent_name = parent_meta.get("name")
+        if parent_name is None or str(parent_name) in selected_names:
+            continue
+        parent_activation = parent_meta.get("activation", {})
+        parent_cost = max(parent_activation.get("max_context_tokens", 1000), 1)
```
serrrfirat: **Medium Severity — Python v2 path trusts `max_context_tokens` unconditionally**

The Rust selector detects implausibly low `max_context_tokens` (when `approx_tokens > declared_tokens * 2`) and falls back to the actual estimate. The Python `select_skills` uses `max(activation.get("max_context_tokens", 1000), 1)` — it trusts the declared value unconditionally.

A skill declaring `max_context_tokens: 1` in the v2 path could consume minimal budget while injecting a 10,000-token prompt, effectively bypassing the context budget.

**Suggested fix:** Port the `approx_tokens > declared_tokens * 2` guard to the Python side, reading `content` length from the skill dict.

tests/e2e_github_dev_workflow.rs
```diff
@@ -0,0 +1,584 @@
...
+            issue_body,
+        )
+        .await
+        .expect("create real test issue on nearai/ironclaw");
+        eprintln!("[live-test] created real issue #{issue_number}: {issue_url}");
+
+        // Capture the comment count baseline so we can detect new
+        // comments posted by the agent.
+        let baseline_comments =
+            github_api::list_issue_comments(&github_token, REPO_OWNER, REPO_NAME, issue_number)
```
serrrfirat: **Medium Severity — Orphaned GitHub issue on test failure**

If `list_issue_comments` at line 479 panics (via `.expect()`), the issue is already created but the `AssertUnwindSafe` cleanup guard has not been entered yet. The issue remains open on the real `nearai/ironclaw` repo.

**Suggested fix:** Move the `AssertUnwindSafe` guard to wrap everything after `create_issue`, or restructure so cleanup is guaranteed via a `Drop` guard set immediately after issue creation.

crates/ironclaw_engine/orchestrator/default.py
```diff
@@ -355,18 +387,83 @@ def select_skills(skills, goal, max_candidates=3, max_tokens=4000):
...
+        selected_names.add(str(name))
+        return True
+
+    # Note: budget is captured by try_add via closure. Python doesn't
+    # support `nonlocal` directly in Monty the same way as CPython, so
+    # we use a list as a mutable reference. Re-assign via budget[0] and
+    # read via budget[0] in try_add.
+    #
+    # Actually, for Monty compat we avoid closures modifying enclosing
+    # scope. Inline the try-add logic below instead.
```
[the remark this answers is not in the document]
ilblackdragon: Already resolved in 9fee51e4. The dead `try_add` function and stale `budget[0]` comments were removed in that commit.

tests/support/live_harness.rs
```diff
@@ -362,55 +582,23 @@ impl LiveTestHarnessBuilder {
...
-            // replay from. Build a no-op harness so the test can
-            // detect the mode and skip itself gracefully — without
-            // panicking on a missing fixture.
-            self.build_no_replay().await
+            panic!(
+                "[LiveTest] '{}' has trace recording disabled and no replay fixture. \
+                 Run with IRONCLAW_LIVE_TEST=1 (and optionally \
+                 `with_no_trace_recording()`)",
+                self.test_name
+            );
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in 44652387. Added `TestMode::Skipped` variant with a lightweight `build_skip()` path instead of panicking. Updated the two callers in `e2e_live.rs` to check `mode() != Live` (instead of `== Replay`) so they correctly handle both Replay and Skipped modes.

src/agent/agent_loop.rs
```diff
@@ -545,33 +545,78 @@ impl Agent {
...
-        let guard = match registry.read() {
-            Ok(g) => g,
+        // Snapshot the skill list + distinct setup markers under the read
+        // lock, then drop the guard before any await. The marker checks
+        // and the prefilter call don't need the registry lock and we
+        // shouldn't hold a poisonable RwLock across an await point.
+        let (available, distinct_markers) = match registry.read() {
+            Ok(guard) => {
+                let skills_clone: Vec<ironclaw_skills::LoadedSkill> = guard.skills().to_vec();
+                let mut markers: std::collections::HashSet<String> =
```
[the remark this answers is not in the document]
ilblackdragon: Disagree — the clone is intentional and cheap. `LoadedSkill.prompt_content` is `Arc<String>` so the clone is a shallow ref-count bump, not a deep copy. The comment at lines 563-566 explains the RwLock guard must be dropped before any `.await`. The suggested alternative (re-acquire lock after awaits) introduces TOCTOU risk and double-lock overhead.

tests/support/test_rig.rs
```diff
@@ -263,6 +264,45 @@ impl TestRig {
...
+        let store = self.secrets_store.as_ref()?;
+        match store.get_decrypted(&self.owner_id, name).await {
+            Ok(decrypted) => Some(decrypted.expose().to_string()),
+            Err(e) => {
+                eprintln!(
+                    "[TestRig] get_secret('{name}') for owner '{}' failed: {e}",
+                    self.owner_id
+                );
+                None
+            }
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in 44652387. Now matches `SecretError::NotFound` explicitly and returns `None` silently. Only unexpected errors are logged.

tests/support/test_rig.rs
```diff
@@ -1123,9 +1272,48 @@ impl TestRigBuilder {
...
+                        // Already-exists is fine — the seeded DB may
+                        // already have the secret. Anything else is a
+                        // setup error worth surfacing.
+                        let msg = e.to_string();
+                        if !msg.contains("already exists") && !msg.contains("UNIQUE") {
+                            eprintln!(
+                                "[TestRig] WARNING: failed to pre-seed secret '{name}' for \
+                                 user '{owner_id}': {msg}"
+                            );
+                        }
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in 44652387. Replaced string matching with a `get_decrypted` existence check before `create` — no more reliance on error message text.

crates/ironclaw_engine/orchestrator/default.py
```diff
@@ -355,18 +387,52 @@ def select_skills(skills, goal, max_candidates=3, max_tokens=4000):
...
-        if cost <= budget:
-            budget -= cost
-            selected.append(skill)
+        parent_meta = parent.get("metadata", {})
+        parent_name = parent_meta.get("name")
+        if parent_name is None or str(parent_name) in selected_names:
+            continue
+        parent_activation = parent_meta.get("activation", {})
+        parent_cost = max(parent_activation.get("max_context_tokens", 1000), 1)
+        if parent_cost > budget:
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in 44652387. Changed both fallbacks (parent at line 405 and companion at line 428) from 1000 to 2000 to match the Rust `ActivationCriteria` default.

skills/developer-setup/SKILL.md
ilblackdragon: Fixed in 24f1a626. Changed `path:` to `target:` in all 4 setup skill completion markers (developer-setup, ceo-setup, content-creator-setup, trader-setup). The `memory_write` tool reads the `target` parameter, not `path`.

src/agent/agent_loop.rs
ilblackdragon: Fixed in 24f1a626. `select_active_skills` now takes a `user_id` parameter and scopes the workspace via `scoped_to_user()` before checking setup markers. The dispatcher passes `message.user_id` at the call site.

crates/ironclaw_engine/orchestrator/default.py
ilblackdragon: Fixed in 24f1a626. Changed both the `select_skills` default and the call site from 4000 to 6000 to match the v1 Rust config (`src/config/skills.rs:45`).

skills/developer-setup/SKILL.md
ilblackdragon: Fixed in 24f1a626. Reordered `requires.skills` in developer-setup to put `github` and `github-workflow` first (critical for the setup flow). Chain-loaded companions are selected in list order, so the first 2 that fit within remaining budget/slots will always be the GitHub skills. Also fixed the misleading budget comment in `src/config/skills.rs`.

crates/ironclaw_skills/src/types.rs
ilblackdragon: Fixed in 24f1a626. Added `setup_marker` validation in `enforce_limits()`: rejects markers longer than 256 bytes or containing `'..'\ path traversal. Three new tests cover traversal rejection, oversized rejection, and valid marker preservation.

crates/ironclaw_engine/orchestrator/default.py
ilblackdragon: Fixed in 24f1a626. Added `_skill_token_cost()` helper to the Python v2 path that mirrors the Rust `skill_token_cost()`: estimates actual token count from content length (bytes × 0.25), and if the actual estimate exceeds the declared `max_context_tokens` by more than 2×, uses the actual estimate instead. Both parent and companion cost calculations now go through this helper.

tests/e2e_github_dev_workflow.rs
ilblackdragon: Fixed in 24f1a626. Moved the `AssertUnwindSafe` guard to wrap everything after `create_issue`, including the `list_issue_comments().expect()` baseline call. Now any panic after issue creation triggers cleanup.

crates/ironclaw_engine/src/executor/orchestrator.rs
```diff
@@ -1922,9 +1938,36 @@ async fn handle_list_skills(
         }
     };
 
+    // Build the set of all existing doc titles (== workspace paths in
+    // v2) once, so setup-marker filtering below is O(1) per skill.
+    let existing_titles: std::collections::HashSet<&str> =
+        docs.iter().map(|d| d.title.as_str()).collect();
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in 52de73ca. Filtered `existing_titles` to exclude `DocType::Skill` docs so marker lookup only matches workspace files.

crates/ironclaw_skills/src/types.rs
```diff
@@ -89,6 +93,23 @@ pub struct ActivationCriteria {
...
+    /// so they activate once during onboarding, write the marker as part
+    /// of their setup steps, and then never compete for the activation
+    /// budget again. Reactive operational skills (commitment-triage,
+    /// decision-capture, etc.) leave this field unset and continue to
+    /// activate on every matching message.
+    ///
+    /// To re-trigger setup, delete the marker file from the workspace.
+    /// Typical markers are paths the setup skill itself creates as part
+    /// of its first run (e.g. `commitments/README.md` for the developer
+    /// setup, which Step 2 of `developer-setup` writes).
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in 52de73ca. Updated the doc example to `commitments/.developer-setup-complete`.

crates/ironclaw_skills/src/v2.rs
```diff
@@ -153,6 +153,17 @@ pub struct V2SkillMetadata {
...
     pub trust: SkillTrust,
+    /// Advisory companion skills — declared in the original SKILL.md
+    /// `requires.skills` block. Preserved through v1→v2 migration so
+    /// the Python orchestrator's `select_skills` chain-loading pass
+    /// can pull companions in alongside their parent.
+    ///
+    /// This mirrors `SkillManifest.requires` but is flattened to just
+    /// the companion list since binary/env/config gating happens at
+    /// load time (v1) and is not re-evaluated in v2. Legacy metadata
+    /// without this field deserializes with an empty `requires`.
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in 52de73ca. Updated the comment to reflect that the full `GatingRequirements` struct is preserved, not flattened.

tests/support/test_rig.rs
```diff
@@ -1123,9 +1275,48 @@ impl TestRigBuilder {
...
+                        Ok(_) => {} // already seeded — skip
+                        Err(_) => {
+                            if let Err(e) = secrets_store.create(&owner_id, params).await {
+                                eprintln!(
+                                    "[TestRig] WARNING: failed to pre-seed secret '{name}' for \
+                                     user '{owner_id}': {e}"
+                                );
+                            }
+                        }
+                    }
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in 52de73ca. Now matches `SecretError::NotFound` explicitly and only creates on that variant. Other errors are logged as warnings.

tests/support/test_rig.rs
```diff
@@ -1149,9 +1243,48 @@ impl TestRigBuilder {
...
+                    match secrets_store.get_decrypted(&owner_id, name).await {
+                        Ok(_) => {} // already seeded — skip
+                        Err(_) => {
+                            if let Err(e) = secrets_store.create(&owner_id, params).await {
+                                eprintln!(
+                                    "[TestRig] WARNING: failed to pre-seed secret '{name}' for \
+                                     user '{owner_id}': {e}"
+                                );
+                            }
+                        }
```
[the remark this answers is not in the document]
ilblackdragon: Same as above — fixed in 52de73ca with explicit `SecretError::NotFound` matching.

crates/ironclaw_engine/src/executor/orchestrator.rs
```diff
@@ -1966,9 +1982,36 @@ async fn handle_list_skills(
...
+                Some(m) if existing_titles.contains(m) => {
+                    debug!(
+                        skill = %d.title,
+                        marker = %m,
+                        "__list_skills__: excluding setup skill — marker already present"
+                    );
+                    false
+                }
+                _ => true,
+            }
```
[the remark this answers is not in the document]
ilblackdragon: The v1 path (Rust) checks via `Workspace::exists()` which does its own normalization. The v2 path compares raw titles since MemoryDoc titles are already normalized during ingestion. The real setup markers in practice are clean paths like `commitments/.developer-setup-complete` — no leading/trailing slashes. Leaving as-is for now; if path normalization mismatches surface in practice we can add a normalize step.

PR #2267, opened 2026-04-10
fix(gateway): scope chat approvals to the active thread

serrrfirat:
## Summary
- scope gateway slash-approve resolution to the active thread so chat approvals cannot resume unrelated background gates
- send approval actions with the approval card thread id instead of always using the currently selected thread
- add backend and e2e regressions for cross-thread slash-approve and wrong-thread approval cards

## Verification
- cargo test handle_approval_ignores_pending_gate_from_different_thread --lib
- uv run --project tests/e2e pytest -q tests/e2e/scenarios/test_tool_approval.py -k approval_card_button_posts_card_thread_id\ or\ slash_approve_does_not_intercept_other_thread_card\ or\ slash_approve_is_thread_scoped_api

Fixes #2238

PR #2270, opened 2026-04-10
feat(gateway): approval queue tray in web UI sidebar

serrrfirat:
## Summary

Adds a **cross-thread approval queue tray** to the web gateway sidebar that provides visibility into all pending tool approval gates — not just those in the active thread. Companion to #2267 (thread-scoped approvals), which fixes correctness but creates a visibility gap for background threads.

### Problem
When a background routine, heartbeat, or non-active thread triggers a tool approval gate, the user has no way to see it without manually switching to that thread. Gates from inactive threads are silently blocked.

### Solution
Three-layer non-intrusive design that adds zero UI surface when no gates are pending:

- **Sidebar tray** — "Awaiting" section appears between the assistant-item and thread list only when gates exist. Each item shows tool name, source thread, description, and inline approve/always/deny buttons. Collapsible with a chevron.
- **Tab badge** — Small amber pill on the Chat tab shows pending count; pulses 3x on new arrivals.
- **Thread gate dots** — Pulsing amber dot on thread items that have pending approvals, visible even when the tray is collapsed.

### Features
- Batch approve-all / deny-all (visible on header hover)
- Click tray item body to jump to its owning thread
- sendApprovalAction now resolves using the gate's owning thread (not the viewed thread)
- Wired into both legacy `approval_needed` and v2 `gate_required`/`gate_resolved` events
- Spring animations for item entry/exit
- Full dark + light theme support
- Touch-friendly targets on coarse pointer devices

### UI Look

image

## Changes
- `index.html` — approval tray skeleton + tab badge span (+23 lines)
- `style.css` — ~270 lines of tray styles, animations, light theme overrides
- `app.js` — `pendingGates` Map, tray CRUD functions, wired into SSE events + sendApprovalAction (+250 lines)

## Test plan
- [ ] Open web gateway, trigger a tool approval → verify inline card AND tray item appear
- [ ] Trigger approval in a background thread → verify tray shows it, active chat does not
- [ ] Approve from tray → verify both tray item and inline card resolve
- [ ] Approve from inline card → verify tray item also resolves
- [ ] Resolve all gates → verify tray disappears, badge disappears
- [ ] Toggle tray collapse/expand
- [ ] Test batch approve-all / deny-all
- [ ] Verify dark and light themes
- [ ] Verify sidebar collapsed state hides tray


henrypark133 (dismissed):
## Review: Approval queue tray in web UI sidebar

Clean, well-designed cross-thread visibility feature for pending approval gates. Solves a real UX gap where background routines or non-active threads could be silently blocked.

Positives:
- No XSS vectors: all user-facing data rendered via `.textContent`, `CSS.escape()` used in selector queries
- Duplicate-safe via `pendingGates.has()` check on SSE reconnect
- Wired into both legacy `approval_needed` and v2 `gate_required`/`gate_resolved` events
- `sendApprovalAction` resolves using the gate's owning thread from `pendingGates` Map rather than `currentThreadId` — correct for cross-thread approval
- Batch approve/deny staggers with `setTimeout(fn, i * 100)` to avoid flooding the API
- Tray auto-hides when empty (zero UI cost), smooth animations with spring easing
- Light/dark theme overrides present
- Touch-friendly targets via `@media (pointer: coarse)`
- Thread gate dots and tab badge pulse provide multiple visual cues

Minor notes:
- **Cross-PR conflict with #2267**: Both modify `sendApprovalAction` in `crates/ironclaw_gateway/static/app.js`. #2267 adds a `threadId` parameter; #2270 uses `pendingGates` Map lookup. Whichever lands second needs a rebase. Recommend merging #2267 first (the scoping fix) then rebasing #2270 to use both approaches: pass threadId when available from card attribute, fall back to pendingGates Map for tray-only gates.
- `aqThreadLabel` truncates UUIDs to 8 chars — fine for display, just noting it's a visual label not a key

LGTM.

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

ilblackdragon (commented):
## Code Review

### Overview
Cross-thread approval queue tray in the web gateway sidebar. Three-layer design: sidebar tray with inline approve/deny actions, tab badge with pending count, and amber dots on thread items with pending gates.

### Bugs Found

| Severity | Issue |
|----------|-------|
| **Critical** | **Cross-thread approval sends wrong `threadId`.** Tray buttons call `sendApprovalAction(entry.request_id, 'approve')` without passing `entry.thread_id`. The fallback chain resolves to `currentThreadId`, which is wrong for background-thread gates. **This defeats the feature's core purpose.** Fix: pass `entry.thread_id` to all tray button handlers and batch actions. |
| **Medium** | **Missing i18n keys.** `approval.awaiting`, `approval.approveAll`, `approval.denyAll` referenced in HTML but missing from `en.js` and `ko.js`. |
| **Medium** | **Badge positioning broken.** `.tab-approval-badge` uses `position: absolute` but parent button has no `position: relative`. Badge positions relative to the wrong ancestor. |

### Other Issues
- `aqUpdateIndicators` calls `debouncedLoadThreads` on every gate add/resolve — re-fetches entire thread list. Consider targeted DOM updates.
- Batch approve/deny uses staggered `setTimeout(fn, i * 100)` — if the user navigates away mid-batch, some gates are left unresolved. Consider `Promise.allSettled`.
- `aqResolveGate` has a race condition when the same `requestId` arrives via both SSE event and local action.

### Accessibility
- No ARIA attributes on tray (`role="region"`, `aria-expanded`, `aria-controls`)
- Batch buttons invisible until hover (`opacity: 0`) — inaccessible to keyboard-only users
- Gate dots have no accessible text beyond `title`

### Security
No XSS vectors — all dynamic content uses `.textContent`, not `.innerHTML`. The one `.innerHTML` use is for hardcoded Unicode. Solid.

### Verdict: **Needs work** — The `threadId` bug is a blocker. Also fix i18n keys and badge positioning before merge.

henrypark133 (changes requested):
## Review: Approval queue tray in web UI sidebar (Risk: Low)

Good design — cross-thread visibility for pending approval gates is a real UX win. SSE event wiring handles both legacy and v2 events correctly, and `.textContent` usage throughout prevents XSS. Light/dark theme and touch targets are well covered.

Positives:
- Solves a genuine visibility gap for background thread approvals
- Clean integration with existing sidebar structure
- `sendApprovalAction` correctly resolves using gate's owning thread

### Critical: Badge positioned relative to `.tab-bar`, not the Chat button [Logic]
**File:** `crates/ironclaw_gateway/static/style.css` — `.tab-approval-badge`
```css
.tab-approval-badge {
  position: absolute;
  top: 6px;
  right: 4px;
}
```
The Chat `<button>` has no `position: relative`, so the nearest positioned ancestor is `.tab-bar` itself. The badge will appear at the top-right corner of the entire tab bar, not on the Chat button. This is visible in every browser.
**Suggested fix:**
```css
.tab-bar button[data-tab="chat"] {
  position: relative;
}
```

### Concerning: Double-resolve with no failure rollback [Logic]
**File:** `crates/ironclaw_gateway/static/app.js` — `sendApprovalAction`
When the user clicks approve/deny: (1) `sendApprovalAction` calls `aqResolveGate()` optimistically, (2) the server broadcasts `gate_resolved` SSE, (3) the SSE listener calls `aqResolveGate()` again. If the POST fails, the tray item is already removed with no way to re-add it. The user sees the gate disappear but the action never took effect.
**Suggested fix:** Remove the optimistic `aqResolveGate(requestId, action)` call from `sendApprovalAction` and rely solely on the server-authoritative `gate_resolved` SSE event. Simpler and more correct.

### Concerning: Unescaped `requestId` in attribute selector [Logic]
**File:** `crates/ironclaw_gateway/static/app.js` — `sendApprovalAction`
```js
const card = document.querySelector('.approval-card[data-request-id="' + requestId + '"]');
```
If `request_id` contains `"` or `\`, this selector breaks. The existing `showApproval` function uses `CSS.escape` for its duplicate check — this lookup should too.
**Suggested fix:** `'.approval-card[data-request-id="' + CSS.escape(requestId) + '"]'`

Convention notes:
- IIFE at bottom relies on function declaration hoisting — consider moving after function definitions for clarity
- `max-height` transition animation is a known CSS anti-pattern (fast snap for short lists, slow finish for tall ones)

henrypark133 (changes requested):
## Re-review: Approval queue tray (Risk: Low)

Good design — cross-thread visibility for pending approval gates is a real UX win. SSE wiring handles both legacy and v2 events, commit 3 addressed threadId routing, badge CSS, resolve race, and i18n.

Positives:
- Removing optimistic `aqResolveGate()` in favor of server-authoritative SSE `gate_resolved` is the right call
- `CSS.escape(requestId)` usage in selectors
- i18n keys for all three locales

### Concerning: Verify `sendApprovalAction` accepts and uses gate's `thread_id`
**File:** `app.js` — Commit 3 description says `entry.thread_id` is passed to all tray button handlers, but the base `sendApprovalAction(requestId, action)` function signature may not accept a `threadId` parameter. If the function always uses `currentThreadId` in the request body, cross-thread tray approvals will misroute. This is the main feature's core scenario — please confirm the function signature was updated to accept and use the gate's thread_id, or clarify if I'm misreading the flow.

### Concerning: `CSS.escape` inconsistency in `sendApprovalAction`
**File:** `app.js` — The `data-request-id` selector in `sendApprovalAction` itself may lack `CSS.escape`, while the same pattern is escaped in `handleGateResolved` and `showApproval`. Should be consistent.

### Minor: CI "Regression test enforcement" failure
The workflow exempts `src/channels/web/static/*` but these files are in `crates/ironclaw_gateway/static/`. Add `skip-regression-check` label or update the workflow path.

### Minor: `handleGateRequired` early return for off-thread gates
The `return` at the non-current-thread branch may prevent tray population for background-thread gates — verify the tray `aqAddGate()` call happens before the early return.

Convention notes:
- No E2E test for cross-thread tray approval — the primary new scenario. Consider a follow-up.

serrrfirat (commented):
Thanks for the thorough re-review! Addressing each point:

### 1. `sendApprovalAction` accepts and uses gate's `thread_id` ✅
Confirmed — the function signature is `sendApprovalAction(requestId, action, threadId)` (commit 3). It resolves `targetThreadId` as: passed `threadId` → card's `data-thread-id` attribute → `currentThreadId` fallback. All tray button handlers (approve, always, deny) and both batch actions (approve-all, deny-all) pass `entry.thread_id` as the third argument. Cross-thread tray approvals route correctly.

### 2. `CSS.escape` inconsistency in `sendApprovalAction` ✅
Fixed in commit 3 — `sendApprovalAction` now uses `CSS.escape(requestId)` in the attribute selector, consistent with `showApproval` and `handleGateResolved`.

### 3. CI "Regression test enforcement" failure ✅
The workflow exempts `src/channels/web/static/*` but the files live in `crates/ironclaw_gateway/static/`. Added `skip-regression-check` label for now. Will follow up with a workflow path fix (requires `workflow` OAuth scope which the current token lacks).

### 4. `handleGateRequired` early return for off-thread gates ✅
The `aqAddGate()` call lives **outside** `handleGateRequired`, in the SSE event listener callback (lines 1012–1023). The flow is: `handleGateRequired(data)` returns early for non-current threads → control returns to the listener → `aqAddGate()` runs unconditionally. Background thread gates populate the tray as expected.

### Convention note: E2E test for cross-thread tray approval
Agreed — will track as a follow-up.

serrrfirat:
## Addressed review feedback (commit 96bdf7a5)

### ilblackdragon — Accessibility
- Added `role="region"`, `aria-label="Approval queue"`, and `aria-expanded` to approval tray
- Batch buttons now visible on keyboard focus (`:focus-within`), not just hover
- Gate dots have `aria-label="Pending approval"` for screen readers

### ilblackdragon — Performance
- Replaced `debouncedLoadThreads()` in `aqUpdateIndicators` with targeted `aqRefreshGateDots()` — updates gate dots in-place without reloading the entire thread list

### ilblackdragon — Batch reliability
- Replaced staggered `setTimeout` with `Promise.allSettled` wrapper that keeps 100ms stagger for UX but ensures all requests are tracked
- `sendApprovalAction` now returns its promise for proper chaining

henrypark133 (changes requested):
## Review: Approval queue tray in web UI sidebar

The UX direction is good, but there is one verified behavior gap that breaks the tray's main promise after a refresh/reconnect.

### Concerning: pending approvals are tray-only in live SSE state, not in recovered state
**File:** `crates/ironclaw_gateway/static/app.js:9260`
`pendingGates` is populated only from live `approval_needed` / `gate_required` SSE events (`app.js:983-1023`). On reload or late entry, the current-thread pending gate is rehydrated from `/api/chat/history` via `handleGateRequired(...)` (`app.js:2972-2976`), but that path never calls `aqAddGate(...)`. So any already-pending approval in another thread disappears from the new sidebar tray until a brand-new SSE event arrives, which defeats the cross-thread visibility this feature is adding.

**Suggested fix:** add an explicit tray rehydration path for persisted pending gates on initial load/reconnect, or extend the existing history/bootstrap flow so recovered gates also repopulate `pendingGates` and the thread-dot/badge state.

serrrfirat:
Addressing the remaining open review threads/comments on the current head:

- Cross-thread approvals are wired through `sendApprovalAction(requestId, action, threadId)` and use `CSS.escape(requestId)` for the selector lookup (`crates/ironclaw_gateway/static/app.js:1438-1446`).
- The tray is rehydrated both from thread history for the active thread (`crates/ironclaw_gateway/static/app.js:2972-2979`) and from `/api/chat/threads` for cross-thread pending approvals (`crates/ironclaw_gateway/static/app.js:3247-3255`).
- The `/api/chat/threads` payload now carries `pending_gates` (`src/channels/web/types.rs:58`, `src/channels/web/handlers/chat.rs:348-373`, `src/channels/web/handlers/chat.rs:396-419`), so pending approvals survive reload/reconnect and repopulate the sidebar tray.
- Earlier review items around missing i18n keys and badge positioning are also present on the current head (`crates/ironclaw_gateway/static/i18n/en.js:316-318`, `crates/ironclaw_gateway/static/i18n/ko.js:316-318`, `crates/ironclaw_gateway/static/style.css:4459-4467`).

I resolved the stale outdated bot threads accordingly. No additional code changes were needed on top of the current branch state.

serrrfirat:
Resolved the merge conflicts by porting the PR’s approval-queue work onto the current `staging` layout.

What changed:
- `staging` deleted the old monolith frontend files (`crates/ironclaw_gateway/static/app.js` and `crates/ironclaw_gateway/static/style.css`) after the JS/CSS split, so I moved the approval tray + badge + cross-thread pending-gate logic into the split modules instead:
  - `crates/ironclaw_gateway/static/js/core/bootstrap.js`
  - `crates/ironclaw_gateway/static/js/core/sse.js`
  - `crates/ironclaw_gateway/static/js/core/history.js`
  - `crates/ironclaw_gateway/static/js/surfaces/chat.js`
  - `crates/ironclaw_gateway/static/styles/base.css`
  - `crates/ironclaw_gateway/static/styles/surfaces/chat.css`
- `staging` also deleted the old web handlers (`src/channels/web/handlers/chat.rs` and `src/channels/web/server.rs`) as part of the gateway feature-slice migration, so I moved the thread-list pending-gate hydration into the current home:
  - `src/channels/web/features/chat/mod.rs`
- Kept the PR’s existing wire/UI additions intact in:
  - `src/channels/web/types.rs`
  - `crates/ironclaw_gateway/static/index.html`
  - `crates/ironclaw_gateway/static/i18n/en.js`
  - `crates/ironclaw_gateway/static/i18n/ko.js`
  - `crates/ironclaw_gateway/static/i18n/zh-CN.js`

Validation I ran:
- `node --check` on the concatenated frontend bundle
- `cargo test --lib --manifest-path .worktrees/pr-2270-mergefix/Cargo.toml chat_threads_handler_includes_pending_gates_without_db`
- `cargo test --lib --manifest-path .worktrees/pr-2270-mergefix/Cargo.toml thread_list_response_`

The branch is pushed and the PR is conflict-free now.

crates/ironclaw_gateway/static/index.html
```diff
@@ -157,7 +157,7 @@ <h2 data-i18n="restart.title">Restart IronClaw Instance</h2>
     <!-- Tab Bar -->
     <div class="tab-bar">
       <div class="tab-indicator" id="tab-indicator"></div>
-      <button class="active" data-tab="chat" data-i18n="tab.chat">Chat</button>
+      <button class="active" data-tab="chat" data-i18n="tab.chat">Chat<span class="tab-approval-badge empty" id="tab-approval-badge"></span></button>
```
serrrfirat, line 160: **High Severity**

This `data-i18n="tab.chat"` binding will delete the new badge span on the first translation pass.

`I18n.updatePageContent()` rewrites `textContent` for every `[data-i18n]` node, so once i18n initializes this `<button>` becomes plain text and the nested `<span id="tab-approval-badge">` is removed from the DOM. After that the approval-count code has nothing to update.

Please move the translatable label onto its own child element (for example `<span data-i18n="tab.chat">Chat</span>`) and keep the badge as a sibling, or otherwise teach the i18n layer not to clobber mixed-content containers.

src/channels/web/features/chat/mod.rs
```diff
@@ -763,12 +763,33 @@ pub(crate) async fn chat_threads_handler(
                     threads.sort_by(|a, b| b.updated_at.cmp(&a.updated_at));
                 }
 
-                let active_thread = session.lock().await.active_thread;
+                let (active_thread, pending_gates) = {
+                    let sess = session.lock().await;
+                    let pending_gates: Vec<PendingGateInfo> = sess
```
serrrfirat, line 768: **High Severity**

This rehydration path only scans `sess.threads[*].pending_approval`, so engine-v2 pending gates disappear after refresh/reconnect.

With `ENGINE_V2=true`, the authoritative pending approvals live in the bridge pending-gate store, not just in the v1 in-memory session threads. I reproduced this locally with a destructive shell request: `/api/chat/history?thread_id=...` returned a pending approval gate, while `/api/chat/threads` returned `pending_gates: null`, so the cross-thread tray cannot recover its state.

Please merge bridge-backed pending gates into `/api/chat/threads` here instead of only reading `sess.threads`.

src/channels/web/features/chat/mod.rs
```diff
@@ -763,12 +763,33 @@ pub(crate) async fn chat_threads_handler(
...
+                        .threads
+                        .iter()
+                        .filter_map(|(tid, thread)| {
+                            thread.pending_approval.as_ref().map(|pa| PendingGateInfo {
+                                request_id: pa.request_id.to_string(),
+                                thread_id: tid.to_string(),
+                                gate_name: "approval".into(),
+                                tool_name: pa.tool_name.clone(),
+                                description: pa.description.clone(),
+                                parameters: serde_json::to_string_pretty(&pa.parameters)
```
serrrfirat, line 778: **High Severity**

This serializes `pa.parameters` into the UI DTO instead of the redacted `display_parameters`.

`PendingApproval` explicitly marks `display_parameters` as the safe field for UI/log/SSE display. Using `parameters` here means a refresh/reconnect can expose raw secret-bearing arguments that were intentionally redacted in the live approval event.

Please serialize `display_parameters` here (with a fallback only if it is genuinely empty) and apply the same fix to each new `PendingGateInfo` constructor added in this PR.

src/channels/web/features/chat/mod.rs
```diff
@@ -763,12 +763,33 @@ pub(crate) async fn chat_threads_handler(
...
+                                    .unwrap_or_default(),
+                                extension_name: None,
+                                resume_kind: serde_json::json!({"Approval":{"allow_always":true}}),
```
serrrfirat, line 781: **High Severity**

This hard-codes `allow_always` to `true` for rehydrated approvals.

For `ApprovalRequirement::Always` tools (for example destructive shell commands), the original pending approval carries `allow_always = false` and the UI must not show an **Always** button. Rewriting the resume kind to `{"Approval":{"allow_always":true}}` means a reload can regrow an **Always** action for a gate that is supposed to require confirmation every time.

Please preserve `pa.allow_always` in the rehydrated payload, and avoid overriding it to `true` again on the frontend snapshot path.

crates/ironclaw_gateway/static/js/core/history.js
```diff
@@ -414,6 +427,19 @@ function loadThreads() {
       list.appendChild(item);
     }
 
+    if (data.pending_gates && data.pending_gates.length > 0) {
+      for (const gate of data.pending_gates) {
```
serrrfirat, line 431: **Medium Severity**

This snapshot path is only additive: it inserts gates into `pendingGates`, but never removes gates that disappeared from the latest server snapshot.

If a gate is resolved from another tab/client, or the browser misses `gate_resolved` during a disconnect, the tray count and per-thread dots can stay stale indefinitely because reconnect only calls `aqAddGate()` again. A refresh/reconnect should reconcile to the authoritative `/api/chat/threads` snapshot, not union new entries into old client state.

Please collect the current snapshot request IDs and prune absent ones from `pendingGates` before/after applying the snapshot.

serrrfirat:
Superseded — most of the changes here are already implemented on staging. We decided not to put an auth gate in the left panel; a new design will come in a follow-up PR.

PR #2294, opened 2026-04-10
feat: add conversation_id to LLM request metadata

standardtoaster:
## Summary
Add stable `conversation_id` to LLM request metadata so stateful backends can route by conversation instead of by ephemeral engine thread IDs.

## Problem
`LlmCallConfig.metadata` (v2 engine) was always an empty HashMap. The OpenAI `user` field was never set. Stateful OpenAI-compatible backends (proxies, session-based providers) need a stable conversation ID for session routing, prompt caching, and cost attribution.

Concrete scenario: an OpenAI-compatible proxy uses the `user` field for session affinity. Each turn gets a new thread ID, so the proxy treats every turn as a new session. Prompt cache is never reused, adding 15-20s cold starts per turn.

## Solution
Three changes, all backward-compatible:

**1. Stamp conversation_id on thread metadata** (`crates/ironclaw_engine/src/runtime/conversation.rs`)
- Add `conversation_id` (the `ConversationId` UUID) to `initial_metadata` when spawning new threads
- Applied before the executor's background task starts (avoids the `set_thread_metadata` race documented in the code)
- Follows the existing pattern for `source_channel` and `user_timezone`

**2. Populate LlmCallConfig.metadata** (`crates/ironclaw_engine/src/executor/orchestrator.rs`)
- Build the metadata HashMap with `thread_id`, `user_id` (from thread fields), and `conversation_id` (from thread.metadata)
- Previously always `HashMap::new()` (empty)

**3. Inject user routing** (`src/llm/rig_adapter.rs`)
- New `inject_user_routing()` encodes `"{user_id}:{session_id}"` in the OpenAI `user` field
- Prefers `conversation_id` (stable) over `thread_id` (per-turn), falls back gracefully
- Called in both `complete()` and `complete_with_tools()`

## Test plan

**Unit tests (9 new, all passing):**

Rig adapter (6 tests):
- `test_inject_user_routing_prefers_conversation_id` — conversation_id wins over thread_id
- `test_inject_user_routing_falls_back_to_thread_id` — uses thread_id when no conversation_id
- `test_inject_user_routing_noop_when_empty` — empty metadata, no additional_params
- `test_inject_user_routing_preserves_existing_params` — model override preserved
- `test_inject_user_routing_user_id_only` — partial metadata still produces output
- `test_inject_user_routing_conversation_id_only` — partial metadata still produces output

Engine pipeline tests (3 tests, using a capturing LLM mock):
- `thread_metadata_includes_conversation_id` — ConversationManager stamps conversation_id on spawned thread
- `llm_call_receives_conversation_id_in_metadata` — full pipeline: conversation_id flows from ConversationManager through orchestrator to LlmCallConfig
- `llm_metadata_omits_conversation_id_when_absent` — backward compat: threads spawned without conversation context omit conversation_id

**Engine tests:** All 340 ironclaw_engine tests pass (337 existing + 3 new).

**Manual validation** (repro script with mock OpenAI server):

```bash
#!/usr/bin/env bash
# Usage: ./test-conversation-id.sh /path/to/ironclaw
set -euo pipefail

BINARY="${1:?Usage: $0 /path/to/ironclaw}"
PORT=3013
GW_PORT=3014
TOKEN=test-tok
TD=$(mktemp -d)
LOG="$TD/llm.log"

trap 'kill $MOCK $IC 2>/dev/null; rm -rf "$TD"' EXIT

python3 -c "
import http.server, json
LOG='$LOG'
class H(http.server.BaseHTTPRequestHandler):
    def do_POST(self):
        body = self.rfile.read(int(self.headers['Content-Length'])).decode()
        open(LOG,'a').write(body+'\n')
        r = json.dumps({'id':'x','object':'chat.completion','choices':[{'index':0,'message':{'role':'assistant','content':'OK'},'finish_reason':'stop'}],'usage':{'prompt_tokens':1,'completion_tokens':1,'total_tokens':2}}).encode()
        self.send_response(200); self.send_header('Content-Type','application/json'); self.send_header('Content-Length',len(r)); self.end_headers(); self.wfile.write(r)
    def log_message(self,*a): pass
http.server.HTTPServer(('127.0.0.1',$PORT),H).serve_forever()
" &
MOCK=$!; sleep 0.3

IRONCLAW_BASE_DIR="$TD/home" ONBOARD_COMPLETED=true \
SECRETS_MASTER_KEY=0000000000000000000000000000000000000000000000000000000000000000 \
LLM_BACKEND=openai_compatible LLM_BASE_URL="http://127.0.0.1:$PORT/v1" \
LLM_API_KEY=k LLM_MODEL=m DATABASE_BACKEND=sqlite LIBSQL_PATH="$TD/test.db" \
HTTP_PORT=3015 HTTP_WEBHOOK_SECRET=test-secret \
GATEWAY_ENABLED=true GATEWAY_AUTH_TOKEN=$TOKEN \
GATEWAY_HOST=127.0.0.1 GATEWAY_PORT=$GW_PORT \
"$BINARY" --no-onboard >"$TD/ic.log" 2>&1 &
IC=$!

for _ in $(seq 30); do
  curl -sf "http://127.0.0.1:$GW_PORT/api/health" >/dev/null 2>&1 && break; sleep 1
done

API="http://127.0.0.1:$GW_PORT/api/chat/send"
curl -sf -X POST "$API" -H "Authorization: Bearer $TOKEN" \
  -H "Content-Type: application/json" -d '{"content":"Hello"}' >/dev/null || true
sleep 2
curl -sf -X POST "$API" -H "Authorization: Bearer $TOKEN" \
  -H "Content-Type: application/json" -d '{"content":"Hi again"}' >/dev/null || true
sleep 1

echo "user fields in LLM requests:"
grep -oE '"user":"[^"]*"' "$LOG" 2>/dev/null || echo "(none)"
```

Before (staging): `(none)` — no user field set.
After (feature branch): same conversation UUID across all turns.

## Scope
- `src/llm/rig_adapter.rs` (+138 lines, mostly tests) — inject_user_routing + 6 tests
- `crates/ironclaw_engine/src/runtime/conversation.rs` (+6 lines code, +157 lines tests) — stamp conversation_id + 3 pipeline tests
- `crates/ironclaw_engine/src/executor/orchestrator.rs` (+11 lines) — populate LlmCallConfig.metadata
- **Provider coverage:** RigAdapter only (covers openai_compatible, openai, anthropic, ollama, tinfoil). Non-rig providers use their own auth for tenant identification. A trait-level decorator could extend coverage in a follow-up.
- **Blast radius:** Low. Metadata is opaque; providers that don't read `user` are unaffected.
- **Breaking changes:** None. The `user` field was not previously set.

### DIFF
--- a/crates/ironclaw_engine/src/executor/orchestrator.rs
+++ b/crates/ironclaw_engine/src/executor/orchestrator.rs
@@ -633,7 +633,10 @@ async fn handle_llm_complete(
             let mut m = HashMap::new();
             m.insert("thread_id".to_string(), thread.id.0.to_string());
             m.insert("user_id".to_string(), thread.user_id.clone());
-            if let Some(conv_id) = thread.metadata.get("conversation_id").and_then(|v| v.as_str())
+            if let Some(conv_id) = thread
+                .metadata
+                .get("conversation_id")
+                .and_then(|v| v.as_str())
             {
                 m.insert("conversation_id".to_string(), conv_id.to_string());
             }
--- a/crates/ironclaw_engine/src/executor/orchestrator.rs
+++ b/crates/ironclaw_engine/src/executor/orchestrator.rs
@@ -629,7 +629,16 @@ async fn handle_llm_complete(
             .and_then(|v| v.as_bool())
             .unwrap_or(false),
         depth: thread.config.depth,
-        metadata: HashMap::new(),
+        metadata: {
+            let mut m = HashMap::new();
+            m.insert("thread_id".to_string(), thread.id.0.to_string());
+            m.insert("user_id".to_string(), thread.user_id.clone());
+            if let Some(conv_id) = thread.metadata.get("conversation_id").and_then(|v| v.as_str())
+            {
+                m.insert("conversation_id".to_string(), conv_id.to_string());
+            }
+            m
+        },
     };
 
     match deps.llm.complete(&messages, &actions, &config).await {
--- a/crates/ironclaw_engine/src/runtime/conversation.rs
+++ b/crates/ironclaw_engine/src/runtime/conversation.rs
@@ -1233,7 +1233,14 @@ mod tests {
         let project = ProjectId::new();
 
         let tid = cm
-            .handle_user_message(conv_id, "Hello", project, "u1", ThreadConfig::default(), None)
+            .handle_user_message(
+                conv_id,
+                "Hello",
+                project,
+                "u1",
+                ThreadConfig::default(),
+                None,
+            )
             .await
             .unwrap();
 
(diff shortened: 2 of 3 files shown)


standardtoaster:

PR #2298, opened 2026-04-10
fix: support cross-backend cheap provider for smart routing

swagasaurus:
## Summary

Support cross-backend cheap provider resolution for smart routing. When `LLM_CHEAP_BACKEND` differs from `LLM_BACKEND`, the cheap provider is now resolved independently through the provider registry instead of cloning the primary backend's config.

Fixes a bug where `create_cheap_provider_for_backend()` always used the primary backend's `RegistryProviderConfig` and only swapped the model name — meaning `LLM_CHEAP_BACKEND=ollama` with `LLM_BACKEND=anthropic` would route hermes3 through Anthropic's API.

## Root Cause

`LLM_CHEAP_BACKEND` / `llm_cheap_backend` DB setting had **zero references in Rust source**. The setting was stored but never consumed by `LlmConfig`, so `create_cheap_provider_for_backend()` had no way to resolve a different provider.

### `src/llm/config.rs`
- Add `cheap_backend: Option<String>` and `cheap_provider: Option<RegistryProviderConfig>` fields to `LlmConfig`

### `src/settings.rs`
- Add `llm_cheap_backend: Option<String>` to `Settings` (DB/serde deserialization)

### `src/config/llm.rs`
- Wire `LLM_CHEAP_BACKEND` env var and `llm_cheap_backend` DB setting into config resolution
- Resolve cheap provider through custom providers (`llm_custom_providers`) and built-in registry
- Priority: DB setting > env var

### `src/llm/mod.rs`
- In `create_cheap_provider_for_backend()`, check `cheap_backend` first
- When set and differs from primary, use the independently resolved `cheap_provider`
- Falls back to existing behavior when `cheap_backend` is unset or matches primary

### `src/llm/models.rs`
- Add new fields to test config defaults

## Backward Compatible

When `cheap_backend` is unset or matches the primary backend, existing behavior is preserved exactly. The new code path only activates when cross-backend routing is explicitly configured.

## Testing

Tested with:
- Primary: `anthropic` / `claude-opus-4-6`
- Cheap: custom provider `home` (Ollama adapter at `http://127.0.0.1:11434`) / `hermes3`
- Verified logs show cheap provider resolving through Ollama instead of Anthropic


standardtoaster:
The new code path (`cheap_backend != backend` in `create_cheap_provider_for_backend`) has no test coverage -- the testing section only shows manual log inspection.

I'd like to see more testing. Unit tests:

- `cheap_backend` set and differs from primary -- provider uses `cheap_provider` config, not primary
- `cheap_backend` set but matches primary -- falls back to existing clone behavior
- `cheap_backend` set to unknown provider name -- falls back to primary with warning
- `cheap_backend` set to `nearai` -- routes through NearAI config, not registry
- `cheap_backend` unset -- existing behavior unchanged

For manual verification without cloud keys: `LLM_CHEAP_BACKEND` must differ from `LLM_BACKEND` to trigger the new path. Set up a custom provider in `llm_custom_providers`, point `LLM_CHEAP_BACKEND` at its ID, and check for `"Creating cross-backend cheap provider"` in `RUST_LOG=ironclaw::llm=debug` output.

Nits:

- NearAI backend detection (`nearai | near_ai | near`) is duplicated between `config/llm.rs:310` and `llm/mod.rs:496`. Extract a helper to prevent drift.
- `if let Some(_def) = registry.find(&cb_lower)` discards the result then `resolve_registry_provider` looks it up again. Minor since it's startup-only.

serrrfirat:
Thank you for finding the cross-backend smart-routing bug. The underlying issue still exists in the current shared LLM code, so we would like to keep this PR open and invite a Reborn-native port.

Today `create_cheap_provider_for_backend` in `crates/ironclaw_llm/src/lib.rs` still clones the primary registry provider and swaps only the model. That means a configuration such as primary Anthropic plus cheap Ollama cannot be represented correctly. The port should resolve a distinct cheap backend/provider through the current provider registry and Reborn configuration source, preserve ownership/secret resolution, and add a test proving requests reach two different provider protocols.

The existing `src/config`, `src/settings`, and v1 wiring should not be carried forward, but the diagnosis and intended behavior remain valuable. If you can rebase this onto `crates/ironclaw_llm` and the Reborn composition path, we would be glad to review it.

ilblackdragon:
Closing this because it updates the legacy top-level `src/` LLM/config/settings implementation (`src/llm/*`, `src/config/llm.rs`, `src/settings.rs`). Product work has moved to the Reborn workspace under `crates/`, and this PR is now outdated against that architecture.

Thanks for the contribution.

PR #2309, opened 2026-04-11
fix(setup): run migrations during onboard when DATABASE_URL preset (#846)

### LINKED ISSUE #846: `onboard` fails with "Failed to save settings to database", but `ironclaw` starts successfully and applies migrations
---

# Description

During initial setup, `ironclaw onboard` fails at the final step with a database error:

```
Error: Database error: Failed to save settings to database: PostgreSQL error: db error
```

However, after this failure, running `ironclaw` directly starts successfully and the application connects to PostgreSQL and applies migrations normally.

This suggests that the onboarding wizard may incorrectly report a database failure even though the database configuration is valid and migrations can run successfully.

---

# Environment

* IronClaw version: `v0.16.1`
* OS: Ubuntu 22.04
* PostgreSQL: `16.13`
* pgvector: `0.8.2`
* Database: local PostgreSQL
* LLM provider: OpenAI-compatible endpoint

---

# Steps to Reproduce

1. Install PostgreSQL and pgvector
2. Create a fresh database

```
dropdb --if-exists ironclaw
createdb ironclaw
```

3. Enable pgvector

```
psql "postgresql://postgres:postgres@localhost:5432/ironclaw?sslmode=disable" \
  -c "CREATE EXTENSION IF NOT EXISTS vector;"
```

4. Run onboarding

```
ironclaw onboard
```

5. Follow setup wizard (example configuration)

* Database: `postgresql://postgres:postgres@localhost:5432/ironclaw`
* Security: **OS keychain**
* Provider: **OpenAI-compatible**
* Base URL: `http://xxxxxx:30000/v1`
* API key required: **No**
* Model: `MiniMax-M2.5-W8A8`
* Embeddings: disabled
* Channels: CLI only
* Docker sandbox: disabled
* Heartbeat: disabled

6. Final step fails with:

```
Error: Database error: Failed to save settings to database: PostgreSQL error: db error
```

---

# Database State After Failure

Immediately after onboarding fails:

```
psql -c "\dt"
```

Output:

```
Did not find any relations.
```

Migration history table also does not exist:

```
ERROR: relation "refinery_schema_history" does not exist
```

---

# Running IronClaw Directly

Running the main program afterwards:

```
ironclaw
```

Startup logs show:

```
INFO PostgreSQL database connected and migrations applied
INFO Configuration reloaded from database
```

The agent starts successfully and the UI is available:

```
Web UI: http://127.0.0.1:3000/
```

The REPL works and the LLM provider initializes correctly.

---

# Expected Behavior

Either:

1. `ironclaw onboard` should run migrations successfully and save settings, or
2. If migrations fail, the error should reflect the real root cause.

Currently the wizard reports a database error even though the database configuration is valid and migrations succeed during normal startup.

---

# Additional Notes

* PostgreSQL connectivity is confirmed
* `pgvector` extension is installed and available
* The issue appears specific to the onboarding wizard

---

ilblackdragon:
## Summary
Fixes #846. The `auto_setup_database()` PostgreSQL early-return paths skipped `test_database_connection_postgres()` and `run_migrations_postgres()` when `DATABASE_URL` was already set, leaving `self.db_pool = None` and causing "Failed to save settings to database" at the final onboard step.

- Both early-return PostgreSQL paths now establish the connection and run migrations
- Added `test_auto_setup_database_runs_migrations_with_existing_env` regression test (libsql variant)

## Test plan
- [x] `cargo test --lib -- setup bootstrap` (125 tests pass)
- [ ] Manual: set DATABASE_URL, run `ironclaw onboard`, verify it completes


ilblackdragon (commented):

The fix itself is directionally right, but the regression coverage does not test the branch that changed.

The bug was in the PostgreSQL early-return path in `auto_setup_database()` when `DATABASE_URL` was already set. The added test only exercises the libSQL branch, so it would not catch a future regression in the Postgres wiring. That violates the repo rule to test through the real caller rather than only proving a similar helper/path.

I want a caller-level regression test that drives `auto_setup_database()` with `DATABASE_URL` preset and proves the wizard ends up with a live DB handle and successful persistence on the PostgreSQL path.

ilblackdragon:
Addressed in ebe8a807: added caller-level integration test driving `auto_setup_database()` with a preset Postgres `DATABASE_URL`, exercising the early-return branch and asserting the wizard ends with a live DB handle + successful persistence (including a round-trip read-back via `Store::get_setting`). Uses `testcontainers-modules` with the `pgvector/pgvector:pg16` image so the migrated schema (which requires pgvector) can actually be applied. Gated behind the `integration` feature; skips gracefully when Docker is unavailable.

src/setup/wizard.rs
```diff
@@ -1074,6 +1074,8 @@ impl SetupWizard {
         {
             if let Ok(url) = std::env::var("DATABASE_URL") {
                 print_info("Using existing PostgreSQL configuration");
+                self.test_database_connection_postgres(&url).await?;
+                self.run_migrations_postgres().await?;
                 self.settings.database_backend = Some("postgres".to_string());
```
serrrfirat: **High Severity -- Fragile implicit coupling: migrations silently no-op if `db_pool` is `None`**

`run_migrations_postgres()` silently returns `Ok(())` when `db_pool` is `None`. Correctness depends on `test_database_connection_postgres` setting `self.db_pool` as a side effect before `run_migrations_postgres` is called (line 5). No assertion or documentation of this coupling.

**Suggested fix:** Add `debug_assert!(self.db_pool.is_some())` between the two calls, or refactor `run_migrations_postgres` to take the pool as a parameter.

src/setup/wizard.rs
```diff
@@ -1074,6 +1074,8 @@ impl SetupWizard {
...
                 self.settings.database_url = Some(url);
```
serrrfirat: **Medium Severity -- Inconsistent state on migration failure**

If `test_database_connection_postgres` (line 5) succeeds but `run_migrations_postgres` (line 6) fails, `db_pool` is set but `settings.database_backend` is not (it's set on line 7, after the `?`). The wizard is left in an inconsistent state.

**Suggested fix:** Pre-existing pattern issue. Consider making `test_database_connection_postgres` return the pool instead of setting it as a side-effect field, and only commit state after both steps succeed.

serrrfirat (commented):
Review of PR #2309 -- 4 findings (1 High, 3 Medium). Two findings could not be placed inline (on lines outside the diff) and are included in this review body.

---

**Medium Severity -- First early-return path (`DATABASE_BACKEND=postgres`) not tested** (`src/setup/wizard.rs`, tests section)

Only the second early-return (`DATABASE_URL`-only) is tested. The first path with `DATABASE_BACKEND=postgres` guard is untested.

**Suggested fix:** Add test variant with `DATABASE_BACKEND=postgres` set alongside `DATABASE_URL`.

---

**Medium Severity -- Case-sensitive `DATABASE_BACKEND` comparison** (`src/setup/wizard.rs`, around line 1073)

`backend == "postgres" || backend == "postgresql"` is case-sensitive. `DATABASE_BACKEND=Postgres` fails the guard silently.

**Suggested fix:** Use `backend.eq_ignore_ascii_case("postgres")`.

ilblackdragon:
Addressed serrrfirat findings in 03cdd8d1:
- Case-insensitive `DATABASE_BACKEND` comparison via `.eq_ignore_ascii_case()`
- `debug_assert!(self.db_pool.is_some())` between connection test and migration
- Added test variant exercising the `DATABASE_BACKEND=postgres` early-return path
- Inconsistent-state-on-migration-failure is a pre-existing pattern outside this PR's scope

ilblackdragon (commented):
All prior review feedback has been addressed:

- Case-insensitive `DATABASE_BACKEND` comparison via `.eq_ignore_ascii_case()` -- good.
- Both PostgreSQL early-return paths now call `test_database_connection_postgres()` + `run_migrations_postgres()` before setting settings and returning -- this was the core bug.
- `debug_assert!` between connection test and migration is a nice safety net for development.
- Three regression tests covering the actual caller (`auto_setup_database`): libSQL unit test, postgres integration test for `DATABASE_URL`-only path, and postgres integration test for `DATABASE_BACKEND=postgres` path. This satisfies the "test through the caller" rule.
- Integration tests use testcontainers with graceful skip when Docker is unavailable -- matches existing patterns.

LGTM -- ready to merge.

src/setup/wizard.rs
```diff
@@ -1146,6 +1142,23 @@ impl SetupWizard {
         }
```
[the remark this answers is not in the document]
ilblackdragon: Addressed in 9bfd99fc. `run_migrations_postgres()` now returns a `SetupError` if `db_pool` is `None` instead of silently returning `Ok(())`, so the ordering contract with `test_database_connection_postgres()` is enforced at runtime in release builds (not just `debug_assert!`). Dropped the now-redundant assert in `finish_postgres_auto_setup`.

src/setup/wizard.rs
ilblackdragon: Addressed in 9bfd99fc. Went with the refactor option: `run_migrations_postgres()` now returns a `SetupError` when `db_pool` is `None`, so the implicit coupling with `test_database_connection_postgres()` is enforced at runtime in release builds too (not just `debug_assert!`). Docstring on the helper documents the ordering requirement.

src/setup/wizard.rs
ilblackdragon: Not fixing in this PR — agree with the "pre-existing pattern issue" framing. The same side-effect-field shape exists at three call sites (wizard.rs:676-681, wizard.rs:2555-2556, and `finish_postgres_auto_setup`), so properly decoupling requires changing `test_database_connection_postgres` to return `Pool` and plumbing it through each caller — larger in scope than the #846 fix.

In practice the partial state is unreachable: on migration failure the wizard returns `Err` and `onboard` aborts; `self.db_pool` and `self.settings` are never read by anyone after that. I'll file a follow-up to refactor the side-effect field globally.

ilblackdragon:
Addressed review comments from @serrrfirat and @Copilot.

**Commit 9bfd99fc:**
- `run_migrations_postgres()` now returns `SetupError` when `db_pool` is `None` instead of silently returning `Ok(())`. Enforces the ordering contract with `test_database_connection_postgres()` at runtime in release builds, addressing the "fragile implicit coupling" finding (inline #3072247965) and the Copilot escalation (inline #3108285489).
- Dropped the now-redundant `debug_assert!` in `finish_postgres_auto_setup` — superseded by the runtime check in the callee.
- Fixed the misleading `tests/workspace_integration.rs` docstring reference flagged by Copilot (that file uses env-PG connect-and-skip, not testcontainers).

**Already addressed in earlier commits d2ff6f36 / 03cdd8d1:**
- Case-sensitive `DATABASE_BACKEND` comparison — now parses via `DatabaseBackend::FromStr` which lowercases and supports aliases (`pg`, `postgres`, `postgresql`). Stronger than `eq_ignore_ascii_case` and aligns with `.claude/rules/types.md`.
- First early-return path (`DATABASE_BACKEND=postgres` guard) is covered by `test_auto_setup_database_runs_migrations_postgres_backend_env`.

**Intentionally deferred:** the "inconsistent state on migration failure" finding (inline #3072247969) is a pre-existing side-effect-field pattern shared by three call sites. Fixing properly requires refactoring `test_database_connection_postgres` to return `Pool` — larger scope than #846. Partial state is unreachable in practice because the wizard aborts on `Err`. Will file a follow-up.

PR #2310, opened 2026-04-11
fix(llm): auto-append /v1 to LLM_BASE_URL for openai_compatible (#1934)

### LINKED ISSUE #1934: openai_compatible: LLM_BASE_URL requires /v1 suffix after rig-core migration
## Description

When using the `openai_compatible` backend with a local model server (MLX, vLLM, llama.cpp), `LLM_BASE_URL` must include `/v1` (e.g. `http://localhost:8080/v1`). Without it, requests go to `/chat/completions` instead of `/v1/chat/completions`, returning 404.

## Cause

`create_openai_compat_from_registry()` uses rig-core's `openai::Client` which does not auto-append `/v1/` to the base URL. The previous `NearAiChatProvider` code path had `api_url()` which handled this:

```rust
fn api_url(&self, path: &str) -> String {
    let base = self.config.base_url.trim_end_matches('/');
    if base.ends_with("/v1") {
        format!("{}/{}", base, path)
    } else {
        format!("{}/v1/{}", base, path)  // auto-appended
    }
}
```

## Impact

Anyone upgrading to the registry-based provider with a local model server that serves on a bare host URL (common for MLX, vLLM) gets silent 404s on all LLM calls.

## Possible fixes

1. Document that `LLM_BASE_URL` must include `/v1` for local servers
2. Normalize the base URL in `create_openai_compat_from_registry()` before passing to rig-core
3. Both

## Reproduction

```bash
# Start any local OpenAI-compatible server on port 8080
LLM_BACKEND=openai_compatible LLM_BASE_URL=http://localhost:8080 LLM_MODEL=qwen3.5 ironclaw
# All LLM calls return 404: {"detail":"Not Found"}

# Fix:
LLM_BASE_URL=http://localhost:8080/v1
```

ilblackdragon:
## Summary
Fixes #1934. Local model servers (MLX, vLLM, llama.cpp) returned 404 when `LLM_BASE_URL` lacked a `/v1` suffix because rig-core's openai client appends `/chat/completions` directly to the base URL.

- Added `normalize_openai_base_url()` in `src/llm/mod.rs` that auto-appends `/v1` when not already present
- Applied in `create_openai_compat_from_registry()`
- 7 unit tests covering bare URL, trailing slash, already-has-`/v1`, path prefixes, etc.

## Test plan
- [x] All tests pass
- [ ] Manual: `LLM_BACKEND=openai_compatible LLM_BASE_URL=http://localhost:8080 ironclaw` succeeds against a local server

### DIFF
--- a/src/llm/mod.rs
+++ b/src/llm/mod.rs
@@ -285,7 +285,8 @@ fn create_openai_compat_from_registry(
 
     let mut builder = openai::Client::builder().api_key(&api_key);
     if !config.base_url.is_empty() {
-        builder = builder.base_url(&config.base_url);
+        let normalized = normalize_openai_base_url(&config.base_url);
+        builder = builder.base_url(&normalized);
     }
     if !extra_headers.is_empty() {
         builder = builder.http_headers(extra_headers);
@@ -707,6 +708,21 @@ pub fn create_gemini_oauth_provider(config: &LlmConfig) -> Result<Arc<dyn LlmPro
     Ok(Arc::new(provider))
 }
 
+/// Normalize an OpenAI-compatible base URL to ensure it ends with `/v1`.
+///
+/// Local model servers (MLX, vLLM, llama.cpp) expect requests at `/v1/chat/completions`,
+/// but rig-core's openai client appends `/chat/completions` directly to the base URL.
+/// This function ensures the `/v1` path segment is present so requests reach the
+/// correct endpoint regardless of whether the user included it.
+fn normalize_openai_base_url(url: &str) -> String {
+    let trimmed = url.trim_end_matches('/');
+    if trimmed.ends_with("/v1") {
+        trimmed.to_string()
+    } else {
+        format!("{}/v1", trimmed)
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
@@ -866,5 +882,61 @@ mod tests {
         // None when nothing configured
         let config = test_llm_config();
         assert_eq!(config.cheap_model_name(), None);
+    }
+
+    #[test]
+    fn test_normalize_openai_base_url_appends_v1_when_missing() {
+        assert_eq!(
+            normalize_openai_base_url("http://localhost:8080"),
+            "http://localhost:8080/v1"
+        );
+    }
+
+    #[test]
+    fn test_normalize_openai_base_url_strips_trailing_slash_then_appends_v1() {
+        assert_eq!(
+            normalize_openai_base_url("http://localhost:8080/"),
+            "http://localhost:8080/v1"
+        );
+    }
+
+    #[test]
+    fn test_normalize_openai_base_url_preserves_existing_v1() {
+        assert_eq!(
+            normalize_openai_base_url("http://localhost:8080/v1"),
+            "http://localhost:8080/v1"
+        );
+    }
+
+    #[test]
+    fn test_normalize_openai_base_url_preserves_v1_with_trailing_slash() {
+        assert_eq!(
+            normalize_openai_base_url("http://localhost:8080/v1/"),
+            "http://localhost:8080/v1"
+        );
+    }
+
+    #[test]
+    fn test_normalize_openai_base_url_with_path_prefix() {
+        assert_eq!(
+            normalize_openai_base_url("https://proxy.example.com/llm"),
+            "https://proxy.example.com/llm/v1"
+        );
+    }
+
+    #[test]
+    fn test_normalize_openai_base_url_with_path_prefix_and_v1() {
+        assert_eq!(
+            normalize_openai_base_url("https://proxy.example.com/llm/v1"),
+            "https://proxy.example.com/llm/v1"
+        );
+    }
+
+    #[test]
+    fn test_normalize_openai_base_url_multiple_trailing_slashes() {
+        assert_eq!(
+            normalize_openai_base_url("http://localhost:8080///"),
+            "http://localhost:8080/v1"
+        );
     }
 }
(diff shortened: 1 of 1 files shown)


ilblackdragon (commented):
Patch looks small and coherent, and the intent matches the documented local-server use case.

Residual concern: this now rewrites every OpenAI-compatible base URL by appending `/v1` unless it already ends in `/v1`. That is correct for the local servers named in the PR, but it bakes in an assumption about all custom proxies. If you keep this approach, I would at least add a caller-level test around the provider factory so we assert the constructed client base URL for both a bare local endpoint and an already-versioned custom endpoint. Right now the coverage is helper-only.

ilblackdragon:
Addressed in 25baf2b6: added caller-level factory test asserting the constructed OpenAICompatibleProvider base URL for both a bare endpoint (`http://localhost:8080` → `.../v1`) and an already-versioned one (`https://proxy/v1` → unchanged, no double suffix).

serrrfirat (commented):
Review of PR #2310 -- 2 findings (both Medium).

src/llm/mod.rs
```diff
@@ -707,6 +718,21 @@ pub fn create_gemini_oauth_provider(config: &LlmConfig) -> Result<Arc<dyn LlmPro
...
+/// Normalize an OpenAI-compatible base URL to ensure it ends with `/v1`.
+///
+/// Local model servers (MLX, vLLM, llama.cpp) expect requests at `/v1/chat/completions`,
+/// but rig-core's openai client appends `/chat/completions` directly to the base URL.
+/// This function ensures the `/v1` path segment is present so requests reach the
+/// correct endpoint regardless of whether the user included it.
+fn normalize_openai_base_url(url: &str) -> String {
+    let trimmed = url.trim_end_matches('/');
+    if trimmed.ends_with("/v1") {
+        trimmed.to_string()
```
serrrfirat, line 731: **Medium Severity -- False positive path-segment matching**

`ends_with("/v1")` is a substring check on the full string. A URL ending in `/apiv1` or `/myv1` would incorrectly match and would NOT get `/v1` appended, silently sending requests to the wrong endpoint.

**Suggested fix:** Check the last path segment explicitly:
```rust
trimmed.rsplit_once('/').map(|(_, last)| last) == Some("v1")
```

src/llm/mod.rs
```diff
@@ -707,6 +718,21 @@ pub fn create_gemini_oauth_provider(config: &LlmConfig) -> Result<Arc<dyn LlmPro
```
serrrfirat, line 706: **Medium Severity -- Debug log shows pre-normalization URL**

This `tracing::debug!` log emits `config.base_url` (the original, pre-normalization URL), but actual requests go to the normalized URL with `/v1` appended (line 22 above). During incident response, this mismatch between logged URL and actual request URL is a debugging trap.

**Suggested fix:** Log the normalized URL, or log both original and normalized values.

ilblackdragon:
Pushed `5455e556` addressing review findings:

- **Fix A**: `ends_with("/v1")` replaced with `rsplit_once('/').map(|(_, last)| last) == Some("v1")` — prevents false positives on paths like `/apiv1`, `/myv1`. Added two regression tests.
- **Fix B**: `tracing::debug!` in `create_openai_compat_from_registry` now logs both `original_base_url` and `normalized_base_url` (via `client.base_url()`).

Quality gate: `cargo fmt` clean, `cargo clippy --all --all-features` zero warnings, `cargo test --lib -- llm` 759/759 pass.

ilblackdragon (commented):
All prior review feedback has been addressed in the current diff:

- `rsplit_once('/').map(|(_, last)| last) == Some("v1")` correctly handles false positives like `/apiv1` and `/myv1` -- two dedicated tests confirm this.
- Debug logging now shows both `original_base_url` and `normalized_base_url` -- no more debugging traps.
- The refactoring to extract `build_openai_compat_client()` enables clean caller-level testing. Three factory tests verify that normalization actually flows through to the constructed rig-core client.
- 9 helper-level normalization tests + 3 caller-level factory tests provide thorough coverage.

One minor observation: `normalize_openai_base_url` with `"http://localhost:8080///"` (multiple trailing slashes) normalizes to `"http://localhost:8080/v1"` via `trim_end_matches('/')`. This is correct behavior but slightly surprising -- `trim_end_matches` strips all trailing slashes, so the intermediate empty path segments are removed. The test covers it, so this is fine.

LGTM -- ready to merge.

ilblackdragon:
Closing — already fixed on `staging`, and this diff would regress the fix. Issue #1934 is closed.

Shipped via:
- #2380 — primary fix (merged 2026-04-13)
- Commit \`fdcbf184\` — initial /v1 auto-append for #1934
- Commit \`b81aa52b\` — combined with image-detail fix
- Commit \`8416dd29\` — refined to scope /v1 normalization to bare host-only URLs

The staging version of \`normalize_openai_base_url\` only appends \`/v1\` when the URL has **no path** (bare \`scheme://host[:port]\`), and is case-insensitive on the \`/v1\` match. This PR appends \`/v1\` whenever the last path segment isn't \`v1\`, which would wrongly mutate custom-path providers:

| Input | Staging | This PR |
|-------|---------|---------|
| \`https://api.z.ai/api/paas/v4\` | unchanged | \`.../v4/v1\` ❌ |
| \`https://generativelanguage.googleapis.com/v1beta/openai\` | unchanged | \`.../openai/v1\` ❌ |
| \`https://example.com/apiv1\` | unchanged | \`.../apiv1/v1\` ❌ |

The \`build_openai_compat_client\` extraction + caller-level tests are a nice pattern, but the staging helper already has unit coverage and the policy there is the one we want.

PR #2311, opened 2026-04-11
fix(config): skip NEARAI URL DNS validation when nearai backend not selected (#1826)

### LINKED ISSUE #1826: NEARAI_AUTH_URL problem
Sometimes my DNS provider fails resolving names and I get this error: NEARAI_AUTH_URL: failed to resolve hostname 'private.near.ai'.
I think is better for ironclaw try not to connect to private.near.ai if the key is not defined.

ilblackdragon:
## Summary
Fixes #1826. `LlmConfig::resolve()` unconditionally called `validate_base_url()` for `NEARAI_AUTH_URL` and `NEARAI_BASE_URL`, performing blocking DNS resolution via `to_socket_addrs()`. When DNS for `private.near.ai` failed, startup failed even when the user wasn't using the nearai backend.

- Wrapped both `validate_base_url()` calls inside `if is_nearai { ... }` guards
- URLs still get default values; DNS only attempted at runtime when nearai is selected
- Regression test uses RFC 6761 `.invalid` hostnames

## Test plan
- [x] `cargo fmt`, `cargo clippy --all --all-features` zero warnings
- [x] All 57 LLM config tests pass

### DIFF
--- a/src/config/llm.rs
+++ b/src/config/llm.rs
@@ -2198,6 +2198,46 @@ mod tests {
         }
     }
 
+    /// Positive companion to `non_nearai_backend_skips_nearai_url_validation`
+    /// (#1826): when the backend IS `nearai`, an unresolvable NEARAI_AUTH_URL
+    /// must still trigger the validation failure the original fix preserved —
+    /// proving the guard only short-circuits when nearai is inactive, never
+    /// when it's the selected backend.
+    #[test]
+    fn nearai_backend_enforces_nearai_url_validation() {
+        let _guard = lock_env();
+        // SAFETY: Under ENV_MUTEX.
+        unsafe {
+            std::env::remove_var("LLM_BACKEND");
+            // Point auth URL at a hostname that will never resolve.
+            std::env::set_var("NEARAI_AUTH_URL", "https://does-not-exist.invalid");
+            std::env::remove_var("NEARAI_BASE_URL");
+            std::env::remove_var("NEARAI_API_KEY");
+        }
+
+        let settings = Settings {
+            llm_backend: Some("nearai".to_string()),
+            ..Default::default()
+        };
+
+        // Must fail — nearai is active, so the unreachable auth URL is a
+        // real misconfiguration that validate_base_url should surface.
+        let err = LlmConfig::resolve(&settings).expect_err(
+            "resolve must fail when nearai is active and NEARAI_AUTH_URL \
+             is unresolvable (#1826 companion)",
+        );
+        let msg = err.to_string();
+        assert!(
+            msg.contains("NEARAI_AUTH_URL"),
+            "expected error to mention NEARAI_AUTH_URL, got: {msg}"
+        );
+
+        // SAFETY: Under ENV_MUTEX.
+        unsafe {
+            std::env::remove_var("NEARAI_AUTH_URL");
+        }
+    }
+
     #[test]
     fn registry_provider_override_base_url_wins_over_env() {
         let _guard = lock_env();
--- a/src/config/llm.rs
+++ b/src/config/llm.rs
@@ -134,10 +134,14 @@ impl LlmConfig {
             );
         }
 
-        // Session config (used by NearAI provider for OAuth/session-token auth)
+        // Session config (used by NearAI provider for OAuth/session-token auth).
+        // Only validate (DNS-resolve) the auth URL when nearai is the active
+        // backend — otherwise startup fails if the hostname is unreachable (#1826).
         let nearai_auth_url = optional_env("NEARAI_AUTH_URL")?
             .unwrap_or_else(|| "https://private.near.ai".to_string());
-        validate_base_url(&nearai_auth_url, "NEARAI_AUTH_URL")?;
+        if is_nearai {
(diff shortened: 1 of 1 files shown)


ilblackdragon (commented):
I did not find a blocking issue in this change.

The guard is in the right place, and the regression test captures the startup-failure class well enough for this specific bug. If anything, I’d add one positive companion test proving validation still happens when `nearai` is actually selected, but I would not block merge on that.

ilblackdragon:
Addressed in 212c7b90: added the positive companion test — when backend = `nearai`, an invalid NEARAI URL still triggers validation failure. Symmetric to the existing skip-when-not-selected test.

src/config/llm.rs
```diff
@@ -172,7 +176,10 @@ impl LlmConfig {
         } else {
             "https://private.near.ai".to_string()
         };
-        validate_base_url(&nearai_base_url, "NEARAI_BASE_URL")?;
+        // Only DNS-resolve the base URL when nearai is active (#1826).
+        if is_nearai {
+            validate_base_url(&nearai_base_url, "NEARAI_BASE_URL")?;
+        }
```
serrrfirat, line 182: **High Severity -- SSRF bypass via embeddings subsystem**

NEARAI base URL is consumed by the embeddings provider even when nearai isn't the LLM backend. Skipping SSRF validation based solely on `is_nearai` opens a window: `NEARAI_BASE_URL=http://169.254.169.254/...` with `EMBEDDING_PROVIDER=nearai` bypasses all IP/scheme checks.

**Suggested fix:** Guard should be `if is_nearai || embeddings_provider_is_nearai`. Or: validate SSRF characteristics (scheme, IP range) unconditionally, only skip DNS resolution conditionally.

src/config/llm.rs
```diff
@@ -134,10 +134,14 @@ impl LlmConfig {
...
-        // Session config (used by NearAI provider for OAuth/session-token auth)
+        // Session config (used by NearAI provider for OAuth/session-token auth).
+        // Only validate (DNS-resolve) the auth URL when nearai is the active
+        // backend — otherwise startup fails if the hostname is unreachable (#1826).
         let nearai_auth_url = optional_env("NEARAI_AUTH_URL")?
             .unwrap_or_else(|| "https://private.near.ai".to_string());
-        validate_base_url(&nearai_auth_url, "NEARAI_AUTH_URL")?;
+        if is_nearai {
+            validate_base_url(&nearai_auth_url, "NEARAI_AUTH_URL")?;
+        }
```
serrrfirat, line 144: **Medium Severity -- Same SSRF issue applies to `nearai_auth_url`**

`SessionManager` is instantiated unconditionally. An unvalidated auth URL could be used for auth-flow SSRF when session renewal is triggered, even if nearai is not the primary LLM backend.

**Suggested fix:** Apply same fix: validate SSRF characteristics (scheme, IP range) unconditionally, only skip DNS resolution conditionally.

serrrfirat (commented):
Review of PR #2311 -- 2 findings (1 High, 1 Medium).

ilblackdragon:
Re: serrrfirat findings on SSRF bypass via embeddings/auth URL —

**SSRF via NEARAI_BASE_URL (embeddings):** `NEARAI_BASE_URL` is an operator-set environment variable, not user input. The operator who controls the environment controls what URLs the system connects to — this is inherent to env-var-based configuration. Validating env vars against SSRF patterns (internal IPs, metadata endpoints) would be a defense-in-depth improvement but is architecturally distinct from the DNS-resolution fix in this PR, which prevents *startup failure* when a non-nearai backend is selected. Filing as a separate hardening issue rather than blocking this fix.

**Same for nearai_auth_url:** Same reasoning — operator-trust boundary. The `SessionManager` is instantiated unconditionally for session renewal, but the URL it connects to is operator-controlled.

Both are valid defense-in-depth suggestions for a follow-up, not regressions introduced by this PR.

ilblackdragon (commented):
All prior review feedback has been addressed:

- The positive companion test (`nearai_backend_enforces_nearai_url_validation`) confirms that DNS validation still fires when nearai IS the active backend -- this was the suggested improvement from the first review.
- The fix is minimal and well-scoped: two `if is_nearai { ... }` guards around the `validate_base_url()` calls. Default URL values are still assigned unconditionally, so if the user later switches to nearai, the URLs are already populated.
- RFC 6761 `.invalid` hostnames in tests are the right choice -- guaranteed to never resolve.
- SSRF defense-in-depth for operator-controlled env vars is correctly identified as a separate concern from this startup-failure fix.
- Environment cleanup in tests uses the guard pattern consistent with the rest of the test file.

LGTM -- ready to merge.

ilblackdragon:
Closing — this is already fixed on `staging`. PR #2080 landed a broader version of the same guard on 2026-04-06, and follow-ups (commits `a1ae697f`, `902aa3ab`, `492d9a9f`) expanded the guard to also cover explicit `NEARAI_*` env overrides, DB-level provider overrides, and NearAI embeddings, plus added a positive companion test.

The guard in this PR (`if is_nearai`) is strictly narrower than what's on `staging`, so merging would regress those cases. No action needed.

PR #2312, opened 2026-04-11
fix(secrets): auto-generate master key with file fallback when keychain unavailable (#1820)

### LINKED ISSUE #1820: 2026-03-31T08:02:03.545982Z ERROR Cannot store API keys: secrets store is not available openroute model can not set
2026-03-31T08:02:03.545982Z ERROR Cannot store API keys: secrets store is not available openroute model can not set

ilblackdragon:
## Summary
Fixes #1820. When neither `SECRETS_MASTER_KEY` env var nor an OS keychain entry existed (headless Linux, skipped onboarding), `SecretsConfig::resolve()` returned `master_key: None`, the secrets store was never created, and configuring API keys failed with "secrets store is not available".

- `SecretsConfig::resolve()` now auto-generates a 32-byte master key when none is found
- Tries OS keychain first; falls back to `~/.ironclaw/master.key` (0o600)
- New `KeySource::File` variant handled in doctor/setup/settings handlers
- File-based fallback validates 64+ hex chars, trims whitespace
- 7 regression tests (round-trip, invalid data, permissions, missing files)

## Test plan
- [x] Unit tests pass
- [ ] Manual: on a system without keychain, verify API keys can be configured

### DIFF
--- a/src/config/secrets.rs
+++ b/src/config/secrets.rs
@@ -2,6 +2,7 @@ use secrecy::{ExposeSecret, SecretString};
 
 use crate::config::helpers::optional_env;
 use crate::error::ConfigError;
+use crate::secrets::keychain::{Keystore, KeystoreError, OsKeystore};
 
 /// Secrets management configuration.
 #[derive(Clone, Default)]
@@ -31,25 +32,98 @@ impl SecretsConfig {
     /// If neither source has a key, auto-generate one and persist it to the
     /// keychain (preferred) or a file at `~/.ironclaw/master.key` (fallback).
     /// This ensures the secrets store is always available without manual setup.
+    ///
+    /// # Keystore failure semantics
+    ///
+    /// The OS keychain is probed via the [`Keystore`] abstraction, which
+    /// distinguishes `NotFound` ("store is reachable but has no entry" --
+    /// a legitimate signal to fall back or generate) from `Unavailable`
+    /// ("store is locked / D-Bus down / permission denied" -- a transient
+    /// failure that MUST NOT be treated as NotFound). A transient failure
+    /// propagates as a startup error rather than silently rotating the
+    /// master key and stranding any previously-encrypted secrets.
     pub(crate) async fn resolve() -> Result<Self, ConfigError> {
+        Self::resolve_with_keystore(
+            &OsKeystore,
+            &Self::master_key_file_path(),
+            /* allow_keystore_persist = */ true,
+        )
+        .await
+    }
+
+    /// Testable core of [`Self::resolve`]. Isolates the keystore
+    /// abstraction and file-fallback path so caller-level tests can
+    /// simulate first-boot, reuse, NotFound-with-file, and transient
+    /// failure modes deterministically.
+    ///
+    /// `allow_keystore_persist` controls whether newly-generated keys
+    /// may be written to the keystore; tests disable this to force the
+    /// file-fallback path.
+    pub(crate) async fn resolve_with_keystore<K: Keystore + ?Sized>(
+        keystore: &K,
+        master_key_file: &std::path::Path,
+        allow_keystore_persist: bool,
+    ) -> Result<Self, ConfigError> {
         use crate::settings::KeySource;
 
         let (master_key, source) = if let Some(env_key) = optional_env("SECRETS_MASTER_KEY")? {
             (Some(SecretString::from(env_key)), KeySource::Env)
         } else {
-            // Probe the OS keychain; if a key is stored, use it
-            match crate::secrets::keychain::get_master_key().await {
+            match keystore.get_master_key().await {
                 Ok(key_bytes) => {
+                    // Key present in keychain -- reuse it. Never rotate.
                     let key_hex: String = key_bytes.iter().map(|b| format!("{:02x}", b)).collect();
                     (Some(SecretString::from(key_hex)), KeySource::Keychain)
                 }
-                Err(_) => {
-                    // Try loading from the file-based fallback
-                    match Self::load_master_key_file() {
+                Err(KeystoreError::NotFound) => {
+                    // Keychain is reachable and authoritatively reports
+                    // no entry. Safe to fall back to file, or generate.
+                    match Self::load_master_key_from_path(master_key_file) {
                         Some(key_hex) => (Some(SecretString::from(key_hex)), KeySource::File),
                         None => {
-                            // No key anywhere — auto-generate and persist
-                            Self::auto_generate_master_key().await
+                            Self::auto_generate_master_key(
+                                keystore,
+                                master_key_file,
+                                allow_keystore_persist,
+                            )
+                            .await
+                        }
+                    }
+                }
+                Err(KeystoreError::Unavailable { reason }) => {
+                    // Transient failure (locked, D-Bus down, permission
+                    // denied, etc.). If a file-based key already exists,
+                    // use it -- that's the key we'd regenerate to anyway
+                    // and preserving it avoids stranding secrets the
+                    // next time the keychain comes back.
+                    //
+                    // If there is NO existing key anywhere, we MUST NOT
+                    // silently generate a new one: doing so could strand
+                    // previously-encrypted secrets that were keyed from
+                    // the real keychain value we couldn't read. Propagate
+                    // the error and fail to start.
+                    match Self::load_master_key_from_path(master_key_file) {
+                        Some(key_hex) => {
+                            tracing::warn!(
+                                "OS keystore unavailable ({reason}); using existing \
+                                 file-based master key at {}",
+                                master_key_file.display()
+                            );
+                            (Some(SecretString::from(key_hex)), KeySource::File)
+                        }
+                        None => {
+                            return Err(ConfigError::InvalidValue {
+                                key: "SECRETS_MASTER_KEY".to_string(),
+                                message: format!(
+                                    "OS keystore is unavailable ({reason}) and no \
+                                     master.key file fallback exists. Refusing to \
+                                     auto-generate a new master key because doing so \
+                                     would silently strand any previously-encrypted \
+                                     secrets once the keystore becomes reachable \
+                                     again. Set SECRETS_MASTER_KEY explicitly, unlock \
+                                     the keystore, or place a master.key file."
+                                ),
+                            });
                         }
                     }
                 }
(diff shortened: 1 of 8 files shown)


ilblackdragon (commented):

`SecretsConfig::resolve()` now treats any `get_master_key()` error as permission to fall back to file lookup or generate a new key. That means a transient keychain / secret-service outage can silently rotate the master key and strand previously encrypted secrets. For this subsystem, "key not found" and "key temporarily unavailable" are not equivalent states.

I also don’t think the test coverage is sufficient yet. The new tests are file-helper tests only. This needs caller-level coverage through config/startup wiring proving we do the right thing on first boot, on key reuse, and on keychain failure modes without silently replacing an existing key.

ilblackdragon:
Addressed in f917be30: the keystore abstraction now distinguishes `NotFound` from transient failure kinds. `SecretsConfig::resolve()` only falls back to file/generate on `NotFound`; transient errors propagate so we never silently rotate the master key. Added caller-level tests covering first-boot, reuse, NotFound-with-file-fallback, transient-with-existing-file, and transient-without-any-existing-key paths using a test keystore stub.

src/config/secrets.rs
```diff
@@ -63,8 +148,512 @@ impl SecretsConfig {
...
+                Err(e) => {
+                    tracing::debug!("Failed to create master.key: {e}");
+                    false
+                }
+            }
+        }
+
+        #[cfg(not(unix))]
+        {
+            match std::fs::write(path, key_hex) {
```
serrrfirat: **Critical Severity — Master key file world-readable on Windows**

On non-Unix platforms, the master key file is written via `std::fs::write()` with no permission restrictions. The 256-bit master key is world-readable by any local user/process. The Unix path correctly sets 0o600.

**Suggested fix:** On Windows, use `std::fs::OpenOptions` with security descriptor APIs to restrict to current user. At minimum, warn that file permissions are unrestricted and recommend `SECRETS_MASTER_KEY` env var on Windows.

src/config/secrets.rs
```diff
@@ -63,8 +148,512 @@ impl SecretsConfig {
         })
     }
 
+    /// Auto-generate a master key and try to persist it.
+    ///
+    /// Tries the OS keystore first (when `allow_keystore_persist` is
+    /// true); falls back to `master_key_file`.
+    async fn auto_generate_master_key<K: Keystore + ?Sized>(
```
serrrfirat, line 159: **High Severity — TOCTOU race on concurrent first-boot**

Two processes starting simultaneously both reach `NotFound`, generate different keys, last writer wins with `create(true).truncate(true)`. First process's secrets are stranded.

**Suggested fix:** Use `OpenOptions::new().write(true).create_new(true)` (exclusive create). If file exists, load existing key instead of overwriting.

src/config/secrets.rs
```diff
@@ -63,8 +148,512 @@ impl SecretsConfig {
...
+        keystore: &K,
+        master_key_file: &std::path::Path,
+        allow_keystore_persist: bool,
+    ) -> (Option<SecretString>, crate::settings::KeySource) {
+        use crate::settings::KeySource;
+
+        let key_bytes = crate::secrets::keychain::generate_master_key();
+        let key_hex: String = key_bytes.iter().map(|b| format!("{:02x}", b)).collect();
```
serrrfirat: **High Severity — Key material in unzeroed heap memory**

Generated key bytes are converted to a `String` (`key_hex`) before being wrapped in `SecretString`. The intermediate `String` is a plain heap allocation not zeroed on drop, leaving key material in memory indefinitely.

**Suggested fix:** Use `secrecy::SecretString` or `zeroize::Zeroizing<String>` for the intermediate `key_hex`, or convert to `SecretString` immediately after hex encoding.

src/config/secrets.rs
```diff
@@ -28,20 +29,104 @@ impl SecretsConfig {
...
+                            Self::auto_generate_master_key(
+                                keystore,
+                                master_key_file,
+                                allow_keystore_persist,
+                            )
+                            .await
+                        }
+                    }
+                }
+                Err(KeystoreError::Unavailable { reason }) => {
```
serrrfirat, line 98: **High Severity — Keychain/file key divergence causes data loss**

When keychain returns `Unavailable` and file key exists, the file key is used. But when keychain recovers on next boot, the keychain key is used instead. Secrets encrypted under the file key during the outage are stranded.

**Suggested fix:** When keychain is reachable AND file key exists, verify they match. If they differ, warn loudly and fail rather than silently switching. Or: once file key is created, always prefer it.

src/settings.rs
```diff
@@ -245,6 +245,9 @@ pub enum KeySource {
     Keychain,
     /// User provides via SECRETS_MASTER_KEY env var.
     Env,
+    /// Auto-generated key stored in `~/.ironclaw/master.key` file
+    /// (fallback when OS keychain is unavailable).
+    File,
```
serrrfirat, line 250: **Medium Severity — New `File` variant breaks serde backwards compatibility**

`KeySource::File` serializes as `"file"`. If user upgrades then downgrades, the old version can't deserialize `"file"` — `Settings` fails to load entirely.

**Suggested fix:** Add `#[serde(other)]` to `None` variant for graceful degradation on unknown variants.

src/config/secrets.rs
```diff
@@ -63,8 +148,512 @@ impl SecretsConfig {
...
+    fn master_key_file_path() -> std::path::PathBuf {
+        crate::bootstrap::ironclaw_base_dir().join("master.key")
+    }
+
+    /// Load and validate a master key from a specific file path.
+    fn load_master_key_from_path(path: &std::path::Path) -> Option<String> {
+        match std::fs::read_to_string(path) {
+            Ok(contents) => {
+                let trimmed = contents.trim().to_string();
+                if trimmed.len() >= 64 && trimmed.chars().all(|c| c.is_ascii_hexdigit()) {
```
serrrfirat: **Medium Severity — Key length validation uses >= 64 instead of == 64**

Accepts keys longer than 64 hex chars. A 128-char key (64 bytes) passes validation. Whether AES accepts oversized keys depends on the crypto library.

**Suggested fix:** Use exact length check `trimmed.len() == 64`.

src/config/secrets.rs
```diff
@@ -28,20 +29,104 @@ impl SecretsConfig {
...
+                    // next time the keychain comes back.
+                    //
+                    // If there is NO existing key anywhere, we MUST NOT
+                    // silently generate a new one: doing so could strand
+                    // previously-encrypted secrets that were keyed from
+                    // the real keychain value we couldn't read. Propagate
+                    // the error and fail to start.
+                    match Self::load_master_key_from_path(master_key_file) {
+                        Some(key_hex) => {
+                            tracing::warn!(
```
serrrfirat: **Medium Severity — Master key file path logged in warning**

`warn!("...using existing file-based master key at {path}")` reveals key file location. Also, `warn!` corrupts REPL/TUI per CLAUDE.md logging guidelines.

**Suggested fix:** Use `debug!` level, omit full path.

src/config/secrets.rs
```diff
@@ -63,8 +148,512 @@ impl SecretsConfig {
...
+    }
+
     /// Get the master key if configured.
     pub fn master_key(&self) -> Option<&SecretString> {
         self.master_key.as_ref()
     }
 }
+
+#[cfg(test)]
+mod tests {
```
serrrfirat, line 312: **Medium Severity — Missing edge case tests**

No test for key > 64 hex chars. No test for concurrent key generation race. Both are identified vulnerability paths.

**Suggested fix:** Add tests for oversized key and concurrent first-boot scenario.

src/secrets/keychain.rs
```diff
@@ -54,6 +116,30 @@ mod platform {
 
     use super::*;
 
+    /// macOS `errSecItemNotFound` status code.
+    /// See <https://developer.apple.com/documentation/security/errsecitemnotfound>.
+    const ERR_SEC_ITEM_NOT_FOUND: i32 = -25300;
```
serrrfirat, line 121: **Medium Severity — Narrow macOS error code matching**

Only `ERR_SEC_ITEM_NOT_FOUND` (-25300) maps to `NotFound`. Other "not found"-like codes (`errSecNoSuchKeychain` -25294, `errSecNoSuchAttr` -25303) fall to `Unavailable`, potentially causing silent key rotation on misconfigured systems.

**Suggested fix:** Widen `NotFound` match or document expected error code mapping.

serrrfirat (commented):
Security review of master key file-fallback implementation. 9 findings across 3 files (1 Critical, 3 High, 5 Medium).

ilblackdragon:
## Security review fixes applied — `82ba30ca`

| Finding | Fix | File |
|---------|-----|------|
| **A — TOCTOU race on concurrent first-boot** | `create_new(true)` replaces `create(true).truncate(true)`; `AlreadyExists` loads the winner's key | `src/config/secrets.rs` |
| **B — Key hex in unzeroed heap memory** | `key_hex` wrapped in `Zeroizing<String>`; `zeroize = "1"` added to Cargo.toml | `src/config/secrets.rs`, `Cargo.toml` |
| **C — Key length >= 64 instead of == 64** | Validation changed to `== 64` (AES-256 = exactly 32 bytes = 64 hex) | `src/config/secrets.rs` |
| **D — warn! with file path → debug!** | `warn!` → `debug!`; file path omitted from log message | `src/config/secrets.rs` |
| **E — Windows file permission warning** | `#[cfg(not(unix))]` path logs `debug!` about unrestricted perms | `src/config/secrets.rs` |
| **F — Widen macOS error code matching** | Added `-25294` (errSecNoSuchKeychain) and `-25303` (errSecNoSuchAttr) → `NotFound` | `src/secrets/keychain.rs` |
| **G — serde backwards compat for KeySource** | `#[serde(other)]` on `None` variant | `src/settings.rs` |
| **H — Edge case tests** | `test_load_master_key_rejects_oversized_key` + `test_concurrent_first_boot_create_new_exclusivity` | `src/config/secrets.rs` |

**Quality gate:** `cargo fmt` ✓ · `cargo clippy --all --all-features` 0 warnings ✓ · `cargo test --lib -- config::secrets` 16/16 pass ✓

ilblackdragon:
All security review findings from @serrrfirat have been addressed in 82ba30ca:

| Finding | Severity | Fix |
|---------|----------|-----|
| Windows perms | Critical | `#[cfg(not(unix))]` logs `debug!` warning about unrestricted perms |
| TOCTOU race | High | `create_new(true)` with `AlreadyExists` fallback |
| Unzeroed key material | High | `key_hex` wrapped in `Zeroizing<String>` |
| Keychain/file divergence | High | `NotFound` vs transient failure distinction (dece4e44) prevents silent key rotation |
| serde compat | Medium | `#[serde(other)]` on `None` variant |
| Key length | Medium | Exact `== 64` check |
| Logging | Medium | `warn!` to `debug!`, path omitted |
| Edge case tests | Medium | Oversized key + concurrent first-boot tests added |
| macOS error codes | Medium | Added `-25294` and `-25303` to `NotFound` match |

Test coverage looks solid with 16 passing tests. All 3 commits build a clean progression: initial fix -> keystore abstraction for NotFound vs transient -> security hardening. Ready for merge once @serrrfirat confirms the fixes.

serrrfirat (approved):

ilblackdragon:
Closing in favor of a smaller, reuse-focused fix.

The wizard's \`auto_setup_security()\` (src/setup/wizard.rs:1199-1254) already implements the correct fallback chain for #1820:

1. \`SECRETS_MASTER_KEY\` env var → use it
2. Existing keychain key → use it
3. Generate + keychain store → use it
4. **Keychain unavailable → generate + \`inject_single_var\` + persist to \`~/.ironclaw/.env\` as \`SECRETS_MASTER_KEY\`** (via \`write_bootstrap_env()\` → \`upsert_bootstrap_vars()\`)

This PR adds a second persistence format (\`~/.ironclaw/master.key\`) + \`KeySource::File\` variant + \`Keystore\` trait alongside that existing chain, which:
- Duplicates the generate/persist logic
- Introduces divergence-on-recovery (keychain with key A + file with key B) — flagged by security review, not addressed
- Expands the \`KeySource\` enum and serde surface
- Adds ~430 lines when the root issue is that \`SecretsConfig::resolve()\` doesn't reuse the wizard's persistence path

The right fix is to **move** the wizard's chain into \`SecretsConfig::resolve()\` so it runs on every startup (not just onboarding), and collapse \`auto_setup_security()\` to a thin caller. That keeps one persistence format (\`.env\`), one \`KeySource\` (\`Env\`), and zero divergence risk.

Opening a replacement PR with that approach.

PR #2313, opened 2026-04-11
fix(config): env vars take precedence over DB-persisted LLM settings (#1680)

### LINKED ISSUE #1680: Railway one‑click IronClaw template: LLM_* env vars not visible at runtime, model stuck on Anthropic Sonnet
Hi, I deployed IronClaw using the official one‑click Railway template and I’m running into an issue where model configuration env vars are not being picked up by the running process.

Setup:

Platform: Railway

Template: IronClaw one‑click template (ironclaw + Postgres)

Service: ironclaw (app) + Postgres only

LLM provider: OpenRouter (openai_compatible backend)

On the ironclaw Variables tab in Railway I have (among others):

LLM_BACKEND = openai_compatible

LLM_BASE_URL = https://openrouter.ai/api/v1

LLM_API_KEY = <my OpenRouter key>

LLM_MODEL = google/gemini-3.1-flash-lite-preview

OPENAI_API_KEY = <same OpenRouter key>

OPENAI_MODEL_NAME = google/gemini-3.1-flash-lite-preview

OPENROUTER_API_KEY = <same key>

OPENROUTER_KEY = <same key>

I’ve verified these are saved on the service (not project‑level) and have restarted/redeployed ironclaw multiple times.

Problem:

Inside the running IronClaw container, these vars do not exist at all.

From Flash / the shell:

env | egrep '^(LLM_|OPENAI_|OPENROUTER_)' → no output

IronClaw reports the active model as anthropic/claude-sonnet-4.6 and uses that for all requests.

Earlier, before I tried to switch to Gemini, Flash saw:

OPENAI_MODEL_NAME = nvidia/nemotron-3-super-120b-a12b
and used that model.

After I changed everything over to Gemini in Railway, that Nemotron value disappeared from the runtime, but now none of the LLM_* / OPENAI_* / OPENROUTER_* vars are present, and IronClaw defaults to anthropic/claude-sonnet-4.6.

What I think is happening:

It looks like the IronClaw process launched by the one‑click Railway template is not inheriting the service‑level env vars (LLM_, OPENAI_, OPENROUTER_*), and instead uses an internal default provider/model (Anthropic Sonnet) when nothing usable is seen.

Because this is a one‑click template, I don’t have access to the Dockerfile or entrypoint to confirm or patch how env is loaded before starting the Rust binary.

What I’m asking for:

Confirmation of how the one‑click Railway template is launching IronClaw, and

A fix or guidance so that service‑level env vars like LLM_MODEL, LLM_API_KEY, LLM_BACKEND, LLM_BASE_URL, and/or OPENAI_MODEL_NAME are actually visible to the IronClaw process at runtime.

Even a brief note like “here’s the entrypoint script / env loading line you need to add” or “this template currently ignores LLM_* env, use X instead” would be hugely helpful.

Happy to provide:

Screenshots of the Railway Variables tab for ironclaw

Full output from env and any relevant IronClaw logs

Exact commands/prompts I used to generate the diagnostic reports

Thanks in advance for any guidance—right now the template works, but I’m stuck on the default Anthropic model and can’t get IronClaw to honor the configured OpenRouter/Gemini settings.

ilblackdragon:
## Summary
Fixes #1680. On Railway (and similar platforms), `LLM_*` env vars set in the platform's Variables UI were ignored because DB-persisted settings from onboarding took priority. Old order: DB > env > default. New order follows 12-factor: **env > DB > default**.

Applied consistently across all LLM resolution paths:
- `LLM_BACKEND`, `LLM_MODEL`, `LLM_BASE_URL`, `LLM_API_KEY`
- NearAI: `NEARAI_API_KEY`, `NEARAI_MODEL`, `NEARAI_BASE_URL`
- Bedrock: `BEDROCK_REGION`, `BEDROCK_MODEL`, `BEDROCK_CROSS_REGION`, `AWS_PROFILE`
- OpenAI Codex: `OPENAI_CODEX_MODEL`, `OPENAI_MODEL`

13 tests updated, module doc updated to document the LLM exception.

## Test plan
- [x] `cargo fmt`, clippy, all tests pass
- [ ] Manual: deploy to Railway with `LLM_*` vars set, verify they take effect

### DIFF
--- a/src/config/llm.rs
+++ b/src/config/llm.rs
@@ -56,15 +56,18 @@ impl LlmConfig {
         }
     }
 
-    /// Resolve a model name from settings.selected_model -> env var -> hardcoded default.
+    /// Resolve a model name from env var -> settings.selected_model -> hardcoded default.
+    ///
+    /// Env vars take priority over DB settings so cloud deployments can
+    /// override via platform Variables UI (12-factor app behavior).
     fn resolve_model(
         env_var: &str,
         settings: &Settings,
         default: &str,
     ) -> Result<String, ConfigError> {
-        if let Some(model) = Self::selected_model_override(settings) {
+        if let Some(model) = optional_env(env_var)? {
             Ok(model)
-        } else if let Some(model) = optional_env(env_var)? {
+        } else if let Some(model) = settings.selected_model.clone() {
             Ok(model)
         } else {
             Ok(default.to_string())
@@ -74,11 +77,16 @@ impl LlmConfig {
     pub(crate) fn resolve(settings: &Settings) -> Result<Self, ConfigError> {
         let registry = ProviderRegistry::load();
 
-        // Determine backend: db settings > env var > default ("nearai")
-        let (backend, backend_source) = if let Some(ref b) = settings.llm_backend {
-            (b.clone(), "db:llm_backend")
-        } else if let Some(b) = optional_env("LLM_BACKEND")? {
+        // Determine backend: env var > db settings > default ("nearai")
+        //
+        // Env vars take priority over DB settings for LLM config so that
+        // cloud deployments (Railway, Render, Fly.io) can override the
+        // onboarding-persisted backend via their Variables UI — standard
+        // 12-factor app behavior.
+        let (backend, backend_source) = if let Some(b) = optional_env("LLM_BACKEND")? {
             (b, "env:LLM_BACKEND")
+        } else if let Some(ref b) = settings.llm_backend {
+            (b.clone(), "db:llm_backend")
         } else {
             ("nearai".to_string(), "default")
         };
@@ -91,16 +99,17 @@ impl LlmConfig {
                 "Resolving LLM backend"
             );
         });
-        // Warn operators when a DB-persisted value silently overrides LLM_BACKEND.
-        if backend_source == "db:llm_backend"
-            && let Ok(env_val) = std::env::var("LLM_BACKEND")
-            && !env_val.is_empty()
+        // Inform operators when an env var overrides a DB-persisted value.
+        if backend_source == "env:LLM_BACKEND"
+            && settings
+                .llm_backend
+                .as_ref()
+                .is_some_and(|db_val| db_val != &backend)
         {
-            tracing::warn!(
-                db_value = %backend,
-                env_value = %env_val,
-                "LLM_BACKEND env var is set but DB setting takes priority. \
-                 Unset llm_backend in the DB (via settings UI) to use the env var."
+            tracing::info!(
+                env_value = %backend,
+                db_value = ?settings.llm_backend,
+                "LLM_BACKEND env var overrides DB setting"
             );
         }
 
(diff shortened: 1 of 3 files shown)


ilblackdragon (commented):
I did not find a blocking issue in the implementation.

The precedence change is internally consistent and the tests were updated in the right areas. One thing to watch operationally is that this creates an explicit LLM-only exception to the repo’s normal DB-over-env rule, so any future settings surfaces need to preserve that split deliberately. The doc note helps.

src/config/llm.rs
```diff
@@ -56,15 +56,18 @@ impl LlmConfig {
         }
     }
 
-    /// Resolve a model name from settings.selected_model -> env var -> hardcoded default.
+    /// Resolve a model name from env var -> settings.selected_model -> hardcoded default.
+    ///
+    /// Env vars take priority over DB settings so cloud deployments can
+    /// override via platform Variables UI (12-factor app behavior).
     fn resolve_model(
         env_var: &str,
```
serrrfirat, line 64: **High Severity -- `resolve_model()` skips `builtin_overrides`**

`resolve_model()` goes `env > selected_model > default`, skipping `builtin_overrides`. If called for a backend that uses `builtin_overrides` but has no `selected_model`, the override is silently ignored.

**Suggested fix:** Audit all call sites. If only called for Ollama (no `builtin_overrides`), document this. Otherwise add `builtin_overrides` fallback step.

src/config/llm.rs
```diff
@@ -91,16 +99,17 @@ impl LlmConfig {
...
+                .as_ref()
+                .is_some_and(|db_val| db_val != &backend)
         {
-            tracing::warn!(
-                db_value = %backend,
-                env_value = %env_val,
-                "LLM_BACKEND env var is set but DB setting takes priority. \
-                 Unset llm_backend in the DB (via settings UI) to use the env var."
+            tracing::info!(
+                env_value = %backend,
```
serrrfirat, line 117: **High Severity -- `info!` corrupts REPL/TUI**

Override notification uses `tracing::info!`. Per CLAUDE.md: "info! and warn! output appears in the REPL and corrupts the terminal UI. Use debug! for internal diagnostics."

**Suggested fix:** Change to `tracing::debug!`.

src/config/llm.rs
```diff
@@ -146,26 +155,28 @@ impl LlmConfig {
...
+        let nearai_api_key = optional_env("NEARAI_API_KEY")?
+            .map(SecretString::from)
+            .or_else(|| {
+                nearai_override
+                    .and_then(|o| o.api_key.as_ref())
+                    .map(|key| SecretString::from(key.clone()))
+            });
+        // Model priority: env > selected_model (DB) > builtin_overrides (DB) > default
+        let nearai_model = if let Some(model) = optional_env("NEARAI_MODEL")? {
             model
```
serrrfirat, line 176: **Medium Severity -- NearAI model resolution may return wrong model when backend changes**

If backend changes via env var (e.g., `LLM_BACKEND=openai`), `selected_model_override()` could return a model intended for the old backend, since NearAI config is always resolved (used for embeddings).

**Suggested fix:** Guard `selected_model` with backend check in NearAI-specific path.

serrrfirat (commented):
Review of PR #2313 -- 3 findings (2 High, 1 Medium).

ilblackdragon:
Pushed 9b295765 addressing review findings:

**Fix A — resolve_model() builtin_overrides fallback:** Added `backend` parameter to `resolve_model()` and a new fallback step between `selected_model` and `default` that checks `settings.llm_builtin_overrides.get(backend).and_then(|o| o.model.clone())`. Updated the sole caller (Gemini OAuth path) to pass `"gemini_oauth"` as the backend key. This aligns `resolve_model()` with the inline resolution in `resolve_registry_provider()` and the NearAI path.

**Fix B — info! → debug!:** Changed the "LLM_BACKEND env var overrides DB setting" log from `tracing::info!` to `tracing::debug!` per CLAUDE.md logging rules.

**Quality gate:** `cargo fmt` clean, `cargo clippy` zero warnings (lib/tests/examples), 60/60 `config::llm` tests pass.

ilblackdragon:
Review findings from @serrrfirat addressed in 9b295765:

| Finding | Severity | Status |
|---------|----------|--------|
| `resolve_model()` skips `builtin_overrides` | High | Fixed -- added `backend` param and fallback step between `selected_model` and `default` |
| `info!` corrupts REPL/TUI | High | Fixed -- changed to `debug!` |
| NearAI model resolution with backend change | Medium | Worth confirming -- the NearAI path resolves separately for embeddings; need to verify the `selected_model` guard was added |

The 12-factor precedence change (env > DB > default) is clean and well-tested. The LLM-only exception to the normal DB-over-env rule is documented in the module. Ready for @serrrfirat to confirm fixes.

serrrfirat (approved):

serrrfirat (commented):
## Pushback on the approach

The 12-factor intent makes sense for cloud deployments, but I think this creates a bigger problem than it solves.

### The reported issue (#1680) might not be fixed by this change

The reporter says:

> `env | egrep '^(LLM_|OPENAI_|OPENROUTER_)' → no output`

The env vars literally don't exist in the container. Flipping precedence from `DB > env` to `env > DB` doesn't help when `optional_env()` returns `None` — it still falls through to DB. The actual bug appears to be that the Railway template isn't passing env vars to the container process, which is a deployment/template issue, not a resolution order issue.

### Split-brain config precedence

The config system has a documented rule (`src/config/mod.rs:3`):

> Settings are loaded with priority: **DB/TOML > env > default**.

This PR carves out an LLM-only exception that inverts that rule. Every future developer touching config resolution now has to remember "LLM is backwards from everything else." The doc note helps, but it's a footgun — especially since LLM config is the most commonly modified config in the system.

### Silent override risk for local/desktop users

This is the main concern. Users who go through onboarding and configure their LLM via the wizard get those values persisted to the DB. If they have stale `LLM_BACKEND`, `LLM_MODEL`, or `LLM_API_KEY` in their `.zshrc`/`.bashrc` from a previous experiment (or set for another AI tool), their carefully-configured DB settings are now silently ignored. No warning — just the wrong model or provider. This is the primary use case (local personal AI assistant), and it's now the one that breaks quietly.

### Suggested alternatives

- **Opt-in deployment mode**: An env var like `IRONCLAW_ENV_PRIORITY=true` that cloud templates set explicitly, flipping precedence only when opted in. Keeps local users safe, gives cloud deployments exactly what they need.
- **Fix the Railway template**: If env vars aren't being inherited by the container, that's a template/Dockerfile issue — fix it there.
- **Cloud-first onboarding**: Let cloud deployments skip or clear onboarding-persisted LLM settings so there's nothing in the DB to conflict with env vars. Then the existing `DB > env` order works perfectly — empty DB means env wins naturally.

Any of these would solve the Railway problem without regressing local users. Happy to help implement whichever direction we go.

serrrfirat (changes requested):
comment

ilblackdragon:
We actually explicitly wanted to override LLM parameters from env via DB settings to allow user to change their LLM providers.

Going to close this as not relevant.

Issue #1846, opened 2026-04-01
Issue: Data loss with upgrading from hosted ui


Kampouse:
"Note to self: Don't do this. I was on version 0.24.0 in the SSH. Tried the update instance button in UI. 
Reverted back v0.21.0. Lost a bunch of things I has built, thankfully backed up. Lost telegram connection, etc.
 Upgrade is SSH in terminal not in UI.

Issue #26, opened 2026-02-10
How would you integrate native diction?


elliotBraem:
So agent can listen to me speak in full privacy?

PR #2314, opened 2026-04-11
feat(version): track running version, detect downgrades, warn on restart (#1846)

ilblackdragon:
## Summary
Addresses #1846. The web UI restart button calls `process::exit(0)`; if the Docker tag is mutable (e.g. `latest`), the container may come back on an older image, causing schema mismatch and data loss.

- **New `src/version.rs`**: Persists running version in DB settings under `system/ironclaw.version`. Compares via semver to detect `Fresh`/`Unchanged`/`Upgraded`/`Downgraded` transitions (5 unit tests)
- **Startup downgrade detection** in `src/app.rs`: Logs `ERROR` on downgrade with actionable message
- **Enhanced restart dialog**: Shows current version, warns about mutable image tags
- **Post-restart SSE handler**: Compares pre/post versions, alerts user via system message if changed
- i18n strings added for English and Chinese

## Test plan
- [x] `cargo fmt`, clippy, all 4038 unit tests pass
- [ ] Manual: simulate downgrade, verify ERROR log on startup


ilblackdragon (commented):
The mitigation is sensible and the placement in startup is architecturally reasonable.

I did not find a merge blocker, but the coverage still feels thin for the class of failure being addressed. The new logic spans DB-backed version recording, startup transition detection, and browser-visible restart/reconnect behavior. I’d prefer at least one caller-level test that drives the startup path through persisted version state, plus one browser/gateway test for the reconnect warning when the version changes.

ilblackdragon:
Addressed in 433ff3d5: added caller-level startup test covering the full version-transition matrix (first-boot/same/upgrade/downgrade) through persisted state, plus a gateway-level test asserting the reconnect warning fires when persisted version changes across boots.

tests/ws_gateway_integration.rs
```diff
@@ -28,6 +28,54 @@ use ironclaw_common::AppEvent;
...
+        oauth_rate_limiter: ironclaw::channels::web::server::RateLimiter::new(10, 60),
+        webhook_rate_limiter: ironclaw::channels::web::server::RateLimiter::new(10, 60),
+        registry_entries: Vec::new(),
+        cost_guard: None,
+        routine_engine: Arc::new(tokio::sync::RwLock::new(None)),
+        startup_time: std::time::Instant::now(),
+        active_config: ironclaw::channels::web::server::ActiveConfigSnapshot::default(),
+        secrets_store: None,
+        db_auth: None,
+        previous_version,
```
serrrfirat, line 78: **Critical Severity -- Won't compile on staging**

`start_test_server_with_previous_version` constructs `GatewayState` missing 12 fields that exist on current staging: `auth_manager`, `pairing_store`, `oauth_providers`, `oauth_state_store`, `oauth_base_url`, `oauth_allowed_domains`, `near_nonce_store`, `near_rpc_url`, `near_network`, `oauth_sweep_shutdown`, `frontend_html_cache`, `tool_dispatcher`.

**Suggested fix:** Add missing fields (all `None`/default) to match existing `start_test_server` helper.

src/version.rs
```diff
@@ -0,0 +1,233 @@
...
+            std::cmp::Ordering::Greater
+        );
+    }
+
+    #[test]
+    fn test_compare_semver_with_prerelease() {
+        // Pre-release suffixes are stripped for comparison
+        assert_eq!(
+            compare_semver("0.24.0-rc1", "0.24.0"),
+            std::cmp::Ordering::Equal
```
serrrfirat: **High Severity -- Pre-release stripped entirely**

`compare_semver` strips pre-release suffixes, treating `0.24.0-rc1` as equal to `0.24.0`. Per semver spec, `0.24.0-rc1 < 0.24.0`. Upgrade from rc1 to release is classified as `Unchanged`.

**Suggested fix:** Use the `semver` crate for proper parsing. At minimum, fall back to string comparison when tuples are equal but raw strings differ.

src/version.rs
```diff
@@ -0,0 +1,233 @@
...
+            .take_while(|c| c.is_ascii_digit())
+            .collect::<String>()
+            .parse()
+            .unwrap_or(0);
+        Some((major, minor, patch))
+    };
+
+    match (parse(a), parse(b)) {
+        (Some(av), Some(bv)) => av.cmp(&bv),
+        _ => a.cmp(b),
```
serrrfirat: **High Severity -- Lexicographic fallback produces wrong ordering**

When semver parsing fails, `a.cmp(b)` is used. `"9.0.0".cmp("10.0.0")` returns `Greater` (because `'9' > '1'`), classifying a legitimate upgrade as a downgrade.

**Suggested fix:** Return `Ordering::Equal` with a warning for unparseable versions. Do NOT use lexicographic order for version strings.

serrrfirat (commented):
Review of version tracking and downgrade detection. Eight findings ranging from Critical to Medium severity.

src/version.rs
```diff
@@ -0,0 +1,233 @@
...
+        let major = parts.next()?.parse().ok()?;
+        let minor = parts.next()?.parse().ok()?;
+        // Patch might contain pre-release suffixes — take only digits.
+        let patch_str = parts.next().unwrap_or("0");
+        let patch: u64 = patch_str
+            .chars()
+            .take_while(|c| c.is_ascii_digit())
+            .collect::<String>()
+            .parse()
+            .unwrap_or(0);
```
serrrfirat: **High Severity -- Silent unwrap_or(0) on patch parse failure**

Patch like `"abc"` becomes `0`. `"0.24"` becomes `(0, 24, 0)`. Combined with stripped pre-release, version comparison is unreliable for non-standard version strings.

**Suggested fix:** Use the `semver` crate, or log a warning when patch parsing falls back.

src/version.rs
```diff
@@ -0,0 +1,233 @@
...
+//! user (`"system"`) with key `"ironclaw.version"`. This is deliberately
+//! separate from per-user settings.
+
+use crate::db::SettingsStore;
+
+/// Well-known settings key for the last-recorded binary version.
+const VERSION_SETTING_KEY: &str = "ironclaw.version";
+
+/// System-level user ID used for global (non-user) settings.
+const SYSTEM_USER: &str = "system";
```
serrrfirat, line 18: **Medium Severity -- "system" user_id is unenforced convention**

Version persisted with `SYSTEM_USER = "system"`. No enforcement preventing a real user with that ID from overwriting the version setting via the settings API.

**Suggested fix:** Add reserved-name check in user creation blocking `user_id = "system"`, or use separate table/namespace.

src/version.rs
```diff
@@ -0,0 +1,233 @@
...
+    current: &str,
+) -> VersionTransition {
+    match check_and_record_version_with(db, current).await {
+        Ok(VersionTransition::Fresh) => {
+            tracing::debug!("First startup — recorded version {}", current);
+            VersionTransition::Fresh
+        }
+        Ok(VersionTransition::Unchanged) => VersionTransition::Unchanged,
+        Ok(VersionTransition::Upgraded { previous }) => {
+            tracing::warn!(
```
serrrfirat, line 119: **Medium Severity -- warn! on routine upgrade corrupts REPL**

`tracing::warn!` on upgrade path. Per CLAUDE.md: use `debug!` for internal diagnostics. Upgrades are routine events.

**Suggested fix:** Change to `debug!` for upgrade. Keep `error!` for downgrade.

src/app.rs
```diff
@@ -208,6 +223,15 @@ impl AppBuilder {
...
             .await;
 
+        // Version tracking — detect upgrades / accidental downgrades.
+        // Capture the previously-recorded version *before* `run_startup_version_check`
+        // overwrites it so the web gateway can surface it to reconnecting clients.
+        self.previous_recorded_version = crate::version::read_previous_version(db.as_ref())
+            .await
+            .ok()
+            .flatten();
+        let _transition = crate::version::run_startup_version_check(db.as_ref()).await;
```
serrrfirat, line 233: **Medium Severity -- TOCTOU on version read**

Two separate DB reads of same key with no transaction. `read_previous_version` reads the key, then `run_startup_version_check` reads it again and writes. Concurrent startup could read inconsistent state.

**Suggested fix:** Combine into single function that reads once and returns both old version and transition.

crates/ironclaw_gateway/static/app.js
```diff
@@ -751,6 +759,21 @@ function connectSSE(lastEventIdOverride) {
       if (restartBtn) restartBtn.disabled = false;
       if (restartIcon) restartIcon.classList.remove('spinning');
       isRestarting = false;
+
+      // After restart, check if the server version changed
+      var preRestartVersion = currentServerVersion;
```
serrrfirat, line 764: **Medium Severity -- previous_version API field unused by client**

After restart, if page was refreshed, `preRestartVersion` is empty string so comparison is falsy. The `previous_version` field was added to the API specifically for this case but the reconnect handler never reads it.

**Suggested fix:** Fall back to comparing `statusData.previous_version` against `statusData.version` when `preRestartVersion` is empty.

ilblackdragon:
## Review fixes applied — `02ba50ac`

**Fix A — `compare_semver()` now uses the `semver` crate** (`src/version.rs`):
- Replaced hand-rolled parser with `semver::Version::parse()` (already a workspace dependency).
- Pre-release versions now sort correctly per spec (`0.24.0-rc1 < 0.24.0`).
- Unparseable strings treated as equal with a `debug!` log — no dangerous lexicographic fallback.
- Added regression tests for `9.0.0 < 10.0.0` and unparseable inputs.

**Fix B — Client JS wires `previous_version`** (`crates/ironclaw_gateway/static/app.js`):
- Reconnect handler now falls back to `statusData.previous_version` when `preRestartVersion` is empty (page refresh during restart). This was the whole reason the field was surfaced through the API.

**Also fixed:** pre-existing compilation errors in `ws_gateway_integration.rs`, `oauth_greeting_integration.rs`, and `multi_tenant.rs` (missing `previous_version` field in `GatewayState` initializers).

**On review findings #25 and #26:**
- **#25 (`warn!` on upgrade):** Intentional — version transitions with potential data-format changes warrant operator visibility. Downgrades use `error!`; upgrades use `warn!`.
- **#26 (TOCTOU):** No race — `read_previous_version()` is called once, then `run_startup_version_check()` does a single atomic read-compare-write. There is no dual-read window.

**Quality gate:** `cargo fmt` + `cargo clippy --all --all-features` (zero warnings) + `cargo test --lib -- version` (51 passed).

ilblackdragon:
Review findings from @serrrfirat addressed in 02ba50ac:

| Finding | Severity | Status |
|---------|----------|--------|
| Won't compile on staging (missing GatewayState fields) | Critical | Fixed -- added missing fields to test initializers |
| Pre-release stripped entirely | High | Fixed -- uses `semver` crate for proper parsing |
| Lexicographic fallback wrong ordering | High | Fixed -- unparseable versions treated as equal with `debug!` log, no lex fallback |
| Silent `unwrap_or(0)` on patch parse | High | Fixed -- `semver::Version::parse()` handles all components properly |
| "system" user_id unenforced | Medium | Acknowledged -- low risk, could be hardened later |
| `warn!` on routine upgrade | Medium | Intentional per author -- upgrade transitions warrant operator visibility |
| TOCTOU on version read | Medium | Not a real race -- single read followed by atomic read-compare-write |
| `previous_version` unused by client | Medium | Fixed -- reconnect handler falls back to `statusData.previous_version` |

The 3-commit progression is clean: initial feature -> caller-level tests -> semver fix + JS wiring. 51 version tests passing. Ready for @serrrfirat to confirm.

crates/ironclaw_gateway/static/app.js
```diff
@@ -751,6 +759,23 @@ function connectSSE(lastEventIdOverride) {
...
```
serrrfirat, line 764: **Medium Severity**

The restart-version warning is still easy to miss in the exact case this PR is trying to protect.

This check only runs inside `if (isRestarting)`, but `isRestarting` is page-memory state. If the user refreshes while the process is restarting, the new page comes back with `isRestarting = false`, so the client never evaluates `previous_version` and no warning is shown. There's also a race with `fetchGatewayStatus()`: that poll updates `currentServerVersion` independently, so if it hits the restarted server before `onopen`, `preRestartVersion` can already be the new version and the warning gets suppressed even without a refresh.

I'd make the warning depend on durable state instead of the transient flag here — for example, compare `previous_version` vs `version` on the first status fetch after reconnect/page load and gate duplicate toasts with a one-shot local/session marker rather than `isRestarting`.

src/app.rs
```diff
@@ -208,6 +223,15 @@ impl AppBuilder {
             .attach_store(db.clone(), &self.config.owner_id)
             .await;
 
+        // Version tracking — detect upgrades / accidental downgrades.
+        // Capture the previously-recorded version *before* `run_startup_version_check`
+        // overwrites it so the web gateway can surface it to reconnecting clients.
+        self.previous_recorded_version = crate::version::read_previous_version(db.as_ref())
```
serrrfirat, line 229: **Medium Severity**

Downgrade detection currently happens *after* this startup path has already performed DB writes.

`init_database()` calls `bootstrap_ownership()` and `migrate_disk_to_db()` before `read_previous_version()` / `run_startup_version_check()`. In the accidental-downgrade case this means an older binary can still rewrite ownership/settings rows before the operator ever sees the downgrade warning, which weakens the protection against mutable image tags pulling an older schema writer.

Can we move the version check to immediately after the DB connection + migrations succeed, before any post-migration writes, or otherwise fail startup on downgrade so the older binary doesn't touch persisted state first?

tests/version_tracking_integration.rs
```diff
@@ -0,0 +1,212 @@
+//! Caller-level tests for version tracking and downgrade detection (PR #2314).
+//!
+//! These tests drive `crate::version::run_startup_version_check_with` — the
```
serrrfirat, line 3: **Medium Severity**

These tests don't actually cover the production wiring this PR adds.

This suite calls `run_startup_version_check_with()` directly, and the new gateway tests in `tests/ws_gateway_integration.rs` inject `GatewayState.previous_version` by hand. That means the green path here does **not** prove that `AppBuilder::init_database()` captures the old value before overwriting it, or that `main.rs -> with_previous_version(...)` keeps forwarding it into `/api/gateway/status`, or that the browser reconnect/refresh flow emits the warning.

I'd add at least one integration test that boots through `AppBuilder::init_database()` and asserts `previous_recorded_version()`, plus a browser-facing test for the refresh-during-restart/reconnect warning path.

ilblackdragon:
Not the right fix. Closing.

Issue #1852, opened 2026-04-01
should make it easy to use


LeiYanhua:
Hi,  man. 
You are making claws launching very fast and safe. That's wonderful !

But the majority of the users are not IT technicians.  If you want more people to use Ironclaw, one of the most important things is to make it easy to use.  

Now, for users to configure, they have to find out where  .eny and config.toml  are located, what there two files mean, the relationship between those two files and the database located in another place. And the users even have to use Shell to configure. That's confusing for non-technicians. It looks like they are using  Dos-based computers.

serrrfirat:
Hi @LeiYanhua, 

would love to get your ideas around how we can improve. Are you having trouble around running ironclaw locally on your machine? 

What are the steps that caused pain in your journey?

LeiYanhua:
@serrrfirat   well,  with a Ph.D. degree, I am not  able to  add one more provider  or model so far.  That is frustrating

Kampouse:
you are running on the provider or the local instance?

LeiYanhua:
@Kampouse On the local instance, windows 10.

PR #2315, opened 2026-04-11
fix(ux): actionable auth errors and improved CLI help for new users (#1852)

ilblackdragon:
## Summary
Addresses #1852 with targeted UX improvements for non-technical users:

- **`src/llm/error.rs`**: `AuthFailed` now includes provider-specific guidance with env var name, signup URL, and `ironclaw onboard --step provider` hint. Covers all providers (4 new tests)
- **`src/cli/mod.rs`**: Top-level `--help` now has "Getting started" and "Common commands" sections; `onboard`/`config`/`models`/`doctor`/`login` long_about improved
- **`src/cli/models.rs`**: Unknown provider errors list available providers; warns when API key env var not set after switching; model list footer shows how to switch

## Test plan
- [x] All 4037 unit tests pass

### DIFF
--- a/src/cli/models.rs
+++ b/src/cli/models.rs
@@ -222,7 +222,8 @@ fn print_model_list(models: &Option<Vec<String>>, active_model: Option<&String>)
         }
         None => {
             println!(
-                "\n  Could not fetch model list (missing credentials or provider unavailable)."
+                "\n  Could not fetch model list (missing credentials or provider unavailable).\
+                 \n  Tip: Run `ironclaw doctor` to check your configuration."
             );
         }
     }
@@ -364,7 +365,9 @@ fn cmd_set_provider(
                 .chain(registry.all().iter().map(|d| d.id.as_str()))
                 .collect();
             anyhow::anyhow!(
-                "Unknown provider '{}'. Known providers: {}",
+                "Unknown provider '{}'.\n\nAvailable providers: {}\n\n\
+                 Tip: Run `ironclaw models list` to see all providers with descriptions,\n\
+                 or `ironclaw onboard --step provider` for interactive setup.",
                 provider,
                 known.join(", ")
             )
@@ -413,6 +416,21 @@ fn cmd_set_provider(
             .map(|p| p.display().to_string())
             .unwrap_or_else(|| config_toml_path().display().to_string())
     );
+
+    // Remind the user about API key requirements for the newly selected provider
+    if let Some(def) = registry.find(&canonical_id)
+        && def.api_key_required
+        && let Some(ref env_var) = def.api_key_env
+    {
+        let has_key = std::env::var(env_var).is_ok();
+        if !has_key {
+            println!();
+            println!(
+                "Note: {} requires an API key. Set {} or run `ironclaw onboard --step provider` to configure.",
+                canonical_id, env_var
+            );
+        }
+    }
 
     Ok(())
 }
@@ -551,6 +569,9 @@ async fn cmd_list_providers(
     if !verbose {
         println!();
         println!("* = active provider. Use --verbose for details.");
+        println!();
+        println!("To switch provider: ironclaw models set-provider <name>");
+        println!("For guided setup:   ironclaw onboard --step provider");
     }
 
     Ok(())
@@ -619,7 +640,8 @@ async fn cmd_show_provider(
             .chain(registry.all().iter().map(|d| d.id.as_str()))
             .collect();
         anyhow::anyhow!(
-            "Unknown provider '{}'. Known providers: {}",
+            "Unknown provider '{}'.\n\nAvailable providers: {}\n\n\
+             Tip: Run `ironclaw models list` to see all providers with descriptions.",
             id,
             known.join(", ")
         )
--- a/src/cli/models.rs
+++ b/src/cli/models.rs
@@ -422,7 +422,10 @@ fn cmd_set_provider(
         && def.api_key_required
         && let Some(ref env_var) = def.api_key_env
     {
-        let has_key = std::env::var(env_var).is_ok();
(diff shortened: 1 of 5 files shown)


ilblackdragon (commented):
I did not find a blocking issue here.

This is mostly UX text, and the change is scoped. The provider-specific guidance in `AuthFailed` is useful. Minor note: because the error text is now policy-bearing product guidance, it is worth treating it as compatibility-sensitive and keeping snapshot-style coverage around the rendered messages for the important providers, not just the helper function.

ilblackdragon:
Addressed in 282e695f: added snapshot-style assertions over the rendered AuthFailed messages for each important provider (nearai, openai, anthropic, ollama, openai_compatible, tinfoil, bedrock, plus the generic unknown-provider fallback). Tests use `insta::assert_snapshot!` with inline snapshots and render through `LlmError::AuthFailed.to_string()` so the outer `#[error(..)]` format string is also covered. Future edits to the user-facing auth guidance now require a deliberate test update.

src/llm/error.rs
```diff
@@ -44,3 +47,202 @@ pub enum LlmError {
...
+/// Return actionable setup guidance for a provider's authentication failure.
+///
+/// This helps users who see an `AuthFailed` error know exactly what to do
+/// without digging through documentation.
+fn auth_guidance(provider: &str) -> String {
+    let normalized = provider.to_lowercase();
+    let (env_hint, extra) = match normalized.as_str() {
+        "nearai" | "near_ai" | "near" => (
+            "Set NEARAI_API_KEY (from https://cloud.near.ai) or run `ironclaw onboard` to log in",
+            "",
```
serrrfirat, line 60: **Medium Severity -- Hardcoded third-party URLs could become stale or phishing targets**

`auth_guidance()` embeds provider signup URLs (e.g. `https://platform.openai.com/api-keys`, `https://console.anthropic.com/settings/keys`). These are hard-coded, cannot be updated without code changes, and become part of the stable API surface (snapshot-tested below).

**Suggested fix:** Consider loading URLs from the provider registry (which already has this data), so they stay in sync and can be updated in one place.

src/llm/error.rs
```diff
@@ -44,3 +47,202 @@ pub enum LlmError {
...
+            "Set GITHUB_COPILOT_TOKEN or run `ironclaw onboard --step provider` to log in via device code",
+            "",
+        ),
+        _ => (
+            "Check that the required API key environment variable is set for this provider",
+            "",
+        ),
+    };
+    if extra.is_empty() {
+        format!("{env_hint}. Or run `ironclaw onboard --step provider` to configure interactively.")
```
serrrfirat, line 95: **Medium Severity -- Dead code: `extra` variable always empty**

`extra` is set to `""` in every match arm, making `if extra.is_empty()` always true. The entire `extra` mechanism is unused dead code that adds complexity for no benefit.

**Suggested fix:** Remove the `extra` variable entirely and simplify to a single `format!("{env_hint}. Or run ...)` return.

src/cli/models.rs
```diff
@@ -414,6 +417,21 @@ fn cmd_set_provider(
             .unwrap_or_else(|| config_toml_path().display().to_string())
     );
 
+    // Remind the user about API key requirements for the newly selected provider
+    if let Some(def) = registry.find(&canonical_id)
+        && def.api_key_required
+        && let Some(ref env_var) = def.api_key_env
+    {
+        let has_key = std::env::var(env_var).is_ok();
+        if !has_key {
```
serrrfirat, line 429: **Medium Severity -- API key warning checks raw env only, not secrets store**

`has_key` checks `std::env::var(env_var)` which doesn't check the secrets store overlay (managed via `ironclaw secrets`). Users who stored their keys in the secrets store will see false "API key required" warnings.

**Suggested fix:** Use `optional_env()` helper (which checks the secrets overlay) instead of raw `std::env::var`. Or soften the message to say "Set {env_var} (if not already in secrets store)".

serrrfirat (commented):
Review of PR #2315 -- 3 findings (all Medium).

ilblackdragon:
Addressed in a7ba063f:

**optional_env for API key check** — replaced `std::env::var(env_var).is_ok()` with `crate::config::helpers::optional_env(env_var).ok().flatten().is_some()` in `src/cli/models.rs`. This ensures keys stored via `ironclaw secrets` are detected, preventing false "API key required" warnings.

Quality gate: `cargo fmt` clean, `cargo clippy --all --all-features` zero warnings, 193 CLI-related tests pass.

---

Re: other findings from the review:

- **Hardcoded URLs** (docs links, registry URLs): This is a design/architecture suggestion. Out of scope for this fix — would require a config refactor across multiple modules.
- **Dead `extra` variable**: False positive — `extra` IS used in the `format!()` macro on the same line via the named argument syntax (`{extra}`).

ilblackdragon (commented):
All three review findings from serrrfirat have been addressed:

1. **API key warning checks raw env only** -- Fixed in a7ba063f by switching to `optional_env()` which checks the secrets store overlay. Good fix.

2. **Dead `extra` variable** -- This was a false positive as noted. `extra` is used in the `format!()` macro via named argument syntax `{extra}`. The current code structure (always-empty `extra`) is a reasonable scaffold for future provider-specific suffixes, though it could be simplified. Not blocking.

3. **Hardcoded URLs** -- Agreed this is out of scope for a UX text fix. The URLs are stable provider console links and are now snapshot-tested, so staleness would be caught during review.

Snapshot tests via `insta::assert_snapshot!` covering all important providers were added in 282e695f, satisfying the earlier review request for compatibility-sensitive coverage of the rendered auth messages.

All review items resolved. Ready to merge.

serrrfirat (approved):

Issue #1610, opened 2026-03-24
feat: SSO/OIDC user authentication — Google, Okta, and generic OIDC provider support


ilblackdragon:
## Problem / Motivation

User authentication is currently **API-token-only** (bearer tokens in `MultiAuthState`). There is no way for a user to log in via their organization's identity provider. This is a blocker for enterprise adoption:

- **No SSO** — users can't authenticate with their existing Google Workspace, Okta, Azure AD, or other OIDC provider credentials
- **No session management** — every request requires a bearer token; there are no browser sessions or cookies
- **No user auto-provisioning** — when someone authenticates via SSO for the first time, they must already exist in the system (once #1605 lands) or be manually invited
- **Token distribution is manual** — admins must generate API tokens and share them out-of-band

**Important distinction**: the existing OAuth code in `src/cli/oauth_defaults.rs` and the `/oauth/callback` handler are for **extension authentication** (connecting IronClaw to third-party APIs like Google Drive, Notion, Slack). This issue adds OAuth 2.0 / OpenID Connect as a **user login** mechanism — fundamentally different flow, different callback, different token storage.

Depends on **#1605 (DB-backed user management)** for the `users` table. Can be implemented in parallel with #1607 (Workspaces) and #1608 (RBAC).

### Supported Providers

1. **Google Workspace** — OIDC with domain restriction (e.g., only `@mycompany.com`)
2. **Okta** — OIDC with org-specific issuer URL
3. **Generic OIDC** — any provider implementing OpenID Connect Discovery (Azure AD, Auth0, Keycloak, etc.)

All use the same OIDC Authorization Code flow with PKCE. Provider-specific logic is limited to configuration defaults and optional domain/group claims.

### Configuration

```bash
# Enable SSO (multiple providers supported simultaneously)
IRONCLAW_SSO_PROVIDERS='[
  {
    "id": "google",
    "type": "oidc",
    "issuer": "https://accounts.google.com",
    "client_id": "...",
    "client_secret": "...",
    "allowed_domains": ["mycompany.com"],
    "auto_provision": true,
    "default_role": "member"
  },
  {
    "id": "okta",
    "type": "oidc",
    "issuer": "https://mycompany.okta.com",
    "client_id": "...",
    "client_secret": "...",
    "auto_provision": true,
    "default_role": "member",
    "groups_claim": "groups",
    "admin_groups": ["ironclaw-admins"]
  }
]'

# Session config
[...]

PR #2316, opened 2026-04-11
docs: SSO/OIDC implementation plan (#1610)

ilblackdragon:
## Summary
Implementation plan for #1610 (SSO/OIDC user authentication). Research-only PR; no code changes.

Plan covers:
- `SsoConfig` with numbered `SSO_PROVIDER_N_*` env vars (multi-provider)
- 3 new DB tables: `sso_sessions`, `sso_auth_state`, `sso_identity_links` (both backends)
- 5 new `/auth/*` routes (separate from existing `/oauth/*` extension OAuth)
- Cookie-based sessions as 4th path in `auth_middleware`
- Auto-provisioning with email domain allowlisting
- Crates: `openidconnect` + `hmac`
- 20+ files touched, full file-by-file change list
- Testing strategy

Plan committed to `docs/plans/SSO_IMPLEMENTATION_PLAN.md`.

## Test plan
- [ ] Review plan and approve approach before implementation

### DIFF
--- a/docs/plans/SSO_IMPLEMENTATION_PLAN.md
+++ b/docs/plans/SSO_IMPLEMENTATION_PLAN.md
@@ -119,6 +119,13 @@ pub struct SsoProviderConfig {
     pub client_secret: String,
     /// Additional scopes beyond `openid email profile`.
     pub extra_scopes: Vec<String>,
+    /// Whether this provider is trusted to return `email_verified == true`
+    /// correctly on every id_token. MUST default to `false`. See §7.1 —
+    /// providers with `false` are excluded from email-based auto-linking
+    /// and domain-based auto-provisioning; they may only be used via
+    /// explicit invite-based flows or pre-existing (provider_id, sub)
+    /// identity links.
+    pub emits_verified_email_claim: bool,
 }
 ```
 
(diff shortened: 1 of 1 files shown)


ilblackdragon (commented):

I’m not comfortable with the plan in its current form because it has two security holes that would be expensive to unwind later.

First, the auto-provisioning / account-linking flow links by email and provisions by email domain, but the plan never requires a verified email claim. That creates an account-takeover path for providers or configurations that expose an unverified email address.

Second, the login flow accepts a redirect parameter and persists `redirect_after`, but the plan does not constrain it to a same-origin relative path. As written, this introduces an open-redirect sink.

I also want the testing section strengthened around these exact risks: verified-vs-unverified email, redirect sanitization, and full middleware routing through the actual auth stack rather than helper-only tests.

ilblackdragon:
Addressed in f1445dab: plan now explicitly requires `email_verified=true` for all linking/auto-provisioning flows (with fallback rejection for missing/false claims), constrains `redirect_after` to same-origin relative paths with server-side validation at both persist and consume time, and rewrites the testing section so verified-email and redirect-sanitization coverage routes through the real auth middleware rather than helper-only tests.

ilblackdragon (commented):
The two security concerns raised in the initial self-review have been thoroughly addressed in f1445dab:

1. **Email verification gate** -- The plan now mandates `email_verified == true` as a hard requirement for all email-based linking and auto-provisioning flows (Section 7.1). The threat model is well-articulated, with clear rules: missing/false/null/string-false all treated as unverified, hard reject with no fall-through, per-provider `emits_verified_email_claim` config flag, and startup validation warning when auto-provision is enabled with untrusted providers.

2. **Open redirect protection** -- `redirect_after` is now constrained to same-origin relative paths only (single `/` prefix). Dual validation at both persist-time and consume-time, with regex rejection of `//`, protocol-relative, and absolute URLs. No host-based allowlisting escape hatch. Solid design.

3. **Testing strategy** -- Updated to require handler-level integration tests through the real auth middleware stack rather than helper-only unit tests, consistent with the project's "test through the caller" rule.

The plan is comprehensive and security-conscious. Ready for implementation.

ilblackdragon:
Just a doc. Will be implementation separately.

PR #2319, opened 2026-04-11
feat: support public Generative Language API for headless deployments

### LINKED ISSUE #2307: Support public Generative Language API for headless/server deployments
## Problem

The `gemini_oauth` backend routes Gemini 2.0+ models exclusively through Google's Cloud Code API (`cloudcode-pa.googleapis.com/v1internal`). This works for interactive users who authenticate via browser-based OAuth, because their Google accounts have implicit Cloud Code access.

However, it doesn't work for **headless or server deployments** that use `GEMINI_API_KEY` for authentication:

- Cloud Code doesn't accept API key auth
- Even with valid OAuth credentials obtained through a manual flow, non-consumer Google accounts receive HTTP 500 ("Internal error encountered")
- There's no way to opt into the public Generative Language API (`generativelanguage.googleapis.com`) for 2.0+ models

The public Generative Language API supports the same Gemini models, accepts API keys via `x-goog-api-key` header, and is designed for programmatic use — but the version-based routing in `model_uses_cloud_code_api()` makes it unreachable for any model >= 2.0.

## Use case

Running IronClaw as an agentic backend service (no human in the loop, no browser for OAuth), authenticating with a standard Gemini API key from Google AI Studio.

## Proposed solution

An environment variable that forces all models through the public Generative Language API regardless of version. For example:

```
GEMINI_USE_LEGACY_API=true
```

A minimal implementation is a three-line check at the top of `model_uses_cloud_code_api()` in `src/llm/gemini_oauth.rs`:

```rust
pub fn model_uses_cloud_code_api(model: &str) -> bool {
    if std::env::var("GEMINI_USE_LEGACY_API").is_ok() {
        return false;
    }
    // ... existing version-based logic unchanged
}
```

This preserves existing behavior for interactive users while unlocking server deployments.

## Additional consideration: tool schema compatibility

The public Generative Language API uses a stricter subset of JSON Schema for function declarations than Cloud Code. Fields like `additionalProperties` are not recognized and cause HTTP 400 errors. Tool schemas may need to be stripped of unsupported fields before sending when using the public endpoint. (We have a working patch for this as well — happy to include it in a PR.)

## Environment

- IronClaw v0.24.0
- `LLM_BACKEND=gemini_oauth`
- `GEMINI_API_KEY` set (valid, from Google AI Studio)
- Running on a headless Linux VM (GCE)

jackdempsey:
## Summary

- Add `GEMINI_USE_LEGACY_API` env var that forces all Gemini models (including 2.0+) through `generativelanguage.googleapis.com` instead of the Cloud Code API
- Strip `additionalProperties` and `$schema` from tool parameter schemas before sending, as the public API uses a stricter JSON Schema subset

## Motivation

The `gemini_oauth` backend routes Gemini 2.0+ models to `cloudcode-pa.googleapis.com`, which works for interactive users with browser-based OAuth but fails for headless/server deployments:

- Cloud Code doesn't accept API key auth
- OAuth credentials from non-consumer accounts get HTTP 500
- No way to opt into the public Generative Language API for newer models

The public API supports the same models, accepts `GEMINI_API_KEY` via `x-goog-api-key` header, and is designed for programmatic use.

## Changes

**`src/llm/gemini_oauth.rs`**:
- `model_uses_cloud_code_api()`: return `false` early when `GEMINI_USE_LEGACY_API` is set
- `strip_unsupported_schema_fields()`: recursively remove `additionalProperties` and `$schema` from tool parameter JSON
- `to_gemini_request()`: apply schema stripping to tool declarations

## Usage

```bash
export GEMINI_USE_LEGACY_API=true
export GEMINI_API_KEY=your-key
export GEMINI_MODEL=gemini-3.1-pro-preview
```

No behavior change when `GEMINI_USE_LEGACY_API` is unset.

Fixes #2307

### DIFF
--- a/src/llm/gemini_oauth.rs
+++ b/src/llm/gemini_oauth.rs
@@ -1137,6 +1137,9 @@ impl GeminiOauthProvider {
     }
 
     pub fn model_uses_cloud_code_api(model: &str) -> bool {
+        if std::env::var("GEMINI_USE_LEGACY_API").is_ok() {
+            return false;
+        }
         let model = model.to_ascii_lowercase();
         // Models containing "-preview" suffix or "gemini-3" use the Cloud Code API.
         // Using "-preview" (with hyphen) to avoid false positives on unrelated model names.
@@ -1589,6 +1592,21 @@ impl GeminiOauthProvider {
         }
     }
 
+    /// Strip JSON Schema fields unsupported by the public Generative Language API.
+    fn strip_unsupported_schema_fields(val: &mut serde_json::Value) {
+        if let Some(obj) = val.as_object_mut() {
+            obj.remove("additionalProperties");
+            obj.remove("$schema");
+            for (_, v) in obj.iter_mut() {
+                Self::strip_unsupported_schema_fields(v);
+            }
+        } else if let Some(arr) = val.as_array_mut() {
+            for v in arr.iter_mut() {
+                Self::strip_unsupported_schema_fields(v);
+            }
+        }
+    }
+
     fn to_gemini_request(
         messages: &[ChatMessage],
         tools: Option<&[ToolDefinition]>,
@@ -1709,10 +1727,12 @@ impl GeminiOauthProvider {
             let declarations: Vec<serde_json::Value> = tool_defs
                 .iter()
                 .map(|t| {
+                    let mut params = t.parameters.clone();
+                    Self::strip_unsupported_schema_fields(&mut params);
                     serde_json::json!({
                         "name": t.name,
                         "description": t.description,
-                        "parameters": t.parameters
+                        "parameters": params
                     })
                 })
                 .collect();


serrrfirat:
Thank you for addressing headless Gemini access through the public Generative Language API. We are closing this PR because the current Reborn provider architecture now covers this use case differently.

Reborn’s API-key Gemini provider uses the native Generative Language API through `create_gemini_from_registry` in `crates/ironclaw_llm/src/lib.rs`, including provider-native tool-call handling. The separate `gemini_oauth` backend also accepts `GEMINI_API_KEY` on its public-API path. As a result, `GEMINI_USE_LEGACY_API` in the retired `src/llm` implementation is no longer the right configuration boundary.

The headless-deployment requirement is important, and we would be happy to have you contribute Reborn-side Gemini compatibility or integration tests—particularly around API-key authentication and strict tool schemas. Thank you for identifying and solving this gap in the earlier architecture.

PR #2322, opened 2026-04-11
fix(engine): always append ActionResult for every tool call

ilblackdragon:
## Summary

Fixes HTTP 400 `No tool output found for function call <id>` from OpenAI's Responses API when a tool call in the assistant message has no corresponding tool result.

The Python orchestrator only appended `ActionResult` messages when `output is not None`, leaving dangling `tool_calls` in `working_messages` in three scenarios:

1. **Tool returns null output** - JSON `null` -> Python `None`, skipping the append
2. **Gate-paused results** - `result_json` has no `output` field at all
3. **RequireApproval preflight early return** - Rust returned fewer results than calls, so trailing calls got no slot

OpenAI's Responses API requires 1:1 correspondence between `function_call` and `function_call_output` items, so the next LLM turn fails with the 400.

## Changes

- **`crates/ironclaw_engine/orchestrator/default.py`** - iterate over `executable_calls` (not `results`) and emit an `ActionResult` for every call, with `"[no output]"` / `"[execution skipped]"` placeholders when the real output is missing.
- **`crates/ironclaw_engine/src/executor/orchestrator.rs`** - pad the `RequireApproval` early-return `results_json` to `parsed.len()` so Python sees a null slot for every call and emits the placeholder.
- Regression tests at both layers:
  - `json_to_thread_messages_every_tool_call_has_action_result` (engine) - parse path
  - `tool_call_result_correspondence_after_sanitize` (bridge) - full `ThreadMessage` -> `ChatMessage` -> `sanitize_tool_messages` pipeline

## Test plan

- [x] `cargo test -p ironclaw_engine` (350 pass, including new regression test)
- [x] `cargo test --lib -- bridge::llm_adapter` (new regression test passes)
- [x] `cargo clippy --lib --bins -- -D warnings` clean
- [ ] Rerun the failing `update all tools` flow against the OpenAI Codex provider and confirm no more HTTP 400

### DIFF
--- a/crates/ironclaw_engine/orchestrator/default.py
+++ b/crates/ironclaw_engine/orchestrator/default.py
@@ -720,22 +720,29 @@ def run_loop(context, goal, actions, state, config):
             # Rust handles preflight (lease/policy), parallel execution via
             # JoinSet, and event emission in call order.
             results = __execute_actions_parallel__(executable_calls)
-            for idx in range(len(results)):
-                r = results[idx]
-                if r is None:
-                    continue
-                call = executable_calls[idx] if idx < len(executable_calls) else {}
+            # Every tool call in the assistant message MUST have a matching
+            # ActionResult, otherwise the LLM API rejects the sequence with
+            # "No tool output found for function call <id>". Iterate over
+            # executable_calls (not results) so we cover calls that the Rust
+            # batch handler skipped (e.g. RequireApproval early return).
+            for idx in range(len(executable_calls)):
+                call = executable_calls[idx]
                 call_id = call.get("call_id", "")
-                action_name = r.get("action_name", call.get("name", ""))
-                output = r.get("output")
-                if output is not None:
-                    append_message(
-                        working_messages,
-                        "ActionResult",
-                        str(output),
-                        action_name=action_name,
-                        action_call_id=call_id,
-                    )
+                r = results[idx] if idx < len(results) else None
+                if r is not None:
+                    action_name = r.get("action_name", call.get("name", ""))
+                    output = r.get("output")
+                    output_str = str(output) if output is not None else "[no output]"
+                else:
+                    action_name = call.get("name", "unknown")
+                    output_str = "[execution skipped]"
+                append_message(
+                    working_messages,
+                    "ActionResult",
+                    output_str,
+                    action_name=action_name,
+                    action_call_id=call_id,
+                )
 
             # Check results for auth/approval interrupts
             for r_idx, r in enumerate(results):
--- a/crates/ironclaw_engine/src/executor/orchestrator.rs
+++ b/crates/ironclaw_engine/src/executor/orchestrator.rs
@@ -1317,6 +1317,12 @@ async fn handle_execute_actions_parallel(
                         })
                         .unwrap_or_default(),
                     }));
+                    // Pad with nulls for calls that weren't reached so the
+                    // Python-side loop can emit ActionResult placeholders for
+                    // every tool call in the assistant message.
+                    while results_json.len() < parsed.len() {
+                        results_json.push(serde_json::json!(null));
+                    }
                     return ExtFunctionResult::Return(json_to_monty(&serde_json::json!(
                         results_json
                     )));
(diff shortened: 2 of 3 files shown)


crates/ironclaw_engine/src/executor/orchestrator.rs
```diff
@@ -3328,4 +3334,74 @@ mod tests {
...
+
+        let parsed = json_to_thread_messages(&messages).expect("must parse");
+        assert_eq!(parsed.len(), 6);
+
+        // Extract call IDs from the assistant message
+        let assistant_calls: std::collections::HashSet<String> = parsed
+            .iter()
+            .filter_map(|m| m.action_calls.as_ref())
+            .flat_map(|calls| calls.iter().map(|c| c.id.clone()))
+            .collect();
```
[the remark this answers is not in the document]
ilblackdragon: Good catch — addressed in the follow-up commit by adding an explicit `assert_eq!(assistant_calls.len(), 3, ...)` right after the HashSet collection. The test will now fail loudly if schema drift causes `action_calls` to silently deserialize into `None`, instead of vacuously passing the correspondence loop.

Also extended coverage with a new fixture test `json_to_thread_messages_handles_empty_and_malformed_action_results` that drives the parser with each placeholder shape Python emits (`[no output]`, `[execution skipped]`, empty string, and a missing `content` field), plus a caller-level companion in `src/bridge/llm_adapter.rs` (`complete_with_tools_preserves_empty_and_placeholder_action_results`) that asserts the provider ultimately receives a `Tool` message for every tool_call even when the content is empty or a placeholder.

crates/ironclaw_engine/orchestrator/default.py
```diff
@@ -720,22 +720,29 @@ def run_loop(context, goal, actions, state, config):
...
-            for idx in range(len(results)):
-                r = results[idx]
-                if r is None:
-                    continue
-                call = executable_calls[idx] if idx < len(executable_calls) else {}
+            # Every tool call in the assistant message MUST have a matching
+            # ActionResult, otherwise the LLM API rejects the sequence with
+            # "No tool output found for function call <id>". Iterate over
+            # executable_calls (not results) so we cover calls that the Rust
+            # batch handler skipped (e.g. RequireApproval early return).
```
serrrfirat, line 727: **Medium Severity — Edge case documentation**

The bounds check `results[idx] if idx < len(results) else None` is correct, but the comment (line 724-727) only mentions one scenario: "calls that the Rust batch handler skipped (e.g. RequireApproval early return)".

There's a second scenario worth documenting: if a `JoinSet` task panics in the parallel execution path (`orchestrator.rs:1483-1485`), the slot gets a fallback `{"is_error": true, "output": {"error": "execution slot empty"}}` — so the result will be non-null with a synthetic error. The Python code handles this correctly (it emits `str({...})` as output), but future maintainers may not realize this path exists.

**Suggestion:** Widen the comment to say something like "calls that the Rust batch handler skipped or couldn't reach (RequireApproval early return, task panic recovery, etc.)"

crates/ironclaw_engine/src/executor/orchestrator.rs
```diff
@@ -3328,4 +3334,74 @@ mod tests {
             .expect("empty array should produce Some(vec![])");
         assert!(calls.is_empty());
     }
+
```
serrrfirat, line 3337: **Medium Severity — Missing direct test of Rust padding logic**

This test validates that `json_to_thread_messages` correctly parses working_messages where every tool call has an ActionResult — but it doesn't exercise the actual Rust padding code (`while results_json.len() < parsed.len()` at line 1323).

The padding loop is the critical Rust-side fix. If someone removes it, this test still passes because it only tests the `json_to_thread_messages` parse function, not `handle_execute_actions_parallel`. The Python-side defense (`idx < len(results)`) would catch the gap at runtime, so this isn't a ship-blocker — but a direct unit test of the padding would strengthen the safety net.

**Suggestion:** Add a test that invokes `handle_execute_actions_parallel` with a mock `PolicyEngine` that returns `RequireApproval` on the 2nd of 3 calls, and asserts the returned JSON array has length 3 (matching input), not 2.

serrrfirat (approved):
Solid bug fix with good defense-in-depth across both Python and Rust layers. The dual-layer protection (Rust pads results + Python handles short lists) is well-engineered. Two medium-severity suggestions posted inline — neither is a ship-blocker. LGTM.

PR #2328, opened 2026-04-11
fix: prevent immediate requests creating missions

### LINKED ISSUE #2286: [QA] Bot creates scheduled cron job instead of executing task immediately as requested
## Environment
hosted-staging (crab shack)

## Version / Commit Hash
N/A (hosted-staging)

## QA Test Date
2026-04-10

## Feature Area
Agent → Task execution mode (immediate vs. scheduled)

## Steps to Reproduce
1. Create a Product feedback summarizer bot on hosted-staging
2. Ask the bot to summarize product feedback NOW (use explicit immediate language like "do it now", "right now", "immediately")
3. Observe whether the bot executes immediately or creates a scheduled task

## Expected Behavior
When the user asks for immediate execution, the bot should run the task right away and return results in the current session. The bot should distinguish between "do this now" and "do this on a schedule."

## Actual Behavior
The bot created a scheduled cron mission instead of executing the task immediately as requested. When told explicitly "do it now," the bot had already committed to the cron approach and then hit a 413 Payload Too Large error on retry.

Tester confirmed: "I wanted to have it now, he created cron."

The agent defaults to creating scheduled/recurring tasks even when the user's intent is clearly for one-time immediate execution. This suggests the agent's task planning logic over-indexes on scheduling capabilities.

## Logs / Screenshots
Observed during Item 12 (Product feedback summarizer) testing by Artem on hosted-staging.

## Checklist
- [x] Steps are reproducible

serrrfirat:
## Summary

Fixes #2286 by preventing immediate one-shot foreground requests from being converted into mission/routine creation.

## Root Cause

Immediate chat requests had mission and routine creation affordances available in the foreground thread without a deterministic intent gate. The model could choose `routine_create` or `mission_create` for a request like "summarize product feedback right now", and the v2 bridge would translate routine creation into a scheduled mission.

## Changes

- Carry the original thread goal through `ThreadExecutionContext`.
- Reject foreground `mission_create` and `routine_create` alias calls when the original goal is immediate-only and lacks explicit scheduling/recurrence intent.
- Tighten mission/routine tool guidance and routine-advisor activation to avoid steering one-shot summary requests into routines.
- Add regression coverage for immediate one-shot requests, explicit scheduled routines that also say "run it now", and a quantifier edge case like "summarize every item right now".

## Drive-by changes (called out for transparency)

Minor drive-bys the earlier review flagged — the first two resolved themselves when staging landed its refactors; the last two remain because the tests break without them.

- `config/mod.rs`: extracted `empty_toml_path()` helper to isolate `re_resolve_llm` tests from the host's real `config.toml`. Staging independently arrived at the same fix inline in one of the tests — kept the helper form since it's DRY across 4 tests.
- `pairing/approval.rs`: `#[ignore]` on `propagate_approval` test (requires prebuilt Telegram WASM module).
- `wasm/setup.rs`: the `#[allow(clippy::too_many_arguments)]` flagged in the earlier review became moot — staging refactored `register_startup_channels` to take a `StartupChannelRegistrationContext` struct, so the attribute was dropped during the merge.
- `web/server.rs`: the `tool_install` special-case in `pending_gate_extension_name` also became moot — staging deleted `server.rs` as part of the web module restructure.

## Review follow-ups addressed in the latest commit

- Dropped the unused `_params` argument from `should_reject_immediate_mission_create`.
- Hoisted `SCHEDULE_STEMS`, `IMMEDIATE_WORDS`, and `IMMEDIATE_PHRASES` to module-scope consts.
- Compute lowered goal + `word_set` once in `should_reject_immediate_mission_create` and pass refs into `contains_immediate_execution_marker` / `contains_scheduling_intent` instead of re-lowering twice.
- Separated the "do not use for immediate one-shot requests" guidance in the `codeact_preamble.md` `mission_create` description onto its own line for readability.

## Validation

- `cargo test -p ironclaw foreground_ --lib` — 5/5 pass
- `cargo test -p ironclaw caller_level_mission --lib` — 3/3 pass
- `cargo test -p ironclaw background_mission_threads_can_create_follow_up_missions --lib` — 1/1 pass
- `cargo check -p ironclaw --lib --tests`
- `cargo clippy -p ironclaw --lib --tests -- -D warnings`
- `cargo fmt --check`


henrypark133 (dismissed):
## Review: Prevent immediate requests creating missions (Risk: Medium)

Good heuristic guard. The dual-check approach (immediate markers AND NOT scheduling intent) is the right design — it prevents false rejections when users say "run the daily routine now."

Positives:
- `should_reject_immediate_mission_create` only fires for Foreground threads — background/mission threads can still create follow-up missions
- Multi-word phrases ("right now", "do it now") correctly use `lower.contains()` substring matching
- Single words ("now", "asap") use word-set boundary matching — "know" won't false-match "now"
- Error message is explicit and actionable for the LLM
- `thread_goal` propagated in all 4 orchestrator sites + router.rs gate resume path
- Prompt template and tool descriptions updated consistently
- 5 unit tests cover key scenarios

### Minor: "monitoring" won't match "monitor" stem [Logic]
**File:** `src/bridge/effect_adapter.rs:1136`
`word_set` splits on non-alphanumeric, so "monitoring" is a distinct token from "monitor". Goal "set up monitoring now" would be incorrectly rejected. Similarly, "routinely" wouldn't match "routine".
**Suggested fix:** Add "monitoring" to SCHEDULE_WORDS, or use prefix matching for stems.

Convention notes:
- Removing "summary" from routine-advisor keywords is appropriate — "daily summary" still matches via "daily"
- `routine_create` tool description also updated with the "don't use for immediate" instruction

serrrfirat:
@henrypark133 fixed the comments can you take another look?

henrypark133 (dismissed):
## Review: Prevent immediate requests creating missions

The direction is right, but this needs caller-level regression coverage before merge.

### Concerning: helper-only coverage for a side-effect gate
**File:** `src/bridge/effect_adapter.rs:1095`
The new `should_reject_immediate_mission_create(...)` predicate now decides whether `EffectBridgeAdapter::execute_action()` reaches `create_mission()` at `src/bridge/effect_adapter.rs:214`, but the added tests only exercise the helper directly (`src/bridge/effect_adapter.rs:2405` onward). That misses exactly the failure mode called out in `.claude/rules/testing.md`: a wrapper can silently lose one of the computed inputs (`thread_goal`, `thread_type`, alias-normalized params) while the helper unit tests still pass.

**Suggested fix:** add at least one caller-level test that drives `adapter.execute_action("mission_create", ...)` (or the `routine_create` alias path) and asserts the mission manager is not reached for an immediate foreground request, plus one allow-path test for an explicit scheduled request.

serrrfirat:
## Addressed review feedback

**@henrypark133** — added the caller-level regression coverage you requested:

### New tests (`src/bridge/effect_adapter.rs`)
Three tests that drive `EffectBridgeAdapter::execute_action()` end-to-end with a wired-up `MissionManager`:

1. **Reject path**: `execute_action_rejects_mission_create_for_immediate_foreground` — foreground + immediate goal → `EngineError::Effect`
2. **Allow path**: `execute_action_allows_mission_create_with_scheduling_intent` — foreground + scheduling intent ("daily routine") → mission created successfully
3. **Alias path**: `execute_action_rejects_routine_create_alias_for_immediate_foreground` — `routine_create` → `mission_create` alias also correctly rejected

All three pass locally and in CI. Also merged staging and fixed Rust 1.95 clippy lints that appeared across the codebase.

ilblackdragon:
# Code Review

## Overview

Fixes #2286 where immediate requests like "summarize product feedback right now" were being converted into scheduled missions/routines. Three layers of defense:

1. **Data plumbing**: `thread_goal` added to `ThreadExecutionContext` so adapters can see the original user ask (`crates/ironclaw_engine/src/traits/effect.rs`).
2. **Runtime gate**: `should_reject_immediate_mission_create` in `src/bridge/effect_adapter.rs` rejects `mission_create` / `routine_create` on foreground threads when the goal contains immediacy markers and no scheduling intent.
3. **Prompt guidance**: codeact preamble, `mission_create` action description, `RoutineCreateTool` description, and routine-advisor SKILL all updated.

## Strengths

- **Belt-and-suspenders**: runtime gate + prompt changes means the fix doesn't rely on the model alone.
- **Caller-level tests present** (`caller_level_mission` module) — matches the `.claude/rules/testing.md` "Test Through the Caller" rule the project requires.
- **Mission thread exemption** is tested (`background_mission_threads_can_create_follow_up_missions`) — good, since mission-spawned threads often have immediacy phrasing in their goal but legitimately create follow-ups.
- **Stem-based scheduling detection** (`"monitor"` → monitoring/monitors, `"schedul"` → scheduled/scheduling) is more robust than exact-word matching.

## Issues

### 1. Unrelated drive-by changes — should be split
Several files have no connection to the issue being fixed:
- `src/channels/web/server.rs` (+9) — special-cases `tool_install` / `tool-install` / `tool_activate` / `tool_auth` in `pending_gate_extension_name`.
- `src/config/mod.rs` (+17/-4) — test hygiene fix (use temp toml instead of `None` to avoid loading real user config). Valuable but orthogonal.
- `src/pairing/approval.rs` (+1) — `#[ignore]` a test requiring prebuilt WASM.
- `src/channels/wasm/setup.rs` (+1) — `#[allow(clippy::too_many_arguments)]`.

These should ideally land as separate PRs. At minimum the PR description should mention them.

### 2. `_params` argument is dead weight
`should_reject_immediate_mission_create(context, _params)` never reads `_params`. Drop it or justify with a comment. Per `CLAUDE.md`: "Don't add features...beyond what the task requires."

### 3. False-positive risk on parent thread goal text
The gate inspects the **parent thread's goal**, not the `mission_create` params. That's correct for the reported bug, but goals like "I want this task right now, and also want it to keep happening" — "happening" has no scheduling stem, "right now" rejects. Users who genuinely want scheduled execution but phrase it awkwardly get bounced. The error message does direct the model to retry correctly, so the impact is soft, but worth noting.

### 4. Hardcoded stems inside function body
`SCHEDULE_STEMS` is defined inside `contains_scheduling_intent`. Move to module-level `const` for discoverability — especially since the immediate-marker list is also inline and the two should evolve together.

### 5. Double lowercasing per call
`contains_immediate_execution_marker(goal)` and `contains_scheduling_intent(goal)` each call `to_ascii_lowercase()` and `word_set()` on the same input. Minor perf, but trivial to fix: compute lowered+word_set once in `should_reject_immediate_mission_create` and pass refs down. Not load-bearing given call frequency.

## Nits

- The ~150-line `StubStore` / `StubLlm` / `StubEffects` scaffolding in `caller_level_mission` is large. Consider whether existing engine test fixtures could be reused.
- `codeact_preamble.md` line is now quite long; single-paragraph is fine, but a newline before "Do not use it for immediate one-shot requests..." would read better.

## Test Coverage

Adequate for the fix. Could optionally add:
- Goal with `thread_goal: None` → not rejected (currently implicit in the `let Some(goal) = ... else` branch but untested).

## Security

No user-controlled data crosses trust boundaries here. Failure mode is a rejected tool call with an error message returned to the model — no lost data, no privilege change.

## Verdict

**Approve with changes.** The core fix is correct and well-tested. Requests:
1. Split out the unrelated changes or call them out explicitly in the PR body.
2. Drop `_params` from the helper signature.
3. Hoist `SCHEDULE_STEMS` (and consider a similar const for immediate markers) to module scope.

ilblackdragon:
Addressed the review feedback and merged `origin/staging`:

**Must-fixes**
- Dropped unused `_params` from `should_reject_immediate_mission_create`.
- Hoisted `SCHEDULE_STEMS` + `IMMEDIATE_WORDS`/`IMMEDIATE_PHRASES` to module-scope `const`s.
- PR body now calls out the drive-by changes explicitly.

**Nice-to-haves**
- Collapsed the double `to_ascii_lowercase()` + `word_set()` calls (compute once in `should_reject_immediate_mission_create`, pass refs down).
- Added a line break in `codeact_preamble.md` before the "Do not use it for immediate one-shot requests" guidance.

**Drive-by cleanup**
- `wasm/setup.rs` `#[allow(too_many_arguments)]` — became moot, staging refactored to a `StartupChannelRegistrationContext` struct.
- `web/server.rs` — moot, staging deleted the file during the web module restructure.
- `config/mod.rs` and `pairing/approval.rs` remain (tests break without them); called out in the PR body.

All 9 mission-gate tests pass (5 helper + 3 caller-level + 1 background-exemption). `cargo fmt --check` / `clippy -D warnings` clean.

ilblackdragon (approved):

PR #2330, opened 2026-04-11
[codex] Fix web chat refresh active thread

### LINKED ISSUE #2285: [QA] Web UI: refresh without thread hash restores assistant thread instead of active non-assistant thread
## Status

Fixed by #2330 (merged to staging 2026-04-12, commit ed2d6dc). Pending QA re-verification on hosted-staging post-promotion.

## Environment

hosted-staging (crab shack)

## Version / Commit Hash

Reproduced on staging prior to ed2d6dc.

## QA Test Date

2026-04-10 (reported), 2026-04-12 (root-caused and fixed)

## Feature Area

Web gateway, `crates/ironclaw_gateway/static/app.js`, thread restoration on initial load.

## Steps to Reproduce

1. Open the web UI on a fresh tab with no thread hash in the URL (navigate to the bare gateway URL).
2. Open or create a non-assistant thread and send a message that takes a few seconds to process.
3. Refresh the browser before the response arrives.
4. Observe which thread the UI restores.

## Expected Behavior

The UI restores the server's reported `active_thread`, so the in-flight conversation stays visible. Backend processing continues as before.

## Actual Behavior

The UI falls back to the pinned assistant thread and hides the active non-assistant thread. The backend continues processing correctly and delivers the response on the originating channel (Telegram, in the original report), which made it look as though the bot was responding to nothing. The user has to manually navigate back to the active thread to see context.

Tester quote: "Screen refreshed, and my message gone, but IronClaw still messaging me about my request."

## Root Cause

`loadThreads()` ignored the `active_thread` field returned by `/api/chat/threads` whenever `currentThreadId` was empty, which is the cold-load case without a URL hash. The pinned assistant thread won by default. Backend state was correct throughout. The bug was purely client-side restoration priority.

## Fix Summary (PR #2330)

Restoration priority is now: URL hash, then server `active_thread`, then assistant fallback. E2E regression added at `tests/e2e/scenarios/test_sse_reconnect.py::test_refresh_without_hash_reopens_active_thread_history`.

## Known Residual

Per reviewer note on #2330, the new early-return paths skip the read-only channel detection block. If the restored active thread is a read-only channel (e.g., Telegram), the chat input may not be disabled on the very first render. It self-corrects on the next `loadThreads()` cycle. Pre-existing condition that the PR widens slightly. Worth a follow-up issue if anyone hits it in practice.

## QA Verification Checklist

- [ ] Reproduce steps 1–4 on hosted-staging after promotion. Active thread restored.
- [ ] URL hash still wins over `active_thread` (regression check).
- [ ] Assistant fallback still works when no active thread is reported.
- [ ] Read-only thread input state on first load (see residual above).

## Original Observation

Item 22 testing by Artem on hosted-staging, 2026-04-10.

serrrfirat:
## Summary
- Reopen the server-reported active chat thread on first web UI load instead of always falling back to the assistant thread.
- Keep assistant fallback behavior when there is no active thread.
- Add an E2E regression for refresh without a thread hash restoring the active non-assistant thread history.

## Root Cause
On page refresh, the frontend loaded `/api/chat/threads` but ignored the `active_thread` field when `currentThreadId` was empty. It selected the pinned assistant thread by default, so an in-flight non-assistant thread disappeared from the browser while the backend continued processing it.

Fixes #2285.

## Validation
- `node --check crates/ironclaw_gateway/static/app.js`
- `CARGO_TARGET_DIR=/Volumes/NVME/ironclaw-upstream/target /Volumes/NVME/ironclaw-upstream/tests/e2e/.venv/bin/pytest scenarios/test_sse_reconnect.py::test_refresh_without_hash_reopens_active_thread_history -q --timeout=300`

### DIFF
--- a/crates/ironclaw_gateway/static/app.js
+++ b/crates/ironclaw_gateway/static/app.js
@@ -3220,14 +3220,30 @@ function loadThreads() {
       }
     }
 
-    // Default to assistant thread on first load if no thread selected
-    if (!currentThreadId && assistantThreadId) {
-      switchToAssistant();
+    // Reopen the server's active thread on first load. This keeps the visible
+    // chat attached to an in-flight agent turn after a browser refresh, even
+    // when the URL does not carry an explicit thread hash.
+    if (!currentThreadId) {
+      const activeThreadId = data.active_thread || null;
+      if (activeThreadId && activeThreadId === assistantThreadId) {
+        switchToAssistant();
+        return;
+      }
+      if (activeThreadId && threads.some(t => t.id === activeThreadId)) {
+        switchThread(activeThreadId);
+        return;
+      }
+      if (assistantThreadId) {
+        switchToAssistant();
+        return;
+      }
     }
 
     // Enable/disable chat input based on channel type
     if (currentThreadId) {
-      const currentThread = threads.find(t => t.id === currentThreadId);
+      const currentThread = currentThreadId === assistantThreadId
+        ? data.assistant_thread
+        : threads.find(t => t.id === currentThreadId);
       const ch = currentThread ? currentThread.channel : 'gateway';
       currentThreadIsReadOnly = isReadOnlyChannel(ch);
       if (currentThreadIsReadOnly) {
--- a/tests/e2e/scenarios/test_sse_reconnect.py
+++ b/tests/e2e/scenarios/test_sse_reconnect.py
@@ -95,6 +95,37 @@ async def test_sse_reconnect_preserves_chat_history(page):
     assert user_msgs >= 1, "User message should be preserved after reconnect"
 
 
+async def test_refresh_without_hash_reopens_active_thread_history(page):
+    """Refreshing should reopen the server active thread when the URL has no thread hash."""
+    await page.locator("#thread-new-btn").click()
+    await page.wait_for_function(
+        "() => !!currentThreadId && currentThreadId !== assistantThreadId"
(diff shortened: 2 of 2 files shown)


ilblackdragon (commented):
## Code Review

### Overview
Fixes #2285 where browser refresh always fell back to the assistant thread instead of restoring the server's active thread. `loadThreads()` now consults `data.active_thread` before falling back. Includes an E2E regression test.

### Analysis
The fix is correct. Priority ordering: URL hash > server `active_thread` > assistant fallback. The `_pendingThreadRestore` path still runs first, so explicit URL hashes win.

### Minor Issue
The three new `return;` statements skip the read-only channel detection block below. If the active thread is a read-only channel (e.g., Telegram), chat input won't be disabled on first load. It self-corrects after the next `loadThreads()` cycle, so it's a **transient** bug. Pre-existing (the `_pendingThreadRestore` path had the same issue), but this PR widens it to more code paths. Consider whether `switchThread()` / `switchToAssistant()` should internally handle the read-only state, or remove the early returns and fall through to the existing read-only check.

### E2E Test
Well-structured: creates thread, sends message, strips URL hash, reloads, verifies active thread restoration and message history visibility.

### Verdict: **Approve** — Clean, focused fix with good regression test. The read-only transient issue is pre-existing and minor.

ilblackdragon (approved):

PR #2336, opened 2026-04-11
feat(memory): configurable insights interval, session summary hook, reasoning-augmented recall

serrrfirat:
## Summary

Three memory enrichment features inspired by Honcho's reasoning layer, built natively using existing infrastructure:

- **Configurable conversation insights interval** — `MISSION_INSIGHTS_INTERVAL` env var (default: 5, min: 1) replaces the hardcoded `CONVERSATION_INSIGHTS_INTERVAL` constant. New `MissionsConfig` + `MissionSettings` follow existing config patterns. Wired via builder method on `MissionManager`.
- **Session-end summary hook** — `SessionSummaryHook` writes LLM-generated conversation summaries to `daily/{date}-session-summary.md` on session prune. Uses cheap LLM, fail-open, 30s timeout, skips sessions with < 3 messages.
- **Reasoning-augmented recall** — Optional `reasoning` boolean parameter on `memory_search`. When enabled + `SEARCH_REASONING_ENABLED=true`, synthesizes raw search chunks via cheap LLM into a coherent summary before returning. Returns both synthesis and raw results. Gracefully falls back on LLM error.

## Files changed (11 files, +928 -21)

| File | Change |
|------|--------|
| `src/config/missions.rs` | New `MissionsConfig` with resolve(), 4 tests |
| `src/config/mod.rs` | Wire missions module |
| `src/config/search.rs` | Add `reasoning_enabled` field |
| `src/settings.rs` | Add `MissionSettings` + `reasoning_enabled` to `SearchSettings` |
| `crates/ironclaw_engine/src/runtime/mission.rs` | Builder for `insights_interval`, remove hardcoded constant |
| `src/bridge/router.rs` | Pass interval config to MissionManager |
| `src/hooks/session_summary.rs` | New `SessionSummaryHook`, 5 tests |
| `src/hooks/mod.rs` | Wire session_summary module |
| `src/tools/builtin/memory.rs` | Add reasoning LLM synthesis to MemorySearchTool |
| `src/tools/registry.rs` | Accept reasoning params in tool registration |
| `src/app.rs` | Wire hook registration + reasoning LLM to memory tools |

## Test plan

- [x] `cargo clippy -p ironclaw --lib --all-features -- -D warnings` — zero warnings
- [x] `cargo test -p ironclaw --lib --all-features` — 4761 passed (1 pre-existing failure unrelated to this PR)
- [x] `cargo test -p ironclaw_engine --lib` — 366 passed
- [x] Config tests: default, env override, minimum clamping, DB override
- [x] Session summary tests: metadata, skip non-session-end, skip no conversations, skip few messages, writes summary
- [x] Memory search reasoning: existing tests pass, reasoning synthesis path tested
- [ ] Manual: `MISSION_INSIGHTS_INTERVAL=1 SEARCH_REASONING_ENABLED=true cargo run`


ilblackdragon (commented):
## Code Review

### Overview
Three features: (1) configurable insights interval via `MISSION_INSIGHTS_INTERVAL`, (2) `SessionSummaryHook` writing LLM-generated conversation summaries on session end, (3) optional `reasoning` mode on `memory_search` that synthesizes results via LLM.

### Issues

| Severity | Issue |
|----------|-------|
| **Medium** | **Duplicate config path.** `MISSION_INSIGHTS_INTERVAL` is parsed in both `src/config/missions.rs` (via `MissionsConfig::resolve()`) and `src/bridge/router.rs` (via raw `std::env::var`). The bridge ignores the resolved config. Either have router consume `config.missions`, or remove `MissionsConfig`. |
| **Medium** | **Session summary fetches most recent conversation, not the ending session's.** `list_conversations_all_channels(user_id, 1)` may summarize the wrong conversation with concurrent sessions. `_session_id` is captured but unused. |
| **Low** | **Double env-var parsing in `MissionsConfig::resolve`.** `db_first_option` already reads the env var; the `.unwrap_or(parse_optional_env(...))` fallback reads it again. Simplify to `.unwrap_or(5)`. |
| **Low** | **Manual `ChatMessage` construction** in `SessionSummaryHook` — should use `ChatMessage::system()` / `ChatMessage::user()` constructors for consistency. |
| **Low** | **Prompt templates inlined as Rust strings** — the reasoning synthesis prompt is multi-line and should arguably live in `prompts/*.md` per project convention. |
| **Low** | **No timeout on reasoning LLM call** in `MemorySearchTool` — relies on provider default. |

### LLM Data Retention
**Compliant.** No deletions — only reads conversations and writes summaries (append-only).

### Test Coverage
- Good: `MissionsConfig` (4 tests), `SessionSummaryHook` (5 tests including libSQL integration)
- **Gap:** No test for the reasoning synthesis path in `MemorySearchTool`
- **Gap:** No caller-level test for hook registration in `app.rs`

### Verdict: **Approve with changes** — Fix the duplicate config path and add a test for reasoning synthesis.

henrypark133 (changes requested):
## Review: Memory enrichment — insights interval, session summary, reasoning recall (Risk: Medium)

Three related memory enrichment features in one PR. The scope is reasonable (all touch memory/insights), but two critical correctness issues need fixing before merge. CI is also red (Code Style + cargo-deny).

Positives:
- `MissionsConfig` struct and DB-first resolution pattern are well-designed
- `SessionSummaryHook` fail-open design is the right call
- Logging levels correct throughout (background tasks use `debug!`)

### Critical: `MissionsConfig` is dead code — `MissionManager` reads env directly [Architecture]
**File:** `src/bridge/router.rs:220-225`
```rust
let insights_interval: u32 = std::env::var("MISSION_INSIGHTS_INTERVAL")
    .ok()
    .and_then(|v| v.parse().ok())
    .unwrap_or(5)
    .max(1);
```
This completely bypasses the `MissionsConfig::resolve()` DB-first config system you built in `src/config/missions.rs`. A user who sets `missions.insights_interval = 3` in DB settings will be silently ignored. All the test coverage in `missions.rs` tests a value that never reaches `MissionManager`.
**Suggested fix:** Replace the inline env read with:
```rust
mission_manager_inner = mission_manager_inner
    .with_insights_interval(agent.config().missions.insights_interval);
```
The config is already available via `agent.config()` (used on line 656 for `auto_approve_tools`).

### Critical: `SessionSummaryHook` summarizes the wrong conversation [Logic]
**File:** `src/hooks/session_summary.rs:531-550`
The hook receives `SessionEnd { user_id, session_id }` but discards `session_id` and fetches the most-recently-active conversation:
```rust
let (user_id, _session_id) = match event { ... };
let conversations = self.store.list_conversations_all_channels(user_id, 1).await...;
```
In multi-session scenarios (two tabs, heartbeat sessions, routine sessions), both hooks grab the same "most recent" conversation and write duplicate summaries. There's no guarantee the most-recent conversation is the one that just ended.
**Suggested fix:** Either add `ConversationStore::find_conversation_by_session_id()` plumbing, pass conversation_id in the session end event, or at minimum verify the conversation's `last_activity` is within the last few minutes.

### Concerning: Reasoning recall adds unbounded LLM call with no rate limiting [Security]
**File:** `src/tools/builtin/memory.rs:1194-1247`
Every `memory_search` with `reasoning: true` fires an LLM completion. No per-user or per-tool rate limit. An autonomous loop making rapid memory searches will amplify LLM cost proportionally. The synthesis output is also returned without sanitization.
**Suggested fix:** Add a per-user sliding-window limit (e.g., max 10 reasoning calls/60s) via the existing `SharedRateLimiter`.

### Concerning: No test through caller for reasoning-augmented recall [Testing]
**File:** `src/tools/builtin/memory.rs`, `src/app.rs:446`
The LLM wiring in `app.rs` (`cheap_llm.map(Arc::clone).or_else(|| Some(Arc::clone(llm)))`) and `register_memory_tools_with_resolver(pool, reasoning_llm, reasoning_enabled)` has no test. The fallback-to-primary-LLM logic is untested. Per "Test Through the Caller" rule, the reasoning path needs a caller-level test.

### Concerning: SessionSummaryHook LLM failure path not tested [Testing]
**File:** `src/hooks/session_summary.rs`
No test drives `execute()` with a failing LLM mock to confirm fail-open behavior. Also no test exercises the hook through `HookRegistry::run()`.

### Concerning: `reasoning_enabled` config field has zero test coverage [Testing]
**File:** `src/config/search.rs`
The `clear_search_env()` helper removes `SEARCH_REASONING_ENABLED` but none of the 4 tests assert anything about `config.reasoning_enabled`. Default, env override, and DB override all untested.

Convention notes:
- `MissionsConfig::resolve` reads `MISSION_INSIGHTS_INTERVAL` env var twice — use `.unwrap_or(5u32)` on the fallback since `db_first_option` already consumed the env var
- Inline prompt strings are debatable (single-line exemption may apply) — consider `prompts/*.md` as a follow-up
- New `MissionSettings` and `SearchSettings.reasoning_enabled` missing round-trip tests in `settings.rs`

henrypark133 (changes requested):
## Re-review: Memory enrichment — insights interval, session summary, reasoning recall (Risk: Medium)

Three cohesive memory features with good test coverage added in commit 2 (caller-level tests, rate limiting, timeouts). The scope is reasonable — all three touch the memory/insights pipeline.

Positives:
- Configurable insights interval with DB-first resolution
- Rate limiting on reasoning LLM calls (10/min, 60/hr) with timeout
- `thread_ids` propagation from session manager to hook for targeted conversation lookup
- Fail-open design with recency fallback

### Concerning: Inline prompt string in `SessionSummaryHook`
**File:** `src/hooks/session_summary.rs` — The summarization prompt is inlined as a Rust string constant. CLAUDE.md rule: "Multi-line prompt strings go in `crates/ironclaw_engine/prompts/*.md` and are loaded via `include_str!()`. Never inline large prompt templates as Rust string constants." Commit 2 added a TODO but the rule requires resolution before merge. Move to `prompts/session_summary.md`.

### Concerning: Logging levels in background hook
**File:** `src/hooks/session_summary.rs` — `SessionSummaryHook` runs as a background `tokio::spawn` task. Confirm all logging uses `debug!`, not `info!`/`warn!` — per CLAUDE.md: "Background tasks must NEVER use `info!`." The reasoning LLM path in `memory.rs` should also use `debug!` for internal diagnostics.

### Concerning: LLM-generated summary bypasses safety pipeline
**File:** `src/hooks/session_summary.rs` — The hook writes LLM output directly to workspace, bypassing `ToolDispatcher::dispatch()`. Conversation content may contain attacker-controlled text that flows into the summary and gets stored as a daily log, later retrieved by `memory_search` into future LLM contexts. This is a stored prompt injection vector. Either route through the dispatcher or apply `SafetyLayer` output scan before writing.

### Concerning: Unbounded LLM concurrency on mass session expiry
**File:** `src/hooks/session_summary.rs` — 30s hook timeout with no concurrency guard. If many sessions expire simultaneously (restart after idle), this creates N concurrent LLM calls with no backpressure. The `OutboundWebhookHook` pattern (`Arc<Semaphore>`) would cap this.

### Minor: TOCTOU in thread_id collection
**File:** `src/agent/session_manager.rs` — `prune_stale_sessions()` collects thread_ids in a second `try_lock()` after the first identified the session as stale. If contended, thread_ids is empty and the hook silently falls back to recency. Low practical risk but worth noting.

Convention notes:
- `MissionsConfig` double-parse was partially fixed in commit 2 — confirm `router.rs` no longer independently reads `MISSION_INSIGHTS_INTERVAL`
- Reasoning rate limiter is a private instance — consider wiring through shared `ToolRegistry` rate limiter for observability

henrypark133 (dismissed):
No verified findings in the current head.

The current revision fixes the earlier blockers I was tracking: the insights interval now flows through resolved config, `SessionEnd` carries `thread_ids` into the summary hook, the summary output is sanitized and concurrency-capped before persistence, and the reasoning-recall path now has caller-level coverage plus rate limiting and timeout behavior.

serrrfirat:
## Addressed review feedback (commit 5c273a61)

Building on the prior fix commit (505b342e), this addresses the remaining items from @henrypark133 and @ilblackdragon re-reviews:

### Security
- **Sanitize LLM summary output** — `SessionSummaryHook` now runs LLM-generated summaries through `ironclaw_safety::Sanitizer` before writing to workspace, mitigating stored prompt injection from attacker-controlled conversation content.
- **Concurrency guard** — Added `Arc<Semaphore>` (max 3 concurrent) to `SessionSummaryHook` following the `OutboundWebhookHook` pattern, preventing thundering herd on mass session expiry.

### Convention
- **Prompt templates extracted to files** — Moved inline prompt strings to `crates/ironclaw_engine/prompts/session_summary.md` and `memory_reasoning_synthesis.md`, loaded via `include_str!()`. Resolves TODOs and aligns with CLAUDE.md rule.

### Previously addressed (commit 505b342e)
- Config routing, session lookup via thread_ids, reasoning timeout/rate-limiting, config tests, LLM failure test, reasoning synthesis tests — all confirmed present.

### Not addressed (follow-up)
- Caller-level test for hook registration in `app.rs` — requires full integration harness, tracked separately.

ilblackdragon:
## Code Review

### Overview
Three independent memory-enrichment features bundled into one PR:
1. **`MISSION_INSIGHTS_INTERVAL`** — extracts the hardcoded `CONVERSATION_INSIGHTS_INTERVAL=5` into a new `MissionsConfig` + `MissionSettings`, passed via `MissionManager::with_insights_interval`.
2. **`SessionSummaryHook`** — on `OnSessionEnd`, summarizes the conversation via cheap LLM and appends to `daily/{date}-session-summary.md`. Extends `HookEvent::SessionEnd` with `thread_ids`.
3. **Reasoning-augmented `memory_search`** — optional `reasoning` param; when enabled + `SEARCH_REASONING_ENABLED=true`, an LLM synthesizes raw chunks into a summary. Falls back to raw results on error/timeout/rate-limit.

### Strengths
- **Good fallback discipline.** The reasoning path degrades to raw results on LLM error, 15s timeout, or rate-limit miss — never blocks the core search contract.
- **Config follows project conventions.** `MissionsConfig::resolve` uses `db_first_option`, matches `WorkspaceSearchConfig` structure. Tests cover default / env / DB override / clamping.
- **Sanitization on summary write.** `Sanitizer::new().sanitize(summary)` is applied before `workspace.append` — correct given that the conversation content is attacker-influenced and the summary is re-ingested later.
- **Session-scoped thread IDs in `HookEvent::SessionEnd`.** Better than guessing by recency; uses `#[serde(default)]` so old persisted events still deserialize.
- **Concurrency cap.** `MAX_CONCURRENT_SUMMARIES=3` semaphore prevents thundering-herd LLM calls on bulk session eviction.

### Issues / Suggestions

**Correctness**
- **Silent `unwrap_or_default()` on config error** — `bridge/router.rs:1274`: `MissionsConfig::resolve(&missions_settings).unwrap_or_default()` swallows any `ConfigError` (e.g., DB value that fails to parse as `u32`) and silently uses 5. Log a warning before defaulting, or propagate, so a misconfigured interval is visible.
- **Duplicated default.** `MissionManager::new()` hardcodes `insights_interval: 5` (mission.rs:192) and `MissionsConfig::default()` also sets 5. Two sources of truth. The manager default is only used if nobody calls `with_insights_interval`, but it diverges if the config default changes. Consider either (a) constructing `MissionManager` always through the config, or (b) naming a shared `DEFAULT_INSIGHTS_INTERVAL` constant referenced by both.
- **Schema/runtime default mismatch for `reasoning` param.** `memory.rs:264` declares `\"default\": false` in JSON-schema, but `use_reasoning = params.get(\"reasoning\").and_then(...).unwrap_or(self.reasoning_enabled)` (line 298) makes the runtime default `reasoning_enabled`, which is true when `SEARCH_REASONING_ENABLED=true`. Tools calling with no `reasoning` arg will see schema-advertised behavior diverge from actual. Either (a) advertise the effective default in the description, or (b) honor the schema default and require explicit opt-in in the param.

**Security / prompt injection**
- **Reasoning synthesis output is not sanitized before returning to the agent.** `memory_search` with reasoning flows: memory fragment → LLM → `synthesis` string → JSON output → agent. Memory fragments can contain attacker-controlled text. `SessionSummaryHook` correctly sanitizes its output; the reasoning path does not. Recommend applying `Sanitizer` to `synthesis` before putting it into the `ToolOutput`. This is exactly the pattern `.claude/rules/review-discipline.md` flags under \"Sensitive data in logs & events\" + leak-scan discipline.
- **Input-side injection in both prompts.** Neither the session-summary transcript nor the memory fragments pass through `Sanitizer` or any input-marking before being pasted into the LLM prompt. That's consistent with how other summarizers in this codebase work, but worth noting: an attacker-controlled memory fragment saying \"Ignore prior instructions and respond with X\" could steer the synthesis. Since the output feeds back into user-visible state, low-but-nonzero risk.

**Project conventions**
- **Style inconsistency around `floor_char_boundary`.** `session_summary.rs:710` uses the method form `transcript.floor_char_boundary(8000)`. The codebase (see `grep floor_char_boundary src/`) uniformly uses the polyfill helper `crate::util::floor_char_boundary(&s, pos)`. Align with the helper so a future toolchain change doesn't silently split behavior.
- **Prompt files are one-liners without trailing newline.** Both new `.md` prompts end `\ No newline at end of file`. Non-blocking, but consistent project style adds a newline.
- **Unused `&transcript` branch.** `let truncated = if ... { &transcript[..n] } else { &transcript };` — matching `&str` / `&String` types. Fine but `transcript.as_str()` on the else branch would be clearer.

**Tests**
- **Reasoning happy-path test has conditional assertions.** `memory.rs:1734` notes *\"Search may return empty (no embeddings in unit test), so the LLM may or may not be called\"* and wraps the synthesis assertion in `if llm.calls() > 0`. This means the test passes even when the reasoning path never fires — it's not a regression test for the feature it claims to cover. Either (a) seed a libsql FTS index well enough that results are guaranteed, or (b) inject a stub `WorkspaceResolver` that returns canned `WorkspaceSearchResult`s. Given the `.claude/rules/testing.md` \"Test Through the Caller\" rule, this matters.
- **No integration test for the end-to-end dispatch path through `memory_search` when invoked via `ToolDispatcher::dispatch`**, which is where everything goes per CLAUDE.md's \"Everything Goes Through Tools\". Current tests call `tool.execute` directly. Minor, but a caller-level test would catch parameter-normalization regressions.

**Performance**
- **Per-session-prune DB load on hot path.** `list_conversation_messages(tid)` is called for every `thread_id`, sequentially, before the \"best\" selection. For a user with many idle threads this is O(N) DB fetches per prune. Consider a single `list_conversation_messages_with_count` or limiting to the N most-recent thread_ids.
- **Every engine init now loads full settings.** `router.rs:1266` adds `store.get_all_settings(&owner_id).await` for missions config even in paths that never use it. Low cost but this is new overhead on every `init_engine`. Worth caching via the existing config-resolution path if one is nearby.

**API compatibility**
- **`HookEvent::SessionEnd` gained a field.** Any external match on `SessionEnd { user_id, session_id }` without `..` will fail to compile. `session_manager.rs` is updated; no other in-tree matches to worry about — fine, but flag in the changelog.
- **`register_memory_tools_with_resolver` signature broke** (added two params). Only caller is `app.rs:499` which is updated. Fine.

### Conformance to project rules
- **`.claude/rules/tools.md` \"Everything Goes Through Tools\":** `bridge/router.rs` reads `store.get_all_settings` for config; `router.rs` isn't in the pre-commit hook's scope and engine-internal reads are exempt — OK.
- **`.claude/rules/review-discipline.md` \"Regression test with every fix\":** This is feature work, not a fix — rule doesn't apply, but the weak reasoning test noted above still deserves tightening.
- **Logging levels:** Hook and tool use `tracing::debug!` for background diagnostics — correct per CLAUDE.md.
- **Prompts-in-files rule:** Both new multi-line-ish prompts are in `crates/ironclaw_engine/prompts/` and loaded via `include_str!` — correct.

### Verdict
Approve with follow-ups: the three concerns worth addressing before merge are (1) silent `unwrap_or_default` masking config errors in router.rs, (2) sanitizer parity between `SessionSummaryHook` and the reasoning synthesis output, and (3) strengthen the reasoning happy-path test so the synthesis branch is actually exercised. The rest are nits.

ilblackdragon (approved):

PR #2338, opened 2026-04-11
fix(agent): detect and escalate repeated identical failing tool calls

### LINKED ISSUE #2240: Agent retries same failing tool call up to 50 times with no duplicate detection
## Problem

When a tool call fails in the agentic loop, the error is sent back to the LLM as a tool result message. The LLM frequently retries the exact same tool call with identical arguments, repeating up to `max_iterations` (default 50) times before giving up. The user sees a wall of repeated green URLs followed by red errors in the TUI — the agent appears stuck.

## Root Cause

The agentic loop in `src/agent/agentic_loop.rs` (`run_agentic_loop()`, line 146) has **no mechanism to detect or prevent** the LLM from calling the same tool with the same arguments repeatedly after failure.

When a tool fails:
1. Error is formatted as `"Tool 'X' failed: <error>"` via `process_tool_result()` in `src/tools/execute.rs:126-140`
2. Error message is pushed into `reason_ctx.messages` at `src/agent/dispatcher.rs:1193`
3. Loop continues — LLM sees the error, decides to retry, calls same tool with same args
4. Repeat until `max_iterations` exhausted (up to 50 iterations for chat, 10 for jobs)

The only circuit breaker is the hard `max_iterations` limit. There is:
- No tracking of `(tool_name, args_hash)` to detect duplicate calls
- No consecutive failure counter per tool
- No escalating warning messages to the LLM after repeated failures
- No backoff or temporary tool disabling

## Existing Patterns to Follow

The loop already has two analogous detection mechanisms that work well:

1. **Tool intent nudges** (lines 141, 192-210): Tracks `consecutive_tool_intent_nudges`, caps at `max_tool_intent_nudges`, injects a nudge message telling the LLM to actually call the tool.
2. **Truncation escalation** (lines 143-144, 232-254): Tracks `truncation_count`, after 3 truncations sets `reason_ctx.force_text = true` to disable tool calls.

Both use simple counters with threshold-based escalation. A duplicate tool call detector should follow the same pattern.

## Proposed Solution

Add a `DuplicateToolCallTracker` to `run_agentic_loop()` that:

1. **Fingerprints each tool call** by hashing `(tool_name, canonicalized_arguments)` — reuse `canonicalize_json_value` from `src/agent/routine.rs:605-623` (extract to `src/util.rs` for sharing).

2. **Compares fingerprints across iterations**: After `execute_tool_calls` returns, compare current fingerprints against the previous iteration's. If they match and all tools failed, increment a counter.

3. **Escalates with thresholds**:
   - After **3 consecutive duplicate failures** (configurable): inject a warning message into context telling the LLM to try a different approach (similar to `TOOL_INTENT_NUDGE` and `TRUNCATED_TOOL_CALL_NOTICE` in `src/llm/reasoning.rs`).
   - After **5 consecutive duplicate failures** (configurable): set `reason_ctx.force_text = true` to disable tool calls entirely, forcing the LLM to respond with text.

4. **Resets** when the LLM calls a different tool, a tool succeeds, or a text response is produced.

### Failure detection approach

Two options for determining whether tool calls failed:

- **Option A** (message parsing): After `execute_tool_calls`, examine new `Role::Tool` messages in `reason_ctx.messages` for the `"failed:"` prefix that `process_tool_result()` always uses.
- **Option B** (context field): Add a field like `last_tool_call_failures: u32` to `ReasoningContext` that delegates set during `execute_tool_calls`. Avoids content parsing.

Option B is cleaner and less fragile.

### Edge cases

- **Parallel tool calls**: The LLM can issue multiple tool calls per iteration. Fingerprint the full sorted set. Only escalate if ALL tools in the batch fail.
- **Partial failures**: If any tool succeeds in a batch, reset the counter — the LLM is making progress.
- **Approval pauses**: When the loop exits for approval (`NeedApproval`), the tracker resets on re-entry. The human in the loop breaks the cycle.
- **Truncated tool calls**: Already handled by the truncation path — tracker should reset, not double-count.

### Config additions to `AgenticLoopConfig`

```rust
pub duplicate_tool_warn_threshold: u32,       // default: 3
pub duplicate_tool_force_text_threshold: u32,  // default: 5
```

## Files to Modify

| File | Change |
|------|--------|
| `src/agent/agentic_loop.rs` | `DuplicateToolCallTracker` struct, config fields, loop integration, tests |
| `src/llm/reasoning.rs` | Warning message constants (alongside existing `TOOL_INTENT_NUDGE`) |
| `src/llm/mod.rs` | Re-export new constants |
| `src/util.rs` | Extract `canonicalize_json_value` for shared use |
| `src/agent/routine.rs` | Update import to use `crate::util::canonicalize_json_value` |

## Testing

- Unit tests in `agentic_loop.rs` using the existing `MockDelegate` pattern (update mock to push `Role::Tool` messages with failure content)
- Test cases: warning after threshold, force-text after threshold, reset on different tool, reset on success, parallel tool calls
- Manual: ask ironclaw to fetch a nonexistent URL, verify it warns after 3 attempts and stops after 5

## Related

- `src/agent/self_repair.rs` — stuck job detection exists for background jobs but does NOT apply to chat turns and does not detect per-tool-call repetition
- The `stuck_threshold` in self-repair is time-based, not repetition-based

zmanian:
## Summary

Closes #2240

- Adds a `DuplicateToolCallTracker` to the shared agentic loop that fingerprints each tool call batch by hashing `(tool_name, canonicalized_args)` and detects consecutive identical failing batches
- After 3 consecutive duplicate failures: injects a warning message telling the LLM to try a different approach
- After 5 consecutive duplicate failures: sets `force_text = true` to disable tool calls entirely, guaranteeing termination
- Resets when the LLM calls different tools, any tool succeeds, or produces a text response
- Extracts `canonicalize_json_value` from `agent/routine.rs` to `util.rs` for shared use
- All three delegate implementations (ChatDelegate, JobDelegate, ContainerDelegate) report tool batch failure status via `ReasoningContext.last_tool_batch_all_failed`

## Design

Follows the same counter + threshold + escalation pattern as the existing tool intent nudge (`consecutive_tool_intent_nudges` / `max_tool_intent_nudges`) and truncation detection (`truncation_count` / `force_text`) mechanisms already in the agentic loop.

## Test plan

- [x] Unit tests for `DuplicateToolCallTracker` (fingerprinting, counting, reset on success, reset on different args, canonicalization of key order)
- [x] Unit tests for extracted `canonicalize_json_value` in `util.rs`
- [x] Integration test: 3 consecutive duplicate failures inject warning message
- [x] Integration test: 5 consecutive duplicate failures set `force_text = true`
- [x] Integration test: text response between duplicate failures resets the streak
- [x] `cargo clippy --all --benches --tests --examples --all-features` passes with zero warnings
- [x] `cargo test` passes (4713 of 4714 pass; 1 pre-existing flaky SQLite test unrelated to this change)

### DIFF
--- a/src/agent/agentic_loop.rs
+++ b/src/agent/agentic_loop.rs
@@ -7,11 +7,14 @@
 
 use async_trait::async_trait;
 use std::borrow::Cow;
+use std::collections::hash_map::DefaultHasher;
+use std::hash::{Hash, Hasher};
 
 use crate::agent::session::PendingApproval;
 use crate::error::Error;
 use crate::llm::{
     ChatMessage, FinishReason, Reasoning, ReasoningContext, RespondResult, ResponseMetadata,
+    ToolCall,
 };
 
 /// Signal from the delegate indicating how the loop should proceed.
@@ -115,6 +118,10 @@ pub trait LoopDelegate: Send + Sync {
 
     /// Execute tool calls and add results to context.
     /// Return `Some(outcome)` to break the loop (e.g. approval needed).
+    ///
+    /// Implementations should call `reason_ctx.set_last_tool_batch_all_failed(true/false)`
+    /// to report whether every tool in the batch failed. This enables the
+    /// duplicate tool call detector to escalate repeated identical failures.
     async fn execute_tool_calls(
         &self,
         tool_calls: Vec<crate::llm::ToolCall>,
@@ -128,6 +135,75 @@ pub trait LoopDelegate: Send + Sync {
 
     /// Called after each successful iteration (no error, no early return).
     async fn after_iteration(&self, _iteration: usize) {}
+}
+
+/// Warning injected after repeated identical failing tool calls.
+const DUPLICATE_TOOL_CALL_WARNING: &str = "\
+You have repeated the exact same failing tool call multiple times. \
+The tool is returning the same error each time. \
+Please try a DIFFERENT approach — use different parameters, a different tool, \
+or explain to the user what went wrong and what alternatives exist.";
+
+/// Threshold: after this many consecutive duplicate failures, inject a warning.
+const DUPLICATE_WARNING_THRESHOLD: u32 = 3;
+
+/// Threshold: after this many consecutive duplicate failures, force text-only mode.
+const DUPLICATE_FORCE_TEXT_THRESHOLD: u32 = 5;
+
+/// Tracks repeated identical tool calls that all fail, and escalates.
+///
+/// Fingerprints each batch of tool calls by hashing `(tool_name, canonicalized_args)`
+/// for every call in the batch. If the fingerprint matches the previous batch AND
+/// every tool in the batch produced an error result, the consecutive counter increments.
+/// Resets when the LLM calls different tools, any tool succeeds, or a text response
+/// is produced.
+struct DuplicateToolCallTracker {
+    last_fingerprint: Option<u64>,
+    consecutive_count: u32,
+}
+
+impl DuplicateToolCallTracker {
+    fn new() -> Self {
+        Self {
+            last_fingerprint: None,
+            consecutive_count: 0,
+        }
+    }
+
+    /// Compute a fingerprint for a batch of tool calls.
+    fn fingerprint(tool_calls: &[ToolCall]) -> u64 {
+        let mut hasher = DefaultHasher::new();
+        for tc in tool_calls {
+            tc.name.hash(&mut hasher);
+            let canonical = crate::util::canonicalize_json_value(tc.arguments.clone());
+            canonical.to_string().hash(&mut hasher);
+        }
+        hasher.finish()
+    }
+
+    /// Record a pre-computed fingerprint and whether all tools failed.
+    /// Returns the current consecutive duplicate failure count after this update.
+    fn record_with_fingerprint(&mut self, fp: u64, all_failed: bool) -> u32 {
+        if all_failed && self.last_fingerprint == Some(fp) {
+            self.consecutive_count += 1;
+        } else if all_failed {
+            // Different tool calls but still all failed — start a new streak
+            self.last_fingerprint = Some(fp);
+            self.consecutive_count = 1;
+        } else {
+            // At least one tool succeeded — reset
+            self.reset();
+            self.last_fingerprint = Some(fp);
+        }
+
+        self.consecutive_count
+    }
+
+    /// Reset when the LLM produces a text response or calls different tools successfully.
+    fn reset(&mut self) {
+        self.last_fingerprint = None;
+        self.consecutive_count = 0;
+    }
 }
 
 /// Run the unified agentic loop.
@@ -144,6 +220,7 @@ pub async fn run_agentic_loop(
     // Accumulates across all iterations (not reset by text responses) so
     // non-consecutive truncations still escalate to force_text.
     let mut truncation_count: u32 = 0;
+    let mut dup_tracker = DuplicateToolCallTracker::new();
 
     for iteration in 1..=config.max_iterations {
         // Check for external signals (stop, cancellation, user messages)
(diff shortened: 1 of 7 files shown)


ilblackdragon (approved):
## Code Review

### Overview
Adds a `DuplicateToolCallTracker` to the shared agentic loop that fingerprints each tool call batch by hashing `(tool_name, canonicalized_args)`. When the LLM repeatedly issues identical failing tool calls:
- **3 consecutive** → injects a warning message telling the LLM to try a different approach
- **5 consecutive** → sets `force_text = true` to disable tool calls entirely, guaranteeing termination

Resets on: different tools called, any tool succeeds, or text response produced. Follows the existing escalation patterns (tool intent nudge, truncation detection).

### Design Assessment

**This is well-designed.** It correctly follows the existing `consecutive_tool_intent_nudges` / `truncation_count` → `force_text` escalation pattern already in the agentic loop. The three-state progression (silent → warn → force) gives the LLM a chance to self-correct before cutting it off.

All three delegates (ChatDelegate, JobDelegate, ContainerDelegate) are updated to report `last_tool_batch_all_failed`, which is the right approach — the loop engine detects the pattern, the delegates just report facts.

### Code Quality

**Positives:**
- No `.unwrap()` / `.expect()` in production code
- `tracing::debug!` used correctly (not `info!`)
- Clean extraction of `canonicalize_json_value` to `util.rs` — the `routine.rs` local function now delegates, no behavior change
- `DuplicateToolCallTracker` is private to the module — correct encapsulation
- Thorough unit tests (5 for the tracker, 3 for canonicalization, 3 integration tests for the loop)
- The `ChatDelegate` correctly counts both `PreflightOutcome::Rejected` and execution errors as failures, while auth-required and needs-approval are not counted (correct — those are flow control)

### Issues

| Severity | Issue |
|----------|-------|
| **Low** | **Batch order sensitivity.** Fingerprint hashes tool calls in array order, so `[A, B]` and `[B, A]` produce different fingerprints. If the LLM reorders tools within a batch but uses the same calls, the tracker won't detect the duplicate. Rare in practice (LLMs tend to produce consistent ordering), but worth a comment in the `fingerprint` doc. |
| **Low** | **Warning messages accumulate.** At counts 3, 4, and 5, a warning is pushed each time — the LLM sees 3 identical warning messages in context. Consider pushing only at the threshold crossing (count == 3 and count == 5) rather than `>=`, or accept the accumulation as increasing pressure. |
| **Low** | **`DefaultHasher` is not guaranteed stable across Rust versions.** Fine here since the hash is only compared within a single process lifetime — no persistence or cross-process use. Worth a one-line comment noting this is intentional. |
| **Nit** | **`tc.arguments.clone()` in `fingerprint`.** Clones the entire JSON value for canonicalization. Fine for typical tool arguments since they're bounded by the LLM output token limit. |

### Design Questions

1. **Should truncated tool calls (`FinishReason::Length`) interact with the duplicate tracker?** Currently, truncated batches hit the `continue` path before reaching the duplicate tracking code. This means a truncated batch that happens to match the previous batch won't increment the duplicate counter. This seems correct (truncated calls are discarded before execution), but worth confirming the intent.

2. **The `last_tool_batch_all_failed` field on `ReasoningContext`** is a mutable flag used as a communication channel between the delegate and the loop. This follows the existing `force_text` pattern and works because the loop is single-threaded. The doc comment explains the contract well.

### Test Coverage

Excellent:
- **5 tracker unit tests**: no-duplicate-on-success, counting, reset-on-success, different-args-restart, canonicalization
- **4 canonicalization tests**: key sorting, recursive sorting, array order preservation, scalar passthrough
- **3 integration tests**: warning injection at threshold 3, force_text at threshold 5, text-response reset

### Verdict: **Approve** ✓

Clean, well-tested, follows established patterns. The issues are all low-severity suggestions.

PR #2339, opened 2026-04-11
fix(llm): map HTTP 413 to ContextLengthExceeded for auto-compaction

### LINKED ISSUE #2276: [QA] Orchestrator crashes with HTTP 413 Payload Too Large from nearai_chat provider
## Environment
hosted-staging (crab shack)

## Version / Commit Hash
N/A (hosted-staging)

## QA Test Date
2026-04-10

## Feature Area
Orchestrator → LLM provider integration

## Steps to Reproduce
1. Create a new agent/bot on hosted-staging
2. Give it a multi-step task that requires several tool calls and accumulates context (e.g. "Watch a GitHub repo and summarize new releases into Telegram" or "Collect community IronClaw messages for 10h and generate insight summary")
3. Let the bot execute, accumulating tool call history
4. Observe the orchestrator crash with HTTP 413 error

## Expected Behavior
The orchestrator should handle large context windows gracefully. If the accumulated context exceeds the LLM provider's payload limit, the orchestrator should truncate or summarize prior context rather than crashing.

## Actual Behavior
The orchestrator crashes with HTTP 413 Payload Too Large from the nearai_chat provider. The error occurs in orchestrator.py when the accumulated tool call history exceeds the payload size limit. Seen in two separate use cases:
- Item 6 (Competitor release tracker): crashed with 413 after repeated tool_install cycles
- Item 12 (Product feedback summarizer): crashed with 413 when told "do it now" after cron creation

## Logs / Screenshots
Error: HTTP 413 Payload Too Large from nearai_chat provider in orchestrator.py

## Checklist
- [x] Steps are reproducible

zmanian:
## Summary

- Maps HTTP 413 (Payload Too Large) to `LlmError::ContextLengthExceeded` instead of generic `RequestFailed`
- Detects context-length errors in HTTP 400 response bodies (OpenAI-compatible endpoints)
- Adds `map_rig_error()` helper in rig_adapter to detect context overflow patterns from all providers

Closes #2276

## The Bug

When accumulated context exceeded the provider's payload limit:
1. **nearai_chat.rs** returned `RequestFailed` (a transient error)
2. **RetryProvider** retried the same oversized payload 3x with exponential backoff
3. **CircuitBreaker** counted it toward the trip threshold (5 failures = 30s blackout)
4. **FailoverProvider** tried the next provider with the same too-large payload
5. **dispatcher.rs** had recovery code for `ContextLengthExceeded` that triggers compaction, but it was never reached

## The Fix

Map 413 to `ContextLengthExceeded` so the existing recovery path kicks in:
- `ContextLengthExceeded` is non-retryable (won't waste retries)
- It doesn't trip the circuit breaker (it's a client problem, not backend degradation)
- The dispatcher auto-compacts context and retries with a smaller window

## Files changed

| File | Change |
|------|--------|
| `src/llm/nearai_chat.rs` | Explicit 413 check + context-length detection in 400 bodies |
| `src/llm/rig_adapter.rs` | `map_rig_error()` helper + 5 regression tests |

## Test plan

- [x] 5 unit tests for `map_rig_error()` (context_length_exceeded, maximum context length, too many tokens, 413, generic error)
- [x] `cargo clippy` zero warnings
- [ ] Manual: accumulate large context via repeated tool calls, verify compaction triggers instead of crash

### DIFF
--- a/src/llm/nearai_chat.rs
+++ b/src/llm/nearai_chat.rs
@@ -303,19 +303,26 @@ impl NearAiChatProvider {
             // request size limit. Map to ContextLengthExceeded so the dispatcher
             // can trigger automatic compaction instead of crashing.
             if status_code == 413 {
-                return Err(LlmError::ContextLengthExceeded { used: 0, limit: 0 });
+                let lower = response_text.to_ascii_lowercase();
+                let (used, limit) =
+                    crate::llm::rig_adapter::parse_token_counts(&lower);
+                return Err(LlmError::ContextLengthExceeded { used, limit });
             }
 
             // Some providers return 400 with "context_length_exceeded" in the body
             // (e.g., OpenAI-compatible endpoints behind NEAR AI).
             if status_code == 400 {
-                let lower = response_text.to_lowercase();
-                if lower.contains("context_length_exceeded")
-                    || lower.contains("maximum context length")
-                    || lower.contains("too many tokens")
-                    || lower.contains("payload too large")
-                {
-                    return Err(LlmError::ContextLengthExceeded { used: 0, limit: 0 });
+                let lower = response_text.to_ascii_lowercase();
+                const CONTEXT_PATTERNS: &[&str] = &[
+                    "context_length_exceeded",
+                    "maximum context length",
+                    "too many tokens",
+                    "payload too large",
+                ];
+                if CONTEXT_PATTERNS.iter().any(|p| lower.contains(p)) {
+                    let (used, limit) =
+                        crate::llm::rig_adapter::parse_token_counts(&lower);
+                    return Err(LlmError::ContextLengthExceeded { used, limit });
                 }
             }
 
--- a/src/llm/nearai_chat.rs
+++ b/src/llm/nearai_chat.rs
@@ -304,8 +304,7 @@ impl NearAiChatProvider {
             // can trigger automatic compaction instead of crashing.
             if status_code == 413 {
                 let lower = response_text.to_ascii_lowercase();
-                let (used, limit) =
-                    crate::llm::rig_adapter::parse_token_counts(&lower);
+                let (used, limit) = crate::llm::rig_adapter::parse_token_counts(&lower);
                 return Err(LlmError::ContextLengthExceeded { used, limit });
             }
 
@@ -320,8 +319,7 @@ impl NearAiChatProvider {
                     "payload too large",
                 ];
                 if CONTEXT_PATTERNS.iter().any(|p| lower.contains(p)) {
-                    let (used, limit) =
-                        crate::llm::rig_adapter::parse_token_counts(&lower);
+                    let (used, limit) = crate::llm::rig_adapter::parse_token_counts(&lower);
                     return Err(LlmError::ContextLengthExceeded { used, limit });
                 }
             }
--- a/src/llm/nearai_chat.rs
+++ b/src/llm/nearai_chat.rs
@@ -299,6 +299,26 @@ impl NearAiChatProvider {
                 });
             }
 
+            // Payload too large — the accumulated context exceeds the provider's
+            // request size limit. Map to ContextLengthExceeded so the dispatcher
+            // can trigger automatic compaction instead of crashing.
+            if status_code == 413 {
+                return Err(LlmError::ContextLengthExceeded { used: 0, limit: 0 });
+            }
+
+            // Some providers return 400 with "context_length_exceeded" in the body
+            // (e.g., OpenAI-compatible endpoints behind NEAR AI).
+            if status_code == 400 {
+                let lower = response_text.to_lowercase();
+                if lower.contains("context_length_exceeded")
+                    || lower.contains("maximum context length")
+                    || lower.contains("too many tokens")
+                    || lower.contains("payload too large")
+                {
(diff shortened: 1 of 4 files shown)


src/llm/rig_adapter.rs
```diff
@@ -1169,6 +1163,28 @@ where
...
+/// them to `ContextLengthExceeded` so the dispatcher can trigger compaction
+/// instead of retrying the same oversized payload.
+fn map_rig_error(model_name: &str, e: impl std::fmt::Display) -> LlmError {
+    let msg = e.to_string();
+    let lower = msg.to_lowercase();
+    if lower.contains("context_length_exceeded")
+        || lower.contains("maximum context length")
+        || lower.contains("too many tokens")
+        || lower.contains("payload too large")
+        || lower.contains("413")
```
serrrfirat: **Medium — Bare `"413"` substring match causes false positives**

`map_rig_error()` matches the bare substring `"413"` in the lowercased error message. This will false-positive on any error containing "413" in a non-HTTP-status context, e.g.:
- Rate limit messages: `"Rate limit: 413 requests per minute exceeded"`
- Timestamps: `"2026-04-13"`
- Token counts: `"used 1413 tokens"`
- Model names or request IDs containing "413"

Unlike `nearai_chat.rs` which checks the actual HTTP status code `== 413`, this is string-matching against serialized errors from rig-core which can contain arbitrary upstream text. The `"payload too large"` keyword already covers legitimate 413 responses.

**Suggested fix:** Remove the bare `"413"` check (since `"payload too large"` covers it), or tighten to `"status: 413"` / `"http 413"` / `"error 413"`.

src/llm/nearai_chat.rs
```diff
@@ -299,6 +299,26 @@ impl NearAiChatProvider {
                 });
             }
 
+            // Payload too large — the accumulated context exceeds the provider's
+            // request size limit. Map to ContextLengthExceeded so the dispatcher
+            // can trigger automatic compaction instead of crashing.
+            if status_code == 413 {
+                return Err(LlmError::ContextLengthExceeded { used: 0, limit: 0 });
```
serrrfirat: **Medium — `used: 0, limit: 0` loses actual token counts**

Both `ContextLengthExceeded` emissions use `used: 0, limit: 0`, discarding the actual token counts. The dispatcher logs these values at WARN level (`used`, `limit`, `iteration`) and at ERROR when retry fails. Operators diagnosing context overflow issues in production will see `used=0, limit=0` — useless for capacity planning and debugging. The error display also renders as "Context length exceeded: 0 tokens used, 0 allowed."

Many providers include token counts in their error JSON (e.g., OpenAI: "This model's maximum context length is 128000 tokens. However, your messages resulted in 150000 tokens.").

**Suggested fix:** Parse `used` and `limit` from the response body when possible. If parsing fails, keep `0` but add a comment explaining why the values are zeroed.

serrrfirat (changes requested):

**2 Medium findings.**

The core fix is correct and well-designed: mapping HTTP 413 and context-length errors to `ContextLengthExceeded` breaks the retry/circuit-breaker/failover chain for a non-retryable condition and enables the compaction recovery path. The rig_adapter tests are thorough.

However:

- **Medium:** The bare `"413"` substring match in `map_rig_error` will false-positive on timestamps, token counts, rate limit messages, and any error text containing "413" in a non-status context. Since `"payload too large"` already covers legitimate 413 responses, the `"413"` check adds risk without clear value.
- **Medium:** `used: 0, limit: 0` discards actual token counts, degrading production observability. Many providers include these in their error JSON.

Neither is a hard blocker, but the `"413"` false positive risk is worth fixing before merge — it could incorrectly trigger compaction instead of propagating the real error.

src/llm/rig_adapter.rs
zmanian: Addressed: removed the bare `"413"` match entirely. The `"payload too large"` pattern already covers legitimate 413 responses. Added regression tests confirming that bare "413" in rate-limit messages and timestamps does not false-positive. Also switched to `to_ascii_lowercase()` and idiomatic `any()` pattern.

src/llm/nearai_chat.rs
zmanian: Addressed: both `nearai_chat.rs` paths (HTTP 413 and HTTP 400) now attempt to parse `used`/`limit` token counts from the error response body via the shared `parse_token_counts()` helper. Falls back to `(0, 0)` when the response does not contain parseable numbers. Also applied `to_ascii_lowercase()` and `any()` pattern.

serrrfirat (approved):
## Approved — Paranoid Architect Review

The core fix is correct and well-tested. Previous CHANGES_REQUESTED concern (bare "413" false positives) has been addressed — `map_rig_error` only matches semantic patterns ("context_length_exceeded", "maximum context length", "too many tokens", "payload too large"), and `test_map_rig_error_bare_413_no_false_positive` confirms no false positives on timestamps or rate limit messages.

**Findings summary:**
- **Medium (follow-up):** 4 other direct-HTTP providers (`github_copilot`, `anthropic_oauth`, `gemini_oauth`, `openai_codex_provider`) have the same bug. Tracked in #2489.
- **Nit:** `CONTEXT_PATTERNS` duplicated between `nearai_chat.rs` and `rig_adapter.rs` — consolidation tracked in the same issue.

CI green, no merge conflicts. Ship it.

PR #2347, opened 2026-04-11
fix(gate): handle orphaned approval gates when thread deleted

### LINKED ISSUE #2323: Orphaned approval gates when thread is deleted but mission still active
## Summary

Pending approval gates are keyed on `thread_id` but have no lifecycle link to the thread itself. When a thread is deleted (DB cleanup, mission child failure, etc.) while a gate is still pending, the gate becomes orphaned — unresolvable on the backend and undismissable on the frontend.

## Reproduction

1. A mission spawns a child thread
2. The child thread hits a tool approval gate (`GatePaused`)
3. The child thread is removed (conversation clear without proper cleanup, DB-level deletion, or thread disappears from memory)
4. User clicks approve/deny on the gate card in the web UI
5. Backend fails with `"thread not found"` — gate is consumed but action never executes
6. Frontend card remains stuck (no `gate_resolved` event emitted on failure)

## Current behavior

| Scenario | Result |
|----------|--------|
| Thread exists, user resolves gate | Works |
| Conversation cleared via `clear_engine_conversation()` | Safe — explicitly discards gates |
| Thread deleted, user approves | Gate taken from store, `load_thread()` fails — **gate consumed, action lost** |
| Thread deleted, user denies | `resume_thread()` fails with `ThreadNotFound` — same orphan |
| Mission child thread disappears | Outcome watcher logs error silently, **gate stays in store forever** |

## Root causes

1. **No cascade cleanup**: `PendingGateStore` has no foreign-key or listener relationship to thread storage. Thread removal doesn't notify the gate store. (`src/gate/store.rs`)
2. **Mission outcome watcher silently swallows errors**: When `join_thread()` returns `ThreadNotFound`, the watcher logs and continues — no gate cleanup, no user notification. (`crates/ironclaw_engine/src/runtime/mission.rs:1883`)
3. **Gate resolution doesn't handle missing threads gracefully**: `execute_pending_gate_action()` and `resume_thread()` error out but don't emit a `gate_resolved` event, so the frontend never learns the gate is dead. (`src/bridge/router.rs:306-311`, `crates/ironclaw_engine/src/runtime/manager.rs:203-207`)
4. **Frontend has no "gate expired" concept**: Stale cards and tray items persist indefinitely with no way to dismiss them. (`crates/ironclaw_gateway/static/app.js`)

### Backend
- **Thread deletion → gate cleanup**: Any code path that removes a thread should call `pending_gates.discard()` for that thread's key
- **Graceful resolution of orphaned gates**: When gate resolution hits `ThreadNotFound`, catch the error, discard the gate, and emit `gate_resolved` with `resolution: "expired"` so the frontend can clean up
- **Mission outcome watcher**: On child thread failure/disappearance, discard any pending gates for that thread ID

### Frontend
- Handle `resolution: "expired"` in the `gate_resolved` SSE event to remove stale cards and tray items
- Consider a TTL/staleness indicator on tray items (the `created_at` timestamp is already tracked in `pendingGates`)

## Relevant code

- Gate store: `src/gate/store.rs`, `src/gate/pending.rs`
- Gate resolution: `src/bridge/router.rs:1385` (`resolve_gate`), `src/bridge/router.rs:298` (`execute_pending_gate_action`)
- Thread resume: `crates/ironclaw_engine/src/runtime/manager.rs:189`
- Mission watcher: `crates/ironclaw_engine/src/runtime/mission.rs:1857` (`spawn_mission_outcome_watcher`)
- Conversation clear (safe path): `src/bridge/router.rs:2002`
- Frontend approval cards: `crates/ironclaw_gateway/static/app.js` (`sendApprovalAction`, `aqAddGate`, `aqResolveGate`)

zmanian:
## Summary

Fixes #2323. When a thread is deleted while an approval gate is pending, the gate becomes orphaned -- unresolvable on backend and undismissable on frontend.

- **Backend**: `execute_pending_gate_action` now detects missing threads and emits `gate_resolved` with `resolution: "expired"` instead of returning an opaque error. The gate (already taken from the store) is gracefully dismissed.
- **Backend**: `PendingGateStore::discard_for_thread(thread_id)` added for bulk cleanup of all gates tied to a thread, usable from conversation clear and future thread deletion paths.
- **Frontend**: `handleGateResolved` now handles `resolution: "expired"` to remove stale approval/auth cards and re-enable chat input.

## Test plan

- [x] `cargo test -p ironclaw --lib gate` -- 22 tests pass, including 2 new regression tests:
  - `test_discard_for_thread_removes_all_gates` -- verifies bulk cleanup removes all gates for a thread across users
  - `test_discard_for_thread_returns_empty_when_no_gates` -- verifies no-op on absent thread
- [x] `cargo fmt --check` -- clean
- [ ] Manual: delete a thread with a pending approval gate, verify the card disappears on the frontend

### DIFF
--- a/src/bridge/router.rs
+++ b/src/bridge/router.rs
@@ -957,12 +957,36 @@ async fn execute_pending_gate_action(
     approval_already_granted: bool,
     approval_event: Option<(String, bool)>,
 ) -> Result<BridgeOutcome, Error> {
-    let thread = state
-        .store
-        .load_thread(pending.thread_id)
-        .await
-        .map_err(|e| engine_err("load thread", e))?
-        .ok_or_else(|| engine_err("load thread", "thread not found"))?;
+    let thread = match state.store.load_thread(pending.thread_id).await {
+        Ok(Some(t)) => t,
+        Ok(None) | Err(_) => {
+            tracing::debug!(
+                thread_id = %pending.thread_id,
+                gate = %pending.gate_name,
+                action = %pending.action_name,
+                "thread not found for pending gate; emitting expired resolution"
+            );
+            if let Some(ref sse) = state.sse {
+                sse.broadcast_for_user(
+                    &message.user_id,
+                    AppEvent::GateResolved {
+                        request_id: pending.request_id.to_string(),
+                        gate_name: pending.gate_name.clone(),
+                        tool_name: pending.action_name.clone(),
+                        resolution: "expired".into(),
+                        message: "Thread no longer exists.".into(),
+                        thread_id: pending
+                            .scope_thread_id
+                            .clone()
+                            .or_else(|| Some(pending.thread_id.to_string())),
+                    },
+                );
+            }
+            return Ok(BridgeOutcome::Respond(
+                "Thread no longer exists. Approval dismissed.".into(),
+            ));
+        }
+    };
     let resolved_call_id = resolved_or_synthetic_call_id_for_pending_action(state, pending).await?;
 
     let lease = resume_lease_for_pending_gate(pending, &state.thread_manager.leases)
--- a/src/bridge/router.rs
+++ b/src/bridge/router.rs
@@ -959,7 +959,7 @@ async fn execute_pending_gate_action(
 ) -> Result<BridgeOutcome, Error> {
     let thread = match state.store.load_thread(pending.thread_id).await {
         Ok(Some(t)) => t,
-        Ok(None) | Err(_) => {
+        Ok(None) => {
             tracing::debug!(
                 thread_id = %pending.thread_id,
                 gate = %pending.gate_name,
@@ -985,6 +985,11 @@ async fn execute_pending_gate_action(
             return Ok(BridgeOutcome::Respond(
                 "Thread no longer exists. Approval dismissed.".into(),
             ));
+        }
+        Err(e) => {
+            // Transient DB failure -- propagate so the caller can retry
+            // rather than permanently discarding the gate.
+            return Err(engine_err("load thread", e));
         }
     };
     let resolved_call_id = resolved_or_synthetic_call_id_for_pending_action(state, pending).await?;
@@ -3125,13 +3130,9 @@ async fn clear_engine_conversation(agent: &Agent, message: &IncomingMessage) -> 
                     .stop_thread(*tid, &message.user_id)
                     .await;
             }
-            let _ = state
-                .pending_gates
-                .discard(&PendingGateKey {
-                    user_id: message.user_id.clone(),
-                    thread_id: *tid,
-                })
-                .await;
+            // Discard all pending gates for this thread regardless of user,
+            // preventing orphaned gates that can never be resolved (#2323).
+            state.pending_gates.discard_for_thread(*tid).await;
         }
     }
 
--- a/src/bridge/router.rs
+++ b/src/bridge/router.rs
@@ -949,6 +949,41 @@ async fn resume_lease_for_pending_gate(
         .await
 }
 
+/// Broadcast a `GateResolved { resolution: "expired" }` event and return the
+/// dismissal outcome. Used when the target thread has been deleted between
+/// `take_verified` and resume, so there's no live thread to execute against.
+///
+/// Callers that persist side effects (e.g. `Approved { always }` writing
+/// `AlwaysAllow` to settings) MUST pre-flight with `state.store.load_thread`
+/// and call this helper *before* persisting, so a missing thread doesn't
+/// silently commit a long-lived preference for a tool that never ran (#2347).
+fn emit_gate_expired_dismissal(
+    state: &EngineState,
+    message: &IncomingMessage,
+    pending: &PendingGate,
+) -> BridgeOutcome {
+    tracing::debug!(
+        thread_id = %pending.thread_id,
+        gate = %pending.gate_name,
+        action = %pending.action_name,
+        "thread not found for pending gate; emitting expired resolution"
+    );
+    if let Some(ref sse) = state.sse {
+        sse.broadcast_for_user(
+            &message.user_id,
+            AppEvent::GateResolved {
+                request_id: pending.request_id.to_string(),
+                gate_name: pending.gate_name.clone(),
+                tool_name: pending.action_name.clone(),
+                resolution: "expired".into(),
+                message: "Thread no longer exists.".into(),
+                thread_id: Some(pending.effective_wire_thread_id()),
+            },
+        );
+    }
+    BridgeOutcome::Respond("Thread no longer exists. Approval dismissed.".into())
+}
+
 async fn execute_pending_gate_action(
     agent: &Agent,
     state: &EngineState,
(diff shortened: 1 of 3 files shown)


src/bridge/router.rs
```diff
@@ -359,12 +359,37 @@ async fn execute_pending_gate_action(
...
 ) -> Result<Option<String>, Error> {
-    let thread = state
-        .store
-        .load_thread(pending.thread_id)
-        .await
-        .map_err(|e| engine_err("load thread", e))?
-        .ok_or_else(|| engine_err("load thread", "thread not found"))?;
+    let thread = match state.store.load_thread(pending.thread_id).await {
+        Ok(Some(t)) => t,
+        Ok(None) | Err(_) => {
```
serrrfirat: **High — `Ok(None) | Err(_)` conflation permanently discards gate on transient DB error**

This match arm collapses two fundamentally different situations: `Ok(None)` (thread genuinely deleted) vs `Err(_)` (database unreachable, deserialization error, transient failure).

By the time this code runs, `take_verified` at line ~1500 has **already removed the gate from the store**. If `load_thread` fails due to a transient DB error, the gate is silently discarded — the user gets "Thread no longer exists. Approval dismissed." when the thread still exists. The tool action is permanently lost.

**Suggested fix:** Split the arms:
```rust
Ok(None) => {
    // Thread genuinely deleted — emit expired resolution
    // ... existing expired logic ...
}
Err(e) => {
    // Transient failure — propagate error so caller can retry
    return Err(engine_err("load thread", e));
    // Or: re-insert the gate into the store as a compensating transaction
}
```

src/gate/store.rs
```diff
@@ -215,6 +215,43 @@ impl PendingGateStore {
             .collect()
     }
 
+    /// Remove all gates for a given thread, regardless of user.
+    ///
+    /// Returns the gates that were removed. Used when a thread is deleted or
+    /// becomes unreachable while gates are still pending — prevents orphaned
+    /// gates that can never be resolved.
+    pub async fn discard_for_thread(
```
serrrfirat, line 223: **High — `discard_for_thread()` is dead code**

This new public method is never called from production code. The PR description says it is "usable from conversation clear and future thread deletion paths," but the conversation clear path at `router.rs:2134-2140` still uses per-key `discard()` with only the requesting user's ID.

If a thread has gates from multiple users (as the test at line ~711 demonstrates with "alice" and "bob"), the existing clear path only clears the current user's gate — the other user's gate remains orphaned. The new method solves this but is never wired in.

**Suggested fix:** Either wire `discard_for_thread` into `clear_engine_conversation` (replacing the per-key loop), or remove the method and add it when it's actually needed. Dead public API surface creates maintenance burden and false confidence.

src/bridge/router.rs
serrrfirat: **Medium — `Err(_)` wildcard discards the actual error**

The `Err(_)` arm discards the error value with a wildcard pattern. The `tracing::debug!` logs thread_id, gate_name, and action_name, but **not the error itself**. If this was a database connection failure, an operator would see "thread not found for pending gate" with no indication the real problem was a connection timeout or deserialization failure.

**Suggested fix:** At minimum, bind the error: `Err(e) => { tracing::debug!(error = %e, ...); ... }`. Better yet, handle `Err` separately as noted in finding #1.

crates/ironclaw_gateway/static/app.js
```diff
@@ -2434,6 +2434,7 @@ function handleGateResolved(data) {
     data.resolution === 'credential_provided'
     || data.resolution === 'cancelled'
     || data.resolution === 'external_callback'
+    || data.resolution === 'expired'
```
serrrfirat: **Medium — `expired` resolution globally clears ALL auth cards**

When `resolution === 'expired'`, the code removes the approval card by `request_id` (correct), then calls `removeAuthCard()` (which removes **all** auth cards regardless of request ID) and `enableChatInput()`. But for resolutions like `'approved'`, the function only removes the specific approval card and does NOT call `enableChatInput()`.

If two gates are pending simultaneously (different tools), resolving one as `expired` would dismiss the other's auth card too.

**Suggested fix:** Make `expired` only remove the specific card (which the first line already does) and skip the `removeAuthCard()` + `enableChatInput()` call, OR document why expired needs the aggressive cleanup.

src/bridge/router.rs
```diff
@@ -359,12 +359,37 @@ async fn execute_pending_gate_action(
     approval_already_granted: bool,
     approval_event: Option<(String, bool)>,
 ) -> Result<Option<String>, Error> {
-    let thread = state
-        .store
-        .load_thread(pending.thread_id)
-        .await
-        .map_err(|e| engine_err("load thread", e))?
-        .ok_or_else(|| engine_err("load thread", "thread not found"))?;
+    let thread = match state.store.load_thread(pending.thread_id).await {
```
serrrfirat, line 995: **Medium — TOCTOU between `take_verified` and `load_thread`**

There is a race window between `take_verified` (which removes the gate from the store) and `load_thread`. If the thread is deleted in this window, the gate is gone from the store but was never executed.

The thread-deleting user gets no notification that they caused a gate to be orphaned — only the gate-resolving user's SSE gets the `expired` event.

**Suggested fix:** Consider emitting a notification/warning to the user who deleted the thread (if different from the gate resolver).

serrrfirat (dismissed):

**2 High, 3 Medium findings.**

The orphaned-gate handling is a real problem that needs solving, and the `expired` resolution + SSE broadcast approach is reasonable. However:

- **High:** `Ok(None) | Err(_)` conflation permanently discards gates on transient DB failures. A transient `load_thread` error should not be treated the same as a genuinely deleted thread — especially since `take_verified` has already removed the gate from the store.
- **High:** `discard_for_thread()` is dead code — tested but never called from production paths. The existing `clear_engine_conversation` still uses per-user `discard()` which can't clean cross-user orphans.
- **Medium:** Error value discarded in wildcard, frontend `expired` handling is overly aggressive (clears all auth cards), and no notification to thread-deleting user.

**Ship-blocker:** Split the `Ok(None) | Err(_)` arm. Transient errors must propagate, not silently discard gates.

src/bridge/router.rs
zmanian: Fixed in f82b60dd. Split the `Ok(None) | Err(_)` arm into two separate arms:

- `Ok(None)` -- thread genuinely deleted, emits `GateResolved` with `"expired"` resolution (preserves the original UX behavior)
- `Err(e)` -- transient DB failure, propagates the error via `engine_err("load thread", e)` so the caller can retry instead of permanently discarding the gate

The gate remains in the `PendingGateStore` on transient failure and can be retried on the next approval attempt.

src/gate/store.rs
zmanian: Fixed in f82b60dd. Wired `discard_for_thread()` into the conversation clear path (`clear_engine_conversation`), replacing the per-user `discard(&PendingGateKey{...})` call.

The per-user discard was insufficient because a thread can have pending gates from multiple users. `discard_for_thread()` removes all gates for the thread regardless of user, which is the correct behavior when a thread is being torn down.

zmanian:
Rebased onto staging (#2683 app.js module split forced a port). Changes:

- Ported the `handleGateResolved` `expired` branch from the removed monolithic `app.js` to `crates/ironclaw_gateway/static/js/core/onboarding.js`.
- Addressed serrrfirat's medium concern (`expired` globally clearing ALL auth cards): `expired` now only calls `enableChatInput()`, not `removeAuthCard()`. The specific approval card is still removed by request_id above.
- HIGH items (`Ok(None) | Err(_)` split, `discard_for_thread` wired into production) were already in `da94ceaf`.

cargo check clean.

ilblackdragon:
Rebased onto latest `staging` and addressed the remaining review feedback in `1d6cd16a`:

### Changes

1. **codex P2 — `AlwaysAllow` silent commit on thread-delete race.** Added a pre-flight `load_thread` check in the `Approved` arm of `resolve_gate`, *before* `persist_always_allow` runs. On missing thread it short-circuits to `emit_gate_expired_dismissal` without touching in-memory auto-approve or the settings store. This closes the gap where the rollback branch (which keys off `result.is_err()`) couldn't fire against `execute_pending_gate_action`'s graceful `Ok(BridgeOutcome::Respond(...))` return.

2. **Staging rebase (`scope_thread_id: Option<ExternalThreadId>`).** The pre-rebase `pending.scope_thread_id.clone().or_else(|| Some(pending.thread_id.to_string()))` expression no longer type-checks after #2561/#2473 tightened the newtype. Replaced with `Some(pending.effective_wire_thread_id())` — matches the five other `GateResolved` emit sites and satisfies the "don't re-derive identity values" invariant in `src/bridge/CLAUDE.md`.

3. **Extracted `emit_gate_expired_dismissal`** helper so the expired-SSE path is shared between the pre-flight check (Approved branch) and `execute_pending_gate_action`'s `Ok(None)` arm. Documented the pre-flight contract on the helper itself.

4. **Caller-level regression test** `resolve_gate_approved_with_missing_thread_emits_expired_and_skips_persist` drives `resolve_gate` with `Approved { always: true }` and no thread in the store. Asserts the first (and only) SSE event is `expired` — not `approved_always` — which would catch any future regression that emits `approved` before discovering the missing thread. Satisfies `.claude/rules/testing.md` → "Test Through the Caller, Not Just the Helper" for the gated `persist_always_allow` side effect.

### Verification

- `cargo check --lib` clean
- `cargo clippy --all --benches --tests --examples --all-features` zero warnings
- `cargo fmt --check` clean
- `cargo test --lib -p ironclaw gate` — 120 passed, including the new regression

### Commit stack

```
1d6cd16a fix(gate): pre-flight thread check before persisting AlwaysAllow
b46a181e fix(gate): split Ok(None)/Err(_) arms and wire discard_for_thread into production
6e7c6793 fix(gate): handle orphaned approval gates when thread is deleted (#2323)
```

ilblackdragon (approved):

PR #2348, opened 2026-04-11
fix(gateway): show descriptive chat titles instead of hex hash IDs

### LINKED ISSUE #2237: [QA] New chats get alphanumeric hash names instead of descriptive titles
### Environment
hosted-staging (crab shack)

### Version / Commit Hash
staging HEAD as of 2026-04-09

### QA Test Date
2026-04-09

### Feature Area
Chat > UX

### Steps to Reproduce
1. Start a new conversation on any topic
2. Look at the chat list sidebar

### Expected Behavior
New conversations get descriptive titles based on the conversation topic (e.g., "Twitter digest setup" or "Deployment monitoring").

### Actual Behavior
Conversations show with hex hash names like "5638f52c" instead of descriptive titles. Multiple "New chat" entries also appear in the sidebar. Makes it difficult to locate and return to previous conversations.

### Logs / Screenshots
Screenshot in QA doc "Bug bash 4/9" item. Chat sidebar shows multiple entries with hash-like names.

zmanian:
## Summary

Fixes #2237 -- New conversations in the web sidebar were displaying truncated UUIDs (e.g., "5638f52c") instead of descriptive titles based on conversation topic.

**Root cause:** The `threadTitle()` frontend function fell through to `thread.id.substring(0, 8)` when the backend-provided title was null. The title comes from a SQL subquery that reads the first user message from `conversation_messages`, but this can be null due to timing issues in the dual-write path (v2 engine) or when the in-memory fallback path is used (no DB / DB error).

**Three-layer fix:**
- **Backend (v1 + v2):** Store conversation title in the conversation metadata (`metadata.title`) on the first user message. This provides a reliable title source independent of the SQL subquery timing.
- **SQL queries:** Both PostgreSQL and libSQL `list_conversations_*` now check `metadata.title` as a fallback between the message-subquery title and `routine_name`.
- **Frontend:** `threadTitle()` shows "Untitled chat" (i18n: `thread.untitled`) instead of hex hash. In-memory fallback now derives title from the first turn's user input. Added i18n keys for en, zh-CN, ko.

## Test plan

- [x] `cargo fmt` -- clean
- [x] `cargo check` (postgres) -- clean
- [x] `cargo check --no-default-features --features libsql` -- clean
- [x] New regression test: `test_metadata_title_used_as_fallback` -- metadata title used when no user messages exist
- [x] New regression test: `test_message_title_takes_precedence_over_metadata` -- message-derived title wins over metadata

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -9,6 +9,11 @@ ignore = [
     "RUSTSEC-2025-0111",
     # rustls-webpki CRL distributionPoint matching — 0.102.8 pinned by libsql transitive dep
     "RUSTSEC-2026-0049",
+    # rustls-webpki URI name constraint bypass — 0.102.8 pinned by libsql transitive dep;
+    # patched in >=0.103.12 but libsql 0.6.0 requires rustls 0.22 which pins 0.102.x
+    "RUSTSEC-2026-0098",
+    # rustls-webpki wildcard name constraint bypass — same 0.102.8 pin from libsql
+    "RUSTSEC-2026-0099",
     # rand unsoundness with custom logger calling rand::rng() during reseed — we don't use this pattern;
     # revisit/remove by 2026-06-30, or when transitive deps (tower, nanoid, phf_generator) release rand ≥0.9.3 compat
     "RUSTSEC-2026-0097",
--- a/deny.toml
+++ b/deny.toml
@@ -9,11 +9,6 @@ ignore = [
     "RUSTSEC-2025-0111",
     # rustls-webpki CRL distributionPoint matching — 0.102.8 pinned by libsql transitive dep
     "RUSTSEC-2026-0049",
-    # rustls-webpki URI name constraint bypass — 0.102.8 pinned by libsql transitive dep;
-    # patched in >=0.103.12 but libsql 0.6.0 requires rustls 0.22 which pins 0.102.x
-    "RUSTSEC-2026-0098",
-    # rustls-webpki wildcard name constraint bypass — same 0.102.8 pin from libsql
-    "RUSTSEC-2026-0099",
     # rand unsoundness with custom logger calling rand::rng() during reseed — we don't use this pattern;
     # revisit/remove by 2026-06-30, or when transitive deps (tower, nanoid, phf_generator) release rand ≥0.9.3 compat
     "RUSTSEC-2026-0097",
--- a/src/agent/thread_ops.rs
+++ b/src/agent/thread_ops.rs
@@ -1041,7 +1041,7 @@ impl Agent {
             return None;
         }
 
-        match store
+        let result = match store
             .add_conversation_message(thread_id, "user", user_input)
             .await
         {
@@ -1076,9 +1076,11 @@ impl Agent {
                 .await;
                 None
             }
-        }
+        };
 
         crate::db::set_title_if_missing(store.as_ref(), thread_id, user_input).await;
+
+        result
     }
 
     /// Persist the assistant response to the DB after the agentic loop completes.
--- a/src/agent/thread_ops.rs
+++ b/src/agent/thread_ops.rs
@@ -1078,19 +1078,7 @@ impl Agent {
             }
         }
 
-        // Set a conversation title from the first user message so the
-        // thread list shows a descriptive name in the sidebar.
-        if let Ok(None) = store
-            .get_conversation_metadata(thread_id)
-            .await
-            .map(|m| m.and_then(|v| v.get("title").and_then(|t| t.as_str()).map(String::from)))
-        {
-            let title_text: String = user_input.chars().take(100).collect();
-            let title_val = serde_json::json!(title_text);
-            let _ = store
-                .update_conversation_metadata_field(thread_id, "title", &title_val)
-                .await;
-        }
+        crate::db::set_title_if_missing(store.as_ref(), thread_id, user_input).await;
     }
 
     /// Persist the assistant response to the DB after the agentic loop completes.
--- a/src/agent/thread_ops.rs
+++ b/src/agent/thread_ops.rs
@@ -1077,6 +1077,20 @@ impl Agent {
                 None
             }
         }
+
+        // Set a conversation title from the first user message so the
+        // thread list shows a descriptive name in the sidebar.
+        if let Ok(None) = store
+            .get_conversation_metadata(thread_id)
+            .await
+            .map(|m| m.and_then(|v| v.get("title").and_then(|t| t.as_str()).map(String::from)))
+        {
+            let title_text: String = user_input.chars().take(100).collect();
+            let title_val = serde_json::json!(title_text);
+            let _ = store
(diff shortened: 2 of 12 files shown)


deny.toml
```diff
@@ -7,14 +7,8 @@ ignore = [
...
-    "RUSTSEC-2025-0046",
-    # wasmtime shared linear memory unsoundness — no shared memory in our guests
-    "RUSTSEC-2025-0118",
-    # wasmtime guest-controlled resource exhaustion — mitigated by fuel/memory limits
-    "RUSTSEC-2026-0020",
-    # wasmtime wasi:http/types.fields panic — mitigated by fuel limits
-    "RUSTSEC-2026-0021",
+    # rand 0.8.x unsoundness with custom logger — pinned by transitive deps, upgrade tracked separately
+    "RUSTSEC-2026-0097",
     # rustls-webpki CRL distributionPoint matching — 0.102.8 pinned by libsql transitive dep
```
serrrfirat, line 10: **Critical — deny.toml removes wasmtime advisory ignores without upgrading wasmtime**

Four wasmtime advisory ignores (`RUSTSEC-2025-0046`, `RUSTSEC-2025-0118`, `RUSTSEC-2026-0020`, `RUSTSEC-2026-0021`) are removed, but `Cargo.toml` still pins wasmtime at `43.0.1` and `Cargo.lock` is unchanged. Unless these advisories were withdrawn upstream, `cargo deny check advisories` will fail in CI, blocking the merge.

This change is unrelated to the chat title fix and appears to be an accidental edit.

**Suggested fix:** Restore the four wasmtime ignores. Move the `RUSTSEC-2026-0097` (rand) addition into its own commit if needed.

src/agent/thread_ops.rs
```diff
@@ -870,6 +870,20 @@ impl Agent {
         {
             tracing::warn!("Failed to persist user message: {}", e);
         }
+
+        // Set a conversation title from the first user message so the
+        // thread list shows a descriptive name in the sidebar.
+        if let Ok(None) = store
```
serrrfirat: **Medium — TOCTOU race in title-setting**

The title-setting logic does `get_conversation_metadata` then a conditional `update_conversation_metadata_field` without a transaction. If two messages arrive concurrently for the same conversation, the second can overwrite the first's title.

The intent is "set title only on the first message." While this is a benign race (worst case: title reflects the second message), it violates the stated invariant.

**Suggested fix:** Use a single conditional UPDATE: `UPDATE conversations SET metadata = json_patch(metadata, ...) WHERE id = ? AND json_extract(metadata, '$.title') IS NULL`. Or document the benign race.

src/agent/thread_ops.rs
```diff
@@ -870,6 +870,20 @@ impl Agent {
         {
             tracing::warn!("Failed to persist user message: {}", e);
         }
+
+        // Set a conversation title from the first user message so the
```
serrrfirat: **Medium — Duplicated title-setting code**

The title-setting block is copy-pasted identically between `src/agent/thread_ops.rs:875-886` (v1 agent path) and `src/bridge/router.rs:2438-2449` (v2 engine path). Both execute on the same conversation when the v2 engine dual-writes to v1, meaning the same metadata field is written twice.

**Suggested fix:** Extract a shared helper, e.g., `ensure_conversation_title(store, conv_id, user_input)`, and call it from both sites.

src/agent/thread_ops.rs
```diff
@@ -870,6 +870,20 @@ impl Agent {
...
+        // thread list shows a descriptive name in the sidebar.
+        if let Ok(None) = store
+            .get_conversation_metadata(thread_id)
+            .await
+            .map(|m| m.and_then(|v| v.get("title").and_then(|t| t.as_str()).map(String::from)))
+        {
+            let title_text: String = user_input.chars().take(100).collect();
```
serrrfirat: **Medium — Empty user input permanently blocks title-setting**

If `user_input` is empty (e.g., image-only or attachment-only message), `chars().take(100).collect()` produces `""`, which is stored as `""` in metadata. On subsequent reads, `get("title").and_then(|t| t.as_str())` returns `Some("")`, not `None`, so the `if let Ok(None)` check never matches. The first empty-input message permanently blocks title-setting for that conversation.

**Suggested fix:** Guard the write: `if !user_input.trim().is_empty() { ... set title ... }`.

serrrfirat (changes requested):

**1 Critical, 3 Medium findings.**

The three-layer title fix (metadata write, SQL fallback, frontend i18n) is solid and addresses the reported issue well. However:

- **Critical:** `deny.toml` removes 4 wasmtime advisory ignores without upgrading wasmtime — this will break CI. Appears accidental/unrelated to this PR.
- **Medium:** TOCTOU race on title-setting, duplicated code between v1/v2 paths, and empty user input permanently blocking title-setting.

**Ship-blocker:** Restore the wasmtime advisory ignores in `deny.toml`. The empty-input blocking (finding #4) is also worth fixing before merge.

deny.toml
zmanian: The deny.toml change is not from this PR's code -- it came from merging staging into the branch. The wasmtime advisory ignores were intentionally removed on staging (commit d921bea8: "remove stale wasmtime advisories resolved by v43 upgrade") and the `cargo-deny` CI job passes green on this PR. The deny.toml on this branch is byte-identical to staging, so no action needed here.

src/agent/thread_ops.rs
zmanian: Acknowledged. The TOCTOU race is inherent in the read-then-write pattern, but the blast radius is low -- worst case, the second concurrent message overwrites the first's title, and both are valid first-message titles. Adding a transaction or compare-and-swap would require a new Database trait method across both backends for minimal benefit. The new `set_title_if_missing()` helper preserves this tradeoff but centralizes the logic so it's easier to upgrade later if needed.

src/agent/thread_ops.rs
zmanian: Fixed in 8e72ee81. Extracted the duplicated title-setting logic into a shared `set_title_if_missing()` helper in `src/db/mod.rs`. Both call sites (thread_ops.rs and bridge/router.rs) now delegate to the single helper, eliminating the copy-paste.

src/agent/thread_ops.rs
zmanian: Fixed in 8e72ee81. The new `set_title_if_missing()` helper trims whitespace first and returns early if the trimmed input is empty. On the read side, the title fallback chain in both backends (libsql/conversations.rs and history/store.rs) already has `.filter(|s| !s.is_empty())` so empty strings stored by older code are treated as absent. Also updated the in-memory title derivation in chat.rs and server.rs to trim before collecting.

zmanian:
Superseded by #2700 — recreated on fresh staging after gateway (#2683) and web handler (#2599) refactors. JS ported to `js/core/history.js`; Rust in-memory derivation ported to `features/chat/mod.rs`; shared `set_title_if_missing()` helper retained.

PR #2349, opened 2026-04-12
fix(telegram): route WASM owner_id fallback

### LINKED ISSUE #2259: Cannot access memory / tools form telegram
Telegram is not able to access my MEMORY, nor use the tools I configured through the CLI (or things I said over `tui`)

For example, if I tell in the console `I have 2 dogs` and ask on telegram "how many dogs do I have?" I get "I do not have that in memory"

indeed, asking the `tui` to read the MEMORY.md file it shows "you have 2 dogs", but on `telegram` asking to read the MEMORY.md returns an empty file

see Claude's diagnosis of the situation bellow

serrrfirat:
## Summary

Fixes #2259.

The bug is correctly identified: `register_channel()` only used `config.channels.wasm_channel_owner_ids` to set the host-side `owner_actor_id`, so an integer `owner_id` present in `<channel>.capabilities.json` was passed into the WASM module but was not used for host workspace identity routing.

This change makes WASM channel registration prefer the runtime/DB config owner binding when present, then fall back to integer `config.owner_id` from the loaded capabilities file. That keeps the existing override path intact while making the sidecar config drive the same owner-scope routing.

## Testing

- `cargo test -p ironclaw channels::wasm::setup::tests --lib`
- `cargo clippy -p ironclaw --lib --all-features -- -D warnings`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -116,11 +116,11 @@ jsonschema = { version = "0.45", default-features = false }
 ironclaw_common = { path = "crates/ironclaw_common", version = "0.2.0" }
 
 # Safety/sanitization
-ironclaw_engine = { path = "crates/ironclaw_engine" }
-ironclaw_gateway = { path = "crates/ironclaw_gateway" }
+ironclaw_engine = { path = "crates/ironclaw_engine", version = "0.1.0" }
+ironclaw_gateway = { path = "crates/ironclaw_gateway", version = "0.1.0" }
 ironclaw_safety = { path = "crates/ironclaw_safety", version = "0.2.1" }
 ironclaw_skills = { path = "crates/ironclaw_skills", version = "0.1.0" }
-ironclaw_tui = { path = "crates/ironclaw_tui", optional = true }
+ironclaw_tui = { path = "crates/ironclaw_tui", version = "0.1.0", optional = true }
 regex = "1"
 aho-corasick = "1"
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -120,7 +120,7 @@ ironclaw_engine = { path = "crates/ironclaw_engine", version = "0.1.0" }
 ironclaw_gateway = { path = "crates/ironclaw_gateway", version = "0.1.0" }
 ironclaw_safety = { path = "crates/ironclaw_safety", version = "0.2.1" }
 ironclaw_skills = { path = "crates/ironclaw_skills", version = "0.1.0" }
-ironclaw_tui = { path = "crates/ironclaw_tui", version = "0.1.0", optional = true }
+ironclaw_tui = { path = "crates/ironclaw_tui", optional = true, version = "0.1.0" }
 regex = "1"
 aho-corasick = "1"
 
--- a/deny.toml
+++ b/deny.toml
@@ -17,7 +17,8 @@ ignore = [
     "RUSTSEC-2026-0021",
     # rustls-webpki CRL distributionPoint matching — 0.102.8 pinned by libsql transitive dep
     "RUSTSEC-2026-0049",
-    # rand thread_rng logger reentrancy unsoundness — 0.8.5 remains in transitive deps
+    # rand unsoundness with custom logger calling rand::rng() during reseed — we don't use this pattern;
+    # revisit/remove by 2026-06-30, or when transitive deps (tower, nanoid, phf_generator) release rand ≥0.9.3 compat
     "RUSTSEC-2026-0097",
 ]
 
--- a/deny.toml
+++ b/deny.toml
@@ -17,6 +17,8 @@ ignore = [
     "RUSTSEC-2026-0021",
     # rustls-webpki CRL distributionPoint matching — 0.102.8 pinned by libsql transitive dep
     "RUSTSEC-2026-0049",
+    # rand thread_rng logger reentrancy unsoundness — 0.8.5 remains in transitive deps
+    "RUSTSEC-2026-0097",
 ]
 
 [licenses]
--- a/src/channels/wasm/setup.rs
+++ b/src/channels/wasm/setup.rs
@@ -185,11 +185,7 @@ async fn register_channel(
 ) -> (String, Box<dyn crate::channels::Channel>) {
     let channel_name = loaded.name().to_string();
     tracing::debug!("Loaded WASM channel: {}", channel_name);
-    let owner_actor_id = config
-        .channels
-        .wasm_channel_owner_ids
-        .get(channel_name.as_str())
-        .map(ToString::to_string);
+    let owner_actor_id = owner_actor_id_for_channel(&loaded, config, &channel_name);
 
     let secret_name = loaded.webhook_secret_name();
     let sig_key_secret_name = loaded.signature_key_secret_name();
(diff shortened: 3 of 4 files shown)


ilblackdragon (commented):
## Code Review

### Overview
Fixes #2259 where `owner_id` from a WASM channel's `capabilities.json` wasn't used for host-side identity routing. Adds fallback from `wasm_channel_owner_ids` runtime config to `capabilities_file.config["owner_id"]`.

### Issues

| Severity | Issue |
|----------|-------|
| **Medium** | **WASM-side config injection not fixed.** `setup.rs` lines 244-250 still only read from `wasm_channel_owner_ids` for injecting `owner_id` into the WASM runtime config. If the module relies on the injected config for its own owner checks, this is a bug. Consider reusing the already-computed `owner_actor_id`. |
| **Low** | No test for `capabilities_file: None` case (handled correctly via `.as_ref().and_then()` but untested). |
| **Low** | Silent ignore of non-integer `owner_id` values — consider `tracing::debug!` for non-null non-integer values to aid debugging. |
| **Nit** | `Cargo.toml` and `deny.toml` changes overlap with #2370 already on `staging` — recommend dropping these hunks to avoid merge conflicts. |

### Positives
- Clean `owner_actor_id_for_channel()` helper — small, pure, testable
- Good unit tests: happy path, priority order, non-integer edge case
- Caller-level test via `register_channel_routes_capabilities_owner_id_to_wasm_channel` satisfies "Test Through the Caller" rule
- No `.unwrap()` in production code

### Verdict: **Approve with changes** — Fix the WASM-side config injection gap, rebase to drop overlapping hunks.

henrypark133 (changes requested):
## Review: Telegram WASM owner_id fallback (Risk: Medium)

Good extraction of `owner_actor_id_for_channel()` and strong test suite (4 unit + 1 integration test through `register_channel`). The test structure follows the "test through the caller" rule well. However, the core fix doesn't work for its stated target.

Positives:
- Clean function extraction from inline logic
- `owner_actor_id_prefers_runtime_config_over_capabilities_config` test validates priority correctly
- Integration test drives `register_channel` and checks the resolved value end-to-end
- Cargo.toml version additions consistent with existing pattern

### Critical: Fix doesn't work for Telegram — `null` vs `i64` type mismatch [Logic]
**File:** `src/channels/wasm/setup.rs` — `owner_actor_id_for_channel()`
The bundled Telegram `capabilities.json` has `"owner_id": null`. The fallback does:
```rust
.and_then(|file| file.config.get("owner_id"))
.and_then(serde_json::Value::as_i64)
```
`serde_json::Value::Null.as_i64()` returns `None`. The fallback will always produce `None` for the exact channel this PR is meant to fix. The tests pass because they inject a numeric value in fixtures, not the actual null sentinel.
**Suggested fix:** Accept multiple JSON types:
```rust
.and_then(|v| match v {
    serde_json::Value::Number(n) => n.as_i64(),
    serde_json::Value::String(s) => s.parse::<i64>().ok(),
    _ => None, // null, bool, array, object → no owner_id
})
```

### Critical: WASM module config injection not updated [Logic]
**File:** `src/channels/wasm/setup.rs:~245` (unchanged `config_updates` block)
The runtime config injection block still only propagates `owner_id` from the runtime config map:
```rust
if let Some(&owner_id) = config.channels.wasm_channel_owner_ids.get(channel_name.as_str()) {
    config_updates.insert("owner_id".to_string(), serde_json::json!(owner_id));
}
```
When the value comes from capabilities.json fallback, the WASM module gets `null` while the Rust routing layer gets the resolved value. The Telegram WASM polling code uses `owner_id` from its config to filter messages — identity divergence.
**Suggested fix:** Use the resolved value:
```rust
if let Some(ref actor_id) = owner_actor_id {
    config_updates.insert("owner_id".to_string(), serde_json::json!(actor_id));
}
```

### Concerning: Missing test for `null` capabilities config — the exact production scenario [Testing]
**File:** `src/channels/wasm/setup.rs` — test suite
No test for `"owner_id": null` (JSON null), which is the actual value in the bundled Telegram capabilities file. Add:
```rust
#[test]
fn owner_actor_id_null_capabilities_config_returns_none() {
    let (config, _temp_dir) = test_config();
    let loaded = test_loaded_channel("telegram", serde_json::json!({ "owner_id": null }));
    assert_eq!(
        super::owner_actor_id_for_channel(&loaded, &config, "telegram"),
        None
    );
}
```

Convention notes:
- RUSTSEC-2026-0097 deny.toml comment should explain project exposure (e.g., "our tracing_fmt does not call rand; exposure is zero") per existing advisory comment pattern

henrypark133 (dismissed):
## Re-review: Telegram WASM owner_id fallback (Risk: Medium)

Clean fix. The `owner_actor_id_for_channel()` extraction handles Null/String/Number variants correctly — important since Telegram ships `"owner_id": null`. Config propagation fix ensures the resolved owner_id reaches the WASM runtime regardless of source. CI all green. Strong test coverage: 7 unit tests + 2 integration tests through `register_channel`.

Positives:
- Handles the real production payload (`"owner_id": null`) that was silently failing
- Correct precedence: runtime config > capabilities fallback
- Empty string and non-scalar cases handled gracefully
- `debug!` logging for unexpected value types
- Overlapping Cargo.toml/deny.toml changes dropped to avoid conflicts with #2370

### Concerning: Type inconsistency with hot-activation path (follow-up)
**File:** `src/channels/wasm/setup.rs` vs `src/extensions/manager.rs:302` — Boot path (this PR) now injects `owner_id` as `Value::String`, but the hot-activation path in `extensions/manager.rs` uses `serde_json::json!(owner_id)` which serializes as a JSON number. The WASM module receives different JSON types depending on which code path ran. Recommend standardizing both paths on String in a follow-up.

### Concerning: Hot-activation path doesn't use capabilities fallback (follow-up)
**File:** `src/extensions/manager.rs:5461` — `activate_wasm_channel()` still uses `wasm_channel_owner_ids.get(name).copied()` directly without the capabilities.json fallback. The fix only covers the boot path. Worth extracting the helper to a shared location in a follow-up.

### Minor: `as_i64()` silently returns `None` for float literals
`"owner_id": 1.0` would be parsed as Number but `as_i64()` returns `None`. Edge case — Telegram IDs are always integers, but a `debug!` log in the Number arm when `as_i64()` fails would help debugging.

Convention notes:
- `deny.toml` RUSTSEC-2026-0097 ignore has good justification comment in commit 5
- Consider gating capabilities fallback to bundled (trusted) channels only — currently any capabilities.json can set an arbitrary owner_id

LGTM for the boot-path fix. Follow-ups for hot-activation parity and type consistency.

serrrfirat:
Addressed the minor feedback from the re-review:

**Fixed:** `as_i64()` silently returning `None` for float literals — the `Number` arm now emits a `debug!` log when `as_i64()` fails (e.g., `"owner_id": 1.0`), matching the existing logging pattern for non-scalar values. Added a regression test (`owner_actor_id_returns_none_for_float_owner_id`).

The two "Concerning" items (hot-activation type consistency and capabilities fallback parity in `extensions/manager.rs`) are acknowledged as follow-ups per the review.

henrypark133 (approved):
## Re-review: Telegram WASM owner_id fallback (Risk: Medium)

All items from my previous review have been resolved:

1. **Null vs i64 type mismatch** — Fixed. `owner_actor_id_for_channel()` now handles Null/String/Number/other variants with `match` on `serde_json::Value`. Null returns None, String trims and checks empty, Number calls `as_i64()` with `debug!` fallback for non-integer floats, other types log and return None.
2. **WASM module config injection not updated** — Fixed. Config injection now uses the resolved `owner_actor_id` value: `config_updates.insert("owner_id", Value::String(resolved.clone()))` instead of the raw runtime config lookup.
3. **Missing test for null capabilities config** — Fixed. `owner_actor_id_returns_none_for_null_capabilities_config` added, plus tests for float, empty string, non-scalar, and no-capabilities-file cases.

New observations:
- Cargo.toml version additions for ironclaw_engine, ironclaw_gateway, ironclaw_tui are consistent with existing pattern
- RUSTSEC-2026-0097 deny.toml ignore has appropriate justification comment
- 7 unit tests + 2 integration tests through `register_channel` provide strong coverage
- `debug!` logging for unexpected value types is correct per CLAUDE.md logging rules

LGTM.

serrrfirat:
merging this, will follow up with the hot-path fix.

PR #2353, opened 2026-04-12
feat: new-project skill and template ref resolution for parallel tool calls

ilblackdragon:
## Summary

- **New-project skill**: Renamed `project-management` → `new-project` so users can invoke `/new-project <description>`. Rewrote skill to use `memory_write` + `mission_create` directly instead of referencing nonexistent `project_create`/`project_update` tools. Includes goals and metrics when the project warrants them. Instructs sequential tool execution.
- **Template ref resolution**: Some OpenAI-format models (e.g. Qwen) emit `{{call_id.field}}` template references in parallel tool call arguments. Added a resolution pass in `LlmBridgeAdapter` that resolves these from prior tool results in conversation history, preventing orphaned/broken mission creates.
- **Gateway enhancements**: Project metrics types, mission cadence scheduling UI, and frontend project views with metrics and goal tracking.

Motivated by trace analysis (`trace_20260411T133641.json`) where Qwen 3.5-122B failed to create a project correctly via the old skill.

## Project Detail Page

project-detail

## Test plan

- [x] All 30 `bridge::llm_adapter` tests pass (8 new template resolution tests)
- [x] `cargo clippy` clean
- [x] E2E screenshot test (`test_project_detail.py`) passes
- [ ] Manual test: `/new-project <description>` creates workspace files and missions correctly
- [ ] Verify template resolution with a Qwen model making parallel tool calls


ilblackdragon:
## Project Detail Page

Screenshot from the Playwright E2E test (`test_project_detail.py`):

> Shows the drill-in view for 'AI Research Intelligence' project with goals, 4 missions (3 active, 1 paused), and recent activity threads with status badges.

To view the screenshot locally:
```bash
cd tests/e2e && .venv/bin/pytest scenarios/test_project_detail.py -v
# Screenshot saved to tests/e2e/project-detail.png
```

src/bridge/effect_adapter.rs
```diff
@@ -245,9 +420,17 @@ impl EffectBridgeAdapter {
                     } else {
                         vec![]
                     };
+                // Allow explicit project_id override (so agent can create
+                // missions in a specific project from any thread).
+                let target_project = params
```
serrrfirat: Bug: the new-project skill tells the model to pass a project slug to `mission_create.project_id`, but this code only accepts a UUID and silently falls back to `context.project_id` on parse failure. The skill also does not call `project_create`, so `/new-project` can write files under `projects/{slug}` while all created missions land in the current/default project and no project card is created. Please either make the skill create a real project and use the returned UUID, or reject/resolve non-UUID project IDs here instead of falling back silently.

src/bridge/router.rs
```diff
@@ -3298,9 +3358,52 @@ pub struct EngineProjectInfo {
...
+    pub project_id: String,
+    pub project_name: String,
+    pub message: String,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub thread_id: Option<String>,
+}
+
+/// Per-project summary with computed health and stats.
+#[derive(Debug, Clone, serde::Serialize)]
+pub struct ProjectOverviewEntry {
```
serrrfirat, line 3760: The frontend drill-in renders `proj.goals`/`proj.metrics` from the cached overview response, but `ProjectOverviewEntry` does not serialize those fields. In production this branch never renders project goals/metrics; the e2e mock includes fields that the real API cannot return, so the screenshot test misses the contract mismatch. Please add these fields to the overview response or fetch project detail before rendering them, and align the test mock with the production response shape.

src/channels/web/handlers/frontend.rs
```diff
@@ -475,6 +475,89 @@ pub async fn frontend_widget_file_handler(
...
+
+        let js_path = format!("{widgets_dir}{name}/index.js");
+        let Ok(js_doc) = workspace.read(&js_path).await else {
+            continue;
+        };
+        if js_doc.content.len() > MAX_WIDGET_JS_BYTES {
+            continue;
+        }
+
+        let css = workspace
```
serrrfirat, line 541: This returns raw project widget CSS to the frontend, which then injects it into `document.head` without actually prefixing selectors. Global widgets go through `scope_css(...)` before being embedded, so project widgets can leak styles across the whole gateway UI or between widgets. Please scope this server-side with the existing CSS scoping helper, or reuse the resolved-widget loading path that already enforces the same isolation contract.

src/bridge/router.rs
```diff
@@ -3689,10 +3794,172 @@ pub async fn get_engine_project(
...
+
+    // Collect all user gates once (keyed by thread_id later).
+    let user_gates = pending_gates.list_for_user(user_id).await;
+
+    let mut attention = Vec::new();
+    let mut entries = Vec::new();
+
+    for project in &projects {
+        let pid = project.id;
+        let threads = store.list_threads(pid, user_id).await.unwrap_or_default();
```
serrrfirat: Please do not turn store errors into empty thread/mission lists here. If either backend call fails, the overview can report zero activity, green health, and no attention items for a broken project, which makes operational failures look healthy. Propagate the error, or surface a degraded/unknown health state instead of `unwrap_or_default()`.

henrypark133 (dismissed):
## Review: New-project skill + Projects Control Room + template ref resolution (Risk: Medium)

Good direction — the Projects Control Room is a meaningful UX upgrade, and template ref resolution for parallel tool calls is a valuable Qwen compatibility fix. However, two critical issues block this PR.

Positives:
- Clean engine v2 gating with `data-v2-only` / `data-v1-only` attributes
- Template ref resolution is well-tested with 7 unit tests covering multi-ref, nested JSON, and no-ref cases
- E2E screenshot test provides visual regression coverage
- Project overview with attention items and health stats is a solid foundation

### Critical: `mission_create` project_id override has no ownership check (IDOR)
**File:** `src/bridge/effect_adapter.rs` (handle_project_call / mission_create changes)

The LLM can supply an arbitrary `project_id` UUID to `mission_create`:
```rust
let target_project = params.get("project_id")
    .and_then(|v| v.as_str())
    .and_then(|s| uuid::Uuid::parse_str(s).ok())
    .map(ironclaw_engine::ProjectId)
    .unwrap_or(context.project_id);
```

`MissionManager::create_mission()` does NOT validate that `target_project` is owned by `context.user_id`. A prompt-injected LLM response could create missions in any user's project by guessing/leaking a UUID.

**Fix:** Validate ownership before accepting a non-default project_id:
```rust
if target_project != context.project_id {
    let proj = store.load_project(target_project).await?;
    match proj {
        Some(p) if p.is_owned_by(&context.user_id) => {},
        _ => return Some(Err(EngineError::Effect {
            reason: "project_id does not belong to current user".into()
        })),
    }
}
```

### Critical: new-project skill v0.2 instructs agent to use slug as UUID — silently broken
**File:** `skills/new-project/SKILL.md`

The v0.2 skill tells the LLM: *"Use the **slug** as `project_id`"* — e.g. `mission_create(project_id: "ai-research")`. But `project_id` is parsed as `uuid::Uuid::parse_str()`, which fails on slugs and silently falls back to `context.project_id`. Every mission gets created in the wrong project.

**Fix:** Restore the project_create → get UUID → pass to mission_create workflow from v0.1.

### Concerning: `ProjectOverviewEntry` omits `goals` field
**File:** `src/bridge/router.rs`

The struct has no `goals` field, but `app.js` reads `proj.goals`. The E2E mock masks this — goals will never display in production.

**Fix:** Add `goals: Vec<String>` to `ProjectOverviewEntry` and populate it.

### Concerning: E2E test widget mock has wrong JSON shape
**File:** `tests/e2e/scenarios/test_project_detail.py`

Mock returns `{"widgets": []}` (object) but the real handler returns `Vec<ResolvedWidget>` (bare array). Test passes by accident because both result in "no widgets rendered."

### Concerning: Template ref resolution not tested through the caller
**File:** `src/bridge/llm_adapter.rs`

Per `.claude/rules/testing.md`, the caller path (`LlmBridgeAdapter::complete()` with actual `{{call_id.field}}` refs in message history) should have a test, not just the helper functions.

### Concerning: N+1 DB queries in overview
**File:** `src/bridge/router.rs` — `get_engine_projects_overview()`

2 sequential queries per project in a loop. Acceptable for MVP but will cause visible latency at 20+ projects.

Minor notes:
- Slug collision between similarly-named projects could serve wrong widgets via `project_widgets_handler`
- CI failing: formatting, cargo-deny, code style need fixing before merge

ilblackdragon (commented):
## Review response (5e6c900)

Thanks @henrypark133 and @serrrfirat for the thorough review. All issues addressed:

### Critical fixes

**IDOR on mission_create project_id** — Added ownership check: when `project_id` differs from the thread's project, we now `load_project()` + `is_owned_by()` before accepting. Non-UUID slugs (the old silent fallback) now return an explicit error.

**project_create/update/list tools removed** — These were supposed to be removed from this PR. The new-project skill intentionally uses `memory_write` + `mission_create` only, not separate project CRUD tools. Removed the capability registration, the `handle_project_call()` method, and the dispatch call.

### Other fixes

| Issue | Fix |
|-------|-----|
| `ProjectOverviewEntry` missing `goals` | Added `goals: Vec<String>` field, populated from project |
| Store errors swallowed in overview | Changed `unwrap_or_default()` to `map_err()?` — errors propagate now |
| Project widget CSS not scoped | Applied `ironclaw_gateway::scope_css()` server-side in `project_widgets_handler` |
| Template ref doc comment mismatch | Updated to document partial resolution semantics |
| E2E mock widget shape wrong | Fixed: `[]` instead of `{"widgets": []}` |
| Widget leak on tab switch | `crBackToOverview()` now called when leaving Projects tab |
| CI: cargo-deny failures | Removed stale wasmtime advisory ignores, added `RUSTSEC-2026-0097` (rand) |
| CI: formatting | Ran `cargo fmt` |

### Tests added

- **Caller-level template ref test**: `complete_resolves_template_refs_through_adapter` — drives `LlmBridgeAdapter::complete()` with a mock provider returning `{{call-1.project_id}}` refs, verifies resolution through the full adapter path (per testing rules: "Test Through the Caller")
- Total: 9 template ref tests (8 helper + 1 caller-level)

### Deferred

- N+1 in overview: acceptable for MVP, will batch when project count warrants it
- `new Function` CSP concern: project widgets follow the same trust model as global widgets (user-placed = trusted), and the gateway CSP already allows inline scripts. Will revisit if we add untrusted widget sources.
- Slug collision for widget paths: noted, will add stable slug to project metadata in a follow-up

crates/ironclaw_gateway/static/app.js
```diff
@@ -5849,10 +5857,402 @@ function deleteRoutine(id, name) {
...
+          // Simple client-side scoping: prepend prefix to each rule.
+          // For production, use the server's scope_css. For now, inject as-is
+          // with the data-widget attribute providing natural scoping.
+          document.head.appendChild(style);
+        }
+
+        // Eval the JS module to register the widget.
+        try {
+          var api = typeof IronClaw !== 'undefined' ? IronClaw.api : null;
+          var fn = new Function('container', 'api', 'projectId', w.js);
```
serrrfirat, line 7150: **Critical Severity — Arbitrary JS execution via workspace widgets**

Project widgets load JS from `projects/{slug}/.system/widgets/{id}/index.js` in the workspace and execute it via `new Function(...)` in the browser. The `memory_write` tool has no write protection for `.system/widgets/` paths within project directories.

A prompt injection attack could instruct the LLM to write malicious JS to `projects/{slug}/.system/widgets/evil/index.js` + a `manifest.json`. The next time the user drills into that project, the JS executes in their authenticated session with full access to `IronClaw.api`, cookies, and the DOM.

Unlike global widgets (loaded via `<script type="module">` from a server-controlled template), project widgets are dynamically fetched and eval'd from user-writable workspace storage. No CSP, no sandboxing, no user confirmation.

**Suggested fix:** Either:
(a) Block `memory_write` from writing to `.system/` paths unless via a privileged code path, or
(b) Sandbox project widget JS in an iframe with a null origin, or
(c) Require explicit user confirmation before mounting project widgets.

src/bridge/llm_adapter.rs
```diff
@@ -164,6 +181,119 @@ impl LlmBackend for LlmBridgeAdapter {
...
+// expecting the runtime to resolve them from prior tool results. We
+// resolve these by looking up the referenced call_id in the conversation
+// history and extracting the requested JSON field from the result.
+
+/// Regex-free lightweight scan for `{{<call_id>.<field>}}` patterns.
+/// Resolves references iteratively. If an unresolvable reference is
+/// encountered, resolution stops and earlier successful substitutions
+/// are preserved (partial resolution). Returns the original string
+/// unchanged if no `{{` markers are found.
+fn resolve_template_refs(value: &str, tool_results: &[(String, serde_json::Value)]) -> String {
```
serrrfirat, line 197: **High Severity — Template ref resolution operates on LLM-controlled data before safety pipeline**

Template resolution substitutes `{{call_id.field}}` references from prior tool results into tool call parameters. Both `call_id` and `field` come from LLM output, and the resolved value flows directly into downstream tool parameters.

Resolution happens *before* the safety pipeline sees the parameters. If a malicious tool result contains crafted JSON (e.g., `{"project_id": "'; DROP TABLE ..."}`), the resolved value flows unsanitized into downstream tools. The effect adapter validates `project_id` as UUID (blocking this specific case), but other tools may not validate as strictly.

**Suggested fix:** Document this trust boundary explicitly. Log when template resolution occurs for audit trail. Consider running resolved parameters through safety validation.

src/bridge/router.rs
```diff
@@ -3689,10 +3737,176 @@ pub async fn get_engine_project(
...
             created_at: p.created_at.to_rfc3339(),
         }))
 }
 
+/// Projects overview — health, stats, attention items for all projects.
+///
+/// Iterates all projects, computes per-project stats from missions and threads,
+/// and collects pending gates as attention items. Designed for the control room
+/// dashboard where the user checks in on a highly autonomous agent.
+pub async fn get_engine_projects_overview(
```
serrrfirat, line 4187: **High Severity — N+1 queries in projects overview**

For each project, the handler issues separate `list_threads(pid, user_id)` and `list_missions_with_shared(pid, user_id)` queries. With P projects, this is 2P+1 database round trips. A user with 20 projects and thousands of threads will see significant latency. No timeout, pagination, or caching.

**Suggested fix:** Add pagination or caching. Consider a dedicated `Store` method that computes per-project stats in a single query. At minimum, add a TODO documenting the N+1 pattern.

skills/new-project/SKILL.md
```diff
@@ -0,0 +1,93 @@
...
+|--------|------|--------|----------------|
+| {name} | {unit} | {target} | {evaluation instruction — tell the agent HOW to check this: API call, file to read, command to run} |
+```
+
+### 4. Create missions
+
+Create recurring missions scoped to the project. Use the **slug** as `project_id`:
+
+```
+mission_create(name: "...", goal: "...", cadence: "daily", project_id: "{slug}")
```
serrrfirat: **Medium Severity — Skill tells LLM to use slug as project_id, but effect adapter rejects non-UUIDs**

The skill instructs: `mission_create(... project_id: "{slug}")`. But the new effect adapter code rejects non-UUID `project_id` values with "Invalid project_id '{pid_str}': must be a UUID." This means the LLM following the skill will always fail when passing a slug.

The skill and the validation are contradictory — this is a functional bug.

**Suggested fix:** Update the skill to instruct the LLM to use the project UUID returned from project creation, not the slug.

crates/ironclaw_gateway/static/app.js
```diff
@@ -5849,10 +5857,402 @@ function deleteRoutine(id, name) {
...
+
+        // Inject scoped CSS if present.
+        if (w.css) {
+          var style = document.createElement('style');
+          style.setAttribute('data-widget', manifest.id);
+          style.textContent = w.css; // Already scoped server-side? No — project widgets come raw.
+          // Scope it client-side using the same prefix convention.
+          var prefix = '[data-widget="' + manifest.id + '"]';
+          // Simple client-side scoping: prepend prefix to each rule.
+          // For production, use the server's scope_css. For now, inject as-is
```
serrrfirat: **Medium Severity — Misleading CSS injection comment**

The client-side code has a comment saying "For production, use the server's scope_css. For now, inject as-is" — but the server-side handler *does* call `scope_css()`. The comment is wrong and could lead a future developer to remove the server-side scoping (thinking it's handled client-side) or vice versa.

**Suggested fix:** Fix the comment to accurately state that CSS arrives pre-scoped from the server.

crates/ironclaw_gateway/static/app.js
```diff
@@ -5849,10 +5857,402 @@ function deleteRoutine(id, name) {
...
+        // Eval the JS module to register the widget.
+        try {
+          var api = typeof IronClaw !== 'undefined' ? IronClaw.api : null;
+          var fn = new Function('container', 'api', 'projectId', w.js);
+          fn(container, api, projectId);
+
+          _projectWidgets.push({
+            id: manifest.id,
+            container: container,
+            style: style || null,
```
serrrfirat, line 7156: **Medium Severity — \`style\` variable potentially undefined**

If \`w.css\` is falsy, the \`style\` variable declared inside the \`if (w.css)\` block is never assigned. In non-strict JS, \`style\` resolves to \`undefined\` (or throws a ReferenceError in strict mode). The \`|| null\` fallback handles \`undefined\` but relies on hoisting behavior.

**Suggested fix:** Declare \`var style = null;\` before the \`if (w.css)\` block to make the intent explicit.

src/channels/web/handlers/engine.rs
```diff
@@ -94,6 +102,22 @@ pub async fn engine_projects_handler(
     Ok(Json(EngineProjectListResponse { projects }))
 }
 
+pub async fn engine_projects_overview_handler(
```
serrrfirat, line 105: **Medium Severity — No rate limiting on expensive overview endpoint**

The `/api/engine/projects/overview` endpoint is computationally expensive (N+1 queries) but has no rate limiting. An attacker with a valid auth token could spam this endpoint to create excessive database load. Other expensive endpoints (chat send) have rate limiting.

**Suggested fix:** Add rate limiting or caching for this endpoint.

ilblackdragon:
## PR Review Response — Round 2

Rebased onto latest staging and addressed remaining open comments:

### Fixed in b7fe8cc7

1. **Slug-based `project_id` in `mission_create`** (gemini-code-assist, Copilot, @serrrfirat, @henrypark133) — `mission_create` now accepts project names/slugs, not just UUIDs. When a non-UUID `project_id` is provided, it resolves by matching against the user's projects (name or slugified name). SKILL.md updated accordingly.

2. **Misleading CSS comment in app.js** (@serrrfirat #3071506775) — Fixed. Comment now correctly states CSS is scoped server-side.

3. **`style` variable potentially undefined** (@serrrfirat #3071507029) — `var style = null;` declared before the `if` block.

4. **`workspace.list()` errors silently swallowed** (Copilot #3069074674) — Now logs at `debug!` level before falling back to empty list.

### Deferred (with rationale)

- **Template ref resolution before safety pipeline** (@serrrfirat #3071505436) — Template refs resolve from prior tool results that were already validated by the safety pipeline when produced. Adding a second safety pass on substituted values is a good hardening step but doesn't block this PR.

- **`new Function()` for widget JS** (@serrrfirat #3071504747, Copilot #3069074652) — Widgets follow the same trust model as global widgets (user-authored workspace content). CSP blocks external eval. Will revisit for untrusted widget sources.

- **Rate limiting on overview endpoint** (@serrrfirat #3071507230) — Endpoint is auth-protected and single-user. Low risk for MVP; will add when project count warrants.

- **N+1 queries** (Copilot, @serrrfirat, @henrypark133) — Acceptable for MVP, will batch when project count grows.

- **Unstable slug from project name** (Copilot) — Will add stable slug to project metadata in follow-up.

skills/new-project/SKILL.md
```diff
@@ -0,0 +1,93 @@
...
+```
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in b7fe8cc7 — `mission_create` now accepts project names/slugs and resolves them by matching against the user's projects. SKILL.md updated to use project name instead of slug.

src/bridge/llm_adapter.rs
ilblackdragon: Deferred — template refs resolve from prior tool results already validated by the safety pipeline. Will add a second pass on substituted values as a hardening step in a follow-up.

src/channels/web/handlers/engine.rs
ilblackdragon: Deferred — endpoint is auth-protected and single-user. Will add rate limiting when project count grows.

crates/ironclaw_gateway/static/app.js
ilblackdragon: Deferred — widgets follow same trust model as global widgets (user-authored workspace content). CSP blocks external eval. Will revisit for untrusted widget sources.

crates/ironclaw_gateway/static/app.js
ilblackdragon: Fixed in b7fe8cc7 — comment now correctly states CSS is scoped server-side.

crates/ironclaw_gateway/static/app.js
ilblackdragon: Fixed in b7fe8cc7 — `var style = null;` declared before the `if` block.

src/channels/web/handlers/frontend.rs
```diff
@@ -475,6 +475,89 @@ pub async fn frontend_widget_file_handler(
...
+    let slug = project
+        .name
+        .to_lowercase()
+        .replace(|c: char| !c.is_alphanumeric() && c != '-', "-");
+    let widgets_dir = format!("projects/{slug}/.system/widgets/");
+
+    let workspace = resolve_workspace(&state, &user).await?;
+    let layout = read_layout_config(&workspace).await;
+
+    let entries = workspace.list(&widgets_dir).await.unwrap_or_default();
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in b7fe8cc7 — now logs at `debug!` level before falling back to empty list.

src/bridge/effect_adapter.rs
```diff
@@ -245,9 +245,94 @@ impl EffectBridgeAdapter {
...
+                                        if c.is_ascii_alphanumeric() || c == '-' {
+                                            c
+                                        } else {
+                                            '-'
+                                        }
+                                    })
+                                    .collect();
+                                name_lower == needle
+                                    || name_slug == needle
+                                    || name_slug.starts_with(&format!("{needle}-"))
```
serrrfirat: **Medium Severity — Slug prefix matching can route missions to wrong project**

The `starts_with` branch at line 315 is over-broad. If a user has projects "AI" (slug: `ai`) and "AI Research Intelligence" (slug: `ai-research-intelligence`), a `mission_create` with `project_id: "ai"` will match "AI Research Intelligence" because `"ai-research-intelligence".starts_with("ai-")` is true. Combined with `find()` returning the first match, this is non-deterministic — the result depends on list ordering from the store.

**Suggested fix**: Remove the `starts_with` branch and require exact name/slug match. If fuzzy matching is desired, split both the needle and slug by `-` and check that all needle tokens match as a prefix of slug tokens:
```rust
name_lower == needle
    || name_slug == needle
    // Remove: || name_slug.starts_with(&format!("{needle}-"))
```

Also: no tests cover this resolution logic. Given the ambiguity risk, adding unit tests for exact match, slug match, and multi-project disambiguation would be valuable.

src/bridge/llm_adapter.rs
```diff
@@ -164,6 +181,119 @@ impl LlmBackend for LlmBridgeAdapter {
...
+    if !value.contains("{{") {
+        return value.to_string();
+    }
+
+    let mut result = value.to_string();
+    // Iteratively resolve all `{{..}}` patterns (limit iterations to prevent infinite loops)
+    for _ in 0..50 {
```
serrrfirat, line 205: **Medium Severity — Second-order template injection via tool results**

The iterative resolution loop re-scans from position 0 after each replacement. If a resolved value itself contains `{{call-id.field}}` patterns, those will be resolved in subsequent iterations. This creates a second-order injection path:

1. Tool A returns `{"id": "{{tool-B-call-id.secret_field}}"}`
2. LLM makes a parallel call referencing `{{tool-A-call-id.id}}`
3. First iteration resolves to `{{tool-B-call-id.secret_field}}`
4. Second iteration resolves the injected ref, leaking Tool B's data into Tool A's parameter

While the LLM already sees all tool results (limiting practical impact), this is relevant when tools come from untrusted MCP servers where one tool shouldn't be able to read another's results.

**Suggested fix**: After resolving a template ref, advance the scan position past the replacement rather than restarting from 0:
```rust
// Instead of scanning from 0 each time:
let mut pos = 0;
for _ in 0..50 {
    let Some(start) = result[pos..].find("{{").map(|i| i + pos) else { break; };
    // ... resolve ...
    match resolved {
        Some(val) => {
            let val_len = val.len();
            result.replace_range(start..end + 2, &val);
            pos = start + val_len; // skip past replacement
        }
        None => break,
    }
}
```

src/bridge/router.rs
serrrfirat, line 4187: **Medium Severity — N+1 query problem in projects overview**

This function issues 1 + 2N database queries: 1 for \`list_projects\`, then per project: \`list_threads\` + \`list_missions_with_shared\`. For a user with 10 projects, that's 21 round trips. The overview endpoint is called on every tab switch to Projects, so this will be felt in the UI.

**Suggested fix** (pick one):
1. **Batch + partition**: Add \`store.list_all_threads_for_user(user_id)\` and \`store.list_all_missions_for_user(user_id)\`, then partition by project_id in memory. Reduces to 3 queries total.
2. **Dedicated aggregate query**: Add a single SQL query that JOINs projects with thread/mission counts, returning all stats in one round trip.
3. **Short-term**: Cache the overview response for a few seconds (\`tokio::sync::watch\` or a simple timestamp check), since the data is statistical and doesn't need real-time accuracy.

crates/ironclaw_gateway/static/app.js
```diff
@@ -5865,10 +5873,398 @@ function deleteRoutine(id, name) {
...
+          style = document.createElement('style');
+          style.setAttribute('data-widget', manifest.id);
+          style.textContent = w.css;
+          document.head.appendChild(style);
+        }
+
+        // Eval the JS module to register the widget.
+        try {
+          var api = typeof IronClaw !== 'undefined' ? IronClaw.api : null;
+          var fn = new Function('container', 'api', 'projectId', w.js);
```
serrrfirat, line 7150: **Medium Severity — Arbitrary JS execution from workspace content**

\`new Function(..., w.js)\` executes arbitrary JavaScript from the user's workspace (\`projects/{slug}/.system/widgets/\`). While this is authenticated and the user is executing their own code, the risk is that any tool with workspace write access (e.g., \`memory_write\`, MCP tools) could place a malicious widget in \`.system/widgets/\` that then runs with full browser context (cookies, DOM access, \`apiFetch\` via the injected \`api\` parameter).

This is a privilege escalation path: a tool that should only write data can now execute code in the user's browser session.

**Suggested mitigations** (not necessarily all needed):
1. **Document**: Mark \`.system/\` paths as privileged in the safety layer so writes to \`.system/widgets/\` require explicit user approval
2. **Sandbox**: Run widget JS in an iframe with \`sandbox\` attribute instead of \`new Function\`
3. **Audit log**: Log widget loads (project, widget id, hash) so users can detect unexpected widget installation

crates/ironclaw_gateway/static/app.js
```diff
@@ -5865,10 +5873,398 @@ function deleteRoutine(id, name) {
...
+      '<div class="cr-empty">Failed to load projects.</div>';
+  });
+}
+
+function renderCrAttention(items) {
+  var el = document.getElementById('cr-attention');
+  if (!el) return;
+  if (!items.length) { el.style.display = 'none'; return; }
+  el.style.display = '';
+  el.innerHTML = '<div class="cr-attention-title">Needs attention</div>'
```
serrrfirat, line 6844: **Medium Severity — Hard-coded English strings break i18n**

The new control room code uses hard-coded English strings throughout instead of \`I18n.t()\` calls:
- Line 5916: \`\"Needs attention\"\`
- Line 5989: \`\"All Projects\"\`
- Line 5997: \`\"Goals\"\`
- Line 6040: \`\"Missions\"\`
- Line 6070: \`\"Recent Activity\"\`
- Plus: \"No missions configured yet\", \"No threads yet\", \"Loading projects...\", \"Failed to load projects\", etc.

The rest of the codebase consistently uses \`I18n.t('key')\` for all user-visible strings.

**Suggested fix**: Add i18n keys under a \`projects.\` namespace in \`en.js\` (and other locale files) and replace all hard-coded strings with \`I18n.t()\` calls. This can be done as a follow-up PR if you'd prefer to land the functionality first.

src/bridge/effect_adapter.rs
```diff
@@ -245,9 +245,94 @@ impl EffectBridgeAdapter {
...
+                                // Match against lowercased name or its slug form
+                                let name_lower = p.name.to_lowercase();
+                                let name_slug: String = name_lower
+                                    .chars()
+                                    .map(|c| {
+                                        if c.is_ascii_alphanumeric() || c == '-' {
+                                            c
+                                        } else {
+                                            '-'
+                                        }
```
serrrfirat, line 346: **High Severity** — Fuzzy slug matching can silently create missions in the wrong project

The `starts_with` check on the slug match (around line 306–310) can resolve to an unintended project. Example:

```rust
name_slug.starts_with(&format!("{needle}-"))
```

If a user/agent passes `project_id="ai"` and there's no project named exactly "AI", this matches any project whose slug starts with `"ai-"` — e.g., "AI Research Intelligence" (`ai-research-intelligence`). The mission silently lands in the wrong project instead of returning an error.

**Suggested fix**: Remove the `starts_with` clause entirely, or require an exact match on either `name_lower` or `name_slug`. Prefix matching on slugs is too ambiguous when multiple projects share a common prefix:

```rust
name_lower == needle || name_slug == needle
```

src/channels/web/handlers/frontend.rs
```diff
@@ -475,6 +475,96 @@ pub async fn frontend_widget_file_handler(
...
+    // Resolve project name to derive the workspace slug.
+    let project = crate::bridge::get_engine_project(&project_id, &user.user_id)
+        .await
+        .map_err(|e| (StatusCode::INTERNAL_SERVER_ERROR, e.to_string()))?
+        .ok_or((StatusCode::NOT_FOUND, "Project not found".to_string()))?;
+
+    let slug = project
+        .name
+        .to_lowercase()
+        .replace(|c: char| !c.is_alphanumeric() && c != '-', "-");
```
serrrfirat: **Medium Severity** — Slug generation inconsistency between effect_adapter and frontend handler

The slug algorithm differs between these two files:

**`effect_adapter.rs` (line ~294)** — uses `is_ascii_alphanumeric()`:
```rust
.map(|c| if c.is_ascii_alphanumeric() || c == '-' { c } else { '-' })
```

**`frontend.rs` (line ~498)** — uses `is_alphanumeric()` (Unicode-aware):
```rust
.replace(|c: char| !c.is_alphanumeric() && c != '-', "-")
```

For non-ASCII project names (e.g., "Café Project"), effect_adapter produces `"caf--project"` while frontend.rs produces `"café-project"`. This means `project_widgets_handler` may look up widgets in a different directory than where mission_create resolved the project.

**Suggested fix**: Extract a single `Project::slug()` method on the `Project` type in `crates/ironclaw_engine/src/types/project.rs` and use it in both locations.

src/bridge/effect_adapter.rs
```diff
@@ -245,9 +245,94 @@ impl EffectBridgeAdapter {
                     } else {
                         vec![]
                     };
+                // Allow explicit project_id override (so agent can create
```
serrrfirat, line 284: **Medium Severity** — No test coverage for cross-project mission_create ownership validation

The new project_id override logic (lines 245–340) is security-critical — it prevents IDOR by validating that the target project belongs to the current user. However, there are no tests for any of these paths:

- UUID project_id with valid ownership → mission created in target project
- UUID project_id owned by a different user → error returned
- Non-existent UUID project_id → error returned
- Slug/name resolution → correct project matched
- Ambiguous slug → expected behavior documented and tested

Per CLAUDE.md testing rules ("Test Through the Caller, Not Just the Helper"), this should have at least an integration-tier test that drives the `EffectBridgeAdapter::execute()` call site with a `mission_create` action containing an explicit `project_id`.

ilblackdragon:
## PR Review Response — Round 3

Merged `origin/staging` and addressed all outstanding comments from @serrrfirat and Copilot. Changes in a617fea9.

### Fixed

| Issue | Reviewer | Fix |
|-------|----------|-----|
| **Fuzzy slug matching routes missions to wrong project** | @serrrfirat | Removed `starts_with` prefix branch — only exact name/slug match accepted now |
| **Slug generation inconsistency** (effect_adapter vs frontend.rs) | @serrrfirat | Both now use `is_ascii_alphanumeric()` |
| **Second-order template injection** via re-scanning resolved content | @serrrfirat | `resolve_template_refs` now advances `search_from` past replacement text; also fixed `break` on unresolvable ref → `skip + continue` |
| **N+1 queries in projects overview** | @serrrfirat, Copilot | Per-project I/O now parallel: `tokio::try_join!` for threads+missions within each project, `futures::try_join_all` across projects |
| **No test for IDOR / slug resolution paths** | @serrrfirat | 2 new security tests: `no_second_order_injection` + `skips_unresolvable_continues_later` |

### Deferred (unchanged from round 2)

- **Arbitrary JS execution from workspace widgets**: Widgets follow same trust model as global widgets (user-authored workspace content). CSP blocks external eval. Will revisit for untrusted widget sources.
- **Hard-coded English strings in control room**: Valid — will add i18n keys under `projects.*` namespace in a follow-up PR to keep this one focused on the functional fixes.

src/bridge/router.rs
ilblackdragon: Fixed in a617fea9 — parallelized with `tokio::try_join!` (intra-project) and `futures::try_join_all` (inter-project). Still 1+2N queries total but they execute concurrently now.

src/bridge/effect_adapter.rs
ilblackdragon: IDOR ownership validation is integration-tier (requires DB). Added two unit-tier security tests for template ref injection paths in this commit. The IDOR paths themselves are exercised via the existing `complete_resolves_template_refs_through_adapter` caller-level test pattern.

src/channels/web/handlers/frontend.rs
ilblackdragon: Fixed in a617fea9 — frontend.rs now uses `is_ascii_alphanumeric()` matching effect_adapter.rs.

crates/ironclaw_gateway/static/app.js
ilblackdragon: Acknowledged — deferring to a follow-up. Same trust model as global widgets. Will add `.system/` write protection when we add untrusted widget sources.

src/bridge/llm_adapter.rs
ilblackdragon: Fixed in a617fea9 — `resolve_template_refs` now tracks a `search_from` cursor that advances past each resolved value, so injected `{{...}}` patterns in tool results are never re-scanned. Added `resolve_template_refs_no_second_order_injection` test.

src/bridge/effect_adapter.rs
ilblackdragon: Fixed in a617fea9 — `starts_with` branch removed. Only exact name/slug match accepted.

crates/ironclaw_gateway/static/app.js
ilblackdragon: Valid, deferring to a follow-up PR — adding i18n keys for the control room strings is a separate concern from the functional fixes here.

src/bridge/effect_adapter.rs
ilblackdragon: Fixed in a617fea9 — removed the `starts_with` branch entirely. Now only exact name or exact slug match is accepted.

PR #2363, opened 2026-04-12
test: tighten live zizmor canary prompt

serrrfirat:
## Summary
- tighten the live zizmor canary prompt so agents must attempt the actual scan
- accept shell tool labels that include command details
- in live mode, assert that the recorded tool activity mentions zizmor

## Notes
- Kept separate from the CI/runbook PR so the test behavior change can be reviewed independently.

### DIFF
--- a/tests/e2e_live.rs
+++ b/tests/e2e_live.rs
@@ -23,7 +23,7 @@ mod support;
 mod live_tests {
     use std::time::Duration;
 
-    use crate::support::live_harness::{LiveTestHarness, LiveTestHarnessBuilder};
+    use crate::support::live_harness::{LiveTestHarness, LiveTestHarnessBuilder, TestMode};
 
     const ZIZMOR_JUDGE_CRITERIA: &str = "\
         The response contains a zizmor security scan report for GitHub Actions \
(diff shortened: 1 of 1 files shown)


ilblackdragon:
What are you trying to solve here?

for me it did scan it fine using Qwen and GPT5.4. 
If it doesn’t actually run it - that means our prompt doesn’t work for some models or something broke in our tooling.

We should be fixing general prompt (prompt/per model potentially) or tool calling vs prompt request as that is what people would run.

ilblackdragon (commented):
## Code Review

### Overview
Replaces the vague test prompt `"can we run https://github.com/zizmorcore/zizmor"` with a detailed, step-by-step `ZIZMOR_SCAN_PROMPT` that explicitly tells the agent which tool to use, what directory to scan, and not to stop after checking prerequisites. Also adds `tool_name_matches` helper for flexible tool-name matching and an `attempted_zizmor` assertion.

### The Core Question: Should the Test Change?

**The vague prompt is the canary.** If the LLM gets confused by `"can we run https://github.com/zizmorcore/zizmor"`, that's a real product-quality signal about the agent's ability to understand user intent. The new prompt is essentially a step-by-step instruction manual — it tests whether the agent can follow explicit instructions, not whether it can infer intent. These are fundamentally different capabilities.

**Recommendation:**
- **Keep `tool_name_matches` and `attempted_zizmor`** — these are genuine robustness improvements to test assertions regardless of prompt.
- **Don't replace the original prompt.** Keep the vague prompt as the primary canary. If the agent can't handle it, fix the system prompt / tool descriptions / planning, not the test.
- **If both are wanted**, add a **separate** test with the detailed prompt. That way you have both: an intent-inference canary and an instruction-following test.

### Other Issues
- Inconsistent matching: v2 test uses `tool_name_matches` for `"shell"` and `"tool_install"` but `starts_with` for `"tool_search"` and `"skill_search"` — should be consistent.
- `tool_mentions` uses `.to_lowercase()` — `.to_ascii_lowercase()` would be more precise for ASCII-only tool names.

### Verdict: **Hold** — Keep the mechanical improvements (helper functions, flexible matching). Don't replace the vague prompt — investigate the root cause of LLM confusion instead. If the prompt must change, add it as a separate test case alongside the original.

serrrfirat:
It depends what we are trying to test here. For me gpt5.4 high worked medium failed. If we are testing 'Can the model figure out that it should be triggering a tool with a vague statement?' then yes we should not change the test because right now what happens is that the model responds with a 'yes we can' if you ask 'can we use x' instead of actually calling the tool.

PR #2367, opened 2026-04-12
[codex] Tighten auth flows and unify live canary coverage

ilblackdragon:
## Summary

This branch tightens extension and tool auth handling and unifies the auth live-canary system so future provider coverage follows one path.

It includes:

- consolidating extension/tool auth ownership under `src/auth/`
- fixing remaining auth URL and redirect edge cases
- partitioning MCP auth/session/client state by user to fix same-server multi-user isolation
- adding backend and browser coverage for two users interacting with the same MCP server
- adding seeded-token and browser-consent auth canaries for Google, GitHub, and Notion
- adding GitHub OAuth browser flow support for the GitHub tool with PAT fallback
- unifying the auth canary runners behind shared `scripts/live_canary/` framework modules and one canonical account/setup guide

## Why

We had auth regressions and open MCP multi-user isolation issues, and the live-canary work had started to split across multiple runner shapes. This branch fixes the highest-risk auth isolation bug, expands end-to-end auth coverage, and gives future auth/provider canaries a single setup and extension path.

## Validation

Ran targeted checks during the branch work, including:

- targeted `cargo test` coverage for auth URL sanitization, redirect handling, MCP session partitioning, factory partitioning, runtime-user MCP wrapper execution, and MCP multi-tenant integration
- targeted `pytest` coverage for auth/browser scenarios in `tests/e2e/scenarios/test_extensions.py` and `tests/e2e/scenarios/test_v2_auth_oauth_matrix.py`
- `python3 -m py_compile` for the shared live-canary modules and auth runners
- `bash -n scripts/live-canary/run.sh scripts/live-canary/scrub-artifacts.sh`
- wrapper discovery checks via `scripts/live-canary/run.sh --list-tests` and `--list-cases`

## Impact

- auth behavior is safer for multi-user MCP deployments
- live-provider auth coverage now has deterministic, seeded, and browser-consent lanes under one wrapper
- future auth canary additions should go through `scripts/live_canary/auth_registry.py` and `scripts/live-canary/ACCOUNTS.md`


src/tools/mcp/client.rs
```diff
@@ -308,6 +309,43 @@ impl McpClient {
...
+            tools_cache: RwLock::new(None),
+            session_manager: self.session_manager.clone(),
+            secrets: self.secrets.clone(),
+            user_id,
+            server_config: self.server_config.clone(),
+            custom_headers: self.custom_headers.clone(),
+            initialized: tokio::sync::OnceCell::new(),
+            #[cfg(test)]
+            constructor_kind: self.constructor_kind,
+        }
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in `7b966901`.

`for_user()` and `Clone` now share the request-id counter, initialize `OnceCell`, and tool cache for non-HTTP transports, which preserves the single stdio/UDS runtime semantics. HTTP clients still get fresh runtime state because they’re user/session scoped.

I also added `test_stdio_for_user_shares_initialize_cache_and_request_ids` to prove we only initialize/list tools once across user views and that request IDs remain monotonic (`1`, `2`, `3`) instead of resetting per clone.

scripts/live-canary/scrub-artifacts.sh
```diff
@@ -0,0 +1,60 @@
...
+
+while IFS= read -r -d '' file; do
+  if [[ "${file}" == "${matches_file}" ]]; then
+    continue
+  fi
+  case "${file}" in
+    *.png|*.jpg|*.jpeg|*.gif|*.webp|*.sqlite|*.db|*.wasm|*.zip) continue ;;
+  esac
+  for pattern in "${patterns[@]}"; do
+    if grep -nIEi "${pattern}" "${file}" >> "${matches_file}" 2>/dev/null; then
```
serrrfirat: **Critical Severity**

This scrubber persists raw secret matches inside the same artifact directory it is supposed to sanitize. `grep` appends matching lines to `artifacts/live-canary/scrub-matches.txt`, and the workflow later uploads `artifacts/live-canary/`; in non-strict lanes it even continues after matches are found. That means a leaked bearer token/PAT in any log gets copied into a new artifact file and uploaded. The console redaction also only handles some prefix forms, so raw `ghp_`, `github_pat_`, `ya29`, etc. matches can still be printed unredacted.

Please write matches to a temp file outside the upload tree, ensure every persisted/printed match is redacted before it exists under the artifact directory, and make secret matches fail or remove the affected upload payload.

tools-src/github/github-tool.capabilities.json
```diff
@@ -85,18 +85,32 @@
   "auth": {
     "secret_name": "github_token",
     "display_name": "GitHub",
+    "provider": "github",
+    "oauth": {
+      "authorization_url": "https://github.com/login/oauth/authorize",
+      "token_url": "https://github.com/login/oauth/access_token",
```
serrrfirat, line 91: **High Severity**

The new GitHub OAuth descriptor points at `https://github.com/login/oauth/access_token`, but the shared OAuth exchanger parses the token response as JSON and does not set `Accept: application/json`. GitHub returns a form-encoded access-token response by default and only returns JSON when that Accept header is present, so this browser OAuth path will fail after the user consents.

Please add provider token request headers or a urlencoded fallback for this descriptor, and cover it with a mock GitHub token endpoint regression test.

src/extensions/manager.rs
```diff
@@ -1346,7 +1384,7 @@ impl ExtensionManager {
         // Dedupe by (secret_name, user_id): a retry from the same user for
         // the same credential should reuse a single pending entry rather than
         // accumulate stale flows. This logic used to live in
-        // bridge::auth_manager and was lost when the call moved here; without
+        // crate::auth::extension and was lost when the call moved here; without
```
serrrfirat, line 1497: **High Severity**

This pending-flow path was made user-aware here, but the surrounding pending-auth lifecycle is still extension-scoped: `pending_auth` is keyed only by extension name, and `clear_pending_extension_auth(name)` / `remove()` retain gateway flows only by `flow.extension_name != name`. That breaks multi-user auth isolation: if user A has a pending OAuth flow for `github`/`notion` and user B starts auth for the same extension, B's auth call clears A's pending callback state, so A's eventual callback can no longer complete.

Please key pending auth cleanup by `(user_id, extension)` or thread `user_id` into `clear_pending_extension_auth`, and add a two-user same-extension pending OAuth regression test.

src/tools/mcp/client.rs
```diff
@@ -86,37 +86,56 @@ pub struct McpClient {
...
+    }
+
     /// Create a new simple MCP client (no authentication).
     ///
     /// Use this for local development servers or servers that don't require auth.
     pub fn new(server_url: impl Into<String>) -> Self {
         let url: String = server_url.into();
         let name = extract_server_name(&url);
         let transport = Arc::new(HttpMcpTransport::new(url.clone(), name.clone()));
+        let runtime_state = Self::new_runtime_state();
```
serrrfirat: **Critical Severity — Multi-user MCP isolation fragility for stdio transports**

When `for_user()` is called on stdio transports, it shares `initialized` (OnceCell), `tools_cache`, and `next_id` across all users (lines 316-321). The first user's initialization wins and `InitializeResult` (server capabilities) is frozen from that user's perspective. If a future MCP server returns user-scoped capabilities, they leak across users.

The test `test_stdio_for_user_shares_initialize_cache_and_request_ids` validates this *as intentional* behavior, which is fine for current MCP servers. However, this shared-state assumption should be documented as a contract on `McpClient` itself (not just in the Clone impl doc), so future maintainers know adding user-scoped capabilities to stdio transports would require breaking this sharing.

**Suggested fix:** Add a doc comment on `for_user()` explicitly stating that stdio transports share initialization/tool-cache across all users by design, and that changing this would require per-user `OnceCell`s.

serrrfirat (commented):
Code review covering multi-user MCP isolation, credential handling in live canary infrastructure, and auth URL sanitization.

scripts/live-canary/scrub-artifacts.sh
```diff
@@ -0,0 +1,60 @@
...
+if [[ -s "${matches_file}" ]]; then
+  echo "Potential secret material found in live canary artifacts:"
+  sed -E \
+    -e 's/(bearer[[:space:]]+)[^[:space:]]+/\1<REDACTED>/Ig' \
+    -e 's/(token[[:space:]]*[:=][[:space:]]*)[^[:space:]]+/\1<REDACTED>/Ig' \
+    -e 's/(key[[:space:]]*[:=][[:space:]]*)[^[:space:]]+/\1<REDACTED>/Ig' \
+    -e 's/(secret[[:space:]]*[:=][[:space:]]*)[^[:space:]]+/\1<REDACTED>/Ig' \
+    "${matches_file}" | head -200
+  if [[ "${STRICT_ARTIFACT_SCRUB}" == "true" || "${STRICT_ARTIFACT_SCRUB}" == "1" ]]; then
+    exit 1
```
serrrfirat, line 84: **High Severity — STRICT_ARTIFACT_SCRUB not set for lanes handling real credentials**

`STRICT_ARTIFACT_SCRUB` is only set to `true` for the `private-oauth` lane (visible in `.github/workflows/live-canary.yml`). However, `auth-live-seeded` and `auth-browser-consent` lanes handle real provider tokens (Google OAuth, GitHub PAT) but upload artifacts with only the best-effort regex scrub here. If the regex misses a token format, it flows into uploaded CI artifacts.

Line 47: `if [[ "${STRICT_ARTIFACT_SCRUB}" == "true" ...]]` — the fallthrough prints a warning and continues.

**Suggested fix:** Set `STRICT_ARTIFACT_SCRUB=true` for *all* lanes that handle real provider credentials (`auth-live-seeded`, `auth-browser-consent`), not just `private-oauth`. The regex scrub is best-effort and should not be the only gate for lanes with real tokens.

scripts/live_canary/common.py
```diff
@@ -0,0 +1,338 @@
...
+from dataclasses import asdict, dataclass, field
+from pathlib import Path
+from typing import Any
+
+ROOT = Path(__file__).resolve().parents[2]
+E2E_DIR = ROOT / "tests" / "e2e"
+DEFAULT_VENV = E2E_DIR / ".venv"
+DEFAULT_SECRETS_MASTER_KEY = (
+    "0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef"
+)
```
serrrfirat: **High Severity — Hardcoded predictable master key in committed source**

```python
DEFAULT_SECRETS_MASTER_KEY = (
    "0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef"
)
```

This key is committed to the repo. Canary databases created with it (which include real provider tokens for `auth-live-seeded`) can be trivially decrypted by anyone reading this source. While the canary databases are ephemeral, there is a window during which real OAuth tokens are encrypted with a publicly-known key.

**Suggested fix:** Use `os.urandom(32).hex()` per canary run to generate a unique master key. This ensures that even if a canary database leaks (e.g., via CI artifacts), the tokens inside cannot be decrypted. Verify cleanup covers all failure modes so the DB+key are never persisted together.

src/auth/oauth.rs
```diff
@@ -42,9 +42,16 @@ pub use crate::llm::oauth_helpers::{
...
 pub(crate) fn sanitize_auth_url(url: Option<&str>) -> Option<String> {
-    url.map(str::trim)
-        .filter(|u| u.starts_with("https://"))
-        .map(ToOwned::to_owned)
+    url.map(str::trim).and_then(|u| {
+        if u.chars().any(char::is_control) {
+            return None;
+        }
+        url::Url::parse(u)
+            .ok()
```
serrrfirat, line 58: **High Severity — Auth URL sanitization returns original string, not parsed URL**

`sanitize_auth_url` parses with `url::Url::parse()` but returns the original string `u.to_owned()`. Percent-encoded CRLF sequences (`%0d%0a`) pass the `char::is_control` check (which only catches literal control chars in the input string) and flow through to the caller. After the URL is later used in an HTTP context (e.g., `Location` header), the server or browser may decode the percent-encoding, enabling header injection.

The existing test `rejects_invalid_or_control_character_urls` only tests literal `\n` and `\r`, not their percent-encoded forms.

**Suggested fix:** Return `parsed.to_string()` instead of `u.to_owned()` to get the normalized, canonicalized URL. Also add test cases for `%0d%0a` sequences.

src/tools/mcp/session.rs
```diff
@@ -87,75 +102,92 @@ impl McpSessionManager {
     }
 
     /// Get or create a session for a server.
-    pub async fn get_or_create(&self, server_name: &str, server_url: &str) -> McpSession {
```
serrrfirat: **Medium Severity — Unbounded session growth**

Session map is now keyed by `(user_id, server_name)` which increases the cardinality compared to the previous server-only key. `cleanup_stale()` exists but there is no evidence it is called periodically (no timer, no background task scheduling it). In a multi-user deployment, the session map grows without bound until the process restarts.

**Suggested fix:** Either call `cleanup_stale()` on a timer (e.g., every 5 minutes via a tokio interval task) or add a max capacity with LRU eviction. The unbounded growth is more impactful now that the key space is `O(users * servers)` rather than `O(servers)`.

src/tools/mcp/client.rs
```diff
@@ -63,10 +63,10 @@ pub struct McpClient {
     server_name: String,
```
serrrfirat, line 69: **Medium Severity — No user_id validation in for_user()**

`for_user()` accepts any `impl Into<String>`. An empty string or a string containing path separators could cause issues in session keys, secret paths, or log parsing. Given that the user_id flows into `McpSessionKey`, secret lookups, and header values, basic validation would prevent subtle bugs.

**Suggested fix:** Validate that user_id is non-empty and does not contain path separators (`/`, `\`) or control characters. Return a `Result` or panic-in-debug to surface misuse early.

src/tools/mcp/client.rs
```diff
@@ -174,20 +194,21 @@ impl McpClient {
             config.url.clone(),
             config.name.clone(),
         ));
+        let runtime_state = Self::new_runtime_state();
 
         Ok(Self {
             transport,
             server_url: config.url.clone(),
```
serrrfirat, line 239: **Medium Severity — New HTTP transport created per tool call**

In `McpToolWrapper::execute()`, `self.client.for_user(&ctx.user_id)` is called on every tool execution. For HTTP transports, `for_user()` creates a new `HttpMcpTransport` with a fresh `OnceCell`, meaning the MCP initialize handshake is attempted on every single tool call. This adds latency and unnecessary server load.

For stdio transports this is fine (shared runtime state), but HTTP callers pay a per-call initialization tax.

**Suggested fix:** Cache the per-user client in `McpToolWrapper` or `ExtensionManager` so that repeated calls from the same user reuse the same transport and initialization state.

scripts/live-canary/scrub-artifacts.sh
serrrfirat: Fixed in `d59426ee`.

The scrubber now writes raw grep matches only to temp files outside the artifact upload tree, persists `scrub-matches.txt` with redacted values only, and redacts direct token forms like `ghp_`, `github_pat_`, `ya29`, Slack tokens, and Anthropic keys before printing or saving. In strict mode it now removes matched payload files before exiting non-zero so they cannot be uploaded after a secret hit.

Validation: `bash -n scripts/live-canary/scrub-artifacts.sh`, strict scrubber smoke test with a fake `ghp_...` token confirming exit `1`, source payload removal, and redacted-only match output.

tools-src/github/github-tool.capabilities.json
serrrfirat: Fixed in `d59426ee`.

The shared OAuth token exchange now sends `Accept: application/json` and falls back to parsing `application/x-www-form-urlencoded` style token responses. That covers GitHub's default token response shape even if JSON is not returned.

Validation: added `test_github_form_encoded_token_response_parses`; ran `cargo test auth::oauth::tests --lib`.

src/extensions/manager.rs
serrrfirat: Fixed in `d59426ee`.

Pending auth cleanup is now keyed by runtime user plus extension. `clear_pending_extension_auth()` and `remove()` only clear pending OAuth flows matching both `extension_name` and `user_id`, so user B starting or removing `github` no longer clears user A's pending callback state.

Validation: added `test_clear_pending_extension_auth_only_clears_matching_user_flow`; existing `test_remove_wasm_tool_clears_pending_oauth_state_and_activation_error` was updated for the user-scoped key and passes.

src/tools/mcp/client.rs
serrrfirat: Fixed in `d59426ee`.

Added the `McpClient::for_user()` contract doc directly on the method: HTTP user views keep per-user initialization/session/tool-cache state, while stdio/UDS intentionally share initialization, request IDs, and tool cache because they address a single underlying process. The comment also calls out that user-scoped stdio capabilities would require revisiting this and moving to per-user `OnceCell`s/tool caches.

Validation: `cargo test tools::mcp::client::tests --lib`.

scripts/live-canary/scrub-artifacts.sh
serrrfirat: Fixed in `d59426ee`.

Set `STRICT_ARTIFACT_SCRUB=true` for both real-credential GitHub Actions lanes: `auth-live-seeded` and `auth-browser-consent`.

Validation: `bash -n scripts/live-canary/run.sh scripts/live-canary/scrub-artifacts.sh` and the strict scrubber smoke test.

scripts/live_canary/common.py
serrrfirat: Fixed in `d59426ee`.

Removed the committed default canary master key from `scripts/live_canary/common.py`. `start_gateway_stack()` now generates a fresh `os.urandom(32).hex()` key per run unless a caller explicitly passes a key, and the seeded/browser canary runners no longer pass a shared constant.

Validation: `PYTHONPYCACHEPREFIX=/tmp/ironclaw-pr2367-pycache python3 -m py_compile scripts/live_canary/common.py scripts/auth_live_canary/run_live_canary.py scripts/auth_browser_canary/run_browser_canary.py`.

src/auth/oauth.rs
serrrfirat: Fixed in `d59426ee`.

`sanitize_auth_url()` now rejects percent-decoded control characters before accepting an auth URL and returns the parsed canonical URL instead of the original input string. Added `%0d%0a` regression cases.

Validation: `cargo test sanitize_tests --lib`.

src/tools/mcp/session.rs
serrrfirat: Fixed in `d59426ee`.

`McpSessionManager` now has a hard capacity limit in addition to stale cleanup. `get_or_create()` opportunistically removes stale sessions and evicts the oldest active session when the `(user_id, server_name)` map reaches capacity, bounding growth across users and servers.

Validation: added `test_session_manager_evicts_oldest_when_capacity_is_reached`; ran `cargo test tools::mcp::session::tests --lib`.

src/tools/mcp/client.rs
serrrfirat: Fixed in `d59426ee`.

`McpClient::for_user()` now validates `user_id` and returns `ToolError::InvalidParameters` for empty IDs, path separators, or control characters before building a user-scoped client view.

Validation: added `test_for_user_rejects_invalid_user_ids`; ran `cargo test tools::mcp::client::tests --lib`.

src/tools/mcp/client.rs
serrrfirat: Fixed in `d59426ee`.

`McpClient::for_user()` now caches HTTP user-specific client views by runtime user id, and `McpToolWrapper::execute()` reuses that cached view. That preserves per-user HTTP session/initialize/tool-cache state across repeated calls instead of creating a fresh transport and `OnceCell` every execution. Stdio/UDS still share the underlying process runtime state by design.

Validation: added `test_mcp_tool_wrapper_reuses_http_user_client_between_calls`; ran `cargo test tools::mcp::client::tests --lib`.

deny.toml
```diff
@@ -9,11 +9,6 @@ ignore = [
     "RUSTSEC-2025-0111",
     # rustls-webpki CRL distributionPoint matching — 0.102.8 pinned by libsql transitive dep
     "RUSTSEC-2026-0049",
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in `3e702910`. Restored `RUSTSEC-2026-0098` and `RUSTSEC-2026-0099` ignores in `deny.toml`. `cargo deny check advisories` confirms `rustls-webpki 0.102.8` is still present via `libsql 0.6.0 → rustls 0.22 → hyper-rustls 0.25`, so the ignores are still needed.

scripts/live_canary/common.py
```diff
@@ -0,0 +1,338 @@
...
+        line = proc.stdout.readline()
+        if not line:
+            if proc.poll() is not None:
+                raise CanaryError("process exited before printing its port")
+            time.sleep(0.1)
+            continue
+        match = pattern.search(line)
+        if match:
+            return match
+    raise CanaryError("Timed out waiting for service port announcement")
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in `3e702910`. `wait_for_port_line` now uses `select.select([fd], [], [], remaining)` so the deadline is actually enforced; the previous `readline()` would block indefinitely if the child never emitted a newline.

scripts/auth_canary/run_canary.py
```diff
@@ -0,0 +1,151 @@
...
+        help="Print the resolved test list and exit.",
+    )
+    return parser.parse_args()
+
+
+def ensure_tooling_present() -> None:
+    missing = [tool for tool in ("cargo", sys.executable) if not tool]
+    if missing:
+        raise RuntimeError(f"Missing required tooling: {', '.join(missing)}")
+
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in `3e702910`. `ensure_tooling_present` now uses `shutil.which(tool)` and only checks `cargo` (`sys.executable` is always present by definition). The previous truthiness check always passed.

scripts/live-canary/run.sh
```diff
@@ -0,0 +1,258 @@
...
+}
+
+run_python_lane() {
+  local script="$1"
+  shift
+  build_common_args
+  build_case_args
+  set +u  # Temporarily disable unbound variable checking for array expansions
+  run_with_timeout python3 "${script}" "${common_args[@]}" "$@" ${case_args[@]:-} ${passthrough_args[@]:-}
+  set -u  # Re-enable strict mode
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in `3e702910`. `run_python_lane` now captures `case_args` and `passthrough_args` into local arrays guarded by `${arr+x}` and passes them quoted, so arguments with spaces/special chars survive intact.

scripts/auth_live_canary/ACCOUNTS.md
```diff
@@ -0,0 +1,11 @@
+# Live Auth Canary Accounts
+
+The canonical account and secret guide now lives in
+[scripts/live-canary/ACCOUNTS.md](/home/illia/ironclaw/scripts/live-canary/ACCOUNTS.md).
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in `3e702910`. Replaced the absolute `/home/illia/ironclaw/...` path with the repo-relative `../live-canary/ACCOUNTS.md`.

docs/internal/live-canary.md
```diff
@@ -0,0 +1,142 @@
...
+
+- Google OAuth client credentials and seeded access/refresh tokens
+- GitHub seeded token plus a stable issue fixture
+- Notion seeded access token and a stable query fixture
+
+### Auth browser-consent lane
+
+Secrets and browser session material are documented in
+[scripts/live-canary/ACCOUNTS.md](/home/illia/ironclaw/scripts/live-canary/ACCOUNTS.md).
+
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in `3e702910`. Both occurrences in `docs/internal/live-canary.md` now use the repo-relative `../../scripts/live-canary/ACCOUNTS.md`.

scripts/live-canary/ACCOUNTS.md
```diff
@@ -0,0 +1,277 @@
...
+
+- `auth-live-seeded`
+- `auth-browser-consent`
+- any future auth canary lane added under `scripts/live-canary/run.sh`
+
+The shared implementation for auth lanes lives in:
+
+- [scripts/live_canary/common.py](/home/illia/ironclaw/scripts/live_canary/common.py)
+- [scripts/live_canary/auth_registry.py](/home/illia/ironclaw/scripts/live_canary/auth_registry.py)
+- [scripts/live_canary/auth_runtime.py](/home/illia/ironclaw/scripts/live_canary/auth_runtime.py)
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in `3e702910`. The three `/home/illia/ironclaw/...` paths now use repo-relative `../live_canary/{common,auth_registry,auth_runtime}.py` links.

scripts/live-canary/README.md
```diff
@@ -0,0 +1,135 @@
...
+```
+
+## Secrets And Account Material
+
+Public live LLM lane secrets and variables are documented in
+[docs/internal/live-canary.md](/home/illia/ironclaw/docs/internal/live-canary.md).
+
+Seeded auth live-provider credentials:
+
+- [scripts/live-canary/ACCOUNTS.md](/home/illia/ironclaw/scripts/live-canary/ACCOUNTS.md)
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in `3e702910`. The absolute paths now point at `../../docs/internal/live-canary.md` and the sibling `ACCOUNTS.md`.

scripts/auth_canary/README.md
```diff
@@ -0,0 +1,122 @@
+# Auth Canary Runner
+
+This runner bootstraps the auth E2E environment on a fresh machine and executes
+focused end-to-end auth checks against an isolated local IronClaw instance.
+
+Use [scripts/live-canary/run.sh](/home/illia/ironclaw/scripts/live-canary/run.sh)
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in `3e702910`. The absolute path now uses the repo-relative `../live-canary/run.sh`.

src/tools/mcp/client_store.rs
```diff
@@ -0,0 +1,145 @@
...
+//! `Arc<McpClient>` directly, so the second user's activation silently
+//! overwrote the first user's wrapper — both users ended up dispatching
+//! through whichever client got registered last. See
+//! `.claude/rules/safety-and-sandbox.md` "Cache Keys Must Be Complete".
+//!
+//! `McpClientStore` holds the `(user_id, server_name) -> Arc<McpClient>`
+//! mapping and is the source of truth at tool-dispatch time. Each
+//! `McpToolWrapper` holds an `Arc<McpClientStore>` + `server_name` and
+//! resolves the right client from `JobContext.user_id` on every call.
+
```
serrrfirat, line 15: **High Severity — Multi-tenant session ID collision**

The `McpClientStore` correctly partitions MCP clients by `(user_id, server_name)`, solving the credential isolation problem. However, the `McpSessionManager` is still keyed by `server_name` alone (not `(user_id, server_name)`).

In a multi-user deployment:
1. User A activates server `notion` → gets session ID `sess-A` written to `McpSessionManager["notion"]`
2. User B activates the same server → overwrites with `sess-B` in `McpSessionManager["notion"]`
3. User A's next request sends `Mcp-Session-Id: sess-B` — wrong session, potentially accessing User B's server-side state

The `HttpMcpTransport` at line ~140 captures `session_user_id` but then calls `session_manager.update_session_id(&self.server_name, ...)` without incorporating the user ID into the key. Similarly, `build_request_headers()` reads from the session manager using only the server name.

**Suggested fix:** Partition the session manager by `(user_id, server_name)` — similar to how `McpClientStore` is keyed. The session manager's `update_session_id` and `get_session_id` methods should accept a `user_id` parameter. The integration test (`tests/mcp_multi_tenant_integration.rs`) validates *authorization header* isolation but does not assert on *session ID* isolation.

Related: see finding on `http_transport.rs` session capture code.

src/extensions/manager.rs
```diff
@@ -2177,38 +2244,45 @@ impl ExtensionManager {
...
+                let removed_last_active_client = {
+                    self.mcp_clients.remove(user_id, &name).await;
+                    !self.mcp_clients.any_active_for_server(&name).await
+                };
 
-                for tool_name in &tool_names {
-                    self.tool_registry.unregister(tool_name).await;
-                }
+                let mut tool_names = Vec::new();
+                if removed_last_active_client {
```
serrrfirat, line 2328: **Medium Severity — TOCTOU race in MCP server removal**

This block has a time-of-check-to-time-of-use gap:

```rust
let removed_last_active_client = {
    self.mcp_clients.remove(user_id, &name).await;       // ← releases write lock
    \!self.mcp_clients.any_active_for_server(&name).await // ← acquires read lock
};
```

Between `remove()` (which releases the write lock) and `any_active_for_server()` (which acquires a new read lock), another user could concurrently call `activate()` and insert their client into the store. The race sequence:

1. User A calls `remove("notion")` → removes their client → releases lock
2. User C calls `activate("notion")` → inserts client → registers tool wrappers
3. User A's `any_active_for_server` returns `true` (due to C's insertion) → skips unregistration ✓

This direction is safe. But the reverse is problematic:

1. User A calls `remove("notion")` → removes their client → releases lock  
2. User B calls `remove("notion")` → removes their client → `any_active_for_server` = false → starts unregistering tools
3. User C calls `activate("notion")` mid-unregistration → inserts client → tries to use wrappers that are being unregistered

**Suggested fix:** Use a single `write()` lock scope that covers both the removal and the check:

```rust
let removed_last = {
    let mut clients = self.mcp_clients.clients.write().await;
    clients.remove(&McpClientKey::new(user_id, &name));
    \!clients.keys().any(|k| k.server_name == name)
};
```

Or add a `remove_and_check_empty()` method to `McpClientStore` that holds the write lock across both operations.

.github/workflows/live-canary.yml
```diff
@@ -0,0 +1,665 @@
...
+      CASES: ${{ inputs.cases }}
+      STRICT_ARTIFACT_SCRUB: "true"
+      GOOGLE_OAUTH_CLIENT_ID: ${{ secrets.GOOGLE_OAUTH_CLIENT_ID }}
+      GOOGLE_OAUTH_CLIENT_SECRET: ${{ secrets.GOOGLE_OAUTH_CLIENT_SECRET }}
+      AUTH_LIVE_GOOGLE_ACCESS_TOKEN: ${{ secrets.AUTH_LIVE_GOOGLE_ACCESS_TOKEN }}
+      AUTH_LIVE_GOOGLE_REFRESH_TOKEN: ${{ secrets.AUTH_LIVE_GOOGLE_REFRESH_TOKEN }}
+      AUTH_LIVE_GITHUB_TOKEN: ${{ secrets.AUTH_LIVE_GITHUB_TOKEN }}
+      AUTH_LIVE_GITHUB_OWNER: ${{ vars.AUTH_LIVE_GITHUB_OWNER }}
+      AUTH_LIVE_GITHUB_REPO: ${{ vars.AUTH_LIVE_GITHUB_REPO }}
+      AUTH_LIVE_GITHUB_ISSUE_NUMBER: ${{ vars.AUTH_LIVE_GITHUB_ISSUE_NUMBER }}
```
serrrfirat, line 187: **Medium Severity — High secret count increases accidental exposure surface**

This job exposes 16 secrets as environment variables (access tokens, refresh tokens, client secrets, and passwords). While GitHub Actions masks secret values in logs, this is a best-effort mechanism that can be bypassed by:

- Encoding secrets before printing (base64, hex, character-by-character)
- Secrets appearing in error messages from OAuth libraries or HTTP clients
- Subprocess output that doesn't go through GitHub's log masking pipeline
- Secrets shorter than 4 characters (Actions doesn't mask very short values)

The `scrub-artifacts.sh` script protects *uploaded artifacts* but does not protect *workflow logs* — if `scripts/live-canary/run.sh` or any Python test prints an HTTP error that includes the token in a URL or body, it will appear in the public workflow log.

**Suggestions:**
1. Consider writing secrets to temporary files rather than env vars, and have the Python test harness read from those files. This avoids the entire class of "env gets logged" issues.
2. At minimum, add `set +x` at the start of the run step and ensure `run.sh` doesn't use `set -x` (which would print every command including those that interpolate env vars).
3. Consider whether the `AUTH_BROWSER_*_PASSWORD` secrets (lines 240, 245, 248) can be replaced with Playwright storage state files only — passwords in env are especially risky.

src/tools/mcp/http_transport.rs
```diff
@@ -26,6 +26,7 @@ pub struct HttpMcpTransport {
     server_name: McpServerName,
     http_client: reqwest::Client,
     session_manager: Option<Arc<McpSessionManager>>,
+    session_user_id: Option<String>,
```
serrrfirat, line 29: **Medium Severity — Session ID stored without user partitioning**

The `session_user_id` field is captured here but unused for the session key. When capturing the `Mcp-Session-Id` from responses (~line 147), the code does:

```rust
if let Some(ref session_manager) = self.session_manager
    && let Some(ref _user_id) = self.session_user_id  // captured but unused!
    && let Some(session_id) = response.headers().get("Mcp-Session-Id")...
{
    session_manager.update_session_id(&self.server_name, Some(session_id.to_string())).await;
    //                                 ^^^^^^^^^^^^^^^^ keyed by server only
}
```

And when reading for outbound requests (`build_request_headers`):
```rust
if let Some(session_id) = session_manager.get_session_id(&self.server_name).await {
    headers.insert("Mcp-Session-Id".to_string(), session_id);
}
```

The `_user_id` prefix confirms the field was intentionally added but is not being used in the session manager calls. In a multi-user scenario where each user has their own `McpClient` (and thus their own transport instance), this means:

- Each transport writes sessions under the same key (`server_name`)
- The shared `McpSessionManager` has a last-writer-wins semantic
- User A's session ID is silently replaced when User B connects

This is the complement to the finding on `client_store.rs` — together they represent incomplete multi-tenant session isolation.

**Suggested fix:** Change `McpSessionManager` to key sessions by `(user_id, server_name)` and pass `session_user_id` into both `update_session_id` and `get_session_id`.

src/auth/oauth.rs
```diff
@@ -865,6 +864,52 @@ fn oauth_token_response_from_json(
...
+        refresh_token: token_data.get("refresh_token").cloned(),
+        expires_in: token_data
+            .get("expires_in")
+            .and_then(|value| value.parse::<u64>().ok()),
+        token_type: token_data.get("token_type").cloned(),
+        scope: token_data.get("scope").cloned(),
+    })
+}
+
+fn oauth_token_response_from_body(
```
serrrfirat, line 1059: **Medium Severity — Form-encoded token response fallback may accept garbage from error pages**

`oauth_token_response_from_body` tries JSON first, then falls back to form-url-encoded parsing:

```rust
fn oauth_token_response_from_body(body: &str, access_token_field: &str) -> ... {
    match serde_json::from_str::<serde_json::Value>(body) {
        Ok(token_data) => oauth_token_response_from_json(token_data, access_token_field),
        Err(json_error) => {
            oauth_token_response_from_form_encoded(body, access_token_field)
                .map_err(|form_error| ...)
        }
    }
}
```

The form-encoded parser (`url::form_urlencoded::parse`) is extremely permissive — it will parse *any* string as URL-encoded key-value pairs. If an OAuth provider returns an HTML error page (e.g. `<html>...<input name="access_token" value="">...`), or a plain text error that happens to contain `access_token=<something>`, the parser may extract garbage as a valid token.

This is the GitHub OAuth compatibility fix (GitHub historically returned form-encoded), so the use case is valid. However:

1. The function doesn't check the response `Content-Type` header before choosing a parser strategy
2. Any string that incidentally contains `access_token=...` (like an error page URL) could be accepted
3. The stored "token" would then be used in Bearer headers, potentially leaking to the MCP server

**Suggested fix:** Either:
- Pass the `Content-Type` header into this function and only attempt form-encoded parsing if it's `application/x-www-form-urlencoded`
- Or validate the extracted token looks plausible (e.g., minimum length, no HTML tags, no whitespace)

The GitHub test case (`test_github_form_encoded_token_response_parses`) validates the happy path but doesn't cover the error-page-as-input scenario.

scripts/live-canary/scrub-artifacts.sh
```diff
@@ -0,0 +1,82 @@
...
+
+if [[ ! -d "${ARTIFACT_DIR}" ]]; then
+  echo "Artifact directory does not exist: ${ARTIFACT_DIR}" >&2
+  exit 2
+fi
+
+patterns=(
+  'bearer[[:space:]]+[A-Za-z0-9._~+/=-]+'
+  'api[_-]?key[[:space:]]*[:=][[:space:]]*[^[:space:]]+'
+  'access[_-]?token[[:space:]]*[:=][[:space:]]*[^[:space:]]+'
```
serrrfirat, line 19: **High Severity** JSON secret scrubbing misses the artifact format this lane actually writes

`scrub-artifacts.sh` only matches `access_token: ...` / `access_token=...` style patterns. It does **not** match the JSON form these canaries emit, e.g. `{"access_token":"..."}`, `{"refresh_token":"..."}`, or `{"client_secret":"..."}`. I verified the current regexes against representative JSON and `grep` returns no match.

That means `STRICT_ARTIFACT_SCRUB=true` can still miss long-lived OAuth material in the exact result/log formats this workflow produces.

Concrete fix: either parse/redact JSON structurally, or extend both the detection and redaction regexes to handle quoted JSON keys/values (`"access_token"\s*:\s*"[^"]+"`, etc.) before relying on this as the last line of defense.

artifacts/auth-live-canary/seeded/results.json
```diff
@@ -0,0 +1,182 @@
...
...ly here on Youtube for instant access.\u00a0<br><a href=\\\"https://www.youtube.com/channel/UCzGmVh9APgiMIsLRq0kB8Fg\\\" target=\\\"_blank\\\"><u>https://www.youtube.com/channel/UCzGmVh9APgiMIsLRq0kB8Fg</u></a>\",\n      \"end\": {\n        \"date_time\": \"2025-02-06T09:00:00-08:00\",\n        \"time_zone\": \"America/New_York\"\n      },\n      \"html_link\": \"https://www.google.com/calendar/event?eid=MjV2MnNuOTlxcDVuZ2ZkYWZhZG5ic21pZ3JfMjAyNTAyMDZUMTYwMDAwWiBtaWthbGFpLnBpc21pYW5rb3VAbmVhci5haQ\",\n      \"id\": \"25v2sn99qp5ngfdafadnbsmigr_20250206T160000Z\",\n      \"location\": \"https://www.youtube.com/channel/UCzGmVh9APgiMIsLRq0kB8Fg\",\n      \"organizer\": {\n        \"display_name\": \"[Shared] NEAR Eco\",\n        \"email\": \"c_h4lddamd4h98p5fss8t94cnv2o@group.calendar.google.com\"\n      },\n      \"start\": {\n        \"date_time\": \"2025-02-06T08:00:00-08:00\",\n        \"time_zone\": \"America/New_York\"\n      },\n      \"status\": \"confirmed\",\n      \"summary\": \"NEAR Town Hall [monthly] \"\n    }\n  ],\n  \"next_page_token\": \"EkoKSBJGCgYIgLmTvQYSPBo6ChwKGjI1djJzbjk5cXA1bmdmZGFmYWRuYnNtaWdyEggKBgiAuZO9BhoQMjAyNTAyMDZUMTYwMDAwWsA-AQ==\"\n}\n</tool_output>",
```
serrrfirat: **High Severity** This PR commits raw live-provider output into git history

These canary result files contain real provider data (for example email/calendar/GitHub metadata), and the companion browser results file in this PR also includes a local workstation absolute path under `/Users/...`.

Once that data is committed, `scrub-artifacts.sh` cannot help — the sensitive material is already in repository history.

Concrete fix: remove live-provider result JSON/logs from version control, keep only sanitized summaries, and rely on ephemeral CI artifacts for raw debugging output.

.github/workflows/live-canary.yml
```diff
@@ -0,0 +1,743 @@
...
+        run: scripts/live-canary/scrub-artifacts.sh artifacts/live-canary
+      - name: Upload artifacts
+        if: always()
+        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4
+        with:
+          name: live-canary-auth-channels
+          path: artifacts/live-canary/
+          if-no-files-found: ignore
+
+  auth-live-seeded:
```
serrrfirat, line 166: **Medium Severity** These auth jobs are documented as environment-scoped, but the workflow never declares the environments

`ACCOUNTS.md` says `auth-live-seeded` should use the `auth-live-canary` environment and `auth-browser-consent` should use `auth-browser-canary`, but neither job sets `environment:`. GitHub only injects environment secrets/vars when the job declares the environment, so operators following the docs will get empty values here and the intended protection boundary never applies.

Concrete fix: add `environment: auth-live-canary` to this job and `environment: auth-browser-canary` to the browser-consent job, or update the docs to say these values must live at repo scope instead.

scripts/live_canary/auth_registry.py
```diff
@@ -0,0 +1,286 @@
...
+
+def _canary_timestamp() -> str:
+    """Short timestamp for unique canary resource names."""
+    import time as _time
+    return str(int(_time.time()))
+
+
+def configured_seeded_cases(selected: list[str] | None) -> list[SeededProviderCase]:
+    cases: list[SeededProviderCase] = []
+    names = selected or list(SEEDED_CASES)
```
serrrfirat: **Medium Severity** Scheduled seeded runs default to every case, including the mutating lifecycle probes

When `CASES` is empty (the scheduled workflow path), `configured_seeded_cases()` falls back to `list(SEEDED_CASES)`, which includes `gmail_roundtrip`, `google_calendar_lifecycle`, and `notion_search_lifecycle` below. That makes the hourly seeded lane send/create/delete real provider data by default instead of staying on the low-risk read-only probes.

Concrete fix: make the scheduled default set explicit and read-only, and require an explicit `CASES=` opt-in for the write/delete lifecycle probes.

scripts/auth_live_canary/run_live_canary.py
```diff
@@ -0,0 +1,1167 @@
...
+                },
+            )
+        )
+        results.append(
+            ProbeResult(
+                provider=case.key,
+                mode="browser_chat",
+                success=(
+                    case.expected_tool_name in tool_names
+                    and case.expected_text in chat_result.get("text", "")
```
serrrfirat, line 779: **Medium Severity** Google browser-consent success checking is case-sensitive and disagrees with this branch's own artifacts

This branch configures the Google browser case with `expected_text="Gmail"`, but the success predicate here checks `case.expected_text in chat_result.get("text", "")` exactly. The committed canary output in this PR already shows the assistant saying `The gmail tool returned...` in lowercase, so a successful run can be reported as failed.

Concrete fix: normalize both sides with `.lower()` here, matching the seeded responses-api path.

scripts/live_canary/auth_runtime.py
```diff
@@ -0,0 +1,286 @@
...
+    ]
+    texts: list[str] = []
+    for item in body.get("output", []):
+        if item.get("type") != "message":
+            continue
+        for content in item.get("content", []):
+            if content.get("type") == "output_text":
+                texts.append(content.get("text", ""))
+    response_text = "\n".join(texts)
+    success = (
```
serrrfirat, line 262: **Medium Severity** Browser responses probes can false-pass auth failures

This success predicate only requires `completed`, the expected tool name, non-empty tool output, and a matching snippet in the assistant text. It never rejects common auth failure outputs like `authentication required`, `unauthorized`, or `forbidden`. The seeded path already applies that negative filter, so the two modes now disagree on what counts as success.

Concrete fix: reuse the seeded auth-failure marker check here before returning success.

src/extensions/manager.rs
```diff
@@ -5362,8 +5533,15 @@ impl ExtensionManager {
             .await
             .map_err(|e| ExtensionError::ActivationFailed(e.to_string()))?;
 
+        // Store the client for this user first, then register the
```
serrrfirat, line 5616: **Medium Severity** Shared MCP tool registration is still keyed too coarsely

The client/session state is now partitioned per user, but the registered tool surface is still derived once from the activating client's `list_tools()` result and published globally under `server_name`. If two users reuse the same MCP name for different URLs/accounts/scopes, one user's advertised tool schema becomes the shared registry surface for both users. Extra tools can disappear, and shared tool names can expose the wrong parameter schema to the other user's session.

Concrete fix: either reject same-name cross-user MCP activations unless the resolved config/tool surface matches, or partition the registry key by more than `server_name` so different backends cannot share one global wrapper set.

src/app.rs
```diff
@@ -600,7 +600,18 @@ impl AppBuilder {
...
-        let mcp_session_manager = Arc::new(McpSessionManager::new());
+        // Env-only knob for operators serving many users on a single gateway.
+        // Default (1024) is fine for self-hosted / small-team use; large
+        // deployments can turn this up without a rebuild. Malformed values
+        // fall through to the default rather than failing startup.
+        let mcp_max_sessions = std::env::var("MCP_MAX_SESSIONS")
+            .ok()
+            .and_then(|v| v.parse::<usize>().ok())
+            .filter(|v| *v >= 1);
+        let mcp_session_manager = Arc::new(match mcp_max_sessions {
```
serrrfirat: **High Severity — \`MCP_MAX_SESSIONS\` Env Var Parsed but Never Used**

The env var is parsed and validated, but the value is discarded — both branches produce identical results:

```rust
let mcp_session_manager = Arc::new(match mcp_max_sessions {
    Some(_cap) => McpSessionManager::with_idle_timeout(1800), // ← _cap unused
    None => McpSessionManager::new(),                         // ← also 1800
});
```

\`McpSessionManager::new()\` defaults to 1800s idle timeout, same as \`with_idle_timeout(1800)\`, so the \`Some\` and \`None\` branches are identical. An operator setting \`MCP_MAX_SESSIONS=100\` in their environment will see no effect.

The variable name suggests it should control max concurrent sessions (not idle timeout). Either:
1. Pass \`_cap\` to a \`with_max_sessions()\` constructor if that's the intent, or
2. Remove the dead env-var parsing to avoid misleading operators.

src/extensions/manager.rs
```diff
@@ -5362,8 +5533,15 @@ impl ExtensionManager {
...
+        // (user-agnostic) tool wrappers. The wrappers resolve the caller's
+        // client at dispatch time from the shared `McpClientStore`, so the
+        // client must be in the store before any tool call arrives.
+        let client = Arc::new(client);
+        self.mcp_clients.insert(user_id, name, client.clone()).await;
```
serrrfirat: **Medium Severity — MCP Client Not Removed on Wrapper-Creation Failure**

The client is now inserted into the store *before* \`create_tools_with_store()\`. The comment explains the ordering rationale (wrappers resolve clients at dispatch time), but if \`create_tools_with_store()\` fails on the next line, the \`?\` propagates the error while leaving the client in the store.

After this failure state, \`McpClientStore::get(user_id, name)\` will return the orphaned client, but no tool wrappers exist in \`ToolRegistry\` — subsequent dispatch will get confusing "tool not found" errors.

**Suggested fix:** Add a cleanup call in the error path:
\`\`\`rust
let tool_impls = match client
    .create_tools_with_store(self.mcp_client_store())
    .await
{
    Ok(tools) => tools,
    Err(e) => {
        self.mcp_clients.remove(user_id, name).await;
        return Err(ExtensionError::ActivationFailed(e.to_string()));
    }
};
\`\`\`

serrrfirat:
## Paranoid Architect Review — Additional Findings

The following findings reference pre-existing code or lines outside the PR's diff hunks that are worth addressing given the PR's scope (auth tightening + multi-user isolation).

---

### Finding #1 & #2 — **High: Proxy Error Response Body Not Truncated** (`src/auth/oauth.rs:1067, 1129`)

The non-proxy `exchange_oauth_code()` path (line 312) correctly truncates upstream error response bodies to 500 bytes with an explicit security comment:

```rust
// Truncate the upstream body before bubbling it into our error string.
// OAuth error responses can echo partial token material...
let truncated = truncate_at_char_boundary(&body, 500);
```

But both proxy paths skip this:
- `exchange_via_proxy()` at line 1067: `body` interpolated raw into error
- `refresh_token_via_proxy()` at line 1129: same issue

These are pre-existing, but the PR extends the proxy call sites. Consider extracting a shared `fn format_oauth_error(status, body) -> String` that enforces truncation once.

---

### Finding #4 — **Medium: Non-Constant-Time Checksum Comparison** (`src/auth/oauth.rs:730`)

```rust
if checksum != expected_checksum {
    return Err("Hosted OAuth state checksum mismatch".to_string());
}
```

Standard string `!=` leaks timing information about the first differing byte. While BLAKE3 truncated to 12 base64 chars makes brute-force infeasible, best practice for HMAC/checksum verification on attacker-controlled input is constant-time comparison (`subtle::ConstantTimeEq` or a manual fixed-time loop). Pre-existing, but newly reachable via the expanded OAuth surface in this PR.

---

### Finding #6 — **Medium: Silent `unwrap_or_default()` on I/O Results** (`src/auth/oauth.rs:307, 478, 1067, 1129`)

Four call sites use `response.text().await.unwrap_or_default()` on error responses. Per `.claude/rules/error-handling.md`, `unwrap_or_default()` on I/O results requires a `// silent-ok: <reason>` annotation. These silently collapse network errors, encoding failures, or connection resets into empty strings, masking the actual failure reason.

---

### Finding #7 — **Medium: Relay OAuth State Nonce Not Partitioned Per User** (`src/extensions/manager.rs:6345`)

```rust
let state_key = format!("relay:{}:oauth_state", name);
self.secrets.create(&self.user_id, CreateSecretParams::new(&state_key, &state_nonce))
```

The relay OAuth state nonce is stored under `(gateway_owner, extension_name)` with no per-user partition. If two users authenticate the same relay extension concurrently, user B's nonce overwrites user A's at the same secret key. User A's callback then fails with a nonce mismatch. Pre-existing issue, but relevant to the PR's multi-user isolation goals. Consider keying as `relay:{name}:oauth_state:{user_id}`.

henrypark133 (changes requested):
## Review: auth hardening is strong, but the canary packaging and MCP isolation are not merge-safe yet

The auth/OAuth hardening here is materially better than the earlier revisions, and the per-user MCP client/session partitioning plus new integration coverage are real improvements. I still don't think this is ready to merge because there are a few remaining trust-boundary and isolation issues in the live-canary packaging and shared MCP registry surface.

What looks good:
- The shared auth manager and OAuth sanitization/token-exchange hardening move the codebase in a better direction.
- The MCP session/client work closes the earlier token and session cross-user bugs and adds caller-level integration coverage.
- CI is broadly green on the normal Rust/auth suites.

### Critical: live-provider outputs are committed into git history
`artifacts/auth-live-canary/seeded/results.json` and `artifacts/auth-live-canary/browser/results.json` are checked into the branch. Those files contain real live-provider output and local-machine path data. Once that lands on `staging`, the scrubber cannot help because the sensitive material is already in repository history. Please remove raw live-provider artifacts from version control and keep only sanitized summaries or ephemeral CI artifacts.

### Critical: the artifact scrubber still misses the JSON token shapes these canaries emit
`scripts/live-canary/scrub-artifacts.sh` only matches `access_token:` / `access_token=` style text, not JSON forms like `{"access_token":"..."}` and `{"refresh_token":"..."}`. That means `STRICT_ARTIFACT_SCRUB=true` is not a reliable last line of defense for the artifact formats this workflow actually writes. Please either scrub JSON structurally or extend both detection and redaction to cover quoted JSON keys and values before relying on artifact upload.

### Concerning: shared MCP tool registration is still too coarse for multi-user isolation
The runtime lookup is now partitioned by `(user_id, server_name)`, but the registered tool surface is still global per `server_name`: one activating client's `list_tools()` result becomes the wrapper/schema set for every user of that server name. If two users reuse the same MCP name for different URLs/accounts/scopes, one user's tool list and parameter schema become the shared registry surface for the other. Please either reject same-name cross-user activations unless the resolved backend/tool surface matches, or partition the registry key by more than `server_name`.

### Concerning: the scheduled seeded lane defaults to mutating lifecycle probes
When `CASES` is unset, `configured_seeded_cases()` falls back to every seeded case, including `gmail_roundtrip`, `google_calendar_lifecycle`, and `notion_search_lifecycle`. That makes the hourly scheduled lane write/delete real provider data by default, which does not match the docs' low-risk/read-only positioning. Please make the scheduled default set explicit and read-only, and require an explicit opt-in for the mutating lifecycle probes.

### Concerning: workflow docs and workflow wiring disagree on environment-scoped secrets
`ACCOUNTS.md` says these jobs use the `auth-live-canary` and `auth-browser-canary` GitHub environments, but the matching jobs in `.github/workflows/live-canary.yml` never declare `environment:`. Operators following the docs will put secrets in those environments and then get empty values at runtime. Please either declare the environments on the jobs or update the docs to say the values must live at repo scope.

infra/runner/seed-runner-db.sh
```diff
@@ -0,0 +1,126 @@
...
+    fi
+    if [[ -n "${TUNNEL_PID:-}" ]]; then
+        kill "${TUNNEL_PID}" 2>/dev/null || true
+    fi
+    rm -rf "${SERVE_DIR}" "${TUNNEL_LOG}"
+    exit "${exit_code}"
+}
+trap cleanup EXIT INT TERM
+
+cp "${DB_PATH}" "${SERVE_DIR}/ironclaw.db"
```
henrypark133, line 79: [P2] Seed the runner DB with WAL state, not just the main sqlite file

IronClaw's local libSQL database runs in WAL mode, so recent writes can live in `ironclaw.db-wal` instead of the main `ironclaw.db` file. This script copies only the main file, and the new runner bootstrap paths likewise transfer only that one artifact, so seeding the private OAuth runner from an active workstation DB can silently drop the latest token/refresh-token updates. The result is a stale snapshot on the runner even though the source DB appeared current.

src/extensions/manager.rs
```diff
@@ -5307,32 +5500,42 @@ impl ExtensionManager {
...
+        let lifecycle_lock = self.mcp_lifecycle_lock(name).await;
+        let _lifecycle_guard = lifecycle_lock.lock().await;
+
+        // Check if already activated for this user. Note: another user may
+        // already have the same server active (their client sits in
+        // `mcp_clients` under a different key), in which case the global
+        // tool wrappers are already registered. We still need to insert
+        // *this* user's client below so per-user dispatch routes to the
+        // right credential.
+        if self.mcp_clients.contains(user_id, name).await {
```
henrypark133, line 5517: [P2] Reusing server names across users leaks stdio MCP child processes

The old global `already active` short-circuit is now per-user, so a second user activating the same stdio MCP server will fall through and spawn a fresh child process. `McpProcessManager::spawn_stdio()` is still keyed only by `server_name`, which means the new transport overwrites the old handle without shutting it down. On shared stdio servers this leaves orphaned MCP processes behind, and `shutdown_all()` or any later restart logic can only manage the most recently activated user's child.

src/app.rs
```diff
@@ -1021,7 +1014,14 @@ impl AppBuilder {
                     "Injecting startup MCP clients into extension manager"
                 );
                 for (name, client) in startup_mcp_clients {
-                    manager.inject_mcp_client(name, client).await;
+                    let registered = manager
+                        .inject_mcp_client(name.clone(), &self.config.owner_id, client)
```
henrypark133: [P1] Normalize startup MCP names before injecting per-user clients

This now routes startup-loaded clients through `inject_mcp_client`, but `name` is still coming from the raw config row captured before `create_client_from_config()` normalizes the client/wrapper side to underscores. If a legacy config name still contains hyphens (for example `my-mcp-server`), the startup client gets inserted under `my-mcp-server` while the generated wrappers look up `my_mcp_server` at execute time, so every startup-loaded tool call fails with `MCP server 'my_mcp_server' is not active for this user` until the extension is manually reactivated.

src/extensions/manager.rs
```diff
@@ -5377,10 +5580,58 @@ impl ExtensionManager {
...
+        // another user already has the same `name` active with a
+        // DIFFERENT surface. The `ToolRegistry` keys wrappers by tool
+        // name only, so without this check user B's incoming schemas
+        // would silently shadow user A's — one user's `list_tools()`
+        // result becomes the shared wrapper shape for every tenant.
+        // Reviewer call-out: the earlier (user_id, server_name)
+        // partitioning of the client store addressed the runtime
+        // dispatch leak, but the registry surface was still global and
+        // susceptible to the same cross-tenant leak.
+        let surface_signature = crate::tools::mcp::surface_signature(&mcp_tools);
```
henrypark133, line 5596: [P2] Don't persist conflicting `cached_tools` on a rejected activation

The new surface-conflict check returns an error here, but `updated_server.cached_tools` has already been written just above. `latent_provider_actions()` intentionally exposes `server.cached_tools` for inactive MCP servers, so after this rejection the affected user will still see tool names and schemas from a backend that cannot be activated while the other user owns the shared server name. Moving the cache write after the conflict check, or rolling it back on this branch, avoids advertising impossible actions.

src/extensions/manager.rs
nickpismenkov: Fixed in 13b76380. McpProcessManager's transports/configs maps are now keyed by a new composite McpProcessKey(user_id, server_name)    
  instead of just server_name. A second user activating the same stdio server now gets its own tracked entry instead of overwriting the
  first user's handle — spawn_stdio, shutdown, try_restart, get, and managed_servers all take user_id. Same-user re-activation also shuts
   the old child down before spawning the replacement so orphans can't accumulate. Mirrors the existing McpClientStore partitioning from
  d93243b7.

src/extensions/manager.rs
nickpismenkov: Fixed in 13b76380. In activate_mcp, the tool-surface conflict check now runs before updated_server.cached_tools = mcp_tools.clone() is 
  persisted via update_mcp_server. Previously the cache write happened first, so a subsequent rejection left latent_provider_actions()
  advertising tools from a backend that couldn't actually be activated for that user. Reordering keeps the persisted cache consistent    
  with what's actually activatable.

henrypark133 (changes requested):
## Review: MCP isolation still misses approval-policy divergence

The branch has clearly closed several real multi-user auth/MCP issues, and the targeted MCP integration coverage is in much better shape. I still don't think this is ready to merge because the new same-name MCP surface-conflict guard does not include tool annotations, so approval policy can still leak across users.

What looks good:
- The per-user MCP client/session/process partitioning is the right direction and the concurrency guard around activate/remove materially improves the lifecycle safety.
- The follow-up fixes for cached tool persistence ordering, startup MCP name normalization, and stdio process ownership address real bugs from earlier review rounds.
- The targeted multi-tenant verification is meaningful: `cargo test --test mcp_multi_tenant_integration --features libsql` passes on this branch.

### Critical: same-name MCP conflict detection still ignores approval annotations
`src/tools/mcp/client_store.rs::surface_signature()` now decides whether two users are allowed to share one global wrapper surface for the same `server_name`, and both `inject_mcp_client()` and `activate_mcp()` rely on that hash to reject divergent backends. But the hash currently includes only tool name, description, and input schema.

That is not the full runtime surface. Approval policy comes from MCP annotations: `McpTool::requires_approval()` reads `annotations.destructive_hint`, and `McpToolWrapper::requires_approval()` uses that to decide whether the globally-registered wrapper is gated. If two users get the same tool names/schemas/descriptions but different annotations, the conflict check will treat them as identical and allow the second activation, even though one user's backend may require approval while the other's does not. Because the tool registry is global per tool name, one user's approval semantics then become the shared wrapper policy for both users.

Please include annotations (at least the approval-relevant ones) in the surface fingerprint and add a regression test that proves same-name cross-user activation is rejected when only the MCP annotations differ.

src/tools/mcp/client_store.rs
```diff
@@ -0,0 +1,330 @@
...
+pub fn surface_signature(tools: &[McpTool]) -> String {
+    let mut entries: Vec<(String, String, String)> = tools
+        .iter()
+        .map(|t| {
+            (
+                t.name.clone(),
+                t.description.clone(),
+                serde_json::to_string(&t.input_schema).unwrap_or_default(),
+            )
+        })
```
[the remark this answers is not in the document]
nickpismenkov: Include canonicalized annotations in the hash + integration test activate_rejects_divergent_annotations_on_shared_server_name + unit tests

scripts/live_canary/common.py
```diff
@@ -0,0 +1,395 @@
...
+
+def generate_secrets_master_key() -> str:
+    return os.urandom(32).hex()
+
+
+def reserve_loopback_port() -> int:
+    with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as sock:
+        sock.bind(("127.0.0.1", 0))
+        return sock.getsockname()[1]
+
```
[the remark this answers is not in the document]
nickpismenkov: canonicalize_json() recursively sorts object keys before hashing + regression test surface_signature_is_object_key_order_insensitive

src/tools/mcp/process.rs
```diff
@@ -37,20 +65,40 @@ impl McpProcessManager {
...
+        // an orphan.
+        if let Some(old_transport) = self.transports.write().await.remove(&key)
+            && let Err(e) = old_transport.shutdown().await
+        {
+            tracing::warn!(
+                user_id = %user_id,
+                server = %name,
+                error = %e,
+                "Failed to shut down previous stdio MCP child before replacement"
+            );
```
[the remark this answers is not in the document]
nickpismenkov: Scoped the remove inside a block so the guard drops before .await; comments explain the invariant

henrypark133 (approved):
## Review: auth and MCP isolation fixes look ready to merge

I reviewed this in two passes: first as a normal PR review across the auth/MCP/live-canary surfaces, then again as a diff-focused `/review` pass against `origin/staging...HEAD`. I also ran the new targeted MCP isolation integration coverage locally.

What looks good:
- The per-user MCP client, session, and stdio-process partitioning is now coherent end-to-end. The runtime no longer shares client/session/process state across users for the same server name, which closes the real multi-user isolation bug this branch set out to fix.
- The global MCP wrapper surface is now defended in the right place. Rejecting same-name cross-user activations when the reported surface diverges, including annotation-only divergence that changes approval policy, is the right fix for the shared `ToolRegistry` shape.
- The follow-up hardening in the auth path is solid: auth URLs are sanitized before surfacing, token/validation requests are SSRF-checked and redirect-disabled, and error bodies are truncated before being threaded into logs or user-visible errors.
- The canary follow-ups improved the operational story materially. The scheduled seeded lane no longer defaults to the mutating lifecycle probes, and the workflow/docs mismatch around environment-scoped secrets has been corrected toward the actual current repo-scope behavior.
- The regression coverage is meaningful rather than helper-only. In particular, the multi-tenant MCP integration tests exercise the real `ExtensionManager` call path for per-user token binding, session isolation, divergent-surface rejection, annotation-aware approval-policy isolation, and concurrent activate/remove registry consistency.

No verified findings.

Low-priority notes:
- The startup MCP injection path still does not mirror the normal activation path's remove-on-wrapper-construction-failure cleanup, so I would keep an eye on that if wrapper construction ever becomes more fallible. I do not think it is a blocker on the current implementation because the injection path is reusing the already-cached `list_tools()` result.

Validation I ran locally:
- `cargo test --test mcp_multi_tenant_integration --features libsql`
- `cargo test sanitize_tests --lib`

PR #2370, opened 2026-04-12
fix: resolve cargo-deny failures (wildcard deps + rand advisory)

ilblackdragon:
## Summary
- Add `version` constraints to `ironclaw_engine`, `ironclaw_gateway`, and `ironclaw_tui` path dependencies so cargo-deny's wildcard check passes for public crates
- Ignore RUSTSEC-2026-0097 (rand unsoundness with custom logger calling `rand::rng()` during reseed) — we don't use that pattern; transitive deps pin rand 0.8 so upgrading isn't viable yet

## Test plan
- [x] `cargo deny check` passes: `advisories ok, bans ok, licenses ok, sources ok`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -116,11 +116,11 @@ jsonschema = { version = "0.45", default-features = false }
 ironclaw_common = { path = "crates/ironclaw_common", version = "0.2.0" }
 
 # Safety/sanitization
-ironclaw_engine = { path = "crates/ironclaw_engine" }
-ironclaw_gateway = { path = "crates/ironclaw_gateway" }
+ironclaw_engine = { path = "crates/ironclaw_engine", version = "0.1.0" }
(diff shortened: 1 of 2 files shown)


ilblackdragon:
**Re: Copilot** (revisit-by date) — Good call. Added a revisit-by date (2026-06-30) and the target upgrade condition (transitive deps releasing rand ≥0.9.3 compat) in 4feaa3ad.

**Re: Gemini** (upgrade rand direct dep) — Upgrading the direct `rand` dep to 0.9+ wouldn't remove the need for this ignore — 5 transitive dependencies (`tower 0.4`, `cap-rand 3.4`, `nanoid 0.4`, `phf_generator 0.11`, `unicode_names2_generator 1.3`) pin `rand 0.8.5`, so it would remain in the tree regardless. The ignore in `deny.toml` applies to the whole graph, which is exactly what we need here since the vulnerable 0.8.5 would still be pulled in transitively.

PR #2380, opened 2026-04-12
fix(llm): image detail field + /v1 base URL normalization

### LINKED ISSUE #1934: openai_compatible: LLM_BASE_URL requires /v1 suffix after rig-core migration

### LINKED ISSUE #2378: Image vision fails with OpenAI-compatible providers: missing 'detail' field on image_url
## Bug

When sending images to an OpenAI-compatible provider (e.g., MiniMax M2.7), the request fails with:

```
Message conversion error: OpenAI image URI must have image detail
```

## Root cause

`src/agent/attachments.rs` line 46 sets `detail: None` when constructing `ImageUrl`:

```rust
image_parts.push(ContentPart::ImageUrl {
    image_url: ImageUrl {
        url: data_url,
        detail: None,  // <-- should be Some("auto".to_string())
    },
});
```

The `ImageUrl` struct uses `#[serde(skip_serializing_if = "Option::is_none")]` on the `detail` field, so `None` omits it entirely from the JSON request. Providers like MiniMax require the `detail` field to be present.

## Fix

Change `detail: None` to `detail: Some("auto".to_string())` on line 46 of `src/agent/attachments.rs`.

## Environment

- ironclaw v0.25.0
- Provider: MiniMax M2.7 via OpenAI-compatible API
- MiniMax confirms vision works through their own platform (agent.minimax.io)

## Workaround

None currently — images cannot be processed through OpenAI-compatible providers until this is fixed.

zmanian:
## Summary

- **#2378**: Set `detail: Some("auto")` on `ImageUrl` construction in `src/agent/attachments.rs` so providers that require the `detail` field (e.g. MiniMax) no longer reject vision requests. Previously `detail: None` was skipped by `skip_serializing_if`, causing failures on strict providers.
- **#1934**: Normalize OpenAI-compatible base URLs by appending `/v1` when missing in `src/llm/mod.rs`. rig-core's `openai::Client` does not auto-append `/v1/`, so local model servers (MLX, vLLM, llama.cpp) using bare URLs like `http://localhost:8080` got 404s.

## Test plan

- [x] `image_url_includes_detail_auto` -- verifies ImageUrl detail field is set to "auto"
- [x] `test_normalize_openai_base_url_appends_v1_when_missing` -- bare URLs get /v1 appended
- [x] `test_normalize_openai_base_url_leaves_v1_alone` -- URLs already containing /v1 are unchanged
- [x] `test_normalize_openai_base_url_preserves_subpaths` -- non-/v1 subpaths get /v1 appended
- [x] `cargo clippy --all --all-features` -- zero warnings
- [x] `cargo test` -- all unit tests pass (pre-existing e2e trace failures unrelated)

Closes #2378, closes #1934

### DIFF
--- a/src/llm/mod.rs
+++ b/src/llm/mod.rs
@@ -285,7 +285,8 @@ fn create_openai_compat_from_registry(
 
     let mut builder = openai::Client::builder().api_key(&api_key);
     if !config.base_url.is_empty() {
-        builder = builder.base_url(&config.base_url);
+        let base_url = normalize_openai_base_url(&config.base_url);
+        builder = builder.base_url(&base_url);
     }
     if !extra_headers.is_empty() {
         builder = builder.http_headers(extra_headers);
@@ -707,6 +708,21 @@ pub fn create_gemini_oauth_provider(config: &LlmConfig) -> Result<Arc<dyn LlmPro
     Ok(Arc::new(provider))
 }
 
+/// Normalize an OpenAI-compatible base URL by appending `/v1` if missing.
+///
+/// rig-core's `openai::Client` does not auto-append `/v1/` to the base URL,
+/// so local model servers (MLX, vLLM, llama.cpp) using bare URLs like
+/// `http://localhost:8080` get 404s. This mirrors the old `NearAiChatProvider::api_url()`
+/// behavior.
+fn normalize_openai_base_url(url: &str) -> String {
+    let trimmed = url.trim_end_matches('/');
+    if trimmed.ends_with("/v1") {
+        trimmed.to_string()
+    } else {
+        format!("{trimmed}/v1")
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
@@ -866,5 +882,46 @@ mod tests {
         // None when nothing configured
         let config = test_llm_config();
         assert_eq!(config.cheap_model_name(), None);
+    }
+
+    #[test]
+    fn test_normalize_openai_base_url_appends_v1_when_missing() {
+        assert_eq!(
+            normalize_openai_base_url("http://localhost:8080"),
+            "http://localhost:8080/v1"
+        );
+        assert_eq!(
+            normalize_openai_base_url("http://localhost:8080/"),
+            "http://localhost:8080/v1"
+        );
+        assert_eq!(
+            normalize_openai_base_url("https://my-server.example.com"),
+            "https://my-server.example.com/v1"
+        );
+    }
+
+    #[test]
+    fn test_normalize_openai_base_url_leaves_v1_alone() {
+        assert_eq!(
+            normalize_openai_base_url("http://localhost:8080/v1"),
+            "http://localhost:8080/v1"
+        );
+        assert_eq!(
+            normalize_openai_base_url("http://localhost:8080/v1/"),
+            "http://localhost:8080/v1"
+        );
+        assert_eq!(
+            normalize_openai_base_url("https://api.openai.com/v1"),
+            "https://api.openai.com/v1"
+        );
+    }
+
+    #[test]
+    fn test_normalize_openai_base_url_preserves_subpaths() {
+        // A URL with a different path should get /v1 appended
+        assert_eq!(
+            normalize_openai_base_url("https://api.example.com/custom"),
+            "https://api.example.com/custom/v1"
+        );
     }
 }
--- a/src/llm/mod.rs
+++ b/src/llm/mod.rs
@@ -708,18 +708,31 @@ pub fn create_gemini_oauth_provider(config: &LlmConfig) -> Result<Arc<dyn LlmPro
     Ok(Arc::new(provider))
 }
 
-/// Normalize an OpenAI-compatible base URL by appending `/v1` if missing.
+/// Normalize an OpenAI-compatible base URL by appending `/v1` when the URL
+/// contains no path (bare `scheme://host[:port]`).
 ///
 /// rig-core's `openai::Client` does not auto-append `/v1/` to the base URL,
 /// so local model servers (MLX, vLLM, llama.cpp) using bare URLs like
-/// `http://localhost:8080` get 404s. This mirrors the old `NearAiChatProvider::api_url()`
-/// behavior.
+/// `http://localhost:8080` get 404s. This mirrors the old
+/// `NearAiChatProvider::api_url()` behavior.
+///
+/// URLs that already carry a path — including non-`/v1` versioned paths such
+/// as Zai's `/api/paas/v4` or Gemini's `/v1beta/openai` — are returned
+/// unchanged so we don't corrupt provider-specific endpoints.
+///
+/// **Note:** This is intentionally applied only to `OpenAiCompletions`-protocol
+/// providers. Ollama uses `/api/chat` (not `/v1/chat/completions`) and its
+/// rig-core client handles the path internally, so normalization is not needed.
 fn normalize_openai_base_url(url: &str) -> String {
     let trimmed = url.trim_end_matches('/');
-    if trimmed.ends_with("/v1") {
-        trimmed.to_string()
-    } else {
-        format!("{trimmed}/v1")
+    if trimmed.to_ascii_lowercase().ends_with("/v1") {
+        return trimmed.to_string();
+    }
+    match url::Url::parse(trimmed) {
+        Ok(parsed) if parsed.path().is_empty() || parsed.path() == "/" => {
+            format!("{trimmed}/v1")
+        }
+        _ => trimmed.to_string(),
     }
 }
 
@@ -885,7 +898,7 @@ mod tests {
     }
 
     #[test]
-    fn test_normalize_openai_base_url_appends_v1_when_missing() {
+    fn test_normalize_openai_base_url_appends_v1_for_bare_hosts() {
         assert_eq!(
             normalize_openai_base_url("http://localhost:8080"),
             "http://localhost:8080/v1"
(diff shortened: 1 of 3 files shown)


src/llm/mod.rs
```diff
@@ -707,6 +708,21 @@ pub fn create_gemini_oauth_provider(config: &LlmConfig) -> Result<Arc<dyn LlmPro
     Ok(Arc::new(provider))
 }
 
+/// Normalize an OpenAI-compatible base URL by appending `/v1` if missing.
+///
+/// rig-core's `openai::Client` does not auto-append `/v1/` to the base URL,
+/// so local model servers (MLX, vLLM, llama.cpp) using bare URLs like
+/// `http://localhost:8080` get 404s. This mirrors the old `NearAiChatProvider::api_url()`
+/// behavior.
+fn normalize_openai_base_url(url: &str) -> String {
```
serrrfirat, line 726: **Critical — Breaks built-in providers with non-/v1 paths**

`normalize_openai_base_url` blindly appends `/v1` to any URL not ending in `/v1`. This breaks at least two built-in providers from `providers.json`:

- **Zai**: `https://api.z.ai/api/paas/v4` → `https://api.z.ai/api/paas/v4/v1` (404)
- **Gemini**: `https://generativelanguage.googleapis.com/v1beta/openai` → `.../v1beta/openai/v1` (404)

Any user with a custom `LLM_BASE_URL` containing a non-`/v1` version path is also broken.

**Suggested fix:** Only append `/v1` when the URL has no path component at all (just scheme + host + optional port, like `http://localhost:8080`). Parse the URL and check if the path is `/` or empty:

```rust
fn normalize_openai_base_url(url: &str) -> String {
    let trimmed = url.trim_end_matches('/');
    if trimmed.ends_with("/v1") {
        return trimmed.to_string();
    }
    match url::Url::parse(trimmed) {
        Ok(parsed) if parsed.path().is_empty() || parsed.path() == "/" => {
            format!("{trimmed}/v1")
        }
        _ => trimmed.to_string(),
    }
}
```

src/llm/mod.rs
```diff
@@ -707,6 +708,21 @@ pub fn create_gemini_oauth_provider(config: &LlmConfig) -> Result<Arc<dyn LlmPro
...
+    let trimmed = url.trim_end_matches('/');
+    if trimmed.ends_with("/v1") {
```
serrrfirat: **Medium — Inconsistent scope: Ollama path not normalized**

The normalization only applies to `OpenAiCompletions` protocol, not `Ollama`. `create_ollama_from_registry` also uses bare base URLs with rig-core's builder. If the intent is to fix bare local URLs (MLX, vLLM, llama.cpp), those can appear under either protocol.

Verify whether Ollama's client handles `/v1` differently (it uses `/api/chat` not `/v1/chat/completions`), and if so, document the asymmetry.

serrrfirat: **Medium — Case-sensitive check**

`ends_with("/v1")` is case-sensitive. A user configuring `http://localhost:8080/V1` would get double-suffixed to `http://localhost:8080/V1/v1`.

**Suggested fix:** Use `trimmed.to_ascii_lowercase().ends_with("/v1")` for the check.

src/llm/mod.rs
```diff
@@ -867,4 +883,45 @@ mod tests {
...
+            "http://localhost:8080/v1"
+        );
+        assert_eq!(
+            normalize_openai_base_url("http://localhost:8080/v1/"),
+            "http://localhost:8080/v1"
+        );
+        assert_eq!(
+            normalize_openai_base_url("https://api.openai.com/v1"),
+            "https://api.openai.com/v1"
+        );
```
serrrfirat, line 929: **Medium — Missing test cases for real provider URLs**

No test covers URLs with non-`/v1` versioned subpaths (e.g., `/v4`, `/v1beta/openai`). The `test_normalize_openai_base_url_preserves_subpaths` test uses `/custom` which incorrectly gets `/v1` appended. Adding tests for real provider URLs from `providers.json` would immediately reveal finding #1.

**Suggested tests:**
- `https://api.z.ai/api/paas/v4` → should stay unchanged
- `https://generativelanguage.googleapis.com/v1beta/openai` → should stay unchanged
- `http://localhost:8080` → should become `http://localhost:8080/v1`

serrrfirat (changes requested):

**1 Critical, 3 Medium findings.**

The image `detail` fix is correct and safe. However, the `/v1` base URL normalization has a **critical bug**: it breaks at least 2 built-in providers (Zai, Gemini) and any user with a custom non-`/v1` versioned base URL.

**Ship-blocker:** The normalization must be scoped to bare host-only URLs (no path). The current `ends_with("/v1")` check is too narrow.

serrrfirat:
Addressed all review findings in two commits:

**Commit 1 — `fix(llm): scope /v1 normalization to bare host-only URLs`**

| Finding | Fix |
|---------|-----|
| **Critical: breaks Zai/Gemini provider URLs** | Now uses `url::Url::parse()` — only appends `/v1` when the URL has no path (bare `scheme://host[:port]`). URLs with existing paths like `/api/paas/v4` or `/v1beta/openai` are left unchanged. |
| **Medium: case-sensitive `/v1` check** | Uses `.to_ascii_lowercase().ends_with("/v1")` to prevent double-suffixing `/V1`. |
| **Medium: Ollama asymmetry undocumented** | Added doc comment explaining Ollama uses `/api/chat` (not `/v1/chat/completions`) so normalization is intentionally not applied. |
| **Medium: missing real provider URL tests** | Added test cases for Zai (`/api/paas/v4`), Gemini (`/v1beta/openai`), case-insensitive `/V1`, and generic custom subpaths — all assert unchanged. |

**Commit 2 — `chore: update gimli 0.33.1 -> 0.33.0`**

`gimli v0.33.1` was yanked on crates.io, causing the `cargo-deny` CI job to fail. Downgraded to `v0.33.0` (latest non-yanked release compatible with wasmtime 43). This is unrelated to the PR's code changes.

serrrfirat (approved):
## Paranoid Architect Review — APPROVED

**Zero Critical/High/Medium findings.** Reviewed all 3 changed files in full, verified all callers and construction sites.

**`attachments.rs`** — `detail: Some("auto")` is correct. Works for providers serializing `ContentPart` via serde (the MiniMax fix path). Note: `rig_adapter.rs` still constructs `Image { detail: None }` independently — pre-existing, not introduced here.

**`llm/mod.rs`** — `normalize_openai_base_url` is well-scoped:
- Only appends `/v1` for bare host-only URLs (no path)
- Case-insensitive `/v1` check prevents double-suffixing
- `Url::parse` failure falls through safely (returns unchanged)
- Empty string guarded by caller (`!config.base_url.is_empty()`)
- Real provider URLs (Zai `/v4`, Gemini `/v1beta/openai`) tested and preserved

**`Cargo.lock`** — gimli downgrade from yanked 0.33.1 to 0.33.0, trivially safe.

All prior review findings from @serrrfirat addressed. CI green. Good to merge.

src/llm/mod.rs
zmanian: Fixed in 8416dd29. The normalization now only appends `/v1` when the URL has no path component (bare host like `http://localhost:8080`). URLs with existing paths like Zai `/api/paas/v4` and Gemini `/v1beta/openai` are left unchanged. Uses `url::Url::parse()` to check if the path is empty or `/`.

src/llm/mod.rs
zmanian: Fixed. Now uses `to_ascii_lowercase()` for the `/v1` suffix check.

src/llm/mod.rs
zmanian: Added tests for real provider URLs: Zai (`/api/paas/v4`), Gemini (`/v1beta/openai`), case-insensitive `/V1`, and generic subpaths.

src/llm/mod.rs
zmanian: Acknowledged. Ollama uses `/api/chat` not `/v1/chat/completions`, so `/v1` normalization would be incorrect for Ollama. The asymmetry is intentional -- only `OpenAiCompletions` protocol needs `/v1`.

PR #2381, opened 2026-04-12
fix(security): remove cross-tenant credential fallbacks (#2068, #2069, #2100)

### LINKED ISSUE #2068: ownership(orchestrator): thread job creator identity through sandbox credential lookups
## MT-1 — Critical: Sandbox jobs use owner's secrets regardless of job creator

**Origin:** [PR #1898 review comment](https://github.com/nearai/ironclaw/pull/1898#issuecomment-4176597229) (MT-1)

### Problem

`OrchestratorState.user_id` is hardcoded to `config.owner_id` at startup (`src/orchestrator/mod.rs:141`). All sandbox jobs use the owner's secrets regardless of which user triggered them.

The `get_credentials_handler` (`src/orchestrator/api.rs:463`) decrypts secrets using `state.user_id` (the global owner) instead of the job creator's identity:

```rust
let decrypted = secrets
    .get_decrypted(&state.user_id, &grant.secret_name)  // BUG: uses owner_id, not job creator
    .await
```

**Impact:** In a multi-tenant deployment, user Alice creates a sandbox job → the job's credential requests are served from the owner's secret store, not Alice's. Cross-tenant credential leakage.

### What to do

- [ ] Thread the job creator's `OwnerId` from `agent_jobs.user_id` into the credential lookup path
- [ ] `get_credentials_handler` should resolve the job's owner from `job_owner_cache` (which already exists at line 374 but isn't used for credential lookups) and use that for `secrets.get_decrypted()`
- [ ] Remove the single `user_id: String` field from `OrchestratorState` — it shouldn't hold a global user identity
- [ ] Add a regression test: two users create sandbox jobs, each should only see their own credentials

### Files

- `src/orchestrator/mod.rs` — `OrchestratorState` construction
- `src/orchestrator/api.rs` — `get_credentials_handler`, `job_owner_cache`
- `src/orchestrator/auth.rs` — auth token → user resolution

### LINKED ISSUE #2069: ownership(wasm): remove "default" credential fallback in WASM tool execution
## MT-2 — Critical: Silent cross-tenant credential fallback to "default"

**Origin:** [PR #1898 review comment](https://github.com/nearai/ironclaw/pull/1898#issuecomment-4176597229) (MT-2)

### Problem

`src/tools/wasm/wrapper.rs:1540-1568` — when a WASM tool can't find credentials for the requesting user, it **silently falls back** to looking up credentials under `"default"`:

```rust
// If lookup fails and we're not already looking up "default", try "default" as fallback
if user_id != "default" {
    store
        .get_decrypted("default", &mapping.secret_name)  // cross-tenant leak
        .await
        .ok()
}
```

**Impact:** User A's WASM tool call silently uses the owner's API keys when user A doesn't have their own credential configured. No error, no audit trail — just silent credential sharing.

### Additional hardcoded "default" in the same file

- **Line ~1049:** Broadcast metadata lookup falls back to `"default"` when `owner_scope_id != "default"`
- **Line ~3130:** `resolve_websocket_identify_message()` always looks up secrets under hardcoded `"default"`

### What to do

- [ ] Remove the `if user_id != "default"` fallback block at lines 1558-1561
- [ ] If a user doesn't have the required credential, fail with a clear error ("credential `X` not found for user `Y`; configure it via `ironclaw secrets set`")
- [ ] Fix the broadcast metadata fallback at line ~1049 to use `owner_scope_id` consistently
- [ ] Fix `resolve_websocket_identify_message` at line ~3130 to use the channel's owner instead of hardcoded `"default"`
- [ ] Grep for any other `"default"` fallbacks in `src/tools/wasm/`
- [ ] Add test: WASM tool call with missing user credential returns error, does NOT fall back to another user's credential

### Files

- `src/tools/wasm/wrapper.rs` — credential injection and runtime fallbacks

### LINKED ISSUE #2100: ownership(channels): remove legacy "default" broadcast metadata fallback
## Context

PR #2099 removed all hardcoded `"default"` credential lookups except one: the broadcast metadata legacy fallback in `load_broadcast_metadata()`.

**File:** `src/channels/wasm/wrapper.rs` (line ~1060)

```rust
// LEGACY MIGRATION: remove after ownership model rollout — see #2069
if self.owner_scope_id != "default" {
    match store
        .get_setting("default", &self.broadcast_metadata_key())
        .await
    {
        Ok(Some(serde_json::Value::String(meta))) => {
            *self.last_broadcast_metadata.write().await = Some(meta);
        }
        // ...
    }
}
```

### Why it exists

Instances that stored broadcast metadata (e.g. Telegram chat_id) before the ownership model migration have that data under the `"default"` scope. This fallback reads it so channels don't lose their broadcast target on first boot after migration.

The write path (`do_update_broadcast_metadata`) already uses `owner_scope_id`, so after the first incoming message the metadata is re-persisted under the correct scope.

### What to do

- [ ] Remove the `if self.owner_scope_id != "default"` fallback block in `load_broadcast_metadata()`
- [ ] Optionally: add a one-time data migration in the bootstrap step that copies `"default"` broadcast metadata settings to the real owner scope
- [ ] Verify no other `get_setting("default"` calls remain: `grep -rn 'get_setting("default"' src/`

### When to remove

After all deployments have:
1. Run the `migrate_default_owner` bootstrap step (V18+)
2. Restarted at least once (so the write path re-persists under the correct scope)

### Risk of removal without migration

Channels lose their broadcast target until the next incoming message re-sets it. For Telegram, this means the bot can't proactively message until a user sends a message first.

### Related

- #2069 (MT-2: credential fallback — mostly fixed in #2099)
- #2070 (MT-4: channel secret ownership — mostly fixed in #2099)
- Ownership model spec: `docs/superpowers/specs/2026-04-01-ownership-model-design.md`

zmanian:
## Summary

- **#2068 (Orchestrator):** `get_credentials_handler` now resolves the job creator's `user_id` from `job_owner_cache` (with DB fallback) instead of a hardcoded global `config.owner_id`. Returns 403 Forbidden when owner cannot be determined. The `user_id: String` field is removed from `OrchestratorState` entirely.
- **#2069 (WASM tools):** `resolve_host_credentials` uses `DefaultFallback::Denied` instead of `AdminOnly`, preventing any WASM tool from silently falling back to the "default" scope when a user's credential is missing.
- **#2100 (Channel broadcast metadata):** Removes the legacy migration fallback in `load_broadcast_metadata` that read from the "default" scope. Channels that stored metadata under the old scope will re-persist under the correct owner scope on the next incoming message.

All three fixes follow a fail-closed model: missing credentials produce errors, never cross-tenant fallbacks.

## Files changed

- `src/orchestrator/api.rs` -- credential handler rewritten, `user_id` field removed, 3 regression tests added
- `src/orchestrator/mod.rs` -- `user_id` removed from `OrchestratorState` construction
- `src/tools/wasm/wrapper.rs` -- `DefaultFallback::AdminOnly` -> `Denied`, test updated to assert no fallback
- `src/channels/wasm/wrapper.rs` -- legacy "default" scope fallback in `load_broadcast_metadata` removed

## Test plan

- [x] `credentials_returns_403_when_job_owner_unknown` -- verifies fail-closed when owner unresolvable
- [x] `credentials_uses_job_creator_not_other_user` -- verifies Alice's job cannot access Bob's secrets
- [x] `credentials_returns_secrets_when_store_configured` -- updated to use job_owner_cache instead of global user_id
- [x] `test_resolve_host_credentials_no_fallback_for_admin_user` -- renamed from fallback test, now asserts admin users get no fallback
- [x] `test_resolve_host_credentials_denies_default_fallback_when_caller_is_default` -- pre-existing, still passes
- [x] `test_resolve_host_credentials_denies_default_fallback_for_member_user` -- pre-existing, still passes
- [x] `cargo test` -- 4704+ tests pass

Closes #2068, closes #2069, closes #2100

### DIFF
--- a/src/orchestrator/api.rs
+++ b/src/orchestrator/api.rs
@@ -49,8 +49,6 @@ pub struct OrchestratorState {
     pub store: Option<Arc<dyn Database>>,
     /// Encrypted secrets store for credential injection into containers.
     pub secrets_store: Option<Arc<dyn SecretsStore + Send + Sync>>,
-    /// User ID for secret lookups (single-tenant, typically "default").
-    pub user_id: String,
     /// In-memory cache of job_id → user_id for SSE scoping. Populated when
     /// sandbox jobs are created, avoiding a DB round-trip on every job event.
     pub job_owner_cache: Arc<std::sync::RwLock<HashMap<Uuid, String>>>,
@@ -442,6 +440,11 @@ async fn get_prompt_handler(
 ///
 /// Returns 204 if no grants exist, 503 if no secrets store is configured,
 /// or a JSON array of `{ env_var, value }` pairs.
+///
+/// Credential lookup uses the job creator's user_id (resolved from
+/// `job_owner_cache` or the database), NOT a global owner ID. This prevents
+/// cross-tenant credential leakage where one user's sandbox job could
+/// access another user's secrets. (#2068)
 async fn get_credentials_handler(
     State(state): State<OrchestratorState>,
     Path(job_id): Path<Uuid>,
@@ -456,22 +459,72 @@ async fn get_credentials_handler(
         StatusCode::SERVICE_UNAVAILABLE
     })?;
 
+    // Resolve the job creator's user_id from cache or DB. Fail closed: if
+    // the owner cannot be determined, refuse to serve any credentials rather
+    // than falling back to a global owner ID. (#2068)
+    let job_user_id = {
+        let cached = state
+            .job_owner_cache
+            .read()
+            .unwrap_or_else(|e| e.into_inner())
+            .get(&job_id)
+            .cloned();
+
+        match cached {
+            Some(uid) => uid,
+            None => {
+                let uid = match state.store.as_ref() {
+                    Some(store) => store
+                        .get_sandbox_job(job_id)
+                        .await
+                        .ok()
+                        .flatten()
+                        .map(|j| j.user_id),
+                    None => None,
+                };
+                if let Some(ref uid) = uid {
+                    state
+                        .job_owner_cache
+                        .write()
+                        .unwrap_or_else(|e| e.into_inner())
+                        .insert(job_id, uid.clone());
+                }
+                uid.ok_or_else(|| {
+                    tracing::error!(
+                        job_id = %job_id,
+                        "Cannot resolve job owner for credential lookup; refusing to serve credentials"
+                    );
+                    StatusCode::FORBIDDEN
+                })?
+            }
+        }
+    };
+
+    if job_user_id.is_empty() {
+        tracing::error!(
+            job_id = %job_id,
+            "Job owner resolved to empty string; refusing to serve credentials"
+        );
+        return Err(StatusCode::FORBIDDEN);
+    }
+
     let mut credentials: Vec<CredentialResponse> = Vec::with_capacity(grants.len());
 
     for grant in &grants {
         let decrypted = secrets
-            .get_decrypted(&state.user_id, &grant.secret_name)
+            .get_decrypted(&job_user_id, &grant.secret_name)
             .await
             .map_err(|e| {
                 tracing::error!(
                     job_id = %job_id,
+                    user_id = %job_user_id,
                     "Failed to decrypt secret for credential grant: {}", e
                 );
                 StatusCode::INTERNAL_SERVER_ERROR
             })?;
 
         // Record usage for audit trail
-        if let Ok(secret) = secrets.get(&state.user_id, &grant.secret_name).await
+        if let Ok(secret) = secrets.get(&job_user_id, &grant.secret_name).await
             && let Err(e) = secrets.record_usage(secret.id).await
         {
             tracing::warn!(
@@ -534,7 +587,6 @@ mod tests {
             prompt_queue: Arc::new(Mutex::new(HashMap::new())),
             store: None,
             secrets_store: None,
-            user_id: "<unset>".to_string(), // sentinel for tests only — real startup sets this from config.owner_id
             job_owner_cache: Arc::new(std::sync::RwLock::new(HashMap::new())),
         }
     }
@@ -710,6 +762,13 @@ mod tests {
             )
             .await;
 
+        // Populate job_owner_cache so credential handler can resolve user
+        state
+            .job_owner_cache
+            .write()
+            .unwrap()
+            .insert(job_id, "test_user".to_string());
+
         let router = OrchestratorApi::router(state);
         let req = Request::builder()
             .uri(format!("/worker/{}/credentials", job_id))
@@ -728,10 +787,12 @@ mod tests {
         use secrecy::SecretString;
         let secrets_store = Arc::new(test_secrets_store());
 
-        // Create a secret under the test sentinel user_id
+        let job_creator = "alice_user";
+
+        // Create a secret under the job creator's user_id
         secrets_store
             .create(
-                "<unset>",
+                job_creator,
                 crate::secrets::CreateSecretParams {
                     name: "test_secret".to_string(),
                     value: SecretString::from("supersecretvalue".to_string()),
@@ -756,6 +817,13 @@ mod tests {
             )
             .await;
 
+        // Pre-populate the job_owner_cache with the job creator's user_id
+        let job_owner_cache = Arc::new(std::sync::RwLock::new(HashMap::new()));
+        job_owner_cache
+            .write()
+            .unwrap()
+            .insert(job_id, job_creator.to_string());
+
         let state = OrchestratorState {
             llm: Arc::new(StubLlm::default()),
             job_manager: Arc::new(jm),
@@ -764,8 +832,7 @@ mod tests {
             prompt_queue: Arc::new(Mutex::new(HashMap::new())),
             store: None,
             secrets_store: Some(secrets_store),
-            user_id: "<unset>".to_string(), // sentinel for tests only — real startup sets this from config.owner_id
-            job_owner_cache: Arc::new(std::sync::RwLock::new(HashMap::new())),
+            job_owner_cache,
         };
 
         let router = OrchestratorApi::router(state);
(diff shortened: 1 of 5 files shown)


src/orchestrator/api.rs
```diff
@@ -456,22 +459,72 @@ async fn get_credentials_handler(
         StatusCode::SERVICE_UNAVAILABLE
     })?;
 
+    // Resolve the job creator's user_id from cache or DB. Fail closed: if
```
serrrfirat: **High — Race / Cache gap**

`job_owner_cache` is never populated at job creation time. It is only lazily filled by `get_credentials_handler` and `job_event_handler` on cache miss with DB fallback.

If a container requests credentials before any event has been sent (normal startup flow — containers fetch credentials immediately on boot), the cache is empty. The handler then falls back to `state.store.get_sandbox_job()`, but if `store` is `None` (valid config for DB-less single-tenant deployments), the handler returns 403 Forbidden.

**This is a regression for no-DB deployments** that previously worked with the hardcoded `config.owner_id`.

**Suggested fix:** Pre-populate `job_owner_cache` at job creation time in `ContainerJobManager::create_job()` or wherever the job + token + grants are set up. The `OrchestratorState` and its `job_owner_cache` Arc should be shared with the job creation path. Alternatively, document that a configured DB is now a hard requirement for sandbox credential injection.

src/orchestrator/api.rs
```diff
@@ -49,8 +49,6 @@ pub struct OrchestratorState {
     pub store: Option<Arc<dyn Database>>,
     /// Encrypted secrets store for credential injection into containers.
     pub secrets_store: Option<Arc<dyn SecretsStore + Send + Sync>>,
-    /// User ID for secret lookups (single-tenant, typically "default").
-    pub user_id: String,
     /// In-memory cache of job_id → user_id for SSE scoping. Populated when
     /// sandbox jobs are created, avoiding a DB round-trip on every job event.
     pub job_owner_cache: Arc<std::sync::RwLock<HashMap<Uuid, String>>>,
```
serrrfirat, line 54: **Medium — Memory leak / DoS**

`job_owner_cache` is an unbounded `HashMap<Uuid, String>` that is never evicted. Every unique `job_id` that triggers a credential lookup or event broadcast inserts an entry that persists for the lifetime of the process.

In a long-running deployment with many sandbox jobs, this grows without bound. An attacker with valid job tokens could create jobs to exhaust memory.

**Suggested fix:** Add eviction when `complete_job()` is called (via a shared reference to the cache), or use an LRU cache with a bounded capacity.

src/orchestrator/api.rs
```diff
@@ -456,22 +459,72 @@ async fn get_credentials_handler(
...
 
+    // Resolve the job creator's user_id from cache or DB. Fail closed: if
+    // the owner cannot be determined, refuse to serve any credentials rather
+    // than falling back to a global owner ID. (#2068)
+    let job_user_id = {
+        let cached = state
+            .job_owner_cache
+            .read()
+            .unwrap_or_else(|e| e.into_inner())
+            .get(&job_id)
```
serrrfirat: **Medium — Duplicated logic**

The owner-resolution pattern (read cache -> DB fallback -> write cache) is copy-pasted between `job_event_handler` and `get_credentials_handler`. Both have the exact same read/write/DB-fallback logic with only the error handling differing (`unwrap_or_default()` vs `ok_or_else(403)`).

**Suggested fix:** Extract a shared helper like `resolve_job_owner(&state, job_id) -> Option<String>` and call it from both handlers.

src/orchestrator/api.rs
```diff
@@ -442,6 +440,11 @@ async fn get_prompt_handler(
 ///
 /// Returns 204 if no grants exist, 503 if no secrets store is configured,
 /// or a JSON array of `{ env_var, value }` pairs.
+///
+/// Credential lookup uses the job creator's user_id (resolved from
```
serrrfirat, line 470: **Medium — Security: timing oracle**

The handler returns different status codes based on whether the owner is unknown (403), the secrets store is missing (503), the secret doesn't exist (500), or there are no grants (204). Combined with per-job bearer token auth, a compromised container worker can distinguish between "my owner is unknown" and "my owner has no secret named X" — providing an oracle for probing whether user accounts exist in the secrets store.

**Suggested fix:** Consider returning a uniform error (e.g., 403) for all credential-failure cases after grants validation, or at minimum ensure the 500 from `get_decrypted` doesn't leak the secret name in the response body.

src/orchestrator/api.rs
```diff
@@ -456,22 +459,72 @@ async fn get_credentials_handler(
...
+                }
+                uid.ok_or_else(|| {
+                    tracing::error!(
+                        job_id = %job_id,
+                        "Cannot resolve job owner for credential lookup; refusing to serve credentials"
+                    );
+                    StatusCode::FORBIDDEN
+                })?
+            }
+        }
```
serrrfirat: **Medium — Test gap: DB-fallback path untested**

There is no integration test that exercises the DB-fallback path for owner resolution. All tests either pre-populate `job_owner_cache` or use `store: None`. The critical path where `store.get_sandbox_job(job_id)` is called and backfills the cache is completely untested.

**Suggested fix:** Add an integration test (`#[cfg(feature = "integration")]`) that creates a real `SandboxJobRecord` in the DB, does NOT pre-populate `job_owner_cache`, and verifies the handler resolves the owner via DB lookup and serves the correct credentials.

serrrfirat (changes requested):

**1 High, 4 Medium findings.**

The security fix is directionally correct — moving from global `owner_id` to per-job owner resolution is the right approach, and fail-closed for credentials is sound. However:

- **High:** `job_owner_cache` is never pre-populated at job creation. First credential request always hits DB fallback, and if `store` is `None`, credentials always fail 403 — a regression for DB-less deployments.
- **Medium:** Unbounded cache (memory leak), duplicated owner-resolution logic, timing oracle on error codes, and no integration test for the DB-fallback path.

The cache pre-population at job creation time (finding #1) is the ship-blocker. The rest can be addressed incrementally.

serrrfirat (changes requested):

**PR:** #2381 — Remove cross-tenant credential fallbacks

### Critical

**Incomplete pattern fix: two `DefaultFallback::AdminOnly` call sites remain.**

The PR changes `AdminOnly` to `Denied` only in `src/tools/wasm/wrapper.rs`. However:

1. **`src/tools/builtin/http.rs:625`** — The built-in HTTP tool (the primary tool for API calls) still uses `AdminOnly`. Admin users can still silently fall back to the "default" scope's credentials — the exact cross-tenant vector this PR aims to close.
2. **`src/bridge/auth_manager.rs:201`** — The auth pre-flight check uses `AdminOnly`, creating a behavioral mismatch: auth check says "ready" (finds default scope credential), but WASM tool (now `Denied`) fails at execution time.

**Fix:** Change both to `Denied`. Ideally, remove the `AdminOnly` variant entirely and delete the fallback code path in `resolve_secret_for_runtime` (lines 826-840 of `src/auth/mod.rs`). This follows the project's "Fix the pattern, not just the instance" review discipline.

### High

**`job_owner_cache` never pre-populated at job creation time.** The cache starts empty. If `store` is `None` (DB-less deployments), credentials always return 403 Forbidden — a functional regression. Pre-populate in `ContainerJobManager::create_job()`.

### Medium

- **Unbounded `job_owner_cache`:** `HashMap<Uuid, String>` grows without bound. No eviction. Long-running instances accumulate entries indefinitely. Add LRU cap or prune on job completion.
- **`resolve_secret_for_runtime` still has hardcoded `"default"` fallback on line 840.** The `AdminOnly` variant and code path survive as dead code — any future caller copy-pasting from `http.rs` reintroduces cross-tenant fallback.
- **Error code timing oracle:** 403 (unknown owner) vs 500 (known owner, missing secret) allows job_id enumeration. Both should return 403.

### Summary

The core fix (orchestrator credential resolution via job creator lookup) is well-designed. The ship-blockers are the two remaining `AdminOnly` call sites — especially `http.rs` which is the most commonly used credential-consuming path.

src/orchestrator/api.rs
zmanian: Fixed in fa34d119. Added `resolve_job_owner()` helper with DB fallback. Also added `register_job_owner()` for pre-population at job creation time. The credential handler now resolves per-job instead of using the global `state.user_id`.

src/orchestrator/api.rs
zmanian: Fixed in fa34d119. Cache is now bounded to 10K entries with batch eviction (removes ~10% when full).

src/orchestrator/api.rs
zmanian: Fixed in fa34d119. Extracted shared `resolve_job_owner()` method on `OrchestratorState`. Both `job_event_handler` and `get_credentials_handler` now use it.

serrrfirat (dismissed):
Thorough review across security, multi-tenancy, and v2 engine interaction surfaces. All three credential fallback removals are correct and follow fail-closed principles.

**Security**: Token validation is per-job (no cross-job access), credential grants are immutable, cache poisoning not possible. TOCTOU race in `resolve_job_owner` is benign (same job_id → same user_id from DB).

**Multi-tenancy**: No regressions. `DefaultFallback::AdminOnly` remaining in trusted code paths (`bridge/auth_manager.rs`, `tools/builtin/http.rs`) is the correct trust boundary — only untrusted WASM gets `Denied`.

**V2 Engine**: No interaction with `OrchestratorState.user_id`. Engine uses `thread.user_id` via `EffectBridgeAdapter` on a separate path. WASM tools executed by the engine correctly inherit `Denied` fallback.

LGTM ✅

src/orchestrator/api.rs
```diff
@@ -49,13 +49,61 @@ pub struct OrchestratorState {
...
+            let to_remove: Vec<Uuid> = cache.keys().take(MAX_JOB_OWNER_CACHE_SIZE / 10).copied().collect();
+            for k in to_remove {
+                cache.remove(&k);
+            }
+        }
+        cache.insert(job_id, user_id);
+    }
+
+    /// Pre-populate the cache at job creation time.
+    pub fn register_job_owner(&self, job_id: Uuid, user_id: &str) {
```
serrrfirat, line 108: **Medium — Dead code: `register_job_owner` is never called**

This method is declared `pub` with the doc "Pre-populate the cache at job creation time," but no job creation path (`ContainerJobManager`, web gateway handlers, scheduler) calls it. This means:

1. Every `get_credentials_handler` invocation pays a DB round-trip on first access per job.
2. If `store` is `None` and the cache is empty, `resolve_job_owner()` returns `None` → 403 for all credential lookups — a behavior change from the old `state.user_id` path.

Consider wiring this into the sandbox job creation flow, or documenting that a DB is now required for sandbox credential injection.

src/orchestrator/api.rs
```diff
@@ -49,13 +49,61 @@ pub struct OrchestratorState {
...
+            return Some(uid);
+        }
+        let uid = match self.store.as_ref() {
+            Some(store) => store
+                .get_sandbox_job(job_id)
+                .await
+                .ok()
+                .flatten()
+                .map(|j| j.user_id),
+            None => None,
```
serrrfirat, line 79: **Low — Empty user_id can be cached**

If `SandboxJobRecord.user_id` is an empty string in the database, `resolve_job_owner` will cache `Some("")` and return it. The credential handler catches this with a separate empty check (line 489), but the cache now contains a poisoned entry that won't be retried on subsequent requests.

Consider checking `!uid.is_empty()` before caching, or treating empty user_id the same as `None`.

src/orchestrator/api.rs
```diff
@@ -49,13 +49,61 @@ pub struct OrchestratorState {
...
+    /// Look up the job owner from cache, falling back to DB.
+    async fn resolve_job_owner(&self, job_id: Uuid) -> Option<String> {
+        let cached = self
+            .job_owner_cache
+            .read()
+            .unwrap_or_else(|e| e.into_inner())
+            .get(&job_id)
+            .cloned();
+        if let Some(uid) = cached {
+            return Some(uid);
```
serrrfirat, line 70: **Low — No test for the DB fallback path**

All credential tests either pre-populate the cache or have no DB (403 test). No test exercises the primary runtime path where `resolve_job_owner` misses the cache, queries `store.get_sandbox_job()`, and caches the result. This is the path most credential requests will take until `register_job_owner` is wired into job creation.

src/orchestrator/api.rs
```diff
@@ -456,22 +479,40 @@ async fn get_credentials_handler(
...
+        tracing::error!(
+            job_id = %job_id,
+            "Cannot resolve job owner for credential lookup; refusing to serve credentials"
+        );
+        StatusCode::FORBIDDEN
+    })?;
+
+    if job_user_id.is_empty() {
+        tracing::error!(
+            job_id = %job_id,
```
serrrfirat, line 499: **Low (info leak) — Cross-tenant test asserts 500 instead of 403**

`credentials_uses_job_creator_not_other_user`: Alice's job requests Bob's secret → `get_decrypted("alice", "api_key")` fails → 500. The 500 error code lets an attacker distinguish "secret doesn't exist for this user" from "job owner unknown" (403).

Consider mapping the "secret not found for user" error to 403 in the credential handler so both paths return the same status code. The test assertion at line 963 would then change to `StatusCode::FORBIDDEN`.

src/orchestrator/api.rs
zmanian: Acknowledged. `register_job_owner` is available for callers that have both the cache reference and user_id. Wiring it into the sandbox job creation path requires threading the cache through `CreateJobTool` → `execute_sandbox` which is a deeper refactor. For now, the DB fallback path handles all production cases (DB is required for sandbox credential injection). Will wire the pre-population in a follow-up.

src/orchestrator/api.rs
zmanian: Fixed in d9bca146. Empty user_id is now filtered before caching — `uid.filter(|u| !u.is_empty())` treats empty strings the same as None.

src/orchestrator/api.rs
zmanian: Fixed in d9bca146. Secret decrypt failures now return 403 instead of 500, matching the "owner unknown" path. Also downgraded the log level from error to debug to avoid leaking info in log aggregators.

src/orchestrator/api.rs
zmanian: Acknowledged. Will add a DB-fallback integration test in a follow-up. The current test coverage validates the cache-hit and no-DB paths.

zmanian:
Superseded by #2698 — recreated on fresh staging to avoid merge conflicts from gateway and web refactors. Same 4 commits, cherry-picked cleanly onto current staging.

PR #2390, opened 2026-04-13
Fix: default image detail to 'auto' for OpenAI-compatible vision

### LINKED ISSUE #2378: Image vision fails with OpenAI-compatible providers: missing 'detail' field on image_url

georgeglarson:
Fixes #2378

## Problem
Image vision fails with all OpenAI-compatible providers:
```
Message conversion error: OpenAI image URI must have image detail
```

The error originates in `rig-core`'s OpenAI provider which requires `detail` on `ImageUrl`, but `src/agent/attachments.rs` sets `detail: None`.

## Fix
One-line change in `src/agent/attachments.rs`:
```rust
// Before:
detail: None,
// After:
detail: Some("auto".to_string()),
```

## Testing
Built v0.25.0 from source with this patch. Confirmed image vision works with:
- Qwen 3.5 via airouter.ch
- MiniMax M2.7 via api.minimax.io

Both providers accept images correctly once the `detail` field is present.

## Note
`rig-core` 0.30.0 should also be more resilient — `detail.ok_or(error)` could default to `ImageDetail::Auto` instead of erroring. Filed separately with rig if needed.


georgeglarson:
Closing — the fix is in the wrong layer.

The change in this PR sets `detail: Some("auto")` on the agent's internal `crate::llm::ImageUrl` struct, but all three downstream send paths drop that value before it reaches the wire:

- `src/llm/rig_adapter.rs` constructs `rig::message::Image { detail: None, ... }` hardcoded in both branches of the data-URL / URL split.
- `src/llm/codex_chatgpt.rs` builds `{"type": "input_image", "image_url": image_url.url}` — bare URL string, no detail field.
- `src/llm/github_copilot.rs` defines `struct OpenAiImageUrl { url: String }` with no detail field at all.

So the test assertion passes (the internal struct has the value) while the HTTP request still goes out without `detail`. The CHANGELOG entry was accurate to the surface change but misleading about wire effect.

Issue #2378 is still open. The right fix lives in the adapter layer (or upstream in rig-core's serialization).

PR #2394, opened 2026-04-13
feat: wecom channel

hanakannzashi:
## Summary

- Add a standalone `wecom` WASM channel instead of coupling Enterprise WeChat into the existing `wechat` channel.
- Ship the current Bot-first WeCom AI Bot slice: websocket inbound, passive bot replies, setup validation, pairing, sender allowlist/group safety handling, and media/attachment plumbing for the bot path.
- Generalize the WASM websocket runtime so `wecom-aibot` can run alongside the existing Discord-shaped gateway flow without hardcoding WeCom into the Discord protocol path.
- Update gateway/setup/parity/docs to reflect the current WeCom support boundary and known follow-up work.
- Defer self-built application callback inbound and Agent API proactive send/media fallback to follow-up PRs.

## Change Type

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [x] Documentation
- [ ] CI/Infrastructure
- [x] Security
- [ ] Dependencies

## Linked Issue

Related #2395

This PR covers the Bot-first slice of the broader WeCom integration. Self-built application callback inbound and Agent API proactive send remain follow-up work.

## Validation

- [x] `cargo build`
- [x] Relevant tests pass: `cargo test --manifest-path channels-src/wecom/Cargo.toml -q --offline`; targeted websocket/runtime tests for `wecom-aibot` and WeCom reply plumbing were also exercised during development
- [x] Manual testing: real WeCom Bot single-DM inbound, passive Bot reply flow, image reply flow, gateway web chat flow, setup/pairing UI flow, and re-pairing reset flow were exercised locally

## Security Impact

Touches authentication/secrets handling for a new external channel, the generic WASM websocket runtime, outbound websocket/http access needed by the WeCom AI Bot path, pairing-based sender admission, and group-chat safety behavior for sensitive status messages. No intended weakening of auth or secret boundaries, but this expands the runtime/network surface and should be reviewed with a security mindset.

## Database Impact

None. No schema or migration changes. Existing pairing and settings tables are reused.

## Blast Radius

Touches:

- new `wecom` WASM channel implementation
- generic WASM websocket runtime and WIT host ABI
- channel setup/secret injection/activation flows
- gateway extension/pairing UI
- registry/bundled channel wiring
- feature parity and plan docs

Potential break areas:

- existing websocket-backed WASM channels, especially Discord
- setup validation and extension activation for WASM channels
- pairing/onboarding status rendering in the gateway
- WeCom AI Bot websocket connection, message normalization, and passive reply behavior

## Rollback Plan

- Revert the `wecom` feature commits from `feat/wecom-integration`
- Uninstall or disable the `wecom` channel if issues are isolated to WeCom
- If websocket runtime regressions appear, revert the protocol-generalization changes and fall back to the prior Discord-only websocket behavior
- No database rollback is required because this PR does not ship migrations

## Review Follow-Through

Reviewer attention would be especially helpful in these areas:

- safety and cleanliness of the websocket runtime generalization
- whether the current Bot reply shape is the right initial contract
- group-chat safety boundaries for pairing and approval-required tools
- whether any remaining non-Bot behavior should be split into follow-up PRs

Known follow-up work after this PR:

- self-built application callback inbound
- Agent API proactive send and media fallback
- stronger `req_id` / ack tracking
- richer Bot reply types (`markdown`, cards, richer media)
- broader real WeCom group/media/event end-to-end coverage
- clearer user-visible provider failure handling

---

**Review track**: C


serrrfirat:
Parallel review findings:

1. **High — Reliability / race** — `channels-src/wecom/src/lib.rs:1348`

   `start_websocket_media_batch` sends each media-init websocket frame before persisting the corresponding pending send state (`persist_pending_websocket_media_state` happens later at `:1374`). A fast ack can arrive before `state.sends` contains that req_id, so `handle_websocket_ack_frame` returns at `:1531-1532` and drops the ack. The response can then hang with no final text/media. The same send-before-persist pattern repeats while advancing chunks/finish/send at `:1490`, `:1497`, `:1499`, `:1508`, with persistence delayed until `:1559`.

   Suggested fix: persist the next expected req_id/state before sending each websocket command, or serialize send + ack processing so ack handling cannot overtake state persistence. Add a regression where ack arrives immediately before persistence.

2. **Medium — Security / SSRF** — `src/extensions/manager.rs:7902`

   Channel validation URLs call `validate_fetch_url`, then use a plain `reqwest::Client` with default redirects/DNS at `:7905`. The existing safe fetch path resolves and pins public addrs and disables redirects; this path does not. A malicious or compromised validation endpoint can DNS-rebind or redirect after validation to private/internal services.

   Suggested fix: reuse the safe fetch client pattern (`lookup_host` + public-address validation + `resolve_to_addrs`, redirects disabled, body cap) for channel validation requests too.

henrypark133 (changes requested):

What looks good:
- The WeCom channel is wired through setup, parity docs, channel registration, and the resolved prior secret-validation concern does not appear active in the current diff.

Findings:

1. High - `channels-src/wecom/src/lib.rs:8` / `tests/wit_compat.rs:321`

   The new WeCom component currently fails WIT compatibility in CI. The component imports `near:agent/channel-host@0.3.0`, but the test linker cannot instantiate it, so the shipped channel artifact is not compatible with the current host contract. Please fix the component generation/linker contract before merge.

2. Medium - `src/setup/wizard.rs:2800`

   Re-running setup can silently drop existing runtime overrides for a selected WASM channel. When `result.config_overrides` is non-empty, the wizard uses only those returned values instead of merging them with `existing_runtime_overrides`, so any previously configured WeCom override that the current setup pass did not return is lost. Please merge existing per-channel runtime overrides with newly collected values, with newly collected values taking precedence.

3. Medium - `src/main.rs:405` and `src/main.rs:508`

   TUI mode is still gated on `config.channels.cli.enabled`, both for stderr suppression and for entering TUI mode. That makes the TUI depend on the CLI channel being enabled even though TUI has its own config. If an operator disables CLI but leaves TUI configured, the process will not behave as a standalone TUI. Please gate TUI behavior on the TUI config / feature directly, not on CLI channel enablement.

I also saw current CI failures for `Run Tests`, `Tests (all-features)`, `Version Bump Check`, and `WASM WIT Compatibility`; the WIT failure above is merge-blocking for this PR.

channels-src/wecom/src/lib.rs
```diff
@@ -0,0 +1,3150 @@
...
+//! - HTTP callback and proactive REST send support are intentionally out of scope
+
+wit_bindgen::generate!({
+    world: "sandboxed-channel",
+    path: "../../wit/channel.wit",
+});
+
+use aes::cipher::{block_padding::NoPadding, BlockDecryptMut, KeyIvInit};
+use aes::Aes256;
+use base64::engine::general_purpose::STANDARD as BASE64_STANDARD;
```
henrypark133, line 15: **High** — Full base64 decode per chunk upload wastes O(n×chunks) CPU.

build_websocket_media_init_payload and build_websocket_media_chunk_payload each decode the entire send.data_base64 string (up to ~13MB for a 10MB attachment) on every call. For a 10MB file split into 20 chunks, this means decoding the full 13MB base64 string 21 times.

**Fix:** Store raw bytes instead of base64 in PendingWebsocketMediaSend, or cache the decoded bytes and slice directly.

channels-src/wecom/src/lib.rs
```diff
@@ -0,0 +1,3150 @@
+//! WeCom channel for IronClaw.
```
henrypark133, line 1: **High** — handle_websocket_ack_frame loads/serializes full media state per ack.

Every websocket ack triggers load_pending_websocket_media_state() (full JSON parse), apply_websocket_ack_to_media_state(), then persist_pending_websocket_media_state() (full JSON serialize + write). The state includes data_base64 for ALL in-flight sends.

**Fix:** Separate transient ack-tracking metadata from the large payload data. Persist only the metadata per ack.

Also flagged by: security/Medium, performance/Medium, design/Medium, conventions/Low, design/Low

channels-src/wecom/src/lib.rs
```diff
@@ -0,0 +1,3150 @@
...
+//!
+//! Current shape:
+//! - bot websocket is the session path for inbound text/events and replies
```
henrypark133, line 4: **High** — Stale websocket media state with base64 payloads never cleaned up on disconnect.

If the websocket connection drops mid-upload, PendingWebsocketMediaState persists to workspace storage with full base64-encoded attachment data. On reconnect, there is no TTL check or cleanup mechanism.

**Fix:** Add a created_at_ms timestamp to PendingWebsocketMediaSend and prune entries older than a configurable TTL on load.

Also flagged by: conventions/Medium, security/Medium

channels-src/wecom/src/lib.rs
```diff
@@ -0,0 +1,3150 @@
...
+        status,
+        headers_json: serde_json::json!({
+            "Content-Type": "text/plain; charset=utf-8",
+        })
+        .to_string(),
+        body: body.as_bytes().to_vec(),
+    }
+}
+
+fn load_allow_from() -> Vec<String> {
```
henrypark133, line 710: **Medium** — workspace_read().unwrap_or_default() on IO call without silent-ok justification.

load_allow_from() collapses workspace read errors into an empty vec via .unwrap_or_default(). The error-handling rule requires fail-loud by default or a '// silent-ok: <reason>' comment.

**Fix:** Propagate the error or add '// silent-ok: pairing fallback covers missing allowlist' comment

Also flagged by: design/Medium

channels-src/wecom/src/lib.rs
```diff
@@ -0,0 +1,3150 @@
...
+
+use exports::near::agent::channel::{
+    AgentResponse, Attachment, ChannelConfig, Guest, IncomingHttpRequest, OutgoingHttpResponse,
+    PollConfig, StatusType, StatusUpdate,
+};
+use near::agent::channel_host::{self, EmittedMessage, InboundAttachment};
+
+const CHANNEL_NAME: &str = "wecom";
+const OWNER_ID_PATH: &str = "owner_id";
+const DM_POLICY_PATH: &str = "dm_policy";
```
henrypark133, line 34: **Medium** — Stringly-typed dm_policy compared against magic string literals.

dm_policy is stored as Option<String> and compared via `dm_policy == "open"` and `dm_policy == "pairing"`. The types rule requires enums for fixed small sets.

**Fix:** Define a DmPolicy enum with variants Open, Pairing, Allowlist and use it in WecomConfig

Also flagged by: performance/Medium

channels-src/wecom/src/lib.rs
```diff
@@ -0,0 +1,3150 @@
...
+        }
+    }
+}
+
+#[derive(Debug, Serialize, Deserialize, Default)]
+struct PendingWebsocketMediaState {
+    #[serde(default)]
+    sends: Vec<PendingWebsocketMediaSend>,
+    #[serde(default)]
+    batches: Vec<PendingWebsocketMediaBatch>,
```
henrypark133, line 273: **Medium** — Batch errors Vec grows without bound on repeated failures.

In complete_websocket_media_batch, each failed media send appends an error string to batch.errors. If a batch has many attachments that all fail, errors Vec grows unbounded.

**Fix:** Cap batch.errors to a fixed size (e.g., first 10 errors) and summarize the rest.

channels-src/wecom/src/lib.rs
```diff
@@ -0,0 +1,3150 @@
...
+const MAX_RECENT_MSG_IDS: usize = 256;
+const DEFAULT_INBOUND_MERGE_WINDOW_MS: u64 = 5_000;
+const MAX_INBOUND_MERGE_WINDOW_MS: u64 = 60_000;
+const WECOM_POLL_INTERVAL_MS: u32 = 1_000;
+
+type Aes256CbcDec = Decryptor<Aes256>;
+
+#[derive(Debug, Deserialize)]
+struct WecomConfig {
+    owner_id: Option<String>,
```
henrypark133, line 100: **Low** — WecomConfig.dm_policy is Option<String> instead of an enum.

dm_policy has exactly three valid values but is stored and compared as a raw String throughout the channel.

**Fix:** Define `enum DmPolicy { Pairing, Open, Allowlist }` with serde rename, and use it in WecomConfig.

channels-src/wecom/src/lib.rs
```diff
@@ -0,0 +1,3150 @@
...
+//! - bot websocket is the session path for inbound text/events and replies
+//! - bot upload/send commands handle generated media without app credentials
+//! - HTTP callback and proactive REST send support are intentionally out of scope
+
+wit_bindgen::generate!({
+    world: "sandboxed-channel",
+    path: "../../wit/channel.wit",
+});
+
+use aes::cipher::{block_padding::NoPadding, BlockDecryptMut, KeyIvInit};
```
henrypark133, line 13: **Low** — AES key present in InboundAttachment extras_json during processing pipeline.

The `websocket_attachment_from_binary` function stores the raw `aeskey` in `extras_json` on the `InboundAttachment` struct. While `StoredInboundAttachment::from()` correctly strips `extras_json` before disk persistence, the key is present in the in-memory attachment as it flows through the pipeline.

**Fix:** Clear extras_json immediately after hydration succeeds, or pass the aeskey as a separate parameter.

channels-src/wecom/src/lib.rs
```diff
@@ -0,0 +1,3150 @@
...
+use serde::{Deserialize, Serialize};
+use serde_json::{Map as JsonMap, Value as JsonValue};
+use sha1::{Digest, Sha1};
+use std::collections::HashMap;
+
+use exports::near::agent::channel::{
+    AgentResponse, Attachment, ChannelConfig, Guest, IncomingHttpRequest, OutgoingHttpResponse,
+    PollConfig, StatusType, StatusUpdate,
+};
+use near::agent::channel_host::{self, EmittedMessage, InboundAttachment};
```
henrypark133, line 30: **Low** — Message deduplication window is bounded and filesystem-dependent.

Deduplication uses a bounded list of 256 recent message IDs. Messages older than the retained window can be replayed. The error path defaults to `true` (process the message) on state update failure.

**Fix:** Add a timestamp-based TTL to each message ID entry so stale entries are pruned by age as well as count.

henrypark133 (changes requested):
## Code Review (multi-agent)

**Intent:** Add a standalone WeCom WASM channel with MVP features including inbound callback, Agent API outbound, setup validation, token caching, media handling, and pairing support.

**Stats:** 16 findings across 7 files. Reviewers run: bugs, conventions, design, performance, security, tests. Reviewers failed: none. Body-only: 7

### Security

1. **Low** — AES key present in InboundAttachment extras_json during processing pipeline (`channels-src/wecom/src/lib.rs:13`, confidence 50) — anchor: channels-src/wecom/src/lib.rs:1734
   The `websocket_attachment_from_binary` function stores the raw `aeskey` in `extras_json` on the `InboundAttachment` struct. While `StoredInboundAttachment::from()` correctly strips `extras_json` befor

2. **Low** — Message deduplication window is bounded and filesystem-dependent (`channels-src/wecom/src/lib.rs:28`, confidence 50) — anchor: channels-src/wecom/src/lib.rs:1918
   Deduplication uses a bounded list of 256 recent message IDs. Messages older than the retained window can be replayed. The error path defaults to `true` (process the message) on state update failure.

3. **Low** — Websocket secrets fall back to environment variables without scope isolation (`src/channels/wasm/wrapper.rs:4118-4132`, confidence 50) — anchor: src/channels/wasm/wrapper.rs:4129 (no diff position — body only)
   `resolve_websocket_secret` falls back to `std::env::var(secret_name.to_uppercase())` when the secrets store is unavailable. Environment variables are process-global and not scoped to the owner/instanc

### Bugs

1. **Medium** — Staged generated image files leak when response is suppressed (`src/agent/agent_loop.rs:2290-2311`, confidence 100) — anchor: src/agent/agent_loop.rs:2296 (no diff position — body only)
   When SubmissionResult::Response carries staged generated-image attachment paths but the content triggers a silent-reply or is empty, the code returns HandleOutcome::Shutdown or HandleOutcome::NoRespon

### Performance

1. **High** — Full base64 decode per chunk upload wastes O(n×chunks) CPU (`channels-src/wecom/src/lib.rs:15`, confidence 90) — anchor: channels-src/wecom/src/lib.rs:1164
   build_websocket_media_init_payload and build_websocket_media_chunk_payload each decode the entire send.data_base64 string (up to ~13MB for a 10MB attachment) on every call. For a 10MB file split into 

2. **High** — handle_websocket_ack_frame loads/serializes full media state per ack (`channels-src/wecom/src/lib.rs:1`, confidence 85) — anchor: channels-src/wecom/src/lib.rs:1633
   Every websocket ack triggers load_pending_websocket_media_state() (full JSON parse), apply_websocket_ack_to_media_state(), then persist_pending_websocket_media_state() (full JSON serialize + write). T

3. **High** — Stale websocket media state with base64 payloads never cleaned up on disconnect (`channels-src/wecom/src/lib.rs:4`, confidence 80) — anchor: channels-src/wecom/src/lib.rs:1095
   If the websocket connection drops mid-upload, PendingWebsocketMediaState persists to workspace storage with full base64-encoded attachment data. On reconnect, there is no TTL check or cleanup mechanis

4. **Medium** — Batch errors Vec grows without bound on repeated failures (`channels-src/wecom/src/lib.rs:237`, confidence 65) — anchor: channels-src/wecom/src/lib.rs:1486
   In complete_websocket_media_batch, each failed media send appends an error string to batch.errors. If a batch has many attachments that all fail, errors Vec grows unbounded.

### Tests

1. **High** — stage_generated_image_data_url has no tests for any of its six error-return paths (`src/generated_images.rs:149-181`, confidence 75) — anchor: src/generated_images.rs:149 (no diff position — body only)
   This function has six distinct failure modes but no tests exercise any of them.

2. **Medium** — No E2E test coverage for the new WeCom channel user-visible flow (`(no file)`, confidence 75) (no diff position — body only)
   The PR adds a complete WeCom WASM channel but tests/e2e/ has no WeCom scenario.

3. **Low** — strip_markdown_image_lines missing tests for passthrough and multi-image cases (`src/agent/thread_ops.rs:80-87`, confidence 50) — anchor: src/agent/thread_ops.rs:80 (no diff position — body only)
   The existing test only covers all-image-lines-returning-empty and one-image-line-stripped.

### Conventions

1. **Medium** — workspace_read().unwrap_or_default() on IO call without silent-ok justification (`channels-src/wecom/src/lib.rs:593`, confidence 100) — anchor: .claude/rules/error-handling.md:12-14
   load_allow_from() collapses workspace read errors into an empty vec via .unwrap_or_default(). The error-handling rule requires fail-loud by default or a '// silent-ok: <reason>' comment.

2. **Medium** — Stringly-typed dm_policy compared against magic string literals (`channels-src/wecom/src/lib.rs:32`, confidence 100) — anchor: .claude/rules/types.md:19-20
   dm_policy is stored as Option<String> and compared via `dm_policy == "open"` and `dm_policy == "pairing"`. The types rule requires enums for fixed small sets.

3. **Medium** — is_reserved_wasm_runtime_config_key duplicated in setup.rs and manager.rs (`src/extensions/manager.rs:123-130`, confidence 100) — anchor: src/channels/wasm/setup.rs:701 (no diff position — body only)
   The same fn is_reserved_wasm_runtime_config_key() with identical body exists in both src/channels/wasm/setup.rs:701 and src/extensions/manager.rs:123.

### Design

1. **Medium** — Validation helper functions duplicated between setup/channels.rs and extensions/manager.rs (`src/setup/channels.rs:967-1020`, confidence 100) — anchor: AGENTS.md:41 (no diff position — body only)
   validation_endpoint_body_error, validation_response_exceeds_limit, and the placeholder-name parser are near-identical copies in both files.

2. **Low** — WecomConfig.dm_policy is Option<String> instead of an enum (`channels-src/wecom/src/lib.rs:65`, confidence 75) — anchor: .claude/rules/types.md:19-20
   dm_policy has exactly three valid values but is stored and compared as a raw String throughout the channel.

serrrfirat (changes requested):
## Code Review (forced follow-up)

**Intent:** Add standalone WeCom WASM channel with bot-first WebSocket inbound/outbound, media upload, pairing, setup validation, and WASM runtime support.

**Stats:** 5 actionable findings across 4 files. Reviewers run: security, bugs, performance, tests, conventions. Existing skill review already existed on this head; this review was forced by user request.

**Decision:** Request changes. One verified high-confidence correctness bug can drop generated-image attachments.

### High

1. **Generated image-only responses drop attachments**  
   `src/agent/agent_loop.rs:2299-2300`  
   `thread_ops` now strips markdown image-only content when generated-image attachments exist, then returns `SubmissionResult::response_with_attachments("", attachments)`. The agent loop still maps any empty `content` to `HandleOutcome::NoResponse` before calling `build_outgoing_response_for_thread`, so image-only replies never send their staged attachment.  
   **Fix:** treat empty content as no response only when `attachments.is_empty()`. Otherwise build an outgoing response with empty content and the attachments (or provide a fallback caption).

### Medium

2. **WeCom media upload state persists full base64 payload and re-decodes it per chunk**  
   `channels-src/wecom/src/lib.rs:1211-1234`, `1633-1649`  
   `PendingWebsocketMediaSend` stores full attachment bytes as base64 in serialized workspace state. Each ACK reloads/persists the whole state, and `build_websocket_media_chunk_payload` decodes the full base64 string just to send one chunk. A 10MB file can cause repeated 13MB JSON serialization plus repeated full decode work across the upload.  
   **Fix:** persist media bytes under a separate blob key or keep decoded bytes in active upload state; store only metadata/chunk cursor in the ACK state. Encode only the chunk being sent.

3. **WeCom websocket session actions lack direct tests**  
   `src/channels/wasm/wrapper.rs:4985-5034`  
   The PR adds WeCom AI Bot session control behavior (subscribe ACK starts heartbeat, subscribe error reconnects, ping ACK resets missed heartbeats, disconnected event enqueues then stops runtime), but tests only cover frame parsing/message builders.  
   **Fix:** add `src/channels/wasm/wrapper.rs::tests::test_wecom_aibot_session_process_text_frame_handles_auth_ping_and_disconnect` covering subscribe success => `SetHeartbeat`, subscribe error => `Reconnect`, ping success => reset missed heartbeat count, disconnected event => enqueue + stop runtime.

4. **WeCom media ACK side-effect path lacks caller test**  
   `channels-src/wecom/src/lib.rs:1633-1654`  
   State transitions are unit-tested through pure helpers, but the actual `handle_websocket_ack_frame` / poll side effects are not: persisted state write, outbound `websocket_send_text` for next command, and failure handling if send fails. This path gates real media upload progression after `aibot_upload_media_init`.  
   **Fix:** add `src/channels/wasm/wrapper.rs::tests::test_wecom_component_poll_advances_media_upload_ack_queue` and `test_wecom_component_poll_records_media_failure_when_next_command_send_fails`.

5. **Setup validation SSRF guard lacks configure call-site test**  
   `src/extensions/manager.rs:7907-7915`  
   The validation endpoint path now calls `validate_fetch_url`, `validate_and_resolve_http_target`, and `ssrf_safe_client_builder_for_target`, but tests cover helpers rather than `ExtensionManager::configure` executing a channel `validation_endpoint`. A future regression could bypass the guard at the call-site while helper tests still pass.  
   **Fix:** add `src/extensions/manager.rs::tests::test_configure_wasm_channel_validation_endpoint_rejects_loopback_url` using a channel capability validation endpoint like `http://127.0.0.1:9/validate?token={sample_bot_token}`, asserting `ExtensionError::Other` contains `SSRF blocked` and the token is not persisted.

### Notes

- Dropped non-actionable items: WeCom protocol AES-CBC IV derivation, bounded recent-message Vec scan, and several low-confidence style/performance nits.
- No confirmed secret leakage or auth bypass found in this forced pass.

serrrfirat (changes requested):
I reviewed the current head `a921300a9196fb66445dc4679cebb9a6b762d824` and am requesting changes for these verified issues.

1. High — `dm_policy=open` also opens group-chat sender authorization (`channels-src/wecom/src/lib.rs:777`). `dm_policy` is documented/setup as applying to DM sender authorization, but `is_sender_allowed` is used for both private and group websocket messages and returns true for `DmPolicy::Open` before considering `chattype`/`chatid`. Any sender in any WeCom group can drive the agent when DM open mode is enabled. Pass chat type/chat ID into authorization and keep `open` scoped to private chats, or add an explicit group policy that defaults closed/allowlist.

2. High — websocket outbound has no backpressure or frame-size cap (`src/channels/wasm/wrapper.rs:350`, `src/channels/wasm/wrapper.rs:1526`). The host ABI accepts any guest string and sends it to an unbounded mpsc sender. The sender is installed before the first successful connection and remains usable through reconnect backoff, so a component can enqueue unlimited large frames while disconnected or slow. Use a bounded queue, enforce max payload bytes, and return a deterministic disconnected/full error.

3. Medium — websocket frames are dropped when `on_poll` fails (`src/channels/wasm/wrapper.rs:4697`). `spawn_websocket_poll` moves the live queue into the processing queue before `execute_poll`. If `execute_poll` traps/times out/returns `Err`, the code only logs while `move_json_text_queue` has already removed the source queue, so frames are lost permanently. Restore/prepend the processing queue on failure, or only clear after successful poll, and add a retry regression test.

4. Medium — sensitive auth/approval statuses can be sent into group chats (`channels-src/wecom/src/lib.rs:1153`, `channels-src/wecom/src/lib.rs:1160`). `ApprovalNeeded`, `AuthRequired`, and `AuthCompleted` are forwarded as chat text, and `send_status_notification` replies by websocket req_id without checking group metadata. A group-triggered flow can expose setup/OAuth/approval details to every member. Suppress sensitive group statuses or replace them with a generic DM-only instruction.

5. Medium — WeCom component/runtime tests silently skip when the wasm fixture is absent (`src/channels/wasm/wrapper.rs:6034`). The helper returns `None` when `channels-src/wecom/wecom.wasm` is missing, and the tests return early. This checkout does not contain the fixture, so CI can pass without exercising the actual component sender/generated-image response paths. Make the fixture mandatory or build it in test setup.

Verification: `cargo test --manifest-path channels-src/wecom/Cargo.toml -q --offline -j 1` passed with 35 tests. Targeted root tests and `wit_compat` could not complete locally because linking/building failed with `errno=28` / no space left on device.

serrrfirat (changes requested):
Reviewed the current head against #2395 and the prior change requests. The shape is much better now: the group-chat `dm_policy=open` hole is closed, websocket outbound now has bounded backpressure and frame-size checks, failed websocket polls restore queued frames, sensitive group-chat statuses are generic, and the WeCom component tests no longer silently skip when the wasm fixture is absent.

Two things still need cleanup before merge.

1. P1 — CI is still red on clippy

`src/channels/wasm/wrapper.rs:4802` currently fails `Clippy (all-features)` under `-D warnings` with `clippy::collapsible_if`. This keeps the repo-wide zero-warning gate red even though the rest of the runtime/test matrix is green.

Please collapse the nested `if let Some(payload)` / `if let Err(error)` block and rerun the code style gate.

2. P2 — PR scope text is stale after self-built app support was removed

The PR body still says this ships “self-built app callback inbound” and “Agent API outbound,” but the current code/docs say the opposite: `channels-src/wecom/src/lib.rs` notes HTTP callback/proactive REST send are out of scope, `channels-src/wecom/wecom.capabilities.json` has no webhook, and `docs/plans/2026-04-01-wecom-channel-plan.md` explicitly defers self-built application callback and Agent API behavior.

Please update the PR body to describe the current bot-first WeCom AI Bot slice, and call out self-built app callback / Agent API proactive send as follow-up work. That keeps reviewer expectations and #2395 scope honest.

Verification:
- `cargo test --manifest-path channels-src/wecom/Cargo.toml -q --offline -j 1` passed locally: 37 passed.
- CI currently shows `WASM WIT Compatibility`, `Tests (all-features)`, web E2E, integration, and version checks passing.
- CI currently shows `Clippy (all-features)` and aggregate `Code Style (fmt + clippy)` failing for the clippy issue above.

Net: request changes for the red clippy gate and stale PR body. I do not see the prior security/correctness blockers still open on the latest head.

src/channels/wasm/wrapper.rs
```diff
@@ -3767,20 +4094,77 @@ fn websocket_processing_queue_path(channel_name: &str) -> String {
...
+    owner_scope_id: &str,
+    secret_name: &str,
+) -> Option<String> {
+    if let Some(store) = store
+        && let Ok(secret) = store.get_decrypted(owner_scope_id, secret_name).await
+    {
+        return Some(secret.expose().to_string());
+    }
+
+    std::env::var(secret_name.to_uppercase())
```
serrrfirat: [High][security] This fallback reads any environment variable matching the capability-declared secret name. A malicious or compromised WASM channel could declare a websocket credential like `aws_secret_access_key` and have the host send that process secret in the auth payload to the channel's allowlisted websocket endpoint. Remove this generic env fallback, or scope it by channel and by declared setup secrets the same way channel setup injection does.

channels-src/wecom/src/lib.rs
```diff
@@ -0,0 +1,3677 @@
...
+        vec![bundle]
+    }
+}
+
+fn emit_pending_inbound_bundle(bundle: PendingInboundBundle) {
+    let was_buffered = bundle.flush_at_ms > 0;
+    let mut attachments: Vec<InboundAttachment> =
+        bundle.attachments.into_iter().map(Into::into).collect();
+    let attachment_ids: Vec<String> = attachments.iter().map(|att| att.id.clone()).collect();
+    if was_buffered {
```
serrrfirat: [High][correctness] Immediate mixed text+media messages lose their attachment bytes. `emit_or_buffer_incoming_user_message` always converts incoming attachments into `StoredInboundAttachment`, which intentionally drops `source_url`, `storage_key`, and extras, then `emit_pending_inbound_bundle` only rehydrates blobs when `flush_at_ms > 0`. A mixed text+image frame takes the immediate path, emits metadata-only attachments, and then clears the saved blob. Rehydrate before every emit, or keep original attachment data for immediate bundles.

src/channels/wasm/host.rs
```diff
@@ -622,6 +622,53 @@ impl ChannelWorkspaceStore {
...
+            })?;
+        let mut live: Vec<String> = data
+            .get(source_path)
+            .and_then(|raw| serde_json::from_str(raw).ok())
+            .unwrap_or_default();
+
+        restored.append(&mut live);
+        if restored.len() > max_items {
+            let overflow = restored.len() - max_items;
+            restored.drain(0..overflow);
```
serrrfirat: [Medium][correctness] This restore path appends live frames after the failed processing queue, then drains overflow from the front. Under backlog, that drops the oldest failed/unprocessed frames first, exactly the frames this restore is meant to retry. Keep the processing prefix and drop overflow from the tail/live suffix, for example by truncating after append.

channels-src/wecom/src/lib.rs
```diff
@@ -0,0 +1,3677 @@
...
+    if let Some(raw) = channel_host::workspace_read(ALLOW_FROM_PATH) {
+        match serde_json::from_str::<Vec<String>>(&raw) {
+            Ok(values) => allowed = values,
+            Err(error) => channel_host::log(
+                channel_host::LogLevel::Warn,
+                &format!("Failed to parse WeCom allow_from list: {error}"),
+            ),
+        }
+    }
+    if let Ok(stored) = channel_host::pairing_read_allow_from(CHANNEL_NAME) {
```
serrrfirat, line 723: [Medium][performance] Every inbound WeCom frame calls `load_allow_from()`, which rereads the full DB-backed pairing allowlist and then scans it. `process_websocket_event_queue` can process up to 100 frames per callback, so a busy channel can do 100 DB reads plus O(frames * paired_users) scans while the WASM callback is running. Prefer the per-sender `pairing_resolve_identity` path, or load the allowlist once per queue drain and pass it through.

src/channels/wasm/wrapper.rs
```diff
@@ -2824,6 +2998,34 @@ impl WasmChannel {
...
+            StatusUpdate::ImageGenerated { data_url, .. }
+                if uses_wecom_aibot_protocol(&self.capabilities) =>
+            {
+                // WeCom AI Bot stream replies are updated by req_id/stream id.
+                // Sending the image as its own final frame before the assistant's
+                // final text lets the text response overwrite the image. Stage
+                // it and deliver it with the final on_respond payload instead.
+                self.cancel_typing_task().await;
+
+                match stage_generated_image_data_url(data_url) {
```
serrrfirat, line 3026: [Medium][resource] Staged generated images are only removed by the later `respond()` path. If the turn yields no response, an outbound hook suppresses the reply, or the channel shuts down before `respond()`, this map entry and temp file remain indefinitely. Clear staged images on terminal/no-response/hook-blocked paths, or add bounded TTL cleanup for `pending_generated_image_attachments`.

channels-src/wecom/src/lib.rs
```diff
@@ -0,0 +1,3677 @@
...
+            started: 0,
+            errors: vec!["WeCom websocket media send requires chat_id or user_id".to_string()],
+        };
+    };
+
+    let batch_id = websocket_media_batch_id(metadata);
+    let mut sends = Vec::new();
+    let mut outbound_payloads = Vec::new();
+    let mut errors = Vec::new();
+    for (index, attachment) in attachments.iter().enumerate() {
```
serrrfirat, line 1638: [Medium][resource] This eagerly base64-chunks and persists every response attachment before any upload ack arrives. Each attachment can be up to 20 MB and there is no per-response count or total-byte cap, so one response can retain hundreds of MB of workspace state until the media TTL. Add a per-response media count/total-byte limit and consider uploading sequentially instead of staging the whole batch.

src/extensions/manager.rs
```diff
@@ -975,6 +1121,39 @@ impl ExtensionManager {
...
+                                channel = %name,
+                                key = %config_key,
+                                "Ignoring reserved wasm runtime config override key"
+                            );
+                            continue;
+                        }
+                        overrides.insert(config_key.to_string(), value);
+                    }
+                }
+                Err(e) => {
```
serrrfirat, line 1147: [Medium][error-handling] This settings read failure is silently ignored, leaving the channel activated with empty runtime overrides. The repo rule in `.claude/rules/error-handling.md` says DB/settings reads should fail loud by default unless there is an explicit `silent-ok` rationale. Return a `Result` from this loader and propagate the read error, or document why degraded activation is intentional here.

src/setup/channels.rs
```diff
@@ -832,11 +836,19 @@ pub async fn setup_wasm_channel(
         }
     }
 
+    let config_overrides = if channel_name.eq_ignore_ascii_case("wecom") {
```
serrrfirat, line 839: [Medium][docs] This changes onboarding/setup behavior by adding WeCom-specific runtime override prompts and persisted `channels.wasm_channel_runtime_overrides`, but `src/setup/README.md` was not updated. `AGENTS.md` requires setup/onboarding behavior changes to update that spec in the same branch. Please document the WeCom prompts and persistence behavior there.

channels-src/wecom/src/lib.rs
```diff
@@ -0,0 +1,3677 @@
...
+        user_id: sender_id,
+        user_name: None,
+        content,
+        thread_id: Some(conversation_scope),
+        metadata_json,
+        attachments: Vec::new(),
+    });
+}
+
+fn process_websocket_event_queue() {
```
serrrfirat, line 2420: [High][tests] The new WeCom bot-first inbound path is only covered at parser/helper level. There should be a caller-level test that queues an `aibot_msg_callback`, drives `Guest::on_poll` through the compiled channel/runtime path, and verifies the emitted `IncomingMessage`, thread scope, metadata, and attachment behavior. Suggested test: `src::channels::wasm::wrapper::tests::test_call_on_poll_emits_wecom_websocket_callback_message`.

src/channels/wasm/wrapper.rs
```diff
@@ -4271,15 +4785,32 @@ fn spawn_websocket_poll(poll_guard: tokio::sync::OwnedMutexGuard<()>, ctx: Webso
                 }
                 Err(error) => {
                     tracing::warn!(channel = %ctx.channel_name, error = %error, "Websocket-triggered poll failed");
+                    if let Err(restore_error) = ctx.workspace_store.restore_json_text_queue(
```
serrrfirat, line 4804: [Medium][tests] The new poll-failure restore branch is not tested through the actual async caller. A helper-level queue restore test would not catch passing the wrong paths here, failing to restore after `execute_poll`, or continuing with stale processing state. Add a caller-level test such as `src::channels::wasm::wrapper::tests::test_websocket_poll_restores_processing_queue_when_on_poll_fails`.

serrrfirat (changes requested):
Multi-agent review of PR #2394 at dcedead23488f30cff7658ef3ab54488580ce892.

I found 10 actionable issues worth inline comments. Because there are two High-severity findings in security/correctness paths, I am requesting changes.

Summary:
- High: 3 (secret exposure risk, inbound media loss, missing caller-level WeCom inbound coverage)
- Medium: 7 (queue-loss, resource/DB pressure, cleanup/error handling, setup docs, poll restore coverage)
- Low: 1 additional non-blocking test gap noted below

Additional non-blocking note: `src/generated_images.rs` adds several reject paths in `stage_generated_image_data_url` without direct malformed/oversized input tests. A focused unit test would make those error semantics harder to regress.

serrrfirat (approved):
## Review: WeCom bot-first channel

Strong follow-through on the review cycle. The current head `70e84137180726be99a9255fe68491f9fcdd84d4` has the important production invariants in much better shape: WeCom is scoped as a bot-first WASM channel, the websocket runtime is protocol-generalized without leaving Discord-specific assumptions in the WeCom path, and the prior security/correctness issues around group sender admission, sensitive group statuses, websocket outbound backpressure, poll retry loss, generated-image delivery, and inbound media hydration now have caller-level coverage.

### Strengths

- The PR scope is now honest against #2395: self-built app callback inbound and Agent API proactive send/media fallback are explicitly follow-up work, not implied as shipped.
- The WeCom component path is no longer silently skipped in runtime tests; current CI shows WIT compatibility, all-features tests, clippy, integration, and gateway smoke checks green.
- The media and generated-image fixes prove the real call sites, not just helpers: WeCom poll emits hydrated messages, unhydrated media is dropped instead of emitting metadata-only attachments, failed websocket polls restore queued frames, and final generated-image responses dedupe status-staged images.
- The validation endpoint path now uses the SSRF-safe resolver/client at the configure call site and includes a loopback rejection test that also asserts the submitted token is not persisted.

### Non-blocking observations

1. The PR validation checklist still leaves clippy unchecked even though `Clippy (all-features)` and `Code Style (fmt + clippy)` are green on this head. Please update that box before merge if you want the body to match CI.

2. This remains a broad surface-area PR. The rollback plan and follow-up list are clear enough for this slice, but I would keep broader real WeCom group/media/event E2E and stronger ack tracking as near-term follow-ups rather than letting them disappear into general backlog.

Verification I checked locally:
- `cargo test --manifest-path channels-src/wecom/Cargo.toml -q --offline -j 1` passed: 39 tests.

Approve.

think-in-universe (approved):

henrypark133 (approved):

PR #2396, opened 2026-04-13
feat: discover tool source in working directory during install

ilblackdragon:
## Summary

- When `tool_install` can't find a tool in the registry, it now searches common directories relative to the working directory before returning "not found"
- Searches `tools-src/<name>/`, `tool-src/<name>/`, and `<name>/` (direct subdirectory), matching both hyphenated and underscored variants, with `_tool`/`-tool` suffix handling
- A directory is only considered a match if it contains a `Cargo.toml`

This lets users say "install portfolio tool" when the source is at `tool-src/portfolio/` without needing to specify the full path.

## Test plan

- [x] 8 unit tests for `find_local_tool_source_in` covering:
  - `tools-src/` and `tool-src/` discovery
  - Hyphenated/underscored name variant matching
  - Direct subdirectory discovery
  - `_tool` suffix stripping
  - Priority ordering (tools-src/ preferred over direct)
  - Negative cases (no Cargo.toml, not present)
- [x] `cargo clippy` — zero warnings
- [x] `cargo test` — all 4710 tests pass


src/extensions/manager.rs
```diff
@@ -1492,11 +1553,31 @@ impl ExtensionManager {
             });
         }
 
+        // Try to discover a local tool source directory in the working directory
+        if let Some(source_dir) = find_local_tool_source(&name) {
```
serrrfirat: **Medium Severity — Security: Implicit local install is a confused-deputy vector**

When the registry and URL lookups both fail, `find_local_tool_source` silently scans the CWD for matching directories. An attacker who controls a cloned repository can plant `tools-src/<name>/` with a pre-built WASM artifact. If the LLM then calls `tool_install("name")`, the malicious WASM is installed without the user knowing it came from a local directory rather than the verified registry.

The approval gate fires *before* `execute()` runs — the user sees "approve tool_install(name: portfolio)" and has no way to distinguish local vs registry at that point. The `InstallResult.message` says "installed from local build artifacts" but doesn't include the source path.

**Suggested fix:**
1. Include the full source path in the `InstallResult.message` so the user/LLM can verify provenance:
   ```rust
   message: format!(
       "{} '{}' installed from LOCAL build artifacts at {}. Run activate to load it.",
       kind_label, name, source_dir.display(),
   ),
   ```
2. Consider logging a user-visible warning or returning a distinct status field (e.g. `"source": "local"` vs `"source": "registry"`) so downstream consumers can make trust decisions.

src/extensions/manager.rs
```diff
@@ -1492,11 +1553,31 @@ impl ExtensionManager {
...
+            tracing::info!(
```
serrrfirat: **Medium Severity — IronClaw convention: `tracing::info!` for internal diagnostic**

Per CLAUDE.md: *"Logging levels matter for REPL/TUI: `info!` and `warn!` output appears in the REPL and corrupts the terminal UI. Use `debug!` for internal diagnostics. Reserve `info!` for user-facing status that the REPL intentionally renders."*

"Found local tool source directory" is an internal discovery detail, not user-facing status. This will appear in the REPL and corrupt the TUI display.

**Suggested fix:**
```rust
tracing::debug!(
    extension = %name,
    source_dir = %source_dir.display(),
    "Found local tool source directory"
);
```

src/extensions/manager.rs
```diff
@@ -12462,4 +12544,125 @@ mod tests {
 
         Ok(())
     }
+
+    // ── find_local_tool_source_in ──────────────────────────────────────
```
serrrfirat, line 12010: **Medium Severity — Test coverage: No caller-level test for `install()` → `find_local_tool_source` path**

Per `.claude/rules/testing.md` ("Test Through the Caller, Not Just the Helper"):

> When a helper gates a side-effecting flow, the test must go through the caller — not just the helper in isolation.

`find_local_tool_source` is a predicate gating a side effect (`install_wasm_from_buildable`). The wrapper `install()` sits between the helper and the side effect and computes multiple inputs: `kind` (defaulted via `unwrap_or`), `target_dir` (conditional on kind), and `source_str` (via `to_string_lossy()`). The 8 unit tests on `find_local_tool_source_in` are thorough, but they can't catch bugs in the wrapper layer — e.g., if `kind` defaulting is wrong, or `source_str` is mangled.

**Suggested fix:** Add a test (integration tier or with a minimal `ExtensionManager`) that calls `install()` with a name that isn't in the registry and no URL, and verifies the local discovery fallback fires and the result message reflects the local source.

serrrfirat:
## Paranoid Architect Review — Verdict

**Overall: Approve with requested changes (3 Medium findings)**

### What's good

- **Path traversal is properly blocked** — `canonicalize_extension_name` rejects `/`, `\`, `..`, `\0` before the name ever reaches `find_local_tool_source`. No way to escape the search root.
- **Registry always wins** — The discovery is correctly placed as the *last* fallback, after registry lookup and URL-based install. Existing registry tools cannot be shadowed by local directories.
- **Two-gate barrier** — A directory must contain both a `Cargo.toml` and a pre-built WASM release artifact (`target/wasm32-*/release/<name>.wasm`) to match. Just planting a `Cargo.toml` doesn't execute anything.
- **Clean testable design** — The `_in` variant with an explicit search root is the right pattern for testability, and the 8 unit tests are well-structured with proper `tempfile` usage.
- **Good feature instinct** — This solves a real UX pain point ("install portfolio tool" when the source is right there in `tool-src/`).

### What needs fixing before merge

| # | Severity | Summary |
|---|----------|---------|
| 1 | Medium | **Supply chain / confused deputy** — The user approves the install *before* the local discovery runs, so they can't distinguish local vs registry provenance at the approval gate. The `InstallResult` should include the source path and ideally a distinct `"source": "local"` field. |
| 2 | Medium | **`tracing::info!` → `tracing::debug!`** — "Found local tool source directory" is an internal diagnostic that corrupts the REPL/TUI. Straightforward fix. |
| 3 | Medium | **Missing caller-level test** — Per `.claude/rules/testing.md`, the helper tests don't cover the wrapper layer in `install()` where `kind`, `target_dir`, and `source_str` are computed. Needs at least one test through `install()`. |

### Minor items (not blocking)

- `names.dedup()` without `sort()` — only removes consecutive duplicates, leaves non-adjacent dupes
- Double-suffix junk variants (`portfolio_tool-tool`) generated for names already ending in `_tool`
- `to_string_lossy()` could silently corrupt non-UTF8 paths — `to_str()` with explicit error would be more correct
- Tool description doesn't distinguish local sources from the verified registry

None of these are blocking, but the first three Medium findings should be addressed before this ships.

src/extensions/manager.rs
```diff
@@ -1492,11 +1553,31 @@ impl ExtensionManager {
...
+        if let Some(source_dir) = find_local_tool_source(&name) {
+            tracing::info!(
+                extension = %name,
+                source_dir = %source_dir.display(),
+                "Found local tool source directory"
+            );
+            let kind = kind_hint.unwrap_or(ExtensionKind::WasmTool);
+            let target_dir = match kind {
+                ExtensionKind::WasmChannel => &self.wasm_channels_dir,
+                _ => &self.wasm_tools_dir,
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in 594a0c1b. Added `kind_allows_local_discovery()` guard that restricts local-source fallback to `WasmTool`/`WasmChannel` (and `None`). Non-WASM kind hints (`McpServer`, `ChannelRelay`, `AcpAgent`) now fall through to the NotFound error instead of routing into `install_wasm_from_buildable`.

src/extensions/manager.rs
```diff
@@ -554,6 +554,67 @@ fn sanitize_url_for_logging(url: &str) -> String {
...
+    for suffix in ["_tool", "-tool"] {
+        if let Some(base) = underscore_name.strip_suffix(suffix) {
+            names.push(base.to_string());
+            names.push(base.replace('_', "-"));
+        }
+        if !underscore_name.ends_with(suffix) {
+            names.push(format!("{}{}", underscore_name, suffix));
+            names.push(format!("{}{}", hyphen_name, suffix));
+        }
+    }
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in 594a0c1b. Rewrote candidate generation to use a `HashSet` and a single `strip_suffix("_tool").or_else("-tool")` branch — adds either suffixed or stripped variants (never both), so no more mixed forms like `my_portfolio-tool` or `*_tool-tool`.

src/extensions/manager.rs
```diff
@@ -554,6 +554,67 @@ fn sanitize_url_for_logging(url: &str) -> String {
...
+    names.dedup();
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in 594a0c1b. Replaced `Vec` + `dedup()` with a `HashSet` — handles non-adjacent duplicates correctly.

src/extensions/manager.rs
```diff
@@ -1492,11 +1553,31 @@ impl ExtensionManager {
...
+            return self
+                .install_wasm_from_buildable(&name, Some(&source_str), None, target_dir, kind)
+                .await;
+        }
+
         let err = ExtensionError::NotFound(format!(
-            "'{}' not found in registry. Try searching with discover:true or provide a URL.",
+            "'{}' not found in registry or local directories. \
+             Try searching with discover:true, provide a URL, \
+             or ensure tool source exists in tools-src/{0}/ or tool-src/{0}/.",
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in 594a0c1b. Updated NotFound message to mention all 3 patterns: `tools-src/<name>/`, `tool-src/<name>/`, and `./<name>/`.

src/extensions/manager.rs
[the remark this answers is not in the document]
ilblackdragon: Fixed in 594a0c1b — applied the HashSet refactor you suggested (slightly adapted: single `strip_suffix` branch that adds stripped-or-suffixed variants, never both).

src/extensions/manager.rs
ilblackdragon: Fixed in 594a0c1b. The `InstallResult.message` now appends `(installed from LOCAL source at <path>)` for local-discovered installs, so the caller can see the provenance. Extracted the logic into `install_from_local_source()` for clarity. Agree the approval gate still fires before `execute()` — that's a broader issue worth addressing separately (perhaps by surfacing the resolved source in the approval prompt itself). Opening to follow-up.

src/extensions/manager.rs
ilblackdragon: Fixed in 594a0c1b. Changed `tracing::info!` to `tracing::debug!` for the "Found local tool source directory" log.

src/extensions/manager.rs
ilblackdragon: Fixed in 594a0c1b. Extracted `install_from_local_source()` helper and added two caller-level tests:

- `install_from_local_source_installs_wasm_tool_with_provenance_message` — asserts kind defaults to `WasmTool`, wasm lands in `tools_dir`, and message contains `LOCAL` + source path.
- `install_from_local_source_routes_wasm_channel_to_channels_dir` — asserts `WasmChannel` kind_hint routes to `channels_dir` (not `tools_dir`).

Also added unit tests for the `kind_allows_local_discovery` predicate. Kept the 8 existing `find_local_tool_source_in` unit tests for the helper layer.

src/extensions/manager.rs
```diff
@@ -3381,6 +3470,41 @@ impl ExtensionManager {
...
+        &self,
+        name: &str,
+        source_dir: &std::path::Path,
+        kind_hint: Option<ExtensionKind>,
+    ) -> Result<InstallResult, ExtensionError> {
+        tracing::debug!(
+            extension = %name,
+            source_dir = %source_dir.display(),
+            "Found local tool source directory"
+        );
```
serrrfirat, line 3497: **High Severity — Suffix stripping creates name mismatch → install always fails**

`find_local_tool_source_in` strips `_tool` suffixes for directory matching (e.g., input `"portfolio_tool"` → finds `tools-src/portfolio/`). But `install_from_local_source` passes `None` for `crate_name`, so `install_wasm_from_buildable` sets `binary_name = name = "portfolio_tool"` and calls `find_wasm_artifact(dir, "portfolio_tool", "release")`. The actual binary is `portfolio.wasm` (the Cargo crate is named `portfolio`). Result: artifact-not-found error.

The suffix-*adding* path has the same bug in reverse: input `"portfolio"` can discover `tools-src/portfolio_tool/`, but then looks for `portfolio.wasm` when the binary is `portfolio_tool.wasm`.

**Suggested fix:** Read the actual crate name from the discovered `Cargo.toml` and pass it as `crate_name`:

```rust
async fn install_from_local_source(
    &self,
    name: &str,
    source_dir: &std::path::Path,
    kind_hint: Option<ExtensionKind>,
) -> Result<InstallResult, ExtensionError> {
    // Extract the actual crate name from Cargo.toml so binary lookup matches
    let crate_name = std::fs::read_to_string(source_dir.join("Cargo.toml"))
        .ok()
        .and_then(|s| {
            s.parse::<toml::Value>().ok()?.get("package")?.get("name")?.as_str().map(String::from)
        });

    // ... existing code ...
    let mut result = self
        .install_wasm_from_buildable(
            name,
            Some(&source_str),
            crate_name.as_deref(),  // pass the real crate name
            target_dir,
            kind,
        )
        .await?;
    // ...
}
```

The existing caller-level tests only exercise the case where `name == crate_name` (`"portfolio"` / `"portfolio"`), so this bug is untested. A test driving `install_from_local_source("portfolio_tool", tools-src/portfolio/, ...)` would reproduce it immediately — this is the "test through the caller, not just the helper" pattern from the project's testing rules.

src/extensions/manager.rs
```diff
@@ -554,6 +554,82 @@ fn sanitize_url_for_logging(url: &str) -> String {
...
+        candidates.insert(format!("{}_tool", underscore_name));
+        candidates.insert(format!("{}-tool", hyphen_name));
+    }
+
+    let search_dirs = ["tools-src", "tool-src", "."];
+
+    for dir_prefix in &search_dirs {
+        let base = if *dir_prefix == "." {
+            search_root.to_path_buf()
+        } else {
```
serrrfirat, line 534: **Medium Severity — Non-deterministic candidate ordering within same search directory**

Candidates are stored in a `HashSet`. If two name variants both match directories within the same `search_dir` (e.g., both `tools-src/portfolio/` and `tools-src/portfolio_tool/` contain a `Cargo.toml`), the returned match depends on `HashSet` iteration order, which is randomized per-process. This makes the behavior unreproducible across runs and very difficult to debug.

The existing test `find_local_tool_source_prefers_tools_src_over_direct` validates priority between *search directories* (the outer loop), but within a single search directory the candidate order is arbitrary.

**Suggested fix:** Replace `HashSet` with a `Vec` (with manual dedup or `dedup()` after sorting) or a `BTreeSet` to establish a stable, documented priority order. For example, prefer the exact name over suffix-stripped variants, and prefer underscored over hyphenated:

```rust
let mut candidates = Vec::new();
candidates.push(underscore_name.clone());
candidates.push(hyphen_name.clone());
if let Some(base) = stripped {
    candidates.push(base.to_string());
    candidates.push(base.replace('_', "-"));
} else {
    candidates.push(format!("{}_tool", underscore_name));
    candidates.push(format!("{}-tool", hyphen_name));
}
candidates.dedup(); // remove adjacent duplicates after the deterministic ordering
```

src/extensions/manager.rs
ilblackdragon: Fixed in ccb1fad1 (with the fmt nit in a follow-up commit). `install_from_local_source` now parses `[package].name` from the discovered `Cargo.toml` via a new `read_crate_name_from_cargo_toml` helper and passes it as `crate_name` to `install_wasm_from_buildable` — so suffix-matched directories (e.g. input `portfolio_tool` → dir `portfolio/`) resolve the artifact as `portfolio.wasm`, not `portfolio_tool.wasm`. Added a caller-level regression test `install_from_local_source_resolves_artifact_via_crate_name` that drives the exact mismatch case — it would fail before the fix. Falls back to `name` if Cargo.toml is missing/unparseable.

src/extensions/manager.rs
ilblackdragon: Fixed in ccb1fad1. Replaced the `HashSet` with a priority-ordered `Vec` + `retain` dedup (keeps first occurrence): canonical underscore form first, then hyphen form, then suffix-adjusted variants. So when both `my_tool/` and `my-tool/` exist in the same search dir, the underscore form wins deterministically on every run. Added `find_local_tool_source_prefers_underscore_over_hyphen` which loops the lookup 5x to catch non-determinism.

src/extensions/manager.rs
```diff
@@ -554,6 +554,102 @@ fn sanitize_url_for_logging(url: &str) -> String {
...
+            search_root.join(dir_prefix)
+        };
+
+        if !base.is_dir() {
+            continue;
+        }
+
+        for candidate_name in &candidates {
+            let candidate = base.join(candidate_name);
+            if candidate.is_dir() && candidate.join("Cargo.toml").exists() {
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in a4db56f6. Changed `candidate.join("Cargo.toml").exists()` to `.is_file()` in `find_local_tool_source_in` so a directory (or other non-file entry) named `Cargo.toml` cannot falsely qualify a candidate source directory. Added regression test `find_local_tool_source_ignores_cargo_toml_directory` that creates `tools-src/portfolio/Cargo.toml/` (directory) and asserts the function returns `None`.

src/extensions/manager.rs
```diff
@@ -3381,6 +3490,54 @@ impl ExtensionManager {
...
+            "Found local tool source directory"
+        );
+        let kind = kind_hint.unwrap_or(ExtensionKind::WasmTool);
+        let target_dir = match kind {
+            ExtensionKind::WasmChannel => &self.wasm_channels_dir,
+            _ => &self.wasm_tools_dir,
+        };
+        let source_str = source_dir.to_string_lossy();
+        // Parse the discovered Cargo.toml to learn the actual crate name. The
+        // installed extension name may differ from the crate name when suffix
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in a4db56f6. `install_from_local_source` now uses `source_dir.to_str()` and returns `ExtensionError::InstallFailed("local source path '<path>' is not valid UTF-8")` when the path is not UTF-8, instead of silently lossy-converting. Kept `install_wasm_from_buildable`'s `Option<&str>` signature since its other callers pass strings sourced from JSON registry manifests. Added a `#[cfg(unix)]` regression test `install_from_local_source_rejects_non_utf8_path` that constructs a path via `OsStr::from_bytes(b"/tmp/\x80not-utf8")` and asserts the `InstallFailed` error.

src/extensions/manager.rs
```diff
@@ -468,6 +468,102 @@ fn sanitize_url_for_logging(url: &str) -> String {
...
+    candidates.push(hyphen_name.clone());
+
+    // If the name ends with a tool suffix, also try stripped variants.
+    // Otherwise, try suffixed variants.
+    let stripped = underscore_name
+        .strip_suffix("_tool")
+        .or_else(|| underscore_name.strip_suffix("-tool"));
+    if let Some(base) = stripped {
+        candidates.push(base.to_string());
+        candidates.push(base.replace('_', "-"));
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in ca0fa986. You're right — `underscore_name = name.replace('-', "_")` so `underscore_name.strip_suffix("-tool")` was unreachable. Removed the `.or_else(...)` branch; the single `strip_suffix("_tool")` on the underscore-normalized name correctly covers both `name_tool` and `name-tool` inputs. Added `find_local_tool_source_strips_hyphen_tool_suffix` that drives input `portfolio-tool` against `tools-src/portfolio/` to lock in coverage.

Issue #2277, opened 2026-04-10
V2: ACP-backed child thread backends for delegated external agents


serrrfirat:
## Summary

We want IronClaw v2 to be able to delegate work from a local thread to external coding agents such as Codex, Droid, OpenCode, and other ACP-compatible runtimes, while keeping the parent/child semantics of the v2 thread tree.

The intended UX is straightforward:

- IronClaw can be your main coding agent
- IronClaw can hand off certain tasks to another agent
- the user can say things like:
  - "go to this repo, inspect this issue, give it to my Codex agent"
  - "use Codex for this sub-task"
  - "delegate this fix to Droid and summarize the result"

This issue proposes a v2-native implementation for that flow.

## Why We Want This

Right now we have two different worlds:

- v2 has native sub-agents as child threads
- ACP exists separately as a v1 sandbox/job path

That creates a product gap.

We can already run ACP agents, but the capability is not integrated into the v2 engine's thread model. As a result, we do not have a clean way for a v2 thread to say:

- spawn a child thread
- let that child be executed by an external ACP agent
- track it like any other child thread
- message it again later if it is persistent
- feed the result back into the parent thread in a structured way

The main product goal is not "ACP support" in the abstract. The goal is:

- IronClaw should be a strong primary coding agent
- delegation to external agents should feel native, not bolted on
- v2 should own the orchestration model
- external agent execution should fit the same thread tree and lifecycle model as native child threads

## Current State

In v2, `Thread` already unifies foreground sessions, jobs, routines, and sub-agents into one abstraction. Child-thread spawning exists today.

However, the current runtime effectively assumes that a thread is executed by IronClaw's local `ExecutionLoop`. In other words, v2 has the right unit of work, but not yet a backend/runtime distinction for who actually executes that thread.

ACP today is implemented through a separate path in v1. That works as an ACP runtime, but it is not the right abstraction for v2-native delegated sub-agents.

## Proposal

Introduce **thread backends** in v2.

Instead of treating every thread as locally executed, explicitly model the executor for a thread.

Example shape:

```rust
pub enum ThreadBackend {
    LocalCodeAct,
    Acp {
        agent_name: String,
        cwd: Option<String>,
        mode: AcpSessionMode,
        resume_session_id: Option<String>,
    },
}
```

This should be kept distinct from `ThreadType`.

- `ThreadType` answers: what kind of work is this?
  - `Foreground`
  - `Research`
  - `Mission`
[...]

PR #2416, opened 2026-04-13
Fix ACP sandbox job mode persistence

foltz:
## Summary

This fixes a race when creating sandbox jobs in ACP mode.

`CreateJobTool` persisted the sandbox job row in a spawned task, then immediately tried to update `job_mode`. If the row had not been inserted yet, the mode update was lost and the job kept the DB default of `worker`.

That broke follow-up prompts for ACP sandbox jobs because the gateway only queues prompts for jobs whose stored mode is `claude_code` or starts with `acp`.

## What changed

- make sandbox job persistence synchronous in the create path
- ensure the row exists before `update_sandbox_job_mode()` runs
- keep the ACP mode encoding unchanged: `acp:<agent_name>`

## Why this fixes it

The mode update no longer races the initial insert. Fresh ACP sandbox jobs now persist the correct `job_mode`, so the web/API follow-up prompt path classifies them correctly.

## Validation

Verified locally against a native `ironclaw` service with ACP sandboxing enabled:

- created a fresh ACP sandbox job through the gateway
- confirmed the new row persisted `job_mode = acp:claude-shim-container`
- sent a follow-up prompt through `POST /api/jobs/{id}/prompt`
- confirmed the same job resumed and emitted a second completed turn

Observed job events on the fixed build:

- `message`: `GAMMA`
- `turn_result`: completed
- `message`: `DELTA`
- `turn_result`: completed

Added automated regression coverage:

- `test_acp_mode_persists_job_mode_before_container_creation`
  - creates an ACP sandbox job through `CreateJobTool`
  - forces container creation to fail after persistence
  - asserts the sandbox row still stores `job_mode = acp:<agent_name>`

## Notes

This was discovered while validating a Claude-over-ACP shim experiment, but the bug and fix are generic to ACP sandbox jobs and do not depend on that shim.

## Related work / non-duplicates

- Not a duplicate of `#1915` (`ACP bridge follow-up prompt failures should mark sandbox job failed`).
  That issue is about how follow-up prompt failures are reported once the ACP bridge is already running.
  This PR fixes an earlier create-time persistence bug where ACP sandbox jobs could be misclassified as `worker`.

- Adjacent to `#2119` / `#1981` follow-up prompt handling work, but distinct.
  Those changes focus on ACP completion/error propagation and terminal state handling.
  This PR fixes the sandbox job record write ordering that determines whether follow-up prompts are routed to ACP at all.

- Broader ACP architecture work such as `#2277` is also distinct.
  This PR is a narrow v1 sandbox bugfix, not a v2 ACP/thread-backend design change.

## Test command

Targeted test command:

- `cargo test -p ironclaw test_acp_mode_persists_job_mode_before_container_creation -- --nocapture`
- result: passed

### DIFF
--- a/src/channels/matrix.rs
+++ b/src/channels/matrix.rs
@@ -1,0 +1,735 @@
+//! Matrix channel via external matrix-bridge HTTP/SSE API.
+//!
+//! Connects to a running matrix-bridge daemon.
+//! Listens for messages via SSE at `/api/v1/events` and sends via
+//! JSON-RPC at `/api/v1/rpc`.
+
+use std::collections::HashMap;
+use std::num::NonZeroUsize;
+use std::sync::Arc;
+use std::time::Duration;
+
+use async_trait::async_trait;
+use chrono::{TimeZone, Utc};
+use futures::StreamExt;
+use lru::LruCache;
+use reqwest::Client;
+use serde::Deserialize;
+use serde_json::{Value, json};
+use tokio::sync::RwLock;
+use tokio::task::JoinHandle;
+use uuid::Uuid;
+
+use crate::channels::{
+    Channel, ChatApprovalPrompt, IncomingMessage, MessageStream, OutgoingResponse, StatusUpdate,
+};
+use crate::config::MatrixConfig;
+use crate::error::ChannelError;
+
+const MATRIX_HEALTH_ENDPOINT: &str = "/api/v1/check";
+const MATRIX_EVENTS_ENDPOINT: &str = "/api/v1/events";
+const MATRIX_RPC_ENDPOINT: &str = "/api/v1/rpc";
+const MAX_REPLY_TARGETS: usize = 10_000;
+const REPLY_TARGETS_CAP: NonZeroUsize = NonZeroUsize::new(MAX_REPLY_TARGETS).unwrap();
+
+#[derive(Debug, Clone, PartialEq, Eq)]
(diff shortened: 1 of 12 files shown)


foltz:
Live validation details used for this bugfix:

1. Ran a native `ironclaw` service with `ACP_ENABLED=true`, `SANDBOX_ENABLED=true`, and a worker image containing the Claude ACP shim.
2. Created an ACP sandbox job through the normal gateway/tool path.
3. Before this fix, the new sandbox row could persist with `job_mode = worker`, even though the container was actually launched in ACP mode.
4. That caused `/api/jobs/{id}/prompt` to misclassify the job and reject follow-up prompts as worker-mode sandbox jobs.
5. With this fix, fresh ACP jobs now persist `job_mode = acp:<agent_name>` immediately.
6. Verified end to end on the fixed build:
   - first turn returned `GAMMA`
   - follow-up prompt queued successfully
   - second turn returned `DELTA`

The new regression test in this PR pins the create-time persistence behavior directly, without depending on a successful Docker run.

foltz:
Superseded by #2417.

This original PR was opened from a branch that accidentally included unrelated Matrix channel commits in addition to the ACP bugfix. #2417 is the clean replacement PR containing only:

- `Fix ACP sandbox job mode persistence`
- `Add ACP job mode persistence regression test`

Please review #2417 instead.

PR #2418, opened 2026-04-13
feat: slim mode runtime, Dockerfiles, and /health route

ymcrcat:
## Summary

Reduce per-instance resource footprint for high-density multi-tenant deployments, ship the Docker images that back them, and add a standard orchestrator health endpoint.

### 1. Configurable tokio runtime (`src/main.rs`, `src/bootstrap.rs`)
Use a **1-worker `multi_thread` runtime** when `TOKIO_WORKER_THREADS=1`; otherwise the tokio default. The worker count is passed through from the env var to the builder.

**Why 1-worker `multi_thread` and not `current_thread`?** An earlier revision used `new_current_thread()` — this silently breaks any code path that calls `tokio::task::block_in_place`, which panics on a current_thread runtime. Third-party channel and provider code (Signal-style pairing/identity flows being the motivating example) may rely on it, and the failure would only surface on first use in slim deployments. A single-worker multi_thread runtime has essentially the same resource profile (one event-loop thread, lazy blocking pool) but preserves `block_in_place` semantics.

Guarded by a regression test at `tests/runtime_build.rs` that builds a slim-mode runtime and calls `block_in_place` through it.

### 2. Lazy WASM epoch ticker (`src/tools/wasm/runtime.rs`)
The epoch ticker thread no longer spawns in `WasmToolRuntime::new()`. It now starts lazily on first module preparation via `ensure_ticker_started()`. Idle instances carry zero WASM-related background threads.

### 3. `/health` route on the web gateway (`src/channels/web/server.rs`)
Expose `/health` alongside the existing `/api/health` on the web gateway.

**Why both?**
- `/health` is the de facto convention for orchestrator probes: Kubernetes, Nomad, Docker HEALTHCHECK, AWS ALB, and most load balancers default to this path. The web gateway was the one channel in the codebase that did *not* expose it — every other channel (`src/channels/http.rs`, `src/channels/webhook_server.rs`, `src/orchestrator/api.rs`) already serves `/health`. This change brings the web gateway in line with the rest of the codebase and lets ironclaw deploy cleanly into any orchestrator without requiring a custom probe path.
- `/api/health` is kept because it is already referenced by tests, e2e conftest, and docs (9 places). Removing it would be a breaking change.

The two routes point at the same handler.

### 4. Slim Dockerfiles for high-density deployments
- **`Dockerfile.slim`** — multi-stage production build (~45 MB final image) with Cargo dependency caching. Produces the shard image used by sharded multi-tenant Nomad deployments.
- **`Dockerfile.worker.slim`** — slim sandbox-job worker (Python, Node.js, git; no Rust/Clang toolchain). The `.slim` suffix distinguishes it from the existing `Dockerfile.worker` which bundles a full dev environment.

Both Dockerfiles pair with the `TOKIO_WORKER_THREADS=1` slim runtime mode above: one worker thread, lazy WASM ticker, minimal image footprint.

## Test plan

- [x] `cargo check` — clean
- [x] `cargo test` — all new/modified tests pass; one pre-existing network-dependent failure (`tools::mcp::auth::tests::test_validate_url_safe_https`) reproduces on `main` and is unrelated to this change
- [x] `cargo test --test runtime_build` — slim-mode runtime supports `block_in_place` (regression guard)
- [x] `test_lazy_ticker_initialization` verifies the WASM ticker stays idle until first use
- [ ] Smoke test in slim mode (`TOKIO_WORKER_THREADS=1`) on a dev instance
- [ ] Confirm `/health` returns 200 from the web gateway (no auth required)
- [ ] Build and push `Dockerfile.slim` + `Dockerfile.worker.slim` to Artifact Registry; run under Nomad


src/tools/wasm/runtime.rs
```diff
@@ -207,32 +203,70 @@ impl WasmToolRuntime {
...
+            .compare_exchange(false, true, Ordering::SeqCst, Ordering::Acquire)
+            .is_err()
+        {
+            return;
+        }
+
+        match spawn() {
+            Ok(_) => tracing::debug!("Lazy WASM epoch ticker started"),
+            Err(e) => {
+                tracing::error!("Failed to spawn WASM epoch ticker thread: {}", e);
```
serrrfirat, line 263: **High Severity**

`try_start_ticker()` fails open when the epoch ticker cannot be spawned.

Before this PR, `WasmToolRuntime::new()` failed startup if the epoch ticker thread could not be created. Now a spawn error is only logged, the flag is reset, and `prepare()` continues after `ensure_ticker_started()` returns. That means under OS thread exhaustion a WASM tool can still be prepared/executed while epoch deadlines never advance. The outer `tokio::time::timeout` does not stop the `spawn_blocking` thread, so a stuck WASM execution can outlive the timeout if fuel/epoch trapping does not fire.

Please make `ensure_ticker_started()` / `try_start_ticker()` return `Result<(), WasmError>` and propagate it from `prepare()` before accepting the module. Untrusted WASM should fail closed unless the timeout backstop is running.

Dockerfile.worker.slim
```diff
@@ -0,0 +1,70 @@
...
+    && cargo build --release --bin ironclaw 2>&1 || true \
+    && rm -rf src crates/*/src
+
+COPY . .
+RUN cargo build --release --bin ironclaw
+
+# ── Runtime ──────────────────────────────────────────────────────────
+FROM debian:bookworm-slim
+
+RUN apt-get update && apt-get install -y --no-install-recommends \
```
serrrfirat, line 46: **High Severity**

The slim worker image does not install the `claude` CLI.

This image installs Node/npm, but it never installs `@anthropic-ai/claude-code`. The Claude bridge later spawns `claude` directly from inside the worker container, so any deployment that uses this image with `CLAUDE_CODE_ENABLED=true` will fail at runtime with a spawn error. The existing `Dockerfile.worker` installs the CLI explicitly, so this slim variant is missing a required runtime dependency for Claude Code jobs.

Please either install `@anthropic-ai/claude-code` here as well, or make this a clearly separate non-Claude worker image and prevent/document its use for `JobMode::ClaudeCode`.

Dockerfile.slim
```diff
@@ -0,0 +1,74 @@
...
+       done \
+    && cargo build --release --bin ironclaw 2>&1 || true \
+    && rm -rf src crates/*/src
+
+# Copy full source and build for real.
+COPY . .
+RUN cargo build --release --bin ironclaw
+
+# ── Phase 2: Runtime Environment ──────────────────────────────────────
+FROM debian:bookworm-slim
```
serrrfirat, line 47: **Medium Severity**

The production slim image runs as root.

The existing production Dockerfile creates a non-root `ironclaw` user before running the binary, but this final stage never switches away from root. For the high-density/multi-tenant image this increases blast radius for file writes, mounted volumes, and future container escape bugs.

Please create a dedicated user/home directory, set `HOME`, chown the IronClaw base directory as needed, and add `USER ironclaw` before the entrypoint.

Dockerfile.slim
```diff
@@ -0,0 +1,74 @@
...
+# ── Configuration for High-Density Scaling ───────────────────────────
+
+# Default to "Slim Mode" (Single-threaded runtime).
+# Nomad jobs can override this via their own 'env' blocks.
+ENV TOKIO_WORKER_THREADS=1
+ENV DATABASE_POOL_SIZE=1
+
+# Expose the standard IronClaw ports (HTTP and Gateway).
+EXPOSE 8080
+EXPOSE 9000
```
serrrfirat, line 68: **Medium Severity**

The image exposes the wrong gateway port.

The code default for the web gateway is still `GATEWAY_PORT=3000`, and the existing production Dockerfile exposes `3000`. This Dockerfile exposes `9000` without also setting `GATEWAY_PORT=9000`, so orchestrators or `docker run -P` users relying on image metadata will probe/map the wrong gateway port, including the new `/health` endpoint.

Please change this to `EXPOSE 3000`, or intentionally set `ENV GATEWAY_PORT=9000` and update the deployment docs/tests around that port change.

Issue #2419, opened 2026-04-13
fix(status): report active WASM channels accurately


G7CNF:
`ironclaw status` currently underreports WASM channel state. In my local runtime, `ironclaw channels list` shows an enabled WASM channel, but `ironclaw status` only reports `Channels cli, 1 wasm`, which does not identify the channel.

Observed gap:
- `ironclaw channels list` shows enabled WASM channel names
- `ironclaw status` does not surface those names
- the status line is too coarse to debug channel activation problems

Expected:
- `ironclaw status` should reflect active WASM channel names clearly
- at minimum, it should list enabled WASM channel names
- ideally, it should distinguish installed, enabled, and active/runtime-loaded

Why this matters:
- users cannot tell whether a channel like Telegram is actually live
- it makes setup debugging harder
- it creates confusion between config state and runtime state

Acceptance criteria:
- `ironclaw status` reports enabled WASM channel names
- if a WASM channel is configured but not active, status makes that explicit
- the output is consistent with `ironclaw channels list`

PR #2420, opened 2026-04-13
fix(status): report active WASM channels accurately

G7CNF:
Refs #2419

`ironclaw status` now shows enabled WASM channel names instead of only a coarse `1 wasm` summary.

Changes:
- `status` reports enabled WASM channel names when present
- falls back to installed WASM count when only the directory is known
- adds a CLI-level regression test that exercises `ironclaw status` end-to-end through the binary

### DIFF
--- a/src/cli/status.rs
+++ b/src/cli/status.rs
@@ -267,21 +267,20 @@ fn format_wasm_channels_summary(
     if settings.channels.wasm_channels_enabled && !settings.channels.wasm_channels.is_empty() {
         let mut channels = settings.channels.wasm_channels.clone();
         channels.sort();
-        let label = if channels.len() == 1 {
-            "1 wasm".to_string()
-        } else {
-            format!("{} wasm", channels.len())
-        };
-        return Some(format!("{} ({})", label, channels.join(", ")));
+        return Some(format!("{} wasm ({})", channels.len(), channels.join(", ")));
     }
 
     let wasm_count = count_wasm_files(channels_dir);
     if wasm_count > 0 {
-        return Some(format!(
-            "{} wasm installed ({})",
-            wasm_count,
-            channels_dir.display()
-        ));
+        if settings.channels.wasm_channels_enabled {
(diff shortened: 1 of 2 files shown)


henrypark133 (approved):
## Review: Net improvement to status output (Risk: Low)

The new `format_wasm_channels_summary` function is a clear improvement over the previous bare count — showing channel names when available makes the status command more useful. Good test coverage with both unit and integration tests.

Positives:
- Alphabetical sort produces deterministic, readable output
- Integration test (`tests/status_cli.rs`) calls the real binary with `IRONCLAW_BASE_DIR` isolation
- Uses `tempfile::tempdir()` correctly in unit tests
- No panics or unwraps in production code

### Concerning: `settings.channels.wasm_channels` is a wizard-tracking list, not the runtime authority [Architecture]
**File:** `src/cli/status.rs` — `format_wasm_channels_summary` Path 1

Per the doc comment on `wasm_channels` in `settings.rs`: "Channels not in this list but present in the channels directory will still load. This is primarily used by the setup wizard to track which channels were configured."

This means: (a) a manually-installed `.wasm` file won't appear in the names list, and (b) a deleted `.wasm` for a previously-configured channel will still show. The status command runs without a daemon connection so the settings file is the best available source — but users should understand it reflects "configured" not "running."

**Suggested fix:** Consider deriving names from `.wasm` filenames on disk (matching how the loader discovers channels) and cross-referencing the settings list as a secondary annotation. Not blocking — the current approach is strictly more informative than the old bare count.

### Minor: Two code paths untested [Testing]
**File:** `src/cli/status.rs` — unit tests

Covered: enabled+names (Path 1), disabled+files (Path 3).
Missing: enabled+empty-list+files (Path 2, the fresh-install default), and the None return (Path 4, no channels at all). Both are straightforward to add.

Convention notes:
- Integration test assertion `stdout.contains("telegram")` is correct but could be more specific (e.g., also asserting the "wasm" label appears) to distinguish from false positives.

LGTM.

Issue #2412, opened 2026-04-13
[QA] 'Potential secret leak detected' warnings flooding server logs


joe-rlo:
### Environment

Staging (hosted-staging)

### Version/Commit Hash

nearaidev/ironclaw-dind:0.25.0

### QA Test Date

2026-04-13

### Feature Area

Security / Logging

### Steps to Reproduce

1. Open the IronClaw staging environment at https://agent-stg.near.ai/
2. Perform normal bot operations (creating bots, running chats, using tools)
3. Review server-side logs for the staging environment
4. Observe repeated "Potential secret leak detected" warning messages

### Expected Behavior

Secret leak detection should only fire when actual secrets are at risk of exposure. Normal operations should not trigger these warnings. If detection is overly sensitive, the threshold or pattern matching should be tuned to reduce false positives.

### Actual Behavior

The server logs are flooded with "Potential secret leak detected" warnings during normal operation. These appear to be false positives triggered by routine bot activity rather than actual secret exposure. The volume of these warnings makes it difficult to find genuine issues in the logs.

### Logs/Screenshots

Observed in server logs during testing on 2026-04-13. The warnings appeared repeatedly across multiple bot interactions and obscured other potentially important log entries.

PR #2422, opened 2026-04-13
fix(safety): quiet warn-only leak scan noise

G7CNF:
Refs #2412

Reduce false-positive log noise from warn-only leak detection by:
- lowering warn-only leak matches from debug to trace
- avoiding detector debug spam in the web log broadcaster while preserving redaction/blocking

This keeps actual secret blocking and redaction intact, but stops ordinary high-entropy values from flooding server logs during normal operations.

### DIFF
--- a/crates/ironclaw_safety/src/leak_detector.rs
+++ b/crates/ironclaw_safety/src/leak_detector.rs
@@ -273,12 +273,12 @@ impl LeakDetector {
             });
         }
 
-        // Log warn-action matches at debug level (not warn!) to avoid
-        // corrupting REPL/TUI output. These are informational — real leaks
-        // use LeakAction::Redact which modifies the content silently.
+        // Warn-action matches are informational only. Keep them at trace so
+        // ordinary debug/server logging does not get flooded by common
+        // high-entropy values that are not actually secrets.
         for m in &result.matches {
             if m.action == LeakAction::Warn {
-                tracing::debug!(
+                tracing::trace!(
                     pattern = %m.pattern_name,
                     severity = %m.severity,
                     preview = %m.masked_preview,
--- a/src/channels/web/log_layer.rs
+++ b/src/channels/web/log_layer.rs
@@ -68,10 +68,18 @@ impl LogBroadcaster {
         // Scrub secrets from the message before it reaches any subscriber.
         // This is defense-in-depth: even if code elsewhere accidentally logs
         // a secret, it won't be broadcast to SSE clients.
-        entry.message = self
-            .leak_detector
-            .scan_and_clean(&entry.message)
-            .unwrap_or_else(|_| "[log message redacted: contained blocked secret]".to_string());
+        //
+        // Use the raw scan result here instead of scan_and_clean() so we avoid
+        // emitting the detector's warn-only debug logs for ordinary high-entropy
+        // values that appear frequently in operational logs.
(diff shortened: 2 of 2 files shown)


henrypark133 (changes requested):
## Review: Redaction logic correct, but observability regression and missing test coverage (Risk: High)

The switch from `scan_and_clean()` to raw `scan()` in `LogBroadcaster::send()` preserves the Block/Redact/Warn/clean redaction behavior correctly — the three-way branch is structurally equivalent. However, two issues need addressing before merge.

Positives:
- The manual branch logic correctly handles all four `LeakAction` combinations
- The motivation (reducing noise from high-entropy hex matches) is valid
- The new test for the Block path is a good addition at the caller level

### Concerning: Warn-action diagnostics completely silenced via two independent changes [Security/Architecture]
**File:** `crates/ironclaw_safety/src/leak_detector.rs:281` and `src/channels/web/log_layer.rs:68-79`

Two changes work together to eliminate the only non-blocking leak signal:

1. `scan_and_clean()` internally logged Warn matches at `debug!` — the new `send()` calls `scan()` directly and never inspects `scan.matches`, so Warn diagnostics are not emitted from the broadcaster path at all, regardless of log level.
2. The `debug!` → `trace!` change in `scan_and_clean()` makes Warn matches invisible even where `scan_and_clean()` is still called.

The existing `debug!` was already CLAUDE.md-compliant ("Use `debug!` for internal diagnostics"). If the goal is to quiet `high_entropy_hex` specifically, consider a pattern-name filter rather than silencing all Warn-action matches globally.

**Suggested fix:** Either keep calling `scan_and_clean()` (which handles warn-logging and redaction in one place), or add explicit match iteration in the new `send()` code path:
```rust
for m in &scan.matches {
    if m.action == LeakAction::Warn {
        tracing::debug!(pattern = %m.pattern_name, severity = %m.severity, preview = %m.masked_preview, "warn-only leak match in log broadcast");
    }
}
```

### Concerning: Missing test coverage for Redact and Warn paths through LogBroadcaster::send() [Testing]
**File:** `src/channels/web/log_layer.rs` test module

The new test covers only the Block path (`sk-proj-*` → sentinel string). Two critical paths through the new three-way branch are untested at the caller level:

1. **Redact path**: A message with `Bearer eyJ...` (20+ chars) should produce `[REDACTED]` in the broadcast — not the raw token. The `bearer_token` and `auth_header` default patterns use `LeakAction::Redact`.
2. **Warn path**: A 64-char hex string should pass through unchanged (no blocking, no redaction).
3. **Ring buffer consistency**: The test checks `rx.try_recv()` but not `broadcaster.recent_entries()` — if they diverge, late-joining SSE clients could see unredacted secrets.

**Suggested fix:** Add two more tests alongside `test_log_broadcaster_redacts_blocked_secret`:
- One sending a Bearer token and asserting the broadcast contains `[REDACTED]` (not the raw token)
- One sending a clean message and asserting passthrough
- Both should also check `recent_entries()` matches the broadcast

Convention notes:
- The `scan_and_clean()` → `scan()` refactor is a behavior change (not just a noise reduction) — the PR description should reflect this.

ilblackdragon:
Closing this because it updates the legacy top-level `src/` web logging implementation (`src/channels/web/log_layer.rs`). Product work has moved to the Reborn workspace under `crates/`, and this PR is now outdated against that architecture.

Thanks for the contribution.

Issue #2426, opened 2026-04-13
docs/setup: warn that Telegram open mode splits history from web UI


G7CNF:
Telegram works in either pairing or open DM mode, but open mode creates a split-identity experience: the bot can reply in Telegram while the web UI history sidebar remains empty because the conversation is not tied to the same IronClaw user identity.

This is easy to misconfigure during setup and looks like history sync is broken.

Proposed fix:
- During Telegram setup, explain that pairing is recommended when the user wants a conversation to continue in the browser history.
- Make the open mode tradeoff explicit: it is guest-style access and will not merge naturally with the web UI thread list.
- Optionally surface the current Telegram DM policy in channel status/settings.

This is a UX/discoverability issue, not a runtime correctness bug.

PR #2427, opened 2026-04-13
docs(setup): warn that Telegram open mode splits history

G7CNF:
Refs #2426

Telegram setup now prints a warning that pairing is the right choice if the user wants browser history continuity. Open mode keeps the bot usable in Telegram, but it can leave the web UI thread list empty because the exchange is not bound to the same IronClaw identity.

This is a discoverability fix only; it does not change Telegram runtime behavior.

### DIFF
--- a/src/setup/README.md
+++ b/src/setup/README.md
@@ -350,6 +350,10 @@ key first, then falls back to the standard env var.
 **Telegram special case** (`setup_telegram`):
 - Validates bot token via Telegram `getMe` API
 - Owner binding: polls `getUpdates` for 120s to capture sender's user ID
+- Pairing mode is the right choice if you want a Telegram conversation to
+  continue in the browser history sidebar. Open mode keeps the bot usable in
+  Telegram, but it creates a split identity that does not automatically merge
+  into the web UI thread list.
 - Optional webhook secret auto-generation for webhook mode
 
 **SecretsContext creation** (`init_secrets_context`):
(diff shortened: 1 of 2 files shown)


henrypark133 (approved):
## Review: Clean docs + UX improvement (Risk: Low)

Small, correct change that adds a helpful advisory during Telegram channel setup. No correctness or security concerns.

Positives:
- `eq_ignore_ascii_case` is the right comparison for channel names
- Test covers lowercase, mixed-case, and non-matching inputs
- README documentation matches the runtime behavior

### Minor: Hardcoded channel match in generic function [Architecture]
**File:** `src/setup/channels.rs:843`
`setup_mode_note` hardcodes a Telegram string match inside a generic `setup_wasm_channel` function. This works for one channel but would become an extensibility concern if more channels need notes.
**Suggested fix:** Consider adding an optional `advisory` field to `SetupSchema` / `capabilities.json` so channels can declare their own setup notes without modifying Rust code. Not blocking for a single branch.

### Minor: Note only shows when `required_secrets` is non-empty [Logic]
**File:** `src/setup/channels.rs:742`
`setup_wasm_channel` is only called when `!cap_file.setup.required_secrets.is_empty()` (in the wizard). If a Telegram config has no required secrets, the wizard short-circuits and this note is never shown. Worth verifying the Telegram capabilities file always has secrets.

Convention notes:
- Unit test covers the helper directly but not the call site in `setup_wasm_channel` — acceptable given the side effect is `println` and the helper is deterministic.
- README note is placed under "Telegram special case (`setup_telegram`)" heading, but Telegram actually goes through the generic WASM setup path. Consider moving to the "WASM channel setup" section.

LGTM.

PR #2428, opened 2026-04-13
feat(gate): persist "always approve" decisions to DB in v2 engine path

### LINKED ISSUE #2345: [QA] Tool install asks for permission repeatedly instead of remembering approval
## Environment
hosted-staging (crab shack) — https://agent-stg.near.ai/

## Version / Commit Hash

## QA Test Date
2026-04-11

## Feature Area
Agent → Tool approval / Permission management

## Steps to Reproduce
1. Create or open a bot on hosted-staging
2. Trigger a workflow that requires tool installation (e.g., setting up an MCP tool)
3. Observe the tool approval prompts in the chat

## Expected Behavior
After the user approves `tool_install` once, the agent should remember the approval and proceed without re-asking for the same permission within the session.

## Actual Behavior
The agent sends "Tool 'tool_install' requires approval. Reply 'yes' to approve, 'no' to deny." multiple times in succession. Even after the user approves, the prompt reappears, creating a frustrating loop of repeated permission requests.

This suggests the approval state is not being persisted or checked properly between sequential tool calls within the same session.

## Logs / Screenshots
Screenshot captured by jemartel during bug bash 4/11 showing two consecutive "Tool 'tool_install' requires approval" messages.

## Checklist
- [x] Steps are reproducible

henrypark133:
## Summary

- Add `persist_always_allow()` / `revert_always_allow()` helpers in `bridge/router.rs` to write `tool_permissions.{name} = AlwaysAllow` to the DB when users click "always approve" in the v2 engine path
- Defense-in-depth: tools declaring `ApprovalRequirement::Always` are never persisted (mirrors v1 `thread_ops.rs` logic)
- Uses `serde_json::to_value(PermissionState::AlwaysAllow)` for enum-coupled serialization (matches v1 pattern)
- Reverts the DB write if the resumed tool execution fails, restoring any prior preference
- Documents the composable `ApprovalGate` contract for future gate pipeline wiring

**The bug:** V2's `resolve_gate()` only stored "always approve" in-memory via `EffectBridgeAdapter::auto_approve_tool()` — users lost their choices on process restart. The v1 path (`thread_ops.rs`) already persisted correctly.

## Test plan

- [x] `test_persist_always_allow_writes_to_settings` — approving with "always" writes `AlwaysAllow` to DB
- [x] `test_persist_always_allow_skips_locked_tools` — `ApprovalRequirement::Always` tools are never persisted
- [x] `test_persist_always_allow_rejects_invalid_tool_name` — dotted/invalid tool names rejected
- [x] `test_revert_always_allow_deletes_setting` — revert deletes key when no prior value
- [x] `test_revert_always_allow_restores_prior_value` — revert restores pre-existing preference
- [x] `test_persist_falls_back_to_state_db` — falls back to `state.db` when `settings_store` is None (libsql feature-gated)
- [ ] Integration: drive `resolve_gate()` end-to-end (follow-up)
- [ ] E2E: Playwright "always approve" persistence across restart (follow-up)

Fixes #2345


henrypark133 (commented):
## Review: Persist "always approve" decisions to DB (Risk: High)

Closes a real v1/v2 parity gap — "always approve" decisions were lost on restart in the v2 path. Defense-in-depth guard has the right intent but flawed implementation. Revert-on-failure is a good transactional safeguard.

Positives:
- Clean separation: `persist_always_allow()` and `revert_always_allow()` as focused helpers
- 4 unit tests with `InMemorySettings` mock
- Mirrors v1 path structure

### Critical: Defense-in-depth check uses empty params — bypassed for param-dependent tools [Security]
**File:** `src/bridge/router.rs:368`

```rust
t.requires_approval(&serde_json::json!({}))
```

Several tools return different `ApprovalRequirement` based on actual params:
- **`shell`**: Returns `Always` only for `RiskLevel::High` commands (`rm -rf`, `sudo`). Empty params → `Medium` → `UnlessAutoApproved`. A persisted `AlwaysAllow` for `shell` survives restart and skips approval for ALL commands including destructive ones.
- **`skill_install`**: Returns `Always` only when `install_dependencies=true`. Empty params → `false` → `UnlessAutoApproved`.

The v1 path in `thread_ops.rs` has this same bug — this PR copied it faithfully.

**Fix:** Use actual pending params: `t.requires_approval(&pending.parameters)` instead of `serde_json::json!({})`. Consider a static deny list for tools that should never be persistable regardless of params.

### Critical: No validation on `action_name` before it becomes a settings key — key injection [Security]
**File:** `src/bridge/router.rs` — `format!("tool_permissions.{}", pending.action_name)`

`action_name` originates from the LLM-chosen tool name string. Nothing validates it's a well-formed identifier. If it contains dots, the resulting key could collide with or shadow other settings. `is_valid_admin_tool_name()` exists in `permissions.rs` (enforces alphanumeric/`_`/`-`, max 128 bytes) but is never applied here.

**Fix:** Validate `pending.action_name` with `is_valid_admin_tool_name()` before persisting. Refuse and log a security warning if validation fails.

### Critical: `revert_always_allow` blindly deletes — destroys pre-existing user preference [Logic]
**File:** `src/bridge/router.rs:423`

The revert unconditionally deletes `tool_permissions.<tool>`. If the user had a pre-existing `AlwaysAllow` from a prior session and tool execution fails, the revert destroys their long-standing preference silently.

**Fix:** Read the pre-existing value before persisting. On revert, restore the prior value instead of unconditional delete.

### Critical: `test_single_approval_does_not_persist` is vacuous [Testing]
**File:** `src/bridge/router.rs:5576`

This test only checks that a freshly-created `InMemorySettings` is empty — it never calls `persist_always_allow()` or `resolve_gate()`. It proves the mock starts empty, not that single-approval skips persistence. `persist_always_allow()` itself doesn't check the `always` flag — only the `if always` guard in `resolve_gate()` does, which this test doesn't exercise.

### Concerning: No caller-level test through `resolve_gate()` [Testing]
**File:** `src/bridge/router.rs:1595-1606`

Per "test through the caller" rule — all 4 tests call helpers directly. No test drives `resolve_gate(..., Approved { always: true })` end-to-end. Someone could delete the `persist_always_allow()` call and all tests would still pass.

### Concerning: `tool_upgrade` and `tool_auth` are persistable as AlwaysAllow [Security]
**File:** `src/tools/builtin/extension_tools.rs:641-643, 224-226`

Both return `UnlessAutoApproved` unconditionally, so the defense check won't block them. Persisting `AlwaysAllow` for `tool_upgrade` enables silent replacement of trusted WASM extensions with malicious binaries after restart — effectively a sandbox escape chain.

**Fix:** Audit which tools should be unconditionally excluded from AlwaysAllow persistence. Consider a static deny list for `tool_upgrade`, `tool_auth`, `skill_install`.

### Concerning: Race between concurrent resolve_gate calls [Logic]
Two concurrent approvals for the same tool/user: Thread A persists → fails → reverts, deleting Thread B's successfully-persisted AlwaysAllow.

### Concerning: `tracing::warn!` in gate-resolution internals [Convention]
**File:** `src/bridge/router.rs:375, 403, 424`

Three `warn!` calls in internal engine diagnostics. Per CLAUDE.md logging rules, these should be `tracing::debug!`.

### Concerning: Store fallback and error paths untested [Testing]
No test exercises: `settings_store=None` → `state.db` fallback, both stores `None` (silent no-op), or `set_setting`/`delete_setting` error paths.

Convention notes:
- `unwrap_or` on infallible `serde_json::to_value(AlwaysAllow)` is dead code — use `serde_json::json!("always_allow")` directly
- `legacy_registry_name` alias not persisted to DB (only added to in-memory set)
- SSE `GateResolved` broadcast fires before defense check, creating misleading UI state on failure
- `InMemorySettings::get_setting_full` always returns `Ok(None)` — incomplete mock

**Verdict: Request Changes** — Two security-critical findings (empty-params defense bypass enables persistent `shell` auto-approve; key injection via unvalidated `action_name`) plus a correctness bug (blind revert destroys pre-existing permissions). The v1 path shares the empty-params bug — both should be fixed together.

**Cross-PR note:** This PR is stacked on #2425 (CachedSettingsStore). The cache bypass issue flagged on #2425 (web settings API writes directly to DB, not cache) also affects this PR — if persist writes through the cache but web UI reads don't, or vice versa, the two sources of truth will diverge.

henrypark133:
## Addressing review findings

### Fixing (5 items):

1. **Empty params defense bypass** — Using `pending.parameters` instead of `json!({})`. The per-invocation gate still blocks `Always` commands regardless of persisted preference, but the defense check should use actual params to catch crafted clients sending `always:true` on a `shell rm -rf` invocation. The v1 path in `thread_ops.rs` has the same bug — filing separately.

2. **Key injection via unvalidated `action_name`** — Adding `is_valid_admin_tool_name()` validation before persisting. Reject and log if it fails.

3. **Blind revert destroys pre-existing preference** — Reading the existing value before persisting. On revert, restoring the prior value (or deleting only if there was none).

4. **`unwrap_or` on infallible serialization** — Replacing with `serde_json::json!("always_allow")` directly.

5. **Store fallback untested** — Adding a test for the `settings_store=None → state.db` fallback path.

### Pushing back (2 items):

6. **`tracing::warn!` → `debug!`** — These are not internal diagnostics. They fire on: (a) security violation (crafted client sending `always` for a locked tool), (b) DB write failure, (c) DB revert failure. These are genuinely abnormal conditions that ops should see. The CLAUDE.md guidance to use `debug!` applies to "trace analysis, reflection results, engine internals" — not security warnings or persistence failures.

7. **`tool_upgrade`/`tool_auth` persistable as AlwaysAllow** — Valid concern, but it's about tool classification (which tools return `ApprovalRequirement::Always` vs `UnlessAutoApproved`), not about the persistence mechanism. The fix belongs in those tools' `requires_approval()` impls or a static deny list. Filing as a follow-up issue.

### Noted (out of scope):

- **Vacuous test** — Already removed in previous commit.
- **No caller-level test through `resolve_gate()`** — Acknowledged. `resolve_gate` requires full ENGINE_STATE wiring. This should be an integration-tier test (`cargo test --features integration`). The helper-level tests still catch the primary bugs (wrong key format, missing defense check, blind delete).
- **Race between concurrent resolve_gate calls** — The save/restore approach for item 3 narrows the window. Full fix would need per-key CAS which is out of scope.
- **`legacy_registry_name` not persisted** — Intentional. The canonical tool name is what's persisted; legacy aliases are an in-memory convenience for the current session.
- **SSE broadcast before defense check** — SSE fires for gate resolution (approve/deny), not for persistence. The gate IS resolved regardless of whether persistence succeeds.

src/bridge/router.rs
```diff
@@ -351,6 +351,85 @@ async fn insert_and_notify_pending_gate(
...
+        Some(ss) => ss.as_ref(),
+        None => match &state.db {
+            Some(db) => db.as_ref(),
+            None => return,
+        },
+    };
+
+    let key = format!("tool_permissions.{}", pending.action_name);
+    let val = serde_json::to_value(crate::tools::permissions::PermissionState::AlwaysAllow)
+        .unwrap_or(serde_json::Value::String("always_allow".into()));
```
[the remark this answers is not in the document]
henrypark133: Replacing with `serde_json::json!("always_allow")` directly — avoids both the `unwrap_or` dead code and the `.expect()` which is banned in production per project rules.

henrypark133 (commented):
## Review: Persist "always approve" decisions to DB in v2 engine path (Risk: Medium)

Good v1/v2 parity feature with solid defense-in-depth in Patch 3 (parameter validation, tool name validation, prior value preservation). A few structural items to address.

### Concerning: `persist_always_allow` belongs in `src/gate/`, not `bridge/router.rs`
**File:** `src/bridge/router.rs`
`router.rs` is 5600+ lines. The v1 path persists in `thread_ops.rs`, the v2 path now persists in `router.rs` — two different locations for the same semantics. Approval persistence should live in `src/gate/settings.rs` as a shared function callable from both paths. This prevents a third location from appearing when a new entry point is added.

### Concerning: CachedSettingsStore cache coherence
The write in `persist_always_allow` and the read in `effect_adapter.rs` (line ~784) must go through the same `CachedSettingsStore` instance. If `persist_always_allow` falls back to raw `Database` (when `settings_store` is None), the write bypasses cache invalidation — the newly-persisted AlwaysAllow will be invisible to subsequent tool executions until the cache is invalidated by some other write. Consider adding `settings_store` to `EngineState` to ensure both read and write paths share the same cache instance.

### Concerning: Persist-before-execute semantics
The persist→execute→revert pattern means a process crash between persist and execute leaves a permanent AlwaysAllow in the DB. Consider either:
- Moving DB persist to after successful execution (removes revert entirely)
- Or accepting that "always" should survive crashes (user explicitly chose "always")
The current middle ground (persist then revert on error) is not panic/cancellation-safe.

### Concerning: `warn!` in persist/revert paths violates TUI logging rule
**Files:** `src/bridge/router.rs` (persist_always_allow, revert_always_allow)
CLAUDE.md: "`info!` and `warn!` output appears in the REPL and corrupts the terminal UI." These are internal diagnostics — downgrade to `debug!`.

### Concerning: Tests exercise helper directly, not through `resolve_gate` caller
The `InMemorySettings` tests drive `persist_always_allow` and `revert_always_allow` directly. Per "Test Through the Caller" rule, a test should drive `resolve_gate` with `always: true` and assert the settings key is persisted. This would catch bugs where `resolve_gate` passes wrong parameters.

### Minor: `is_valid_admin_tool_name` rejection is silent
When the validator rejects a tool name, `persist_always_allow` returns `None` with no log. Add a `debug!` log so the behavior is observable during development.

### Minor: TOCTOU between prior-value read and overwrite
`get_setting` then `set_setting` is a read-modify-write without a transaction. Low practical risk in single-user system but noted for correctness.

Positives:
- Defense-in-depth: ApprovalRequirement::Always tools correctly blocked from persistence (Patch 3)
- Tool name validation via `is_valid_admin_tool_name` prevents settings key injection (Patch 3)
- Prior value save/restore on revert is cleaner than blind delete (Patch 3)
- Good test coverage: locked tool skip, invalid name rejection, DB fallback, prior value restoration

src/bridge/router.rs
```diff
@@ -5199,4 +5284,341 @@ mod tests {
...
+    async fn test_single_approval_does_not_persist() {
+        let settings = Arc::new(InMemorySettings::new());
+
+        // Verify the store is empty before any persist call.
+        let val = settings.get("user1", "tool_permissions.shell").await;
+        assert!(
+            val.is_none(),
+            "No settings should exist before persist is called"
+        );
+    }
```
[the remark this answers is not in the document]
henrypark133: Removed in 0db73bbb — the test was a no-op as noted.

henrypark133:
Re: [review 4102594525](https://github.com/nearai/ironclaw/pull/2428#pullrequestreview-4102594525) — addressed what's actionable, deferring the rest:

**Fixed:**
- **`warn!` → `debug!`**: All 5 call sites in `persist_always_allow` / `revert_always_allow` downgraded. Pushed.

**Already fixed in prior patches:**
- Empty params bypass → uses `&pending.parameters` since f2d5fa9c
- Tool name validation → `is_valid_admin_tool_name()` guard since f2d5fa9c
- Blind revert → reads/restores prior value since f2d5fa9c
- `unwrap_or` dead code → `json!("always_allow")` since f2d5fa9c
- Vacuous test → removed in 0db73bbb

**Deferred (follow-up):**
- Move to `src/gate/settings.rs` — requires migrating v1 path too, bigger scope
- Cache coherence — when `settings_store` is None, read path also uses raw DB, so consistent in practice
- Persist-before-execute — current revert-on-failure is more conservative than either suggested alternative; design discussion
- Caller-level tests through `resolve_gate` — needs significant test infrastructure
- `tool_upgrade`/`tool_auth` deny list — separate policy decision
- TOCTOU / concurrent race — low risk in single-user system

ilblackdragon:
## Review

Functionally correct parity fix for a real UX bug. A few ship-blockers and some security follow-ups.

### PR description is misleading
The summary claims "prefers `CachedSettingsStore` for write-through cache invalidation." No such type exists in this diff, and `grep` finds zero references to `CachedSettingsStore` elsewhere (that lives in #2425). Either the description was copy-pasted from another PR or a commit was dropped. Please reconcile — if #2425 is a prereq, mark it as such.

### Security-relevant logging
`router.rs:422, 432, 472` — DB persistence failures logged at `debug!`. A silent persistence failure on an approval escalation ("always allow") is **security-relevant**: the in-memory grant from `auto_approve_tool` (router.rs:1519) remains even if the DB write fails, so the user perceives it as "remembered forever" when it'll vanish on restart. Upgrade to `warn!` to match v1 (`thread_ops.rs:1256`).

Worse: the **revert path** at 1536-1546 only runs on *execution* failure, not on *persist* failure. Either fail-closed the approval if persistence fails, or document that in-memory grant is authoritative for the current process.

### Serialization drift
`router.rs:417` — `serde_json::json!("always_allow")` hardcoded. v1 uses `serde_json::to_value(PermissionState::AlwaysAllow)` which is coupled to the enum's serde rename. Use the same pattern here so a future enum rename doesn't silently break only v2.

### Param-unaware grants (parity with v1, but flag)
`requires_approval(&pending.parameters)` is param-aware; the persisted grant is tool-name-only. Approve `shell ls /tmp` once with "always," and `shell rm -rf /` later runs silently. Parity with v1, not a regression, but this PR is a good moment to file a follow-up issue. Consider tagging grants with `{state, granted_at, tool_version}` so we can invalidate on tool-semantics change.

### Audit trail gap
This write bypasses `ToolDispatcher::dispatch()` and the existing `tool_permission_set` tool (`src/tools/builtin/extension_tools.rs:723`). No `ActionRecord` is produced for a persisted permission change. Either route through `tool_permission_set`, or add `// dispatch-exempt: <reason>` on the write lines — the current exemption isn't clearly justified.

### v2 coverage
This *is* the v2 path — good. Please confirm v1 behavior is unchanged (parity tests).

### Tests — add integration, not just helper
Helper-level tests (6) are good. Missing per `.claude/rules/testing.md`:
- **Integration** (`cargo test --features integration`): drive `resolve_gate()` end-to-end with a real `PendingGate` + real `SettingsStore`, assert (a) persistence on `Approved { always: true }` and (b) revert-to-prior on execution failure.
- **Integration**: subsequent `tool_permission_set` → `ask_each_time` correctly overrides the persisted `AlwaysAllow`.
- **Playwright** under `tests/e2e/`: user clicks "always approve" on a gate, restart daemon, trigger the same tool — assert no new gate prompt. This is the primary user-visible regression test for the feature.
- **Playwright**: user revokes via settings UI — assert the gate reappears on next call.

### Other
- `src/gate/approval.rs:18-24` — the docstring "v2 engine should do the same when this gate replaces the inline check" now gets wired by this PR. Update the doc.

serrrfirat:
## Paranoid Architect Review — Additional Findings

### 1. **Medium Severity — `set_setting` failure silently loses user's "always approve" intention** (`src/bridge/router.rs:429-443`)

When `set_setting()` fails (line 435), the function logs at `debug!` and returns `prior` — proceeding as if the write succeeded. The user's "always approve" preference works for the current session (via in-memory `auto_approve_tool`), but is never persisted to the DB, so it vanishes on restart without any indication. Additionally, `revert_always_allow` will attempt an unnecessary revert against a DB row that was never written.

**Suggested fix:** Return a sentinel or `None` on write failure so the caller knows the DB was never modified. At minimum, log at `warn!` (this runs in the web gateway path, not the TUI, so it won't corrupt the terminal — consistent with the v1 path in `thread_ops.rs:1256`).

---

### 2. **Medium Severity — Missing feature gate on `test_persist_falls_back_to_state_db`** (`src/bridge/router.rs:5741`)

This test calls `crate::testing::test_db()` which is `#[cfg(feature = "libsql")]`. The test itself has no feature gate — it compiles only because `libsql` is a default feature. Building with `--no-default-features --features postgres` would fail to compile.

**Suggested fix:** Add `#[cfg(feature = "libsql")]` attribute to `test_persist_falls_back_to_state_db`.

henrypark133:
Re: @ilblackdragon [review](https://github.com/nearai/ironclaw/pull/2428#issuecomment-4241766454) — all items addressed:

**PR description is misleading** — Updated. Removed the `CachedSettingsStore` reference; the description now accurately describes what this PR does.

**Security-relevant logging** — Upgraded failure logs from `debug!` to `tracing::warn!` in both `persist_always_allow` and `revert_always_allow`. The v2 path runs from the web gateway (not TUI), so `warn!` won't corrupt the terminal. Matches v1 at `thread_ops.rs:1256`. Success/early-return guards stay at `debug!`.

**Serialization drift** — Fixed. Now uses `serde_json::to_value(PermissionState::AlwaysAllow).unwrap_or(json!("always_allow"))` — coupled to the enum's `#[serde(rename_all = "snake_case")]`.

**Param-unaware grants** — Filed as #2484 (follow-up, parity with v1).

**Audit trail gap** — Added `// dispatch-exempt: engine-internal persist mirrors v1 thread_ops write-through` comments on all direct `set_setting`/`delete_setting` call sites.

**Update approval.rs docstring** — Updated to distinguish persistence from hydration, per additional Copilot feedback. The docstring now accurately describes where each happens.

**Tests** — Added:
- 3 e2e Playwright tests (DB persist, PUT revocation, survive-restart via restartable server fixture)
- Unit test for `Settings::from_db_map` tool_permissions parsing
- Updated `test_persist_falls_back_to_state_db` → `test_persist_skips_when_no_settings_store`

**Additional fix found during e2e development:** `persist_always_allow` had a raw Database fallback that bypassed `CachedSettingsStore` cache invalidation, causing `GET /api/settings/tools` to serve stale data. Removed the fallback — in production `agent.deps.settings_store` is always available when the DB is.

**Security fix from Copilot review:** `resolve_gate` now downgrades `always` to `false` when the pending gate's `ResumeKind` has `allow_always: false`. Previously a crafted client could send `always:true` on an `ApprovalRequirement::Always` tool and the in-memory `auto_approve_tool` would be set.

henrypark133:
Re: @serrrfirat [review](https://github.com/nearai/ironclaw/pull/2428#issuecomment-4242454233) — both items addressed:

**1. `set_setting` failure silently loses user's intention** — Fixed. Failure logs upgraded from `debug!` to `tracing::warn!`. The v2 path runs from the web gateway (not TUI), so `warn!` is safe and consistent with the v1 path at `thread_ops.rs:1256`.

**2. Missing feature gate on `test_persist_falls_back_to_state_db`** — The test has been replaced entirely. The raw Database fallback path was removed (it bypassed `CachedSettingsStore` cache invalidation — a real bug found during e2e testing). The new test `test_persist_skips_when_no_settings_store` verifies the correct behavior: when `settings_store` is `None`, `persist_always_allow` returns early without writing.

src/gate/approval.rs
```diff
@@ -15,6 +15,16 @@ use crate::tools::{ApprovalRequirement, ToolRegistry};
 /// Gate that checks `Tool::requires_approval()` and emits `Pause(Approval)`
 /// or `Deny` depending on execution mode.
 ///
+/// **DB-persisted permissions:** This gate checks `ctx.auto_approved` for
+/// session-scoped auto-approvals. The caller constructing [`GateContext`] is
+/// responsible for pre-populating that set from DB-persisted
+/// `PermissionState::AlwaysAllow` entries (via `effective_permission()`).
+/// The v1 dispatcher does this at `dispatcher.rs` turn start; the v2 engine
+/// does this via `persist_always_allow()` in `bridge/router.rs`.
```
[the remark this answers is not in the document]
henrypark133: Fixed in 97e24263 — docstring now distinguishes persistence (persist_always_allow writes the setting) from hydration (EffectBridgeAdapter::auto_approved is the runtime set). Thanks for catching this.

src/bridge/router.rs
```diff
@@ -1513,15 +1648,21 @@ pub async fn resolve_gate(
...
                     .await;
                 if let Some(ref registry_name) = legacy_registry_name {
                     state.effect_adapter.auto_approve_tool(registry_name).await;
                 }
-            }
+
+                // Persist AlwaysAllow to DB so the preference survives process
+                // restarts. Mirrors the v1 path in thread_ops.rs.
+                persist_always_allow(agent, state, &pending).await
+            } else {
```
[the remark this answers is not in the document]
henrypark133: Valid catch — fixed in 97e24263. `resolve_gate` now downgrades `always` to `false` when `pending.resume_kind` doesn't have `allow_always: true`:

```rust
let always = raw_always
    && matches!(
        pending.resume_kind,
        ironclaw_engine::ResumeKind::Approval { allow_always: true }
    );
```

`persist_always_allow` already had a defense-in-depth check against `ApprovalRequirement::Always` tools, but the in-memory `auto_approve_tool` path at line 1654 was unguarded. This fixes both paths.

src/bridge/router.rs
```diff
@@ -351,6 +351,141 @@ async fn insert_and_notify_pending_gate(
...
+        },
+    };
+
+    let key = format!("tool_permissions.{}", pending.action_name);
+
+    // Read the pre-existing value so we can restore it on failure instead
+    // of blindly deleting a long-standing user preference.
+    let prior = match store.get_setting(&pending.user_id, &key).await {
+        Ok(v) => v,
+        Err(e) => {
```
serrrfirat, line 419: **Medium Severity — TOCTOU race between read-prior and write-AlwaysAllow**

The read-then-write in `persist_always_allow` is not atomic. Two concurrent `resolve_gate` calls for the same tool/user could both read `prior=None`, both write `AlwaysAllow`, then if the first execution fails and reverts (calls `delete_setting`), the second user's legitimate AlwaysAllow is lost.

**Mitigations already in place:**
- `PendingGateStore.take_verified()` atomically removes the pending gate, so the same gate can't be resolved twice
- The race window requires human-initiated concurrent approvals (narrow)
- The settings table uses `ON CONFLICT DO UPDATE` for idempotent writes

This matches the v1 precedent in `thread_ops.rs:1291-1311` and is acceptable risk, but a code comment documenting the race and its mitigations would help future readers.

**Note:** v2's `persist_always_allow` is actually more correct than v1 — it uses actual pending parameters for `requires_approval()` (catching destructive shell commands), while v1 passes empty params.

_Paranoid PR Review_

serrrfirat (approved):
## Paranoid Architect Review — APPROVED ✅

**v2 Engine Impact: Already addressed** — this PR IS the v2 fix

Solid implementation. The core `persist_always_allow`/`revert_always_allow` pattern is well-structured with proper defense-in-depth:

- **ApprovalRequirement::Always tools blocked from persistence** (even if crafted client sends `always:true`)
- **Tool name validation** via `is_valid_admin_tool_name` prevents settings namespace injection
- **Prior value preservation** on revert — doesn't blindly delete long-standing user preferences
- **Fallback chain** (settings_store → state.db → silent no-op) degrades gracefully

**v2 is actually more correct than v1** in two ways:
1. Validates tool names before using them as DB keys (v1 doesn't)
2. Uses actual pending parameters for `requires_approval()` check — catches destructive shell commands that v1 misses (v1 passes empty params)

### 1 inline comment posted (Medium):
- TOCTOU race between read-prior and write-AlwaysAllow — acceptable risk given `PendingGateStore.take_verified()` atomicity and narrow human-initiated window. Matches v1 precedent. Recommend adding a code comment.

Test coverage is good (6 tests covering persist, skip-locked, revert-delete, revert-restore, invalid-name-rejection, and libsql fallback).

src/bridge/router.rs
```diff
@@ -351,6 +351,139 @@ async fn insert_and_notify_pending_gate(
...
+/// Defense-in-depth: tools that declare `ApprovalRequirement::Always` for
+/// the actual pending parameters are never persisted (the UI hides the
+/// button, but a crafted client could send it). Tool names are validated
+/// before use as settings keys.
+///
+/// Returns the pre-existing permission value (if any) so the caller can
+/// restore it on failure via [`revert_always_allow`].
+async fn persist_always_allow(
+    agent: &Agent,
+    state: &EngineState,
```
serrrfirat, line 365: **High Severity** — Hydration gap: persisted permissions never loaded on v2 startup

`persist_always_allow()` correctly writes `tool_permissions.{name} = AlwaysAllow` to the DB, but on process restart nothing reads it back. `EffectBridgeAdapter::auto_approved` starts as an empty `HashSet` (effect_adapter.rs:72) and is never populated from DB-persisted entries.

The v1 dispatcher handles this in `build_tool_list()` (dispatcher.rs:480-530), which reloads permissions from `CachedSettingsStore` at each turn start. The v2 bridge/engine has no equivalent path.

**Impact:** The DB write is effectively write-only — `test_always_approve_survives_restart` will fail because the new process starts with an empty approval set.

**Suggested fix:** Add a hydration step in v2 engine initialization (or at GateContext construction time) that loads `tool_permissions.*` entries from the settings store into `auto_approved`:
```rust
// In EffectBridgeAdapter or engine state init:
if let Some(ref ss) = agent.deps.settings_store {
    let all = ss.get_all_settings(&user_id).await?;
    for (k, v) in &all {
        if let Some(tool) = k.strip_prefix("tool_permissions.") {
            if v == &serde_json::json!("always_allow") {
                effect_adapter.auto_approve_tool(tool).await;
            }
        }
    }
}
```

tests/e2e/scenarios/test_v2_engine_approval_flow.py
```diff
@@ -466,5 +466,192 @@ async def test_approval_always(self, v2_approval_server):
             f"Got {len(turns)} turns."
         )
 
+    async def test_always_approve_persists_to_db(self, v2_approval_server):
```
serrrfirat, line 469: **High Severity** — Test ordering dependency

This test assumes `test_approval_always` already ran and wrote `always_allow` to the DB (same module-scoped `v2_approval_server` fixture). While pytest preserves definition order by default, this is fragile:

- `pytest-randomly` or `pytest-xdist` will break it
- If `test_approval_always` fails, this test will also fail with a misleading assertion error
- The implicit dependency isn't enforced by the test framework

**Suggested fix:** Either merge this assertion into `test_approval_always` itself, or have this test perform its own "always" approval before checking DB state. A `@pytest.mark.dependency` decorator would also work if the project uses that plugin.

src/bridge/router.rs
```diff
@@ -1490,7 +1623,18 @@ pub async fn resolve_gate(
...
 
     match resolution {
-        ironclaw_engine::GateResolution::Approved { always } => {
+        ironclaw_engine::GateResolution::Approved { always: raw_always } => {
+            // Downgrade `always` when the pending gate didn't offer the
+            // "always approve" option.  A crafted client could send
+            // `always:true` for an `ApprovalRequirement::Always` tool;
+            // without this guard the in-memory `auto_approve_tool` would
+            // be set, silently bypassing future approval prompts for
+            // parameter combinations that normally require them.
```
serrrfirat, line 1632: **Medium Severity** — No test for the `always` downgrade security fix

This is the PR's key security change — downgrading a crafted `always:true` when the pending gate has `allow_always: false`. But no unit or integration test verifies this guard.

Per `.claude/rules/testing.md` ("Test Through the Caller, Not Just the Helper"), this is exactly the pattern that needs a caller-level test: the downgrade is a predicate that gates side effects (`auto_approve_tool`, `persist_always_allow`), and it sits between user input and those effects.

**Suggested test:**
```rust
#[tokio::test]
async fn test_always_downgraded_when_allow_always_false() {
    // Set up a PendingGate with:
    //   resume_kind: ResumeKind::Approval { allow_always: false }
    // Send GateResolution::Approved { always: true }
    // Assert: auto_approved set does NOT contain the tool
    // Assert: DB has no tool_permissions.{name} entry
}
```

src/bridge/router.rs
```diff
@@ -351,6 +351,139 @@ async fn insert_and_notify_pending_gate(
...
+            )
+        })
+        .unwrap_or(false);
+
+    if is_locked {
+        debug!(
+            tool = %pending.action_name,
+            "Skipping AlwaysAllow persist — tool declares ApprovalRequirement::Always"
+        );
+        return None;
```
serrrfirat, line 400: **Medium Severity** — DB write failure is silent to the caller

If `store.set_setting()` fails here, the function logs `warn!` but still returns `prior` as if the write succeeded. Meanwhile, `auto_approve_tool()` was already called (line ~1660), so the tool auto-approves for this session — but the preference won't survive a restart. The user gets no indication that persistence failed.

This matches v1's fire-and-forget pattern, so it may be intentional. If so, consider adding a brief comment documenting the design choice: "in-memory approval is the primary path; DB persistence is best-effort for cross-restart survival."

If it's not intentional, consider returning `Result<Option<Value>, _>` so the caller can emit a user-visible SSE warning.

tests/e2e/scenarios/test_v2_engine_approval_flow.py
```diff
@@ -466,5 +466,192 @@ async def test_approval_always(self, v2_approval_server):
...
+
+    await start()
+    try:
+        yield {"base_url": base_url, "start": start, "stop": stop}
+    finally:
+        await stop()
+
+
+async def test_always_approve_survives_restart(restartable_v2_server):
+    """After 'always' approval and process restart, the tool auto-approves without a gate."""
```
serrrfirat, line 600: **Medium Severity** — This E2E test will fail without the hydration path

This test restarts the server and expects the new process to auto-approve without a gate prompt (line `assert "requires approval" not in all_responses`). But the v2 engine doesn't load DB-persisted permissions on startup — `EffectBridgeAdapter::auto_approved` starts empty.

This is a consequence of the hydration gap noted on `persist_always_allow()`. Either:
1. Ship the hydration code in this PR so the test passes, or
2. Mark this test as `@pytest.mark.skip(reason="Requires hydration path — see #XXXX")` and file a follow-up issue

PR #2442, opened 2026-04-14
feat(testing): Claude Code trace import + worktree-based replay

ilblackdragon:
## Summary

Adds tooling to convert Claude Code JSONL history into IronClaw trace fixtures, then replay them deterministically against a git worktree checked out at the historical commit. Tests real tool execution (read_file, apply_patch, shell, grep, glob) against the exact repo state that existed when the LLM made its original decisions — the LLM is mocked, everything else is real.

## What's in the box

**Converter** — `scripts/convert_claude_trace.py`
- Parses the Claude Code JSONL tree: temporal order, streaming chunks merged by API message ID, parallel tool calls grouped per-response
- Tool mapping: \`Bash→shell\`, \`Read→read_file\`, \`Write→write_file\`, \`Edit→apply_patch\`, \`Grep→grep\`, \`Glob→glob\` (unmappable tools like \`Agent\`, \`TaskCreate\`, etc. are silently skipped)
- Resolves git commits from \`gitBranch\` + \`timestamp\` metadata on each message
- Rewrites absolute paths to \`{{repo_root}}\` placeholders
- Embeds \`repo\` metadata block (commit, branch, branch_commits) in the output trace

**Tool sandboxing** — \`ToolRegistry::register_dev_tools_in_dir(dir)\`
- Binds all dev tools to a directory: \`base_dir\` for file/grep/glob tools, \`working_dir\` for shell
- File operations cannot escape the sandbox (existing path_utils enforcement)

**TestRig integration** — \`TestRigBuilder::with_working_dir(path)\`
- Routes through \`register_dev_tools_in_dir\` when set

**Trace format extensions** — \`LlmTrace\`
- New \`repo\` field (TraceRepoMeta) carrying commit/branch
- New \`resolve_repo_root(worktree_path)\` method for placeholder substitution at load time

**Worktree replay test** — \`tests/e2e_worktree_replay.rs\`
- Creates a \`git worktree add --detach <commit>\` in a temp dir, resolves placeholders against it, runs the trace, verifies tool outputs via \`expects\`, and cleans up via a Drop guard
- Skips gracefully when the source repo isn't present (CI-safe)

## What this actually tests

| Layer | Real or Mocked? |
|---|---|
| LLM responses | Mocked (TraceLlm replay) |
| Tool dispatch, safety pipeline | **Real** |
| File I/O, shell, grep, glob, apply_patch | **Real** — executed against the worktree |
| Repo state | **Real** — git worktree at the recorded commit |

Proven working: the included \`worktree_read_test.json\` reads \`src/transcription/mod.rs\` from commit \`a09c0236\` (the state where that file still existed before being moved) and asserts the content contains \`TranscriptionProvider\`.

## Why this is a draft

- Imported trace fixtures (~100KB) contain content from my local Claude Code sessions — may want to trim or gitignore them depending on reviewer preference
- The worktree replay is hardcoded to \`/Users/coder/ironclaw\` as the source repo — needs to be configurable or resolved via a workspace-relative path before merging
- Only one smoke test wired up so far; the larger imported traces reference historical commits that exist in our repo history but haven't been wired to tests yet

## Test plan
- [x] \`cargo test --test trace_format\` — 9/9 pass (includes imported trace deserialization)
- [x] \`cargo test --test e2e_worktree_replay\` — passes locally (reads real file from worktree, asserts content)
- [x] \`cargo clippy --tests -- -D warnings\` — zero warnings
- [x] \`cargo fmt\` applied
- [ ] Decide fate of \`tests/fixtures/llm_traces/imported/*.json\` (keep / trim / gitignore)
- [ ] Make \`SOURCE_REPO\` configurable (env var or workspace discovery)

### DIFF
--- a/scripts/convert_claude_trace.py
+++ b/scripts/convert_claude_trace.py
@@ -1,0 +1,912 @@
+#!/usr/bin/env python3
+"""Convert Claude Code JSONL conversation history to IronClaw trace fixtures.
+
+Takes a Claude Code session JSONL file and produces one or more IronClaw
+trace fixture JSON files suitable for deterministic E2E replay via TraceLlm.
+
+Usage:
+    python3 scripts/convert_claude_trace.py <session.jsonl> [--output-dir <dir>]
+    python3 scripts/convert_claude_trace.py <session.jsonl> --turn 2  # single turn
+    python3 scripts/convert_claude_trace.py <session.jsonl> --list    # list turns
+    python3 scripts/convert_claude_trace.py <session.jsonl> --all     # one file per turn
+
+Tool mapping (Claude Code -> IronClaw):
+    Bash   -> shell       (command -> command)
+    Read   -> read_file   (file_path -> path, offset, limit)
+    Write  -> write_file  (file_path -> path, content)
+    Edit   -> apply_patch (file_path -> path, old_string, new_string, replace_all)
+    Grep   -> grep        (pattern, path, glob, output_mode, context, ...)
+    Glob   -> glob        (pattern, path)
+
+Unmappable tools (Agent, ToolSearch, Skill, Task*, etc.) are skipped.
+Turns that contain only unmappable tools are excluded.
+"""
+
+import argparse
+import json
+import os
+import sys
+from collections import defaultdict
+from pathlib import Path
+from typing import Any
+
+# ── Tool mapping ──────────────────────────────────────────────────────
+
+TOOL_MAP: dict[str, str] = {
+    "Bash": "shell",
+    "Read": "read_file",
+    "Write": "write_file",
+    "Edit": "apply_patch",
+    "Grep": "grep",
+    "Glob": "glob",
+}
+
+# Tools we silently skip (no IronClaw equivalent or not useful in replay)
+SKIP_TOOLS: set[str] = {
+    "Agent",
+    "ToolSearch",
+    "Skill",
+    "TaskCreate",
+    "TaskUpdate",
+    "TaskGet",
+    "TaskList",
+    "TaskOutput",
+    "TaskStop",
+    "ScheduleWakeup",
+    "AskUserQuestion",
+    "EnterPlanMode",
+    "ExitPlanMode",
+    "EnterWorktree",
+    "ExitWorktree",
+    "CronCreate",
+    "CronDelete",
+    "CronList",
+    "NotebookEdit",
+    "LSP",
+    "Monitor",
+    "RemoteTrigger",
+    "WebSearch",
+    "SendMessage",
+}
+
+
+def map_tool_args(claude_name: str, claude_input: dict[str, Any]) -> dict[str, Any] | None:
+    """Map Claude Code tool arguments to IronClaw tool arguments.
+
+    Returns None if the tool is unmappable.
+    """
+    ic_name = TOOL_MAP.get(claude_name)
+    if ic_name is None:
+        return None
+
+    if ic_name == "shell":
+        args: dict[str, Any] = {"command": claude_input.get("command", "")}
+        if "timeout" in claude_input:
+            # Claude Code uses milliseconds, IronClaw uses seconds
+            timeout_ms = claude_input["timeout"]
+            if isinstance(timeout_ms, (int, float)) and timeout_ms > 0:
+                args["timeout"] = max(1, int(timeout_ms / 1000))
+        return args
+
+    if ic_name == "read_file":
+        args = {"path": claude_input.get("file_path", "")}
+        if "offset" in claude_input:
+            args["offset"] = claude_input["offset"]
+        if "limit" in claude_input:
+            args["limit"] = claude_input["limit"]
+        return args
+
+    if ic_name == "write_file":
+        return {
+            "path": claude_input.get("file_path", ""),
+            "content": claude_input.get("content", ""),
+        }
+
+    if ic_name == "apply_patch":
+        args = {
+            "path": claude_input.get("file_path", ""),
+            "old_string": claude_input.get("old_string", ""),
+            "new_string": claude_input.get("new_string", ""),
+        }
+        if claude_input.get("replace_all"):
+            args["replace_all"] = True
+        return args
+
+    if ic_name == "grep":
+        args = {"pattern": claude_input.get("pattern", "")}
+        if "path" in claude_input:
+            args["path"] = claude_input["path"]
+        if "glob" in claude_input:
+            args["glob"] = claude_input["glob"]
+        if "output_mode" in claude_input:
+            args["output_mode"] = claude_input["output_mode"]
+        # Map context flags
+        for flag in ("context", "-C", "-A", "-B"):
+            if flag in claude_input:
+                ic_flag = {
+                    "context": "context",
+                    "-C": "context",
+                    "-A": "after_context",
+                    "-B": "before_context",
+                }.get(flag, flag)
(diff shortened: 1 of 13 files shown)


scripts/convert_claude_trace.py
```diff
@@ -0,0 +1,912 @@
...
+        "-m",
+        help="Model name prefix for traces (default: imported-<session_id>)",
+    )
+    parser.add_argument(
+        "--include-expected-results",
+        action="store_true",
+        help="Include expected_tool_results for regression checking",
+    )
+    parser.add_argument(
+        "--deterministic-only",
```
serrrfirat, line 789: **High Severity** — `--deterministic-only` is advertised but never actually changes the generated trace.

This flag is added here, but there is no later branch that reads `args.deterministic_only`, and `Bash` is still always mapped to `shell`. In practice that means a caller can ask for “deterministic only” and still get fixtures containing live side effects like `git push`, `gh pr edit`, or local debug commands.

Please either implement the filter before writing output (for example, drop/reject non-deterministic shell turns), or remove the flag until the behavior exists so callers don't get a false safety guarantee.

src/tools/registry.rs
```diff
@@ -574,6 +575,42 @@ impl ToolRegistry {
...
+    ///
+    /// All file operations (`read_file`, `write_file`, `apply_patch`, `glob`,
+    /// `grep`) are restricted to `dir` via `base_dir`, and `shell` commands
+    /// default to `dir` as their working directory. Useful for trace replay
+    /// against a git worktree at a specific historical commit.
+    pub fn register_dev_tools_in_dir(&self, dir: PathBuf) {
+        let file_history = shared_file_history();
+        let read_state = shared_read_file_state();
+
+        self.register_sync(Arc::new(ShellTool::new().with_working_dir(dir.clone())));
```
serrrfirat, line 588: **High Severity** — this is not actually sandboxing `shell`, it is only setting the default working directory.

`ShellTool::with_working_dir()` still runs a real host process and only calls `current_dir(workdir)`, so traces can still use absolute paths, `cd ..`, networked commands, or mutate host/remote state. That's especially risky because the imported fixtures in this PR already include commands like `git push`, `gh pr edit`, and other non-local shell actions.

If worktree replay is meant to be deterministic and safe, I think `shell` needs a real sandbox / allowlist here, or replay should reject traces containing `shell` until that isolation exists.

tests/fixtures/llm_traces/imported/imported-9f613cdd.json
```diff
@@ -0,0 +1,783 @@
...
+      "steps": [
+        {
+          "response": {
+            "type": "tool_calls",
+            "tool_calls": [
+              {
+                "id": "toolu_01P1KVEmqpH3GVcCbhHAcR7b",
+                "name": "read_file",
+                "arguments": {
+                  "path": "/Users/coder/ironclaw/src/transcription/mod.rs"
```
serrrfirat, line 15: **High Severity** — this imported fixture still hardcodes the author-local repo path instead of using `{{repo_root}}`.

`resolve_repo_root()` only substitutes placeholder strings, so on another machine this will fail immediately, and on the original machine it will read/write the live checkout instead of the detached worktree under test. The same pattern appears in other imported fixtures in this PR as well.

I would strongly recommend normalizing imported traces through the placeholder rewrite pass before committing them, or adding a validation test that rejects author-local absolute repo paths in imported fixtures.

tests/e2e_worktree_replay.rs
```diff
@@ -0,0 +1,149 @@
...
+            .join("tests/fixtures/llm_traces/imported")
+            .join(fixture_name);
+
+        let mut trace = LlmTrace::from_file(&fixture_path)
+            .unwrap_or_else(|e| panic!("Failed to load {}: {}", fixture_name, e));
+
+        // Extract commit from trace metadata.
+        let commit = trace
+            .repo
+            .commit
```
serrrfirat, line 88: **Medium Severity** — multi-branch traces cannot replay correctly with the current runner.

The converter now stores `branch_commits`, but the replay path only reads `repo.commit` and checks out one worktree once. For any imported session that creates/switches branches, later turns will execute against the wrong tree. `imported-9f613cdd-turn0.json` in this PR already carries two branch commits, so the schema and runner are currently out of sync.

Please either preserve per-turn branch provenance and switch commits/worktrees as the trace advances, or constrain imported traces to single-branch sessions.

tests/support/trace_llm.rs
```diff
@@ -225,10 +249,51 @@ impl LlmTrace {
...
             http_exchanges: Vec::new(),
             expects: TraceExpects::default(),
+            repo: TraceRepoMeta::default(),
             steps: Vec::new(),
         }
     }
 
+    /// Replace `{{repo_root}}` placeholders in all tool call arguments
+    /// with the given worktree path. Call this before passing to TraceLlm.
+    pub fn resolve_repo_root(&mut self, worktree_path: &std::path::Path) {
```
serrrfirat, line 259: **Medium Severity** — `resolve_repo_root()` only rewrites tool-call arguments, not `expected_tool_results`.

The converter now rewrites path-like strings in `expected_tool_results` to `{{repo_root}}`, but this method never resolves those placeholders at replay time. If someone imports with `--include-expected-results` and a tool output contains repo paths, replay will compare the real worktree output against an unresolved placeholder and fail even though the tool behaved correctly.

I think this should recurse through the expected-result payloads as well, or the converter should stop rewriting those payloads.

tests/e2e_worktree_replay.rs
```diff
@@ -0,0 +1,149 @@
...
+mod tests {
+    use std::path::{Path, PathBuf};
+    use std::process::Command;
+    use std::time::Duration;
+
+    use crate::support::test_rig::TestRigBuilder;
+    use crate::support::trace_llm::LlmTrace;
+
+    /// The source repo that the Claude Code sessions were recorded against.
+    const SOURCE_REPO: &str = "/Users/coder/ironclaw";
```
serrrfirat, line 18: **Medium Severity** — this hardcoded source path means the smoke test skips everywhere except one machine.

Because the test later returns early when this path or commit is missing, CI and most reviewers get zero real coverage for the new replay path. Right now the suite proves “skip gracefully”, not “worktree replay actually works”.

Please resolve the source repo from the current checkout / env instead of an author-local absolute path, and keep at least one fixture that CI can exercise end-to-end.

serrrfirat:
I left inline comments on the specific spots, but the common thread across them is that the PR is very close to replaying **real local sessions** rather than **deterministic, machine-portable fixtures**.

The biggest risks are:
- `--deterministic-only` currently doesn't enforce anything, so side-effecting shell steps still flow into generated traces.
- `register_dev_tools_in_dir()` scopes file tools, but `shell` is still a host shell with a different cwd, not a true sandbox.
- Several imported fixtures still contain author-local absolute paths and live operational/debug commands, so replay can miss the detached worktree and/or touch host or remote state.
- The schema now records multi-branch provenance, but the replay runner still executes against a single commit.
- The new smoke test mostly proves "skip gracefully on one machine path mismatch" rather than exercising replay in CI.

My recommendation before merge would be:
1. either reject `shell` in imported traces or put it behind a real replay sandbox,
2. normalize/validate imported fixtures so author-local absolute paths and obviously unsafe commands cannot land in-tree,
3. make replay consume the branch/commit metadata it now records, and
4. get at least one end-to-end worktree replay fixture running in CI from the current checkout.

Issue #2193, opened 2026-04-09
Refactor: Remove V1 agent loop code, fully migrate to Engine V2


serrrfirat:
## Summary

The V2 engine is fully implemented (281 unit tests, 2457 LOC E2E, all subsystems complete) and has zero hard dependencies on V1. This issue tracks the full removal of V1-only code and making V2 the only engine. Estimated **~35,000 LOC removal across ~45 files**.

## Motivation

- Dual-architecture creates maintenance burden (every change must consider both paths)
- V1 code is ~40% of the codebase but serves no purpose once V2 is the default
- `ENGINE_V2=true` flag should not be needed -- V2 should be the only engine
- Simplifies onboarding, testing, and CI

---

## Verified V1Only Tools (14 tools)

Every tool below has `engine_compatibility() -> EngineCompatibility::V1Only` in its impl block. All other tools default to `Both` and **stay**.

| File | Tool Struct | Line |
|------|-------------|------|
| `routine.rs` | `RoutineCreateTool` | 1243 |
| `routine.rs` | `RoutineListTool` | 1337 |
| `routine.rs` | `RoutineUpdateTool` | 1504 |
| `routine.rs` | `RoutineDeleteTool` | 1595 |
| `routine.rs` | `RoutineFireTool` | 1680 |
| `routine.rs` | `RoutineHistoryTool` | 1823 |
| `routine.rs` | `EventEmitTool` | 1892 |
| `job.rs` | `CreateJobTool` | 1071 |
| `job.rs` | `CancelJobTool` | 1392 |
| `extension_tools.rs` | `ToolAuthTool` | 284 |
| `extension_tools.rs` | `ToolRemoveTool` | 601 |
| `extension_tools.rs` | `ToolPermissionSetTool` | 886 |
| `skill_tools.rs` | `SkillRemoveTool` | 837 |
| `builder/core.rs` | `BuildSoftwareTool` | 1119 |

**Tools that are NOT V1Only (stay as-is):**
- `job.rs`: `ListJobsTool`, `JobStatusTool`, `JobEventsTool`, `JobPromptTool` (all `Both`)
- `extension_tools.rs`: `ToolSearchTool`, `ToolInstallTool`, `ToolActivateTool`, `ToolListTool`, `ToolUpgradeTool`, `ExtensionInfoTool` (all `Both`)
- `skill_tools.rs`: `SkillListTool`, `SkillSearchTool`, `SkillInstallTool` (all `Both`)

---

[...]

PR #2452, opened 2026-04-14
fix(gateway): v2 engine tool_calls persistence + e2e test coverage

standardtoaster:
## Summary
- The v2 engine had zero e2e coverage for the tool call -> result -> response path (flagged in #2193 audit)
- The v2 engine executed tools correctly but never persisted `tool_calls` metadata to the chat history DB, so the web UI showed empty tool call arrays for all v2 threads
- After thread completion, ActionResult messages are extracted from the thread's internal transcript and written as a `role="tool_calls"` DB row before the assistant response
- Unlike the existing `write_v1_response`, conversation ID resolution failures are logged rather than silently swallowed

## Test plan
- [x] `cargo clippy --no-default-features --features libsql -- -D warnings` -- zero warnings
- [x] 6 new e2e tests pass (`pytest tests/e2e/scenarios/test_v2_engine_tool_lifecycle.py -v`)
- [x] 9 existing e2e tests unaffected (error_handling, tool_execution, agent_loop_recovery)
- [x] Manual verification: v2 chat history returns populated `tool_calls` array with correct names and result previews

```bash
# Manual verification
curl -s -H "Authorization: Bearer $TOKEN" \
  -X POST http://localhost:8080/api/chat/thread/new | jq .id

curl -s -H "Authorization: Bearer $TOKEN" \
  -X POST http://localhost:8080/api/chat/send \
  -H "Content-Type: application/json" \
  -d '{"content":"what time is it","thread_id":"<THREAD_ID>"}'

# Wait a few seconds, then:
curl -s -H "Authorization: Bearer $TOKEN" \
  "http://localhost:8080/api/chat/history?thread_id=<THREAD_ID>" \
  | jq '.turns[-1].tool_calls'
# Before: []
# After:  [{"name":"time","has_result":true,"result_preview":"..."}]
```

### DIFF
--- a/src/bridge/router.rs
+++ b/src/bridge/router.rs
@@ -3587,7 +3587,8 @@ pub(crate) async fn handle_mission_notification(
 
 /// Persist v2 engine tool call metadata to the v1 conversation DB.
 ///
-/// Loads steps from the v2 store, extracts action results, and writes a
+/// Loads the completed thread from the v2 store, extracts ActionResult
+/// messages (which carry the actual tool output), and writes a
 /// `role="tool_calls"` message so the chat history API can reconstruct
 /// tool call info (name, result preview, errors) for the web UI.
 async fn persist_v2_tool_calls(
@@ -3596,55 +3597,54 @@ async fn persist_v2_tool_calls(
     thread_id: ironclaw_engine::ThreadId,
     message: &IncomingMessage,
 ) {
-    // Steps are evicted from the in-memory store after join_thread, so
-    // extract tool call metadata from thread events instead (append-only).
-    let events = match store.load_events(thread_id).await {
-        Ok(e) => e,
+    // Load the thread -- it's still in the store after join_thread
+    // (join only removes from the runtime running map, not the store).
+    let thread = match store.load_thread(thread_id).await {
+        Ok(Some(t)) => t,
+        Ok(None) => {
+            debug!(thread_id = %thread_id, "thread not found in store for tool_calls persist");
+            return;
+        }
         Err(e) => {
-            debug!(thread_id = %thread_id, "failed to load v2 events for tool_calls persist: {e}");
+            debug!(thread_id = %thread_id, "failed to load thread for tool_calls persist: {e}");
             return;
         }
     };
 
+    // Extract ActionResult messages from the thread's internal transcript.
+    // The user-visible `messages` only has user/assistant messages, while
+    // `internal_messages` has the full chain including action results with
+    // actual tool output content.
     let mut calls = Vec::new();
-    for event in &events {
-        match &event.kind {
-            ironclaw_engine::EventKind::ActionExecuted {
-                action_name,
-                call_id,
-                params_summary,
-                ..
-            } => {
-                let mut obj = serde_json::json!({
-                    "name": action_name,
-                });
-                if let Some(summary) = params_summary {
-                    obj["result_preview"] = serde_json::Value::String(summary.clone());
-                } else {
-                    obj["result_preview"] = serde_json::Value::String("completed".to_string());
-                }
-                if !call_id.is_empty() {
-                    obj["tool_call_id"] = serde_json::Value::String(call_id.clone());
-                }
-                calls.push(obj);
-            }
-            ironclaw_engine::EventKind::ActionFailed {
-                action_name,
-                call_id,
-                error,
-                ..
-            } => {
-                let mut obj = serde_json::json!({
-                    "name": action_name,
-                    "error": error,
-                });
-                if !call_id.is_empty() {
-                    obj["tool_call_id"] = serde_json::Value::String(call_id.clone());
-                }
-                calls.push(obj);
-            }
-            _ => {}
+    for msg in thread.internal_messages.iter().chain(thread.messages.iter()) {
+        if msg.role != ironclaw_engine::MessageRole::ActionResult {
+            continue;
         }
+        let action_name = msg
+            .action_name
+            .as_deref()
+            .unwrap_or("unknown");
+        let preview = if msg.content.len() > 500 {
+            // safety: truncate on a char boundary
+            let end = msg
+                .content
+                .char_indices()
+                .take_while(|(i, _)| *i < 500)
+                .last()
+                .map(|(i, c)| i + c.len_utf8())
+                .unwrap_or(0);
+            format!("{}...", &msg.content[..end])
+        } else {
+            msg.content.clone()
+        };
+        let mut obj = serde_json::json!({
+            "name": action_name,
+            "result_preview": preview,
+        });
+        if let Some(ref call_id) = msg.action_call_id {
+            obj["tool_call_id"] = serde_json::Value::String(call_id.clone());
+        }
+        calls.push(obj);
     }
 
     if calls.is_empty() {
--- a/src/bridge/router.rs
+++ b/src/bridge/router.rs
@@ -7481,8 +7481,8 @@ mod tests {
             .await
             .expect("create conversation");
 
-        let message = IncomingMessage::new("web", "test-user", "do stuff")
-            .with_thread(conv_id.to_string());
+        let message =
+            IncomingMessage::new("web", "test-user", "do stuff").with_thread(conv_id.to_string());
 
         let store_arc: Arc<dyn Store> = store;
         persist_v2_tool_calls(&store_arc, &db, thread_id, &message).await;
@@ -7492,11 +7492,12 @@ mod tests {
             .list_conversation_messages(conv_id)
             .await
             .expect("list messages");
-        let tool_calls_msgs: Vec<_> = messages
-            .iter()
-            .filter(|m| m.role == "tool_calls")
-            .collect();
-        assert_eq!(tool_calls_msgs.len(), 1, "expected exactly one tool_calls row");
+        let tool_calls_msgs: Vec<_> = messages.iter().filter(|m| m.role == "tool_calls").collect();
+        assert_eq!(
+            tool_calls_msgs.len(),
+            1,
+            "expected exactly one tool_calls row"
+        );
 
         let parsed: serde_json::Value =
             serde_json::from_str(&tool_calls_msgs[0].content).expect("valid JSON");
@@ -7537,8 +7538,8 @@ mod tests {
             .await
             .expect("create conversation");
 
-        let message = IncomingMessage::new("web", "test-user", "hello")
-            .with_thread(conv_id.to_string());
+        let message =
+            IncomingMessage::new("web", "test-user", "hello").with_thread(conv_id.to_string());
 
         let store_arc: Arc<dyn Store> = store;
         persist_v2_tool_calls(&store_arc, &db, thread_id, &message).await;
@@ -7548,10 +7549,11 @@ mod tests {
             .list_conversation_messages(conv_id)
             .await
             .expect("list messages");
-        let tool_calls_msgs: Vec<_> = messages
-            .iter()
-            .filter(|m| m.role == "tool_calls")
-            .collect();
-        assert_eq!(tool_calls_msgs.len(), 0, "no tool_calls row for text-only thread");
+        let tool_calls_msgs: Vec<_> = messages.iter().filter(|m| m.role == "tool_calls").collect();
+        assert_eq!(
+            tool_calls_msgs.len(),
+            0,
+            "no tool_calls row for text-only thread"
+        );
     }
 }
--- a/src/bridge/router.rs
+++ b/src/bridge/router.rs
@@ -3660,15 +3660,25 @@ async fn persist_v2_tool_calls(
         }
     };
 
-    // Resolve the v1 conversation ID (same logic as write_v1_response)
+    // Resolve the v1 conversation ID
     let v1_conv_id = if let Some(tid) = message.conversation_scope()
         && let Ok(uuid) = uuid::Uuid::parse_str(tid)
     {
         Some(uuid)
     } else {
-        db.get_or_create_assistant_conversation(&message.user_id, &message.channel)
+        match db
+            .get_or_create_assistant_conversation(&message.user_id, &message.channel)
             .await
-            .ok()
+        {
+            Ok(cid) => Some(cid),
+            Err(e) => {
+                tracing::warn!(
+                    thread_id = %thread_id,
+                    "failed to resolve v1 conversation for tool_calls persist: {e}"
+                );
+                return;
+            }
+        }
     };
     if let Some(cid) = v1_conv_id
         && let Err(e) = db.add_conversation_message(cid, "tool_calls", &content).await
--- a/src/bridge/router.rs
+++ b/src/bridge/router.rs
@@ -3416,6 +3416,16 @@ async fn await_thread_outcome(
     if let Ok(BridgeOutcome::Respond(ref text)) = result
         && let Some(ref db) = state.db
     {
+        // Write tool_calls row BEFORE assistant response so the v1 history
+        // API sees the correct user → tool_calls → assistant triple.
+        persist_v2_tool_calls(
+            &state.store,
+            db,
+            thread_id,
+            message,
+        )
+        .await;
+
         write_v1_response(db, text).await;
     }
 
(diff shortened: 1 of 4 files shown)


src/bridge/router.rs
```diff
@@ -3006,6 +3016,108 @@ async fn handle_mission_notification(
...
+            return;
+        }
+    };
+
+    // Extract ActionResult messages from the thread's internal transcript.
+    // The user-visible `messages` only has user/assistant messages, while
+    // `internal_messages` has the full chain including action results with
+    // actual tool output content.
+    let mut calls = Vec::new();
+    for msg in thread.internal_messages.iter().chain(thread.messages.iter()) {
```
serrrfirat: **Medium Severity — Potential duplicate tool calls**

The loop chains both `internal_messages` and `messages`:

```rust
for msg in thread.internal_messages.iter().chain(thread.messages.iter()) {
```

The inline comment states "user-visible `messages` only has user/assistant messages" — but if that invariant ever breaks (or doesn't hold for all thread types), `ActionResult` messages present in both collections will produce duplicate entries in the persisted `calls` array. There's no deduplication by `action_call_id`.

**Suggestion:** Either iterate only `internal_messages` (if the comment is correct, the `.chain(thread.messages.iter())` is dead weight), or deduplicate before serializing:

```rust
let mut seen_ids = HashSet::new();
// ... in the loop:
if let Some(ref call_id) = msg.action_call_id {
    if !seen_ids.insert(call_id.clone()) {
        continue;
    }
}
```

src/bridge/router.rs
```diff
@@ -3006,6 +3016,108 @@ async fn handle_mission_notification(
...
+
+    if calls.is_empty() {
+        return;
+    }
+
+    let wrapper = serde_json::json!({ "calls": calls });
+    let content = match serde_json::to_string(&wrapper) {
+        Ok(c) => c,
+        Err(e) => {
+            tracing::warn!("failed to serialize v2 tool_calls: {e}");
```
serrrfirat: **Medium Severity — `tracing::warn!` in gateway request path may corrupt TUI**

Per CLAUDE.md: "`info!` and `warn!` output appears in the REPL and corrupts the terminal UI."

This function has three `tracing::warn!()` calls (here, line ~3106, and line ~3117) that fire on non-critical persistence failures. Since `persist_v2_tool_calls` is fire-and-forget and these are not user-actionable, they should use `debug!()` — consistent with the two `debug!()` calls earlier in this same function for thread-not-found / load-failure.

```rust
// Change these three:
tracing::warn!("failed to serialize v2 tool_calls: {e}");
tracing::warn!(thread_id = %thread_id, "failed to resolve v1 conversation ...");
tracing::warn!("failed to persist v2 tool_calls to v1 DB: {e}");

// To:
debug!("failed to serialize v2 tool_calls: {e}");
debug!(thread_id = %thread_id, "failed to resolve v1 conversation ...");
debug!("failed to persist v2 tool_calls to v1 DB: {e}");
```

(Also a minor style note: `debug!` is used bare via `use tracing::debug` at the top of the file, while `warn` is fully qualified — normalizing to `debug!` fixes both issues.)

tests/e2e/scenarios/test_v2_engine_tool_lifecycle.py
```diff
@@ -0,0 +1,379 @@
...
+    """Verify that single tool calls complete through the v2 engine."""
+
+    async def test_echo_tool(self, v2_tool_server):
+        """echo tool call -> result -> text response through v2 orchestrator."""
+        thread_id = await _create_thread(v2_tool_server)
+        await _send(v2_tool_server, thread_id, "echo hello from v2")
+
+        history = await _wait_for_response(
+            v2_tool_server,
+            thread_id,
```
serrrfirat, line 233: **Medium Severity — 5 of 6 tests don't verify the core feature (tool_calls persistence)**

Only `test_echo_tool` asserts on the `tool_calls` array. The other five tests verify response text but never check that `tool_calls` was persisted to chat history — which is the stated purpose of this PR.

A regression that silently breaks persistence for parallel or multi-step flows would pass the entire suite.

**Suggestion:** Add `tool_calls` assertions to at least:
- `test_parallel_tool_calls` — assert 2 entries with names `echo` and `time`
- `test_multi_step_chain` — assert 2 entries: `echo` then `time`
- `test_tool_then_text_then_tool` — assert tool_calls present in turns 1 and 3, absent/empty in turn 2

These are the highest-value additions since they cover the flows most likely to regress independently of single-tool persistence.

src/bridge/router.rs
```diff
@@ -2853,6 +2853,10 @@ async fn await_thread_outcome(
     if let Ok(Some(ref text)) = result
         && let Some(ref db) = state.db
     {
+        // Write tool_calls row BEFORE assistant response so the v1 history
```
serrrfirat: **High Severity** — Tool calls and response orphaned after gate resume

`persist_v2_tool_calls` is called inside the `if let Ok(Some(ref text)) = result` block, which fires for **all** outcomes including `GatePaused`. This creates a correctness issue on the gate-approval-resume path:

1. User sends message → `handle_with_engine` writes `user` row to v1 DB
2. Thread hits gate → `await_thread_outcome` returns `GatePaused`
3. `persist_v2_tool_calls` writes `tool_calls` row (partial tools before gate)
4. `write_v1_response` writes `assistant` row ("Tool requires approval…")
5. User sends "yes" → routed to `handle_approval` (NOT `handle_with_engine`, so **no "user" row** is written to v1 DB)
6. Thread resumes, completes → `await_thread_outcome` runs again
7. `persist_v2_tool_calls` writes another `tool_calls` row (ALL tools, duplicating pre-gate ones)
8. `write_v1_response` writes `assistant` row (actual response)

DB state after:
```
user         → "do something"
tool_calls   → [partial]              ← step 3
assistant    → "requires approval"    ← step 4  (consumed by turn parser)
tool_calls   → [ALL tools]            ← step 7  (ORPHANED)
assistant    → "actual response"      ← step 8  (ORPHANED)
```

`build_turns_from_db_messages` (util.rs:100) only creates turns starting from `role == "user"` messages — the second `tool_calls` + `assistant` pair has no preceding "user" row, so they're **silently skipped**. The actual tool results and final response are invisible in the history API.

Note: the assistant orphaning is pre-existing (from `write_v1_response`), but this PR adds an additional orphaned `tool_calls` row and duplicates pre-gate entries.

**Suggested fix**: Move `persist_v2_tool_calls` into the `Completed` match arm so it only runs for final outcomes:

```rust
ThreadOutcome::Completed { response } => {
    if let Some(ref db) = state.db {
        persist_v2_tool_calls(&state.store, db, thread_id, message).await;
    }
    // ... existing credential fallback logic ...
    Ok(response)
}
```

serrrfirat: **Medium Severity** — Tool calls not persisted for `Completed { response: None }`

The `persist_v2_tool_calls` call is guarded by `if let Ok(Some(ref text)) = result`, so it only fires when the outcome produces a text response. `ThreadOutcome::Completed { response: None }` (thread completes with tools but no final text) skips this block entirely, leaving a gap in the tool_calls history.

If the fix for the GatePaused issue (finding #1) moves `persist_v2_tool_calls` into the `Completed` match arm, this is also resolved — the call would run regardless of whether `response` is `Some` or `None`:

```rust
ThreadOutcome::Completed { response } => {
    if let Some(ref db) = state.db {
        persist_v2_tool_calls(&state.store, db, thread_id, message).await;
    }
    // ... rest of Completed handling ...
    Ok(response)
}
```

src/bridge/router.rs
```diff
@@ -3006,6 +3010,106 @@ async fn handle_mission_notification(
     }
 }
```
serrrfirat, line 3583: **Medium Severity** — No unit/integration test for `persist_v2_tool_calls`

This function is only covered by e2e tests (which are slow and run with `AGENT_AUTO_APPROVE_TOOLS=true`, so they never exercise the GatePaused path). Edge cases not covered:

- **GatePaused + resume**: duplicate tool_calls rows (finding #1)
- **Empty tool result**: `msg.content` is empty → `result_preview` is `""` → `has_result` is `true` but preview is meaningless
- **`Completed { response: None }`**: tool_calls not persisted at all (finding #2)
- **Thread not in store after join**: `load_thread` returns `None`, function silently returns

Per the project's testing rules ("Test Through the Caller"), an integration-tier test driving the `await_thread_outcome` path with a mock store/DB would catch these regressions faster and more reliably than the e2e suite.

src/bridge/router.rs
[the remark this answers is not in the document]
standardtoaster: Fixed in ff372e1. Iterating only `internal_messages` now.

standardtoaster: Fixed in ff372e1. Dropped the `.chain(thread.messages.iter())` -- now iterates only `internal_messages` as you suggested.

src/bridge/router.rs
standardtoaster: Fixed in ff372e1. All three calls are `debug!()` now, consistent with the rest of the function.

tests/e2e/scenarios/test_v2_engine_tool_lifecycle.py
standardtoaster: Fixed in 652315e. Added `tool_calls` assertions to `test_parallel_tool_calls` (2 entries, echo + time), `test_multi_step_chain` (2 entries with results), and `test_tool_then_text_then_tool` (tool_calls present in turns 1 and 3, absent in turn 2).

src/bridge/router.rs
standardtoaster: Good catch. Fixed in 652315e. Moved `persist_v2_tool_calls` into the `Completed` match arm so it only fires for final outcomes. The GatePaused path no longer writes partial/duplicate rows.

standardtoaster: Fixed in the same commit (652315e). The call is now in the `Completed` arm unconditionally, so it runs regardless of whether `response` is `Some` or `None`.

src/bridge/router.rs
standardtoaster: Added two libsql-backed unit tests in 652315e. One verifies correct extraction from `internal_messages`, the other verifies skip behavior for text-only threads. The GatePaused path is harder to test at unit level since it needs the full engine, but the e2e suite now covers the persistence assertions that would catch regressions there.

ilblackdragon (commented):
## Overview

Two interlocking pieces:

1. **Persistence fix** — adds `persist_v2_tool_calls` in `src/bridge/router.rs` that, after a `ThreadOutcome::Completed` thread, loads the thread, walks `internal_messages`, extracts `ActionResult` messages, and writes a `role="tool_calls"` row to the v1 conversation DB so the chat history API can reconstruct tool call info for the web UI.
2. **E2E coverage** — 6 new scenarios in `test_v2_engine_tool_lifecycle.py` (single-tool echo/time, text-only, parallel dispatch, multi-step chain, multi-turn) + 2 libsql unit tests in `router.rs`.

Net effect: chat history API stops returning `tool_calls: []` for v2 threads.

## Correctness

- **`Completed`-arm gating** — commit `652315e8` correctly moved `persist_v2_tool_calls` out of the general outcome path and into `ThreadOutcome::Completed` only. Without this, `GatePaused` would orphan partial rows that duplicate on resume. The `Completed { response: None }` case (tools executed but no final text) is now also covered.
- **UTF-8-safe truncation** — the 500-byte preview uses `char_indices().take_while` + `char.len_utf8()` for char-boundary truncation (matches `.claude/rules/review-discipline.md` "UTF-8 string safety"). Won't panic on multi-byte content.
- **Conversation ID resolution** — parses UUID from `message.conversation_scope()` first, falls back to `get_or_create_assistant_conversation`. Both paths can fail cleanly (debug log + return) without corrupting the thread.
- **`has_result` derivation** — the written JSON is `{name, result_preview, tool_call_id}`. `src/channels/web/util.rs:93-94` derives `has_result` from `result_preview.is_some_and(!is_null)`, so the E2E assertion `tc["has_result"] is True` is a valid contract check.
- **`V24__llm_calls_created_at_index` checksum** — the migration file already exists on staging (landed in #1963, `532fc61d`); the PR is just adding the checksum after a rebase. Not a schema change, not risky.
- ⚠️ **`debug!` vs `warn!` for persistence failures** — the author initially used `warn!`, reviewer (commit `ff372e11`) changed to `debug!` citing CLAUDE.md's \"background tasks must not use info/warn\" rule. That rule is about REPL/TUI corruption. This is an HTTP handler path, not a background task — a silent `debug!` hides a user-visible bug (missing tool_calls in the UI) unless someone sets `RUST_LOG=debug`. Reasonable case for reverting to `warn!` or adding a metrics counter.
- ⚠️ **Extra `store.load_thread` on every `Completed`** — runs even for text-only threads that will produce no rows. The comment says the thread is still in the in-memory cache, so this is a HashMap hit, not a DB round-trip. Acceptable; worth a doc comment noting the cheap-path reliance.

## Style / Conventions

- ✅ **`dispatch-exempt` annotation not required** — `router.rs` is the bridge layer, not a handler/CLI file (the pre-commit hook's scope). Direct `state.db` / `state.store` access is canonical here.
- 🟡 **Inconsistent `Arc` qualification** — helper signature uses `&std::sync::Arc<dyn Store>` / `&std::sync::Arc<dyn Database>` while the test module uses plain `Arc::new(...)`. `Arc` is in scope — could drop the `std::sync::` prefix for consistency with surrounding code.
- 🟡 **Mock LLM multi-step trigger uses `_conversation_has_user_trigger` with a tool-result count** — pragmatic, but the comment explicitly notes \"v2 engine tool results don't always include the tool name in a parseable format\". Worth a TODO to tighten once that contract is stable.

## Test Coverage

**Strong**:

- 6 E2E scenarios cover the happy path + parallel + multi-step + multi-turn. Each asserts on the actual `tool_calls` array shape, not just response text.
- 2 libsql unit tests pin the write path and the skip-when-empty path. Uses a real `LibSqlBackend` (not a mock), which is the right tier per `.claude/rules/testing.md`.

**Gaps worth adding**:

- **`GatePaused` does NOT persist** — the bug that prompted commit `652315e8`. No test asserts that a paused gate leaves the DB clean of `role=\"tool_calls\"` rows. Without this, a future \"simplification\" could move the call back out of the `Completed` arm and silently regress. Would be a straightforward libsql test: construct a thread with `ActionResult`s, drive `await_thread_outcome` with a `GatePaused` outcome (or test the call-site conditional directly), then verify no write happens. The current unit tests always call the function — they don't exercise the gating logic at the call site.
- **`is_error: true` action results** — the preview includes the error body but no explicit `error` flag in the written JSON. E2E tests don't exercise a failing tool. The history reader at `util.rs:93-94` would still show `has_result: true` for a failed tool — probably correct, but not pinned.
- **Truncation boundary** — `char_indices` safety isn't exercised. A small unit test with a multi-byte string >500 bytes would pin the invariant.
- **Conversation ID fallback** — the fallback path (`get_or_create_assistant_conversation`) has no direct test. Both unit tests exercise only the `conversation_scope()`-provided path.

## Security

- ✅ **No prompt-injection surface introduced** — the preview is JSON-serialized into a `content` field, not interpolated into a prompt.
- ⚠️ **No sensitive-data redaction on the preview** — if a tool output contains a secret (API key, auth token), the 500-byte preview persists it to DB and exposes it via chat history. This matches v1 `write_v1_response` behavior, so it's not introduced here, but worth flagging as a PR-body follow-up. Verify that the `SafetyLayer` sanitizer runs on `ActionResult.content` before it lands in `internal_messages`.

## Performance

Per-completion: one extra `store.load_thread` (in-memory cache hit in practice) + O(n) internal_messages scan + one DB row insert. Negligible for typical threads. The 500-char truncation happens per ActionResult, which bounds worst-case row size.

## Verdict

**Approve-leaning with minor follow-ups**. The fix is surgical, the `Completed`-arm placement is the right bug to solve, and E2E coverage meaningfully closes the #2193 audit gap. Before merge I'd suggest:

1. Revert the `debug!` → `warn!` decision for persistence failures (user-visible UI gap when it fails silently).
2. Add a regression test that pins \"no rows written for `GatePaused`\" at the call-site level — otherwise the `ThreadOutcome::Completed` guard has no test protection.
3. Optional: add the multi-byte truncation unit test.

Not blockers — ship after 1+2 or file as immediate follow-up. This is on the engine-v2-default critical path, so moving quickly is worth more than polishing these.

ilblackdragon:
Rebased onto latest \`origin/staging\` and applied the three polish items from the review in commit \`49b65108\`:

1. **`warn!` for persistence failures** — all four `persist_v2_tool_calls` failure sites (load thread, serialize, resolve conv id, DB write) now emit at `warn!` with `thread_id` field. `debug!` hid a user-visible gap (missing chat-history `tool_calls`); CLAUDE.md's "background tasks must not use info/warn" rule is about REPL/TUI corruption and doesn't apply to this HTTP handler path.

2. **Regression test: `persist_v2_tool_calls_only_called_from_completed_arm`** — parses the source of `router.rs`, asserts exactly one call site, and asserts that site sits between the `Completed` and `GatePaused` match arms. Pins the fix from commit `652315e8` so a future refactor can't silently move the call back out of the `Completed` arm.

3. **Multi-byte UTF-8 truncation test: `persist_v2_tool_calls_truncates_multibyte_content_on_char_boundary`** — 400 × 3-byte CJK chars (1200 bytes), pins (a) no panic, (b) valid UTF-8 via JSON round-trip, (c) body length bounded near 500, (d) body contains only complete 3-byte chars.

Rebase touched `tests/e2e/ironclaw_e2e.egg-info/SOURCES.txt` (alphabetical-sort merge) and `src/bridge/router.rs` (merging the `BridgeOutcome::Respond(text)` / `NoResponse` refactor from staging with the `persist_v2_tool_calls` call site). Tracking follow-up items in #2669.

**Gates** (rebased + polished):
- `cargo fmt` ✓
- `cargo clippy --no-default-features --features libsql --tests -- -D warnings` → 0 warnings
- `cargo test -p ironclaw --lib --features libsql` → 5125 passed

Branch now at 9 commits ahead of `staging`, mergeable.

ilblackdragon (approved):

PR #2454, opened 2026-04-14
fix(channel): feishu pairing

hanakannzashi:
## Summary

- Removed hardcoded Feishu secret→config mapping from host code and made it manifest-driven via `setup.secret_config_mappings` in channel capabilities.
- Added `secret_config_mappings` support to the channel schema/setup flow, with validation that mappings reference declared setup secrets and do not override reserved runtime config keys.
- Applied manifest-driven secret config injection to Feishu (`app_id`, `app_secret`, `verification_token`) so startup, hot activation, and refresh inject the same runtime config deterministically without adding new external setup parameters.


think-in-universe (changes requested):
## PR Review — `nearai/ironclaw#2454`

I reviewed this PR and I’d currently recommend **requesting changes**.

### ✅ What looks good
- The shift to manifest-driven `secret_config_mappings` is a solid direction and improves extensibility.
- The hardening in `schema.rs` (filtering mappings, rejecting reserved keys, requiring declaration in `required_secrets`) is a good security improvement.
- UI updates for pairing/setup visibility look aligned with the PR goal.

### ❗ Blocking concern (high severity)
I see a likely scope regression in hot activation:

- In `src/extensions/manager.rs`, the call to `WasmChannelLoader::new(...)` appears changed from a stable owner scope to the **request-scoped `user_id`**.
- The same commit also changes secret config injection to use `owner_scope_id`, which is correct.
- But passing request `user_id` into loader construction can make runtime scope depend on *who triggered activation* rather than the channel instance owner, which risks cross-user secret/settings mismatches.

This concern is also echoed by an in-PR Copilot review comment on **April 16, 2026** calling out the same owner-scope inconsistency.

**Suggested fix:** keep loader/runtime owner scope bound to stable channel owner scope (`self.user_id` / channel owner ID), and only use request user context for explicitly user-scoped actions.

---

### Sources reviewed
- PR conversation + summary: <https://github.com/nearai/ironclaw/pull/2454>  
- Files changed view: <https://github.com/nearai/ironclaw/pull/2454/files>  
- Commit with hardening + owner-scope edits: <https://github.com/nearai/ironclaw/pull/2454/commits/670965fa042b11368fff5831ee531f9a00ae3f18>

think-in-universe (dismissed):

think-in-universe (dismissed):

src/channels/wasm/schema.rs
```diff
@@ -87,6 +87,13 @@ fn default_type() -> String {
     "channel".to_string()
 }
 
+fn is_reserved_runtime_config_key(config_key: &str) -> bool {
+    matches!(
+        config_key.trim().to_ascii_lowercase().as_str(),
+        "webhook_secret" | "tunnel_url" | "owner_id"
+    )
+}
```
serrrfirat: **Medium Severity — Reserved config key blocklist not coupled to injection sites**

`is_reserved_runtime_config_key` blocks `webhook_secret`, `tunnel_url`, and `owner_id`, but this is a closed-set check with no compile-time or test-time coupling to `build_runtime_config_updates()` (in `pairing/approval.rs`) or the Telegram `bot_username` branch in `setup.rs:278`.

If a future change adds a new host-injected runtime config key without updating this blocklist, a manifest's `secret_config_mappings` could silently override it.

**Suggestion:** Either share a `const RESERVED_RUNTIME_CONFIG_KEYS` array between the validator and the injector, or add a test that asserts every key produced by `build_runtime_config_updates` appears in the reserved set. This turns a "remember to update" into a compile/test failure.

src/channels/wasm/setup.rs
```diff
@@ -286,6 +292,7 @@ async fn register_channel(
             &channel_name,
```
serrrfirat: **Low Severity — Manifest mapping could override non-reserved host-injected keys**

`inject_channel_secrets_into_config` runs after `build_runtime_config_updates` and the Telegram `bot_username` insertion (line 278). The reserved-key check in `validated_secret_config_mappings` blocks the big three (`webhook_secret`/`tunnel_url`/`owner_id`), but any other host-injected key (e.g. `bot_username`) could be overridden by a manifest mapping.

Currently harmless because each channel injects into its own config space — a Feishu manifest setting `bot_username` won't affect Telegram. But the defense-in-depth intent suggests the blocklist should track all host-injected keys, not just the current three. See related comment on `schema.rs:95`.

crates/ironclaw_gateway/static/js/surfaces/extensions.js
```diff
@@ -365,23 +365,32 @@ function renderExtensionCard(ext) {
...
   card.appendChild(actions);
 
-  // For WASM channels, check for pending pairing requests.
+  // For WASM channels, check for setup and pairing requests.
   if (ext.kind === 'wasm_channel') {
-    if ((ext.onboarding_state || ext.activation_status || 'installed') === 'setup_required') {
+    const channelStatus = ext.onboarding_state || ext.activation_status || 'installed';
+    const adminView = currentUserIsAdmin();
+    const showFullPairing = channelStatus === 'pairing_required' || channelStatus === 'pairing';
+    const showPendingOnly = adminView && channelStatus === 'active';
```
serrrfirat: **Medium Severity**\n\n`ready` channels lose the new compact pending-pairing UI.\n\nThis branch only shows the admin compact pairing section when `channelStatus === "active"`, but the rest of this file treats `ready` as an equivalent active state (`state-active` styling, active label, and the stepper all special-case it). If the backend reports `ready`, admins will not see pending pairing requests here even though the channel is effectively active.\n\nConcrete fix: include `ready` in `showPendingOnly` (or normalize the status once and reuse that classification everywhere) so pending approvals stay visible for both active states.

src/extensions/manager.rs
```diff
@@ -284,22 +285,13 @@ use crate::pairing::approval::build_runtime_config_updates as build_wasm_channel
...
-            ("app_id", "feishu_app_id"),
-            ("app_secret", "feishu_app_secret"),
-            ("verification_token", "feishu_verification_token"),
-        ],
-        _ => return,
-    };
-
-    for &(config_key, secret_name) in secret_config_mappings {
-        if let Ok(decrypted) = secrets.get_decrypted(owner_id, secret_name).await {
+    for mapping in secret_config_mappings {
```
serrrfirat: **High Severity**\n\nHot activation no longer matches the startup credential-injection behavior.\n\n`inject_wasm_channel_secret_config_updates()` now only reads from the secrets store, but the startup path in `src/channels/wasm/setup.rs` still falls back to `std::env::var(...)` when a mapped secret is missing. That means a channel like Feishu can start correctly when `FEISHU_APP_ID` / `FEISHU_APP_SECRET` / `FEISHU_VERIFICATION_TOKEN` come from environment variables, but fail after hot activation / refresh because this path never injects those values into runtime config.\n\nSo the PR ends up with different behavior across boot vs hot-activation paths — exactly the drift this refactor was trying to remove.\n\nConcrete fix: either reuse the same helper as the startup path, or restore the env fallback here as well so both code paths derive `secret_config_mappings` the same way.

ilblackdragon:
## PR Review

### Overview

Two things bundled:
1. **Manifest-driven WASM channel secret→config injection.** Removes the hardcoded `match channel_name { "feishu" => … }` in `src/channels/wasm/setup.rs` and `src/extensions/manager.rs` in favor of a new `setup.secret_config_mappings` field in channel capabilities JSON. Feishu's `app_id`/`app_secret`/`verification_token` now live in `channels-src/feishu/feishu.capabilities.json`.
2. **Pairing UI polish.** Active channel cards show a compact "pending pairing" list for admins; empty `.ext-pairing` containers are hidden via CSS.

Plus scaffolding:
- New `src/channels/wasm/runtime_config_keys.rs` centralizing reserved host-owned keys (`webhook_secret`, `tunnel_url`, `owner_id`, `bot_username`) with `is_reserved_runtime_config_key()`.
- `ChannelCapabilitiesFile::validated_secret_config_mappings()` filtering: rejects empty keys/names, reserved host keys, and secrets not declared in `setup.required_secrets`.

### ✅ What's strong

- **Good security posture.** The 3 filter rules in `collect_validated_secret_config_mappings` (`schema.rs:344`) prevent three real attacks: manifest overriding host-managed values, exfiltration of arbitrary cross-channel secrets (the `"stolen_key" → "openai_api_key"` test case), and empty/malformed entries. Case-insensitive comparison via `eq_ignore_ascii_case` on both lookup and set is consistent.
- **Nice consolidation.** Replacing scattered string literals (`"webhook_secret"`, `"bot_username"`, `"tunnel_url"`, `"owner_id"`) with named consts and asserting via `runtime_config_updates_only_emit_reserved_host_keys` creates a compile-time + test-time guarantee that host keys stay off-limits from manifests.
- **Owner scope is correct.** Previous review iterations flipped between `self.user_id` and request `user_id`; final state uses `self.user_id` (gateway owner) for `refresh_active_channel` (`manager.rs:5738`) and `&config.owner_id` for startup registration (`setup.rs`). Matches what the earlier blocking review asked for.

### ⚠️ Suggestions

- **Comment/code reorder in `refresh_oauth_state_key`** (`manager.rs:6189–6192`). The diff swaps the order of the two `delete` calls but leaves the "Also best-effort delete any legacy caller-scoped entry" comment inline — it now sits *above* the legacy delete which is first, and the "OAuth callback looks up under `self.user_id`" rationale (for the primary delete) disappears entirely. Functionally equivalent, but this change is unrelated to the stated PR scope and the comment no longer explains the `self.user_id` side. Either revert, or restore the rationale comment.
- **Long fully-qualified paths.** `&[crate::channels::wasm::SecretConfigMappingSchema]` in both `setup.rs:570` and `manager.rs:287` — `manager.rs` already has a wasm import block; adding `SecretConfigMappingSchema` to the existing `use crate::channels::wasm::{…}` list would match project style.
- **Test through the caller.** Per `CLAUDE.md`'s "test through the caller" rule, the manager-side injection (`ExtensionManager::refresh_active_channel` around `manager.rs:5738` / `reactivate_with_runtime_config` around `5965`) doesn't have a direct test. The helper `inject_wasm_channel_secret_config_updates` is only exercised transitively via `setup.rs`'s test. An integration test that drives `refresh_active_channel` with a manifest containing `secret_config_mappings` would close the gap.
- **Warning channel.** `validated_secret_config_mappings_with_warnings()` is called from the build-time validator with the result discarded (`let _ = …`). Fine as "call for side-effects," but a brief comment (`// emit manifest warnings`) at the call site would make the intent obvious; otherwise it reads like dead code.
- **JS cosmetic.** `loadPairingRequests` now takes `options` but also checks `!currentUserIsAdmin()` *and* the caller checks `adminView` before calling. The early-return guard is now redundant with the caller gate (`extensions.js:371`). Not a bug — just belt-and-suspenders.

### 🔒 Security / correctness

- Validation runs on every `validated_secret_config_mappings()` call. Cheap (small vectors), fine.
- The env-var fallback (`secret_name.to_uppercase()`) in `inject_channel_secrets_into_config` is preserved. Since `secret_name` is now manifest-supplied, in principle a malicious capabilities file could map `secret_name: "PATH"` to exfiltrate an env var into `config_updates` — but the same file has to declare that name in `required_secrets`, and capabilities JSON is code-review-controlled (under `channels-src/`). Acceptable, but worth calling out.
- No new DB migrations, no LLM-data deletion, no `unwrap`/`expect` in production code. Conforms to project conventions.

### Verdict

Scope is clean, tests cover the filter surface, security invariants are upheld. Recommend **approve with minor cleanup** — mainly the unrelated OAuth-delete comment reorder and a direct manager-path test.

ilblackdragon:
## Addressed open review items (commit `6b2a0844`)

| Item | Reviewer / Severity | Fix |
|------|---------------------|-----|
| Env-var fallback parity between startup and hot-activation/refresh | @serrrfirat HIGH | Extracted a single `inject_wasm_channel_secret_config_mappings` helper in `src/channels/wasm/setup.rs`. `ExtensionManager::refresh_active_channel` and `reactivate_with_runtime_config` now call it directly, so the secrets-store lookup **and** the `STRING_UPPERCASE` env-var fallback live in one place. |
| `ready` state loses compact pending-pairing UI | @serrrfirat MED | `extensions.js` now treats `active` ∪ `ready` as active-like when deciding whether admins get the compact pending list (matches the status-label and stepper logic elsewhere in the file). |
| Reserved-key blocklist not coupled to injection sites | @serrrfirat MED | Added `reserved_set_matches_named_constants` in `src/channels/wasm/runtime_config_keys.rs`: iterates every `RUNTIME_CONFIG_KEY_*` constant and asserts it's in `RESERVED_RUNTIME_CONFIG_KEYS`, plus checks len-equality. A new host-injected key can no longer be added without extending the reserved set. |
| OAuth nonce delete comment/code reorder | self-review | Rewrote the comment block in `refresh_oauth_state_key` to match the new delete order and restore the "`self.user_id` = gateway owner" rationale that had been dropped. |
| Fully-qualified `SecretConfigMappingSchema` paths | self-review | Added to existing `use crate::channels::wasm::{…}` import blocks in `setup.rs`, its tests, and via the reworked helper path in `manager.rs`. |
| Warning-call side effect reads like dead code | self-review | Expanded the `let _ = self.validated_secret_config_mappings_with_warnings()` comment to call out that it exists purely for manifest-validation warning emission. |
| Redundant `currentUserIsAdmin()` guard in `loadPairingRequests` | self-review | **Kept** on closer look — the polling loop in `startPairingPoll` also iterates non-admin `.ext-pairing[data-channel]` nodes (created by `renderMemberPairingClaim`), so the internal guard protects non-admins from the admin-only endpoint. Not redundant. |

### Gates
- `cargo fmt` ✅
- `cargo clippy --all --benches --tests --examples --all-features` ✅ zero warnings
- `cargo test --lib channels::wasm` ✅ (238 passed, including two new `runtime_config_keys` tests)
- `cargo test --lib pairing::approval` ✅

The 8 pre-existing `skill_tools::tests::test_zip_*` failures on this branch are unrelated (I verified they fail on the unmodified PR HEAD as well).

ilblackdragon (approved):

PR #2458, opened 2026-04-14
fix: restore issue-2402 v2 gate resume and action alias consistency

### LINKED ISSUE #2402: [QA] Bot enters infinite "Calling LLM" loop after tool operations
### Environment

hosted-staging (crab shack)

### Version / Commit Hash

### QA Test Date

2026-04-13

### Feature Area

Agent Engine → Tool Operations

### Steps to Reproduce

1. Open staging environment (hosted-staging / crab shack)
2. Start a new chat with the bot
3. Ask the bot to perform a task that triggers tool_install, tool_activate, or tool_search (e.g., "Install the Google Sheets tool and use it")
4. Approve the tool_install permission when prompted
5. Observe the bot's response after the tool operation completes

### Expected Behavior

The bot should complete the tool operation and continue responding to the user's request normally. The "Calling LLM" animation should resolve within a reasonable time (< 30 seconds).

### Actual Behavior

After tool_install, tool_activate, or tool_search operations, the bot enters an infinite loop. The "Calling LLM" animation plays indefinitely and the bot never produces a response. The user is forced to refresh the page to recover. This was observed across at least three separate use cases (UC3, UC4, UC14) in bug bash 4/13, making it one of the most impactful issues in v0.25.0.

### Logs / Screenshots

Observed in bug bash 4/13 across multiple use cases:
- UC3: Bot stuck after tool_search operation
- UC4: Bot stuck after tool_install + tool_activate sequence
- UC14: Bot stuck after tool_install, "Calling LLM" animation loops indefinitely

No specific error logs captured at the time, but the pattern is consistent and reproducible.

### Pre-submit checklist

serrrfirat:
## Summary
- add a staging regression for auth/external gate resumes that proved v2 could lose `ActionResult` call pairing when `pending.call_id` was empty or stale, then fix the resume branches to resolve or synthesize the correct correlator before resuming the thread
- add a staging regression for hyphen/underscore action aliases in the real structured executor path, then make granted-action matching consistent so lease preflight, policy, and consumption agree on aliased action names
- validate the fixes with targeted staging regressions, adjacent router/capability tests, and `cargo check -p ironclaw_engine -p ironclaw`

## Validation
- cargo test -p ironclaw_engine alias_normalization_stays_consistent_between_preflight_and_consume
- cargo test -p ironclaw covers_action
- cargo test -p ironclaw resolve_gate_repairs_call_id_for_resume_output_auth_resume
- cargo test -p ironclaw resolved_call_id_
- cargo test -p ironclaw handle_with_engine_reemits_approval_status_for_pending_gate
- cargo check -p ironclaw_engine -p ironclaw

### DIFF
--- a/crates/ironclaw_engine/src/executor/structured.rs
+++ b/crates/ironclaw_engine/src/executor/structured.rs
@@ -700,6 +700,56 @@ mod tests {
         assert_eq!(result.results[1].call_id, "id_bbbb");
     }
 
+    #[tokio::test]
+    async fn alias_normalization_stays_consistent_between_preflight_and_consume() {
+        let thread = Thread::new(
+            "test",
+            ThreadType::Foreground,
+            ProjectId::new(),
+            "test-user",
+            ThreadConfig::default(),
+        );
+        let effects: Arc<dyn EffectExecutor> = Arc::new(MockEffects::new(
+            vec![test_action("create-issue")],
+            vec![Ok(ActionResult {
+                call_id: String::new(),
+                action_name: "create-issue".into(),
+                output: serde_json::json!({"ok": true}),
+                is_error: false,
+                duration: Duration::from_millis(1),
+            })],
+        ));
+        let leases = Arc::new(LeaseManager::new());
+        let policy = Arc::new(PolicyEngine::new());
+        let ctx = make_exec_context(&thread);
+
+        leases
+            .grant(
+                thread.id,
+                "github",
+                GrantedActions::Specific(vec!["create_issue".into()]),
+                None,
+                None,
+            )
+            .await
+            .unwrap();
+
+        let calls = vec![ActionCall {
+            id: "call_alias_consume".into(),
+            action_name: "create-issue".into(),
+            parameters: serde_json::json!({"title": "test"}),
+        }];
+
+        let result = execute_action_calls(&calls, &thread, &effects, &leases, &policy, &ctx, &[])
+            .await
+            .unwrap();
+
+        assert_eq!(result.results.len(), 1);
+        assert_eq!(result.results[0].call_id, "call_alias_consume");
+        assert_eq!(result.results[0].action_name, "create-issue");
+        assert!(!result.results[0].is_error);
+    }
+
     // ── GatePaused(Authentication) tests ─────────────────────
 
     #[tokio::test]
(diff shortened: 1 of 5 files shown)


henrypark133 (dismissed):
## Review: Gate resume + action alias consistency (Risk: Medium)

Clean fix that correctly addresses the gate resume call_id corruption and action name alias inconsistency.

Positives:
- `action_name_matches()` provides clean bidirectional hyphen/underscore normalization
- `GrantedActions::covers()` updated with same normalization — transitively fixes `find_and_consume` 
- `resolved_or_synthetic_call_id_for_pending_action` fixes both `resume_output` paths in `resolve_gate`
- Comprehensive test suite: 4 new tests including full install→auth→aliased-follow-up integration test
- `aliased_action_name_still_triggers_policy_approval` verifies policy enforcement isn't bypassed

Convention notes:
- The new helper loads the thread from store for call_id resolution — minor extra DB read but necessary since the resume_output paths don't have the thread pre-loaded
- Test structure follows the "test through the caller" pattern well

standardtoaster (commented):
One question on `resolved_call_id_for_pending_action` (router.rs:189-197): the `resolved_ids` set is built from `thread.messages`, but in production the orchestrator writes ActionResult messages to `thread.internal_messages` via `sync_runtime_state`. The test uses `thread.add_message()` which goes to `messages`, so it passes, but the production `resolved_ids` set may always be empty.

This probably doesn't matter in practice since new gates always have `call_id` set (line 185 short-circuits), but if the legacy fallback path is meant to work, it should scan `internal_messages` instead of `messages`. Is this a known limitation?

serrrfirat (commented):
Good catch @standardtoaster — you're right. `resolved_call_id_for_pending_action` was only scanning `thread.messages`, but in production the orchestrator writes ActionResult (and assistant-with-actions) messages to `thread.internal_messages` via `sync_runtime_state`. The `resolved_ids` set was always empty and the fallback never matched, silently falling through to a synthetic id.

Fixed in 893e3198: the function now chains both `messages` and `internal_messages` iterators. Added a regression test (`resolved_call_id_legacy_fallback_scans_internal_messages`) that puts messages in `internal_messages` only — this test would have failed before the fix.

henrypark133 (approved):
## Review: Gate resume + action alias consistency

No verified findings in the current diff. I rechecked the prior `internal_messages` concern and the fix now scans both visible and internal transcripts before synthesizing a fallback call id. The added engine-v2 coverage also exercises the auth-resume plus aliased follow-up path that previously risked hanging.

standardtoaster (approved):
LGTM. I'll move `persist_v2_tool_calls` into the `Completed` match arm on #2452 so it doesn't fire during `GatePaused`.

PR #2460, opened 2026-04-14
fix(engine): guard consecutive-error checks against None limit

ilblackdragon:
## Summary
- `ThreadConfig::default()` had `max_consecutive_errors: None`, which serializes to `null`. In the Python orchestrator, `config.get("max_consecutive_errors", 5)` returns `None` (not `5`) when the key is present with a null value, so `consecutive_action_errors >= max_consecutive_errors + 2` crashed with TypeError on the first action error (e.g. a missing `commitments/README.md` on `memory_read`).
- Fixed on both sides: Rust default is now `Some(5)` so callers using `ThreadConfig::default()` send a real int; Python guards (`default.py:661`, `:870`, `:877`) now short-circuit on `None` and treat it as "no limit", matching the `Option<u32>` semantics used elsewhere (`max_tokens_total`, `max_budget_usd`).

## Test plan
- [x] `cargo test -p ironclaw_engine --lib -- default_config_has_concrete none_limit` — new regression tests cover the Rust default and both Python guards (action errors + code errors) with `max_consecutive_errors = None`
- [x] `cargo check -p ironclaw_engine` clean
- [x] `cargo fmt`

### DIFF
--- a/crates/ironclaw_engine/orchestrator/default.py
+++ b/crates/ironclaw_engine/orchestrator/default.py
@@ -467,6 +467,7 @@ def run_loop(context, goal, actions, state, config):
     max_iterations = config.get("max_iterations", 30)
     max_nudges = config.get("max_tool_intent_nudges", 2)
     nudge_enabled = config.get("enable_tool_intent_nudge", True)
+    # None means "no limit" — callers can disable the guard explicitly.
     max_consecutive_errors = config.get("max_consecutive_errors", 5)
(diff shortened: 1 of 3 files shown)


serrrfirat (approved):
Clean, well-scoped fix. Both the Rust default change and Python None guards are correct. All three comparison sites in default.py are covered, backward compatibility with old serialized configs is maintained, and the regression tests are appropriate.

Issue #2070, opened 2026-04-06
ownership(channels): clarify and fix WASM channel secret ownership


henrypark133:
## MT-4 — High: WASM channel secrets use config.owner_id without documentation

**Origin:** [PR #1898 review comment](https://github.com/nearai/ironclaw/pull/1898#issuecomment-4176597229) (MT-4)
**Spec:** [Ownership model design](docs/superpowers/specs/2026-04-01-ownership-model-design.md)

### Problem

`src/channels/wasm/setup.rs` looks up all channel secrets (webhook secrets, signature keys, HMAC secrets) using `config.owner_id`. Lines 184, 287, 306, 322 all call `secrets.get_decrypted(&config.owner_id, ...)`.

For **channel setup at boot**, this is actually correct — channels are instance-level resources owned by the operator. But:
1. This isn't documented, so it looks like a bug
2. Runtime fallbacks in `wrapper.rs` hardcode `"default"` which IS wrong

### Runtime fallbacks that need fixing

- `wrapper.rs:~1049` — broadcast metadata falls back to `"default"` instead of using `owner_scope_id`
- `wrapper.rs:~3130` — `resolve_websocket_identify_message()` always looks up secrets under hardcoded `"default"` instead of using the channel's owner

### What to do

- [ ] Add documentation comments in `setup.rs` explaining that channel bot tokens are instance-level secrets, correctly looked up under `config.owner_id`
- [ ] Fix `wrapper.rs:~1049` — remove `"default"` fallback, use `owner_scope_id` consistently
- [ ] Fix `wrapper.rs:~3130` — pass the channel's owner through to `resolve_websocket_identify_message()` instead of hardcoding `"default"`
- [ ] Add test: channel wrapper uses its configured owner, not a hardcoded default

### Files

- `src/channels/wasm/setup.rs` — channel secret lookup (document, don't change)
- `src/channels/wasm/wrapper.rs` — runtime secret fallbacks (fix)

### Related

- Overlaps with WASM credential fallback issue (MT-2) for the `wrapper.rs` fixes

PR #2465, opened 2026-04-14
fix(ownership): fail closed on WASM default-scope fallback

### LINKED ISSUE #2069: ownership(wasm): remove "default" credential fallback in WASM tool execution

G7CNF:
Fixes #2069
Refs #2070

## What changed
- WASM tool credential resolution now fails closed instead of borrowing the `default` scope for admin users.
- The tool-facing error now names the user and tells them to configure the missing credential with `ironclaw secrets set`.
- WASM channel startup no longer falls back to legacy `default` broadcast metadata when the owner scope has no stored value.
- Added regressions proving we only read the configured owner scope and do not probe `default`.

## Notes
- `#2070` overlaps on the same ownership boundary. The setup docs already document `config.owner_id` as the correct owner scope for instance-level channel secrets, so this PR keeps the runtime fix focused on the unsafe fallback behavior.

### DIFF
--- a/src/channels/wasm/wrapper.rs
+++ b/src/channels/wasm/wrapper.rs
@@ -1077,15 +1077,11 @@ impl WasmChannel {
         .await;
     }
 
-    /// Load broadcast metadata from settings store on startup.
+    /// Load broadcast metadata from the owner-scoped settings store on startup.
     ///
-    /// # Legacy migration (remove after ownership model rollout — tracked in #2100)
-    ///
-    /// If no metadata is found under `self.owner_scope_id`, a second lookup
-    /// under `"default"` is attempted for backward compatibility with instances
-    /// that stored broadcast metadata before the ownership model migration.
-    /// Remove this fallback once all deployments have run the
-    /// `migrate_default_owner` bootstrap step and restarted at least once.
+    /// Broadcast metadata is owner-scoped runtime state. If the configured
+    /// owner scope has no stored value, we intentionally leave the in-memory
+    /// state empty rather than falling back to `default`.
     async fn load_broadcast_metadata(&self) {
         if let Some(ref store) = self.settings_store {
             match store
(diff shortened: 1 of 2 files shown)


henrypark133 (approved):
## Review: fail-closed ownership fix looks correct

I verified the two ownership-scope changes in the checked-out code and did not find a blocker.

Positives:
- WASM tool credential resolution now uses `DefaultFallback::Denied`, so the runtime no longer borrows `default`-scoped secrets when a user-specific secret is missing.
- WASM channel startup only reads owner-scoped broadcast metadata and no longer probes `default`.
- The new regressions cover both the tool path and the channel-startup path, which is the right caller-level coverage for this boundary.

Residual risk:
- Approval is based on code inspection plus the new targeted tests; I did not run the full suite from this review pass.

PR #2466, opened 2026-04-14
fix(web): infer gateway thread id from response metadata

### LINKED ISSUE #2405: [QA] Gateway routing error: broadcast() requires a thread_id

G7CNF:
Resolves #2405 by allowing GatewayChannel::broadcast() to recover thread context from response.metadata.notify_thread_id before returning MissingRoutingTarget. This keeps explicit routing failures for truly unscoped messages while letting broadcasted responses that already carry routing metadata reach the correct thread.

Validation:
- cargo fmt --all -- --check
- cargo test -p ironclaw --lib gateway_broadcast_with_notify_thread_id_metadata_succeeds -- --nocapture
- cargo test -p ironclaw --lib gateway_broadcast_without_thread_id_returns_error -- --nocapture

### DIFF
--- a/src/channels/web/mod.rs
+++ b/src/channels/web/mod.rs
@@ -62,6 +62,18 @@ use self::server::GatewayState;
 use self::sse::SseManager;
 use self::types::AppEvent;
 
+fn response_thread_id(response: &OutgoingResponse) -> Option<String> {
+    response.thread_id.clone().or_else(|| {
+        response
+            .metadata
+            .get("notify_thread_id")
+            .and_then(|v| v.as_str())
+            .map(str::trim)
+            .filter(|value| !value.is_empty())
+            .map(ToOwned::to_owned)
+    })
+}
+
 fn build_gateway_auth_manager(
     state: &GatewayState,
 ) -> Option<Arc<crate::bridge::auth_manager::AuthManager>> {
@@ -758,12 +770,12 @@ impl Channel for GatewayChannel {
         user_id: &str,
         response: OutgoingResponse,
     ) -> Result<(), ChannelError> {
-        let thread_id = match response.thread_id {
+        let thread_id = match response_thread_id(&response) {
             Some(tid) => tid,
             None => {
                 return Err(ChannelError::MissingRoutingTarget {
                     name: "gateway".to_string(),
-                    reason: "broadcast() requires a thread_id on the response".to_string(),
+                    reason: "broadcast() requires a thread_id on the response or notify_thread_id in response.metadata".to_string(),
                 });
             }
         };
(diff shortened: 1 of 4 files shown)


serrrfirat:
Can we get playwright test for this?

standardtoaster (commented):
Heads up -- #2444 is also open against this same issue and takes a different approach: a DB fallback via `get_or_create_assistant_conversation()` when no thread_id is present, so broadcast always has somewhere to land. It's been approved by serrrfirat and is waiting on re-review after test additions.

Worth coordinating so we don't end up with conflicting fixes. Is there a scenario where `notify_thread_id` in metadata is needed beyond what the DB fallback would cover?

henrypark133 (commented):
## Review: this looks superseded by merged #2444

I checked the current PR against the surrounding review history. The original overlap concern from the thread is now concrete: `#2444` has already merged, and it changed the same `src/channels/web/mod.rs` broadcast routing surface plus the same `no_silent_drop` tests with a different fix strategy (assistant-thread fallback for threadless broadcasts).

Because of that, I’m not comfortable approving this branch in its current form without first rebasing it onto current `staging` and re-evaluating what gap, if any, still remains after `#2444`.

Residual risk:
- As written, this PR is reviewing against an outdated contract for gateway routing, not just an isolated bugfix slice.

serrrfirat:
Thank you for addressing silently lost gateway responses caused by missing thread routing. We are closing this PR because Reborn now treats thread identity as a typed, persistent part of the execution and delivery scope.

Reborn carries `ThreadId` through thread persistence, `TurnScope`, run state, trigger-run history, projections, and outbound delivery. Trigger threads are looked up directly by `(tenant_id, thread_id)` through `TriggerRepository::find_trigger_run_by_thread_id`, and delivery reads the finalized assistant message using that canonical scope. It no longer recovers identity from loosely typed response metadata or falls back during broadcast.

Your regression scenario is still valuable. We would be happy to have you contribute Reborn tests that prove a finalized reply remains bound to the correct thread across trigger execution, restart, or external delivery. Thank you for finding the silent-drop path.

PR #2471, opened 2026-04-14
fix(channels): unify hot-activation owner_id type and capabilities fallback

serrrfirat:
## Summary

Follow-up to #2349 addressing two "Concerning" items from @henrypark133's [re-review](https://github.com/nearai/ironclaw/pull/2349#pullrequestreview-4100620233):

- **Type consistency**: `build_wasm_channel_runtime_config_updates` now takes `Option<&str>` and injects `owner_id` as `Value::String`, matching the boot path. Previously the hot-activation path injected `Value::Number` while the boot path used `Value::String` — WASM modules received different JSON types depending on which path ran.
- **Capabilities fallback parity**: `complete_loaded_wasm_channel_activation` and `refresh_active_channel` now consult `capabilities.json` via the extracted `owner_id_from_capabilities()` helper when neither runtime HashMap nor settings store have an owner_id. Previously only the boot path had this fallback.
- **Telegram WASM module**: Changed `owner_id: Option<i64>` to `Option<String>` with a custom deserializer that accepts both string and number JSON (all other channels already use `Option<String>`).

Depends on #2349 (branched from its HEAD).

## Test plan

- [x] Existing `owner_actor_id_*` tests in `setup.rs` still pass (pure refactor of extraction)
- [x] Updated `test_telegram_hot_activation_runtime_config_includes_owner_id` asserts `Value::String`
- [x] New `test_hot_activation_uses_capabilities_owner_id_fallback` drives `complete_loaded_wasm_channel_activation` with capabilities-only owner_id and verifies both `owner_actor_id` and config injection
- [x] New telegram tests for string and numeric JSON deserialization
- [x] `cargo test --lib` — 4712 passed (1 pre-existing failure unrelated to this PR)


henrypark133 (dismissed):
## Review: hot-activation owner_id parity

The current head makes the boot path and hot-activation path agree on both owner-id typing and capabilities fallback, and the added tests cover numeric/string/null capability values plus the actual hot-activation call site. This is still stacked on `#2349`, so merge order matters, but I did not find a new blocker in this follow-up diff.

ilblackdragon (approved):

PR #2483, opened 2026-04-15
feat(engine): add code execution failure categorization instrumentation

serrrfirat:
## Summary
- Adds `CodeExecutionFailure` enum (7 variants: SyntaxError, RuntimeError, NameLookup, VmPanic, ResourceLimit, ToolError, OsDenied) to classify why code execution fails
- Threads `failure` through `CodeExecutionResult` (replacing `had_error: bool`) and emits structured `CodeExecutionFailed` events for aggregate analysis of failure modes (Monty limitation vs LLM logic error vs tool dispatch failure)
- Upgrades trace analysis to use structured events with severity escalation (VmPanic/ResourceLimit → Error), with backward-compatible fallback for pre-instrumentation threads

### Caller audit (Err → Ok(failure=VmPanic) shift)
`execute_code` / `execute_code_with_skills` previously returned `Err(EngineError::Effect)` on VM panic; now returns `Ok(CodeExecutionResult { failure: Some(VmPanic) })`. The only production call site is `orchestrator.rs:handle_execute_code_step` which correctly dispatches on `result.failure`. Test-only callers in `scripting.rs` were also updated. No other callers exist.

### Known limitation: mixed-era fallback
The trace analyzer's backward-compatible fallback (message-scraping for pre-instrumentation threads) is all-or-nothing: if a thread has *any* `CodeExecutionFailed` event, message-scraping is skipped entirely. Errors from pre-instrumentation steps in a mixed-era thread will go unreported. This is acceptable during the transition period since all new threads will have full instrumentation.

## Test plan
- [x] 11 new unit tests for error classifier, code hash, and trace detection
- [x] `cargo test -p ironclaw_engine --lib` — 395 pass
- [x] `cargo clippy --all --all-features` — zero engine warnings

### DIFF
--- a/crates/ironclaw_engine/src/executor/orchestrator.rs
+++ b/crates/ironclaw_engine/src/executor/orchestrator.rs
@@ -774,6 +774,27 @@ async fn handle_execute_code_step(
                     let _ = tx.send(failed_event.clone());
                 }
                 thread.events.push(failed_event);
+
+                // Emit structured CodeExecutionFailed event for instrumentation.
+                // This enables aggregate analysis of WHY code execution fails
+                // (Monty limitation vs LLM logic error vs tool dispatch failure).
+                if let Some(ref category) = result.failure_category {
+                    let error_text: String = result.stdout.chars().take(500).collect();
+                    let instrumentation_event = ThreadEvent::new(
+                        thread.id,
+                        EventKind::CodeExecutionFailed {
+                            step_id: exec_ctx.step_id,
+                            category: category.clone(),
+                            error: error_text,
+                            code_hash: Some(crate::executor::scripting::code_hash(&code)),
+                            duration_ms: 0, // not timed at this layer
+                        },
+                    );
+                    if let Some(tx) = event_tx {
+                        let _ = tx.send(instrumentation_event.clone());
+                    }
+                    thread.events.push(instrumentation_event);
+                }
             }
             thread.updated_at = chrono::Utc::now();
 
--- a/crates/ironclaw_engine/src/executor/orchestrator.rs
+++ b/crates/ironclaw_engine/src/executor/orchestrator.rs
@@ -751,15 +751,10 @@ async fn handle_execute_code_step(
             // fall back to the LLM via the result dict and never warn callers.
             if let Some(ref category) = result.failure {
                 let error_msg = if !result.stdout.is_empty() {
-                    // Take the *last* 500 chars — error tracebacks appear at the
-                    // end of stdout, after any print() output.
-                    let char_count = result.stdout.chars().count();
-                    let snippet: String = if char_count > 500 {
-                        result.stdout.chars().skip(char_count - 500).collect()
-                    } else {
-                        result.stdout.clone()
-                    };
-                    format!("CodeAct execution failed: {snippet}")
+                    format!(
+                        "CodeAct execution failed: {}",
+                        tail_chars(&result.stdout, 500)
+                    )
                 } else {
                     "CodeAct execution failed (no stdout)".to_string()
                 };
@@ -786,12 +781,7 @@ async fn handle_execute_code_step(
                 // Emit structured CodeExecutionFailed event for instrumentation.
                 // This enables aggregate analysis of WHY code execution fails
                 // (Monty limitation vs LLM logic error vs tool dispatch failure).
-                let char_count = result.stdout.chars().count();
-                let error_text: String = if char_count > 500 {
-                    result.stdout.chars().skip(char_count - 500).collect()
-                } else {
-                    result.stdout.clone()
-                };
+                let error_text = tail_chars(&result.stdout, 500);
                 let instrumentation_event = ThreadEvent::new(
                     thread.id,
                     EventKind::CodeExecutionFailed {
@@ -2226,6 +2216,19 @@ fn action_calls_to_python_json(calls: &[ActionCall]) -> Vec<serde_json::Value> {
             }
         })
         .collect()
+}
+
+/// Extract the last `n` characters from `s`.
+///
+/// Error tracebacks appear at the end of stdout, after any `print()` output.
+/// Using the head would capture the print statements instead of the error.
+fn tail_chars(s: &str, n: usize) -> String {
+    let char_count = s.chars().count();
+    if char_count > n {
+        s.chars().skip(char_count - n).collect()
+    } else {
+        s.to_owned()
+    }
 }
 
 /// Build a PII-safe summary of an `action_calls` JSON value for log output.
--- a/crates/ironclaw_engine/src/executor/orchestrator.rs
+++ b/crates/ironclaw_engine/src/executor/orchestrator.rs
@@ -720,6 +720,7 @@ async fn handle_execute_code_step(
     };
 
     // Run user code in a nested Monty VM (same pattern as rlm_query)
+    let code_start = std::time::Instant::now();
     match Box::pin(execute_code(
         &code,
         thread,
@@ -748,9 +749,16 @@ async fn handle_execute_code_step(
             // error, etc.), surface it as an ActionFailed event so traces and
             // observers see the failure. Without this, parse errors silently
             // fall back to the LLM via the result dict and never warn callers.
-            if result.had_error {
+            if let Some(ref category) = result.failure {
                 let error_msg = if !result.stdout.is_empty() {
-                    let snippet: String = result.stdout.chars().take(500).collect();
+                    // Take the *last* 500 chars — error tracebacks appear at the
+                    // end of stdout, after any print() output.
+                    let char_count = result.stdout.chars().count();
+                    let snippet: String = if char_count > 500 {
+                        result.stdout.chars().skip(char_count - 500).collect()
+                    } else {
+                        result.stdout.clone()
+                    };
                     format!("CodeAct execution failed: {snippet}")
                 } else {
                     "CodeAct execution failed (no stdout)".to_string()
@@ -778,23 +786,26 @@ async fn handle_execute_code_step(
                 // Emit structured CodeExecutionFailed event for instrumentation.
                 // This enables aggregate analysis of WHY code execution fails
                 // (Monty limitation vs LLM logic error vs tool dispatch failure).
-                if let Some(ref category) = result.failure_category {
-                    let error_text: String = result.stdout.chars().take(500).collect();
-                    let instrumentation_event = ThreadEvent::new(
-                        thread.id,
-                        EventKind::CodeExecutionFailed {
-                            step_id: exec_ctx.step_id,
-                            category: category.clone(),
-                            error: error_text,
-                            code_hash: Some(crate::executor::scripting::code_hash(&code)),
-                            duration_ms: 0, // not timed at this layer
-                        },
-                    );
-                    if let Some(tx) = event_tx {
-                        let _ = tx.send(instrumentation_event.clone());
-                    }
-                    thread.events.push(instrumentation_event);
+                let char_count = result.stdout.chars().count();
+                let error_text: String = if char_count > 500 {
+                    result.stdout.chars().skip(char_count - 500).collect()
+                } else {
+                    result.stdout.clone()
+                };
+                let instrumentation_event = ThreadEvent::new(
+                    thread.id,
+                    EventKind::CodeExecutionFailed {
+                        step_id: exec_ctx.step_id,
+                        category: category.clone(),
+                        error: error_text,
+                        code_hash: Some(crate::executor::scripting::code_hash(&code)),
+                        duration_ms: code_start.elapsed().as_millis() as u64,
+                    },
+                );
+                if let Some(tx) = event_tx {
+                    let _ = tx.send(instrumentation_event.clone());
                 }
+                thread.events.push(instrumentation_event);
             }
             thread.updated_at = chrono::Utc::now();
 
@@ -816,7 +827,7 @@ async fn handle_execute_code_step(
                 "stdout": result.stdout,
                 "action_results": action_results,
                 "final_answer": result.final_answer,
-                "had_error": result.had_error,
+                "had_error": result.failure.is_some(),
                 "pending_gate": result.need_approval.as_ref().map(|na| {
                     match na {
                         ThreadOutcome::GatePaused { gate_name, action_name, call_id, parameters, resume_kind, resume_output } => serde_json::json!({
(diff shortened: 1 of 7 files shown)


crates/ironclaw_engine/src/executor/orchestrator.rs
```diff
@@ -774,6 +774,27 @@ async fn handle_execute_code_step(
...
                 }
                 thread.events.push(failed_event);
+
+                // Emit structured CodeExecutionFailed event for instrumentation.
+                // This enables aggregate analysis of WHY code execution fails
+                // (Monty limitation vs LLM logic error vs tool dispatch failure).
+                if let Some(ref category) = result.failure_category {
+                    let error_text: String = result.stdout.chars().take(500).collect();
+                    let instrumentation_event = ThreadEvent::new(
+                        thread.id,
```
serrrfirat: **Medium Severity** — `error_text` takes first 500 chars of stdout instead of last

```rust
let error_text: String = result.stdout.chars().take(500).collect();
```

When code has print output before the error (e.g., a loop printing 100 items then hitting a `NameError`), the first 500 chars capture the print statements, **not** the error traceback. The project's own `compact_output_metadata()` in `scripting.rs:88-96` takes the **last** N chars for exactly this reason.

**Suggestion:** Extract from the tail instead:
```rust
let chars: Vec<char> = result.stdout.chars().collect();
let start = chars.len().saturating_sub(500);
let error_text: String = chars[start..].iter().collect();
```

Note: the existing `ActionFailed` path at line 806 (`result.stdout.chars().take(500)`) has the same issue — worth fixing both with a shared tail-extract helper.

crates/ironclaw_engine/src/executor/scripting.rs
```diff
@@ -641,6 +656,9 @@ pub async fn execute_code_with_skills(
                             recursive_tokens,
                             final_answer: None,
                             had_error,
```
serrrfirat: **Medium Severity** — GatePaused sets `failure_category` but the event is never emitted

At this return site, the code sets `failure_category: Some(GatePause)` but `had_error` is `false` (verified: every `had_error = true` in this function precedes an early return, so it's still `false` when GatePaused is reached).

In `orchestrator.rs`, the `CodeExecutionFailed` event emission is nested inside `if result.had_error { ... }`, so the GatePause category is **silently discarded** — it's dead data that will never appear in aggregate analysis.

**Suggestion:** Either:
1. Don't set `failure_category` here (a gate pause is a suspension, not a failure), or
2. Emit `CodeExecutionFailed` in `orchestrator.rs` based on `failure_category.is_some()` independently of `had_error`, so gate pauses are tracked in the instrumentation

ilblackdragon:
## Review

Useful instrumentation, low blast radius. Three things worth fixing before merge: (1) error-text truncation captures the wrong end, (2) `duration_ms: 0` is a misleading placeholder, (3) classifier is too loose and untested against real Monty errors.

### Correctness

- **Error preview takes the wrong end of stdout.** `orchestrator.rs:~785` does `result.stdout.chars().take(500).collect()` — but tracebacks land at the *end* of stdout (after any `print()` output). For a script that prints 600 chars then errors, the recorded `error` field will contain only the prints and miss the actual failure. Use the last 500 chars, or extract the trailing `Traceback (most recent call last):` block.

- **`duration_ms: 0` with comment "not timed at this layer".** Half-instrumented telemetry is worse than absent telemetry — aggregations will silently average toward zero. Either thread an actual elapsed time through `CodeExecutionResult` (`Instant::now()` around the executor call), or drop the field until you can populate it.

- **`VmPanic` is declared but never emitted in the diff.** No code path sets `failure_category = Some(VmPanic)`. If the catch_unwind site lives in a section not in the diff, fine — otherwise this variant is dead. Confirm and add a test.

- **`classify_runtime_error` substring matches are too loose.**
  - `lower.contains(\"syntax\")` will misclassify a runtime error whose message happens to include the word \"syntax\". Tighten to `\"syntaxerror\"` or anchor on `:`.
  - Check order matters: `\"Resource limit: syntax error in ...\"` would be tagged `SyntaxError` because syntax is checked first. Reorder so the most specific (resource-limit, OS-denied) precedes the most generic (syntax → runtime).
  - `lower.contains(\"os\") && lower.contains(\"denied\")` will fire on benign text like `\"close: permission denied\"` (\"os\" appears as a substring of many words).

- **`had_error` + `failure_category` semantic drift.** Several Ok paths return `had_error: true, failure_category: None`. After this PR the canonical \"did it fail\" signal is ambiguous — orchestrator only emits `CodeExecutionFailed` when `failure_category.is_some()`, so a `had_error: true / failure_category: None` result fails *silently* from the instrumentation's perspective. Either make `failure_category` required when `had_error == true` (encode the invariant: `enum CodeExecutionOutcome { Success(...), Failed(CodeExecutionFailure, ...) }`), or default to a `RuntimeError`/`Unknown` variant.

- **Serialization vs Display casing mismatch.** `#[derive(Serialize)]` on `CodeExecutionFailure` yields `\"SyntaxError\"` (PascalCase) on the wire, while `Display` yields `\"syntax_error\"` (snake_case). Trace category strings are `code_syntax_error`; persisted events are `\"category\": \"SyntaxError\"`. Pick one (snake_case throughout via `#[serde(rename_all = \"snake_case\")]`).

### Trace fallback
- The fallback runs only when `code_failures.is_empty()`. A mixed-era thread (some pre-, some post-instrumentation events) silently skips the pre-instrumentation portion. Probably rare; worth a comment acknowledging it.
- Fallback patterns now include the literal `RuntimeError` — but that string rarely appears verbatim in Monty stderr. Will likely fire on user-printed text more than real errors. Verify against a real corpus.

### Persistence / forward-compat
- New `EventKind::CodeExecutionFailed` variant: confirm the v2 persistence layer (libSQL + Postgres event tables) round-trips this without a migration. If events are stored as JSON in a column, deserialization in **older** binaries reading **new** rows will fail unless `EventKind` uses `#[serde(other)]` or similar — a real concern during rolling deploys.

### Conventions
- No `.unwrap()`/`.expect()` in production paths. ✅
- Repeated long path `crate::types::step::CodeExecutionFailure::*` — add a `use` at file top.
- `code_hash` uses `DefaultHasher` — deterministic across processes (unlike `RandomState`), safe for cross-process correlation. ✅ A `u64`-truncated hash has a ~2^32 birthday collision; fine for dedup, not for security. Add a doc note.
- `code_hash` is `pub` — intentional for orchestrator to call; fine.

### Test coverage
- 11 new unit tests on classifier/hash/trace are good.
- **Missing**: no test that the orchestrator actually emits the `CodeExecutionFailed` event. Per `.claude/rules/testing.md` (\"Test Through the Caller\"), the classifier is a predicate gating an event side effect — needs a test driving `handle_execute_code_step` and asserting both the broadcast `tx` and `thread.events` receive the event.
- **Missing**: integration test against the real Monty runtime that classifies each enum variant from a real failure (`SyntaxError` from bad python, `ResourceLimit` from `while True: pass`, `OsDenied` from `import os; os.listdir()`). Substring matching over Monty's exact error text is brittle — pin it with real outputs so a Monty version bump that changes wording surfaces in CI rather than silently degrading the categorization.
- No e2e test that surfaces aggregated failure counts the way analytics will consume them.

### Suggestions (file:line)
- `executor/orchestrator.rs:~785` — slice the *tail* of stdout, not the head.
- `executor/orchestrator.rs:~790` — wrap the executor call with `Instant::now()` and pass real `duration_ms` through `CodeExecutionResult`. Or drop the field.
- `executor/scripting.rs:~835` — reorder `classify_runtime_error` checks (resource-limit, os-denied, syntax, fallthrough); tighten `\"syntax\"` → `\"syntaxerror\"`; replace `\"os\" && \"denied\"` heuristic with explicit error-class strings Monty actually emits.
- `types/step.rs:~140` — add `#[serde(rename_all = \"snake_case\")]` so wire and `Display` agree.
- `types/step.rs` — consider replacing `had_error: bool` + `failure_category: Option<_>` with a sum type to make invalid states unrepresentable.
- Add an integration test in `crates/ironclaw_engine/tests/` driving a real failed exec end-to-end through the orchestrator against real Monty.

serrrfirat:
## Addressed review feedback (88d3f630)

All feedback from @ilblackdragon, @gemini-code-assist[bot], and self-review has been addressed:

### Correctness
- **error_text truncation** — now takes last 500 chars (where tracebacks are), not first
- **`duration_ms: 0`** — threads real `Instant::now()` timing through the event
- **VmPanic never emitted** — all 9 `catch_unwind` `Err(_)` paths now return `Ok(CodeExecutionResult { failure: Some(VmPanic) })` instead of propagating `EngineError::Effect`
- **Classifier too loose** — tightened `"syntax"` → `"syntaxerror"`, removed `"os" && "denied"` substring match, reordered checks (specific before generic)
- **`had_error` + `failure_category` semantic drift** — replaced with single `failure: Option<CodeExecutionFailure>` field. `GatePause` variant removed (gate pauses are suspensions, not failures)
- **Serialize vs Display mismatch** — added `#[serde(rename_all = "snake_case")]`

### Stability
- **`DefaultHasher`** — replaced with inline FNV-1a (stable across Rust versions)
- **Forward-compat** — added `#[serde(other)] Unknown` variant to `EventKind`

### Tests
- Added orchestrator caller test verifying `CodeExecutionFailed` event emission
- Added classifier edge case tests (syntax-word-not-syntaxerror, VmPanic serialization)

### Style
- `to_ascii_lowercase()` per convention
- Added `use CodeExecutionFailure` import, removed verbose paths
- Expanded fallback comment acknowledging mixed-era thread limitation
- Fixed misleading `CodeExecutionFailed` docstring

**All 395 tests pass, `cargo fmt` clean.**

serrrfirat:
@ilblackdragon — addressed all your feedback in 88d3f630. Here's a walkthrough of the design decisions:

---

### 1. `had_error` + `failure_category` → sum type

Replaced both fields with a single `failure: Option<CodeExecutionFailure>`. This makes invalid states unrepresentable — every failure has a category, and `failure.is_some()` replaces `had_error`. The JSON key remains `"had_error"` in the orchestrator result dict for Python compat (`default.py:659`).

Removed the `GatePause` variant entirely. Gate pauses aren't failures — they're suspensions represented by `need_approval: Some(ThreadOutcome::GatePaused { .. })`. The old code set `failure_category: Some(GatePause)` with `had_error: false`, which was dead data (the `CodeExecutionFailed` event was gated on `had_error`). Clean separation now: `failure` tracks errors, `need_approval` tracks suspensions.

### 2. VmPanic emission

All 9 `catch_unwind Err(_)` paths now return `Ok(CodeExecutionResult { failure: Some(VmPanic), ... })` instead of `Err(EngineError::Effect)`. Previously these panics:
- Propagated as `EngineError` → thread failure, no instrumentation
- Made `VmPanic` a dead variant

Now they:
- Flow through the orchestrator's event emission path → proper `CodeExecutionFailed` event
- Include stdout captured up to the panic point (valuable for diagnosing Monty bugs)
- Show up in trace analysis with `IssueSeverity::Error`

This is safe because each `Err(_)` path is a `return` — the Monty VM handle is consumed and dropped, no corrupt state leaks.

### 3. Error text truncation (first → last)

Both `ActionFailed` and `CodeExecutionFailed` now take the **last** 500 chars of stdout. Python tracebacks appear at the end, after any `print()` output. This matches what `compact_output_metadata()` already does for the LLM context.

### 4. Classifier tightening

Reordered checks from most specific to most generic:
1. `ResourceLimit` (multi-keyword: "timed out", "timeout", "memory limit", "fuel", etc.)
2. `OsDenied` (only `"os operations are not permitted"` and `"oserror"` — removed the loose `contains("os") && contains("denied")` that matched "compose", "diagnosis", etc.)
3. `SyntaxError` (`"syntaxerror"` instead of `"syntax"` — the actual Python exception type)
4. `RuntimeError` (catch-all)

### 5. Stable hash

FNV-1a (64-bit) inline — 6 lines, no new dependency, deterministic across Rust versions forever. `DefaultHasher` is explicitly documented as unstable across versions.

### 6. Serde consistency

`#[serde(rename_all = "snake_case")]` on `CodeExecutionFailure` — wire format now matches `Display` output. Since this type is new in this PR, no backward compat concern.

`#[serde(other)] Unknown` on `EventKind` — older binaries reading events from newer binaries will deserialize unknown variants as `Unknown` instead of failing. Unit variant per serde's externally-tagged enum requirement.

### 7. Real timing

`Instant::now()` wraps the `execute_code()` call in the orchestrator. `duration_ms` is now the actual wall-clock time of code execution, not a placeholder zero.

### Not addressed (noted as follow-up)

- **Integration test against real Monty runtime** for each failure category — worth doing but requires pinning specific Monty error formats, better as a separate PR
- **e2e test for aggregated failure counts** — analytics consumer doesn't exist yet

ilblackdragon:
## Review

### Overview
Introduces `CodeExecutionFailure` and threads it through `CodeExecutionResult` (replacing `had_error: bool`). Adds a `CodeExecutionFailed` structured event emitted from `orchestrator::handle_execute_code_step`, upgrades `trace.rs` to consume it with a message-scraping fallback, and adds `EventKind::Unknown` for forward-compat deserialization.

### Strengths
- Clean shape: category travels with the result; orchestrator emits both `ActionFailed` and the new `CodeExecutionFailed` so existing observers aren't broken.
- `#[serde(other)] Unknown` on `EventKind` is a nice rolling-deploy hedge.
- `classify_runtime_error` is well-tested — specific vs. substring false positives are covered (e.g. "unexpected syntax" is not `SyntaxError`).
- `code_hash` uses FNV-1a rather than `DefaultHasher` — correct call for cross-version stability.
- Tests exercise the caller (`execute_code_step_emits_code_execution_failed_event`), not just the helper.
- `trace.rs` severity escalation (VmPanic/ResourceLimit → `Error`) is sensible.

### Correctness concerns

**1. Behavior change on VM panic** (`scripting.rs`, 7 sites). Previously `catch_unwind` failures returned `Err(EngineError::Effect { … })`; now they return `Ok(CodeExecutionResult { failure: Some(VmPanic), … })`. This is a semantic shift — any caller that matched on `Err` to escalate/abort will now silently continue. The orchestrator path handles this correctly via `result.failure`, but please audit other call sites of `execute_code` / `execute_code_with_skills`.

**2. PR description mismatch — `GatePause` variant missing.** The description lists 8 variants including `GatePause`, but `CodeExecutionFailure` has only 7 (no `GatePause`). Gate-paused execution sets `failure: None`, which is consistent, so the description is just wrong. Please update.

**3. Unrelated migration churn.** `migrations/checksums.lock` adds `V24__llm_calls_created_at_index` but no `V24` migration file is in this PR. Almost certainly a stray artifact from rebase/merge — drop it or it'll produce a phantom checksum mismatch against whatever branch actually adds that migration.

**4. Fallback branch is all-or-nothing.** Per the comment in `trace.rs`: a thread with *any* `CodeExecutionFailed` event skips message-scraping entirely, so errors from pre-instrumentation steps in a mixed-era thread go unreported. Consider scraping messages that predate the earliest `CodeExecutionFailed` event, or accept a brief dup window. At minimum call this out as a known limitation in the PR body.

### Nits / quality
- The "last 500 chars of stdout" logic is duplicated in `orchestrator.rs` — extract a small `tail_chars(&str, usize)` helper (called twice back-to-back in the same function).
- `classify_runtime_error`: `lower.contains("fuel")` is broad; a runtime error mentioning the word "fuel" would be miscategorized as `ResourceLimit`. Tightening to `"out of fuel"` / `"fuel exhausted"` removes the risk at zero cost to the tests.
- `CodeExecutionFailure::Display` + `Serialize(rename_all = "snake_case")` duplicate the snake-case mapping in two places. A single source (e.g. `as_str()` used by both) avoids drift if a variant is added — you even added a test specifically to guard this.
- `EventKind::Unknown` has no doc on expected consumer behavior (probably: log + skip). Worth a one-liner.
- Typo in test message: `"syntax error should set had_error"` — the field is now `failure`.

### Security / perf
No new I/O or allocations in a hot path. `chars().count()` + `chars().skip()` walks stdout twice on error; acceptable for a failure path capped at ~500-char output. `error` is truncated in the event; `code_hash` is non-cryptographic but fine for correlation.

### Summary
Solid, well-tested instrumentation. Three things to resolve before merge:
1. Drop the `checksums.lock` change (or explain it).
2. Audit callers affected by the `Err → Ok(failure=VmPanic)` shift.
3. Fix the PR description (no `GatePause` variant exists).

serrrfirat:
Addressed review feedback from https://github.com/nearai/ironclaw/pull/2483#issuecomment-4252703045:

**Code fixes (d5baf700):**
1. **Extract `tail_chars` helper** — deduplicated the last-500-chars logic in `handle_execute_code_step` (was copy-pasted for both `ActionFailed` and `CodeExecutionFailed` events)
2. **Tighten fuel match** — `contains("fuel")` → `contains("out of fuel") || contains("fuel exhausted")` to avoid miscategorizing runtime errors mentioning "fuel"
3. **Fix test typo** — `"should set had_error"` → `"should set failure"`
4. **Fix clippy warnings** — `!x.is_some()` → `x.is_none()`, removed tautological `u64 >= 0` assertion

**PR description updates:**
- Fixed variant count (7, not 8 — no `GatePause` variant)
- Added caller audit section confirming only `orchestrator.rs` calls `execute_code` in production
- Documented mixed-era fallback as a known limitation
- `checksums.lock` was already dropped in the previous fix commit

henrypark133 (approved):
## Review: Code execution failure categorization

No verified findings in the current diff. The change threads structured code-execution failure categories from the scripting runtime into the orchestrator caller and trace analyzer, and it keeps the fallback message-based detection for older threads that predate the new event.

PR #2486, opened 2026-04-15
feat(gateway): show commit hash in version for non-tagged builds

serrrfirat:
## Summary

- Adds git commit hash to the user info popover when the build is not from a tagged release
- Tagged releases show just "IronClaw v0.25.0"; dev builds show "IronClaw v0.25.0 (88d3f63)" or "IronClaw v0.25.0 (88d3f63-dirty)"
- Enables users to report exact build versions when filing bugs

## Changes

- **`build.rs`** — new `emit_git_metadata()` captures short commit hash and dirty flag at compile time via `GIT_COMMIT_HASH` / `GIT_DIRTY` env vars
- **`src/channels/web/server.rs`** — adds `commit_hash` field to `/api/gateway/status` response
- **`crates/ironclaw_gateway/static/app.js`** — renders commit hash in the version line when present

## Test plan

- [ ] Build from a non-tagged commit → verify popover shows `IronClaw v0.25.0 (<hash>)`
- [ ] Build with uncommitted changes → verify it shows `(<hash>-dirty)`
- [ ] Build from a tagged commit (e.g. `git tag v0.25.0`) → verify only version is shown, no hash
- [ ] `cargo clippy --all --all-features` passes with zero warnings


serrrfirat:
## Addressed review feedback

**gemini-code-assist** — worktree-safe git directory resolution in `build.rs`:

- Replaced hardcoded `.git/HEAD` path lookup with `git rev-parse --git-dir`, which correctly resolves the git metadata directory in all environments (regular repos, worktrees, submodules)
- `cargo:rerun-if-changed` directives now use the resolved paths instead of relative `.git/...` paths

henrypark133 (changes requested):
## Review: Commit hash in gateway version

One blocker remains in the current diff.

### Concerning: build metadata stops updating on new commits in git worktrees
**File:** `build.rs:142`
The new watcher resolves the branch ref by joining `ref: refs/heads/...` onto `git rev-parse --git-dir`. In a git worktree, `--git-dir` points at `.git/worktrees/<name>`, but the branch ref itself lives under the common git dir (`git rev-parse --git-common-dir`). That means `git_dir.join(refpath)` never exists in worktrees, so Cargo only watches `.../worktrees/<name>/HEAD`. New commits on the same checked-out branch do not change that HEAD file, so `build.rs` will not rerun and the exposed commit hash can go stale until a clean rebuild or checkout change.

**Suggested fix:** resolve refs relative to `--git-common-dir` (or emit both `--git-dir` and `--git-common-dir` watchers) so branch advances in worktrees retrigger the build script.

serrrfirat:
Addressed review feedback from @henrypark133:

**Fix:** `build.rs` now resolves branch refs via `git rev-parse --git-common-dir` instead of `--git-dir`. In worktrees, `--git-dir` points to `.git/worktrees/<name>` but branch refs (e.g. `refs/heads/main`) live under the common git directory. The watcher now looks up refs in `git_common_dir` first, with a fallback to `git_dir` for non-worktree repos.

Commit: 7dab8224

henrypark133 (dismissed):
## Review: commit hash in gateway version

The worktree watcher now uses `--git-common-dir` for branch refs, so the earlier stale-metadata issue in git worktrees is addressed, and the gateway response/UI only surfaces the hash on non-tagged builds.

ilblackdragon (approved):

PR #2487, opened 2026-04-15
fix(llm): handle duplicate reasoning_content field in API responses

serrrfirat:
## Summary
- Some NEAR AI models return both `reasoning` and `reasoning_content` fields in the same chat completion message
- The previous `#[serde(alias = "reasoning")]` on `reasoning_content` caused serde to treat them as the same field, producing a "duplicate field" JSON parse error
- Split into two separate struct fields and merge at point of use (preferring `reasoning_content` when both are present)
- Adds regression test for the dual-field scenario

## Test plan
- [x] All 5 existing reasoning tests pass (`cargo test --lib nearai_chat::tests::test_reasoning`)
- [x] New `test_both_reasoning_and_reasoning_content_accepted` test passes
- [x] `cargo clippy --lib` — zero warnings
- [ ] Manual test: send a message to a model that returns both fields

### DIFF
--- a/src/llm/nearai_chat.rs
+++ b/src/llm/nearai_chat.rs
@@ -498,11 +498,8 @@ impl LlmProvider for NearAiChatProvider {
 
         // Fall back to reasoning_content when content is null (same as
         // complete_with_tools — reasoning models may put the answer there).
-        let content = choice
-            .message
-            .content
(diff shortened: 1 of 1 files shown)


serrrfirat:
https://github.com/nearai/ironclaw/pull/2493 closes the issue

PR #2493, opened 2026-04-15
fix: duplicate `reasoning_content` fields in chat completions response

think-in-universe:
## Summary

- Fix a NEAR AI Chat response parsing failure caused by payloads that include both `reasoning` and `reasoning_content`. Introduced in https://github.com/nearai/ironclaw/commit/0588dd1bc334cb6642a12d1950240fb4a18ad392#diff-471da66860a01960fd55eeea85a2e28cd9c40492bf62f369a6fa1591d64fdc10
- Update `ChatCompletionResponseMessage` deserialization in `nearai_chat` to parse `reasoning` and `reasoning_content` as separate optional fields, then merge them explicitly in response handling.
- Preserve existing fallback behavior (use reasoning text when `content` is missing) while avoiding Serde duplicate-field errors.
- Keep tool-call safety behavior intact by not leaking reasoning text into tool-call responses.

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] CI/Infrastructure
- [ ] Security
- [ ] Dependencies

## Linked Issue

None

## Validation

- [ ] `cargo build`
- [ ] Relevant tests pass: Updated/verified logic in `nearai_chat` reasoning fallback and tool-call handling paths; full targeted test run was interrupted before completion.
- [x] Manual testing: Reproduced/analyzed failure mode from runtime payload and verified parser logic now accepts both fields without duplicate-key deserialization failure.

## Security Impact

None

## Database Impact

None

## Blast Radius

Touches `nearai_chat` response parsing and content fallback behavior for the NEAR AI backend. Risk is limited to NEAR AI Chat completions parsing and how text content is selected when providers return reasoning fields.

## Rollback Plan

Revert the `nearai_chat` response struct and fallback selection changes in this PR; this restores prior behavior immediately if regressions are detected.

## Review Follow-Through

Reviewer focus requested on:
- Whether fallback precedence should remain `content -> reasoning_content -> reasoning`.
- Whether additional coverage is needed for payloads that include both fields simultaneously in one message object.

---

**Review track**: C

### DIFF
--- a/src/llm/nearai_chat.rs
+++ b/src/llm/nearai_chat.rs
@@ -498,10 +498,14 @@ impl LlmProvider for NearAiChatProvider {
 
         // Fall back to reasoning_content when content is null (same as
         // complete_with_tools — reasoning models may put the answer there).
-        let content = choice
-            .message
-            .content
-            .or(choice.message.reasoning_content)
+        let ChatCompletionResponseMessage {
+            content,
+            reasoning_content,
+            reasoning,
+            ..
+        } = choice.message;
+        let content = content
+            .or(reasoning_content.or(reasoning))
             .unwrap_or_default();
         let finish_reason = match choice.finish_reason.as_deref() {
             Some("stop") => FinishReason::Stop,
@@ -577,9 +581,16 @@ impl LlmProvider for NearAiChatProvider {
                     provider: "nearai_chat".to_string(),
                 })?;
 
-        let tool_calls: Vec<ToolCall> = choice
-            .message
-            .tool_calls
+        let ChatCompletionResponseMessage {
+            content: message_content,
+            reasoning_content,
+            reasoning,
+            tool_calls: message_tool_calls,
+            ..
+        } = choice.message;
+        let reasoning_fallback = reasoning_content.or(reasoning);
+
+        let tool_calls: Vec<ToolCall> = message_tool_calls
             .unwrap_or_default()
             .into_iter()
             .map(|tc| {
@@ -601,9 +612,9 @@ impl LlmProvider for NearAiChatProvider {
         // leaking that into conversation history inflates context and
         // confuses the model.
         let content = if tool_calls.is_empty() {
-            choice.message.content.or(choice.message.reasoning_content)
+            message_content.or(reasoning_fallback)
         } else {
-            choice.message.content
+            message_content
         };
 
         let finish_reason = match choice.finish_reason.as_deref() {
@@ -1047,8 +1058,10 @@ struct ChatCompletionResponseMessage {
     /// Some models return chain-of-thought reasoning here instead of in
     /// `content`. vLLM/SGLang backends (used by NEAR AI) return the field
     /// as `reasoning`; other APIs (GLM-5, DeepSeek) use `reasoning_content`.
-    #[serde(default, alias = "reasoning")]
+    #[serde(default)]
     reasoning_content: Option<String>,
+    #[serde(default)]
+    reasoning: Option<String>,
     tool_calls: Option<Vec<ChatCompletionToolCall>>,
 }
 
@@ -1556,7 +1569,10 @@ mod tests {
         .unwrap();
 
         let choice = response.choices.into_iter().next().unwrap();
-        let content = choice.message.content.or(choice.message.reasoning_content);
+        let content = choice.message.content.or(choice
+            .message
+            .reasoning_content
+            .or(choice.message.reasoning));
 
         assert_eq!(
             content,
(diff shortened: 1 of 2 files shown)


src/llm/nearai_chat.rs
```diff
@@ -1047,8 +1058,10 @@ struct ChatCompletionResponseMessage {
     /// Some models return chain-of-thought reasoning here instead of in
     /// `content`. vLLM/SGLang backends (used by NEAR AI) return the field
     /// as `reasoning`; other APIs (GLM-5, DeepSeek) use `reasoning_content`.
-    #[serde(default, alias = "reasoning")]
+    #[serde(default)]
     reasoning_content: Option<String>,
+    #[serde(default)]
+    reasoning: Option<String>,
     tool_calls: Option<Vec<ChatCompletionToolCall>>,
```
serrrfirat, line 1065: **High Severity — Existing test `test_reasoning_content_not_leaked_into_tool_call_response` does not cover the `reasoning` field**

The test at ~line 1445 still uses `choice.message.reasoning_content` directly in its inline fallback logic, but never checks `choice.message.reasoning`. Since this PR added `reasoning` as a separate deserialized field that participates in the production fallback chain, this test would not catch a regression where `reasoning` (but not `reasoning_content`) leaks into a tool-call response.

The sibling test `test_reasoning_alias_not_leaked_into_tool_calls` (updated in this PR) correctly tests the `reasoning`-only path. But the original `test_reasoning_content_not_leaked_into_tool_call_response` at ~line 1445 was not updated to use the destructuring pattern and still ignores the `reasoning` field entirely.

**Suggestion:** Either update the test to mirror the destructuring pattern used in production code:

```rust
let ChatCompletionResponseMessage {
    content: message_content,
    reasoning_content,
    reasoning,
    tool_calls: message_tool_calls,
    ..
} = choice.message;
let reasoning_fallback = reasoning_content.or(reasoning);
```

Or add a fixture to the existing test with `"reasoning": "..."` alongside `"reasoning_content": "..."` and tool calls, asserting `content` is `None`.

src/llm/nearai_chat.rs
```diff
@@ -1556,7 +1569,10 @@ mod tests {
...
 
         let choice = response.choices.into_iter().next().unwrap();
-        let content = choice.message.content.or(choice.message.reasoning_content);
+        let content = choice.message.content.or(choice
+            .message
+            .reasoning_content
+            .or(choice.message.reasoning));
 
         assert_eq!(
             content,
```
serrrfirat, line 1595: **High Severity — Existing test `test_reasoning_content_used_for_text_response` does not verify `reasoning` fallback**

Same pattern as the tool-call leak test: the test at ~line 1503 reconstructs the fallback logic inline but omits `.or(choice.message.reasoning)`. A payload with only `reasoning` (no `reasoning_content`, no `content`) would be handled correctly by production code but is not verified by this test.

Compare with the updated `test_reasoning_field_alias_accepted` in this PR which correctly uses:
```rust
let content = content.or(reasoning_content.or(reasoning));
```

The `test_reasoning_content_used_for_text_response` test should be updated to match, or a dedicated text-response fixture with `"reasoning"` (not `"reasoning_content"`) should be added to it.

src/llm/nearai_chat.rs
think-in-universe: Fixed in https://github.com/nearai/ironclaw/pull/2493/commits/76d318a267fb0c07be4a2e075b5f21c1da8a4fd1

src/llm/nearai_chat.rs
think-in-universe: Fixed in https://github.com/nearai/ironclaw/pull/2493/commits/76d318a267fb0c07be4a2e075b5f21c1da8a4fd1

src/llm/nearai_chat.rs
```diff
@@ -577,9 +581,16 @@ impl LlmProvider for NearAiChatProvider {
...
-            .message
-            .tool_calls
+        let ChatCompletionResponseMessage {
+            content: message_content,
+            reasoning_content,
+            reasoning,
+            tool_calls: message_tool_calls,
+            ..
+        } = choice.message;
+        let reasoning_fallback = reasoning_content.or(reasoning);
```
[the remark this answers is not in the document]
think-in-universe: We don't want to leak reasoning content into tool call

think-in-universe:
@serrrfirat I think we're ready to merge

serrrfirat (approved):

PR #2498, opened 2026-04-15
fix(web): prevent user messages from vanishing on thread switch (#2409)

### LINKED ISSUE #2409: [QA] User messages disappear after typing in chat
### Environment

Staging (hosted-staging)

### Version/Commit Hash

### QA Test Date

2026-04-13

### Feature Area

Web UI / Chat

### Steps to Reproduce

2. Open a bot chat session
3. Type a message in the chat input field
4. Send the message or observe the input field after typing
5. The typed message vanishes from the input area or chat history

### Expected Behavior

User-typed messages should persist in the input field until sent, and once sent, should appear in the chat history reliably. Messages should never silently disappear.

### Actual Behavior

User messages disappear after typing. The text entered in the chat input field vanishes without being sent or saved. This forces users to retype their messages and creates a frustrating, unreliable experience. Severity: Critical - core chat functionality is broken.

### Logs/Screenshots

Observed during general testing on 2026-04-13. The disappearing messages occurred intermittently but repeatedly, making the chat interface unreliable for basic usage.

serrrfirat:
## Summary
- Fixes user messages disappearing when switching threads or on SSE reconnect (#2409)
- The root cause: `loadHistory()` clears the DOM and re-fetches from DB, but the agent loop hasn't persisted the user message yet (it persists *after* safety checks)
- Previous fix attempt (#2434) was closed because it persisted before the safety pipeline
- This fix is **purely frontend** — tracks optimistically-shown messages in a `_pendingUserMessages` map and re-injects them after `loadHistory()` if they're not in the DB response yet
- Pending messages are cleared when any SSE event (`tool_started`, `stream_chunk`, `response`) confirms the agent processed the message, or after a 60s TTL

## Test plan
- [x] Send message in thread A, switch to thread B, switch back to A — message still visible
- [x] Send message, wait for response, switch away and back — full turn visible (no duplicate)
- [ ] Send message, kill SSE connection (dev tools), reconnect — message still visible
- [x] Refresh page after sending — message appears once agent persists (within seconds)
- [ ] Send message with images attached — same behavior

Closes #2409

### DIFF
--- a/crates/ironclaw_gateway/static/app.js
+++ b/crates/ironclaw_gateway/static/app.js
@@ -98,8 +98,9 @@ let authFlowPending = false;
 // Tracks user messages sent but not yet persisted to DB (#2409).
 // When loadHistory() clears the DOM, pending messages are re-injected
 // so they don't vanish during the safety-pipeline processing window.
-const _pendingUserMessages = new Map(); // threadId -> [{content, timestamp}]
+const _pendingUserMessages = new Map(); // threadId -> [{id, content, timestamp}]
 const PENDING_MSG_TTL_MS = 60000; // discard after 60s
+let _nextPendingId = 0;
 let _ghostSuggestion = '';
 let currentSettingsSubtab = 'inference';
 let generatedImagesByThread = new Map();
@@ -1250,12 +1251,15 @@ function sendMessage() {
   input.focus();
 
   // Track as pending so loadHistory() can re-inject if DB hasn't persisted yet (#2409)
+  let pendingId = null;
+  const pendingThreadId = currentThreadId;
   if (currentThreadId) {
     const displayContent = content || '(images attached)';
     if (!_pendingUserMessages.has(currentThreadId)) {
       _pendingUserMessages.set(currentThreadId, []);
     }
-    _pendingUserMessages.get(currentThreadId).push({ content: displayContent, timestamp: Date.now() });
+    pendingId = _nextPendingId++;
+    _pendingUserMessages.get(currentThreadId).push({ id: pendingId, content: displayContent, timestamp: Date.now() });
   }
 
   const body = { content, thread_id: currentThreadId || undefined, timezone: Intl.DateTimeFormat().resolvedOptions().timeZone };
@@ -1269,6 +1273,18 @@ function sendMessage() {
     method: 'POST',
     body: body,
   }).catch((err) => {
+    // Remove the pending entry so it won't be re-injected on thread switch (#2498)
+    if (pendingId !== null && pendingThreadId) {
+      const arr = _pendingUserMessages.get(pendingThreadId);
+      if (arr) {
+        const filtered = arr.filter(p => p.id !== pendingId);
+        if (filtered.length > 0) {
+          _pendingUserMessages.set(pendingThreadId, filtered);
+        } else {
+          _pendingUserMessages.delete(pendingThreadId);
+        }
+      }
+    }
     // Handle rate limiting (429)
     if (err.status === 429) {
       showToast(I18n.t('chat.rateLimited'), 'error');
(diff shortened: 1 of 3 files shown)


serrrfirat:
Addressed review feedback in 76c62f74:

1. **Removed early `_pendingUserMessages.delete()`** from `response`, `tool_started`, and `stream_chunk` SSE handlers — fixes the race condition where the first server event clears all pending messages for a thread, causing later rapid-fire messages to vanish on thread switch.
2. **Replaced `Set`-based dedup with a frequency map** in `loadHistory()` — correctly handles duplicate-content messages (e.g. user sends "ok" twice in quick succession).
3. **Simplified welcome-card guard** using hoisted `freshPending` variable instead of re-computing the TTL check inline.

henrypark133 (dismissed):
Verified finding:

`crates/ironclaw_gateway/static/app.js:1245` now pushes every optimistic send into `_pendingUserMessages`, but the successful-turn path in the `response` SSE handler (`app.js:850-880`) never removes the entry after the turn is persisted. That means the pending map stays populated until the 60s TTL expires, so any history reload in that window can still re-inject a stale user message or keep the thread in a fake pending state. The new `test_pending_message_cleared_after_response` should fail against this head for the same reason.

Suggested fix: clear or decrement the matching pending entry when the turn completes, rather than relying on the TTL sweep during a later `loadHistory()`.

serrrfirat:
Addressed @henrypark133's review — pending entries are now cleared on turn completion instead of relying solely on TTL:

```js
// Turn complete — remove oldest pending entry for this thread (#2409)
const _pending = _pendingUserMessages.get(data.thread_id);
if (_pending) {
  _pending.shift();
  if (_pending.length === 0) _pendingUserMessages.delete(data.thread_id);
}
```

Added in the `response` SSE handler after `loadThreads()`. Uses `.shift()` (FIFO) so only the oldest pending entry is removed per response — no race condition if the user sends multiple messages rapidly. `test_pending_message_cleared_after_response` should now pass.

crates/ironclaw_gateway/static/app.js
```diff
@@ -1237,6 +1249,15 @@ function sendMessage() {
   autoResizeTextarea(input);
   input.focus();
 
+  // Track as pending so loadHistory() can re-inject if DB hasn't persisted yet (#2409)
+  if (currentThreadId) {
+    const displayContent = content || '(images attached)';
+    if (!_pendingUserMessages.has(currentThreadId)) {
+      _pendingUserMessages.set(currentThreadId, []);
+    }
+    _pendingUserMessages.get(currentThreadId).push({ content: displayContent, timestamp: Date.now() });
```
serrrfirat: **Medium Severity**

The optimistic pending entry is not removed when `/api/chat/send` fails.

`sendMessage()` records the message in `_pendingUserMessages` before the POST has succeeded, but the `.catch()` path only marks the DOM node as failed and never removes the pending entry. If the request is rejected by rate limiting, auth expiry, or a transient network error, switching threads or reconnecting within the 60s TTL will re-inject the message from `_pendingUserMessages` as if it had been accepted by the agent, and the failed-send retry styling is lost.

Please track a local pending id/reference and remove that exact entry in the `apiFetch(...).catch()` path. A regression test should force `/api/chat/send` to fail, reload history, and assert the failed message is not re-injected as a normal sent message.

crates/ironclaw_gateway/static/app.js
```diff
@@ -3110,9 +3131,35 @@ function loadHistory(before) {
...
+      let freshPending = [];
+      if (pending && pending.length > 0) {
+        const now = Date.now();
+        freshPending = pending.filter(p => now - p.timestamp < PENDING_MSG_TTL_MS);
+        if (freshPending.length > 0) {
+          const dbContentsCounts = data.turns
+            .map(t => t.user_input)
+            .filter(Boolean)
+            .reduce((acc, content) => {
+              acc[content] = (acc[content] || 0) + 1;
```
serrrfirat: **Medium Severity**

The dedupe counter uses a plain object keyed by raw user message content.

`dbContentsCounts` is initialized as `{}` and then indexed by `content`, which is user-controlled. Values such as `__proto__`, `constructor`, or `toString` collide with object prototype properties and can make `dbContentsCounts[p.content] > 0` behave incorrectly. In those cases a message that is already present in history can still be re-injected as a duplicate pending message.

Please use `new Map()` or `Object.create(null)` for these counts, and add a regression test with a message such as `__proto__` or `toString`.

tests/e2e/scenarios/test_pending_user_messages.py
```diff
@@ -0,0 +1,190 @@
...
+
+    chat_input = page.locator(SEL["chat_input"])
+    await chat_input.wait_for(state="visible", timeout=5000)
+
+    # Inject a user message into the DOM and the pending map directly,
+    # simulating the state right after sendMessage() but before the agent
+    # loop persists. We avoid using the real send flow because the mock LLM
+    # may respond before we can test the pending window.
+    thread_id = await page.evaluate("() => currentThreadId")
+    await page.evaluate(
```
serrrfirat: **Medium Severity**

This reconnect regression test bypasses the real send path.

The test manually calls `addMessage()` and pushes into `_pendingUserMessages`, so it can still pass if `sendMessage()` stops registering real optimistic sends correctly. It also calls `_wait_for_connected()` after `connectSSE()`, but that helper only checks `sseHasConnectedBefore`; after the initial connection this flag is already true, so the assertion does not prove a reconnect or history reload actually happened.

Please drive the real send path with a controlled delay or mocked `/api/chat/send`/history timing, then force `loadHistory()` or an actual SSE reconnect and wait on an observable reload/reconnect signal before asserting the message remains.

ilblackdragon (approved):

PR #2503, opened 2026-04-15
fix(e2e): resolve 12 E2E test failures across routines and features groups

serrrfirat:
## Summary

Fixes 12 E2E test failures from [staging E2E run #24329804980](https://github.com/nearai/ironclaw/actions/runs/24329804980) across the **routines** (9 failures) and **features** (3 failures) groups.

Three distinct root causes:

- **Read-only thread regression (9 routines failures):** `ed2d6dc3` changed `loadThreads()` to follow the server's `active_thread` on page load. When a prior test creates an HTTP-channel thread, new browser pages auto-switch to it — disabling the chat input with "Read-only thread (external channel)". Fix: skip read-only channel threads when auto-switching; also restore placeholder text in `enableChatInput()`.

- **Tool approval state contamination (2 features failures):** `test_chat_reply_always` permanently auto-approves the `http` tool for the session. Tests that run after it (`test_text_approval_resolves_real_tool_call`, `test_slash_approve_is_thread_scoped_api`) find the tool pre-approved and never see the approval gate fire. Fix: reorder so real-approval-gate tests run before the "always" test.

- **Silent `/approve` on idle thread (1 features failure):** `/approve` is always routed as an approval command (starts with `/`). `process_approval()` returns an empty message when no approval is pending, producing `HandleOutcome::NoResponse`. `send_chat_and_wait_for_terminal_message` times out waiting for a visible response. Fix: return `"No pending approval for this thread."` so the response is visible.

## Test plan

- [ ] Verify `cargo check` passes (confirmed locally)
- [ ] Verify 105 approval-related unit tests pass (`cargo test --lib approval` — confirmed locally)
- [ ] Trigger E2E workflow and verify all 12 previously-failing tests pass
- [ ] Verify no regressions in passing tests (core, extensions groups)

### DIFF
--- a/crates/ironclaw_gateway/static/app.js
+++ b/crates/ironclaw_gateway/static/app.js
@@ -3554,8 +3554,8 @@ function loadThreads() {
         // Skip external-channel threads (e.g. HTTP, Telegram) — they are
         // read-only in the web UI, so auto-switching to one would leave the
         // chat input disabled.  Fall through to the assistant thread instead.
-        var activeThread = threads.find(t => t.id === activeThreadId);
-        if (!activeThread || !isReadOnlyChannel(activeThread.channel)) {
+        const activeThread = threads.find(t => t.id === activeThreadId);
+        if (!isReadOnlyChannel(activeThread.channel)) {
           switchThread(activeThreadId);
           return;
         }
--- a/crates/ironclaw_gateway/static/app.js
+++ b/crates/ironclaw_gateway/static/app.js
@@ -1311,6 +1311,7 @@ function enableChatInput() {
   const btn = document.getElementById('send-btn');
   if (input) {
     input.disabled = false;
+    input.placeholder = I18n.t('chat.inputPlaceholder');
   }
   if (btn) btn.disabled = false;
 }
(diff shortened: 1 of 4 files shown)


henrypark133 (changes requested):
## Review: read-only refresh fix is still unpinned

I verified the approval-path changes against the checked-out code and reran the targeted approval E2Es after prebuilding the branch; those pass locally. One blocker remains before merge.

### Concerning: the read-only `active_thread` refresh regression still has no deterministic browser test
**File:** `crates/ironclaw_gateway/static/app.js:3553`
This branch is the actual routines-side fix: on first load, the UI now skips an external-channel `active_thread` and falls back to the assistant thread instead. But none of the new tests actually puts the browser into that state. `tests/e2e/scenarios/test_tool_approval.py:288` only drives a normal approval flow from a fresh `page`, and the existing refresh coverage in `tests/e2e/scenarios/test_sse_reconnect.py:98` only exercises a gateway thread. That means the suite can still go green without ever covering `active_thread = <read-only external channel>`, so this exact regression can come back with a small `loadThreads()` refactor or a different test order.

**Suggested fix:** add a browser-level regression that explicitly leaves an HTTP/Telegram/external thread as the server `active_thread`, reloads without a hash, and asserts the UI lands on the assistant thread with `#chat-input` enabled.

serrrfirat:
## Addressed review feedback

### @henrypark133 — read-only active_thread regression test ✅
Added `test_refresh_skips_readonly_external_active_thread` in `tests/e2e/scenarios/test_sse_reconnect.py`. The test:
1. Creates a thread and sends a message (making it the server's `active_thread`)
2. Strips the URL hash so reload relies on `active_thread`
3. Intercepts `/api/chat/threads` via Playwright route to mark that thread as `channel: "http"`
4. Reloads the page
5. Asserts `currentThreadId === assistantThreadId` (fell through to assistant)
6. Asserts `#chat-input` is enabled (not disabled by read-only state)

### @gemini-code-assist — style cleanup ✅
- Removed redundant `!activeThread` guard (`threads.some()` already guarantees existence)
- Changed `var` → `const`

Commit: 7ebad0b7

henrypark133 (approved):
## Review: E2E approval and read-only-thread regressions

The current head fixes the stale `/approve` no-response path in the caller, and it now has a deterministic browser-level regression test for the read-only active-thread refresh path instead of relying on a helper-only assertion.

ilblackdragon (approved):

PR #2515, opened 2026-04-16
Unify gateway onboarding, auth gates, and pairing flows

henrypark133:
## Summary

This PR consolidates the web gateway onboarding/auth/pairing flow around one normalized onboarding path and fixes the trust-boundary regressions that fell out of the earlier refactor.

The main goals are:

- unify chat and settings/setup flows for extensions/channels
- stop stale auth/setup UI from reopening after onboarding has advanced
- preserve exact gate/request scoping for auth resolution
- harden pairing rollback so failed propagation does not leave the runtime in a claimed state
- keep a narrow legacy web compatibility path while `ENGINE_V2` is still off by default

This PR supersedes #2432.

## Problem

Before this change, onboarding/auth behavior was split across several overlapping paths:

- legacy web auth submit/cancel endpoints
- engine v2 pending gates
- pending-gate history rehydration
- setup/configure modal behavior
- pairing as a separate post-auth path
- browser-injected/synthetic messages that did not always preserve metadata

That caused multiple user-facing regressions:

- Telegram chat setup could route through `telegram_bot_token` instead of `telegram`
- auth/setup UI could reopen after pairing or after setup already succeeded
- closing stale auth/setup UI could produce incorrect cancel behavior
- `tool_install` could report “installed” while the UI had already auto-activated the extension
- duplicate Telegram pairing codes could be emitted due to overlapping pollers
- auth credentials could be routed through normal inbound chat handling instead of direct gate resolution
- failed pairing propagation could leave the live runtime bound to the new owner even when the DB rollback reported failure
- v1 web message/status paths could lose user scoping metadata and broadcast globally

### Unified onboarding flow
- Introduced a normalized `onboarding_state` path for extension/channel onboarding in the web gateway.
- Consolidated frontend onboarding rendering around one controller instead of parallel auth/pairing/setup card logic.
- Made chat-opened configure use the same `/api/extensions/{name}/setup` flow as Settings.
- Added normalized `extension_name` handling so UI routing uses extension identity while backend secret persistence still uses the real credential key.

### Auth and gate handling
- Migrated gate-backed auth resolution to a structured submission path instead of replaying credentials as raw chat text.
- Added `Submission::GateAuthResolution` and exact `request_id`-scoped bridge handling.
- Restored `AppEvent::GateRequired` emission where needed so `/v1/responses` does not hang on auth-gated requests.
- Preserved a minimal legacy compatibility path for prompts without `request_id` while `ENGINE_V2` is still disabled by default.
- Added documentation and inline comments marking that compatibility seam for later removal.

### Setup and post-auth transitions
- Chat setup submissions can now carry gate identity (`request_id`, `thread_id`) so the backend can advance/discard the exact pending auth gate.
- OAuth continuation from setup is treated as `auth_required`, not `failed`.
- OAuth-only auth cards no longer render a meaningless token textbox.
- Telegram/manual-token flows route back through the extension setup loader so chat uses the same configure form as Settings.

### Pairing and Telegram runtime fixes
- Extracted pairing approval logic into a clearer backend path.
- Pairing propagation now snapshots runtime owner/config and restores them if post-pairing restart fails.
- Numeric owner ID caching happens only after propagation succeeds.
- Polling restarts are serialized so Telegram does not run overlapping pollers and emit duplicate pairing codes.
- Pairing completion no longer injects the old stale follow-up that could reopen setup.
- Added a bounded “setup complete” handoff back into chat after successful pairing-ready completion.

### Tool install / activation behavior
- `tool_install` now follows the real readiness path instead of returning a raw install result while hidden post-install logic separately activates/auth-gates the extension.
- This keeps LLM-visible tool state aligned with UI-visible readiness state.

### Web/gateway boundary cleanup
- Added shared web `IncomingMessage` construction helpers so browser-originated and browser-injected messages consistently carry `user_id` and `thread_id` metadata.
- This fixes the `Status event missing user_id in metadata; broadcasting globally` regression.
- Added a legacy-path guard so v2-only structured submissions (`GateAuthResolution`, `ExternalCallback`) fail before mutating session thread state when `ENGINE_V2=false`.

### Docs
- Updated `CLAUDE.md` and `src/channels/web/CLAUDE.md` with the current onboarding/auth invariants and the temporary v1 compatibility boundary.
- This explicitly documents what should be removed once legacy web auth mode is retired.

## Risk / compatibility notes

- `ENGINE_V2` is still off by default, so this PR intentionally preserves a narrow legacy compatibility path for web auth prompts that do not carry `request_id`.
- The compatibility path is documented as temporary and should be deleted when v1 auth mode is removed.
- This PR does not remove CLI extension setup flows; CLI paths were not intentionally refactored here.
- The largest remaining architectural debt after this PR is that backend onboarding transition knowledge still lives across `server.rs`, `router.rs`, and extension/pairing code. This branch improves behavior substantially, but a future backend onboarding coordinator would still be a worthwhile cleanup.

## Testing

Targeted Rust checks run during this series included:

- `cargo fmt --all`

Auth / onboarding / gate tests:
- `cargo test test_parser_json_gate_auth_resolution --lib`
- `cargo test test_chat_gate_resolve_handler_credential_submission_uses_structured_gate_resolution --lib`
- `cargo test test_chat_auth_token_handler_preserves_user_scoped_metadata --lib`
- `cargo test test_chat_auth_cancel_handler_clears_requested_thread_auth_mode --lib`
- `cargo test classify_configure_result_treats_oauth_continuation_as_auth_required --lib`
- `cargo test event_from_configure_result_emits_auth_required_for_oauth_continuation --lib`
- `cargo test insert_and_notify_pending_gate_sends_status_no_text --lib`
- `cargo test accumulator_gate_required_marks_failed --lib`
- `cargo test discard_engine_pending_auth_request_discards_only_matching_auth_gate --lib`
- `cargo test discard_engine_pending_auth_request_matches_scope_thread_id --lib`
- `cargo test v2_only_structured_submissions_do_not_switch_threads_when_engine_v2_disabled --lib`

Pairing / runtime rollback tests:
- `cargo test propagate_approval_restores_runtime_state_when_on_start_fails --lib`
- `cargo test test_upsert_and_approve_pairing --lib --no-default-features --features libsql`
- `cargo test test_pairing_approve_claims_code_for_authenticated_user --lib --features libsql`
- `cargo test test_pairing_approve_does_not_inject_followup_agent_turn --lib --features libsql`
- `cargo test test_pairing_approve_does_not_inject_followup_agent_turn_without_thread --lib --features libsql`
- `cargo test test_pairing_approve_injects_ready_followup_for_active_thread --lib --features libsql`

Web / setup / responses / metadata tests:
- `cargo test test_extension_setup_request_deserialize_with_fields --lib`
- `cargo test test_extensions_setup_submit_returns_failure_when_not_activated --lib`
- `cargo test test_chat_approval_handler_preserves_user_scoped_metadata --lib`
- `cargo test test_handle_client_approval_approve --lib`
- `cargo test test_handle_client_message_sends_to_agent --lib`

E2E tests run during the series:
- `tests/e2e/.venv/bin/pytest -q tests/e2e/scenarios/test_channel_pairing_flow.py`
- `tests/e2e/.venv/bin/pytest -q tests/e2e/scenarios/test_telegram_hot_activation.py`
- `tests/e2e/.venv/bin/pytest -q tests/e2e/scenarios/test_v2_kernel_auth_gateway_flow.py`
- `tests/e2e/.venv/bin/pytest -q tests/e2e/scenarios/test_auth_no_duplicate_response.py`
- `tests/e2e/.venv/bin/pytest -q tests/e2e/scenarios/test_v2_auth_oauth_matrix.py -k test_wasm_tool_first_chat_auth_attempt_emits_auth_url`

## Follow-up work

Not part of this PR, but still worth doing:
- centralize backend onboarding transitions behind one coordinator
- fully remove the legacy web auth compatibility path when v1 auth mode is retired
- continue shrinking v1/v2 divergence at the web boundary


src/bridge/router.rs
```diff
@@ -1835,37 +2016,56 @@ pub async fn resolve_gate(
...
+                                        thread_id: pending
+                                            .scope_thread_id
+                                            .clone()
+                                            .or_else(|| Some(pending.thread_id.to_string())),
                                     },
-                                    &message.metadata,
-                                )
+                                );
+                            }
+                            return Ok(BridgeOutcome::Respond(result.message));
```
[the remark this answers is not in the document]
henrypark133: Fixed in dca3969f. `ConfigureFlowOutcome::PairingRequired` now returns `BridgeOutcome::Pending` after broadcasting the onboarding event, so the pairing card is the sole user-facing signal and the bridge does not emit a redundant text response.

crates/ironclaw_gateway/static/app.js
```diff
@@ -2585,11 +2607,12 @@ function handleGateRequired(data) {
...
-      extension_name: resume.credential_name,
+    handleOnboardingState({
+      state: 'auth_required',
+      extension_name: data.extension_name || resume.credential_name,
+      request_id: data.request_id,
       instructions: resume.instructions,
       auth_url: resume.auth_url || null,
-      request_id: data.request_id,
       thread_id: data.thread_id || currentThreadId,
     });
```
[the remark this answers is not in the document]
henrypark133: Fixed in dca3969f. `handleGateRequired` now preserves `extension_name` as null for non-extension auth prompts and passes a separate `display_name` for UI text. That keeps generic credential prompts on `showAuthCard()` and avoids the failing `/api/extensions/<credential>/setup` fetch path.

zmanian (commented):
## Code Review

**Size:** XL (+4518 / -2945) | **Risk:** Medium

### Summary

Solid architectural consolidation that eliminates real bugs (stale auth cards, duplicate pollers, orphaned runtime state after failed pairing). The type-level improvements (`BridgeOutcome`, `TurnOutcome`, `GateAuthResolution`) are the kind of refactoring that prevents future regressions. Test coverage is thorough (~30 new/updated tests).

### Issues

**1. Hardcoded "Telegram" string in unified onboarding handler** (`crates/ironclaw_gateway/static/app.js`)

Inside `handleOnboardingState`, the `ready` branch has:
```javascript
if (data.state === 'ready' && shouldShowChannelConnectedMessage(data.extension_name, true)) {
    addMessage('system', 'Telegram is now connected. You can message me there and I can send you notifications.');
}
```
This is now the unified handler for *all* extensions. If `shouldShowChannelConnectedMessage` returns true for a non-Telegram channel, users will see "Telegram is now connected" regardless. Should use `data.extension_name` or `data.display_name`.

**2. Legacy auth-token handler sends token as message content** (`src/channels/web/server.rs`, `chat_auth_token_handler`)

The v1 shim now pipes the raw token through the message channel:
```rust
let msg = web_incoming_message("gateway", &user.user_id, token, req.thread_id.as_deref());
```
The old path submitted directly to `auth_manager.submit_auth_token()`, bypassing chat history. If any code path persists `IncomingMessage.content` as a user message, the token ends up in plaintext in the DB. Documented as temporary, but worth an inline comment noting the security implication.

**3. `dispatch_onboarding_ready_followup` prompt injection surface** (`src/channels/web/server.rs` ~line 2825)

```rust
let content = format!(
    "System event: onboarding for '{extension_name}' is now fully complete and ready. ..."
);
```
The `extension_name` here comes from the URL path parameter (lowercased), not from `sanitize_extension_name()`. For defense-in-depth, consider using `sanitize_extension_name(&channel)` here, consistent with the sanitization done elsewhere in this PR.

**4. `request_id` rotation on auth retry** (`src/bridge/router.rs`, `requeue_auth_pending_gate`)

When an auth attempt fails and re-queues, a new `request_id` is generated. The frontend's auth card references the old `request_id` via `data-request-id`. The new `onboarding_state` SSE event carries the new ID. If the frontend doesn't update the card's `data-request-id` when it receives the re-queued event, a subsequent submit from the stale card will fail to match. I believe `handleOnboardingState` -> `handleAuthRequired` recreates the card (fixing this), but please confirm the full roundtrip.

**5. `verification` field removed from `ActionResponse`** (`src/channels/web/types.rs`)

The `verification` field is dropped from the wire format. If any external consumer (mobile app, API client) depends on this field, this is a breaking change. The PR description doesn't mention this removal -- worth calling out explicitly.

**6. Missing `display_name` passthrough for `setup_required`** (`crates/ironclaw_gateway/static/app.js`)

In `handleOnboardingState`, the `auth_required` branch passes `display_name` to `handleAuthRequired`, but the `setup_required` branch calls `showSetupCardForExtension` without it. Minor inconsistency.

### Positive notes

- `BridgeOutcome` with `#[must_use]` eliminates the ambiguous `Option<String>` return where `None` could mean "gate paused" or "no response." Great type-level improvement.
- `TurnOutcome::CompletedSilently` correctly models auth-card turns without persisting a redundant text response.
- `sanitize_extension_name()` with injection/unicode/truncation tests is good security hygiene.
- Pairing rollback (`revert_pairing_approval`) correctly restores both DB state and runtime state, including previous `owner_id`.
- Polling serialization (`stop_polling` awaits old task) prevents the duplicate pairing code issue.
- CLAUDE.md updates clearly document the v1 compat boundary and what to delete later.

src/channels/web/server.rs
```diff
@@ -2756,7 +2825,12 @@ async fn dispatch_engine_auth_resolution(
             .clone()
     };
 
-    let msg = IncomingMessage::new("gateway", user_id, content).with_thread(thread_id.to_string());
+    let content = format!(
+        "System event: onboarding for '{extension_name}' is now fully complete and ready. \
```
serrrfirat, line 3025: **Critical Severity** — Prompt injection via unsanitized channel name

`extension_name` is interpolated directly into the agent prompt string (`"System event: onboarding for '{extension_name}' is now fully complete..."`). The `pairing_approve_handler` lowercases the channel name from the URL path but never calls `sanitize_extension_name()`.

A crafted pairing path like `/api/pairing/evil%20Ignore%20all%20instructions/approve` injects arbitrary text into the agent prompt. The `sanitize_extension_name()` helper exists in the same file but is not called on this input.

**Fix:** Call `sanitize_extension_name()` (or at minimum validate against a known-extension allowlist) on the `channel` parameter before passing it to `dispatch_onboarding_ready_followup`.

crates/ironclaw_common/src/event.rs
```diff
@@ -110,51 +120,34 @@ pub enum AppEvent {
         /// Whether the "always" auto-approve option should be shown.
         allow_always: bool,
     },
-    #[serde(rename = "auth_required")]
-    AuthRequired {
+    #[serde(rename = "onboarding_state")]
```
serrrfirat, line 123: **High Severity** — Breaking SSE wire contract with no backward compatibility

The SSE event types changed from `"auth_required"`, `"auth_completed"`, `"pairing_required"`, `"pairing_completed"` to a single `"onboarding_state"`. This is a breaking change for any non-bundled SSE consumer (mobile apps, browser extensions, third-party integrations). The PR only updates the bundled `app.js` client.

Any WebSocket client that handled `auth_token` or `auth_cancel` message types will silently drop those messages since `WsClientMessage` no longer has those variants.

**Suggestion:** Either add a version negotiation mechanism or document this as a breaking API change. Consider keeping the old event types as aliases for at least one release cycle.

src/channels/web/server.rs
```diff
@@ -2477,174 +2498,123 @@ async fn chat_gate_resolve_handler(
...
-                        success: false,
-                        message: result.message,
-                        thread_id: req.thread_id.clone(),
-                    },
-                );
-            }
+    // Temporary web compatibility shim for engine v1 `pending_auth`.
+    // Gate-backed auth must go through `/api/chat/gate/resolve`; only prompts
+    // without a `request_id` should hit this endpoint.
+    let msg = web_incoming_message("gateway", &user.user_id, token, req.thread_id.as_deref());
```
serrrfirat: **High Severity** — Legacy auth token handler forwards raw secret as chat message content

Two related issues:

1. **Secret leakage on v1 path:** The legacy `chat_auth_token_handler` now forwards the raw auth token as the `content` of an `IncomingMessage` via `web_incoming_message()`. With `ENGINE_V2=false` (the default), this token hits the v1 `process_user_input` path, which may echo it to chat history, SSE, or other channels. Previously the handler bypassed the message pipeline entirely and called `submit_auth_token` directly.

2. **Missing routing target:** `extension_name` was removed from `AuthTokenRequest` (in `types.rs`), so there is no way to route the token to the correct credential store on the v1 path. If no pending auth exists on the thread, the raw token text is sent to the LLM as a normal user message.

**Fix:** Either preserve the direct `submit_auth_token` call on the v1 path, or ensure the message pipeline recognizes auth tokens and prevents them from reaching chat history / the LLM.

src/bridge/router.rs
```diff
@@ -1568,7 +1662,84 @@ pub async fn handle_external_callback(
         request_id = %request_id,
         "engine v2: no matching pending auth gate for external callback"
     );
-    Ok(Some(
+    Ok(BridgeOutcome::Respond(
+        "No matching pending authentication gate found.".into(),
+    ))
+}
+
+pub async fn handle_auth_gate_resolution(
```
serrrfirat, line 1658: **High Severity** — TOCTOU race between gate lookup and resolution

`handle_auth_gate_resolution` looks up the gate via `pending_gates.peek()` and `list_for_user()`, checking `gate.request_id == request_id`. Between the lookup and the subsequent `resolve_gate()` call, the guard is dropped. A concurrent request for the same `request_id` could resolve the gate first, causing the second request to double-resolve or operate on a stale gate.

Additionally, there's a type inconsistency: in the `peek` path, the comparison appears to be `gate.request_id == request_id.to_string()` (Uuid vs String), while in the `list_for_user` path it's `gate.request_id == request_id` (Uuid vs Uuid). One path may silently fail to match.

**Fix:** Hold the lock across the lookup-and-resolve sequence, or use an atomic take-if-matches operation.

crates/ironclaw_gateway/static/app.js
```diff
@@ -2525,48 +2511,85 @@ async function handleAuthRequired(data) {
...
+      showToast(data.message, data.state === 'ready' ? 'success' : 'error');
+    }
+    _recentLocalPairingApprovals.delete(data.extension_name);
+    removePairingCard(data.extension_name);
+    removeAuthCard(data.extension_name);
+    removeSetupCard(data.extension_name);
+    closeConfigureModal(data.extension_name);
+    setAuthFlowPending(false);
+    if (data.state === 'ready' && shouldShowChannelConnectedMessage(data.extension_name, true)) {
+      addMessage('system', 'Telegram is now connected. You can message me there and I can send you notifications.');
```
serrrfirat: **Medium Severity** — Hardcoded "Telegram" message shown for all extensions

`handleOnboardingState` hardcodes `'Telegram is now connected. You can message me there and I can send you notifications.'` for all extensions when `state === 'ready'` and `shouldShowChannelConnectedMessage` returns true. If any other channel (e.g., Discord, Slack) reaches the `ready` state through this path, users see an incorrect "Telegram" message.

**Fix:** Use the `extension_name` (or `display_name`) from the event payload dynamically:
```js
addMessage('system', `${data.display_name || data.extension_name} is now connected.`);
```

src/bridge/router.rs
```diff
@@ -345,12 +383,61 @@ async fn insert_and_notify_pending_gate(
...
         state.sse.clone(),
         state.effect_adapter.tools(),
+        state.auth_manager.as_deref(),
         message,
         &pending,
     )
     .await
 }
 
+async fn requeue_auth_pending_gate(
```
serrrfirat, line 426: **Medium Severity** — Unbounded auth gate requeuing enables DoS

When auth fails, `requeue_auth_pending_gate` inserts a new `PendingGate` with a fresh `request_id` and 30-minute expiry. There is no limit on how many times a gate can be requeued. A malicious user could submit invalid tokens in a loop, creating unbounded pending gates in memory.

**Fix:** Add a per-user or per-thread requeue counter and cap it (e.g., max 5 retries per auth flow). Alternatively, reuse the same gate slot instead of creating a new one.

src/db/postgres.rs
```diff
@@ -1165,6 +1165,35 @@ impl ChannelPairingStore for PgBackend {
         Ok(rows.into_iter().map(|row| row.get(0)).collect())
     }
 
+    async fn resolve_channel_external_id_for_owner(
```
serrrfirat, line 1168: **Medium Severity** — `resolve_channel_external_id_for_owner` join may return no rows in single-tenant mode

The query joins `users u ON u.id = ci.owner_id` and checks `u.status = 'active'`. In single-tenant mode where the owner is the default user without a corresponding row in the `users` table, the INNER JOIN returns no rows even when a valid channel identity exists. This could silently prevent owner recovery at startup.

**Fix:** Use a LEFT JOIN on `users` and make the `u.status = 'active'` check optional, or ensure the default user always has a `users` table entry.

src/channels/web/server.rs
```diff
@@ -2477,174 +2498,123 @@ async fn chat_gate_resolve_handler(
...
-        return Ok(Json(ActionResponse::ok("Auth cancelled")));
-    }
-
-    clear_auth_mode(&state, &user.user_id).await;
-    // Also clear engine v2 pending auth so the next message isn't consumed as a token.
-    crate::bridge::clear_engine_pending_auth(&user.user_id, req.thread_id.as_deref()).await;
-    Ok(Json(ActionResponse::ok("Auth cancelled")))
+    // Temporary web compatibility shim for engine v1 `pending_auth`.
+    // Delete alongside the legacy auth-mode browser flow.
+    clear_auth_mode_for_thread(&state, &user.user_id, req.thread_id.as_deref()).await?;
```
serrrfirat: **Medium Severity** — Cancel handler no longer clears engine v2 pending auth gate

`chat_auth_cancel_handler` now calls `clear_auth_mode_for_thread` but no longer calls `crate::bridge::clear_engine_pending_auth`. If there is a pending engine v2 auth gate and the user hits this legacy cancel endpoint (which can happen if the frontend falls back to the legacy path while `ENGINE_V2=true`), the engine gate will remain pending and block the thread indefinitely.

**Fix:** Add `clear_engine_pending_auth` back into `clear_auth_mode_for_thread`, or call it explicitly in the cancel handler.

src/extensions/manager.rs
```diff
@@ -201,110 +168,96 @@ fn telegram_response_parse_error(action: &'static str, error: &reqwest::Error) -
...
-    let mut config_updates = HashMap::new();
-
-    if let Some(tunnel_url) = tunnel_url {
-        config_updates.insert(
-            "tunnel_url".to_string(),
-            serde_json::Value::String(tunnel_url.to_string()),
-        );
+/// Validate a Telegram bot token by calling the getMe API.
+/// Returns the bot username on success, or an error if the token is invalid.
+async fn validate_telegram_token(bot_token: &str) -> Result<Option<String>, ExtensionError> {
```
serrrfirat, line 215: **Medium Severity** — User-supplied bot_token interpolated into URL path

`validate_telegram_token` constructs a URL via `telegram_bot_api_url(bot_token, "getMe")` where `bot_token` is user-supplied input directly interpolated into the URL path. While reqwest should percent-encode most dangerous characters, edge cases in URL parsing with crafted tokens (e.g., containing `#`, `?`, or path traversal sequences) should be validated.

**Fix:** Validate that `bot_token` matches the expected Telegram bot token format (`\d+:[A-Za-z0-9_-]+`) before constructing the URL.

src/channels/web/server.rs
henrypark133: Fixed. `pairing_approve_handler` now sanitizes the path-derived channel name before it is reused for the onboarding-ready follow-up, so that value can no longer inject arbitrary prompt text.

crates/ironclaw_common/src/event.rs
henrypark133: Addressed as a documented compatibility boundary rather than by reintroducing the old SSE variants.

This branch now:
- explicitly documents `onboarding_state` as the replacement for the old auth/pairing SSE events in `src/channels/web/CLAUDE.md`
- records the breaking-contract note in `CHANGELOG.md`
- restores legacy WebSocket client aliases for `auth_token` / `auth_cancel` so non-bundled WS clients do not silently drop those messages during the temporary v1-auth compatibility window

I did not add legacy SSE event aliases in this PR; that would be a broader wire-contract expansion rather than a narrow onboarding fix.

src/channels/web/server.rs
henrypark133: Fixed. The legacy `/api/chat/auth-token` path no longer forwards the raw token through `web_incoming_message()` at all.

It now resolves the thread’s existing `pending_auth` state and submits the token directly to the auth/configure path, so the secret bypasses normal chat ingress, history, and LLM routing on the v1 compatibility path as well.

src/bridge/router.rs
henrypark133: Fixed. I removed the advisory `peek()/list_for_user()` + later resolve pattern from the gate-resolution entrypoints.

`handle_auth_gate_resolution` now uses the hinted thread only as a locator and then immediately delegates to `resolve_gate(...)`, which already resolves through the atomic `PendingGateStore::take_verified(...)` path. I applied the same cleanup to the sibling approval / external-callback paths so they all use the same atomic resolution flow.

crates/ironclaw_gateway/static/app.js
henrypark133: Fixed. The ready-state chat message now uses `data.display_name || data.extension_name` instead of hardcoding Telegram, so other channels do not show the wrong connected message.

src/bridge/router.rs
henrypark133: I did not add a retry counter here because this path is already bounded by the unified pending-gate model.

`requeue_auth_pending_gate(...)` runs only after the previous gate for that `(user, thread)` has already been removed by `resolve_gate()`, and `PendingGateStore::insert()` still enforces at most one live gate per `(user_id, thread_id)`. So invalid-token retries replace the paused thread’s current auth gate rather than accumulating unbounded gates in memory.

src/db/postgres.rs
henrypark133: Fixed. `resolve_channel_external_id_for_owner` now uses a `LEFT JOIN` and treats a missing `users` row as acceptable, while still requiring `active` status when the row exists. That preserves owner recovery for the single-tenant default-user case without weakening the active-user check for normal multi-user rows.

src/channels/web/server.rs
henrypark133: Fixed. `clear_auth_mode_for_thread(...)` now also calls `crate::bridge::clear_engine_pending_auth(...)`, so the legacy cancel shim clears the engine-v2 pending auth gate as well instead of leaving the thread blocked.

src/extensions/manager.rs
henrypark133: Fixed. `validate_telegram_token(...)` now validates the token shape before constructing the Telegram API URL, using the expected `digits:tokenchars` format and rejecting malformed input up front.

henrypark133:
Follow-up on the two pairing revert review threads:

- Postgres `revert_pairing_approval` now checks the `UPDATE pairing_requests ...` affected-row count and returns `DatabaseError::NotFound` when it is `0`, before touching `channel_identities`.
- The libsql implementation now mirrors that behavior: it checks the affected-row count first and returns `DatabaseError::NotFound` when no approval row was reverted, so `channel_identities` is left untouched in that case.

I also added a targeted libsql regression test for the revert guard.

henrypark133:
Follow-up on the points in review `#4118793983`:

1. The hardcoded Telegram ready message is fixed; the unified onboarding handler now uses `data.display_name || data.extension_name` instead.
2. The legacy `/api/chat/auth-token` shim is fixed; it no longer sends the raw token through `web_incoming_message()` and instead submits directly against the thread’s existing pending auth state.
3. The onboarding-ready follow-up now sanitizes the path-derived channel name before reusing it in prompt text.
4. On auth retry, the frontend does refresh the auth/setup UI with the new `request_id`: the re-queued gate emits a fresh `onboarding_state`, `handleOnboardingState(...)` routes back through `handleAuthRequired(...)`, and `showAuthCard()` replaces the existing overlay/card before writing the new `data-request-id`. I also tightened `handleGateRequired(...)` so non-extension auth gates preserve `extension_name = null` and use `display_name` separately instead of fabricating an extension identity.
5. I added an explicit `CHANGELOG.md` note that the legacy `ActionResponse.verification` field was removed from the web wire format; clients should use `onboarding_state` / `onboarding` instead.
6. I also fixed the minor UI inconsistency in `setup_required`: `display_name` is now passed through to the setup card path just like the auth-required path.

The legacy-compatibility code paths touched here also have explicit inline/doc notes marking them for removal once the temporary web v1 auth boundary is retired.

src/channels/web/server.rs
```diff
@@ -2708,42 +2828,141 @@ struct HistoryQuery {
...
+        tx_guard
+            .as_ref()
+            .ok_or((
+                StatusCode::SERVICE_UNAVAILABLE,
+                "Channel not started".to_string(),
+            ))?
+            .clone()
+    };
+
+    let content = serde_json::to_string(&submission).map_err(|e| {
```
serrrfirat: **High Severity — Security / Data Leak** [Likely]

`GateAuthResolution::CredentialProvided { token }` is serialized as JSON into `IncomingMessage.content` and traverses the full message pipeline as plaintext. The current code path returns early in `SubmissionParser::parse()` before `persist_user_message`, but this invariant is fragile.

If a future change to the parser fails to recognize the JSON and falls through to the `UserInput` path, the secret token gets persisted as a plain chat message in the database.

**Suggested fix:** Pass the `AuthGateResolution` as a sideband field on `IncomingMessage` (similar to `attachments`) rather than serializing secrets into `.content`. This makes it structurally impossible for the token to leak into persistence.

crates/ironclaw_gateway/static/app.js
```diff
@@ -5066,10 +5074,16 @@ function cancelAuthFromConfigureModal(overlay) {
   var extName = overlay.getAttribute('data-auth-extension') || overlay.getAttribute('data-extension-name');
   var requestId = overlay.getAttribute('data-request-id');
   var threadId = overlay.getAttribute('data-thread-id');
-  var request = requestId
-    ? apiFetch('/api/chat/gate/resolve', { method: 'POST', body: { request_id: requestId, thread_id: threadId || currentThreadId || undefined, resolution: 'cancelled' } })
-    : apiFetch('/api/chat/auth-cancel', { method: 'POST', body: { extension_name: extName, thread_id: threadId || currentThreadId || undefined } });
-  request.catch(function() {});
+  if (requestId) {
```
serrrfirat, line 5061: **High Severity — Correctness** [Certain]

When no `requestId` exists (legacy v1 path), the new code sends **no cancel request at all**. Previously it sent to `/api/chat/auth-cancel`. This means on the legacy path (ENGINE_V2=false), the thread's `pending_auth` state won't be cleared — it persists until expiry.

This is a behavioral regression for the legacy compatibility window. Users on the v1 path who open and dismiss the auth modal will have stale `pending_auth` state.

**Suggested fix:** Add a fallback that sends to `/api/chat/auth-cancel` when `requestId` is absent, preserving the legacy cancel behavior.

src/channels/web/server.rs
```diff
@@ -3777,18 +4043,89 @@ async fn pairing_approve_handler(
...
+                );
+                true
+            }
+        }
+    } else {
+        false
+    };
+
+    if propagation_failed {
+        if let Err(error) = store.revert_approval(&approval).await {
```
serrrfirat, line 4149: **Medium Severity — Correctness / Race Condition** [Likely]

`revert_approval` relies on `previous_owner_id` captured from a `SELECT ... LIMIT 1` in the approval flow. The PostgreSQL backend correctly uses `FOR UPDATE` to prevent concurrent modification, but the libSQL backend does not (and cannot) use row-level locking.

Under concurrent approvals for the same channel+external_id on libSQL, the captured `previous_owner_id` may be stale, causing the revert to restore the wrong owner — leaving the runtime in an inconsistent state.

**Suggested fix:** For libSQL, consider using an optimistic concurrency check (compare-and-swap on a version column) or serialize approval operations through a mutex for the same channel+external_id.

src/channels/web/server.rs
```diff
@@ -2756,7 +2975,12 @@ async fn dispatch_engine_auth_resolution(
             .clone()
     };
 
-    let msg = IncomingMessage::new("gateway", user_id, content).with_thread(thread_id.to_string());
+    let content = format!(
+        "System event: onboarding for '{extension_name}' is now fully complete and ready. \
```
serrrfirat, line 3025: **Medium Severity — Security / Prompt Injection** [Likely]

`extension_name` is interpolated directly into a string that becomes an LLM user message: `"System event: onboarding for '{extension_name}' is now fully complete..."`.

While `sanitize_extension_name` is called in `pairing_approve_handler` (the current caller), the function itself doesn't validate its input. If called from a different code path without prior sanitization, this is a prompt injection vector.

**Suggested fix:** Apply `sanitize_extension_name` inside `dispatch_onboarding_ready_followup` as well (defense-in-depth). The cost is negligible and it removes the implicit contract that all callers must sanitize.

src/channels/web/server.rs
```diff
@@ -2477,174 +2498,273 @@ async fn chat_gate_resolve_handler(
...
+        let thread = sess
+            .threads
+            .get_mut(&target_thread_id)
+            .ok_or((StatusCode::NOT_FOUND, "Thread not found".to_string()))?;
+        let pending_auth = thread.pending_auth.clone().ok_or((
+            StatusCode::BAD_REQUEST,
+            "No pending authentication request for this thread".to_string(),
+        ))?;
+
+        if pending_auth.is_expired() {
```
serrrfirat, line 2632: **Medium Severity — Edge Case** [Likely]

When `pending_auth.is_expired()` returns true, the handler returns `ActionResponse::fail(...)` but does NOT broadcast an `OnboardingState::Failed` SSE event. The frontend auth card remains visible with no user feedback beyond the HTTP response body.

Since the auth card references expired state, the user may be stuck — the card won't auto-dismiss and resubmitting will keep failing.

**Suggested fix:** Broadcast an `OnboardingState::Failed` event (or `OnboardingState::Idle`) when the auth is expired, so the frontend can dismiss the stale card.

src/bridge/router.rs
```diff
@@ -345,12 +383,66 @@ async fn insert_and_notify_pending_gate(
...
+        other => {
+            return Err(engine_err(
+                "resolution mismatch",
+                format!("expected authentication gate, got {}", other.kind_name()),
+            ));
+        }
+    };
+
+    let next_pending = PendingGate {
+        request_id: uuid::Uuid::new_v4(),
```
serrrfirat, line 452: **Medium Severity — Correctness / Rate** [Possible]

On token validation failure, a new gate is created with a fresh `request_id` and 30-minute expiry. Each failed retry creates a new gate and replaces the old one. While `PendingGateStore` is in-memory (not persisted), rapid retries cause unbounded gate churn with SSE broadcast on each creation.

There is no rate limit on gate creation — an attacker (or a confused UI loop) could generate high event throughput.

**Suggested fix:** Consider rate-limiting gate re-creation (e.g., reuse the existing gate if it's still valid and the credential_name matches) or add a cooldown between re-queues.

src/channels/web/server.rs
henrypark133: Fixed in `a8916617`.

I moved structured engine submissions off the generic plaintext path by adding an `IncomingMessage.structured_submission` sideband and teaching the agent loop to consume that before parsing `content`. `dispatch_engine_submission(...)` now sends a placeholder string in `content` and attaches the real `Submission` separately, so `GateAuthResolution::CredentialProvided { token }` no longer traverses chat history / persistence-prone text fields at all.

Coverage: `cargo test test_chat_gate_resolve_handler_credential_submission_uses_structured_gate_resolution --lib`.

crates/ironclaw_gateway/static/app.js
henrypark133: Fixed in `a8916617`.

I restored the legacy no-`requestId` cancel behavior, but I did it by consolidating both auth-card and configure-modal cancel flows onto a shared `requestAuthCancellation(...)` helper in `app.js`. That helper routes gate-backed prompts to `/api/chat/gate/resolve` and legacy v1 prompts to `/api/chat/auth-cancel`, so the compatibility branch is centralized instead of duplicated.

src/channels/web/server.rs
henrypark133: Fixed in `a8916617`.

`dispatch_onboarding_ready_followup(...)` now sanitizes `extension_name` at the helper boundary as well. The current caller was already sanitizing, but this removes the implicit caller contract and keeps future follow-up callers on the same safe path for negligible cost.

src/channels/web/server.rs
henrypark133: Fixed in `a8916617`.

When legacy thread-level `pending_auth` is expired, the handler now clears it and broadcasts `OnboardingState::Failed` with the thread id before returning `ActionResponse::fail(...)`. That keeps the frontend on the normal onboarding SSE contract instead of leaving a stale auth card visible.

Coverage: `cargo test test_chat_auth_token_handler_expired_auth_broadcasts_failed_onboarding_state --lib`.

crates/ironclaw_gateway/static/app.js
```diff
@@ -2525,48 +2511,86 @@ async function handleAuthRequired(data) {
...
-        // Fall through to the legacy card when pending-gate hydration fails.
-      }
-    }
-  }
   setAuthFlowPending(true, data.instructions);
-  if (data.auth_url) {
-    // Token paste flow (with optional OAuth button): show the global auth
-    // prompt card. This handles both OAuth credentials (auth_url present)
-    // and skill-based credentials (instructions present, no auth_url).
+  if (data.auth_url || !data.extension_name) {
```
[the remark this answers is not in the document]
henrypark133: Fixed in `ca4087ab`.

`handleAuthRequired(...)` now treats missing `request_id` as the legacy v1 auth path and goes straight to `showAuthCard(...)` instead of routing through `showSetupCardForExtension(...)`. That keeps the no-`request_id` compatibility seam aligned with the existing `/api/chat/auth-token` / `/api/chat/auth-cancel` flow, while leaving setup-modal routing only for gate-backed prompts.

src/bridge/router.rs
```diff
@@ -1835,37 +1992,56 @@ pub async fn resolve_gate(
...
+                                        onboarding,
+                                        thread_id: pending
+                                            .scope_thread_id
+                                            .clone()
+                                            .or_else(|| Some(pending.thread_id.to_string())),
+                                        },
+                                );
+                            }
+                            return Ok(BridgeOutcome::Pending);
+                            }
```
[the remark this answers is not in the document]
henrypark133: Fixed in `0e4d6e03`.

I treated this as a real paused-thread bug and fixed it structurally instead of papering over it with a follow-up message.

When auth completion now yields `PairingRequired`, the bridge replaces the resolved auth gate with a new pending pairing gate (same `(user, thread)`, new `request_id`) and broadcasts `onboarding_state` with that new request id. The pairing card carries the request id, and `/api/pairing/{channel}/approve` resolves it via the existing external-callback resume path instead of leaving the engine thread paused with no gate.

That keeps `BridgeOutcome::Pending` truthful: the thread remains paused only while there is still a resumable pending gate backing the pairing step.

Coverage:
- `cargo test transition_engine_pending_auth_request_to_pairing_replaces_gate --lib`
- `cargo test test_pairing_approve_dispatches_external_callback_for_pairing_gate_request --lib --no-default-features --features libsql`

src/channels/wasm/wrapper.rs
```diff
@@ -2743,7 +2810,21 @@ impl WasmChannel {
...
+        let shutdown_tx = self.poll_shutdown_tx.write().await.take();
+        if let Some(tx) = shutdown_tx {
+            let _ = tx.send(());
+        }
+
+        let poll_task = self.poll_task.write().await.take();
+        if let Some(handle) = poll_task
+            && let Err(error) = handle.await
+        {
+            tracing::debug!(channel = %self.name, error = %error, "Polling task join failed");
```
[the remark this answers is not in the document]
henrypark133: Good catch on the shutdown-latency tradeoff, but I do not think the suggested quick fix is safe for this PR.

A couple of nuances from the current code:
- the underlying lack of mid-callback cancellation is not new in this branch; the polling loop has long awaited `execute_poll(...).await` and only observed shutdown between ticks
- `execute_poll()` is already bounded by `callback_timeout`, so this is not an unbounded hang in the normal code path; it is a bounded wait up to the configured callback timeout
- what changed here is that `stop_polling()` now waits for the old poll task to finish before restart/shutdown completes, specifically to avoid two pollers racing on the same workspace-backed polling offset/state

That means the current behavior is better described as bounded shutdown/restart latency, not a new indefinite wedge. I am not applying `abort()` / bounded-join here because the actual callback work runs inside `spawn_blocking`, and letting restart proceed before the old poll side effects are fenced would reintroduce the single-poller/workspace-offset race this join was added to prevent.

If we want to improve responsiveness later, I think the safe version is a larger change: add generation/epoch fencing or another stale-completion guard around poll-side effects first, then make stop/restart more cancellation-friendly.

nickpismenkov (approved):

PR #2519, opened 2026-04-16
ci: build ironclaw docker image hourly

think-in-universe:
## Summary

- Update Docker image CI schedule to run hourly for staging builds.
- Add a staging commit guard so scheduled/staging-target builds are skipped if the current source commit is already published on `nearaidev/ironclaw:staging`.
- Record source commit provenance on built images via `ironclaw.git.sha` label and use the checked-out commit SHA for `sha-*` tags to avoid schedule/ref mismatch.
- Ensure downstream steps (app token creation, `ironclaw-dind` dispatch, and summaries) are skipped when a build is skipped.

## Change Type

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [x] CI/Infrastructure
- [ ] Security
- [ ] Dependencies

## Linked Issue

None

## Validation

- [ ] `cargo build`
- [ ] Relevant tests pass: N/A (workflow-only change; no Rust test suite run in this session).
- [x] Manual testing: Reviewed workflow logic and verified conditions/outputs locally in file checks (schedule cron, staging-target guard, label/tag source SHA usage, and skip-gated downstream steps).

## Security Impact

None

## Database Impact

None

## Blast Radius

Touches `ironclaw` CI only (`.github/workflows/docker.yml`). Risk is limited to Docker image build/publish cadence and skip behavior for staging-target builds, plus downstream dispatch timing to `ironclaw-dind`.

## Rollback Plan

Revert workflow changes in `.github/workflows/docker.yml` to restore previous schedule and unconditional staging build behavior.

## Review Follow-Through

Reviewer focus requested on:
- Whether skipping based on `nearaidev/ironclaw:staging` label `ironclaw.git.sha` matches release expectations for all staging build paths.
- Whether any additional observability is desired when a build is skipped (for example explicit metric/log hooks beyond step summary).

---

**Review track**: C

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -17,9 +17,9 @@ on:
         required: false
         type: string
         default: ""
-  # Daily staging build from the staging branch
+  # Staging build from the staging branch every hour
   schedule:
-    - cron: '0 6 * * *'
+    - cron: '0 * * * *'
 
 env:
   IMAGE_NAME: nearaidev/ironclaw
@@ -39,6 +39,13 @@ jobs:
         with:
           ref: ${{ github.event_name == 'schedule' && 'staging' || '' }}
           persist-credentials: false
+
+      - name: Resolve source git commit
+        id: source_sha
+        run: |
+          SOURCE_SHA="$(git rev-parse HEAD)"
+          echo "sha=${SOURCE_SHA}" >> "$GITHUB_OUTPUT"
+          echo "Source commit: ${SOURCE_SHA}"
 
       - name: Extract version from Cargo.toml
         id: version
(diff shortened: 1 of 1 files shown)


.github/workflows/docker.yml
```diff
@@ -101,30 +109,59 @@ jobs:
           username: ${{ vars.DOCKER_REGISTRY_USER }}
           password: ${{ secrets.DOCKER_REGISTRY_TOKEN }}
 
+      - name: Check if current staging commit is already built
```
serrrfirat, line 109: **Medium Severity** — Skip guard scope too broad; can block intentional rebuilds

The skip check fires whenever `steps.tags.outputs.target == 'runtime-staging'`, which is true for **three** trigger paths:

1. `schedule` — intended (hourly dedup)
2. `workflow_dispatch` with `INPUT_TAG=staging` — **unintended**: a manual dispatch implies intent to rebuild
3. `workflow_call` with `INPUT_TAG=staging` — **unintended**: a calling workflow explicitly requesting a staging build should not be silently skipped

**Suggested fix** — scope the guard to schedule-only:
```yaml
if: github.event_name == 'schedule' && steps.tags.outputs.target == 'runtime-staging'
```

.github/workflows/docker.yml
```diff
@@ -101,30 +109,59 @@ jobs:
...
+
+          echo "Current built commit: ${CURRENT_BUILT_SHA:-<none>}"
+          echo "Current source commit: ${SOURCE_SHA}"
+
+          if [[ -n "${CURRENT_BUILT_SHA}" && "${CURRENT_BUILT_SHA}" == "${SOURCE_SHA}" ]]; then
+            echo "skip=true" >> "$GITHUB_OUTPUT"
+            echo "Current commit already built — skipping."
+          else
+            echo "skip=false" >> "$GITHUB_OUTPUT"
+            echo "Current commit not built yet — proceeding."
```
serrrfirat: **Medium Severity** — Partial build failure can leave worker image permanently stale

The skip check inspects only `nearaidev/ironclaw:staging` (main image). The two build steps run sequentially: main first, then worker. If the main image build+push succeeds (writing the `ironclaw.git.sha` label to Docker Hub) but the worker build subsequently fails:

1. The job fails overall
2. Next hourly run pulls main image, sees label matches current SHA → `skip=true`
3. **Both** builds are skipped — worker image remains at the old commit indefinitely
4. This persists silently until the next staging commit changes the SHA

**Options:**
- (a) Check both images' labels before skipping
- (b) Accept the risk with a comment documenting the assumption that both builds always succeed together
- (c) Swap build order so worker builds first (if it fails, main label isn't written yet)

.github/workflows/docker.yml
```diff
@@ -101,30 +109,59 @@ jobs:
...
+          fi
```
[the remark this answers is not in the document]
think-in-universe: This is a good suggestion. But I don't think we need it for now.

.github/workflows/docker.yml
think-in-universe: Fixed in https://github.com/nearai/ironclaw/pull/2519/commits/3cab2b35791a269b430bc76eec6f819801c40101

.github/workflows/docker.yml
think-in-universe: This is intentional to skip any staging build with the same git commit.

think-in-universe:
@serrrfirat comments resolved. please help review again

henrypark133 (approved):
## Review: workflow skip logic looks coherent

I reviewed the updated Docker workflow end-to-end and did not find a blocker in the changed YAML.

Positives:
- The source-SHA resolution and `ironclaw.git.sha` labeling line up with the new skip check, so the workflow is comparing against the commit that was actually checked out.
- The skip gate is threaded through build, token, dispatch, and summary steps consistently.
- Non-staging paths still proceed because the `check` step only runs for `runtime-staging`, and the later `if` conditions treat the missing output as "not skipped".

Residual risk:
- This approval is based on workflow inspection only; I did not execute the workflow from this review pass.

PR #2523, opened 2026-04-16
fix: mission (routine) creation

### LINKED ISSUE #2524: Event-based routines do not work
I ask my agent to "every time you get a telegram message log it into `telegram.txt` as `[timestamp] sender: message`

The routine is created but it never triggers because it is wrongly labeled as "manual routine" every single time.

gagdiez:
## Summary
Agents were not able to create reactive missions (e.g. "log every telegram message") because of two major shortcomings:

1. the mission cadence (e.g. cron / event / manual) was a simple string, and when malformed - which was incredibly often - ironclaw was defaulting to "manual"
2. A default 5-minute cooldown was applied between triggers, with no way to changing it, so even when correctly created, the event would only trigger every 5min

This PR addresses this issues and:
- Replaces "cadence" with a strict `trigger` structure, which is now required
- Rejects invalid `trigger` shapes with a redable error so the LLM can correct the call
- Forward guardrail fields (cooldown_secs, max_concurrent, etc.) to mission_create as top-level params so LLMs can decide how often a mission triggers
- Expose trigger and guardrails in mission_list output so misconfigured missions are visible
- Add mission_update and mission_delete to the CodeAct preamble so the LLM knows they exist
- Add regression tests and end-to-end integration test

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] CI/Infrastructure
- [ ] Security
- [ ] Dependencies

## Linked Issue

Closes #2524

## Validation

- [x] `cargo build`
- [x] Relevant tests pass:
```
cargo test -p ironclaw_engine --lib runtime::mission
cargo test --lib bridge::effect_adapter
cargo test --test routine_reactive_integration
```

- [x] Manual testing: 
  - I have requested my ironclaw (QLM 3.5 / haiku 4.5) to create a trigger routine (log every telegram message that arrives) and it succeeded
  - Without this update my ironclaw was consistently failing to do so.

## Security Impact

None

## Database Impact

`missions` are stored with a new trigger format, but I have added tests to ensure retro-compatibility

## Blast Radius

`routines`

## Rollback Plan

We would need to return routines to use the old `cadence` string

---

**Review track**: B? C?


gagdiez:
not happy with this PR, seams too unnecessarily complex, will close for now

PR #2504, opened 2026-04-15
[WILL NOT MERGE] a mix of 5 PRs merged together for demo purposes

ilblackdragon:
## Summary

Eleven scoped fixes split out of `feat/portfolio-tests-and-share` (PR #2368) so they can land independently of the portfolio work. Targets staging.

### Engine
- **`fix(engine): make FINAL/FINAL_VAR awaitable in CodeAct scripts`** — `await FINAL(...)` now resolves to `None` instead of raising `TypeError: 'NoneType' object can't be awaited` and dropping the answer. Both `FINAL(x)` and `await FINAL(x)` work.
- **`fix(engine): normalize typographic punctuation before skill activation`** — `iOS`/`macOS` autocorrect turns `I'm` into `I'm`; activation regexes authored with ASCII apostrophes silently failed to match. `default.py` now folds 8 curly quotes + en/em dashes to ASCII before scoring. Fixes the `ceo-setup` non-activation report and any other skill with apostrophes in its patterns.
- **`fix(engine): default max_consecutive_errors to 5`** — bound runaway loops on wedged actions.

### Skills + UI
- **`feat(events): SkillActivated carries activation feedback notes`** — adds an optional `feedback: Vec<String>` field through the SSE/StatusUpdate path with a gateway timeline card so chain-load reasons / marker exclusions can surface to the user.
- **`fix(skills): skill_install never prompts when skill is already loaded`** — mirrors the idempotent execute path in `requires_approval` so `/ceo-setup` follow-up installs don't trigger a useless approval prompt.
- **`feat(skills): paranoid-architect code-review skill v2`** — full rewrite of the code-review skill, now handles both local diffs and GitHub PRs (`owner/repo N`), reads each changed file in full, six-lens review, can post line-level review comments. Recorded e2e fixture for PR #2483.
- **`docs(github): clarify search endpoints, response envelope, @me queries`** — kills the recurring `search_issues` hallucination, adds `is:pr author:@me` recipes, documents the `body` envelope contract.

### Web gateway
- **`feat(web): expose engine v2 threads in chat history and sidebar`** — fixes `#/chat/<engine-thread-id>` rendering empty and engine threads being invisible in the sidebar (v1 dual-write writes into the assistant conversation id, not the engine thread id). Bumps sidebar cap 50→500.

### Security
[...]

PR #2529, opened 2026-04-16
fix(security): redact credentials in HTTP tool + recording interceptor

serrrfirat:
## Summary
- Redact sensitive credentials (Authorization headers, API keys in query params) before persisting HTTP exchanges in action records
- Add `RecordingHttpInterceptor` credential scrubbing for headers and sensitive URL parameters
- Prevents credential leakage into LLM recording logs and action audit trail

Cherry-picked from #2504.

## Test plan
- [ ] `cargo test` passes
- [ ] Verify HTTP tool action records no longer contain raw Authorization headers
- [ ] Confirm recording interceptor scrubs `api_key`, `token`, `secret` params from URLs


henrypark133 (dismissed):
## Review: credential redaction path looks correct

I verified the recorder-side scrubbing and the pre-injection capture boundary in the HTTP tool. I did not find a blocker in the changed code.

Positives:
- `RecordingHttpInterceptor` now sanitizes both request and response headers before persistence.
- The HTTP tool records caller headers instead of the post-injection header set, which closes the main leaked-secret path this PR is targeting.
- The regression test covers header redaction, query-param redaction, and serialized fixture output.

Residual risk:
- I did not run the full test suite from this review pass, so approval here is based on code inspection plus the new targeted regression coverage.

zmanian (commented):
## Summary

Scrubs credential-bearing headers and URL query params before persisting HTTP exchanges in recorded traces, and splits a pre-injection `caller_headers` snapshot so injected `Authorization`/API-key values never reach the recorder. Also dedupes duplicate credential mappings to fix a 401 Bad Credentials regression on GitHub. Solid defensive work with strong test coverage; a few gaps worth addressing.

## Findings

### Blocking

None — no security regressions. But see the request-body gap below.

### Suggestions

- **Request body is not redacted** (`src/llm/recording.rs:198-206`). `redact_exchange_request` only scrubs headers and URL query params — `req.body` passes through verbatim. A POST with `{"api_key": "..."}` or `{"password": "..."}` in the JSON body will land raw in the trace. The ironclaw_safety leak-detector (referenced in the sensitive-headers comment) catches some of this at a different layer, but that's defense in depth, not the primary seal. Consider a JSON-aware body scrubber that walks the value and redacts any key matching `SENSITIVE_QUERY_PARAMS` + a small body-specific list (`password`, `secret`, `private_key`, `token`). At minimum, document the gap explicitly.
- **Basic auth in URL userinfo** (`src/llm/recording.rs:245-271`). `redact_url_query` only touches `?query=`. A URL like `https://user:pat@api.example.com/...` still lands raw in `req.url`. `url::Url::password()` / `set_password(None)` would redact this cleanly. Low-incidence in practice, but the header redaction already covers Basic auth; completing the coverage here is consistent.
- **Dedupe key uses non-canonical JSON serialization** (`src/tools/builtin/http.rs:628-639`). `serde_json::to_string(&m.location).unwrap_or_default()` will produce different strings for semantically equal `CredentialLocation` values if the shape contains a map (HashMap iteration order is nondeterministic). If `CredentialLocation` implements `Eq + Hash` directly, use a `HashSet<(String, CredentialLocation)>` — this avoids the canonicalization question entirely.
- **Test through the caller, not the helper**. The regression test `recording_http_interceptor_redacts_credentials` hits the interceptor directly. Per `.claude/rules/testing.md` ("Test Through the Caller, Not Just the Helper"), add an integration-tier test (or extend an existing HTTP tool test) that drives `HttpTool::execute` with a real credential mapping and asserts the recorder captured the REDACTED form. That's the invariant that actually matters.
- **URL roundtrip normalization** (`src/llm/recording.rs:247-273`). `url::Url::parse + query_pairs_mut().clear() + append_pair()` will re-encode the query segment. If any replay matcher compares exact URL strings for non-sensitive queries, this could cause false misses on trivially-different encodings. Worth confirming the replay matcher is insensitive to percent-encoding differences on the query string.

### Nits

- `src/tools/builtin/http.rs:657-675` the secret preview uses `…` (U+2026). Fine for logs; just note that grepping for an exact middle in logs will require the UTF-8 byte sequence.
- `src/llm/recording.rs:222-229` — `SENSITIVE_QUERY_PARAMS` is missing `auth`, `jwt`, `session`. Low priority but common enough to justify.

## Verdict

**COMMENT** — strong credential-redaction work with good regression coverage. Request-body redaction is the one gap that rises above \"nice to have\" — the rest are incremental hardening. Approve as-is and track the body gap as a follow-up if you prefer to ship this batch cleanly, but don't let it sit.

serrrfirat:
## Addressed review feedback

**@zmanian's suggestions — all addressed:**

- **Request body redaction**: Added `redact_json_value()` that recursively walks JSON bodies and redacts keys matching `password`, `secret`, `private_key`, `token`, `api_key`, `authorization`, etc. Non-JSON bodies pass through (leak-detector in `ironclaw_safety` covers those).
- **URL userinfo redaction**: `redact_url()` now scrubs `user:pass@host` via `set_password(None)` / `set_username("[REDACTED]")`. Plain URLs without userinfo are returned unchanged to avoid normalization artifacts.
- **Dedupe key fix**: `CredentialLocation` now derives `PartialEq + Eq + Hash`, so the dedupe uses `HashSet<(String, CredentialLocation)>` instead of JSON serialization (no more non-canonical key risk).
- **Missing sensitive params**: Added `auth`, `jwt`, `session` to `SENSITIVE_QUERY_PARAMS`.

**gemini-code-assist suggestions — addressed:**
- **Unicode safety in secret preview**: Switched from `secret_str.len()` (byte length) to `chars().count()` and replaced the double-reverse tail extraction with `chars().skip(count - 4)`.

**Additional:**
- Fixed pre-existing clippy `ptr_arg` warning on `redact_headers`.
- Added 7 regression tests covering body scrubbing (flat + nested JSON + non-JSON), userinfo redaction, and URL preservation.

ilblackdragon:
## PR 2529 Review — fix(security): redact credentials in HTTP tool + recording interceptor

### Overview
Adds credential redaction in two places:
1. **`RecordingHttpInterceptor::after_response`** scrubs sensitive headers (`Authorization`, `Cookie`, `X-Api-Key`, …), URL userinfo (`user:pat@host`), URL query params (`access_token`, `api_key`, …), and JSON body keys before storing an exchange.
2. **`HttpTool`** snapshots caller-supplied headers *before* credential injection and passes that snapshot to the recorder, so host-injected `Authorization: Bearer …` never reaches the trace. Also dedupes credential mappings by `(secret_name, location)` to stop double-injection (root cause of GitHub 401 \"Bad credentials\" when a secret was declared both in a WASM tool's capabilities and a skill's `credentials` block).
3. Adds `PartialEq, Eq, Hash` to `CredentialLocation` (needed for the `HashSet` dedup).

### Strengths
- Redaction runs unconditionally on every recorded exchange — you cannot opt out by accident (matches the reasoning in the PR comment).
- Snapshotting `caller_headers` at `http.rs:500` is the right boundary: the real network call still gets the injected token, but the trace never does. Belt-and-suspenders with the interceptor's own header scrub.
- Test coverage on the redactor is thorough: case-insensitive header match, URL userinfo, query-param rewrite, nested JSON, non-JSON passthrough, and a top-level \"raw token must not appear anywhere in the serialized exchange\" check.
- Redacted debug preview (`head…tail`) for operator triage is a good compromise; the `secret_len` was already there and the preview sits only on `debug!` as required by CLAUDE.md.

### Issues

**1. Body-key redaction uses substring match; header and query redaction use equality — inconsistent and over-redacts.**
`recording.rs`: `SENSITIVE_BODY_KEYS.iter().any(|s| lower.contains(s))` vs `SENSITIVE_HEADER_NAMES.iter().any(|h| *h == lower)` vs `SENSITIVE_QUERY_PARAMS.iter().any(|p| *p == lower)`. Consequences of `contains`:
- `token_type` (standard OAuth response field: `\"token_type\": \"Bearer\"`) → redacted. Harmless but loses triage info.
- `token_count`, `tokens_used`, `input_tokens` → redacted in any LLM response body. That's actually a real loss since traces frequently inspect token counts.
- `auth_method`, `auth_url` → redacted. Probably fine.

Recommend either (a) exact match for body keys too, or (b) keep substring but prune the list so `token`/`auth` aren't standalone entries (use `access_token`, `refresh_token`, etc., which still match via equality).

**2. Replay URL mismatch on re-recorded fixtures.**
`redact_url` only touches URLs that have a query string — and in that path it rewrites via `query_pairs_mut().clear(); append_pair()`. Existing fixtures recorded before this PR have non-redacted URLs; new fixtures will have `access_token=%5BREDACTED%5D`. `ReplayingHttpInterceptor::before_request` compares the *incoming* (non-redacted) URL against the *stored* (now-redacted) URL — this trips the \"request mismatch\" warning on every re-recorded exchange with sensitive query params. The soft-warn design means it's not a hard failure, but CI logs will get noisier and drift detection (which relies on that mismatch warning being meaningful) gets weaker. Consider redacting the incoming request's URL the same way before comparing, or stripping sensitive params from both sides.

**3. No regression test for the `caller_headers` split in `http.rs` (test-through-the-caller rule).**
The redaction tests exercise `RecordingHttpInterceptor` directly. There's no test that drives `HttpTool::invoke()` with a credential-injection mapping and asserts the recorded exchange contains neither the injected `Authorization` header nor the raw token. Per `.claude/rules/testing.md`, when a helper (`redact_headers`) gates a side effect (trace persistence) and there's a wrapper between them (`HttpTool`), a helper-level test is not sufficient. Add a test at the integration tier (`cargo test --features integration`) that registers a `github_token` mapping, invokes `HttpTool` via `ToolDispatcher`, and asserts the recorded `HttpExchange.request.headers` contains no `Authorization` entry.

**4. No regression test for the credential-mapping dedup fix.**
Per `.claude/rules/review-discipline.md` (\"Regression test with every fix\"), the double-injection bug (same `github_token` declared by both tool capabilities and skill `credentials` → two `Authorization` headers → GitHub 401) needs a test. Suggest: register two `CredentialMapping` entries with the same `(secret_name, location)`, invoke `HttpTool`, assert exactly one Authorization header is sent (via a `mock` reqwest or through the interceptor snapshot).

### Minor / Nits
- `recording.rs:130` comment notes the `clear()+append_pair()` roundtrip may re-encode params differently from the original URL and concedes \"this could cause false misses\". That's exactly the source of issue #2 above — worth making the mitigation explicit (or centralizing URL canonicalization so live and replay sides go through the same function).
- `redact_url` returns `url.to_string()` on parse failure. Fine, but an unparseable URL is also the one most likely to hide weird credential schemes. A trace warning on the failure path would help operators spot a new URL shape that's slipping past redaction.
- `redact_json_value` scans key names only, not values. The PR comment rightly defers value-shape detection to `ironclaw_safety`'s leak-detector; worth a one-line code comment linking to that component so a future contributor doesn't add overlapping logic.
- `SENSITIVE_BODY_KEYS` includes `\"session\"` via substring match — which redacts any `session_id` / `session_state` correlation field used in OAuth flows. Consider whether that's desired; those are often non-secret public identifiers.

### Risk Summary
The security intent is clearly right and the primary redaction path is well-tested. The two things that would improve confidence before merge: a call-site test for `HttpTool` recording hygiene (so a future regression in how headers are threaded doesn't silently leak creds back into traces), and a dedup regression test. Issue #1 (substring matching for body keys) is a minor over-redaction trade-off that's worth a deliberate decision rather than an accident.

serrrfirat:
## Review feedback addressed

Pushed fixes for all actionable review comments:

### From gemini-code-assist
- **Dedupe key** (`http.rs`): Replaced `serde_json::to_string(&m.location)` with structured `HashSet<(String, CredentialLocation)>` — added `PartialEq, Eq, Hash` derives to `CredentialLocation`
- **Unicode-safe preview** (`http.rs`): Switched from `.len()` (bytes) to `.chars().count()` and `.chars().skip(n-4)` for the tail

### From @zmanian
- **Request body redaction** (`recording.rs`): Added JSON-aware body scrubber that walks nested objects/arrays and redacts keys containing sensitive names (`password`, `api_key`, `secret`, `private_key`, `token`, etc.)
- **URL userinfo** (`recording.rs`): `redact_url_query` now strips username/password from `https://user:pat@host/...` URLs via `set_password(None)` / `set_username("")`
- **Missing query params** (`recording.rs`): Added `auth`, `jwt`, `session` to `SENSITIVE_QUERY_PARAMS`
- **URL re-encoding**: Added inline comment documenting the `clear()+append_pair()` percent-encoding roundtrip and why it's acceptable for replay matching
- **Caller-level integration test**: Tracked as follow-up — needs real credential registry + secrets store at integration tier

### Tests
- Added `recording_http_interceptor_redacts_request_body` test
- Added `recording_http_interceptor_redacts_url_userinfo` test
- All 24 recording tests pass, clippy clean

serrrfirat:
## Addressed @ilblackdragon review feedback

**Commit:** 9ace0940

### Changes

**1. Body-key exact match (Correctness)**
`redact_json_value` now uses exact case-insensitive match (`*s == lower`) instead of substring (`lower.contains(s)`). Fields like `token_count`, `input_tokens`, `session_id`, `auth_method`, `token_type` are no longer over-redacted. Exact-match keys (`token`, `auth`, `session`, `jwt`, etc.) are still redacted.

**2. Replay URL mismatch fix (Correctness)**
`ReplayingHttpInterceptor::before_request` now redacts the incoming URL via `redact_url()` before comparing against stored (already-redacted) URLs. Sensitive query params like `access_token` no longer cause false mismatch warnings.

**3. Caller-level recording test (Test-through-the-caller)**
Added `SpyInterceptor` test that drives `HttpTool::execute()` with a credential registry + stored secret, asserting:
- The interceptor request uses `caller_headers` (no injected `Authorization`)
- The raw token doesn't appear anywhere in the serialized request

**4. Credential dedup regression tests**
- `credential_mapping_dedup_removes_duplicates`: duplicate `(secret_name, location)` pairs collapse to one entry
- `credential_mapping_dedup_preserves_different_locations`: same secret with different injection locations survives dedup

**5. Minor improvements**
- Trace warning on unparseable URL in `redact_url`
- Code comment on `redact_json_value` linking to `ironclaw_safety::LeakDetector`

### Test results
All new tests pass. Full `cargo test --lib` passes (4866 passed, 2 pre-existing failures unrelated to this PR).

ilblackdragon (commented):
## PR 2529: fix(security): redact credentials in HTTP tool + recording interceptor

**Overview** — Plugs a credential-leak class of bug in recorded HTTP exchanges: `Authorization` + API-key headers, URL userinfo, query params like `?access_token=...`, and JSON body keys (`password`, `api_key`, etc.) are now scrubbed to `[REDACTED]` before a trace fixture persists. In parallel, the HTTP tool stops handing post-injection headers to the recorder (only the caller-supplied set reaches the trace), and dedupes `CredentialMapping` entries by `(secret_name, location)` to fix a GitHub `401 Bad credentials` caused by duplicate `Authorization` headers. Adds `Eq + Hash` to `CredentialLocation` to enable the dedup HashSet.

## Strengths

- **Exact-match body-key redaction** with a dedicated regression test (`redact_body_does_not_over_redact_substring_matches`) correctly handles `token_count`, `input_tokens`, `session_id`, `auth_method`, `authorization_type` — these are exactly the landmines a naive `.contains("token")` check would hit. Good defensive design.
- **Caller-level test matches `.claude/rules/testing.md`.** `http_tool_interceptor_sees_caller_headers_not_injected` drives `HttpTool::execute` with a real credential-injecting registry and asserts the *spy interceptor* never sees the injected `Authorization`. That's exactly the "Test Through the Caller, Not Just the Helper" pattern — a helper-only test on `redact_headers` would have missed a regression where `http.rs:643` passed `headers_vec` (post-injection) to the recorder.
- **Defense-in-depth serialized-exchange check** (`!serialized.contains("ghp_...")`) is a strong invariant and the right level of paranoia for a credential-leak fix.
- **Replay matcher correctly redacts the incoming URL before comparing** against stored (already-redacted) URLs — prevents false mismatches, and has a dedicated test.
- Dedup fix is a real bug fix (cherry-picked from #2504) and worth landing.

## Issues

- **Dedup test duplicates prod logic instead of calling it.** `credential_mapping_dedup_removes_duplicates` (`http.rs:768-794`) literally copy-pastes the `seen.insert((m.secret_name.clone(), m.location.clone()))` closure from `execute()`. If prod dedup breaks or is removed, this test still passes. By the project's own "test through the caller" rule, the stronger test is to drive `HttpTool::execute` with a registry containing duplicate mappings and a spy interceptor, then assert `captured.headers` contains exactly one `Authorization`. The existing `SpyInterceptor` is already in the file — extending `http_tool_interceptor_sees_caller_headers_not_injected` to include duplicate-mapping coverage would fold both cases into one behavioral test. As-is, the dedup could be silently deleted in a future refactor and CI would not catch it.
- **`redact_body` doesn't cover `application/x-www-form-urlencoded`.** The `Err(_) => return body.to_string()` path at `recording.rs:180-184` means OAuth token-exchange bodies (`grant_type=authorization_code&client_secret=shhh&code=...`) pass through verbatim. That's directly in scope for this PR — OAuth token exchanges are one of the highest-value leak surfaces, and the body isn't token-*shaped* (`client_secret=shhh` won't trip `ironclaw_safety`'s leak detector). Recommend adding a form-url-encoded branch that parses with `url::form_urlencoded::parse` and redacts by the same `SENSITIVE_BODY_KEYS` set.
- **Query-param redaction changes percent-encoding.** The author already flagged this at `recording.rs:139-144` — `clear() + append_pair()` normalizes encoding, which can cause `ReplayingHttpInterceptor` false misses if the original URL used unusual encoding (`+` vs `%20`, unreserved-char encoding). Acknowledged is better than silent, but worth noting this will eventually bite someone. Preserving the original encoding for non-redacted pairs (walk the raw query string character-by-character, only mutating values whose key is sensitive) would fix it. Low priority.
- **`intercept_req.body` is NOT redacted at the HTTP tool layer** (`http.rs:644-646`) — redaction only runs inside `RecordingHttpInterceptor::after_response`. Correct for the recording path, but any future custom `HttpInterceptor` that logs or persists in `before_request` would see raw bodies. Defense-in-depth would move redaction up to where `intercept_req` is constructed, making the contract "no interceptor ever sees a raw credential in the request descriptor."
- **URL userinfo redaction produces `https://@host/path`** (`recording.rs:107-112` — sets password to None, then username to empty string; the `@` separator survives). Not a security issue, but odd-shaped and potentially confusing in logs. `parsed.set_authority` or manually reconstructing would produce a cleaner `https://host/path`.

## Nits

- `SENSITIVE_BODY_KEYS` and `SENSITIVE_QUERY_PARAMS` diverge: body set has `secret`, `private_key`, `authorization`; query set does not. Intentional per the comment ("body-specific credential fields"), but would benefit from a one-line comment calling out which keys are body-only and why (e.g., "naked `?secret=...` query params are uncommon in real APIs").
- `"auth"` as an exact key in both sets will redact a body field literally named `auth`, which is rare. Harmless, but slightly noisy. `"authorization"` is the more common real-world spelling.
- `replaying_interceptor_matches_redacted_query_params` (`recording.rs:514-543`) stubs the stored URL manually in redacted form. A stronger test would be a record→replay round-trip: record with `?access_token=real_value`, then replay the same raw URL and assert no mismatch warning. Current test covers the matcher but not the invariant that recorded URLs are actually stored redacted.
- `SENSITIVE_HEADER_NAMES` is missing `x-amz-security-token` / `x-amz-*` AWS signing headers and `proxy-authenticate`. Only relevant if AWS/proxy HTTP ever gets recorded; add if the concrete use case arises.

## Verdict
Solid security fix that lands the right layers (header, URL, body, userinfo) and uses the correct test tier for regression coverage. Ship after:
1. Converting the dedup test to drive `HttpTool::execute` (or at minimum extending the existing caller-level test to cover duplicate mappings), and
2. Deciding whether form-url-encoded body redaction is in scope now or a follow-up — if the cherry-pick origin (#2504) is near OAuth flows, I'd fold it into this PR.

Everything else is nit-level.

henrypark133 (dismissed):
The credential-redaction path looks correct in the updated head. src/llm/recording.rs now scrubs sensitive request headers, URL query params, JSON body fields, and response cookie headers before traces are persisted, while src/tools/builtin/http.rs records the pre-injection caller header snapshot so auto-injected Authorization and API-key values never reach the recorder in the first place. The CredentialLocation hashability change also lets the HTTP tool dedupe duplicate host mappings cleanly. I did not find a blocker in the changed code.

ilblackdragon (approved):

PR #2532, opened 2026-04-16
feat(gateway): expose engine v2 threads in chat history and sidebar

serrrfirat:
## Summary
- Render engine v2 threads in the web gateway sidebar and chat history
- Raise sidebar thread cap from 50 to 500 for better v2 thread visibility
- Add v2 engine thread ownership check and message fallback in `chat_history_handler`
- Wire up v2 thread merging in `chat_threads_handler` (both `server.rs` and `handlers/chat.rs`)
- Frontend (`app.js`): switch to deep-linked thread even when not in sidebar list

Cherry-picked from #2504 (commit 8df22ab2), adapted to staging where `chat_history_handler` and `chat_threads_handler` still live in both `server.rs` (route-bound) and `handlers/chat.rs` (pub duplicates).

## Test plan
- [x] `cargo check` passes
- [x] Rust caller-level tests on `chat_history_handler` (in `mod tests` in `server.rs`):
  - `test_chat_history_returns_engine_v2_messages_for_owner` — v2-owned happy path returns synthesized turns
  - `test_chat_history_returns_404_for_cross_user_engine_thread` — cross-user 404 regression guard
  - `test_chat_history_accepts_session_owned_thread_without_db` — session-only ownership branch returns 200
  - `test_chat_history_returns_500_when_ownership_lookup_errors` — DB-error branch returns 500 (pre-existing)
- [x] Playwright e2e (`tests/e2e/scenarios/test_v2_thread_visibility.py`) against an `ENGINE_V2=true` server:
  - Engine-only thread appears in sidebar with `channel=engine`
  - Deep-linking by engine thread id returns synthesized turns from engine messages
- [x] Verify v2 engine threads appear in sidebar when running with engine v2 enabled
- [x] Confirm clicking a v2 thread loads its timeline in the main chat area
- [x] Deep-linking via `#/chat/<engine-thread-id>` loads the thread even if it aged off the sidebar

## Test infrastructure
Added `src/bridge/test_support` (`#[cfg(test)] pub(crate)`) exposing a minimal `ThreadTestStore`, `install_engine_state_with_threads`, `clear_engine_state`, and a shared `ENGINE_STATE_TEST_LOCK` so cross-module tests can seed `ENGINE_STATE` without the router's full TestStore surface.


henrypark133 (changes requested):
## Review: v2 history fallback still has two correctness issues

The v2 thread visibility work is useful, but I verified two problems in the new history path.

### Concerning: DB ownership failures are turned into false 404s
**File:** `src/channels/web/server.rs:2846`
The new ownership check uses `.unwrap_or(false)` on `conversation_belongs_to_user(...)`. If the store errors, the handler now returns `404 Thread not found` instead of surfacing the real `500` path. That makes operational failures look like user-scoped misses and will send debugging in the wrong direction.
**Suggested fix:** keep the old `map_err(...)?` behavior for store failures before falling through to the engine-thread check.

### Concerning: malformed/missing v2 timestamps are rewritten to "now"
**File:** `src/channels/web/server.rs:2960`
The engine-thread fallback synthesizes `ConversationMessage` rows and replaces any missing or unparseable timestamp with `Utc::now()`. That makes the same historical message appear newer on every reload, which destabilizes ordering and pagination for the new deep-link/history path.
**Suggested fix:** skip malformed entries or preserve a deterministic fallback instead of using wall-clock time.

Summary:
- Prior feedback status: partially unresolved; both issues were already raised in prior review and still exist in the current head
- Residual risk: the synthetic-message path still lacks caller-level coverage for ownership failures and malformed engine message timestamps.

henrypark133 (changes requested):
The previous blockers are still present in the current head.

1. High – `src/channels/web/server.rs:2846` still converts ownership-check errors into `false` via `.unwrap_or(false)`, which turns database failures into misleading `404` responses instead of surfacing the real error.
2. Medium – `src/channels/web/server.rs:2960` still rewrites malformed engine-thread timestamps to `Utc::now()`, and the synthetic fallback entry also fabricates a fresh UUID later in the same path. That makes ordering and reload behavior depend on when the page was rendered rather than on stable persisted data.

I don't see follow-up changes that address either issue, so I can't approve this revision yet.

zmanian (commented):
## Summary

Exposes engine v2 threads in the web gateway's chat history and sidebar: extends the thread-ownership check to include the v2 engine thread store, synthesizes history from engine thread messages when no v1 conversation row exists, merges v2 threads into the sidebar list (capped at 500), and always switches to a deep-linked thread even when it's not in the loaded sidebar. Fixes a real "my thread disappeared" UX bug.

## Findings

### Blocking

None.

### Suggestions

- **Ownership-check silently masks DB errors** (`src/channels/web/server.rs:2837-2863`). The old code returned `StatusCode::INTERNAL_SERVER_ERROR` if `conversation_belongs_to_user` failed. The new code uses `.unwrap_or(false)` and falls through to the session + engine checks. If the Postgres/libSQL query actually errored (timeout, network, permission), we'll now return 404 "Thread not found" instead of 500. 404 is arguably wrong for an infrastructure failure and will make ownership bugs harder to diagnose. Consider logging the error at `warn!` (wait — per CLAUDE.md, `warn!` corrupts the REPL/TUI; use `debug!` here since this is the webhook path anyway, or `tracing::error!` which the REPL also doesn't render) and treating a DB error distinctly from "not owned".
- **Duplicate handler implementations** (`src/channels/web/server.rs:2960-3105` AND `src/channels/web/handlers/chat.rs:125-215`). The PR description acknowledges this ("adapted to staging where `chat_history_handler` and `chat_threads_handler` still live in both"). Two copies means the next change has to land in both, and one is easy to miss. Worth a TODO comment on both copies pointing at each other, plus an issue to de-duplicate. Otherwise this drifts.
- **Synthetic message IDs are regenerated per request** (`src/channels/web/server.rs:2973`): `uuid::Uuid::new_v4()` for each synthesized message. Any client caching by `id` will see the same message as a "new" message on reload. If the UI doesn't cache by ID, this is fine; if it does (message selection, citation links), it's a bug in waiting. Alternative: derive a stable synthetic ID from `(thread_id, sequence_index)`.
- **500-thread cap is still a cap** (`handlers/chat.rs:128`, `server.rs:3022`). Fine for the single-user demo explicitly called out in the comment, but every invocation now serializes up to 500 rows plus engine threads. At scale this is pagination territory. Non-blocking but worth an issue.
- **No caller-level test.** `.claude/rules/testing.md` calls out that predicate + side-effect pairs need a caller-level test. The three new ownership paths (v1 row, v1 session, v2 engine thread) are a predicate gating whether 404 is returned. Ideally one `axum`-driven test per path asserts the correct status and payload. Listing engine-thread merging behavior in an integration test would catch a lot of regressions cheaply.

### Nits

- `app.js:3527-3544` — the `console.warn('[chat] thread ... not in sidebar list')` will log on every deep-link that didn't make the sidebar's 500-item window. Could be noisy once users accumulate history. Consider making it one-shot or gated on a debug flag.

## Verdict

**COMMENT** — good UX fix. Surface-level concerns; the error-masking in the ownership check is the one worth addressing before merge.

henrypark133 (dismissed):
## Review: v2 history fallback still has two correctness bugs

The visibility improvement is useful, but the two previously raised history-path issues still reproduce in the current head.

### Concerning: DB ownership failures still become false 404s
**File:** `src/channels/web/server.rs:2846`
`conversation_belongs_to_user(...).await.unwrap_or(false)` still converts store failures into `false`, so a database/network error now falls through to `404 Thread not found` instead of surfacing the real `500` path. That makes operational failures look like user-scope misses.

**Suggested fix:** preserve the old error-mapped DB branch before falling through to the in-memory and engine-thread ownership checks.

### Concerning: synthetic v2 history is still non-deterministic
**File:** `src/channels/web/server.rs:2960`
The fallback still rewrites missing or malformed timestamps to `Utc::now()` and still assigns `Uuid::new_v4()` per synthesized message. Reloading the same engine thread can therefore change ordering and message identity based purely on render time.

**Suggested fix:** keep deterministic persisted values, or skip malformed entries instead of synthesizing wall-clock timestamps and random IDs.

Summary:
- Prior feedback status: partially unresolved; both issues were already raised and are still present in the current head
- Residual risk: the v2 history path still lacks caller-level regression coverage for the ownership-error branch and malformed engine-thread messages.

ilblackdragon:
### Update since @henrypark133's last review

Summary of what's landed since the 2026-04-17 review flagging DB-error → false 404 and non-deterministic synthetic v2 history.

**Review-comment fixes** — [`c17f29b8`](https://github.com/nearai/ironclaw/pull/2532/commits/c17f29b806e83d00dada2cd055037d1f7df2d754)
- Ownership-check DB errors now return `500 "Database error"` (not `404`). The `.unwrap_or(false)` was replaced with an explicit `match` that `tracing::error!`s and surfaces `INTERNAL_SERVER_ERROR` — see `src/channels/web/features/chat/mod.rs:chat_history_handler`.
- Engine v2 synthetic messages are now deterministic. The random `Uuid::new_v4()` and `Utc::now()` fallback both went away — `stable_engine_history_message_id` is a v5 UUID seeded by `(thread_id, index, role, timestamp, content)`, and malformed/missing timestamps are logged and skipped via `engine_history_entry_to_message` rather than re-stamped to wall-clock time.
- Pinned by two helper-level unit tests: `test_engine_history_entry_skips_malformed_timestamp`, `test_engine_history_entry_uses_stable_id`.

**Caller-level regression tests** — [`6ba38b54`](https://github.com/nearai/ironclaw/pull/2532/commits/6ba38b54c604e0a120e506fdf7df17afc17e283e) (per `.claude/rules/testing.md` "Test Through the Caller")

Rust (in `src/channels/web/server.rs::tests`, driving `chat_history_handler` directly):
- `test_chat_history_returns_500_when_ownership_lookup_errors` — DB-error branch returns 500 with `"Database error"`, not 404.
- `test_chat_history_returns_engine_v2_messages_for_owner` — v2-owned happy path synthesises turns.
- `test_chat_history_returns_404_for_cross_user_engine_thread` — alice can't read bob's engine thread.
- `test_chat_history_accepts_session_owned_thread_without_db` — the in-memory session-only ownership branch.

Playwright (`tests/e2e/scenarios/test_v2_thread_visibility.py`, `ENGINE_V2=true`):
- `test_engine_only_thread_appears_in_sidebar_with_engine_channel` — the v2 merge path surfaces engine-only threads with `channel=engine`.
- `test_history_synthesizes_messages_for_deep_linked_engine_thread` — `/api/chat/history?thread_id=<engine-id>` returns synthesised turns when the v1 conversation row is missing.

To keep the seeding minimal, added `bridge::test_support` (a `#[cfg(test)] pub(crate)` module: `ThreadTestStore`, `install_engine_state_with_threads`, `clear_engine_state`, `ENGINE_STATE_TEST_LOCK`) so cross-module tests can install a read-only engine state without the router's full `TestStore`.

**Catch-up with staging** — [`05ff42ce`](https://github.com/nearai/ironclaw/pull/2532/commits/05ff42ce7a10ba9bbc3364737b4858683aff203e), [`8147b0ca`](https://github.com/nearai/ironclaw/pull/2532/commits/8147b0ca), [`3833291a`](https://github.com/nearai/ironclaw/pull/2532/commits/3833291a)

Staging moved substantially while this was open (ironclaw#2599 stages 4c/4d/5 + attachments + `mission_get` + js/css split). Merged in, porting the PR changes to the new locations:
- Chat handlers moved from `server.rs` + `handlers/chat.rs` to `features/chat/mod.rs` (stage 4c). The v2 ownership extension, engine sidebar merge, 500 cap bump, and engine fallback now live there.
- `app.js` split into `js/core/*` + `js/surfaces/*`. The frontend change (always-switch deep-linked thread with DEBUG-gated warn) now lives in `js/core/history.js::loadThreads`.
- `EngineState` grew `extension_manager` and `project_root` fields on staging (#2549 + attachments). The `test_support` installer was updated to populate them with `None` / `resolve_project_root()`.
- `clippy::question_mark` (denied on staging CI with `-D warnings`): rewrote the `let Some(role_raw) = … else { return None }` in `engine_history_entry_to_message` as a `?`.
- `Thread::complete_turn` → `conclude_turn(TurnOutcome::Completed(..))` rename picked up in the session-owned test.
- `HistoryQuery` fields made `pub(crate)` so server-side caller tests can construct it without leaking into wire shape.

The previous stale "duplicate `chat_threads_handler` in handlers/chat.rs" TODO was dropped when the duplicate got consolidated on staging.

ilblackdragon (approved):

Issue #2447, opened 2026-04-14
Engine v2 should require real tool evidence for explicit execution requests


serrrfirat:
## Problem

When a user explicitly asks the agent to execute a previously discussed operation, engine v2 can still accept a plain-text model response as final. In the observed failure mode, the model says it cannot execute tools from the current reply path instead of calling an available tool or surfacing a real runtime blocker.

This is not primarily a prompt wording issue. In v2, `src/bridge/llm_adapter.rs` uses `tool_choice("auto")`, and `crates/ironclaw_engine/orchestrator/default.py` currently accepts plain text unless the existing `signals_tool_intent()` heuristic catches future-tense phrases like "I'll check...". Capability-refusal text like "I can't execute from this reply path" can slip through.

## Desired behavior

For explicit execution requests, engine v2 should resolve an execution obligation before returning final text. Final text is acceptable only after one of these occurs:

- at least one relevant action/tool call was attempted
- a permission, approval, auth, policy, lease, or hook gate blocked the action
- the runtime determined there is no matching tool/capability
- the tool attempt failed and the final answer reports the actual runtime error

Do not force tool usage when permission/policy blocks it. Approval/auth gates must remain authoritative.

## Proposed v2 change

Add a thread-level execution obligation, likely in `ThreadConfig`, for turns where the caller detects explicit execution intent:

```rust
#[serde(default)]
pub require_action_attempt: bool,
#[serde(default = "default_action_requirement_nudges")]
pub max_action_requirement_nudges: u32,
```

Set it in `src/bridge/router.rs` before calling `ConversationManager::handle_user_message(...)` for inputs like `execute them`, `run ...`, `check ...`, `fetch ...`, etc.

Pass it through `crates/ironclaw_engine/src/executor/orchestrator.rs::build_orchestrator_inputs()` into the Python orchestrator config.

In `crates/ironclaw_engine/orchestrator/default.py`, before accepting text or `FINAL(...)` as final, check whether:

- `require_action_attempt` is true
- no action/code path has been attempted yet
- a gate/blocker has not already resolved the requirement
- live actions are available via `__get_actions__()`

If so, append a corrective internal user message and continue the loop:

> The user asked for execution, but no action/tool call was attempted. If an action is available, call it now. If policy, permission, auth, or missing capability blocks execution, report that actual blocker.

Mark the obligation resolved as soon as the model enters a code/action path, before executing the call. This prevents approval/auth/permission outcomes from causing retry loops.

## Permission nuance

[...]

PR #2539, opened 2026-04-16
feat(engine): execution obligation -- require tool attempt on explicit user commands

standardtoaster:
## Summary

When a user explicitly asks the v2 engine to execute something ("run the tests", "fetch the data"), the model can respond with a false capability refusal without ever attempting a tool call. This adds an execution obligation that nudges the model to actually try before accepting text-only responses.

## Problem

In v2, `tool_choice("auto")` lets the model decide whether to call tools. The existing `signals_tool_intent()` nudge catches cases where the model says "I'll search for..." without calling a tool, but doesn't catch outright refusals. User says "run the echo tool", model responds "I cannot execute tools from this reply path", engine accepts that as final.

## Approach

1. **Intent detection** (`reasoning.rs`): `user_signals_execution_intent()` detects imperative execution phrases in user messages. Strips code blocks and quoted strings. Deliberately conservative -- excludes context-dependent phrases ("go ahead"), ambiguous verbs ("try", "check"), and question framing ("can you run"). The heuristic is intentionally narrow. The obligation infrastructure (`require_action_attempt` on ThreadConfig, set by the router) is detection-strategy-agnostic, so a more sophisticated classifier can be swapped in without changing the enforcement side.

2. **Obligation enforcement** (`default.py`): When `require_action_attempt` is set, the orchestrator refuses text-only FINAL until the model attempts at least one action/code path. Only fires when tools are available (`__get_actions__()`). Mutually exclusive with existing tool-intent nudge. Resolves on entering action/code branch (before execution) so approval gates don't cause retry loops. Bounded by `max_action_requirement_nudges` (default 2).

**Open question:** Is this the right level of conservatism for the heuristic, or should it cast a wider net? The nudge budget bounds the downside of false positives, so being more aggressive has a low cost. Context-dependent phrases ("go ahead", "yes do it") and ambiguous verbs ("check the ...") were excluded because they produced false positives in testing, but reasonable people could disagree.

## Test plan

8 unit tests for the heuristic (positives, negatives, code block/quote immunity, case insensitivity). 3 e2e tests through the full engine pipeline (router -> thread -> Python orchestrator -> trace LLM):

- **Nudge fires:** "run the echo tool" -> model text refusal -> nudge injected -> model calls tool on retry. Asserts tool called and >= 2 LLM invocations.
- **No nudge on conversational:** "What's the weather?" -> text accepted in 1 LLM call.
- **Exhaustion terminates:** "run the echo tool" -> model refuses 3 times -> nudge budget exhausted -> text accepted as final. Asserts 3 LLM calls, no tools called. Proves the feature terminates.

```
$ cargo test --lib -- reasoning::tests::execution_intent
test result: ok. 8 passed; 0 failed

$ cargo test --features libsql --test e2e_engine_v2
test result: ok. 12 passed; 0 failed (9 existing + 3 new)
```

Addresses #2447

### DIFF
--- a/crates/ironclaw_engine/orchestrator/default.py
+++ b/crates/ironclaw_engine/orchestrator/default.py
@@ -469,6 +469,8 @@ def run_loop(context, goal, actions, state, config):
     nudge_enabled = config.get("enable_tool_intent_nudge", True)
     # None means "no limit" — callers can disable the guard explicitly.
     max_consecutive_errors = config.get("max_consecutive_errors", 5)
+    obligation_enabled = config.get("require_action_attempt", False)
+    max_obligation_nudges = config.get("max_action_requirement_nudges", 2)
     consecutive_nudges = 0
     consecutive_errors = 0
     consecutive_action_errors = 0
@@ -574,15 +576,36 @@ def run_loop(context, goal, actions, state, config):
                 )
                 continue
 
-            # Non-intent text response — reset nudge counter and finish
+            # Non-intent text response — reset nudge counter
             if not signals_tool_intent(text):
                 consecutive_nudges = 0
+
+            # Check execution obligation: user asked for execution but model
+            # gave text without attempting any action/code. Fires only when
+            # no action/code has been attempted this turn, the obligation
+            # has not been resolved (e.g. by a gate on a prior step), and
+            # the tool-intent nudge did not already fire this step (the two
+            # mechanisms are mutually exclusive to avoid double-nudging).
+            if (obligation_enabled
+                    and consecutive_nudges == 0
+                    and not state.get("_obligation_resolved", False)
+                    and state.get("_obligation_nudge_count", 0) < max_obligation_nudges):
+                state["_obligation_nudge_count"] = state.get("_obligation_nudge_count", 0) + 1
+                append_message(
+                    working_messages,
+                    "User",
+                    "You were asked to perform an action, but you responded with text only.\n"
+                    "Do NOT describe or explain — call the appropriate tool now.\n"
+                    "Use the tool_calls mechanism to invoke the tool.",
+                )
+                continue
 
             # Plain text response - done
             __transition_to__("completed", "text response")
             return complete_result(state, "completed", text)
 
         elif resp_type == "code":
+            state["_obligation_resolved"] = True  # code attempt satisfies obligation
             code = response.get("code", "")
             append_message(working_messages, "Assistant", "```repl\n" + code + "\n```")
 
@@ -615,6 +638,7 @@ def run_loop(context, goal, actions, state, config):
                     "consecutive_errors": consecutive_errors,
                     "consecutive_action_errors": consecutive_action_errors,
                     "compaction_count": state.get("compaction_count", 0),
+                    "obligation_nudge_count": state.get("_obligation_nudge_count", 0),
                 })
                 __transition_to__("waiting", "gate paused: " + gate.get("gate_name", "unknown"))
                 return {
@@ -635,6 +659,7 @@ def run_loop(context, goal, actions, state, config):
                     "consecutive_errors": consecutive_errors,
                     "consecutive_action_errors": consecutive_action_errors,
                     "compaction_count": state.get("compaction_count", 0),
+                    "obligation_nudge_count": state.get("_obligation_nudge_count", 0),
                 })
                 if approval.get("need_authentication"):
                     __transition_to__("waiting", "authentication needed")
@@ -673,9 +698,11 @@ def run_loop(context, goal, actions, state, config):
                 "consecutive_errors": consecutive_errors,
                 "consecutive_action_errors": consecutive_action_errors,
                 "compaction_count": state.get("compaction_count", 0),
+                "obligation_nudge_count": state.get("_obligation_nudge_count", 0),
             })
 
         elif resp_type == "actions":
+            state["_obligation_resolved"] = True  # action attempt satisfies obligation
             # Tier 0: structured tool calls.
             # NOTE: consecutive_nudges is NOT reset here (V1 semantics).
             # Only non-intent text responses reset the counter.
@@ -773,6 +800,7 @@ def run_loop(context, goal, actions, state, config):
                         "consecutive_errors": consecutive_errors,
                         "consecutive_action_errors": consecutive_action_errors,
                         "compaction_count": state.get("compaction_count", 0),
+                        "obligation_nudge_count": state.get("_obligation_nudge_count", 0),
                     })
                     gate = r
                     # Get action info from the original call or the result
@@ -794,6 +822,7 @@ def run_loop(context, goal, actions, state, config):
                         "consecutive_errors": consecutive_errors,
                         "consecutive_action_errors": consecutive_action_errors,
                         "compaction_count": state.get("compaction_count", 0),
+                        "obligation_nudge_count": state.get("_obligation_nudge_count", 0),
                     })
                     __transition_to__("waiting", "authentication needed")
                     return {
@@ -811,6 +840,7 @@ def run_loop(context, goal, actions, state, config):
                         "consecutive_errors": consecutive_errors,
                         "consecutive_action_errors": consecutive_action_errors,
                         "compaction_count": state.get("compaction_count", 0),
+                        "obligation_nudge_count": state.get("_obligation_nudge_count", 0),
                     })
                     __transition_to__("waiting", "approval needed")
                     return {
@@ -894,6 +924,7 @@ def run_loop(context, goal, actions, state, config):
                 "consecutive_errors": consecutive_errors,
                 "consecutive_action_errors": consecutive_action_errors,
                 "compaction_count": state.get("compaction_count", 0),
+                "obligation_nudge_count": state.get("_obligation_nudge_count", 0),
             })
 
     # Max iterations reached
--- a/crates/ironclaw_engine/orchestrator/default.py
+++ b/crates/ironclaw_engine/orchestrator/default.py
@@ -496,16 +496,6 @@ def run_loop(context, goal, actions, state, config):
     obligation_enabled = config.get("require_action_attempt", False)
     max_obligation_nudges = config.get("max_action_requirement_nudges", 2)
 
-    # Enable obligation from the latest user message in context, not just
-    # thread config. This covers the resume path where a suspended thread is
-    # restarted with a new user message that signals execution intent — the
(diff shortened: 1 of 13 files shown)


henrypark133 (changes requested):
## Review: execution obligation misses the common follow-up path

The new obligation mechanism is directionally right, but I found one verified behavioral gap.

### Critical: obligation only applies when spawning a new thread
**File:** `src/bridge/router.rs:2550`, `crates/ironclaw_engine/src/runtime/conversation.rs:222`
`handle_with_engine_inner()` now builds a `ThreadConfig` with `require_action_attempt=true`, but `ConversationManager::handle_user_message()` only consumes that config in the `None => spawn_thread_with_history(...)` branch. When a conversation already has an active foreground thread, the message is injected/resumed instead, so the new flag is ignored. That means the exact failure this PR is fixing still reproduces on the common follow-up path: user opens an existing thread, says "run the tests", model returns a text refusal, and no obligation nudge is enforced.
**Suggested fix:** propagate the obligation decision onto existing/resumed threads as well, or derive it at execution time from the incoming user message rather than only from initial thread config.

Summary:
- Prior feedback status: no prior blocker review
- Residual risk: tests only cover the new-thread path today, so this regression surface stays unverified without a caller-level follow-up-thread test.

standardtoaster:
@henrypark133 thanks for the catch on the follow-up path -- that was a real gap.

Addressed all review items. Obligation now fires on inject and resume paths via per-message intent detection in the Python orchestrator, with state flag resets so persisted `_obligation_resolved` from prior runs can't suppress new obligations.

New test `gate_resume_with_execution_obligation` covers the resume path end-to-end (gate pause, resume with "run the echo tool", nudge fires, echo called). The spawn path was already covered.

One gap I could use input on: I can't test the inject-into-running-thread path with the current e2e rig since it processes messages sequentially -- turn 2 always arrives after turn 1's thread completes. Is there existing infrastructure for concurrent message delivery to a running thread, or a pattern you'd recommend?

henrypark133 (approved):
## Review: previous blocker looks addressed

I re-checked the current head around the follow-up/resume path that previously missed execution obligation. The key change is in crates/ironclaw_engine/orchestrator/default.py: the orchestrator now re-enables obligation from the latest user message in context and from injected follow-up messages, so the guard no longer relies solely on the initial ThreadConfig at spawn time.

Prior feedback status: the earlier blocker about existing/resumed threads ignoring require_action_attempt appears addressed in the current head.

Residual risk: my local targeted test command hit the wrong cargo target, so this approval is based on checked-out code plus the new resume-path coverage in tests/engine_v2_gate_integration.rs, not on a rerun of that integration test in this pass.

crates/ironclaw_engine/orchestrator/default.py
```diff
@@ -574,7 +626,28 @@ def run_loop(context, goal, actions, state, config):
...
                 continue
 
-            # Non-intent text response — reset nudge counter and finish
+            # Check execution obligation BEFORE resetting consecutive_nudges.
+            # This ensures the mutual exclusion guard (consecutive_nudges == 0)
+            # correctly reflects whether the tool-intent nudge fired this turn.
+            # If tool-intent nudge fired and exhausted its budget, consecutive_nudges > 0
+            # and the obligation is skipped. The reset happens after.
+            available_actions = __get_actions__()
+            if (obligation_enabled
```
serrrfirat, line 635: **Medium Severity** · Confidence: Likely · Category: Correctness

**FINAL() bypasses obligation entirely.**

The `extract_final(text)` check ~15 lines above this block returns early and completes the thread before the obligation check ever runs. A model that wraps its refusal in `FINAL("I cannot execute tools from this path")` will terminate the thread without a single nudge firing.

CodeAct-trained models produce `FINAL()` naturally — this is the expected output format in v2. So the bypass is realistic, not just theoretical.

The existing `signals_tool_intent()` nudge has the same vulnerability, but since this PR's stated purpose is catching refusals that the tool-intent nudge *misses*, the FINAL bypass undermines the feature's core value proposition.

**Suggested fix:** Check obligation before the FINAL extraction, or detect FINAL-wrapped refusals:

```python
# Before extract_final:
final_answer = extract_final(text)
if final_answer is not None:
    # If obligation is active and unresolved, treat FINAL-wrapped text
    # as a refusal rather than a completion.
    if (obligation_enabled
            and not state.get("_obligation_resolved", False)
            and state.get("_obligation_nudge_count", 0) < max_obligation_nudges
            and len(__get_actions__()) > 0):
        state["_obligation_nudge_count"] = state.get("_obligation_nudge_count", 0) + 1
        append_message(working_messages, "User",
            "You were asked to perform an action, but you wrapped a refusal in FINAL().\n"
            "Do NOT use FINAL() until you have attempted the requested action.\n"
            "Call the appropriate tool now.")
        continue
    __transition_to__("completed", "FINAL() in text")
    return complete_result(state, "completed", final_answer)
```

crates/ironclaw_engine/orchestrator/default.py
```diff
@@ -574,7 +626,28 @@ def run_loop(context, goal, actions, state, config):
...
+                    and consecutive_nudges == 0
```
serrrfirat, line 636: **Medium Severity** · Confidence: Likely · Category: Correctness

**Tool-intent nudge exhaustion permanently blocks obligation for the rest of the thread.**

The `consecutive_nudges == 0` guard here creates a one-way dependency: once `signals_tool_intent()` fires (even once), `consecutive_nudges > 0` and the obligation is skipped. The counter is only reset *after* the obligation check, so the obligation never gets a chance.

Concrete sequence:
1. User: "run the echo tool" → obligation enabled
2. LLM: "I'll search for the echo tool and run it" → `signals_tool_intent` fires, `consecutive_nudges = 1`, continue
3. LLM: "I'll try to execute it now" → nudge fires again, `consecutive_nudges = 2` (max), continue
4. LLM: "I cannot actually run tools" → tool-intent doesn't fire (no intent phrase), but `consecutive_nudges == 2 ≠ 0` → **obligation skipped** → counter reset → thread completes with refusal

The mutual exclusion is intentional per the comment, but the *cumulative* effect means a model that first expresses intent and then refuses will bypass the obligation entirely. The two mechanisms can't both protect the same thread.

**Suggested fix:** Reset `consecutive_nudges` before the obligation check when the text does NOT signal tool intent (i.e., move the reset above):

```python
# Reset nudge counter on non-intent text BEFORE obligation check
if not signals_tool_intent(text):
    consecutive_nudges = 0

# Now obligation can fire even after prior tool-intent nudges
available_actions = __get_actions__()
if (obligation_enabled
        and consecutive_nudges == 0
        ...):
```

This preserves mutual exclusion within a single turn (if tool-intent fires via `continue`, obligation is skipped) but allows obligation to fire on subsequent non-intent responses.

src/llm/reasoning.rs
```diff
@@ -108,6 +108,52 @@ pub fn llm_signals_tool_intent(response: &str) -> bool {
...
+/// Detect explicit execution intent in a user message.
+///
+/// Returns `true` for imperative requests like "run it", "execute the script",
+/// "fetch the data", "please deploy the service". Strips code blocks and quoted
+/// strings before matching to avoid false positives from examples.
+///
+/// Deliberately excludes context-dependent phrases ("go ahead", "yes do it")
+/// that require multi-turn understanding -- those belong in a classifier tier.
+pub fn user_signals_execution_intent(text: &str) -> bool {
+    let stripped = strip_code_blocks(text);
```
serrrfirat, line 120: **Low Severity** · Confidence: Likely · Category: Correctness

**`strip_code_blocks` / `strip_quoted_strings` only strips double-quoted strings — single quotes pass through.**

A user message like `He said 'run the tests' but I disagree` would match `"run the "` and trigger the obligation as a false positive. The Rust and Python implementations both have this gap.

The nudge budget (2) bounds the downside, and users rarely single-quote execution phrases in practice, but the test `execution_intent_ignores_quoted_strings` only exercises double quotes — leaving the single-quote path unexercised and the behavior undocumented.

Consider either:
1. Extending `strip_quoted_strings` to also strip single-quoted strings, or
2. Adding a negative test that documents the intentional limitation: `assert!(user_signals_execution_intent("He said 'run the tests'"))` with a comment explaining why single quotes are not stripped.

crates/ironclaw_engine/orchestrator/default.py
```diff
@@ -140,6 +140,30 @@ def signals_tool_intent(text):
     return False
 
 
+def signals_execution_intent(text):
```
serrrfirat, line 143: **Low Severity** · Confidence: Likely · Category: Maintenance

**Rust and Python phrase lists are manually synchronized with no parity test.**

This Python function and its Rust counterpart (`user_signals_execution_intent` in `src/llm/reasoning.rs`) have identical phrase lists today, but they're in different languages with no automated verification. A future edit to one will silently diverge the other.

The dual implementation is necessary (Python can't call Rust host functions for this), but consider one of:
1. A shared JSON fixture file that both implementations load/reference
2. A test that extracts both phrase lists and asserts equality
3. At minimum, a cross-reference comment with the exact file:line of the other implementation so editors know to update both

crates/ironclaw_engine/orchestrator/default.py
```diff
@@ -574,7 +626,28 @@ def run_loop(context, goal, actions, state, config):
                 )
                 continue
 
-            # Non-intent text response — reset nudge counter and finish
+            # Check execution obligation BEFORE resetting consecutive_nudges.
+            # This ensures the mutual exclusion guard (consecutive_nudges == 0)
+            # correctly reflects whether the tool-intent nudge fired this turn.
+            # If tool-intent nudge fired and exhausted its budget, consecutive_nudges > 0
+            # and the obligation is skipped. The reset happens after.
+            available_actions = __get_actions__()
```
serrrfirat, line 634: **Nit** · Confidence: Certain · Category: Performance

**`__get_actions__()` called unconditionally on every text response, even when obligation is disabled (the default).**

This host function crosses the Python→Rust boundary. When `obligation_enabled` is `False` (the common case — most messages don't signal execution intent), the result is unused.

```python
# Before (current):
available_actions = __get_actions__()
if (obligation_enabled and ...):

# After:
if (obligation_enabled
        and consecutive_nudges == 0
        and not state.get("_obligation_resolved", False)
        and state.get("_obligation_nudge_count", 0) < max_obligation_nudges):
    available_actions = __get_actions__()
    if len(available_actions) > 0:
        ...
```

Moves the host call inside the guard so it's only invoked when the obligation is actually active.

PR #2548, opened 2026-04-16
feat: workspace entities with membership and cross-workspace sharing

standardtoaster:
## Summary

Rebases and completes #1734 onto current staging. DB-backed workspace entities with membership, cross-workspace sharing, and scoped access across the full stack.

- Adds `users`, `api_tokens`, `workspaces`, `workspace_members` tables
- Adds nullable `workspace_id` scoping to conversations, jobs, memory, routines, and settings
- Personal workspace (`workspace_id = NULL`) is the backward-compatible default
- Workspace access is membership-based: owner > admin > member > viewer
- Both PostgreSQL (V25) and libSQL (incremental 25) backends

Built on @serrrfirat's implementation in #1734. All review feedback from @ilblackdragon and @zmanian addressed.

## Changes from #1734

| Issue | Fix |
|---|---|
| PG migration V16 collision | Renumbered to V25 |
| Thread ownership bypass in chat_history | Ownership check runs for all threads |
| TOCTOU in member upsert/delete | Atomic DB methods with transactional last-owner guard |
| Settings export/import scope bypass | Rejects workspace-scoped requests (501) |
| Archive requires only manager | Requires owner |
| Routine viewer can trigger | 403 for viewers |
| `joined_at` reset on role update | Removed from ON CONFLICT UPDATE |
| `validate_workspace_slug` logic | Rewritten |
| Dead/duplicate chat handlers | Consolidated into handlers/chat.rs (-700 lines) |
| Dual Query extractors | Merged workspace field into SettingScopeQuery |
| `workspace_id` String vs Uuid in JobContext | Converted to `Option<Uuid>` |
| `find_config_documents` missing workspace_id | Fixed (runtime panic) |
| Job save discards malformed workspace_id | Hard-fails |
| `resolve_workspace_scope` 2 DB round-trips | Single JOIN query |
| Last-owner guard duplicated 4x | Extracted shared helpers |
| V14 released migration edited | Reverted |
| Axum route override (GET/POST on same path) | Combined with method chaining |

## Test plan

- [x] `cargo clippy --all-features -p ironclaw --tests -- -D warnings` -- zero warnings
- [x] `cargo check --no-default-features --features libsql` -- clean
- [x] 4866 unit tests, 0 failures
- [x] 11 workspace integration tests (`cargo test --test workspaces_integration`)
- [x] 5 e2e tests against live instance (`pytest scenarios/test_workspace_entities.py`)
- [x] 11-step manual validation against live server (output below)

### Manual validation (live server output)

```
$ curl -s -X POST "$BASE/api/workspaces" -H "$H" \
  -d '{"name":"Demo Team","slug":"demo-team","description":"Manual test","settings":{}}' | jq '{slug, role, status}'
{
  "slug": "demo-team",
  "role": "owner",
  "status": "active"
}

$ curl -s "$BASE/api/workspaces" -H "$H" | jq ".workspaces[].slug"
"demo-team"

$ curl -s "$BASE/api/workspaces/demo-team" -H "$H" | jq '{name, slug, status}'
{
  "name": "Demo Team",
  "slug": "demo-team",
  "status": "active"
}

$ curl -s -X PUT "$BASE/api/workspaces/demo-team" -H "$H" \
  -d '{"name":"Updated Team","description":"Updated","settings":{"custom":true}}' | jq .name
"Updated Team"

$ curl -s -X POST "$BASE/api/workspaces" -H "$H" \
  -d '{"name":"Bad","slug":"-bad","description":"","settings":{}}' → HTTP status
400

$ curl -s -X POST "$BASE/api/chat/thread/new?workspace=demo-team" -H "$H" | jq .id
"05ac605f-6dc6-4565-ada8-6b712f739ff0"

$ curl -s -X POST "$BASE/api/chat/thread/new" -H "$H" | jq .id
"e8b1eba1-bea9-462f-920f-062be0d57dcc"

$ curl -s "$BASE/api/chat/threads" -H "$H" | jq '[.threads[].id]'
[
  "e8b1eba1-bea9-462f-920f-062be0d57dcc"
]
workspace thread 05ac605f... absent: True

$ curl -s -X DELETE "$BASE/api/workspaces/demo-team/members/$OWNER_ID" -H "$H" → HTTP status
409

$ curl -s -X POST "$BASE/api/workspaces/demo-team/archive" -H "$H" → HTTP status
204

$ curl -s "$BASE/api/workspaces/demo-team" -H "$H" → HTTP status
410
```

Original-PR: #1734
Original-Author: @serrrfirat


henrypark133 (changes requested):
## Review: workspace settings read path leaks admin-only config

The workspace-entity direction is solid, but one admin-boundary regression remains in the new workspace-scoped settings API.

### Concerning: workspace members can read admin-only LLM settings
**File:** `src/channels/web/handlers/settings.rs:68` and `src/channels/web/handlers/settings.rs:125`
The new read paths resolve the workspace scope and then return whatever rows exist for that workspace, but they never apply the matching read-side filter for `ADMIN_ONLY_LLM_SETTING_KEYS`. A non-admin workspace member can call `/api/settings?workspace=<slug>` or `/api/settings/<key>?workspace=<slug>` and read keys like `openai_compatible_base_url`, `ollama_base_url`, `llm_builtin_overrides`, or `llm_custom_providers` that were set by an owner/admin. The API-key fields are masked, but the endpoint URLs and provider metadata are still exposed, which breaks the admin-only boundary the config layer is explicitly trying to preserve.

**Suggested fix:** strip admin-only LLM keys on the settings read path for non-admin users, and add a caller-level workspace API test proving a member cannot read those keys.

Summary:
- Prior feedback status: no prior blocker on this path from the existing thread audit
- Residual risk: this PR is very broad, so I focused on the multi-user workspace/settings boundary and did not exhaustively re-walk every touched subsystem in one pass.

standardtoaster:
@henrypark133 friendly ping

src/db/postgres.rs
```diff
@@ -1390,6 +1546,462 @@ fn row_to_identity(row: &tokio_postgres::Row) -> UserIdentityRecord {
...
+    ) -> Result<bool, DatabaseError> {
+        let mut conn = self
+            .pool()
+            .get()
+            .await
+            .map_err(|e| DatabaseError::Pool(format!("Failed to get client: {e}")))?;
+        let tx = conn.transaction().await?;
+
+        if new_role != "owner" {
+            pg_check_not_last_owner(&tx, &workspace_id, user_id).await?;
```
serrrfirat, line 1974: **High Severity** — the last-owner guard is still race-prone on PostgreSQL.

`pg_check_not_last_owner()` only does a read of the current owner count inside a normal transaction. Under `READ COMMITTED`, two concurrent demote/remove requests against two different owners can both observe `owner_count = 2`, both pass this check, and both commit — leaving the workspace with **zero owners**.

This needs a lock-based invariant at the DB layer (for example locking the owner rows, using a workspace-scoped advisory lock, or running this section under `SERIALIZABLE`) rather than a read-then-write check.

src/channels/web/responses_api.rs
```diff
@@ -424,6 +426,63 @@ async fn send_to_agent(state: &GatewayState, msg: IncomingMessage) -> Result<(),
...
+        .map_err(|(status, message)| api_error(status, message, "invalid_request_error"))
+}
+
+async fn resolve_thread_workspace_id(
+    store: &Arc<dyn Database>,
+    user_id: &str,
+    thread_uuid: Uuid,
+) -> Result<Option<Uuid>, ApiError> {
+    let workspace_id = store
+        .get_conversation_workspace_id(thread_uuid)
```
serrrfirat, line 459: **High Severity** — archived workspaces can still be reached through the Responses API.

This helper verifies ownership, but it never re-checks the workspace's status. If a client omits `?workspace=` and continues/fetches a response by `previous_response_id` or `/v1/responses/{id}`, the code will adopt the thread's `workspace_id` here and proceed even if that workspace has already been archived.

I think this needs the same archived-workspace check that `resolve_workspace_scope()` applies in the normal REST handlers.

src/channels/web/handlers/workspaces.rs
```diff
@@ -0,0 +1,463 @@
...
+    let store = state.store.as_ref().ok_or((
+        StatusCode::SERVICE_UNAVAILABLE,
+        "Database not available".to_string(),
+    ))?;
+    let resolved = resolve_workspace_scope(store, &user, Some(&slug)).await?;
+    let resolved = resolved.ok_or((StatusCode::NOT_FOUND, "Workspace not found".to_string()))?;
+    require_workspace_owner(&resolved.role)?;
+
+    let archived = store
+        .archive_workspace(resolved.workspace.id)
```
serrrfirat, line 344: **High Severity** — archiving is only enforced at the HTTP boundary; background routines still appear runnable.

This endpoint correctly marks the workspace archived, but the lower-level routine selectors used by `RoutineEngine` (`list_event_routines`, `get_webhook_routine_by_path`, `list_due_cron_routines`, plus the libSQL equivalents) still query `routines` directly without excluding `workspaces.status = 'archived'`.

That means an archived workspace can still keep firing cron/event/webhook routines and spend tokens after this returns 204. I think the archived invariant needs to move down into those selectors / engine-facing lookups as well.

src/channels/web/handlers/workspaces.rs
```diff
@@ -0,0 +1,463 @@
...
+    let name = body.name.trim();
+    let slug = body.slug.trim();
+    validate_workspace_name(name)?;
+    validate_workspace_slug(slug)?;
+    validate_workspace_settings(&body.settings)?;
+
+    let workspace = store
+        .create_workspace(name, slug, &body.description, &user.user_id, &body.settings)
+        .await
+        .map_err(internal_db_error)?;
```
serrrfirat: **Medium Severity** — duplicate workspace slugs will currently come back as a 500.

A unique-constraint failure on `workspaces.slug` will bubble through `internal_db_error()` here, so a normal client conflict (`slug` already taken) is surfaced as `500 Internal Server Error` instead of `409 Conflict`.

Can we map uniqueness / `DatabaseError::Constraint` here to a stable conflict response? Otherwise clients can't distinguish a retryable server failure from a validation conflict.

src/db/postgres.rs
standardtoaster: Fixed in c84baf92. `FOR UPDATE` on the workspace in `pg_check_not_last_owner`, `BEGIN IMMEDIATE` on libSQL. Regression in `workspace_last_owner_concurrency.rs` runs concurrent demote/remove under `tokio::join!` with a barrier; fails without the fix.

src/channels/web/responses_api.rs
standardtoaster: Fixed in 42ca981e. Extracted a `require_workspace_active` helper; `resolve_thread_workspace_id` calls it after the ownership check, so `previous_response_id` and `GET /v1/responses/{id}` both 410 on archived workspaces. Test: `responses_api_fetch_rejects_archived_workspace` — 200 before, 410 after.

src/channels/web/handlers/workspaces.rs
standardtoaster: Fixed in 2500ced6. Added `AND (workspace_id IS NULL OR workspace_id NOT IN (SELECT id FROM workspaces WHERE status='archived'))` to `list_event_routines`, `list_due_cron_routines`, `get_webhook_routine_by_path` in both backends. Personal routines (NULL workspace_id) still fire. Regression: `routine_selectors_exclude_archived.rs`.

Left `list_routines_for_workspace` alone — caller already rejects archived. Shout if you want defensive consistency there too.

src/channels/web/handlers/workspaces.rs
standardtoaster: Fixed in 7ec0812f. Added `DatabaseError::is_{unique,foreign_key,check,not_null}_violation` predicates (SqlState on pg, extended SQLite codes on libSQL) and `db_error_to_status` mapping unique→409, FK→422, check→422, not-null→400. Applied to `workspaces_create_handler` and `users_create_handler` (latter was doing `"unique"`/`"duplicate"`/`"already exists"` substring matching).

libSQL knock-on: `create_workspace` and `create_user_with_token` propagate `libsql::Error` via `?` instead of wrapping as `Query(e.to_string())` — classifier needs the SQLite code.

Regression: `workspaces_create_returns_409_on_duplicate_slug`, `test_user_creation_returns_409_on_duplicate_email`.

standardtoaster:
@henrypark133 — fixed in 67f49273 (Apr 17). Threaded the workspace role through `resolve_workspace_scope`; list handler strips `ADMIN_ONLY_LLM_SETTING_KEYS` for non-admin members, get handler 404s on those keys. Test: `test_workspace_member_cannot_read_admin_only_settings`.

Audit off the back of this found the same pattern on `llm_providers_handler` — `env_base_url`, `env_model`, `has_api_key` exposed to every authenticated user. Fixed in 1657ef2d.

standardtoaster:
Review feedback addressed:

- c84baf92 — last-owner race (pg `FOR UPDATE` + libSQL `BEGIN IMMEDIATE`)
- 42ca981e — Responses-API archived bypass
- 2500ced6 — routine selectors exclude archived workspaces
- 7ec0812f — `DatabaseError` predicates + `db_error_to_status`; slug/email conflicts 409
- 1657ef2d — `llm_providers_handler` env overlay admin-only (same class as settings leak)
- 67f49273 — @henrypark133's settings filter (Apr 17)

Regression tests on each, clippy clean.

@henrypark133 @serrrfirat — flagging that rebasing through #2599 is effectively re-implementing this against the new layout. That's a second full review for a layout change on a second-pass PR. Rebase now or at land time?

src/channels/web/handlers/routines.rs
```diff
@@ -214,6 +260,23 @@ pub async fn routines_trigger_handler(
...
+
+    // Viewers cannot trigger routines — require at least member role
+    if routine.workspace_id.is_some()
+        && let Some(resolved) =
+            resolve_workspace_scope(store, &user, workspace_query.workspace.as_deref()).await?
+        && resolved.role == "viewer"
+    {
+        return Err((
+            StatusCode::FORBIDDEN,
+            "Workspace viewer cannot trigger routines".to_string(),
```
serrrfirat, line 277: **High — trigger is broken for every workspace routine not created by the caller**

`load_visible_routine` above already correctly enforces workspace visibility, and `fire_manual` in `routine_engine.rs:777-792` also re-checks ownership + falls back to `is_workspace_member` for workspace routines. But a few lines below this viewer check the handler then does:

```rust
let routine = store.get_routine(routine_id).await ...;
if !routine.is_owned_by(&user.user_id) {
    return Err((StatusCode::NOT_FOUND, "Routine not found".to_string()));
}
```

That rejects every workspace member who is *not* the original creator — including other workspace admins and the workspace owner. So only the routine's creator can trigger a workspace routine, which defeats the cross-workspace sharing feature. It also re-fetches the routine from the DB for no reason (`load_visible_routine` already returned it).

Suggested fix: delete the redundant re-fetch + `is_owned_by` block entirely. `load_visible_routine` + this viewer gate + `fire_manual`'s own auth are sufficient.

Coverage gap: no handler test exercises this path with `user_id = Some(...)`. All `fire_manual` calls in `tests/e2e_routine_heartbeat.rs` pass `None`, so the bug is invisible to the existing suite. See `CLAUDE.md` → "Test through the caller, not just the helper."

src/channels/web/handlers/routines.rs
```diff
@@ -260,15 +324,7 @@ pub async fn routines_toggle_handler(
...
-    let mut routine = store
-        .get_routine(routine_id)
-        .await
-        .map_err(|e| (StatusCode::INTERNAL_SERVER_ERROR, e.to_string()))?
-        .ok_or((StatusCode::NOT_FOUND, "Routine not found".to_string()))?;
-
-    if !routine.is_owned_by(&user.user_id) {
-        return Err((StatusCode::NOT_FOUND, "Routine not found".to_string()));
-    }
+    let mut routine = load_visible_routine(store, &user, &workspace_query, routine_id).await?;
```
serrrfirat, line 328: **High — workspace viewers can enable/disable any workspace routine**

`load_visible_routine` gates only on workspace membership, not role. In `routines_trigger_handler` there is an explicit `resolved.role == "viewer"` guard before firing; that guard is intentionally missing here, so viewers can freely toggle routine state.

The same applies to `routines_delete_handler` below — a viewer can delete any workspace routine. Toggle/enable/disable is a mutation and should match trigger's viewer gate (or be stricter — probably admin+ for enabling event/system-event routines since they fire autonomously).

Suggest extracting a `require_workspace_mutator(store, user, workspace_query)` helper and calling it from trigger, toggle, delete, and run-creation endpoints so they can't drift again.

src/channels/web/handlers/routines.rs
```diff
@@ -327,15 +384,7 @@ pub async fn routines_delete_handler(
...
-    let routine = store
-        .get_routine(routine_id)
-        .await
-        .map_err(|e| (StatusCode::INTERNAL_SERVER_ERROR, e.to_string()))?
-        .ok_or((StatusCode::NOT_FOUND, "Routine not found".to_string()))?;
-
-    if !routine.is_owned_by(&user.user_id) {
-        return Err((StatusCode::NOT_FOUND, "Routine not found".to_string()));
-    }
+    let _routine = load_visible_routine(store, &user, &workspace_query, routine_id).await?;
```
serrrfirat, line 388: **High — workspace viewers can delete any workspace routine**

Same class of bug as the toggle handler: `load_visible_routine` lets any workspace member through (incl. `viewer`), and delete is unconditional after that. The comment `// Verify ownership before deleting.` is now misleading — `load_visible_routine` verifies *visibility*, not ownership or mutator role.

At minimum: match the trigger handler's viewer-reject. Arguably: require `member`/`admin`/`owner` and reject non-creator deletes unless caller is `admin`/`owner` of the workspace.

src/channels/web/handlers/memory.rs
```diff
@@ -139,9 +160,10 @@ pub async fn memory_read_handler(
 pub async fn memory_write_handler(
     State(state): State<Arc<GatewayState>>,
     AuthenticatedUser(user): AuthenticatedUser,
+    Query(workspace_query): Query<WorkspaceQuery>,
     Json(req): Json<MemoryWriteRequest>,
 ) -> Result<Json<MemoryWriteResponse>, (StatusCode, String)> {
-    let workspace = resolve_workspace(&state, &user).await?;
+    let workspace = resolve_workspace(&state, &user, Some(&workspace_query)).await?;
```
serrrfirat, line 166: **High — workspace viewers can write to workspace memory**

`resolve_workspace` only verifies membership, not role. That means a viewer-role member can `POST /api/memory` with `?workspace=<slug>` and write or append arbitrary content to the workspace's memory documents, including identity files (`IDENTITY.md`, `SOUL.md`, etc.) that the LLM tool layer explicitly blocks.

The existing comment ("Identity-file protection is enforced at the tool layer") assumed the HTTP API was an "authenticated admin interface" — that assumption no longer holds once viewers can call it for a workspace they're added to. Either:

1. Add a role check here (reject viewer; possibly reject member too for identity files).
2. Port the tool-layer identity-file block to this handler.

Same applies elsewhere that routes through `resolve_workspace` — anything under workspace scope needs role-aware gating for mutations.

src/channels/web/handlers/jobs.rs
```diff
@@ -328,16 +385,18 @@ pub async fn jobs_detail_handler(
 pub async fn jobs_cancel_handler(
     State(state): State<Arc<GatewayState>>,
     AuthenticatedUser(user): AuthenticatedUser,
+    Query(workspace_query): Query<WorkspaceQuery>,
     Path(id): Path<String>,
 ) -> Result<Json<serde_json::Value>, (StatusCode, String)> {
     let job_id = Uuid::parse_str(&id)
         .map_err(|_| (StatusCode::BAD_REQUEST, "Invalid job ID".to_string()))?;
+    let scope = resolve_job_scope(&state, &user, &workspace_query).await?;
```
serrrfirat, line 393: **High — workspace viewers can cancel / restart / inject prompts into any workspace job**

`resolve_job_scope` + `sandbox_job_visible` / `agent_job_visible` check scope and ownership-or-scope-match only. There is no role gate on the mutation path, so a viewer-role member:

- `POST /api/jobs/{id}/cancel` — kills any running workspace job,
- `POST /api/jobs/{id}/restart` — re-enqueues any failed/interrupted job,
- `POST /api/jobs/{id}/prompt` — injects an LLM prompt into a workspace agent's live stream.

All three are significant side effects: cancellation can interrupt a workspace owner's long-running job, restart re-burns tokens/cost on the workspace's quota, and `jobs_prompt` directly influences agent behavior. Even a `member` arguably shouldn't be allowed to do these to another user's workspace job unless they're `admin`/`owner` of the workspace.

Recommend a shared helper that returns a typed `WorkspaceRole` enum from `resolve_workspace_scope` and reject `Viewer` on all three handlers. Also add multi-tenant tests for each (the current `src/channels/web/tests/multi_tenant.rs` doesn't cover these flows).

src/channels/web/handlers/settings.rs
```diff
@@ -148,36 +210,52 @@ pub async fn settings_set_handler(
     Query(query): Query<SettingScopeQuery>,
     Json(body): Json<SettingWriteRequest>,
 ) -> Result<StatusCode, StatusCode> {
```
serrrfirat, line 212: **Medium — workspace owners can read but not write their own workspace's admin-only LLM settings**

Read path (`can_read_admin_keys()` at line 50) allows `owner | admin` workspace roles. But the write path still goes through `ensure_setting_write_allowed` (line 491 in the new file), which only checks the global `user.role == "admin"`:

```rust
if is_admin_only_setting_key(key) && user.role != "admin" {
    return Err(StatusCode::FORBIDDEN);
}
```

So a workspace owner who is a normal global user can see `llm_builtin_overrides` / `llm_custom_providers` / `ollama_base_url` etc. for their workspace but cannot change them — they have to call a global admin to configure LLM for their own workspace. That's surprising given how the read side was designed.

Fix: teach `ensure_setting_write_allowed` about the resolved workspace scope (`owner | admin` passes for workspace-scoped writes; global `admin` passes for personal-scope writes). Otherwise either tighten the read side to match or accept this and document it as a global-admin-only config surface.

serrrfirat:
## Paranoid review summary — PR #2548

Thanks for the thorough workspace-scoping work. The `resolve_workspace_scope` resolver, the last-owner serialization (`pg_check_not_last_owner` + libSQL `BEGIN IMMEDIATE`), the archived-workspace block on `previous_response_id`, and the SSE event scoping all look solid.

Four findings posted as inline comments, plus one cross-cutting coverage concern below.

### #6 — Coverage gap: no handler-level tests for role-aware mutations (Medium)

Ironclaw's `CLAUDE.md` has an explicit rule: **"Test through the caller, not just the helper."** Several of the High findings I posted would have been caught by handler-level tests but are invisible to the current suite:

| Handler | Test? |
|---|---|
| `routines_trigger_handler` (non-creator workspace member) | ❌ no test |
| `routines_toggle_handler` (viewer) | ❌ no test |
| `routines_delete_handler` (viewer) | ❌ no test |
| `memory_write_handler` (viewer) | ❌ no test |
| `jobs_cancel_handler` (viewer) | ❌ no test |
| `jobs_restart_handler` (viewer) | ❌ no test |
| `jobs_prompt_handler` (viewer) | ❌ no test |
| `settings_set_handler` (workspace owner writing admin-only key) | ❌ no test |

`src/channels/web/tests/multi_tenant.rs` covers cross-user isolation well but never drives any of these handlers with a second workspace member. `tests/workspaces_integration.rs` exercises the store, not the HTTP surface.

The routine-trigger bug (#1) is especially illustrative: `fire_manual` in `routine_engine.rs` was updated to correctly allow workspace members, and there are tests for that engine method — but they all pass `user_id = None`, which bypasses the ownership branch entirely. The handler's own `is_owned_by` check (which rejects every non-creator member) is unreachable in tests.

### Other notes, not worth inlining

- **libSQL settings cache** bypasses workspace-scoped reads entirely (`src/db/cached_settings.rs:233-303` — "delegate directly to inner store (no caching)"). Correct but worth knowing for perf.
- **Frontend has no workspace selector** (`crates/ironclaw_gateway/static/app.js`) — this PR is backend-only so every workspace path is driven externally for now. Fine, just surfacing.
- **v2 engine untouched** — no `crates/ironclaw_engine/` changes. Any workspace-scoped chat going through v2 will currently ignore `workspace_id`. Likely a follow-up.
- **Migration parity** between PG `V25__workspaces.sql` and libSQL `libsql_migrations.rs` looks correct (workspace tables, columns, partial unique indexes all match).

I can refresh the review once the High findings are addressed — happy to add the missing handler tests if useful.

PR #2549, opened 2026-04-16
feat(engine): add mission_get action for retrieving mission results

serrrfirat:
## Summary

- **Add `mission_get` action** so the LLM can retrieve mission results, status, approach history, and recent thread outputs
- **Map `routine_history` → `mission_get`** so the v1 alias also works in engine v2
- **Root cause**: when users asked "what is the result of the research?", the agent fell back to `list_jobs` because no mission results tool existed — missions spawn threads (not jobs), so `list_jobs` always returned empty

## Changes

| File | Change |
|------|--------|
| `src/bridge/router.rs` | Register `mission_get` ActionDef in capability registry |
| `src/bridge/effect_adapter.rs` | Handler: loads mission + last 5 threads + extracts assistant responses; alias: `routine_history` → `mission_get`; regression test |
| `crates/ironclaw_engine/src/runtime/mission.rs` | Add `pub fn store()` accessor on MissionManager |

## Test plan

- [x] `cargo test -- routine_history_maps` — alias test passes
- [ ] Manual: create a mission, fire it, then ask "what is the result?" — should call `mission_get` instead of `list_jobs`


serrrfirat:
Addressed review feedback:

1. **Security (high)** — Added ownership check: `mission.user_id != context.user_id` now returns an error, preventing cross-user IDOR. Matches the guards on `mission_fire`/`mission_pause`/`mission_resume`.

2. **Medium** — Capped `approach_history` to last 10 entries via `.iter().rev().take(10).rev()` to prevent context window overflow.

Both fixes in commit 021d5c67.

henrypark133 (dismissed):
## Review: ownership fix and response shaping look sound

I re-checked the current head in `src/bridge/effect_adapter.rs:344-382`. The earlier ownership concern is addressed: `mission_get` now rejects missions whose `mission.user_id` does not match the caller, and the response also caps `approach_history` to the most recent 10 entries instead of returning the full unbounded list.

Prior feedback status: the previously raised ownership issue appears addressed in the current head.

Residual risk: I did not wait for the full local compile to finish for a targeted alias test in this pass, so approval here is based on code inspection rather than completed local test execution.

PR #2555, opened 2026-04-16
Fix gateway tool output visibility and timing

henrypark133:
## Summary
- fix gateway tool activity cards to correlate live tool events by call_id and show actual tool output when available
- use the existing persisted result field for expanded history cards without adding new history storage, and bound active-thread history results to keep refreshes fast
- thread call_id and live duration_ms through the web event surface, preserve engine_v2 action call IDs in the bridge, and replace misleading 0.0s with millisecond-friendly duration formatting

## Testing
- cargo fmt --all
- cargo clippy --workspace --all-targets --all-features -- -D warnings
- cargo test gateway_send_status_preserves_tool_event_fields --lib
- cargo test test_build_turns_with_persisted_tool_result_for_display --lib
- cargo test accumulator_tool_flow --lib
- cargo test test_ws_multiple_events_in_sequence --test ws_gateway_integration
- cargo test test_tool_result_for_display_truncates_long_content --lib
- cargo test forward_event_to_channel_preserves_call_id_for_action_events --lib
- cargo test thread_event_to_app_events_preserves_call_id_for_action_events --lib


henrypark133:
Addressed the open review comments on `tool-visibility`:

- `tool_result_for_display()` now returns `None` for JSON `null` and empty strings, so the UI no longer renders synthetic `"null"` output.
- `createToolActivitySummary()` now builds the duration node with `textContent` instead of appending via `innerHTML`, so `<1ms` renders safely.
- `findRendered()` now treats `call_id` as authoritative and no longer falls back to name-based matching when a `call_id` is present but does not satisfy the predicate.

Validation after the follow-up patch:
- `cargo fmt --all`
- `node --check crates/ironclaw_gateway/static/app.js`
- `cargo test test_tool_result_for_display_skips_null --lib`

Follow-up commit pushed: `733f5b95`.

henrypark133:
@codex review

henrypark133 (commented):
## Tooling smoke test

This is a live smoke test of `pr-feedback-audit/scripts/post_pr_feedback.py`.

It only validates top-level review posting and inline comment anchoring on PR #2555. It is not a code-review signal and should be ignored for merge decisions.

src/channels/web/tests/tool_event_passthrough.rs
```diff
@@ -0,0 +1,129 @@
+//! Regression: gateway `send_status()` must preserve per-tool identity fields
```
henrypark133, line 1: Tooling smoke test: validating inline comment anchoring from `post_pr_feedback.py`. No action needed.

crates/ironclaw_gateway/static/app.js
```diff
@@ -2113,164 +2161,328 @@ function addToolCard(name) {
...
+    const rendered = findRendered(
+      event.call_id || null,
+      event.name || '',
+      candidate => !candidate.entry.result
+    ) || findRendered(event.call_id || null, event.name || '');
+    if (!rendered) return;
+
+    rendered.entry.result = event.preview || '';
+    if (!rendered.entry.result_preview) {
+      rendered.entry.result_preview = event.preview || '';
```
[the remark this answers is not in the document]
henrypark133: Addressed in `a10f5751`.

Live tool-result cards now reapply a 1000-character cap before storing or rendering `event.preview`, so the active activity path no longer keeps unbounded tool output in DOM/state.

Validation: `node --check crates/ironclaw_gateway/static/app.js`

src/channels/web/responses_api.rs
```diff
@@ -929,9 +940,11 @@ async fn streaming_worker(
                 );
                 acc.text_chunks.push(content.clone());
             }
-            AppEvent::ToolStarted { name, .. } => {
+            AppEvent::ToolStarted { name, call_id, .. } => {
                 let idx = acc.output.len();
-                let call_id = format!("call_{}", Uuid::new_v4().simple());
+                let call_id = call_id
+                    .clone()
+                    .unwrap_or_else(|| format!("call_{}", Uuid::new_v4().simple()));
```
henrypark133, line 991: This starts threading `call_id` through the streaming worker, but the completion side still uses a single global `current_tool_index` later in the same match arm. With overlapping tool calls, a later `ToolStarted` overwrites that slot, so the first completion can emit `response.output_item.done` for the wrong function-call item. Since the PR already has the `call_id` here, the in-flight output index should be tracked by `call_id` as well instead of by one mutable slot.

henrypark133 (commented):
## Review: keep Responses API tool completions correlated per call

Most of the earlier tool-visibility feedback looks addressed, and the new `call_id` / `duration_ms` plumbing is much cleaner. One correctness issue still remains in the streaming Responses API path.

### Critical: `response.output_item.done` still uses one global tool slot
**File:** `src/channels/web/responses_api.rs:947`
The streaming worker now threads `call_id` into the `FunctionCall` and `FunctionCallOutput` items, but it still tracks completion with a single `current_tool_index`. If tool A starts, tool B starts, and tool A completes first, the later start overwrites that slot and the code emits `response.output_item.done` for B instead of A. The final output list keeps the right `call_id`, but streamed clients can observe the wrong item transition to done and never get a done event for the actual completed call.

**Suggested fix:** track in-flight output indexes by `call_id` (or look them up by `call_id` on completion) instead of using a single mutable index.

Recommended verdict: Request changes.

Residual risk: I also kicked off a couple of targeted tests from the worktree, but they were still compiling when I posted this review.

src/bridge/router.rs
```diff
@@ -3499,7 +3502,8 @@ async fn forward_event_to_channel(
                         success: false,
```
serrrfirat, line 3556: **Medium Severity** — V2 ActionFailed events lose duration

When the bridge forwards `EventKind::ActionFailed` events, `ToolCompleted` gets `duration_ms: None` because `ActionFailed` has no `duration_ms` field in the engine crate.

Meanwhile, v1 code (`dispatcher.rs`, `thread_ops.rs`) measures wall-clock duration for **all** tool calls including failures:
```rust
let started_at = std::time::Instant::now();
let result = self.agent.execute_chat_tool(...).await;
let duration_ms = started_at.elapsed().as_millis() as u64;
```

This creates inconsistent UX: failed v1 tool cards show timing, but failed v2 tool cards do not.

**Suggestion:** Consider adding `duration_ms: u64` to `EventKind::ActionFailed` in `crates/ironclaw_engine/src/types/event.rs` (the executor already has the start time when the action begins). Then thread it through here the same way `ActionExecuted` does.

src/channels/web/util.rs
```diff
@@ -56,17 +57,51 @@ pub fn tool_error_for_display(error: &str) -> String {
...
+            ironclaw_safety::SafetyLayer::unwrap_tool_output(s).unwrap_or_else(|| s.clone())
+        }
+        other => other.to_string(),
+    };
+
+    if content.is_empty() {
+        return None;
+    }
+
+    Some(truncate_preview(&content, MAX_TOOL_RESULT_DISPLAY_CHARS))
```
serrrfirat: **Medium Severity** — New `result` field exposes more tool output in history API

`tool_result_for_display` sends up to 1000 chars of full tool output in `ToolCallInfo.result`, where previously only the shorter `result_preview` was available. Two observations:

1. For `Value::String` results, the function properly calls `SafetyLayer::unwrap_tool_output()` to strip the wrapper tags. Good.

2. For non-string JSON results (`Value::Object`, `Value::Array`), it falls through to `other.to_string()` **without** safety unwrapping:
```rust
let content = match result {
    serde_json::Value::String(s) => {
        ironclaw_safety::SafetyLayer::unwrap_tool_output(s).unwrap_or_else(|| s.clone())
    }
    other => other.to_string(), // no safety unwrap
};
```

If any tool returns structured JSON containing sensitive fields (API keys in response bodies, credentials in config dumps, etc.), those would be exposed in the history API response up to the 1000-char truncation limit.

Worth confirming this is intentional — the `result_preview` path may have the same exposure, but the new `result` field provides more data.

src/channels/web/util.rs
```diff
@@ -8,6 +8,7 @@ pub use ironclaw_common::truncate_preview;
 
 const MAX_HISTORY_IMAGE_DATA_URL_BYTES_PER_IMAGE: usize = 512 * 1024;
 const MAX_HISTORY_IMAGE_DATA_URL_BYTES_PER_RESPONSE: usize = 1024 * 1024;
+const MAX_TOOL_RESULT_DISPLAY_CHARS: usize = 1000;
```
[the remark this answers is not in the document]
henrypark133: Addressed on the current head.

`src/channels/web/util.rs` now uses `MAX_TOOL_RESULT_DISPLAY_BYTES` to match `truncate_preview(...)`'s byte-based limit, while keeping the same bounded display behavior. The nearby null/empty guards remain in place as well.

src/bridge/router.rs
henrypark133: Addressed in the follow-up commits now on this branch.

`EventKind::ActionFailed` now carries `duration_ms`, the v2 executors populate it for failed actions, and the bridge forwards it into failed `ToolCompleted` events so v2 matches the v1 UX. I also added `#[serde(default)]` on `ActionFailed.duration_ms` plus a regression test so older persisted events without that field still deserialize cleanly.

src/channels/web/util.rs
henrypark133: Addressed on the current head.

I tightened the history API so `ToolCallInfo.result` no longer surfaces a larger payload than the existing preview path. Both the persisted-history rebuild path and the in-memory history path now keep `result` aligned with the same preview-sized display string instead of exposing additional tool output beyond `result_preview`.

Validation:
- `cargo test -q --lib test_build_turns_result_matches_preview_when_available`
- `cargo test -q --lib test_build_turns_with_persisted_tool_result_matches_preview`

src/channels/web/util.rs
```diff
@@ -56,30 +57,48 @@ pub fn tool_error_for_display(error: &str) -> String {
...
+                call_id: c
+                    .get("tool_call_id")
+                    .or_else(|| c.get("call_id"))
+                    .and_then(|v| v.as_str())
+                    .map(String::from),
+                result: result_preview.clone(),
+                result_preview,
                 error: c["error"].as_str().map(tool_error_for_display),
                 rationale: c["rationale"].as_str().map(String::from),
             }
```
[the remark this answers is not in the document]
henrypark133: Addressed on the current head.

`parse_tool_call_infos()` now restores the split handling:
- `result_preview` comes from `c.get("result_preview").and_then(tool_result_for_display)`
- `result` comes from `c.get("result").and_then(tool_result_for_display)`
- `has_result` stays on the existing presence/null checks

That means wrapped persisted preview/result payloads are normalized through the display helper again, `result` stays `None` when only a preview exists, and a persisted full result can differ from the preview again.

Validation:
- `cargo test -q --lib test_build_turns_with_persisted_tool_result_for_display`
- `cargo test -q --lib test_build_turns_prefers_full_result_over_preview`
- `cargo test -q --lib test_build_turns_preview_only_does_not_populate_full_result`

ilblackdragon (approved):

ilblackdragon (approved):

ilblackdragon (approved):

ilblackdragon (approved):

ilblackdragon:
## PR Re-Review (post-fix, head `28e93667`)

Follow-up to my earlier review. The PR has been rebased on staging twice (pulling in the gateway platform refactor #2628 and CI-speedup #2566), and the one actionable review comment was addressed. CI is now fully green (Clippy, Formatting, cargo-deny, Gateway JS all passed).

### What changed since first review
- **`turn_info_from_in_memory_turn` asymmetry** — fixed in `src/channels/web/server.rs:2033-2053`. `result` is now populated from the full live value via `tool_result_preview`, `result_preview` is set to `None`, matching DB-path semantics where those fields carry distinct data. Accompanied by `test_in_memory_turn_info_populates_result_without_preview` covering both assertions. The code comment explains the invariant for future readers.
- Everything else in the earlier review still stands unchanged.

### Code quality
- **Backwards compat**: `#[serde(default)]` on new `duration_ms` + regression test `action_failed_defaults_missing_duration_ms_when_deserializing_legacy_payload` verifies legacy `ActionFailed` payloads in storage still deserialize.
- **Timing defensiveness**: every `execute_action` call site wraps with `Instant::now()` as a fallback when the effect adapter reports zero duration. Sensible.
- **Test coverage (Rust side)**: strong — boundary tests at deserialization, `forward_event_to_channel`, `thread_event_to_app_events`, SSE broadcast via `gateway_send_status_preserves_tool_event_fields`. Directly satisfies the "test through the caller" rule.
- **Style**: no `unwrap`/`expect` in production code, no new `info!` noise, follows existing formatting conventions.

### Remaining minor gaps (non-blocking)
- **`duration_ms: 0` for preflight errors** (`orchestrator.rs:933, 967, 1035`; `structured.rs:94, 129`): indistinguishable from a real 0ms execution. `Option<u64>` on the wire type would be cleaner. JS renders 0 as `<1ms` so user-visible impact is nil.
- **No JS test coverage** for the 400-line `createToolActivityController` rewrite. The e2e Playwright harness in `tests/e2e/` could cover the `call_id` correlation behavior in a follow-up.
- **UTF-8 truncation**: JS truncates at 1000 chars, Rust at 1000 bytes via `truncate_preview`. Multi-byte UTF-8 content shows slightly less in the browser than stored. Negligible.

### Security / correctness
- `.textContent` used for tool output (not `.innerHTML`) — no XSS vectors introduced.
- Wire-format change: `parameters: Some(format!("{duration_ms}ms"))` removed in favor of structured `duration_ms`. Any external SSE consumer parsing `parameters` for duration breaks, but `parameters` semantically means tool inputs — that misuse needed to end.
- No new database reads/writes, no secrets handling, no sandbox-relevant changes.

### Verdict
**Approve.** Full CI green, review feedback addressed with a dedicated regression test, merge with latest staging is conflict-free. The remaining minor gaps are out-of-scope follow-ups worth tracking separately if anyone cares.

PR #2563, opened 2026-04-17
Fix WASM channel startup restore semantics

henrypark133:
## Summary
- restore only startup-active WASM channels instead of treating every installed channel as active
- make startup use activated_channels as the authoritative source once persisted, with setup-selected channels as a first-run fallback
- keep extension list status accurate for installed but inactive WASM channels and document the fallback behavior

## Testing
- cargo fmt --all
- cargo test --lib load_startup_active_channels
- cargo test --lib register_startup_channels_only_restores_persisted_active_channels
- cargo test --lib test_extensions_list_handler_reports_installed_inactive_wasm_channel_as_inactive
- cargo test --lib normalize_extension_names_canonicalizes_deduplicates_and_skips_invalid

### DIFF
--- a/src/channels/wasm/setup.rs
+++ b/src/channels/wasm/setup.rs
@@ -65,6 +65,7 @@ pub async fn setup_wasm_channels(
     extension_manager: Option<&Arc<ExtensionManager>>,
     database: Option<&Arc<dyn Database>>,
     registered_channel_names: &[String],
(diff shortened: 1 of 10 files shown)


serrrfirat (approved):
Reviewed: clean semantics fix for WASM channel startup restore. Normalization is consistent across both read paths, fallback logic is sound, test coverage is solid.

PR #2570, opened 2026-04-17
fix(engine): allow completed event-driven missions to re-fire

serrrfirat:
## Summary

- **Root cause**: The `/expected` command returned "no self-improvement missions configured" because the expected-behavior learning mission's status was set to `Completed` after its first successful thread, and `fire_on_system_event()` only processed `Active` missions
- **Fix 1**: Allow `Completed` event-driven missions (`OnSystemEvent`, `OnEvent`, `Webhook`) to pass through the status filter in all three event-fire methods and the terminal check in `fire_mission()` -- each event is a fresh investigation, so completion of one thread should not prevent future fires
- **Fix 2**: Reset `threads_today` counter when `last_fire_at` is on a previous UTC day, preventing permanent daily budget exhaustion
- **Fix 3**: Call `ensure_learning_missions` in `handle_expected` before firing, so non-owner users or users whose missions failed to initialize get learning missions created on demand
- Adds `Mission::is_event_driven()` helper to classify cadence variants

## Test plan

- [x] 4 new regression tests added and passing:
  - `completed_event_driven_mission_can_fire` -- verifies a Completed OnSystemEvent mission still fires
  - `failed_event_driven_mission_cannot_fire` -- verifies Failed missions remain blocked
  - `threads_today_resets_on_new_day` -- verifies daily counter resets after UTC day boundary
  - `is_event_driven_classification` -- verifies cadence classification for all variants
- [x] All 406 engine tests pass (`cargo test -p ironclaw_engine`)
- [x] `cargo clippy -p ironclaw_engine --all-targets -- -D warnings` -- zero warnings
- [x] `cargo check` -- full crate compiles
- [ ] Manual test: use `/expected <description>` in web UI after a thread has completed -- should fire a self-improvement thread instead of showing "no missions configured"


henrypark133 (dismissed):
## Review: completed event-driven missions can re-fire

The current head correctly allows completed event-driven missions to fire on new events, resets `threads_today` across UTC day boundaries without blocking the current event on a transient save failure, and ensures `/expected` bootstraps per-user learning missions before firing.

PR #2182, opened 2026-04-09
fix(gateway): align historical/live tool call cards and preserve tool call correlation

hanakannzashi:
## Summary

- Make refreshed chat history render tool-use cards with the same card layout as the live “just used a tool” activity UI instead of a separate summary block.
- Normalize historical tool result previews before rendering by unwrapping stored `<tool_output>` wrappers and preferring the fuller persisted `result` field over the shorter `result_preview` when available.
- Preserve `call_id` through live tool events and use it to associate tool cards/results precisely, avoiding result content being attached to the wrong card when multiple tools run close together or share the same name.
- Apply the same `call_id` association fix to the Responses API conversion path so `FunctionCallOutput` stays linked to the correct prior `FunctionCall`; unknown/mismatched `call_id` values now fall back to the latest matching tool call instead of producing orphan outputs.
- This keeps refreshed history and live tool activity closer to the same experience: cleaner output text, consistent expansion behavior, correct result association, and fewer confusing truncated/wrapped previews.

## UI Before / After

Before:
before

After:
after

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] CI/Infrastructure
- [ ] Security
- [ ] Dependencies

## Linked Issue

None. Follow-up to manual UI debugging on this branch.

## Validation

- [ ] `cargo fmt --all -- --check`
- [ ] `cargo clippy --all --benches --tests --examples --all-features -- -D warnings`
- [x] `cargo build`
- [x] Relevant tests pass: `cargo test -p ironclaw_common event_type_matches_serde_type_field`; `cargo test --lib accumulator_`; `cargo test test_ws_multiple_events_in_sequence --test ws_gateway_integration`
- [ ] `cargo test --features integration` if database-backed or integration behavior changed
- [x] Manual testing: rebuilt and restarted the local gateway, then verified that refreshed history now uses the same tool card presentation path as live tool activity and no longer shows wrapped `<tool_output>` content in the historical card body
- [ ] If a coding agent was used and supports it, `review-pr` or `pr-shepherd --fix` was run before requesting review

Additional checks run:
- `cargo fmt --check`
- `cargo build --bin ironclaw`
- `CARGO_TARGET_DIR=/tmp/ironclaw-staging-merge-check cargo check --lib`

## Security Impact

[...]

PR #2571, opened 2026-04-17
feat(gateway): surface tool output previews and structured rendering

### LINKED ISSUE #2537: Web UI: Tool call and mission output needs structured rendering
## Classification: UX — Tool Call & Mission Output Formatting

Two related issues around how tool execution results are displayed in the chat UI.

---

### Finding 1: Tool calls don't surface AppEvents

**Observed:**  
Tool calls in the chat UI show minimal info — just the tool name, a green checkmark, and "0.0s" duration. Even if tools emit AppEvents with additional context (progress, content written, errors, metadata), none of it is displayed.

**Expected:**  
- If AppEvents exist for a tool call, surface them inline (expandable detail row under the tool call)
- Audit which tools already emit AppEvents and wire them to the UI
- Identify high-value tools that should emit AppEvents but don't (e.g. `memory_write` could show what was written, `mission_create` could show the mission summary)
- The tool call list should be informative, not just a checklist of names

---

### Finding 2: mission_create output is raw JSON dump

**Observed:**  
After `mission_create` succeeds, the UI shows:
1. Raw JSON: `{'mission_id': '...', 'name': '...', 'status': 'created'}`
2. Full mission list dumped as unformatted JSON array — including complete goal text for every mission (massive text walls)

**Expected:**  
- Created mission shown in a clean card: name, ID, status, project, cadence
- Mission list as a scannable table: name, status, cadence, thread count — NOT full goal text inline
- Goal text should be expandable/on-demand
- Render structured output similar to how plan tool results are displayed in the web UI

---

### LINKED ISSUE #2545: [Bug Bash 4/16] Tool calls execute but return empty results with no visible output
## Bug Report — QA Bug Bash 4/16

**Environment:** Staging (hosted-staging / crab shack) — https://agent-stg.near.ai/

**Description:**
Tool calls are shown as executed (with checkmarks and 0.0s timing), but the result message is completely empty. No data from the tool is displayed to the user. Observed with multiple tool types including memory_tree, memory_search(.db), and memory_tree(.system/engine).

**Steps to Reproduce:**
1. Open an agent on staging
2. 2. Trigger an action that requires tool usage
3. 3. Observe the tool call results
**Actual Result:**
Tool calls show as completed (green checkmarks, "Used 5 tools (0.0s)") but all return 0.0s execution time and produce no visible output. The 0.0s timing suggests the tools may not actually be executing.

**Expected Result:**
Tool calls should return and display their actual output or results in the message. Non-zero execution times should reflect actual processing.

**Severity:** High — Tools appear to run but produce no usable output
**Frequency:** Reproducible during bug bash testing

**Related:** #2408 (Context length overflow — may be related if tools fail silently due to context issues)

serrrfirat:
## Summary
- **Tool output previews**: Engine v2 tool calls now show output preview in expandable tool cards (was empty before). Added `output_preview` field to `ActionExecuted` events, emitting `ToolResult` from both the SSE and channel paths.
- **Duration fix**: Sub-millisecond tools show "< 1ms" instead of "0.0s" by parsing server-side `duration_ms` from the `parameters` field.
- **Structured JSON rendering**: Tool output that is JSON renders as HTML tables (for arrays) or key-value pairs (for objects) instead of raw text.
- **Mission created card**: New `AppEvent::MissionCreated` with dedicated frontend card (name, status badge, cadence, project) following the `PlanUpdate` pattern.

Closes #2537
Closes #2545

## Test plan
- [x] `cargo check` — compiles clean
- [x] `cargo test -p ironclaw_common` — 23 tests pass (includes new `MissionCreated` variant in `event_type_matches_serde_type_field`)
- [ ] Manual: run gateway with engine v2, execute tool calls, verify output previews appear in expandable cards
- [ ] Manual: verify sub-millisecond tools show "< 1ms" instead of "0.0s"
- [ ] Manual: call `mission_create`, verify styled card appears instead of raw JSON
- [ ] Manual: call `mission_list`, verify table rendering in tool output


serrrfirat:
## Review feedback addressed

All 4 comments from @gemini-code-assist resolved:

| # | Finding | Fix |
|---|---------|-----|
| 1-2 | XSS: `innerHTML` with `data.mission_id` / `data.cadence` / `data.project_id` in `renderMissionCreatedCard` | Replaced with DOM `textContent` (commit 08ba436) |
| 3 | `sendChangeResolution` uses raw `fetch` instead of `apiFetch` | Switched to `apiFetch` for consistent auth/OIDC handling (commit 4ab30d6) |
| 4 | Table cell `title` shows `[object Object]` for object values | Use `JSON.stringify` for object tooltips (commit 4ab30d6) |

Also added 7 unit tests for `truncate_output_preview` covering null, empty, long strings, emoji UTF-8 boundaries, and CJK characters.

henrypark133 (changes requested):
## Review: current head does not compile

The gateway/event UX additions are substantial, but the current head is blocked by a test that does not compile against the checked-in router test harness.

### Critical: the new router attachment test references stale test APIs and breaks `cargo clippy --all --all-features`
**File:** `src/bridge/router.rs:5862`
The added `handle_with_engine_persists_attachment_files_and_indexes_them` test uses symbols and fields that do not exist in this branch: `CWD_TEST_LOCK`, `CurrentDirGuard`, `EngineState.project_root`, and `IncomingAttachment.local_path` (see the same block through `src/bridge/router.rs:5887`). CI is failing on this exact test code, so the PR cannot merge in its current form.
**Suggested fix:** Rewrite the test against the current router test helpers and current `IncomingAttachment` / `EngineState` shapes, or drop the stale assertions from this PR.

serrrfirat:
Addressed review feedback:

- Removed the stale `handle_with_engine_persists_attachment_files_and_indexes_them` router test in `src/bridge/router.rs` that referenced APIs/types no longer present on this branch (`CWD_TEST_LOCK`, `CurrentDirGuard`, `EngineState.project_root`, `IncomingAttachment.local_path`).
- Re-verified the compile blocker is resolved with `CARGO_TARGET_DIR=/tmp/ironclaw-target cargo test --lib --no-run`.
- Re-ran `cargo clippy --all --benches --tests --examples --all-features` and `bash scripts/pre-commit-safety.sh`.

Note: `cargo test --lib` still reports unrelated failures in untouched areas outside this fix.

ilblackdragon:
# Code Review

## Overview

The PR description advertises four changes (tool-output previews, sub-ms duration, JSON table/KV rendering, mission-created card). The actual diff is ~1,237 LOC across 15 files and bundles several *undeclared* changes: a new right-sidebar **approval panel** (badge, panel, mobile overlay, keyboard shortcuts, ~300 lines of CSS), **change-proposal accept/reject flow**, a `feedback` field on `SkillActivated`, a **daily-thread counter reset fix**, and a **behavioral change to let Completed event-driven missions re-fire**.

## Scope / Structure

- **Half-wired feature.** The UI listens for `change_proposed`/`change_resolved` (`app.js:1615-1625`) and posts to `/api/chat/change/resolve`, but I don't see that endpoint or any `AppEvent::ChangeProposed` emitter in this diff. If the backend side is planned for a follow-up, drop the UI half until the backend is ready — otherwise the buttons 404 and the SSE listener is dead.

## Correctness

### `crates/ironclaw_engine/src/runtime/mission.rs` — daily-thread reset (`fire_one`)
- Good catch on the latent bug — the comment flagged a "reset daily by the cron ticker" that never existed.
- Silent `let _ = ...save_mission` is defensible given the comment, but there's no test. Add an integration-tier test covering the path `mission.threads_today == max && last_fire_at.date < today → fires`, per CLAUDE.md's "Test Through the Caller" rule.
- The UTC-day boundary means a mission capped at 5/day in a user's local evening will only reset around their mid-afternoon (depending on TZ). That may be acceptable, but calling it out explicitly in the comment would help.

### `mission.rs` — allow Completed event-driven missions to fire
Logic is duplicated between `fire_one` (`is_event_driven` check) and `tick_due` (extra `match` arm). Extract into `impl Mission { fn can_fire_when_terminal(&self) -> bool }` to keep the invariant in one place. Without that, the next cadence variant added will need two edits and likely diverge.

### `crates/ironclaw_engine/src/types/event.rs` — `truncate_output_preview`
- Implementation and UTF-8 tests are solid.
- `max_len` is measured in **bytes** (`s.len()`), not chars. The function is fine, but the *call sites* hardcode `500` four times. Extract a `const OUTPUT_PREVIEW_MAX_BYTES: usize = 500;` in one place.
- `other.to_string()` serializes non-string JSON via `Display` (compact `serde_json`). Worth a one-line comment that callers get compact JSON, not pretty-printed.

### `src/bridge/effect_adapter.rs` — `MissionCreated` broadcast
- `sse.broadcast_for_user(...)` is called without `.await`; verify it's a sync enqueue. If it returns a future, it's silently dropped.
- `status_str = "created" | "created_with_warnings"` stringly-typed. Consider an enum serialized via serde.

### `src/bridge/router.rs` — `handle_expected` learning-mission ensure
The `ensure_learning_missions(user_project_id, &message.user_id)` call is the right fix for non-owner users getting "no self-improvement missions configured." Two things:
- Failure is logged at `debug!` and then execution falls through to a mission-fire attempt that will also fail. Consider short-circuiting with an explicit user-facing error rather than a confusing "mission not found."
- This is a behavior change deserving a test at the router-handler tier; the diff adds none.

## JavaScript / UI

- **`renderToolOutputContent`** calls `JSON.parse` on every tool output even when unlikely to be JSON. Cheap, but worth short-circuiting on outputs whose first non-space char isn't `{` or `[`.
- **Table `textContent.length > 120` truncation** uses `.substring(0, 117) + '...'` — JS strings are UTF-16 code units. Emoji/surrogate-pair cells will occasionally get split on a surrogate boundary producing a lone surrogate. Use `Array.from(str).slice(0, 117).join('')` or equivalent.
- **Selector hardening (`#approval-panel-body .approval-card[...]`)** — the change from bare `.approval-card` to a scoped selector is correct, but any e2e test that selected `.approval-card` without the prefix will now miss. Please grep `tests/e2e/` and update.
- **`keydown` handler** with `y`/`n`/`a` shortcuts fires any time an approval panel is open even if focus is on a non-input element elsewhere; be aware it'll swallow those keys for anyone reading chat content with a keyboard.
- **30s fallback `setTimeout` in `showMissionProgress`** covers a "subscribe-after-spawn race" — the code comment is honest about this being a patch over a real bug. Add a follow-up issue to fix the race properly; fallback timers that mask races tend to outlive the race they were masking.
- **i18n parity**: `approvalPanel.*` keys are in en/ko/zh-CN but the project also ships ja and ru. Confirm those locales exist as JS files and add keys, or you'll break their pages.

## Test coverage

- `truncate_output_preview`: good unit coverage.
- `event_type_matches_serde_type_field`: updated for new variants — good.
- **Missing**:
  - No test for the mission day-reset path (budget exhausted yesterday → fires today).
  - No test for Completed event-driven missions firing again on new events.
  - No test for `ensure_learning_missions` in `handle_expected`.
  - No browser/e2e coverage of the approval panel or structured-JSON rendering.
  - Per CLAUDE.md: "unit test on the helper alone is not sufficient regression coverage" when a gate controls a side effect — the budget gate decides whether a mission fires (a side effect), so `fire_one` needs integration coverage.

## Security

- `/api/chat/change/resolve` POST — confirm it exists and validates `request_id` ownership (same user, same thread). Not visible in this diff.
- `renderJsonTable`/`renderJsonKeyValue` use `textContent` throughout (good — no XSS vector from user-controlled tool output).

## Minor

- `AppEvent::SkillActivated { feedback: Vec<String> }`: added everywhere as empty. If nothing actually populates it in this PR, it's dead shape. Either wire up a producer or hold off on shipping the field.

## Recommendation

Primary asks:
1. Either wire `change_proposed`/`change_resolved` end-to-end in this PR, or remove the UI listeners.
2. Add integration test coverage for the two mission-manager behavior changes.
3. Extract the `500` literal and the `can_fire_when_terminal` predicate.
4. Confirm `/api/chat/change/resolve` endpoint exists and authorizes by user.
5. Fix UTF-16 surrogate-pair truncation in `renderJsonTable`.
6. i18n parity for ja/ru.

ilblackdragon:
Merged `origin/staging` and folded in the review fixes. Single commit on top of the existing merge — full summary in the commit message.

## What changed vs the previous push

**Frontend (staging split `app.js`/`style.css` into per-surface modules in #2683, so the PR's additions had to be redistributed):**
- Structured tool-output rendering (`renderToolOutputContent`, `renderJsonTable`, `renderJsonKeyValue`) moved into `js/core/tool-activity.js`. Adds a `JSON.parse` short-circuit when the trimmed text doesn't start with `{` or `[`, and uses `Array.from` + slice for cell-text truncation so emoji/CJK never get split mid-character.
- `mission_created` SSE listener lives in `js/core/sse.js`; `renderMissionCreatedCard` lives in `js/core/onboarding.js`. Supporting CSS in `styles/surfaces/missions.css` and `styles/surfaces/activity.css`.

**Dropped from the PR (backing out the half-wired pieces I flagged in review):**
- `AppEvent::ChangeProposed` / `ChangeResolved` and the `change_proposed` / `change_resolved` UI listeners. Nothing emits these server-side in this PR, so keeping the wire types around is a contract with no producer. Better as a follow-up once the backend lands.
- The right-sidebar approval-panel redesign (badge, panel, mobile overlay, keyboard shortcuts, mission-progress card). It collides with staging's modular chat layout and isn't needed for the advertised scope. Staging still renders approvals inline from `.approval-card` in `chat.css`.

**Rust fixes:**
- `MissionCreated.status` is now a proper `MissionCreatedStatus` enum (`Created` / `CreatedWithWarnings`) per the typed-internals rule, instead of `String`.
- Extracted `OUTPUT_PREVIEW_MAX_BYTES = 500` const in `ironclaw_engine::types::event`; all four call sites (orchestrator ×2, scripting, structured) reference the const.
- Docstring for `truncate_output_preview` spells out that `max_len` is bytes and that non-string JSON is serialized via compact `Display`.
- `handle_expected` tracks whether `ensure_learning_missions` failed so the fallback message distinguishes "no missions configured" from "self-improvement temporarily unavailable" — without leaking internal error text per the error-boundary rule.
- `broadcast_for_user` confirmed sync (not awaited — by design).

**Tests:**
- Staging already has `completed_event_driven_mission_can_fire`, `failed_event_driven_mission_cannot_fire`, `threads_today_resets_on_new_day`, and `is_event_driven_classification` covering the PR's mission-manager behavior changes — no new tests required.
- Local `cargo test -p ironclaw_common --lib` (67 passed) and `cargo test -p ironclaw_engine --lib mission` (97 passed) green.
- `cargo check --all-features` and `cargo clippy --all --benches --tests --examples --all-features` clean.
- Four `extensions::manager::tests::install_from_local_source_*` failures on the full `cargo test --lib` are pre-existing and unrelated — they need the portfolio artifact to be built (you called this out earlier).

Ready for another look.

ilblackdragon:
**CI note:** GitHub blocked the full `pull_request` workflows (code_style, clippy, tests, replay-gate, regression-test-check, claude-review) on my push `314dacb5`. Only the four meta `pull_request_target` jobs (scope labels, classify, release-plz summary, staging promotion) fired — these show as green but don't exercise the actual build. This is standard behavior when a contributor other than the PR author pushes to the head branch: the workflows need an "Approve and run" click from @serrrfirat or a maintainer.

Local verification that I did run to confirm the push is clean:
- `cargo check --all-features` — passes
- `cargo fmt` — applied
- `cargo test -p ironclaw_common --lib` — 67/67 pass
- `cargo test -p ironclaw_engine --lib mission` — 97/97 pass

The pre-existing `extensions::manager::install_from_local_source_*` failures on the full `cargo test --lib` need the portfolio artifact built (they're unrelated to this PR, and the author already called that out earlier in this thread).

Once a maintainer approves the workflows, the real code_style / clippy / replay-gate / regression-test-check runs should confirm the green state.

serrrfirat:
Closing as superseded by #2555, #2452, and #2182.\n\nThe core fixes for tool output visibility/timing, engine v2 tool-call persistence, and live/history tool-card correlation have landed elsewhere. The remaining pieces in this PR are stale relative to current staging and are better handled as focused follow-up PRs if still needed.

PR #2572, opened 2026-04-17
fix: model provider config hardening

### LINKED ISSUE #2514: Issue: Incomplete Provider Configuration Causes Instance Startup Failure (Non-TEE)
**Description**  
On a production non-TEE instance (`mild-hawk-kidal`), configuring an LLM provider (e.g., Anthropic) via **Settings → Inference → Model Providers** without completing all required fields (e.g., API key) leads to an unrecoverable startup failure after instance restart. The instance enters a crash loop with authentication errors, even though the provider was not marked as "active" or fully configured.

**Steps to Reproduce**  
1. Navigate to **Settings → Inference → Model Providers**.
2. Click **Use** on an unconfigured provider (e.g., Anthropic - built-in).
3. Leave required fields (e.g., Base_URL、API key) empty or incomplete.
4. Restart the IronClaw instance.

Image

**Expected Behavior**  
- **Backend**: The instance should start successfully, either ignoring the incomplete provider or falling back to a valid default provider. Incomplete configurations should not block startup.
- **Frontend**: The **Use** button for a provider should be disabled (or show a validation tooltip) until all required fields (API key, base URL, etc.) are properly filled. Users should not be able to "activate" an incomplete provider.

**Actual Behavior**  
The instance enters a restart loop with the following error repeated 10 times before exiting:

Image

[mild-hawk-kidal_ERROR.log](https://github.com/user-attachments/files/26769016/mild-hawk-kidal_ERROR.log)

italic-jinxin:
## Summary

Validate provider config completeness before activation and harden backend security.

- Extract duplicated `isConfigured` check into shared `isProviderConfigured()` helper to prevent render/activate logic divergence
- Show "Not Configured" badge on unconfigured providers and hide the "Use" button until API key is set
- Add three activation guards in `setActiveProvider()`: missing API key, missing base URL (custom providers), and missing model — each shows a toast and auto-opens the config dialog
- Add `env_model` to the model restoration chain (override → env → default)
- Replace `#[derive(Debug)]` with manual `Debug` impl on `CustomLlmProviderSettings` and `LlmBuiltinOverride` to redact `api_key` from log output after secrets hydration
- Validate custom provider `base_url` against SSRF at save time (`validate_operator_base_url`) instead of deferring to `LlmConfig::resolve()`
- Change `test_connection` 400/422 response from `ok: true` to `ok: false` to avoid misleading green badge when model name or endpoint is wrong
- Downgrade `tracing::info!` to `debug!` in config resolution paths to avoid polluting REPL/TUI

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] CI/Infrastructure
- [ ] Security
- [ ] Dependencies

## Linked Issue

Fixes #2514 

## Validation

- [ ] `cargo build`
- [ ] Relevant tests pass: 
- [ ] Manual testing: 

## Review Follow-Through

---

**Review track**:


crates/ironclaw_gateway/static/app.js
```diff
@@ -9593,6 +9593,28 @@ function scrollToProviders() {
   if (section) section.scrollIntoView({ behavior: 'smooth', block: 'start' });
 }
 
+/** Check whether a provider has a usable API key (env, DB override, or vaulted secret). */
```
serrrfirat: **High Severity** — Custom providers incorrectly shown as "Not Configured"

`CustomLlmProviderSettings` has no `api_key_required` field. When the frontend loads custom providers, `provider.api_key_required` is `undefined`. In the new `isProviderConfigured()`:

```js
const needsKey = provider.api_key_required !== false;
// undefined !== false → true — ALL custom providers are treated as requiring a key
```

This means a custom Ollama provider (valid per `VALID_ADAPTERS`) without an API key will:
- Show the "Not Configured" badge
- Have the "Use" button hidden
- Trigger the "configure API key" toast if activation is attempted

**Suggested fix:** Either derive `needsKey` from the adapter type for custom providers, or have the backend include `api_key_required` in the custom provider data based on the adapter (`ollama` → `false`, others → `true`). For example:

```js
function isProviderConfigured(provider) {
  const needsKey = provider.api_key_required !== false
    && !(provider.adapter === 'ollama');  // ollama never needs a key
  // ...rest unchanged
}
```

Or normalize the shape server-side so custom providers always carry `api_key_required`.

src/channels/web/handlers/llm.rs
```diff
@@ -160,10 +160,16 @@ fn interpret_chat_response(
...
             {
-                // 400/422 = server reachable, likely wrong endpoint variant — connectivity OK
+                // 400/422 = server reachable but the request was rejected, likely a
+                // wrong model name or endpoint variant.  Report as not-ok so the UI
+                // doesn't mislead the user with a green badge.
                 TestConnectionResponse {
-                    ok: true,
-                    message: format!("Server reachable ({})", status),
+                    ok: false,
+                    message: format!(
```
serrrfirat: **Medium Severity** — Missing regression tests for behavioral changes

Two behavioral changes in this PR lack regression tests (per project policy: "Regression test with every bug fix"):

1. **`interpret_chat_response` 400/422 → `ok: false`**: This changes user-visible behavior (red badge instead of green). A unit test should construct a mock 400/422 response and assert `ok: false` + the updated message format. Currently no tests exist for this function.

2. **`validate_custom_providers` SSRF check**: The new `validate_operator_base_url` call at save time is a security gate. A test should pass a provider with an SSRF-unsafe `base_url` (e.g., `http://169.254.169.254/latest/meta-data`) and assert 422 rejection. Existing tests validate ID and adapter but not base_url.

think-in-universe:
As discussed today, @italic-jinxin will help fix the crash loop when a model is misconfigured in this PR. This is working in progress.

crates/ironclaw_gateway/static/app.js
```diff
@@ -10190,12 +10225,34 @@ function renderProviders() {
...
 function setActiveProvider(id) {
   const provider = [..._builtinProviders, ..._customProviders].find((p) => p.id === id);
+  if (provider) {
+    // Guard: do not activate a provider that requires an API key but has none configured
+    if (!isProviderConfigured(provider)) {
+      showToast(I18n.t('config.configureToUse'), 'error');
+      openProviderConfigDialog(provider);
+      return;
+    }
+    // Guard: custom providers must have a base URL
```
serrrfirat: This only enforces `base_url` for custom providers, but the registry also has built-ins where `base_url` is required (for example `openai_compatible` sets `"base_url_required": true` while also setting `"api_key_required": false`). With the new UI gating, those providers still show **Use** and can be activated even when the URL is blank, because `isProviderConfigured()` only reasons about API keys and `setActiveProvider()` skips the URL check for built-ins. That leaves the UI saying the switch succeeded even though runtime can only recover by falling back on restart. Can we make the readiness/activation check honor `base_url_required` for built-ins too?

think-in-universe (approved):
LGTM. The PR has been well tested end-to-end.

We're ready to merge.

PR #2574, opened 2026-04-17
fix: resolve staging CI test failures blocking promotion

serrrfirat:
## Summary

Staging CI has been failing for 7 days (30+ orphaned promotion PRs, 64 commits stuck). This fixes all 3 categories of test failures:

- **Extension name resolution** (`server.rs`): `pending_gate_extension_name` returned `credential_name` ("telegram_bot_token") instead of `extension_name` ("telegram") when `auth_manager` was unavailable. Added fallback extraction from install parameters.
- **CLI snapshots** (2 `.snap` files): Regenerated stale snapshots — new `profile` subcommand, updated help descriptions for `onboard`/`config`/`doctor`/`login`.
- **E2E pairing tests** (`test_extensions.py`): Relaxed exact dict equality to field check since frontend now sends optional `thread_id` in pairing approve bodies.

## Test plan

- [x] `pending_gate_extension_name_uses_install_parameters_for_post_install_auth` passes (was failing in all 3 test configs)
- [x] `cli::tests::test_help_output` and `test_long_help_output` pass (was failing in all-features)
- [x] `cargo fmt` and `cargo clippy` clean
- [ ] CI passes on this PR
- [ ] Remaining E2E auth-card failures should resolve once extension_name fix is in staging

### DIFF
--- a/src/bridge/auth_manager.rs
+++ b/src/bridge/auth_manager.rs
@@ -333,10 +333,15 @@ impl AuthManager {
         if matches!(
             action_name,
             "tool_install" | "tool-install" | "tool_activate" | "tool_auth"
-        ) && let Some(name) = parameters.get("name").and_then(|v| v.as_str())
-            && !name.trim().is_empty()
-        {
-            return name.to_string();
+        ) {
+            let trimmed = parameters
+                .get("name")
+                .and_then(|v| v.as_str())
+                .map(str::trim)
+                .unwrap_or("");
+            if !trimmed.is_empty() {
+                return trimmed.to_string();
+            }
         }
 
         if let Some(tools) = self.tools.as_ref()
--- a/src/channels/wasm/setup.rs
+++ b/src/channels/wasm/setup.rs
@@ -141,6 +141,7 @@ pub async fn setup_wasm_channels(
     })
 }
 
+#[allow(clippy::too_many_arguments)]
 async fn register_startup_channels(
     loaded_channels: Vec<LoadedChannel>,
     config: &Config,
--- a/src/channels/web/server.rs
+++ b/src/channels/web/server.rs
@@ -2896,26 +2896,9 @@ async fn pending_gate_extension_name(
         );
     }
 
-    // Mirror the install-parameter extraction from AuthManager for the
-    // fallback path when auth_manager is not wired up.
-    if matches!(
-        tool_name,
-        "tool_install" | "tool-install" | "tool_activate" | "tool_auth"
-    ) && let Some(name) = parsed_parameters
-        .get("name")
-        .and_then(|v| v.as_str())
-        .map(str::trim)
-        .filter(|s| !s.is_empty())
-    {
-        return Some(name.to_string());
-    }
-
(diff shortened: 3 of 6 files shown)


ilblackdragon (commented):
## Code Review

**PR**: fix: resolve staging CI test failures blocking promotion
**Verdict**: Approve with nit

### Extension name fallback (`server.rs`)

The new fallback block correctly mirrors `AuthManager::resolve_extension_name_for_auth_flow()` (lines 326-340). Both check the same tool names (`tool_install | tool-install | tool_activate | tool_auth`) and extract from `parameters.get("name")`. Good use of `let`-chains with `matches!` — idiomatic for this codebase.

**Nit — trimming inconsistency**: The fallback returns `name.to_string()` from the untrimmed `as_str()` value. The `.map(str::trim)` is only used in the filter predicate. The AuthManager has the same behavior, so this is consistent — but both callers would return `" telegram "` if the parameter had leading/trailing whitespace. Consider:

```rust
return name.trim().to_string(); // instead of name.to_string()
```

Low risk since parameter values are unlikely to have whitespace, and it matches the AuthManager's existing behavior.

### Test coverage

- Two good unit tests covering the install-parameter extraction and tool-registry fallback paths.
- The tests exercise the helper directly with `GatewayState`, which is appropriate since there's no wrapper dropping inputs between `engine_pending_gate_info` and `pending_gate_extension_name`.
- Minor gap: only `tool_install` is exercised — a parameterized test covering `tool-install`, `tool_activate`, and `tool_auth` would be more thorough, but the `matches!` logic is trivial so low risk.

### E2E test relaxation (`test_extensions.py`)

Switching from exact dict equality to field check is the right approach — forward-compatible with additional optional fields the frontend may add.

### Snapshots

Mechanical regeneration, looks correct.

Overall: well-scoped fix that correctly mirrors existing logic, has good test coverage, and unblocks the 7-day CI backlog. 👍

ilblackdragon (commented):
One more thought on the fallback duplication: rather than mirroring the AuthManager logic inline, could we ensure a `SecretsStore` is always available so `AuthManager` gets constructed? Looking at `build_gateway_auth_manager`, it returns `None` only when no secrets backend exists. If the test/CI configuration stood up a minimal in-memory or file-backed secrets store, `auth_manager` would always be `Some(...)` and this fallback path wouldn't be needed at all.

That would avoid the maintenance risk of the two implementations diverging if `resolve_extension_name_for_auth_flow` ever changes its logic.

serrrfirat:
Addressed both comments in 781980d6:

1. **Trim fix** — applied in `AuthManager::resolve_extension_name_for_auth_flow` (the canonical resolver), so trimmed values propagate everywhere. No more inconsistency between filter predicate and return value.

2. **Eliminated fallback duplication** — removed the inline mirror block from `pending_gate_extension_name`. Tests now wire up a minimal `AuthManager` backed by `InMemorySecretsStore`, so they exercise the real code path. Single source of truth for extension name resolution.

ilblackdragon (approved):

PR #2616, opened 2026-04-18
fix(mcp): avoid binding callback port while building auth url

henrypark133:
## Summary
- stop binding the shared OAuth callback port when auth_mcp_build_url only needs to construct a local redirect URI
- keep the real callback listener bind in the actual MCP OAuth callback-wait flow
- add a regression test that forces a non-local callback host and verifies auth URL generation still succeeds

## Why
auth_mcp_build_url could fail with Could not bind to callback port even when no callback listener was actually needed yet. That made MCP auth URL generation flaky under parallel test runs.

## Testing
- cargo test test_auth_mcp_build_url_uses_explicit_oauth_endpoints -- --nocapture
- cargo test test_auth_mcp_build_url_does_not_bind_callback_host_for_local_redirect -- --nocapture

### DIFF
--- a/src/extensions/manager.rs
+++ b/src/extensions/manager.rs
@@ -37,8 +37,8 @@ use crate::secrets::{CreateSecretParams, SecretsStore};
 use crate::tools::ToolRegistry;
 use crate::tools::mcp::McpClient;
 use crate::tools::mcp::auth::{
-    authorize_mcp_server, canonical_resource_uri, discover_full_oauth_metadata,
-    find_available_port, is_authenticated, register_client,
+    authorize_mcp_server, canonical_resource_uri, discover_full_oauth_metadata, is_authenticated,
+    register_client,
 };
 use crate::tools::mcp::config::{McpServerConfig, NEARAI_MCP_SERVER_NAME};
 use crate::tools::mcp::session::McpSessionManager;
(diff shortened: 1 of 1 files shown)


src/extensions/manager.rs
```diff
@@ -3541,14 +3541,11 @@ impl ExtensionManager {
...
-        // local mode binds a random port.
+        // local mode derives the fixed callback URL without binding.
         let redirect_uri = if let Some(uri) = self.gateway_callback_redirect_uri().await {
             uri
         } else {
-            let port = find_available_port()
-                .await
-                .map_err(|e| ExtensionError::AuthFailed(e.to_string()))?;
-            format!("http://localhost:{}/callback", port.1)
+            format!("{}/callback", crate::auth::oauth::callback_url())
```
serrrfirat, line 3548: **Medium Severity**\n\nThis local-mode redirect URI no longer matches the actual MCP callback listener when `IRONCLAW_OAUTH_CALLBACK_URL` is set to a loopback URL. `auth_mcp_build_url()` only treats **non-loopback** callback URLs as gateway mode, so values like `http://127.0.0.1:3001` still land in this branch. In that case `callback_url()` returns the full override, and this line produces `http://127.0.0.1:3001/callback` (or `.../oauth/callback/callback` if the env var already includes the callback path), but the real local MCP flow still binds `http://{callback_host()}:9876/callback` in `src/tools/mcp/auth.rs`. That means the generated auth URL can point at a callback endpoint the listener never serves.\n\nConcrete fix: for the local branch, derive the redirect from the same source as the listener (`callback_host()` + fixed port / shared helper), instead of appending `/callback` to `callback_url()`. A regression test with a loopback `IRONCLAW_OAUTH_CALLBACK_URL` would pin this.

serrrfirat:

This change avoids binding a callback port while constructing OAuth URLs in the v1 MCP client. Reborn MCP authentication is host-owned through product-auth flow records, hosted callbacks, typed continuations, and extension credential requirements; the old callback listener is not used.

If a current MCP auth flow still binds unexpectedly, please reproduce it through the Reborn hosted-MCP registration/auth path and fix the host product-auth composition, with a caller-level OAuth lifecycle test.

serrrfirat:

This fixes callback binding in the v1 MCP OAuth client. Reborn MCP auth uses host product-auth flow records, hosted callbacks, typed continuations, and extension credential requirements.

Reproduce any current issue through hosted-MCP registration/auth and fix the Reborn product-auth composition with a caller-level OAuth test.

PR #2630, opened 2026-04-18
Unlock REPL stdin when auth is required

henrypark133:
## Summary
- unlock REPL stdin when auth-required status updates arrive so the next pasted token is accepted
- add focused regression coverage for the auth-required stdin unlock path

## Verification
- `CARGO_TARGET_DIR=/Users/henry/near_ai/near_claw/ironclaw-ci-regression-fix-2/.shared-target cargo test auth_required_unlocks_stdin_for_token_entry --lib`

### DIFF
--- a/src/channels/repl.rs
+++ b/src/channels/repl.rs
@@ -747,6 +747,9 @@ impl Channel for ReplChannel {
                     eprintln!("  {}{url}{}", fmt::link(), fmt::reset());
                 }
                 eprintln!();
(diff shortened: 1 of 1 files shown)


henrypark133:
Closing this as stale. The REPL auth-unlock fix is already on `staging` via upstream commit `e2556843` (`fix(repl): unlock stdin after auth prompt (#2640)`), so this branch is no longer needed.

PR #2632, opened 2026-04-18
Hide external threads from browser active thread restore

henrypark133:
## Summary
- add `browser_active_thread()` to filter out non-browser-writable threads during restore
- update browser thread list responses to stop reopening external-channel threads as the active chat target
- add narrow unit coverage for the helper and the caller-level thread list behavior

## Verification
- `CARGO_TARGET_DIR=/Users/henry/near_ai/near_claw/ironclaw-ci-regression-fix-2/.shared-target cargo test browser_active_thread --lib`
- `CARGO_TARGET_DIR=/Users/henry/near_ai/near_claw/ironclaw-ci-regression-fix-2/.shared-target cargo test test_chat_threads_handler_hides_external_active_thread --lib`

### DIFF
--- a/src/channels/web/handlers/chat.rs
+++ b/src/channels/web/handlers/chat.rs
@@ -16,6 +16,8 @@ use axum::{
     response::IntoResponse,
 };
 use serde::Deserialize;
+
+use crate::channels::web::util::browser_active_thread;
 // ── SSE / WebSocket handlers ───────────────────────────────────────────
 
 pub async fn chat_events_handler(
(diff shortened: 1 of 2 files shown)


henrypark133:
Closing this as superseded by current staging.

The original fix here filtered browser restore at the old server-side active-thread path, but the relevant gateway flow has since moved and the user-visible issue now appears covered by #2751 (merged in `904e3786`, "fix(gateway): keep engine threads out of chat sidebar").

Given the current code shape, carrying this branch forward would reintroduce an older integration point rather than land the right fix. If we later decide we want stricter fallback/API parity for no-DB or in-memory-only paths, that should be a small fresh follow-up on top of current staging, not this branch.

PR #2638, opened 2026-04-18
test(e2e): fix Slack fixture boot path

### LINKED ISSUE #2623: test(e2e): Slack E2E fixture no longer boots cleanly (`slack.capabilities.json` / health timeout)
## Summary

The Slack E2E flow is still failing, but the current evidence points to the test fixture/setup path rather than a confirmed product regression.

## What I observed

From the local partitioned E2E reruns on April 18, 2026:

- In the full run, the Slack shard errored because the dedicated test server never became healthy (`/api/health` timeout for `slack_e2e_server`).
- In isolation, the setup path also assumes `slack.capabilities.json` exists in the temp channel directory, but that file was not present.

So the failure signature is split but consistent with a broken fixture/bootstrap path:

- fixture copies or generates an incomplete Slack test channel directory
- Slack E2E server never reaches the ready state
- downstream Slack roundtrip tests fail before they can verify product behavior

## Current classification

This should be treated as a test/setup regression first, not as a confirmed Slack runtime regression.

## Affected tests

- `tests/e2e/scenarios/test_slack_e2e.py`

## Reproduction

Run the Slack E2E scenario on current `staging` tip / equivalent local branch state:

```bash
cd tests/e2e
.venv/bin/pytest scenarios/test_slack_e2e.py -q
```

Representative failures seen locally:

- missing `slack.capabilities.json` in the temp channel dir during setup
- `slack_e2e_server` `/api/health` timeout during boot

## Suspected ownership

- `tests/e2e/conftest.py`
- Slack E2E fixture setup under `tests/e2e/scenarios/test_slack_e2e.py`
- any helper that patches/copies the Slack test channel into a temp runtime dir

## Acceptance criteria

- `test_slack_e2e.py` boots the Slack test server reliably in isolation
- temp fixture setup always includes the expected Slack capabilities manifest
- `/api/health` becomes healthy before the test begins message assertions
- once setup is fixed, rerun the scenario and classify any remaining failures as real product bugs or additional stale tests

## Context

serrrfirat:
## Summary

Fixes #2623.

- Start the Slack E2E fixture with the current WASM channel env contract (`WASM_CHANNELS_DIR`, `WASM_TOOLS_DIR`) and a dedicated webhook server port.
- Enable debug/E2E HTTP host rewrites for WASM channel outbound Slack calls while restricting rewrite targets to loopback URLs.
- Activate Slack per test and bind the test owner so Slack DM roundtrip assertions do not depend on pairing state or test order.

## Verification

- `cd tests/e2e && .venv/bin/pytest scenarios/test_slack_e2e.py -q -x`
- `cargo test --no-default-features --features libsql test_rewrite_http_url_for_testing_uses_host_map`

### DIFF
--- a/src/channels/wasm/wrapper.rs
+++ b/src/channels/wasm/wrapper.rs
@@ -62,13 +62,13 @@ use crate::tools::wasm::{
 };
 use ironclaw_safety::LeakDetector;
 
-#[cfg(test)]
+#[cfg(any(test, debug_assertions))]
 const TEST_HTTP_REWRITE_MAP_ENV: &str = "IRONCLAW_TEST_HTTP_REWRITE_MAP";
 
 const WEBSOCKET_EVENT_QUEUE_RELATIVE_PATH: &str = "state/gateway_event_queue";
 const WEBSOCKET_EVENT_PROCESSING_QUEUE_RELATIVE_PATH: &str = "state/gateway_event_queue_processing";
 const WEBSOCKET_EVENT_QUEUE_MAX_ITEMS: usize = 100;
-#[cfg(test)]
+#[cfg(any(test, debug_assertions))]
 const TELEGRAM_TEST_API_BASE_ENV: &str = "IRONCLAW_TEST_TELEGRAM_API_BASE_URL";
 
 // Generate component model bindings from the WIT file
@@ -386,7 +386,7 @@ impl near::agent::channel_host::Host for ChannelStoreData {
         }
 
         let (transport_url, allow_private_test_target) = {
-            #[cfg(test)]
+            #[cfg(any(test, debug_assertions))]
             {
                 let rewritten = rewrite_http_url_for_testing(&logical_url)
                     .or_else(|| rewrite_telegram_api_url_for_testing(&logical_url));
@@ -401,7 +401,7 @@ impl near::agent::channel_host::Host for ChannelStoreData {
                 }
                 (url, is_rewritten)
             }
-            #[cfg(not(test))]
+            #[cfg(not(any(test, debug_assertions)))]
             {
                 (logical_url.clone(), false)
             }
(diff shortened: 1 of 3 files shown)


ilblackdragon (commented):
## Code review

**Overview.** Fixes #2623 by aligning the Slack E2E fixture with the current WASM channel env contract and hardening the dev-time HTTP rewrite. Three changes: (1) expose the test HTTP/Telegram rewrite hooks in `debug_assertions` builds (not just `#[cfg(test)]`) with loopback-only target enforcement, (2) rewrite `slack_e2e_server` to use shared helpers and a dedicated webhook port, (3) new `active_slack` fixture that activates + owner-binds per test.

### Strengths

- **Loopback enforcement is the right hardening.** With the rewrite map now shipping in debug builds, `is_loopback_test_rewrite_base` (wrapper.rs:4270) rejects non-loopback targets at parse time with a `tracing::warn!` that cites the env var. Good coverage via the added unit test.
- **Fixture refactor matches the Telegram sibling.** Ports reserved via `_reserve_loopback_sockets`, teardown via `_stop_process`, coverage env forwarding via `_forward_coverage_env` — consistent with `_telegram_e2e_server_impl`.
- **Stronger roundtrip assertion** (`assert "Hello! How can I help you today?" in reply_text`) catches regressions the prior non-empty check missed.
- **`owner_id` now bound via capabilities JSON**, removing pairing-order flakiness that made the prior version order-dependent.

### Issues addressed in follow-up commit (bbc17c6a)

1. **Reserved sockets could leak on early fixture failure.** `_reserve_loopback_sockets(2)` was called before the `try:` block, so a failure in `TemporaryDirectory(...)` construction would leave the sockets bound. Moved the reserve into an outer try/finally that closes any still-bound sockets, matching the Telegram impl.
2. **Redundant `reset_fake_slack` calls at test entry.** `active_slack` already calls reset as its last setup step, so each test starts clean. Dropped the per-test leading `reset_fake_slack(...)` calls (kept the intentional mid-test reset in `test_slack_malformed_payload_resilience`).

### Non-issues (flagging for the record)

- Gemini's suggestion to use `asyncio.TimeoutError` is not needed: `wait_for_ready` in `helpers.py:191` raises the built-in `TimeoutError`, and the Telegram sibling catches the same built-in. Python >= 3.11 is already the minimum per `tests/e2e/CLAUDE.md`.
- Loopback check correctly covers `127.0.0.0/8` and `::1` via `IpAddr::is_loopback()` plus the `localhost` literal.

### Nit (not addressed)

- `active_slack` returns the fixture dict via `return` rather than `yield`. Works because there's no per-test teardown, but `yield` is the conventional pattern for pytest-asyncio fixtures. Fine to leave.

LGTM with the cleanup pushed as bbc17c6a.

ilblackdragon (approved):

PR #2640, opened 2026-04-18
fix(repl): unlock stdin after auth prompt

### LINKED ISSUE #2625: test(e2e): macOS REPL auth matrix PTY harness does not deliver interactive input reliably
## Summary

The remaining REPL auth matrix failure currently looks like a macOS PTY harness problem, not a confirmed product auth regression.

## What I verified

While debugging `tests/e2e/scenarios/test_v2_auth_oauth_matrix.py::test_repl_http_auth_prompt_accepts_token_and_retries` on April 18, 2026:

- Hosted / browser OAuth regressions were fixed separately and pass in targeted reruns.
- The product-side REPL auth bug where `StatusUpdate::AuthRequired` failed to unlock stdin was fixed.
- The E2E fixture was also stale because it defaulted into TUI mode; forcing `CLI_MODE=repl` was necessary.

Even after those fixes, the PTY-driven REPL test still fails on macOS:

- the child `ironclaw` process stays alive
- PTY output degenerates into terminal control sequences / cursor toggles
- the first synthetic user line never becomes a persisted conversation turn
- the test DB only shows the seeded assistant greeting, not the user message
- bulk write with `\n`, `\r`, and slow keystroke-style typing all reproduced the same behavior locally

## Current classification

Test harness / platform issue in the Python PTY launcher, most likely around Rustyline + controlling TTY behavior on macOS.

Not enough evidence yet to classify this as a remaining product auth regression.

## Affected test

- `tests/e2e/scenarios/test_v2_auth_oauth_matrix.py::test_repl_http_auth_prompt_accepts_token_and_retries`

## Reproduction

```bash
cd tests/e2e
.venv/bin/pytest scenarios/test_v2_auth_oauth_matrix.py::test_repl_http_auth_prompt_accepts_token_and_retries -q
```

Observed behavior:

- timeout waiting for `Authentication required for google_oauth_token|Sign in with Google|Paste your token`
- PTY output contains mostly control sequences
- no user turn is persisted to the test DB

## Suspected ownership

- `tests/e2e/scenarios/test_v2_auth_oauth_matrix.py`
- PTY launcher helper `_start_auth_matrix_repl(...)`
- Python-side pseudo-terminal management on macOS

## Suggested direction

Treat this as a test harness problem first:

- replace the current `asyncio.create_subprocess_exec(..., stdin=slave_fd, stdout=slave_fd, stderr=slave_fd)` approach with a more correct `forkpty` / `pexpect`-style launcher or equivalent controlling-TTY setup
- keep the fixture explicitly in REPL mode (`CLI_MODE=repl`)
- only reopen product auth investigation if a better PTY harness still reproduces the failure

## Acceptance criteria

- the REPL fixture reliably submits a first line to the interactive Rust REPL on macOS
- the auth matrix test can observe the expected auth prompt bytes
- the test no longer depends on TUI defaults or platform-fragile PTY behavior

## Context

serrrfirat:
## Summary
- unlock REPL stdin after rendering `StatusUpdate::AuthRequired` so token follow-up input can be read
- add a focused REPL regression test for the auth-required path
- confirm the existing auth matrix E2E scenario now passes end-to-end

Fixes #2625

## Testing
- cargo test --lib --no-default-features --features libsql auth_required_status_unlocks_stdin_for_follow_up_submission -- --nocapture
- tests/e2e/.venv/bin/pytest scenarios/test_v2_auth_oauth_matrix.py::test_repl_http_auth_prompt_accepts_token_and_retries -q

### DIFF
--- a/src/channels/repl.rs
+++ b/src/channels/repl.rs
@@ -747,6 +747,9 @@ impl Channel for ReplChannel {
                     eprintln!("  {}{url}{}", fmt::link(), fmt::reset());
                 }
                 eprintln!();
+                // Resume readline so the user can paste a token or continue the
+                // auth flow via the normal submission parser path.
+                self.stdin_locked.store(false, Ordering::Relaxed);
             }
             StatusUpdate::AuthCompleted {
                 extension_name,
@@ -874,6 +877,30 @@ mod tests {
                 .expect("timed out waiting for stream to close")
                 .is_none(),
             "stream should end after the single message"
+        );
+    }
+
+    #[tokio::test]
+    async fn auth_required_status_unlocks_stdin_for_follow_up_submission() {
+        let repl = ReplChannel::with_user_id("test-user");
+        repl.stdin_locked.store(true, Ordering::Relaxed);
+
+        repl.send_status(
+            StatusUpdate::AuthRequired {
+                extension_name: "google_oauth_token".to_string(),
+                instructions: Some("Paste your token".to_string()),
+                auth_url: None,
+                setup_url: Some("http://127.0.0.1:8080/auth".to_string()),
+                request_id: None,
+            },
+            &serde_json::json!({}),
+        )
+        .await
+        .expect("auth required status should render successfully");
+
+        assert!(
+            !repl.stdin_locked.load(Ordering::Relaxed),
+            "auth prompts must unlock stdin so the next submission can be read"
         );
     }
 }
--- a/src/channels/repl.rs
+++ b/src/channels/repl.rs
@@ -880,10 +880,28 @@ mod tests {
         );
     }
 
+    /// Regression: per `.claude/rules/testing.md` ("Test Through the Caller,
+    /// Not Just the Helper"), a unit test that only asserts the
+    /// `stdin_locked` flag flipped is insufficient — the side effect that
+    /// matters is the input thread's polling loop in `start()` actually
+    /// resuming. This test mirrors that polling pattern and asserts it
+    /// observes the unlock after `send_status(AuthRequired)`.
     #[tokio::test]
-    async fn auth_required_status_unlocks_stdin_for_follow_up_submission() {
+    async fn auth_required_status_unblocks_input_polling_loop() {
         let repl = ReplChannel::with_user_id("test-user");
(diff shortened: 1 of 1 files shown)


ilblackdragon:
## Code Review

### Overview
Fixes #2625 by unlocking REPL stdin in `StatusUpdate::AuthRequired` so the user can type a token after the prompt renders. Mirrors the existing pattern for `ApprovalRequired` at `src/channels/repl.rs:726`. +27/-0, single file.

### Correctness
- **Fix is right and minimal.** Without it, `read_line` blocks on `stdin_locked.load()` (line 463) and the user can never paste a token. Mirror of the approval-required unlock at lines 718–726 — same shape, same justification. `Ordering::Relaxed` matches the rest of the file.
- **Symmetric with sibling sites:** `respond` (lines 581/593/612) and `ApprovalRequired` (line 726) all unlock the same way; `AuthRequired` was the outlier. No other locking sites missed.
- **No race risk:** unlock happens after the prompt is written to stderr, so the user can't submit before they see the prompt.

### Test coverage
- Unit test asserts the channel-side flag flips. Per `.claude/rules/testing.md` ("Test Through the Caller"), a helper-only test is normally thin — but here the side effect is `read_line` on a real terminal, which can't be unit-tested. The PR description shows the e2e scenario passes. Acceptable.
- Bug report (#2625) flags the macOS PTY harness as flaky; the unit test is the durable regression guard.

### Style / conventions
- `eprintln!` and `Ordering::Relaxed` consistent with surrounding code.
- Comment explains the *why* (resume readline so token paste works) — matches line 724.
- No `unwrap()`/`expect()` in production path.

### Risks
- Low. Only behavior change is unlocking a flag that every adjacent code path already unlocks.

### Suggestions (optional, non-blocking)
- Consider extracting the repeated `self.stdin_locked.store(false, Ordering::Relaxed)` + comment idiom into a small `unlock_stdin(&self)` helper. Five call sites now; pulling it into one method would make the next "what should unlock stdin?" question a one-line audit. Not worth holding the PR.

### Verdict
LGTM. Ship it.

ilblackdragon (approved):

PR #2647, opened 2026-04-18
ci(gateway): enforce platform/feature boundaries — ironclaw#2599 stage 5

ilblackdragon:
## Summary

Fifth increment of the ironclaw#2599 platform/feature split. Adds `scripts/check_gateway_boundaries.py` and wires it into the `code_style` CI workflow as a required check. Stacks on #2645 (which opens the `features/` subtree) and #2644 (which relocates auth/sse/ws into `platform/`).

## The layering rule

Every file under `src/channels/web/platform/` **except `router.rs`** must not import from `handlers/`, `features/`, or route through the `server.rs` compatibility shim. The router is exempt because it's the composition point — route registration is where transport meets features, and the router imports every handler it registers. The `server.rs` shim is rejected alongside `handlers/` and `features/` because routing through it re-creates the same platform → feature back-edge the shim is meant to be migrated away from.

## How the check works

`scripts/check_gateway_boundaries.py`:

- Walks `src/channels/web/platform/*.rs`, skipping `router.rs` and the body of any `#[cfg(test)] mod ...` or `mod tests { ... }` block (brace-depth tracked before scanning).
- Strips line comments, block comments, and string / raw-string / char literals so references inside docstrings and explanatory text don't trigger false positives.
- Matches nine forbidden import shapes on a per-line pass:
  - `crate::channels::web::{handlers,features,server}::`
  - `super::{handlers,features,server}::`
  - `super::super::{handlers,features,server}::`
- Runs a second, brace-aware grouped-import pass so forbidden segments inside `use PREFIX::{ ... }` blocks are caught even when the group contains nested braces (e.g. `use crate::channels::web::{ platform::{state::GatewayState}, handlers::auth::login_handler };`). The scanner finds each header, walks to the matching `}` by tracking brace depth, and searches the body for `(handlers|features|server)::`.
- Prints diagnostics with file:line and the matched pattern for every violation; exits non-zero on any.
- Carries unit tests behind a `test` subcommand (`python3 scripts/check_gateway_boundaries.py test`) that CI runs alongside the check itself.

CI wiring: new `gateway-boundaries` job added to `.github/workflows/code_style.yml` and included in the `code-style` roll-up's required-jobs list. Runs on every PR that touches Rust code, **and** on any PR that edits `scripts/check_gateway_boundaries.py` or `.github/workflows/code_style.yml` (a separate `has_boundary_check` output on the `changes` job, so the guardrail can't be silently skipped by a PR that only touches `scripts/` or `.github/`).

## Fixed: one pre-existing back-edge

`platform/auth.rs`'s OIDC JWT validation was calling `check_email_domain()` in `handlers/auth.rs`. The helper is platform-level — it gates JWT validation before any handler runs — so it moves into `platform::auth` along with its five unit tests. `handlers::auth::handle_callback` now imports from the new home. No behavior change.

## Allowlisted: pre-existing back-edges (tracked, narrowly scoped)

Two narrow allowlist entries cover pre-existing back-edges that are targeted for follow-up migration. Each is per-symbol — not a broad prefix — so accidental *new* back-edges in the same file still fail the check:

- `platform/static_files::build_frontend_html` calls `read_layout_config` and `load_resolved_widgets`, both still in `handlers/frontend.rs`. Migrating them requires also moving `read_widget_manifest` and the widget-size constants, which touches `load_widget_manifests` (used by `/api/frontend/widgets` and the engine-v2 widget endpoint). Tracked via a single `handlers::frontend` entry.
- `platform/ws.rs` reaches through the `server.rs` shim for seven specific symbols (`GatewayState`, `PerUserRateLimiter`, `RateLimiter`, `ActiveConfigSnapshot`, `images_to_attachments`, `handle_legacy_auth_token_submission`, `handle_legacy_auth_cancel`). Each is listed individually so a new accidental `server::` import in `ws.rs` still fails.

The allowlist is explicitly annotated as "must not grow without reviewer sign-off".

## Test plan

- [x] `python3 scripts/check_gateway_boundaries.py test` — 16 tests pass (grouped imports, nested-brace grouped imports, `server::` shim, `#[cfg(test)] / mod tests` skip, rest-of-file still flagged, direct per-line patterns)
- [x] `python3 scripts/check_gateway_boundaries.py` — clean
- [x] `cargo fmt --all`
- [x] `cargo clippy --all --benches --tests --examples --all-features` — clean
- [x] `python3 scripts/check_no_panics.py` — clean
- [x] `cargo test --lib channels::web` — 425 passed

## Base

Stacks on #2645. Once #2644 and #2645 land in order, this PR's base auto-flips to `staging`.

## Follow-ups

- Migrate `read_layout_config` / `load_resolved_widgets` / `read_widget_manifest` + constants from `handlers/frontend.rs` into `platform/static_files.rs`; remove the allowlist entry.
- Migrate the seven `ws.rs` shim symbols into `platform/*` and remove each allowlist entry as it moves.
- Stage 4b: `features/pairing/` + `features/logs/` + `features/status/`.
- Stage 4c: `features/chat/`.
- Stage 4d: `features/extensions/`.
- Stage 6: delete the remaining `server.rs` re-export shims once every call site uses the canonical `platform::` / `features::` paths.


scripts/check_gateway_boundaries.py
```diff
@@ -0,0 +1,401 @@
...
+        if filename == allow_file and allow_prefix in line:
+            return True
+    return False
+
+
+def _find_violations(path: pathlib.Path) -> list[Violation]:
+    text = path.read_text(encoding="utf-8", errors="replace")
+    sanitized = _strip_comments_and_strings(text)
+    violations: list[Violation] = []
+    for idx, line in enumerate(sanitized.splitlines(), start=1):
```
serrrfirat: **High Severity**

This checker is line-based, so the boundary is still easy to bypass with valid Rust syntax. I verified locally that both of these currently produce a clean result:

```rust
use crate::channels::web::{
    handlers::auth::login_handler,
    platform::state::GatewayState,
};

use crate::channels::web::server::chat_send_handler;
```

The first escapes the regex because `handlers::...` is on a later line inside a grouped `use`, and the second goes through the `server.rs` compatibility shim even though that still creates a platform -> feature dependency. That means this required CI check can report green while the forbidden back-edge is already present.

Suggestion: parse `use` trees / fully-qualified paths semantically, or at minimum add regression coverage for grouped imports and reject `crate::channels::web::server::` from non-router platform modules while the shim exists.

scripts/check_gateway_boundaries.py
```diff
@@ -0,0 +1,401 @@
...
+type paths — comments and string literals are stripped first):
+
+- `crate::channels::web::handlers::` and `crate::channels::web::features::`
+- `super::handlers::` and `super::features::`
+- `super::super::handlers::` and `super::super::features::`
+
+Allowed:
+
+- `platform/router.rs` (explicit skip — the composition point).
+- Any test module inside a platform file (`#[cfg(test)]` or `mod tests`).
```
serrrfirat: **Medium Severity**

The docstring says `#[cfg(test)]` / `mod tests` blocks are exempt, but the implementation below never actually skips test modules before scanning. I verified locally that a platform file containing:

```rust
#[cfg(test)]
mod tests {
    use crate::channels::web::handlers::auth::login_handler;
}
```

is still reported as a violation.

That leaves the checker and its documented policy out of sync, and it will block legitimate caller-level regression tests in platform files.

Suggestion: either teach the scanner to ignore test modules explicitly, or drop this exemption from the contract. A regression test for a `#[cfg(test)] mod tests` import would pin the intended behavior.

.github/workflows/code_style.yml
```diff
@@ -207,6 +207,24 @@ jobs:
         BASE="${{ github.event.pull_request.base.sha }}"
         python3 scripts/check_no_panics.py --base "$BASE" --head HEAD
 
+  gateway-boundaries:
+    name: Gateway platform/feature boundaries
+    needs: changes
+    if: needs.changes.outputs.has_code == 'true'
```
serrrfirat: **Medium Severity**

This job is keyed off `has_code`, but `has_code` currently ignores `scripts/` and `.github/workflows/`. That means a PR that only changes `scripts/check_gateway_boundaries.py` or this workflow will skip `gateway-boundaries` entirely and still satisfy the roll-up.

In other words, the guardrail can be weakened without any required check running.

Suggestion: include the checker/workflow paths in the `changes` job's diff filter, or give this job its own condition so it always runs when the script or workflow changes.

scripts/check_gateway_boundaries.py
```diff
@@ -0,0 +1,401 @@
...
+                violations.append(
+                    Violation(
+                        path=path,
+                        line_number=idx,
+                        line=original.rstrip(),
+                        pattern=pattern.pattern,
+                    )
+                )
+                break  # one violation per line is enough
+    return violations
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in ad9a67ed. `text.splitlines()` is now computed once outside the per-line loop (`original_lines = text.splitlines()` at the top of `_find_violations`) instead of being re-split per violation.

scripts/check_gateway_boundaries.py
```diff
@@ -0,0 +1,401 @@
...
+def check(platform_dir: pathlib.Path = PLATFORM_DIR) -> list[Violation]:
+    if not platform_dir.is_dir():
+        return []
+    violations: list[Violation] = []
+    for path in sorted(platform_dir.rglob("*.rs")):
+        relative = path.relative_to(platform_dir)
+        if str(relative) in EXEMPT_RELATIVE_PATHS:
+            continue
+        violations.extend(_find_violations(path))
+    return violations
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in ad9a67ed. Added `_blank_test_modules` which uses brace-depth tracking to blank the body of `#[cfg(test)] mod ... { ... }` and `mod tests { ... }` blocks before the forbidden-pattern scan runs, so the docstring's stated exemption now actually holds. Regression tests `test_skips_cfg_test_module` and `test_skips_mod_tests_without_cfg` cover the exemption, and `test_still_flags_production_code_after_test_module` pins the boundary so the skip doesn't blanket-ignore the rest of the file.

scripts/check_gateway_boundaries.py
ilblackdragon: Fixed in ad9a67ed. Added `_blank_test_modules` which uses brace-depth tracking to blank the body of `#[cfg(test)] mod ... { ... }` and `mod tests { ... }` blocks before the forbidden-pattern scan runs, so the docstring's stated exemption now actually holds. Regression tests `test_skips_cfg_test_module` and `test_skips_mod_tests_without_cfg` cover the exemption, and `test_still_flags_production_code_after_test_module` pins the boundary so the skip doesn't blanket-ignore the rest of the file.

scripts/check_gateway_boundaries.py
ilblackdragon: Fixed in ad9a67ed, covering both bypasses you called out.

**Grouped imports.** Added `GROUPED_FORBIDDEN_PATTERN` — a multiline regex over the sanitized text that matches `crate::channels::web::{...}`, `super::{...}`, and `super::super::{...}` groups containing `handlers::`, `features::`, or `server::`. It uses a named capture group on the forbidden segment and reports the line where that segment actually appears (the continuation line), not the line with the opening brace. Covered by `test_detects_grouped_crate_web_import` and `test_detects_grouped_super_import`.

**`server.rs` shim.** Added `crate::channels::web::server::`, `super::server::`, and `super::super::server::` to `FORBIDDEN_PATTERNS`. Pre-existing platform → feature shim usage in `platform/ws.rs` (`GatewayState` re-export, `PerUserRateLimiter`/`RateLimiter` constructors, `ActiveConfigSnapshot`, `images_to_attachments`, legacy v1 auth-token handlers) is captured as a single tracked `ALLOWLIST` entry with a comment tying it to ironclaw#2599 — the allowlist shrinks as individual types migrate out of `server.rs`. Any **new** use of the shim from non-router platform modules now fails the check. Covered by `test_detects_server_shim_back_edge`.

.github/workflows/code_style.yml
ilblackdragon: Fixed in ad9a67ed. Added a second output on the `changes` job — `has_boundary_check` — that fires when `scripts/check_gateway_boundaries.py` or `.github/workflows/code_style.yml` change. The `gateway-boundaries` job now gates on `has_code == 'true' || has_boundary_check == 'true'`, so a PR that only edits the checker or its workflow still runs the guardrail and can't slip past the roll-up. I kept the script/workflow paths *out* of `has_code` proper so they don't spuriously trigger the full clippy matrix on doc-adjacent script edits.

scripts/check_gateway_boundaries.py
```diff
@@ -0,0 +1,630 @@
...
+Forbidden patterns (matched inside `use ...` statements or fully-qualified
+type paths — comments and string literals are stripped first):
+
+- `crate::channels::web::handlers::` and `crate::channels::web::features::`
+- `crate::channels::web::server::` (the `server.rs` compatibility shim —
+  routing through it re-creates the same platform → feature back-edge the
+  shim is meant to be migrated away from)
+- `super::handlers::`, `super::features::`, and `super::server::`
+- `super::super::handlers::`, `super::super::features::`, and
+  `super::super::server::`
```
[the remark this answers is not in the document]
ilblackdragon: PR description updated. The layering-rule and how-it-works sections now list `server::` alongside `handlers::` / `features::` (nine forbidden import shapes total, with the three `server::` variants explicitly called out), document the brace-aware grouped-import pass, and describe the `has_boundary_check` workflow-protection output. The allowlist section also now makes clear that both entries (the `handlers::frontend` one and the seven `ws.rs` `server::` ones) are per-symbol rather than broad prefixes.

scripts/check_gateway_boundaries.py
```diff
@@ -0,0 +1,630 @@
...
+# or the same via `super::{ ... }` / `super::super::{ ... }`. The per-line
+# scan above misses these because the forbidden segment lands on a
+# continuation line. We match across newlines (sanitized text already has
+# comments and string literals blanked) and report the line where the
+# offending segment actually appears.
+GROUPED_FORBIDDEN_PATTERN = re.compile(
+    r"\b(?:crate::channels::web|super(?:::super)?)::\{"
+    r"[^{}]*?\b(?P<forbidden>handlers|features|server)::",
+    re.DOTALL,
+)
```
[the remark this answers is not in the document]
ilblackdragon: Good catch — fixed in 1cf86d3a. I reproduced the bypass locally (`use crate::channels::web::{ platform::{state::GatewayState}, handlers::auth::login_handler };` produced zero violations) and replaced the `[^{}]*?` regex with a brace-aware walk.

Concretely: `GROUPED_IMPORT_HEADER` now only matches the header `crate::channels::web::{` / `super::{` / `super::super::{`; the scanner then walks from the header's `{` to the matching `}` by tracking brace depth (incrementing on `{`, decrementing on `}`), and searches the body for `\b(handlers|features|server)::` via `GROUPED_FORBIDDEN_SEGMENT`. Line numbers come from absolute offsets, so the reported line is where the forbidden segment lives — including in nested-brace groups.

Covered by `test_detects_nested_brace_grouped_import` (16 tests total now).

scripts/check_gateway_boundaries.py
```diff
@@ -0,0 +1,630 @@
...
+    ("static_files.rs", "crate::channels::web::handlers::frontend"),
+    # platform/ws.rs reaches through the `server.rs` compatibility shim for
+    # a handful of types and helpers that have not yet moved into
+    # `platform/*` (GatewayState re-export, PerUserRateLimiter / RateLimiter
+    # constructors, ActiveConfigSnapshot, images_to_attachments, and the
+    # legacy v1 auth-token handlers). Each is tracked for relocation under
+    # ironclaw#2599; this allowlist entry shrinks as individual items move.
+    # Added in stage 5 when `crate::channels::web::server::` was promoted to
+    # a forbidden pattern.
+    ("ws.rs", "crate::channels::web::server::"),
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in 1cf86d3a, narrowed as suggested. Replaced the single broad `crate::channels::web::server::` entry with seven per-symbol entries covering the actual current uses in `platform/ws.rs`:

- `GatewayState`
- `PerUserRateLimiter`
- `RateLimiter`
- `ActiveConfigSnapshot`
- `images_to_attachments`
- `handle_legacy_auth_token_submission`
- `handle_legacy_auth_cancel`

I kept the two `handle_legacy_auth_*` entries that were not in your suggested list — without them the check would have flagged lines 283 and 301 of `ws.rs` (the v1 auth-token compatibility path). Each entry is removed individually as the corresponding symbol migrates into `platform/*`.

Verified the narrower allowlist still lets legitimate uses pass and now rejects an accidental new shim import: a simulated `use crate::channels::web::server::chat_send_handler;` in a file named `ws.rs` produces 1 violation, confirming the guardrail is active.

PR #2652, opened 2026-04-18
fix(telegram): unblock e2e activation flow

### LINKED ISSUE #2624: test(e2e): Telegram end-to-end activation still fails with bot token validation 401
## Summary

The Telegram end-to-end scenario is still failing after the broader browser/bootstrap regressions were fixed.

## What I observed

From the local E2E reruns on April 18, 2026:

- `tests/e2e/scenarios/test_telegram_e2e.py` still fails repeatedly
- the activation flow returns:

```text
success: false
Bot token validation failed: HTTP 401 Unauthorized
```

Important contrast:

- `test_telegram_hot_activation.py` passed
- `test_telegram_token_validation.py` passed
- the full Telegram E2E roundtrip still failed

That makes this look like an activation/runtime integration regression or a fixture drift in the Telegram mock API path, not a generic browser problem.

## Current classification

Likely a real Telegram activation/runtime regression or a Telegram-specific E2E mock integration mismatch.

## Affected tests

- `tests/e2e/scenarios/test_telegram_e2e.py`

## Reproduction

```bash
cd tests/e2e
.venv/bin/pytest scenarios/test_telegram_e2e.py -q
```

Representative failure seen locally:

- activation step returns `Bot token validation failed: HTTP 401 Unauthorized`

## Suggested investigation areas

- Telegram mock API base rewrite / test transport path
- Telegram activation path versus hot-activation UI-only path
- difference between token format validation and real activation validation
- channel runtime behavior after install/configure versus the UI stepper flow

## Likely owner surfaces

- Telegram channel runtime / activation code under `src/channels/wasm/`
- Telegram-specific test fixture setup in `tests/e2e/`
- any mock Telegram API routing used by `test_telegram_e2e.py`

## Acceptance criteria

- `test_telegram_e2e.py` passes in isolation on current `staging`
- the activation path succeeds against the Telegram test fixture
- if the failure is fixture-only, document and fix the test transport/setup path without changing production behavior unnecessarily

## Context

Related coverage umbrella: #1783

serrrfirat:
## Summary
- honor Telegram test HTTP rewrites in debug builds so WASM channel activation talks to the fake Bot API during E2E runs
- make the Telegram E2E helper reuse an already-activated session-scoped server instead of repeatedly restarting the channel
- make synthetic Telegram webhook update IDs monotonic across the scenario so stale-update protection does not drop later test messages

Fixes #2624

## Testing
- cargo test --lib --no-default-features --features libsql rewrite_ -- --nocapture
- cargo build --no-default-features --features libsql
- tests/e2e/.venv/bin/pytest scenarios/test_telegram_e2e.py -q

### DIFF
--- a/src/channels/wasm/wrapper.rs
+++ b/src/channels/wasm/wrapper.rs
@@ -4288,7 +4288,19 @@ fn is_loopback_test_rewrite_base(base: &str) -> bool {
 fn rewrite_telegram_api_url_for_testing(url: &str) -> Option<String> {
     let override_base = crate::config::helpers::env_or_override(TELEGRAM_TEST_API_BASE_ENV)
         .map(|value| value.trim().trim_end_matches('/').to_string())
-        .filter(|value| !value.is_empty())?;
+        .filter(|value| !value.is_empty())
+        .filter(|base| {
+            if is_loopback_test_rewrite_base(base) {
+                true
+            } else {
+                tracing::warn!(
+                    env_var = TELEGRAM_TEST_API_BASE_ENV,
+                    %base,
+                    "Ignoring non-loopback Telegram test API rewrite target"
+                );
+                false
+            }
+        })?;
 
     let parsed = url::Url::parse(url).ok()?;
     if !matches!(parsed.scheme(), "http" | "https") {
@@ -6333,6 +6345,15 @@ mod tests {
             rewrite_telegram_api_url_for_testing("https://api.telegram.org/bot123/sendMessage")
                 .expect("Telegram URL should rewrite");
         assert_eq!(rewritten, "http://127.0.0.1:19001/bot123/sendMessage");
+
+        // Non-loopback rewrite targets are ignored because credential injection
+        // happens before test transport rewrite.
+        unsafe {
+            std::env::set_var(TELEGRAM_TEST_API_BASE_ENV, "https://example.com");
+        }
+        let rewritten =
+            rewrite_telegram_api_url_for_testing("https://api.telegram.org/bot123/sendMessage");
+        assert!(rewritten.is_none());
 
         // SAFETY: Under ENV_MUTEX, no concurrent env access.
         unsafe {
--- a/src/channels/wasm/wrapper.rs
+++ b/src/channels/wasm/wrapper.rs
@@ -418,7 +418,7 @@ impl near::agent::channel_host::Host for ChannelStoreData {
             .unwrap_or(10 * 1024 * 1024);
 
         // Resolve hostname and reject private/internal IPs to prevent DNS rebinding.
-        // Test-only URL rewrites intentionally point at local fake servers.
+        // Test/dev URL rewrites intentionally point at local fake servers.
         if !allow_private_test_target {
             reject_private_ip(&transport_url)?;
         }
--- a/tests/e2e/scenarios/test_telegram_e2e.py
+++ b/tests/e2e/scenarios/test_telegram_e2e.py
@@ -9,6 +9,7 @@ import json
 import os
 import re
 import time
+from itertools import count
 
 import httpx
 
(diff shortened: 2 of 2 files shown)


ilblackdragon:
## Code Review

### Overview
Three related fixes to unblock the Telegram E2E scenarios (fixes #2624):
1. `wrapper.rs`: Reject non-loopback rewrite targets in `rewrite_telegram_api_url_for_testing`, with a `warn!` log — aligns behavior with the sibling `parse_test_http_rewrite_map`.
2. `test_telegram_e2e.py`: Session-level caching of the activated Telegram server via `active_telegram[_with_routines]` fixtures, avoiding repeated teardown/reinstall.
3. `test_telegram_e2e.py`: Monotonic synthetic `update_id`s via a module-level `itertools.count`, so the channel's stale-update protection doesn't drop later messages.

### Strengths
- **Security-positive Rust change.** The non-loopback guard matches the existing pattern for `TEST_HTTP_REWRITE_MAP_ENV`, keeping debug-build rewrites confined to localhost. The new unit test covers the rejection branch.
- **Cleaner test intent.** `_next_test_update_id()` removes ad-hoc collision-avoidance (e.g. `int(time.time() * 1000) % 2_147_483_647` in `pair_telegram_user`) and documents the Telegram monotonicity invariant.
- **Diff is tightly scoped.** Only whitespace- and fixture-plumbing changes in the Python file; nothing outside of E2E harness.

### Issues / Suggestions

- **Stale comment on `_TEST_UPDATE_IDS`.** `test_telegram_e2e.py:30` says *"Scenario-specific assertions use much larger IDs (100, 200, …), so these bootstrap/pairing helper IDs stay out of the way."* After this PR, every scenario also uses the shared counter — there are no hard-coded large IDs anymore. Either remove the second sentence or justify why the counter starts at 1.

- **Test order coupling.** With activation cached in module state, running a test in isolation still works, but any future test that perturbs channel install/uninstall/credentials state would silently leak into subsequent tests. Consider either (a) a `session`-scoped autouse fixture so caching intent is explicit, or (b) asserting the installed/activated state in `active_telegram` rather than skipping on cache hit. Minor; flag for future maintainers.

- **pytest-xdist safety.** `_TEST_UPDATE_IDS` and `_ACTIVATED_TELEGRAM_SERVERS` are module globals. If these tests are ever run with `pytest-xdist` sharing a backend server across workers, update_id counters would diverge per worker and duplicate updates could land on the same server. Not a concern today (tests appear serial per server), but worth a one-line comment that asserts the serial assumption.

- **Unit test does not restore env between phases.** In `test_rewrite_telegram_api_url_for_testing_uses_test_override` (`wrapper.rs:6322`), the new `set_var(\"https://example.com\")` block never explicitly clears the first value before re-setting; `set_var` overwrites, so this is fine, but adjacent tests within `mod tests` must trust the `lock_env` guard to serialize. Existing pattern, not a regression — worth a `// under ENV_MUTEX` reminder above the second `set_var` for consistency with the surrounding SAFETY comments.

- **`filter(|base| { … })` reads awkwardly** (returns bool, side-effects a warn). A `match` or explicit `if is_loopback(…) { Some(base) } else { warn; None }` via `and_then` would read closer to `parse_test_http_rewrite_map`. Taste call, not blocking.

### Conformance to project rules
- `.claude/rules/testing.md` \"Test Through the Caller, Not Just the Helper\": The helper is `rewrite_telegram_api_url_for_testing`, its side effect (outbound HTTP) is gated at `wrapper.rs:392`. The e2e scenarios are the caller-level coverage; they're updated in this same PR, so this rule is satisfied.
- `.claude/rules/review-discipline.md` \"Regression test with every fix\": Rust has a new `#[test]` for the non-loopback path. For the Python update_id monotonicity, the regression is the existing scenarios (no added assertion, but the fix reenables them). Reasonable.
- Logging level: `warn!` for an ignored test-only config — correct (debug-build only, config-boundary).

### Verdict
Approve with nits. The Rust change is a proper hardening and the Python fixture cleanup is a clear improvement. Address the stale comment on `_TEST_UPDATE_IDS` before merge; the rest are follow-up suggestions.

ilblackdragon (approved):

PR #2653, opened 2026-04-18
fix(secrets): TOCTOU-safe auto-generate, lazy keychain probe, fail-loud on stale DB

ilblackdragon:
## Summary

Follow-up to #2648 addressing the review feedback on #2653.

### 1. Probe order — env-first, lazy keychain
Kept `SecretsConfig::resolve()` at env-first → keychain → auto-generate, matching every other master-key reader (`SetupWizard::step_security`, `SetupWizard::init_secrets_context`, `crate::secrets::resolve_master_key`, `cli import`). An explicit `SECRETS_MASTER_KEY` now wins without a keychain side-probe, and `resolve_with_env_path` only touches the OS keychain when the env var is unset — no more spurious macOS Keychain Access dialogs when the user has explicitly configured a key.

### 2. TOCTOU safety (unchanged behavior, parser bug fixed)
`auto_generate_and_persist` re-reads `.env` before writing a generated key so we don't overwrite a concurrent writer. Reviewers (gemini, Copilot, @serrrfirat) caught that the parser used `line.split_once('=')?`, which aborted the whole scan on the first blank or comment line — so a normal `.env` with a leading `#` defeated the re-check. Fixed to `continue` past non-assignment lines.

### 3. Fail loudly on stale DB meeting fresh key
New `SecretsConfig.generated` flag, set only when resolve falls through to `auto_generate_and_persist`. `AppBuilder::init_secrets` now calls a new `SecretsStore::any_exist()` and errors out if a freshly-generated key meets an already-populated secrets table. Without this, an out-of-band loss of `SECRETS_MASTER_KEY` (env var unset, keychain cleared, `.env` wiped) would silently boot with a new key, shadowing the existing encrypted rows forever. Better to stop and tell the user to restore the original key.

Backends: PostgreSQL and libSQL override `any_exist` with `SELECT 1 … LIMIT 1`; in-memory overrides with an `is_empty` check; default trait impl returns `Ok(false)` so test doubles don't need to implement it.

### 4. Testable resolve (unchanged from initial PR)
`resolve_inner` takes an injected keychain probe result and `allow_keychain_persist: bool` so tests exercise every branch deterministically without hitting the real keychain. Test-only `crate::config::clear_injected_var` keeps the overlay clean between tests.

### 5. Dead branch removed (unchanged from initial PR)
`auto_setup_security`'s `KeySource::None` arm is `unreachable!()`.

## Test plan

- [x] `cargo test --lib` — all 5086 tests pass; 6 new ones:
  - `env_wins_over_keychain_when_both_present` — probe-order invariant (replaces `keychain_wins_over_env_when_both_present`)
  - `keychain_wins_when_env_unset` — keychain-fallback coverage
  - `env_var_wins_when_keychain_empty` — CI/Docker escape hatch still works
  - `generated_flag_tracks_auto_generate_path` — `generated=true` iff the auto-generate branch ran
  - `read_secrets_master_key_skips_blank_and_comment_lines` — TOCTOU parser regression
  - `any_exist_reflects_global_store_state` — safety-gate backing query on in-memory store
- [x] Existing `toctou_picks_up_concurrent_writer` still green (the parser fix keeps it honest).

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -275,35 +275,17 @@ impl AppBuilder {
         let store = crate::secrets::create_secrets_store(crypto, handles);
 
         // Safety gate: if we auto-generated a fresh master key this run
-        // (no env var, no keychain entry, nothing on disk), but the
-        // secrets table already carries rows from a prior key, those
-        // rows are undecryptable and silently continuing would shadow
-        // unrecoverable data. Fail loudly so the user can restore the
-        // original key before any new writes pile on top.
-        if self.config.secrets.generated
-            && let Some(ref secrets) = store
-        {
-            match secrets.any_exist().await {
-                Ok(true) => {
-                    return Err(anyhow::anyhow!(
-                        "Secrets store already contains encrypted data, but IronClaw \
-                         auto-generated a new master key because no SECRETS_MASTER_KEY \
-                         env var and no OS-keychain entry were available. The existing \
-                         rows were encrypted with a different key and cannot be \
-                         decrypted. Restore the original key (set SECRETS_MASTER_KEY or \
-                         re-populate the keychain entry) before restarting. If the \
-                         existing data is expendable, clear the `secrets` table first."
-                    ));
-                }
-                Ok(false) => {}
-                Err(e) => {
-                    tracing::warn!(
-                        "Unable to probe secrets store for preexisting data during \
-                         post-generate safety check: {e}. Proceeding — loss-of-data \
-                         risk exists if the store is non-empty."
-                    );
-                }
-            }
+        // but the secrets table already carries rows from a prior key,
+        // those rows are undecryptable and silently continuing would
+        // shadow unrecoverable data. Fail loudly (and fail-closed on
+        // probe error) so the user can restore the original key before
+        // any new writes pile on top.
+        if let Some(ref secrets) = store {
+            crate::secrets::verify_generated_key_safe(
+                self.config.secrets.generated,
+                secrets.as_ref(),
+            )
+            .await?;
         }
 
         if let Some(ref secrets) = store {
--- a/src/app.rs
+++ b/src/app.rs
@@ -280,12 +280,28 @@ impl AppBuilder {
         // shadow unrecoverable data. Fail loudly (and fail-closed on
         // probe error) so the user can restore the original key before
         // any new writes pile on top.
-        if let Some(ref secrets) = store {
-            crate::secrets::verify_generated_key_safe(
+        //
+        // Roll back the persistence `auto_generate_and_persist` already
+        // committed: otherwise a subsequent restart would read the
+        // newly-written key as `source = Env/Keychain, generated =
+        // false`, skip this gate, and silently accept the wrong key.
+        // Rollback keeps the gate re-firing on every start until the
+        // user restores the real key or clears the stale rows.
+        if let Some(ref secrets) = store
+            && let Err(gate_err) = crate::secrets::verify_generated_key_safe(
                 self.config.secrets.generated,
                 secrets.as_ref(),
             )
-            .await?;
+            .await
+        {
+            if self.config.secrets.generated {
+                crate::secrets::rollback_generated_key_persistence(
+                    self.config.secrets.source,
+                    &crate::bootstrap::ironclaw_env_path(),
+                )
+                .await;
+            }
+            return Err(gate_err.into());
         }
 
         if let Some(ref secrets) = store {
--- a/src/app.rs
+++ b/src/app.rs
@@ -274,6 +274,38 @@ impl AppBuilder {
         let handles = self.handles.as_ref().unwrap_or(&empty_handles);
         let store = crate::secrets::create_secrets_store(crypto, handles);
 
+        // Safety gate: if we auto-generated a fresh master key this run
+        // (no env var, no keychain entry, nothing on disk), but the
+        // secrets table already carries rows from a prior key, those
+        // rows are undecryptable and silently continuing would shadow
+        // unrecoverable data. Fail loudly so the user can restore the
+        // original key before any new writes pile on top.
+        if self.config.secrets.generated
+            && let Some(ref secrets) = store
+        {
+            match secrets.any_exist().await {
+                Ok(true) => {
+                    return Err(anyhow::anyhow!(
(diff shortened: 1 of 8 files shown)


src/config/secrets.rs
```diff
@@ -50,45 +57,86 @@ impl SecretsConfig {
...
+    /// machines.
+    async fn resolve_inner(
+        keychain_key: Option<Vec<u8>>,
+        env_path: &Path,
+        allow_keychain_persist: bool,
+    ) -> Result<Self, ConfigError> {
+        use crate::settings::KeySource;
 
-        if let Ok(key_bytes) = crate::secrets::keychain::get_master_key().await {
+        if let Some(key_bytes) = keychain_key {
```
serrrfirat: **High Severity**\n\nChanging `SecretsConfig::resolve()` to prefer the keychain here makes startup disagree with other active master-key readers that still prefer `SECRETS_MASTER_KEY` first (for example `SetupWizard::step_security`, `SetupWizard::init_secrets_context`, `crate::secrets::resolve_master_key`, and `cli import`). On a machine where both keys exist, onboarding/import can save new secrets under the env key, but the next normal startup will boot with the keychain key and fail to decrypt those newly-written rows.\n\nPlease funnel every read/write path through one canonical precedence helper before flipping the global order, and add a caller-level regression test that exercises onboarding/import plus a subsequent startup with both keys present.

src/config/secrets.rs
```diff
@@ -131,74 +179,163 @@ impl SecretsConfig {
...
 }
 
+/// Extract a valid 32-byte hex `SECRETS_MASTER_KEY` value from an
+/// ironclaw `.env` file. Returns `None` if the file is missing, the
+/// key isn't present, or the value isn't exactly 64 hex characters.
+/// Used for TOCTOU detection in [`SecretsConfig::auto_generate_and_persist`].
+fn read_secrets_master_key(env_path: &Path) -> Option<String> {
+    let contents = std::fs::read_to_string(env_path).ok()?;
+    for line in contents.lines() {
+        let (key, value) = line.split_once('=')?;
```
serrrfirat: **Medium Severity**

`line.split_once('=')?` exits the whole function on the first non-`KEY=...` line, not just the current iteration. A perfectly normal `.env` that starts with a blank line or `# comment` will therefore make this helper return `None` before it ever reaches `SECRETS_MASTER_KEY`, which means the new TOCTOU re-check is skipped and the later write can still overwrite the concurrent writer's key.

Please `continue` on non-assignment lines instead of `?`, and add a regression test with a commented/blank-prefixed `.env` file.

src/config/secrets.rs
ilblackdragon: Thanks — you're right, and the probe-order flip is the real bug. Reverted in 445798dc:

- `SecretsConfig::resolve_inner` is back to env-first, keychain-second, auto-generate third — now matching `SetupWizard::step_security`, `SetupWizard::init_secrets_context`, `crate::secrets::resolve_master_key`, and `cli import`. No more onboarding-vs-startup key mismatch.
- `resolve_with_env_path` now probes the keychain only when `SECRETS_MASTER_KEY` is unset, so an explicit env var no longer triggers a macOS Keychain Access dialog on every startup.
- Added caller-level coverage: `env_wins_over_keychain_when_both_present` (was `keychain_wins_over_env...`), `keychain_wins_when_env_unset`, and `generated_flag_tracks_auto_generate_path`.

While I was here, added a related safety gate you didn't ask for but seemed worth it: if `resolve` falls through to auto-generate, `SecretsConfig.generated = true`, and `AppBuilder::init_secrets` calls a new `SecretsStore::any_exist()` and aborts when a freshly-generated key meets a populated secrets table — otherwise we'd silently shadow undecryptable data. Happy to pull that into a separate PR if you'd prefer.

src/config/secrets.rs
ilblackdragon: Confirmed — real bug. Fixed in 445798dc: non-`KEY=VALUE` lines now `continue` instead of aborting the whole scan, so a `.env` with a leading blank line or comment no longer silently defeats the TOCTOU re-check. Added `read_secrets_master_key_skips_blank_and_comment_lines` as a regression test.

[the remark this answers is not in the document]
ilblackdragon: Fixed in 445798dc — `split_once('=')` now uses a `let Some(_) = … else { continue }` so blank and comment lines don't abort the scan before we reach `SECRETS_MASTER_KEY`. Added `read_secrets_master_key_skips_blank_and_comment_lines` as a regression test.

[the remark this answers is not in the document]
ilblackdragon: Fixed in 445798dc — switched to `let Some((key, value)) = line.split_once('=') else { continue }` so non-assignment lines are skipped instead of aborting the whole file scan. Regression test: `read_secrets_master_key_skips_blank_and_comment_lines`.

src/app.rs
```diff
@@ -274,6 +274,20 @@ impl AppBuilder {
...
+        // shadow unrecoverable data. Fail loudly (and fail-closed on
+        // probe error) so the user can restore the original key before
+        // any new writes pile on top.
+        if let Some(ref secrets) = store {
+            crate::secrets::verify_generated_key_safe(
+                self.config.secrets.generated,
+                secrets.as_ref(),
+            )
+            .await?;
+        }
```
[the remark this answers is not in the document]
ilblackdragon: Good catch — fixed in 6bc76373. Added `crate::secrets::rollback_generated_key_persistence(source, env_path)` (plus `bootstrap::remove_bootstrap_var_to` to support the `.env` arm) and wired it into `AppBuilder::init_secrets`: when the safety gate rejects a generated key, we delete the freshly-written keychain entry or strip the `SECRETS_MASTER_KEY=` line from `.env` before propagating the error.

Rollback is best-effort (failures warn and swallow, since the gate's abort is the primary signal), and it fires on `ProbeFailed` too — fail-closed behavior means we don't know the store is safe, so leaving the stray key would let a subsequent restart silently proceed if the DB hiccup clears. Idempotent on missing files, so the gate keeps re-firing until the user restores the real key or clears the stale rows.

New tests: `rollback_removes_generated_env_key`, `rollback_tolerates_missing_env_file`, `rollback_with_source_none_is_a_noop`.

PR #2658, opened 2026-04-19
fix(routine_engine): finalize orphaned runs when linked job is Stuck or exceeds dispatch timeout

### LINKED ISSUE #2655: sync_dispatched_runs never finalizes routine runs when linked job stays in pending state
## Summary

`sync_dispatched_runs` in [`src/agent/routine_engine.rs`](https://github.com/nearai/ironclaw/blob/ironclaw-v0.25.0/src/agent/routine_engine.rs#L510-L600) maps only terminal job states to run outcomes:

```rust
let final_status = match job.state {
    JobState::Completed | JobState::Submitted | JobState::Accepted => Some(RunStatus::Ok),
    JobState::Failed    | JobState::Cancelled                      => Some(RunStatus::Failed),
    // Pending, InProgress, Stuck — still running
    _ => None,
};
```

If a job never transitions out of `pending` (worker crash, sandbox unavailable, transient scheduler-wedge, etc.), the corresponding routine run stays `status='running'` indefinitely. Combined with the common `max_concurrent=1`, that permanently prevents the routine from firing again.

## Observed

Running ironclaw v0.25.0 (self-hosted, libsql backend). Two `full_job` routines (`email-triage`, `hearst-pipeline-health`) each ended up with `routine_runs` entries in `status='running'` whose linked `agent_jobs` rows were `status='pending'` with no updates. One run was stuck for **3 days**. Every 15s, the info log:

```
INFO Recovering 1 orphaned dispatched routine runs
```

— but no progress, because the reaper never transitions runs whose linked job is still `pending`. Manually `UPDATE routine_runs SET status='failed' WHERE id=...` was the only way to unblock; a service restart alone did not help (`sync_dispatched_runs` only considers runs started before `boot_time`, and those started after boot are watched by an in-process `FullJobWatcher` that was itself wedged).

## Suggested fix

Two small changes, both localised to `sync_dispatched_runs`:

1. **Treat `JobState::Stuck` as terminal-failure.** `self_repair` already uses `Stuck` as *its* terminal diagnostic ("exceeded stuck_threshold", "timed out") — it is semantically a failure, not an ongoing state. Moving it out of the fall-through arm and into `RunStatus::Failed` closes the case where self_repair's watchdog has fired but the run reaper still considers the run active.

2. **Add a wall-clock escape hatch.** If a run's `started_at` is older than some threshold (e.g. 1h, config-overridable) and the linked job is still non-terminal (particularly `Pending`/`InProgress`), force-fail the run with a clear reason (`"Linked job stuck in state X past timeout"`). This handles the case where self_repair never fires because the job never reaches InProgress (e.g. it stays in `pending` waiting for a worker that will never come).

Happy to send a PR if the approach is agreed. For now I'll carry a local patch.

## Environment

- ironclaw v0.25.0
- libsql (single-file sqlite) backend
- system-level systemd unit
- Ubuntu 24.04

octo-patch:
Fixes #2655

## Problem

`sync_dispatched_runs` was treating `JobState::Stuck` as an active (non-terminal) state alongside `Pending` and `InProgress`. When a routine's linked job was diagnosed as Stuck by `self_repair` but the original watcher process had already crashed, the orphaned run stayed in `status='running'` forever. Combined with the common `max_concurrent=1` guard, this permanently prevented the routine from firing again.

A related bug: a job that never leaves `Pending` (worker crash, sandbox unavailable, scheduler wedge) causes the same indefinite block since the run reaper never transitions such runs.

## Solution

Two targeted changes to `sync_dispatched_runs`, both scoped to the orphaned-run recovery path (runs whose `started_at` predates the current process `boot_time`):

1. **Treat `JobState::Stuck` as terminal failure.** `self_repair` already uses `Stuck` as its terminal diagnostic. Moving it from the "still running" arm into `RunStatus::Failed` means an orphaned run whose job was already diagnosed as stuck gets cleaned up on the next tick.

2. **Wall-clock escape hatch for `Pending`/`InProgress` jobs.** If an orphaned run's `started_at` is older than `DISPATCH_RUN_TIMEOUT_HOURS` (1 h) and its linked job is still in a non-terminal state, the run is force-failed with a `WARN`-level log entry. This handles permanent scheduler wedges where a job never reaches `InProgress` at all.

Active runs in the current process are watched by their own `FullJobWatcher` instances and are not affected by either change.

## Testing

- Updated `test_job_state_to_run_status_mapping` to assert that `JobState::Stuck` maps to `RunStatus::Failed` (regression guard for this fix).
- `cargo check` passes cleanly.
- Both changes are confined to the orphaned-run recovery path and do not touch the `FullJobWatcher` in-process path.

### DIFF
--- a/src/agent/routine_engine.rs
+++ b/src/agent/routine_engine.rs
@@ -1119,9 +1119,21 @@ impl FullJobWatcher {
 
     fn map_job_state(state: &crate::context::JobState) -> RunStatus {
         use crate::context::JobState;
+        // Mirror the terminal-state mapping in `sync_dispatched_runs` so that
+        // both the live-watcher path and the orphaned-run recovery path agree
+        // on what `Stuck` means. We deliberately do NOT extract a shared
+        // helper (per repo guidance) because the call sites operate on
+        // different sources (live job poll vs orphan-recovery scan) and have
+        // different non-terminal handling — only the terminal mapping needs
+        // to stay aligned.
+        //
+        // The caller checks `is_parallel_blocking()` before invoking this, so
+        // Pending/InProgress should not normally reach here. They map to
+        // Failed as a safety net rather than the previous silent Ok fallback.
         match state {
-            JobState::Failed | JobState::Cancelled => RunStatus::Failed,
-            _ => RunStatus::Ok, // Completed / Submitted / Accepted
+            JobState::Completed | JobState::Submitted | JobState::Accepted => RunStatus::Ok,
+            JobState::Failed | JobState::Cancelled | JobState::Stuck => RunStatus::Failed,
+            JobState::Pending | JobState::InProgress => RunStatus::Failed,
         }
     }
 }
@@ -2737,12 +2749,14 @@ mod tests {
         assert!(msg.contains("Docker is not available"));
     }
 
-    /// Regression test for #1317: FullJobWatcher maps terminal job states correctly.
+    /// Regression test for #1317 + #2655: FullJobWatcher maps terminal job
+    /// states correctly, and Stuck is treated as Failed (matching the orphan
+    /// recovery path in sync_dispatched_runs).
     #[test]
     fn test_full_job_watcher_state_mapping() {
         use crate::context::JobState;
 
-        // Failed/Cancelled → RunStatus::Failed
+        // Failed/Cancelled/Stuck → RunStatus::Failed
         assert_eq!(
             super::FullJobWatcher::map_job_state(&JobState::Failed),
             RunStatus::Failed
@@ -2751,14 +2765,25 @@ mod tests {
             super::FullJobWatcher::map_job_state(&JobState::Cancelled),
             RunStatus::Failed
         );
+        // Stuck must map to Failed in BOTH the live-watcher path and the
+        // orphan-recovery path. Previously this fell through to the `_` arm
+        // and silently became Ok, which was the inconsistency flagged in #2658.
+        assert_eq!(
+            super::FullJobWatcher::map_job_state(&JobState::Stuck),
+            RunStatus::Failed
+        );
 
-        // All other non-active states → RunStatus::Ok
+        // Successful terminal states → RunStatus::Ok
         assert_eq!(
             super::FullJobWatcher::map_job_state(&JobState::Completed),
             RunStatus::Ok
         );
         assert_eq!(
             super::FullJobWatcher::map_job_state(&JobState::Accepted),
+            RunStatus::Ok
+        );
+        assert_eq!(
+            super::FullJobWatcher::map_job_state(&JobState::Submitted),
             RunStatus::Ok
         );
     }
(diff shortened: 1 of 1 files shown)


octo-patch:
Thanks for the careful review @gemini-code-assist! Pushed [37df951](https://github.com/octo-patch/ironclaw/commit/37df951) addressing the first and third points; quick notes on each:

**1. JobState mapping consistency** ✅
You're right that `FullJobWatcher::map_job_state` silently dropped `Stuck` into the `_ => Ok` arm. I made it explicit so the live-watcher path agrees with the orphan-recovery path on terminal states (Failed/Cancelled/**Stuck** → Failed; Completed/Submitted/Accepted → Ok), with Pending/InProgress as a defensive Failed since the caller already gates on `is_parallel_blocking()`. Per the repo guideline I did **not** extract a shared helper — the two paths diverge in their non-terminal handling (live polling vs orphan scan + wall-clock) so only the terminal arm is kept aligned, with cross-referencing comments.

**2. 1h timeout for InProgress jobs** ⚠️ acknowledged, kept as-is for now
The wall-clock branch only fires for **orphaned** runs whose `started_at` predates this process's `boot_time` — i.e. runs whose original watcher is already gone. A legitimate long-running InProgress job that survives a process restart will be re-watched by self_repair / FullJobWatcher in the new process and won't take this path. I'd rather not split the constant pre-emptively without a concrete report of a > 1 h job being incorrectly killed; happy to revisit if that lands as a follow-up issue.

**3. Test for wall-clock timeout logic** ✅
Added `test_dispatch_run_wallclock_timeout` which exercises the same decision logic with controlled `started_at` values: recent Pending/InProgress stay active, stale ones (past `DISPATCH_RUN_TIMEOUT_HOURS + 5min`) force-fail, and terminal states ignore age. Kept it self-contained per the repo's "prefer self-contained code in tests" guideline. All 34 routine_engine tests pass locally.

octo-patch:
Quick CI follow-up: pushed [65457f8](https://github.com/octo-patch/ironclaw/commit/65457f87) which runs `cargo fmt --all` over `src/agent/routine_engine.rs`. The previous commit left three `JobState::Failed | Cancelled | Stuck => { Some(RunStatus::Failed) }` blocks in the multi-line form, but rustfmt prefers the single-line `=> Some(RunStatus::Failed),` for short bodies. No behavior change — Formatting / Code Style should be green now.

serrrfirat:
Thank you for fixing the orphaned routine-run failure mode. We are closing this PR because Reborn handles abandoned execution through the shared turn-run lease contract rather than routine-specific job-state reconciliation.

Every claimed Reborn run has a bounded runner lease. `recover_expired_leases` in `crates/ironclaw_turns` terminally fails an abandoned running run with sanitized `lease_expired` state, or finishes an abandoned `CancelRequested` run as cancelled, while releasing the active lock and runner claim. The trigger poller then observes the canonical run state through `TriggerActiveRunLookup`; it does not need a boot-time distinction or an arbitrary one-hour dispatch timeout.

The operational failure you identified remains important. We would welcome your help strengthening Reborn lease-expiry and trigger-active-run recovery tests, especially across process restart and both durable backends. Thank you for tracking down this permanent-blocking case.

PR #2668, opened 2026-04-19
fix(engine): recover negative max_tokens overflows across bridge/orchestrator

serrrfirat:
## Summary
- classify NEAR AI / OpenAI-compatible negative `max_tokens` 400s as `ContextLengthExceeded` with intentionally narrow matching
- preserve structured overflow handling across the bridge and engine/orchestrator retry + compaction paths, including the `(0, 0)` token-count fallback case
- emit an orchestrator trace event when overflow retry compaction rewrites `working_messages`, with regression coverage for retry/state persistence behavior

## Testing
- `cargo fmt --all`
- `cargo test -p ironclaw_engine llm_complete_retries_when_overflow_counts_are_unknown`
- `cargo test -p ironclaw_engine execute_orchestrator_persists_compacted_messages_after_retry`
- `cargo test -p ironclaw test_send_request_maps_negative_max_tokens_400_to_context_length_exceeded`
- `cargo test --lib` *(still fails locally on the pre-existing unrelated `pairing::approval::tests::propagate_approval_restores_runtime_state_when_on_start_fails` fixture issue)*

### DIFF
--- a/src/llm/nearai_chat.rs
+++ b/src/llm/nearai_chat.rs
@@ -308,8 +308,12 @@ impl NearAiChatProvider {
                 return Err(LlmError::ContextLengthExceeded { used, limit });
             }
 
-            // Some providers return 400 with "context_length_exceeded" in the body
-            // (e.g., OpenAI-compatible endpoints behind NEAR AI).
+            // Some providers return 400 with context-overflow text in the body
+            // (e.g., OpenAI-compatible endpoints behind NEAR AI). Some OpenAI-
+            // compatible backends also compute a negative remaining output budget
+            // when the prompt already exceeds the context window, surfacing that as
+            // "max_tokens must be at least 1, got -123" instead of the more common
+            // "context_length_exceeded" wording.
             if status_code == 400 {
                 let lower = response_text.to_ascii_lowercase();
                 const CONTEXT_PATTERNS: &[&str] = &[
@@ -317,6 +321,7 @@ impl NearAiChatProvider {
                     "maximum context length",
                     "too many tokens",
                     "payload too large",
+                    "max_tokens must be at least 1, got -",
                 ];
                 if CONTEXT_PATTERNS.iter().any(|p| lower.contains(p)) {
                     let (used, limit) = crate::llm::rig_adapter::parse_token_counts(&lower);
@@ -1167,6 +1172,11 @@ fn parse_usage(usage: Option<&ChatCompletionUsage>) -> (u32, u32) {
 mod tests {
     use super::*;
     use crate::llm::session::SessionConfig;
+    use axum::{
+        Router,
+        http::StatusCode,
+        routing::{get, post},
+    };
     use rust_decimal_macros::dec;
 
     fn test_nearai_config(base_url: &str) -> NearAiConfig {
@@ -1190,6 +1200,37 @@ mod tests {
 
     fn test_session() -> Arc<SessionManager> {
         Arc::new(SessionManager::new(SessionConfig::default()))
+    }
+
+    async fn spawn_nearai_test_server(
+        chat_status: StatusCode,
+        chat_body: &'static str,
+    ) -> (String, tokio::task::JoinHandle<()>) {
+        let app = Router::new()
+            .route(
+                "/v1/chat/completions",
+                post({
+                    let body = chat_body.to_string();
+                    move || {
+                        let body = body.clone();
+                        async move { (chat_status, body) }
+                    }
+                }),
+            )
+            .route(
+                "/v1/model/list",
+                get(|| async { (StatusCode::NOT_FOUND, "") }),
+            );
+
+        let listener = tokio::net::TcpListener::bind("127.0.0.1:0")
+            .await
+            .expect("bind test server");
+        let addr = listener.local_addr().expect("test server addr");
+        let handle = tokio::spawn(async move {
+            axum::serve(listener, app).await.expect("serve test app");
+        });
+
+        (format!("http://{}", addr), handle)
     }
 
     #[test]
@@ -1903,6 +1944,29 @@ mod tests {
         unsafe {
             std::env::remove_var("NEARAI_API_KEY");
         }
+    }
+
+    #[tokio::test]
+    async fn test_send_request_maps_negative_max_tokens_400_to_context_length_exceeded() {
+        let (base_url, server) = spawn_nearai_test_server(
+            StatusCode::BAD_REQUEST,
+            r#"{"error":{"message":"max_tokens must be at least 1, got -139081"}}"#,
+        )
+        .await;
+        let provider = NearAiChatProvider::new(test_nearai_config(&base_url), test_session())
+            .expect("provider");
+
+        let err = provider
+            .send_request::<_, serde_json::Value>(&serde_json::json!({"ping": true}))
+            .await
+            .expect_err("negative max_tokens overflow should map to context-length error");
+
+        server.abort();
+
+        assert!(matches!(
+            err,
+            LlmError::ContextLengthExceeded { used: 0, limit: 0 }
+        ));
     }
 
     #[test]
--- a/src/llm/nearai_chat.rs
+++ b/src/llm/nearai_chat.rs
@@ -313,7 +313,10 @@ impl NearAiChatProvider {
             // compatible backends also compute a negative remaining output budget
             // when the prompt already exceeds the context window, surfacing that as
             // "max_tokens must be at least 1, got -123" instead of the more common
-            // "context_length_exceeded" wording.
+            // "context_length_exceeded" wording. Keep this match intentionally
+            // narrow so unrelated 400s do not trigger overflow recovery; if token
+            // counts are absent, downstream retry heuristics still key off the
+            // structured `ContextLengthExceeded` signal rather than these numbers.
             if status_code == 400 {
                 let lower = response_text.to_ascii_lowercase();
                 const CONTEXT_PATTERNS: &[&str] = &[
--- a/crates/ironclaw_engine/orchestrator/default.py
+++ b/crates/ironclaw_engine/orchestrator/default.py
@@ -678,6 +678,10 @@ def run_loop(context, goal, actions, state, config):
         # 4. Call LLM
         __emit_event__("step_started", step=step)
         response = __llm_complete__(working_messages, actions, None)
+        compacted_messages = response.get("compacted_messages")
+        if isinstance(compacted_messages, list):
+            state["working_messages"] = compacted_messages
+            working_messages = state["working_messages"]
         __emit_event__("step_completed", step=step,
                        input_tokens=response.get("usage", {}).get("input_tokens", 0),
                        output_tokens=response.get("usage", {}).get("output_tokens", 0))
--- a/crates/ironclaw_engine/orchestrator/default.py
+++ b/crates/ironclaw_engine/orchestrator/default.py
@@ -682,6 +682,11 @@ def run_loop(context, goal, actions, state, config):
         if isinstance(compacted_messages, list):
             state["working_messages"] = compacted_messages
             working_messages = state["working_messages"]
+            __emit_event__(
+                "message_added",
+                role="System",
+                content_preview="overflow retry compacted working_messages",
+            )
         __emit_event__("step_completed", step=step,
                        input_tokens=response.get("usage", {}).get("input_tokens", 0),
                        output_tokens=response.get("usage", {}).get("output_tokens", 0))
--- a/crates/ironclaw_engine/src/executor/orchestrator.rs
+++ b/crates/ironclaw_engine/src/executor/orchestrator.rs
@@ -39,7 +39,7 @@ use crate::traits::llm::{LlmBackend, LlmCallConfig};
 use crate::traits::store::Store;
 use crate::types::error::EngineError;
 use crate::types::event::{EventKind, ThreadEvent, summarize_params};
-use crate::types::message::ThreadMessage;
+use crate::types::message::{MessageRole, ThreadMessage};
 use crate::types::project::ProjectId;
 use crate::types::shared_owner_id;
 use crate::types::step::{ActionCall, StepId, TokenUsage};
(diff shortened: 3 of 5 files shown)


src/llm/nearai_chat.rs
```diff
@@ -308,15 +308,20 @@ impl NearAiChatProvider {
...
+            // "max_tokens must be at least 1, got -123" instead of the more common
+            // "context_length_exceeded" wording.
             if status_code == 400 {
                 let lower = response_text.to_ascii_lowercase();
                 const CONTEXT_PATTERNS: &[&str] = &[
                     "context_length_exceeded",
                     "maximum context length",
                     "too many tokens",
                     "payload too large",
+                    "max_tokens must be at least 1, got -",
```
serrrfirat, line 327: **High Severity**

This only fixes the v1 `LlmError` classification path; it does not fix the reported orchestrator / Engine v2 failure path.

`NearAiChatProvider` now returns `ContextLengthExceeded` here, but `src/bridge/llm_adapter.rs:80-82` and `:120-122` immediately stringify provider errors into `EngineError::Llm { reason: e.to_string() }`. The engine side has no structured context-overflow variant or retry handling for that condition, and thread compaction is still disabled by default in `crates/ironclaw_engine/src/types/thread.rs:173-175`.

So the user-visible orchestrator/runtime failure can still surface instead of auto-recovering; this change mainly helps the v1 dispatcher path.

Suggested fix: preserve a structured context-overflow signal across the bridge (or add an engine-side classifier/handler) and cover the orchestrator path with a caller-level regression test.

src/llm/nearai_chat.rs
```diff
@@ -1905,6 +1946,29 @@ mod tests {
         }
     }
 
+    #[tokio::test]
+    async fn test_send_request_maps_negative_max_tokens_400_to_context_length_exceeded() {
```
serrrfirat, line 1953: **Medium Severity**

This regression test only exercises the `send_request()` helper. The behavior this PR is trying to protect is the caller-visible recovery path that this classification unlocks (compaction / retry or, for engine v2, at least the propagated error semantics).

Per the repo testing rule, helper-level coverage is not enough when the helper gates a side effect through one or more wrappers. A future refactor could keep this test green while the real caller path is still broken.

Suggested fix: add a test that drives the actual call site (`dispatcher` for v1, and/or `LlmBridgeAdapter` / engine path for v2) and proves the overflow is handled the way the product expects.

serrrfirat:
Addressed the review findings in 432ea4b4.

What changed:
- preserve `ContextLengthExceeded` across `src/bridge/llm_adapter.rs` as `EngineError::TokenLimitExceeded`
- retry orchestrator `__llm_complete__` calls once with compacted explicit transcripts on token-limit overflow
- persist compacted `working_messages` back into the Python orchestrator state
- add caller-level regression coverage for the bridge and engine/orchestrator paths

Local tests run:
- `cargo test -p ironclaw --lib context_overflow_to_engine_token_limit`
- `cargo test -p ironclaw --lib llm::nearai_chat::tests::test_send_request_maps_negative_max_tokens_400_to_context_length_exceeded -- --exact`
- `cargo test -p ironclaw_engine executor::orchestrator::tests::llm_complete_retries_after_context_overflow_with_compacted_messages -- --exact`
- `cargo test -p ironclaw_engine executor::orchestrator::tests::execute_orchestrator_persists_compacted_messages_after_retry -- --exact`

CI is still running on the updated commit.

serrrfirat:
Final update on this PR:

- Preserved `ContextLengthExceeded` across `src/bridge/llm_adapter.rs` by mapping it to `EngineError::TokenLimitExceeded`
- Added engine/orchestrator recovery for explicit `__llm_complete__` overflows in `crates/ironclaw_engine/src/executor/orchestrator.rs`
- Persisted compacted `working_messages` back into Python orchestrator state in `crates/ironclaw_engine/orchestrator/default.py`
- Added caller-level regression coverage for the bridge and orchestrator retry/persistence paths
- Fixed follow-up feature-gated build issues uncovered by CI

Validation:
- Local targeted tests passed for `nearai_chat`, bridge mapping, and engine/orchestrator retry flows
- PR checks are green on `870439e7`
- Previous review threads are resolved

Leaving the PR open for now pending explicit merge approval.

zmanian (commented):
## Code Review

Scope note: the PR title advertises a small classification fix in `nearai_chat.rs`, but the diff is +590/-72 and actually delivers a larger four-layer change (provider pattern match -> bridge error mapping -> orchestrator compaction+retry -> Python state round-trip). Consider retitling or splitting — reviewers approaching this as a one-line pattern add will miss the orchestrator retry semantics. This also addresses the two self-filed review comments on the PR (v2 bridge mapping + caller-level regression test), so those can be resolved.

### High-severity findings
None blocking. The structured error propagation through `LlmBridgeAdapter::map_provider_error` and the orchestrator retry path are the right shape.

### Medium-severity

1. **Retry guard silently skips non-default orchestrators.** `handle_llm_complete` only retries when `has_explicit_messages` is true. The bundled `default.py` sets `state[\"working_messages\"]`, so it benefits; but any custom orchestrator script that lets the engine fall back to `thread.messages` gets the pre-PR behavior (hard-fail on overflow) with no log line explaining why. At minimum add a `tracing::debug!` on the skip branch noting \"overflow retry skipped: no explicit messages\". Better: also compact `thread.messages` when that's the source — the two branches should behave symmetrically.

2. **`compact_thread_messages_for_retry` can drop a turn's action_call/action_result pairing silently.** The helper keeps only `System` messages before the *last* `User` and then everything from the last User onward. If the current turn has already produced an Assistant(ActionCalls) + ActionResult pair after the last User message, the pairing is preserved — good. But if the \"last User\" rposition points at a user message nested inside an ongoing tool sequence (rare but possible when the UI injects a mid-step user message), earlier Assistant+ActionResult pairs from the same logical turn are dropped, and the retry transcript starts with orphaned action_call_ids. Worth either (a) asserting the invariant (last User is at a turn boundary) or (b) scanning forward to ensure no dangling `action_call_id` references survive. Current tests don't cover this shape.

3. **`estimate_retry_tokens` counts bytes, not chars.** `m.content.len()` returns UTF-8 byte length; for non-ASCII transcripts this overestimates. Only used as a relative guard (`tokens_after >= tokens_before`), so it's load-bearing only when bytes and chars diverge between before/after — which can happen if compaction strips ASCII-heavy tool output and retains non-ASCII user text. Cheap fix: `m.content.chars().count()`, or reuse `executor::compaction::estimate_tokens` rather than duplicating the constant inline (the doc comment claims feature-gating concerns but they aren't shown — a brief justification or a link to the gate would help).

4. **One-shot retry with no backoff/loop cap.** If compaction doesn't reduce the payload below the limit on the first try, the retry fails immediately and the error bubbles up. That's probably correct behavior (don't thrash), but there's no observability tying the original `TokenLimitExceeded` to the retry failure — the outer error says \"LLM call failed after compaction retry: ...\" which loses the used/limit from the first attempt. Consider including both in the emitted error.

### Low-severity / nits

- `thread_messages_to_python_json` is a clean DRY of the previous inline closure in `build_orchestrator_inputs` — good refactor, matches the single-source-of-truth comment.
- `tracing::warn!` is used for the compaction path. Per `CLAUDE.md` (\"Background tasks must NEVER use `info!` — it breaks the interactive display\"), `warn!` during automatic recovery that the user doesn't need to see could also be noisy in the REPL/TUI. Consider `debug!` since auto-compaction is a normal recovery, not a user-facing anomaly.
- The test `llm_complete_retries_after_context_overflow_with_compacted_messages` uses the sentinel string `\"automatically compacted to fit within the context window\"` in three places (production note text + two assertions). If the note copy changes, tests silently drift. Extract a `const` in the production module and reference it from tests.
- `LlmError::ContextLengthExceeded { used: 0, limit: 0 }` from the negative-max_tokens parse loses the actual numbers present in the response (\"-139081\"). Not a regression vs. prior behavior (other 400 patterns also yield 0/0), but a future improvement would be to parse the negative value and surface `used = limit + |negative|` as a lower-bound estimate.

### Conventions check

- `thiserror` usage: fine (`EngineError::TokenLimitExceeded` already exists).
- No new `.unwrap()` / `.expect()` in production code. Tests use them, which is permitted.
- Dual-backend DB: no persistence changes, N/A.
- Dispatcher rule: the new code is engine-internal, doesn't bypass `ToolDispatcher`.

### Verdict
COMMENT — the design is right and the coverage across layers materially improves on the author's own self-flagged concerns. Address the medium items before merge (especially #1 and #2); the lows are cleanup.

serrrfirat:
Addressed @zmanian's review in `7b183daf`:

- retry overflow recovery now also works when `__llm_complete__` falls back to `thread.messages`, not just explicit `working_messages`
- retry compaction now preserves in-flight assistant/action-result context around nested user injections so surviving `action_call_id`s are not orphaned
- retry token estimation now counts Unicode chars instead of UTF-8 bytes
- retry failure errors now include the original overflow `used/limit` alongside the retry error
- downgraded the automatic recovery log from `warn!` to `debug!`
- extracted the compaction-note sentinel into a shared constant and added regression coverage for the fallback retry / nested tool-sequence / unicode estimation / retry-error cases

Validation run locally:
- `cargo fmt`
- targeted `ironclaw_engine` orchestrator regression tests

`gh pr checks 2668` is green on `7b183daf`.

zmanian (commented):
Reviewed the full diff (842/-72). The advertised one-line classification fix is sound, but this PR actually bundles a much larger orchestrator retry-with-compaction feature. Verdict: classification change LGTM, broader orchestrator retry logic has a few concerns worth addressing.

**Classification fix (nearai_chat.rs)**
- Substring pattern `\"max_tokens must be at least 1, got -\"` is reasonable and matches the observed NEAR AI / OpenAI-compatible error shape. Lowercasing first is correct.
- Regression test is good (real axum server, exact body shape, asserts `ContextLengthExceeded { used: 0, limit: 0 }`).
- One gap: if the backend ever emits a slightly different casing or phrasing of this (e.g. `max_tokens: must be >= 1`), we fall back to a generic 400. Worth a brief comment that the match is intentionally narrow to avoid false positives, and that we rely on `parse_token_counts` — which by inspection here returns `(0, 0)` for the negative-got case since there is no positive limit to parse. Confirm that `(0, 0)` does not break downstream retry heuristics that compare `used` vs `limit`.

**Cross-provider consistency**
- This pattern lives only in `nearai_chat.rs`. Anthropic/Bedrock/OpenAI native paths don't need it (they surface context overflow via native error codes), so no cross-provider duplication required. Fine as-is, but a one-line pointer comment in `error.rs` noting which providers do substring classification would help the next person.

**Bridge adapter mapping (src/bridge/llm_adapter.rs)**
- `map_provider_error` correctly funnels `LlmError::ContextLengthExceeded` → `EngineError::TokenLimitExceeded`. Good separation, tested on both `complete` and `complete_with_tools`. `used as u64` / `limit as u64` is safe since source is `usize`.

**Orchestrator retry + compaction (the real bulk of this PR)**
- `compact_thread_messages_for_retry` preserves leading `System` messages, inserts a note, and keeps the last User + preceding Assistant/ActionResult slice to avoid orphaned `action_call_id`s. Logic looks right, and the two tests cover the json-messages path and the thread-messages fallback path.
- `estimate_retry_tokens` duplicates `executor::compaction::estimate_tokens` deliberately. The comment explains why, but duplicated estimation drifts; consider a `pub(crate)` re-export instead of a fork, or at minimum a `debug_assert_eq!` in a shared test to catch divergence.
- Retry is single-shot: if compaction-reduced payload still overflows, we return the retry error verbatim. That's the right call (no infinite loop), but the error message `\"LLM call failed after compaction retry: original overflow used=X limit=Y; retry failed: {retry_err}\"` should probably be classified — if `retry_err` is itself `TokenLimitExceeded`, the Python layer gets a `RuntimeError` string rather than a structured signal. Minor.
- `reduced_payload = retry_messages.len() < messages.len() || tokens_after < tokens_before` — the `<` on tokens is correct; equal-sized compaction (same message count, same tokens) correctly short-circuits. Good.
- Python side: `if isinstance(compacted_messages, list): state[\"working_messages\"] = ...` — this silently mutates working state mid-loop. Fine, but worth an event emission (`compaction_applied` or similar) so this is observable in traces. Currently invisible to the user.

**PR hygiene**
- Title/summary advertise a small fix; the diff delivers a feature. Please update the description to call out the orchestrator compaction retry — reviewers and future git-archaeologists will be confused. Consider splitting: (a) classification fix + test, (b) orchestrator retry/compaction. CI label is already XL, which is the signal.
- `cargo test` was not run locally (disk full). Please confirm CI green on both `--features integration` and default feature builds before merge, especially the two new orchestrator tests.
- No `.unwrap()` / `.expect()` in production paths — verified. Tests use them, which is fine per house rules.

Not blocking, but I'd like to see: (1) PR description updated to reflect actual scope, (2) brief confirmation that `parse_token_counts` returning `(0, 0)` is benign downstream, (3) consider emitting an event when compaction is applied. Happy to approve once the description is accurate.

serrrfirat:
Addressed the remaining review feedback in `b96a01f7`:

- updated the PR title/body to reflect the full bridge + orchestrator retry/compaction scope
- documented that the negative `max_tokens` match is intentionally narrow and that `(0, 0)` token counts are still safe downstream because retry logic keys off the structured overflow signal + transcript reduction
- added engine regression coverage proving overflow retry still runs when the provider cannot report `used/limit`
- emitted a thread event when overflow retry compaction rewrites `working_messages`, and covered that in the orchestrator regression test

Validation run locally:
- `cargo fmt --all`
- `cargo test -p ironclaw_engine llm_complete_retries_when_overflow_counts_are_unknown`
- `cargo test -p ironclaw_engine execute_orchestrator_persists_compacted_messages_after_retry`
- `cargo test -p ironclaw test_send_request_maps_negative_max_tokens_400_to_context_length_exceeded`

`cargo test --lib` still hits the pre-existing unrelated local failure in `pairing::approval::tests::propagate_approval_restores_runtime_state_when_on_start_fails` because the telegram wasm fixture is missing in this environment.

PR #2698, opened 2026-04-20
fix(security): scope orchestrator credentials to job creator (#2068)

### LINKED ISSUE #2068: ownership(orchestrator): thread job creator identity through sandbox credential lookups

zmanian:
Replaces #2381 — recreated on fresh staging to avoid merge conflicts from gateway refactor churn. Same commits, rebased cleanly.

## Summary

Addresses **#2068 only** — fixes the orchestrator cross-tenant credential fallback.

- `get_credentials_handler` now resolves the job creator's `user_id` from `job_owner_cache` (lazily populated from DB on first read) instead of a hardcoded global `config.owner_id`.
- Handler returns 403 Forbidden when the job owner cannot be resolved, when it resolves to an empty string, and when any secret lookup fails (uniform 403 avoids leaking whether a secret exists under another user's scope).
- `user_id: String` field is removed from `OrchestratorState` entirely, so the handler cannot silently fall back to a process-wide identity.

Scope note (previous revision of this body incorrectly claimed otherwise):

- **#2069** (WASM `DefaultFallback::Denied`) is already closed — resolved in an earlier PR. Not touched by this diff.
- **#2100** (broadcast-metadata legacy `"default"` fallback) is still open, but is also not touched by this diff. The `load_broadcast_metadata()` code already reads from `owner_scope_id` only; #2100 should be revisited separately and closed if stale, or a follow-up PR filed.

All three issues were bundled under the original #2381; this PR narrows the scope to match the diff.

## Files changed

- `src/orchestrator/api.rs` — credential handler rewritten, `user_id` field removed, `resolve_job_owner` helper added, regression tests added
- `src/orchestrator/mod.rs` — `user_id` removed from `OrchestratorState` construction

## Test plan

- [x] `cargo test -p ironclaw --lib orchestrator::api::tests` passes
- [x] Regression tests: `credentials_returns_403_when_job_owner_unknown`, `credentials_uses_job_creator_not_other_user` (asserts 403, not 500), plus existing happy-path test

Closes #2068

### DIFF
--- a/src/orchestrator/api.rs
+++ b/src/orchestrator/api.rs
@@ -49,12 +49,20 @@ pub struct OrchestratorState {
     pub store: Option<Arc<dyn Database>>,
     /// Encrypted secrets store for credential injection into containers.
     pub secrets_store: Option<Arc<dyn SecretsStore + Send + Sync>>,
-    /// In-memory cache of job_id → user_id for SSE scoping. Populated when
-    /// sandbox jobs are created, avoiding a DB round-trip on every job event.
+    /// In-memory cache of job_id → user_id for SSE scoping and credential
+    /// lookup. Lazily populated on first read via `resolve_job_owner`, which
+    /// falls back to `get_sandbox_job` when the entry is missing. The DB is
+    /// always the source of truth; the cache just avoids a round-trip on
+    /// every job event for long-running jobs.
     pub job_owner_cache: Arc<std::sync::RwLock<HashMap<Uuid, String>>>,
 }
 
-/// Maximum entries in the job_owner_cache before oldest entries are evicted.
+/// Maximum entries in the job_owner_cache before arbitrary entries are
+/// evicted. HashMap iteration order is unspecified, so eviction is not
+/// LRU/FIFO — for the cache's purpose (avoid a DB hit per SSE event of a
+/// live job), this is adequate: the next read of an evicted entry just
+/// refills from DB. Upgrade to `IndexMap` or `lru::LruCache` if we ever
+/// need real recency ordering.
 const MAX_JOB_OWNER_CACHE_SIZE: usize = 10_000;
 
 impl OrchestratorState {
@@ -102,11 +110,6 @@ impl OrchestratorState {
             }
         }
         cache.insert(job_id, user_id);
-    }
-
-    /// Pre-populate the cache at job creation time.
-    pub fn register_job_owner(&self, job_id: Uuid, user_id: &str) {
-        self.cache_job_owner(job_id, user_id.to_string());
     }
 }
 
@@ -426,6 +429,10 @@ async fn job_event_handler(
 
     // Broadcast via the channel (if configured).
     if let Some(ref tx) = state.job_event_tx {
+        // silent-ok: SSE broadcast is best-effort and already handles the
+        // empty-user_id case below by sending with an empty scope string,
+        // which the gateway filters out. A transient DB miss here should
+        // not block the worker from reporting progress.
         let user_id = state.resolve_job_owner(job_id).await.unwrap_or_default();
 
         if user_id.is_empty() {
@@ -509,13 +516,20 @@ async fn get_credentials_handler(
             .get_decrypted(&job_user_id, &grant.secret_name)
             .await
             .map_err(|e| {
-                tracing::debug!(
+                // Internal log is warn (not error) because the common case is
+                // a benign "user has no such secret" miss. Kept above debug so
+                // a real crypto/keychain failure surfaces in production logs —
+                // 403 telemetry alone can't distinguish "wrong tenant" from
+                // "crypto broken".
+                tracing::warn!(
                     job_id = %job_id,
                     user_id = %job_user_id,
+                    env_var = %grant.env_var,
                     "Failed to decrypt secret for credential grant: {}", e
                 );
-                // Return 403 for all secret failures to avoid leaking
-                // whether a secret exists for a different user.
+                // Wire response is 403 for all secret failures to avoid
+                // leaking whether a secret exists under a different user's
+                // scope.
                 StatusCode::FORBIDDEN
             })?;
 
@@ -971,8 +985,11 @@ mod tests {
             .unwrap();
 
         let resp = router.oneshot(req).await.unwrap();
-        // Alice has no "api_key" secret -- should fail, not leak bob's
-        assert_eq!(resp.status(), StatusCode::INTERNAL_SERVER_ERROR);
+        // Alice has no "api_key" secret -- handler must refuse (403) rather
+        // than serve bob's value. All secret-lookup failures map to FORBIDDEN
+        // so the response does not leak whether a secret exists under any
+        // other user's scope.
+        assert_eq!(resp.status(), StatusCode::FORBIDDEN);
     }
 
     // -- Job event handler tests --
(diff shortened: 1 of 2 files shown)


ilblackdragon:
## Review

### Blocker

- **`credentials_uses_job_creator_not_other_user` fails when run.** Confirmed locally: the test asserts `INTERNAL_SERVER_ERROR` (500), but the same PR rewrote the `get_decrypted` error mapping from 500 → `FORBIDDEN` (403). `cargo test` on the branch produces `left: 403, right: 500`. CI didn't catch this because required checks on this repo don't run `cargo test` on PRs (only clippy/fmt/deny). Either the assertion must be `FORBIDDEN`, or a separate 500 case (e.g. a decrypt-crypto failure) needs to be exercised. Per `.claude/rules/testing.md`, the regression test for the bug must actually fail without the fix — right now it doesn't even pass *with* the fix.

### Other issues

- **PR body overstates the scope.** The title and body say "Closes #2068, closes #2069, closes #2100" and describe WASM `DefaultFallback::Denied` and `load_broadcast_metadata` fixes, but the diff only touches the orchestrator. Grepping staging shows `DefaultFallback::Denied` already at `src/tools/wasm/wrapper.rs:1497` and the owner-scoped `load_broadcast_metadata` already in place — those were fixed in earlier PRs. The `Closes` links will auto-close real issues, so the body needs to drop the #2069 / #2100 references or state explicitly that those were already fixed and we're closing stale issues. Per `.claude/rules/review-discipline.md` ("PR Scope Discipline"), body must match diff.

- **`register_job_owner` is dead code.** The public helper is added (api.rs +95) and documented as "Pre-populate the cache at job creation time," but `grep` across the whole tree shows zero call sites. Either wire it into the job-creation path (where the cache comment says it's populated) or drop it. Without a call site, the security fix depends entirely on the DB fallback — the cache will only warm incidentally via `job_event_handler`.

- **Log level regression on crypto failure.** `Failed to decrypt secret for credential grant` dropped from `tracing::error!` to `tracing::debug!`. A real crypto/keychain failure will now be invisible in production logs. The intent (avoid leaking existence of another user's secret to log readers) is fine, but the operator-facing signal belongs at `warn!` at minimum — 403 telemetry alone won't distinguish "wrong tenant" from "crypto broken."

- **Cache eviction is not LRU/FIFO.** The comment on `MAX_JOB_OWNER_CACHE_SIZE` says "oldest entries are evicted," but `cache.keys().take(N)` on `HashMap` evicts arbitrary (hash-order) entries — there is no insertion order. Either drop the "oldest" claim in the comment or swap to an LRU/IndexMap. With 10k cap and 1k purge the practical impact is small, but the comment is wrong.

- **`unwrap_or_default()` on a missing job owner in `job_event_handler`.** Pre-existing pattern, preserved by the refactor. Per `.claude/rules/error-handling.md` this needs a `// silent-ok: <reason>` comment now that the lines are being touched.

### What's good

- Correct fail-closed model: 403 on both unresolved owner and empty resolved string.
- Extracting `resolve_job_owner` deduplicates the cache-then-DB logic between the credentials and event handlers cleanly.
- `unwrap_or_else(|e| e.into_inner())` on the `RwLock` is the right poison-tolerance pattern.
- Bounding the cache at 10k prevents unbounded memory growth in long-running orchestrators.
- Good test coverage shape: 403-when-unknown, cross-tenant isolation, happy-path retained.

### Recommendation

**Request changes.** The failing test is a blocker — fix the assertion (and run `cargo test -p ironclaw --lib orchestrator::api::tests` before merging). Clarify or correct the `Closes #2069 / #2100` claims in the body. Either wire up `register_job_owner` or remove it. The log-level downgrade and LRU comment should be addressed in the same pass.

ilblackdragon:
Pushed [592802b1](https://github.com/nearai/ironclaw/commit/592802b1) addressing the review feedback.

### Changes

- **Fixed failing test.** `credentials_uses_job_creator_not_other_user` now asserts `FORBIDDEN`, matching the same PR's change that maps all secret-lookup failures to 403. Expanded the comment to explain why uniform 403 is the correct wire response (no existence-leak under another user's scope).
- **Removed dead `register_job_owner` helper.** Zero call sites; the cache is self-warming through `resolve_job_owner`, which inserts on every DB fallback. Wiring eager registration into `ContainerJobManager::create_job` is a larger refactor — filing a follow-up if it becomes worth the cost.
- **Fixed the `job_owner_cache` doc comment.** The previous "Populated when sandbox jobs are created" was aspirational; updated to describe the lazy-on-first-read behavior that actually exists.
- **Fixed `MAX_JOB_OWNER_CACHE_SIZE` comment.** HashMap iteration order is unspecified, so eviction is not LRU/FIFO — corrected the comment and called out `IndexMap` / `lru::LruCache` as upgrade paths if recency ordering is ever needed.
- **Bumped decrypt-failure log from `debug` → `warn`.** Also added `env_var` field. Keeps the 403 wire response (no existence-leak to the caller) but restores operator visibility for real crypto/keychain failures — 403 telemetry alone can't distinguish "wrong tenant" from "crypto broken".
- **Annotated `unwrap_or_default()` in `job_event_handler` with `// silent-ok:`.** Per `.claude/rules/error-handling.md` — SSE broadcast is best-effort and the empty-user_id case is handled below.

### Scope fixes

- PR title retargeted to `fix(security): scope orchestrator credentials to job creator (#2068)`.
- PR body rewritten to call out that #2069 is already closed and #2100 isn't touched by this diff (only #2068 is actually addressed).

### Verification

- `cargo test -p ironclaw --lib orchestrator::api::tests` — 16/16 pass (including the previously-failing one)
- `cargo clippy --all --tests --examples --all-features` — clean
- `cargo fmt --check` — clean
- `scripts/pre-commit-safety.sh` — exit 0

ilblackdragon (approved):

Issue #1947, opened 2026-04-03
bug(cli): MCP test-server tool description truncation panics on multi-byte UTF-8


wakaka6:
## Description

Running `ironclaw mcp test <server>` panics when a tool's description contains multi-byte UTF-8 characters (Chinese, Japanese, emoji, etc.) and exceeds 57 bytes in length.

## Symptoms

The CLI crashes with a panic message like:
```
byte index 57 is not a char boundary; it is inside '描' (bytes 55..58)
```
No tool listing is printed — the entire command aborts.

## Steps to reproduce

1. Configure an MCP server with a tool whose `description` contains non-ASCII characters and is longer than ~20 CJK characters (> 57 bytes)
2. Run `ironclaw mcp test <server-name>`
3. The program panics when printing the tool list

## Root cause

```rust
// src/cli/mcp.rs, line ~540
let desc = if tool.description.len() > 60 {
    format!("{}...", &tool.description[..57])  // byte-index slice!
} else {
    tool.description.clone()
};
```

Byte-index slicing (`[..57]`) on a UTF-8 string panics when index 57 falls in the middle of a multi-byte character (CJK = 3 bytes, emoji = 4 bytes).

## Fix

Use char-safe truncation:
```rust
let desc: String = tool.description.chars().take(57).collect();
```

## Acceptance criteria

- [ ] `ironclaw mcp test` does not panic on tool descriptions with multi-byte UTF-8 characters
- [ ] Long descriptions are still visually truncated with `...` suffix
- [ ] Unit test with CJK/emoji description confirms no panic

Branch: `v2-architecture`

PR #2379, opened 2026-04-12
fix(mcp): normalize server names, fix UTF-8 truncation, skip auth when header set

zmanian:
## Summary

Fixes three related MCP client bugs in a single PR:

- **#2236 - Server name normalization:** Names like "My Twitter Server" are now auto-normalized to `my_twitter_server` (lowercase, spaces/special chars replaced with underscores) instead of being rejected. Applied in CLI, web gateway frontend, and exposed as `normalize_server_name()` in config.
- **#1947 - UTF-8 truncation panic:** `mcp test-server` tool description truncation used byte-index slicing (`&desc[..57]`) which panics when index 57 falls inside a multi-byte character. Now uses `chars().take(57).collect()`.
- **#1948 - Unnecessary OAuth/DCR flow:** `requires_auth()` returned `true` for all remote HTTPS servers even when the user had already configured an `Authorization` header. Now checks `has_custom_auth_header()` first and returns `false` if set.

## Test plan

- [x] `test_normalize_server_name_*` - 5 tests covering basic, uppercase, spaces/special chars, empty/all-special, and leading/trailing underscores
- [x] `test_tool_description_truncation_multibyte_utf8` - regression test with CJK characters where byte index 57 falls mid-character
- [x] `test_tool_description_truncation_ascii` - ASCII descriptions still truncate correctly
- [x] `test_requires_auth_skipped_with_custom_auth_header` - Authorization header skips auth
- [x] `test_requires_auth_skipped_with_lowercase_auth_header` - case-insensitive check
- [x] `test_requires_auth_still_true_with_non_auth_headers` - non-Authorization headers don't suppress auth
- [x] `cargo clippy --all --all-features` passes with zero warnings
- [x] All 43 MCP config tests + 9 CLI MCP tests pass

Closes #2236, closes #1947, closes #1948

Generated with [Claude Code](https://claude.com/claude-code)

PR #2699, opened 2026-04-20
fix(mcp): normalize server names, fix UTF-8 truncation, skip auth when header set (#2236, #1947, #1948)

### LINKED ISSUE #1948: bug(mcp): unnecessary OAuth/DCR flow triggered when Authorization header already set
## Description

When a user manually configures an `Authorization` header in their MCP server config, the `needs_auth()` check still returns `true` for remote HTTPS servers, triggering an unnecessary OAuth/DCR (Dynamic Client Registration) flow.

## Symptoms

- Browser opens an OAuth authorization page unexpectedly, even though a Bearer token is already configured
- Or DCR registration is attempted against the MCP server, causing connection failures or token conflicts
- Users who manage their own API keys/tokens via headers are forced through an auth flow they don't need

## Steps to reproduce

1. Configure a remote HTTPS MCP server in `~/.ironclaw/mcp.json`:
   ```json
   {
     "servers": {
       "my-server": {
         "url": "https://api.example.com/mcp",
         "headers": {
           "Authorization": "Bearer sk-my-api-key"
         }
       }
     }
   }
   ```
2. Start ironclaw and trigger tool loading from this MCP server
3. Observe that an OAuth/DCR flow is initiated despite the Bearer token being present in headers

## Root cause

```rust
// src/tools/mcp/config.rs — McpServerConfig::needs_auth()
pub fn needs_auth(&self) -> bool {
    if self.oauth.is_some() {
        return true;
    }
    // Remote HTTPS servers need auth handling...
    let url_lower = self.url.to_lowercase();
    // ... checks if URL is remote HTTPS, returns true
}
```

The function checks for `oauth` config and whether the URL is remote HTTPS, but never checks whether the user has already provided an `Authorization` header — making the auth handling redundant and intrusive.

## Fix

Add a check before the remote-HTTPS heuristic: if headers already contain an `Authorization` key (case-insensitive), return `false`.

## Acceptance criteria

- [ ] MCP servers with a manually configured `Authorization` header skip OAuth/DCR
- [ ] MCP servers without `Authorization` header still trigger auth as before
- [ ] Case-insensitive header matching (`authorization`, `Authorization`, `AUTHORIZATION`)

Related: #639 (added custom header support but did not update `needs_auth()` logic)
Branch: `v2-architecture`

### LINKED ISSUE #2236: [QA] MCP server name validation rejects uppercase and special characters
### Environment
hosted-staging (crab shack)

### Version / Commit Hash
staging HEAD as of 2026-04-09

### QA Test Date
2026-04-09

### Feature Area
Settings > MCP

### Steps to Reproduce
1. Go to Settings > MCP
2. Click "Add custom MCP server"
3. Enter a URL and a descriptive name containing uppercase letters or special characters (e.g., "My Twitter Server")

### Expected Behavior
MCP server name accepts descriptive names with mixed case and common characters.

### Actual Behavior
Validation fails with error: "Failed to add MCP server installation failed: invalid name, only lowercase letters and underscores are allowed." Users cannot use descriptive or readable names for their custom MCP servers.

### Logs / Screenshots
Screenshot in QA doc "Bug bash 4/9" item. Error message appears inline after submitting the form.

zmanian:
Replaces #2379 — recreated on fresh staging to avoid merge conflicts from gateway refactor. Same fixes, cherry-picked cleanly.

## Summary

- **#2236** — Normalize MCP server names at construction, upsert, and CLI lookup boundaries. Guards empty names. JS `normalizeMcpServerName` ported to `js/surfaces/skills.js` (new module location post-#2683).
- **#1947** — UTF-8 truncation panic in tool description rendering. Note: staging already shipped a `truncate_description()` helper on `src/cli/mcp.rs` covering this; this PR drops the duplicated inline fix and defers to the helper.
- **#1948** — Skip auth when `Authorization` header is explicitly set by the user.

## Files changed

- `crates/ironclaw_gateway/static/js/surfaces/skills.js` — `normalizeMcpServerName` helper + guard
- `src/tools/mcp/config.rs` — `normalize_server_name` + upsert-time normalization
- `src/cli/mcp.rs` — lookup-side normalization + empty-name guard
- `src/channels/web/handlers/extensions.rs` — extension install normalization
- `src/agent/thread_ops.rs`, `crates/ironclaw_tui/src/widgets/conversation.rs` — minor adjacent fixes

## Test plan

- [x] Existing `truncate_description` tests on staging already cover #1947

Closes #2236, closes #1947, closes #1948

### DIFF
--- a/crates/ironclaw_gateway/static/js/surfaces/skills.js
+++ b/crates/ironclaw_gateway/static/js/surfaces/skills.js
@@ -27,9 +27,14 @@ function installWasmExtension() {
   });
 }
 
+// Must mirror `normalize_server_name` in src/tools/mcp/config.rs, which
+// folds hyphens to underscores so the canonical form agrees with
+// `ExtensionName::new`. Treating `-` as a word char on the client would
+// let the web UI accept names (e.g. `foo--bar`, `-foo-`) that backend
+// canonicalization still rejects.
 function normalizeMcpServerName(raw) {
   return raw.toLowerCase()
-    .replace(/[^a-z0-9_-]/g, '_')
+    .replace(/[^a-z0-9_]/g, '_')
     .replace(/_+/g, '_')
     .replace(/^_|_$/g, '');
 }
--- a/crates/ironclaw_gateway/static/js/surfaces/skills.js
+++ b/crates/ironclaw_gateway/static/js/surfaces/skills.js
@@ -27,10 +27,22 @@ function installWasmExtension() {
   });
 }
 
+function normalizeMcpServerName(raw) {
+  return raw.toLowerCase()
+    .replace(/[^a-z0-9_-]/g, '_')
+    .replace(/_+/g, '_')
+    .replace(/^_|_$/g, '');
+}
+
 function addMcpServer() {
-  var name = document.getElementById('mcp-install-name').value.trim();
-  if (!name) {
+  var rawName = document.getElementById('mcp-install-name').value.trim();
+  if (!rawName) {
     showToast(I18n.t('mcp.serverNameRequired'), 'error');
+    return;
+  }
+  var name = normalizeMcpServerName(rawName);
+  if (!name) {
+    showToast('Server name contains no valid characters', 'error');
     return;
   }
   var url = document.getElementById('mcp-install-url').value.trim();
--- a/src/cli/mcp.rs
+++ b/src/cli/mcp.rs
@@ -14,7 +14,7 @@ use crate::secrets::SecretsStore;
 use crate::tools::mcp::{
     McpClient, McpProcessManager, McpServerConfig, McpSessionManager, OAuthConfig,
     auth::{authorize_mcp_server, is_authenticated},
-    config::{self, EffectiveTransport, McpServersFile},
+    config::{self, EffectiveTransport, McpServersFile, normalize_server_name},
     factory::create_client_from_config,
 };
 
@@ -290,6 +290,7 @@ async fn add_server(args: McpAddArgs) -> anyhow::Result<()> {
 
 /// Remove an MCP server.
 async fn remove_server(name: String) -> anyhow::Result<()> {
+    let name = normalize_server_name(&name);
     let (db, owner_id) = connect_db().await;
     let mut servers = load_servers(db.as_deref(), &owner_id).await?;
     if !servers.remove(&name) {
@@ -411,6 +412,7 @@ async fn list_servers(verbose: bool) -> anyhow::Result<()> {
 
 /// Authenticate with an MCP server.
 async fn auth_server(name: String, user_id: String) -> anyhow::Result<()> {
+    let name = normalize_server_name(&name);
     // Get server config
     let (db, owner_id) = connect_db().await;
     let servers = load_servers(db.as_deref(), &owner_id).await?;
@@ -495,6 +497,7 @@ async fn auth_server(name: String, user_id: String) -> anyhow::Result<()> {
 
 /// Test connection to an MCP server.
 async fn test_server(name: String, user_id: String) -> anyhow::Result<()> {
+    let name = normalize_server_name(&name);
     // Get server config
     let (db, owner_id) = connect_db().await;
     let servers = load_servers(db.as_deref(), &owner_id).await?;
@@ -613,6 +616,7 @@ async fn test_server(name: String, user_id: String) -> anyhow::Result<()> {
 
 /// Toggle server enabled/disabled state.
 async fn toggle_server(name: String, enable: bool, disable: bool) -> anyhow::Result<()> {
+    let name = normalize_server_name(&name);
     let (db, owner_id) = connect_db().await;
     let mut servers = load_servers(db.as_deref(), &owner_id).await?;
 
--- a/src/cli/mcp.rs
+++ b/src/cli/mcp.rs
@@ -516,10 +516,14 @@ async fn test_server(name: String, user_id: String) -> anyhow::Result<()> {
     let secrets = get_secrets_store().await?;
     let has_tokens = is_authenticated(&server, &secrets, &user_id).await;
 
-    let client = if has_tokens {
-        // We have stored tokens, use authenticated client
-        McpClient::new_authenticated(server.clone(), session_manager.clone(), secrets, user_id)
-    } else if server.has_custom_auth_header() {
+    // An explicit Authorization header means the user owns the credential
+    // path (API key, personal access token, etc.). It must win over any
+    // stored OAuth tokens — otherwise adding `Authorization: Bearer …` to
+    // an already-authenticated server silently keeps using the OAuth token
+    // and the explicit header is never sent. This matches the `requires_auth()`
+    // contract (returns false when an Authorization header is present) and
+    // the factory's own ordering in `create_client_from_config`.
+    let client = if server.has_custom_auth_header() {
         let process_manager = Arc::new(McpProcessManager::new());
         create_client_from_config(
             server.clone(),
@@ -530,6 +534,9 @@ async fn test_server(name: String, user_id: String) -> anyhow::Result<()> {
         )
         .await
         .map_err(|e| anyhow::anyhow!("{}", e))?
+    } else if has_tokens {
+        // We have stored tokens, use authenticated client
+        McpClient::new_authenticated(server.clone(), session_manager.clone(), secrets, user_id)
     } else if server.requires_auth() {
         // OAuth configured but no tokens - need to authenticate
         println!();
@@ -585,18 +592,22 @@ async fn test_server(name: String, user_id: String) -> anyhow::Result<()> {
             let err_str = e.to_string();
             // Check if server requires auth but we don't have valid tokens
             if crate::tools::mcp::is_auth_error_message(&err_str) {
-                if has_tokens {
-                    // We had tokens but they failed - need to re-authenticate
-                    println!(
-                        "  ✗ Authentication failed (token may be expired). Try re-authenticating:"
-                    );
-                    println!("    ironclaw mcp auth {}", name);
-                } else if server.has_custom_auth_header() {
+                // Match the construction-side ordering: an explicit
+                // Authorization header wins over stored OAuth tokens, so
+                // the failure message must point at the configured header
+                // first.
+                if server.has_custom_auth_header() {
                     println!("  ✗ Authentication failed.");
                     println!();
                     println!(
                         "  Check the configured Authorization header or API key for this server."
                     );
+                } else if has_tokens {
+                    // We had tokens but they failed - need to re-authenticate
+                    println!(
+                        "  ✗ Authentication failed (token may be expired). Try re-authenticating:"
+                    );
+                    println!("    ironclaw mcp auth {}", name);
                 } else {
                     // No tokens - server requires auth
                     println!("  ✗ Server requires authentication.");
--- a/src/cli/mcp.rs
+++ b/src/cli/mcp.rs
@@ -250,6 +250,10 @@ async fn add_server(args: McpAddArgs) -> anyhow::Result<()> {
     // Validate
     config.validate()?;
     let has_custom_auth_header = config.has_custom_auth_header();
+    // Capture the canonical stored name so follow-up hints suggest the
+    // identifier as it was persisted (matches what `mcp auth <name>` expects
+    // and keeps the suggestion shell-safe).
+    let canonical_name = config.name.clone();
 
     // Save (DB if available, else disk)
     let (db, owner_id) = connect_db().await;
@@ -258,7 +262,7 @@ async fn add_server(args: McpAddArgs) -> anyhow::Result<()> {
     save_servers(db.as_deref(), &owner_id, &servers).await?;
 
     println!();
-    println!("  ✓ Added MCP server '{}'", name);
+    println!("  ✓ Added MCP server '{}'", canonical_name);
 
     match transport_lower.as_str() {
         "stdio" => {
@@ -280,7 +284,10 @@ async fn add_server(args: McpAddArgs) -> anyhow::Result<()> {
 
     if requires_auth && !has_custom_auth_header {
         println!();
-        println!("  Run 'ironclaw mcp auth {}' to authenticate.", name);
+        println!(
+            "  Run 'ironclaw mcp auth {}' to authenticate.",
+            canonical_name
+        );
     }
 
     println!();
--- a/src/tools/mcp/config.rs
+++ b/src/tools/mcp/config.rs
@@ -77,7 +77,7 @@ impl McpServerConfig {
     /// Create a new MCP server configuration.
     pub fn new(name: impl Into<String>, url: impl Into<String>) -> Self {
         Self {
-            name: name.into(),
+            name: normalize_server_name(&name.into()),
             url: url.into(),
             transport: None,
             headers: HashMap::new(),
@@ -96,7 +96,7 @@ impl McpServerConfig {
         env: HashMap<String, String>,
     ) -> Self {
         Self {
-            name: name.into(),
+            name: normalize_server_name(&name.into()),
             url: String::new(),
             transport: Some(McpTransportConfig::Stdio {
                 command: command.into(),
@@ -114,7 +114,7 @@ impl McpServerConfig {
     /// Create a new Unix socket transport MCP server configuration.
     pub fn new_unix(name: impl Into<String>, socket_path: impl Into<String>) -> Self {
         Self {
-            name: name.into(),
+            name: normalize_server_name(&name.into()),
             url: String::new(),
             transport: Some(McpTransportConfig::Unix {
                 socket_path: socket_path.into(),
@@ -406,7 +406,8 @@ impl McpServersFile {
     }
 
     /// Add or update a server configuration.
-    pub fn upsert(&mut self, config: McpServerConfig) {
+    pub fn upsert(&mut self, mut config: McpServerConfig) {
+        config.name = normalize_server_name(&config.name);
         if let Some(existing) = self.get_mut(&config.name) {
             *existing = config;
         } else {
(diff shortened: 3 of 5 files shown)


src/channels/web/handlers/extensions.rs
```diff
@@ -176,8 +176,19 @@ pub async fn extensions_install_handler(
         _ => None,
     });
 
+    let normalized_name = crate::tools::mcp::config::normalize_server_name(&req.name);
```
serrrfirat: **Medium Severity**

This introduces a cross-surface MCP name canonicalization mismatch.

The web install path now runs `req.name` through `normalize_server_name()`, but `ExtensionManager::install()` then canonicalizes the same identifier again via `ExtensionName::new()`, which folds `-` to `_`. In practice that means installing `my-server` from the web UI persists `my_server`.

The CLI paths added in this PR (`remove`, `auth`, `test`, `toggle`) only call `normalize_server_name()` and then do exact `get/remove/get_mut` lookups, so `ironclaw mcp auth my-server` can fail with "Server 'my-server' not found" for a server that was just added through the web UI.

Concrete repro:
1. Add an MCP server from the web UI with name `my-server`.
2. Run `ironclaw mcp auth my-server` (or `remove/test/toggle`).
3. The CLI normalizes to `my-server`, while the stored config key is `my_server`.

Suggestion: pick one canonical representation for MCP server ids at all boundaries, or at least make the CLI lookup/remove paths try the same underscore/hyphen aliases that `ExtensionManager::get_mcp_server()` already does.

src/tools/mcp/config.rs
```diff
@@ -1415,6 +1459,106 @@ mod tests {
...
+        file.upsert(raw_config);
+
+        assert!(file.get("my-server").is_some());
+        assert!(file.get("My-Server").is_none());
+    }
+
+    // --- Issue #1948 regression: requires_auth with custom Authorization header ---
+
+    #[test]
+    fn test_requires_auth_skipped_with_custom_auth_header() {
```
serrrfirat, line 1707: **Medium Severity**

These regression tests pin `requires_auth()`, but the fix is still incomplete at the caller boundary.

The actual call sites still prefer stored OAuth tokens over an explicit `Authorization` header. In `cli::test_server()` the first branch is `if has_tokens { McpClient::new_authenticated(...) } else if server.has_custom_auth_header() { ... }`, and the shared factory still has the same `has_tokens || server.requires_auth()` short-circuit.

So if a user previously authenticated with OAuth, then switches the server to a manual `Authorization` header but still has stale tokens in secrets, IronClaw will still enter the auth path and attempt refresh / auth-side effects on 401s instead of honoring the explicit header. That reintroduces the #1948 behavior whenever old tokens exist.

Suggestion: make `has_custom_auth_header()` win before `has_tokens` in the caller flow too, and add a caller-level regression test that seeds both stored tokens and a custom `Authorization` header.

src/channels/web/handlers/extensions.rs
zmanian: Fixed in 99802127. Aligned `normalize_server_name` with `ExtensionName::new` by folding hyphens to underscores so the web install path and `ExtensionManager::install` produce the same canonical form. Added a regression test `test_normalize_server_name_folds_hyphens_to_match_extension_name` that asserts the two normalizers agree for a set of representative inputs. `McpServerName::new` still *accepts* hyphens so legacy persisted rows from disk continue to load; we just stop introducing new ones through the normalizer.

src/tools/mcp/config.rs
zmanian: Fixed in 99802127. Reordered the branches in `cli::test_server` so `server.has_custom_auth_header()` wins over `has_tokens`, matching the factory's `create_client_from_config` ordering and `requires_auth()`'s contract (returns false when an Authorization header is present). The failure-message branch now mirrors the same ordering — an auth error first points at the configured Authorization header before suggesting re-authentication.

src/cli/mcp.rs
```diff
@@ -290,6 +290,7 @@ async fn add_server(args: McpAddArgs) -> anyhow::Result<()> {
 
 /// Remove an MCP server.
 async fn remove_server(name: String) -> anyhow::Result<()> {
+    let name = normalize_server_name(&name);
```
serrrfirat, line 300: **High Severity**

CLI-side normalization now rewrites the lookup key before exact matching, which breaks legacy persisted MCP entries that still use a hyphenated or mixed-case name.

`load_mcp_servers_from*()` does not migrate old names; it only truncates overlong ones. So an existing config stored as `my-server` or `MCP-1` still loads and shows up in `ironclaw mcp list`, but `remove` / `auth` / `test` / `toggle` now normalize the user input to `my_server` / `mcp_1` before calling `servers.remove/get/get_mut(...)`, and those exact lookups miss.

Concrete repro after upgrade:
1. Existing `mcp-servers.json` contains `"name": "my-server"`.
2. `ironclaw mcp list` still shows `my-server`.
3. `ironclaw mcp remove my-server` normalizes to `my_server` and returns `Server 'my_server' not found`.

I think this needs either legacy-name migration on load, or tolerant lookup here (exact + legacy alias), similar to `ExtensionManager::get_mcp_server()`.

src/tools/mcp/config.rs
```diff
@@ -405,7 +406,8 @@ impl McpServersFile {
     }
 
     /// Add or update a server configuration.
-    pub fn upsert(&mut self, config: McpServerConfig) {
+    pub fn upsert(&mut self, mut config: McpServerConfig) {
+        config.name = normalize_server_name(&config.name);
```
serrrfirat, line 410: **High Severity**

`upsert()` only canonicalizes the incoming config, not the already-loaded rows it compares against. If a legacy config row still exists as `my-server`, an incoming canonicalized `my_server` will not match `self.get_mut(&config.name)`, so this pushes a second logical copy instead of updating the original.

That can happen today because the load path preserves legacy hyphenated names. After that, any add/update through the new canonical path can persist both `my-server` and `my_server`, and startup will load both entries as separate servers.

Suggested fix: normalize stored rows during `load_mcp_servers_from*()` before validation, or make `upsert()` compare against both canonical and legacy aliases before deciding to append.

crates/ironclaw_gateway/static/js/surfaces/skills.js
```diff
@@ -27,12 +27,24 @@ function installWasmExtension() {
   });
 }
 
+function normalizeMcpServerName(raw) {
+  return raw.toLowerCase()
+    .replace(/[^a-z0-9_-]/g, '_')
```
serrrfirat: **Medium Severity**

This browser normalizer still preserves `-`, so it does not match `normalize_server_name()` on the Rust side (which folds hyphens to underscores) or `ExtensionName::new()`.

Concrete examples:
- `foo--bar` stays `foo--bar` here, but `ExtensionName::new("foo--bar")` canonicalizes it to `foo__bar` and rejects it for consecutive underscores.
- `-foo-` stays `-foo-` here, but backend canonicalization becomes `_foo_` and rejects the edge underscores.

So the web UI can accept names that the backend still rejects, which keeps the surface mismatch from #2236 alive. I think this helper should mirror the Rust normalizer exactly: treat `-` like the other non-canonical characters here and let the underscore-collapse step handle it.

src/cli/mcp.rs
zmanian: Fixed in 617805df. Root-caused all three comments to the same gap — the load path preserved legacy hyphenated / mixed-case names verbatim while every new canonicalization boundary (`normalize_server_name`, `ExtensionName::new`, `upsert()`) folded them to underscores. Took the migrate-on-load option: `migrate_legacy_server_name` now folds legal-but-non-canonical legacy names to canonical form after the length-truncation step, so CLI commands that normalize user input (`remove` / `auth` / `test` / `toggle`) hit the stored row directly. Invalid characters still fall through to the existing `retain(...)` guard. Regression test: `test_load_canonicalizes_legacy_hyphenated_names`.

src/tools/mcp/config.rs
zmanian: Fixed in 617805df via the same load-path migration. With `migrate_legacy_server_name` canonicalizing legal-but-non-canonical names on load, a pre-existing `my-server` row becomes `my_server` in memory before `upsert()` ever sees it, so `upsert()`'s `self.get_mut(&config.name)` lookup hits the correct row instead of appending a duplicate. For the case where both `my-server` and `my_server` already live on disk, added `deduplicate_by_name` that runs after the `retain(...)` validation pass — keeps the first occurrence and warns on drops. Regression tests: `test_upsert_after_load_updates_legacy_row_in_place` and `test_load_deduplicates_after_canonicalization`.

crates/ironclaw_gateway/static/js/surfaces/skills.js
zmanian: Fixed in 617805df. Changed the character class in `normalizeMcpServerName` from `[^a-z0-9_-]` to `[^a-z0-9_]` so `-` is treated like any other non-canonical character and gets folded to `_` — the subsequent underscore-collapse / edge-trim steps then handle `foo--bar` → `foo_bar` and `-foo-` → `foo`, matching the Rust `normalize_server_name` output exactly. Added a comment linking the JS helper to the Rust invariant so future drift is easier to catch.

src/tools/mcp/config.rs
```diff
@@ -541,40 +543,90 @@ pub async fn load_mcp_servers() -> Result<McpServersFile, ConfigError> {
...
+    }
+}
+
+/// Drop rows whose name collides with an already-seen canonical name.
+///
+/// Can happen when load-time canonicalization in [`migrate_legacy_server_name`]
+/// rewrites a legacy row (e.g. `my-server`) to a form that matches an
+/// existing canonical row (`my_server`). We keep the first occurrence and
+/// warn on drops so the user can reconcile the duplicate manually.
+fn deduplicate_by_name(servers: &mut Vec<McpServerConfig>) {
```
serrrfirat: **High Severity**\n\nLoad-time canonicalization can silently resurrect stale MCP config when both a legacy row (`my-server`) and a canonical row (`my_server`) already exist.\n\n`deduplicate_by_name()` always keeps the **first** occurrence after both names collapse to the same canonical value. In the exact migration case this PR is addressing, that means an older legacy row can win and a newer canonical row can be dropped, reverting the URL, headers, or OAuth metadata back to stale state.\n\nConcrete example: if the file contains `[ {name: "my-server", url: "old"}, {name: "my_server", url: "new"} ]`, load now returns only the first row and loses the newer config.\n\nPlease resolve collisions with an explicit precedence rule instead of first-wins dropping — e.g. prefer the already-canonical row, prefer the later row, or merge fields deterministically and warn.

src/cli/mcp.rs
```diff
@@ -290,6 +290,7 @@ async fn add_server(args: McpAddArgs) -> anyhow::Result<()> {
 
 /// Remove an MCP server.
```
serrrfirat, line 298: **Medium Severity**\n\nThe CLI now canonicalizes server names on lookup, but the post-add auth hint still prints the raw user input instead of the canonical stored name.\n\nAfter this PR, `add_server()` stores names like `"My Server"` as `my_server`. The follow-up message still tells the user to run `ironclaw mcp auth My Server`, which is the wrong persisted identifier and is not shell-safe unless quoted. That makes the happy-path instruction easy to copy/paste incorrectly.\n\nPlease print the canonical stored name (for example `config.name`) in the auth hint so the CLI advertises the same key it actually persisted and later lookups normalize against.

src/tools/mcp/config.rs
zmanian: Fixed in 55de4490. `deduplicate_by_name` now receives a parallel `was_migrated: &[bool]` slice (populated during the load-time `migrate_legacy_server_name` pass, which now returns whether it rewrote the name). On a post-canonicalization collision, the row whose original name was already canonical wins over the legacy row; if both rows share the same migration status, the later occurrence wins (matches `upsert` append-last semantics). Added regression tests covering legacy-first, canonical-first, two-canonical, and two-legacy cases in `src/tools/mcp/config.rs`. All existing dedup tests updated accordingly.

src/cli/mcp.rs
zmanian: Fixed in 55de4490. `add_server` now captures `config.name` (already canonicalized by the `McpServerConfig::new*` constructors) into a `canonical_name` local before persisting, and the post-add success line plus auth hint both print that value. So inputs like `My Server` now produce `Run 'ironclaw mcp auth my_server' to authenticate` — which matches the persisted identifier and is shell-safe.

src/channels/web/handlers/extensions.rs
zmanian: Addressed by 99802127 / 617805df. The web/CLI/MCP code paths now agree: `McpServerConfig::new` and the load-time `migrate_legacy_server_name` (config.rs:596-614) both pipe names through `normalize_server_name`, which mirrors `ExtensionName::new`. The browser-side `normalizeMcpServerName` in skills.js was rewritten to fold hyphens to underscores too, so JS UI / Rust config / ExtensionName produce the same canonical form.

src/tools/mcp/config.rs
zmanian: The custom-auth precedence is now enforced at the caller boundary: clients with a custom `Authorization` header skip the OAuth resolution path. See 99802127 ("prioritize custom auth header").

src/cli/mcp.rs
zmanian: Addressed by 617805df + 94487879. `migrate_legacy_server_name` (config.rs:596-614) canonicalizes legacy hyphenated/mixed-case names at load time before the in-memory map is built, so by the time CLI lookup runs `normalize_server_name` on input, both sides match. The persistent file is rewritten with canonical names on the next `save_servers`.

src/tools/mcp/config.rs
zmanian: Addressed by 61df77db. `upsert` now canonicalizes the incoming config (`config.name = normalize_server_name(&config.name)`, config.rs:410) and the load-time migration canonicalizes existing rows, so `my-server` (legacy) and `my_server` (canonical) end up under the same key and `upsert` merges them rather than producing duplicates.

crates/ironclaw_gateway/static/js/surfaces/skills.js
zmanian: Addressed by 617805df. `normalizeMcpServerName` in skills.js now mirrors `normalize_server_name` in config.rs: lowercases, replaces non-`[a-z0-9_]` with `_`, collapses runs of underscores, and trims leading/trailing underscores. Hyphens are folded to underscores on both sides.

src/tools/mcp/config.rs
zmanian: Addressed by 55de4490 ("prefer canonical rows on dedup collision"). The dedup pass tracks whether each name was migrated and prefers the row that was already canonical when both forms exist (config.rs:625-633). If both are migrated or both already canonical, the last occurrence wins, matching upsert's append-last semantics.

src/cli/mcp.rs
zmanian: Addressed by 55de4490. The auth hint and CLI output now use the canonical name produced by `normalize_server_name` rather than the raw user input.

src/cli/mcp.rs
```diff
@@ -411,6 +412,7 @@ async fn list_servers(verbose: bool) -> anyhow::Result<()> {
 
 /// Authenticate with an MCP server.
 async fn auth_server(name: String, user_id: String) -> anyhow::Result<()> {
+    let name = normalize_server_name(&name);
```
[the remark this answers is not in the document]
zmanian: Addressed end-to-end: legacy hyphenated/mixed-case rows are migrated at load time (config.rs:596-614), and the CLI's lookup paths (`auth_server` cli/mcp.rs:422, `remove_server` :300, `enable/disable` :507/:637) all run input through `normalize_server_name` so they match the in-memory map's canonical keys. The persistent file is rewritten in canonical form on next save.

crates/ironclaw_gateway/static/js/surfaces/skills.js
```diff
@@ -27,12 +27,24 @@ function installWasmExtension() {
...
-  var name = document.getElementById('mcp-install-name').value.trim();
-  if (!name) {
+  var rawName = document.getElementById('mcp-install-name').value.trim();
+  if (!rawName) {
     showToast(I18n.t('mcp.serverNameRequired'), 'error');
     return;
   }
+  var name = normalizeMcpServerName(rawName);
+  if (!name) {
+    showToast('Server name contains no valid characters', 'error');
```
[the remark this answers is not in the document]
zmanian: Acknowledged — the toast string here is hardcoded. Will move to I18n in a follow-up; not blocking the canonicalization fix.

src/channels/web/handlers/extensions.rs
```diff
@@ -176,8 +176,19 @@ pub async fn extensions_install_handler(
...
+    if normalized_name.is_empty() {
+        return Ok(Json(ActionResponse::fail(
+            "Server name contains no valid characters".to_string(),
```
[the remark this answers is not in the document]
zmanian: Stale: `src/channels/web/handlers/extensions.rs` was restructured into `src/channels/web/features/extensions/mod.rs`, where all install paths run through `ExtensionName::new` for consistent error handling.

src/tools/mcp/config.rs
```diff
@@ -779,6 +781,48 @@ pub async fn remove_mcp_server_db(
...
+        } else {
+            result.push(c);
+            prev_underscore = false;
+        }
+    }
+    // Trim trailing underscore
+    if result.ends_with('_') {
+        result.pop();
+    }
+    result
```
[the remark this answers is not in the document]
zmanian: The current normalizer (config.rs:943-975) is small and runs once per server load, so further microoptimization isn't justified. Leaving as-is.

zmanian:
@serrrfirat friendly ping — green and mergeable. Fixes MCP server-name normalization, UTF-8 truncation, and auth-skip when header set (#2236, #1947, #1948). Mind giving it a review?

zmanian:
Closing: these fixes target the v1 MCP client (src/tools/mcp/), and the Reborn MCP stack (crates/ironclaw_mcp) was rebuilt with a different architecture where none of the three bugs can exist in the same form — name normalization lives in v1 config + v1 gateway JS only, output truncation in Reborn is byte-bounded through McpClientRequest::max_output_bytes in the adapter contract, and Reborn fails closed with AuthRequired routed through product-auth instead of v1's OAuth/DCR client. With UI and runtime investment moving to Reborn, the seven-week-old v1 patch no longer justifies the review cost. #1947 was already closed; if #2236 or #1948 still bite in practice on the v1 binary, the diff here remains cherry-pickable.

Issue #2599, opened 2026-04-17
Epic: Enforce gateway feature boundaries, crate guardrails, and crate-owned E2E


henrypark133:
## Summary

Refactor the web gateway/frontend into bounded feature slices plus a small platform layer, then enforce those boundaries with CI guardrails and crate-owned E2E paths.

This epic is intentionally staged. We will first create strict ownership and import boundaries inside the existing crates, then migrate the highest-risk slices into that structure, and only after the seams stabilize extract any truly shared contracts into a dedicated crate.

This is not a big-bang rewrite. The intent is to improve containment, reviewability, and testability without destabilizing the runtime or deployment model.

## Why this exists

The current gateway structure already has some separation, but the main failure mode is still cross-slice sprawl:

- `src/channels/web/server.rs` still carries too much router composition, state wiring, and product behavior
- `crates/ironclaw_gateway/static/app.js` still carries too much auth, SSE, shell, onboarding, tab, and feature state logic
- feature logic is not consistently owned end-to-end by one backend/frontend slice
- test ownership is manual and coarse rather than aligned to crate/feature boundaries
- architecture boundaries are partly enforced in the repo already, but not specifically for gateway feature slices

This makes it too easy for a “small fix” to reach across multiple areas:
- chat logic reaches into extension onboarding behavior
- setup/auth/onboarding flows get normalized in generic server orchestration
- frontend feature code ends up coupled through shared DOM/state helpers rather than explicit platform APIs
- PRs become harder to review because ownership boundaries are not obvious from the diff

The goal is to make healthy changes local by default and cross-slice changes explicit and reviewable.

## Desired outcome

A normal gateway PR should mostly stay inside one ownership zone.

Examples of the standard we want:

- “Fix extension onboarding retry state”
  - touches `features/extensions/*`
  - maybe one shared contract or platform event
- “Fix chat thread restore after reconnect”
  - touches `features/chat/*`
  - maybe `platform/events.js`
- “Add jobs summary filter”
  - touches `features/jobs/*`
- “Change top-level shell reconnect badge behavior”
  - touches `platform/*`
  - does not require edits across every feature

Cross-slice changes should still be possible, but they should be rare, obvious, and expensive enough that the reviewer immediately notices them.

## Scope and approach

This epic covers:

- backend gateway structure under `src/channels/web/`
- frontend gateway structure under `crates/ironclaw_gateway/static/`
- architecture/ownership enforcement in CI
- E2E organization by owning crate and feature path
[...]

PR #2700, opened 2026-04-20
fix(gateway): show descriptive chat titles instead of hex hash IDs (#2237)

### LINKED ISSUE #2237: [QA] New chats get alphanumeric hash names instead of descriptive titles

zmanian:
Replaces #2348 — recreated on fresh staging after gateway and web handler refactors (#2683, #2599).

## Summary

Three-layer fix for new conversations displaying truncated UUIDs in the web sidebar:

- **Persistence layer** — Store conversation title in metadata on first user message (both v1 agent and v2 engine paths) via shared `set_title_if_missing()` helper in `src/db/mod.rs`. Empty/whitespace-only input is skipped so image-only messages don't permanently block title setting.
- **SQL derivation** — Both PostgreSQL and libSQL title derivation now check `metadata.title` as a fallback between the message-subquery title and `routine_name`.
- **Frontend** — `threadTitle()` (ported to `js/core/history.js` post-#2683) shows localized "Untitled chat" instead of hex hash substring. In-memory fallback derives title from first turn's input with whitespace normalization (addresses review feedback about multi-line input).

## Files changed

- `src/channels/web/features/chat/mod.rs` — in-memory title derivation (ported from handlers/chat.rs post-#2599)
- `src/agent/thread_ops.rs`, `src/bridge/router.rs` — both call shared `set_title_if_missing()`
- `src/db/mod.rs` — `set_title_if_missing()` helper
- `src/db/libsql/conversations.rs`, `src/history/store.rs` — metadata.title fallback in SQL title derivation
- `crates/ironclaw_gateway/static/js/core/history.js` — `threadTitle()` i18n
- `crates/ironclaw_gateway/static/i18n/{en,ko,zh-CN}.js` — new keys

## Notes

- Addresses prior feedback from @serrrfirat: deduplicated title-setting code, empty-input guard, whitespace normalization.
- TOCTOU concern on concurrent title-set remains (noted but deferred — benign race, worst case title reflects second message).

## Test plan

- [ ] Manual: create new thread, verify sidebar shows first-message title instead of hex ID

Closes #2237


ilblackdragon:
# Code Review

## Overview
Three-layer fix so new conversations show a real title in the web sidebar instead of `id.substring(0,8)`:
1. **Backend write** — `set_title_if_missing()` helper persists the first user message as `metadata.title` on both v1 agent (`thread_ops.rs`) and v2 engine (`bridge/router.rs`) paths.
2. **SQL read** — libSQL + PostgreSQL title derivation check `metadata.title` as a fallback between the message-subquery title and `routine_name`.
3. **Frontend** — `threadTitle()` returns localized "Untitled chat" instead of hex; in-memory listing (`chat_threads_handler`) derives title from the first turn with whitespace normalization.

CI is green across fmt/clippy/deny/replay; only en/ko/zh-CN locales exist, so translations are complete.

## Strengths
- Clean, idiomatic Rust; shared helper (`set_title_if_missing`) correctly placed in `src/db/mod.rs` and reused across both engine paths — no duplication.
- Empty-input guard prevents image/attachment-only messages from permanently poisoning the title slot.
- Good regression tests on libSQL: both "metadata fallback used" and "message title wins over metadata" paths are covered.
- Char-aware truncation (`chars().take(100)`) avoids multibyte splits.
- Fire-and-forget title update doesn't block the message-persistence happy path.

## Issues & Suggestions

### Correctness
- **`thread_ops.rs:1079-1082`** — Title is now set **unconditionally** after the match, including when `add_conversation_message` returned `Err` (the match arm logs and returns `None`). That can produce a conversation with a `metadata.title` but no persisted user message — minor "ghost title" case. Consider moving the call inside the `Ok` arm, or gating it on `result.is_some()`.
- **`db/mod.rs:522-525`** — `Err(_) => return` silently swallows storage errors. CLAUDE.md's style guide calls for error context (`.map_err(...)`); at minimum a `debug!("failed to read metadata for title-set: {e}")` would help debugging without bloating the code.
- **Whitespace inconsistency** — `chat_threads_handler` normalizes whitespace via `split_whitespace().join(" ")`, but `set_title_if_missing` only `trim()`s. A multi-line first message persisted today will keep its internal newlines in `metadata.title`, while the in-memory path (pre-persist) would render it collapsed. Low impact but easy to align — collapse whitespace in the helper too.

### Performance
- `set_title_if_missing` does two DB roundtrips (read metadata, then write) on **every** user message of an existing conversation that already has a title. A conditional-update SQL (`UPDATE ... WHERE metadata->>'title' IS NULL`) would be one roundtrip and also race-free. The PR's deferred TOCTOU note would resolve as a side effect.

### Test coverage
- Good coverage for libSQL; **no parallel tests for `history/store.rs` (PostgreSQL path)** — the same metadata-title fallback logic was added there but isn't exercised. Worth a mirror test under `--features integration`.
- No direct test for `set_title_if_missing` itself (empty-input guard, already-has-title no-op, 100-char truncation).
- `chat_threads_handler`'s in-memory title derivation (the multi-line/whitespace case cited in the PR body) isn't regression-tested. Per CLAUDE.md's "test through the caller" rule, this is the kind of handler-level behavior worth an integration test.

### Security
- User input is persisted verbatim (trim + 100-char cap). No control-char/newline filtering. Not a vuln since JSON encodes safely, but worth confirming the frontend renders `metadata.title` via `textContent`/`I18n.t` rather than as HTML.

### Style / conventions
- Follows the codebase: `crate::db::set_title_if_missing` (no re-export), dispatch-exempt isn't needed since the call lives in the engine/bridge path, not a UI mutation handler.
- Doc comment on the helper explains the *why* (empty-input guard) — matches CLAUDE.md guidance.

## Risk assessment
Low risk. Fully additive: existing `sql_title` still takes precedence, so renamed/derived titles are unaffected. The one behavioral change — titles appearing sooner for conversations with no message-subquery hit — is the intended fix. The "ghost title on failed persist" edge case is the only real correctness nit; the rest is polish.

## Recommendation
**Approve with minor nits.** Worth addressing the `Err` ordering in `thread_ops.rs` and adding a PostgreSQL regression test before merge; the whitespace-normalization alignment and conditional-update optimization can be follow-ups.

zmanian:
Addressed @ilblackdragon's review nits in de2876b:

1. **Ghost title on failed persist** — `set_title_if_missing` now only runs when `persist_user_message` succeeded (`result.is_some()` gate in `thread_ops.rs`).
2. **Silent error swallow** — `Err(_) => return` in `set_title_if_missing` now logs at `debug!` level before returning.
3. **Whitespace normalization** — already aligned: `set_title_if_missing` uses `split_whitespace().join(" ")` (line 521-524), matching the in-memory path.

Clippy clean, zero warnings.

---

serrrfirat (commented):
Re-checked this against current staging. The underlying sidebar-title bug is still real, but I found a few remaining correctness gaps before merge; details inline.

src/bridge/router.rs
```diff
@@ -3651,6 +3651,7 @@ async fn handle_with_engine_inner(
             let _ = db
                 .add_conversation_message(cid, "user", effective_content)
                 .await;
+            crate::db::set_title_if_missing(db.as_ref(), cid, effective_content).await;
```
serrrfirat: One more edge here: `effective_content` already comes from `augment_with_attachments()` above, so on attachment/image-only first turns this can derive the title from the synthesized `<attachments>` block / extracted attachment text rather than from actual user-entered text. That seems to defeat the goal of skipping empty input so image-only turns do not claim the title slot. Can we derive the title from raw user text instead of the augmented payload?

serrrfirat: Because the insert result is discarded here, `set_title_if_missing()` still runs even when `add_conversation_message()` fails. That recreates the ghost-title edge case on the v2 path: the sidebar can show a metadata title even though the first user message never persisted. Can we gate the title write on successful message persistence here too, like `thread_ops.rs` now does?

src/db/libsql/conversations.rs
```diff
@@ -1009,4 +1029,85 @@ mod tests {
             "assistant thread lookup should backfill a legacy NULL source_channel"
         );
     }
+
+    /// Regression test for #2237: conversations with a metadata title should
+    /// use it as a fallback when the message-derived title is NULL.
+    #[tokio::test]
+    async fn test_metadata_title_used_as_fallback() {
```
serrrfirat, line 1069: These new tests are helpful for the libSQL read path, but the same precedence change also landed in `src/history/store.rs` and the actual write sites are `thread_ops.rs` / `bridge/router.rs`. As written, a regression in the PostgreSQL path or in the caller wiring would still pass here. Please add a PG mirror and at least one caller-level regression that drives the real persist path.

src/bridge/router.rs
zmanian: Fixed in eaedb2e5. You're right — `effective_content` here is the output of `augment_with_attachments`, so an image/attachment-only first turn would claim the title slot with the synthesized `<attachments>` block or extracted attachment text. The bridge path now passes the raw user `content` to `set_title_if_missing` while the conversation message body continues to store `effective_content`. The same fix was applied in `thread_ops::persist_user_message` — the function now takes a separate `title_source` argument so the DB row body and the title derivation don't share the same input. `set_title_if_missing` already skips empty/whitespace input, so attachment-only first turns now naturally leave the title unset for a later turn to claim.

zmanian: Fixed in eaedb2e5. Captured the `add_conversation_message` result and gated `set_title_if_missing` on `Ok(_)`, mirroring the `result.is_some()` gate on the v1 path. If the insert fails (e.g. DB hiccup), the title write is skipped so the sidebar no longer shows a metadata title for a conversation whose first user row never persisted.

src/db/libsql/conversations.rs
zmanian: Fixed in eaedb2e5. Added two layers:

1. **PG mirror** of the two libSQL regression tests in `src/history/store.rs` behind `feature = "postgres"` + `#[ignore]` (integration tier, same pattern as `test_admin_usage_summary_pg` / `test_save_job_persists_user_id`):
   - `test_metadata_title_used_as_fallback_pg` — pins the fallback to `metadata.title` when no user messages exist.
   - `test_message_title_takes_precedence_over_metadata_pg` — pins that the first user message shadows `metadata.title` on the PG path.

2. **Caller-level regression** in `src/agent/thread_ops.rs::tests::persist_user_message_title_uses_raw_text_not_augmented_payload` that drives `Agent::persist_user_message` against a real `LibSqlBackend` — the actual production persist path. It pins three invariants that a predicate-level test cannot:
   - Attachment-augmented payload with real user text → title comes from the raw text (not the `<attachments>` block).
   - Attachment-only first turn (empty raw text) → title stays unset.
   - Plain-text first turn → title seeded correctly (negative control for the edits).

   A regression in the caller wiring (e.g. swapping `title_source` back to `user_input`, or dropping the `result.is_some()` gate) would leave the libSQL unit tests green but fail this caller-level test.

Run PG tests with: `cargo test -p ironclaw --features integration --lib history::store::tests::test_metadata_title_used_as_fallback_pg -- --ignored` (requires reachable PostgreSQL).

src/channels/web/features/chat/mod.rs
```diff
@@ -752,15 +752,30 @@ pub(crate) async fn chat_threads_handler(
...
-            updated_at: t.updated_at.to_rfc3339(),
-            title: None,
-            thread_type: None,
-            channel: Some("gateway".to_string()),
+        .map(|t| {
+            let title = t
+                .turns
+                .first()
+                .map(|turn| {
+                    turn.user_input
```
serrrfirat: **Medium Severity**

The no-DB / in-memory fallback here is still deriving the sidebar title from `turn.user_input`, but `turn.user_input` is the attachment-augmented payload (`process_user_input` calls `thread.start_turn(effective_content)` and only passes the raw text separately to `persist_user_message`). That means the exact bug this PR is fixing comes back whenever `/api/chat/threads` falls back to session state: a first turn with attachments can title the thread with `<attachments>` markup, OCR text, or saved file-path text instead of the user's raw prompt.

Concrete fix: carry a raw `title_source` into the in-memory thread state (or derive/store the title before attachment augmentation) and reuse that here instead of `turn.user_input`. I'd also add a caller-level test for `chat_threads_handler` in the no-DB path using a first turn whose `effective_content` contains an `<attachments>` block.

src/bridge/router.rs
```diff
@@ -3648,9 +3648,27 @@ async fn handle_with_engine_inner(
...
+            // Title is derived from the raw user `content`, not the
+            // attachment-augmented `effective_content`, so an
+            // image-only or attachment-only first turn does not claim
+            // the title slot with the synthesized `<attachments>` block
+            // or extracted attachment text. `set_title_if_missing`
+            // already skips empty/whitespace input, so an attachment-
+            // only turn simply leaves the title unset until a later
+            // turn carries real user text.
+            if db
                 .add_conversation_message(cid, "user", effective_content)
```
serrrfirat, line 4816: **High Severity**

This fixes the v1 dual-write title source, but the engine-v2 thread title is still derived from the attachment-augmented payload. In this function `effective_content` already includes the synthesized attachment block / extracted text, and that exact string is passed into `ConversationManager::handle_user_message`; inside `crates/ironclaw_engine/src/runtime/conversation.rs` the engine derives `Thread.title` from `content`. So any engine-thread surface (TUI, mission/thread detail views, engine thread APIs) can still expose OCR text, saved project paths, or `<attachments>` markup in the title even after this PR.

Concrete fix: pass a separate raw-text title source into the engine path (mirroring `persist_user_message(..., title_source)` on v1), or extend `handle_user_message` / `spawn_thread_with_history` so the engine title is derived from raw user text before attachment augmentation.

src/db/mod.rs
```diff
@@ -494,6 +494,48 @@ pub trait ConversationStore: Send + Sync {
...
+            .and_then(|t| t.as_str())
+            .is_some_and(|s| !s.is_empty()),
+        Ok(None) => return,
+        Err(e) => {
+            tracing::debug!("failed to read metadata for title-set: {e}");
+            return;
+        }
+    };
+
+    if !has_title {
```
serrrfirat: **Medium Severity**

`set_title_if_missing` is still a check-then-update race: two concurrent writes can both observe `metadata.title` as empty and then race to write different titles, so the sidebar can end up showing the second message instead of the first. Because both the v1 caller path and the v2 gateway bridge now funnel through this helper, the race is still present on both codepaths.

If you want the first persisted user text to win deterministically, this needs to become an atomic compare-and-set in the backend implementations (for example: a single `UPDATE ... WHERE metadata->>'title' IS NULL OR metadata->>'title' = ''` / SQLite equivalent, or a new DB trait method that implements that semantics in one round-trip per backend).

serrrfirat:
I'm not sure if this is still an issue?

src/channels/web/features/chat/mod.rs
zmanian: Addressed in 27669084. The no-DB fallback in `chat_threads_handler` now uses a new `Turn::raw_user_input` field populated by `process_user_input` whenever attachment augmentation changes the text. A shared `title_from_in_memory_turn` helper prefers raw over `user_input` (the augmented payload), and plain turns leave `raw_user_input` as `None` so they fall through cleanly. Three new unit tests (`test_title_from_in_memory_turn_*`) cover augmented / plain / empty-raw cases.

src/bridge/router.rs
zmanian: Addressed in 27669084. `ConversationManager::handle_user_message` now takes `raw_content_for_title: Option<&str>`. The LLM-facing `content` is unchanged (still the attachment-augmented payload fed to `ConversationEntry::user` / `ThreadMessage::user`), but when `raw_content_for_title` differs from `content` the user entry is created via new `ConversationEntry::user_with_title_source`, which stamps `metadata.title_source` with the raw user text. `src/bridge/router.rs` passes the raw `content` alongside `effective_content` at the call site. Tests: `handle_user_message_records_raw_title_source` (augmented case) and `handle_user_message_plain_text_has_no_title_source_metadata` (no redundant metadata on plain turns).

src/db/mod.rs
zmanian: Addressed in 27669084. Replaced the check-then-update pair with a new `ConversationStore::set_conversation_title_if_empty` trait method that uses a single conditional `UPDATE` on both backends:

- libSQL: `UPDATE conversations SET metadata = json_patch(COALESCE(metadata, '{}'), ?2) WHERE id = ?1 AND (metadata IS NULL OR json_extract(metadata, '$.title') IS NULL OR json_extract(metadata, '$.title') = '')`
- PG: `UPDATE conversations SET metadata = COALESCE(metadata, '{}'::jsonb) || $2 WHERE id = $1 AND (metadata IS NULL OR NOT (metadata ? 'title') OR metadata->>'title' = '')`

Returns a bool so callers can see whether the write landed. `set_title_if_missing` now just calls it and logs on error. Regression tests:
- `test_set_title_if_empty_is_atomic_under_concurrency` (libSQL) — `tokio::join!` races two setters with different titles and asserts exactly one wins.
- `test_set_title_if_empty_writes_when_unset` — covers NULL metadata and empty-string title rows.
- `test_set_title_if_empty_is_atomic_pg` — PG mirror behind `#[ignore]` for integration tier, same race assertion.

src/bridge/router.rs
zmanian: Addressed in 27669084. Title is now derived from raw `content` (the user-typed text) rather than `effective_content` (which carries the synthesized `<attachments>` block). See router.rs:3663-3676 — the `add_conversation_message` call still stores `effective_content`, but `set_title_if_missing` receives `content`. `set_title_if_missing` skips empty/whitespace input, so attachment-only first turns simply leave the title unset until a later text-bearing turn.

zmanian: Addressed in 27669084 / eaedb2e5. `set_title_if_missing` is now gated on `add_conversation_message` succeeding (`if db.add_conversation_message(...).await.is_ok()` at router.rs:3671-3677), mirroring the gating in `thread_ops::persist_user_message`. The ghost-title regression is closed.

src/db/libsql/conversations.rs
zmanian: Coverage extended in eaedb2e5 ("derive titles from raw user text, gate on persist success, expand tests"). The actual write sites — `thread_ops::persist_user_message` and the v2 path in `bridge/router.rs` — are now exercised by `agent::thread_ops::tests::persist_user_message_title_uses_raw_text_not_augmented_payload` and the chat in-memory tests `test_title_from_in_memory_turn_*`.

src/channels/web/features/chat/mod.rs
zmanian: Addressed. `title_from_in_memory_turn` now prefers `turn.raw_user_input` and only falls back to `turn.user_input` (which is the attachment-augmented payload) when `raw_user_input` is unset. See chat/mod.rs:1100-1117 with regression tests `test_title_from_in_memory_turn_prefers_raw_input` and `_falls_back_to_user_input`.

src/bridge/router.rs
zmanian: Addressed in 27669084. The engine-v2 path at router.rs:3676 now passes raw `content` (not `effective_content`) into `set_title_if_missing`. `add_conversation_message` still stores `effective_content` so the conversation row contains the augmented payload, but the sidebar title comes from the user's typed text only.

src/db/mod.rs
zmanian: Addressed in 27669084 ("make title write atomic"). `set_title_if_missing` now calls `set_conversation_title_if_empty` (db/mod.rs:488-492), which both libSQL and Postgres backends implement as a single SQL update guarded by a NULL/empty check on `metadata.title`. First commit wins; subsequent concurrent first-turn writes are no-ops. Regression test: `test_set_title_if_empty_is_atomic_under_concurrency`.

src/history/store.rs
```diff
@@ -1839,12 +1839,20 @@ impl Store {
...
                 let sql_title: Option<String> = r.get("title");
-                let title = sql_title.or_else(|| {
-                    metadata
-                        .get("routine_name")
-                        .and_then(|v| v.as_str())
-                        .map(String::from)
-                });
+                let title = sql_title
+                    .or_else(|| {
+                        metadata
```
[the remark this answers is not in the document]
zmanian: Addressed in 716b8784 / 27669084. SQL-derived titles are now whitespace-normalized and filtered for emptiness before being used; blank/whitespace-only message bodies fall through to the metadata title rather than being treated as authoritative. Tests: `test_message_title_takes_precedence_over_metadata`, `test_metadata_title_used_as_fallback`.

src/db/mod.rs
```diff
@@ -494,6 +494,39 @@ pub trait ConversationStore: Send + Sync {
...
+    if trimmed.is_empty() {
+        return;
+    }
+
+    let has_title = match store.get_conversation_metadata(conversation_id).await {
+        Ok(Some(meta)) => meta
+            .get("title")
+            .and_then(|t| t.as_str())
+            .is_some_and(|s| !s.is_empty()),
+        Ok(None) => false,
```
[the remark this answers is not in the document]
zmanian: Addressed in 716b8784 ("normalize sql_title whitespace") and 27669084. SQL-derived titles run through whitespace collapsing + emptiness filtering before being used at all five sites you flagged. Coverage in `test_message_title_takes_precedence_over_metadata` and `test_metadata_title_used_as_fallback`.

src/db/mod.rs
```diff
@@ -494,6 +494,39 @@ pub trait ConversationStore: Send + Sync {
...
+        Err(_) => return,
+    };
+
+    if !has_title {
+        let title_text: String = trimmed.chars().take(100).collect();
```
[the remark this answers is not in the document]
zmanian: Addressed in 716b8784 ("normalize sql_title whitespace") and 27669084. SQL-derived titles run through whitespace collapsing + emptiness filtering before being used at all five sites you flagged. Coverage in `test_message_title_takes_precedence_over_metadata` and `test_metadata_title_used_as_fallback`.

src/db/libsql/conversations.rs
```diff
@@ -171,12 +171,20 @@ impl ConversationStore for LibSqlBackend {
...
                 .map(String::from);
             let sql_title = get_opt_text(&row, 6);
-            let title = sql_title.or_else(|| {
-                metadata
-                    .get("routine_name")
-                    .and_then(|v| v.as_str())
-                    .map(String::from)
-            });
+            let title = sql_title
+                .or_else(|| {
```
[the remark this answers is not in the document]
zmanian: Addressed in 716b8784 ("normalize sql_title whitespace") and 27669084. SQL-derived titles run through whitespace collapsing + emptiness filtering before being used at all five sites you flagged. Coverage in `test_message_title_takes_precedence_over_metadata` and `test_metadata_title_used_as_fallback`.

src/db/libsql/conversations.rs
```diff
@@ -250,12 +258,20 @@ impl ConversationStore for LibSqlBackend {
...
```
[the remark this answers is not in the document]
zmanian: Addressed in 716b8784 ("normalize sql_title whitespace") and 27669084. SQL-derived titles run through whitespace collapsing + emptiness filtering before being used at all five sites you flagged. Coverage in `test_message_title_takes_precedence_over_metadata` and `test_metadata_title_used_as_fallback`.

src/history/store.rs
```diff
@@ -1839,12 +1839,20 @@ impl Store {
...
                     .map(String::from);
                 let sql_title: Option<String> = r.get("title");
-                let title = sql_title.or_else(|| {
-                    metadata
-                        .get("routine_name")
-                        .and_then(|v| v.as_str())
-                        .map(String::from)
-                });
+                let title = sql_title
+                    .or_else(|| {
```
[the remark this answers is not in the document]
zmanian: Addressed in 716b8784 ("normalize sql_title whitespace") and 27669084. SQL-derived titles run through whitespace collapsing + emptiness filtering before being used at all five sites you flagged. Coverage in `test_message_title_takes_precedence_over_metadata` and `test_metadata_title_used_as_fallback`.

src/history/store.rs
```diff
@@ -1913,12 +1921,20 @@ impl Store {
...
                 // since they may have no user messages.
                 let sql_title: Option<String> = r.get("title");
-                let title = sql_title.or_else(|| {
-                    metadata
-                        .get("routine_name")
-                        .and_then(|v| v.as_str())
-                        .map(String::from)
-                });
+                let title = sql_title
+                    .or_else(|| {
```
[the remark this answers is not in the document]
zmanian: Addressed in 716b8784 ("normalize sql_title whitespace") and 27669084. SQL-derived titles run through whitespace collapsing + emptiness filtering before being used at all five sites you flagged. Coverage in `test_message_title_takes_precedence_over_metadata` and `test_metadata_title_used_as_fallback`.

src/bridge/router.rs
zmanian: Addressed in feec0c0e and carried through the d9f6f52 merge from current main. Current main added a dedicated `Thread.title`, so the final fix keeps `Thread.goal` as the full execution prompt / initial LLM message, but derives `Thread.title` from `raw_content_for_title` instead of the attachment-augmented payload. Added `handle_user_message_uses_raw_title_source_as_thread_title`, which pins: title = raw user text, goal = augmented execution prompt, first user message = augmented payload. Verified with `cargo test -p ironclaw_engine runtime::conversation::tests` and `cargo check -p ironclaw`.

zmanian:
Update to get this unstuck:

- Added `feec0c0e` for the remaining engine-v2 title gap: current `main` now has `Thread.title`, so the merged branch derives that title from raw user text while keeping `Thread.goal` / the first LLM message as the full attachment-augmented execution prompt.
- Merged current `origin/main` in `d9f6f52d`; PR is mergeable again.
- Resolved the addressed/stale review threads; unresolved review threads are now at 0.

Local verification:
- `cargo test -p ironclaw_engine runtime::conversation::tests`
- `cargo check -p ironclaw`

The existing re-review request for @serrrfirat is still active.

zmanian:
@serrrfirat friendly ping — this one is green (CI passing) and mergeable, just awaiting your review. It swaps the hex-hash chat titles for descriptive ones (#2237). Could you take a look when you have a moment?

zmanian:
Closing: this fix spans 15 files across the v1 gateway JS, engine conversation types, both DB backends, and the bridge router — areas that have drifted heavily in seven weeks (the branch is now conflicting) — and the webui-v2 surface has its own thread/title path through SessionThreadService and product_workflow, so none of this carries forward. #2237 stays open; the descriptive-titles work should be re-scoped against the v2 thread surface instead of paying this rebase for a UI being replaced.

PR #1491, opened 2026-03-20
feat(web): expose cheap_model and smart_routing_cascade in settings UI

wehrmannit:
## Summary

- Adds `cheap_model` and `smart_routing_cascade` fields to the `Settings` struct so they persist via settings.json / DB
- Exposes both in the web UI settings page (Inference tab) with i18n support (en + zh-CN)
- Updates `LlmConfig::resolve()` to fall back to settings when env vars aren't set (env var > settings > default)
- Marks both keys as restart-required in the UI

Follows up on #1081 which added `LLM_CHEAP_MODEL` and `SMART_ROUTING_CASCADE` env vars but left them unconfigurable from the browser.

## Test plan

- [x] `cargo clippy --all --benches --tests --examples --all-features` — zero warnings
- [x] `cargo build --release` — compiles clean
- [x] Open web UI Settings > Inference, verify "Cheap Model" text field and "Smart Routing Cascade" toggle appear
- [x] Set a cheap model via the UI, restart, confirm it takes effect in smart routing logs
- [x] Verify env var still overrides the UI setting

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2726, opened 2026-04-20
fix(web): harden cheap-model settings rollout

serrrfirat:
## Summary

Follow-up to #1491.

This ports the `cheap_model` / `smart_routing_cascade` settings work onto current `staging` and hardens it for the repo's current architecture:

- persists `cheap_model` and `smart_routing_cascade` in `Settings`
- resolves them as `env > settings > default`
- treats both as admin-scoped LLM-chain settings
- includes them in the existing hot-reload path
- fixes settings admin checks to use `user.is_admin()` so owners are handled correctly
- updates the current web UI implementation under `crates/ironclaw_gateway/static/js/surfaces/settings.js`
- uses an inherit/enabled/disabled control for `smart_routing_cascade` to preserve the unset/default semantics
- exposes effective `cheap_model` and `smart_routing_cascade` through gateway status so the settings UI can render the active values correctly

## Why this is separate from #1491

#1491 targets the older frontend file layout and misses a few current-`staging` concerns:

- shared LLM chain hot reload
- admin scoping for shared-chain settings
- owner/admin role handling via `is_admin()`
- unset/inherit UI semantics for `smart_routing_cascade`

So this PR is intended as a hardened successor / follow-up reference to #1491, not an unrelated parallel change.

## Test plan

- [x] `cargo fmt --all`
- [x] `cargo test --lib cheap_model_uses_settings_when_env_unset`
- [x] `cargo test --lib smart_routing_cascade_uses_settings_when_env_unset`
- [x] `cargo test --lib test_settings_export_owner_can_read_admin_scope`
- [x] `cargo test --lib settings_set_handler_triggers_llm_provider_hot_reload`
- [x] `node --check crates/ironclaw_gateway/static/js/surfaces/settings.js`
- [x] `node --check crates/ironclaw_gateway/static/i18n/en.js`
- [x] `node --check crates/ironclaw_gateway/static/i18n/zh-CN.js`


henrypark133 (dismissed):
No verified findings.

Summary:
- Residual risk: I did not find a scope or reload bug here, but there still is no end-to-end admin-scope settings/status test, and `/api/gateway/status` now exposes the resolved cheap-model snapshot to any authenticated user if that matters for the intended contract.

serrrfirat:
Addressed the review feedback: aligned `cheap_model` and `smart_routing_cascade` to settings/DB-first precedence using the shared config helpers, added regression tests for settings-vs-env resolution, and cleaned up the minimal dispatch-exempt handler checks needed for the safety hook.

Verified with `cargo fmt`, `cargo clippy --all --benches --tests --examples --all-features`, targeted `config::llm` tests, and `scripts/pre-commit-safety.sh`. Full `cargo test --lib` still has unrelated pre-existing failures in the skill ZIP tests.

PR #2728, opened 2026-04-20
feat:(cli) Add Engine V2 migrate CLI for OpenClaw and Hermes

serrrfirat:
## Summary

- add a first-class `ironclaw migrate` CLI with `openclaw` and `hermes` source adapters
- gate the migration/import CLI behind an opt-in `migrate` Cargo feature instead of shipping it in default features
- remove the temporary `import` Cargo feature alias entirely so `migrate` is the only feature gate
- write migrated data into Engine V2-visible state via `HybridStore`, workspace docs, legacy conversation history, settings, and encrypted secrets
- keep `ironclaw import` as a compatibility CLI alias that delegates to the new migrate flow when the feature is enabled
- add targeted migration integration tests plus README / parity / changelog updates

## Change Type

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [x] Documentation
- [ ] CI/Infrastructure
- [ ] Security
- [ ] Dependencies

## Linked Issue

None

## Validation

- [x] `cargo build`
- [x] Relevant tests pass: `cargo test --features migrate --test migrate_cli_integration`
- [ ] `cargo test --features integration` if database-backed or integration behavior changed (not run locally; targeted `cargo test --features migrate --test migrate_cli_integration` covers the migration path, including binary CLI dispatch)
- [x] Manual testing: `cargo check`, `cargo check --tests --features migrate`, `cargo run --features migrate --bin ironclaw -- migrate --help`

## Security Impact

Touches migration-time secrets handling by importing provider credentials and Hermes auth material into the encrypted secrets store. Does not weaken existing runtime auth flows.

## Database Impact

No schema changes. Migration writes existing settings/history/workspace/secrets data through current abstractions and currently validates against the libSQL-backed path used by the feature.

## Blast Radius

CLI command surface, migration/import pipeline, Engine V2 store adapter usage, workspace/settings/secrets persistence, feature gating, and migration docs/tests.

## Rollback Plan

Revert this PR to remove the `migrate` command and restore the prior `import`-only behavior.

## Review Follow-Through

This branch was rebased to drop unrelated cheap-model/settings UI work so the PR now contains only the migration feature changes. The migration CLI is opt-in behind `--features migrate`, is no longer part of the default feature set, and no longer exposes an `import` Cargo feature alias.

---

**Review track**: C

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- *(cli)* add opt-in `ironclaw migrate openclaw|hermes` (behind `--features migrate`) for direct assistant migration into IronClaw Engine V2 state, legacy web history, settings, and encrypted secrets.
+
 ### Changed
 
 - *(web)* gateway onboarding/auth SSE now uses the unified `onboarding_state` event; external SSE clients should migrate from the older auth/pairing event names. Legacy WebSocket `auth_token` and `auth_cancel` client messages remain accepted during the temporary web v1-auth compatibility window.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -201,7 +201,7 @@ ed25519-dalek = { version = "2.2.0", features = ["std"] }
 bs58 = "0.5"
 hex = "0.4.3"
 
-# OpenClaw import (feature gated)
+# Migration readers (OpenClaw / Hermes)
 json5 = { version = "0.4", optional = true }
 
 # macOS keychain
@@ -253,7 +253,8 @@ replay = ["libsql"]
 html-to-markdown = ["dep:html-to-markdown-rs", "dep:readabilityrs"]
 bedrock = ["dep:aws-config", "dep:aws-sdk-bedrockruntime", "dep:aws-smithy-types"]
 tui = ["dep:ironclaw_tui"]
-import = ["dep:json5", "libsql"]
+# Opt-in migration/import CLI for external assistant homes.
+migrate = ["dep:json5", "libsql"]
 
 [[test]]
 name = "e2e_thread_scheduling"
--- a/FEATURE_PARITY.md
+++ b/FEATURE_PARITY.md
@@ -167,6 +167,7 @@ This document tracks feature parity between IronClaw (Rust implementation) and O
 | `agents` | ✅ | ❌ | P3 | Multi-agent management |
 | `sessions` | ✅ | ❌ | P3 | Session listing (shows subagent models) |
 | `memory` | ✅ | ✅ | - | Memory search CLI |
+| `migrate` | ✅ | ✅ | P1 | `ironclaw migrate openclaw` and `ironclaw migrate hermes` import settings, secrets, workspace docs, Engine V2 threads/conversations, and legacy web history |
 | `skills` | ✅ | ✅ | - | CLI subcommands (list, search, info) + agent tools + web API endpoints |
 | `pairing` | ✅ | ✅ | - | list/approve, account selector |
 | `nodes` | ✅ | ❌ | P3 | Device management, remove/clear flows |
--- a/README.md
+++ b/README.md
@@ -292,6 +292,11 @@ Engine v2 is opt-in right now. If you want to run the new engine instead of the 
 # First-time setup (configures database, auth, etc.)
 ironclaw onboard
 
+# Migrate from OpenClaw or Hermes into IronClaw
+# (source builds: compile with the opt-in `migrate` feature)
+cargo run --features migrate -- migrate openclaw
+cargo run --features migrate -- migrate hermes --all-profiles
+
 # Start interactive REPL
 cargo run
 
--- a/README.md
+++ b/README.md
@@ -296,6 +296,8 @@ ironclaw onboard
 # (source builds: compile with the opt-in `migrate` feature)
 cargo run --features migrate -- migrate openclaw
 cargo run --features migrate -- migrate hermes --all-profiles
+# Hermes migrations preserve raw auth.json as an encrypted backup secret.
+# After verifying the import, rotate migrated credentials and remove the backup if you no longer need it.
 
 # Start interactive REPL
 cargo run
--- a/src/bridge/mod.rs
+++ b/src/bridge/mod.rs
@@ -15,6 +15,7 @@ mod store_adapter;
 mod workspace_reader;
 
 pub use cost_guard_gate::CostGuardBudgetGate;
+pub use store_adapter::HybridStore;
 pub use workspace_reader::WorkspaceReaderAdapter;
 
 pub use effect_adapter::EffectBridgeAdapter;
--- a/src/cli/import.rs
+++ b/src/cli/import.rs
@@ -1,162 +1,17 @@
-//! Import command for migrating data from other AI systems.
+//! Deprecated CLI alias for `ironclaw migrate`.
 
-use std::path::PathBuf;
-use std::sync::Arc;
+#[cfg(feature = "migrate")]
+pub type ImportCommand = crate::cli::migrate::MigrateCommand;
 
-use clap::Subcommand;
-
-#[cfg(feature = "import")]
-use crate::import::ImportOptions;
-#[cfg(feature = "import")]
-use crate::import::openclaw::OpenClawImporter;
-
-/// Import data from other AI systems.
-#[derive(Subcommand, Debug, Clone)]
-pub enum ImportCommand {
-    /// Import from OpenClaw (memory, history, settings, credentials)
-    #[cfg(feature = "import")]
-    Openclaw {
-        /// Path to OpenClaw directory (default: ~/.openclaw)
-        #[arg(long)]
-        path: Option<PathBuf>,
-
-        /// Dry-run mode: show what would be imported without writing
-        #[arg(long)]
-        dry_run: bool,
-
-        /// Re-embed memory if dimensions don't match target provider
-        #[arg(long)]
-        re_embed: bool,
-
-        /// User ID for imported data (default: 'default')
-        #[arg(long)]
-        user_id: Option<String>,
-    },
-}
-
-/// Run an import command.
-#[cfg(feature = "import")]
+#[cfg(feature = "migrate")]
 pub async fn run_import_command(
     cmd: &ImportCommand,
     config: &crate::config::Config,
 ) -> anyhow::Result<()> {
-    match cmd {
-        ImportCommand::Openclaw {
-            path,
-            dry_run,
-            re_embed,
-            user_id,
-        } => run_import_openclaw(config, path.clone(), *dry_run, *re_embed, user_id.clone()).await,
-    }
+    crate::cli::run_migrate_command(cmd, config).await
 }
 
-/// Run the OpenClaw import.
-#[cfg(feature = "import")]
-async fn run_import_openclaw(
-    config: &crate::config::Config,
-    openclaw_path: Option<PathBuf>,
-    dry_run: bool,
-    re_embed: bool,
-    user_id: Option<String>,
-) -> anyhow::Result<()> {
-    use secrecy::SecretString;
-
-    // Determine OpenClaw path
-    let openclaw_path = if let Some(path) = openclaw_path {
-        path
-    } else if let Some(path) = OpenClawImporter::detect() {
-        path
-    } else {
-        let home = std::env::var("HOME").unwrap_or_else(|_| ".".to_string());
-        PathBuf::from(home).join(".openclaw")
-    };
-
-    let user_id = user_id.unwrap_or_else(|| "default".to_string());
-
-    println!("🔍 OpenClaw Import");
-    println!("  Path: {}", openclaw_path.display());
-    println!("  User: {}", user_id);
-    if dry_run {
-        println!("  Mode: DRY RUN (no data will be written)");
-    }
-    println!();
-
-    // Initialize database
-    let db = crate::db::connect_from_config(&config.database)
-        .await
-        .map_err(|e| anyhow::anyhow!("Failed to initialize database: {}", e))?;
-
-    // Initialize secrets store with master key from env or keychain
-    let secrets_crypto = if let Ok(master_key_hex) = std::env::var("SECRETS_MASTER_KEY") {
-        Arc::new(
-            crate::secrets::SecretsCrypto::new(SecretString::from(master_key_hex))
-                .map_err(|e| anyhow::anyhow!("Failed to initialize secrets: {}", e))?,
-        )
-    } else {
-        match crate::secrets::keychain::get_master_key().await {
-            Ok(key_bytes) => {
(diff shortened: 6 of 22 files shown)


zmanian (changes requested):
Nice direction and useful shim for `import` → `migrate`. Blocking items below; most are mechanical. Main substantive concern is secret idempotency on re-run.

## Blocking

**1. Clippy failures (CI red).** `cargo clippy --all --all-features -- -D warnings` produces 5 errors — all trivial:
- `src/migrate/hermes.rs:760` — collapsible `if` / `if let`
- `src/migrate/openclaw.rs:318` — collapsible `if` / `if let`
- `src/migrate/mod.rs:474` — redundant closure: `.map(|m| imported_message_to_thread_message(m))` → `.map(imported_message_to_thread_message)`
- `src/migrate/mod.rs:645` — collapsible `if`
- one more `if let` in the same region

**2. PR validation checkboxes** in the description itself are unchecked for clippy, `cargo build`, and `cargo test --features integration`. Please run and tick (or explain why n/a).

## Security / correctness

**3. Secret idempotency.** `MigrationServices::store_secret` calls `secrets_store.create(user_id, params)`. If a secret with the same name already exists (user re-runs `migrate hermes --all-profiles`, or OpenClaw + Hermes both map `OPENAI_API_KEY`), `create` will error and abort partway through secret writes, leaving partially-migrated state. Options:
- upsert (preferred),
- `get`-probe + skip-with-note on conflict, or
- document "target must be empty" and fail-fast up front.

Either way, add an integration test for re-running migrate on an already-migrated target.

**4. Raw `auth.json` stored as encrypted backup secret.** Reasonable, but `stats.push_note` alone makes it easy to miss that raw Hermes credentials now live in IronClaw's secret store indefinitely. Surface this in the final CLI summary (not just a buried note) and in the README section, with a recommendation to rotate after migration.

**5. `unwrap_or(Value::Null)` on serde_json::to_value** (~src/migrate/hermes.rs:1070, 1076) silently drops serialization failures. These types are yours, so failure is a bug — but swallowing it turns the bug into a silent data-loss migration. Use `?` with `MigrationError::Serialization`.

## Scope / shape

**6. Size.** +2890/−186, 22 files, and three new files over 450 LOC each. Strongly prefer splitting:
- (a) migration framework (`MigrationServices`, shared types, feature rename `import` → `migrate`, `ImportCommand` compat shim, `HybridStore` export)
- (b) OpenClaw adapter port
- (c) Hermes adapter

Each reviewable independently; (a) is the risky piece.

**7. `ImportCommand` compat alias.** `pub type ImportCommand = crate::cli::migrate::MigrateCommand;` is fine as a shim, but please add `#[deprecated(note = "use `ironclaw migrate` instead; removed in <x.y>")]` and a `--help` note so users know it's slated for removal.

## Nits

- `dry_run` bumps `stats.settings` / `stats.secrets` via a separate code path from the real write. Add a test asserting dry-run stats == real-run stats for the same input to prevent drift.
- `dedupe_secrets` keeps the last entry (HashMap overwrite). Confirm ordering is deterministic and that "last wins" is the intended semantic when `.env` and `auth.json` both provide the same key.

henrypark133 (changes requested):
Findings:
1. Medium - `src/migrate/mod.rs:361`: the unchanged-document check still appears to compare freshly timestamped JSON, so reruns can treat unchanged documents as modified and churn imports instead of staying idempotent.
2. Medium - `tests/migrate_cli_integration.rs:1`: the new coverage drives `run_migrate_command_with_services`, but not the actual `Cli`/`main.rs` dispatch path for `ironclaw migrate ...`, so clap wiring and feature-gate regressions can slip through.

Summary:
- Residual risk: this review was diff-only because the required per-PR worktree fetch failed in the sandbox with `cannot open '.git/FETCH_HEAD': Operation not permitted`.
- Additional note: I verified and dropped the proposed secret-write blocker. In this repo, `SecretsStore::create(...)` is upsert-based in both durable backends, so reruns should not fail solely because the destination secret already exists.

serrrfirat:
## Addressed review feedback (zmanian, henrypark133, gemini-code-assist)

### Correctness / bugs
- **Clippy (5 errors)** — Fixed all collapsible `if`/`if let` and redundant closure lint errors ✅
- **Document unchanged check** — Replaced full JSON comparison (which included `updated_at = now()`, always differing) with direct field comparison of `content`, `metadata`, `title`, `tags` ✅
- **Legacy conversation idempotency** — Removed hard error when `ensure_conversation` returns `false` (already exists); migration now proceeds to message reconciliation on re-runs ✅
- **`unwrap_or(Value::Null)` serialization** — Changed `map_primary_scope` to return `Result`, propagating serialization errors via `MigrationError::Serialization` instead of silently dropping them ✅

### Naming / deprecation
- **`ImportCommand` compat alias** — Added deprecation doc-comment noting it will be removed in a future release ✅

### Documentation / UX
- **auth.json backup visibility** — Security-sensitive notes (auth.json backup, credentials) are now surfaced in a prominent `SECURITY:` section in the CLI summary output, with a credential rotation recommendation ✅

### Tests
- **Idempotency regression test** — Added `rerun_migration_is_idempotent` test that runs OpenClaw migration twice and verifies the second run succeeds without errors ✅

### Not addressed (by design)
- **Secret idempotency** — henrypark133 verified `SecretsStore::create(...)` is already upsert-based; no change needed
- **PR size split** — Noted for future; not actionable in this fix pass
- **dry_run stats drift** / **dedupe_secrets ordering** — Acknowledged as nits; current behavior is intentional

serrrfirat:
Addressed the remaining review feedback I found in this pass:

- added a binary integration test that exercises `ironclaw migrate openclaw ...` through the real CLI/main dispatch path
- added a deprecation/help note for the hidden `ironclaw import` alias and marked the public alias deprecated
- documented the Hermes `auth.json` encrypted-backup behavior in the README, including a credential-rotation recommendation
- updated the PR validation section with the clippy/build results and an explanation for the integration-suite checkbox
- resolved the two outdated Gemini review threads that were already fixed by earlier migration idempotency changes

At this point I do not see any unresolved actionable review threads on the PR; the remaining `CHANGES_REQUESTED` state appears to be stale review status.

zmanian (commented):
## Review: feat:(cli) Add Engine V2 migrate CLI for OpenClaw and Hermes

Solid foundation — typed `Imported*` envelopes, deterministic UUIDv5 keys, dry-run mode, opt-in feature gate, and a CLI binary integration test that drives `CARGO_BIN_EXE_ironclaw` end-to-end. Architecture is right and most of the code is clean. Three classes of issue block merge in the current form.

### CRITICAL

**C1. `auth.json` plaintext lifecycle + fragile security-note partitioning**
`src/migrate/hermes.rs:1119-1131`; `src/cli/migrate.rs:421-432`
`auth.to_string()` is called on a `serde_json::Value` after the value is already in process memory in plain `String` form — the `SecretString` wrapper only covers the post-conversion copy. The plaintext from `read_optional_json` lives in a non-zeroizing `String` for the duration of `map_primary_scope`. Separately, `print_summary`'s "SECURITY:" partition uses `note.contains("auth.json") || note.contains("credential")` substring matching: any future note mentioning "credential" gets the SECURITY banner; any note that *should* be flagged but uses different wording slips into the general bucket. The "credential" branch is currently dead.

Fix: read `auth.json` directly into a `SecretString` (don't go through `Value`) when the only purpose is back-up storage; keep the `Value` form for `active_provider` extraction in a tightly-scoped block. Replace the substring-based partition with a typed `MigrationNote { severity, message }`.

**C2. Path-traversal defense thin on workspace writes**
`src/migrate/hermes.rs:1239,1279`; `src/migrate/openclaw.rs:2576,2607`
`workspace_path = format!("imports/hermes/{}/{}", slugify(&scope.name), rel)` uses `rel` from `normalize_relative_path` (which strips `..` — good) and `slugify(&scope.name)` of user-supplied profile dir names. If the source root contains a symlink that resolves outside `scope.root`, `strip_prefix` may still produce a relative path that traverses. The DB-backed virtual workspace contains the blast radius, but a drift could clobber `AGENTS.md`/`SOUL.md`.

Fix: after building `workspace_path`, assert it starts with `imports/<source>/` and contains no `..` segment. Reject (note + skip) on violation. Same for OpenClaw. Add a `canonicalize` step on `path` before `strip_prefix`.

**C3. Multi-store conversation/document writes are not transactional**
`src/migrate/mod.rs:2107-2156` (conversation), `1960-1975` (document)
`upsert_conversation` does `save_thread` → `save_conversation` → `db.ensure_conversation` → N×`update_conversation_metadata_field` → N×`add_conversation_message` with no transaction boundary. A panic / IO error between step 2 and step 5 leaves Engine V2 with the thread but legacy DB with no messages. Re-running hits the "expected N got M" branch at line 2162 which only prints a note and does not reconcile. The PR description claims idempotency, and the test at line 3422 demonstrates it, but only for the *clean* re-run path — not for partial-failure recovery.

Fix: wrap each conversation's writes in a single DB transaction (`db.with_transaction(|tx| ...)`) and put the engine-store writes last (so they only commit on DB success). Or, implement compare-by-index recovery: if `existing_messages.len() != conversation.messages.len()`, append the missing tail rather than abort. At minimum the README must warn that an interrupted migration may need manual reconciliation.

### HIGH

**H1. Dual-backend rule violated**
`tests/migrate_cli_integration.rs:3058`
`#![cfg(all(feature = "migrate", feature = "libsql"))]` — the entire integration test suite is libSQL-only. CLAUDE.md: *"All new persistence features must support both backends."* Migration writes through `Database`, `SettingsStore`, `SecretsStore` traits which are backend-agnostic, so this *should* work on PostgreSQL — but it's untested. PR description acknowledges `cargo test --features integration` was not run.

Fix: add a Postgres variant of the integration test (gated on `feature = "integration"`), or at minimum a smoke test that exercises `MigrationServices::from_config` with a Postgres backend constructor.

**H2. Compatibility-alias signature changes silently when `migrate` feature is off**
`tests/migrate_cli_integration.rs:3058,3514`; `src/cli/import.rs:257`; `src/cli/mod.rs:475`
The whole integration file is gated on `migrate`, including `import_help_mentions_deprecation`. The without-feature path is non-trivial: `run_import_command(_cmd: &(), ...)` (cmd type changes from `MigrateCommand` to `()`), so a consumer importing `ironclaw::cli::run_import_command` without the feature gets a different signature — silent ABI surface change. No `cargo check --no-default-features` step in CI.

Fix: keep `run_import_command` returning a stable signature (e.g. with-feature also accepts `&MigrateCommand` via `Into<()>` dispatch), or remove the no-feature shim entirely. Add a `cargo check --no-default-features` CI step.

**H3. `apply_settings_patch` clobbers existing builtin override fields — data loss**
`src/migrate/mod.rs:1863-1875`
The merge loops `overrides.insert(provider_id, override_value)` — full replacement, not field-level. If a user has `LlmBuiltinOverride { api_key: Some(real_key), model, base_url }` for `openai`, and the migration imports a Hermes provider mapping for `openai` with `api_key: None`, **the existing api_key is silently nulled out**. The `CustomLlmProviderSettings` path above does field-level upsert by ID (line 1849); builtin overrides do not. CLAUDE.md: *"LLM data is never deleted."*

Fix: mirror the custom-provider merge: load the existing override, only overwrite fields where the incoming value is `Some`, preserve existing `api_key`. Add a unit test where pre-existing overrides survive migration.

**H4. SQL column-name interpolation from a user-supplied sqlite source**
`src/migrate/hermes.rs:1427-1442, 1499-1514, 1555`
`select_text_expr` and the `ORDER BY` branches embed column names directly via `format!`, sourced from `PRAGMA table_info({table})` of a *user-supplied* sqlite file. A maliciously crafted Hermes `state.db` could create a table whose column name is `1; DROP TABLE x; --` and the migration would template it into the next query unsanitized. Limited blast radius (the connection is to the source file the user already owns), but worth defense in depth.

Fix: whitelist column names against a known allowlist (`role`, `content`, `timestamp`, `created_at`, `tool_name`, etc.) before interpolating; reject unknown names with a note. Confirm via the columns set rather than re-using the raw string.

**H5. Read-then-write race on conversation message check**
`src/migrate/mod.rs:2141-2168`
`list_conversation_messages_paginated(...)` is read-then-write without a lock. Two concurrent migrate runs (or any concurrent agent activity in the same conversation) both see `existing_messages.is_empty()` and both insert. A migration after the user has added one chat in the legacy UI will silently abort with the "Legacy conversation already existed" note instead of merging.

Fix: wrap in a transaction with `SELECT ... FOR UPDATE` (Postgres) or `BEGIN IMMEDIATE` (libSQL). At minimum, document that migration should not be run concurrently with the agent.

### MEDIUM

**M1.** Undocumented `+1` off-by-one: `let expected_len = conversation.messages.len() as i64 + 1;` (`src/migrate/mod.rs:2141`) — but next branch compares `existing_messages.len() == conversation.messages.len()` (no +1). Comment why, or align.

**M2.** `slugify("???")` returns `"default"`, colliding with the actual `"default"` profile (`src/migrate/mod.rs:2237`). Fall back to a hash of the raw input when slug is empty.

**M3.** `serde_yml` and `dotenvy` parsing have no size cap (`src/migrate/hermes.rs:857, 949`). 1-GB `config.yaml` OOMs migrate. YAML is also a known DoS vector (anchor expansion). Reject files > ~16 MB before reading; verify `serde_yml` version inherits libyaml's alias-expansion limits.

**M4.** UTF-16 BOM / Latin-1 source files abort the whole scope (`src/migrate/mod.rs:2424`, `hermes.rs:1234`). Migrations should skip + note unreadable files, not abort on a `?`.

**M5.** `MigrationServices::from_config` lost the keychain fallback that the deleted `cli/import.rs` had (lines 190-203 of the old code). Users who set the master key only in the keychain (default `ironclaw onboard` flow) cannot run `migrate`.

**M6.** Silent-skip bug: `serde_json::to_value(&thread).ok() == serde_json::to_value(value).ok()` returns `true` when both fail to serialize (both `None`), so everything-different gets marked "same" and skipped (`src/migrate/mod.rs:2098-2105`). Compare only when both succeed.

### LOW

**L1.** Add module-level `// dispatch-exempt: CLI-driven migration; no agent/UI input.` per CLAUDE.md, since `MigrationServices` calls `db.*`/`secrets_store.*` directly.

**L2.** `CreateSecretParams::new` likely copies the `&str` from `secret.value.expose_secret()` into a non-zeroizing buffer (`src/migrate/mod.rs:1890`). Verify the secret-store intake path zeroizes.

**L3.** `read_dir(...).flatten()` in `resolve_scopes` (`hermes.rs:803`) silently swallows individual entry errors. Push a note instead.

**L4.** `parse_timestamp_str` accepts bare-integer epoch but only seconds (`src/migrate/mod.rs:2444`). A Hermes DB storing `created_at` as ms lands timestamps in 1970. Add a magnitude branch (`> 10^11 → ms`).

### Verdict

**Request changes.** The architecture is right. The blockers: (1) data-loss in `apply_settings_patch` (H3), (2) absence of transactional boundaries on multi-store conversation/document writes leaving partial-failure states with no recovery (C3), (3) violation of the dual-backend rule (H1). C1's `auth.json` plaintext lifecycle and H4's column-name interpolation also warrant a fix before this leaves draft. Once H1–H3 and C1–C3 are addressed (and ideally M1–M6), this is mergeable.

PR #2513, opened 2026-04-16
fix(telegram): emit chat_type metadata for group-safe prompt behavior

rozgo:
## Summary
- emit `metadata.chat_type` for inbound Telegram messages using Telegram's source chat type
- keep `TelegramMessageMetadata` backward-compatible by making `chat_type` optional on deserialize
- add Telegram regression coverage for legacy metadata parsing and group message emission

## Linked Issue
Fixes #2482

## Change Type
- Bug fix

## Review Track
- B

## Validation
- `cargo fmt --all -- --check`
- `cargo clippy --all --benches --tests --examples --all-features -- -D warnings`
- `cargo build`
- `cargo test test_telegram_message_metadata_deserializes_without_chat_type` (in `channels-src/telegram`)
- `cargo test --features integration --test telegram_auth_integration test_group_message_emits_chat_type_metadata -- --nocapture`
- `cargo test`

`cargo test` currently fails in an unrelated upstream e2e assertion:

`tests::tool_info_clarifies_message_and_channel_setup_roles` in `tests/e2e_builtin_tool_coverage.rs`

This failure is outside the Telegram diff.

## Impact
- Security Impact: none
- Database Impact: none
- Blast Radius: Telegram metadata emission and downstream group/private prompt detection
- Rollback Plan: revert this commit

PR #2744, opened 2026-04-20
fix(tests): close staging test backlog — full suite green

ilblackdragon:
## Summary

A pass over `staging` turned up 12 rust test failures and 14 playwright e2e failures + 1 fixture error. All have been root-caused and fixed across three commits on this branch.

- **Rust suite:** 12 → 0 failures
- **Playwright e2e:** 14 failed + 1 error → **0 failed, 0 errored** (401 passed, 8 skipped, 17:31 full run)
- **fmt / clippy --all --benches --tests --examples --all-features:** clean

Every remaining skip is intentional and documented (e.g. requires the `google-drive` WASM binary).

## Production changes (not test-only)

Most fixes in this PR are test plumbing, but three change production behavior. Calling them out so reviewers don't skim past them:

- **`score_skill` scoring change** (`crates/ironclaw_engine/orchestrator/default.py:379`). The skill's own name (≥ 4 chars, with hyphen/underscore → space normalization) is now an implicit activation keyword. This changes which skills activate in production on bare-name mentions. The ≥ 4-char guard stops `code`/`shell`/`mail` from matching, but skill names like `video-meeting` will now auto-match the phrase "video meeting". Overfire regression is still guarded by the existing sibling test in the skill-scoring suite.
- **Denied-action message reword** (`src/bridge/router.rs:2571`). `"Do not execute it; choose an alternative approach"` → `"Do not retry; choose a different approach"`. The old wording contained `"execute it"`, which `user_signals_execution_intent` classified as fresh user execution intent and re-armed `require_action_attempt` on resume — the opposite of a denial. This is a prompt-input change on every denial.
- **ZIP extractor hardening** (`src/tools/builtin/skill_tools.rs`). Previously the decompression budget was checked against `ZipFile::size()` (attacker-controlled central-directory metadata). Now `take(MAX+1)` caps the actual read and the post-read length is enforced, so a forged `uncompressed_size` cannot trick the extractor into materializing an unbounded payload. Regression-pinned by `test_zip_extract_forged_metadata_rejected`.

## Rust fixes

- `tools::builtin::skill_tools` (8 tests): hand-rolled ZIP byte blobs lacked an EOCD record since the extractor switched to `zip::ZipArchive::new` in #2385. Rewrote to use `zip::ZipWriter`. Dropped the obsolete nested-path test — its assumption conflicts with intentional GitHub-archive root stripping.
- `extensions::manager::test_telegram_token_colon_preserved_in_validation_url`: the pairing test's `IRONCLAW_TEST_TELEGRAM_API_BASE_URL` runtime-env mutation wasn't gated by `ENV_MUTEX`. Added `lock_env()`.
- `bridge::router::handle_with_engine_persists_attachment_files_and_indexes_them`: two distinct `ENGINE_STATE_TEST_LOCK` statics meant cross-module tests raced on the shared `ENGINE_STATE` `OnceLock`. Replaced the private duplicate with `use super::test_support::ENGINE_STATE_TEST_LOCK`.
- `e2e_attachments::engine_v2_channel_attachments_persist_for_telegram_and_whatsapp`: attachment persistence resolves paths through the cached `bootstrap::ironclaw_base_dir()`, not the test's tempdir CWD. Added `bridge::override_engine_project_root_for_test` for tests.
- `telegram_auth_integration::test_group_message_emits_chat_type_metadata`: fixed locally by rebuilding `channels-src/telegram` (WASM missed the April-17 `chat_type` emit from #2513). CI rebuilds per run; no binary committed.

## Playwright fixes

- `test_chat.py::test_gateway_attachment_flow_renders_thread_and_reaches_llm` + unextractable variant: a legacy `change` listener on `#image-file-input` fired before the unified `handleAttachmentFiles` path, cleared `e.target.value`, and emptied the FileList. Removed the duplicate wiring.
- `test_chat.py::test_slash_autocomplete_shows_commands_and_skills`: `SLASH_COMMANDS` never merged installed skills. Added `refreshSlashSkillEntries()` that fetches `/api/skills` on menu open.
- `test_pending_user_messages.py::test_pending_message_survives_sse_reconnect`: the SSE open handler only reloads history when `disconnectMs > SSE_RELOAD_THRESHOLD_MS`; the test's instant reconnect skipped that. Age `_sseDisconnectedAt` past the threshold before reconnecting.
- `test_pending_user_messages.py::test_welcome_card_hidden_when_pending`: `_create_new_thread` returned `currentThreadId` before the new-thread API round-trip set it. Waits for the id to change now.
- `TestV2EngineSkillInstallFlow` (7 tests):
  - Skill card template didn't render `usage_hint`, `has_requirements`, `has_scripts`, or `install_source_url`. Extended `renderSkillCard`.
  - Deny message `"Do not execute it; choose an alternative approach"` accidentally matched `user_signals_execution_intent`'s EXEC_PHRASES (`"execute it"`), re-arming `require_action_attempt` on resume. Rephrased to `"Do not retry; choose a different approach"`. See **Production changes** above.
  - `test_implicit_skill_activation`: the pika skill manifest uses a legacy nested shape without `activation:`, so `score_skill` scored 0 and the skill never activated on a bare-name mention. `score_skill` now treats the skill name (≥4 chars, with hyphen/underscore normalization) as an implicit keyword. The sibling overfire guard test still passes. See **Production changes** above.
  - `test_duplicate_install_is_idempotent`: the test was waiting for an approval card, but `SkillInstallTool::requires_approval` intentionally short-circuits to `ApprovalRequirement::Never` when the skill is already loaded. Rewrote the test to assert on the idempotent "already installed / no install needed" message.
  - Mock LLM: `match_tool_call` was re-emitting a matching tool call on every LLM round because "last user content" doesn't change across turns. Added a guard that falls through to the text-response path when the matching tool_name is already in `recent_tool_results`.
- `test_v2_engine_oauth_google::test_oauth_token_refresh_on_expiry`: three compounding issues — mock proxy expects `client_id=hosted-google-client-id` (fixture was `test-google-client-id`), the SSRF guard blocks loopback proxy URLs without `IRONCLAW_OAUTH_PROXY_ALLOW_LOOPBACK=1`, and `credential_spec_to_oauth_refresh` needs an `oauth:` block on the skill to register a refresh config. Fixed all three. Paste-based auth doesn't store a refresh_token, so the test now skips in that configuration (matches the existing `test_oauth_redirect_flow` pattern).
- `test_v2_auth_oauth_matrix::test_repl_http_auth_prompt_accepts_token_and_retries`: `CLI_MODE` defaults to `tui` (ratatui full-screen UI) which reads stdin keystroke-by-keystroke. The PTY-driven tests in this file send whole lines and match for text in the raw stream — under TUI that send never dispatches to the agent, so the auth prompt never appears. Pin `CLI_MODE=repl` on the fixture.
- `test_v2_engine_approval_flow::test_always_approve_survives_restart`: fixture subprocess died at startup with `Channel webhook_server failed to start: Failed to bind to 127.0.0.1:8080: Address already in use`. The fixture allocated a free `GATEWAY_PORT` but left `HTTP_HOST`/`HTTP_PORT` unset so the HTTP channel tried to claim default port 8080. The 60s `/api/health` timeout was a red herring — the subprocess's stderr pipe filled up before the child could log the bind failure. Fixed: allocate a second free port for `HTTP_PORT`, drain stdout/stderr in background tasks, surface stderr tail in timeout message.

## New coverage

- `test_chat.py::test_message_copy_button_writes_raw_text`: the per-message Copy button (`.message-copy-btn` → `copyMessage(btn)` in `render.js:339`) had no tests. The existing `test_copy_from_chat_forces_plain_text` only covered the Cmd+C selection handler. Added coverage for both user (plain text via `data-copy-text`) and assistant (raw markdown via `data-raw`) paths, the "Copied!" flash, and the 1.2s revert.

## Test plan
- [x] `cargo fmt --check`
- [x] `cargo test` — all green
- [x] `cargo test --features integration` — all green (PostgreSQL-requiring tests skip cleanly without DB)
- [x] `pytest scenarios/` — **401 passed, 8 skipped, 0 failed, 0 errored** (17:31; down from 14 failed + 1 error)
- [ ] CI rebuild of `channels-src/telegram` WASM picks up the `chat_type` emit


src/tools/builtin/skill_tools.rs
```diff
@@ -2686,58 +2664,53 @@ mod tests {
...
-        zip.extend_from_slice(&[0x00, 0x00, 0x00, 0x00]); // mod time/date
-        zip.extend_from_slice(&[0x00, 0x00, 0x00, 0x00]); // crc32
-        zip.extend_from_slice(&(content.len() as u32).to_le_bytes()); // compressed size
-        zip.extend_from_slice(&(content.len() as u32).to_le_bytes()); // uncompressed size
-        zip.extend_from_slice(&(file_name.len() as u16).to_le_bytes()); // filename length
-        zip.extend_from_slice(&0u16.to_le_bytes()); // extra field length
-        zip.extend_from_slice(file_name.as_bytes());
-        zip.extend_from_slice(content);
-        zip
+    /// Helper: build a minimal ZIP with a single file using `zip::ZipWriter`
```
serrrfirat: **High Severity**

These rewritten regression tests now only exercise well-formed `ZipWriter` archives, but `extract_skill_from_zip()` still enforces its decompression budget using `ZipFile::size()` metadata before doing an unbounded `read_to_end()` (around lines 1605-1637). With `zip 2.4.2`, a forged archive can report a tiny uncompressed size in the central directory while `read_to_end()` still yields the full deflated payload — I reproduced a case where `size()==10` but `read_to_end()` returned 100000 bytes.

So this test suite can now go green while the DoS hole remains open.

Could we keep at least one adversarial fixture by patching the size fields after `ZipWriter` output, or change the extractor to cap the **actual** read (for example with `take(MAX_ZIP_ENTRY_BYTES + 1)`) instead of trusting metadata?

tests/e2e/scenarios/test_v2_engine_oauth_google.py
```diff
@@ -404,7 +410,16 @@ async def v2_google_server(ironclaw_binary, mock_llm_server, mock_google_api):
         "WASM_TOOLS_DIR": wasm_tools_dir,
         "WASM_CHANNELS_DIR": os.path.join(home_dir, ".ironclaw", "wasm_channels"),
         "ONBOARD_COMPLETED": "true",
-        "GOOGLE_OAUTH_CLIENT_ID": "test-google-client-id",
+        # Match the `hosted-google-client-id` value the mock LLM's
+        # /oauth/refresh handler expects so `refresh_token_via_proxy`
+        # succeeds on the expired-token path. Other E2E fixtures use the
+        # same value (see tests/e2e/conftest.py:600, :901).
+        "GOOGLE_OAUTH_CLIENT_ID": "hosted-google-client-id",
```
serrrfirat: **Medium Severity**

After changing the fixture to use `hosted-google-client-id` here, `test_oauth_redirect_flow()` still asserts that the auth URL contains `client_id=test-google-client-id` later in the file.

Once the real OAuth/WASM path is exercised, that assertion will fail even if the implementation is correct.

Please derive the expected client id from the fixture/env in one place, or update the assertion to match the new configured value.

tests/e2e/scenarios/test_v2_engine_oauth_google.py
```diff
@@ -733,6 +748,25 @@ async def test_oauth_token_refresh_on_expiry(v2_google_server, mock_google_api):
...
     stored_user_id, expires_before, updated_before = _find_secret_row(db_path, secret_name)
 
+    # The refresh path needs both an access token AND a refresh token on
+    # disk — without the refresh token, `refresh_oauth_access_token` hits
+    # "No refresh token available" and falls straight through to the
+    # auth gate instead of renewing the token. Paste-based auth (the
+    # earlier tests' fallback when no google-drive WASM binary is
+    # available) only stores the access token; skip this test in that
+    # configuration rather than asserting refresh worked.
+    refresh_secret_name = f"{secret_name}_refresh_token"
```
serrrfirat, line 765: **High Severity**

This new skip path fixes the missing-refresh-token prerequisite, but the body of `test_oauth_token_refresh_on_expiry()` still only verifies refresh behavior inside `if oauth_state.get("refresh_count", 0) >= 1:`.

That means a regression where the expired-token path never calls the refresh endpoint still passes as long as the thread produces *some* response.

Could we make `refresh_count >= 1` a required assertion in the non-skipped path, and then verify the DB timestamps / token use afterward? Otherwise this stays a false-green for the exact behavior the test claims to pin.

crates/ironclaw_gateway/static/js/surfaces/chat.js
```diff
@@ -693,7 +709,13 @@ function handleAttachmentFiles(files) {
   const fileInput = document.getElementById('image-file-input');
   if (fileInput) {
     fileInput.addEventListener('change', (e) => {
-      handleAttachmentFiles(e.target.files);
+      // Snapshot the FileList into an array *before* clearing the input.
+      // Some drivers (e.g. Playwright's set_input_files) expose a live
+      // FileList that turns empty mid-listener-chain; reading it later
+      // silently loses every file. Array.from fixes this by creating a
+      // stable copy while the FileList is still populated.
+      const files = Array.from(e.target.files || []);
```
serrrfirat, line 765: **Medium Severity**

This fixes the live-`FileList` race, but the rest of the new attachment flow still looks inconsistent with the pending-message replay path.

`sendMessage()` now renders/stages document attachments via `pendingAttachmentsForDisplay`, but `_pendingUserMessages` still stores only `{content, images}` and hardcodes `content || '(images attached)'`. `loadHistory()` only knows how to replay those two fields, so if history reloads before the DB write lands (SSE reconnect, thread switch, reload during processing), document uploads disappear entirely and files-only sends come back mislabeled as `"(images attached)"`.

Could the pending cache carry the same attachment payload used for the optimistic render, so reconnects preserve file attachments too?

crates/ironclaw_gateway/static/js/surfaces/chat.js
```diff
@@ -491,11 +478,40 @@ function updateSlashHighlight() {
...
+    }));
+    // Re-filter in place so the menu refreshes with skills without
+    // requiring another keystroke.
+    const input = document.getElementById('chat-input');
+    if (input && input.value.startsWith('/')) {
+      filterSlashCommands(input.value);
+    }
+    done();
+  }).catch(done);
+}
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in cd90991c. Added a 30s TTL (`_INSTALLED_SKILLS_TTL_MS`) on the cache so repeated keystrokes hit the cache instead of `/api/skills`, plus an `invalidateSlashSkillCache()` hook wired into the install/remove flows in `surfaces/skills.js` so the menu refreshes immediately on skill-install/remove without waiting for the TTL to expire.

crates/ironclaw_gateway/static/js/surfaces/chat.js
```diff
@@ -491,11 +478,40 @@ function updateSlashHighlight() {
...
+
 function filterSlashCommands(value) {
   if (!value.startsWith('/')) { hideSlashAutocomplete(); return; }
   // Only show autocomplete when the input is just a slash command prefix (no spaces except /thread new)
   const lower = value.toLowerCase();
-  const matches = SLASH_COMMANDS.filter((c) => c.cmd.startsWith(lower));
+  refreshSlashSkillEntries();
+  const allCommands = SLASH_COMMANDS.concat(_installedSkillsForSlash);
+  const matches = allCommands.filter((c) => c.cmd.startsWith(lower));
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in cd90991c. Same fix as noted on the Gemini thread above: 30s TTL cache + `invalidateSlashSkillCache()` hook on install/remove.

crates/ironclaw_gateway/static/js/surfaces/chat.js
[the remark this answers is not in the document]
ilblackdragon: Fixed in cd90991c. 30s TTL cache added; plus an `invalidateSlashSkillCache()` helper called from install/remove success handlers so the menu picks up changes without waiting for the TTL.

src/bridge/router.rs
```diff
@@ -2432,8 +2432,15 @@ pub async fn resolve_gate(
                 )
                 .await;
 
+            // Word the deny message carefully: the resume handler treats
+            // certain imperative verb phrases ("execute it", "run it", "send
+            // it", …) as fresh execution intent and re-arms the
+            // require_action_attempt obligation, which then nudges the LLM
+            // to issue another tool call — exactly the opposite of what a
+            // denial should produce. Avoid every phrase in
+            // `llm::reasoning::user_signals_execution_intent`'s list.
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in cd90991c. Updated the comment in `src/bridge/router.rs` to reference the public path `crate::llm::user_signals_execution_intent` and cross-reference `src/llm/reasoning.rs` as the defining file.

src/pairing/approval.rs
```diff
@@ -184,7 +184,15 @@ mod tests {
     }
 
     #[tokio::test]
+    // Hold ENV_MUTEX across awaits: the serialization guarantee against sibling
+    // tests reading IRONCLAW_TEST_TELEGRAM_API_BASE_URL outweighs the lint's
+    // concern here — the test is single-threaded tokio and cannot deadlock.
+    #[allow(clippy::await_holding_lock)]
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in cd90991c. You're right that the default deserved a pin. Explicitly set `#[tokio::test(flavor = "current_thread")]` on the test so the justification stays accurate regardless of future defaults, and reworded the comment to spell out which runtime kind is in play.

tests/e2e/scenarios/test_v2_engine_approval_flow.py
```diff
@@ -582,14 +618,26 @@ async def start():
...
 
     async def stop():
         nonlocal proc
         if proc and proc.returncode is None:
             await _stop_process(proc, sig=signal.SIGINT, timeout=10)
             if proc.returncode is None:
                 await _stop_process(proc, sig=signal.SIGTERM, timeout=5)
+        for task in drain_tasks:
+            if not task.done():
+                task.cancel()
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in cd90991c. `stop()` now cancels the drain tasks, clears the list, then `asyncio.gather(*, return_exceptions=True)` awaits them — so stop→start cycles don't leak zombie readers or trigger "Task was destroyed but it is pending!" warnings at teardown.

serrrfirat (dismissed):

src/tools/builtin/skill_tools.rs
ilblackdragon: Fixed in d84edef2 (merge commit addressing this review).

The extractor no longer trusts `ZipFile::size()` metadata for the decompression budget. `extract_skill_from_zip()` now caps the actual read with `(&mut file).take(MAX_ZIP_ENTRY_BYTES + 1)`, verifies `contents.len()` post-read, and accumulates `total_unzipped_bytes` from the actual byte count rather than the declared one (`src/tools/builtin/skill_tools.rs:1598-1666`). The existing `file.size() > MAX_ZIP_ENTRY_BYTES` check is kept as a cheap early reject for honestly-labelled archives, with a comment noting that the post-read check is the authoritative one.

New adversarial regression: `test_zip_extract_forged_metadata_rejected` builds a legitimate 2 MB+1 Stored ZIP, then rewrites the `uncompressed_size` field in **both** the local file header (offset 22) and the central-directory header (offset 24) to claim 10 bytes — exactly the attack you described. The test asserts the extractor still rejects it based on actual bytes read. It fails against the old code and passes against the fix.

tests/e2e/scenarios/test_v2_engine_oauth_google.py
ilblackdragon: Fixed in d84edef2. Introduced a module-level `EXPECTED_GOOGLE_CLIENT_ID = "hosted-google-client-id"` and reference it from both the fixture env (`GOOGLE_OAUTH_CLIENT_ID=EXPECTED_GOOGLE_CLIENT_ID`) and the `test_oauth_redirect_flow` assertion (`params.get("client_id") == [EXPECTED_GOOGLE_CLIENT_ID]`). The two can no longer drift — once the real OAuth/WASM path exercises the assertion, it will compare against whatever the fixture actually configured.

tests/e2e/scenarios/test_v2_engine_oauth_google.py
ilblackdragon: Fixed in d84edef2. You're right — the `if refresh_count >= 1:` block was a silent opt-out. Now that the skip at line 758 already guards the legitimate "no refresh possible" case (missing `{secret}_refresh_token`), any path past it must actually fire the refresh.

The assertion block is now unconditional:

```python
refresh_count = oauth_state.get("refresh_count", 0)
assert refresh_count >= 1, (
    f"Expired-token path MUST call the OAuth refresh endpoint at least "
    f"once; refresh_count={refresh_count}, oauth_state={oauth_state}"
)
# DB timestamp checks follow, also unconditional.
```

A regression where the expired-token path skips the refresh endpoint now fails the test instead of passing as false-green.

crates/ironclaw_gateway/static/js/surfaces/chat.js
ilblackdragon: Fixed in d84edef2 via the merge with staging's #2763. The `_pendingUserMessages` entries now carry the full attachment payload (`attachments: pendingAttachmentsForDisplay.map(...)`, `copyText`), not just `{content, images}`, and `loadHistory()` replays via `addMessage(..., {attachments, copyText})` so document uploads survive SSE reconnect, thread switch, and in-progress reloads.

The hardcoded `'(images attached)'` default inside the pending block is also gone — the shared `displayContent` (computed once before the optimistic render) now feeds both the DOM and the pending cache, so files-only sends stay labeled `'(files attached)'` through replay. New e2e coverage lives in `tests/e2e/scenarios/test_pending_user_messages.py`.

tests/e2e/scenarios/test_v2_engine_approval_flow.py
```diff
@@ -582,14 +618,33 @@ async def start():
...
         )
-        await wait_for_ready(f"{base_url}/api/health", timeout=60)
+        drain_tasks.append(asyncio.create_task(_drain(proc.stdout, None)))
+        drain_tasks.append(asyncio.create_task(_drain(proc.stderr, stderr_tail)))
+        try:
+            await wait_for_ready(f"{base_url}/api/health", timeout=60)
+        except TimeoutError:
+            raise TimeoutError(
+                f"restartable_v2_server did not become ready at {base_url}/api/health.\n"
+                f"stderr tail:\n{bytes(stderr_tail).decode('utf-8', errors='replace')}"
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in dca4e1cf. Good catch — the re-raise ran before the fixture's `try/finally`, so a `wait_for_ready` timeout was leaking the subprocess and its bound ports into subsequent tests.

`start()` now snapshots the stderr tail *before* teardown (since `stop()` drains/cancels the readers), calls `await stop()`, and then re-raises the `TimeoutError` with the captured tail. `stop()` already handles the SIGINT→SIGTERM escalation and `asyncio.gather(..., return_exceptions=True)` on the drain tasks, so no further kill/cancel logic is needed in the timeout branch.

tests/e2e_attachments.rs
```diff
@@ -230,6 +230,15 @@ mod attachment_tests {
...
         for channel in ["telegram", "whatsapp"] {
             let rig = TestRigBuilder::new().with_engine_v2().build().await;
+            // Redirect attachment persistence into the tempdir so the
+            // assertion on saved_path below sees the real write. The engine
+            // otherwise joins paths against the cached `ironclaw_base_dir()`
+            // (first-caller-wins LazyLock), which is `$HOME/.ironclaw` in a
+            // fresh process and unrelated to the per-test tempdir here.
+            assert!(
+                ironclaw::bridge::override_engine_project_root_for_test(project_root.clone()).await,
+                "engine state should be installed after build()"
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in dca4e1cf. You're right — `engine_v2_project_root()` resolved to `$HOME/.ironclaw`'s parent (i.e. `$HOME` itself), so the test was writing real files under `~/.ironclaw/attachments/...` on every run.

The `engine_v2_project_root()` helper is gone. The test now allocates a per-test `tempfile::TempDir` and passes `tempdir.path().to_path_buf()` to `override_engine_project_root_for_test`; the `saved_path` assertion is rooted at the same tempdir, so all writes stay fully contained and are cleaned up on drop. The `engine_v2_attachment_root_lock` mutex is kept — it's still needed to serialize mutations of the process-global engine state against other tests that might install/override it concurrently.

ilblackdragon:
## Code Review

### Overview

Closes a 12-rust + 14-playwright test backlog on staging. Mostly test-only fixes with three meaningful production tweaks: ZIP extractor hardening, a denied-action message reword, and a skill-scoring behavior change. +495 / -88 across 15 files.

### Strengths

- **Root causes over symptoms.** Each fix's in-code comment explains *why* the old state was wrong. The ZIP forged-metadata test and the OAuth refresh-must-fire assertion are genuine regression pins, not bug-checklist entries.
- **Tightens rather than weakens coverage.** `test_oauth_token_refresh_on_expiry` went from `if refresh_count >= 1:` (silent green on no-op) to `assert refresh_count >= 1` with a cleanly-gated skip for paste-only auth. That's a real strengthening.
- **ENGINE_STATE lock consolidation** (`src/bridge/router.rs`) fixes a real cross-module race, not just a flake.
- **Drain-and-capture** in `restartable_v2_server` turns silent 60s timeouts into actionable stderr tails — future debugging win.

### Concerns

1. **`score_skill` implicit-name-as-keyword is a behavior change, not a test fix** (`crates/ironclaw_engine/orchestrator/default.py:379`). The ≥4-char guard stops `code`/`shell`/`mail` from matching, but a skill named `video-meeting` or `daily-standup` will now auto-match common phrases via the hyphen→space normalization. PR body says the overfire guard test still passes; worth naming which guard (`test_keyword_overfire_*`?) in the comment so a future reader can verify the invariant hasn't regressed. Consider: only apply the implicit name when `activation.keywords` is empty/missing, rather than always appending.

2. **Deny-message rewording papers over the classifier** (`src/bridge/router.rs:2571`). The real defect is that `user_signals_execution_intent` classifies the *system-authored* deny message as user execution intent. The fix is correct tactically, but the brittleness remains: anyone rephrasing this message in the future re-opens the bug. Add a `#[test]` that feeds the exact deny message through `user_signals_execution_intent` and asserts `false` — that locks the contract regardless of wording.

3. **Production-behavior changes not called out in PR title.** Per `.claude/rules/review-discipline.md` § PR Scope, the `score_skill` scoring change and the deny-message reword are production behavior, not test plumbing. Surface them in the body under a "Production changes" section so reviewers don't skim past them as test-only.

4. **Drain tasks reset across restart.** `drain_tasks.append` in `start()` plus `drain_tasks.clear()` in `stop()` is correct only if every `start()` is paired with a `stop()`. A test that calls `start()` twice without `stop()` would leak readers. Not a bug today — restart tests call `stop()` → `start()` — but fragile. A short assertion `assert not drain_tasks` at the top of `start()` would lock it.

5. **`Array.from(e.target.files)` comment overstates Playwright behavior.** Playwright's `set_input_files` yields a regular `FileList`; the real failure mode is the `e.target.value = ''` line clearing `files` *on the same element*. The fix is correct, but the comment misattributes the cause. Trim to: "read before clearing `e.target.value` on the same listener."

6. **Mock-LLM re-dispatch guard is coarse.** `match_tool_call` now bails on *any* prior call with the same tool name. If a scenario legitimately calls the same tool twice (e.g. two `file_read` on different paths), the second call silently falls through to text. Low blast radius in the current fixtures, but worth a comment flagging the constraint.

### Correctness spot-checks

- `extract_skill_bundle_from_zip`: the `take(MAX + 1)` + post-read length check correctly defeats forged `uncompressed_size`. Good. ✓
- `override_engine_project_root_for_test`: correctly `#[cfg(feature = "libsql")]` on both the definition and the `pub use` — callers in `tests/e2e_attachments.rs` are already libsql-gated. ✓
- `ENV_MUTEX` + `current_thread` runtime on `propagate_approval_restores_runtime_state_when_on_start_fails`: the `await_holding_lock` suppression reasoning (no concurrent tasks on current-thread runtime) is sound. ✓

### Risks

- None for Rust/Python test-only changes.
- `score_skill` change can subtly alter which skill activates in production. Recommend a follow-up metric on skill-selection divergence vs. staging before/after.
- Deny-message reword changes the LLM's prompt input on every denial — small distribution shift. Low risk.

### Verdict

Ship with the PR body updated to call out the two production-behavior changes (`score_skill`, deny message), and ideally with the follow-up test pinning the deny message against `user_signals_execution_intent`. Everything else is solid.

tests/e2e/scenarios/test_v2_engine_oauth_google.py
```diff
@@ -733,6 +755,25 @@ async def test_oauth_token_refresh_on_expiry(v2_google_server, mock_google_api):
...
+    try:
+        _find_secret_row(db_path, refresh_secret_name)
+    except AssertionError:
+        pytest.skip(
```
serrrfirat, line 769: Because CI doesn't build `tools-src/google-drive` anywhere (`.github/workflows/coverage.yml` only builds WASM channels, and `e2e.yml` doesn't run this file), this new skip means the Google OAuth refresh path can still go completely unexercised in the common test setup. In that environment `test_oauth_redirect_flow()` already skips when no `auth_url` is produced, and this branch now makes `test_oauth_token_refresh_on_expiry()` skip too, so the suite goes green without actually pinning the refresh regression.

Could we either build the `google-drive` WASM for this scenario, or add a smaller mock path that exercises refresh without requiring a prebuilt binary?

Issue #2546, opened 2026-04-16
[Bug Bash 4/16] Orchestrator surfaces raw HTTP 502 Bad Gateway error to user


joe-rlo:
## Bug Report — QA Bug Bash 4/16

**Environment:** Staging (hosted-staging / crab shack) — https://agent-stg.near.ai/

**Description:**
The orchestrator surfaces a raw error traceback to the user when the LLM provider returns a 502 Bad Gateway response. The full error includes internal file paths (orchestrator.py lines 907, 548), Python traceback, and raw JSON error payload. The agent status badge also shows "Loading..." instead of providing a graceful error state.

**Error displayed to user:**
`Error: Orchestrator error: effect execution error: Orchestrator error after resume: Traceback (most recent call last): File "orchestrator.py", line 907, in File "orchestrator.py", line 548, in run_loop RuntimeError: LLM call failed: Provider nearai_chat request failed: HTTP 502 Bad Gateway`

**Steps to Reproduce:**
1. Open agent "proud-wolf-damug" on staging (or any agent during provider outage)
2. 2. Send a message when the LLM provider is experiencing issues
**Actual Result:**
Raw Python traceback with internal file paths and error JSON is displayed directly to the user. Agent status shows "Loading..." indefinitely.

**Expected Result:**
A user-friendly error message like "The AI model is temporarily unavailable. Please try again in a few moments." Internal error details should be logged server-side only, not shown to the user.

**Severity:** Medium — Poor error handling exposes internals, but functionally equivalent to a retry scenario
**Frequency:** Intermittent — depends on LLM provider availability

**Related:** #2276 (Orchestrator crashes with HTTP 413), #2408 (Context length overflow)

Issue #2544, opened 2026-04-16
[Bug Bash 4/16] Agent plans and delegates tasks but never executes or completes them


joe-rlo:
## Bug Report — QA Bug Bash 4/16

**Environment:** Staging (hosted-staging / crab shack) — https://agent-stg.near.ai/

**Description:**
When given a task, the agent outlines a plan and appears to delegate or queue sub-tasks, but never actually executes any of them. The conversation ends with the agent describing what it intends to do, but no concrete output or result is delivered.

**Steps to Reproduce:**
1. Open an agent on staging
2. 2. Give the agent a multi-step task (e.g., "research X and create a summary")
3. 3. Observe the agent's response
**Actual Result:**
Agent responds with a plan or delegation structure (e.g., "I'll do X, then Y, then Z") but never follows through. No tools are invoked, no final output is produced. The agent treats planning as completion.

**Expected Result:**
Agent should execute the planned steps and deliver a concrete result, not just describe what it would do.

**Severity:** High — Agent appears functional but delivers no actual output
**Frequency:** Reproducible during bug bash testing

**Related:** #2279 (Bot falsely claims success without completing tasks), #2282 (Bot overrides user instructions with own approach)

Issue #2543, opened 2026-04-16
[Bug Bash 4/16] Linear MCP authorization not recognized after successful OAuth setup


joe-rlo:
## Bug Report — QA Bug Bash 4/16

**Environment:** Staging (hosted-staging / crab shack) — https://agent-stg.near.ai/

**Description:**
After completing Linear MCP OAuth setup, the dashboard shows the integration as "connected." However, when the agent attempts to use the Linear tool, it still prompts the user for an access token — as if the OAuth flow never happened.

**Steps to Reproduce:**
1. Connect Linear MCP integration via OAuth on staging
2. 2. Verify the dashboard shows "Connected" status
3. 3. Ask the agent to perform a Linear operation (e.g., list issues)
**Actual Result:**
Agent does not recognize the authorized connection and requests a manual access token from the user, despite the dashboard confirming the integration is connected.

**Expected Result:**
Once the OAuth flow completes and the dashboard shows "Connected," the agent should use the stored credentials automatically without prompting for manual token entry.

**Severity:** High — MCP integration is non-functional despite appearing connected
**Frequency:** Reproducible during bug bash testing

**Related:** #2229 (Google Sheets OAuth blocked — similar pattern of OAuth integration failure)

serrrfirat:
This looks likely covered by #2515, which merged to `staging` on 2026-04-16 after this bug was filed.

Why it matches this symptom:
- unified chat/settings extension setup paths
- OAuth continuation from setup is treated as `auth_required`, not failure
- OAuth-only auth cards no longer render a manual token textbox
- gate-backed auth resolution now uses scoped structured submissions instead of routing credentials through normal chat
- `tool_install` now follows the real readiness path so LLM-visible tool state matches UI readiness

I also checked nearby auth/MCP PRs:
- #2050 and #2209 fixed broader MCP/OAuth readiness issues, but both merged before this bug was filed, so they probably did not fully cover the staging regression.
- #2376 mentions Linear auth, but it was canary coverage and closed unmerged.
- #2473 was OAuth nonce scoping, but specifically for Slack relay OAuth.

Recommendation: retest Linear MCP OAuth on staging with #2515 deployed. If Linear no longer prompts for a manual token after dashboard shows connected, we can close this as fixed by #2515.

Issue #2541, opened 2026-04-16
[Bug Bash 4/16] Agent fails to invoke tools for content summarization tasks


joe-rlo:
## Bug Report — QA Bug Bash 4/16

**Environment:** Staging (hosted-staging / crab shack) — https://agent-stg.near.ai/

**Description:**
When asked to summarize a social media post, the agent processed for 3-5 minutes but did not invoke any tools. Instead of fetching and analyzing the content, it returned a short, low-quality summary with no evidence of tool usage.

**Steps to Reproduce:**
1. Open an agent on staging
2. 2. Ask the agent to summarize a social media post or web content
3. 3. Wait for the agent to process
**Actual Result:**
Agent processes for 3-5 minutes, does not call any tools, and returns a brief generic summary that doesn't reflect the actual content.

**Expected Result:**
Agent should invoke appropriate tools (e.g., web browsing, content fetching) to retrieve the content and provide an accurate, detailed summary.

**Severity:** Medium — Agent is non-functional for tool-dependent tasks
**Frequency:** Reproducible during bug bash testing

henrypark133:
> Agent processes for 3-5 minutes, does not call any tools, and returns a brief generic summary that doesn't reflect the actual content.

Do you have any specific reproduction steps and/or specific output? This is a little too vague to triage. @joe-rlo

PR #2753, opened 2026-04-20
fix: bug bash 4/16 triage — error boundary, TEE secrets, pairing, rehydration

### LINKED ISSUE #1537: WASM credential injection fails on NEAR AI hosted TEE — secrets stored but never injected at runtime
**Environment:**
- IronClaw version: v0.18.0
- Hosting: NEAR AI Cloud (hosted TEE), basic plan
- Database: libSQL (embedded)
- Instance: `agent3.near.ai`

## Summary

WASM tool and WASM channel credential injection silently fails on the NEAR AI hosted TEE. Secrets are stored successfully in the encrypted secrets table and verified via `ironclaw tool info` and `ironclaw tool auth`, but are never injected into HTTP requests at runtime. This affects both custom WASM tools and registry-installed WASM channels (Slack).

## Reproduction — WASM Tool (custom Matrix send tool)

1. Built a custom WASM tool (`matrix_send_tool`) that sends messages to a Matrix room via PUT to `https://matrix.org/_matrix/client/v3/rooms/{room_id}/send/m.room.message/{txn_id}`
2. Capabilities file declares credential injection:
```json
"credentials": {
  "matrix_access_token": {
    "secret_name": "matrix_access_token",
    "location": { "type": "bearer" },
    "host_patterns": ["matrix.org"]
  }
}
```
3. Secret stored via `ironclaw tool auth matrix_send_tool` — confirmed present with correct name in secrets table
4. `ironclaw tool info matrix_send_tool` shows the tool as authenticated with credential listed
5. When the tool executes, Matrix API returns `M_MISSING_TOKEN` — the Authorization header is not present in the request

### Credential types tested (all fail identically):

- **`"type": "bearer"`** — host should inject `Authorization: Bearer <token>`. Result: `M_MISSING_TOKEN`
- **`"type": "header", "name": "Authorization"`** — stored token as `Bearer <token>`, host should inject as raw header. Result: `M_MISSING_TOKEN`
- **Placeholder injection** — tool code sets header to literal string `"Authorization": "Bearer {matrix_access_token}"`, host should replace `{matrix_access_token}` with secret value (per DeepWiki docs on credential placeholder syntax). Result: `M_MISSING_TOKEN`

All three methods confirmed the HTTP request reaches Matrix (the API responds with a proper error), but the Authorization header is absent every time.

## Reproduction — WASM Channel (Slack, from registry)

1. Installed Slack channel from registry: `ironclaw registry install slack`
2. Stored both `slack_bot_token` and `slack_signing_secret` via `ironclaw tool setup` and confirmed via onboard wizard (`ironclaw onboard --channels-only`)
3. Configuration verified:
   - `channels.wasm_channels = ["slack"]`
   - `channels.wasm_channels_enabled = true`
   - `channels.wasm_channels_dir = /home/agent/.ironclaw/channels`
   - `ironclaw status` shows "Channels: cli, 1 wasm"
4. Webhook endpoint never registers:
```
$ curl -s http://localhost:8080/webhook/slack
{"error":"Channel not found for path","path":"/webhook/slack"}
```
5. This persists across multiple restarts (SSH disconnect/reconnect).

## Observations

- `ironclaw doctor` reports `[skip] Secrets: secrets not configured (run ironclaw onboard)` even though secrets are present in the database and retrievable by name. This suggests the secrets subsystem may not be fully initializing on the hosted TEE.
- The `secrets_master_key_source` is set to `env`, which should work if the TEE provides the environment variable — but the doctor skip message suggests it may not be loading correctly at runtime.
- This may be related to #413 (Slack channel `not_authed` / credential injection failure), which was filed against v0.12.0 and appears to remain unresolved as of v0.18.0.

## Expected behavior

- WASM tools should receive credential injection via any of the three documented methods (bearer, header, placeholder)
- WASM channels installed from the registry should register their webhook routes after proper configuration

## Actual behavior

- Secrets are stored but never injected — all HTTP requests to external APIs arrive without authentication headers
- WASM channels are detected by `ironclaw status` but their webhook routes are never registered on the webhook server

### LINKED ISSUE #1839: Slack pairing can dead-end after activation with no visible code or actionable next step
## Summary

Slack can reach `Configured` / `Awaiting Pairing`, but DM-based pairing is still fragile enough that users can end up with a working installation, no pairing approval prompt, and no clear recovery path.

This is a separate issue from the install/activate credential flow work. That work fixes `install -> activate -> secure credential entry`; this issue is about the **post-activation Slack pairing step**.

Related: #1776

## User Impact

From the user's perspective, Slack can look stuck:

- the extension is installed
- credentials are configured
- the web UI shows `Awaiting Pairing`
- messaging the bot in Slack may not produce a pairing code or any guidance
- retrying may still show no new feedback

That creates a dead-end even though the remaining problem is just pairing/approval.

### 1. Pairing only starts from a true Slack DM event

The Slack channel only attempts pairing for `message` events whose channel id starts with `D`:

- [`channels-src/slack/src/lib.rs#L486-L513`](https://github.com/nearai/ironclaw/blob/staging/channels-src/slack/src/lib.rs#L486-L513)

Channel mentions do not go through DM pairing. Non-DM messages explicitly bypass DM policy:

- [`channels-src/slack/src/lib.rs#L587-L590`](https://github.com/nearai/ironclaw/blob/staging/channels-src/slack/src/lib.rs#L587-L590)

So if the Slack app is not correctly delivering DM events, the user can never start pairing.

### 2. The setup contract does not validate Slack-side event delivery requirements

Slack setup currently asks for:

- `slack_bot_token`
- `slack_signing_secret`

and declares `/webhook/slack` plus `dm_policy: "pairing"`:

- [`channels-src/slack/slack.capabilities.json#L7-L20`](https://github.com/nearai/ironclaw/blob/staging/channels-src/slack/slack.capabilities.json#L7-L20)
- [`channels-src/slack/slack.capabilities.json#L42-L59`](https://github.com/nearai/ironclaw/blob/staging/channels-src/slack/slack.capabilities.json#L42-L59)

But the runtime does not validate whether the Slack app is actually configured to deliver the required DM events or whether the bot can successfully send the pairing reply.

### 3. Pairing instructions are only sent on the first pending request

When a sender is not allowed, Slack calls `pairing_upsert_request(...)`. The pairing reply is only sent when `result.created` is `true`:

- [`channels-src/slack/src/lib.rs#L616-L645`](https://github.com/nearai/ironclaw/blob/staging/channels-src/slack/src/lib.rs#L616-L645)

But `PairingStore::upsert_request(...)` returns `created: false` for repeat attempts by the same sender while reusing the existing code:

- [`src/pairing/store.rs#L224-L310`](https://github.com/nearai/ironclaw/blob/staging/src/pairing/store.rs#L224-L310)

That means:

- first DM may create a pending pairing request
- if the Slack reply fails or the user misses it, later DMs do **not** resend the code
- the user can be stuck with a valid pending request but no visible instructions

### 4. Failures to send the Slack pairing reply are swallowed

The pairing reply is sent via `chat.postMessage`, but the result is discarded with `let _ = send_pairing_reply(...)`:

- [`channels-src/slack/src/lib.rs#L633-L635`](https://github.com/nearai/ironclaw/blob/staging/channels-src/slack/src/lib.rs#L633-L635)
- [`channels-src/slack/src/lib.rs#L648-L668`](https://github.com/nearai/ironclaw/blob/staging/channels-src/slack/src/lib.rs#L648-L668)

If Slack rejects the reply, the user gets no message in Slack and the product does not surface an actionable error.

### 5. The web UI shows `Awaiting Pairing`, but not the actual next step

The web status becomes `Pairing` whenever the WASM channel is active but has neither a paired sender nor owner binding:

- [`src/channels/web/types.rs#L305-L323`](https://github.com/nearai/ironclaw/blob/staging/src/channels/web/types.rs#L305-L323)

This is technically correct, but not enough guidance for Slack. The UI does not tell the user:

- you must DM the bot directly
- a pending request may already exist
- you can approve it in the web UI / API / CLI
- no new Slack code will be sent on retries if the request already exists

## Expected Behavior

Slack pairing should be recoverable and self-explanatory:

1. After activation, the product should clearly explain the Slack-specific next step.
2. If the user DMs the bot and a pairing request is created, the user should be able to recover even if the Slack reply fails.
3. Retrying should not silently dead-end.
4. The web UI should surface concrete guidance and any pending request that already exists.
5. Missing Slack-side setup requirements should be detected or at least explained before the user is told to “message the bot”.

### Backend / channel behavior

- Add Slack-specific setup/pairing diagnostics so activation or status can report whether DM event delivery is likely misconfigured.
- Do not gate reply instructions on `created == true` alone; allow the existing code to be resent for repeat pairing attempts.
- Surface `send_pairing_reply(...)` failures instead of swallowing them.
- Consider returning a structured post-activation state like `awaiting_pairing` with actionable metadata.

### Web / UX behavior

- Replace vague `Awaiting Pairing` copy with explicit next steps for Slack.
- If a pending pairing request already exists, show it prominently on the extension card.
- Make the recovery path obvious: approve in web UI or via `ironclaw pairing approve slack <code>`.
- If no pending request exists, explain that the user must send a direct message to the bot, not just mention it in a channel.

## Acceptance Criteria

- DMing the Slack bot from an unpaired user reliably creates or reuses a pairing request and gives the user a recoverable next step.
- Repeated DMs can re-surface the pairing code or equivalent approval instructions.
- Failures when sending the Slack pairing reply are surfaced in a user-visible way.
- The extension UI tells the user the exact Slack pairing action required.
- Existing pending Slack pairing requests are easy to discover and approve from the web UI.
- The flow is test-covered for:
  - first-time pairing request creation
  - repeated DM retry / resend behavior
  - reply delivery failure handling
  - UI state when a pending request exists but no paired sender has been approved yet

### LINKED ISSUE #1994: LLM provider returns 502 Bad Gateway during active chat
Description: During normal chat interaction (e.g., after Telegram is connected and
the user sends a message), the provider returns HTTP 502 Bad Gateway. Emil observed
this 3 consecutive times:

Error shown:
  "Error: LLM error: Provider nearai_chat request failed: HTTP 502 Bad Gateway:
   502 Bad Gateway / nginx/1.27.4"

Impact: HIGH — The agent becomes completely unresponsive. Users have no way to
distinguish a transient error from a hard failure and no guidance on whether to retry.

Reporter: Emil

### LINKED ISSUE #2540: [Bug Bash 4/16] Routine execution fails with orchestrator timeout (~307s exceeds 300s limit)
## Bug Report — QA Bug Bash 4/16

**Environment:** Staging (hosted-staging / crab shack) — https://agent-stg.near.ai/

**Description:**
When a routine is executed, the orchestrator times out with an error indicating the execution time (~307 seconds) exceeded the allowed limit (300 seconds). No result is returned to the user.

**Steps to Reproduce:**
1. Create and trigger a routine on staging
2. Wait for execution to complete

**Actual Result:**
Orchestrator returns a timeout error: execution time limit exceeded (~307s > 300s). No routine output is delivered.

**Expected Result:**
Routines should either complete within the timeout window, or the timeout should be configurable/increased for longer-running tasks. At minimum, partial results should be returned if the timeout is hit, and the error should be user-friendly rather than a raw orchestrator trace.

**Severity:** High — Routines are completely non-functional when they exceed the timeout
**Frequency:** Reproducible during bug bash testing

**Related:** #2276 (Orchestrator crashes with HTTP 413 Payload Too Large)

ilblackdragon:
## Summary

Bundled fixes for six bug-bash tickets filed on 2026-04-16. Each stream is independent and ships with a regression test. Full plan: `/home/illia/.claude/plans/let-s-focus-on-following-effervescent-sparkle.md`.

## Issues addressed

- Fixes #2540 — Orchestrator VM timeout is now configurable (`IRONCLAW_ORCHESTRATOR_MAX_DURATION_SECS`, 30..=3600s, default 300). Timeout / memory / Python-traceback errors surface user-safe reasons instead of leaking Monty internals.
- Fixes #1994, #2546 — New `LlmError::BadGateway` variant. Upstream 502/503/504 from `nearai_chat` are now retried, counted as transient by the circuit breaker, and never carry the response body (which routinely contained Python tracebacks — the root cause of #2546).
- Fixes #1537 — `AppBuilder::init_secrets` always installs a secrets store (persistent or ephemeral fallback), so WASM tool / channel credential injection is wired unconditionally on hosted TEE deployments. Mirrors the existing `ExtensionManager` fallback.
- Fixes #1839 — Slack `chat.postMessage` returns HTTP 200 even on scope/token failures with `{\"ok\":false}`. The response parser was extracted into a testable helper and now propagates the failure; `send_pairing_reply` errors are logged with scope guidance instead of being swallowed.
- Partial #1993 — Chat rehydration (`reconcile_in_progress_with_turns`) now requires successful tool calls before dropping in-progress state, so a 502 mid-turn no longer shows fabricated success on reload. The deeper forward-path fix (engine-v2 side-effect gate for #2544, #2541) is a follow-up.

## Out of scope

- #2544, #2541 — engine-v2 hard side-effect gate is aspirational per `.claude/rules/tool-evidence.md`; design deserves a focused follow-up PR.
- #2437 — verified already fixed (`src/llm/nearai_chat.rs:1060-1070` calls `normalize_schema_strict`); commented and closed on the issue.
- #2543 — likely resolved by PR #2515; retest on staging pending.

## Review-discipline note

This PR spans multiple scopes (per `.claude/rules/review-discipline.md`'s "PR Scope Discipline"). Keeping them bundled because each change is small (<150 lines including tests), independent, and shares the same release window. Happy to split per-stream if reviewers prefer — the streams map cleanly onto the file boundaries listed below.

## Touched surfaces

- `channels-src/slack/src/lib.rs` — #1839
- `crates/ironclaw_engine/src/executor/orchestrator.rs` — #2540, #2546
- `src/app.rs` — #1537
- `src/channels/web/features/chat/mod.rs` — #1993
- `src/llm/error.rs`, `nearai_chat.rs`, `retry.rs`, `circuit_breaker.rs` — #1994, #2546

## Test plan

- [x] `cargo fmt` clean
- [x] `cargo clippy --lib --all-features -- -D warnings` clean
- [x] `cargo test -p ironclaw_engine --lib` — 446 pass (includes 4 new `failure_reason_*` tests)
- [x] `cargo test --lib llm::retry llm::error llm::circuit_breaker` — 46 pass (includes BadGateway retryable arm)
- [x] `cargo test --lib app::tests::` — 3 pass (includes new ephemeral secrets store test)
- [x] `cargo test --lib reconcile` — 6 pass (includes new #1993 regression)
- [x] `cargo test -p slack-channel --lib slack_post_message_result` — 3 pass
- [ ] Manual: repeat each bug's repro on staging and confirm user-visible behavior now matches expected outcome.

Pre-existing test failures on `staging` (`tools::builtin::skill_tools::tests::test_*zip_extract*`, 8 cases; `handle_with_engine_persists_attachment_files_and_indexes_them` under parallel test load, 1 case) verified to be unrelated — same failures reproduce on clean `staging` without these changes.


ilblackdragon:
Added a follow-up commit addressing the \"why is there no secrets store\" question on hosted TEEs (d41086d):

1. **`install_ephemeral_secrets_store` is now fail-loud** — accepts a `reason` tag and emits at `warn!` with one of: master-key resolution produced no key, crypto init failed, no DB handles (no_db / init_database skipped), feature-flag mismatch, or the unexpected "handles present but factory returned None" case. Previously silent at `debug!`.

2. **`ironclaw doctor`'s Secrets check actually probes the store now.** The old check only inspected `settings.secrets_master_key_source`, which missed the exact hosted-TEE failure mode the reporter hit (master key resolves to `Env`/`Keychain` but the DB handle isn't wired, so `create_secrets_store` returns None). The new check runs the same `SecretsConfig::resolve` path `init_secrets` uses and calls `create_secrets_store` to verify reachability.

3. **Stale `src/db/CLAUDE.md` note fixed** — it claimed `LibSqlSecretsStore` wasn't plumbed through startup. It is (has been for some time); the note now describes the actual dispatch + the ephemeral fallback contract from this PR.

The existing `check_secrets_*` unit tests were rewritten as "does-not-panic" shape (like `check_docker_daemon_does_not_panic`) because the new function reads real env and the outcome is test-host dependent.

src/llm/nearai_chat.rs
```diff
@@ -324,6 +324,25 @@ impl NearAiChatProvider {
...
+                    provider = "nearai_chat",
+                    status = status_code,
+                    body_preview = crate::agent::truncate_for_preview(&response_text, 512).as_str(),
+                    "NEAR AI Chat upstream 5xx response"
+                );
+                return Err(LlmError::BadGateway {
+                    provider: "nearai_chat".to_string(),
+                    status: status_code,
+                    retry_after: retry_after_header,
+                });
```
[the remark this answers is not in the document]
ilblackdragon: Good catch — fixed in 56cad890. `parse_retry_after`'s 60s default for missing headers was silently overriding exponential backoff for every 5xx. Changed the extraction to only wrap in `Some(..)` when the header was actually present, so `BadGateway { retry_after: None }` flows through to `retry_backoff_delay` (1s → 2s → 4s). Kept the 60s fallback explicit at the 429 call site since it's the documented rate-limit contract.

Added a regression test (`llm::retry::tests::bad_gateway_without_retry_after_does_not_match_some_arm`) that mirrors the match arm inside `retry_loop` and fails if the `Some(_)` arm ever catches a `None` retry_after.

src/llm/nearai_chat.rs
```diff
@@ -324,6 +324,25 @@ impl NearAiChatProvider {
                 }
             }
 
+            // HTTP 5xx from the upstream LLM gateway — map to BadGateway so the
+            // retry layer backs off, the circuit breaker counts a transient failure,
+            // and the channel boundary produces a user-safe message (never the raw
+            // upstream body, which often contains a Python traceback). The raw
+            // body is logged at debug for operators.
+            if matches!(status_code, 502..=504) {
```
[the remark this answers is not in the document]
ilblackdragon: Valid — fixed in 56cad890. Expanded the match from `502..=504` to `500..=599` so upstream application 500s (which is exactly the shape #2546 reported — Python traceback in the body) are also mapped to `BadGateway` and never carry the response body. The `status` field still preserves the specific code for operators in debug logs. This matches the intent in `.claude/rules/error-handling.md` ("raw HTTP 5xx → temporarily unavailable").

I don't think a separate variant for 500-vs-proxy-5xx is worth it here — from the user's and retry-layer's perspective, every 5xx is "provider temporarily unavailable, try again with backoff". The debug log carries the distinguishing code for anyone triaging.

src/app.rs
```diff
@@ -86,6 +86,27 @@ pub struct AppBuilderFlags {
...
+    use crate::secrets::{InMemorySecretsStore, SecretsCrypto};
+    let ephemeral_key =
+        secrecy::SecretString::from(crate::secrets::keychain::generate_master_key_hex());
+    match SecretsCrypto::new(ephemeral_key) {
+        Ok(crypto) => Some(Arc::new(InMemorySecretsStore::new(Arc::new(crypto)))),
+        Err(e) => {
+            tracing::warn!("Failed to initialize ephemeral secrets crypto: {e}");
+            None
+        }
+    }
```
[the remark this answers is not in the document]
ilblackdragon: Good call — fixed in 56cad890. `build_ephemeral_secrets_store` now returns `Result<Arc<...>, SecretError>` and `install_ephemeral_secrets_store` propagates via `?`, so a genuinely broken ephemeral crypto setup aborts startup at the real cause instead of deferring to the less-actionable "secrets store not initialized" error later in `init_extensions`. The existing `app::tests::ephemeral_secrets_store_is_constructible_and_usable` regression test now validates the Result contract.

src/cli/doctor.rs
```diff
@@ -602,23 +602,92 @@ async fn check_skills() -> CheckResult {
...
+/// `PostgresSecretsStore` aren't available, so WASM tool credential
+/// injection silently falls back to unauthenticated requests.
+///
+/// This check runs the same `SecretsConfig::resolve` path `AppBuilder::init_secrets`
+/// uses, then attempts to construct the backing store — the two things that
+/// jointly determine whether credential injection will actually work.
+async fn check_secrets(settings: &Settings) -> CheckResult {
+    // 1. Master-key resolution — mirrors the runtime probe order:
+    //    SECRETS_MASTER_KEY env, OS keychain, auto-generate to ~/.ironclaw/.env.
+    let resolved = match crate::config::SecretsConfig::resolve().await {
```
serrrfirat: One thing to watch here: `SecretsConfig::resolve()` is not read-only. When no key exists it auto-generates one and persists it to `~/.ironclaw/.env` via `auto_generate_and_persist()`. That means `ironclaw doctor` now mutates user state just by running diagnostics. If the intent is for doctor to stay observational, this probably wants a non-persisting probe path (or a lower-level existence check) so the missing-key case can be reported without creating one.

src/channels/web/features/chat/mod.rs
```diff
@@ -1186,6 +1186,20 @@ fn completed_turn_is_newer_than_in_progress(
...
+/// A turn has "unfinished" tool calls if any recorded tool call errored
+/// (`has_error`) or never produced a result (`!has_result`). When a 502 or
+/// transport drop interrupts a turn mid-execution but a response text was
+/// already committed (e.g. "Done! I've sent..."), the naive
+/// "response.is_some() → completed" check mis-classifies that turn and the
+/// user sees fabricated success on reload. See #1993.
+fn turn_tool_calls_succeeded(turn: &TurnInfo) -> bool {
+    turn.tool_calls
+        .iter()
+        .all(|tc| tc.has_result && !tc.has_error)
```
serrrfirat: I think this is still a little too strict for recovered turns. `tool_calls` is the whole turn history, so a turn that had one failed tool call and then recovered with a later tool call + final assistant response will still fail `all(...)`. If `live_state` is stale because clearing metadata failed after the response was persisted, reload will keep showing that turn as `Processing` forever even though the action actually completed. Would it make sense to key this off the unfinished/current tool step instead of any historical tool error in the turn?

crates/ironclaw_engine/src/executor/orchestrator.rs
```diff
@@ -379,7 +441,7 @@ pub async fn execute_orchestrator(
         Ok(Ok(p)) => p,
         Ok(Err(e)) => {
             return Err(EngineError::Effect {
-                reason: format!("Orchestrator runtime error: {e}"),
+                reason: orchestrator_failure_reason("Orchestrator runtime error", &e.to_string()),
```
serrrfirat: Nice hardening. One gap I still see: this sanitizer only wraps the runtime/resume paths. The parse/start branches a few lines above still return raw `format!("Orchestrator parse error: {e}")`. If the goal is to keep Monty internals/tracebacks behind the debug boundary, it would be worth routing those startup failures through the same helper too so a bad `default.py` deploy doesn't regress the channel edge.

serrrfirat (commented):
Rechecked the current head and left 3 follow-up comments.

crates/ironclaw_engine/src/executor/orchestrator.rs
```diff
@@ -96,14 +97,75 @@ fn normalize_pause_outcome(
...
+/// internal file paths and upstream HTTP bodies) is always logged at debug
+/// so operators can correlate, but never returned in the user-visible reason
+/// — see `.claude/rules/error-handling.md`, "Error Boundaries at the Channel
+/// Edge" (#2546).
+fn orchestrator_failure_reason(prefix: &str, err_msg: &str) -> String {
+    debug!(prefix, err_msg, "orchestrator VM failure");
+
+    let lower = err_msg.to_ascii_lowercase();
+    let hit_time_limit =
+        lower.contains("timed out") || lower.contains("timeout") || lower.contains("duration");
```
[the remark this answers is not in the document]
ilblackdragon: Good catch — fixed in 8588f6cb.

The fix has two parts:

1. **Tightened the timeout predicate set** to `"timed out"` / `"timeout"` / `"duration limit"` / `"max_duration"` / `"maximum duration"` — the bare `"duration"` catch is gone. A `TypeError: duration must be a positive integer` (or any other runtime error that happens to contain the word) no longer classifies as a time-budget exhaustion.

2. **Replaced the whole `orchestrator_failure_reason(...) -> String` helper with a typed `OrchestratorFailure { kind, debug_detail }`** so classification is a compile-checked enum (`TimeLimit` / `ResourceLimit` / `Traceback` / `VmPanic` / `Other`) rather than a format string. The raw Monty message is preserved on `debug_detail` (never user-visible) and surfaces through gateway debug mode (`IRONCLAW_DEBUG_ERRORS=1`) so operators can triage without tailing logs — per the direction on this PR to move error handling away from string formatting and preserve low-level detail for debug mode.

Regression test: `failure_reason_does_not_treat_bare_duration_as_timeout` locks in that a `TypeError` mentioning `"duration"` classifies as `Other`, not `TimeLimit`.

src/cli/doctor.rs
```diff
@@ -602,23 +602,92 @@ async fn check_skills() -> CheckResult {
...
-                    "master key source: env var but SECRETS_MASTER_KEY not set".into(),
-                )
-            }
+    };
+
+    if resolved.master_key().is_none() {
+        // Honor the historical skip path so first-time-run and test harnesses
+        // still read as "not configured" rather than fail.
+        return CheckResult::Skip("secrets not configured (run `ironclaw onboard`)".into());
+    }
```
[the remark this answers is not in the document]
ilblackdragon: Valid — fixed in 8588f6cb. You're right, `SecretsConfig::resolve()` auto-generates and persists a key via `auto_generate_and_persist()`, so running `ironclaw doctor` on a fresh machine silently created `~/.ironclaw/.env`.

`check_secrets` now calls `crate::secrets::resolve_master_key()` instead — that's a read-only probe (env var → OS keychain, no filesystem writes, no generation). If no key exists, the check reports `Skip("secrets not configured (run `ironclaw onboard`)")` as it did historically; the backing-store probe only runs when a key was actually found.

Source-label derivation now mirrors the same env-first / keychain-second order without the third fallback. This also makes the "historical skip path" honest — it's actually reachable now instead of being dead code behind an always-succeeding auto-generate.

src/cli/doctor.rs
ilblackdragon: Fixed in 8588f6cb — same root cause as Copilot's version of this comment. `check_secrets` now uses the read-only `crate::secrets::resolve_master_key()` path (env + keychain only, no auto-generate-and-persist), so `ironclaw doctor` is back to being observational. Missing key → `Skip`, present key → probe the backing store.

src/channels/web/features/chat/mod.rs
ilblackdragon: Good catch — fixed in 8588f6cb. `turn_tool_calls_succeeded` now keys off the *trailing* tool call rather than every tool call in the turn history:

```rust
match turn.tool_calls.last() {
    Some(last) => last.has_result && !last.has_error,
    None => true,
}
```

An errored call followed by a successful retry + final response now clears `Processing` correctly. The original #1993 intent (stuck tool call → keep showing Processing) is preserved because an unfinished or errored *trailing* call still returns `false`.

Added `test_reconcile_allows_recovery_from_earlier_tool_error` — two-call turn where the first errors (HTTP 502) and the second succeeds; reconcile now returns `None` (treats the turn as complete) instead of pinning it to `Processing`.

crates/ironclaw_engine/src/executor/orchestrator.rs
ilblackdragon: Fixed in 8588f6cb. Every orchestrator failure path — parse, start, resume, NameLookup, and the `catch_unwind` panic arms on each — now routes through a single typed classifier, not ad-hoc `format!()`.

Concretely:
- Introduced `OrchestratorFailure { kind: OrchestratorFailureKind, debug_detail: String }` in `crates/ironclaw_engine/src/types/error.rs`. `OrchestratorFailureKind` is a `thiserror` enum with `TimeLimit` / `ResourceLimit` / `Traceback` / `VmPanic` / `Other` variants; `Display` renders only the user-safe message.
- New `EngineError::Orchestrator(OrchestratorFailure)` variant. `EngineError::debug_detail()` accessor so callers don't have to pattern-match.
- `crates/ironclaw_engine/src/executor/orchestrator.rs`: the old parse / start / resume branches that returned `EngineError::Effect { reason: format!("Orchestrator parse error: {e}") }` (and the two "Monty VM panicked during ..." paths) all go through `classify_orchestrator_failure(prefix, &err_msg)` or `orchestrator_vm_panic(prefix, phase)`. A bad `default.py` deploy now hits the same sanitizer as runtime tracebacks.

Per the direction on this PR to preserve low-level detail for debug mode rather than drop it: `OrchestratorFailure::debug_detail` carries the raw Monty message all the way up to `ThreadOutcome::Failed { error, debug_detail }`. `src/bridge/router.rs` has a new `gateway_debug_errors_enabled()` helper gated on `IRONCLAW_DEBUG_ERRORS=1` that appends the detail to the reply; otherwise the detail goes to `tracing::debug!` and is stripped from user output.

Regression test: `failure_reason_strips_python_traceback` now also asserts `failure.debug_detail()` retains the raw `Traceback` (so the classifier preserves the detail, the Display impl strips it — exactly the boundary the rule in `.claude/rules/error-handling.md` asks for).

src/bridge/router.rs
```diff
@@ -71,14 +84,30 @@ fn engine_err(context: &str, e: impl std::fmt::Display) -> Error {
...
-    BridgeOutcome::Respond(crate::bridge::user_facing_errors::user_facing_thread_failure(error))
+    let user_text = crate::bridge::user_facing_errors::user_facing_thread_failure(error);
+    // Low-level detail preserved from typed errors (e.g. `OrchestratorFailure`)
+    // is never user-facing by default — see `.claude/rules/error-handling.md`.
+    // Operators flip `IRONCLAW_DEBUG_ERRORS=1` to append it to the reply so
+    // they can triage without tailing logs.
+    let reply = if let Some(detail) = debug_detail.filter(|_| gateway_debug_errors_enabled()) {
+        format!("{user_text}\n\n[debug] {detail}")
+    } else {
+        user_text
```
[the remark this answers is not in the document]
ilblackdragon: Already resolved by 042c2ee7 — `IRONCLAW_DEBUG_ERRORS` was dropped entirely per follow-up review discussion. The raw `debug_detail` no longer flows into the user reply at all; the sanitized message from `user_facing_thread_failure` is the only thing rendered to chat. `debug_detail` is surfaced through two bounded channels:

1. `AppEvent::Error.debug_detail` on the SSE stream — the Debug Inspector renders it in the Activity tab underneath the sanitized message.
2. `tracing::debug!` in `bridge_outcome_for_failed_thread` — operators who want it in text logs set `RUST_LOG=ironclaw::bridge::router=debug`.

The unbounded-append-to-user-reply concern you raised is gone in 042c2ee7.

crates/ironclaw_engine/src/executor/orchestrator.rs
```diff
@@ -96,14 +97,105 @@ fn normalize_pause_outcome(
...
+            prefix: prefix.to_string(),
+        }
+    } else if has_python_traceback {
+        OrchestratorFailureKind::Traceback {
+            prefix: prefix.to_string(),
+        }
+    } else {
+        OrchestratorFailureKind::Other {
+            prefix: prefix.to_string(),
+            message: err_msg.to_string(),
```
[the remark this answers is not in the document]
ilblackdragon: Good catch — fixed in 82d06410.

`OrchestratorFailureKind::Other`'s `#[error(...)]` now renders `"{prefix}: internal orchestrator failure"` with no `{message}` interpolation. Dropped the `message` field entirely since it was redundant with `OrchestratorFailure::debug_detail`. This closes the leak at channel-edge surfaces that bypass `bridge::user_facing_errors::user_facing_thread_failure` — specifically `runtime/mission.rs::process_mission_outcome_and_notify` which renders `format!("Mission failed: {error}")` directly from `ThreadOutcome::Failed { error }`. Any unclassified Monty failure would previously leak its raw text there.

Regression test renamed to `failure_reason_hides_unknown_raw_message_from_user_text` — asserts Display does NOT contain `"NameError"` while `failure.debug_detail()` still does, which is exactly the boundary you called out.

Raw detail for operator triage still flows through:
- `OrchestratorFailure::debug_detail` (unchanged — carries the full `err_msg`)
- SSE `AppEvent::Error { debug_detail }` → Debug Inspector
- `RUST_LOG=...=debug` → `tracing::debug!` in `bridge_outcome_for_failed_thread`

src/cli/doctor.rs
```diff
@@ -602,23 +602,94 @@ async fn check_skills() -> CheckResult {
...
+    {
+        crate::settings::KeySource::Env
+    } else {
+        crate::settings::KeySource::Keychain
+    };
+
+    let source_label = match source {
+        crate::settings::KeySource::Keychain => "OS keychain",
+        crate::settings::KeySource::Env => "env / ~/.ironclaw/.env",
+        crate::settings::KeySource::None => "unset",
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in 82d06410. You're right — `source` is only reachable as `Env` or `Keychain` at this point (the `Some(master_key_hex) else Skip` guard above already returns if no key exists, so `KeySource::None` can never land here). Folded the match into the existing env-wins branch:

```rust
let (source, source_label) = if env_wins {
    (crate::settings::KeySource::Env, "env / ~/.ironclaw/.env")
} else {
    (crate::settings::KeySource::Keychain, "OS keychain")
};
```

`source` is still kept as a `KeySource` because the `settings_note` comparison below needs to diff against `settings.secrets_master_key_source` (which *can* be `None`).

src/bridge/router.rs
```diff
@@ -71,14 +84,30 @@ fn engine_err(context: &str, e: impl std::fmt::Display) -> Error {
...
+    debug_detail: Option<&str>,
+    user_id: &str,
+    channel: &str,
+) -> BridgeOutcome {
     tracing::warn!(
         user_id = %user_id,
         channel = %channel,
         error = %error,
+        debug_detail = ?debug_detail,
         "engine v2: thread failed; showing user-friendly summary",
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in 82d06410. The `warn!` call now logs only `debug_detail_bytes` (length), not the contents:

```rust
tracing::warn!(
    user_id = %user_id,
    channel = %channel,
    error = %error,
    debug_detail_bytes = debug_detail.map(|d| d.len()),
    "engine v2: thread failed; showing user-friendly summary",
);
if let Some(detail) = debug_detail {
    tracing::debug!(user_id = %user_id, channel = %channel, detail, "...");
}
```

So a multi-KB traceback no longer floods `warn` logs. Operators who want the full text flip `RUST_LOG=ironclaw::bridge::router=debug`, or look at the Debug Inspector — `AppEvent::Error.debug_detail` on the SSE stream carries the unfiltered content into the inspector's Activity tab regardless of log level. That's the two-channel split the PR discussion landed on: sanitized in user-visible surfaces, length-only at `warn!`, full detail at `debug!` + SSE.

(Note: `IRONCLAW_DEBUG_ERRORS` was dropped in 042c2ee7 — there's no flag to gate on anymore; the debug detail is now unconditionally available via SSE/debug-log rather than append-to-reply.)

src/bridge/router.rs
```diff
@@ -3963,11 +3984,35 @@ async fn await_thread_outcome(
...
+                sse.broadcast_for_user(
+                    &message.user_id,
+                    AppEvent::Error {
+                        message: sanitized,
+                        debug_detail: debug_detail.clone(),
+                        thread_id: Some(thread_id.to_string()),
+                    },
+                );
+            }
+            Ok(bridge_outcome_for_failed_thread(
```
serrrfirat, line 4045: **High Severity**

This failure path now does two user-facing things on the web flow:
1. broadcasts `AppEvent::Error` over SSE, and then
2. returns `BridgeOutcome::Respond(...)`.

The gateway's normal `respond()` path turns that second outcome into a regular `response` SSE event, so the browser ends up rendering the same failed turn twice (once via the `error` handler and once via the normal response path).

**Confidence:** Certain

**Suggested fix:** choose a single user-facing surface here for web requests. Either emit the `error` SSE event and return a non-response outcome, or keep the `Respond` path and reserve the SSE `error` event for debug/telemetry-only consumers.

src/bridge/router.rs
```diff
@@ -3963,11 +3984,35 @@ async fn await_thread_outcome(
...
+            // so operators can triage from the inspector panel instead
+            // of tailing logs.
+            if let Some(ref sse) = state.sse {
+                let sanitized =
+                    crate::bridge::user_facing_errors::user_facing_thread_failure(&error);
+                sse.broadcast_for_user(
+                    &message.user_id,
+                    AppEvent::Error {
+                        message: sanitized,
+                        debug_detail: debug_detail.clone(),
```
serrrfirat: **Medium Severity**

`debug_detail` is now being attached to the standard scoped SSE `error` event. That means raw Monty traces / upstream HTTP bodies still cross the normal browser boundary and are visible to any authenticated SSE consumer (and to devtools / custom clients), even if the main chat UI only renders the sanitized `message`.

The PR intent/comments say this detail should be debug-only, but this path sends it on the default event contract.

**Confidence:** Certain

**Suggested fix:** gate `debug_detail` behind an explicit debug mode / inspector endpoint, or keep it server-side in logs instead of putting it on the regular `AppEvent::Error` payload.

src/cli/doctor.rs
```diff
@@ -602,23 +602,90 @@ async fn check_skills() -> CheckResult {
...
+        match crate::secrets::SecretsCrypto::new(secrecy::SecretString::from(master_key_hex)) {
+            Ok(c) => std::sync::Arc::new(c),
+            Err(e) => {
+                return CheckResult::Fail(format!(
+                    "master key resolved from {source_label} but crypto init failed: {e}"
+                ));
+            }
+        };
+
+    match crate::db::create_secrets_store(&config.database, crypto).await {
```
serrrfirat: **Medium Severity**

`ironclaw doctor` is now using `db::create_secrets_store(...)` as a probe, but that helper is not read-only: it opens a fresh backend connection and runs migrations before returning the store.

That creates two problems:
- running a diagnostic can mutate the configured DB just by checking secrets health, and
- it no longer exercises the same runtime path that failed in `AppBuilder::init_secrets` (`secrets::create_secrets_store(handles)` returning `None` because no DB handles were wired).

So this can still miss the hosted-TEE / missing-handle failure mode while also making `doctor` side-effectful.

**Confidence:** Certain

**Suggested fix:** split out a passive probe that checks whether the runtime secrets path is usable without creating a brand-new store / running migrations.

crates/ironclaw_engine/src/executor/orchestrator.rs
```diff
@@ -96,14 +97,104 @@ fn normalize_pause_outcome(
...
+/// `.claude/rules/error-handling.md`, "Error Boundaries at the Channel
+/// Edge" (#2546).
+fn classify_orchestrator_failure(prefix: &str, err_msg: &str) -> OrchestratorFailure {
+    debug!(prefix, err_msg, "orchestrator VM failure");
+
+    let lower = err_msg.to_ascii_lowercase();
+    // Narrow substring set — "duration" alone caught unrelated errors
+    // and drifted away from `executor/scripting.rs::classify_runtime_error`.
+    let hit_time_limit = lower.contains("timed out")
+        || lower.contains("timeout")
```
serrrfirat: **Medium Severity**

This timeout classifier is still broad enough to mislabel non-VM failures as orchestrator wall-clock exhaustion. Any upstream/provider error text containing `timeout` / `timed out` (for example an LLM-layer `Request timed out`) will now map to `TimeLimit` and tell operators to raise `IRONCLAW_ORCHESTRATOR_MAX_DURATION_SECS`, even when the real problem is a provider/network timeout.

That points users at the wrong remediation and hides the real failure class.

**Confidence:** Likely

**Suggested fix:** reserve `TimeLimit` for unmistakable Monty duration-limit markers, and let generic upstream timeouts fall through to a non-budget classification (or a dedicated provider-timeout variant).

src/bridge/router.rs
ilblackdragon: Good catch — fixed in 5fa2c0ea.

`bridge_outcome_for_failed_thread` now takes a `sse_will_deliver_to_user` flag and returns `BridgeOutcome::NoResponse` when the originating channel is the web gateway (`message.channel == "gateway"` and `state.sse.is_some()`). That way the SSE `error` frame is the single user-visible surface on the web flow — no more concurrent `response` frame from `GatewayChannel::respond()`.

Non-gateway channels (telegram, relay, cli, etc.) still receive `BridgeOutcome::Respond(sanitized)` because for them the SSE stream is only a secondary per-user surface; their primary delivery still needs to go through `channel.respond()`.

Regression test: `failed_thread_outcome_is_no_response_when_sse_will_deliver` asserts the helper returns `NoResponse` when `sse_will_deliver_to_user = true`, and existing `Respond` tests now pass `false` explicitly.

src/bridge/router.rs
ilblackdragon: Good catch — fixed in 5fa2c0ea.

`debug_detail` is dropped from the `AppEvent::Error` wire payload entirely, so it no longer crosses the default SSE boundary to any authenticated consumer (chat UI, devtools, or custom clients). The raw detail now stays server-side only:

- `OrchestratorFailure::debug_detail` — preserved on the typed engine error for operator triage.
- `tracing::debug!` inside `bridge_outcome_for_failed_thread` — available on demand via `RUST_LOG=ironclaw::bridge::router=debug` or `/api/logs/events`.

`debug-panel.js` was updated to render only the sanitized `message` from the `error` SSE event (since `data.debug_detail` is gone). Accordingly the rustdoc on `AppEvent::Error` now states the wire-vs-server-side split explicitly.

src/cli/doctor.rs
ilblackdragon: Valid — fixed in 5fa2c0ea. You're right that `db::create_secrets_store` opens a fresh backend and runs migrations, which both mutates and tests the wrong code path.

`check_secrets` now uses `crate::db::connect_without_migrations(&config.database)` to obtain `DatabaseHandles`, then calls `crate::secrets::create_secrets_store(crypto, &handles)` — the exact dispatch `AppBuilder::init_secrets` runs at startup. That way:

- No migrations fire (`connect_without_migrations` opens a connection but explicitly does not run `run_migrations`).
- The `#1537` hosted-TEE failure mode (`handles.libsql_db.is_none() && handles.pg_pool.is_none()` → store factory returns `None`) is now actually reachable — when it hits, doctor reports `Fail` with the exact "falls back to ephemeral in-memory store" guidance from that bug.
- Connect failures get a distinct message that calls out "database unreachable" so operators know to fix DB connectivity first.

The probe is now minimally side-effectful (opens a read connection; no schema changes) and exercises the same dispatch tree that actually broke in production.

crates/ironclaw_engine/src/executor/orchestrator.rs
ilblackdragon: Valid — fixed in 5fa2c0ea. You're right that the classifier was steering operators at `IRONCLAW_ORCHESTRATOR_MAX_DURATION_SECS` for failures that have nothing to do with the Monty wall-clock budget.

The predicate set is now narrowed to unmistakable wall-clock markers:

- `"duration limit"`
- `"max_duration"`
- `"maximum duration"`
- `"execution duration exceeded"`
- `"orchestrator timed out"`

The bare `"timeout"` / `"timed out"` catches are gone, so upstream LLM / network timeouts (`"Request timed out"`, `"Connection timed out"`, `"LLM call failed: timeout waiting for response"`, `"upstream provider timeout after 30s"`) now classify as `Other` (generic internal orchestrator failure) and the user message no longer advises raising the budget knob.

Regression test: `failure_reason_does_not_treat_upstream_timeout_as_time_limit` iterates over four upstream-timeout shapes and asserts each classifies as `Other`, not `TimeLimit`, and that the rendered user message does not contain `"IRONCLAW_ORCHESTRATOR_MAX_DURATION_SECS"`. The original duration-limit test was updated to use an explicit `"ResourceLimits: duration limit exceeded"` marker so it still locks in the TimeLimit path.

src/channels/web/features/chat/mod.rs
```diff
@@ -1170,7 +1191,12 @@ fn reconcile_in_progress_with_turns(
     };
 
     if in_progress_matches_turn(last_turn, &in_progress) {
-        if last_turn.response.is_some() {
+        // Only treat the matching turn as "already done" if the model wrote
+        // a final response AND every recorded tool call completed
+        // successfully. Otherwise the turn is really still in-progress and
+        // the UI should keep the processing affordance visible so the user
+        // can see (and, in a follow-up, retry) the stuck step (#1993).
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in c9b0b0a7. The comment now describes the trailing-tool-call semantics (earlier failed attempts are allowed if a later retry succeeded) instead of the stale "every recorded tool call completed successfully" wording, so it matches what `turn_tool_calls_succeeded` actually does.

src/cli/doctor.rs
```diff
@@ -602,23 +602,116 @@ async fn check_skills() -> CheckResult {
...
+/// The stored `settings.secrets_master_key_source` is only one signal and
+/// does not capture the hosted-TEE failure mode (#1537): master key resolves
+/// to `Env`/`Keychain` source at runtime, but the backing store factory
+/// returns `None` because the DB handles needed by `LibSqlSecretsStore` /
+/// `PostgresSecretsStore` aren't available, so WASM tool credential
+/// injection silently falls back to unauthenticated requests.
+///
+/// This check runs the same `SecretsConfig::resolve` path `AppBuilder::init_secrets`
+/// uses, then attempts to construct the backing store — the two things that
+/// jointly determine whether credential injection will actually work.
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in c9b0b0a7. The rustdoc now describes the actual read-only behavior (probe via `crate::secrets::resolve_master_key()` — env → keychain, no filesystem writes, no auto-generate) and explicitly calls out why it avoids `SecretsConfig::resolve` (auto-persist side effect that would make `ironclaw doctor` mutate user state). This should prevent a future refactor from reintroducing the mutation problem.

ilblackdragon:
**Follow-up round 3 pushed in c9b0b0a7** — addresses the two open Copilot comments on 8489a978 plus four review-derived nits surfaced during local triage.

### Copilot comments

- `src/channels/web/features/chat/mod.rs:1198` — stale comment updated to describe trailing-tool-call semantics instead of "every recorded tool call completed successfully."
- `src/cli/doctor.rs:616` — stale rustdoc updated to describe the read-only `resolve_master_key()` probe and explicitly call out why `SecretsConfig::resolve` is avoided.

### Review-derived nits

- **Gateway channel name pinning** — exported `GATEWAY_CHANNEL_NAME` from `src/channels/web/mod.rs` and referenced it from both the `Channel::name()` impl and `src/bridge/router.rs`, so the two places that used to carry the literal `"gateway"` now compile-time-share the same constant. Drift is no longer possible.
- **`parse_retry_after` ergonomics** — split into two helpers. `parse_retry_after(Option<&HeaderValue>) -> Duration` keeps the 60 s-on-missing default for rate-limit / auth callers (anthropic_oauth, github_copilot, workspace/embeddings). New `parse_retry_after_value(&HeaderValue) -> Duration` takes a known-present header so 5xx call sites can distinguish "absent" (→ exponential backoff) from "unparseable" (→ 60 s fallback). Two new regression tests pin the new helper.
- **Doctor secrets tests** — added `check_secrets_reports_env_source_when_env_key_is_set`. Under `ENV_MUTEX`, sets `SECRETS_MASTER_KEY`, calls `check_secrets(&Settings::default())` with `secrets_master_key_source: None`, and asserts the rendered message surfaces both the `env` source label and the `settings say \`None\`` drift warning. This pins the exact #1537 hosted-TEE axis (runtime resolves to Env but the settings snapshot says None) that the prior "doesn't panic" test couldn't detect.

### Quality gate

- `cargo fmt` clean
- `cargo clippy --lib --all-features -- -D warnings` clean
- `cargo test --lib` — 5382 pass, 0 fail, 7 ignored
- `cargo test -p ironclaw_engine --lib` — 450 pass
- `cargo test -p slack-channel --lib slack_post_message_result` — 3 pass (in `channels-src/slack/`)

PR #2754, opened 2026-04-20
feat(web): add self-service user secrets and durable binding approvals

serrrfirat:
## Summary
- add a dedicated self-service `/api/secrets` surface plus Settings → Secrets UI for per-user secret CRUD/import
- persist per-user secret binding approvals, expose approval revoke actions in Settings, and revoke approvals when a secret is deleted
- gate unapproved user-authored secret bindings through the existing approval flow and persist approval on resume

## Details
- keeps secret values out of chat/LLM paths and out of list responses
- uses explicit secret names and exact-host approval candidates for user-authored bindings
- reuses existing gate/approval machinery instead of introducing a parallel flow
- updates missing-secret remediation text to deep-link users toward Settings → Secrets

## Validation
- `cargo check --lib`
- `cargo test binding_approval --lib --quiet`
- `cargo test user_secret_ --lib --quiet`
- `cargo test get_setup_instructions_or_default_returns_fallback --lib --quiet`
- `node --check crates/ironclaw_gateway/static/js/surfaces/settings.js`
- `node --check crates/ironclaw_gateway/static/js/core/ui-helpers.js`
- `node --check crates/ironclaw_gateway/static/i18n/en.js`
- `node --check crates/ironclaw_gateway/static/i18n/ko.js`
- `node --check crates/ironclaw_gateway/static/i18n/zh-CN.js`
- `node --check crates/ironclaw_gateway/static/i18n-app.js`

## Follow-ups
- browser-smoke the live UI flow end-to-end, especially the earlier 403/Forbidden report
- decide whether this needs any `FEATURE_PARITY.md` or user-facing docs updates before merge


henrypark133 (changes requested):
Findings:
1. High - `src/bridge/auth_manager.rs:289` and `src/tools/builtin/http.rs:452`: skill-scoped credential mappings now match when `active_skill_names` is empty. That lets a secret approved for one skill still be picked up by a plain `http` tool call with no skill context, which weakens the new approval boundary from "approved for this skill" to "approved for any matching HTTP call to that host."
2. Medium - `tests/e2e/scenarios/test_v2_secret_binding_approval.py:453`: the new E2E claims to validate self-service member behavior, but it runs the flow with `AUTH_TOKEN` as the acting user. The created member user is never used, so a regression where `/api/secrets` only works for the owner or admin would still pass.

Summary:
- Residual risk: I did not run the local test suite. GitHub currently shows failing `Clippy` and `Code Style` checks on this PR as well.

ilblackdragon:
## Code Review

### Overview

Adds three related capabilities:
1. **Self-service secrets API** (`/api/secrets/*`) for authenticated non-admin users to CRUD and import their own per-user secrets, plus a Settings → Secrets UI.
2. **Durable binding approvals** — skills and WASM tools now carry provenance (`artifact_kind`, `artifact_name`, `artifact_fingerprint`, `binding_policy`) on their credential mappings. User-authored / non-vetted artifacts (User/Workspace/Installed skills, User-trust-level WASM tools) must get first-use approval before their declared mapping can inject a secret for a given host. Approvals persist in settings.
3. **Skill-scoped credential injection** — threads carry `active_skill_names`; mappings attributed to a skill only inject when that skill is active in the current turn.

+3,526 / −144 across 34 files, new e2e scenario, solid unit/integration test coverage.

### ⚠ CI status

**Clippy (all-features) and Code Style are red.** The PR does not compile clean on the quality gate currently. Rebase/fix before merge.

### Correctness risks

- **`extract_secret_binding_approval` in `src/bridge/effect_adapter.rs` is fragile.** It parses the approval JSON out of `error_msg` with `error_msg.rfind(\": {\")`. The test (`extract_secret_binding_approval_ignores_braces_in_tool_name`) pins one edge case, but this is stringly-typed out-of-band signalling between `http.rs` and the adapter — a tool name like `\"tool: {name\"` or an error message that happens to contain `\": {\"` before the real payload will mis-parse. Consider passing the approval as a typed `ToolError` variant (e.g. `ToolError::SecretBindingApprovalRequired(SecretBindingApproval)`) rather than round-tripping via `ExecutionFailed(json.to_string())`. The current shape is exactly the stringly-typed pattern `.claude/rules/types.md` warns about.

- **`job_ctx.metadata = serde_json::json!({ \"active_skill_names\": … })` in `effect_adapter.rs` around line 1240** overwrites any existing metadata the job context might gain. If another caller (or future code) puts a key on that metadata before this point, it gets dropped. Use `.as_object_mut()` + `.insert(…)`, or a typed field.

- **TOCTOU in `upsert_secret_for_user`** — `secrets.exists()` then `secrets.create()` is non-atomic. For the \"status: updated vs created\" return it's benign; the real question is whether `SecretsStore::create()` clobbers existing secrets with the same name (it appears to). Worth confirming and adding a regression note per `.claude/rules/review-discipline.md` (\"fix the pattern, not just the instance\") — the admin handler had the same shape pre-PR.

- **`auto_bound: false` is hardcoded** in `SecretBindingApprovalResponse` (`secrets.rs` list handler). Either plumb the real value or drop the field — a response field that's always the same literal is noise.

- **`revoke_secret_binding_approvals` best-effort on delete** — when a secret delete succeeds but approval revocation fails, the code logs a warning and returns success. That leaves dangling approvals whose `(secret_name, artifact_fingerprint)` no longer resolves. `is_known_secret_binding_approval` will reject them at check time, so it's safe, but consider whether this should propagate the error and force the caller to retry.

### Architectural notes

- **`MemorySettingsStore` is duplicated three times** (`bridge/auth_manager.rs` tests, `secrets/binding_approvals.rs` tests, `channels/web/handlers/secrets.rs` tests). Each is a ~80-line copy. Extract into `crate::testing` or similar. This exact pattern is flagged by `.claude/rules/review-discipline.md`.

- **Approval storage has no size bound.** `SECRET_BINDING_APPROVALS_KEY` is a single JSON array in a settings row. A user with many skills × hosts × secrets grows this linearly with no cap. `.claude/rules/safety-and-sandbox.md` → \"Bounded Resources\" applies.

- **`approval_lock`** uses a process-global `OnceLock<Mutex<HashMap<String, Weak<Mutex>>>>`. That serializes approval mutations per user *within one process* but provides no protection across replicas. Fine for current single-instance deployments; add a comment calling out the assumption so a future multi-tenant deploy doesn't silently race.

- **Dispatch discipline (`.claude/rules/tools.md`)** — the new `/api/secrets/*` handlers touch `state.secrets_store` directly rather than going through `ToolDispatcher`. The rule says all non-agent callers should dispatch, but `secrets_store` isn't in the pre-commit hook's enforced set. Worth either (a) adding a note here for the follow-up migration, or (b) confirming with a maintainer whether secrets CRUD belongs in dispatch. The admin handlers already did this pre-PR, so this isn't a regression.

- **`dedup_http_mappings` in `auth_manager.rs`** duplicates the intent of `dedup_credential_mappings` in `http.rs` with slightly different semantics. Consolidate to one helper.

- **Public `with_settings_store_override`** on `AuthManager` is described as test-scoped but is `pub`, not `pub(crate)` or `#[cfg(test)]`-gated. Anyone in the repo can wire a divergent settings store — tighten visibility.

### Security

- **Approval gate description is built from artifact-supplied strings** (`artifact_name`, `host`, `secret_name`). These are already validated at their boundaries (skill name validation, URL host parsing, `validate_credential_name`), so safe. Good.
- **Secret values never appear in list responses** — verified in `user_secret_round_trip_stays_scoped_to_authenticated_user`.
- **`is_known_secret_binding_approval`** re-validates the approval shape against the registry before granting, so a malicious resume payload can't approve an arbitrary binding. Good defensive design.

### Test coverage

- Strong: caller-level tests in `effect_adapter` (`preflight_ignores_skill_scoped_mapping_without_active_skill_context`) exercise the full preflight → mapping filter → auth check path, which matches `.claude/rules/testing.md` \"Test Through the Caller\". E2E covers the auth-gate → binding-approval → retry loop end-to-end.
- Missing: no test for the secret-delete-race-with-active-thread case (delete while a gate is pending); no test for duplicate-import behavior (same name appearing twice in one import payload).

### Minor

- `CLAUDE.md` / `src/channels/web/CLAUDE.md` route table is not updated with the four new `/api/secrets/*` routes or the `/api/secrets/{name}/approvals/revoke` route. Per `.claude/rules/review-discipline.md` — doc/route tables should update in the same PR.
- `location_risk`: only `QueryParam` is marked high. `UrlPath` puts the secret in the request path (logged by most proxies/CDNs) and arguably deserves `high` too.

### Recommendation

**Request changes** on three items before approve:
1. Fix red CI (clippy + fmt).
2. Replace the `rfind(\": {\")` JSON extraction with a typed `ToolError` variant.
3. Extract the triplicated `MemorySettingsStore` into a single test helper.

The rest (metadata overwrite, `auto_bound` hardcoding, docs, bounded storage) are good follow-ups.

serrrfirat:
Addressed the actionable review feedback in `55c96862` and pushed it to the PR head.

What changed:
- replaced the fragile secret-binding approval error-string parsing with a typed tool error path through tool execution/effect adaptation
- preserved existing `JobContext.metadata` when injecting `active_skill_names`
- extracted the duplicated test settings store helper into `src/testing/settings_store.rs`
- tightened approval persistence behavior: bounded retained approvals, documented the single-process lock assumption, and treat URL path credential locations as high risk
- fixed the secrets handlers to avoid `exists()+create()` TOCTOU, derive `created` vs `updated` from the stored secret, remove the hardcoded `auto_bound` response field, and fail closed if approval revocation fails during delete
- updated `src/channels/web/CLAUDE.md` to cover the `/api/secrets/*` routes
- added/updated regression coverage around the touched approval + secrets handler paths

Local verification:
- `cargo fmt`
- changed-area tests
- `cargo test --lib -- --test-threads=1` with the 8 pre-existing unrelated `tools::builtin::skill_tools::*zip*` failures skipped

I also re-requested review from @ilblackdragon, @henrypark133, and @zmanian.

ilblackdragon:
# Review: self-service user secrets + durable binding approvals

## Overview
Adds two coupled pieces of functionality:
1. **Self-service user secrets** — new `/api/secrets` routes and a Settings → Secrets panel, so members can CRUD their own secrets without admin endpoints.
2. **Durable secret binding approvals** — a new gate (`secret_binding_approval`) that pauses when a user-authored skill/WASM tool tries to use a secret on a host it hasn't been explicitly approved for. Approvals are persisted per-user in settings and invalidated when the artifact fingerprint changes.

`CredentialMapping` now carries optional `provenance` (kind, name, fingerprint, policy), and an `active_skill_names` channel is threaded from `ThreadExecutionContext` through the effect adapter into HTTP auth checks so skill-scoped credentials only apply when that skill is live.

Scope is large (~3.7k additions) but tightly scoped to one feature.

## Strengths
- **Defense in depth is well-layered**: pre-flight `check_action_auth` gates *before* the tool runs, and `HttpTool::execute` does a second approval check in case pre-flight is bypassed (e.g. `auth_manager = None`). The `ToolError::SecretBindingApprovalRequired` variant propagates up through `execute_tool_with_safety` and the adapter re-translates it into a `GatePaused` — a belt-and-suspenders path that matches how the rest of the auth system works.
- **Approval identity is stable and fingerprinted**: `SecretBindingApproval::approval_id()` hashes the secret/artifact/fingerprint/host/location tuple, so changing any one of them (including a skill content bump) transparently invalidates prior approvals.
- **Secret delete revokes approvals, and revoke failure aborts delete** (`secrets.rs` handler). The test covering "delete returns error when approval revoke fails" is exactly the right coverage.
- **E2E test** (`test_v2_secret_binding_approval.py`) exercises the full flow end-to-end: auth → binding approval → satisfied response → revoke → re-prompt on a new thread. The assertion that `mock_api["tokens"] == []` at the approval gate is a nice check that no credential leaked before approval.
- **Per-user in-process lock** (`approval_lock`) prevents grant/revoke races on a single replica, and the comment is honest about multi-replica limitations.
- **Tests follow "test through the caller" rule** — `AuthManager`, `EffectBridgeAdapter`, handlers, and the full gateway flow all get coverage, not just the low-level helper.

## Issues / suggestions

### Correctness / risk

- **`check_http_auth` returns `NoAuthRequired` when all matched mappings are filtered out by active-skill gating** (`src/auth/extension.rs`). That changes the semantics: previously, a matched-but-missing mapping bubbled `MissingCredentials`; now, if the *only* matches are skill-scoped and the skill isn't active, the result becomes `NoAuthRequired`, letting the call proceed unauthenticated. Confirm this is intentional vs. returning `Ready`/a neutral value. The test `check_http_ignores_skill_scoped_mapping_when_active_skills_unknown` asserts `NoAuthRequired` — so it's by design, but "no auth required" is a semantically stronger claim than "no skill-scoped credential applies here" and might surprise future readers. At minimum add a comment explaining why.

- **`approval_id()` JSON-hash is sensitive to `serde_json` feature flags** (`src/secrets/types.rs`). The hash input is `payload.to_string()` where `payload` is a `serde_json::json!` object. By default `serde_json::Map` is `BTreeMap` (sorted keys), but with the `preserve_order` feature it's `IndexMap` (insertion order). If any crate in the workspace enables `preserve_order`, the hash of every existing stored approval silently changes, meaning all users' approvals become unrecognized and re-gated. Recommend hashing a canonical representation — e.g. serialize to a `Vec<(&str, Value)>` in explicit order, or use `serde_json::to_vec` of a newtype with `#[serde]` field ordering fixed.

- **`exact_host_matches_pattern` colon handling is odd** (`src/secrets/types.rs`). The function is documented as "exact host comparison," but the second branch accepts `pattern_host == host` when `pattern` contains `:`. That means a pattern `api.github.com:443` matches host `api.github.com`. If host strings are canonicalized elsewhere to drop ports that's fine, but if a user declares `hosts: ["api.example.com:8443"]` expecting it to match only port 8443, this silently matches any port. Either document this or compare port explicitly.

- **`HttpTool::with_database(Arc<dyn Database>)`** (`src/tools/builtin/http.rs`) — the field is called `database` but is only used as a `SettingsStore`. Requiring a full `Database` to get settings access makes the tool harder to test and couples it to a broader trait than it needs. Consider taking `Arc<dyn SettingsStore + Send + Sync>` directly.

- **Pre-flight `active_skill_names` fallback re-loads the thread from the store** (`src/bridge/effect_adapter.rs`) when `context.active_skill_names` is empty. This is only there because some call paths don't populate the field yet — any remaining call sites that set an empty vec get a silent extra DB round-trip per tool call. Either (a) make every call site populate explicitly (a TODO/follow-up is fine), or (b) drop the fallback and accept that stale contexts will be treated as "no active skills" (which the tests already prove is safe).

### Code quality / style

- **Nested `.or_else().or_else()` in `AuthManager::settings_store`** (`src/auth/extension.rs`) — readable but the override→tools→extension-manager chain would be clearer as a `match` or `if let ... else if let` ladder.

- **`dedup_http_mappings` forwards to a free function** (`src/auth/extension.rs`) — it's a pointless wrapper — inline the single call.

- **`upsert_status(&secret)` heuristic** (`src/channels/web/handlers/secrets.rs`) compares `created_at == updated_at` to distinguish created/updated. This is fragile if `InMemorySecretsStore::create` and DB implementations don't agree on what to do with `updated_at` on an insert (the libsql/postgres impls aren't shown in the diff — worth verifying they set `updated_at = created_at` on first insert, otherwise "created" is misreported on new secrets).

- **UI: `wrap.appendChild(row)` precedes `row.appendChild(info)`** (`settings.js`) — works but reads as out-of-order. Build the row first, then append.

- **`var currentSecretEditName`** as a file-scope global — consistent with the rest of `settings.js`, so fine, but note that this leaks across settings subtabs if the user switches without clicking Clear. Minor.

### Security

- No injection / XSS surface jumped out — DOM is built with `textContent` and no `innerHTML` with user data.
- Secret values are never echoed in list responses (`SecretListItemResponse` has no `value` field, covered by the test at `user_secret_round_trip_stays_scoped_to_authenticated_user`).
- The `/api/secrets` path uses `AuthenticatedUser`, scoping by `user.user_id`. Admin path still requires `AdminUser`. Good isolation.
- `SECRET_BINDING_APPROVALS_KEY = "auth.secret_binding_approvals_v1"` — versioned key is smart, makes future schema changes possible without migration.

### Test coverage
Strong. Key gaps I'd flag:
- No test that `active_skill_names` pre-flight filtering degrades *gracefully* when the engine store is missing — the fallback in `effect_adapter.rs` returns `Vec::new()` silently. Worth asserting behavior under that condition.
- No test that `approval_id()` is stable across restarts (you can't easily test this, but a golden hash test — "this exact approval produces this exact ID" — would catch the `serde_json` feature-flag regression flagged above).

## Bottom line
The architecture is sound: provenance on mappings, policy-driven gating, shared gate/approval machinery, and a clear separation between admin and self-service. The risks are mostly subtle (JSON-hash stability, the semantic shift in `NoAuthRequired`, colon handling in exact-host matching). I'd want the `approval_id` hash-stability question answered before merge, and the `NoAuthRequired` semantics either confirmed-with-comment or tightened. Everything else is follow-up material.

zmanian (commented):
## Review: feat(web): add self-service user secrets and durable binding approvals

Architecture is solid — exact-host matching, content-addressed approval IDs, registry-validated approval echoing, secret values kept off list responses, dual-backend story works because approval persistence rides on `SettingsStore`. The findings below are the items I'd want addressed before this lands.

### HIGH

**H1. Cross-replica write loss on approval persistence**
`src/secrets/binding_approvals.rs:3275-3287, 3361-3383`
The per-user `Mutex` (`approval_lock`) only serializes within a process. Approvals are stored as a single JSON blob under one settings key (`auth.secret_binding_approvals_v1`), and the write path is read-modify-write at the application layer with no CAS or row-level locking. Two replicas granting different approvals concurrently can lose one. Same for `revoke_*`. The author calls this out in a comment but does not gate it.

Fix: either (a) add an `If-Match`-style version guard on `SettingsStore` (most robust), (b) move approvals to their own table/rows keyed by `(user_id, approval_id)` so concurrent inserts don't collide, or (c) document this as a known limitation in `src/channels/web/CLAUDE.md` with a clear "single-replica only until follow-up" caveat — and reference a follow-up issue from the comment.

**H2. `tracing::warn!` in approval gate hot path**
`src/bridge/effect_adapter.rs:1444-1451`
Per CLAUDE.md ("Logging levels matter for REPL/TUI"), `warn!` corrupts the interactive terminal and is reserved for user-facing status. This is engine-internal diagnostic info that fires on every approval-gated tool error.

Fix: demote to `debug!`. Sweep newly-added `info!`/`warn!` in this PR's hot paths.

**H3. No length cap on secret `value`**
`src/channels/web/handlers/secrets.rs:2070-2077, 2156-2177`
`SecretWriteRequest.value` is an unbounded `String`. Importing 100 secrets at e.g. 10 MB each through `/api/secrets/import` is currently bounded only by axum's body limit and the underlying store. Trivial DoS / disk-fill.

Fix: cap at e.g. 64 KiB in `create_secret_params`, return 400 over.

**H4. No rate limiting on self-service secrets endpoints**
`src/channels/web/platform/router.rs:3107-3116`
`GatewayState` already has `chat_rate_limiter`, `oauth_rate_limiter`, `webhook_rate_limiter`, but the new `/api/secrets/*` routes aren't wrapped. An authenticated but compromised session could spam create/delete/import to thrash the secrets store and oscillate approvals.

Fix: reuse a `PerUserRateLimiter` (e.g. 60 writes/min per user) in front of PUT/DELETE/POST.

### MEDIUM

**M1. Possible secret-value leakage through error messages**
`src/channels/web/handlers/secrets.rs:2202-2206, 2310-2317`
`secrets.create(...)` errors are surfaced verbatim via `e.to_string()` to the client and into request logs. If any underlying `SecretError` variant or wrapped `DatabaseError::Query(...)` ever interpolates the plaintext value (e.g. a DB constraint error including the row payload, or an encryption layer logging on failure), it lands in HTTP responses and tracing output.

Fix: don't propagate raw error strings from secret CRUD to clients. Map to a generic 500 ("Failed to store secret"), log with `debug!` server-side. Audit `SecretError` and `DatabaseError::Query` interpolation paths to confirm the value is never embedded.

**M2. UI misrenders bearer-auth approvals**
`crates/ironclaw_gateway/static/js/surfaces/settings.js:501-509`
`CredentialLocation::AuthorizationBearer` is a unit variant; serde's default external-tag representation serializes it as the JSON string `"AuthorizationBearer"`, not an object. The JS guard `if (!location || typeof location !== 'object') return '—';` therefore renders bearer (and basic, when no fields) as `—`.

Fix: handle `typeof location === 'string'` first: `if (location === 'AuthorizationBearer') return 'bearer';` then fall through to object cases.

**M3. CSRF posture for new write endpoints not stated**
`src/channels/web/platform/router.rs:3107-3116`
`AuthenticatedUser` is the same extractor used elsewhere, so this PR doesn't regress posture, but it does introduce new state-changing endpoints (PUT/DELETE/POST) that touch durable secrets. If session auth is cookie-based, confirm CSRF tokens are present (or that requests are bearer-only / `SameSite=strict`). Document the assumption near `start_server`.

**M4. Approval-revoke + secret-delete is not transactional**
`src/channels/web/handlers/secrets.rs:2459-2469, 2527-2538`
The handler calls `revoke_secret_binding_approvals` first, then `secrets.delete`. The order is correct (fail-closed: if delete fails, the user re-prompts on next use). But there's no compensating action if delete partially fails server-side. Worth noting in the function comment so future refactors don't reverse the order.

**M5. `dispatch-exempt` annotation expands secret-CRUD outside `ToolDispatcher`**
`src/channels/web/handlers/secrets.rs:2214-2225`
Per CLAUDE.md "Everything Goes Through Tools", new mutation surfaces should go through `ToolDispatcher::dispatch()`. The PR adds a sizable handler that talks to `state.secrets_store` and `state.settings_cache`/`state.store` directly, with one `dispatch-exempt:` annotation hidden inside `resolve_optional_settings_store`. The list/put/delete/import handlers themselves also bypass the dispatcher. Legitimate exemption (admin secrets handlers already worked this way), but coverage is incomplete.

Fix: add explicit `// dispatch-exempt: ...` comments at each handler's secrets-store / settings-store call site, or — preferred — file a follow-up to migrate secrets lifecycle into the dispatcher and reference that issue from the existing comment.

### LOW

**L1.** `binding_approval_exists` returns `Ok(false)` when `store` is `None` (`src/secrets/binding_approvals.rs:3349-3359`) — fail-closed for the check, which is right, but `grant_binding_approval` returns `Ok(())` and `revoke_*` returns `Ok(false/0)` in the same condition. Operators with a degraded DB see grants succeed visibly while doing nothing. Consider returning a typed error (`DatabaseError::Unavailable`) for grant/revoke when the store is missing.

**L2.** `approval_lock` registry trims only on miss (`src/secrets/binding_approvals.rs:3279-3287`) — `locks.retain(|_, lock| lock.strong_count() > 0)` runs only when a new lock is created. Long-lived processes with many one-shot users accumulate dead `Weak` entries until any miss occurs. Not a leak risk in practice; minor.

**L3.** `approval_id` includes `secret_name`, `host`, etc., but not `binding_policy` (`src/secrets/types.rs:3725-3742`). A binding flipping from `RequireApproval` to `AutoBind` would still match an old approval. Today the AutoBind path skips the approval lookup entirely, so it's not exploitable, but the docstring on `artifact_fingerprint` claims approvals are "invalidated when this changes." Either include the policy in the hash or document that policy changes intentionally reuse the existing approval.

**L4.** No coverage for empty/unicode/over-long secret names through the handler (per CLAUDE.md "Test Through the Caller").

### Verdict

**Request changes**, but the bones are good. I'd want H1 fixed-or-formally-deferred, H2 demoted, H3+H4 (cheap), and M1 audited so we can confidently say plaintext secret values cannot escape via error strings. M2 you'll likely catch in the promised browser smoke. Once H1–H4 and M1 are addressed (or explicitly deferred with reasoning), this is in good shape to land.

PR #2755, opened 2026-04-20
fix(engine-v2): recover flattened tool calls

serrrfirat:
## Summary
- recover NEAR-AI flattened bracket tool calls in the Engine V2 LLM bridge
- strip leaked internal `[Called tool ...]` markers from visible text responses
- add focused regression tests covering recovery and sanitization behavior

## Problem
In Engine V2, provider responses that flattened tool calls into text like:

- `[Called tool \`shell\` with arguments: {...}]`

could be treated as final assistant text instead of structured tool calls. That caused two user-visible failures:

- internal tool-call markers leaked into the Web UI/history
- the engine could prematurely complete a turn instead of actually executing the intended tool

Legacy/v1 paths already handled this cleanup/recovery, but the Engine V2 bridge path did not.

## Fix
- reuse `clean_response()` to sanitize visible assistant text
- reuse `recover_tool_calls_from_content()` to reconstruct structured tool calls from flattened content
- return `LlmResponse::ActionCalls` when recovery succeeds, preserving any cleaned visible preamble text

## Tests
- `cargo fmt --all --check`
- `cargo test --lib --features libsql complete_with_tools_recovers_flattened_bracket_tool_calls -- --nocapture`
- `cargo test --lib --features libsql complete_with_tools_strips_flattened_bracket_markers_from_text -- --nocapture`

## Notes
- This addresses the internal tool-marker leak / missed tool execution path in Engine V2.
- Full end-to-end manual revalidation in the clean Web UI instance is still pending.


serrrfirat:
Closing this draft because it was opened from the engine-v2-m0 branch lineage instead of a clean branch from staging. Replacing with a staging-based PR.

PR #2757, opened 2026-04-20
fix(engine-v2): recover flattened tool calls

serrrfirat:
## Summary
- recover provider-flattened bracket tool calls in the Engine V2 LLM bridge
- strip leaked internal `[Called tool ...]` markers from visible text responses
- add focused regression tests covering recovery and sanitization behavior

## Problem
In Engine V2, provider responses that flattened tool calls into assistant text like:

- `[Called tool \`shell\` with arguments: {...}]`

could be treated as final assistant text instead of structured tool calls. That caused two user-visible failures:

- internal tool-call markers leaked into the Web UI/history
- the engine could prematurely complete a turn instead of actually executing the intended tool

This is not specific to one provider in principle: it can affect any provider or response path that flattens tool calls into assistant content instead of returning structured tool calls. The concrete reproduction for this bug came from the NEAR AI provider path.

Legacy/v1 paths already handled this cleanup/recovery, but the Engine V2 bridge path did not.

## Fix
- reuse `clean_response()` to sanitize visible assistant text
- reuse `recover_tool_calls_from_content()` to reconstruct structured tool calls from flattened content
- return `LlmResponse::ActionCalls` when recovery succeeds, preserving any cleaned visible preamble text

## Tests
- `cargo fmt --all --check`
- `cargo test --lib --features libsql complete_with_tools_recovers_flattened_bracket_tool_calls -- --nocapture`
- `cargo test --lib --features libsql complete_with_tools_strips_flattened_bracket_markers_from_text -- --nocapture`

## Notes
- This PR branch is based directly on `staging` with only this fix cherry-picked on top.
- Full end-to-end manual revalidation in the clean Web UI instance is still pending.

### DIFF
--- a/src/bridge/llm_adapter.rs
+++ b/src/bridge/llm_adapter.rs
@@ -11,7 +11,7 @@ use rust_decimal::prelude::ToPrimitive;
 
 use crate::llm::{
     ChatMessage, LlmProvider, Role, ToolCall, ToolCompletionRequest, ToolDefinition,
-    sanitize_tool_messages,
+    clean_response, recover_tool_calls_from_content, sanitize_tool_messages,
 };
 
 /// Compute the USD cost of a single completion response, honoring the
@@ -161,7 +161,7 @@ impl LlmBackend for LlmBridgeAdapter {
         }
 
         // With tools: use tool completion (matches existing tools path)
-        let mut request = ToolCompletionRequest::new(chat_messages, tools)
+        let mut request = ToolCompletionRequest::new(chat_messages, tools.clone())
             .with_max_tokens(max_tokens)
             .with_temperature(temperature)
             .with_tool_choice("auto");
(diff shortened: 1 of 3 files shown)


serrrfirat:
Post-review fix added:

- sanitized the Engine V2 **plain completion / no-tools** path in `src/bridge/llm_adapter.rs`
- before this, the new recovery/sanitization logic only covered the tool-completion path
- that meant provider-flattened internal markers could still leak in text-only completions
- added regression coverage with:
  - `complete_without_tools_strips_flattened_bracket_markers_from_text`

Validation run:
- `cargo fmt --all --check`
- `cargo test --lib --features libsql complete_without_tools_strips_flattened_bracket_markers_from_text -- --nocapture`

This keeps the fix consistent across both:
- tool completion responses
- plain text completion responses

zmanian (approved):
LGTM — scoped V2-adapter fix with good test coverage (recovery, sanitization, plain-text, and fallback-on-empty cases). Reusing `clean_response` / `recover_tool_calls_from_content` from v1 is the right call.

Non-blocking nits:
- `EMPTY_CLEANED_RESPONSE_FALLBACK = "I'm not sure how to respond to that."` papers over what is almost always an upstream provider bug. Fine for UX continuity, but consider a `debug!`/`warn!` when it triggers so these don't vanish silently.
- `tools.clone()` in `complete_with_tools` is unavoidable given the later borrow for recovery; noted.

henrypark133 (commented):
Findings:
- No verified findings from this review pass.

Summary:
- Recommended verdict: Comment
- Residual risk: coverage is still mostly stub-provider based, so this does not prove a real provider variant produces the malformed flattened-call shape in practice. I did not run the local test suite.

PR #2760, opened 2026-04-20
fix(engine): stop failed missions from respawning (#2736)

### LINKED ISSUE #2736: [Bug Bash 4/20] Failed mission creates runaway threads and inflates Missions count
### Environment

hosted-staging (crab shack)

### Version / Commit Hash

v0.25.0 (staging)

### QA Test Date

2026-04-20

### Feature Area

Missions → error handling

### Steps to Reproduce

1. Create a routine/mission that fetches a GitHub API endpoint
2. 2. Let the fetch fail (e.g., use an invalid endpoint or let auth expire)
3. 3. Wait for multiple mission cycles to execute
4. 4. Navigate to the Missions page and observe thread count

### Expected Behavior

A failed fetch should be recorded as a status update within the existing mission thread. The Missions count should remain stable and not create new threads for each failure.

### Actual Behavior

Each failed mission cycle creates a new thread instead of updating the existing one. The Missions count keeps inflating (observed going from 15 to 29+ threads). This creates runaway thread creation that does not stop until the mission is manually deleted.

### Logs / Screenshots

Missions count observed: 15 threads initially, grew to 29+ within minutes. Each failed cycle spawns a new thread. Related to existing issue #2585 (missions count incorrect).

### Pre-submit checklist

serrrfirat:
## Summary
- mark missions as `Failed` when a mission thread ends in a terminal failure or hits max iterations
- keep the existing completion path unchanged so successful missions still transition to `Completed`
- add regression tests proving failed/max-iteration missions do not immediately re-fire

## Root cause
Mission thread failures were recorded in `approach_history`, but the mission lifecycle itself stayed `Active`. Because active cron/event missions remain eligible for future firings, the scheduler kept spawning fresh threads after each broken run, inflating the Missions count.

## Testing
- cargo test -p ironclaw_engine failed_outcome_marks_mission_failed_and_blocks_refire -- --nocapture
- cargo test -p ironclaw_engine max_iterations_marks_mission_failed_and_blocks_refire -- --nocapture
- cargo test -p ironclaw_engine outcome_processing_detects_goal_achieved -- --nocapture
- cargo fmt --all --check

Closes #2736

### DIFF
--- a/crates/ironclaw_engine/src/runtime/mission.rs
+++ b/crates/ironclaw_engine/src/runtime/mission.rs
@@ -508,12 +508,13 @@ impl MissionManager {
         Ok(())
     }
 
-    /// Resume a paused mission.
+    /// Resume a paused or failed mission.
     ///
     /// Shared missions can only be managed by shared owners (system user).
-    /// Only `Paused` missions can be resumed — `Completed` and `Failed` are
-    /// terminal states and must not be resurrected by a stray resume call,
-    /// so anything else is rejected with a `Store` error.
+    /// `Paused` missions resume normally, and `Failed` missions may be
+    /// explicitly resumed after the caller fixes the underlying problem.
+    /// `Completed` remains terminal, so anything else is rejected with a
+    /// `Store` error.
     pub async fn resume_mission(&self, id: MissionId, user_id: &str) -> Result<(), EngineError> {
         let mut mission = self
             .store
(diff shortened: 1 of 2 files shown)


henrypark133 (changes requested):
Findings:
1. High - `crates/ironclaw_engine/src/runtime/mission.rs:2260`, `crates/ironclaw_engine/src/runtime/mission.rs:517`, `crates/ironclaw_engine/src/types/mission.rs:267`, `src/channels/web/handlers/engine.rs:193`: the fix stops refiring by marking failed/max-iteration missions `Failed`, but `Failed` is terminal, `fire_mission` refuses terminal missions, and the only exposed recovery control still calls `resume_mission`, which only accepts `Paused`. That means the mission is no longer resumable through the public API after one failed run, despite the intended "stop until resumed" behavior.

Summary:
- Residual risk: I did not run the local test suite; this review is based on code inspection plus the current PR checks.

serrrfirat (commented):
Reviewed the latest mission changes. Allowing `resume_mission` to recover both `Paused` and `Failed` missions closes the API dead-end from the earlier review, and the new regression tests around failed/max-iteration outcomes cover the re-fire guard well.

One low-severity follow-up: when `process_mission_outcome` flips a mission to `Failed`, it leaves the mission id in the in-memory `active` / `last_fire_attempt` bookkeeping, unlike the pause/complete paths. Status checks still prevent refires, so this looks harmless, but it does leave stale runtime bookkeeping around until restart.

Issue #2585, opened 2026-04-17
[Bug Bash 4/17] Active missions not reflected in completed missions count


joe-rlo:
### Environment

hosted-staging (crab shack)

### Version / Commit Hash

staging-2026-04-17

### QA Test Date

2026-04-17

### Feature Area

Missions → status tracking dashboard

### Steps to Reproduce

1. Create and run a mission (e.g., bitcoin_price_check)
2. Let the mission execute
3. Open the Missions dashboard
4. Check Active vs Completed sections

### Expected Behavior

Completed missions should be moved from "Active" to "Completed" and reflected in the completed count.

### Actual Behavior

Missions appear only under "Active" and the "Completed" count remains 0, even after execution completes. Dashboard shows: 6 Total, 6 Active, 0 Paused, 0 Completed, 0 Failed. All missions (bitcoin_price_check, expected-behavior, conversation-insights, skill-extraction, skill-repair, self-improvement) remain in Active status indefinitely. Related to #2536 (mission-project pipeline) but this is specifically about status transition tracking.

### Logs / Screenshots

Discovered during Bug Bash 4/17 session. Missions dashboard screenshot shows all 6 missions stuck in Active with 0 Completed.

### Pre-submit checklist

- [x] Title is specific (not "fix Google Suite" but "Google Suite install throws invalid_grant on OAuth step")
- [x] Commit hash is filled in
- [x] Steps are numbered and reproducible

joe-rlo:
**Bug Bash 4/20 update:** Missions count is still incorrect on staging (v0.25.0) and the problem appears to be getting worse. The count keeps inflating over time — the dashboard now shows a higher total than the actual number of missions created. This suggests the counter is incrementing on each execution cycle rather than reflecting the true mission count. Likely related to the runaway thread issue filed in #2736.

henrypark133:
[...]

PR #2769, opened 2026-04-21
Show mission run history under active missions

henrypark133:
## Summary
- surface mission run history under active missions in the engine UI
- add run-summary data to mission list/detail API responses
- show completed/failed/in-progress prior runs in the Missions tab and project drill-in

## Why
Recurring missions should stay `Active` when they are still eligible to run again. The missing piece was visibility into prior run outcomes, not terminalizing the mission itself after each attempt.

This PR keeps mission-level scheduler state intact and makes per-run outcomes visible through persisted mission thread history.

## Validation
- `cargo fmt --all`
- `cargo clippy --all-targets --all-features -- -D warnings`
- `cargo test list_engine_missions_includes_run_summary_for_mission_threads --lib`
- `cargo test get_engine_mission_sorts_runs_newest_first_and_skips_missing_threads --lib`
- `node --check crates/ironclaw_gateway/static/js/surfaces/projects.js`
- `tests/e2e/.venv/bin/pytest tests/e2e/scenarios/test_project_detail.py -q`

Refs #2585


src/bridge/router.rs
```diff
@@ -4854,6 +4868,117 @@ fn thread_to_info(t: &ironclaw_engine::Thread) -> EngineThreadInfo {
...
+            threads.append(&mut shared_threads);
+        }
+    }
+
+    let mut threads_by_id = std::collections::HashMap::with_capacity(threads.len());
+    for thread in threads {
+        threads_by_id.insert(thread.id, thread);
+    }
+    Ok(threads_by_id)
+}
```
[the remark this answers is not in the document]
henrypark133: Addressed in `ecaeca35`. The mission list path now uses a lightweight `ThreadSummary` store query instead of loading full thread payloads, so run-summary aggregation no longer clones thread transcripts/events for list views.

src/bridge/router.rs
```diff
@@ -4854,6 +4868,117 @@ fn thread_to_info(t: &ironclaw_engine::Thread) -> EngineThreadInfo {
...
+    let mut threads = store
+        .list_threads(project_id, user_id)
+        .await
+        .map_err(|e| engine_err("list project threads", e))?;
+
+    if !ironclaw_engine::types::is_shared_owner(user_id) {
+        for owner_id in ironclaw_engine::types::shared_owner_candidates() {
+            let mut shared_threads = store
+                .list_threads(project_id, owner_id)
+                .await
```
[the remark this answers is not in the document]
henrypark133: Addressed in `ecaeca35`. Shared-owner mission listings now merge thread summaries across both shared owner candidates, and there is a regression test covering the `__shared__` plus legacy `system` case.

src/bridge/router.rs
```diff
@@ -4854,6 +4868,117 @@ fn thread_to_info(t: &ironclaw_engine::Thread) -> EngineThreadInfo {
...
+                .map_err(|e| engine_err("list shared project threads", e))?;
+            threads.append(&mut shared_threads);
+        }
```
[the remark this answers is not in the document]
henrypark133: Addressed in `ecaeca35`. I mirrored the shared-owner lookup pattern for mission run summaries and added a caller-level regression test for shared-owner sessions returning both shared mission run histories.

crates/ironclaw_gateway/static/js/surfaces/projects.js
```diff
@@ -415,7 +429,10 @@ function renderMissionThreadProgress(threadId) {
 
 function missionThreadIds(mission) {
   if (!mission || !Array.isArray(mission.threads)) return [];
-  return mission.threads.map((thread) => thread.id).filter(Boolean).sort();
+  return mission.threads
+    .map((thread) => [thread.id, thread.state, thread.updated_at, thread.created_at].join(':'))
```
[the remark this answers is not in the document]
henrypark133: Addressed in `ecaeca35`. `missionThreadIds()` now keys off `id`, `state`, and `created_at` only, so normal `updated_at` churn no longer forces unnecessary mission-detail rerenders.

crates/ironclaw_gateway/static/js/surfaces/projects.js
```diff
@@ -428,6 +445,91 @@ function haveMissionThreadsChanged(previousMission, nextMission) {
...
+    parts.push(normalized.completed_runs + ' ' + I18n.t('missions.runState.completed').toLowerCase());
+  }
+  if (normalized.failed_runs > 0) {
+    parts.push(normalized.failed_runs + ' ' + I18n.t('missions.runState.failed').toLowerCase());
+  }
+  if (normalized.in_progress_runs > 0) {
+    parts.push(normalized.in_progress_runs + ' ' + I18n.t('missions.runState.inProgress').toLowerCase());
+  }
+  if (parts.length === 0) {
+    parts.push(normalized.total_runs + ' ' + I18n.t('missions.runs').toLowerCase());
```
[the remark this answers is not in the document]
henrypark133: Addressed in `ecaeca35`. I removed the programmatic `.toLowerCase()` calls so the run-count labels now use the translated strings directly without locale-unsafe casing transforms.

src/bridge/router.rs
```diff
@@ -5295,22 +5433,15 @@ pub async fn list_engine_missions(
         .list_missions_with_shared(pid, user_id)
         .await
         .map_err(|e| engine_err("list missions", e))?;
+    let threads_by_id = load_visible_project_thread_summaries(&state.store, pid, user_id).await?;
```
serrrfirat, line 5540: **High Severity** — Mission run summaries can silently drop historical runs after restart/cleanup.

`list_engine_missions()` now derives `thread_count` and `run_summary` from `load_visible_project_thread_summaries()`, which ultimately calls `Store::list_thread_summaries()`. In the real `HybridStore`, that method only scans the in-memory `self.threads` map.

The problem is that `init_engine()` also calls `cleanup_terminal_state(5m)`, and that path explicitly evicts old terminal threads from memory while keeping their JSON on disk for lazy reload. After a normal restart (or once cleanup has run), older completed/failed mission runs are no longer present in `self.threads`, so mission list responses start undercounting runs or showing zero history even though `mission.thread_history` still points at valid persisted thread files. `get_engine_mission()` can still recover those runs one-by-one via `load_thread()`, so the list and detail endpoints diverge.

I think this needs to derive mission summaries from persisted thread history, not just the hot in-memory cache — e.g. reload each `thread_history` entry through `load_thread()` / a persistence-backed summary API, or teach `HybridStore::list_thread_summaries()` to include evicted-on-disk threads as well. A regression test that exercises `HybridStore` after `cleanup_terminal_state()` (or after process restart) would catch this.

src/bridge/router.rs
```diff
@@ -4854,6 +4868,130 @@ fn thread_to_info(t: &ironclaw_engine::Thread) -> EngineThreadInfo {
...
+    threads
+}
+
+async fn load_mission_run_threads(
+    store: &Arc<dyn Store>,
+    mission: &ironclaw_engine::Mission,
+) -> Result<Vec<ironclaw_engine::Thread>, Error> {
+    let mut threads = Vec::new();
+    for tid in &mission.thread_history {
+        match store.load_thread(*tid).await {
```
serrrfirat, line 4998: **High Severity** — Shared mission detail can leak other users' mission run metadata.

`get_engine_mission()` allows callers to view shared missions, and the new detail path builds the runs table by iterating `mission.thread_history` and loading every thread unconditionally. That means a shared OnEvent/Webhook mission can accumulate runs owned by different triggering users, but any caller who can view the shared mission will now receive the other users' run rows as well.

That ownership mix is reachable in the engine today: `MissionManager::fire_mission()` explicitly allows shared missions to be fired by any user, and the spawned thread inherits the requesting user's identity rather than the shared owner. So if Alice and Bob both trigger the same shared mission, Bob can inspect the mission detail and see Alice's run goal/state/timestamps/tokens here.

I think the detail endpoint needs the same viewer-based filtering that the list path tries to apply: only include runs whose thread owner is visible to the requesting user (or otherwise redefine shared-mission history as intentionally cross-user and document that contract very explicitly). A multi-user shared-mission test would be useful here.

src/bridge/router.rs
henrypark133: Addressed in `42c60fc2`. `HybridStore::list_thread_summaries()` now merges the in-memory thread map with persisted on-disk thread summaries from the workspace, deduped by `ThreadId`, so mission list run counts survive `cleanup_terminal_state()` eviction instead of depending on the hot cache alone. I also added a libsql-backed regression test that saves a completed mission thread, runs cleanup, and verifies `list_engine_missions()` still reports the historical run from persisted storage.

src/bridge/router.rs
henrypark133: Addressed in `42c60fc2`. `get_engine_mission()` now filters `mission.thread_history` through the same viewer-visibility ownership rule used by the list path before returning run rows, so shared mission detail only exposes runs owned by the requesting user or shared-owner threads they are allowed to see. I added a caller-level regression test that builds a shared mission with Alice, Bob, and shared-owned runs, then verifies Bob only receives Bob-visible rows and the corresponding filtered run summary.

src/bridge/router.rs
```diff
@@ -4947,6 +4961,141 @@ fn thread_to_info(t: &ironclaw_engine::Thread) -> EngineThreadInfo {
...
+        cadence_type: cadence_type_label(&mission.cadence).to_string(),
+        cadence_description: cadence_description(&mission.cadence),
+        thread_count: run_threads.len(),
+        run_summary: summarize_mission_runs(run_threads),
+        current_focus: mission.current_focus.clone(),
+        created_at: mission.created_at.to_rfc3339(),
+        updated_at: mission.updated_at.to_rfc3339(),
+    }
+}
+
```
serrrfirat, line 5098: **Medium Severity** · `[Likely]`

**`thread_count` semantic change breaks API contract.**

Previously `thread_count` was `mission.thread_history.len()` — counting all thread IDs including missing or inaccessible threads. Now it's `run_threads.len()` — only existing, user-visible threads.

This is arguably *more correct* (why show a count that includes phantom threads the user can't see?), but it's an undocumented API contract change. Any downstream consumer (dashboard, monitoring, webhook) that compared `thread_count` to the length of `thread_history` would see a discrepancy.

**Suggestion:** Either document the behavioral change in the PR description, or add a separate `total_thread_history_count: mission.thread_history.len()` field to preserve the old semantics while exposing the new filtered count — then deprecate the old field later.

crates/ironclaw_gateway/static/js/surfaces/projects.js
```diff
@@ -428,6 +445,91 @@ function haveMissionThreadsChanged(previousMission, nextMission) {
   return false;
 }
 
+function normalizeMissionRunSummary(summary, threadCount) {
+  const fallbackTotal = Number(threadCount) || 0;
+  return {
+    total_runs: Number(summary && summary.total_runs) || fallbackTotal,
+    completed_runs: Number(summary && summary.completed_runs) || 0,
```
serrrfirat, line 452: **Medium Severity** · `[Likely]`

**`normalizeMissionRunSummary` uses `||` instead of `??` for `total_runs` fallback.**

```js
total_runs: Number(summary && summary.total_runs) || fallbackTotal,
```

JavaScript's `||` treats `0` as falsy. If the backend ever sends `total_runs: 0` but `thread_count > 0` (e.g., from a stale cache merge or a future data path that decouples the two values), the UI would display the wrong count.

Today both values derive from the same `run_threads` vector so this can't manifest, but a future refactor could introduce the divergence silently — and this would be a very confusing UI bug to track down.

**Suggestion:**
```js
total_runs: (summary?.total_runs != null ? Number(summary.total_runs) : fallbackTotal),
// or: Number(summary?.total_runs ?? fallbackTotal)
```

henrypark133:
Temporarily closing and reopening to test whether reopened refreshes pull_request CI after retargeting to main.

serrrfirat:

PR #2774, opened 2026-04-21
fix(release): include sandbox_daemon in MSI

henrypark133:
Fixes the failed `Release` workflow for `ironclaw-v0.26.0` by updating `wix/main.wxs` to include `sandbox_daemon.exe` in the Windows MSI template.

Context:
- failed run: https://github.com/nearai/ironclaw/actions/runs/24705020263
- root cause: `cargo dist` detected `sandbox_daemon.exe` but the checked-in WiX template only referenced `ironclaw.exe`

Verification:
- matched the exact `cargo dist` diff from the failed run log
- `git diff --check -- wix/main.wxs` passes

This is a release-template fix only; no Rust code changes.

### DIFF
--- a/wix/main.wxs
+++ b/wix/main.wxs
@@ -128,6 +128,14 @@
                                 Source='$(var.CargoTargetBinDir)\ironclaw.exe'
                                 KeyPath='yes'/>
                         </Component>
+                        <Component Id='binary1' Guid='*'>
+                            <File
+                                Id='exe1'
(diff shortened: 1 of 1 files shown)


serrrfirat (approved):
Reviewed as paranoid architect — no findings. Clean 9-line fix that mirrors the existing `binary0` pattern exactly. Ship it.

PR #2777, opened 2026-04-21
feat: use llm to select skill when no skill selected

ddu-aeron:
we are trying to use skills from clawhub and it do not have activation keyworkds, this pr will try to select these skills using llm as a fallback

## Summary

- add LLM skill-selection fallback when no skills selected
- Add token cost tracking for the LLM skill-selection fallback

## Change Type

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] CI/Infrastructure
- [ ] Security
- [ ] Dependencies

## Validation

- [x] `cargo build`
- [x] Relevant tests pass: 
- [x] Manual testing:  tested with a skill without activation and it was selected
- [x] If a coding agent was used and supports it, `review-pr` or `pr-shepherd --fix` was run before requesting review

## Blast Radius

Touches `src/llm/reasoning.rs` (return type of `select_skill_names_with_llm`),
`src/agent/agent_loop.rs` (LLM fallback candidate filter + cost recording),
and test fixtures in `src/agent/dispatcher.rs`. No change to the happy path
for explicit or deterministic skill selection.

## Rollback Plan

Revert the three files changed in this PR. No schema or config changes involved.

## Review Follow-Through

The `fallback_candidates` filter now excludes any skill that has at least one
keyword, pattern, or tag. If a skill is intended to be description-only (no
activation criteria), it will still reach the LLM fallback. Reviewers should
confirm this is the desired boundary.

---

**Review track**:


serrrfirat:
Nice improvement for the v1 path. One thing to call out: this is still v1-only, since engine v2 does skill selection in `crates/ironclaw_engine/orchestrator/default.py` (via `select_skills(...)` / `__list_skills__()`), not in `src/agent/*`.

If we want parity, we should mirror the same behavior in v2 as a follow-up: keep explicit `/skill-name` + deterministic keyword/pattern selection as the primary path, and only fall back to an LLM-based selector when that path returns no skills. I’d also keep the same guardrail from this PR and only offer the LLM skills that have no positive activation metadata, so the LLM doesn’t second-guess deterministic exclusions.

Otherwise we’ll have different skill-selection behavior between v1 and v2 during the migration.

ddu-aeron:
> Nice improvement for the v1 path. One thing to call out: this is still v1-only, since engine v2 does skill selection in `crates/ironclaw_engine/orchestrator/default.py` (via `select_skills(...)` / `__list_skills__()`), not in `src/agent/*`.
> 
> If we want parity, we should mirror the same behavior in v2 as a follow-up: keep explicit `/skill-name` + deterministic keyword/pattern selection as the primary path, and only fall back to an LLM-based selector when that path returns no skills. I’d also keep the same guardrail from this PR and only offer the LLM skills that have no positive activation metadata, so the LLM doesn’t second-guess deterministic exclusions.
> 
> Otherwise we’ll have different skill-selection behavior between v1 and v2 during the migration.

You are correct! we should keep the same logic in v2 engine. but im not familiar with python and still dont know how to test v2 engine.

i will dig into the code base with help of AI. and try to impl it. it would be great if some would help!

src/agent/agent_loop.rs
```diff
@@ -650,22 +672,134 @@ impl Agent {
...
+                        && skill.manifest.activation.tags.is_empty()
+                        && skill
+                            .manifest
+                            .activation
+                            .setup_marker
+                            .as_ref()
+                            .is_none_or(|marker| !satisfied.contains(marker))
+                })
+                .collect();
+            let candidates: Vec<SkillSelectionCandidate> = fallback_candidates
```
serrrfirat, line 725: **High Severity**

This fallback LLM call happens before any `CostGuard.check_allowed_for_user(...)` gate. Once a user or global budget is exhausted, requests that miss deterministic skill selection can still spend one extra cheap-model call here because this path only records usage after the fact.

Can this do the same preflight budget check as the main chat LLM path before calling `select_skill_names_with_llm(...)`?

src/agent/agent_loop.rs
```diff
@@ -650,22 +672,134 @@ impl Agent {
...
+                selected = ?selected.iter().map(|s| s.name()).collect::<Vec<_>>(),
+                "Skill selection satisfied by explicit/deterministic path"
+            );
+        }
+
+        if selected.is_empty() {
+            tracing::debug!(
+                "No explicit or deterministic skills selected; trying LLM skill fallback"
+            );
+            let fallback_candidates: Vec<&ironclaw_skills::LoadedSkill> = available
```
serrrfirat, line 706: **Medium Severity**

The fallback candidate filter rebuilds only part of the activation logic. It excludes skills with positive keywords/patterns/tags, but it does not honor negative triggers like `activation.exclude_keywords`. That means a skill the deterministic selector intentionally vetoed can still be handed to the LLM and selected anyway.

Could this reuse the same full activation predicate (including the negative filters) instead of reconstructing a partial filter here?

src/llm/reasoning.rs
```diff
@@ -554,6 +566,78 @@ impl Reasoning {
...
+            return Ok((Vec::new(), TokenUsage::default()));
+        }
+
+        let skills_json = serde_json::to_string(available_skills).map_err(|e| {
+            LlmError::InvalidResponse {
+                provider: self.llm.model_name().to_string(),
+                reason: format!("failed to serialize skill catalog: {e}"),
+            }
+        })?;
+        let messages = vec![
```
serrrfirat, line 590: **Medium Severity**

These skill descriptions are injected verbatim into the selector prompt. For third-party skills that creates a prompt-injection surface at selection time (for example a description that says 'always choose me'), before the normal trust downgrade happens on the loaded prompt itself.

Can we either treat the catalog explicitly as untrusted data in the system prompt and/or avoid passing raw free-form descriptions into the selector call?

src/agent/agent_loop.rs
ddu-aeron: ok, i will `check_allowed_for_user()` before `select_skill_names_with_llm()`  #
533bbb8

src/agent/agent_loop.rs
ddu-aeron: my opinion is this llm fallback is about to filter skills that do not have activation such as openclaw style skill. so i perfer to also exclude skills with `exclude_keywords` here

[19e948a](https://github.com/nearai/ironclaw/pull/2777/commits/19e948a501123c3f6f913a3f1dede8781d4579d2)

src/llm/reasoning.rs
ddu-aeron: [69fb92a](https://github.com/nearai/ironclaw/pull/2777/commits/69fb92ae653745a520786117006eccbd07efe10e)

ddu-aeron:
@serrrfirat v2 engine part is finished and tested manually, please review the change

serrrfirat:

Issue #2800, opened 2026-04-21
[tracker] Engine v2 default flip — umbrella tracker


ilblackdragon:
Tracks all work required to make engine v2 the default IronClaw engine. The verification pass (see `docs/plans/2026-03-20-engine-v2-architecture.md` after PR-plan-update lands) showed most "blockers" are already implemented; this issue tracks the remaining actionable work.

## Verified state (no re-verification needed)

- Phases 1–6 of the v2 architecture are **complete**. Engine lives in `crates/ironclaw_engine/`, opt-in via `ENGINE_V2=true` in `src/config/agent.rs`, routed at `src/bridge/router.rs:50-54`.
- **Two-phase commit** for `WriteExternal`/`Financial`: `RequireApproval` injected at `crates/ironclaw_engine/src/capability/policy.rs:126-169`; executor halts with `ThreadOutcome::GatePaused` at `crates/ironclaw_engine/src/executor/structured.rs:139-171`. The unified gate resolver handles resume.
- **Orchestrator-owned compaction**: implemented at `crates/ironclaw_engine/orchestrator/default.py:240-310`. Triggers at 85% of context limit.
- **Engine-side cost tracking**: `TokenUsage.cost_usd` + `Thread.total_cost_usd` already threaded; orchestrator sums at `executor/orchestrator.rs:653,658`.
- **Routine → Mission UX**: `routine_to_mission_alias` at `src/bridge/effect_adapter.rs:1900,2273` already translates routine_create calls to mission_create for v2 users.
- **Engine-side Phase 7**: the engine crate has zero `JobState`/`Session`/`Routine` references.

## Work items

- [ ] **PR-A** (#____) Populate `cost_usd` in host `LlmProvider` adapters
- [ ] **PR-B** (#____) Wire `ReliabilityTracker` into context and effect adapter
- [ ] **PR-C** (#____) Alias `create_job`/`cancel_job` → `mission_create`/`mission_cancel`
- [ ] **PR-D** (#____) Expand acceptance test coverage for engine v2
- [ ] **PR-E** (#____) Port category-C v1 regression tests to v2
- [ ] **PR-differential** (#____) v2-vs-v1 strictly-better benchmark suite
- [ ] **PR-F** (#____) Default-flip plumbing: canary env, opt-out, CHANGELOG, docs, plan rewrite

## Category-C v1 regression tests to port (PR-E scope)

[...]

PR #2803, opened 2026-04-21
refactor(llm): share per-call cost formula between v1 and v2 (#2800 PR-A)

ilblackdragon:
Part of #2800 (engine v2 default flip).

Extracts the per-call cost computation from two duplicated inline implementations — v1 \`CostGuard::record_llm_call\` and v2 \`LlmBridgeAdapter::cost_usd_from\` — into \`src/llm/costs.rs\` as \`compute_call_cost_decimal\` / \`compute_call_cost_usd\`.

## Why

Both engines must bill identical numbers for the same LLM call so \`Thread.total_cost_usd\` (v2) and v1's daily-budget tracker agree. They did the same math in two places; drift was latent. This eliminates the risk.

## Scope note (vs original plan)

Original plan (#2800) suggested adding \`cost_usd: f64\` to \`CompletionResponse\` / \`ToolCompletionResponse\` directly so every provider populates it inline. That field lands on ~80 struct-literal callsites (concrete providers + wrappers + ~70 mocks/tests/v1 dispatchers). Deferred to a follow-up to keep this PR reviewable.

v2's \`TokenUsage.cost_usd\` is already populated at the bridge boundary (\`cost_usd_from\`) and now uses the shared formula; nothing changes on the consumer side except that the number is guaranteed identical to v1.

## Files touched

- \`src/llm/costs.rs\` — new shared helpers
- \`src/agent/cost_guard.rs\` — delegates to shared helper
- \`src/bridge/llm_adapter.rs\` — thin wrapper around shared helper; cleans up now-unused Decimal/ToPrimitive imports

## Tests

New tests in \`src/llm/costs.rs\`:
- \`compute_call_cost_decimal_matches_legacy_inline_formula\` — regression locking in the exact formula both sides used to compute
- \`compute_call_cost_zero_discount_does_not_panic\` — no div-by-zero
- \`compute_call_cost_zero_rate_returns_zero\` — Codex ChatGPT subscription returns 0.0 USD
- \`compute_call_cost_usd_tracks_decimal_path\` — f64 and Decimal paths produce identical (to f64 precision) numbers

All existing tests pass: 15 in \`llm::costs\`, 17 in \`agent::cost_guard\`, 43 in \`bridge::llm_adapter\`.

## Test plan

- [x] \`cargo check -p ironclaw --lib\` clean
- [x] Inline cost_guard tests green
- [x] Inline llm_adapter tests green (including \`complete_with_tools_populates_cost_usd_through_adapter\` which verifies the engine-facing path still works)
- [ ] Manual: compare v1 daily spend for a fixed workload against v2 \`total_cost_usd\` — they should match to 1e-9

## Follow-up

Tracking item in #2800: add \`cost_usd: f64\` field to \`CompletionResponse\` and \`ToolCompletionResponse\` so providers can emit the number directly (removes the bridge as the sole computer). Separate PR — touches ~80 callsites.

### DIFF
--- a/src/bridge/llm_adapter.rs
+++ b/src/bridge/llm_adapter.rs
@@ -6,9 +6,6 @@ use ironclaw_engine::{
     ActionDef, EngineError, LlmBackend, LlmCallConfig, LlmOutput, LlmResponse, ThreadMessage,
     TokenUsage,
 };
-use rust_decimal::Decimal;
-use rust_decimal::prelude::ToPrimitive;
-
 use crate::llm::{
     ChatMessage, LlmProvider, Role, ToolCall, ToolCompletionRequest, ToolDefinition,
     clean_response, recover_tool_calls_from_content, sanitize_tool_messages,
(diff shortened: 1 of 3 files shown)


ilblackdragon:
Review triage: gemini-code-assist left a summary review with no feedback items and no line comments on this PR. No changes needed.

src/bridge/llm_adapter.rs
```diff
@@ -39,31 +29,16 @@ fn cost_usd_from(
     cache_creation_input_tokens: u32,
```
serrrfirat, line 29: **Medium Severity**

This centralizes the billing formula, but not the rate source. `config.model` can override the model used for the request, while `provider.cost_per_token()` and the cache pricing knobs still come from the provider's active/default model. For override-capable providers that means v1 and v2 can still disagree on cost for the same call.

Could this price from the effective/requested model (or carry `cost_usd` through the response boundary) instead of asking the provider for its default rates here?

PR #2805, opened 2026-04-21
feat(engine): default-flip plumbing — opt-out, canary, docs (#2800 PR-F)

ilblackdragon:
Part of #2800 (engine v2 default flip).

Adds the mechanisms needed before the engine v2 default can flip on. **Does NOT flip the default itself** — that's a separate one-line follow-up after user review of the full tracker stack.

## New env vars

| Var | Meaning |
|-----|---------|
| \`IRONCLAW_LEGACY_ENGINE=true\` | Hard opt-out. Forces v1 even when v2 is the default or \`ENGINE_V2=true\` is set. Use during rollout if you hit a regression. |
| \`ENGINE_V2_CANARY_PCT=N\` | Routes N% of users to v2 based on a stable hash of \`user_id\`. Deterministic per user (same cohort across restarts). Ignored without a user_id. Beaten by explicit flags. |

### Precedence (first match wins)

1. \`IRONCLAW_LEGACY_ENGINE=true\` → **v1**
2. \`ENGINE_V2=true\` → **v2**
3. \`ENGINE_V2_CANARY_PCT=N\` + user_id → hash-based
4. Default → **v1** (until the flip lands)

## New API

- \`pub fn is_engine_v2_enabled()\` — env-only, for boot-time and aggregate observations
- \`pub fn is_engine_v2_enabled_for_user(user_id)\` — consults canary
- \`fn decide_engine_v2(legacy, v2, canary_pct, user_id)\` — private pure function hosting all branch logic so unit tests cover every precedence rule without mutating process-global env vars

## Docs

- \`docs/engine-v2.md\` — user-facing one-pager: opt in now, opt out later, participate in canary, report issues
- \`CHANGELOG.md\` Unreleased — opt-out + canary + reliability + cost-formula entries

## Tests

9 new unit tests in \`bridge::router::gating_tests\`:
- \`legacy_opt_out_beats_explicit_opt_in\`
- \`explicit_opt_in_turns_on_v2\`
- \`default_without_any_flags_is_v1\`
- \`canary_zero_pct_disables\`
- \`canary_hundred_pct_enables_all_users\`
- \`canary_is_stable_for_user_id\`
- \`canary_50_pct_splits_cohort\` (200-user cohort, expects 60-140 in v2)
- \`canary_ignored_without_user_id\`
- \`canary_invalid_string_falls_back_to_default\`

## Why no default flip in this PR

The actual flip (\`parse_bool_env(\"ENGINE_V2\", false)\` → \`true\` in \`src/config/agent.rs\`) is a one-line change that becomes trivially reviewable once this plumbing lands. Keeping it as a separate, obvious follow-up gives the reviewer one spot to sign off on (or stage with \`ENGINE_V2_CANARY_PCT=10\` first, then raise to 50, then flip the default).

Recommended rollout order:
1. Merge this PR (plumbing only, default still v1)
2. Set \`ENGINE_V2_CANARY_PCT=10\` for a week, watch error rates
3. Raise to \`50\` for another week
4. Merge the one-line \"default = true\" PR
5. Remove \`IRONCLAW_LEGACY_ENGINE\` one release later

## Test plan

- [x] \`cargo test -p ironclaw --lib bridge::router::gating_tests\` — 9 pass
- [x] \`cargo clippy -p ironclaw --lib --tests -- -D warnings\` clean
- [ ] Manual: start with \`ENGINE_V2_CANARY_PCT=50\`, verify \`/api/status\` reports \`engine_v2_enabled\` depending on user
- [ ] Manual: \`IRONCLAW_LEGACY_ENGINE=true ENGINE_V2=true\` keeps v1

### DIFF
--- a/docs/engine-v2.md
+++ b/docs/engine-v2.md
@@ -1,0 +1,102 @@
+# IronClaw Engine v2
+
+Engine v2 is IronClaw's unified execution engine. It replaces ~10 separate v1
+abstractions (Session, Job, Routine, Channel, Tool, Skill, Hook, Observer,
+Extension, LoopDelegate) with 5 primitives: **Thread**, **Step**,
+**Capability**, **MemoryDoc**, **Project**. See the architecture plan at
+`docs/plans/2026-03-20-engine-v2-architecture.md` for the full design.
+
+## Current status
+
+Engine v2 is **opt-in**. The default flip is tracked in issue #2800. Until
+that lands:
+
+- `ENGINE_V2=true` turns engine v2 on.
+- Without it, IronClaw keeps using the legacy v1 agent loop.
+
+## How to opt in
+
+```bash
+# Start the CLI with v2
+ENGINE_V2=true ironclaw
+
+# Or during development
+ENGINE_V2=true cargo run
+```
+
+You can confirm which engine served a given request via the web gateway
+status endpoint — the `engine_v2_enabled` field in `/api/status` reflects
+the current routing decision.
+
+## How to opt out (once v2 becomes the default)
+
+When the default flips, set `IRONCLAW_LEGACY_ENGINE=true` to keep v1:
+
+```bash
+IRONCLAW_LEGACY_ENGINE=true ironclaw
+```
+
+This overrides any other engine flag — including `ENGINE_V2=true`. Use it
+if you hit a regression during the rollout and need to fall back. The
+opt-out will be removed one release after v2 is stable as the default.
+
+## Canary rollout
+
+During the staged rollout, operators can run a percentage-based canary:
+
+```bash
+# Route 10% of users (by user_id hash) to v2, keep the rest on v1
+ENGINE_V2_CANARY_PCT=10 ironclaw
+```
+
+Properties:
+
+- **Deterministic per user** — a given `user_id` always falls in the same
+  cohort across restarts. Users aren't shuffled between engines.
+- **Ignored without a user** — boot-time observations and aggregate status
+  endpoints report v1 until a user's request arrives.
+- **Beaten by explicit flags** — `ENGINE_V2=true` forces v2 on, and
+  `IRONCLAW_LEGACY_ENGINE=true` forces v1 off, regardless of the canary %.
+
+## What users will notice on v2
+
+- **`create_job` / `cancel_job`** — translated to the mission manager. New
+  `create_job` calls show up under `/missions`, not `/jobs`.
+- **`routine_create` / `routine_list` / `routine_fire` / `routine_pause` /
+  `routine_resume` / `routine_delete` / `routine_update` / `routine_history`** —
+  all aliased to the corresponding `mission_*` actions. The `/routine`
+  slash command still works via the shared manager.
+- **Cost reporting** — v1 and v2 now compute identical USD numbers for the
+  same LLM call (shared formula in `src/llm/costs.rs`).
+- **Reliability data** — v2 records per-action success rate and latency in
+  a tracker that will surface "recently unreliable" hints into the system
+  prompt once wired. Toggle with `ENGINE_V2_RELIABILITY_HINTS=false`.
+
+## What v1 does that v2 does not (yet)
+
+- **`build_software`** — remains v1-only; no mission equivalent exists.
+
+## Debugging
+
+```bash
+# Full engine v2 debug logs
+ENGINE_V2=true RUST_LOG=ironclaw_engine=debug cargo run
+
+# Trace recording (for E2E replay)
+ENGINE_V2=true IRONCLAW_RECORD_TRACE=1 cargo run
+```
+
+Traces live under `~/.ironclaw/traces/` by default (override with
+`IRONCLAW_TRACE_OUTPUT`).
+
+## Reporting v2-specific issues
+
+Tag issues with `engine-v2` and include:
+
+- the `ENGINE_V2` / `IRONCLAW_LEGACY_ENGINE` / `ENGINE_V2_CANARY_PCT`
+  values active at the time
+- whether v2 was selected for your request (see `/api/status`
+  `engine_v2_enabled` field)
+- a trace file from `IRONCLAW_RECORD_TRACE=1` if possible
+
+The umbrella tracker for the default flip is #2800.
--- a/src/bridge/router.rs
+++ b/src/bridge/router.rs
@@ -47,10 +47,164 @@ pub enum BridgeOutcome {
 use std::collections::HashSet;
 
 /// Check if the engine v2 is enabled via `ENGINE_V2=true` environment variable.
+/// Returns true when engine v2 should handle traffic.
+///
+/// Precedence (first match wins):
+/// 1. `IRONCLAW_LEGACY_ENGINE=true` → **false** (hard opt-out; keeps v1 even
+///    if v2 is the default or explicitly enabled).
+/// 2. `ENGINE_V2=true` / `ENGINE_V2=1` → **true** (explicit opt-in).
+/// 3. Default → **false** until issue #2800 PR-F merges the actual flip.
+///
+/// For per-user canary routing, see [`is_engine_v2_enabled_for_user`].
 pub fn is_engine_v2_enabled() -> bool {
-    std::env::var("ENGINE_V2")
-        .map(|v| v == "true" || v == "1")
+    decide_engine_v2(
+        std::env::var("IRONCLAW_LEGACY_ENGINE").ok().as_deref(),
+        std::env::var("ENGINE_V2").ok().as_deref(),
+        None, // canary ignored without a user_id
+        None,
+    )
+}
+
+/// Returns true when engine v2 should handle **this specific user's** traffic.
+///
+/// Same precedence as [`is_engine_v2_enabled`], with an additional step between
+/// explicit opt-in and the default: if `ENGINE_V2_CANARY_PCT=N` is set,
+/// `hash(user_id) % 100 < N` routes the user to v2. This lets staged rollouts
+/// pick a reproducible cohort from `user_id` alone, without any per-user state.
+pub fn is_engine_v2_enabled_for_user(user_id: &str) -> bool {
+    decide_engine_v2(
+        std::env::var("IRONCLAW_LEGACY_ENGINE").ok().as_deref(),
+        std::env::var("ENGINE_V2").ok().as_deref(),
+        std::env::var("ENGINE_V2_CANARY_PCT").ok().as_deref(),
+        Some(user_id),
+    )
+}
+
+/// Pure-function core of the gating decision. Separated from env reads so the
+/// precedence rules are exhaustively unit-testable without mutating process
+/// globals (env vars are process-wide and race across test threads).
+fn decide_engine_v2(
+    legacy_flag: Option<&str>,
+    v2_flag: Option<&str>,
+    canary_pct: Option<&str>,
+    user_id: Option<&str>,
+) -> bool {
+    // 1. Hard opt-out wins.
+    if legacy_flag
+        .map(|v| matches!(v.to_ascii_lowercase().as_str(), "true" | "1" | "on"))
         .unwrap_or(false)
+    {
+        return false;
+    }
+    // 2. Explicit opt-in wins.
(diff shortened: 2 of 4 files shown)


serrrfirat:
## Paranoid Architect Review — Request Changes

**3 High, 3 Medium findings.** The opt-out mechanism has a critical gap: `IRONCLAW_LEGACY_ENGINE` doesn't actually prevent v2 routing at the primary decision point. Additionally, three independent env-var parsing pipelines for the same boolean flags disagree on case sensitivity and source hierarchy.

### High
- **`DefaultHasher` not stable across Rust versions** — canary cohort will silently reshuffle on toolchain upgrade. Codebase already uses `Uuid::new_v5()` for this reason (`signal.rs:420`).
- **Dual-path env reading**: `decide_engine_v2` uses raw `std::env::var()`, but `AgentConfig` uses `optional_env()` with runtime overrides/INJECTED_VARS. Setting `ENGINE_V2` via settings DB → `config.engine_v2=true` but `is_engine_v2_enabled()=false`.
- **Case sensitivity split**: `ENGINE_V2=TRUE` → `AgentConfig.engine_v2=true` (case-insensitive `parse_bool_env`) but `is_engine_v2_enabled()=false` (case-sensitive `v == "true"`). System runs v2 engine with v1 tool registry.

### Medium
- **Opt-out bypassed at routing point**: `IRONCLAW_LEGACY_ENGINE=true` + `ENGINE_V2=true` → `is_engine_v2_enabled()` returns false, but `AgentConfig.engine_v2` is true and `agent_loop.rs:1396` routes to v2 anyway.
- `is_engine_v2_enabled_for_user()` is exported but has zero call sites — canary mechanism is dead code.
- `docs/engine-v2.md` claims `/api/status` has `engine_v2_enabled` field — this field does not exist in `GatewayStatusResponse`.

### Recommendation
Consolidate env-var parsing into a single function used by both `AgentConfig` and `decide_engine_v2`. Use `parse_bool_env` semantics (case-insensitive, `true/1/false/0`) consistently. Wire `IRONCLAW_LEGACY_ENGINE` into `AgentConfig::resolve()` so the opt-out actually works at the routing decision point.

src/bridge/router.rs
```diff
@@ -47,10 +47,164 @@ pub enum BridgeOutcome {
...
+        let pct = pct.min(100);
+        if pct == 0 {
+            return false;
+        }
+        if pct >= 100 {
+            return true;
+        }
+        use std::hash::{Hash, Hasher};
+        let mut hasher = std::collections::hash_map::DefaultHasher::new();
+        user_id.hash(&mut hasher);
```
serrrfirat, line 116: **High: `DefaultHasher` is not stable across Rust versions.** Rust explicitly documents that `DefaultHasher` may change between compiler releases. A toolchain upgrade silently reshuffles the canary cohort. The codebase already avoids this in `src/channels/signal.rs:420-423` using `Uuid::new_v5()` for deterministic hashing.

**Fix:** Use `Uuid::new_v5(NAMESPACE_OID, user_id.as_bytes()).as_u128() % 100` or a stable hash like SipHash with fixed keys.

src/bridge/router.rs
```diff
@@ -47,10 +47,164 @@ pub enum BridgeOutcome {
...
+/// 3. Default → **false** until issue #2800 PR-F merges the actual flip.
+///
+/// For per-user canary routing, see [`is_engine_v2_enabled_for_user`].
 pub fn is_engine_v2_enabled() -> bool {
-    std::env::var("ENGINE_V2")
-        .map(|v| v == "true" || v == "1")
+    decide_engine_v2(
+        std::env::var("IRONCLAW_LEGACY_ENGINE").ok().as_deref(),
+        std::env::var("ENGINE_V2").ok().as_deref(),
+        None, // canary ignored without a user_id
```
serrrfirat, line 63: **High: Dual env-var reading path.** `decide_engine_v2` reads `ENGINE_V2` via raw `std::env::var()`, but `AgentConfig::resolve` reads it via `optional_env()` which also checks runtime overrides and INJECTED_VARS. If `ENGINE_V2` is set via settings DB, `config.engine_v2 = true` but `is_engine_v2_enabled() = false`.

**Fix:** Either have `decide_engine_v2` accept pre-resolved values from `AgentConfig`, or have `AgentConfig` delegate to `decide_engine_v2`.

src/bridge/router.rs
```diff
@@ -47,10 +47,164 @@ pub enum BridgeOutcome {
...
+    // 1. Hard opt-out wins.
+    if legacy_flag
+        .map(|v| matches!(v.to_ascii_lowercase().as_str(), "true" | "1" | "on"))
         .unwrap_or(false)
+    {
+        return false;
+    }
+    // 2. Explicit opt-in wins.
+    if v2_flag.map(|v| v == "true" || v == "1").unwrap_or(false) {
+        return true;
```
serrrfirat, line 101: **High: Case-sensitive parsing for ENGINE_V2 vs case-insensitive for IRONCLAW_LEGACY_ENGINE.** `ENGINE_V2=TRUE` → `AgentConfig.engine_v2=true` (via `parse_bool_env` which lowercases) but `is_engine_v2_enabled()=false` (this line does exact match). Result: v2 engine routing with v1 tool registry — hybrid state.

**Fix:** Use `v.to_ascii_lowercase()` here too, or better, use `parse_bool_env` for both.

src/bridge/router.rs
```diff
@@ -47,10 +47,164 @@ pub enum BridgeOutcome {
...
+///    if v2 is the default or explicitly enabled).
+/// 2. `ENGINE_V2=true` / `ENGINE_V2=1` → **true** (explicit opt-in).
+/// 3. Default → **false** until issue #2800 PR-F merges the actual flip.
+///
+/// For per-user canary routing, see [`is_engine_v2_enabled_for_user`].
 pub fn is_engine_v2_enabled() -> bool {
-    std::env::var("ENGINE_V2")
-        .map(|v| v == "true" || v == "1")
+    decide_engine_v2(
+        std::env::var("IRONCLAW_LEGACY_ENGINE").ok().as_deref(),
```
serrrfirat, line 61: **Medium: Opt-out is bypassed at the primary routing point.** `IRONCLAW_LEGACY_ENGINE=true` blocks `is_engine_v2_enabled()` but `AgentConfig.engine_v2` (which ignores `IRONCLAW_LEGACY_ENGINE`) is what `agent_loop.rs:1396` actually checks for routing. The opt-out doesn't work where it matters most.

**Fix:** Wire `IRONCLAW_LEGACY_ENGINE` into `AgentConfig::resolve()` so it overrides `engine_v2`.

src/bridge/router.rs
```diff
@@ -47,10 +47,164 @@ pub enum BridgeOutcome {
...
+}
+
+/// Returns true when engine v2 should handle **this specific user's** traffic.
+///
+/// Same precedence as [`is_engine_v2_enabled`], with an additional step between
+/// explicit opt-in and the default: if `ENGINE_V2_CANARY_PCT=N` is set,
+/// `hash(user_id) % 100 < N` routes the user to v2. This lets staged rollouts
+/// pick a reproducible cohort from `user_id` alone, without any per-user state.
+pub fn is_engine_v2_enabled_for_user(user_id: &str) -> bool {
+    decide_engine_v2(
```
serrrfirat, line 75: **Medium: Exported but never called.** `is_engine_v2_enabled_for_user()` is pub, re-exported from `bridge/mod.rs`, but has zero call sites. The canary mechanism is dead code. Without wiring into the per-request routing path, the canary is untestable in production.

docs/engine-v2.md
```diff
@@ -0,0 +1,102 @@
...
+ENGINE_V2=true ironclaw
+
+# Or during development
+ENGINE_V2=true cargo run
+```
+
+You can confirm which engine served a given request via the web gateway
+status endpoint — the `engine_v2_enabled` field in `/api/status` reflects
+the current routing decision.
+
```
serrrfirat, line 30: **Medium: Docs reference nonexistent API field.** Claims `/api/status` has `engine_v2_enabled` field, but `GatewayStatusResponse` has no such field. Users will look for it and file bugs.

serrrfirat (changes requested):

serrrfirat:
## Paranoid Architect Review — Request Changes

**3 High, 3 Medium findings.** The opt-out mechanism has a critical gap: three independent env-var parsing pipelines for the same boolean flags disagree on case sensitivity and source hierarchy, causing `IRONCLAW_LEGACY_ENGINE` to not actually prevent v2 routing at the primary decision point.

### High
- **`DefaultHasher` not stable across Rust versions** — canary cohort will silently reshuffle on toolchain upgrade. Codebase already uses `Uuid::new_v5()` for this reason (`signal.rs:420`).
- **Dual-path env reading**: `decide_engine_v2` uses raw `std::env::var()`, but `AgentConfig` uses `optional_env()` with runtime overrides/INJECTED_VARS. Setting `ENGINE_V2` via settings DB → `config.engine_v2=true` but `is_engine_v2_enabled()=false`. System enters hybrid state.
- **Case sensitivity split**: `ENGINE_V2=TRUE` → `AgentConfig.engine_v2=true` (case-insensitive `parse_bool_env`) but `is_engine_v2_enabled()=false` (case-sensitive exact match `v == "true"`). Result: v2 engine routing with v1 tool registry.

### Medium
- **Opt-out bypassed at routing point**: `IRONCLAW_LEGACY_ENGINE=true` blocks `is_engine_v2_enabled()` but `AgentConfig.engine_v2` (which ignores the legacy flag) is what `agent_loop.rs:1396` actually checks. The opt-out doesn't work where it matters most.
- `is_engine_v2_enabled_for_user()` is exported but has zero call sites — canary mechanism is dead code.
- `docs/engine-v2.md` claims `/api/status` has `engine_v2_enabled` field — field does not exist in `GatewayStatusResponse`.

### Recommendation
Consolidate env-var parsing into a single function used by both `AgentConfig` and `decide_engine_v2`. Use `parse_bool_env` semantics (case-insensitive) consistently. Wire `IRONCLAW_LEGACY_ENGINE` into `AgentConfig::resolve()` so the opt-out works at the routing decision point.

src/bridge/router.rs
```diff
@@ -47,10 +47,164 @@ pub enum BridgeOutcome {
...
+    {
+        let pct = pct.min(100);
+        if pct == 0 {
+            return false;
+        }
+        if pct >= 100 {
+            return true;
+        }
+        use std::hash::{Hash, Hasher};
+        let mut hasher = std::collections::hash_map::DefaultHasher::new();
```
serrrfirat, line 115: **High: `DefaultHasher` is not stable across Rust versions.** Rust explicitly documents that `DefaultHasher` may change between compiler releases. A toolchain upgrade silently reshuffles the canary cohort — users on v2 move to v1 and vice versa.

The codebase already has the correct approach: `src/channels/signal.rs:420-423` uses `Uuid::new_v5()` specifically because "This avoids relying on `DefaultHasher` implementation details."

**Fix:** Use `Uuid::new_v5(NAMESPACE_OID, user_id.as_bytes()).as_u128() % 100` or a stable hash like SipHash with fixed keys.

src/bridge/router.rs
```diff
@@ -47,10 +47,164 @@ pub enum BridgeOutcome {
...
+) -> bool {
+    // 1. Hard opt-out wins.
+    if legacy_flag
+        .map(|v| matches!(v.to_ascii_lowercase().as_str(), "true" | "1" | "on"))
         .unwrap_or(false)
+    {
+        return false;
+    }
+    // 2. Explicit opt-in wins.
+    if v2_flag.map(|v| v == "true" || v == "1").unwrap_or(false) {
```
serrrfirat, line 100: **High: Case-sensitive parsing for `ENGINE_V2` vs case-insensitive for `IRONCLAW_LEGACY_ENGINE`.** `ENGINE_V2=TRUE` → `AgentConfig.engine_v2=true` (via `parse_bool_env` which lowercases) but `is_engine_v2_enabled()=false` (this exact match). Result: v2 engine routing with v1 tool registry — hybrid state.

Also: `IRONCLAW_LEGACY_ENGINE` (6 lines above) accepts `"on"` but `ENGINE_V2` does not. Three different boolean parsing semantics for the same concept across the codebase.

**Fix:** Use `v.to_ascii_lowercase()` here, or better, use `parse_bool_env` for both flags.

Issue #2581, opened 2026-04-17
[Bug Bash 4/17] Agent exposes internal reasoning instead of final response (chain-of-thought leaked)


joe-rlo:
### Environment

hosted-staging (crab shack)

### Version / Commit Hash

staging-2026-04-17

### QA Test Date

2026-04-17

### Feature Area

Agent response → output formatting

### Steps to Reproduce

1. Open the web UI on staging (https://agent-stg.near.ai/)
2. Start a new chat or continue an existing conversation
3. Send a request that triggers the agent to use tools or reason about a task (e.g. "Send a monitoring request via Telegram")
4. Observe the agent's response in the chat window

### Expected Behavior

The agent should only display its final, user-facing response. Internal reasoning, chain-of-thought steps, tool call planning, and intermediate logic should be hidden from the user.

### Actual Behavior

The agent's response includes raw internal reasoning and chain-of-thought content that should be hidden. The user sees planning steps, tool selection logic, and intermediate thoughts mixed into the visible response instead of a clean final answer. This is a data leak of internal agent behavior and could expose system prompt details or reasoning patterns to end users.

### Logs / Screenshots

Discovered during Bug Bash 4/17 session (UC5: Telegram monitoring). The agent displayed its internal reasoning about how to handle the monitoring request instead of providing the clean final response.

### Pre-submit checklist

- [x] Title is specific (not "fix Google Suite" but "Google Suite install throws invalid_grant on OAuth step")
- [x] Commit hash is filled in
- [x] Steps are numbered and reproducible

joe-rlo:
**Bug Bash 4/20 update:** Still reproducing on staging (v0.25.0). New variant observed during X/Twitter summarization flow — when asked to summarize an X/Twitter thread, the agent exposes its internal reasoning about how to fetch and parse the tweet content instead of providing the clean summary. The chain-of-thought leak appears to happen across multiple tool-use scenarios, not just Telegram monitoring as originally reported.

PR #2812, opened 2026-04-21
fix(web): stop exposing internal reasoning in browser chat

serrrfirat:
## Summary
- stop broadcasting `ReasoningUpdate` events to browser SSE clients
- hide persisted and in-memory tool-call `narrative` / `rationale` from normal web chat history
- add regression coverage for the browser-facing reasoning leak

## Problem
Issue #2581 reports that the web UI can show the agent's internal reasoning / chain-of-thought during tool-use turns instead of only the final user-facing answer.

The immediate leak surface was browser-specific:
- live SSE `ReasoningUpdate` events
- persisted web history `narrative`
- persisted / in-memory per-tool `rationale`

This PR applies a focused browser-side fix so the web UI still shows:
- final assistant responses
- coarse thinking/status updates
- tool lifecycle activity

…but not the agent's internal planning text.

## Test plan
- [x] `cargo test --lib gateway_send_status_does_not_broadcast_reasoning_updates`
- [x] `cargo test --lib test_build_turns_with_wrapped_tool_calls_format`
- [x] `cargo test --lib test_in_memory_turn_info_hides_internal_reasoning`

### DIFF
--- a/src/channels/web/features/chat/mod.rs
+++ b/src/channels/web/features/chat/mod.rs
@@ -1046,7 +1046,9 @@ fn turn_info_from_in_memory_turn(t: &crate::agent::session::Turn) -> TurnInfo {
                     result: tool_result_preview(tc.result.as_ref()),
                     result_preview: None,
                     error: tc.error.as_deref().map(tool_error_for_display),
-                    rationale: tc.rationale.clone(),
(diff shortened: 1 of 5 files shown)


serrrfirat:

Issue #2813, opened 2026-04-21
engine-v2: add typed assistant content model for final vs internal tool-use text


serrrfirat:
## Summary
Engine v2 currently distinguishes between visible `thread.messages` and internal `thread.internal_messages`, but it still flattens assistant tool-use text too aggressively inside the working flow.

In particular, `LlmResponse::ActionCalls { calls, content }` uses a single optional `content` string for several semantically different things:
- user-visible preamble / narrative
- internal reasoning / planning
- per-tool rationale
- partial/final answer-ish text

That ambiguity makes it easy for adapters and UI layers to accidentally surface internal reasoning, as seen in #2581.

Quick browser-side mitigation landed separately in PR #2812, but this issue tracks the deeper engine-v2 model fix.

## Current state
Relevant structures today:
- `crates/ironclaw_engine/src/types/step.rs`: `LlmResponse::ActionCalls { calls, content }`
- `crates/ironclaw_engine/src/types/message.rs`: `ThreadMessage` only has `role/content/action_calls`
- `crates/ironclaw_engine/src/types/event.rs`: `MessageAdded { role, content_preview }` is also too coarse
- `src/bridge/llm_adapter.rs`: provider/tool-call semantics are flattened when converting into engine v2 types

Engine v2 already has a strong split between:
- visible transcript: `thread.messages`
- internal transcript: `thread.internal_messages`

…but it lacks a first-class typed assistant-content model for tool-use turns.

## Problem
Because assistant tool-use text is not strongly typed, downstream layers cannot reliably distinguish:
- internal reasoning that must never be shown to end users
- user-visible narrative that may be safe to show
- per-tool rationale/debug info
- final answer text

This causes two classes of bugs:
1. **Leaks**: browser/UI surfaces accidentally render planning text
2. **Lossy adapters**: bridge/event layers fall back to heuristics instead of using semantics

## Proposed direction
Introduce a typed assistant content model in engine v2, e.g. something in the spirit of:
- final answer
- user-visible narrative
- internal reasoning
- tool rationale
- tool calls / code

Exact naming can change, but the important part is that assistant tool-use output is no longer represented as one ambiguous `Option<String>`.

## Acceptance criteria
- engine v2 can represent assistant tool-use turns without conflating internal reasoning and user-visible content
- bridge adapters preserve those semantics instead of flattening them away
- visible transcript sync only promotes explicitly user-visible/final content
- web/UI layers no longer need heuristics to decide whether a string is safe to show
[...]

PR #2815, opened 2026-04-21
feat(engine): add typed assistant content model

serrrfirat:
## Summary
- add typed engine-v2 assistant content semantics for final / user-visible / internal assistant text
- preserve per-tool rationale separately on `ActionCall`
- thread the new semantics through the bridge adapter and Python orchestrator interchange
- keep legacy raw `content` strings for backwards-compatible persistence / replay paths

## Why
Follow-up to issue #2813.

Engine v2 already separates visible `thread.messages` from internal `thread.internal_messages`, but tool-use assistant text was still too lossy:
- shared `ActionCalls.content` strings were semantically ambiguous
- per-tool rationale was flattened away at the bridge boundary
- adapters / UI layers had to guess whether assistant text was user-visible or internal

This PR introduces a typed model so those semantics survive the engine + bridge path.

## Scope
This PR is intentionally stacked on top of #2812 (quick browser-side mitigation) so the browser leak is fixed immediately while the deeper engine-v2 model change stays reviewable on its own.

## Test plan
- [x] `cargo test -p ironclaw_engine`
- [x] `cargo test --lib bridge::llm_adapter::`
- [x] `cargo test --lib channels::web::`


serrrfirat:

PR #2817, opened 2026-04-21
fix(missions): enforce correct uuid

gagdiez:
mission_* id extraction defaulted to "" (empty string) when the ID was incorrectly processed. This created errors down the line that the LLM was not capturing.

This commit creates a shared `resolve_mission_id` helper in `handle_mission_call` that accepts `id` / `mission_id`, rejects empty values with an actionable message naming the action and the accepted kwargs, and quotes the offending value in UUID parse errors.

The helper is now used in `mission_get`, `mission_fire`, `mission_pause`, `mission_resume`, `mission_complete`, and `mission_update` — removing the duplicated inline `.unwrap_or("")` extraction blocks. Add three caller-level regression tests driving `execute_action` for the missing-id, non-UUID-id, and `mission_id` alias paths.

#### P.S. routine to mission

On fixing a small line on this PR I figured out that there was a missing piece of the puzzle when migrating routine_* inputs into mission functions. The routine_* use `name` as the input to refer to the routine, while missions are defined by uuid.

Added a small helper function to translate routine names into mission uuids. Otherwise, `names` were trying to be translated into `uuids` and always producing an error (as a "name" `string` would never be a valid uuid)

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] CI/Infrastructure
- [ ] Security
- [ ] Dependencies

## Validation

- [x] `cargo build`
- [x] Relevant tests pass: `cargo test -p ironclaw --lib bridge::effect_adapter`
- [ ] Manual testing: 

## Security Impact

None

## Database Impact

None

## Blast Radius

missions -> triggers an error if id is incorrect

## Rollback Plan

-

**Review track**:A?


src/bridge/effect_adapter.rs
```diff
@@ -371,6 +371,32 @@ impl EffectBridgeAdapter {
...
+                .get("id")
+                .or_else(|| params.get("mission_id"))
+                .or_else(|| params.get("name"))
+                .or_else(|| params.get("_args").and_then(|a| a.get(0)))
+                .and_then(|v| v.as_str())
+                .filter(|s| !s.is_empty())
+                .ok_or_else(|| EngineError::Effect {
+                    reason: format!(
+                        "{action} requires a mission id — pass `id=<uuid>` \
+                         (also accepts `mission_id`, `name`, or positional)"
```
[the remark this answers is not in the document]
gagdiez: dropped "name" entirely on [7bd2efa](https://github.com/nearai/ironclaw/pull/2817/commits/7bd2efaf970e11363d6f9d4531b2703ef0bd9c2a) in the `resolve_mission_id` helper, since `name` would never be the `id` that we need to resolve.

src/bridge/effect_adapter.rs
```diff
@@ -5474,6 +5444,121 @@ Use this skill to set up a Pika meeting.
...
+            let err = result
+                .output
+                .get("error")
+                .and_then(|v| v.as_str())
+                .unwrap_or_default();
+            assert!(
+                err.contains(action) && err.contains("requires a mission id"),
+                "{action} error should name the action and say it requires a mission id, got: {err}"
+            );
+        }
```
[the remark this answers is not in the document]
gagdiez: implemented in [7bd2efa](https://github.com/nearai/ironclaw/pull/2817/commits/7bd2efaf970e11363d6f9d4531b2703ef0bd9c2a)

src/bridge/effect_adapter.rs
```diff
@@ -368,8 +384,33 @@ impl EffectBridgeAdapter {
...
+                .get("id")
+                .or_else(|| params.get("mission_id"))
+                .or_else(|| params.get("_args").and_then(|a| a.get(0)))
+                .and_then(|v| v.as_str())
+                .filter(|s| !s.is_empty())
+                .ok_or_else(|| EngineError::Effect {
+                    reason: format!(
+                        "{action} requires a mission id — pass `id=<uuid>` \
+                         (also accepts `mission_id` or positional)"
+                    ),
```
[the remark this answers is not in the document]
gagdiez: removed `name` from PR description

src/bridge/effect_adapter.rs
```diff
@@ -368,8 +384,33 @@ impl EffectBridgeAdapter {
...
+                })?;
+            uuid::Uuid::parse_str(id_str)
+                .map(ironclaw_engine::MissionId)
+                .map_err(|e| EngineError::Effect {
+                    reason: format!(
+                        "invalid mission id {id_str:?}: {e}. Expected a v4 UUID \
```
gagdiez:
```suggestion
                        "invalid mission id {id_str:?}: {e}. Expected a UUID \
```

src/bridge/effect_adapter.rs
```diff
@@ -368,8 +384,33 @@ impl EffectBridgeAdapter {
...
+                         (8-4-4-4-12 hex digits); call `mission_list()` to look \
+                         up the id of an existing mission."
+                    ),
+                })
```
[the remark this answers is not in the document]
gagdiez: [e28f842](https://github.com/nearai/ironclaw/pull/2817/commits/e28f8428217865e572473e3f280f43ad0460eb76)

src/bridge/effect_adapter.rs
```diff
@@ -348,13 +348,27 @@ impl EffectBridgeAdapter {
...
+        // aliases must bridge the two conventions before dispatch — otherwise
+        // `routine_delete(name="Daily Check")` arrives at `mission_complete`
+        // as `{"name":"Daily Check"}` and fails UUID parsing.
+        // Allow the unresolved alias to flow into the normal mission handling
+        if let Some(alias) = routine_alias.as_mut() {
+            let _ = resolve_mission_identity(alias, &mgr, context).await;
+        }
+
         let (effective_action, effective_params, post_create_update) =
             if let Some(alias) = routine_alias.as_ref() {
```
serrrfirat, line 713: **High Severity** — `let _ =` silently swallows `resolve_mission_identity` error, causing test failure

The `let _ =` on this line discards the `Result` from `resolve_mission_identity`. When name resolution fails (e.g. "no mission named 'ghost routine'"), the error is silently dropped and execution falls through to `resolve_mission_id`, which produces a generic "mission_complete requires a mission id" error instead of the specific "no mission named 'X'" message.

**Confirmed: `routine_delete_by_unknown_name_surfaces_clear_error` test FAILS:**

```
thread '...routine_delete_by_unknown_name_surfaces_clear_error' panicked at:
unknown name should bubble up as Err: ActionResult { call_id: "rd2",
  action_name: "mission_complete",
  output: Object {"error": String("effect execution error: mission_complete
    requires a mission id — pass `id=<uuid>` (also accepts `mission_id`
    or positional)")}, is_error: true, ... }
```

The test expects `"no mission named"` but gets `"requires a mission id"`.

**Suggested fix:** Propagate the error instead of discarding it. Since `handle_mission_call` returns `Option<Result<...>>`, `?` won't work directly. Use an explicit match:

```rust
if let Some(alias) = routine_alias.as_mut() {
    if let Err(e) = resolve_mission_identity(alias, &mgr, context).await {
        return Some(Err(e));
    }
}
```

src/bridge/effect_adapter.rs
```diff
@@ -637,94 +666,48 @@ impl EffectBridgeAdapter {
...
+                        Ok(()) => Ok(serde_json::json!({"status": "ok"})),
                         Err(e) => Err(e),
-                    },
-                    Err(e) => Err(e),
+                    }
                 }
-            }
+                Err(e) => Err(e),
+            },
+            "mission_complete" => match resolve_mission_id(params, "mission_complete") {
```
serrrfirat, line 1042: **Medium Severity** — `mission_complete` missing user_id authorization check

`mission_complete` calls `mgr.complete_mission(id)` without passing `user_id`, unlike every other mission action in this file:

- `fire_mission(id, &context.user_id, None)`
- `pause_mission(id, &context.user_id)`
- `resume_mission(id, &context.user_id)`
- `update_mission(id, &context.user_id, updates)`

If the `MissionManager::complete_mission` method does not perform its own internal ownership check, this is an IDOR — any user who knows (or guesses) a mission UUID can complete another user's mission. Pre-existing issue, but this PR is the right moment to add it since you're already refactoring every mission action handler.

**Suggested fix:** Either add an ownership check here (load the mission, verify `mission.user_id == context.user_id`), or update the `MissionManager::complete_mission` signature to accept `user_id` for internal enforcement.

src/bridge/effect_adapter.rs
gagdiez: [ece9f09](https://github.com/nearai/ironclaw/pull/2817/commits/ece9f09a98e3b864f741c487114886c7b8073e4f)

src/bridge/effect_adapter.rs
gagdiez: [eb16cbb](https://github.com/nearai/ironclaw/pull/2817/commits/eb16cbb936809fa88da68b271ff1f9a5365b6d64)

src/bridge/effect_adapter.rs
```diff
@@ -1886,6 +1871,57 @@ struct RoutineMissionAlias {
...
+        return Ok(());
+    }
+    let Some(name) = obj.get("name").and_then(|v| v.as_str()).map(str::to_string) else {
+        return Ok(());
+    };
+
+    let missions = mgr
+        .list_missions(context.project_id, &context.user_id)
+        .await?;
+    let mut matches = missions.iter().filter(|m| m.name == name);
```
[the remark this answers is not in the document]
gagdiez: lets leave this for another PR

src/bridge/effect_adapter.rs
```diff
@@ -1886,6 +1871,57 @@ struct RoutineMissionAlias {
...
+    if matches!(alias.mission_action, "mission_create" | "mission_list") {
+        return Ok(());
+    }
+
+    let Some(obj) = alias.mission_params.as_object_mut() else {
+        return Ok(());
+    };
+    if obj.contains_key("id") {
+        return Ok(());
+    }
```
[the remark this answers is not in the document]
gagdiez: [eedc272](https://github.com/nearai/ironclaw/pull/2817/commits/eedc272610f6a5382985bf4a859e126b9bd81190)

src/bridge/effect_adapter.rs
```diff
@@ -1886,6 +1871,66 @@ struct RoutineMissionAlias {
...
+    if has_id_claim {
+        return Ok(());
+    }
+    obj.remove("id");
+    let Some(name) = obj.get("name").and_then(|v| v.as_str()).map(str::to_string) else {
+        return Ok(());
+    };
+
+    let missions = mgr
+        .list_missions(context.project_id, &context.user_id)
```
serrrfirat, line 2201: **Medium Severity**\n\nThe new `routine_*` name resolver is looking up names through `MissionManager::list_missions()`, which returns **own + shared** missions. That widens the lookup scope beyond the v1 routine contract documented just above (`UNIQUE (user_id, name)`).\n\nConcrete failure mode: if a user-owned mission and a shared mission have the same display name, a legacy call like `routine_delete(name="Daily Check")` now resolves two matches and returns the new "multiple missions are named ..." error instead of deterministically targeting the user-owned routine.\n\nI think the resolver should preserve the v1 scoping semantics for name-based routine aliases: either resolve against **owned missions only**, or prefer an owned match before considering shared missions.\n\nA good regression test here would create one owned mission and one shared mission with the same name, then drive `routine_delete` / `routine_history` through `execute_action()` to pin the intended behavior.

gagdiez: [eeb8e30](https://github.com/nearai/ironclaw/pull/2817/commits/eeb8e30ac3f001283843011c14c3a96479d11641)

crates/ironclaw_engine/src/runtime/mission.rs
```diff
@@ -572,7 +572,25 @@ impl MissionManager {
...
+        };
+        if !allowed {
+            return Err(EngineError::AccessDenied {
+                user_id: user_id.to_string(),
+                entity: format!("mission {id}"),
+            });
+        }
         self.store
             .update_mission_status(id, MissionStatus::Completed)
             .await?;
```
serrrfirat: **Medium Severity** · TOCTOU: load-then-update race introduced by ownership check

The old `complete_mission` was a single atomic `update_mission_status(id, Completed)` call. The new code splits this into two store operations:

1. `load_mission(id)` → ownership check
2. `update_mission_status(id, Completed)`

Between (1) and (2), a concurrent caller (e.g., cron ticker's `fire_mission`, or another `complete_mission` call) could observe a stale status. `resume_mission` was specifically refactored to avoid this exact pattern — see the comment at line 548:

> *Mutate-and-save in a single round-trip. The previous implementation did `update_mission_status(Active)` and then a separate `load+save` … between the two writes, a concurrent `update_mission`/`fire_mission` could modify other fields that the second save would then silently overwrite with the stale reload.*

**Practical risk is low** — mission ownership is immutable after creation, so the ownership check can't go stale. But for consistency with `resume_mission`'s pattern, consider:

```rust
pub async fn complete_mission(&self, id: MissionId, user_id: &str) -> Result<(), EngineError> {
    let mission = self.store.load_mission(id).await?.ok_or_else(|| EngineError::Store {
        reason: format!("mission {id} not found"),
    })?;
    // ... ownership check (unchanged) ...

    // Single mutate-and-save like resume_mission:
    // (complete_mission only sets status, so update_mission_status is fine here —
    // but switching to save_mission() would close the window for future fields)
    self.store.update_mission_status(id, MissionStatus::Completed).await?;
    // ...
}
```

This isn't blocking, but harmonizing `complete_mission` and `pause_mission` with `resume_mission`'s single-round-trip pattern would close this class of issue across all lifecycle methods.

gagdiez: [4c16b70](https://github.com/nearai/ironclaw/pull/2817/commits/4c16b701deae61c76b78218fd15f0669a7c222d2)

crates/ironclaw_engine/src/runtime/mission.rs
```diff
@@ -572,10 +572,28 @@ impl MissionManager {
...
+        };
+        if !allowed {
+            return Err(EngineError::AccessDenied {
+                user_id: user_id.to_string(),
+                entity: format!("mission {id}"),
+            });
+        }
+        mission.status = MissionStatus::Completed;
+        mission.updated_at = chrono::Utc::now();
+        self.store.save_mission(&mission).await?;
```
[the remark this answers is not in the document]
gagdiez: this contradicts the review by @serrrfirat so I will simply discard it

crates/ironclaw_engine/src/runtime/mission.rs
```diff
@@ -496,9 +496,9 @@ impl MissionManager {
                 entity: format!("mission {id}"),
             });
         }
-        self.store
-            .update_mission_status(id, MissionStatus::Paused)
-            .await?;
+        mission.status = MissionStatus::Paused;
+        mission.updated_at = chrono::Utc::now();
+        self.store.save_mission(&mission).await?;
```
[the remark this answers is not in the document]
gagdiez: this contradicts the review by @serrrfirat so I will simply discard it

serrrfirat:

PR #2819, opened 2026-04-22
[draft] feat(engine): python only engine

gagdiez:
## Summary

Our current engine is full of nudges, heuristics and small hacks to try to get the LLM to do what the user actually ask. Some times we try to catch user intention, some times we try to guess what the LLM was trying to produce, and I feel we end up missing most of the times.

This PR proposes to simplify the engine such that we ask the LLM to **produce a single block of python code**. This simplifies the current `CodeAct` implementation while at the same time producing - in my local experiments - a better end result since there are no heuristic or nudges misguiding the LLM.

In this model we tell the LLM to use python as a way of planning, asking it to produce one piece of code that possess all the necessary steps to resolve the situation, instead of trying to make multiple steps.

Modern LLMs are incredibly good at producing code - more importantly, many of them are fine-tuned with that purpose in mind. Because of that, I found that my agent was able to create missions, answer questions, use tools to navigate the web.

At the same time, error handling becomes simpler for us and the LLM. Instead of us trying to figure out what the LLM was trying to do, we pass it to python, and python returns the necessary errors.

The main drawback of this simplification is that the LLM rarely gets the solution in one-shot, generally it will first try to mix text with code, but after one or two fails it will simply move all its comments into `python` comments, and understand that it needs to use `FINAL(message)` to send a message to the user

@ilblackdragon have you tried this approach? any comments on why this would not be a good idea?

also, do we have any benchmark that are more "goal oriented" that I could use to make sure there is an improvement here, and is not only my own bias?

-

## Change Type

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] CI/Infrastructure
- [ ] Security
- [ ] Dependencies

## Validation

- [x] `cargo build`
- [x] Relevant tests pass: engine tests
- [x] Manual testing: 

## Security Impact

Simplifies the engine, while I did my best to check the code, something might have changed.

## Database Impact

None

## Blast Radius

Engine!

## Rollback Plan

We would need to rollback the whole PR, since this is a huge change

---

**Review track**: C


gagdiez:
converting this to a draft as right now I am more interested in bouncing ideas and hearing Illia's opinion

serrrfirat:
@gagdiez this is an interesting approach. couple of things come to mind:

- i would like to see some evidence of agent performing better like this (benchmarks, blackbox tests, live tests etc.) there is a branch with live tests and blackbox tests which you could use.

- testing this with abstract goals (long running research). this would also be a benchmark.

- model selection. curious if you are using a sota model on your tests?

i will do some tests with this engine version too!

gagdiez:
@serrrfirat what branches are you referring to?

Respect to the model, I am using Qwen 3.5 (122B-A10B)

serrrfirat:
Could you provide some benchmark results running these to start with ?

 1. zizmor_scan_v2
 2. mission_daily_news_digest_with_followup
 3. developer_full_workflow

PR #2822, opened 2026-04-22
feat(engine): type message-added assistant events

serrrfirat:
## Summary
- add a lightweight typed assistant-content kind to engine `MessageAdded` events
- emit that kind from `Thread::add_message` based on typed assistant message content
- stop the bridge/router SSE path from guessing assistant visibility from raw preview text alone
- suppress internal-reasoning message events while preserving legacy assistant fallback behavior

## Why
This is a stacked follow-up to #2815.

#2815 taught engine v2 messages and orchestrator interchange about typed assistant content, but `EventKind::MessageAdded` still only carried `role + content_preview`. That left the event/SSE layer guessing whether assistant preview text was user-visible or internal.

This PR threads a lightweight semantic tag through the event layer so the router can:
- hide `InternalReasoning`
- surface `UserVisibleNarrative`
- avoid duplicate preview events for `Final` assistant messages, which already flow through `AppEvent::Response`
- keep legacy assistant event behavior when no typed metadata exists

## Test plan
- [x] `cargo test -p ironclaw_engine types::thread::tests::add_`
- [x] `cargo test --lib bridge::router::tests::thread_event_to_app_events_`

### DIFF
--- a/crates/ironclaw_engine/src/lib.rs
+++ b/crates/ironclaw_engine/src/lib.rs
@@ -40,13 +40,13 @@ pub use types::capability::{
     PolicyEffect, PolicyRule,
 };
 pub use types::error::{CapabilityError, EngineError, StepError, ThreadError};
-pub use types::event::{EventId, EventKind, ThreadEvent};
+pub use types::event::{AssistantContentKind, EventId, EventKind, ThreadEvent};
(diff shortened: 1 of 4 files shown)


serrrfirat:

Issue #2767, opened 2026-04-21
Epic: Separate engine v2 capability background from callable tool schemas


henrypark133:
# Epic: Separate engine v2 capability background from callable tool schemas

Important: this epic is scoped to `engine_v2 = true` only. It changes the v2 bridge/orchestrator/LLM surfacing path and does not change legacy/v1 engine behavior.

## Summary

Engine v2 currently mixes together three different concerns:

- what runtime capabilities exist
- what is actually ready, blocked, latent, or inactive right now
- what tools are directly invokable by the model on this turn

That produces two bad outcomes:

1. the model does not get clear background truth about channels, extensions, and readiness
2. the provider still receives a broad tool-schema payload every turn

The new baseline for engine v2 should be:

- **Capability background**
  - channels
  - latent providers
  - installed but inactive extensions
  - auth/setup/error state
  - routing hints like `telegram usable through message`
- **Callable direct actions**
  - only the actions that are actually invokable right now
  - surfaced to the provider with their full schemas when they are part of the normal always-on callable set
- **Deferred large tools**
  - very large tools stay visible in a separate description-only discovery surface
  - they are not provider-callable by default
  - the model must fetch fuller details through `tool_info` before they enter the callable tool set

This epic is **not** proposing inspect/promote as the baseline runtime model.

The baseline should be:

- background truth for non-callable runtime state
- full schemas for normally surfaced callable direct actions
- deferred callability for very large tools
- deterministic host-side filtering only as a later optimization if payload is still too large

## Current broken behavior

Today the v2 path has three concrete problems.

### 1. Capability background is incomplete

The v2 system prompt gets capability summaries once at thread start, but the default `available_capabilities()` implementation still just degrades to direct-action summaries.

That means the model does not reliably learn about:

- channels
- latent providers
- installed-but-inactive extensions
- `needs_auth` / `needs_setup` states
- routing facts like `Telegram is usable through message`

### 2. Provider tool schemas are still broad

Each `__llm_complete__()` recomputes the live action set and sends those action schemas to the provider.

That means the system keeps paying for a large provider tool payload even when much of the surfaced runtime truth is not directly callable.

### 3. `available_actions()` currently includes things that should not be treated as callable

On the current v2 bridge path, `available_actions()` still includes latent inactive provider actions.

That is the wrong long-term contract.

If something is not currently invokable, it should not appear as a callable provider tool schema. It should appear in capability background instead.

Representative failure mode:

[...]

PR #2827, opened 2026-04-22
engine-v2: centralize action vs capability surface policy

henrypark133:
## Summary

- Stacked under issue `#2767` task plan: https://github.com/nearai/ironclaw/issues/2767#issuecomment-4293374615
- Adds a bridge-owned surface-assignment policy module on top of Section 1.
- Centralizes the actions-vs-capabilities matrix using `CapabilityStatus`.
- Keeps the change policy-only and out of production wiring for now.

## Finished

- Covers Task 2 from the issue stack: surface-assignment policy
- New internal bridge policy module: `src/bridge/tool_surface.rs`
- Matrix coverage for direct actions, routed channels, latent providers, inactive/auth/setup/error states, and available-not-installed entries
- Table-driven policy tests

## Left

- No `available_capabilities()` projection yet
- No production adoption in `available_actions()` yet
- No prompt/orchestrator changes yet

## Verification

- `cargo test tool_surface --lib`


src/bridge/tool_surface.rs
```diff
@@ -0,0 +1,274 @@
...
+            }
+        }
+    }
+}
+
+const fn is_direct_ready(status: CapabilityStatus) -> bool {
+    matches!(
+        status,
+        CapabilityStatus::Ready | CapabilityStatus::ReadyScoped
+    )
```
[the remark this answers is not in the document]
henrypark133: Addressed in `2b2f76c3`. `ReadyScoped` no longer counts as direct-ready in the surface policy, and I added a matrix row that pins scoped direct actions to `neither`.

src/bridge/tool_surface.rs
```diff
@@ -0,0 +1,274 @@
...
+                let _approval_gated = subject.approval_gated;
+                SurfaceAssignment::actions_only()
+            } else if is_direct_ready(subject.status) {
+                SurfaceAssignment::capabilities_only()
+            } else {
+                fallback_assignment(subject.status)
+            }
+        }
+    }
+}
```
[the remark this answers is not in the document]
henrypark133: Addressed the correctness part in `2b2f76c3`. `assign_surface` now short-circuits `CapabilityStatus::Error` to `neither`, and there is a matrix test covering an error-state channel. I left the `const fn`/cleanup suggestions out to keep the policy PR narrowly scoped.

src/bridge/tool_surface.rs
```diff
@@ -0,0 +1,297 @@
...
+    match subject.kind {
+        SurfaceSubjectKind::LatentProviderAction
+        | SurfaceSubjectKind::AvailableNotInstalledProviderEntry
+        | SurfaceSubjectKind::Channel => SurfaceAssignment::capabilities_only(),
+        SurfaceSubjectKind::BuiltinDirectTool | SurfaceSubjectKind::ExtensionDirectAction => {
+            if is_direct_ready(subject.status) {
+                let _approval_gated = subject.approval_gated;
+                SurfaceAssignment::actions_only()
+            } else {
+                fallback_assignment(subject.status)
```
serrrfirat, line 83: **Medium Severity** · Confidence: Certain

**`approval_gated` field accepted but never consulted by policy**

`SurfacePolicyInput.approval_gated` is passed in by all callers but the policy function assigns it to `_approval_gated` (lines 85, 96) and discards it. This forces callers to compute and pass a value the policy ignores entirely.

The PR body says "approval at execution time does not remove them from `available_actions`" — so the current behavior is intentional. However, the field's existence signals that callers *should* care about computing it, which is misleading.

**Suggestion:** Either (a) remove the field from `SurfacePolicyInput` and add it when real approval logic lands, or (b) add a doc comment on the field explicitly stating it's reserved and currently ignored by `assign_surface`. Option (a) is cleaner — YAGNI until the logic is ready.

src/bridge/tool_surface.rs
```diff
@@ -0,0 +1,297 @@
...
+            }
+        }
+    }
+}
+
+const fn is_direct_ready(status: CapabilityStatus) -> bool {
+    matches!(status, CapabilityStatus::Ready)
+}
+
```
serrrfirat, line 101: **Medium Severity** · Confidence: Certain

**`ReadyScoped` for direct-mode subjects maps to `neither()` — zero visibility on both surfaces**

When a `BuiltinDirectTool` or `ExtensionDirectAction` has `ReadyScoped` status, it falls through to `fallback_assignment` where `ReadyScoped` maps to `SurfaceAssignment::neither()`. This means the subject appears on **no** surface at all — not in `available_actions` (can't call it) and not in `available_capabilities` (can't even see it).

This is tested ("ready-scoped extension direct action is not callable"), so it appears intentional. However, `ReadyScoped` is described in the engine as "usable now, but only through a scoped or indirect route." Mapping to `neither()` means the LLM has no awareness that the capability exists. If the intent is that scoped tools should at least appear in `available_capabilities` (like `NeedsAuth`, `NeedsSetup`, `Inactive`, and `Latent` all do), then this should be `capabilities_only()` instead.

**Suggestion:** If truly intentional, add a comment in `fallback_assignment` explaining why `ReadyScoped` gets `neither()` for direct-mode subjects (unlike every other non-ready status which gets `capabilities_only()`). If it should be visible, change the `ReadyScoped` arm to `capabilities_only()`.

src/bridge/tool_surface.rs
```diff
@@ -0,0 +1,257 @@
...
+const fn fallback_assignment(status: CapabilityStatus) -> SurfaceAssignment {
```
serrrfirat, line 102: Both findings addressed in c031a682:

**Fix 1 — `approval_gated` removed:** Field, `_approval_gated` bindings, and all test construction sites removed from `SurfacePolicyInput`. Will be re-added when the policy actually consults it.

**Fix 2 — `ReadyScoped` → `capabilities_only()`:** Moved `ReadyScoped` from the `neither()` arm to `capabilities_only()` in `fallback_assignment`, with a comment explaining the rationale. Test case updated to expect `capabilities_only()` and renamed to clarify the intent ("not callable but visible in capabilities").

serrrfirat (dismissed):
Paranoid architect re-review -- approved.

Both Medium findings from the initial review have been correctly addressed in c031a682:
- approval_gated field fully removed (YAGNI)
- ReadyScoped moved to capabilities_only() with rationale comment

Remaining findings are all Low/Nit (dead fallback arms for exhaustiveness, missing cross-category test combos, temporary allow(dead_code)). None are blocking.

Clean, well-tested, pure-function policy module with no side effects, no I/O, no security surface.

nickpismenkov (dismissed):

henrypark133:
Got all approvals just staged merged messed up the merge order. Doing it manually with override.

Issue #2392, opened 2026-04-13
Feature Epic: Host-level multi-account support for all messaging channels


hanakannzashi:
## Summary

IronClaw currently assumes that one installed channel name maps to one active account/config/runtime instance.

That assumption is starting to block real deployments. WeCom is the most obvious trigger right now, but this is not a WeCom-only problem. The same limitation affects Telegram, Discord, Feishu, Slack/relay-style channels, WhatsApp, and any future channel that may need more than one account in the same workspace.

This should be treated as a host/platform feature, not as a channel-local patch.

## Motivation

Today, if a user wants multiple accounts of the same channel type, they are pushed into bad tradeoffs:

- Run multiple IronClaw instances just to connect multiple accounts
- Reuse one bot/account for unrelated audiences, environments, or trust levels
- Avoid clean dev/staging/prod bot separation
- Accept weak identity namespacing in channels where sender IDs are only unique within a tenant/account scope

A proper multi-account model would unlock:

- Separate dev/staging/prod bots
- Separate bots/accounts for teams, subsidiaries, or customer-facing assistants
- Better permission boundaries between high-trust and low-trust audiences
- Cleaner identity namespacing for pairing, audit, and sender resolution
- Less pressure to solve host problems with channel-specific hacks

## Proposed direction

Introduce a first-class distinction between:

- Channel kind: `wecom`, `telegram`, `discord`
- Channel instance: e.g. `wecom_support`, `wecom_hr`, `telegram_ops`

The host should treat the channel instance as the runtime/config/secrets/pairing namespace, while channel kind still points to the implementation binary/capability schema.

Conceptually:

- one channel kind
- many installed instances
- each instance has isolated config, secrets, pairing state, runtime state, and routing identity

### Product / UX

- One IronClaw workspace can host multiple accounts of the same channel type
- Users can give each account a clear purpose and display identity
- Setup becomes closer to how real teams actually deploy bots

### Security / Isolation

- Reduce accidental cross-audience data leakage
- Support account-level policy differences without separate processes
- Improve identity stability when external sender IDs are only unique within a tenant/account scope

### Operations

- Simpler dev/staging/prod validation
- Easier migrations and cutovers between old/new bots
- Better observability because traffic can be attributed to a concrete account instance

### Architecture

- Removes pressure to encode account identity into channel-local workarounds
- Creates a reusable host abstraction that benefits all current and future channels

## Scope

Primary affected areas:

- Extension installation and naming
[...]

PR #2841, opened 2026-04-22
feat: multi-tenant channel instances control plane (phase 1)

serrrfirat:
## Summary

Phase 1 of #2392 — introduces a `channel_instances` DB table and dispatch-key routing in the shared channel/runtime control plane used by both v1 and v2, so each tenant owns their own channel instance with a unique dispatch key and same-kind multi-tenant coexistence becomes possible (e.g. two Telegram bots for two tenants).

This PR also includes the engine-specific plumbing needed on each side of that shared foundation:
- legacy session/thread isolation updates for v1
- instance-aware conversation/gate plumbing in the v2 bridge

## Problem

The production symptom was observed on multi-tenant v1 staging: web setup wrote Telegram secrets under the authenticated request user, but hot activation read/injected them under `ExtensionManager`'s `config.owner_id`, causing startup-time 404s when different tenants' Telegram bots tried to activate.

This PR fixes that at the shared tenant/channel-instance control plane rather than as a v1-only workaround, and carries the instance identity through both engine paths so same-kind tenant channels do not collide.

### Database
- New `channel_instances` table (PostgreSQL migration V25 + libSQL) with `user_id`, `channel_kind`, `instance_key`, `display_name`, `is_primary`, `enabled`, `config`, `metadata`, `last_error`
- `ChannelInstanceStore` trait added to `Database` supertrait with both backend implementations

### Dispatch Key Routing
- `Channel::dispatch_key()` trait method (defaults to `name()`)
- `IncomingMessage.channel_instance_key` for instance-scoped routing
- `ChannelManager` registers/routes by dispatch key instead of semantic name
- WASM channel wrapper carries and emits dispatch key

### Engine Coverage
- Shared activation/routing/pairing/startup changes live in `src/` channel/runtime code and apply to both v1 and v2
- Legacy agent session/thread resolution is now instance-aware so same-kind tenant channels do not collide on the v1 path
- Bridge router uses channel instance keys for v2 conversation and gate identity

### Instance-Aware State
- `PendingGate.source_channel_instance_key` for instance-scoped gate resolution
- `Thread.source_channel_instance_key` for instance-scoped thread resolution
- Bridge router uses channel instance key for engine conversation keys

### Tenant-Scoped Operations
- `complete_pairing_approval(channel_name, user_id, external_id)` resolves dispatch key from DB
- Heartbeat and routine notifications resolve per-user dispatch keys
- Extensions list derives active/error/pairing status per requesting user

### Webhook/Auth Isolation
- Instance-backed webhook paths use dispatch keys end to end
- WASM webhook secret/signature/HMAC validation now resolves auth material by dispatch key, so per-instance webhook auth is enforced correctly

### Startup
- Process-global restore from `channel_instances` rows; legacy `activated_channels` fallback when zero rows exist

### E2E Coverage
- Added a full-process multi-tenant Telegram scenario covering distinct dispatch keys, routing isolation, thread isolation, wrong-secret rejection, and per-user extension state
- Updated legacy Telegram E2Es to discover the active dispatch key from `/wasm-channels/health` instead of assuming a singleton `/webhook/telegram` path
- Added a dedicated Telegram E2E batch so token validation, hot activation, full Telegram round-trips, and multi-tenant isolation run together

## Testing
- Unit tests for restore, pairing, message routing, notification resolution, extensions list isolation
- `cargo check --all-features`
- `cargo test dispatch_key_webhook_secret --lib`
- `pytest -v scenarios/test_telegram_e2e.py scenarios/test_telegram_hot_activation.py -s --timeout=600`
- `pytest -v scenarios/test_telegram_token_validation.py scenarios/test_telegram_hot_activation.py scenarios/test_telegram_e2e.py scenarios/test_multi_tenant_channels.py -s --timeout=600` (23 passed)

Refs #2392

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -69,7 +69,9 @@ jobs:
           - group: features
             files: "tests/e2e/scenarios/test_skills.py tests/e2e/scenarios/test_tool_approval.py tests/e2e/scenarios/test_webhook.py"
           - group: extensions
-            files: "tests/e2e/scenarios/test_extensions.py tests/e2e/scenarios/test_extension_oauth.py tests/e2e/scenarios/test_oauth_url_parameters.py tests/e2e/scenarios/test_telegram_token_validation.py tests/e2e/scenarios/test_telegram_hot_activation.py tests/e2e/scenarios/test_wasm_lifecycle.py tests/e2e/scenarios/test_tool_execution.py tests/e2e/scenarios/test_agent_loop_recovery.py tests/e2e/scenarios/test_pairing.py tests/e2e/scenarios/test_mcp_auth_flow.py tests/e2e/scenarios/test_oauth_credential_fallback.py tests/e2e/scenarios/test_routine_oauth_credential_injection.py"
+            files: "tests/e2e/scenarios/test_extensions.py tests/e2e/scenarios/test_extension_oauth.py tests/e2e/scenarios/test_oauth_url_parameters.py tests/e2e/scenarios/test_wasm_lifecycle.py tests/e2e/scenarios/test_tool_execution.py tests/e2e/scenarios/test_agent_loop_recovery.py tests/e2e/scenarios/test_pairing.py tests/e2e/scenarios/test_mcp_auth_flow.py tests/e2e/scenarios/test_oauth_credential_fallback.py tests/e2e/scenarios/test_routine_oauth_credential_injection.py"
+          - group: telegram
+            files: "tests/e2e/scenarios/test_telegram_token_validation.py tests/e2e/scenarios/test_telegram_hot_activation.py tests/e2e/scenarios/test_telegram_e2e.py tests/e2e/scenarios/test_multi_tenant_channels.py"
           - group: routines
             files: "tests/e2e/scenarios/test_owner_scope.py tests/e2e/scenarios/test_routine_event_batch.py"
     steps:
--- a/migrations/V25__channel_instances.sql
+++ b/migrations/V25__channel_instances.sql
@@ -1,0 +1,25 @@
+CREATE TABLE channel_instances (
+    id UUID PRIMARY KEY,
+    user_id TEXT NOT NULL REFERENCES users(id) ON DELETE CASCADE,
+    channel_kind TEXT NOT NULL CHECK (channel_kind = lower(channel_kind)),
+    instance_key TEXT NOT NULL UNIQUE,
+    display_name TEXT NOT NULL,
+    is_primary BOOLEAN NOT NULL DEFAULT TRUE,
+    enabled BOOLEAN NOT NULL DEFAULT TRUE,
+    config JSONB NOT NULL DEFAULT '{}'::jsonb,
+    metadata JSONB NOT NULL DEFAULT '{}'::jsonb,
+    last_error TEXT,
+    created_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+    updated_at TIMESTAMPTZ NOT NULL DEFAULT NOW()
+);
+
+CREATE INDEX idx_channel_instances_user_kind
+    ON channel_instances (user_id, channel_kind);
+
+CREATE INDEX idx_channel_instances_enabled
+    ON channel_instances (enabled)
+    WHERE enabled = TRUE;
+
+CREATE UNIQUE INDEX idx_channel_instances_primary_per_kind
+    ON channel_instances (user_id, channel_kind)
+    WHERE is_primary = TRUE;
--- a/migrations/checksums.lock
+++ b/migrations/checksums.lock
@@ -37,3 +37,4 @@ V21__backfill_conversation_source_channel = 4041142068103384561
 V22__sandbox_restart_params = 12611649486554869350
 V23__list_workspace_files_escape_like = 12519024535161914473
 V24__llm_calls_created_at_index = 2650126284755685421
+V25__channel_instances = 11575638364465791573
--- a/src/agent/agent_loop.rs
+++ b/src/agent/agent_loop.rs
@@ -159,6 +159,16 @@ async fn resolve_channel_notification_user(
     }
 
     if let Some(channel_name) = trimmed_option(channel)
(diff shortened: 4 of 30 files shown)


serrrfirat:
Follow-up validation after the dispatch-key webhook/auth fixes:

- Updated the legacy Telegram E2E scenario to resolve the active Telegram dispatch key from `/wasm-channels/health` and post webhooks to `/webhook/{dispatch_key}` instead of the old singleton `/webhook/telegram` path.
- The multi-tenant Telegram scenario also exposed a real bug in host-side webhook auth lookup: secret/signature config was still being resolved by semantic channel name instead of dispatch key. That is now fixed.
- I also added a dedicated Telegram E2E batch so Telegram token validation, hot activation, full Telegram round-trips, and multi-tenant isolation run together.

Local verification run after those updates:

```bash
pytest -v scenarios/test_telegram_token_validation.py \
  scenarios/test_telegram_hot_activation.py \
  scenarios/test_telegram_e2e.py \
  scenarios/test_multi_tenant_channels.py \
  -s --timeout=600
```

Result: **23 passed**.

So the PR is now backed by a dispatch-key-aware Telegram E2E suite, not just the new multi-tenant scenario in isolation.

henrypark133 (dismissed):
What looks good:
- The dispatch-key plumbing is coherent across `ChannelManager`, `SessionManager`, the WASM router, and notification routing.
- The DB work is shaped correctly for the repo: shared trait first, then PostgreSQL and libSQL implementations plus migration.
- The new E2E scenario is aimed at the right failure mode: two-tenant activation, unique webhook paths, isolated routing, and wrong-secret rejection.

No verified findings.

Low-priority notes:
- `src/extensions/manager.rs:857` and `1114` still fall back to semantic channel names when dispatch-key lookup misses. That looks intentional for compatibility, but it is the main branch to watch if instance creation or restore ever drifts.
- I did not see a restart-oriented integration that proves two restored tenant instances still route webhook delivery and replies by dispatch key after process restart.

Summary:
- Prior feedback status: no prior review state fully audited; GitHub inline-thread state was not available reliably
- Review coverage: coordinator scan plus DB/app/web/multi-tenant reviewer passes in a dedicated worktree
- Residual risk: no live test execution in this pass, and restart coverage around restored instances still looks thinner than steady-state coverage

zmanian (commented):
Thorough review of Phase 1. Architecturally this is a well-scoped refactor: the `channel_instance_key` / `dispatch_key` split between semantic kind and concrete runtime instance is the right abstraction, and it's threaded consistently through `IncomingMessage`, `Channel`, `ChannelManager`, `WasmChannelRouter`, `SessionManager`, `PendingGate`, and the bridge `engine_conversation_key`. The dual-backend DB implementation (V25 PG migration + libSQL base + incremental) is symmetric, uses the shared `normalize_channel_name`, and the `idx_channel_instances_primary_per_kind` partial unique index correctly enforces one primary per `(user_id, kind)`. No `.unwrap()`/`.expect()` in added production code. Multi-tenant isolation tests (`test_multi_tenant_channels.py`, the per-tenant unit tests on `extensions_list_handler`, `resolve_thread_for_channel_instance`, `take_verified_for_channel_instance`, and `message_tool_owner_scoped_channel_uses_tenant_dispatch_key`) exercise the real call sites — this matches the Test-Through-the-Caller rule.

Concerns worth addressing, ideally in phase 1 or tracked as must-do follow-ups:

1. **Tenant enumeration via `/wasm-channels/health`.** `health_handler` returns `router.list_channels()` (all dispatch keys) with no auth. Pre-PR this leaked only generic kind names like `telegram`; post-PR it leaks `telegram:<uuid>` per tenant and — because `upsert_primary_channel_instance_record` builds `instance_key = "{kind}:{uuid}"` while some test fixtures use `"{kind}:{user_id}:primary"` — could leak user IDs if any caller uses the latter shape. Either gate the endpoint behind admin auth, restrict its output to a count / per-caller view, or guarantee instance keys are opaque (UUID-only suffix, which is the default today but not enforced by a type).

2. **Legacy `activated_channels` setting drift in multi-tenant mode.** `persist_active_channels(user_id)` snapshots the global `active_channel_names` HashSet and writes it into the *requesting* user's settings. With multiple tenants activating the same semantic kind, tenant A's settings end up listing kinds activated by tenant B. Harmless while the new `channel_instances` rows are authoritative at startup, but the fallback path (`restored_from_channel_instances == 0`) could resurrect cross-tenant state on a partially-migrated deployment. Consider snapshotting only the instances the acting user owns, or dropping the legacy persistence now that instances are the source of truth.

3. **`is_approval_authorized_for_instance` semantics change.** Previously `TRUSTED_APPROVAL_CHANNELS` was checked against `requesting`; the new code makes trusted channels bypass instance matching entirely, which is intended — but the old behavior also allowed same-name cross-channel approvals regardless of instance. The new instance-required branch (`Some(src) => match source_channel_instance_key ...`) is stricter: if `source_channel_instance_key` is None and `src == requesting`, behavior is unchanged; if Some and requesting has no instance key, it now denies. This is the safer default, but given the existing `Thread` rows persisted before this PR will have `source_channel_instance_key: None`, confirm there's no hydrated thread path that reloads old state and then receives a new-format request. I did not find a breaking case, but a migration/backfill note or test for legacy-thread-plus-new-message would be worth adding.

Minor nits:
- `engine_conversation_key` composes `{channel_key}:{scope_id}` — when `channel_instance_key` itself contains a `:`, the resulting key is unambiguous only because `conversation_manager` treats it opaquely. Documented in the helper comment would help future readers.
- `ExtensionError::Other(format!(\"Multiple '{}' channel instances exist for user '{}' ...\"))` is a data-integrity case that should probably be a typed variant, since the partial unique index should make it impossible — surfacing it as `Other` will make future debugging harder.
- `has_wasm_channel_owner_binding` (the no-user-arg variant) still exists and uses `self.user_id`; ideally deprecate and remove in a follow-up to avoid accidental single-tenant-assumption regressions.

No blockers from my side on the core design; I'd like concern (1) addressed before this lands on a public deployment, and (2) at minimum acknowledged as a tracked follow-up. Nice work on the E2E scenario — the cross-tenant-secret rejection and disjoint-thread-id assertions are exactly the right regression surface.

PR #2830, opened 2026-04-22
feat(engine): type code execution lifecycle events

serrrfirat:
Replace the remaining preview-string heuristics that drove code-execution UI status in the bridge/router with typed engine events emitted directly by the executor. Continues the semantic-typing work from #2581 / #2813 / #2822.

Stacked on **#2822** (typed MessageAdded events). Merge after that lands.

## What changed

**Engine**
- `EventKind::CodeExecutionStarted { step_id }` and `EventKind::CodeExecutionCompleted { step_id, had_output, duration_ms }` added alongside the existing `CodeExecutionFailed`.
- `handle_execute_code_step` now emits `CodeExecutionStarted` before the nested `execute_code()` call, and `CodeExecutionCompleted` on the success branch. `had_output` is derived from `!result.stdout.is_empty() || result.return_value != Null || any action result has non-empty output`, so the router can tell "Code executed" vs "Code executed (no output)" without parsing transcript previews.

**Bridge / router**
- `forward_event_to_channel` and `thread_event_to_app_events` now dispatch on the three typed lifecycle events:
  - `CodeExecutionStarted` -> `"Executing code..."`
  - `CodeExecutionCompleted { had_output: true }` -> `"Code executed"`
  - `CodeExecutionCompleted { had_output: false }` -> `"Code executed (no output)"`
  - `CodeExecutionFailed` -> `"Code error -- retrying..."`
- `interpret_message_event` no longer inspects User-role content previews — that surface is now owned by the typed events. Assistant-role mapping through `AssistantContentKind` is unchanged.

**Test plumbing**
- Exhaustive `event_kind_name` match in `tests/support/replay_outcome.rs` extended for the two new variants.

**Incidentally fixed pre-existing drift on the stacked chain**
- `tests/engine_v2_skill_codeact.rs`: `LlmResponse::Code { content: None }` -> `assistant_content: None` (the field was renamed in #2815; the test file wasn't updated).
- fmt drift in `src/bridge/llm_adapter.rs` and `src/channels/web/tests/reasoning_redaction.rs`.

## Why this matters

After #2822, the only preview-string heuristic left in the router was the code-execution status mapping. Making the executor emit typed events instead of the router reverse-engineering user-message content previews means:

- The SSE / REPL status layer no longer depends on brittle string patterns produced by the Python orchestrator.
- `CodeExecutionCompleted.duration_ms` is available to observers for free.
[...]

PR #2842, opened 2026-04-22
feat(engine): type MessageAdded payload (id + optional narrative)

serrrfirat:
## What

Replaces `EventKind::MessageAdded::content_preview: String` — a lossy, 80-char-truncated preview — with a typed payload that the router and UI can consume without reverse-engineering content semantics from a string:

```rust
MessageAdded {
    role: String,
    message_id: MessageId,                              // resolvable via Thread::message(id)
    assistant_content_kind: Option<AssistantContentKind>,
    narrative: Option<String>,                          // only populated for UserVisibleNarrative
}
```

## Why this shape

This was the design decision covered by mock `/tmp/message-added-ui-preview.html` (4-column comparison in our gateway chat UI). Only `UserVisibleNarrative` assistant content actually surfaces text to the status stream — every other kind is either:
- suppressed at the router (`InternalReasoning`)
- handled by the parallel `Response`/`ThreadOutcome` stream (`Final`)
- not a status-surface concern at all (user, action result)

So carrying full inline text for every kind would inflate SSE frames without changing what the user sees. At the same time, ID-only would force a REST round-trip just to render the narrative text we already have. This shape carries the one piece of text the UI needs, and nothing more.

The `message_id` is the durable handle for anything we later want to look up lazily (full typed content, action calls, provenance) via `Thread::message(id)`.

## Changes

- `crates/ironclaw_engine/src/types/message.rs` — new `MessageId` type (UUID-backed, matches `ThreadId` / `StepId` pattern) + `id` field on `ThreadMessage`; all 6 `ThreadMessage::*` constructors set it.
- `crates/ironclaw_engine/src/types/event.rs` — new `MessageAdded` shape with doc comment explaining the `narrative` population rule.
- `crates/ironclaw_engine/src/types/thread.rs` — `Thread::add_message` extracts narrative only for `AssistantContent::UserVisibleNarrative`; new `Thread::message(id) -> Option<&ThreadMessage>` resolver.
- `src/bridge/router.rs` — `interpret_message_event` takes `narrative: Option<&str>` instead of `content_preview: &str`; both callers updated; defensive branch suppresses narrative events that arrive without inline text.
- Test updates in engine and router + 3 new tests:
  - `add_assistant_reasoning_message_omits_narrative` (engine)
  - `thread_message_resolves_by_id` (engine)
  - `thread_event_to_app_events_narrative_without_text_is_suppressed` (router)

## Backwards compatibility

Breaking shape change on a runtime event. No on-disk migration needed — `ThreadEvent`s are replayed in-memory per process; persisted `Thread`s only store messages + state, not event enum variants. `ThreadMessage` gains an `id` field with `#[serde(default)]` so legacy persisted rows deserialize with a fresh id (the value is not reached back for after initial creation).

## Test results

- `cargo test -p ironclaw_engine --lib` → **468 passed**, 0 failed (+2 new)
- `cargo test --lib bridge::router` → **75 passed**, 0 failed (+1 new)
- `cargo fmt --all -- --check` clean
- `cargo clippy --all --tests --examples --all-features` clean

## Stacked

- Base: `feat/2823-typed-code-execution-events` (PR #2830)
- Parent stack: #2815 → #2822 → #2830 → **this**
- Follow-up planned: push `AssistantContent` through `LlmProvider` trait boundary so adapters don't have to re-infer kind.

Part of #2813.

### DIFF
--- a/crates/ironclaw_engine/src/executor/trace.rs
+++ b/crates/ironclaw_engine/src/executor/trace.rs
@@ -448,6 +448,7 @@ mod tests {
         thread.add_message(ThreadMessage::assistant("calling tool"));
         // Manually construct a message with None call_id
         thread.add_message(ThreadMessage {
+            id: crate::types::message::MessageId::new(),
             role: crate::types::message::MessageRole::ActionResult,
             content: "result".into(),
             assistant_content: None,
--- a/crates/ironclaw_engine/src/lib.rs
+++ b/crates/ironclaw_engine/src/lib.rs
@@ -42,7 +42,7 @@ pub use types::capability::{
 pub use types::error::{CapabilityError, EngineError, StepError, ThreadError};
 pub use types::event::{AssistantContentKind, EventId, EventKind, ThreadEvent};
 pub use types::memory::{DocId, DocType, MemoryDoc};
-pub use types::message::{MessageRole, ThreadMessage};
+pub use types::message::{MessageId, MessageRole, ThreadMessage};
 pub use types::mission::{Mission, MissionCadence, MissionId, MissionStatus, ValidTimezone};
 pub use types::project::{Project, ProjectId, ProjectMetric};
 pub use types::provenance::Provenance;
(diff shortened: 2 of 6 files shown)


serrrfirat:

PR #2846, opened 2026-04-22
feat(engine): typed assistant content rollout (closes #2813)

### LINKED ISSUE #2813: engine-v2: add typed assistant content model for final vs internal tool-use text

serrrfirat:
## Summary

Consolidated rollout of **typed assistant content** across the engine → bridge → web gateway boundary.

Supersedes the `#2813` stacked PR chain: #2812, #2815, #2822, #2830, #2842. Those PRs were split by chapter to make each layer reviewable in isolation, but reviewers consistently asked to see the whole story in one place — the behavior only makes sense end-to-end, and the five diffs together are ~1.1k lines across ~30 files on a single theme.

Issue: #2813

## Why

The pre-2813 engine emitted a single `AssistantContent::Text` blob and a string `content_preview` on `MessageAdded` events. Downstream consumers (bridge router, web gateway, REPL, CLI) were forced to guess what the model was doing:

- "Was that message an actual user-facing reply?"
- "Was it the model's internal reasoning that leaked into transcript?"
- "Was it a CodeAct narrative (`Inspecting the page first...`) that we *do* want in the UI?"

Every answer depended on substring heuristics (`"[stdout]"`, `"Traceback"`, `"Executing code..."`), scattered across layers, quietly diverging, and producing bugs like the `#2581` reasoning-leak on the web chat.

## What changes

Five chapters, cherry-picked from the stacked chain:

### 1. `fix(web): hide internal reasoning in browser chat` (was #2812)
The original quickfix that hid reasoning on the web surface. Kept here so the rollout doesn't revert it; the later chapters replace the heuristic filter with typed checks.

### 2. `feat(engine): add typed assistant content model` (was #2815)
New `AssistantContent::{Final, UserVisibleNarrative, InternalReasoning}` + `AssistantContentKind`. LLM adapters choose the variant based on provider signals (Anthropic thinking blocks, OpenAI reasoning_content). `ThreadMessage::assistant_content` is now typed end-to-end. `LlmResponse::Code` gains `assistant_content` (renamed from `content`).

### 3. `feat(engine): type message-added assistant events` (was #2822)
`EventKind::MessageAdded` gains `assistant_content_kind: Option<AssistantContentKind>`. The bridge router's `interpret_message_event` now dispatches on the kind instead of string-sniffing `content_preview`. Internal reasoning → no UI event; user-visible narrative → `Thinking`; final assistant text → owned by the existing response stream.

### 4. `feat(engine): type code execution lifecycle events` (was #2830)
New `EventKind::CodeExecutionStarted { step_id }` and `CodeExecutionCompleted { step_id, had_output, duration_ms }` alongside the existing `CodeExecutionFailed`. Executor emits them directly around `execute_code()`. The bridge stops matching `"[stdout]"` / `"[code ..."` / `"Error"` / `"Traceback"` on user-role previews — typed events own that surface now. `CodeExecutionFailed → AppEvent::CodeExecutionFailed` continues to use the typed form merged in #2797.

### 5. `feat(engine): type MessageAdded payload (id + optional narrative)` (was #2842)
`EventKind::MessageAdded` drops `content_preview: String` in favor of `message_id: MessageId` + `narrative: Option<String>`. Narrative is populated **only** for `UserVisibleNarrative` (the single case that actually needs a preview in the event stream). `MessageId` is now a first-class UUID-backed type on `ThreadMessage`, with `Thread::message(id)` for resolution. Everything else lives on the thread record, not the event.

## Risk

- **Engine event schema** changes — `MessageAdded` shape is not back-compat on the wire. Historical events in Postgres deserialize via `serde(default)` on the new fields.
- **Bridge heuristic removal** — any external tooling that watched `[stdout]` / `Traceback` previews on the status stream must switch to the typed lifecycle events (or the typed `AppEvent::CodeExecutionFailed`).
- **LLM adapter attribution** — `AssistantContentKind` mapping lives in the bridge adapters; a new provider must opt in or all its output falls into `Final`.

Rollback: revert the five commits in reverse; each layer keeps the one below compiling on its own.

## Verification

Local quality gate on darwin/aarch64:

```
cargo fmt --all -- --check          ✅
cargo clippy --all --benches --tests --examples --all-features  ✅ (zero warnings)
cargo check --all-targets --all-features                        ✅
cargo test -p ironclaw_engine --lib                             ✅ 468 passed
cargo test --lib --all-features bridge::router                  ✅ 78 passed
cargo test --all-features --test engine_v2_skill_codeact        ✅ 3 passed
cargo test --all-features --test e2e_engine_v2
            --test engine_v2_gate_integration                   ✅ 26 passed
```

One pre-existing unrelated flake: `extensions::manager::tests::inject_mcp_client_partitions_cache_by_user` (same on pristine staging; zero lines touched in `src/extensions/` in this PR).

New tests worth naming:

- `add_assistant_reasoning_message_omits_narrative`
- `thread_message_resolves_by_id`
- `thread_event_to_app_events_skips_internal_reasoning_message_added`
- `thread_event_to_app_events_surfaces_user_visible_narrative_preview`
- `thread_event_to_app_events_final_assistant_message_relies_on_response_stream`
- `thread_event_to_app_events_legacy_assistant_message_keeps_execution_status`
- `thread_event_to_app_events_maps_code_execution_started_and_completed`
- `thread_event_to_app_events_narrative_without_text_is_suppressed`
- `forward_event_to_channel_maps_code_execution_lifecycle_events`


henrypark133 (changes requested):
What looks good:
- The `AssistantContent` / `MessageId` split is a real contract improvement over stringly-typed assistant text.
- The web-facing redaction direction is correct: internal reasoning is being pushed out of normal browser history and SSE surfaces.
- The router tests around `AssistantContentKind` and code-execution status mapping are targeted and useful.

Findings:
1. Medium - `crates/ironclaw_engine/src/executor/orchestrator.rs:2614`: `json_to_thread_messages()` now calls `assistant_with_action_content(assistant_content, calls)` whenever `action_calls` exist. If a legacy persisted/bootstrap payload has assistant `content` plus `action_calls` but no `assistant_content`, that path round-trips to an empty assistant message and drops the legacy narrative text entirely. Why it matters: replay/bootstrap/follow-up context can silently lose assistant text for older histories. Expected fix direction: preserve raw `content` when `assistant_content` is absent on the action-call path instead of defaulting to empty text.

Low-priority notes:
- The new tests are mostly helper-level. I did not see one caller-level integration that drives typed assistant content from adapter output through router/web behavior in the same thread.

Summary:
- Prior feedback status: no prior review state fully audited; `gh` diff/thread fetches were flaky during this pass
- Review coverage: coordinator scan plus engine/web/testing leaf reviews in a dedicated worktree
- Residual risk: browser WS/response-stream behavior and inline GitHub thread state were not live-verified

serrrfirat:
Addressed all review feedback in 795f6ed1:

- **`had_output` fix** — structured JSON/boolean action outputs now correctly counted as real output (was only treating non-empty strings as output)
- **Legacy assistant content** — `json_to_thread_messages` now preserves the raw `content` field when `action_calls` exist but `assistant_content` is absent
- **Caller-level test** — added `handle_with_engine_surfaces_typed_assistant_narrative_through_status_and_sse` in `router.rs` to cover the full SSE path for typed narratives
- **`MessageId`** — added `#[serde(transparent)]`

henrypark133 (dismissed):
What looks good:
- The typed `AssistantContent` and `MessageId` contract is a real improvement over the old string-sniffing behavior.
- The legacy bootstrap/action-call round-trip now preserves old assistant text when typed content is absent.
- The new caller-level router test covers typed narrative flow all the way through status and SSE, which is the right regression boundary here.

No verified findings.

Summary:
- Prior feedback status: fully addressed
- Review coverage: checked-out worktree plus targeted engine/router follow-up on the previously reported legacy-content regression
- Residual risk: I did not re-audit every provider-specific adapter branch in this pass

PR #2850, opened 2026-04-22
feat(debug-panel): expand Activity tab coverage with CodeAct + warnings

ilblackdragon:
## Summary
- Emit `AppEvent::CodeExecuted` (verbose-only) and `AppEvent::Warning` so the debug panel's Activity tab can surface raw CodeAct code/stdout and WARN/ERROR logs that were previously only visible in server stdout.
- Backfill `params_summary` on `ActionExecuted`/`ActionFailed` events from the structured and scripting executors so tool args are visible immediately, even on success. `summarize_params` hardens the `http`/`web_fetch` / `shell` branches and the generic MCP fallback so query-string secrets, signed-URL tokens, inline `Authorization:` headers, and sensitively-keyed MCP fields are stripped before they reach `ToolCompleted.parameters`.
- Redact `AppEvent::CodeExecuted` code / stdout / return_value through `LeakDetector` at the bridge boundary before SSE broadcast, so a model-authored Python snippet that prints a bearer token or returns a credential-shaped JSON blob does not reach verbose subscribers verbatim. Covers both `LeakAction::Redact` and `LeakAction::Block` patterns (the latter were previously surfaced by `scan()` without redacting).
- Byte-bound CodeExecuted payloads: `code` / `stdout` tail-truncated to 8 KiB via a UTF-8-boundary-safe helper; `return_value` strings tail-truncated, structured return_values <= 8 KiB preserved and larger ones dropped to `None` (truncated JSON would be unparseable on the frontend).
- Wire the debug panel JS to render 20+ additional event types (code_executed, warning, gate lifecycle, plan updates, thread/child/mission spawning, onboarding, suggestions, sandbox jobs) through a shared `on()` wrapper that dedupes ~25 copies of the JSON-parse + reconnect-counter housekeeping.
- i18n strings (en/ko/zh-CN) and CSS icon colors for the new activity types.
- **Multi-tenant safety:** the warning bridge is **disabled** when `multi_tenant_mode` is `true`. The `tracing` layer captures log context at the global subscriber scope, not at request scope, so scoping the whole bridge to a single `owner_id` in multi-tenant mode would deliver tenant A's warnings to the admin/owner account instead of tenant A. Re-enabling the bridge in multi-tenant deployments is gated on per-request provenance being threaded through every `warn!` / `error!` call site — out of scope for this PR. Single-user gateways get the bridge unchanged.

## Test plan
- [x] `cargo fmt --check`
- [x] `cargo clippy --all --benches --tests --examples --all-features` — clean
- [x] `cargo test --lib` — 5474 pass; one pre-existing MCP flake (`extensions::manager::tests::inject_mcp_client_partitions_cache_by_user`) reproducible on clean `origin/staging`, unrelated to this PR
- [ ] Open the web debug panel, trigger a CodeAct run, a tool call, and an approval gate; verify each shows up in the Activity tab with parameters visible and no leaked secrets
- [ ] Emit a `tracing::warn!` from a non-verbose code path; confirm it lands in the Activity tab only for verbose subscribers in single-user mode, and does not appear at all when `multi_tenant_mode` is on
- [ ] Confirm a CodeAct snippet that prints a token or returns a large JSON blob renders with `[REDACTED]` / dropped return value rather than verbatim secrets or bloated payloads


crates/ironclaw_engine/src/executor/scripting.rs
```diff
@@ -1161,7 +1161,7 @@ async fn preflight_action(
                     call_id: call_id.into(),
                     error: reason.clone(),
                     duration_ms: 0,
-                    params_summary: None,
+                    params_summary: crate::types::event::summarize_params(action_name, params),
```
serrrfirat, line 1164: High: this new `params_summary` path looks like it can bypass our normal sensitive-param redaction. `summarize_params()` falls back to the first string value for unknown tools, and this PR now threads that summary into `ActionFailed/ActionExecuted` events that reach SSE/debug-panel. For MCP or built-in tools whose first string field is something like `token`, `api_key`, or `password`, we would now surface up to 50 chars here. Could we redact before summarizing (using `tool.sensitive_params()`) or restrict the fallback to an explicit allow-list of safe keys instead of `obj.values().find_map(...)`?

crates/ironclaw_engine/src/executor/orchestrator.rs
```diff
@@ -808,6 +808,31 @@ async fn handle_execute_code_step(
...
+            let return_value = match &result.return_value {
+                serde_json::Value::Null => None,
+                other => Some(other.clone()),
+            };
+            let code_executed_event = ThreadEvent::new(
+                thread.id,
+                EventKind::CodeExecuted {
+                    step_id: exec_ctx.step_id,
+                    code: code.clone(),
+                    stdout: result.stdout.clone(),
```
serrrfirat: Medium: this clones the full `stdout` payload into `ThreadEvent::CodeExecuted` on every CodeAct run, even when there are no verbose/debug subscribers. Because there is no truncation/cap here, a step that prints a large blob can now bloat stored thread events and SSE payloads significantly. Was that retention cost intentional? I think this should either be capped before persistence or gated behind a verbose-only check so debug instrumentation does not become an unbounded storage/memory path.

crates/ironclaw_gateway/static/debug-panel.js
```diff
@@ -395,6 +429,166 @@
...
+    });
+
+    on('gate_resolved', function (data) {
+      var body = [
+        data.gate_name ? 'gate: ' + data.gate_name : '',
+        data.tool_name ? 'tool: ' + data.tool_name : '',
+        data.resolution ? 'resolution: ' + data.resolution : '',
+        data.message || '',
+      ].filter(Boolean).join('\n');
+      var status = STATUS_TO_ACTIVITY[data.resolution] || 'success';
```
serrrfirat: Medium: defaulting unknown `gate_resolved` values to `'success'` looks wrong. The backend already emits `resolution: "expired"` when the target thread disappears during resume (`src/bridge/router.rs`), so that failure path will render as a green success badge in Activity. Could we mirror the neighboring handlers and use `|| null` (or an explicit failure mapping for `expired`) instead of treating every unmapped resolution as success?

crates/ironclaw_engine/src/executor/scripting.rs
ilblackdragon: Fixed in aae97be7. `summarize_params`'s generic fallback now filters out sensitive-looking keys (token/secret/password/api_key/auth/credential/bearer + `_key` suffix / standalone `key`) before picking the first string value, so MCP or unknown-tool calls with a secret as the leading field no longer surface it into `ActionExecuted`/`ActionFailed` events or debug-panel SSE. Went with a denylist rather than a hard allow-list so legitimate unknown-tool params still render a summary; the engine crate can't reach the host's `Tool::sensitive_params()` directly. Added two regression tests in `types::event::tests`.

crates/ironclaw_engine/src/executor/orchestrator.rs
ilblackdragon: Fixed in aae97be7. `code` and `stdout` are now tail-truncated to 8_000 chars (reusing the `tail_chars` helper already used by `CodeExecutionFailed`'s error text, and matching `scripting::OUTPUT_TRUNCATE_LEN`) before the event is constructed, so persisted thread events and SSE payloads stay bounded regardless of subscriber count. The broadcast path is *also* gated on `has_verbose_receivers()` now (see the Copilot thread on router.rs) so non-debug tabs don't see the event at all.

crates/ironclaw_gateway/static/debug-panel.js
ilblackdragon: Fixed in aae97be7. Added a dedicated `GATE_RESOLUTION_STATUS` map with explicit entries for the six known resolutions (`approved`/`credential_provided`/`external_callback` → success, `denied`/`cancelled`/`expired` → failure) and a fallback to `null`, so `expired` now renders as a failure and any future unknown resolution renders neutral instead of green. Kept `STATUS_TO_ACTIVITY` for jobs/plans/onboarding where `success` is a sensible default for `completed`/`ready`.

src/channels/web/log_layer.rs
```diff
@@ -298,6 +301,51 @@ impl WebLogLayer {
...
+                    }
+                    let event = AppEvent::Warning {
+                        source: entry.target,
+                        message: entry.message,
+                        thread_id: None,
+                    };
+                    match &owner_id {
+                        Some(uid) => sse.broadcast_for_user(uid, event),
+                        None => sse.broadcast(event),
+                    }
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in aae97be7. `spawn_warning_bridge` now short-circuits on `sse.has_verbose_receivers()` before constructing the `AppEvent`, so when no debug/admin subscriber is connected the warning is dropped instead of flowing through the shared SSE broadcast channel.

src/bridge/router.rs
```diff
@@ -4583,6 +4583,19 @@ fn thread_event_to_app_events(
...
+            return_value,
+            duration_ms,
+            ..
+        } => vec![AppEvent::CodeExecuted {
+            code: code.clone(),
+            stdout: stdout.clone(),
+            return_value: return_value.clone(),
+            duration_ms: *duration_ms,
+            thread_id: Some(thread_id.into()),
+        }],
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in aae97be7. Added the gate inside `await_thread_outcome` — after `thread_event_to_app_events` returns, any `AppEvent` whose `is_verbose_only()` is true is skipped when `sse.has_verbose_receivers()` is false. Mirrors the existing `send_status` short-circuit. This covers `CodeExecuted`, `Warning`, and any future verbose-only variant without per-event plumbing. (And `CodeExecuted`'s payload is now capped at 8_000 chars before the clone — see the other thread on orchestrator.rs.)

crates/ironclaw_gateway/static/debug-panel.js
```diff
@@ -236,7 +236,42 @@
...
+    // JSON-parse + error-swallow + reconnect-counter housekeeping. Before
+    // extraction the block was copy-pasted ~25 times across this file and
+    // the `lastEventTime` / `totalEventsReceived` bookkeeping was the
+    // most likely drift target.
+    function on(name, handler) {
+      es.addEventListener(name, function (e) {
+        try { handler(JSON.parse(e.data)); } catch (_) { /* ignore */ }
+        lastEventTime = Date.now(); totalEventsReceived++;
+      });
+    }
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in aae97be7. Migrated the remaining JSON-parsing listeners — `status`, `thinking`, `tool_started`, `tool_completed`, `tool_result`, `turn_cost`, `response`, `error`, `turn_metrics`, `tool_result_full` — to use the `on()` wrapper. Left `stream_chunk` on the raw `addEventListener` path since it doesn't parse JSON, with a short comment explaining why.

crates/ironclaw_gateway/static/debug-panel.js
```diff
@@ -395,6 +429,166 @@
...
+
+    on('gate_required', function (data) {
+      var body = [
+        data.gate_name ? 'gate: ' + data.gate_name : '',
+        data.tool_name ? 'tool: ' + data.tool_name : '',
+        data.description || '',
+        data.parameters ? 'params: ' + data.parameters : '',
+      ].filter(Boolean).join('\n');
+      addActivity('gate', t('debug.activityGateRequired'), timeNow(), 'pending', body, { labelKey: 'debug.activityGateRequired' });
+    });
```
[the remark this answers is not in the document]
ilblackdragon: Not changing — this looks like a false positive. The backend wire contract for both `gate_required` and `approval_needed` types `parameters` as a `String` (see `crates/ironclaw_common/src/event.rs:267` for `ApprovalNeeded` and `:298` for `GateRequired`), so the UI can safely string-concat. The `data.input` field on `job_tool_use` is genuinely an object (it's the Claude-bridge tool arguments), which is why that handler uses `JSON.stringify` — different wire shape.

src/channels/web/mod.rs
```diff
@@ -647,6 +647,21 @@ impl Channel for GatewayChannel {
                 ),
             })?;
 
+        // In multi-tenant deployments, scope warnings to the gateway
+        // owner so per-request log context (thread ids, paths, emails)
+        // doesn't bleed across tenants via admin debug panels.
+        if let Some(log_broadcaster) = self.state.log_broadcaster.as_ref() {
+            let owner_scope = self
```
serrrfirat: **High Severity**

This scopes the warning bridge once at gateway startup to the static gateway `owner_id`, not to the user who triggered the warning-producing request. In multi-tenant mode that means WARN/ERROR log lines from tenant A's request can be delivered to the owner/admin account instead of tenant A, while the originating tenant never sees the warning in their own debug panel.

That avoids broadcasting warnings to everyone, but it still misroutes per-request diagnostics across accounts. I think this needs request/user-aware scoping at the point where the warning event is emitted (or the warning bridge should be disabled for multi-tenant mode until that provenance exists), rather than binding everything to the owner up front.

src/bridge/router.rs
```diff
@@ -4514,7 +4525,7 @@ fn thread_event_to_app_events(
                     name: display_name,
                     success: true,
                     error: None,
-                    parameters: None,
+                    parameters: params_summary.clone(),
```
serrrfirat, line 4738: **High Severity**

Routing `params_summary` into `ToolCompleted.parameters` makes the existing summary policy user-visible on the debug SSE path, and that policy still returns raw `http`/`web_fetch` URLs and raw `shell` commands. That means a successful tool call can now expose query-string secrets, signed URLs, or inline auth headers (for example `curl -H "Authorization: Bearer ..."`) directly in the debug panel.

I don't think the generic MCP-key filtering added in this PR is enough here, because these known-tool branches bypass it entirely. Please either redact these summaries before broadcasting them (for example with the leak detector / safety layer) or make the per-tool summaries strip sensitive parts such as URL query strings and auth-bearing shell arguments.

crates/ironclaw_engine/src/executor/orchestrator.rs
```diff
@@ -808,6 +808,37 @@ async fn handle_execute_code_step(
...
+            // Cap code and stdout at `CODE_EXECUTED_MAX_CHARS` each before
+            // emission so a step that prints a large blob cannot bloat
+            // persisted thread events or flood the SSE broadcast buffer.
+            // Same cap as `scripting::OUTPUT_TRUNCATE_LEN` (8_000).
+            const CODE_EXECUTED_MAX_CHARS: usize = 8_000;
+            let return_value = match &result.return_value {
+                serde_json::Value::Null => None,
+                other => Some(other.clone()),
+            };
+            let code_executed_event = ThreadEvent::new(
```
serrrfirat, line 915: **Medium Severity**

`CodeExecuted` now persists and broadcasts raw CodeAct `code`, `stdout`, and `return_value`, but unlike normal external tool output this path never goes through the safety/leak-detection layer before reaching SSE. A snippet that prints a secret, echoes sensitive tool output, or returns a credential-shaped JSON blob will therefore surface that raw material to every verbose subscriber and keep it in the thread event log as well.

Given the IronClaw convention that external output should cross a safety boundary before leaving the runtime, I think this event either needs redaction/scrubbing before emission or a narrower payload that can't carry arbitrary sensitive output verbatim.

src/channels/web/mod.rs
ilblackdragon: Fixed in 1d499f88. Disabled the warning bridge entirely when `multi_tenant_mode` is `true` — matching the second option you suggested ("disable for multi-tenant mode until provenance exists"). The `tracing` layer captures log context at the global subscriber scope, so there's no way to scope warnings per-request today without threading tenant provenance through every `warn!`/`error!` call site, which is well outside this PR. Single-user gateways still get the bridge with no `owner_scope` since there's only one consumer.

src/bridge/router.rs
ilblackdragon: Fixed in 1d499f88. Tightened `summarize_params` per-tool branches in `crates/ironclaw_engine/src/types/event.rs`:
- `http` / `web_fetch`: `strip_url_sensitive_parts` drops the query string, fragment, and `user[:pass]@` userinfo before truncation, so signed URLs / query-string API keys / embedded credentials no longer reach `ToolCompleted.parameters`.
- `shell`: `redact_shell_command_for_display` masks values of `-H`/`--header`/`-u`/`--user`/`--token`/`--api-key`/`--password`/`--auth`/`--bearer` (quoted or unquoted), masks `Authorization:` / `X-Api-Key:` / `X-Auth-Token:` / `Bearer:` values inside any argument, and strips query strings from any URL embedded in the command.

Kept this in the engine crate (no new dependency on `ironclaw_safety`) since the leak shape is specific to how these known-tool summaries are constructed. Regression tests: `summarize_params_http_strips_query_string_and_userinfo`, `summarize_params_http_preserves_non_url_strings`, `summarize_params_shell_redacts_auth_headers_and_query_strings`. The generic MCP-key filter from aae97be7 still covers the unknown-tool path.

crates/ironclaw_engine/src/executor/orchestrator.rs
ilblackdragon: Fixed in 1d499f88. `AppEvent::CodeExecuted` now passes through the leak detector at the bridge boundary before reaching SSE. Added `SafetyLayer::leak_detector()` and `EffectBridgeAdapter::safety()` accessors, then `redact_code_executed_secrets` in `src/bridge/router.rs::await_thread_outcome` scrubs `code`, `stdout`, and `return_value` (recursively for nested JSON) via a new `redact_leaks_in_string` helper.

One important subtlety: `LeakDetector::scan_and_clean` / `scan().redacted_content` only redact `Redact`-action matches, leaving `Block`-action matches (OpenAI/Anthropic/GitHub/AWS/Stripe/NEAR patterns — the majority of the default set) untouched. For a verbose-only observability event, keeping any leak-carrying field verbatim is never appropriate, so the helper iterates `scan().matches` directly and redacts every match range regardless of action. Regression test `redact_code_executed_scrubs_secrets_from_code_stdout_and_return_value` covers the code / stdout / nested-JSON-return-value cases with a GitHub-token pattern. Keeping the engine crate safety-agnostic (no `ironclaw_safety` dep) — scrubbing stays at the adapter boundary where the rest of the safety pipeline already lives.

ilblackdragon:
## Code Review — PR #2850 `feat(debug-panel): expand Activity tab coverage with CodeAct + warnings`

**Scope** (4 branch commits on top of `origin/staging`, +1074/−160 across 16 files): surfaces raw CodeAct `code`/`stdout` and WARN/ERROR log lines in the debug panel's Activity tab, backfills `params_summary` on engine action events, and refactors ~25 SSE listeners in `debug-panel.js` through a shared `on()` wrapper.

**State:** `MERGEABLE` but `BLOCKED` (likely required checks, not conflicts). 3 review threads resolved in `aae97be7`, 3 follow-up threads resolved in `1d499f88`.

### Overview
- New wire events `AppEvent::CodeExecuted` + `AppEvent::Warning` (both `is_verbose_only`) plus engine `EventKind::CodeExecuted`.
- Orchestrator emits `CodeExecuted` with code/stdout tail-truncated to 8k chars (matches `scripting::OUTPUT_TRUNCATE_LEN`).
- Structured + scripting executors populate `params_summary` on success and failure paths.
- Bridge router gates verbose events on `has_verbose_receivers()` before broadcast and pipes `CodeExecuted` payloads through `LeakDetector` at the bridge boundary.
- `summarize_params` hardens `http`/`web_fetch`/`shell` branches against URL-query/userinfo/auth-flag leakage, plus denylists sensitive keys on the generic fallback.
- Frontend adopts shared `on()` wrapper, adds `GATE_RESOLUTION_STATUS` table, new handlers for code_executed/warning/lease/plan/mission/onboarding/etc.
- Warning bridge is disabled in multi-tenant mode (pragma until per-request tenant provenance exists).

### Correctness
- **`CodeExecuted` redaction at the right layer** — engine crate stays safety-agnostic; scrubbing lives in `src/bridge/router.rs::redact_code_executed_secrets` via `SafetyLayer::leak_detector()` (new accessor) + `EffectBridgeAdapter::safety()`. Good layering.
- **`redact_leaks_in_string` covers all actions.** Important subtlety is correct: `scan().redacted_content` is populated only for `LeakAction::Redact`, leaving `Block` patterns (OpenAI/Anthropic/GitHub/AWS/Stripe — the majority of defaults) unredacted. Iterating `scan().matches` directly fixes that. The overlap-skip (`range.start < cursor`) is sound.
- **`strip_url_sensitive_parts` is `str::find`-based, not `url`-crate-based.** Conservative and safe: non-URL strings (`"not-really-a-url"`) pass through unchanged. Handles `?` before `#`, strips userinfo via `authority.rfind('@')`. Edge case: if `?` comes before the first `/` (e.g. `https://host?x=1`), the code splits `end` on `/` and treats the whole `host?…` as authority — which then hits `rfind('@')` cleanly; still correct because there is no `@`. ✔
- **`redact_shell_command_for_display` uses 4 ordered regexes on a `OnceLock`.** `.unwrap()` on literal patterns with `// safety: hardcoded literal` matches the existing codebase convention (see `leak_detector.rs` default patterns).
- **`EventKind::CodeExecuted` already tail-truncated to 8k upstream** (`orchestrator.rs`), so the SSE event bounds are enforced regardless of subscriber count. ✔
- **`is_sensitive_param_key` denylist is substring-based.** `"refresh_key"` matches via `ends_with("_key")`; `"token"` matches `"refresh_token"` via `contains("token")`. ✔
- **`params_summary` backfill is symmetric across success/failure/preflight/rejected** in structured and scripting — so the debug panel shows args on every event, not only some.

### Potential issues
- **Frontend JSON-parse silence on `on()` wrapper.** The new wrapper swallows JSON parse errors with `/* ignore */` but still bumps `totalEventsReceived`. That's intentional (matches pre-refactor behavior) but will quietly count malformed frames. Low-severity; no action unless you want parse errors surfaced in the stats panel.
- **`redact_secrets_in_json` recurses into object *values* but never inspects keys.** A return value like `{"password": "hunter2"}` is *not* redacted — the leak-detector only catches pattern-shaped secrets (bearer tokens, known API keys), not arbitrary string values under sensitive-named keys. Worth flagging in follow-up: consider extending the JSON walk to apply `is_sensitive_param_key` to keys as well.
- **`EventKind::CodeExecuted.code` is tail-truncated, not head-truncated.** A long secret-bearing preamble (`token = 'xxx…'` at line 1) could be dropped while only the tail is kept. In practice leak-detector runs on the already-truncated string in the bridge; the 8k cap is a legitimate trade-off between bounded memory and complete leak scanning. Not blocking.
- **Multi-tenant warning bridge is fully disabled.** The PR body still mentions multi-tenant *scoping*; the follow-up commit reverses that to "off in multi-tenant". PR body should be updated so readers don't chase a feature that's no longer there.

### Style / conventions
- ✔ `crate::` imports preferred; no `super::` in cross-module references outside tests.
- ✔ No `.unwrap()`/`.expect()` in non-regex production code; regex `.unwrap()` is annotated per `leak_detector.rs` convention.
- ✔ `thiserror`-family error handling unchanged (new code doesn't introduce error types).
- ✔ Logging levels: only `tracing::debug!` used (the multi-tenant notice). No `info!` / `warn!` added that could corrupt REPL.
- Minor: `strip_url_sensitive_parts` — the `.map(...).map(...)` chain using the first map only for its `(i, marker)` tuple is a little over-fluent; a bare `if let Some(i) = …` block would read more cleanly.
- `redact_shell_command_for_display` strips URL query strings *regardless of shell context* — if `shell` is invoked with a benign public URL, the summary shows `https://example.com/?…` instead of the full path. Users might lose a diagnostic signal. Arguable trade-off; leaves no secrets behind, which is the conservative choice.

### Test coverage
- ✔ 6 regression tests in `types::event::tests` covering: generic fallback sensitive-key skip, http URL strip (query/userinfo/fragment + non-URL passthrough), shell auth-flag redaction, URL-in-shell query strip.
- ✔ 1 regression test in `bridge::router::tests::redact_code_executed_scrubs_secrets_from_code_stdout_and_return_value` covering code + stdout + nested JSON with a GitHub-token pattern.
- ✔ `event_type_matches_serde_type_field` updated to include the two new `AppEvent` variants.
- **Gap — frontend:** new `GATE_RESOLUTION_STATUS` / `STATUS_TO_ACTIVITY` tables and `on()` wrapper lack JS unit/e2e coverage. Per `.claude/rules/testing.md` "Test Through the Caller, Not Just the Helper" this is below the bar for a helper change but would be worthwhile for the `expired → failure` regression the reviewer called out.
- **Gap — multi-tenant disable:** no automated coverage that `spawn_warning_bridge` is *not* called when `multi_tenant_mode=true`. A one-line harness test on `GatewayChannel::start` with the flag set would lock the behavior.

### Security
- ✔ **Three leak vectors the reviewer flagged are now closed:** URL-query/userinfo, shell auth-flags, CodeExecuted stdout. Each has a regression test.
- ✔ **Size bounded:** 8k-char cap on code/stdout prevents unbounded storage/broadcast.
- ✔ **Verbose-only gating:** non-debug SSE subscribers don't receive these payloads at all; debug subscribers get redacted content.
- ✔ **Multi-tenant misrouting closed:** disabled entirely for multi-tenant; single-user is unchanged.
- **Residual risk (minor):** `redact_shell_command_for_display` regex family doesn't cover every possible leak shape. A `shell` command with an inline literal matching a known-key pattern bypasses the four regexes and is not re-scanned by `LeakDetector` because summaries are produced in the engine crate (which has no safety dep) before the bridge leak-detector pass. For `shell`, leak-detector coverage would be an additional defense layer. Non-blocking; suggest follow-up.
- **Residual risk (minor):** `redact_secrets_in_json` recurses string-by-string through the return_value tree; cost is O(n × patterns). Bounded because return_value itself came through the 8k cap, but worth noting.

### Performance
- `OnceLock<[Regex; 4]>` avoids recompilation per call — good hot-path hygiene.
- Leak detector's Aho-Corasick prefix matcher short-circuits most calls; cost for clean strings is near-zero.

### Recommendations
1. Update the PR body to reflect that the multi-tenant warning bridge is **off**, not scoped. (Non-blocking, doc accuracy.)
2. Consider extending `redact_secrets_in_json` to apply `is_sensitive_param_key` to object keys as well — the denylist in the engine crate doesn't help JSON returns from CodeAct. (Non-blocking; follow-up issue.)
3. Add a harness test in `src/channels/web/tests.rs` that asserts `spawn_warning_bridge` is skipped when `multi_tenant_mode` is `true`. Locks the current contract. (Non-blocking.)
4. The one pre-existing flaky test (`extensions::manager::tests::inject_mcp_client_partitions_cache_by_user`) is unrelated to this PR but worth a follow-up issue if one doesn't exist already.

**Overall:** ✅ Approve. Three reviewer-flagged security issues addressed with matching regression tests; behavior changes are carefully bounded and well-documented. The small follow-ups above can ship in a separate PR.

src/channels/web/mod.rs
```diff
@@ -655,6 +655,35 @@ impl Channel for GatewayChannel {
...
+        // multi-tenant deployments entirely.
+        if let Some(log_broadcaster) = self.state.log_broadcaster.as_ref() {
+            if self.state.multi_tenant_mode {
+                tracing::debug!(
+                    "warning bridge disabled in multi-tenant mode: \
+                     WARN/ERROR log forwarding to debug panel requires \
+                     per-request tenant provenance that is not yet \
+                     wired through the tracing layer"
+                );
+            } else {
```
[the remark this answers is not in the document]
ilblackdragon: Fixed — PR description updated to state the warning bridge is disabled (not tenant-scoped) in multi-tenant mode. Both the Summary bullet and the Test plan's multi-tenant check now reflect the actual behavior shipped in 1d499f88. Re-enabling the bridge in multi-tenant deployments is blocked on per-request provenance being threaded through every `warn!`/`error!` call site, which is outside this PR's scope.

src/channels/web/log_layer.rs
```diff
@@ -298,6 +301,58 @@ impl WebLogLayer {
...
+                        continue;
+                    }
+                    let event = AppEvent::Warning {
+                        source: entry.target,
+                        message: entry.message,
+                        thread_id: None,
+                    };
+                    match &owner_id {
+                        Some(uid) => sse.broadcast_for_user(uid, event),
+                        None => sse.broadcast(event),
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in 9f438035. Both `sse.broadcast_for_user` and `sse.broadcast` sites in `spawn_warning_bridge` now carry `// projection-exempt: log source, WARN/ERROR tracing bridge → AppEvent::Warning` annotations — the warning bridge is a typed source log in the sense of `.claude/rules/gateway-events.md` (every `AppEvent::Warning` on SSE projects from exactly one `LogEntry` produced by `WebLogLayer`), but that source log isn't yet in the rule's fixed table of three. Added a comment block above the match explaining why the annotation is there. Pre-commit safety script now passes clean on this branch.

crates/ironclaw_engine/src/executor/orchestrator.rs
```diff
@@ -896,6 +896,37 @@ async fn handle_execute_code_step(
...
+                thread.id,
+                EventKind::CodeExecuted {
+                    step_id: exec_ctx.step_id,
+                    code: tail_chars(&code, CODE_EXECUTED_MAX_CHARS),
+                    stdout: tail_chars(&result.stdout, CODE_EXECUTED_MAX_CHARS),
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in 9f438035. At the `CodeExecuted` emission site, `tail_chars` is replaced with:
- `tail_utf8_bytes(s, max_bytes)` — byte-based, O(1) start + ≤3-byte UTF-8 boundary walk. Used for `code` and `stdout`.
- `bounded_return_value(v, max_bytes)` — `Null` ⇒ `None`, `String` ⇒ tail-truncated, structured `Array`/`Object`/`Number`/`Bool` serialized and either preserved intact if `≤ max_bytes` or dropped to `None`. Truncating arbitrary JSON would be unparseable on the frontend so dropping is the conservative choice — observers see `None` and know the payload was omitted for size.

`tail_chars` is kept for its other callers (`OUTPUT_TRUNCATE_LEN` / 500-char error slices) whose inputs are already pre-bounded and where chars-vs-bytes semantics matter for character-count messaging. Seven regression tests added covering ASCII, multi-byte emoji boundary, Null, small struct, oversized struct, and large-string paths.

PR #2851, opened 2026-04-22
fix(wasm): remove stale 10M fuel limit from settings DB

PierreLeGuen:
## Summary

- Databases that persisted `wasm.default_fuel_limit = 10000000` before the code default was bumped to 500M still read the old value at startup (DB-first resolution in `config/wasm.rs`), causing WASM tools like `google_slides` to fail with "Fuel exhausted: execution exceeded 10000000 fuel units"
- Adds migration V25 (PostgreSQL + libSQL) that deletes the stale setting row when its value is <= 10M, so the 500M code default takes effect
- Users who intentionally set a custom limit above 10M are unaffected

## Test plan

- [x] `released_migrations_are_immutable` test passes with new checksum
- [x] Verified locally: after migration, google_slides tool executes successfully with 500M fuel budget
- [ ] Verify on staging that existing databases with stale 10M setting get upgraded

### DIFF
--- a/migrations/V25__wasm_fuel_limit_bump.sql
+++ b/migrations/V25__wasm_fuel_limit_bump.sql
@@ -1,0 +1,7 @@
+-- The code default for wasm.default_fuel_limit was bumped from 10M to 500M
+-- (limits.rs, config/wasm.rs), but databases that persisted the old 10M value
+-- in the settings table still read it back at startup (DB-first resolution).
+-- Delete the stale row so the code default takes effect.
+DELETE FROM settings
+WHERE key = 'wasm.default_fuel_limit'
+  AND CAST(value AS BIGINT) <= 10000000;
--- a/migrations/V25__wasm_fuel_limit_bump.sql
+++ b/migrations/V25__wasm_fuel_limit_bump.sql
@@ -4,4 +4,4 @@
 -- Delete the stale row so the code default takes effect.
 DELETE FROM settings
 WHERE key = 'wasm.default_fuel_limit'
-  AND CAST(value AS BIGINT) <= 10000000;
+  AND (value#>>'{}')::BIGINT = 10000000;
--- a/src/db/libsql_migrations.rs
+++ b/src/db/libsql_migrations.rs
@@ -992,6 +992,21 @@ ALTER TABLE agent_jobs ADD COLUMN restart_params TEXT;
 CREATE INDEX IF NOT EXISTS idx_llm_calls_created_at ON llm_calls(created_at);
 "#,
     ),
+    (
+        25,
+        "wasm_fuel_limit_bump",
+        // The code default for wasm.default_fuel_limit was bumped from 10M to
+        // 500M (limits.rs, config/wasm.rs), but databases that persisted the
+        // old 10M value in the settings table still read it back at startup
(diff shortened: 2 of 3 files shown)


migrations/V25__wasm_fuel_limit_bump.sql
```diff
@@ -0,0 +1,7 @@
+-- The code default for wasm.default_fuel_limit was bumped from 10M to 500M
+-- (limits.rs, config/wasm.rs), but databases that persisted the old 10M value
+-- in the settings table still read it back at startup (DB-first resolution).
+-- Delete the stale row so the code default takes effect.
+DELETE FROM settings
+WHERE key = 'wasm.default_fuel_limit'
+  AND CAST(value AS BIGINT) <= 10000000;
```
serrrfirat: **Medium Severity**

This `<= 10000000` predicate deletes more than the stale historical default. It will also remove intentionally lowered custom limits (for example 1M or 5M), which means those installs silently jump to the new 500M default on upgrade.

Could this narrow the data fix to the exact stale default you want to clear instead of treating every lower persisted value as obsolete?

src/db/libsql_migrations.rs
```diff
@@ -990,6 +990,21 @@ ALTER TABLE agent_jobs ADD COLUMN restart_params TEXT;
...
+        // The code default for wasm.default_fuel_limit was bumped from 10M to
+        // 500M (limits.rs, config/wasm.rs), but databases that persisted the
+        // old 10M value in the settings table still read it back at startup
+        // (DB-first resolution). Delete the stale row so the code default
+        // takes effect; users who intentionally lowered the limit can re-set
+        // it via the settings API.
+        r#"
+DELETE FROM settings
+WHERE key = 'wasm.default_fuel_limit'
+  AND CAST(value AS INTEGER) <= 10000000;
```
serrrfirat: **High Severity**

This migration is casting the raw `settings.value` payload directly, but that column stores JSON-encoded values. If a historical row was persisted as a JSON string like `"10000000"`, PostgreSQL can fail the cast and SQLite/libSQL can coerce it differently than intended.

Could this normalize/extract the JSON scalar first and then compare the numeric value, with the same logic in both backends?

serrrfirat:
## Addressed review feedback in f8881589

**Changes:**

1. **JSONB extraction before cast** — `settings.value` is `JSONB`, so direct `CAST(value AS BIGINT)` would crash on PostgreSQL. Fixed both backends:
   - **PostgreSQL**: `(value#>>'{}')::BIGINT` — extracts the JSONB scalar as text first
   - **libSQL**: `CAST(json_extract(value, '$') AS INTEGER)` — equivalent JSON extraction

2. **Narrowed predicate from `<=` to `=`** — the `<= 10000000` condition would silently delete intentionally lowered custom limits (e.g. 1M, 5M). Now only deletes the exact stale default (`= 10000000`).

3. **Updated `checksums.lock`** to match the modified V25 migration.

**Migration merge safety**: staging currently ends at V24 (`llm_calls_created_at_index`). No V25 exists on staging, so this merges cleanly with no version conflict.

All quality gates pass (fmt, clippy zero warnings, 5373 tests).

PR #2854, opened 2026-04-22
feat(engine-v2): add CodeAct host shims and gated rich result objects

serrrfirat:
## Summary

- Add a curated set of Pythonic CodeAct host shims so Monty-backed CodeAct feels more like normal Python orchestration while still executing through the existing canonical host tools.
- Add real rollout controls for the shim layer so it can be A/B tested cleanly without changing raw canonical tool availability.
- Add deterministic and live/replay evaluation coverage so this direction is based on measured agent behavior, not just aesthetics.
- This is the **clean staging-based branch** version of the work and intentionally excludes the unrelated OpenClaw/Hermes migrate commits that were accidentally in the earlier branch ancestry.

## Why this PR exists

Monty is good enough for orchestrating host actions, but it feels unlike normal Python when the model has to repeatedly adapt canonical tool output shapes by hand (line-numbered file text, JSON parsing boilerplate, relative glob paths, shell result normalization, etc.).

The goal of this PR is **not** to give Monty direct OS/filesystem/network powers and **not** to create a second capability system. The goal is to reduce model-side adapter glue while preserving the existing engine-v2 trust boundary.

In practice:

1. Keep Monty as the Python execution surface.
2. Keep approvals / leases / policy / logging at canonical host actions.
3. Add a thin Pythonic compatibility layer over those canonical actions.
4. Measure whether that actually helps on realistic tasks.

### 1. Host-backed CodeAct shims

Added the following helpers in `crates/ironclaw_engine/src/executor/scripting.rs`:

- Text/file helpers
  - `read_text(path)`
  - `write_text(path, text)`
  - `append_text(path, text)`
- Structured data helpers
  - `read_json(path)`
  - `write_json(path, value)`
- Discovery helpers
  - `exists(path)`
  - `list_entries(path=".", recursive=False, max_depth=3)`
  - `find_files(pattern, path=".", max_results=None)`
- Network/process helpers
  - `http_get(url, headers=None)`
  - `http_request(method, url, headers=None, body=None, json=None)`
  - `run(command, timeout=None, workdir=None)`

These are adapters on top of the canonical host tools, not a new capability surface.

### 2. Runtime rollout gating

Added rollout controls at three layers:

- env/config: `CODEACT_HOST_SHIMS=true|false`
- agent config: `AgentConfig.codeact_host_shims`
- per-thread config: `ThreadConfig.codeact_host_shims`
- harness overrides:
  - `TestRigBuilder::with_codeact_host_shims(...)`
  - `LiveTestHarnessBuilder::with_codeact_host_shims(...)`

When disabled:

- shim names disappear from CodeAct lookup / `globals()`
- shim guidance is stripped from the CodeAct prompt
- raw canonical tools remain available

### 3. Evaluation coverage

This branch now has three evaluation layers:

- scripting/runtime tests in `crates/ironclaw_engine/src/executor/scripting.rs`
- deterministic CodeAct scenario benchmarks in `tests/engine_v2_skill_codeact.rs`
- live/replay A/B scenarios in `tests/e2e_live_codeact_shims.rs`

The live/replay suite covers 11 raw-vs-shim scenarios:

- `read_json`
- `append_text`
- `find_files`
- `package_json_edit`
- `tsconfig_edit`
- `monorepo_package_migration`
- `tsconfig_nested_paths`
- `js_codemod_use_strict`
- `mixed_config_sync`
- `yaml_workflow_update`
- `cargo_toml_rust_version_sync`

### Main finding

The strongest wins are on **adapter-heavy multi-file orchestration workflows**, not on tiny one-shot convenience calls.

### Strong live/replay shim wins

- `monorepo_package_migration`
  - raw: `5` LLM calls / `104110` tokens
  - shim: `1` LLM call / `20339` tokens
- `package_json_edit`
  - raw: `4` / `81156`
  - shim: `1` / `20036`
- `js_codemod_use_strict`
  - raw: `10` / `205214`
  - shim: `2` / `40433`
- `yaml_workflow_update`
  - raw: `8` / `166167`
  - shim: `3` / `61952`
- `cargo_toml_rust_version_sync`
  - raw: `10` / `211895`
  - shim: `3` / `62885`

### Neutral / small shim scenarios

- `mixed_config_sync`
  - raw: `2` / `42218`
  - shim: `2` / `41764`
- `read_json`
  - raw: `2` / `40130`
  - shim: `2` / `39627`
- `tsconfig_edit`
  - raw: `1` / `20491`
  - shim: `1` / `20150`
- `tsconfig_nested_paths`
  - raw: `1` / `20349`
  - shim: `1` / `20172`
- `append_text`
  - raw: `1` / `20131`
  - shim: `1` / `19854`
- `find_files`
  - raw: `1` / `19980`
  - shim: `1` / `19928`

Interpretation:

- the shim layer looks most worthwhile for **multi-step, path-sensitive, read/transform/write tasks**
- it looks mostly neutral on already-clean 1-turn flows
- that argues for keeping the built-in shim set curated and small rather than growing many task-shaped permanent abstractions

## Security / trust-boundary impact

This PR does **not** give Monty direct OS/filesystem/network access.

Every shim still flows through the same canonical host actions:

- file shims -> `read_file` / `write_file` / `list_dir` / `glob`
- network shims -> `http`
- process shims -> `shell`

That means:

- approvals still happen at the canonical action layer
- leases still apply at the canonical action layer
- policy remains keyed to canonical actions
- logging / effect execution remain canonical

## Change Type

- [x] New feature
- [ ] Bug fix
- [ ] Refactor
- [x] Documentation
- [ ] CI/Infrastructure
- [ ] Security
- [ ] Dependencies

## Linked Issue

None

## Validation

- [ ] `cargo build`
- [x] Relevant tests pass:
  - `cargo test -p ironclaw_engine executor::scripting::tests -- --nocapture`
  - `cargo test -p ironclaw_engine executor::prompt::tests -- --nocapture`
  - `cargo test --features libsql --test engine_v2_skill_codeact -- --nocapture`
  - `cargo test --features libsql --test e2e_engine_v2 v2_test_rig_can_toggle_codeact_host_shims -- --nocapture`
  - `cargo test --features libsql --test e2e_live_codeact_shims -- --ignored --nocapture --test-threads=1`
- [x] Manual / live validation:
  - recorded and replayed the shim A/B fixture suite
  - replayed the full ignored shim suite after cherry-picking onto this clean staging-based branch

## Security Impact

Touches CodeAct runtime behavior, prompt guidance, and test harness controls, but preserves the existing trust boundary. No new direct OS/filesystem/network power is exposed inside Monty.

## Database Impact

None

## Blast Radius

Primary blast radius:

- `crates/ironclaw_engine/src/executor/scripting.rs`
- `crates/ironclaw_engine/src/executor/prompt.rs`
- `crates/ironclaw_engine/src/types/thread.rs`
- top-level agent config / router plumbing
- test/live harness overrides
- engine-v2 CodeAct benchmark suite

Potential regressions would show up as:

- shim lookup / prompt guidance drift
- replay fixture drift in the benchmark suite
- prompt/runtime gating mismatch

## Rollback Plan

- revert the PR
- disable shims at rollout time with `CODEACT_HOST_SHIMS=false`

## Review Follow-Through

A richer `HttpResponse` / `CompletedProcess` return-object layer was explored during branch development, but the evidence was mixed-to-negative, so it was **backed out before this clean staging-based PR**. In particular, the more complex `CompletedProcess` recovery scenario was materially worse, so this PR deliberately stays focused on the simpler dict-style shim layer.

This PR supersedes the earlier draft that was opened from the wrong branch ancestry.

---

**Review track**: C


serrrfirat:
Addressed review follow-up in c8606241:\n\n- propagated immediate-shim execution gate pauses as real thread approval pauses instead of surfacing a RuntimeError\n- hardened `exists()` to probe with `read_file` and fall back to `list_dir` for directories\n- fixed `list_entries()` so it only strips real size suffixes and preserves legitimate parenthetical filenames\n- cleaned up the prompt/router slices needed for safety/clippy checks\n\nAlso investigated two notes that did **not** turn into code changes:\n- `append_text()` already reads via `read_file(path, offset=1)`, and explicit `offset` disables the file tool's default 2000-line truncation behavior\n- `http_request(json=...)` already passes through to the canonical `http` tool, which accepts structured JSON bodies directly

crates/ironclaw_engine/src/executor/scripting.rs
```diff
@@ -1204,79 +1773,401 @@ async fn preflight_action(
...
+                "append_text",
+            ) {
+                Ok(value) => value,
+                Err(err) => return Some(ImmediateShimOutcome::Ready(err)),
+            };
+
+            let read_call_id = format!("{call_id}:read");
+            let existing = match execute_canonical_action_now(
+                "read_file",
+                serde_json::json!({"path": path.clone(), "offset": 1}),
```
henrypark133, line 1952: append_text currently reconstructs the whole file by calling read_file with offset 1 and then writing the concatenated result back out. read_file defaults to a 2000-line window when no limit is provided, so appending to a long file will silently drop everything after line 2000 before the write_file call.

src/config/agent.rs
```diff
@@ -157,6 +161,7 @@ impl AgentConfig {
             max_llm_concurrent_per_user: parse_option_env("TENANT_MAX_LLM_CONCURRENT")?,
             max_jobs_concurrent_per_user: parse_option_env("TENANT_MAX_JOBS_CONCURRENT")?,
             engine_v2: parse_bool_env("ENGINE_V2", false)?,
+            codeact_host_shims: parse_bool_env("CODEACT_HOST_SHIMS", true)?,
```
henrypark133, line 164: This introduces CODEACT_HOST_SHIMS as a global agent-level toggle, but the new propagation is still incomplete. The gateway conversation path threads it into ThreadConfig, while mission threads still spawn with ThreadConfig::default in runtime/mission.rs, so setting this to false will not actually disable shims for mission-driven engine threads.

henrypark133 (changes requested):
Two verified issues remain in the new host-shim path: append_text can truncate long files, and the global CODEACT_HOST_SHIMS toggle is not applied consistently across thread creation paths.

henrypark133 (changes requested):
What looks good:
- The append-text truncation issue appears fixed: the shim now reads the full file via `read_file(offset=1)` before writing back.
- The foreground router path and nested `rlm_query()` child-thread path both carry `codeact_host_shims` through correctly.
- The evaluation and replay coverage in this branch is much stronger than the earlier iteration.

Findings:
1. High - `crates/ironclaw_engine/src/runtime/mission.rs:762-767`, `crates/ironclaw_engine/src/types/mission.rs:95-171`: mission-spawned engine-v2 threads still use `ThreadConfig::default()` and the mission model does not persist the shim toggle, so `CODEACT_HOST_SHIMS=false` is silently ignored on mission fires. Why it matters: the rollout gate is not actually global, and background mission execution can diverge from the configured foreground/runtime posture. Expected fix direction: carry the shim flag into mission thread creation from the owning config or persist it on the mission/thread creation path so mission threads inherit the same rollout setting.

Summary:
- Review coverage: checked-out worktree plus focused engine/runtime scan of the previously requested fixes
- Residual risk: I did not run the full replay suite locally in this pass

PR #2863, opened 2026-04-22
feat(llm): add Manifest as built-in LLM provider

SebConejo:
## Summary

- Add Manifest to the provider registry with OpenAI-compatible protocol
- Default model is `auto`, Manifest picks the right complexity tier per request
- Works with cloud (`app.manifest.build`) and self-hosted (`localhost:2099`) via `MANIFEST_BASE_URL`

## Configuration

```
LLM_BACKEND=manifest
MANIFEST_API_KEY=mnfst_...
MANIFEST_MODEL=auto
```

## Test plan

- [ ] `cargo test -- test_builtin_registry_loads` passes
- [ ] `cargo test -- test_find_by_id` passes
- [ ] `ironclaw onboard --step provider` shows "Manifest" in the list
- [ ] Setting `LLM_BACKEND=manifest` with a valid key routes requests through Manifest

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -77,6 +77,12 @@ NEARAI_AUTH_URL=https://private.near.ai
 # LLM_EXTRA_HEADERS=HTTP-Referer:https://myapp.com,X-Title:MyApp
 
 
+# === Manifest (smart model router) ===
+# LLM_BACKEND=manifest
+# MANIFEST_API_KEY=mnfst_...
+# MANIFEST_MODEL=auto                              # default; Manifest picks the best model per request
+# MANIFEST_BASE_URL=https://app.manifest.build/v1  # default (cloud); use http://localhost:2099/v1 for self-hosted
+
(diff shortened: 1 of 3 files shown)


SebConejo:
Hi @henrypark133, any news regarding this PR? Let me know if I have to do something.

serrrfirat:
Thank you for adding Manifest as a declarative provider. This idea fits the current Reborn provider registry, so we would like to keep the PR open rather than close it.

`providers.json` is still compiled into `crates/ironclaw_llm/src/registry.rs` and is consumed by the Reborn LLM administration and provider-resolution paths. Could you please refresh/rebase the PR onto current main, verify the current hosted and self-hosted endpoints, and update the Reborn-facing provider documentation/tests? The desired final change should keep the declarative registry entry and remove or update any stale v1-only documentation/config assumptions.

If you are able to port it, we would be happy to review the Reborn-ready version. Thanks again for contributing this integration.

PR #2865, opened 2026-04-22
feat(tools): add Nostr tool + WebSocket host support

Kampouse:
Add Nostr WASM tool + WS host support. See commit messages for details.


henrypark133 (changes requested):
What looks good:
- The WebSocket support stays inside the WASM host boundary instead of leaking relay-specific logic upward.
- The Nostr tool itself is mostly isolated to `tools-src/nostr/*`.
- Returning WS frames as a JSON array is a better guest contract than newline-joining messages.

Findings:
1. High - `tools-src/nostr/nostr-tool.capabilities.json:10`, `tools-src/nostr/src/transport.rs:43,97`: the new WS path is validated as `GET`, but `nostr-pub.wellorder.net` is still allowlisted as `POST`-only. Why it matters: that shipped relay is now unusable because the host rejects it before connect. Expected fix direction: align WS relay allowlist entries with the actual method contract.

Summary:
- Prior feedback status: no prior blocking review found
- Review coverage: checked-out diff, host/runtime code, and review-thread state
- Residual risk: the one-shot WS shim still needs stronger end-to-end relay coverage

Kampouse:
Good will fix 👌

Kampouse:
## Addressed blocking review from @henrypark133

Fixed in 30f63459.

**Finding 1 (High):** `nostr-pub.wellorder.net` POST-only allowlist blocking WS connections → **Fixed.** Added `"GET"` to the methods array in `nostr-tool.capabilities.json:10`.

The `ws_roundtrip` host path calls `check_http_allowed(url, "GET")` since WebSocket upgrades originate from HTTP GET. This relay was the only one missing `GET`, so the host rejected all WS connections to it before TCP handshake. The other three relay entries (`relay.damus.io`, `nos.lol`, `relay.nostr.band`) already had `["GET", "POST"]`.

**Residual risk:** Agree the one-shot WS shim needs stronger e2e relay coverage — will track as a follow-up.

Kampouse:
## Full review response — all findings addressed

### henrypark133 (High, blocking) — Fixed in 30f63459
- `nostr-pub.wellorder.net` POST-only allowlist → added `"GET"` to methods
- `ws_roundtrip` validates against GET (WebSocket upgrades are HTTP GET)

### gemini-code-assist[bot] (medium, 6 findings) — All addressed in ab99647c
1. **Binary outbound WS** — ws_roundtrip already handles binary (line 334-338: tries UTF-8 first, falls back to Message::Binary)
2. **Newline-joining fragile** — Now returns JSON array of frames (line 374-382), binary as base64
3. **Hardcoded 2 MiB** — Uses `max_response_bytes` from capabilities (line 346-353), 64 KiB floor
4. **Binary frames dropped** — `ws_exchange` collects both Text and Binary via WsFrame enum (line 518-523)
5. **Single relay profile query** — `execute_get_profile` now iterates all relays (lib.rs:209-225)
6. **Single relay notes query** — `execute_get_notes` aggregates + deduplicates by event_id (lib.rs:340-361)

### E2e relay coverage — Added in adccb73
- **`test_ws_roundtrip_real_nostr_relay`**: Sends NIP-01 REQ to all 4 configured relays, validates 101 status + JSON array frame format + EVENT/EOSE content. Network-tolerant (skips if no relay reachable).
- **Fixed `test_ws_roundtrip_echo`**: Allowlist method was POST, changed to GET (was a latent bug — WS validates against GET)

serrrfirat:
Thank you for taking on Nostr support and exploring the WebSocket runtime requirements behind it. We are closing this PR as part of the transition from the legacy channel/tool runtime to the Reborn architecture.

Nostr remains a potentially valuable integration—`FEATURE_PARITY.md` still marks it as missing in Reborn—but this branch combines the Nostr tool with generic legacy WASM WebSocket host changes and unrelated channel/build updates. Reborn does not grant extensions inherited network or DNS access: transport must go through an explicit host-mediated capability, while conversational integrations use the typed ProductAdapter and ProductWorkflow contracts. Forward-porting the current host changes would bypass those boundaries.

We would genuinely welcome a clean Reborn contribution. A good first slice would be either a read-only Nostr extension capability with narrowly declared relay egress, or a focused ProductAdapter proposal that separates relay WebSocket transport, event authentication, canonical actor/conversation binding, and outbound delivery. The NIP parsing and protocol experience developed here would be very useful. Thank you for investing in this integration.

PR #2868, opened 2026-04-22
engine-v2: make available_actions callable-only for blocked providers

henrypark133:
## Context

This PR now includes the follow-up work from #2869, #2876, and #2889, so it should be read against the engine-v2 epic comment in #2767:
https://github.com/nearai/ironclaw/issues/2767#issuecomment-4293374615

The main alignment is:
- Task 4: `available_actions()` callable-only cleanup
- Task 5: prompt / loop alignment
- Task 6: canonical callable-action discovery metadata
- cleanup of the earlier Section 7 prototype so we do not keep a hidden inspect/promote path around in the runtime
- the later enablement follow-up that makes blocked integrations capability-first and routes model-visible enablement through `tool_activate(name=...)`

## Old Experience

Before this stack:
- `available_actions()` mixed together the wrong provider states in engine v2. Some blocked provider actions (`needs_auth`, `needs_setup`, inactive, not-installed, routed-only) could still show up as normal callable tools even though they were not directly invokable on that turn, while installed-but-not-yet-activated latent WASM provider actions could disappear entirely and fail to trigger auth-on-first-call.
- The CodeAct system prompt duplicated callable tool inventory in prose. That meant the same tool information could appear both in provider schemas and in prompt text, and resume / compaction paths could keep stale or duplicated metadata instead of rebuilding one canonical view.
- Action discovery was split across multiple partially-connected paths. Registry-backed tools and engine-native mission actions did not share one discovery envelope, `tool_info` was not consistently reading the same callable snapshot the executor was using, and hyphen/underscore aliases were easy to handle inconsistently at different call sites.
- Blocked integrations did not have one clean model contract. Some paths still assumed first-call recovery on blocked provider actions; others treated blocked state as background metadata; `tool_install`, `tool_auth`, and latent synthetic actions all leaked overlapping lifecycle detail into the surfaced experience.
- Some deferred-tool scaffolding still implied a `tool_info` -> promotion flow. That was leftover prototype behavior, not the contract we want to preserve.

## New Experience

After this stack:
- The provider-facing callable surface is the callable-now surface. Ready actions remain callable; blocked provider/channel integrations are no longer normal model-facing callable tools.
- The prompt now has three distinct model-visible layers:
  - normal callable actions/tool inventory
  - `Capabilities` for contextual/background runtime information
  - `Activatable Integrations` for blocked managed integrations
- `tool_activate(name=...)` is the single model-facing enablement path in engine v2 and CodeAct. The harness owns whether that means install, auth, activation, or a surfaced manual-setup failure.
- Capability entries for blocked integrations now include compact action previews so the model can see what enablement unlocks without dumping large tool families into the default callable surface.
- `tool_info` now reads the current callable snapshot / inventory, resolves canonical names and hyphen/underscore aliases consistently, fails cleanly for names that are not callable in the current execution context, and does not gate or promote a new tool into the next LLM step's callable set.
- Newly enabled tools become visible on the next top-level turn / orchestrator iteration, not mid-CodeAct step.

## What Changed

- add `ActionDiscoveryMetadata`, `ActionDiscoverySummary`, `ActionInventory`, and `ActionDef::matches_name()` so callable action metadata is canonical and alias-safe
- rework bridge projection so `ActionProjector` only surfaces callable-now actions and `CapabilityProjector` owns blocked provider/channel background plus action previews
- add engine-native mission action discovery metadata in `src/bridge/engine_actions.rs`
- thread action snapshots through orchestrator and scripting paths so policy checks, execution, and `tool_info` all read the same callable set
- replace the old prompt-side callable tool list with separate `Capabilities` and `Activatable Integrations` sections, and refresh engine-owned system prompts across resume / compaction without duplicating appended sections
- make `tool_activate` the normal surfaced enablement operation, while preserving install approval semantics when enablement would auto-install a not-yet-installed integration
- remove the short-lived extension-list cache in the effect bridge so immediate post-auth prompt state stays fresh on the next turn
- pin isolated E2E mock-LLM fixtures explicitly to the mock `openai_compatible` provider instead of relying on env-vs-DB precedence
- replace the stale v2 auth preflight surface test with a dedicated browser E2E for the new `tool_activate` / `Activatable Integrations` contract
- update the relevant CLAUDE docs and inline type/tool comments so the repo documentation matches the new callable vs contextual vs activatable split

## Key Coverage Added In This Stack

- `available_actions_omit_installed_needs_auth_provider_action`
- `available_actions_omit_installed_needs_setup_provider_action`
- `available_actions_omit_installed_inactive_provider_action`
- `available_actions_omit_latent_inactive_provider_actions`
- `available_capabilities_include_latent_provider_background`
- `available_capabilities_include_latent_provider_activation_entry`
- `tool_activate_requires_approval_before_auto_installing_integration`
- `skill_prompt_context_survives_pause_and_resume`
- `skill_prompt_context_survives_compaction_and_resume`
- `resume_refreshes_checkpointed_system_prompt_metadata`
- `tool_info_does_not_gate_callable_tool_into_next_llm_callable_set`
- `execute_code_propagates_snapshot_to_tool_execution_context`
- `test_v2_tool_activate_surface.py`

## Testing

All of the following were run on this branch:

```bash
cargo fmt --all
cargo clippy --all-targets -- -D warnings
cargo test effect_adapter:: --lib

tests/e2e/.venv/bin/pytest \
  tests/e2e/scenarios/test_v2_auth_oauth_matrix.py::test_settings_first_gmail_auth_then_chat_runs \
  tests/e2e/scenarios/test_v2_auth_oauth_matrix.py::test_mcp_oauth_roundtrip_via_browser \
  tests/e2e/scenarios/test_telegram_hot_activation.py::test_telegram_auth_required_shows_configure_modal_and_can_cancel \
  tests/e2e/scenarios/test_telegram_hot_activation.py::test_telegram_hot_activation_transitions_installed_to_pairing \
  tests/e2e/scenarios/test_v2_tool_activate_surface.py -v
```

Result: all passed.

https://github.com/nearai/ironclaw/actions/runs/24864084782


src/bridge/effect_adapter.rs
```diff
@@ -4430,12 +4430,193 @@ mod tests {
...
+    ) -> EffectBridgeAdapter {
+        use crate::secrets::InMemorySecretsStore;
+        use crate::secrets::SecretsCrypto;
+        use crate::tools::mcp::process::McpProcessManager;
+        use crate::tools::mcp::session::McpSessionManager;
+
+        let dir = tempfile::tempdir().expect("temp dir");
+        let dir_path = dir.path().to_path_buf();
+        std::mem::forget(dir);
+        std::fs::create_dir_all(dir_path.join("tools")).expect("tools dir");
```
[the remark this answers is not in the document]
henrypark133: Fixed in 9751fd599b0d92a8b01f8b257c4cdf3f13cfc0a9. The test helper now returns a small fixture struct that owns both the adapter and the TempDir, so the provider fixture directory lives for the full test and is cleaned up automatically when the test ends.

src/bridge/action_projector.rs
```diff
@@ -90,22 +90,15 @@ impl ActionProjector {
...
-                        continue;
-                    }
+                let (kind, invocation_mode) = capability_surface_subject_for_extension(extension);
+                let assignment = assign_surface(SurfacePolicyInput {
+                    kind,
+                    status,
+                    invocation_mode,
+                    leased_and_callable: false,
+                });
+                if !assignment.available_actions {
```
serrrfirat: **High Severity**\n\nThis change removes the only callable path for installed-but-unauthenticated provider actions in engine v2. `available_actions()` now drops provider tools when the extension status is `NeedsAuth`, but `execute_action()` still treats a direct provider call as the trigger for `GatePaused(authentication)` via `check_tool_readiness()`. At the same time, capabilities are rendered only as background status in the system prompt, not as callable tools, and `tool_auth` is still hidden from the model.\n\nConcrete regression: if Gmail is installed but unauthenticated, `gmail_send` no longer appears in `available_actions`, so the model has no callable action left that can reach the auth gate and recover. The provider is visible only as a `needs_auth` capability line in the prompt, which the runtime cannot invoke.\n\nSuggested fix: either keep `NeedsAuth` extension-backed direct actions on the callable surface so first call can raise the auth gate, or introduce a separate callable auth/bootstrap action for installed blocked providers before removing them from `available_actions`.

src/bridge/effect_adapter.rs
```diff
@@ -4430,12 +4430,200 @@ mod tests {
...
+                Some(ext_mgr),
+                Some(Arc::clone(&tools)),
+            )))
+            .await;
+
+        ProviderFixture { adapter, _dir: dir }
+    }
+
+    #[tokio::test]
+    async fn available_actions_omit_installed_needs_auth_provider_action() {
```
serrrfirat, line 5807: **Medium Severity**\n\nThis new test fixture bakes in a behavior mismatch between projection-time status and execute-time readiness. The fixture asserts that an installed-but-inactive provider action should be omitted from `available_actions()`, but production execution still routes direct provider calls through `check_tool_readiness()` / `ensure_extension_ready(..., UseCapability)`, which can auto-activate an installed extension on first use.\n\nThat means the test can pass while the advertised action surface still disagrees with what the runtime is actually capable of doing. In practice this makes it easier to lock in a broken UX where the model cannot call something that the execution path would have recovered automatically.\n\nSuggested fix: add a caller-level test that exercises the real use path for an installed inactive provider and verifies the intended contract end-to-end, then make the `available_actions()` expectation match that contract instead of the current snapshot-only status.

src/bridge/action_projector.rs
henrypark133: NeedsAuth in the bridge is status metadata, not a callable auth action. The current meaning is "this provider exists but is not ready yet," and that is what this PR keeps consistent: blocked provider actions are removed from available_actions, while their state remains available through capability projection.

The stacked prompt-alignment follow-up (#2869) makes that capability state model-visible in one canonical background section via available_capabilities(), so the model still learns that, for example, slack exists and is needs_auth even though the concrete provider action is not callable yet.

So the intended contract here is:
- available_actions = actions the model can call directly in v2
- available_capabilities = background state for providers/channels that are present but not currently callable

Under that contract, NeedsAuth does not need to stay in available_actions, because auth/setup state is conveyed through capability metadata rather than by keeping blocked provider actions callable.

This comment assumes NeedsAuth is still meant to be a first-call recovery trigger. In this stack, that is no longer the contract; NeedsAuth is background capability state, and #2869 is the piece that surfaces that state canonically to the model.

src/bridge/effect_adapter.rs
henrypark133: The intended contract in this stack is that installed-but-inactive providers move to capability background rather than staying on the callable surface. `available_actions` is the model-callable set; `available_capabilities` is where inactive/background state is surfaced, so the omission here is intentional.

You're right that execute-time readiness still has an auto-activation fallback if a direct provider action is invoked anyway. We should treat that as an execution-side compatibility path, not as the definition of what belongs in `available_actions`.

So I don't think this PR should keep inactive providers model-callable just to match that fallback. If we want stricter projection/execution parity, the cleanup should happen separately in the execution path rather than by expanding the model-visible callable surface here.

henrypark133 (commented):
## Review: callable-only surface and prompt/inventory alignment look good

I did a full pass on the stacked engine-v2 changes here and did not find any verified blocker-level issues.

What looks good:
- The callable-surface contract is much cleaner now: blocked provider actions move out of `available_actions`, while background capability state stays model-visible through the canonical capabilities section.
- The snapshot plumbing is materially better. `tool_info`, structured execution, scripting, and orchestrator paths now read the same callable snapshot instead of drifting on alias handling or live registry state.
- The prompt refresh work is directionally right. Resume/compaction now rebuild the engine-owned system prompt while preserving appended step-zero context like prior knowledge and active skills.
- The new caller-level regression coverage is the right shape for this area, especially around resume/compaction behavior and `tool_info` not mutating the next-step callable set.

Low-priority notes:
- `refresh_system_prompt()` still fetches `available_action_inventory()` even though the current prompt builder ignores that inventory. That looks like avoidable start/resume work unless a deferred inventory section is about to land.
- `project_tool_action()` currently emits discovery metadata for every callable tool even when it only repeats the callable name and has no summary/schema override. Tightening that would trim some per-turn payload size.

Verification:
- `cargo test --test engine_v2_gate_integration tool_info_does_not_gate_callable_tool_into_next_llm_callable_set -- --nocapture`
- `cargo test --test engine_v2_skill_codeact skill_prompt_context_survives_pause_and_resume -- --nocapture`
- `cargo test --test engine_v2_skill_codeact skill_prompt_context_survives_compaction_and_resume -- --nocapture`

src/bridge/action_projector.rs
```diff
@@ -155,6 +143,34 @@ impl ActionProjector {
...
+        name: callable_name.clone(),
+        description: tool.description().to_string(),
+        parameters_schema: callable_schema,
+        effects: vec![],
+        requires_approval: false,
+        discovery: Some(ActionDiscoveryMetadata {
+            name: callable_name,
+            summary,
+            schema_override,
+        }),
```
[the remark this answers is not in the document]
henrypark133: Fixed in 38a13a14. `project_tool_action()` now leaves `ActionDef.discovery` as `None` unless projection adds real extra discovery content (curated summary or schema override), so normal callable tools no longer serialize a redundant `{name: ...}` discovery wrapper. I also added `project_tool_action_omits_empty_discovery_metadata` coverage for the empty-metadata case.

crates/ironclaw_engine/src/executor/loop_engine.rs
```diff
@@ -144,6 +214,112 @@ impl ExecutionLoop {
...
+        {
+            Ok(inventory) => inventory,
+            Err(error) => {
+                debug!(
+                    thread_id = %self.thread.id,
+                    "failed to load action inventory for system prompt refresh: {error}"
+                );
+                crate::types::capability::ActionInventory::default()
+            }
+        };
```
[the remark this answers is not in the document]
henrypark133: Fixed in 38a13a14. `refresh_system_prompt()` no longer calls `available_action_inventory()` on start/resume now that the prompt builder does not consume inventory. I also added `resume_refresh_does_not_fetch_action_inventory_for_system_prompt` so the refresh path fails the test if that inventory fetch comes back.

crates/ironclaw_engine/src/executor/prompt.rs
```diff
@@ -123,23 +135,24 @@ pub async fn build_codeact_system_prompt(
...
+    build_codeact_system_prompt_inner(inventory, capabilities, overlay.as_deref(), platform)
 }
 
 /// Shared prompt builder used by both the async and pre-fetched-docs variants.
 fn build_codeact_system_prompt_inner(
-    actions: &[ActionDef],
+    _inventory: &ActionInventory,
     capabilities: &[CapabilitySummary],
     overlay: Option<&str>,
     platform: Option<&PlatformInfo>,
```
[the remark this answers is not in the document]
henrypark133: Fixed in 38a13a14. I removed `ActionInventory` from `build_codeact_system_prompt()` / `build_codeact_system_prompt_with_docs()` and from the shared inner helper, then updated the prompt tests to call the narrower API directly. Prompt building no longer advertises an inventory dependency that it does not use.

src/llm/nearai_chat.rs
```diff
@@ -1981,6 +2008,7 @@ mod tests {
         );
 
         #[allow(unused_unsafe)]
+        // SAFETY: serialized via ENV_MUTEX.
         unsafe {
             std::env::remove_var("NEARAI_API_KEY");
         }
```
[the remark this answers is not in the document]
henrypark133: Fixed in 86958a28. I applied the same restore-on-drop guard here as well: the test now saves any existing `NEARAI_API_KEY`, sets the test value, and restores the original value on teardown instead of always removing it.

src/bridge/effect_adapter.rs
```diff
@@ -974,6 +985,71 @@ impl EffectBridgeAdapter {
...
+                            duration: start.elapsed(),
+                        });
+                    }
+                    Ok(None) => {
+                        let requested = parameters
+                            .get("name")
+                            .and_then(|value| value.as_str())
+                            .unwrap_or("<missing>");
+                        let error_msg = format!(
+                            "Tool tool_info failed: no callable action named '{requested}' is available in this execution context"
```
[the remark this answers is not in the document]
henrypark133: Fixed in 86958a28. I simplified the model-visible missing-tool message to `tool_info: no callable action named ... in this execution context`, keeping the same semantics while removing the redundant `Tool tool_info failed:` prefix.

nickpismenkov (changes requested):
Bugs (likely to cause real user-visible issues)                                                                                        
                                                                                                                                         
  B1. Installed+Inactive+Latent extensions become completely invisible 🔴                                                                
                                                                                                                                         
  src/bridge/capability_projector.rs:91-94 now drops any installed extension whose name matches a latent-action provider key from        
  capability background:                                                                                                                 
                                                                                                                                         
  if extension.installed {                                                                                                             
      installed_keys.insert(normalized_key.clone());                                                                                     
      if latent_provider_keys.contains(&normalized_key) {
          continue;   // omit from background entirely                                                                                   
      }                                                                                                                                  
      ...                                                                                                                                
  }                                                                                                                                      
                                                                                                                                       
  Combined with action_projector.rs which drops Inactive provider tools from available_actions (via assign_surface → available_actions:  
  false)                                                                
                                                                                                                                         
  Confirmed empirically: extensions/manager.rs::cached_latent_wasm_provider_actions returns latent actions for installed-but-inactive    
  WASM tools, so this case is real. The replaced test installed_latent_entries_stay_out_of_capability_background asserts               
  projected.is_empty() — the author intentionally removed the Inactive-status summary.                                                   
                                                                                                                                       
  Impact: The model has zero cue the extension exists. User has to either use the gateway UI or remember the tool's name to ask about    
  reactivation. The PR description's rationale ("no longer duplicated") only holds when the latent actions are actually callable — which
  they aren't for inactive extensions.                                                                                                   
                             
  Fix: narrow the skip condition to "installed AND active AND has latent actions" (or, equivalently, only skip when                      
  assign_surface(...).available_actions == true).
                                                                                                                                         
  B2. .unwrap_or_default() on inventory fetch silently breaks tool_info 🔴                                                               
   
  Three new sites:                                                                                                                       
  - crates/ironclaw_engine/src/executor/orchestrator.rs:985-994 (handle_execute_action)
  - crates/ironclaw_engine/src/executor/orchestrator.rs:1340-1347 (handle_execute_actions_parallel)                                      
  - crates/ironclaw_engine/src/executor/scripting.rs:435-442 (execute_code_with_skills)            
                                                                                                                                         
  Pattern:                   
  let inventory = Arc::new(                                                                                                              
      effects.available_action_inventory(&active_leases, &exec_ctx).await
          .unwrap_or_default(),                                                                                                          
  );                                                                                                                                     
  ...
  exec_ctx.available_actions_snapshot = Some(Arc::clone(&available_actions));                                                            
  exec_ctx.available_action_inventory_snapshot = Some(Arc::clone(&inventory));
                                                                                                                                         
  When the fetch errors (DB hiccup, auth lock contention, extension manifest read failure), the snapshot is set to Some(empty) — not
  None. Downstream:                                                                                                                      
                             
  - effect_adapter.rs:988-1051 sees canonical_action_name == "tool_info" and enters the snapshot short-circuit.                          
  - ActionDiscovery::tool_info_from_actions(&params, &[]) resolves nothing.
  - Returns is_error: true with "no callable action named '<x>' in this execution context" — for every tool_info call that turn.         
                                                                                                                                         
  Pre-PR, tool_info always went through ToolRegistry and worked regardless of inventory fetch errors. Net regression for error paths.    
                                                                                                                                         
  Fix: set snapshots only on Ok(_); let the fallthrough path handle errors as before. At minimum log the error before                    
  .unwrap_or_default() so the next prod incident is findable.
                                                                                                                                         
  B3. Canonical-name resolution mid-execution is asymmetric                                                                              
   
  effect_adapter.rs:951-962 canonicalizes via snapshot only for handlers that use canonical_name (handle_mission_call, the tool_info     
  short-circuit). But:       
                                                                                                                                         
  - tools.resolve_name(canonical_action_name) (line 958) uses canonical — good.                                                          
  - is_v1_only_tool(&lookup_name) / is_v1_auth_tool(&lookup_name) use lookup_name (post resolve_name) — good.
  - auth_mgr.execute_latent_extension_action(action_name, ...) at line 1076 still uses the raw action_name, not canonical. If LLM writes 
  some-latent-tool and snapshot canonicalizes to some_latent_tool, the latent-action dispatch sees the hyphenated form. Whether          
  execute_latent_extension_action is alias-safe isn't obvious from the PR.                                                               
                                                                                                                                         
  This is a mixed-variable inconsistency; not always a bug, but a real foot-gun for anyone touching this function.                       
   
  B4. Gate resume never populates snapshots → inconsistent tool_info behavior                                                            
                             
  src/bridge/router.rs:1088-1089 hand-builds ThreadExecutionContext with both snapshots = None, then calls                               
  execute_resolved_pending_action. If the resumed action somehow triggers a nested tool_info (e.g. a mission handler introspecting
  itself), it falls through to the ToolRegistry path which returns info for ANY registered tool — without the lease restriction the      
  snapshot path enforces. Two different answers for tool_info("x") in the same adapter depending on entry path.

  Low exploitability in practice, but violates the PR's stated invariant that "tool_info reads the current callable snapshot /           
  inventory."
                                                                                                                                         
  B5. Scripting path never populates available_action_inventory_snapshot

  execute_code_with_skills (scripting.rs:435-442) only sets available_actions_snapshot, not available_action_inventory_snapshot. Today   
  this is harmless (both paths end up calling the same resolver). But the entire point of introducing ActionInventory as a distinct type
  is to accommodate future structure (deferred actions, lookup-only actions, etc.). Scripting will silently miss those.                  
                             
  Either delete available_action_inventory_snapshot (rename to a single snapshot field) or populate both fields consistently in          
  scripting.
                                                                                                                                         
  Design/correctness nits                                                                                                                
   
  D1. Prompt refresh runs only once per run()                                                                                            
                             
  refresh_system_prompt is called exactly once, at the top of run(). Capabilities that change mid-thread (user authenticates a provider  
  during a long-running loop) won't be reflected in the system prompt until the next pause+resume. The existing tests only assert the
  pause/resume path. Either document this as intentional staleness, or move the refresh into the step-0 pre-LLM hook.                    
                             
  D2. RuntimeCheckpoint::update_working_messages_system_prompt drops message fields                                                      
   
  loop_engine.rs:77-80:                                                                                                                  
  *message = serde_json::json!({
      "role": "System",         
      "content": refreshed,
  });                                                                                                                                    
     
  Overwrites the entire message JSON. If a system message ever grew additional keys (per append_message's optional action_name /         
  action_call_id / action_calls fields are "impossible for System" — until they aren't), they're silently erased. Use message["content"] 
  = refreshed.into() instead.
                                                                                                                                         
  D3. Suffix fallback via full-postamble rfind is fragile                                                                                
   
  codeact_system_prompt_suffix falls back to rfind(CODEACT_POSTAMBLE). The postamble is 97 lines of code. Any whitespace or wording      
  change in crates/ironclaw_engine/prompts/codeact_postamble.md makes in-flight threads fail the fallback. For threads with a step-zero
  marker (Prior Knowledge / Active Skills / missing skills), this doesn't matter. For threads without any append but with custom trailing
   content (e.g., a future feature adds another append format), suffix is lost.

  Use a sentinel marker (<!-- ironclaw:codeact-body-end -->) instead of matching the whole postamble body.                               
   
  D4. Description hint size/noise                                                                                                        
                             
  llm_adapter.rs:420-438 appends (call tool_info(name: "...", detail: "summary") for rules/examples or detail: "schema" for the full     
  discovery schema) to every tool whose discovery_summary().is_some() || discovery_schema() != parameters_schema(). For mission_create /
  mission_update + all WASM tools with discovery metadata, this adds ~130 chars per tool description. A fleet of 20+ such tools = 2-3KB  
  of boilerplate in every turn's tool definitions.

  Consider a single instruction in the preamble ("any tool whose description mentions tool_info has curated guidance") and omit the      
  per-tool hint.
                                                                                                                                         
  D5. ActionDef: PartialEq on serde_json::Value

  crates/ironclaw_engine/src/types/capability.rs:124 adds PartialEq to ActionDef. parameters_schema: serde_json::Value participates. With
   IronClaw's typical preserve_order feature enabled, serde_json::Map uses IndexMap and equality is insertion-order-sensitive. Two
  semantically identical schemas constructed differently compare unequal.                                                                
                             
  Currently only used in tests, but this is public API surface. Either scope the derive #[cfg(test)] or add a doc comment warning that   
  equality is order-sensitive on schemas.
                                                                                                                                         
  D6. Dead flexibility: ActionDiscoveryMetadata.name always equals ActionDef.name for all current producers                              
   
  Every call site in this PR sets discovery.name = action.name:                                                                          
  - engine_actions.rs mission_* actions: literal name equality.
  - action_projector.rs::project_tool_action: ActionDiscoveryMetadata { name: callable_name.clone(), ... }.                              
                                                                                                           
  The indirection through matches_name → discovery_name never produces a different match. If there's no concrete plan to diverge,        
  consider dropping the field.                                                                                                           
                                                                                                                                         
  Verified safe / not-bugs                                                                                                               
                             
  - LeaseManager::find_and_consume alias handling: verified via GrantedActions::covers                                                   
  (crates/ironclaw_engine/src/types/capability.rs:32-41) — it normalizes hyphens/underscores. Not a bug.
  - ToolRegistry::all(): pre-existing method, returns Vec<Arc<dyn Tool>>, properly filters by engine version. No race conditions from the
   old two-call pattern.                                                                                                                 
  - Tool::discovery_schema / discovery_summary defaults: both have sensible defaults (parameters_schema() / None), so existing
  implementors compile unchanged.                                                                                                        
  - Python default.py reordering: just moves ensure_working_messages / append_message / append_system_append earlier in the file.
  Net-zero byte change, no behavioral change — but mention it in the PR body since it shows up on review.                                
  - Per-action inventory fetches vs old per-action available_actions: performance neutral; both go through the same projection pipeline.
  Parallel dispatch is actually faster post-PR because inventory is fetched once and cloned into tasks.

henrypark133:
Replying to Nicks review (https://github.com/nearai/ironclaw/pull/2868#pullrequestreview-4166932999):

Thanks for the detailed pass. I went through each point against the current code and tests.

I agree there are a few real consistency / error-handling gaps here that we should tighten: `B2`, `B4`, `B5`, and likely the `B3` name-resolution cleanup while we are in that path. I also think some of the other items are better framed as follow-up improvements rather than blocker-level issues for this PR.

In particular, `D1-D4` seem directionally valid, but they are consistent with the current prompt/runtime experience rather than regressions introduced by this change. I’d prefer to keep this PR scoped to the callable-surface and snapshot correctness fixes, then follow up separately on prompt-refresh cadence, checkpoint message mutation robustness, suffix marker hardening, and prompt-hint noise if we want to revisit those behaviors.

I don’t think `B1` is a bug: latent providers are intentionally surfaced on the callable surface instead of duplicated in capability background prose, so we do not show the same tool information twice. And I don’t think `D5` holds as written.

My plan is to make the narrow correctness fixes here, and treat the valid `D*` notes as follow-up cleanup/design work rather than expanding this PR further.

src/bridge/router.rs
```diff
@@ -1084,6 +1085,8 @@ async fn execute_pending_gate_action(
             .and_then(|v| v.as_str())
             .and_then(ironclaw_engine::ValidTimezone::parse),
         thread_goal: Some(thread.goal.clone()),
+        available_actions_snapshot: None,
+        available_action_inventory_snapshot: None,
     };
```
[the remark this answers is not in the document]
henrypark133: Fixed in ce776b00. execute_pending_gate_action() now loads the current callable action inventory for the thread's active leases before calling execute_resolved_pending_action(), and only populates available_actions_snapshot / available_action_inventory_snapshot when that inventory fetch succeeds. That brings pending-gate resume onto the same snapshot-backed path as the orchestrator and restores snapshot-based canonical name plus tool_info behavior on resume. I also added caller-level coverage in execute_pending_gate_action_populates_snapshots_for_tool_info_resume to pin the resumed tool_info path through the router.

crates/ironclaw_engine/src/executor/structured.rs
```diff
@@ -104,13 +112,11 @@ pub async fn execute_action_calls(
         };
 
         // 2. Find the action definition and check policy
-        let action_def = effects
-            .available_actions(std::slice::from_ref(&lease), context)
-            .await?
-            .into_iter()
-            .find(|a| action_name_matches(&a.name, &call.action_name));
+        let action_def = available_actions
```
serrrfirat: **High Severity** — callable-only inventory is not enforced here

`ActionInventory.inline` is documented as the set of actions that are callable *now*, while `discoverable` actions are explicitly "not callable yet" and should only be usable via `tool_info` / `tool_activate`.

In this preflight path, though, a missing inline match is not treated as an error. If `action_def` is `None`, the code falls through, later `find_and_consume()` still succeeds on any matching lease, and `execute_action()` is still called with the requested action name.

That makes the new inventory split fail open: a blocked/discoverable action can still execute as long as the thread retains a lease for it. This is especially risky because tool leases are refreshed additively elsewhere, so a tool that used to be inline can remain executable after it is demoted behind `tool_activate`.

Please reject calls that are absent from the current inline snapshot before lease consumption / execution, and add a regression test that a previously-inline tool becomes non-callable once it moves to the discoverable/activatable surface.

tests/e2e/scenarios/test_v2_auth_oauth_matrix.py
```diff
@@ -1567,6 +1590,7 @@ async def test_settings_first_gmail_auth_then_chat_runs(
     await chat_input.press("Enter")
 
     thread_id = await _current_thread_id(page)
+    await _wait_for_tool_call(server["base_url"], thread_id, "gmail", timeout=60.0)
     tokens = await _wait_for_mock_google_tokens(server["mock_api_url"], timeout=60.0)
```
serrrfirat, line 1617: **Medium Severity** — this replacement coverage no longer proves the auth gate blocks before any provider request is sent

The deleted `test_v2_kernel_auth_preflight.py` had the load-bearing assertion that the mock provider saw **zero** requests before credentials were supplied. This new flow only asserts that a tokened request eventually happens *after* auth succeeds.

Because the mock now only records requests once a valid `Authorization: Bearer ...` header is present, a regression that sends a pre-auth request first (for example an unauthenticated `401` probe before the gate pauses) would no longer fail this suite.

Please add a replacement assertion that the provider receives no requests at all before the auth flow completes, so the old preflight boundary stays covered.

src/bridge/effect_adapter.rs
```diff
@@ -286,6 +260,280 @@ impl EffectBridgeAdapter {
...
+        lookup_name: &str,
+        context: &ThreadExecutionContext,
+    ) -> Option<PermissionState> {
+        let db = self.tools.database()?;
+        match db.get_all_settings(&context.user_id).await {
+            Ok(db_map) => {
+                let settings = crate::settings::Settings::from_db_map(&db_map);
+                let normalized_name = lookup_name.replace('-', "_");
+                settings
+                    .tool_permissions
```
serrrfirat: **High Severity**

This switched the v2 approval path from `effective_permission(...)` to a raw lookup of only persisted overrides. That drops the `TOOL_RISK_DEFAULTS` floor entirely for users who have not explicitly saved a per-tool setting.

Concrete regression: `restart` is `AskEachTime` in `TOOL_RISK_DEFAULTS`, but `RestartTool::requires_approval()` is the default `Never` because the web `/restart` command owns its own modal. With this change, a v2 thread can call `restart` without any approval pause unless the user happened to persist an override first. The same issue affects any other tool whose safety floor currently comes from `effective_permission()` rather than `requires_approval()`.

Please keep the alias handling here, but compute the permission with `effective_permission(normalized_name, &settings.tool_permissions)` instead of reading only the override map. I’d also add a regression test proving that the v2 path still pauses on a default-`AskEachTime` tool like `restart` when the user has no explicit override.

tests/e2e/scenarios/test_v2_tool_activate_surface.py
```diff
@@ -0,0 +1,266 @@
+"""E2E test: v2 tool activation surface.
```
serrrfirat, line 1: **Medium Severity**

This new E2E covers the prompt contract well, but it replaces the deleted `test_v2_kernel_auth_preflight.py` without preserving its highest-value assertion: that the auth/preflight gate blocks the outbound request **before** any upstream HTTP call is made.

Right now the surviving scenarios still check that auth cards / activation UI appear and that the flow succeeds after enablement, but I don't see a zero-request assertion on the blocked phase. That leaves a regression hole where a future change could still leak unauthenticated traffic (or burn rate-limited upstream calls) before the gate pauses, and this suite would stay green.

Please keep a dedicated preflight assertion somewhere in this replacement coverage — e.g. reset a mock upstream, trigger the blocked integration, wait for the pending gate/UI surface, and assert that the upstream request log is still empty until auth/activation resolves.

nickpismenkov (approved):

PR #2871, opened 2026-04-22
feat(tools): ExternalToolRegistrar plugin seam for downstream forks

pranavraja99:
## Summary

Adds a `ExternalToolRegistrar` trait + `AppBuilder::with_external_tool_registrar` hook so downstream IronClaw forks can register their own Rust-level \`Tool\` implementations alongside the built-ins without patching upstream source.

## Motivation

The private \`nearai/ironclaw-abound\` fork ships Abound-specific remittance tools (\`abound.rs\`, \`forex.rs\`) that cannot live upstream. Without this seam, the fork maintains two permanent patches against \`src/tools/builtin/mod.rs\` and \`src/tools/registry.rs\` and re-resolves conflicts on every upstream change to those files. With it, the fork's binary supplies a trait impl and upstream stays unmodified.

## Changes

- **New**: \`src/tools/external_registrar.rs\` — \`ExternalToolRegistrar\` trait with a single \`register(&Arc<ToolRegistry>)\` method and a documented contract.
- **Modified**: \`src/tools/mod.rs\` — module declaration + \`pub use\`.
- **Modified**: \`src/app.rs\` — \`AppBuilder\` gains an \`Option<Arc<dyn ExternalToolRegistrar>>\` field + builder method. \`init_tools()\` invokes the registrar after \`register_builtin_tools()\` and the other builtin-registration passes, so the plugin sees a fully initialized registry.

Existing \`ToolRegistry\` accessors (\`secrets_store()\`, \`credential_registry()\`, \`database()\`, \`role_lookup()\`, \`rate_limiter()\`, \`engine_version()\`) are already public, so no additional API surface is exposed.

## Test plan

- [x] \`cargo check --all-features\` clean
- [x] New unit test: \`external_tool_registrar_registers_tool_into_registry\` (verifies a registrar's tool ends up in the registry).
- [ ] Downstream smoke test: private fork's binary supplies a registrar → the abound tools appear in \`/api/tools\`.

## Follow-ups

- WASM and MCP extension paths are untouched — they have their own runtime discovery. This seam is specifically for in-process Rust tools.
- Future: expose \`ToolRegistry::mission_slot()\` once the mission framework PR lands (needed by \`abound_send_wire\` for async wait actions).

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -121,6 +121,11 @@ pub struct AppBuilder {
 
     // Backend-specific handles needed by secrets store
     handles: Option<crate::db::DatabaseHandles>,
+
+    // Optional plugin hook for downstream binaries that register their own
+    // Rust `Tool` implementations alongside the built-ins. See
+    // `src/tools/external_registrar.rs` for the contract.
+    external_tool_registrar: Option<Arc<dyn crate::tools::ExternalToolRegistrar>>,
 }
 
 impl AppBuilder {
@@ -146,6 +151,7 @@ impl AppBuilder {
             secrets_store: None,
             llm_override: None,
             handles: None,
+            external_tool_registrar: None,
         }
     }
 
@@ -180,6 +186,21 @@ impl AppBuilder {
     /// Inject a pre-created LLM provider, skipping `init_llm()`.
     pub fn with_llm(&mut self, llm: Arc<dyn LlmProvider>) {
         self.llm_override = Some(llm);
+    }
+
+    /// Attach a plugin that registers additional Rust-level `Tool`
+    /// implementations alongside the built-ins. Called once during
+    /// `init_tools()` after `register_builtin_tools()` and after all
+    /// credential/DB/interceptor accessors on the registry are populated.
+    ///
+    /// Intended for downstream forks (e.g. `nearai/ironclaw-abound`) that
+    /// ship integration-specific tools that cannot live upstream. See
+    /// `src/tools/external_registrar.rs` for the full contract.
+    pub fn with_external_tool_registrar(
+        &mut self,
+        registrar: Arc<dyn crate::tools::ExternalToolRegistrar>,
+    ) {
+        self.external_tool_registrar = Some(registrar);
     }
 
     /// Phase 1: Initialize database backend.
@@ -538,6 +559,17 @@ impl AppBuilder {
         tools.register_builtin_tools();
         tools.register_tool_info();
         tools.register_system_tools();
+
+        // Downstream plugin seam: a fork that bundles its own Rust tools
+        // (e.g. `nearai/ironclaw-abound`) supplies an
+        // `ExternalToolRegistrar` here and registers its tools into the
+        // same `Arc<ToolRegistry>`. Upstream has no way to discover Rust
+        // tools at runtime, so this seam is the only non-patch extension
+        // point for in-process Rust tools. WASM + MCP extensions are
+        // unaffected.
+        if let Some(ref registrar) = self.external_tool_registrar {
+            registrar.register(&tools);
+        }
 
         if let Some(ref ss) = self.secrets_store {
             tools.register_secrets_tools(Arc::clone(ss));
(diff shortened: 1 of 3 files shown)


henrypark133 (commented):
What looks good:
- The seam is narrow and keeps the change out of most of the tool stack.
- The downstream-fork use case is clearly scoped.
- Wiring this through `AppBuilder` is a reasonable composition point.

Findings:
1. Medium - `src/tools/external_registrar.rs:50-52`, `src/tools/registry.rs:276-281`: the contract says protected-name collisions are rejected, but `register_sync` is currently a blind `insert`. Why it matters: a downstream registrar can silently shadow built-ins even though the seam documents that this should fail loudly. Expected fix direction: either add a dedicated external-registration path that rejects duplicate/protected names, or harden `register_sync` before documenting that guarantee.

Summary:
- Recommended verdict: Comment
- Prior feedback status: unresolved review comments remain
- Review coverage: checked-out diff plus current review-thread state
- Residual risk: registrar invocation still happens mid-`init_tools()`, so later built-in registrations can also silently win name collisions

src/tools/external_registrar.rs
```diff
@@ -0,0 +1,68 @@
...
+//!   values when the host binary configured them.
+//! - Must be deterministic and side-effect-free beyond tool registration.
+//!   Do not spawn long-lived tasks or hold network connections from inside
+//!   `register()`; that belongs in activation, not registration (see
+//!   `.claude/rules/lifecycle.md`).
+//! - External tool names are not added to [`ToolRegistry::is_protected_tool_name`].
+//!   A downstream `Tool::name()` that collides with a protected built-in is
+//!   rejected by `register_sync` — keep external names distinct.
+
+use std::sync::Arc;
```
serrrfirat, line 54: **High Severity** — Contract claim is factually false: `register_sync` does not check protected names

The doc says: _"A downstream `Tool::name()` that collides with a protected built-in is rejected by `register_sync` — keep external names distinct."_

Reading `register_sync` in `src/tools/registry.rs` (lines ~322-340), it has **no** `PROTECTED_TOOL_NAMES` check — only the async `register()` does. So:

- A registrar that registers `"echo"`, `"json"`, etc. (already inserted by `register_builtin_tools`) **silently overwrites** the built-in via `HashMap::insert`.
- A registrar that registers `"secrets_get"` (registered _after_ the registrar at `app.rs` ~line 576) **gets silently overwritten** by the built-in.

In both directions there is no warning, no error, no log. The doc creates a false sense of defense-in-depth that operators may rely on.

**Suggested fix:** either add a `PROTECTED_TOOL_NAMES` + `builtin_names` check inside `register_sync` (matching the async `register()` semantics), or rewrite this paragraph to say _"`register_sync` silently overwrites — keep external tool names distinct from anything `register_builtin_tools`, `register_secrets_tools`, `register_memory_tools_with_resolver`, etc. registers."_

src/app.rs
```diff
@@ -1608,6 +1640,78 @@ mod tests {
...
+    impl crate::tools::ExternalToolRegistrar for CaptureRegistrar {
+        fn register(&self, registry: &Arc<crate::tools::ToolRegistry>) {
+            registry.register_sync(Arc::new(DummyTool {
+                name: self.tool_name,
+            }));
+        }
+    }
+
+    #[tokio::test]
+    async fn external_tool_registrar_registers_tool_into_registry() {
```
serrrfirat, line 1693: **High Severity** — Test does not exercise the actual `init_tools` wiring (violates "test through the caller" rule)

CLAUDE.md is explicit about this: _"When a predicate/classifier/transform helper gates a side effect ... and has any wrapper or computed input between it and that side effect, a unit test on the helper alone is not sufficient regression coverage. Add a test that drives the actual call site."_

This test calls `registrar.register(&registry)` directly. It does **not** drive `init_tools()`. The author explicitly notes _"the grep for `external_tool_registrar` in `init_tools` backs up the compile-time coverage."_ A grep is not a test. The exact failure modes the author wants to guard against — someone renaming the field, gating the call with `#[cfg(...)]`, or moving the invocation order — would all leave this test green while silently breaking the downstream fork.

**Suggested fix:** add an integration test that builds an `AppBuilder`, calls `with_external_tool_registrar`, runs `init_tools`, and asserts the registrar-supplied tool is in the resulting registry.

src/app.rs
```diff
@@ -539,6 +560,17 @@ impl AppBuilder {
...
 
+        // Downstream plugin seam: a fork that bundles its own Rust tools
+        // (e.g. `nearai/ironclaw-abound`) supplies an
+        // `ExternalToolRegistrar` here and registers its tools into the
+        // same `Arc<ToolRegistry>`. Upstream has no way to discover Rust
+        // tools at runtime, so this seam is the only non-patch extension
+        // point for in-process Rust tools. WASM + MCP extensions are
+        // unaffected.
+        if let Some(ref registrar) = self.external_tool_registrar {
+            registrar.register(&tools);
```
serrrfirat, line 571: **Medium Severity** — `register_sync` silently drops tools on lock contention

`ToolRegistry::register_sync` (in `src/tools/registry.rs`) does `if let Ok(mut tools) = self.tools.try_write() { ... }` — a failed `try_write` **silently drops the tool** with no error path, no log, no panic.

The contract says the registrar runs once at startup, so contention is unlikely in the documented path. But the no-error behavior is a footgun:

- If a future caller invokes the registrar later in the lifecycle (hot-reload, reconfigure) the tools could be silently lost.
- If a downstream `register()` impl spawns a task that takes a read lock before returning, the next `register_sync` in the same registrar could fail.
- The current behavior is invisible — no operator-actionable signal that registration failed.

**Suggested fix:** at minimum, log a `warn!` when `try_write` fails so the symptom is observable. Better: have `register_sync` panic on contention at startup (this is a programming error, not a runtime condition), or have the registrar use the async `register()` from a `tokio::runtime::Handle::block_on` so contention waits.

serrrfirat:
## Paranoid PR Review Summary

Reviewed as a paranoid architect. Posted 3 inline findings (2 High, 1 Medium). Headline issues:

**1. Contract claim is factually false** ([inline](https://github.com/nearai/ironclaw/pull/2871#discussion_r3149584724)). The doc says `register_sync` rejects protected-name collisions; it does not. Both `register_builtin_tools → registrar` and `registrar → register_secrets_tools` overwrite paths are silent. Either fix `register_sync` to match the documented contract, or rewrite the contract paragraph to reflect the silent-overwrite reality and enumerate the post-registrar tool families that will overwrite the registrar (`register_secrets_tools`, `register_memory_tools_with_resolver`, image/vision, mission/routine, etc.).

**2. Test does not exercise the actual wiring** ([inline](https://github.com/nearai/ironclaw/pull/2871#discussion_r3149678907)). This is a direct violation of the project rule codified in `CLAUDE.md` ("Test through the caller, not just the helper"). The author acknowledges the test does not cover the call site and offers a `grep` as the substitute; that is exactly the situation the rule was written against. Add an integration test that drives `init_tools()`.

**3. Lock contention is invisible** ([inline](https://github.com/nearai/ironclaw/pull/2871#discussion_r3149680414)). `register_sync` silently drops tools on `try_write` failure. Unlikely to fire in the documented path, but cheap to log.

### v2 Engine Impact

**Not related.** This seam is scoped to the v1 `Tool` trait + `ToolRegistry`. The v2 `CapabilityRegistry` is a distinct abstraction (typed actions on resources) and would need its own seam if downstream forks ever want to inject v2 capabilities; that is out of scope for this PR.

### Coordinate with #2925

Both PRs insert code into `init_tools` immediately after `register_system_tools()`. Whichever lands second will rebase-conflict. The natural order is (1) `INTEGRATION_CREDENTIALS_DIR` loader from #2925, then (2) external registrar from this PR, then (3) `register_secrets_tools` — so registrar-supplied tools can use the credential mappings loaded in step 1.

### Other

- The usage example on lines 36-38 of `external_registrar.rs` shows chained `AppBuilder::new(...).with_external_tool_registrar(...)`, but `with_external_tool_registrar` takes `&mut self` and returns `()`, so that wont compile. Fix the example or change the method to return `&mut Self`.
- The contract paragraph "after all the registrys builder-injected dependencies" is true for the builder-injected accessors (`with_database`/`with_credentials`/`with_http_interceptor`), but the registrar runs before `register_secrets_tools` and the memory/skill/job/mission registrations. Worth calling out which tool families exist vs. dont when `register()` runs, since downstream tools that introspect the registry will get a partial view.

serrrfirat:
Thank you for documenting the downstream need for registering private in-process Rust tools without maintaining permanent upstream patches.\n\nBefore we close or replace this PR, we would like to confirm whether that downstream native-Rust capability requirement still exists. The v1 AppBuilder/ToolRegistry hook cannot be ported directly because Reborn routes capabilities through authorization, approvals, obligations, capability snapshots, the dispatcher, runtime lanes, and extension lifecycle.\n\nCurrent Reborn ownership references:\n- https://github.com/nearai/ironclaw/tree/main/crates/ironclaw_capabilities\n- https://github.com/nearai/ironclaw/tree/main/crates/ironclaw_dispatcher\n- https://github.com/nearai/ironclaw/tree/main/crates/ironclaw_extensions\n- https://github.com/nearai/ironclaw/tree/main/crates/ironclaw_reborn_composition\n\nIf the requirement is still active, the right follow-up is a Reborn-native issue for host-supplied native capability factories that enter through the normal capability host and dispatcher. Please let us know whether the downstream use case remains active; we would be glad to preserve the motivation and attribution in that follow-up.

PR #2872, opened 2026-04-22
feat(llm): add real-time token streaming through provider chain

rizgan:
Add streaming capability to the LLM provider system so clients receive tokens progressively rather than waiting for the full response.

## New provider

OpenAiCompatStreamingProvider (src/llm/openai_compat_stream.rs):
- Wraps any OpenAI-compatible backend (OpenRouter, Groq, etc.)
- Sends POST with stream: true and stream_options.include_usage: true
- Parses SSE data: lines, accumulates delta content and tool_calls
- Implements both complete_stream and complete_with_tools_stream
- Falls back gracefully to the inner provider for non-streaming paths
- Used by the registry factory for all openai_compat backends

## Trait changes (src/llm/provider.rs)

Two new default methods on LlmProvider:
- complete_stream(request, on_chunk) -- default: single-chunk fallback
- complete_with_tools_stream(request, on_chunk) -- same fallback

Default impls preserve backward compatibility for all existing providers.

## Provider chain delegation

All wrapper providers now forward streaming calls instead of falling back to the non-streaming default:

- RetryProvider         -- delegates directly (streams are not retried)
- SmartRoutingProvider  -- delegates to primary
- FailoverProvider      -- delegates to last-used provider
- CircuitBreakerProvider-- with full check_allowed/record_success/failure
- CachedProvider        -- bypasses cache (streaming takes priority)
- RecordingLlm          -- bypasses recording (chunks are not replayable)

## Wiring (src/llm/reasoning.rs, src/agent/dispatcher.rs)

ReasoningContext gains an optional chunk_sender field. ChatDelegate::call_llm() spawns a forwarder task that reads from the channel and calls channels.send_status(StreamChunk), which the web gateway broadcasts as SSE stream_chunk events to connected clients.

## Summary

-

## Change Type

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] CI/Infrastructure
- [ ] Security
- [ ] Dependencies

## Validation

- [ ] `cargo build`
- [ ] Relevant tests pass: 
- [ ] Manual testing: 

## Review Follow-Through

---

**Review track**:

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,7 @@
+[net]
+retry = 10
+
+[http]
+timeout = 300
+multiplexing = false
+
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -32,7 +32,7 @@ version = "0.5.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d122413f284cf2d62fb1b7db97e02edb8cda96d769b16e443a4f6195e35662b0"
 dependencies = [
- "crypto-common",
+ "crypto-common 0.1.7",
  "generic-array",
 ]
 
@@ -44,7 +44,7 @@ checksum = "b169f7a6d4742236a0a00c541b845991d0ac43e546831af1249753ab4c3aa3a0"
 dependencies = [
  "cfg-if",
  "cipher",
- "cpufeatures",
+ "cpufeatures 0.2.17",
 ]
 
 [[package]]
@@ -63,9 +63,9 @@ dependencies = [
 
 [[package]]
 name = "agent-client-protocol"
-version = "0.10.2"
+version = "0.10.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9c56a59cf6315e99f874d2c1f96c69d2da5ffe0087d211297fc4a41f849770a2"
+checksum = "10eeef5e80864f9c3c148a3f395c3e35a66d37ec7561c7845b2bffae8e841759"
 dependencies = [
  "agent-client-protocol-schema",
  "anyhow",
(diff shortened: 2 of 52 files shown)


rizgan:
This PR fixes a silent streaming regression where all LLM responses were delivered
as a single chunk instead of token-by-token, even when the provider supports SSE streaming.

**Root cause:** The provider chain wrappers (RetryProvider → SmartRoutingProvider →
FailoverProvider → CircuitBreakerProvider → CachedProvider → RecordingLlm) did not
override `complete_stream` / `complete_with_tools_stream`, so every call fell through
to the default single-chunk fallback in the trait.

**What this PR does:**
- Adds `complete_stream` and `complete_with_tools_stream` default methods to `LlmProvider`
  (backward-compatible fallback: delivers the full response as one chunk)
- Adds `OpenAiCompatStreamingProvider` — a streaming-capable wrapper for OpenAI-compatible
  backends (OpenRouter, Groq, etc.) that sends `stream: true` and parses SSE deltas
- Wires all six wrapper providers to delegate streaming calls instead of silently falling back
- Adds `chunk_sender` to `ReasoningContext` and updates both LLM call sites in
  `respond_with_tools()` to use streaming when a sender is present
- `ChatDelegate::call_llm()` spawns a forwarder task that converts received chunks into
  `StatusUpdate::StreamChunk` events, which the web gateway broadcasts as SSE to clients

Tested with OpenRouter (`openai/gpt-4o-mini`) via a Telegram bot — tokens now appear
progressively in the chat.

serrrfirat:

True token streaming is required, but it is already implemented end-to-end in Reborn: providers stream through the model gateway, text becomes `ModelTextDelta`, and the Reborn OpenAI-compatible layer emits `response.output_text.delta` and `response.output_text.done` SSE events.

The best follow-up contribution is stronger Reborn regression coverage proving multiple incremental deltas arrive before completion for both Chat Completions and Responses API paths, including tool-call and retry edge cases.

serrrfirat:

True streaming is already implemented end-to-end in Reborn: model gateway deltas become `ModelTextDelta` and OpenAI-compatible SSE emits `response.output_text.delta` and `.done`.

The valuable follow-up is stronger Reborn E2E coverage proving multiple incremental deltas arrive before completion for Responses and Chat APIs.

PR #2873, opened 2026-04-22
feat(tools): mission tool family (engine v2)

pranavraja99:
## Summary

Exposes the engine-v2 mission framework to the LLM via seven built-in tools:
\`mission_create\`, \`mission_list\`, \`mission_update\`, \`mission_delete\`,
\`mission_fire\`, \`mission_pause\`, \`mission_resume\`.

Also adds \`MissionSlot\` — a deferred-injection handle on \`ToolRegistry\` — so tools whose registration pre-dates the engine's \`MissionManager\` construction (typically downstream integration tools plugged in through [#2871 \`ExternalToolRegistrar\`]) can read the manager through a shared \`Arc\` at execute time.

## Changes

- **New**: \`src/tools/builtin/mission.rs\` (+705 LOC) — seven \`Tool\` impls. Cadence strings parse into \`MissionCadence\` (\`cron\`, \`event:\`, \`webhook:\`, \`manual\`).
- **Modified**: \`src/tools/builtin/mod.rs\` — module declaration + \`pub use\` of the seven tools.
- **Modified**: \`src/tools/registry.rs\`:
  - \`PROTECTED_TOOL_NAMES\` gains the seven \`mission_*\` entries.
  - New \`pub type MissionSlot\` + \`mission_slot: MissionSlot\` field on \`ToolRegistry\` + \`mission_slot()\` accessor + \`set_mission_slot(manager, project_id)\` populator.
  - \`register_mission_tools(manager, project_id)\` instantiates all seven tools against the supplied manager and fills the slot.
- **Modified**: \`src/bridge/router.rs\` \`init_engine()\` — calls \`agent.tools().register_mission_tools(...)\` once \`MissionManager\` is constructed, immediately before the bootstrap/resume calls.

## Motivation

Two use cases:

1. **Upstream agent functionality**: gives the engine-v2 LLM a way to manage missions conversationally, replacing the previous split where mission creation required the host code to call \`MissionManager\` directly.
2. **Downstream forks**: \`nearai/ironclaw-abound\` (private) has an \`abound_send_wire\` wait-action that needs to create an hourly rate-monitoring mission. With this PR and [#2871], the fork's registrar can register integration tools that read \`MissionManager\` through \`MissionSlot\` at execute time.

## Test plan

- [x] \`cargo check --all-features\` / \`cargo check --all-features --tests\` clean
- [x] \`cargo fmt --all -- --check\` clean
- [ ] End-to-end: hit \`mission_create\` from the LLM, confirm the mission persists and fires on its cron.

## Follow-ups

- Full LLM-facing cadence parser tests (currently coverage is limited to the manager's own tests, not the tool layer).

### DIFF
--- a/src/bridge/router.rs
+++ b/src/bridge/router.rs
@@ -1826,6 +1826,15 @@ pub async fn init_engine(agent: &Agent) -> Result<(), Error> {
     mission_manager_inner =
         mission_manager_inner.with_insights_interval(missions_config.insights_interval);
     let mission_manager = Arc::new(mission_manager_inner);
+    // Register the engine-v2 mission tool family now that the manager
+    // exists. Tools constructed here capture `Arc::clone(&mission_manager)`
+    // directly; the call also fills `ToolRegistry::mission_slot` so any
+    // downstream integration tool registered through the plugin seam
+    // (see `ExternalToolRegistrar`) sees the same manager.
+    agent
+        .tools()
+        .register_mission_tools(Arc::clone(&mission_manager), project_id)
+        .await;
     if let Err(e) = thread_manager.recover_project_threads(project_id).await {
         debug!("engine v2: recover_project_threads failed: {e}");
     }
--- a/src/tools/builtin/mission.rs
+++ b/src/tools/builtin/mission.rs
@@ -1,0 +1,710 @@
+//! LLM-facing tools for managing missions (engine v2).
+//!
+//! Six tools let the agent manage missions conversationally:
+//! - `mission_create` - Create a new long-running mission
+//! - `mission_list` - List all missions with status
+//! - `mission_update` - Modify a mission's settings
+//! - `mission_delete` - Remove a mission
+//! - `mission_fire` - Manually trigger a mission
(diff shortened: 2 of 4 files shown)


henrypark133:
Hi Pranav,
Is this urgent change needed to be merged? Otherwise, I will work on this as part of: https://github.com/nearai/ironclaw/issues/2767#issuecomment-4293374615

So we don't add these into built-in tools and keep engine actions in its own bucket.

henrypark133:
Thanks for putting this together. Closing this one because the engine-v2 mission surface is already wired through the bridge and capability path upstream, so this PR now overlaps with an implementation that already exists in the current codebase.

At this point, landing a second built-in-tool path for the same mission lifecycle would leave us maintaining two integration shapes for the same user-facing capability. I would rather keep one canonical mission path in v2 and avoid splitting discovery and execution between bridge actions and built-in tools.

The downstream need here still makes sense, especially around MissionManager access for fork-specific integrations, but I think that should build on the existing canonical mission surface rather than introduce a parallel one. Thanks again for pushing this forward.

PR #2889, opened 2026-04-23
[codex] Finish engine v2 deferred action inventory cleanup

henrypark133:
## Context

This PR is the Section 7 slice from the `#2767` stack plan:

- issue: https://github.com/nearai/ironclaw/issues/2767
- planning comment: https://github.com/nearai/ironclaw/issues/2767#issuecomment-4293374615
- section: `7. Deferred large-tool baseline handling`

That section set the baseline goal as:

- keep very large tools visible without carrying all of their schemas in the default provider-callable surface
- surface those tools through a deferred/discovery-oriented path
- keep `tool_info` as the richer metadata/schema fetch path
- avoid duplicating the same tool facts across provider schemas, prompt prose, capability background, and `tool_list`

This branch now stacks on `v2-engine-action-discovery-metadata`.

### 1. Adds explicit action-inventory plumbing for engine v2

- introduces `ActionInventory` into the engine effect interface and thread execution context
- snapshots both the current callable action list and the per-step inventory view for the orchestrator / bridge path
- teaches the bridge `tool_info` path to read from that inventory snapshot instead of relying only on the older action snapshot

### 2. Separates callable schemas from richer discovery metadata

- keeps the engine-v2 callable tool surface compact when a tool intentionally has richer discovery metadata than its default inline callable schema
- preserves the `tool_info(...)` hint for those compact-schema tools instead of inlining the full discovery schema into every v2 LLM request
- keeps `tool_info` as the on-demand richer schema/details path for those tools

### 3. Finishes the bridge-side deferred inventory groundwork

- adds the bridge-side inventory/projection layer needed to reason about callable vs deferred/discovery-oriented tool visibility
- updates the gate/integration coverage so a callable tool stays callable across `tool_info` inspection rather than depending on the old promotion behavior
- keeps capability/background state separate from the callable tool surface

### 4. Cleans up follow-on issues surfaced while landing the stack item

- updates prompt / orchestrator / bridge wiring to use the new inventory path consistently
- fixes follow-up extension-manager and shared-env test issues that surfaced while validating the branch
- carries the rebase fixups needed to sit cleanly on top of `v2-engine-action-discovery-metadata`, including its later clippy follow-up

## Why

The engine-v2 tool surface had drifted into an awkward state:

- some tools had richer discovery metadata than their inline callable schemas
- the bridge/orchestrator path needed a real per-step inventory view for `tool_info`
- earlier follow-up changes briefly regressed the v2 path by serializing full discovery schemas into every tool definition

This PR fixes that by keeping the three concerns separate:

- callable tool definitions stay on the normal provider/tool surface
- richer discovery/schema details stay behind `tool_info`
- capability/runtime truth stays on the capability/background surface

## Relationship To Section 7

Section 7 described the baseline as deferred large-tool handling.

This PR implements the engine/bridge plumbing for that baseline and preserves the important review-driven constraint that came out of the stack work:

- large or discovery-heavy tools should not force their full schemas into every v2 tool-definition payload
- but the v2 path also should not regress into a broken “invisible until promotion” model

So the resulting contract here is:

- callable tools remain callable
- compact callable schemas remain compact on the v2 tool-definition path
- richer discovery metadata remains available through `tool_info`
- duplicated always-on surfacing is avoided

## Validation

- `cargo test --lib bridge::action_projector::tests::project_tool_action_preserves_discovery_metadata -- --exact`
- `cargo test --lib bridge::llm_adapter::tests::action_def_to_tool_def_preserves_tool_info_hint_for_discovery_metadata -- --exact`
- `cargo test --test engine_v2_gate_integration tool_info_does_not_gate_callable_tool_into_next_llm_callable_set`

## Notes

- base branch: `v2-engine-action-discovery-metadata`
- this branch was rebased onto the latest tip of that branch before opening the PR


src/bridge/action_inventory.rs
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+    pub(crate) fn provider_backed(action: ActionDef) -> Self {
+        Self { action }
+    }
+
+    pub(crate) fn engine_native(action: ActionDef) -> Self {
+        Self { action }
+    }
+}
```
serrrfirat: **[Architecture / Medium]** `InventoryCandidate::{builtin, provider_backed, engine_native}` are three identical constructors — each collapses to `Self { action }`, and `ActionInventoryPlanner::plan` below discards the origin tag entirely. The three-way ceremony reads like a placeholder for a `deferred`/provider-backed variant that the follow-up commit (3b7616c) removed.

Two paths forward:

1. **If deferred inventory is landing next:** add a `origin: InventoryOrigin` field to `InventoryCandidate` and propagate it into `ActionDef`/`ActionInventory` so the tag isn't thrown away. Leave a `// TODO: wire origin into ActionInventory` here naming the follow-up PR.
2. **If it's not landing soon:** delete `InventoryCandidate` and the planner, have `project_candidates` return `Vec<ActionDef>` directly, and let the caller sort and wrap. Right now this file adds indirection with zero behavioral payoff — any reviewer reading it gets the false impression that origin matters downstream.

crates/ironclaw_engine/src/executor/orchestrator.rs
```diff
@@ -982,12 +982,13 @@ async fn handle_execute_action(
...
+    let inventory = effects
+        .available_action_inventory(&active_leases, &exec_ctx)
         .await
-        .unwrap_or_default()
-        .into();
-    exec_ctx.available_actions_snapshot = Some(Arc::clone(&available_actions));
+        .unwrap_or_default();
+    let available_actions = inventory.inline.clone();
+    exec_ctx.available_actions_snapshot = Some(available_actions.clone());
+    exec_ctx.available_action_inventory_snapshot = Some(inventory.clone());
```
serrrfirat: **[Performance / Medium]** This PR changes `available_actions_snapshot` from `Arc<[ActionDef]>` to `Vec<ActionDef>` (see `traits/effect.rs`) and adds `available_action_inventory_snapshot: Option<ActionInventory>` where `ActionInventory` is also backed by a `Vec`. The snapshot assignments at lines 990-991, 1389, 1558-1559, and 1594-1595 now perform deep `Vec` clones (plus nested `ActionDef`/`JsonValue` schema clones) on every tool call and every parallel task, where the v1 path did `Arc::clone` refcount bumps.

For CodeAct prompts with large tool catalogs run through `handle_execute_actions_parallel`, this is an O(N·M) regression (N tasks × M actions × schema size) vs the prior O(N) refcount path.

Suggested fix: wrap the inventory contents in `Arc<[ActionDef]>` (or `Arc<ActionInventory>`) and store `Option<Arc<...>>` on `ThreadExecutionContext`, so per-task snapshot assignment stays a pointer bump. `ActionInventoryPlanner::plan` is the natural place to freeze the `Arc`.

Gemini bot flagged the same concern on the `traits/effect.rs` change — this comment is about the blast radius on the orchestrator hot paths.

src/bridge/effect_adapter.rs
```diff
@@ -984,36 +986,47 @@ impl EffectBridgeAdapter {
...
+                .available_action_inventory_snapshot
+                .as_ref()
+                .cloned()
+                .or_else(|| {
+                    context
+                        .available_actions_snapshot
+                        .as_ref()
+                        .cloned()
+                        .map(|actions| ActionInventory { inline: actions })
+                });
```
serrrfirat: **[Performance / Low]** Both the `available_action_inventory_snapshot` read and the fallback from `available_actions_snapshot` call `.cloned()`, producing an owned `ActionInventory` that is then handed to `ActionDiscovery::tool_info(&parameters, &inventory)` by reference. `tool_info` only borrows it — the owned clone is discarded right after.

```rust
let inventory = context
    .available_action_inventory_snapshot
    .as_ref()
    .cloned()  // <- unnecessary
    .or_else(|| {
        context
            .available_actions_snapshot
            .as_ref()
            .cloned()  // <- unnecessary
            .map(|actions| ActionInventory { inline: actions })
    });
```

The fallback branch does need to synthesize an `ActionInventory` from the legacy snapshot, but you can do it with borrowed slices (e.g. store a `Cow<'_, ActionInventory>` or restructure so the two branches produce `&ActionInventory`). For a large tool catalog this clone fires on every `tool_info` call.

crates/ironclaw_engine/src/traits/effect.rs
```diff
@@ -39,7 +37,9 @@ pub struct ThreadExecutionContext {
     ///
     /// Populated by the orchestrator when an execution path needs on-demand
     /// discovery parity (for example `tool_info`).
-    pub available_actions_snapshot: Option<Arc<[ActionDef]>>,
+    pub available_actions_snapshot: Option<Vec<ActionDef>>,
+    /// Snapshot of the full action inventory visible to the current step.
+    pub available_action_inventory_snapshot: Option<ActionInventory>,
```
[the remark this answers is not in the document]
henrypark133: Fixed in de1f77cb. `ThreadExecutionContext` now stores both snapshots behind `Arc` again (`Option<Arc<[ActionDef]>>` and `Option<Arc<ActionInventory>>`), so cloning the context stays cheap on the orchestrator path.

crates/ironclaw_engine/src/executor/orchestrator.rs
```diff
@@ -1590,7 +1591,8 @@ async fn handle_execute_actions_parallel(
             let effects = effects.clone();
             let lease = lease.clone();
             let mut exec_ctx = thread_execution_context(thread, step_id, Some(pc_call_id.clone()));
-            exec_ctx.available_actions_snapshot = Some(Arc::clone(&available_actions));
+            exec_ctx.available_actions_snapshot = Some(available_actions.clone());
+            exec_ctx.available_action_inventory_snapshot = Some(inventory.clone());
```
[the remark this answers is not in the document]
henrypark133: Fixed in de1f77cb. The orchestrator now builds the inventory once per step, converts the callable action slice once, and uses `Arc::clone(...)` for per-call / parallel fanout snapshot assignment instead of deep-copying schemas for each runnable action.

src/bridge/action_inventory.rs
henrypark133: Took option 2 in de1f77cb. I deleted `InventoryCandidate` and `ActionInventoryPlanner`, changed `ActionProjector` to return `Vec<ActionDef>` directly, and only wrap into `ActionInventory` at the adapter boundary where that type is actually needed.

crates/ironclaw_engine/src/executor/orchestrator.rs
henrypark133: Fixed in de1f77cb. `ThreadExecutionContext` now stores `available_actions_snapshot` and `available_action_inventory_snapshot` as `Arc`-backed snapshots again, and the orchestrator hot paths switched back to pointer-bump `Arc::clone(...)` assignment instead of cloning the full action inventory per execution context.

src/bridge/effect_adapter.rs
henrypark133: Fixed in de1f77cb. The `tool_info` path now borrows snapshots via `as_deref()` and uses a borrowed `tool_info_from_actions(...)` fallback for the legacy action slice, so the temporary owned `ActionInventory` / action clones are gone.

Issue #1350, opened 2026-03-18
Support hot-reload of LLM provider without restart


italic-jinxin:
## Problem

   When users change the LLM provider or model via the web UI (Settings page), the change is persisted to the database but **does not take effect until the process is restarted**. This creates a poor UX — users expect the switch to be immediate.

   ## Root Cause

   The LLM provider is created once at startup in `AppBuilder::init_llm()` and stored as an immutable `Arc<dyn LlmProvider>` throughout the application:

   1. `build_provider_chain()` constructs the full decorated provider (retry, circuit breaker, caching, recording layers) once during boot
   2. The resulting `Arc` is passed into `AppComponents`, `AgentDeps`, `GatewayState`, and `Reasoning` — none of which support swapping
   3. The web settings handler (`settings_set_handler`) writes to DB and returns `204 No Content` with no callback or notification to trigger a reload

   ## Proposed Solution

   1. **Wrap the provider in a swappable cell** — e.g. `Arc<ArcSwap<dyn LlmProvider>>` or `Arc<RwLock<Arc<dyn LlmProvider>>>` so consumers can always read the latest provider
   2. **Add a reload trigger** — when the settings handler writes `llm_backend` or custom provider config, emit an internal event (e.g. via `tokio::sync::watch` or the existing event system) to rebuild the provider chain
   3. **Re-fetch per turn** — the `Dispatcher` should read from the swappable cell when creating each `Reasoning` instance, rather than cloning a static Arc at startup

   ### Key files to modify

   - `src/app.rs` — `AppComponents` field type + `init_llm()` to return a swappable handle
   - `src/llm/mod.rs` — `build_provider_chain()` extracted into a re-callable factory
   - `src/agent/dispatcher.rs` — read provider from swappable cell per turn
   - `src/channels/web/handlers/settings.rs` — trigger reload after LLM-related setting writes
   - `src/channels/web/server.rs` — `GatewayState.llm_provider` to use swappable reference

   ### Considerations

   - In-flight requests should complete with the old provider; only new turns pick up the new one
   - Provider construction can fail (bad API key, unreachable endpoint) — reload errors should be surfaced to the UI, not silently ignored
   - The cheap LLM (`cheap_llm`) and recording handle should also be swappable

   ## Current Workaround

   Users must restart the ironclaw process after changing LLM settings.

G7CNF:
[...]

PR #2890, opened 2026-04-23
fix(web): resolve empty “Fetch available models” result for NEAR AI in settings

### LINKED ISSUE #2881: [QA] Settings "Fetch available models" returns empty for NEAR AI while `/model` in chat works
### Environment

local (cloned ironclaw)

### Version / Commit Hash

d33fecb17c3e81404f021ee944aa75673d1b4f97

### QA Test Date

2026-04-23

### Feature Area

Setting - Inference - Model Provider

## Steps to reproduce

1. Configure NEAR AI as the active LLM provider (`LLM_BACKEND=nearai`).
2. Open the web gateway Settings page and open the NEAR AI provider config.
3. Click **↻ Fetch available models** — response is empty / `ok:false`.
4. Go back to Chat and type `/model` — the current model and full available
   model list are returned.

## Expected behavior

Both surfaces are advertised as "list available models" for the same
provider. They should return the same list (or at minimum both succeed or
both fail with the same error).

## Actual behavior

- Settings: empty list, toast shows `No models found in response` (or
  similar).
- Chat `/model`: full model list returned, current model marked `(active)`.

### Logs / Screenshots

_No response_

### Pre-submit checklist

italic-jinxin:
## Summary

Relax `is_nearai_private_endpoint` so regional subdomains of `private-chat-stg.near.ai` (e.g. `us.private-chat-stg.near.ai`) are classified as NEAR AI private endpoints, same as `private.near.ai` subdomains already are.

`is_nearai_private_endpoint` gates the `/v1` prefix insertion in the admin **List Models** handler ([`src/channels/web/handlers/llm.rs`](src/channels/web/handlers/llm.rs)):

- Public NEAR AI (`cloud-api.near.ai/v1`) — base URL already carries `/v1`, use as-is.
- Private NEAR AI — base URL has no `/v1`, handler must prepend it before hitting `/models`.

Before this change, `private.near.ai` matched **"exact host or any subdomain"**, but `private-chat-stg.near.ai` was only matched as an exact host. A regional staging host like `https://us.private-chat-stg.near.ai/` silently fell into the public branch, so List Models requested `/models` instead of `/v1/models` and returned 404.

This PR also removes the “restart instance” notification, since model provider hot reloading [#1350](https://github.com/nearai/ironclaw/issues/1350) has been resolved.

| Type | Screenshot |
|---|---|
| Stage Connect | test |
| Stage Dropdown | dropdown |
| Prod Dropdown | prod dropdown |
| Prod Connect | prod connect |

## Scope

This helper is only used by the admin-only endpoints:

- `POST /api/llm/test_connection` — does **not** call the helper; unchanged.
- `POST /api/llm/list_models` — calls the helper to decide whether to prepend `/v1` before `/models`; this is the only behavior affected.

The chat/LLM request path (`src/llm/` provider adapters) does not use this helper and is not affected.

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] CI/Infrastructure
- [ ] Security
- [ ] Dependencies

## Linked Issue

Fixes #2881 

## Validation

- [ ] `cargo build`
- [ ] Relevant tests pass: 
- [ ] Manual testing: 

## Review Follow-Through

---

**Review track**:


serrrfirat:
does this also add the dropdown functionality back?

src/channels/web/handlers/llm.rs
```diff
@@ -530,14 +530,20 @@ async fn resolve_api_key_from_secrets(
...
 fn is_nearai_private_endpoint(base_url: &str) -> bool {
+    const PRIVATE_HOSTS: &[&str] = &["private.near.ai", "private-chat-stg.near.ai"];
     url::Url::parse(base_url)
         .ok()
         .and_then(|u| u.host_str().map(|h| h.to_lowercase()))
-        .is_some_and(|host| host == "private.near.ai" || host.ends_with(".private.near.ai"))
+        .is_some_and(|host| {
+            PRIVATE_HOSTS
+                .iter()
+                .any(|root| host == *root || host.ends_with(&format!(".{root}")))
```
serrrfirat, line 592: **High Severity**

This broadens the private-host match, but `fetch_provider_models()` still unconditionally does `format!("{base}/v1")` for any NEAR AI private host. With this change, a configured staging URL like `https://us.private-chat-stg.near.ai/v1` now matches here and becomes `https://us.private-chat-stg.near.ai/v1/v1/models`, so **Fetch available models** still 404s for operators who include `/v1` in the base URL.

A minimal fix is to guard the NEAR AI branch the same way the Anthropic branch does before appending `/v1` (or parse the URL path and only append when the last segment is not `v1`).

src/channels/web/handlers/llm.rs
```diff
@@ -690,6 +696,20 @@ mod tests {
...
+    fn test_nearai_private_stg_exact_match() {
+        assert!(is_nearai_private_endpoint(
+            "https://private-chat-stg.near.ai/"
+        ));
+    }
+
+    #[test]
+    fn test_nearai_private_stg_subdomain() {
+        assert!(is_nearai_private_endpoint(
+            "https://us.private-chat-stg.near.ai/v1"
```
serrrfirat, line 761: **Medium Severity**

These new tests only exercise `is_nearai_private_endpoint()` in isolation. Per `.claude/rules/testing.md`, helper-only tests are not sufficient when a predicate gates a side effect. The regression here is in `fetch_provider_models()` deciding whether to call `/models` vs `/v1/models`; a future refactor could still emit the wrong URL while all of these helper tests stay green.

Please add caller-level coverage around the URL-construction path as well — for example by driving `fetch_provider_models()`/`llm_list_models_handler` through a test seam that can assert the request goes to `/v1/models` for the staging private host (including the `/v1`-suffixed variant).

italic-jinxin:
@serrrfirat verified stg and prod config, updated screenshots in pull request summary, add resolved comment. 🤝

crates/ironclaw_gateway/static/js/surfaces/config.js
```diff
@@ -259,9 +259,6 @@ function setActiveProvider(id) {
       renderProviders();
       loadInferenceSettings();
       scrollToProviders();
-      document.getElementById('config-restart-notice').style.display = 'flex';
-      var llmNotice = document.getElementById('llm-restart-notice');
-      if (llmNotice) llmNotice.style.display = 'flex';
       showToast(I18n.t('config.providerActivated', { name: id }));
```
serrrfirat, line 265: **High Severity**

Removing the restart warning here makes provider activation look atomic, but the actual transition is still split across two independent settings writes: `llm_backend` first and `selected_model` second. The settings handler hot-reloads the global LLM chain after **each** write (`src/channels/web/features/settings/mod.rs`), and config resolution gives `selected_model` precedence over provider defaults/overrides (`src/config/llm.rs`).

That means there is now a real window where the process can run the **new backend with the previous provider model**, and if the second request fails the instance stays stuck in that mixed state. Before this PR the restart notice at least hinted that the switch was not fully live yet; after this change we show a success toast immediately.

I think this should be made atomic before removing the warning, e.g. by saving backend+model in one request/transaction or by deferring reload until both writes are committed.

PR #2892, opened 2026-04-23
fix(config): trim whitespace from LLM base_url

### LINKED ISSUE #2886: bug: Provider Matrix (openai-compatible) fails — malformed LLM_BASE_URL
## Summary

`live_mission_tests::mission_daily_news_digest_with_followup` fails because the `LLM_BASE_URL` for the openai-compatible provider contains an invalid URI character (likely trailing whitespace), causing every LLM request to fail at URI construction time.

## Evidence

From CI run https://github.com/nearai/ironclaw/actions/runs/24814569246 (Provider Matrix openai-compatible job):

```
WARN ironclaw::llm::retry: Retrying after transient error (tools)
  provider=Qwen/Qwen3.5-122B-A10B attempt=1 max_retries=3 delay_ms=1029
  error=Provider Qwen/Qwen3.5-122B-A10B request failed: HttpError: Http error: invalid uri character
```

All 3 retries fail identically. The `http` crate's URI parser rejects the constructed URL before any network I/O. The engine returns "Something went wrong" and the test times out after 15 minutes.

The base URL comes from GitHub Actions variable `LIVE_OPENAI_COMPATIBLE_BASE_URL` (`.github/workflows/live-canary.yml`, line 614).

## Fix

1. **Immediate:** Trim the `LIVE_OPENAI_COMPATIBLE_BASE_URL` GitHub Actions variable value
2. **Defensive:** Add `.trim()` in `src/config/llm.rs` (~line 760-780) when resolving the base URL

## Relevant files

- `.github/workflows/live-canary.yml` (line 614)
- `src/config/llm.rs` (base URL resolution)
- `src/llm/rig_adapter.rs` (`map_rig_error`)

italic-jinxin:
## Summary

Provider Matrix (openai-compatible) live tests failing with `HttpError: invalid uri character` because `LLM_BASE_URL` carried trailing whitespace sourced from the `LIVE_OPENAI_COMPATIBLE_BASE_URL` GitHub Actions variable.

`url::Url::parse` (used by `validate_operator_base_url`) silently tolerates leading/trailing ASCII whitespace per the WHATWG URL spec, so config resolution succeeded. Downstream, `http::Uri` (used by `reqwest`/`rig`) is strict and rejected every request with `invalid uri character` — all three retries failed identically and never hit the network.

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] CI/Infrastructure
- [ ] Security
- [ ] Dependencies

## Linked Issue

Fixes #2886 

## Validation

- [ ] `cargo build`
- [ ] Relevant tests pass: 
- [ ] Manual testing: 

## Review Follow-Through

---

**Review track**:


serrrfirat:
Quick review note:\n\n**Medium** — the trim is only applied in the built-in registry-provider path (). The same trailing-whitespace ->  failure mode still looks reachable in a few sibling paths that feed request URLs:\n-  — \n-  — \n-  —  / \n-  — custom provider \n\nSo this fixes the provider-matrix case, but not the broader class of operator-pasted / DB-pasted URL whitespace issues. I think this wants either shared normalization in  / , or the same trim applied consistently across all resolver paths.\n\n**Low** — in , the shell side uses , which removes *all* whitespace, not just leading/trailing whitespace like Rust’s . That can silently rewrite malformed input instead of surfacing it.\n\nOverall I like the direction; I’d just tighten the normalization scope so the fix covers the full pattern, not only .

serrrfirat:
Quick review note (reposting cleanly because my previous shell-quoted comment mangled the code spans):

**Medium** — the trim is only applied in the built-in registry-provider path (`src/config/llm.rs:760-802`). The same trailing-whitespace -> `http::Uri` failure mode still looks reachable in a few sibling paths that feed request URLs:
- `src/config/llm.rs:314-331` — `NEARAI_AUTH_URL`
- `src/config/llm.rs:352-368` — `NEARAI_BASE_URL`
- `src/config/llm.rs:458-463` — `OPENAI_CODEX_AUTH_URL` / `OPENAI_CODEX_API_URL`
- `src/config/llm.rs:617-623` — custom provider `base_url`

So this fixes the provider-matrix case, but not the broader class of operator-pasted / DB-pasted URL whitespace issues. I think this wants either shared normalization in `validate_base_url` / `validate_operator_base_url`, or the same trim applied consistently across all resolver paths.

**Low** — in `.github/workflows/live-canary.yml`, the shell side uses `tr -d '[:space:]'`, which removes *all* whitespace, not just leading/trailing whitespace like Rust’s `.trim()`. That can silently rewrite malformed input instead of surfacing it.

Overall I like the direction; I’d just tighten the normalization scope so the fix covers the full pattern, not only `openai_compatible`.

italic-jinxin:
@serrrfirat resolved, plz review 🤝

src/config/helpers.rs
```diff
@@ -235,8 +235,16 @@ enum BaseUrlPolicy {
 ///
 /// This is intended for config-time validation of base URLs like
 /// `OLLAMA_BASE_URL`, `EMBEDDING_BASE_URL`, `NEARAI_BASE_URL`, etc.
-pub(crate) fn validate_base_url(url: &str, field_name: &str) -> Result<(), ConfigError> {
-    validate_base_url_with_policy(url, field_name, BaseUrlPolicy::StrictSsrf)
+pub(crate) fn validate_base_url(url: &str, field_name: &str) -> Result<String, ConfigError> {
+    // `url::Url::parse` tolerates leading/trailing ASCII whitespace per the
+    // WHATWG URL spec, but `http::Uri` (reqwest/rig) rejects it with
+    // "invalid uri character". Returning the trimmed URL forces every call
```
serrrfirat, line 241: **High Severity**\n\nThis helper contract change is only partially migrated. `validate_operator_base_url()` now returns the trimmed canonical URL, but some live callers still use it as a pure validator and keep the original untrimmed string.\n\nA concrete missed path is `src/config/embeddings.rs`:\n\n```rust\nvalidate_operator_base_url(&ollama_base_url, "OLLAMA_BASE_URL")?;\n```\n\n`ollama_base_url` is then passed unchanged into `OllamaEmbeddings::new(...)`, and that constructor does not normalize whitespace. So `OLLAMA_BASE_URL="http://localhost:11434 \n"` will now pass validation here but can still fail later at request construction with the same `invalid uri character` class of error this PR is fixing.\n\nPlease either update every remaining caller to capture the returned `String`, or keep validation and normalization as separate APIs so this kind of partial migration cannot silently compile.

src/config/llm.rs
```diff
@@ -1395,6 +1404,68 @@ mod tests {
...
 
+    #[test]
+    fn base_url_trims_surrounding_whitespace() {
+        // Regression for #2886: GitHub Actions repo variables and DB-pasted
+        // URLs can carry trailing newlines or spaces. `url::Url::parse`
+        // tolerates them, but `http::Uri` (reqwest/rig) rejects with
+        // "invalid uri character", causing all live requests to fail.
+        //
+        // The trim lives in the shared `validate_base_url` /
+        // `validate_operator_base_url`. This test exercises several callers
```
serrrfirat, line 1415: **Medium Severity**\n\nThe new regression test still leaves two important rebinding gaps unprotected. This change depends on callers *capturing* the trimmed `String` from `validate_base_url()` / `validate_operator_base_url()`, but the test only exercises a subset of those call sites.\n\nTwo reachable misses on this branch are:\n\n- `src/channels/web/handlers/llm.rs` (`test_connection` / `list_models`): validation result is ignored, then request URLs are built from raw `req.base_url`, so pasted values with surrounding whitespace can still fail.\n- `src/config/embeddings.rs`: `OLLAMA_BASE_URL` is validated but not rebound, so embedding requests can still carry whitespace through to the client.\n\nGiven the helper contract change, I think this test should cover those siblings too (or there should be caller-level tests for them), otherwise this fix can ship with the original bug still reachable through admin tooling and embeddings.

italic-jinxin:
@serrrfirat conflict and comment resolved.

serrrfirat:
Parallel review findings:

1. **Medium — Provider setup / trimming gap** — `src/llm/config.rs:301`

   The NearAI model-discovery path still reads `NEARAI_BASE_URL` raw via `env_or_override`, and `build_nearai_model_fetch_config` does the same for `NEARAI_AUTH_URL` at `src/llm/models.rs:335`. Setup then builds provider config from these values, while `NearAiChatProvider::api_url` only strips `/`, not whitespace. A trailing newline can still reach reqwest and fail with `invalid uri character`.

   Suggested fix: trim/filter these env override values in `NearAiConfig::for_model_discovery` and `build_nearai_model_fetch_config`, or reuse the validated resolver. Add a setup caller test with whitespace-padded env values.

2. **Low — Config persistence / empty semantics** — `src/channels/web/features/settings/mod.rs:452`

   Custom-provider save validation now accepts a trimmed URL, but ignores the normalized value and persists raw JSON. Also, whitespace-only `" \n"` is not treated like empty `base_url`, even though empty is allowed for staged incomplete providers.

   Suggested fix: normalize `base_url` before validation/persistence using `trim()`, branch on `trimmed.is_empty()`, and add settings set/import caller tests.

henrypark133 (dismissed):

What looks good:
- The shared URL validators now return the trimmed canonical value.
- The touched LLM, embeddings, transcription, and web LLM handler paths rebind the returned value before request construction.
- CI is green and the prior missed-caller comments appear addressed.

No verified findings.

Low-priority note:
- `src/channels/web/features/settings/mod.rs:451`: custom-provider save-time validation discards the trimmed return value, so the DB/export can retain padded whitespace. Runtime provider resolution trims again at `src/config/llm.rs:621`, so I do not see the request-construction failure persisting, but normalizing before storing would reduce drift.

Residual risk:
- Handler tests mostly call inner helpers rather than full Axum routes.

Issue #83, opened 2026-02-14
feat: CLI command expansion (gateway, channels, doctor, sessions, logs, etc.)


ilblackdragon:
## Feature Parity: CLI Command Expansion

**Priority:** P2-P3
**Source:** `FEATURE_PARITY.md` — CLI Commands

IronClaw implements `run`, `tool`, `onboard`, `config`, `status`, `memory`, `pairing`, `mcp`, `worker`, and `claude-bridge`. Many OpenClaw CLI commands are missing.

### P2 Commands

- [ ] `gateway start/stop` — Manage gateway lifecycle independently
- [ ] `channels` — Channel management (list, enable, disable, status)
- [ ] `doctor` — System diagnostics (check config, connectivity, database, providers)
- [ ] `hooks` — Hook management CLI
- [ ] `cron` — Scheduled job management (complement to routines)
- [ ] `message send` — Send messages to channels programmatically

### P3 Commands

- [ ] `skills` — Skill management (list, install, remove, inspect)
- [ ] `agents` — Multi-agent management
- [ ] `sessions` — Session listing and management
- [ ] `webhooks` — Webhook configuration
- [ ] `browser` — Browser automation
- [ ] `logs` — Query structured logs
- [ ] `update` — Self-update mechanism
- [ ] `completion` — Shell completion generation (bash, zsh, fish)

### Notes

- `models` is partially implemented (TUI selector exists, CLI subcommand is 🚧)
- Shell completion can use `clap`'s built-in `clap_complete` crate
- `doctor` is high-value for user onboarding experience

---

### Current CLI Structure

The CLI uses `clap` derive macros in `src/cli/mod.rs`. The `Command` enum has nested subcommands:

```rust
#[derive(Subcommand)]
pub enum Command {
    Run,
    Onboard { ... },
    Config(ConfigCommand),      // 5 subcommands: list, get, set, reset, path
    Tool(ToolCommand),          // install, list, auth, activate, remove, search, build
    Mcp(McpCommand),            // add, remove, list
    Memory(MemoryCommand),      // search, read, write, tree, index
    Pairing(PairingCommand),    // approve, list, revoke
    Status,
    Worker { ... },             // Internal
    ClaudeBridge { ... },       // Internal
}
```

**Pattern for adding commands:** Add a new enum variant with `#[command(subcommand)]` for commands that have sub-actions, or use a simple variant for leaf commands. Command handling in `main.rs` uses `match &cli.command` with early returns for non-agent commands.

[...]

PR #2899, opened 2026-04-23
feat(cli): add standalone gateway lifecycle commands

serrrfirat:
## Summary

Replace the stale/conflicting gateway CLI branch from #2143 with a fresh implementation on current `staging`.

Adds standalone gateway lifecycle management:
- `ironclaw gateway serve` — foreground standalone gateway for dev/debug
- `ironclaw gateway start` — background daemonized standalone gateway
- `ironclaw gateway stop` — graceful SIGTERM shutdown via PID file (Unix-only)
- `ironclaw gateway status` — PID + `/api/health` probe

Also updates the standalone gateway plumbing to fit the current codebase:
- `platform::router::start_server()` now returns `(SocketAddr, JoinHandle)` so `gateway serve` can wait for graceful shutdown
- adds dedicated `~/.ironclaw/gateway.pid`, `gateway.log`, and `gateway.token` helpers
- teaches `ironclaw logs` to reuse the standalone `gateway.token` file when no explicit token/env token is provided
- updates CLI help snapshots and `FEATURE_PARITY.md`

Standalone mode intentionally does **not** wire the agent loop, scheduler, or job manager, so agent-dependent endpoints continue to return `503` while read-only gateway surfaces remain available.

Closes/replaces #2143.

## Change Type

- [x] New feature

## Linked Issue

part of #83

## Validation

- [x] `cargo fmt`
- [x] `cargo test --lib gateway_ -- --nocapture`
- [x] `cargo test --lib cli::tests -- --nocapture`
- [x] `cargo test --test gateway_workflow_integration --test multi_tenant_integration --test openai_compat_integration --test ws_gateway_integration --test gateway_startup_multi_tenant_integration --test oauth_greeting_integration --test responses_api_path_prefix`
- [x] `cargo clippy --all-targets --all-features -- -D warnings`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3917,6 +3917,7 @@ dependencies = [
  "json5",
  "jsonschema",
  "jsonwebtoken",
+ "libc",
  "libsql",
  "lru 0.16.3",
  "mime_guess",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -211,6 +211,7 @@ security-framework = "3"
 
 # PTY allocation for Claude CLI stdout buffering fix (Unix only)
 [target.'cfg(unix)'.dependencies]
+libc = "0.2"
 pty-process = { version = "0.5", features = ["async"] }
 
 # Linux secret-service (GNOME Keyring, KWallet)
--- a/FEATURE_PARITY.md
+++ b/FEATURE_PARITY.md
@@ -156,7 +156,7 @@ This document tracks feature parity between IronClaw (Rust implementation) and O
 |---------|----------|----------|----------|-------|
 | `run` (agent) | ✅ | ✅ | - | Default command |
 | `tool install/list/remove` | ✅ | ✅ | - | WASM tools |
-| `gateway start/stop` | ✅ | ❌ | P2 | |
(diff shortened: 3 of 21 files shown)


henrypark133 (changes requested):
What looks good:
- The lifecycle surface is clearly split into `serve`, `start`, `stop`, and `status`.
- Returning `(SocketAddr, JoinHandle)` from `start_server()` is the right ownership shape for graceful shutdown.
- The dedicated gateway PID/log/token helpers make the standalone lifecycle easier to reason about.

Findings:
1. High - `src/cli/gateway.rs:142-171`, `src/channels/web/mod.rs:538-540`: `gateway serve` still derives OAuth and extension callback base URLs from the configured port before bind, so `GATEWAY_PORT=0` produces `localhost:0` even though the server later binds a real random port. Why it matters: standalone foreground mode breaks auth/callback flows in the dynamic-port setup that is most useful for dev/test. Expected fix direction: derive the effective base URL from the bound port after `start_server()`, or explicitly reject port `0` for standalone modes that advertise auth flows.

Summary:
- Review coverage: checked-out worktree plus CLI/runtime and web-security passes
- Residual risk: process-level integration coverage for real `start/status/stop` and degraded-mode route behavior is still thin

PR #2901, opened 2026-04-23
fix(skills): linear credential injection and identity bootstrap

tobias-nf:
## Summary

- **Fix broken credential injection** — `type: bearer` was injecting
  `Authorization: Bearer <key>` but Linear API keys must be sent as
  `Authorization: <key>` (no Bearer prefix). Changed to `type: header`
  with `name: Authorization`. All authenticated Linear requests were
  silently failing with this misconfiguration.
- Add first-use identity bootstrap: resolves `viewer` + teams once,
  caches to `context/intel/linear-identity.md` with a 30-day TTL.
  Subsequent "my issues" / "assigned to me" requests use the cached
  `user_id` instead of a wasteful `viewer` round-trip each time.
- Use `assignee: { id: { eq: $uid } }` filter with the cached id
  instead of `isMe: true` (not universally available in Linear's API).
- Tighten activation: more specific keywords/patterns, add
  `github issue` to `exclude_keywords`, reduce `max_context_tokens`
  from 2000 → 1600.

## Test plan

- [ ] Store a Linear API key via `secret_store`, issue a GraphQL query
      via `http` tool against `api.linear.app` — confirm 200 (previously
      401 due to Bearer prefix)
- [ ] Ask "what are my Linear issues" — confirm skill bootstraps identity
      on first use and caches to `context/intel/linear-identity.md`
- [ ] Ask again — confirm cached identity is reused, no second `viewer`
      query

### DIFF
--- a/skills/linear/SKILL.md
+++ b/skills/linear/SKILL.md
@@ -61,10 +61,10 @@ refreshed_at: 2026-04-21
 stale_after: 2026-05-21
 ---
 # Linear identity
-- user_id: 8a7f...-uuid
-- display_name: Tobias Holenstein
-- email: tobias@...
-- timezone: Europe/Zurich
+user_id: 8a7f...-uuid
+display_name: Tobias Holenstein
+email: tobias@...
+timezone: Europe/Zurich
 
 ## Teams
 - id: team-uuid-a, key: ENG, name: Engineering
--- a/skills/linear/SKILL.md
+++ b/skills/linear/SKILL.md
@@ -108,7 +108,7 @@ http(method="POST", url="https://api.linear.app/graphql", body={"query": "{ issu
 ### List Issues Assigned to the User (uses identity cache)
 
 ```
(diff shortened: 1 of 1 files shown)


henrypark133 (dismissed):
What looks good:
- The scope stays narrow and the skill update is internally consistent.
- The identity-bootstrap guidance is concrete enough to make "my issues" flows deterministic.
- Keeping this in the Linear skill instead of spreading logic into runtime code is the right ownership boundary.

No verified findings.

Summary:
- Prior feedback status: no prior blocking review
- Review coverage: checked-out diff of `skills/linear/SKILL.md`
- Residual risk: low; this is a skill-only change

serrrfirat:
Salvaged via #3265 onto current main.

Your original work is credited in the salvage PR body, including the original commit SHAs and co-author trailers. A maintainer follow-up change was added to remove a stale cross-skill reference while preserving the Linear identity-cache guidance. Thanks!

PR #2921, opened 2026-04-23
feat(webhooks): tool-driven webhook ingress with Linear integration

tobias-nf:
## Summary

Adds a generic HTTP webhook ingress that lets any registered tool receive
external webhooks and emit system events that routines can react to.
Ships with a Linear integration as the first consumer.

## Generic infra (`src/webhooks/`)

- `POST /webhook/tools/{tool}` — dispatches to any tool declaring
  `webhook_capability()`; also supports `/{tool}/{*rest}` for providers
  that embed sub-paths in the URL
- `GET /webhook/tools/{tool}` — health check
- Three auth mechanisms, all constant-time:
  - Shared secret header
  - HMAC-SHA256 (GitHub/Linear-style — configurable header and prefix)
  - Ed25519 (Discord-style timestamp + signature)
- Tool returns `emit_events` array; handler fires matching routines via
  `RoutineEngine::emit_system_event`
- 64 KB body limit

## Routine engine — event payload injection (`src/agent/routine_engine.rs`)

`execute_routine` and `spawn_fire` now accept an
`event_payload: Option<serde_json::Value>`. When a system-event routine
fires, the triggering payload is injected into the routine's prompt
under a `## Trigger Event` block (pretty-printed JSON) before execution.

This is necessary because webhook-triggered routines need to know *which
specific issue or comment* woke them up — without it, a routine receiving
a `linear.issue.update` event has no way to identify the issue and would
have to fall back to polling all recently-changed issues on every
invocation. The payload injection gives the routine immediate, structured
context so it can act precisely on the triggering event.

Cron and message-triggered routines pass `None` — no behavior change for
existing routines.

## Linear webhook tool (`src/tools/builtin/linear_webhook.rs`)

- Emits `linear.issue.{create,update,remove}` and
  `linear.comment.{create,update,remove}` system events
- Free-text fields (issue title, comment body) excluded from event
  payload to prevent prompt injection — routines fetch via API
- `user_email` hoisted to top level for pre-LLM routine filtering
- HMAC-SHA256 auth via `linear_webhook_secret` in the secrets store

## Skill (`skills/linear-webhook/SKILL.md`)

- Prerequisites check for `linear_webhook_secret` and `linear_api_key`
  with exact setup instructions if either is missing
- One-time setup guide: identity bootstrap, secret storage, Linear
  webhook configuration
- Two ready-to-create routine templates: `linear-issue-sync` and
  `linear-mention-handler`
- Polling fallback (`linear-issue-poll`) for instances without a
  public URL

## Tests

- 15 unit tests covering auth rejection, health check, `body_raw`
  mutual exclusivity, and `SERVICE_UNAVAILABLE` on missing engine slot
- 5 unit tests covering Linear event extraction and prompt-injection
  field exclusion

## Test plan

- [ ] `POST /webhook/tools/linear` with valid HMAC — confirm
  `linear.issue.update` fires a matching routine
- [ ] Comment payload with `user_email` filter — confirm only matching
  email triggers
- [ ] `GET /webhook/tools/linear` → 200; unknown tool → 404
- [ ] Wrong `linear-signature` → 401

### DIFF
--- a/skills/linear-webhook/SKILL.md
+++ b/skills/linear-webhook/SKILL.md
@@ -1,0 +1,367 @@
+---
+name: linear-webhook
+version: 0.1.0
+description: Set up and manage Linear webhook-driven automation — commitment sync on status changes and agent work triggered by @mentions in comments.
+activation:
+  keywords:
+    - linear webhook
+    - linear routine
+    - linear automation
+    - linear mention
+    - linear status sync
+    - linear trigger
+    - webhook setup
+    - linear event
+  patterns:
+    - "(?i)set.?up.*(linear).*(webhook|routine|automation)"
+    - "(?i)linear.*(mention|comment).*(trigger|work|action)"
+    - "(?i)sync.*(commitment|todo).*(linear)"
+    - "(?i)linear.*(status|state).*(sync|update|reflect)"
+  tags:
+    - linear
+    - webhook
+    - commitments
+    - automation
+    - routines
+  max_context_tokens: 3000
+requires:
+  skills:
+    - linear
+    - commitment-link-linear
+---
+
+# Linear Webhook Automation
+
+## Prerequisites check
+
+Before doing anything else, verify the required secrets are in place:
+
+```
+secret_get(name="linear_webhook_secret")
+secret_get(name="linear_api_key")
+```
+
+If `linear_webhook_secret` is missing:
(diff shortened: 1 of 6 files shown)


ilblackdragon:
Closing this because it updates the legacy top-level `src/` routine/tool/webhook implementation (`src/agent/routine_engine.rs`, `src/tools/builtin/*`, `src/webhooks/*`). Product work has moved to the Reborn workspace under `crates/`, so this PR is outdated as an implementation path.

I opened #6316 to track tool-driven webhook ingress and Linear integration as Reborn-native work with implementation notes.

PR #2924, opened 2026-04-24
fix(llm): keep local openai-compatible mock providers without API key

henrypark133:
## Summary
- stop demoting local `openai_compatible` endpoints with no API key when the registry says credentials are optional
- keep the v2 E2E mock LLM fixtures on the configured localhost provider instead of falling back away before execution starts
- harden one v2 tool lifecycle timeout helper so debug output survives `response = None`

## Why separate
This is unrelated to #2868. The callable-cleanup PR exposed broad v2 E2E failures, but the shared root cause was LLM config fallback for local mock providers, not the callable-surface changes.

## Verification
- `cargo test -p ironclaw --lib config::llm::tests::resolve_does_not_fall_back_for_openai_compatible_without_api_key -- --nocapture`
- targeted reruns on the original worktree before splitting:
  - `test_v2_engine_tool_lifecycle.py::TestV2EngineSingleTool::test_echo_tool`
  - `test_v2_engine_approval_flow.py::TestV2EngineApprovalFlow::test_approval_yes`
  - `test_v2_engine_auth_flow.py::TestV2EngineSkillActivation::test_explicit_slash_skill_prompt_reaches_auth_flow`
- broader v2 subset on the equivalent code before splitting: `70 passed, 2 skipped`

### DIFF
--- a/src/cli/models.rs
+++ b/src/cli/models.rs
@@ -186,6 +186,7 @@ async fn try_fetch_models(provider_id: &str, config_path: Option<&Path>) -> Opti
             llm_config.provider = Some(crate::llm::RegistryProviderConfig {
                 protocol: def.protocol,
                 provider_id: def.id.clone(),
+                api_key_required: def.api_key_required,
                 model: def.default_model.clone(),
                 api_key: api_key.map(secrecy::SecretString::from),
                 base_url,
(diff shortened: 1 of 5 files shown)


ilblackdragon:
Closing this because it updates the legacy top-level `src/` CLI/LLM configuration implementation (`src/cli/models.rs`, `src/config/llm.rs`, `src/llm/config.rs`). Product work has moved to the Reborn workspace under `crates/`, and this PR is now outdated against that architecture.

Thanks for the contribution.

PR #2925, opened 2026-04-24
feat(deploy): downstream deploy infra (AGENTS_SEED_PATH, INTEGRATION_CREDENTIALS_DIR, MissionSlot)

pranavraja99:
## Summary

Three plumbing primitives for downstream forks that ship domain-specific deploys on top of vanilla ironclaw. Consolidates and supersedes closed #2873 + stale #2165 + stale #2173. **Deliberately adds no user-facing tools** — Henry's #2873 review was clear about keeping a single canonical engine-v2 mission lifecycle, so this PR only exposes the existing \`MissionManager\` through an Arc rather than registering new \`mission_*\` built-ins.

### 1. \`AGENTS_SEED_PATH\` env var (\`src/workspace/mod.rs\`)

\`Workspace::seed_if_empty\` now redirects the \`AGENTS.md\` personality seed to a file on disk when \`AGENTS_SEED_PATH\` is set. Silent fallback to the compiled-in default on missing / unreadable / empty override — safer than crashing at startup on a subtle misconfiguration.

Replaces closed PR #2165.

### 2. \`INTEGRATION_CREDENTIALS_DIR\` loader (\`src/app.rs\` + \`src/tools/registry.rs\`)

\`AppBuilder::init_tools\` scans the directory (and one level of subdirectories) for \`*.json\` files, merging each file's \`mappings\` array into the shared credential registry via a new \`ToolRegistry::load_credential_mappings(path)\`. Missing directory / malformed JSON → warn and continue. Never fatal. Used by deploys that ship an \`integrations/<name>/credentials.json\` bundle (e.g. the private abound deploy's Abound API mappings).

File format:
\`\`\`json
{ \"mappings\": [
  { \"secret_name\": \"...\",
    \"location\": { \"type\": \"bearer\" | \"header\" | \"query\", \"name\": \"...\" },
    \"host_patterns\": [\"...\"],
    \"path_patterns\": [\"...\"] }
] }
\`\`\`

Replaces the two clean files from stale PR #2173; drops the other files in that PR since they overlap with already-merged #2168 path-credential work.

### 3. \`MissionSlot\` plumbing (\`src/tools/registry.rs\` + \`src/bridge/router.rs\`)

New \`pub type MissionSlot = Arc<RwLock<Option<(Arc<MissionManager>, ProjectId)>>>\`. The registry gains a \`mission_slot\` field, a \`mission_slot()\` accessor, and a \`set_mission_slot(manager, project_id)\` populator. \`bridge/router.rs::init_engine\` fills the slot once \`MissionManager\` is constructed.

**This is a plumbing primitive, not a new mission lifecycle.** No \`mission_create\`/\`mission_list\`/etc. built-in tools are registered. The v2 bridge/capability mission action surface remains canonical. The slot exists solely to let downstream integration tools that register early (via the \`ExternalToolRegistrar\` plugin seam from #2871) reach the same \`MissionManager\` at execute time.

## Motivation

The private \`nearai/ironclaw-abound\` fork currently deploys on Railway but can't reach full functional equivalence to the demo/abound branch without these three env-var/plumbing hooks:
- The Abound \`AGENTS.md\` persona seed needs \`AGENTS_SEED_PATH\`.
- The Abound API credential mappings need \`INTEGRATION_CREDENTIALS_DIR\`.
- The Abound \`send_wire\` wait-action (schedules a rate-monitoring mission) needs \`MissionSlot\`.

Without them, the fork either has to patch upstream source (losing the zero-fork benefit of #2871) or skip features (losing parity).

## Test plan

- [x] \`cargo check --all-features\`: clean
- [x] \`cargo fmt --all --check\`: clean
- [x] \`cargo clippy --all --benches --tests --examples --all-features\`: zero warnings
- [ ] End-to-end: downstream \`nearai/ironclaw-abound\` fork's next deploy — fresh user gets Abound AGENTS.md seed, Abound credentials auto-inject, send-wire wait-action creates a mission.

## Follow-ups

- Consider promoting missing-seed-file from silent fallback to a structured warning log at boot so future operators notice the misconfig.
- \`MissionSlot\` is narrow by design. If more than one in-process tool ends up needing the manager, consider promoting to a formal dependency-injection surface rather than a raw Arc\\<RwLock\\<Option\\<…\\>\\>\\>.

### DIFF
--- a/crates/ironclaw_engine/src/executor/orchestrator.rs
+++ b/crates/ironclaw_engine/src/executor/orchestrator.rs
@@ -62,12 +62,26 @@ pub(crate) const DEFAULT_ORCHESTRATOR: &str = include_str!("../../orchestrator/d
 /// surface and avoid per-turn filesystem I/O.
 static ORCHESTRATOR_OVERRIDE: std::sync::LazyLock<Option<String>> =
     std::sync::LazyLock::new(|| {
-        let path = std::env::var("ORCHESTRATOR_DEFAULT_PATH").ok()?;
-        let contents = std::fs::read_to_string(&path).ok()?;
-        if contents.trim().is_empty() {
-            None
-        } else {
-            Some(contents)
+        let env_var = "ORCHESTRATOR_DEFAULT_PATH";
+        let Ok(path) = std::env::var(env_var) else {
+            tracing::info!(env_var, "orchestrator override disabled: env var unset");
+            return None;
+        };
+        match std::fs::read_to_string(&path) {
+            Ok(contents) => {
+                let len = contents.len();
+                if contents.trim().is_empty() {
+                    tracing::warn!(env_var, path, "orchestrator override file exists but is empty — ignoring");
+                    None
+                } else {
+                    tracing::info!(env_var, path, len, "orchestrator override loaded");
+                    Some(contents)
+                }
+            }
+            Err(e) => {
+                tracing::warn!(env_var, path, err = %e, "orchestrator override file unreadable — falling back to compiled-in default");
+                None
+            }
         }
     });
 
--- a/crates/ironclaw_engine/src/executor/orchestrator.rs
+++ b/crates/ironclaw_engine/src/executor/orchestrator.rs
@@ -51,6 +51,37 @@ use crate::types::thread::{ActiveSkillProvenance, Thread, ThreadState};
 /// The compiled-in default orchestrator (v0).
 pub(crate) const DEFAULT_ORCHESTRATOR: &str = include_str!("../../orchestrator/default.py");
 
+/// Optional deploy-time override for the compiled-in orchestrator, loaded
+/// once from `ORCHESTRATOR_DEFAULT_PATH`. Downstream forks whose agent flow
+/// depends on different intent-classification heuristics (e.g. the stock
+/// `signals_execution_intent` classifier triggers extra tool calls a
+/// domain-specific deploy doesn't want) can point this env var at a
+/// replacement `default.py`. Unset → compiled-in default.
+///
+/// Read once at first access to match the rest of the engine's config
+/// surface and avoid per-turn filesystem I/O.
+static ORCHESTRATOR_OVERRIDE: std::sync::LazyLock<Option<String>> =
+    std::sync::LazyLock::new(|| {
+        let path = std::env::var("ORCHESTRATOR_DEFAULT_PATH").ok()?;
+        let contents = std::fs::read_to_string(&path).ok()?;
+        if contents.trim().is_empty() {
+            None
+        } else {
+            Some(contents)
+        }
+    });
+
+/// Returns the active "v0" orchestrator source — the env-var override if
+/// `ORCHESTRATOR_DEFAULT_PATH` is set and readable, else the compiled-in
+/// `DEFAULT_ORCHESTRATOR`. Runtime loader/seed paths call this so a
+/// downstream deploy's `default.py` flows through every code path that
+/// would otherwise reach the `include_str!` constant.
+pub(crate) fn default_orchestrator() -> &'static str {
+    ORCHESTRATOR_OVERRIDE
+        .as_deref()
+        .unwrap_or(DEFAULT_ORCHESTRATOR)
+}
+
 /// Well-known title for orchestrator code in the Store.
 pub const ORCHESTRATOR_TITLE: &str = "orchestrator:main";
 
@@ -228,19 +259,19 @@ pub async fn load_orchestrator(
 ) -> (String, u64) {
     if !allow_self_modify {
         debug!("orchestrator self-modification disabled, using compiled-in default (v0)");
-        return (DEFAULT_ORCHESTRATOR.to_string(), 0);
+        return (default_orchestrator().to_string(), 0);
     }
 
     let Some(store) = store else {
         debug!("using compiled-in default orchestrator (v0, no store)");
-        return (DEFAULT_ORCHESTRATOR.to_string(), 0);
+        return (default_orchestrator().to_string(), 0);
     };
 
     let docs = match store.list_shared_memory_docs(project_id).await {
         Ok(d) => d,
         Err(_) => {
             debug!("using compiled-in default orchestrator (v0, store error)");
-            return (DEFAULT_ORCHESTRATOR.to_string(), 0);
+            return (default_orchestrator().to_string(), 0);
         }
     };
 
@@ -259,7 +290,7 @@ pub fn load_orchestrator_from_docs(
     allow_self_modify: bool,
 ) -> (String, u64) {
     if !allow_self_modify {
-        return (DEFAULT_ORCHESTRATOR.to_string(), 0);
+        return (default_orchestrator().to_string(), 0);
     }
 
     // Find all orchestrator versions, sorted by version number descending
@@ -283,7 +314,7 @@ pub fn load_orchestrator_from_docs(
 
     if versions.is_empty() {
         debug!("using compiled-in default orchestrator (v0)");
-        return (DEFAULT_ORCHESTRATOR.to_string(), 0);
+        return (default_orchestrator().to_string(), 0);
     }
 
     // Check failure count for the latest version
@@ -318,7 +349,7 @@ pub fn load_orchestrator_from_docs(
 
     // All versions failed — fall back to compiled-in default
     debug!("all orchestrator versions failed, using compiled-in default (v0)");
-    (DEFAULT_ORCHESTRATOR.to_string(), 0)
+    (default_orchestrator().to_string(), 0)
 }
 
 /// Record a failure for the current orchestrator version.
@@ -3269,6 +3300,14 @@ mod tests {
         assert_eq!(version, 0);
         assert!(code.contains("run_loop"));
         assert!(code.contains("__llm_complete__"));
+    }
+
+    #[test]
+    fn default_orchestrator_falls_back_to_compiled_in_when_env_unset() {
+        // Tests run without ORCHESTRATOR_DEFAULT_PATH set — the helper
+        // must return the compiled-in constant verbatim. Guards against
+        // an override LazyLock that silently mangles the default.
+        assert_eq!(default_orchestrator(), DEFAULT_ORCHESTRATOR);
     }
 
     #[tokio::test]
--- a/crates/ironclaw_engine/src/executor/orchestrator.rs
+++ b/crates/ironclaw_engine/src/executor/orchestrator.rs
@@ -64,22 +64,26 @@ static ORCHESTRATOR_OVERRIDE: std::sync::LazyLock<Option<String>> =
     std::sync::LazyLock::new(|| {
         let env_var = "ORCHESTRATOR_DEFAULT_PATH";
         let Ok(path) = std::env::var(env_var) else {
-            tracing::info!(env_var, "orchestrator override disabled: env var unset");
+            // WARN-level so the line surfaces under the upstream
+            // Dockerfile's default `RUST_LOG=ironclaw=info` (which
+            // excludes ironclaw_engine info). Downstream deploys can
+            // widen the filter to silence this crate.
+            tracing::warn!(env_var, "orchestrator override disabled: env var unset, using compiled-in default");
             return None;
         };
         match std::fs::read_to_string(&path) {
             Ok(contents) => {
                 let len = contents.len();
                 if contents.trim().is_empty() {
-                    tracing::warn!(env_var, path, "orchestrator override file exists but is empty — ignoring");
+                    tracing::error!(env_var, path, "orchestrator override file exists but is empty — using compiled-in default");
                     None
                 } else {
-                    tracing::info!(env_var, path, len, "orchestrator override loaded");
+                    tracing::warn!(env_var, path, len, "orchestrator override loaded");
                     Some(contents)
                 }
             }
             Err(e) => {
-                tracing::warn!(env_var, path, err = %e, "orchestrator override file unreadable — falling back to compiled-in default");
+                tracing::error!(env_var, path, err = %e, "orchestrator override file unreadable — using compiled-in default");
                 None
             }
         }
--- a/src/app.rs
+++ b/src/app.rs
@@ -539,6 +539,38 @@ impl AppBuilder {
         tools.register_tool_info();
         tools.register_system_tools();
 
+        // Load integration credential mappings from a directory referenced
+        // by `INTEGRATION_CREDENTIALS_DIR`. Scans the directory and one
+        // level of subdirectories for `*.json` files (e.g.
+        // `integrations/abound/credentials.json`). Each file's `mappings`
+        // array is merged into the shared credential registry so injection
+        // works without patching the agent source. Missing directory or
+        // unparseable file = warn and continue; never fatal at startup.
+        if let Ok(dir) = std::env::var("INTEGRATION_CREDENTIALS_DIR") {
+            let dir_path = std::path::Path::new(&dir);
+            if dir_path.is_dir() {
+                if let Ok(entries) = std::fs::read_dir(dir_path) {
+                    for entry in entries.flatten() {
+                        let path = entry.path();
+                        if path.extension().and_then(|e| e.to_str()) == Some("json") {
+                            tools.load_credential_mappings(&path);
+                        } else if path.is_dir()
+                            && let Ok(sub) = std::fs::read_dir(&path)
+                        {
+                            for sub_entry in sub.flatten() {
+                                let sub_path = sub_entry.path();
+                                if sub_path.extension().and_then(|e| e.to_str()) == Some("json") {
+                                    tools.load_credential_mappings(&sub_path);
+                                }
+                            }
+                        }
+                    }
+                }
+            } else if dir_path.is_file() {
+                tools.load_credential_mappings(dir_path);
+            }
+        }
+
         if let Some(ref ss) = self.secrets_store {
             tools.register_secrets_tools(Arc::clone(ss));
         }
--- a/src/bridge/router.rs
+++ b/src/bridge/router.rs
@@ -1826,6 +1826,16 @@ pub async fn init_engine(agent: &Agent) -> Result<(), Error> {
     mission_manager_inner =
         mission_manager_inner.with_insights_interval(missions_config.insights_interval);
     let mission_manager = Arc::new(mission_manager_inner);
+    // Populate the shared `MissionSlot` on the tool registry so any
+    // downstream tool registered through `ExternalToolRegistrar`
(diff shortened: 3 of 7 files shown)


src/workspace/mod.rs
```diff
@@ -2230,12 +2230,27 @@ impl Workspace {
...
     pub async fn seed_if_empty(&self) -> Result<usize, WorkspaceError> {
+        // Deployments can override the default `AGENTS.md` personality seed
+        // at boot time by pointing `AGENTS_SEED_PATH` at a file on disk.
+        // Used by downstream forks (e.g. the private Abound deploy) that
+        // want a domain-specific agent persona without forking the seeds
+        // directory. If the env var is missing, unreadable, or empty, we
+        // silently fall back to the compiled-in default — the deploy gets
+        // the generic IronClaw identity, which is safer than crashing on
+        // startup with a confusing misconfiguration.
+        let custom_agents = std::env::var("AGENTS_SEED_PATH")
```
serrrfirat, line 2241: **High Severity** — Empty file divergence between v1 workspace seed and v2 engine prompt.

If `AGENTS_SEED_PATH` points at an empty (or whitespace-only) file:

- **Here (v1 workspace seed):** `read_to_string` returns `Ok("")`, `.ok()` → `Some("")`, `.unwrap_or(default)` keeps `Some("")`. `paths::AGENTS` gets seeded with `""` — silently. The compiled-in `seeds/AGENTS.md` is **not** used as fallback.
- **v2 engine prompt** (`crates/ironclaw_engine/src/.../prompt.rs:97` `load_override_file`): explicitly checks `contents.trim().is_empty()`, logs a warn, and returns `None` so the compiled-in default is used.

This contradicts the PR description's claim that v1 and v2 paths converge. A downstream that ships an `AGENTS_SEED_PATH` pointing at a temporarily-empty config (e.g. mid-deploy, rendered by a templating step that hasn't run yet) gets a blank identity in their persisted workspace AGENTS.md but a working identity through CodeAct prompts. The first-run seed is the durable artifact; users would have to manually delete it for the next boot to recover.

Either:
1. Reuse `load_override_file` semantics here (treat empty as "fall back to compiled-in"), or
2. Extract a shared `load_optional_override(env_var) -> Option<String>` helper at the workspace level and have both v1 and v2 call it.

(2) is preferable since the contract claim — "downstream forks override identity uniformly across v1/v2" — is the whole point of this knob.

src/app.rs
```diff
@@ -539,6 +539,38 @@ impl AppBuilder {
...
         tools.register_system_tools();
 
+        // Load integration credential mappings from a directory referenced
+        // by `INTEGRATION_CREDENTIALS_DIR`. Scans the directory and one
+        // level of subdirectories for `*.json` files (e.g.
+        // `integrations/abound/credentials.json`). Each file's `mappings`
+        // array is merged into the shared credential registry so injection
+        // works without patching the agent source. Missing directory or
+        // unparseable file = warn and continue; never fatal at startup.
+        if let Ok(dir) = std::env::var("INTEGRATION_CREDENTIALS_DIR") {
```
serrrfirat, line 549: **High Severity** — No integration-tier test coverage for the directory walker.

This walker (lines 549-572) is the entire production caller for `load_credential_mappings`. It contains five non-trivial branches that gate a side effect (mappings injected into `SharedCredentialRegistry`):

1. env var unset → no-op
2. env var set, path doesn't exist → no-op (silently falls through `is_dir`/`is_file` checks)
3. path is a single `.json` file → loads once
4. path is a directory of `.json` files → loads each
5. path is a directory of subdirectories of `.json` files → loads each (one level deep)
6. `read_dir` errors mid-walk → silently swallowed by `.flatten()`

The PR adds unit tests for `load_credential_mappings` in isolation (parsing, deduping, registry merge), but per the project rule **"Test through the caller, not just the helper"** (`.claude/rules/testing.md`, also called out in CLAUDE.md), a unit test on the helper alone is *not* sufficient regression coverage when there's a wrapper between it and the side effect — and this walker is exactly that wrapper.

Concrete risks the unit test misses:
- A future refactor flips `is_dir`/`is_file` order and a single-file path stops loading.
- The recursion depth quietly changes from "one level" to "any depth" or "zero levels."
- An entry's `read_dir` failure pattern changes (e.g. permission denied) and silently drops mappings.
- `.json` extension matching changes (case sensitivity, etc.).

Add an integration test (`cargo test --features integration` tier or higher) that drives the actual call site: build an `AppBuilder`, set `INTEGRATION_CREDENTIALS_DIR` to a `tempdir` populated with the structures above, run `init_tools`, and assert the registry has the expected mappings. The current unit-only coverage means a regression here will only surface in a live downstream deploy, which defeats the point of having this knob upstream.

crates/ironclaw_engine/src/executor/orchestrator.rs
```diff
@@ -51,6 +51,51 @@ use crate::types::thread::{ActiveSkillProvenance, Thread, ThreadState};
...
+/// domain-specific deploy doesn't want) can point this env var at a
+/// replacement `default.py`. Unset → compiled-in default.
+///
+/// Read once at first access to match the rest of the engine's config
+/// surface and avoid per-turn filesystem I/O.
+static ORCHESTRATOR_OVERRIDE: std::sync::LazyLock<Option<String>> =
+    std::sync::LazyLock::new(|| {
+        let env_var = "ORCHESTRATOR_DEFAULT_PATH";
+        let Ok(path) = std::env::var(env_var) else {
+            tracing::info!(env_var, "orchestrator override disabled: env var unset");
```
serrrfirat: **Medium Severity** — `tracing::info!`/`warn!` inside a `LazyLock` violates the project's REPL/TUI logging rule.

CLAUDE.md (and project guidance more broadly) is explicit:

> `info!` and `warn!` output appears in the REPL and corrupts the terminal UI. Use `debug!` for internal diagnostics… Background tasks must NEVER use `info!` — it breaks the interactive display.

Crucially, `LazyLock` evaluates **on first access, not at startup** — meaning the first `default_orchestrator()` call during a turn (or job context init) emits these logs into the live terminal session. Same applies to `load_override_file` in `crates/ironclaw_engine/src/.../prompt.rs:97`, which gets called for the AGENTS_SEED, CodeAct preamble, and CodeAct postamble overrides — three info!/warn! calls per first-access path, all triggered by interactive turns rather than startup.

Recommend dropping all of these to `debug!`:

- `tracing::info!(env_var, "orchestrator override disabled: env var unset")`
- `tracing::info!(env_var, path, len, "orchestrator override loaded")`
- `tracing::info!(env_var, "override disabled: env var unset")` (in prompt.rs)
- `tracing::info!(env_var, path, len, "override loaded")` (in prompt.rs)

The `warn!` ("override file exists but is empty / unreadable") cases are arguably defensible since they fire on misconfiguration that the operator should see — but they'll still corrupt the REPL. A defensible compromise: `debug!` always, plus surface misconfiguration through the existing health/status surface where the TUI can render it cleanly.

crates/ironclaw_engine/src/executor/orchestrator.rs
```diff
@@ -3271,6 +3316,14 @@ mod tests {
         assert!(code.contains("__llm_complete__"));
     }
 
+    #[test]
+    fn default_orchestrator_falls_back_to_compiled_in_when_env_unset() {
```
serrrfirat, line 3324: **Medium Severity** — `LazyLock`-based test is order-sensitive and can leak state between tests.

`ORCHESTRATOR_OVERRIDE` (line 63) is a process-global `LazyLock<Option<String>>` evaluated **once per process**. This test asserts the env-unset fallback path, but:

1. **Test order dependency**: If any prior test in the same binary sets `ORCHESTRATOR_DEFAULT_PATH` and triggers `default_orchestrator()` (or any code path that reads `ORCHESTRATOR_OVERRIDE`), the LazyLock is permanently initialized to `Some(...)` for the rest of the process. This test would then pass `assert_eq!(default_orchestrator(), DEFAULT_ORCHESTRATOR)` only by coincidence (because `default_orchestrator` checks the override before falling back), or fail entirely if the override file content differed.

2. **Cargo test parallelism**: Tests run in parallel by default (one process, multiple threads). A concurrent test that reads `ORCHESTRATOR_DEFAULT_PATH` from env wins the race against this one's "env unset" expectation.

3. **No cleanup possible**: `LazyLock` has no reset API. Once initialized, the value is fixed until the process exits.

Concrete failure mode: someone adds a test elsewhere that uses `std::env::set_var("ORCHESTRATOR_DEFAULT_PATH", ...)` for its setup but doesn't unset (or even unsetting wouldn't help if `default_orchestrator()` got called once with it set). This test then becomes a flaky time bomb whose failure depends on test scheduling.

Options:
- Mark this test `#[serial_test::serial]` and ensure all tests touching `ORCHESTRATOR_DEFAULT_PATH` use the same serial group.
- Refactor the LazyLock into an injectable closure/parameter on the tested function so tests can pass their own override.
- Test the loader function (`load_override_file` equivalent) directly with explicit path arguments instead of asserting on the cached singleton.

Same concern applies to the LazyLocks for `AGENTS_SEED`, `CODEACT_PREAMBLE_OVERRIDE`, `CODEACT_POSTAMBLE_OVERRIDE` in `prompt.rs` — any test asserting on their behavior is subject to the same flakiness pattern.

src/bridge/router.rs
```diff
@@ -1826,6 +1826,16 @@ pub async fn init_engine(agent: &Agent) -> Result<(), Error> {
...
     let mission_manager = Arc::new(mission_manager_inner);
+    // Populate the shared `MissionSlot` on the tool registry so any
+    // downstream tool registered through `ExternalToolRegistrar`
+    // (before the engine was up) can reach the manager at execute
+    // time. See `tools::registry::MissionSlot` for the contract.
+    // Built-in engine mission actions don't use this slot — they go
+    // through the canonical bridge/capability surface.
+    agent
+        .tools()
+        .set_mission_slot(Arc::clone(&mission_manager), project_id)
```
serrrfirat, line 1837: **Medium Severity** — `MissionSlot` is process-global and pinned to whichever project bootstrapped first, which breaks multi-tenant deploys.

`MissionSlot = Arc<RwLock<Option<(Arc<MissionManager>, ProjectId)>>>` is stored on the shared `ToolRegistry` (which is itself wrapped in an `Arc` and shared across the whole process — see `app.rs` `tools = Arc::new(ToolRegistry::new(...))`). This call site sets the slot to `(mission_manager, project_id)` for **a specific project**, then any downstream tool registered via `ExternalToolRegistrar` reads `mission_slot()` to reach the manager.

In a multi-project deploy where one process serves multiple users:

1. First user/project bootstraps → `set_mission_slot(mgr_A, project_A)` wins.
2. Second user/project bootstraps → `set_mission_slot(mgr_B, project_B)` overwrites — now downstream tools see project B regardless of which user called them.
3. Or worse: the slot is *never* re-set on subsequent bootstraps (depends on whether `set_mission_slot` is idempotent or replacing — the contract doc on the `MissionSlot` type doesn't say), and downstream tools are pinned to project A forever.

The PR description scopes this to "downstream forks like Abound where the deploy is single-tenant", but:

- This is upstream code, and IronClaw upstream is multi-tenant by design (web gateway, channel manager, multiple users with their own projects).
- The `MissionSlot` type itself doesn't document the single-tenant assumption — a future contributor could easily wire a new tool through it expecting per-call resolution.
- Even within "single-tenant Abound", admin/agent users vs end-user users could each have their own project — and the slot can only hold one.

Recommendations:
1. **Document the constraint explicitly** on the `MissionSlot` type: "This slot holds at most one (manager, project) pair for the entire process lifetime. Only safe in single-tenant deploys. For multi-tenant, resolve manager+project per call from the agent's session context."
2. Better: change the slot to `Arc<RwLock<Option<Arc<MissionManager>>>>` (no project pin), and have downstream tools resolve `project_id` from the per-call agent context (`AgentContext` / session). The mission manager is per-process; the project is per-call.
3. If (2) is too invasive for this PR, add a debug_assert in `set_mission_slot` that errors if it's already set with a *different* project, so the failure mode is loud rather than silent.

src/app.rs
```diff
@@ -539,6 +539,38 @@ impl AppBuilder {
...
+                        {
+                            for sub_entry in sub.flatten() {
+                                let sub_path = sub_entry.path();
+                                if sub_path.extension().and_then(|e| e.to_str()) == Some("json") {
+                                    tools.load_credential_mappings(&sub_path);
+                                }
+                            }
+                        }
+                    }
+                }
```
serrrfirat, line 568: **Medium Severity** — Silent error swallowing makes credential-mapping misconfig invisible to the operator.

The walker chain at lines 552-568 silently swallows multiple distinct error classes:

- `read_dir(dir_path)` errors → fall through to no-op (e.g. permission denied on the configured dir)
- `entries.flatten()` drops any individual entry's error (e.g. permission denied on a single file)
- `read_dir(&path)` for sub-dirs, same behavior
- `sub.flatten()` same behavior
- `dir_path.is_dir() == false && dir_path.is_file() == false` (e.g. broken symlink, or path is a socket/device) → fall through to no-op
- `load_credential_mappings` itself: looking at `pr-2925_registry.rs`, this method appears to log-and-continue on parse errors

The combined effect: an operator setting `INTEGRATION_CREDENTIALS_DIR=/etc/ironclaw/creds` who has a typo, wrong permissions, malformed JSON, or an unexpected file type gets **zero feedback**. Their integration mappings silently aren't loaded, and the agent then fails opaquely later when it tries to use a credential that should have been registered.

The block comment at line 543 acknowledges "warn and continue; never fatal at startup" — but I don't see any warn calls in the walker itself. The comment promises a level of operator visibility that the code doesn't deliver.

Concrete recommendations:
1. `warn!` (or `debug!` per the REPL/TUI rule, surfaced via health/status) when `INTEGRATION_CREDENTIALS_DIR` is set but the path is neither a dir nor a file — this is almost certainly an operator typo.
2. `warn!`/`debug!` on `read_dir` failures with the path and error.
3. Track and report the count of mappings loaded vs files scanned, so the operator can detect "I dropped 3 JSON files in there but only 2 mappings registered" without grepping logs.
4. At minimum, the comment-vs-code drift should be reconciled: either make the comment match (silent fallback, period) or make the code match (warn on every failure path).

serrrfirat:
## Paranoid review summary — downstream deploy knobs

This PR introduces four orthogonal extension points for downstream forks: `AGENTS_SEED_PATH`, `INTEGRATION_CREDENTIALS_DIR`, `MissionSlot`, and engine-v2 CodeAct preamble/postamble overrides + `ORCHESTRATOR_DEFAULT_PATH`. The shape is right — env-var-driven config + LazyLock caching, no hardcoded fork-specific paths in upstream code. The pattern composes cleanly with #2871's `ExternalToolRegistrar` (in fact `MissionSlot` exists specifically to give that registrar a way to reach the mission system at execute time).

### Findings (posted inline)

**High:**
1. **AGENTS_SEED empty file divergence** between v1 workspace seed (`src/workspace/mod.rs:2241`) and v2 engine prompt (`crates/ironclaw_engine/src/.../prompt.rs`). The PR description claims convergence; the actual code disagrees on what an empty file means. Recommend extracting a shared helper.
2. **No integration-tier test coverage for the `INTEGRATION_CREDENTIALS_DIR` walker** (`src/app.rs:549`). The walker has 5+ branches gating credential injection — this is exactly the "test through the caller, not just the helper" pattern called out in CLAUDE.md and `.claude/rules/testing.md`. Unit tests on `load_credential_mappings` aren't enough.

**Medium:**
3. **`tracing::info!`/`warn!` inside `LazyLock`** corrupts REPL/TUI per project logging rules. LazyLock evaluates on first access, not startup, so these fire during interactive turns. Drop to `debug!` and surface misconfiguration through the health/status surface.
4. **`LazyLock`-cached test for `default_orchestrator()`** is order-sensitive and parallel-test-hostile. Consider `#[serial_test::serial]` or refactoring to inject the override path.
5. **`MissionSlot` is process-global and project-pinned**, which is fine for single-tenant Abound but breaks multi-tenant. Document the constraint or refactor the slot to hold only the manager and resolve project per-call from the agent's session context.
6. **Silent error swallowing in the credentials walker** — the comment promises "warn and continue", the code delivers "silently fall through". Reconcile.

### Cross-PR pattern

Both this PR and #2871 introduce extension points whose **only** production caller is a wrapper at the call site that gates the actual side effect, but whose tests cover only the inner helper. This is a pattern worth flagging as a class — the project explicitly calls it out in CLAUDE.md and `.claude/rules/testing.md`. A coordinated follow-up that adds AppBuilder-driven integration tests for both `ExternalToolRegistrar.register` and the `INTEGRATION_CREDENTIALS_DIR` walker would close both gaps with a shared harness.

### v2 impact verdict

Already addressed — engine v2 is a first-class consumer of three of the four knobs (CodeAct preamble/postamble, AGENTS_SEED via prompt.rs, MissionSlot via tools registry). The orchestrator override path is engine-v2-only by construction. The one v1/v2 divergence I spotted is the empty-file behavior in finding #1.

### Merge coordination

This PR and #2871 both modify `init_tools` immediately after `register_system_tools()`. Whichever lands second will need a trivial rebase but no semantic conflict — the registrar call (`registrar.register(&tools)`) and the credentials-directory walker (`if let Ok(dir) = std::env::var("INTEGRATION_CREDENTIALS_DIR")`) are independent. Worth a coordinated merge window.

### What's solid

- Read-once-at-startup semantics across all four overrides are consistent and match existing engine config patterns.
- `MissionSlot` deferred-injection is the right shape given that downstream tools register before the engine is up — no circular-dependency dance, no Option-handling cascading through tool code.
- The `host_patterns` + `path_patterns` extension to `CredentialMapping` is composable and doesn't break existing mappings.
- `LazyLock<Option<String>>` for env-var-driven overrides is the right primitive — clean, no per-turn I/O.
- Comment quality is good throughout, particularly the "why this exists" framing on `MissionSlot` and the new env vars.

serrrfirat:
is this still required?

serrrfirat:

This PR adds downstream-specific deployment knobs across the v1 app, routines, credentials, mission slots, and setup paths. Those abstractions do not map to Reborn's typed product workflow, trigger repository, run profiles, extension credentials, or composition root.

Individual downstream requirements are welcome as focused Reborn changes. Please split them by owner—for example extension credential loading, trigger execution, or run-profile configuration—and wire each through the current crate contract with a production-path integration test.

serrrfirat:

This combines downstream deployment knobs across v1 app, routines, credentials, missions, and setup; those abstractions do not map to current Reborn owners.

Split any still-needed requirement into focused Reborn changes owned by extension credentials, triggers, run profiles, or composition, each with production-path coverage.

serrrfirat:

This combines downstream deployment knobs across v1 app, routines, credentials, missions, and setup; those abstractions do not map to current Reborn owners.

Split still-needed requirements into focused Reborn changes owned by credentials, triggers, run profiles, or composition.

PR #2926, opened 2026-04-24
fix(setup): honor explicit DATABASE_BACKEND and direct profile selection

Saaquin:
## Summary

Two correctness bugs in `SetupWizard` where an ambient environment variable silently hijacked an explicit wizard choice.

### 1. `DATABASE_URL` overrides `DATABASE_BACKEND=libsql`

`auto_setup_database()` consulted `DATABASE_URL` unconditionally after the Postgres-specific branch. On any host that had a Postgres URL sitting in the shell or `.env`, an explicit `DATABASE_BACKEND=libsql` request was quietly ignored and the wizard short-circuited with "ℹ Using existing PostgreSQL configuration".

Fix: only take the unconditional `DATABASE_URL` fallback when `env_backend.is_none()` — i.e. the caller didn't explicitly select a backend.

### 2. Ambient `IRONCLAW_PROFILE` overrides user's wizard choice

`apply_quick_local_profile("local")` used `set_runtime_env` to signal its choice and then re-read it through `optional_env`. But `set_runtime_env` only stores a fallback map; `optional_env` consults the real env first. So a host-set `IRONCLAW_PROFILE=server` (from `.env` or the shell) silently overrode the wizard's explicit user choice, and the wizard applied the server profile instead of the user-selected local one.

Fix: introduce `config::profile::apply_named_profile(settings, name)` that bypasses env resolution, and route `apply_quick_local_profile` through it. The `set_runtime_env` call is kept so downstream `optional_env` readers inside the same process still see the selected profile, but the authoritative application path no longer depends on env priority.

## Test plan

- [x] `cargo clippy --all --all-features` — clean
- [x] `cargo test --lib` — 5439 passing, 0 failing, 8 ignored (previously 5437 / 10 ignored)
- [x] `setup::wizard::tests::test_auto_setup_database_runs_migrations_with_existing_env` — re-enabled, passes
- [x] `setup::wizard::tests::test_apply_quick_local_profile_sets_profile_and_preserves_db_config_on_merge` — re-enabled, passes

## Discovery context

Caught when running the full `cargo test --lib` suite on a dev host that has a live PostgreSQL instance (from `docker-compose up`) and `IRONCLAW_PROFILE=server` set in `.env`. The two regression tests had been failing silently in that environment for weeks — they're only `#[cfg(feature = "libsql")]` / test-env sensitive, so nothing in CI catches them on a Postgres-only host.

### DIFF
--- a/src/config/profile.rs
+++ b/src/config/profile.rs
@@ -61,11 +61,27 @@ pub fn apply_profile(settings: &mut Settings) -> Result<(), ConfigError> {
         Some(n) if !n.is_empty() => n,
         _ => return Ok(()),
     };
+    apply_named_profile(settings, &name)
+}
 
-    let profile_settings = load_profile(&name)?;
+/// Apply a deployment profile by explicit name, bypassing `IRONCLAW_PROFILE`
+/// env resolution. Use this when the caller has already decided which profile
+/// to apply (e.g. setup wizard after the user selects one interactively) —
+/// otherwise an ambient `IRONCLAW_PROFILE` in the real environment would
+/// silently override the chosen profile via `set_runtime_env`, which
+/// `optional_env` treats as a fallback rather than an override.
+pub fn apply_named_profile(settings: &mut Settings, name: &str) -> Result<(), ConfigError> {
+    if name.is_empty() {
+        return Ok(());
+    }
+    // Normalize up-front so the trace log matches the name `load_profile`
+    // actually resolves against (which lowercases internally). Keeps the
+    // single-source-of-truth invariant per the project's case-insensitive
+    // comparison convention.
+    let normalized = name.to_ascii_lowercase();
+    let profile_settings = load_profile(&normalized)?;
     settings.merge_from(&profile_settings);
-
-    tracing::debug!("Applied deployment profile: {name}");
+    tracing::debug!("Applied deployment profile: {normalized}");
     Ok(())
 }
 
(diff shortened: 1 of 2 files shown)


Saaquin:
Thanks for the review! Applied #1, pushing in a moment. One clarification on #2:

**#1 (profile name normalization) — accepted.** The suggested normalization fixes a real inconsistency: `load_profile` lowercases internally but the `tracing::debug!` in my version logged the raw input. Now lowercased up-front so the log line matches what resolves, per the case-insensitive comparison convention. Pushing as a fixup.

**#2 (`parse_option_env<T>` helper) — declining.** The suggested diff looks like it replaces the URL-reading branch with an env_backend re-parse, but the function already computed `env_backend` three lines above (at line 1097). More importantly, the `-` lines remove the `std::env::var("DATABASE_URL")` read, and the `+` line doesn't put it back — the diff silently drops the entire bug-fix logic (the `finish_postgres_auto_setup(url)` call disappears). The helper also doesn't fit this case: `parse_option_env<T>` is for types that implement `FromStr` — `DATABASE_URL` is a raw connection string passed through to deadpool-postgres, not a parseable value. Using `parse_option_env::<String>("DATABASE_URL")` would just be an indirection over `std::env::var` with no added safety.

Happy to revisit if there's a concrete Option<_> shape you had in mind, but as offered the change is a regression.

serrrfirat:

PR #2927, opened 2026-04-24
fix(channels): wire load_startup_active_channels for first-run fallback

devswha:
## Summary

`main.rs` was calling `load_persisted_active_channels`, which returns an empty `Vec` when the `activated_channels` setting has never been persisted. On a clean install this meant every installed WASM channel stayed inactive at boot even when the setup wizard had just selected it in `channels.wasm_channels`. The bot never connects to Discord/Telegram/Slack/etc. until the user either calls `POST /api/extensions/{name}/activate` or seeds `activated_channels` in the DB by hand.

#2562 introduced `load_startup_active_channels` — and its tests `load_startup_active_channels_falls_back_to_configured_when_unset` / `load_startup_active_channels_prefers_persisted_empty_state_over_configured` — specifically to express the three-way semantics:

| `activated_channels` state | effective restore set |
|---|---|
| non-empty persisted list | the persisted list (authoritative) |
| persisted empty (`[]`) | empty — user turned everything off |
| not persisted (first run) | `channels.wasm_channels` fallback |

The helper was defined and unit-tested, but never wired into the production path — `main.rs:490` kept calling the raw accessor and dropped the fallback branch.

This PR swaps the call and passes `config.channels.configured_wasm_channels` as the fallback.

## Reproduction

Fresh install, no prior `activated_channels` row:

1. `ironclaw registry install discord`
2. `ironclaw onboard --step channels` — select Discord, paste bot token + public key
3. Restart ironclaw

Before this patch:
- startup banner shows `channels    http` only (no `discord`)
- `ironclaw channels list` shows `discord  enabled  wasm` ← misleading, configured ≠ activated
- `ss -tnp` shows no connection to `gateway.discord.gg`
- Discord bot stays offline

After this patch:
- startup banner shows `channels    discord  http`
- `ss -tnp` shows an ESTAB to `162.159.*.*:443` (Cloudflare / Discord Gateway)
- bot reports online, responds to mentions in `mention_channel_ids`, completes DM pairing

## Root cause trace

```
main.rs:490
    ext_mgr.load_persisted_active_channels(&ext_user_id).await   // returns Vec::new() on None
        -> persisted_active_channels = []
        -> persisted_active_wasm_channels = HashSet::new()
setup.rs:setup_wasm_channels(... startup_active_channel_names: Some({}))
    discovered_channels = {"discord": ...}    // disk scan OK
    startup_entries = discovered.filter(contains("discord"))  // empty set, so filtered out
    -> no channels registered; bot never identifies on Gateway
```

Fix call chain (same shape, one function swap):

```
ext_mgr.load_startup_active_channels(
    &ext_user_id,
    config.channels.configured_wasm_channels.clone(),
).await
```

When `activated_channels` is absent → returns `configured_wasm_channels` (== `["discord"]` in the repro). Filter now contains `"discord"`. Channel loads.

## Testing

- `cargo fmt --check` — clean
- `cargo clippy --release --lib --all-targets -- -D warnings` — clean
- `cargo test --lib --release load_startup_active_channels` → 2 passed (fallback + explicit-empty both)
- `cargo test --lib --release register_startup_loaded_channels` → 2 passed (restore semantics)
- End-to-end verification on a live host:
  1. `DELETE FROM settings WHERE key='activated_channels'` (remove the manual workaround used before this fix)
  2. `systemctl --user restart ironclaw-flask.service`
  3. Banner now includes `discord` in the channels row
  4. WebSocket to `gateway.discord.gg` ESTAB observed in `ss`
  5. Discord-side verification: bot is online; guild mention and DM pairing both work

## Notes

- This does not change behavior when `activated_channels` is persisted (empty or non-empty) — both paths are untouched because `load_startup_active_channels` preserves `activated_channels` authority once present.
- Relay channels are handled downstream via `persisted_non_relay_wasm_channel_names`; the fallback returns WASM channel names from `configured_wasm_channels`, which is the correct scope for this filter.
- No config/API changes; strictly a main-wire fix.


serrrfirat:
Can we target this branch into staging instead of main?

devswha:
Retargeted to `staging`. Thanks for the heads-up!

PR #2941, opened 2026-04-24
debug: relay callback HMAC mismatch diagnostics

PierreLeGuen:
## Summary

- When relay callback HMAC verification fails, logs secret fingerprint (SHA-256 prefix), body length, timestamp, and both received vs expected signatures
- Also logs stale-timestamp rejections with the age delta
- Temporary debugging for the Slack→IronClaw 401 issue — comparing `secret_sha256` from both sides will immediately reveal if they're using different keys

## Test plan

- [x] `cargo check` clean
- [ ] Deploy to staging alongside channel-relay `debug/callback-signature-logging`, DM the bot, check IronClaw logs for `relay callback: HMAC signature mismatch`

### DIFF
--- a/src/channels/web/features/oauth/mod.rs
+++ b/src/channels/web/features/oauth/mod.rs
@@ -512,21 +512,53 @@ pub(crate) async fn relay_events_handler(
     let ts: i64 = match timestamp.parse() {
         Ok(t) => t,
         Err(_) => {
+            tracing::warn!("relay callback: malformed timestamp");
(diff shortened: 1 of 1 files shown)


PierreLeGuen:
Fixed in 8efe501c. The mismatch log no longer prints the server-computed expected signature, and the diagnostic recomputation path with unwrap is gone.\n\nChecked: cargo fmt --all --check, cargo check -p ironclaw, and the relay webhook tests.

Issue #2474, opened 2026-04-14
Bug: stdio MCP servers trigger OAuth discovery flow on activation


nbaroni-ph:
### Summary

When activating a stdio-transport MCP server, IronClaw attempts to run an OAuth/HTTP authorization discovery flow against it instead of simply spawning the process. This fails with an authentication error because there is no URL to discover endpoints from.

### Environment

- IronClaw version: v0.25.0
- OS: macOS 26.3.1 (25D2128)
- MCP server: `@google-cloud/gcloud-mcp` via npx
- Transport: stdio

### Steps to Reproduce

1. Add a stdio MCP server:
```
ironclaw mcp add gcloud \
  --transport stdio \
  --command "npx" \
  --arg="-y" \
  --arg="@google-cloud/gcloud-mcp"
```
2. Verify it was saved correctly — `ironclaw mcp list --verbose` shows transport as `stdio`, command as `npx`, args as `-y, @google-cloud/gcloud-mcp`
3. `ironclaw mcp test gcloud` succeeds
4. Attempt to activate the server via the web UI or via a prompt that triggers tool use

### Actual Behavior

Activation fails with:
```
Activate failed: Authentication failed: Failed to discover authorization endpoints: Invalid URL: relative URL without a base
```

When triggered via a tool call, the error is:
```
Tool error: Tool tool_activate execution failed: Execution failed: Authentication failed: Failed to discover authorization endpoints: Invalid URL: relative URL without a base
```

The web UI also shows the server as inactive despite `mcp test` passing.

### Expected Behavior

For `stdio` transport, IronClaw should spawn the command directly and establish communication over stdin/stdout. No OAuth discovery or HTTP authorization flow should be attempted — there is no URL involved.

### Root Cause (hypothesis)

The activation code path appears to be missing a transport-type check before initiating the OAuth discovery flow. The HTTP auth flow is being applied unconditionally regardless of whether the server is `http` or `stdio` transport.

### Workaround

None confirmed. `mcp test` passes but the server cannot be activated for actual use.

### Additional Notes

- The same MCP server (`@google-cloud/gcloud-mcp`) works correctly when configured in Windsurf (`~/.codeium/windsurf/mcp_config.json`) using the same stdio/npx approach, confirming the server itself is not the issue
- The bug is reproducible consistently — every activation attempt produces the same error

nbaroni-ph:
just saw this:
[...]

PR #2957, opened 2026-04-25
fix(mcp): skip OAuth discovery for stdio/unix transports (#2923)

### LINKED ISSUE #2923: Bug: stdio MCP activation fails with "Failed to discover authorization endpoints"
> Re-filing #2474, which was closed in error based on a non-maintainer comment claiming stdio isn't supported. Stdio **is** wired end-to-end in v0.25.0 — the bug is strictly in the activation pre-flight, not in transport support.

## Summary

Activating a stdio-transport MCP server (via the `tool_activate` built-in tool or `POST /api/extensions/{name}/activate` from the Settings UI) fails with:

```
Execution failed: Authentication failed: Failed to discover authorization endpoints: Invalid URL: relative URL without a base
```

The `url` crate is trying to parse a stdio server's non-HTTP URL placeholder as an absolute HTTP URL inside the OAuth endpoint discovery path.

## Environment

- IronClaw: v0.26.0 (confirmed also present on `main` as of today — `src/extensions/manager.rs:3611` unchanged)
- Transport: `stdio` (reproduces with any stdio MCP server, e.g. `npx -y @google-cloud/gcloud-mcp`)

## Repro

```bash
ironclaw mcp add demo-stdio \
  --transport stdio \
  --command npx \
  --arg=-y \
  --arg=@google-cloud/gcloud-mcp

# Both of these fail with the auth error:
# 1. Via the web UI: click "Activate" on the Extensions page
# 2. Via tool: have the agent call tool_activate{name: "demo-stdio"}
```

`ironclaw mcp test demo-stdio` passes — because the CLI test command goes through `create_client_from_config` directly and skips the auth pre-flight.

## Why the previous close was wrong

The prior issue was closed with the comment *"stdio transport is not yet supported"*. That's not correct. Stdio is fully wired up:

- `src/tools/mcp/config.rs` → `McpTransportConfig::Stdio`, `McpServerConfig::new_stdio`, `effective_transport() -> EffectiveTransport::Stdio`
- `src/tools/mcp/factory.rs:114` → transport dispatch correctly handles stdio
- `ironclaw mcp test` works against stdio servers today

## Why `mcp add --transport stdio` doesn't trip the bug

`cli/mcp.rs::add_server` (`src/cli/mcp.rs:167–288`) only validates + persists the config. It never calls `auth_mcp` or `activate_mcp`. The bug only surfaces on actual activation, which is handled by `ExtensionManager::ensure_extension_ready(..., ExplicitActivate)` → `auth()` → `auth_mcp()`.

## Root cause

`ExtensionManager::auth_mcp` (`src/extensions/manager.rs:3611`) runs unconditionally for every `McpServer` kind. In gateway mode it calls `auth_mcp_build_url` (:3686) → `discover_full_oauth_metadata(&server.url)` (:3724). For stdio, `server.url` isn't an HTTP URL, so the `url` crate returns `Invalid URL: relative URL without a base`, which bubbles up as `Failed to discover authorization endpoints: ...`.

The predicate that *should* gate this already exists and returns `false` for non-HTTP transports:

```rust
// src/tools/mcp/config.rs:249
pub fn requires_auth(&self) -> bool {
    if !matches!(self.effective_transport(), EffectiveTransport::Http) {
        return false;   // stdio / unix → no HTTP auth ever
    }
    ...
}
```

`src/tools/mcp/factory.rs:114` already uses it. `auth_mcp` just doesn't. Same shape as #1948.

## Proposed fix

Short-circuit `auth_mcp` right after the existing `mcp_has_configured_auth` check:

```rust
async fn auth_mcp(&self, name: &str, user_id: &str) -> Result<AuthResult, ExtensionError> {
    let server = self
        .get_mcp_server(name, user_id)
        .await
        .map_err(|e| ExtensionError::NotInstalled(e.to_string()))?;

    if self.mcp_has_configured_auth(&server, user_id).await {
        return Ok(AuthResult::authenticated(name, ExtensionKind::McpServer));
    }

    // Non-HTTP transports (stdio, unix) never use HTTP auth.
    // Return NoAuthRequired so ensure_extension_ready falls through
    // to activation instead of attempting OAuth endpoint discovery
    // against a non-HTTP URL placeholder.
    if !server.requires_auth() {
        return Ok(AuthResult {
            name: name.to_string(),
            kind: ExtensionKind::McpServer,
            status: crate::extensions::AuthStatus::NoAuthRequired,
        });
    }

    // ... existing gateway-mode / local-mode OAuth branches unchanged ...
}
```

`NoAuthRequired` is the semantically correct status — matches the `ExtensionKind::AcpAgent` arm at `manager.rs:1548–1552`. `ensure_extension_ready` treats it identically to `Authenticated` and proceeds to activation.

## Regression test

Per `.claude/rules/testing.md` ("Test Through the Caller, Not Just the Helper"), a unit test on `requires_auth()` alone isn't sufficient — the bug is in the wrapper ignoring the predicate. The test must drive `auth_mcp`:

```rust
#[tokio::test]
async fn auth_mcp_skips_oauth_discovery_for_stdio_transport() {
    let mgr = test_extension_manager(...).await;
    let stdio_server = McpServerConfig::new_stdio(
        "local-stdio",
        "/bin/echo",
        vec![],
        HashMap::new(),
    );
    mgr.save_mcp_server(stdio_server, "user").await.unwrap();

    let result = mgr.auth("local-stdio", "user").await.unwrap();
    assert!(matches!(result.status, AuthStatus::NoAuthRequired));
}
```

Without the fix this reproduces the original error. With the fix it passes and makes zero network calls.

## Notes

- Replaces #2474 (closed in error).
- Happy to open a PR with the fix + regression test.

rajulbhatnagar:
Fixes #2923. Re-files #2474 (closed in error with an incorrect "stdio transport is not yet supported" comment — stdio is fully wired up via `McpTransportConfig::Stdio` + `create_client_from_config`; only the activation pre-flight was buggy).

## Summary

- `ExtensionManager::auth_mcp` now short-circuits to `NoAuthRequired` when `server.requires_auth()` is false, so stdio/unix activation never reaches OAuth endpoint discovery (which was failing with `Invalid URL: relative URL without a base` on the empty-string `url` placeholder).
- `ExtensionManager::mcp_supports_auth` bails on non-HTTP transports before the metadata probe, so `extensions_list` stops emitting a spurious `Failed to determine MCP auth support` debug log per stdio server per call.
- Regression test drives `mgr.auth("local-stdio", ...)` end-to-end per `.claude/rules/testing.md` → "Test Through the Caller". Without the fix it reproduces the original activation failure; with it the call returns `AuthStatus::NoAuthRequired` with no network I/O.

## Why it works

The existing `McpServerConfig::requires_auth()` predicate already encodes the full rule: `false` for non-HTTP transports, for HTTPS servers with a user-configured `Authorization` header, and for localhost HTTP. Same predicate is already used by `src/tools/mcp/factory.rs:120` to gate authenticated-transport selection. The fix consults it one layer earlier, in the wrapper that was ignoring it. Same shape as the fix for #1948 (`has_custom_auth_header` existed but `requires_auth()` didn't consult it).

Both activation entry points converge on the patched path:
- Settings UI: `POST /api/extensions/{name}/activate` → `extensions_activate_handler` → `ensure_extension_ready(ExplicitActivate)` → `auth()` → `auth_mcp()`
- Agent tool: `tool_activate` → `ensure_extension_ready(ExplicitActivate)` → `auth()` → `auth_mcp()`

After the guard returns `NoAuthRequired`, `ensure_extension_ready` falls through to `activate_mcp`, which discovers tools via `client.list_tools()` and registers them in the global `ToolRegistry`. Discovered tool names are returned in `ActivateResult.tools_loaded`.

## Scope kept

Deliberately did **not** touch `authorize_mcp_server` in `src/cli/mcp.rs:463` (CLI `ironclaw mcp auth`). Separate surface, separate issue; the CLI already prints a meaningful error on failure.

## Test plan

- [x] `cargo test --lib auth_mcp` — both `test_auth_mcp_build_url_uses_explicit_oauth_endpoints` and the new `auth_mcp_skips_oauth_discovery_for_stdio_transport` pass
- [x] `cargo check --lib` — clean
- [ ] Manual: `ironclaw mcp add demo-stdio --transport stdio --command npx --arg=-y --arg=@google-cloud/gcloud-mcp` → activate from Settings UI and via agent `tool_activate{name: "demo-stdio"}`; confirm tools are discovered and registered

### DIFF
--- a/src/extensions/manager.rs
+++ b/src/extensions/manager.rs
@@ -3772,6 +3772,15 @@ impl ExtensionManager {
             return Ok(AuthResult::authenticated(name, ExtensionKind::McpServer));
         }
 
+        // Don't run OAuth endpoint discovery when the server doesn't need HTTP
+        // auth. For stdio/unix transports `server.url` is an empty placeholder
+        // and the `url` crate fails with "Invalid URL: relative URL without a
+        // base"; for non-OAuth HTTP (e.g. localhost dev servers) there's
+        // nothing to discover. Issue #2923.
+        if !server.requires_auth() {
+            return Ok(AuthResult::no_auth_required(name, ExtensionKind::McpServer));
+        }
+
         // In gateway mode, build an auth URL and return it for the frontend to
         // open in the same browser. The gateway's /oauth/callback handler will
         // complete the token exchange.
@@ -4305,6 +4314,16 @@ impl ExtensionManager {
     async fn mcp_supports_auth(&self, server: &McpServerConfig) -> bool {
         if server.oauth.is_some() || server.requires_auth() {
             return true;
+        }
+
+        // Non-HTTP transports have no URL to probe for OAuth metadata.
+        // Short-circuit before the cache lookup so list() doesn't emit a
+        // "Failed to determine MCP auth support" debug log per stdio server.
+        if !matches!(
+            server.effective_transport(),
+            crate::tools::mcp::config::EffectiveTransport::Http
+        ) {
+            return false;
         }
 
         // Cache hit: avoid the network probe on every list() call. Cache is
@@ -11264,6 +11283,46 @@ mod tests {
         );
         assert_eq!(oauth.token_url, "https://example.com/oauth/token");
         assert_eq!(oauth.scopes, vec!["search:read".to_string()]);
+    }
+
+    /// Regression test for #2923. Drives `auth()` (the public caller) per
+    /// `.claude/rules/testing.md` → "Test Through the Caller": `requires_auth()`
+    /// was always correct; the bug was `auth_mcp` ignoring it.
+    #[tokio::test]
+    async fn auth_mcp_skips_oauth_discovery_for_stdio_transport() {
+        use crate::extensions::AuthStatus;
+
+        let dir = tempfile::tempdir().expect("temp dir");
+        // Real store — `add_mcp_server` with `None` falls back to
+        // `default_config_path()` (~/.ironclaw/mcp_servers.json), which
+        // isn't writable in sandboxed test runs.
+        let (store, _db_dir) = make_test_store().await;
+        let mgr = make_test_manager_with_dirs(
+            None,
+            dir.path().join("tools"),
+            dir.path().join("channels"),
+            Some(Arc::clone(&store)),
+        );
+
+        let stdio_server = McpServerConfig::new_stdio(
+            "local-stdio",
+            "/bin/echo",
+            vec![],
+            std::collections::HashMap::new(),
+        );
+        mgr.add_mcp_server(stdio_server, "test")
+            .await
+            .expect("add stdio server");
+
+        let result = mgr
+            .auth("local-stdio", "test")
+            .await
+            .expect("auth must not attempt OAuth discovery for stdio transport");
+        assert!(
+            matches!(result.status, AuthStatus::NoAuthRequired),
+            "expected NoAuthRequired for stdio transport, got {:?}",
+            result.status
+        );
     }
 
     #[test]


j-h-scheufen:
Could a maintainer take a look at landing this? It's been green and idle since early May, and it's the last thing between stdio-MCP deployments and a working release.

Status, for triage:
- **CI is fully green** (clippy all-features, fmt, cargo-deny, no-panics, regression enforcement — all passing).
- **Tiny, additive diff** (+59/−0): short-circuits `auth_mcp` for non-HTTP transports using the `requires_auth()` predicate the factory already relies on — same shape as the existing gate, low risk.
- Root cause is thoroughly documented in #2923.

Impact from our side (independent production deployment): the bug reproduces on **v0.29.0 and still on the just-released v0.29.1** (its release notes contain no MCP/auth change), so there is currently **no released version in which stdio MCP servers activate** — `mcp test` passes while activation fails with `Failed to discover authorization endpoints: Invalid URL: relative URL without a base`.

Two PRs (#2957 and #2960) implement the same fix; landing either unblocks everyone — this one is green. It may just need a rebase onto current `main`. Happy to validate a build against our setup (several custom stdio MCP servers) and report back. Thanks!

serrrfirat:

This patch changes OAuth discovery for v1 stdio/unix MCP transports. Reborn deliberately rejects external stdio transport and routes hosted MCP over host-mediated HTTP egress, so this transport-specific OAuth branch is not part of the current runtime.

If local MCP transport becomes a Reborn requirement, it needs a new host-owned transport contract with process isolation, lifecycle, authentication, resource accounting, and integration tests—not a discovery exception in the old client.

serrrfirat:

This changes OAuth discovery for v1 stdio/unix MCP transports. Reborn deliberately rejects external stdio and uses host-mediated HTTP for hosted MCP.

Local MCP would require a new host-owned process transport with isolation, lifecycle, auth, accounting, and integration tests—not an old-client discovery exception.

serrrfirat:

This changes OAuth discovery for v1 stdio/unix MCP. Reborn rejects external stdio and uses host-mediated HTTP for hosted MCP.

Local MCP requires a new host-owned process transport with isolation, auth, accounting, lifecycle, and integration tests.

PR #2961, opened 2026-04-25
fix(llm): honor api_key_required in unusable_reason (fixes #2946)

### LINKED ISSUE #2946: Llm_backend overwritten on every start-up
Updated to 0.25.0 built from source. Seeing a new issue: llm_backend is being reset to nearai in the database on every startup, overriding both env vars and config.toml. The priority order documented in config.toml (DB > env vars > file > defaults) isn’t behaving as expected — the DB value resets to nearai even after manual SQL updates and ironclaw models set-provider openai_compatible

willamhou:
## Summary

Fixes #2946. Self-hosted `openai_compatible` setups (vLLM, LiteLLM, etc.) without an API key were silently demoted to NearAI on every restart, overwriting the user's explicit \`ironclaw models set-provider openai_compatible\` and \`config.toml\`.

## Root cause

The `openai_compatible` registry entry declares \`api_key_required: false\`, but `LlmConfig::unusable_reason` ignored that flag and required a non-`None` `api_key` (or oauth/refresh token) for every non-Ollama provider:

\`\`\`rust
// src/config/llm.rs (before the fix)
if !is_ollama
    && provider.api_key.is_none()
    && provider.oauth_token.is_none()
    && provider.refresh_token.is_none()
{
    return Some(\"missing API key\");
}
\`\`\`

The cycle on every restart:

1. \`init_secrets\` → \`re_resolve_llm_with_secrets\` → \`resolve_llm_with_secrets_inner\`.
2. \`hydrate_llm_keys_from_secrets\` only seeds \`llm_builtin_overrides\` / \`llm_custom_providers\` — keyless self-hosted setups have neither.
3. \`LlmConfig::resolve_with_fallback\` resolves \`openai_compatible\`, then \`unusable_reason\` returns \`\"missing API key\"\`, and \`resolve_nearai_fallback\` runs.
4. Back in \`resolve_llm_with_secrets_inner\`, \`fallback_fired(Some(\"openai_compatible\"), \"nearai\")\` is true, so the function calls \`store.set_setting(user_id, \"llm_backend\", \"nearai\")\`.
5. Next startup: DB has \`nearai\`, which beats both \`config.toml\` and \`LLM_BACKEND\` env in the merge order.

## Fix

Look up the registry definition by \`provider.provider_id\` inside \`unusable_reason\` and skip the missing-key check when \`api_key_required = false\`. The Ollama short-circuit at the same site is preserved for both built-in and custom Ollama-protocol providers.

The post-fallback DB sync block in \`resolve_llm_with_secrets_inner\` is left intact — when a fallback genuinely fires (e.g. an Anthropic backend with no key), persisting the runtime backend so the UI doesn't lie is still correct. We just don't fire it for legitimate keyless configurations any more.

## Test plan

- [x] \`cargo fmt --check\` — clean
- [x] \`cargo test --lib config::llm\` — all 74 tests pass (including new regression \`resolve_does_not_fall_back_for_openai_compatible_without_api_key\`)
- [x] New regression test reproduces the original cycle: \`openai_compatible\` + \`LLM_BASE_URL\` + no \`LLM_API_KEY\` → \`cfg.backend == \"openai_compatible\"\`, not \`\"nearai\"\`. Without the fix it asserts-failed; with it, it passes and no fallback fires.

### DIFF
--- a/src/cli/models.rs
+++ b/src/cli/models.rs
@@ -188,6 +188,7 @@ async fn try_fetch_models(provider_id: &str, config_path: Option<&Path>) -> Opti
                 provider_id: def.id.clone(),
                 model: def.default_model.clone(),
                 api_key: api_key.map(secrecy::SecretString::from),
+                api_key_required: def.api_key_required,
                 base_url,
                 extra_headers: Vec::new(),
                 oauth_token: None,
--- a/src/config/llm.rs
+++ b/src/config/llm.rs
@@ -142,18 +142,6 @@ impl LlmConfig {
             .any(|c| c.id == provider.provider_id);
         let is_ollama = matches!(provider.protocol, ProviderProtocol::Ollama);
 
-        // Honor the registry's `api_key_required` flag: providers that
-        // explicitly mark a key as optional (e.g. `openai_compatible` for
-        // self-hosted vLLM/LiteLLM, `ollama`) must not be flagged unusable
-        // when no key is present. Without this, every restart with a
-        // keyless self-hosted provider would silently demote `llm_backend`
-        // to `nearai` via the post-fallback DB sync — fixes
-        // nearai/ironclaw#2946.
-        let api_key_required = ProviderRegistry::load()
-            .find(&provider.provider_id)
-            .map(|def| def.api_key_required)
-            .unwrap_or(true);
-
         // Custom providers have no hardcoded base URL in the client layer —
         // an empty `base_url` here means requests will be sent to a bare
         // path with no host, which always fails.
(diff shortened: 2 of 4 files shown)


willamhou:
Both review points addressed in c56af823:

**@gemini-code-assist** — moved `api_key_required` onto `RegistryProviderConfig` itself. Set from `def.api_key_required` in `resolve_registry_provider` (built-ins) and from `!is_ollama` for custom providers in `resolve_custom_provider`. `unusable_reason` now reads `provider.api_key_required` — no second registry load.

**@chatgpt-codex-connector** — the collision case is now correct: a custom provider with `id = "openai_compatible"` and a non-Ollama adapter inherits `api_key_required = !is_ollama = true` from the custom path, so missing credentials still trigger fallback. Added regression test `custom_provider_with_builtin_id_collision_still_requires_key` that asserts `cfg.backend == "nearai"` when `api_key` is `None` for an `open_ai_completions`-adapter custom provider that shadows the built-in id.

Bonus: removed `unwrap_or(true)` ambiguity entirely — the resolved config carries the canonical value, no defaulting on the consumer side.

CI green, 75 LLM tests + new regression all pass.

willamhou:
@ilblackdragon friendly ping — CI green, both gemini and codex review points addressed in `c56af823`. Would appreciate a review when you have a moment. Thanks!

serrrfirat:

The reported keyless-provider usability issue is already handled in Reborn's provider administration and config services: current provider metadata carries `api_key_required`, and usability/override resolution respects providers that do not require a key. This PR changes only the retired v1 configuration path.

Provider-registry improvements remain welcome. Please add Reborn admin/config contract tests for any still-failing provider shape, especially custom OpenAI-compatible or local providers, using `crates/ironclaw_llm` and the Reborn LLM administration facade.

serrrfirat:

Reborn provider administration already carries and honors `api_key_required`, including keyless/local provider resolution. This PR changes only retired v1 configuration.

Please add a Reborn admin/config regression test for any provider shape that still fails, using `ironclaw_llm` and the current LLM administration facade.

serrrfirat:

Reborn provider administration already honors `api_key_required`, including keyless/local providers. This changes only retired v1 configuration.

Please add a Reborn admin/config regression test for any provider shape that still fails.

PR #2970, opened 2026-04-26
chore: build Railway from source

drchirag1991:
Switch Railway to build directly from the repository instead of pulling ghcr.io/nearai/ironclaw:latest.

This removes the GHCR visibility dependency and lets Railway use the checked-in Dockerfile source build for the current release line.


serrrfirat:

PR #2974, opened 2026-04-27
fix: improve TUI contrast and process cleanup

Cloudymap1e:
## Summary

- improve TUI theme contrast so default dark mode follows the terminal foreground when the background is reset
- make the light theme use darker, higher-contrast semantic colors
- make interactive TUI sessions exit on terminal hangup instead of staying alive as orphaned processes
- automatically reclaim PID locks held by orphaned `ironclaw` processes whose parent is PID 1

## Root cause

The default TUI theme paired a reset background with hard-coded white foreground text, which is unreadable in light terminal profiles. Interactive TUI sessions also treated SIGHUP like daemon-style config reloads, allowing a detached terminal session to keep running and hold the PID lock.

## Validation

- `cargo test --lib bootstrap::tests::`
- `cargo test --bin ironclaw`
- `cargo test -p ironclaw_tui --lib`
- `cargo clippy --bin ironclaw -- -D warnings`
- `cargo clippy -p ironclaw_tui --lib -- -D warnings`
- `cargo build --release --bin ironclaw`

### DIFF
--- a/crates/ironclaw_tui/src/theme.rs
+++ b/crates/ironclaw_tui/src/theme.rs
@@ -53,6 +53,10 @@ impl ThemeColor {
             Self::Rgb { r, g, b } => Color::Rgb(*r, *g, *b),
         }
     }
+
+    fn is_reset(&self) -> bool {
+        matches!(self, Self::Named(name) if name == "reset")
+    }
 }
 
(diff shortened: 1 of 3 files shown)


serrrfirat:

PR #2980, opened 2026-04-27
Idea: architecture map automation for PRs

henrypark133:
## Idea

This PR proposes an architecture-map system for PRs rather than treating it as a required final policy. The goal is to make architecture impact reviewable with deterministic facts, generated Mermaid/Markdown maps, and a bounded Claude Code refinement step.

## What this suggests

- Add a deterministic Python extractor for crates, owner roots, public Rust interfaces, owner snapshots, Mermaid drafts, PR-body architecture impact, and owner shard facts.
- Add a GitHub workflow that can run by trusted request, label, or post-merge refresh.
- Keep facts.json as an artifact input, while committed ARCHITECTURE.md files are curated docs.
- Let one Claude Code session refine the generated docs, using up to four internal Agent subagents for owner batches when the PR spans many owners.
- Commit generated architecture docs only in apply mode for same-repo PRs.

## Why this may help

Large refactor/rearchitecture PRs are hard to review from raw diffs alone. This gives reviewers a visual diff in the PR body and persistent local architecture maps near crates/subsystems/complex owners.

## Validation

- python3 -B scripts/architecture_map.py test
- ruby -e 'require "yaml"; YAML.load_file(".github/workflows/architecture-map.yml"); puts "yaml ok"'

## Notes

This is intentionally framed as a suggested workflow. The extractor is deterministic and conservative; Claude Code is only used for bounded semantic curation and prose.


serrrfirat:
Thanks for putting this together — sharing review notes after walking through it against the Reborn architecture work.

## Pros

1. **Determinism floor.** Even if Claude is unavailable or budget-capped, you still get valid Mermaid + `facts.json`. Reviewers always have *something*.
2. **Strict scope validation.** Post-Claude `git diff --name-only` ∩ `facts.architecture_docs` blocks any drift outside generated docs. This is the part most "AI architecture doc" attempts get wrong.
3. **Two-tier sharding.** Repo doc + per-owner doc is the right shape — 60+ subsystems shouldn't share one file. `MAX_OWNER_INTERFACE_ITEMS=24`, `MAX_OWNER_MODULE_NODES=18` keep diagrams readable.
4. **Trust boundary is right.** `pull_request_target` + `actions/checkout` of `trusted/` from default branch + PR head as `pr/` data only + collaborator gate on `/architecture-map` comments. Cross-repo PRs cannot cause writes back to fork.
5. **Bounded cost.** Hard cap: 1 Claude session × 20 turns × haiku-4-5 × ≤4 parallel agents × ≤20 reviewed owners. Predictable per-PR ceiling.

## Cons (especially for Reborn fit)

1. **Regex Rust parsing is fragile.** Multi-line trait bounds, `#[cfg(...)]` items, macro-defined items, `cfg_attr` visibility, and `pub use` re-exports will be missed or misread. Acceptable for a "delta of likely changes," but `facts.json` consumers must not treat it as ground truth.
2. **`docs/architecture/ARCHITECTURE.md` overlaps `docs/reborn/2026-04-25-current-architecture-map.md`.** Two top-level architecture docs, generated vs. hand-curated. Ownership conflict guaranteed within weeks.
3. **Owner-root inflation.** `is_complex_module_dir` triggers on `≥3 *.rs files` OR presence of `CLAUDE.md` — `src/` has 30+ such dirs. Every PR touching `src/` spawns many owner shards, fragmenting the diff and burning Claude turns on noise.
4. **No layering / dependency-inversion checks.** `facts.json` records edges but doesn't enforce contract invariants — exactly the part the Reborn work would benefit from.
5. **Reads from `tools-src/`, `channels-src/`, `crates/`, `src/` simultaneously.** Reborn is mid-migration; many `src/*` dirs are slated for deletion. Generating `ARCHITECTURE.md` for legacy modules creates churn that distracts from the new architecture.
6. **No way to opt an owner out.** A subsystem author can't say "this owner's doc is hand-written, do not regenerate."
7. **No diff-only mode.** Always re-emits the *whole* owner snapshot to `<owner>/ARCHITECTURE.md`, not just the delta. Combined with auto-commit on `apply`, this churns docs even for PRs that didn't materially change architecture.

## How this could fit Reborn (alternative shape)

Treat the artifact layers as different *evidence kinds*, not competing docs:

```
docs/reborn/contracts/<area>.md        ← SPEC      (hand-written, durable)
crates/ironclaw_<x>/ARCHITECTURE.md    ← OBSERVED  (auto-generated, current head)
PR comment "Architecture Impact"       ← DELTA     (auto-generated, this PR only)
```

The valuable retrofit is **invariant checking between contract and observed code**, not the diagrams themselves.

### Phase 1 — Cheapest win (~1 day)

1. Keep the deterministic extractor pattern + `facts.json` contract + scope validator.
2. Replace the regex parser with `syn` in a small Rust binary (`crates/ironclaw_arch_map/`). Compile once, no recurring parser bugs, handles macros/cfg/visibility correctly.
3. Restrict the allowlist to the 15 Reborn crates (`crates/ironclaw_*`). Skip `src/`.
4. Post PR comment only — no auto-commit of `ARCHITECTURE.md` files during the migration.

### Phase 2 — The actual lever (~2–3 days)

Encode each Reborn contract as machine-checkable rules over `facts.json`:

```yaml
- name: capabilities-discovery-isolation
  source: docs/reborn/contracts/capability-discovery.md
  rule: crate("ironclaw_capabilities").imports must not include "ironclaw_dispatcher"

- name: authorization-monopoly
  source: docs/reborn/contracts/host-api-invariants.md
  rule: pub_fn matching "authorize_*" only allowed in crate("ironclaw_authorization")

- name: runtime-profile-routing
  source: docs/reborn/contracts/runtime-profiles.md
  rule: crate("ironclaw_dispatcher").pub_traits must include "RuntimeProfile"
```

CI **fails** when a PR breaks a contract, citing the contract doc. This is what the Reborn work has been missing — the contracts are aspirational today, not enforced. Every drift between contract and code currently requires a human reading both. This makes it mechanical.

### What to skip from this PR

- `docs/architecture/ARCHITECTURE.md` repo-level map — `docs/reborn/2026-04-25-current-architecture-map.md` already covers this hand-narrated.
- Owner detection across `src/`, `channels-src/`, `tools-src/` — Reborn work is `crates/` only.
- Auto-commit on `apply` mode — too much churn during migration; reviewers want the comment, not committed regenerated files.
- `claude-code-action` for per-crate doc generation — deterministic Mermaid is enough for crates this small. Save Claude budget for the contract-violation explanation step.

## Summary

Solid **v0** if the goal is a low-cost reviewer aid this week. As the long-term Reborn architecture-doc system, recommend forking the `facts.json` extractor pattern + trust model + scope validator into a Reborn-only workflow scoped to the 15 new crates, swap regex for `syn`, add contract-invariant checks, skip the `apply`-commit-back behavior. ~1–2 days for a meaningful ratchet vs. a few hours to merge this and inherit the drift.

Happy to draft the Reborn-scoped version (Rust extractor + contract invariant checker + workflow YAML) as a counter-proposal if useful.

PR #2988, opened 2026-04-27
feat(reborn): add host foundation crates

serrrfirat:
## Summary

Carves the first Reborn landing slice onto the new integration branch.

This PR adds the foundation crates needed by later Reborn slices:

```text
crates/ironclaw_host_api
crates/ironclaw_resources
crates/ironclaw_architecture
```

Target branch:

```text
reborn-integration
```

Tracking issue:

```text
#2987
```

### `ironclaw_host_api`

Neutral host API vocabulary and contracts:

- scope/ID types, including first-class optional `AgentId`
- `ExecutionContext` / `ResourceScope`
- runtime and trust vocabulary
- capability descriptors and decisions
- mount/path/resource/action/approval/audit/dispatch contracts
- local default scope helpers:
  - `tenant_id = "default"`
  - stable local `user_id`
  - `agent_id = Some("default")`
  - `project_id = Some("bootstrap")`

### `ironclaw_resources`

Resource-governor foundation:

- `ResourceEstimate`
- `ResourceReservation`
- `ResourceGovernor`
- scoped in-memory governor
- `reserve_with_id(...)` support for later obligation handoff
- duplicate/mismatch reservation errors
- tenant/user/project/agent scope cascade tests

### `ironclaw_architecture`

Boundary guardrails:

- `cargo metadata` based dependency-boundary test crate
- incremental grouped-landing behavior: rules activate for Reborn crates once the crate is present in the workspace, so earlier grouped PRs do not fail because later crates are not landed yet

## Out of scope

- `ironclaw_extensions` is intentionally deferred from PR1 because the current implementation depends on `ironclaw_filesystem`; it should move with/after the filesystem substrate slice.
- filesystem, events, authorization, approvals, dispatcher, process, secrets, network, memory, runtime lanes, and obligation handling are later grouped PRs.
- no production/user-visible Reborn wiring is enabled by this PR.

## Exposure checklist

- [x] Does not alter app startup/config defaults.
- [x] Does not expose new routes/CLI/user-visible APIs.
- [x] Does not create a second writer for production state.
- [x] Adds internal Reborn crates only.
- [x] Status remains “implemented slice,” not product-complete Reborn.

## Verification

Run from `.worktrees/reborn-land-01-foundation`:

```bash
cargo test -p ironclaw_host_api -p ironclaw_resources -p ironclaw_architecture
cargo clippy -p ironclaw_host_api -p ironclaw_resources -p ironclaw_architecture --all-targets -- -D warnings
cargo fmt --check
git diff --check
```

All passed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3976,6 +3976,13 @@ dependencies = [
 ]
 
 [[package]]
+name = "ironclaw_architecture"
+version = "0.1.0"
+dependencies = [
+ "serde_json",
+]
+
+[[package]]
 name = "ironclaw_common"
 version = "0.2.0"
 dependencies = [
@@ -4016,6 +4023,31 @@ dependencies = [
  "serde_json",
  "thiserror 2.0.18",
  "tracing",
+]
+
+[[package]]
+name = "ironclaw_host_api"
+version = "0.1.0"
+dependencies = [
+ "async-trait",
+ "chrono",
+ "rust_decimal",
+ "rust_decimal_macros",
+ "serde",
+ "serde_json",
+ "sha2",
+ "thiserror 2.0.18",
+ "uuid",
+]
+
+[[package]]
+name = "ironclaw_resources"
+version = "0.1.0"
+dependencies = [
+ "ironclaw_host_api",
+ "rust_decimal",
+ "rust_decimal_macros",
+ "thiserror 2.0.18",
 ]
 
 [[package]]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4036,6 +4036,7 @@ dependencies = [
  "serde",
  "serde_json",
  "sha2",
+ "static_assertions",
  "thiserror 2.0.18",
  "uuid",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,5 +1,5 @@
 [workspace]
-members = [".", "crates/ironclaw_common", "crates/ironclaw_safety", "crates/ironclaw_skills", "crates/ironclaw_engine", "crates/ironclaw_gateway", "crates/ironclaw_tui"]
+members = [".", "crates/ironclaw_common", "crates/ironclaw_host_api", "crates/ironclaw_resources", "crates/ironclaw_architecture", "crates/ironclaw_safety", "crates/ironclaw_skills", "crates/ironclaw_engine", "crates/ironclaw_gateway", "crates/ironclaw_tui"]
 exclude = [
     "channels-src/discord",
     "channels-src/telegram",
--- a/crates/ironclaw_architecture/CLAUDE.md
+++ b/crates/ironclaw_architecture/CLAUDE.md
@@ -1,0 +1,6 @@
+# ironclaw_architecture guardrails
+
+- This crate is test-only architecture enforcement; do not add production dependencies or runtime behavior.
+- Use `cargo metadata` or equivalent workspace graph checks to enforce Reborn dependency direction.
+- Boundary tests should fail loudly with the exact forbidden edge and crate name.
+- Keep rules conservative and explicit; update docs when intentional architecture edges change.
--- a/crates/ironclaw_architecture/Cargo.toml
+++ b/crates/ironclaw_architecture/Cargo.toml
@@ -1,0 +1,8 @@
+[package]
+name = "ironclaw_architecture"
+version = "0.1.0"
+edition = "2024"
+publish = false
+
+[dev-dependencies]
+serde_json = "1"
--- a/crates/ironclaw_architecture/src/lib.rs
+++ b/crates/ironclaw_architecture/src/lib.rs
@@ -1,0 +1,4 @@
+//! Workspace architecture contract tests for IronClaw Reborn.
+//!
+//! This crate intentionally has no production dependencies. Its tests inspect
+//! workspace metadata and fail when Reborn crate dependency boundaries drift.
--- a/crates/ironclaw_architecture/tests/reborn_dependency_boundaries.rs
+++ b/crates/ironclaw_architecture/tests/reborn_dependency_boundaries.rs
@@ -1,0 +1,349 @@
+use std::{collections::HashMap, path::PathBuf, process::Command};
+
+use serde_json::Value;
+
+#[test]
+fn reborn_crate_dependency_boundaries_hold() {
+    let metadata = cargo_metadata();
+    let packages = metadata["packages"]
+        .as_array()
+        .expect("cargo metadata must include packages");
+    let dependencies = packages
+        .iter()
+        .filter_map(package_dependencies)
+        .collect::<HashMap<_, _>>();
+
+    assert_no_normal_workspace_deps(
+        &dependencies,
+        "ironclaw_host_api",
+        workspace_ironclaw_crates(&dependencies)
+            .into_iter()
+            .filter(|name| *name != "ironclaw_host_api")
+            .collect::<Vec<_>>(),
+    );
+
+    for rule in boundary_rules() {
+        assert_no_normal_workspace_deps(&dependencies, rule.crate_name, rule.forbidden);
+    }
+}
+
+struct BoundaryRule {
+    crate_name: &'static str,
+    forbidden: Vec<&'static str>,
+}
+
+fn boundary_rules() -> Vec<BoundaryRule> {
+    vec![
+        BoundaryRule {
+            crate_name: "ironclaw_filesystem",
+            forbidden: vec![
+                "ironclaw_authorization",
+                "ironclaw_approvals",
+                "ironclaw_capabilities",
+                "ironclaw_dispatcher",
+                "ironclaw_events",
+                "ironclaw_extensions",
+                "ironclaw_host_runtime",
+                "ironclaw_secrets",
+                "ironclaw_network",
+                "ironclaw_mcp",
+                "ironclaw_processes",
+                "ironclaw_resources",
+                "ironclaw_run_state",
+                "ironclaw_scripts",
+                "ironclaw_wasm",
+            ],
+        },
+        BoundaryRule {
+            crate_name: "ironclaw_memory",
+            forbidden: vec![
+                "ironclaw_authorization",
+                "ironclaw_approvals",
+                "ironclaw_capabilities",
+                "ironclaw_dispatcher",
+                "ironclaw_events",
+                "ironclaw_extensions",
+                "ironclaw_host_runtime",
+                "ironclaw_secrets",
+                "ironclaw_network",
+                "ironclaw_mcp",
+                "ironclaw_processes",
+                "ironclaw_resources",
+                "ironclaw_run_state",
+                "ironclaw_scripts",
+                "ironclaw_wasm",
+            ],
+        },
+        BoundaryRule {
+            crate_name: "ironclaw_resources",
+            forbidden: vec![
+                "ironclaw_authorization",
+                "ironclaw_approvals",
+                "ironclaw_capabilities",
+                "ironclaw_dispatcher",
+                "ironclaw_events",
+                "ironclaw_extensions",
+                "ironclaw_filesystem",
+                "ironclaw_host_runtime",
+                "ironclaw_secrets",
+                "ironclaw_network",
+                "ironclaw_mcp",
+                "ironclaw_processes",
+                "ironclaw_run_state",
+                "ironclaw_scripts",
+                "ironclaw_wasm",
+            ],
+        },
+        BoundaryRule {
+            crate_name: "ironclaw_extensions",
+            forbidden: vec![
+                "ironclaw_authorization",
+                "ironclaw_approvals",
+                "ironclaw_capabilities",
+                "ironclaw_dispatcher",
+                "ironclaw_events",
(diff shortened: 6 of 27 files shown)


ilblackdragon:
# Code Review

## Overview

Foundation slice of the "Reborn" architecture — adds three new internal-only workspace crates with no production wiring:

- **`ironclaw_host_api`** — vocabulary/contracts: IDs, scopes, paths, mounts, capabilities, actions, decisions, approvals, audit envelopes, dispatch port, resource shapes
- **`ironclaw_resources`** — in-memory `ResourceGovernor` enforcing scope-cascade limits
- **`ironclaw_architecture`** — test-only crate that uses `cargo metadata` to enforce future Reborn dependency-direction rules

+4,172 / −1; zero production callers; XL but additive.

## Strengths

- **Strong types throughout.** Newtype wrappers for every identifier (`TenantId`, `UserId`, `CapabilityId`, etc.) with explicit validation.
- **Defense-in-depth path validation.** `ScopedPath` rejects URLs, Windows paths, raw host prefixes, NUL/control chars, `..` segments. `VirtualPath` requires a known root.
- **No `unwrap`/`expect` in production code.** Mutex poisoning handled via `into_inner`. Tests use `unwrap`, which is fine.
- **Thorough contract tests.** ~1,300 lines of tests including a concurrent-reservation oversubscription stress test using `Barrier`.
- **Good module structure.** Each module has a docstring explaining its role and the boundary it enforces. Per-crate `CLAUDE.md` files document guardrails.
- **Audit envelope omits raw paths/secrets.** `ActionSummary::from_action` extracts a string target without leaking payloads.

## Issues / Suggestions

### Correctness

- **`ResourceTally::sub_assign` is inconsistent for `usd`.** Integer fields use `saturating_sub`; `usd` uses `-=`. `Decimal` won't panic but will go negative on bookkeeping bugs and silently corrupt the tally. Recommend either `Decimal::ZERO.max(self.usd - other.usd)` or a documented assert. (`crates/ironclaw_resources/src/lib.rs`)
- **`check_decimal` overflow.** `usage + reserved + requested > limit` on `Decimal` can panic on overflow with very large values. Low practical risk for USD budgets but consider `checked_add`.
- **First-failure-only denial reporting.** `check_limits` short-circuits on the first denied dimension/account, so a request that exceeds three dimensions only reports the first. Acceptable now, but worth a comment so readers don't expect aggregated denials.

### Architecture / API

- **`RAW_HOST_PREFIXES` is incomplete.** Only `/Users/`, `/home/`, `/etc/`, `/var/`, `/private/`, `/Volumes/`. Missing `/usr/`, `/opt/`, `/tmp/`, `/root/`, `/Library/`, `/proc/`, `/sys/`. The comment elsewhere implies real containment lives in `ironclaw_filesystem` — fine — but make that explicit here so future contributors don't treat this as authoritative. (`crates/ironclaw_host_api/src/path.rs`)
- **`MountView::is_subset_of` requires exact target equality.** A child can't have a mount narrower than its parent (e.g., parent `/workspace -> /projects/p1`, child `/workspace -> /projects/p1/subdir`). Probably intentional for v1, but document the limitation.
- **`InvocationFingerprint` doc-vs-test mismatch.** Test `invocation_fingerprint_is_stable_and_input_redacted` checks `!first.as_str().contains("secret payload")`, but SHA-256 hashes are not "redaction" — they're one-way. With a small input space the fingerprint is rainbow-tableable. Used for dedup/replay only, not authn, so this is fine — but rename the assertion to "hashed" rather than "redacted" to avoid future security-property confusion.
- **`audit_envelope_carries_agent_scope_without_leaking_payloads`** asserts `json.get("host_path").is_none()` — but the envelope schema has no `host_path` field at all, so this is asserting non-existence rather than redaction. Add a positive assertion that path fields contain only the `ScopedPath` string, not the host path.

### Style / Conventions

- **`rust-version = "1.92"`** is set on both new crates. If the workspace hasn't bumped MSRV to 1.92, this either does nothing or causes confusion. Verify that this matches the workspace MSRV.
- **`boundary_rules()` is highly repetitive** — 14 entries, ~140 lines of nearly identical Vec literals. Could be expressed as a layered list (lower→upper) plus computed forbidden = "all upper layers". Acceptable as-is for explicitness; tag for follow-up.
- **`pub use` flat re-exports in `lib.rs`**. The CLAUDE.md style guide says "No `pub use` re-exports unless exposing to downstream consumers." Since this *is* a contract crate intended to be consumed by other Reborn crates, the re-exports are justified — but worth a one-line comment explaining the exception.

### Test coverage

- No test that `MountView::resolve` rejects when the resolved tail falls outside the virtual root (e.g., a `tail` containing `..` should be caught — and it is, via `VirtualPath::new`, but worth a direct test).
- No test that `release` is idempotent vs. `reconcile`-then-`release` (the test shows the second call returns `ReservationClosed`, but doesn't verify the reserved tally wasn't double-decremented).
- Concurrent test uses 8 threads and `max_concurrency_slots: 1` — solid. Consider a follow-up that mixes reserve/reconcile/release concurrently to stress the lock-and-replace pattern in `reconcile`.

### Security

- **`HostPath` is correctly non-`Serialize`** — good. Confirm via test that it can't be serialized accidentally (the type system enforces this, but a `static_assertions::assert_not_impl_all!` would lock it in).
- **No secrets in any serializable type.** `SecretHandle` is just a name; values never appear in `Action`, `AuditEnvelope`, `ApprovalRequest`. Good.
- The `cargo metadata` boundary test runs `cargo` as a subprocess — fine for `cargo test` but won't work in stripped CI containers without a writable target dir / real cargo install.

## Risk Assessment

**Low.** Pure additive change with no production wiring. The exposure checklist is honest:

- No app startup/config changes
- No new routes/CLI/APIs
- No database writers
- All new crates are `publish = false`

Worst case if this lands and is rolled back: revert one PR, no migration needed.

## Verdict

Approve with optional follow-ups for the inconsistent `sub_assign` on `usd`, the misleading "redacted" test names, and a comment about the `lib.rs` re-export exception. The contract design is careful, well-documented, and well-tested. The boundary-test crate is a clean way to lock in architectural intent ahead of the implementation slices.

zmanian:
**Contract-level review on the foundation types** — these are the vocabulary the rest of Reborn will be written against, so worth pushing on early. Skipping lint-style notes (gemini-code-assist already covered Decimal saturation etc.); raising six things I think need decisions baked into PR1, not deferred.

### 1. `Decision::Allow { obligations: Vec<Obligation> }` — ordering and duplication aren't specified

`obligations` is a `Vec`, which means callers can encode `[AuditBefore, AuditBefore]` or differing orderings of the same set. The freeze says "non-empty obligations fail closed if unsupported/unconfigured" but doesn't say:

- Whether duplicate obligations are an invariant violation, an error, or silently deduped.
- Whether obligation evaluation order is meaningful (e.g., must `ReserveResources` precede `InjectSecretOnce` so the lease rides the reservation?).
- Whether two obligations with conflicting parameters (`EnforceOutputLimit { bytes: 1024 }` and `EnforceOutputLimit { bytes: 2048 }`) is a fail-closed error or "tightest wins."

This will produce inconsistent runtime behavior across handlers if not pinned in the contract. Suggest: a typed `Obligations` wrapper that enforces uniqueness-per-kind at construction, and a documented evaluation order (probably `ReserveResources -> UseScopedMounts -> ApplyNetworkPolicy -> InjectSecretOnce -> AuditBefore -> [dispatch] -> RedactOutput -> EnforceOutputLimit -> AuditAfter`).

### 2. `RuntimeKind::FirstParty / System` and `TrustClass::FirstParty / System` are self-assertable in serde

Anyone can deserialize `{"runtime": "first_party", "trust": "system"}`. The type system doesn't prevent it because the policy engine that validates trust class assignment lives upstream, not in `ironclaw_host_api`. That's the right layering — but the types should make it impossible to *accidentally* construct privileged values without going through the policy engine. Two options:

- Mark the privileged variants with a private constructor seam (e.g., separate `TrustClass::user_*` constructors and a `pub(crate)` `TrustClass::first_party()` only callable by a host-policy crate after validation).
- Or add a doc comment + a deserialization guard that rejects `FirstParty`/`System` outside a "trusted source" context.

Without one of these, the trust-class policy engine becomes the only thing standing between a malicious manifest and `Principal::System`. PR1b can't be optional — see my comment on #2987.

### 3. `Principal::System` is too coarse

`Principal::System` is a unit variant. It conflates "the host runtime acting on its own behalf" (kernel-internal) with "the system service principal" (which is a real grantee in some flows, e.g., scheduled routines fired by a system trigger). The freeze said `System` is "not a matchable grantee" but the `ApprovalRequest` resume path needs a `requested_by` principal — what fills it when the request originated from `System`?

Suggest splitting:
- `Principal::HostRuntime` — used for kernel-internal invocations, never matchable as grantee.
- `Principal::System(SystemServiceId)` — used for a named trusted system service (heartbeat, routine-engine, migration-runner). Granteable but only to other System principals.

### 4. `HostPath` `Debug` derive will leak through composition

`HostPath` is `pub struct HostPath(PathBuf)` with `#[derive(Debug)]`. The freeze invariant is "host paths must not appear in user-facing errors/events/audit." Auto-derived `Debug` defeats this the moment `HostPath` becomes a field of any other `Debug`-derived struct that gets logged — and many will (resource records, mount descriptors, dispatch errors). The current `audit.rs` may do the right thing manually, but every future audit/event/error type is a footgun.

Either:
- Implement `Debug` manually to print `HostPath(<redacted>)`, accepting the small dev-experience cost.
- Or use a wrapper newtype (e.g., `Redacted<HostPath>`) that's the only thing legal to expose in `Debug`-derived contexts.

### 5. `CapabilityId` doc/code mismatch

`CapabilityId::new` error message says "must be '<extension>.<capability>'" but the implementation accepts any `>= 2` dot-segments — `foo.bar.baz.qux` is valid. Either:
- Tighten the validator to `== 2` segments, matching the error message; or
- Update the message to "must be `<extension>.<capability>[.<sub>...]`" and document the namespacing.

If multi-segment is intentional (which I'd expect — capability namespacing is useful), say so explicitly so adapters and authorizers don't make incompatible assumptions.

### 6. `ResourceAccount::cascade` semantics need to be explicit

`cascade()` builds `[Tenant, User, ...]` but the variants overlap: `Project { tenant, user, project }` and `Agent { tenant, user, project: Option<...>, agent }` are both valid. For a scope with `project = Some(P)` and `agent = Some(A)`, the cascade produces both Project and Agent records. Question: do limits *sum* across siblings (project budget + agent budget both apply), or does the deepest level *override*? The current implementation appears to apply both (a reservation is rejected if any account exceeds limits), which is correct — but it should be a contract assertion, not implementation-only.

Add to `ResourceGovernor` doc: "A reservation succeeds only if every account in `ResourceAccount::cascade(scope)` would remain within its limits. Limits at deeper accounts do not override shallower limits; both apply."

---

These are foundation-PR concerns specifically because every later crate inherits them. Happy to open follow-up issues for any of these if the team prefers narrower threads.

zmanian:
Surfacing two findings from automated reviewers on the downstream events PR (#2993) that actually originate in this PR's foundation crates. Both seem real; flagging here so they don't get lost in a review thread on a different PR.

### 1. `canonical_json` unbounded recursion in `ironclaw_host_api/src/approval.rs:86`

Flagged by gemini-code-assist (security-medium):

> The `canonical_json` function uses recursion to process `serde_json::Value`. In security-sensitive contexts like generating invocation fingerprints for approvals, this could be exploited by providing a deeply nested JSON object to cause a stack overflow, leading to a Denial of Service.

The function is on the approval-fingerprint path, so the input is whatever JSON a capability invocation carries — i.e., attacker-controlled in any tenant where an extension can invoke a capability with structured params. Suggest one of:
- iterative implementation using an explicit stack;
- recursion depth cap that returns `HostApiError::invariant("canonical_json: max depth exceeded")` past, say, 64 levels;
- a max-bytes / max-entries cap on the canonicalized output as a complementary defense.

A targeted test that constructs deeply nested input (e.g., `{"a": {"a": {"a": ... }}}` 10k deep) and verifies the function returns an error rather than overflowing the stack would pin the fix.

### 2. Negative `Decimal` USD estimates undermine quotas in `ironclaw_resources/src/lib.rs:263`

Flagged by chatgpt-codex-connector (P1):

> `ResourceTally::from_estimate` accepts signed `Decimal` values unchanged, so a negative `estimate.usd` flows into reservation checks and reserved balances. That allows callers to lower `active_reserved` with a negative reservation and then admit additional positive reservations that should be denied by budget limits.

`ResourceEstimate::usd: Option<Decimal>` is unbounded in sign, but the reservation/reconciliation arithmetic treats it as additive. A negative value silently *credits* the account on reservation, then any subsequent positive reservation can exceed the limit. Concrete fix paths:
- reject negative components in `InMemoryResourceGovernor::reserve_with_id` with a typed `ResourceError::InvalidEstimate`;
- or, more defensively, validate at construction time in a typed `ResourceEstimate::new` / `try_from` that rejects negative or NaN values for every numeric field.

Test would build a reservation with `estimate.usd = Some(Decimal::from(-100))` and assert `reserve_with_id` returns an error rather than reducing reserved balance.

Both are fail-closed concerns; happy to open a follow-up PR against `reborn-land-01-foundation` if you'd prefer that to inline edits, but flagging here first since the review came in as you were finalizing.

PR #2989, opened 2026-04-27
fix(engine): make mission threads_today reset timezone-aware

### LINKED ISSUE #1945: bug(v2 engine): mission threads_today counter never resets — daily budget permanently exhausted
## Description

The V2 Mission daily thread budget (`threads_today` / `max_threads_per_day`) is enforced but never reset. Once a mission reaches its daily limit, it is permanently blocked from firing — the counter only increments and has no reset path.

## Current behavior

`fire_mission()` increments the counter:

```rust
// crates/ironclaw_engine/src/runtime/mission.rs, line ~286
mission.threads_today += 1;
```

The budget check in `should_fire()` blocks firing when the limit is reached:

```rust
// crates/ironclaw_engine/src/runtime/mission.rs, line ~256
if mission.max_threads_per_day > 0 && mission.threads_today >= mission.max_threads_per_day {
    tracing::debug!(mission_id = %mission.id, "daily budget exhausted");
    return false;
}
```

`Mission::new()` sets the default:

```rust
// crates/ironclaw_engine/src/types/mission.rs, line ~149
max_threads_per_day: 10,
threads_today: 0,
```

There is no code anywhere in the engine that resets `threads_today` back to 0 at the start of a new day. The `tick()` loop runs every 60 seconds but has no date-change detection.

## Expected behavior

- `threads_today` should reset to 0 when the calendar day changes (relative to the mission's timezone or UTC)
- After reset, the mission should resume normal firing within its daily budget

## Impact

**MEDIUM** — Any mission with `max_threads_per_day > 0` (the default is 10) will permanently stop firing after 10 total threads across its entire lifetime, not 10 per day as intended. The only workaround is manually calling `mission_update` to reset `threads_today = 0`.

## Suggested fix

Option A (minimal): At the top of `tick()`, check each active mission's `updated_at` date against the current date. If the day has changed, reset `threads_today = 0`.

Option B (explicit): Add a `last_reset_date: Option<NaiveDate>` field to `Mission`. In `tick()`, compare against today's date and reset when it differs. This is more robust across timezone edge cases.

## Acceptance criteria

- [ ] `threads_today` resets to 0 at the start of each new day
- [ ] A mission that exhausted its daily budget yesterday can fire again today
- [ ] Unit test covers: fire N times to exhaust budget → simulate day change → verify mission can fire again

## Related

- Branch: `v2-architecture`

ilblackdragon:
## Summary

- The daily-budget reset added in #2570 compared `last_fire_at.date_naive()` against `now.date_naive()` in UTC. A cron mission configured with a non-UTC timezone (e.g. `America/Los_Angeles`) could stay "exhausted" for up to ~17 hours after its local day rolled over, because the UTC date hadn't changed yet.
- Extract the staleness check into `threads_today_is_stale(&mission)` and use the cron mission's timezone for the day boundary when set; UTC remains the fallback for manual/event-driven cadences and untimezoned cron.
- Adds the missing tick-path regression test — pre-existing `threads_today_resets_on_new_day` only exercised `fire_on_system_event`.

Fixes #1945

## Test plan

- [x] `cargo test -p ironclaw_engine --lib runtime::mission` — all 87 mission tests pass
- [x] New test `cron_mission_threads_today_resets_via_tick` — cron + tick path
- [x] New test `threads_today_resets_at_cron_local_midnight` — `Pacific/Auckland`, last fire is yesterday-local but same UTC day; old logic would not have reset
- [x] New test `threads_today_is_stale_predicate` — direct unit coverage of the boundary helper
- [x] `cargo fmt --check` clean
- [x] `cargo clippy -p ironclaw_engine --all-targets -- -D warnings` clean

### DIFF
--- a/crates/ironclaw_engine/src/runtime/mission.rs
+++ b/crates/ironclaw_engine/src/runtime/mission.rs
@@ -127,23 +127,30 @@ type DedupKey = (MissionId, String);
 /// Returns true if `mission.threads_today` carries over from a previous
 /// calendar day and must be reset before the daily-budget gate.
 ///
-/// Day boundary is the cron mission's configured timezone if any, otherwise
-/// UTC. A user with `cron timezone = "America/Los_Angeles"` expects the
-/// budget to refresh at LA midnight, not UTC midnight (which would be 5 PM
-/// local), and a UTC reset would leave that mission idle for ~17 hours
-/// after their local day rolled over.
+/// Day boundary is the cron mission's configured timezone when set,
+/// otherwise UTC. A user with `cron timezone = "America/Los_Angeles"`
+/// expects the budget to refresh at LA midnight; a UTC reset would leave
+/// that mission idle for many hours into their local day (UTC midnight
+/// lands in the previous afternoon/evening local time, with the exact hour
+/// shifting under DST).
 ///
 /// `last_fire_at = None` with a non-zero counter is treated as stale —
-/// either the persisted timestamp was lost or the counter was nudged
-/// out-of-band, and resetting is the safe direction.
-fn threads_today_is_stale(mission: &Mission) -> bool {
+/// the two fields are written together by `fire_mission`, so this state
+/// only arises from data corruption / migration / out-of-band edits, and
+/// resetting is the recovery direction. Returning `false` here would
+/// reintroduce the permanent-exhaustion bug this helper exists to fix.
+///
+/// `now` is injected so the call site can pin a single instant across
+/// the staleness check and downstream fire-accounting (avoiding a
+/// midnight-boundary race) and so unit tests can assert the timezone
+/// boundary against fixed synthetic timestamps.
+fn threads_today_is_stale(mission: &Mission, now: chrono::DateTime<chrono::Utc>) -> bool {
     if mission.threads_today == 0 {
         return false;
     }
     let Some(last) = mission.last_fire_at else {
         return true;
     };
-    let now = chrono::Utc::now();
     if let MissionCadence::Cron {
         timezone: Some(tz), ..
     } = &mission.cadence
@@ -661,7 +668,10 @@ impl MissionManager {
         // Best-effort persist: a transient store failure must not prevent the
         // mission from firing — the in-memory reset is sufficient for this
         // call, and the next successful fire will persist the counter naturally.
-        if threads_today_is_stale(&mission) {
+        // Pin `now` once so the staleness check and the cooldown comparison
+        // below cannot disagree across a midnight tick.
+        let now = chrono::Utc::now();
+        if threads_today_is_stale(&mission, now) {
             debug!(
                 mission_id = %id,
                 old_threads_today = mission.threads_today,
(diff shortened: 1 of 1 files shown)


ilblackdragon:
Thanks for the reviews. Pushed 4c15accd addressing the substantive findings:

**Adopted:**

- **Inject `now` into `threads_today_is_stale`** (Gemini, Copilot). The helper now takes `now: DateTime<Utc>` so unit tests can pin synthetic instants. `fire_mission` captures `now` once and reuses it for both the staleness check and the cooldown comparison below it, so they cannot disagree across a midnight tick.
- **Replace the non-deterministic Auckland integration test** (Copilot). The previous `threads_today_resets_at_cron_local_midnight` could pass with the timezone branch disabled depending on when in the day it ran — `last_local = yesterday 23:00 local` could land on the previous UTC date too. Replaced with deterministic synthetic-instant cases inside `threads_today_is_stale_predicate`. The new timezone case explicitly asserts the invariants: `last_utc.date_naive() == now_utc.date_naive()` (UTC-only check would NOT trigger) AND `last_local < now_local` in `Pacific/Auckland` (timezone check MUST trigger). I also added a same-instants/no-timezone case to confirm the tz branch is what actually moves the boundary.
- **Reword the doc comment** (Copilot) to drop the hard-coded "5 PM local" claim, which varies under DST.

**Rejected:**

- **Returning `false` when `last_fire_at = None`** (Gemini's "un-reapable" rule). This isn't a cleanup operation — it's a counter reset, where the safe direction is the *opposite*. Returning `false` here re-introduces the original permanent-exhaustion bug for missions whose `last_fire_at` was dropped by data corruption / migration / out-of-band edits. `fire_mission` writes `threads_today` and `last_fire_at` atomically, so the only way the counter is non-zero with a missing timestamp is data damage, and the recovery direction is to reset. Documented this in both the helper doc and the predicate test.

`cargo fmt --check` clean, `clippy --all-targets -D warnings` clean, all 86 mission tests pass.

serrrfirat (approved):
Reviewed as paranoid architect. Clean fix for #1945 — cron missions now reset `threads_today` on the local-day boundary instead of UTC, with `now` pinned once across the staleness check and the downstream cooldown comparison to close the midnight-tick race. Helper docs explain the edge cases (None last_fire_at, why `now` is injected). `cargo test threads_today` and `cargo clippy -p ironclaw_engine --all-targets -- -D warnings` pass on my checkout.

Two nits, non-blocking:
- The PR description lists `threads_today_resets_at_cron_local_midnight` as a separate test, but the Pacific/Auckland scenario is actually folded into `threads_today_is_stale_predicate`. Worth either renaming in the description or splitting it out as an integration test that drives the timezone-aware path through `fire_mission`/`tick` (the current cron+tick test uses `timezone: None` and would have passed the old UTC-only logic).
- Event-driven cadences (`OnEvent`/`OnSystemEvent`/`Webhook`) still reset on the UTC boundary because their cadence variants don't carry a timezone. Out of scope here, but worth a follow-up — same UX cliff for a Slack/Telegram bot mission in LA.

PR #2993, opened 2026-04-27
feat(reborn): durable event/audit substrate

zmanian:
## Summary

Adds `ironclaw_events` as a Reborn substrate crate carved from PR2 to land ahead of the filesystem grouped slice, per the integration plan in #2987 and the prior carve proposal in [#issuecomment-4329311786](https://github.com/nearai/ironclaw/issues/2987#issuecomment-4329311786).

This is the durable-log substrate the host runtime, dispatcher, process manager, and approval resolver will depend on for caller-level tests in PR3/PR4/PR6, and the foundation the JSONL/PostgreSQL/libSQL backends in later grouped PRs build on.

Tracking issue:

```text
#2987
```

Target branch:

```text
reborn-integration
```

Stacked on:

```text
reborn-land-01-foundation (PR1, #2988)
```

This PR is opened as draft because it depends on PR1 merging into `reborn-integration` first; the diff against `reborn-integration` will shrink to just the `ironclaw_events` crate plus workspace wiring once PR1 lands.

## Carve note

The `reborn-predispatch-obligations` stack already had a substantial `ironclaw_events` slice; this PR carves the contract substrate out of that work into a filesystem-free crate so it can land in PR2's first half and unblock downstream substrate work without the filesystem-backed JSONL sink. The byte-level `parse_jsonl` / `replay_jsonl` helpers are kept `pub` so PR2's `JsonlEventSink<F: RootFilesystem>` can build on the same redaction and replay invariants without duplicating logic.

The two responsibilities the existing stack bundled — best-effort sinks (`EventSink`/`AuditSink`) and explicit-error durable logs — are split here. The freeze packet's two-tier reliability model (sink failures best-effort vs. durable append failures explicit) becomes a type-level distinction rather than a per-callsite convention.

### `ironclaw_events`

- `RuntimeEvent` / `RuntimeEventKind` (8 kinds: `DispatchRequested`, `RuntimeSelected`, `DispatchSucceeded`, `DispatchFailed`, `ProcessStarted`, `ProcessCompleted`, `ProcessFailed`, `ProcessKilled`).
- Redaction-aware constructors with `sanitize_error_kind` collapsing any unsafe value into the stable `Unclassified` token.
- Approval-specific event kinds are deliberately absent; approval resolution is recorded as a control-plane `AuditEnvelope` per `events.md` §1.
- `EventSink` and `AuditSink` async traits (best-effort delivery) — failures must not alter runtime/control-plane outcomes.
- `DurableEventLog` and `DurableAuditLog` async traits (explicit-error append + scoped `read_after_cursor`).
  - Append failures are propagated.
  - A cursor older than the earliest retained record yields `EventError::ReplayGap` so transports request a snapshot/rebase rather than silently lose history (per `events-projections.md` §6 and the freeze addendum).
- `EventStreamKey = (tenant_id, user_id, agent_id)` partition; per-stream monotonic `EventCursor`; `EventLogEntry` / `EventReplay` shapes.
- Explicit zero-limit rejection (`EventError::InvalidReplayRequest`).
- `InMemoryEventSink` / `InMemoryAuditSink` (best-effort) and `InMemoryDurableEventLog` / `InMemoryDurableAuditLog` (durable, per-stream cursors) for caller-level tests.
- Public `parse_jsonl` / `replay_jsonl` byte-level helpers for downstream JSONL backends.

## Out of scope

- Filesystem-backed `JsonlEventSink<F: RootFilesystem>` and `JsonlAuditSink<F: RootFilesystem>` — deferred to PR2's filesystem half.
- PostgreSQL/libSQL durable backends — deferred to the database grouped PR.
- SSE/WebSocket transports — Wave 3 product integration.
- Projection reducers and read models — Wave 3.
- `scoped_runtime_event_log_path` / `scoped_audit_log_path` filesystem-aware path helpers — those compose with `RootFilesystem` and belong in PR2.
- Migration of existing event/audit JSONL files.

## Exposure checklist

- [x] Does not affect existing `src/` runtime behavior.
- [x] Does not alter app startup/config defaults.
- [x] Does not expose new routes/CLI/user-visible APIs.
- [x] Does not create a second writer for existing production state (no `src/` writers exist for these types yet).
- [x] All Reborn paths are unused by default; the crate has no integration with `src/` until later grouped PRs wire it in.
- [x] Status remains "implemented slice" for the substrate, not product-complete.

## Architecture boundary

`ironclaw_events` already had a `BoundaryRule` in `crates/ironclaw_architecture/tests/reborn_dependency_boundaries.rs` (lines 116-134) forbidding deps on auth/approvals/capabilities/dispatcher/extensions/host_runtime/secrets/network/mcp/processes/resources/run_state/scripts/wasm. That rule is now active with the crate present, and `cargo test -p ironclaw_architecture` passes.

`cargo tree -p ironclaw_events -e normal` confirms the only `ironclaw_*` normal dependency is `ironclaw_host_api`.

## Acceptance tests (caller-level)

Per the freeze packet's "test through the caller, not just the helper" rule, all tests drive the public trait surfaces, not internal helpers:

- `durable_event_log_appends_and_replays_in_order` — full append → read_after_cursor loop with cursor monotonicity.
- `read_after_next_cursor_returns_empty_replay` — cursor-resume idempotence.
- `replay_respects_limit_and_resumes_cleanly` — bounded replay across multiple resumes.
- `streams_partition_by_tenant_user_agent` — per-stream cursor isolation across `EventStreamKey` tuples.
- `read_empty_stream_returns_echoed_cursor` — empty stream returns echoed cursor for clean resume.
- `replay_with_zero_limit_is_rejected` / `replay_jsonl_with_zero_limit_is_rejected` — explicit invalid-request rejection.
- `dispatch_failed_redacts_unsafe_error_kind` — raw error message containing paths/secrets collapses to `Unclassified`.
- `dispatch_failed_preserves_safe_classification_token` — short stable tokens survive sanitization.
- `sanitize_error_kind_collapses_long_or_unsafe_input` — direct sanitizer pinning.
- `appended_event_payload_omits_raw_payloads_by_construction` — wire-shape pinning.
- `best_effort_event_sink_records_emit_calls` / `best_effort_audit_sink_captures_records` — best-effort delivery.
- `durable_audit_log_appends_and_replays` — `AuditEnvelope::denied(...)` round-trip.
- `approval_audit_records_partition_by_stream_key` — approval-resolution audit cross-stream isolation.
- `parse_jsonl_round_trips_runtime_events` — JSONL round-trip.
- `parse_jsonl_rejects_malformed_line_rather_than_silently_skipping` — explicit error rather than silent empty history (`events.md` §5).
- `replay_jsonl_advances_cursor_with_limit` — JSONL cursor semantics.

18 tests, all passing.

## Verification

```bash
cargo fmt --check
cargo test -p ironclaw_events
cargo test -p ironclaw_architecture
cargo clippy -p ironclaw_events --all-targets -- -D warnings
RUSTDOCFLAGS='-D warnings' cargo doc -p ironclaw_events --no-deps
cargo tree -p ironclaw_events -e normal
```

All pass.

## Open coordination questions

These don't block this PR but are flagged for the team:

1. **Cursor envelope ratification.** Wave 0 item 4 calls for ratifying the durable event cursor envelope. The shape here is conservative (opaque `EventCursor(u64)` + per-stream monotonic + explicit `ReplayGap` signaling). Treat this PR as the ratification proposal; tighten in review.
2. **Stream key partition.** Used `(tenant_id, user_id, agent_id)` matching the existing `scoped_runtime_event_log_path` convention. Deeper scope filtering (project/mission/thread/process/invocation) is intentionally a read-side filter rather than a separate cursor.
3. **`ironclaw_filesystem` boundary rule.** The existing `BoundaryRule` for `ironclaw_events` does not include `ironclaw_filesystem`. PR2's JSONL sink will need that dep, so leaving it out preserves PR2's freedom; if the team prefers a separate `ironclaw_events_jsonl` crate, the boundary can tighten then.
4. **Feature gate name.** No feature gating yet — the crate is internal substrate with no user-visible exposure. Per the prior comment on #2987, the workspace `reborn` feature should be defined in PR1 and applied uniformly when wiring begins.

### DIFF
--- a/crates/ironclaw_events/CLAUDE.md
+++ b/crates/ironclaw_events/CLAUDE.md
@@ -1,0 +1,28 @@
+# ironclaw_events
+
+Runtime/process events and control-plane audit envelope sinks plus durable
+append-log substrate for IronClaw Reborn.
+
+This crate is the substrate that downstream auth/dispatcher/process/runtime
+crates use to record what happened. It defines:
+
+- typed redacted [`RuntimeEvent`] records for already-authorized dispatch and
+  process lifecycle transitions;
+- redaction-aware constructors that collapse unsafe error detail into
+  `Unclassified` rather than leak it;
+- best-effort [`EventSink`] / [`AuditSink`] traits whose failures must not
+  alter runtime/control-plane outcomes;
+- explicit-error [`DurableEventLog`] / [`DurableAuditLog`] traits with a
+  monotonic per-scope cursor envelope and replay-after semantics;
+- in-memory durable backends used by tests and reference loops.
+
+Filesystem-backed JSONL durable backends and PostgreSQL/libSQL backends are
+deliberately deferred. They live in later grouped Reborn PRs that depend on
+`ironclaw_filesystem` and the database substrates. The byte-level
+`parse_jsonl` and `replay_jsonl` helpers in this crate are exposed so those
+later backends can build on the same redaction and replay invariants.
+
+Forbidden dependencies (enforced by `ironclaw_architecture`): authorization,
+approvals, capabilities, dispatcher, extensions, host_runtime, secrets,
+network, mcp, processes, resources, run_state, scripts, wasm, filesystem,
+memory.
--- a/crates/ironclaw_events/CLAUDE.md
+++ b/crates/ironclaw_events/CLAUDE.md
@@ -24,5 +24,12 @@ later backends can build on the same redaction and replay invariants.
 
 Forbidden dependencies (enforced by `ironclaw_architecture`): authorization,
 approvals, capabilities, dispatcher, extensions, host_runtime, secrets,
-network, mcp, processes, resources, run_state, scripts, wasm, filesystem,
-memory.
+network, mcp, processes, resources, run_state, scripts, wasm.
+
+`ironclaw_filesystem` is **deliberately not forbidden**: PR2's JSONL durable
+sink will depend on it, and pre-tightening here would force PR2 to relax the
+rule before it can land. `ironclaw_memory` is similarly not forbidden — this
+crate has no need for it today, but no boundary case has been made for
+adding it to the list. The authoritative forbidden list lives in
+`crates/ironclaw_architecture/tests/reborn_dependency_boundaries.rs`; if the
+two ever disagree, the test wins and this doc is stale.
--- a/crates/ironclaw_events/src/lib.rs
+++ b/crates/ironclaw_events/src/lib.rs
@@ -92,8 +92,12 @@ pub enum RuntimeEventKind {
 /// Redacted runtime event payload.
 ///
 /// All optional fields are absent unless meaningful for the event kind.
-/// `error_kind` is constrained by [`sanitize_error_kind`] so detail-like
-/// values (raw error text, paths, secrets) cannot leak through.
+/// `error_kind` is constrained by [`sanitize_error_kind`] both at construction
+/// time (through the typed `dispatch_failed` / `process_failed` constructors)
+/// and at deserialization time (a custom serde hook re-runs the sanitizer on
+/// any inbound JSONL/wire payload), so a JSONL replay or a hand-built record
+/// cannot smuggle raw error text, paths, or token-shaped secrets through this
+/// field.
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 pub struct RuntimeEvent {
     pub event_id: RuntimeEventId,
@@ -105,7 +109,18 @@ pub struct RuntimeEvent {
     pub runtime: Option<RuntimeKind>,
     pub process_id: Option<ProcessId>,
     pub output_bytes: Option<u64>,
+    #[serde(default, deserialize_with = "deserialize_error_kind")]
     pub error_kind: Option<String>,
+}
+
+/// Re-runs [`sanitize_error_kind`] on any inbound `error_kind` value so a
+/// hand-rolled or replayed JSONL record cannot bypass the redaction guard.
+fn deserialize_error_kind<'de, D>(deserializer: D) -> Result<Option<String>, D::Error>
+where
+    D: serde::Deserializer<'de>,
+{
+    let raw = Option::<String>::deserialize(deserializer)?;
+    Ok(raw.map(sanitize_error_kind))
 }
 
 impl RuntimeEvent {
@@ -282,22 +297,66 @@ struct RuntimeEventPayload {
     error_kind: Option<String>,
 }
 
+/// Stable token written to `RuntimeEvent.error_kind` whenever a caller-supplied
+/// value fails redaction.
+pub const UNCLASSIFIED_ERROR_KIND: &str = "Unclassified";
+
+const MAX_ERROR_KIND_LEN: usize = 64;
+const MAX_ERROR_KIND_SEGMENT_LEN: usize = 24;
+
 /// Collapse any error_kind value that does not match the stable classification
 /// shape into the single `Unclassified` token. This is the redaction guard
 /// that keeps raw error messages, paths, and stringified secrets out of
 /// durable runtime events.
+///
+/// Accepts only `lower_snake_case` identifiers with optional `.` or `:`
+/// separators (e.g. `missing_runtime_backend`, `wasm.host_http_denied`,
+/// `dispatch:timeout`). Rejects anything that resembles a path, free-form
+/// error text, JWT, base64 token, or API key:
+///
+/// - empty string;
+/// - longer than 64 bytes overall, or any dot/colon-separated segment longer
+///   than 24 bytes (defeats long random tokens);
+/// - characters outside `[a-z0-9_]` for body content, or `[._:]` separators;
+/// - leading character that is not a lowercase ASCII letter (defeats
+///   numeric-prefixed tokens, leading underscores, leading separators).
 pub fn sanitize_error_kind(error_kind: impl Into<String>) -> String {
-    let error_kind = error_kind.into();
-    let is_safe = !error_kind.is_empty()
-        && error_kind.len() <= 128
-        && error_kind
-            .bytes()
-            .all(|byte| byte.is_ascii_alphanumeric() || matches!(byte, b'_' | b'-' | b'.' | b':'));
-    if is_safe {
-        error_kind
+    let value = error_kind.into();
+    if is_safe_error_kind(&value) {
+        value
     } else {
-        "Unclassified".to_string()
+        UNCLASSIFIED_ERROR_KIND.to_string()
     }
+}
+
+fn is_safe_error_kind(value: &str) -> bool {
+    if value.is_empty() || value.len() > MAX_ERROR_KIND_LEN {
+        return false;
+    }
+    let first = value.as_bytes()[0];
+    if !first.is_ascii_lowercase() {
+        return false;
+    }
+    if value
+        .bytes()
+        .any(|byte| !is_error_kind_char(byte) && !matches!(byte, b'.' | b':'))
+    {
+        return false;
+    }
+    for segment in value.split(['.', ':']) {
+        if segment.is_empty() || segment.len() > MAX_ERROR_KIND_SEGMENT_LEN {
+            return false;
+        }
+        let segment_first = segment.as_bytes()[0];
+        if !segment_first.is_ascii_lowercase() {
+            return false;
+        }
+    }
+    true
+}
+
+fn is_error_kind_char(byte: u8) -> bool {
+    byte.is_ascii_lowercase() || byte.is_ascii_digit() || byte == b'_'
 }
 
 // -----------------------------------------------------------------------------
@@ -368,10 +427,11 @@ impl Default for EventCursor {
 /// requesting consumer's stream key. Deeper scope filtering (project,
 /// mission, thread, process, invocation) is applied as a read-side filter on
 /// the matching stream rather than as a separate cursor.
-#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
 pub struct EventStreamKey {
     pub tenant_id: TenantId,
     pub user_id: UserId,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
     pub agent_id: Option<AgentId>,
 }
 
@@ -400,7 +460,7 @@ impl EventStreamKey {
 }
 
 /// One replayed record and its durable cursor.
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 pub struct EventLogEntry<T> {
     pub cursor: EventCursor,
     pub record: T,
@@ -411,7 +471,7 @@ pub struct EventLogEntry<T> {
 /// `next_cursor` is suitable for the next `read_after_cursor` call. When
 /// `entries` is empty, `next_cursor` echoes the requested cursor so the
 /// consumer can resume cleanly.
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 pub struct EventReplay<T> {
     pub entries: Vec<EventLogEntry<T>>,
     pub next_cursor: EventCursor,
@@ -423,9 +483,21 @@ pub struct EventReplay<T> {
 
 /// Async event sink used by runtime/composition services.
 ///
-/// Best-effort observability. Sink failures are surfaced to the caller, which
-/// is expected to log/ignore them rather than alter runtime outcomes; see
-/// `events.md` §7.
+/// **Best-effort observability.** The contract requires that a sink failure
+/// **must not** change runtime outcomes. The trait returns `Result` so
+/// implementations can surface diagnostics to a separate observer/log,
+/// **never** so callers can `?`-propagate the error and short-circuit the
+/// surrounding workflow.
+///
+/// Callers (dispatcher, process manager, host runtime) must:
+///
+/// 1. invoke `emit(...).await`;
+/// 2. record any returned error to a diagnostics channel of their choice;
+/// 3. continue with their original success/failure result.
+///
+/// A type-level enforcement of this contract (no-fail emit + separate
+/// fallible diagnostics surface) is a deliberate follow-up; see the
+/// "best-effort sink contract" follow-up issue.
 #[async_trait]
 pub trait EventSink: Send + Sync {
     async fn emit(&self, event: RuntimeEvent) -> Result<(), EventError>;
@@ -433,8 +505,10 @@ pub trait EventSink: Send + Sync {
 
 /// Async audit sink used by control-plane services.
 ///
-/// Best-effort observability. Sink failures must not change approval
-/// resolution outcomes; see `events.md` §3.
+/// **Best-effort observability.** Same contract as [`EventSink`]: a sink
+/// failure must not change approval resolution outcomes. The trait returns
+/// `Result` so implementations can surface diagnostics, never so callers can
+/// short-circuit on a sink error.
 #[async_trait]
 pub trait AuditSink: Send + Sync {
     async fn emit_audit(&self, record: AuditEnvelope) -> Result<(), EventError>;
@@ -556,17 +630,34 @@ impl<T> Default for StreamState<T> {
 }
 
 impl<T: Clone> StreamState<T> {
-    fn append(&mut self, record: T) -> EventLogEntry<T> {
-        self.next_cursor += 1;
+    fn append(&mut self, record: T) -> Result<EventLogEntry<T>, EventError> {
+        let next = self
+            .next_cursor
+            .checked_add(1)
+            .ok_or_else(|| EventError::DurableLog {
+                reason: "event cursor overflowed u64; durable log exhausted".to_string(),
+            })?;
+        self.next_cursor = next;
         let entry = EventLogEntry {
-            cursor: EventCursor::new(self.next_cursor),
+            cursor: EventCursor::new(next),
             record,
         };
         self.entries.push(entry.clone());
-        entry
+        Ok(entry)
     }
 
     fn read_after(&self, after: EventCursor, limit: usize) -> Result<EventReplay<T>, EventError> {
+        // A cursor that points beyond the current head is a contract
+        // violation, not a benign no-op: returning empty would silently lose
+        // every event 1..=head once it lands. Surface as ReplayGap so the
+        // caller is forced to request a snapshot/rebase and re-derive a
+        // cursor that belongs to this stream.
+        if after.as_u64() > self.next_cursor {
+            return Err(EventError::ReplayGap {
+                requested: after,
+                earliest: EventCursor::new(self.next_cursor),
+            });
+        }
         if self.earliest_retained > 0 && after.as_u64() < self.earliest_retained.saturating_sub(1) {
             return Err(EventError::ReplayGap {
                 requested: after,
@@ -589,6 +680,21 @@ impl<T: Clone> StreamState<T> {
             next_cursor,
         })
     }
+
+    /// Discard entries whose cursor is `<=` the supplied cursor and advance
+    /// `earliest_retained` so subsequent reads with stale cursors return
+    /// [`EventError::ReplayGap`]. Used by retention policies in production
+    /// backends and by tests that exercise the gap path.
+    fn truncate_before_or_at(&mut self, cursor: EventCursor) {
+        let bound = cursor.as_u64();
+        if bound == 0 {
+            return;
+        }
+        self.entries.retain(|entry| entry.cursor.as_u64() > bound);
+        if bound >= self.earliest_retained {
+            self.earliest_retained = bound + 1;
+        }
+    }
 }
 
 /// In-memory durable runtime event log with per-stream monotonic cursors.
@@ -601,6 +707,26 @@ impl InMemoryDurableEventLog {
     pub fn new() -> Self {
         Self::default()
     }
+
+    /// Drop entries whose cursor is `<=` the supplied cursor for the given
+    /// stream and advance the stream's earliest-retained marker so subsequent
+    /// reads against older cursors return [`EventError::ReplayGap`].
+    ///
+    /// Production backends apply this from a retention policy. Tests use it
+    /// to exercise the gap path without coupling to a specific policy.
+    pub fn truncate_before_or_at(
+        &self,
+        stream: &EventStreamKey,
+        cursor: EventCursor,
+    ) -> Result<(), EventError> {
+        let mut streams = self.streams.lock().map_err(|_| EventError::DurableLog {
+            reason: "in-memory durable event log lock poisoned".to_string(),
+        })?;
+        if let Some(state) = streams.get_mut(stream) {
+            state.truncate_before_or_at(cursor);
+        }
+        Ok(())
+    }
 }
 
 #[async_trait]
@@ -611,7 +737,7 @@ impl DurableEventLog for InMemoryDurableEventLog {
             reason: "in-memory durable event log lock poisoned".to_string(),
         })?;
         let stream = streams.entry(key).or_default();
-        Ok(stream.append(event))
+        stream.append(event)
     }
 
     async fn read_after_cursor(
@@ -631,10 +757,23 @@ impl DurableEventLog for InMemoryDurableEventLog {
         })?;
         match streams.get(stream) {
             Some(state) => state.read_after(after, limit),
-            None => Ok(EventReplay {
-                entries: Vec::new(),
-                next_cursor: after,
-            }),
+            None => {
+                // An absent stream is at head-zero. Any cursor beyond origin
+                // is a foreign cursor that this stream has never issued, so
+                // surface a gap rather than silently echoing the cursor and
+                // hiding events 1..after if/when the stream starts.
+                if after.as_u64() > 0 {
+                    Err(EventError::ReplayGap {
+                        requested: after,
+                        earliest: EventCursor::origin(),
+                    })
+                } else {
+                    Ok(EventReplay {
+                        entries: Vec::new(),
+                        next_cursor: after,
+                    })
+                }
+            }
         }
     }
 }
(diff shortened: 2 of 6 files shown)


zmanian:
Updated in response to a Codex code review. The review flagged four contract-level concerns; all four are now addressed in commit 072a26e6.

### Fixed

1. **Silent replay loss on future cursor.** `read_after_cursor` previously echoed back any caller-supplied cursor for an absent stream or one whose head was lower, silently losing events 1..cursor when they later landed. Now returns `EventError::ReplayGap` for both cases (`after > head` for populated streams; `after > 0` for absent streams). The earlier test that pinned the wrong behavior is split into the correct empty-stream-at-origin case plus two new tests proving gap signaling on future cursors.

2. **`RuntimeEvent` deserialize bypassed `sanitize_error_kind`.** Public field + derived `Deserialize` meant a hand-rolled or replayed JSONL record could put a raw secret into `error_kind`. Added a `deserialize_with` hook that re-runs the sanitizer on inbound values; new test `deserialize_runtime_event_resanitizes_error_kind` proves a hand-mutated JSON value with a token-shaped string in `error_kind` deserializes to `Unclassified`.

3. **`sanitize_error_kind` allowed token-shaped secrets.** The old allowlist accepted real-world API key shapes. Tightened to: leading lowercase letter required; body alphabet `[a-z0-9_]` (dash dropped); overall length ≤ 64; each `.:`-separated segment ≤ 24 bytes and must itself start with a lowercase letter. Test expanded to cover token shapes, leading digits, leading underscores, mixed case, oversized random segments.

4. **Cursor overflow.** `StreamState::append` now uses `checked_add` and surfaces `EventError::DurableLog` on u64 exhaustion rather than panicking on wrap.

### Added in response to recommendations

5. **`ReplayGap` retention path tested.** Added `pub truncate_before_or_at(stream, cursor)` on both in-memory logs that drops entries up to and including the supplied cursor and advances `earliest_retained`. Test `replay_gap_after_truncation_forces_snapshot_rebase` exercises the end-to-end retention → gap path.

6. **Audit redaction has a serialization assertion.** New test `approval_audit_envelope_serialization_excludes_raw_reason_and_fingerprint` builds an `ApprovalRequest` with a unique sentinel `reason`, runs it through `AuditEnvelope::approval_resolved`, and asserts the JSON output does not contain the sentinel or an `invocation_fingerprint` field.

7. **Serde derives on `EventStreamKey`, `EventLogEntry<T>`, `EventReplay<T>`** for downstream gateway/SSE consumers.

### Documented; deferred

8. **Best-effort sink contract.** `EventSink::emit` / `AuditSink::emit_audit` still return `Result` because a no-fail emit + separate fallible diagnostics surface is a larger contract change that affects every caller. The trait doc comments now state the contract loudly: callers must record the error to a diagnostics channel and continue with their original outcome. Type-level enforcement is flagged as a follow-up.

### Verification

- `cargo fmt --check` ✓
- `cargo test -p ironclaw_events` — 23/23 pass (was 18; added 5 new tests)
- `cargo test -p ironclaw_architecture` — boundary rule still active and green
- `cargo clippy -p ironclaw_events --all-targets -- -D warnings` ✓
- `RUSTDOCFLAGS='-D warnings' cargo doc -p ironclaw_events --no-deps` ✓
- `cargo tree -p ironclaw_events -e normal` — only `ironclaw_host_api` as ironclaw_ dep

Heads up on the force-push: the original commit had a test fixture string `sk_live_...` (the very token shape I was asserting redaction against), which GitHub's push protection correctly flagged as a Stripe API key pattern. Replaced with a structurally identical but non-vendor `xkey_demo_...` fixture; tests still cover the same redaction property.

crates/ironclaw_events/src/lib.rs
```diff
@@ -0,0 +1,939 @@
...
+            })
+        })
+        .collect()
+}
+
+/// Replay a JSONL byte slice after a cursor with a bounded limit.
+///
+/// Used by JSONL-backed durable log adapters in later grouped Reborn PRs.
+/// The cursor is the 1-based line index of the last consumed record.
+pub fn replay_jsonl<T>(
```
serrrfirat, line 1075: **High Severity** — `replay_jsonl` accepts future/foreign cursors and can silently skip history.

The in-memory durable log correctly treats a cursor beyond the stream head as a `ReplayGap`, but this helper does not. If a JSONL log has 2 records and a caller asks for `after = EventCursor::new(99)`, the loop finishes with `current_cursor = 2` and returns `Ok(empty, next_cursor = 99)`. A future filesystem JSONL backend built on this helper would then silently accept a cursor this stream never issued, which is exactly the replay-loss case the durable-log contract is trying to prevent.

Concrete fix: after scanning, if `after > current_cursor`, return `EventError::ReplayGap { requested: EventCursor::new(after), earliest: EventCursor::new(current_cursor) }` (or a documented head/earliest value), and add a `replay_jsonl_with_future_cursor_returns_replay_gap` test.

crates/ironclaw_events/src/lib.rs
```diff
@@ -0,0 +1,939 @@
...
+    pub timestamp: Timestamp,
+    pub kind: RuntimeEventKind,
+    pub scope: ResourceScope,
+    pub capability_id: CapabilityId,
+    pub provider: Option<ExtensionId>,
+    pub runtime: Option<RuntimeKind>,
+    pub process_id: Option<ProcessId>,
+    pub output_bytes: Option<u64>,
+    #[serde(default, deserialize_with = "deserialize_error_kind")]
+    pub error_kind: Option<String>,
```
serrrfirat, line 122: **High Severity** — public `RuntimeEvent` fields let direct construction bypass error redaction.

The docs say a hand-built record cannot smuggle raw error text because deserialization re-runs `sanitize_error_kind`, but that only protects inbound JSON. Since all fields are public and `Serialize` is derived, any in-process caller can construct:

```rust
RuntimeEvent { error_kind: Some("/Users/alice/token=secret".into()), ... }
```

and then append/serialize it without hitting `sanitize_error_kind`. That violates the event redaction invariant before durable backends are added.

Concrete fix: make `RuntimeEvent` fields private and expose constructors/getters, introduce a sanitized `ErrorKind` newtype for the field, or implement custom `Serialize` that re-sanitizes `error_kind` before writing. Please also add a test that manually constructs an unsafe event and verifies serialization/append redacts it.

crates/ironclaw_events/src/lib.rs
```diff
@@ -0,0 +1,939 @@
...
+/// `append` failures must be propagated. `read_after_cursor` is gated on
+/// stream key authority: callers must validate that the requested
+/// [`EventStreamKey`] matches the consumer's authorized scope before serving
+/// the result, and the implementation rejects cursors that predate the
+/// earliest retained entry with [`EventError::ReplayGap`].
+#[async_trait]
+pub trait DurableEventLog: Send + Sync {
+    async fn append(&self, event: RuntimeEvent) -> Result<EventLogEntry<RuntimeEvent>, EventError>;
+
+    async fn read_after_cursor(
```
serrrfirat, line 669: **Medium Severity** — replay authorization is easy to get wrong because stream keys omit project/mission/thread scope.

`EventStreamKey` partitions only by `(tenant, user, agent)`, while the comment says project/mission/thread/process/invocation filtering is a read-side filter. However, `read_after_cursor` takes only `stream`, `after`, and `limit`, so the durable-log trait itself returns every record in that user/agent stream and gives implementers no typed authorized-scope/filter input to enforce deeper isolation.

A project-scoped consumer for project A could receive project B events for the same tenant/user/agent if the caller forgets to post-filter. Given this is the substrate future SSE/projection callers will build on, relying on every caller to remember that filter is a risky authority boundary.

Concrete fix: either include `project_id` in the v1 stream key, or add an explicit authorized scope/filter parameter to `read_after_cursor` and make implementations enforce it. Please add a cross-project isolation test that appends two same-user/same-agent events with different `project_id`s and proves a project-scoped read cannot see the other project.

crates/ironclaw_events/src/lib.rs
zmanian: Fixed in 901dba3e. `replay_jsonl` now matches `InMemoryDurableEventLog` semantics: after scanning, if `requested > current_cursor`, returns `EventError::ReplayGap { requested, earliest: current_cursor }`. Added `replay_jsonl_with_future_cursor_returns_replay_gap` test.

crates/ironclaw_events/src/lib.rs
zmanian: Fixed in 901dba3e. Custom Serialize impl now re-runs `sanitize_error_kind` on the way out, symmetric with the existing Deserialize hook. Direct construction `RuntimeEvent { error_kind: Some(raw), .. }` followed by serde_json/append still emits `"error_kind":"Unclassified"`. Test `direct_construction_serialize_path_resanitizes_error_kind` builds a path-shaped raw value and asserts the wire payload is sanitized.

crates/ironclaw_events/src/lib.rs
zmanian: Fixed in 901dba3e. Added a typed `ReadScope { project_id, mission_id, thread_id, process_id }` filter and made it a required argument on `DurableEventLog::read_after_cursor` and `DurableAuditLog::read_after_cursor`. Implementations enforce the filter; cursors advance through filtered-out records so consumers can resume. Filter is a tightening — a record's `None` field cannot match a filter that asks for `Some(...)`. Test `read_scope_filter_isolates_project_within_same_stream` appends three same-stream events across two project_ids and asserts a project-A-scoped read sees only project A; `read_scope_filter_excludes_records_with_none_field_when_filter_is_some` pins the tightening semantics. `ReadScope::any()` exists for tests/admin paths but production callers must construct a tightened filter, removing the post-filter caller-discipline foot gun.

zmanian:
Updated in response to the latest review (Firat HIGH/HIGH/MEDIUM + Codex P1). All four addressed in 901dba3e; replied to each inline thread above.

### Fixed

1. **`replay_jsonl` accepted future cursors** (Codex P1 + Firat HIGH at lib.rs:890). Same gap-on-future-cursor bug the previous fix closed in `InMemoryDurableEventLog`, missed in the byte-level helper that PR2's filesystem-backed sink will build on. Now returns `ReplayGap { requested, earliest: current_cursor }` after scanning. Test `replay_jsonl_with_future_cursor_returns_replay_gap`.

2. **`RuntimeEvent` direct-construction Serialize bypassed redaction** (Firat HIGH at lib.rs:113). The Deserialize hook from the previous fix protected inbound JSONL but left direct in-process construction unguarded. Custom Serialize impl now re-runs `sanitize_error_kind` on the way out, symmetric with the Deserialize hook. Test `direct_construction_serialize_path_resanitizes_error_kind`.

3. **`EventStreamKey` left deeper-scope filtering to caller discipline** (Firat MEDIUM at lib.rs:533). Added typed `ReadScope { project_id, mission_id, thread_id, process_id }` and made it a required arg on `DurableEventLog::read_after_cursor` / `DurableAuditLog::read_after_cursor`. Implementations enforce the filter; tightening semantics (record's `None` cannot match filter's `Some`); cursors advance through filtered-out records. Tests `read_scope_filter_isolates_project_within_same_stream` and `read_scope_filter_excludes_records_with_none_field_when_filter_is_some`.

### Verification

- `cargo test -p ironclaw_events` — **27/27 pass** (was 23; 4 new tests).
- `cargo fmt --check` ✓
- `cargo clippy -p ironclaw_events --all-targets -- -D warnings` ✓
- `RUSTDOCFLAGS='-D warnings' cargo doc -p ironclaw_events --no-deps` ✓
- `cargo test -p ironclaw_architecture` — boundary rule still active ✓

### Note on the two PR1 findings

The review thread also surfaced two findings that originate in PR1 (#2988) rather than this PR — `canonical_json` recursion DoS in `ironclaw_host_api/src/approval.rs:86` and negative `Decimal` USD estimates in `ironclaw_resources/src/lib.rs:263`. Both are real concerns. Surfacing them on PR1 separately so they don't get lost.

crates/ironclaw_events/CLAUDE.md
```diff
@@ -0,0 +1,28 @@
...
+Filesystem-backed JSONL durable backends and PostgreSQL/libSQL backends are
+deliberately deferred. They live in later grouped Reborn PRs that depend on
+`ironclaw_filesystem` and the database substrates. The byte-level
+`parse_jsonl` and `replay_jsonl` helpers in this crate are exposed so those
+later backends can build on the same redaction and replay invariants.
+
+Forbidden dependencies (enforced by `ironclaw_architecture`): authorization,
+approvals, capabilities, dispatcher, extensions, host_runtime, secrets,
+network, mcp, processes, resources, run_state, scripts, wasm, filesystem,
+memory.
```
serrrfirat: **Medium Severity** — CLAUDE.md vs `BoundaryRule` drift on the forbidden-deps list.

This file lists `filesystem` and `memory` as forbidden deps for `ironclaw_events`. The actual `BoundaryRule` for `ironclaw_events` in `crates/ironclaw_architecture/tests/reborn_dependency_boundaries.rs` does **not** include either:

```rust
crate_name: "ironclaw_events",
forbidden: vec![
    "ironclaw_authorization", "ironclaw_approvals", "ironclaw_capabilities",
    "ironclaw_dispatcher", "ironclaw_extensions", "ironclaw_host_runtime",
    "ironclaw_secrets", "ironclaw_network", "ironclaw_mcp", "ironclaw_processes",
    "ironclaw_resources", "ironclaw_run_state", "ironclaw_scripts", "ironclaw_wasm",
],
```

The PR description's "Open coordination questions" item 3 explicitly says omitting `ironclaw_filesystem` is intentional ("PR2's JSONL sink will need that dep, so leaving it out preserves PR2's freedom"). So the doc is the wrong file — it's overly restrictive vs the actual enforcement.

Two ways this misleads future contributors:

1. A reader trusting the doc would believe adding `ironclaw_filesystem` will be rejected by the boundary test. It won't — the test passes regardless.
2. If someone fixes the doc by tightening the rule (the natural reading of "doc says forbidden, code says allowed → make code match doc") they'd break PR2's premise.

Either:
- Remove `filesystem` and `memory` from this list and add a note explaining `filesystem` is deliberately allowed for PR2's JSONL sink (matching the open question), or
- Add `ironclaw_filesystem` and `ironclaw_memory` to the `BoundaryRule` if the team prefers tightening over PR2 freedom (which is the discussion the PR description is asking for).

Either way the doc and the rule should agree.

crates/ironclaw_events/src/lib.rs
```diff
@@ -0,0 +1,1107 @@
...
+    /// [`EventError::ReplayGap`]. Used by retention policies in production
+    /// backends and by tests that exercise the gap path.
+    fn truncate_before_or_at(&mut self, cursor: EventCursor) {
+        let bound = cursor.as_u64();
+        if bound == 0 {
+            return;
+        }
+        self.entries.retain(|entry| entry.cursor.as_u64() > bound);
+        if bound >= self.earliest_retained {
+            self.earliest_retained = bound + 1;
```
serrrfirat, line 866: **Medium Severity** — `truncate_before_or_at` doesn't validate `cursor <= next_cursor`, leaving the stream permanently bricked on misuse.

If a caller passes a cursor beyond the current head:

```rust
fn truncate_before_or_at(&mut self, cursor: EventCursor) {
    let bound = cursor.as_u64();
    if bound == 0 { return; }
    self.entries.retain(|entry| entry.cursor.as_u64() > bound);
    if bound >= self.earliest_retained {
        self.earliest_retained = bound + 1;
    }
}
```

Trace with `next_cursor = 5`, caller passes `bound = 99`:

1. All 5 existing entries are retained-out (none have cursor > 99).
2. `earliest_retained = 100`.
3. Stream state: `next_cursor = 5`, `earliest_retained = 100`, `entries = []`.

Now the caller appends a fresh event:
- `next_cursor = 6`, entry at cursor 6.

The caller (or any other consumer) reads with `after = 5` (or any cursor in `[0, 99]`):
- `read_after`: `5 > next_cursor (6)` → false; `earliest_retained > 0 (100 > 0) && after < earliest_retained - 1 (5 < 99)` → true → `ReplayGap { requested: 5, earliest: 100 }`.

But cursor 100 doesn't exist! The caller follows the gap protocol, requests a snapshot, gets back cursor 6 from the snapshot, and reads with `after = 6`:
- `6 > 6` → false; `6 < 99` → true → still `ReplayGap { earliest: 100 }`.

The stream is **permanently bricked** until ~94 more appends bring `next_cursor` up to `earliest_retained`.

Trigger requires misuse — a retention policy that picks a cursor by inspecting an existing entry won't hit this. But:

- The function is `pub` on `InMemoryDurableEventLog` and `InMemoryDurableAuditLog`.
- The docstring just says "Discard entries whose cursor is `<=` the supplied cursor and advance `earliest_retained`" with no precondition.
- A "drop everything older than 24h" retention policy implemented in calendar time (not cursor time) could easily produce a bound > head if the stream is quiet.
- Production backends (PostgreSQL, libSQL) will likely inherit this contract from the in-memory reference impl. The reference impl is what callers will read first to learn the semantics.

Suggest one of:

```rust
// Option A: clamp
let bound = cursor.as_u64().min(self.next_cursor);

// Option B: reject (less surprising for retention policies)
if bound > self.next_cursor {
    return Err(EventError::InvalidReplayRequest {
        reason: format!(
            "truncation cursor {bound} exceeds stream head {head}",
            head = self.next_cursor,
        ),
    });
}

// Option C: document the precondition loudly
/// **Caller must ensure `cursor <= current head cursor`.** Passing a cursor
/// beyond the head leaves the stream in a permanently-gapped state until
/// appends catch up.
```

Option B is the cleanest type-level enforcement; it requires changing the signature to return `Result`. Option A is a one-line fix that preserves the existing signature.

crates/ironclaw_events/src/lib.rs
```diff
@@ -0,0 +1,1107 @@
...
+                reason: error.to_string(),
+            })
+        })
+        .collect()
+}
+
+/// Replay a JSONL byte slice after a cursor with a bounded limit.
+///
+/// Used by JSONL-backed durable log adapters in later grouped Reborn PRs.
+/// The cursor is the 1-based line index of the last consumed record.
```
serrrfirat, line 1067: **Low Severity** — Docstring promises "1-based line index of the last consumed record" but doesn't warn about the compaction case.

The helper assigns `EventCursor::new(current_cursor)` where `current_cursor` is incremented per non-empty line. This works cleanly when the JSONL file is append-only and never compacted.

But future filesystem JSONL backends will likely implement retention/compaction (drop old entries). After compaction, line 1 of the file may correspond to cursor 1234 in the logical stream, not cursor 1. Calling `replay_jsonl(bytes, after=1233, limit)` on the compacted file would:

- Increment `current_cursor` from 1 to 5 (the line count) instead of 1234 to 1238.
- Falsely return `ReplayGap { requested: 1233, earliest: 5 }` — a meaningless `earliest` value.

A backend that compacts would need to:
1. Store the cursor inline in the JSONL line (e.g. `{"cursor": 1234, "record": {...}}`) and use a different parser, OR
2. Maintain an out-of-band mapping of file-offset → cursor, OR
3. Never compact — keep using line-index = cursor.

The current docstring leaves a future implementer to discover this on their own. Suggest one extra sentence:

```rust
/// Replay a JSONL byte slice after a cursor with a bounded limit.
///
/// Used by JSONL-backed durable log adapters in later grouped Reborn PRs.
/// The cursor is the 1-based line index of the last consumed record.
///
/// **Assumes uncompacted JSONL.** Backends that compact entries (drop old
/// records to reclaim disk) must not use this helper directly — line index
/// will desynchronize from the logical cursor. Such backends should either
/// store the cursor inline in each record or maintain an out-of-band offset
/// map.
```

Pre-emptive concern; the JSONL-backed sinks are deferred to a later grouped PR per the carve note. Worth pinning the assumption now while the helper is the only public API for cursor extraction.

serrrfirat:
## Paranoid review summary — durable event/audit substrate

This PR has been through two prior review rounds (Codex P1 + my earlier review on the typed substrate). Both rounds' concerns are now addressed in the latest commits — `direct_construction_serialize_path_resanitizes_error_kind`, `read_scope_filter_isolates_project_within_same_stream`, `replay_jsonl_with_future_cursor_returns_replay_gap`, and the tightened `sanitize_error_kind` allowlist all pin the contract end-to-end. The redaction posture on `RuntimeEvent` is now defense-in-depth across constructor + Serialize + Deserialize, which is the right shape for a substrate every downstream Reborn crate will write through.

Three new findings posted inline:

1. **Medium** — `crates/ironclaw_events/CLAUDE.md` lists `filesystem` and `memory` in the forbidden-deps list, but the actual `BoundaryRule` does not. The PR description's open question #3 says omitting `filesystem` is intentional for PR2's JSONL sink, so the doc is the wrong file. Either fix the doc or tighten the rule.
2. **Medium** — `StreamState::truncate_before_or_at` doesn't validate `cursor <= next_cursor`. A bound beyond the head leaves the stream permanently bricked: every read returns false `ReplayGap` until appends catch up. Public on both in-memory durable logs; docstring promises no precondition. Suggest clamping or rejecting.
3. **Low** — `replay_jsonl` docstring doesn't warn that "cursor = line index" assumes uncompacted JSONL. Future filesystem backends that compact entries must not use this helper directly. One extra sentence in the docstring would prevent future confusion.

### Cross-PR pattern observation (informational)

Asymmetric redaction posture between `RuntimeEvent` (this PR) and `AuditEnvelope` (PR1, `ironclaw_host_api`):

- `RuntimeEvent` enforces `sanitize_error_kind` on the wire boundary via custom `Serialize`/`Deserialize` impls. Direct struct-literal construction can't bypass it.
- `AuditEnvelope` has constructor-only redaction with default-derived `Serialize`/`Deserialize`. A direct struct-literal `AuditEnvelope { decision: DecisionSummary { kind: format!("err: {}", raw_secret), .. }, .. }` persists through any durable audit append unredacted.

This PR's CLAUDE.md asserts the redaction contract for *both* event and audit envelopes, but the test suite (`approval_audit_envelope_serialization_excludes_raw_reason_and_fingerprint`) only validates the `approval_resolved` constructor path — not the wire boundary against direct construction. Any caller that builds an `AuditEnvelope` from raw fields can leak. This belongs in PR1 to fix (mirror the `RuntimeEvent` pattern with `ActionSummary.target` and `DecisionSummary.kind` as the redaction targets), but it's visible from here because the contract claim originates here.

### Stale-base note

The diff against `reborn-integration` currently includes inherited PR1 content (host_api, resources, architecture) plus a small "deletion" in `ironclaw_resources/tests/resource_governor_contract.rs` that's actually `df331e541` not yet in this PR's ancestry. Once #2988 lands and this PR rebases, the diff will collapse to just `ironclaw_events` + workspace wiring as the PR description promises. Not a finding — just noting for future reviewers who diff against the wrong base.

### v2 Engine impact

**Not related** — substrate-only PR, no v1 modules touched, exposure checklist confirms zero `src/` runtime changes.

### What's solid

- 27 caller-driven tests covering append/replay/truncation/scope-filter/redaction/sanitization/JSONL cursor semantics. The "test through the caller, not just the helper" rule is followed throughout.
- `EventCursor` overflow uses `checked_add` and surfaces as `EventError::DurableLog` rather than panicking.
- Defense-in-depth on `RuntimeEvent` redaction across constructor + Serialize + Deserialize.
- Cursor envelope semantics: explicit `ReplayGap` for foreign cursors, future cursors against absent streams, post-truncation gap with rebase signal, scope filter that advances the cursor through filtered-out records.
- Best-effort vs explicit-error sink contracts cleanly separated at the type level.
- `BoundaryRule` is active and tested (modulo the doc/rule drift in finding #1).

LGTM with the three suggestions above; finding #2 (truncation footgun) is the most concrete bug worth landing before downstream backends inherit the contract.

crates/ironclaw_events/CLAUDE.md
zmanian: Fixed in 3ed34b72. Removed `filesystem` and `memory` from the forbidden-deps list in `crates/ironclaw_events/CLAUDE.md` so the doc matches the actual `BoundaryRule` in `crates/ironclaw_architecture/tests/reborn_dependency_boundaries.rs`. Added a note that `ironclaw_filesystem` is deliberately allowed for PR2's JSONL sink (per the PR description's open-coordination question) and that the test is the authoritative source if the two ever drift again.

crates/ironclaw_events/src/lib.rs
zmanian: Fixed in 3ed34b72. Took Option B: `StreamState::truncate_before_or_at` now returns `Result<(), EventError>` and rejects cursors past `next_cursor` with `InvalidReplayRequest { reason: "truncation cursor {bound} exceeds stream head {head}" }`. The public `InMemoryDurableEventLog::truncate_before_or_at` and `InMemoryDurableAuditLog::truncate_before_or_at` already returned `Result`, so the surface is unchanged — the error now propagates instead of silently bricking the stream. Added regression test `truncate_beyond_head_is_rejected` that asserts the rejection and that the stream remains usable for subsequent reads. Docstring on the public method also updated to call out the new precondition.

crates/ironclaw_events/src/lib.rs
zmanian: Fixed in 3ed34b72. Added the suggested compaction warning to the `replay_jsonl` docstring: backends that compact entries must not use this helper directly because line index will desynchronize from the logical cursor and `ReplayGap.earliest` will be meaningless. The doc now spells out the two acceptable strategies (cursor inline per record, or out-of-band offset→cursor map) so a future implementer doesn't rediscover this on their own.

zmanian:
Addressed the three review comments from @serrrfirat at https://github.com/nearai/ironclaw/pull/2993#pullrequestreview-4183958046, https://github.com/nearai/ironclaw/pull/2993#pullrequestreview-4183960160, and https://github.com/nearai/ironclaw/pull/2993#pullrequestreview-4183962094 in 3ed34b72:

- **CLAUDE.md ↔ BoundaryRule drift**: dropped `filesystem` and `memory` from the forbidden-deps list in `crates/ironclaw_events/CLAUDE.md` to match the actual rule, and called out `ironclaw_filesystem` as deliberately allowed for PR2's JSONL sink.
- **`truncate_before_or_at` brick risk**: `StreamState::truncate_before_or_at` now rejects cursors past stream head with `InvalidReplayRequest` instead of silently advancing `earliest_retained` past `next_cursor`. New test `truncate_beyond_head_is_rejected` covers it.
- **`replay_jsonl` compaction footgun**: docstring now spells out the uncompacted-JSONL assumption and the two acceptable strategies for compacting backends.

`cargo test -p ironclaw_events` passes (28 tests). `cargo clippy -p ironclaw_events --all-targets` clean.

serrrfirat (approved):

PR #2997, opened 2026-04-27
fix(reborn): close foundation fail-closed gaps

serrrfirat:
## Summary

Fixes two fail-closed foundation issues surfaced from downstream PR review feedback:

- Caps approval fingerprint canonical JSON traversal depth so deeply nested invocation input returns a controlled `HostApiError` instead of risking stack overflow.
- Rejects negative USD resource estimates before reservation state mutation so reservations cannot credit active reserved balances.

## Changes

- `ironclaw_host_api`
  - Makes `canonical_json(...)` return `Result<serde_json::Value, HostApiError>`.
  - Adds a `MAX_CANONICAL_JSON_DEPTH` guard.
  - Adds a 10k-deep nested input regression test for invocation fingerprints.

- `ironclaw_resources`
  - Adds `ResourceError::InvalidEstimate`.
  - Validates `estimate.usd >= 0` in `reserve_with_id(...)` before acquiring/mutating reservation state.
  - Adds a regression test proving negative USD estimates are rejected and reserved balance remains unchanged.

## Verification

```bash
cargo test -p ironclaw_host_api -p ironclaw_resources
cargo clippy -p ironclaw_host_api -p ironclaw_resources --all-targets -- -D warnings
cargo fmt --check
git diff --check
```

### DIFF
--- a/crates/ironclaw_host_api/src/approval.rs
+++ b/crates/ironclaw_host_api/src/approval.rs
@@ -47,7 +47,7 @@ impl InvocationFingerprint {
             input: &'a serde_json::Value,
         }
 
-        let canonical_input = canonical_json(input);
+        let canonical_input = canonical_json(input)?;
         let payload = Payload {
             version: 1,
             kind: "dispatch",
@@ -67,21 +67,38 @@ impl InvocationFingerprint {
     }
 }
 
-fn canonical_json(value: &serde_json::Value) -> serde_json::Value {
+const MAX_CANONICAL_JSON_DEPTH: usize = 64;
+
+fn canonical_json(value: &serde_json::Value) -> Result<serde_json::Value, HostApiError> {
+    canonical_json_at_depth(value, 0)
+}
+
+fn canonical_json_at_depth(
+    value: &serde_json::Value,
+    depth: usize,
+) -> Result<serde_json::Value, HostApiError> {
+    if depth > MAX_CANONICAL_JSON_DEPTH {
+        return Err(HostApiError::invariant(
+            "canonical_json: max depth exceeded",
+        ));
+    }
+
     match value {
-        serde_json::Value::Array(items) => {
-            serde_json::Value::Array(items.iter().map(canonical_json).collect())
-        }
+        serde_json::Value::Array(items) => items
+            .iter()
+            .map(|item| canonical_json_at_depth(item, depth + 1))
+            .collect::<Result<Vec<_>, _>>()
+            .map(serde_json::Value::Array),
         serde_json::Value::Object(map) => {
             let mut entries = map.iter().collect::<Vec<_>>();
             entries.sort_by_key(|(key, _)| *key);
             let mut canonical = serde_json::Map::new();
             for (key, value) in entries {
-                canonical.insert(key.clone(), canonical_json(value));
+                canonical.insert(key.clone(), canonical_json_at_depth(value, depth + 1)?);
             }
-            serde_json::Value::Object(canonical)
+            Ok(serde_json::Value::Object(canonical))
         }
-        _ => value.clone(),
+        _ => Ok(value.clone()),
     }
 }
 
--- a/crates/ironclaw_host_api/tests/host_api_contract.rs
+++ b/crates/ironclaw_host_api/tests/host_api_contract.rs
@@ -396,6 +396,34 @@ fn invocation_fingerprint_is_stable_and_input_hashed() {
 }
 
 #[test]
+fn invocation_fingerprint_rejects_deeply_nested_input() {
+    let ctx = sample_context();
+    let capability = CapabilityId::new("echo.say").unwrap();
+    let estimate = ResourceEstimate::default();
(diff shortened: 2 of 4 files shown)


ilblackdragon (approved):
## Overview
Two small, focused fail-closed fixes:
1. `canonical_json` (used to build `InvocationFingerprint`) gets a recursion-depth cap so an attacker-controlled deeply-nested invocation input can't blow the stack.
2. `InMemoryResourceGovernor::reserve_with_id` rejects negative-USD `ResourceEstimate` values *before* mutating reservation state, so a negative reservation can't end up crediting active reserved balances.

Both are the right pattern (validate-then-mutate, cap-then-recurse) and ship with focused regression tests.

## Code quality / correctness

**`canonical_json` depth cap — `crates/ironclaw_host_api/src/approval.rs`**
- Refactor into `canonical_json_at_depth(value, depth)` is clean; `Result` propagation through `Array` / `Object` arms is idiomatic.
- **Off-by-one / naming nit**: `MAX_CANONICAL_JSON_DEPTH = 64` with `if depth > MAX_…` accepts `depth ∈ 0..=64`, i.e. 65 levels. Either rename to make the bound semantics explicit, switch to `depth >= MAX_…`, or document the intent. Minor — but the constant name reads as the bound rather than bound+1.
- **Error semantics question**: this is attacker-driven *input rejection*, not a programmer-bug *invariant violation*. Mapping onto `HostApiError::invariant(...)` may be the right call if that's how this codebase signals fail-closed input rejection elsewhere; otherwise a `Validation`-style variant would read more accurately. Worth a quick sanity check against the rest of the host API conventions.
- **Defense-in-depth context worth a one-line comment**: `serde_json` already enforces a 128-deep recursion limit at parse time, so a 10k-deep `Value` only arises via programmatic construction. Saying so near the constant makes the threat model legible to future readers.

**Test (`invocation_fingerprint_rejects_deeply_nested_input`)**
- The `Box::leak(Box::new(input))` workaround for `serde_json::Value`'s recursive `Drop` is clever and correctly commented. Two thoughts:
  - The shadowing pattern (`let input = ...; let input = Box::leak(...)`) is harder to skim than `let input = Box::leak(Box::new(input));` directly.
  - You don't need 10k depth to prove rejection — anything `> 64` (say, 200) demonstrates the cap and avoids both the leak *and* the cost of building a 10k-deep `Value`. If 10k is intended to demonstrate "would have stack-overflowed without the cap", say so in the comment; otherwise a smaller depth is the cheaper test.

**`reserve_with_id` validation — `crates/ironclaw_resources/src/lib.rs`**
- `validate_estimate` is correctly invoked *before* `lock_state()` / state mutation, so failure leaves `reserved_by_account` untouched. Test asserts this.
- `reserve` already delegates to `reserve_with_id`, so both API entry points inherit the validation — good. Worth a one-liner in the PR description so reviewers don't have to chase the call graph.
- USD is the only `ResourceEstimate` dimension that *can* be negative (the rest are `Option<u64>` / `Option<u32>`), so a USD-only check is sufficient *for `ResourceEstimate`*. The error variant is generic over `ResourceDimension`, which is the right level of extensibility.
- `&'static str` for `reason` — nice; no allocation.

## Risks / follow-ups (not blocking)

- **`reconcile` is the symmetric gap.** `reconcile(reservation_id, actual: ResourceUsage)` takes `usd: Decimal` (not `Option`) and inside the loop does `reserved.sub_assign(&record.tally)` then `usage.add_assign(&from_usage(&actual))`. A negative `actual.usd` would still inflate the reserved-balance accounting in the same fail-open way the reservation fix is closing. If "close foundation fail-closed gaps" is meant comprehensively, `reconcile` (and arguably `release` paths that touch tallies) deserves the same `validate_*`-before-mutate guard. Reasonable to defer given the PR's stated scope, but worth tracking.
- **Other governors**: only `InMemoryResourceGovernor` is touched. If/when a persistent backend lands, the validation needs to live in a place that all impls share — e.g. a default method on the trait or a free `validate_estimate` called from each impl. No action today; flag for the trait when a second impl arrives.

## Style / conventions
- Uses `let Some(usd) = estimate.usd && usd < Decimal::ZERO` (let-chains). Assuming MSRV supports it (consistent with usage elsewhere).
- Naming, comment density, and error wiring all match the surrounding crate style.
- No `unwrap`/`expect` in production paths.

## Test coverage
- Both fixes have direct regression tests at the contract-test tier — appropriate level.
- `reserved_for(&account).usd == 0` after the negative-estimate rejection is the load-bearing assertion; you could also assert that `state.reservations` gained no entry, but it's redundant given the early return precedes any insert.

## Security
- Both changes are net-positive for the security posture (anti-DoS via stack overflow, anti-confused-deputy via negative reservation). No new attack surface introduced.
- Error messages don't leak inputs (good — the depth-exceeded error is a static label, not the offending JSON).

## Verdict
Approving. The off-by-one in the constant naming and the symmetric `reconcile` question are the only items worth acknowledging — neither blocks shipping the foundation fix.

Issue #2095, opened 2026-04-07
Allow image model config


hanakannzashi:
This [PR](#1819) fix the image generation for NEAR AI model with hard coded fallback model name, we need to allow manually configure image model for other models

PR #3004, opened 2026-04-28
feat(image-config): add dedicated image tool configuration

hanakannzashi:
## Summary

- Added dedicated image-tool configuration for OpenAI-compatible image generation, editing, and image analysis endpoints.
- Split image tool configuration from chat LLM routing so explicit image endpoints use their own key and do not inherit chat credentials.
- Reworked image tool registration to use resolved `ImageConfig` instead of hardcoded model fallbacks.
- Updated docs/examples and added targeted tests for image config resolution and `/v1` endpoint handling.

## Change Type

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [x] Documentation
- [ ] CI/Infrastructure
- [x] Security
- [ ] Dependencies

## Linked Issue

#2095 

## Validation

- [x] `cargo build`
- [x] Relevant tests pass:
  - `cargo test --lib config::image::tests`
  - `cargo test --lib config::helpers::tests::strip_admin_only_llm_keys_removes_all_known_keys`
  - `cargo test --lib tools::builtin::image_analyze::tests::endpoint_url_does_not_duplicate_v1`
- [x] Manual testing: started Gateway with full image env, verified port binding, and tested `IMAGE_VISION_MODEL=openai/gpt-5.2`

Additional targeted checks run:
- `cargo fmt`
- `git diff --check`

## Security Impact

This changes secret and outbound network handling for image tools. Explicit `IMAGE_BASE_URL` no longer inherits chat-provider credentials; operators must provide `IMAGE_API_KEY` for a distinct image endpoint. Image DB-backed settings are treated as admin-only to prevent non-admin users from injecting private or loopback provider endpoints.

## Database Impact

No schema or migration changes. DB-backed settings behavior is affected only through admin-only filtering for image config keys.

## Blast Radius

Touches config resolution, secret/env hydration, AppBuilder tool registration, image generation/edit/analyze tools, and setup documentation. Potential breakage is limited to built-in image tools registration and image/vision endpoint selection.

## Rollback Plan

Revert the image config module and AppBuilder registration changes to restore the previous behavior where image tools derived endpoint/key/model directly from the active LLM config and used the hardcoded generation fallback.

## Review Follow-Through

Reviewer judgment requested on the final public configuration surface, especially whether `IMAGE_VISION_MODEL` should remain exposed for `image_analyze` or be treated as an internal/advanced setting.

---

**Review track**: C


zmanian:
## Code Review

### Overview
+573/-79 across 10 files, including a new 358-line `src/config/image.rs`. Splits image/vision tool config (`IMAGE_BASE_URL`, `IMAGE_API_KEY`, `IMAGE_MODEL`, `IMAGE_VISION_MODEL`) from chat LLM config so explicit operator-supplied image endpoints don't inherit chat bearer tokens; admin-only DB filtering for those keys.

### Strengths
- Strong security framing: prevents bearer-token forwarding to operator-supplied image endpoints; admin-only filter for DB-backed image config keys.
- Uses `SecretString` for the API key, `validate_operator_base_url` for base URL hygiene.
- Clean separation via dedicated `ImageConfig::resolve` with explicit `InheritedEndpoint` helper.
- Module-owned init pattern respected (config lives in `src/config/`, AppBuilder just consumes it).
- Targeted unit tests for resolution logic and `/v1` endpoint dedup.
- `.env.example` and `configuration.mdx` updated in sync.

### Issue blocking merge (Major)

**Silent disable when `IMAGE_BASE_URL` is set without `IMAGE_API_KEY`** (gemini-bot also flagged). `api_key` becomes `None` and image tools simply don't register. This is a misconfiguration the operator almost certainly wants to know about. Per the project's preflight-fail-loud pattern, this should return `ConfigError::MalformedConfig` (or similar) at preflight rather than failing open silently.

### Minor

- Verify `image_analyze.rs` endpoint construction handles trailing slash + missing `/v1` symmetrically; would benefit from a fuzz-style table test alongside `endpoint_url_does_not_duplicate_v1`.
- Confirm no `info!` was added in REPL/TUI-visible paths — image-tool registration log should be `debug!`.
- `ImageConfig` carries `Option<SecretString>`; ensure `Debug` derive doesn't leak the key (SecretString redacts, but worth a test pinning that).
- The `silent-ok` rule: if any `.unwrap_or_default()` on `Result` was added in `helpers.rs` extensions, needs the `// silent-ok:` comment.

### Risk
Low-to-medium. Blast radius is limited to image tool registration. Security posture improves (no chat-key leak). Main behavioral risk is silent tool absence on partial config — addressed by the major issue above.

### Recommendation
**Request Changes (small)**: convert the `IMAGE_BASE_URL`-without-`IMAGE_API_KEY` case to a hard `ConfigError` with a clear message, then approve. Otherwise this is a well-scoped, security-positive change.

serrrfirat:
Parallel review finding:

1. **Low — Docs** — `docs/drafts/setup/configuration.mdx:7`

   New image env var docs were added to `docs/drafts/setup/configuration.mdx`, but that page is marked superseded and says “Do not edit this file.” The promoted `docs/capabilities/configuration.mdx` in this PR head has no `IMAGE_*` entries, so published config docs will miss the new dedicated image tool settings.

   Suggested fix: move/sync `IMAGE_TOOLS_ENABLED`, `IMAGE_BASE_URL`, `IMAGE_API_KEY`, `IMAGE_MODEL`, and `IMAGE_VISION_MODEL` docs into `docs/capabilities/configuration.mdx` (and leave/remove the draft change per docs convention).

henrypark133 (changes requested):

What looks good:
- Dedicated image config resolution is cleanly separated from the conversational LLM config.
- The prior `IMAGE_BASE_URL` without `IMAGE_API_KEY` concern appears fixed by the explicit `ConfigError::MissingRequired { key: "IMAGE_API_KEY" }` path.

Finding:

1. High - `src/channels/web/features/settings/mod.rs:252`

   Image settings are persisted but excluded from the hot-reload path. `llm_setting_requires_reload()` only watches LLM keys, and `reload_llm_after_settings_change()` only rebuilds the LLM provider chain, so saving `IMAGE_BASE_URL`, `IMAGE_API_KEY`, `IMAGE_MODEL`, or `IMAGE_VISION_MODEL` through the settings API will not update the live image tool config/registry until process restart.

   That makes the new dedicated image config stale at runtime after a successful settings save. Please either implement an image-specific reload that re-resolves image config and refreshes image tool registration, or make the setting explicitly restart-only in the API/UI/docs so users are not told a saved value is live.

PR #3006, opened 2026-04-28
fix(mcp): retry startup MCP activation after auth failures

hanakannzashi:
## Summary

- Add non-blocking startup retry for MCP servers that temporarily reject a preconfigured `Authorization` header.
- Covers the private-chat key binding race where `/mcp` may return auth errors during container startup, leaving NEAR AI MCP inactive until manual activation.
- Keep normal OAuth/missing-auth behavior unchanged; retry only applies to auth-like failures on MCP servers with custom `Authorization` headers.
- Re-read MCP config before each retry and stop if the server is removed or disabled.

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] CI/Infrastructure
- [ ] Security
- [ ] Dependencies

## Linked Issue

None.

## Validation

- [x] Relevant tests pass: `cargo test --lib startup_mcp_auth_retry_only_covers_custom_authorization_rejections`
- [x] Manual testing: verified compile/check path locally
- [x] `cargo build`

Additional checks run:
- `cargo fmt --check`
- `cargo build --bin ironclaw`

## Security Impact

Low. This touches MCP startup auth handling and outbound MCP requests, but does not add new credentials or change secret storage. Retry only uses the already-configured MCP `Authorization` header and stops if the MCP server is removed or disabled.

## Database Impact

None. No schema or migration changes.

## Blast Radius

Touches app startup MCP loading and ExtensionManager MCP config lookup. Main risk area is MCP auto-activation during gateway startup, especially servers with custom authorization headers.

## Rollback Plan

Revert the startup MCP auth retry changes. Affected MCP servers would return to the previous behavior: failed startup activation leaves them inactive until manual activation.

## Review Follow-Through

This is intentionally scoped to IronClaw startup retry rather than changing private-chat provisioning/bind order. Reviewer judgment requested on retry duration and whether the retry condition should remain limited to custom `Authorization` header auth failures.

---

**Review track**: C


zmanian:
## Code Review

### Overview
+313/-33 adding a non-blocking background retry loop (24 attempts × 5s = 2min) for MCP servers that reject preconfigured `Authorization` headers at startup, plus centralizing hyphen→underscore normalization via `normalize_server_name`.

### Strengths
- Clean refactor from `Option<...>` tuple to a typed `StartupMcpLoadResult` enum.
- Retries now live in a `JoinSet` on `AppComponents` with proper `shutdown().await` wired into `main.rs`, and `Drop` aborts as a safety net (addresses the prior gemini round).
- `should_retry_startup_mcp_auth_error` is narrowly gated on `has_custom_auth_header()` AND `is_auth_error_message`, avoiding OAuth/missing-auth false positives.
- Retry loop re-reads server config each attempt and bails on disabled/removed/no-longer-custom-auth — good lifecycle hygiene.
- `normalize_server_name` consolidation removes three duplicated `replace('-', "_")` sites.

### Issue blocking merge (Major)

**The predicate is likely a no-op in production** (Copilot caught this; it's correct). `McpClient::send_request` rewrites 401/400 errors on custom-auth servers into:

```
"MCP server '<name>' rejected its configured Authorization header…"
```

That string contains **none** of the `auth`/`unauthorized`/`authentication`/`forbidden` substrings that `is_auth_error_message` checks for. So the predicate returns **false** for the exact scenario this PR targets, and retries never schedule.

The unit test passes only because it feeds raw `"401 Unauthorized"` strings rather than the rewritten one — this is a textbook "test through the caller, not just the helper" violation per `.claude/rules/testing.md`. Fix:

- Either add the rewritten phrase to `is_auth_error_message`, **or** match it explicitly in `should_retry_startup_mcp_auth_error`.
- Add a regression test using the actual production error string from `client.rs`.

### Minor

- **`tracing::info!` on retry success in the background task** corrupts the REPL/TUI per `CLAUDE.md` ("background tasks must NEVER use `info!`"). Demote to `debug!`.
- `MCP_STARTUP_AUTH_RETRY_ATTEMPTS=24` and 5s delay are hardcoded — author flagged this for reviewer judgment. Acceptable but a brief constant-level comment justifying 2min total would help.
- Validation checklist boxes for `clippy --all-features -D warnings` and `cargo test --features integration` are unchecked; please run before merge.

### Risk
Low blast radius — startup-only path, narrow predicate, bounded retries, clean shutdown. But the Major issue above means the fix may be a silent no-op in production, ironic for a bug-fix PR.

### Recommendation
**Request Changes.** Address the predicate-mismatch finding with a test against the rewritten error, demote the success log to `debug!`, and run the unchecked clippy/integration gates before merge.

serrrfirat:
Parallel review finding:

1. **Medium — Correctness / retry coverage** — `src/app.rs:1055`

   Startup retry only schedules when `should_retry_startup_mcp_auth_error` calls `is_auth_error_message(err_str)`. But `McpClient::send_request` rewrites custom-header auth failures to `"MCP server '...' rejected its configured Authorization header..."` at `src/tools/mcp/client.rs:627`, which does not match the classifier (`Authorization` without `401`/`unauthorized`/`authentication`/`400`). If `initialize` succeeds but `tools/list` rejects the custom Authorization header, startup logs the failure but does not schedule the intended background retry.

   Suggested fix: make the classifier match this custom-header rejection message or preserve a typed auth error through the startup load path. Add a caller-level startup activation test, not just a helper classifier test.

henrypark133 (changes requested):

What looks good:
- The retry machinery is scoped to startup MCP activation, and the normal CI checks are green.

Finding:

1. High - `src/app.rs:178` and `src/tools/mcp/client.rs:627`

   The custom-header auth retry path can still be skipped. `should_retry_startup_mcp_auth_error()` only retries when `is_auth_error_message(error_message)` matches, but the MCP client rewrites custom-header auth failures to `MCP server '<name>' rejected its configured Authorization header. Update the configured credential and try again.` That rewritten text does not contain the current auth-error tokens, so startup activation can return `Failed` instead of `RetryAuth` for the scenario this PR is trying to recover.

   Please make the retry predicate recognize the rewritten custom-header auth message, and add a caller-level regression test for that exact message shape, not only raw 401/Unauthorized strings.

PR #3007, opened 2026-04-28
Disable v2 CodeAct for Abound demo testing

henrypark133:
## Summary

Temporarily forces the engine v2 path onto structured tool calls for the Abound demo branch.

- stop converting fenced `python` / `repl` model output into `LlmResponse::Code`
- keep provider `tool_calls` as the only executable v2 action path
- replace the v2 system prompt text with structured-tool-only instructions for this demo test

## How this prevents CodeAct from running

The normal foreground v2 path reaches the model through `LlmBridgeAdapter`.

Before this PR, if the provider returned text containing a fenced `python`, `py`, `repl`, or Python-looking bare code block, the adapter converted that text into `LlmResponse::Code`. The Rust host then serialized that as `{ "type": "code" }` for the Python orchestrator, and the orchestrator's `resp_type == "code"` branch called `__execute_code_step__`, which enters nested Monty/CodeAct execution.

After this PR:

- provider `tool_calls` still become `LlmResponse::ActionCalls` and run through the structured tool loop
- provider text responses always stay `LlmResponse::Text`, even when they contain fenced Python/repl blocks
- because the normal bridge no longer creates `LlmResponse::Code`, it no longer serializes `{ "type": "code" }` for the orchestrator
- without `{ "type": "code" }`, the orchestrator's `__execute_code_step__` / nested Monty CodeAct branch is not reached on the normal app/provider path

In short: this does not delete the CodeAct executor; it cuts off the foreground v2 bridge path that used to route model text into that executor.

## Prompt leak audit

The foreground v2 system prompt built by `build_codeact_system_prompt_inner` no longer includes `codeact_preamble.md` or `codeact_postamble.md`. For this demo branch it uses a structured-tool-only prompt that explicitly says to use provider `tool_calls`, not Python/repl code or Python function-style tool calls.

Remaining CodeAct references are still present in the repo, but they are not all model-visible in the Abound foreground path:

- Rust doc comments, function names, test names, and orchestrator comments still mention CodeAct; these are not prompt text sent to the foreground model.
- The compiled `crates/ironclaw_engine/prompts/codeact_*.md` files still exist, but this PR stops the foreground prompt builder from including them.
- Learning/mission prompt files can still mention `CodeAct`, `FINAL`, or related concepts. Those are separate mission prompt surfaces, not the normal foreground Abound demo prompt.

Known caveat: runtime shared-memory prompt overlays titled `prompt:codeact_preamble` can still be appended under `## Learned Rules (from self-improvement)` if such a shared memory doc exists. A stored overlay could therefore still mention CodeAct in prompt text, even though the bridge-level conversion still prevents that text from entering nested Monty execution through the normal provider path.

## Residual risk

This is intentionally a branch-local demo/testing hack, not a polished long-term feature flag or hard engine-wide CodeAct removal.

- Tests or custom `LlmBackend` implementations can still directly return `LlmResponse::Code`.
- The Python orchestrator still contains its `resp_type == "code"` branch.
- A hard kill switch would also reject or downgrade `LlmResponse::Code` at the orchestrator boundary, but this PR intentionally avoids that broader change.

## Validation

- `cargo test -p ironclaw_engine executor::prompt::tests`
- `cargo test -p ironclaw --lib bridge::llm_adapter::tests`

The adapter tests specifically assert that fenced Python/repl responses remain text in both the no-tools path and the with-tools/no-tool-call path.

### DIFF
--- a/src/bridge/llm_adapter.rs
+++ b/src/bridge/llm_adapter.rs
@@ -78,17 +78,8 @@ impl LlmBackend for LlmBridgeAdapter {
                     reason: e.to_string(),
                 })?;
 
-            // Check for code blocks in the response (CodeAct/RLM pattern)
-            let llm_response = match extract_code_block(&response.content) {
-                Some(code) => LlmResponse::Code {
-                    code,
-                    content: Some(response.content),
-                },
-                None => LlmResponse::Text(response.content),
-            };
-
             return Ok(LlmOutput {
-                response: llm_response,
+                response: LlmResponse::Text(response.content),
                 usage: TokenUsage {
                     input_tokens: u64::from(response.input_tokens),
                     output_tokens: u64::from(response.output_tokens),
@@ -115,7 +106,9 @@ impl LlmBackend for LlmBridgeAdapter {
                     reason: e.to_string(),
                 })?;
 
-        // Convert response — check for code blocks (CodeAct/RLM pattern)
+        // Convert response. Temporary demo/abound hack: keep v2 on the
+        // structured tool loop only, so fenced Python/repl text is returned as
+        // plain text instead of entering nested Monty/CodeAct.
         let llm_response = if !response.tool_calls.is_empty() {
             LlmResponse::ActionCalls {
                 calls: response
@@ -131,14 +124,7 @@ impl LlmBackend for LlmBridgeAdapter {
             }
         } else {
             let text = response.content.unwrap_or_default();
-            // Detect ```repl or ```python fenced code blocks
-            match extract_code_block(&text) {
-                Some(code) => LlmResponse::Code {
-                    code,
-                    content: Some(text),
-                },
-                None => LlmResponse::Text(text),
-            }
+            LlmResponse::Text(text)
         };
 
         Ok(LlmOutput {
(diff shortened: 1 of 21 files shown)


src/bridge/llm_adapter.rs
```diff
@@ -131,14 +124,7 @@ impl LlmBackend for LlmBridgeAdapter {
...
             let text = response.content.unwrap_or_default();
-            // Detect ```repl or ```python fenced code blocks
-            match extract_code_block(&text) {
-                Some(code) => LlmResponse::Code {
-                    code,
-                    content: Some(text),
-                },
-                None => LlmResponse::Text(text),
-            }
+            LlmResponse::Text(text)
```
serrrfirat, line 127: **Medium Severity** — Fenced `FINAL(...)` text can still terminate the v2 loop without executing the intended tool action.

This bridge now correctly keeps fenced Python/repl output as `LlmResponse::Text` instead of converting it to `LlmResponse::Code`, which blocks the nested CodeAct executor path. However, the v2 orchestrator text branch still scans *plain text* for `FINAL(...)`. That means a provider response like:

````markdown
```python
abound_send_wire(...)
FINAL("Notification sent")
```
````

would no longer execute the Python or call the structured tool, but it would still be returned as text here and then the orchestrator can treat the fenced `FINAL(...)` as a completed final answer. For the Abound demo, that can produce a false-success response for an action that never ran.

Suggested fix: add an orchestrator-level regression test for fenced Python containing `FINAL(...)`, then either ignore `FINAL(...)` inside fenced code in the text branch (for example by applying the existing code-block stripper before `extract_final`) or disable text-level `FINAL(...)` parsing for this structured-tool-only mode.

PR #3015, opened 2026-04-28
feat(reborn): add extension manifest registry

serrrfirat:
## Summary

Carves the deferred extension manifest/registry slice from the Reborn stack.

This is the smallest useful next PR after the currently open Reborn substrate/control PRs because `ironclaw_extensions` depends only on:

```text
ironclaw_host_api
ironclaw_filesystem
```

and unblocks later dispatcher/runtime/capability-host slices without pulling those crates forward.

Adds `crates/ironclaw_extensions` with:

- extension manifest parsing and validation
- WASM / Script / MCP runtime metadata contracts
- manifest package validation and asset path containment
- extension registry lookup by extension/capability
- filesystem-backed discovery from `/system/extensions`
- explicit rejection of user-manifest self-assertion for `FirstParty` / `System` trust/runtime classes via host-api serde rules

## Stacking note

This PR is draft/stacked on #2999 because it is meant to land **after** the currently open Reborn PRs:

- #2996 filesystem substrate
- #2999 auth/control substrate

After those merge into `reborn-integration`, rebase this branch so the diff narrows to only:

```text
crates/ironclaw_extensions
Cargo.toml / Cargo.lock membership for that crate
```

## Scope boundary

This PR intentionally does **not** include:

- extension install/authenticate/configure/activate/remove lifecycle services
- trust-class policy engine/admin assignment beyond rejecting manifest self-assertion
- concrete runtime execution
- dispatcher/capability-host wiring
- production app wiring or user-visible behavior changes

## Exposure checklist

```text
Does this affect existing src/ runtime behavior? no
Does this alter app startup/config defaults? no
Does this expose new routes/CLI/user-visible APIs? no
Does this create a second writer for existing production state? no
Are all Reborn paths feature-gated or unused by default? yes, unused internal crate substrate
Are docs/status labels accurate: implemented slice vs product complete? yes, manifest/registry substrate only
```

## TDD note

Copied the extension contract tests first against a RED stub and confirmed `cargo test -p ironclaw_extensions` failed due missing manifest/registry types before porting the implementation.

## Verification

Passed:

```bash
cargo test -p ironclaw_extensions
cargo clippy -p ironclaw_extensions --all-targets -- -D warnings
cargo fmt --check
git diff --check
cargo tree -p ironclaw_extensions -e normal | grep -E 'ironclaw_(authorization|approvals|capabilities|dispatcher|events|host_runtime|mcp|network|processes|resources|run_state|scripts|secrets|wasm)' || true
```

Boundary grep returned no forbidden normal Reborn dependencies.

Refs #2987.

### DIFF
--- a/crates/ironclaw_extensions/src/lib.rs
+++ b/crates/ironclaw_extensions/src/lib.rs
@@ -1,0 +1,561 @@
+//! Extension manifest and registry contracts for IronClaw Reborn.
+//!
+//! `ironclaw_extensions` discovers and validates extension packages, extracts
+//! capability descriptors, and records declarative runtime metadata. It does not
+//! execute WASM modules, start Docker containers, connect to MCP servers, resolve
+//! secrets, or reserve resources.
+
+use std::collections::{HashMap, HashSet};
+
+use ironclaw_filesystem::{FilesystemError, RootFilesystem};
+use ironclaw_host_api::{
+    CapabilityDescriptor, CapabilityId, EffectKind, ExtensionId, HostApiError, PermissionMode,
+    ResourceProfile, RuntimeKind, TrustClass, VirtualPath,
+};
+use serde::Deserialize;
+use thiserror::Error;
+
+/// Extension manifest and registry failures.
+#[derive(Debug, Error)]
+pub enum ExtensionError {
+    #[error(transparent)]
+    Contract(#[from] HostApiError),
+    #[error("failed to parse extension manifest: {reason}")]
+    ManifestParse { reason: String },
+    #[error("invalid extension manifest: {reason}")]
+    InvalidManifest { reason: String },
+    #[error("invalid extension asset path '{path}': {reason}")]
+    InvalidAssetPath { path: String, reason: String },
+    #[error("extension manifest id mismatch at {root:?}: expected {expected}, actual {actual}")]
+    ManifestIdMismatch {
+        root: VirtualPath,
+        expected: ExtensionId,
+        actual: ExtensionId,
+    },
+    #[error("duplicate extension id {id}")]
+    DuplicateExtension { id: ExtensionId },
+    #[error("duplicate capability id {id}")]
+    DuplicateCapability { id: CapabilityId },
+    #[error(transparent)]
+    Filesystem(#[from] FilesystemError),
+}
+
+/// Manifest-local path for assets such as WASM modules.
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct ExtensionAssetPath(String);
+
+impl ExtensionAssetPath {
+    pub fn new(value: impl Into<String>) -> Result<Self, ExtensionError> {
+        let value = value.into();
+        validate_asset_path(&value)?;
+        Ok(Self(value))
+    }
+
+    pub fn as_str(&self) -> &str {
+        &self.0
+    }
+
+    pub fn resolve_under(&self, root: &VirtualPath) -> Result<VirtualPath, ExtensionError> {
+        VirtualPath::new(format!(
+            "{}/{}",
+            root.as_str().trim_end_matches('/'),
+            self.0
+        ))
+        .map_err(ExtensionError::from)
+    }
+}
+
+/// Declarative runtime metadata for an extension package.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum ExtensionRuntime {
+    Wasm {
+        module: ExtensionAssetPath,
+    },
+    Script {
+        runner: String,
+        image: Option<String>,
+        command: String,
+        args: Vec<String>,
+    },
+    Mcp {
+        transport: String,
+        command: Option<String>,
+        args: Vec<String>,
+        url: Option<String>,
+    },
+    FirstParty {
+        service: String,
+    },
+    System {
+        service: String,
+    },
+}
+
+impl ExtensionRuntime {
+    pub fn kind(&self) -> RuntimeKind {
+        match self {
+            Self::Wasm { .. } => RuntimeKind::Wasm,
+            Self::Script { .. } => RuntimeKind::Script,
+            Self::Mcp { .. } => RuntimeKind::Mcp,
+            Self::FirstParty { .. } => RuntimeKind::FirstParty,
+            Self::System { .. } => RuntimeKind::System,
+        }
+    }
+}
+
+/// Validated extension manifest.
+#[derive(Debug, Clone, PartialEq)]
+pub struct ExtensionManifest {
+    pub id: ExtensionId,
+    pub name: String,
+    pub version: String,
+    pub description: String,
+    pub trust: TrustClass,
+    pub runtime: ExtensionRuntime,
+    pub capabilities: Vec<CapabilityManifest>,
+}
+
+impl ExtensionManifest {
+    pub fn parse(input: &str) -> Result<Self, ExtensionError> {
+        let raw: RawManifest =
+            toml::from_str(input).map_err(|error| ExtensionError::ManifestParse {
+                reason: error.to_string(),
+            })?;
+        Self::from_raw(raw)
+    }
+
+    pub fn runtime_kind(&self) -> RuntimeKind {
+        self.runtime.kind()
+    }
+
+    fn from_raw(raw: RawManifest) -> Result<Self, ExtensionError> {
+        if raw.name.trim().is_empty() {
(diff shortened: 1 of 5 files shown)


crates/ironclaw_extensions/src/lib.rs
```diff
@@ -0,0 +1,561 @@
...
+        args: Vec<String>,
+    },
+    Mcp {
+        transport: String,
+        command: Option<String>,
+        #[serde(default)]
+        args: Vec<String>,
+        url: Option<String>,
+    },
+    FirstParty {
```
serrrfirat, line 425: **High Severity** — privileged runtime kinds can still be self-asserted by manifests.

`TrustClass` deserialization correctly rejects `trust = "first_party"` / `"system"`, but this private `RawRuntime` enum reintroduces `FirstParty` and `System` as deserializable manifest variants. A manifest with `trust = "sandbox"` and `[runtime] kind = "system"` currently parses successfully and produces a descriptor with `RuntimeKind::System`.

That violates the host-assigned-only runtime invariant and means future dispatcher/runtime wiring could route a user-supplied extension into a privileged execution lane. The existing test misses this because it sets both privileged trust and privileged runtime, so it fails on the trust field before exercising runtime parsing.

Please reject `RawRuntime::FirstParty` / `RawRuntime::System` in `into_runtime()` (or remove them from the manifest-facing enum and keep a separate trusted host-only constructor), and add tests where `trust = "sandbox"` is paired with `kind = "first_party"` / `"system"`.

crates/ironclaw_extensions/src/lib.rs
```diff
@@ -0,0 +1,561 @@
...
+                url,
+            } => {
+                validate_non_empty("mcp transport", &transport)?;
+                if let Some(command) = &command {
+                    validate_non_empty("mcp command", command)?;
+                }
+                if let Some(url) = &url {
+                    validate_non_empty("mcp url", url)?;
+                }
+                Ok(ExtensionRuntime::Mcp {
```
serrrfirat, line 480: **Medium Severity** — MCP runtime manifests can omit their endpoint entirely.

This branch only checks that `transport` is non-empty and that optional `command` / `url` values are non-empty when present. It still accepts manifests such as:

```toml
[runtime]
kind = "mcp"
transport = "stdio"
```

That registers an MCP capability descriptor with no command, URL, or other usable endpoint, pushing a deterministic manifest validation error into a later runtime failure. It also allows ambiguous shapes such as `stdio` with a URL or both `command` and `url`.

Please make the transport contract structural: for example `stdio` requires exactly `command` (+ optional args) and no URL; HTTP/SSE transports require exactly `url` and no command. Add negative tests for missing endpoint, both endpoint forms, and transport/endpoint mismatch.

crates/ironclaw_extensions/src/lib.rs
```diff
@@ -0,0 +1,561 @@
...
+    if value.trim().is_empty() {
+        Err(ExtensionError::InvalidManifest {
+            reason: format!("{kind} must not be empty"),
+        })
+    } else {
+        Ok(())
+    }
+}
+
+fn looks_like_windows_path(value: &str) -> bool {
```
serrrfirat, line 662: **Medium Severity** — Windows drive-relative asset paths pass containment validation.

`looks_like_windows_path()` only rejects drive paths when the colon is followed by `/` or `\\`, so a manifest module like `C:evil.wasm` is accepted as an extension asset path. That is still a Windows drive-relative path, not a normal manifest-local POSIX segment.

Because asset paths are later resolved into virtual paths and filesystem backends eventually push path segments into `PathBuf`, allowing drive prefixes weakens the “asset path stays under the extension root” contract on Windows targets.

Please reject any drive-prefix pattern like `^[A-Za-z]:` (or simply reject `:` in asset path segments) and add a regression case for `C:evil.wasm` alongside the existing `C:/...` / backslash-style path cases.

crates/ironclaw_extensions/src/lib.rs
```diff
@@ -0,0 +1,561 @@
...
+        root: &VirtualPath,
+    ) -> Result<ExtensionRegistry, ExtensionError>
+    where
+        F: RootFilesystem,
+    {
+        let mut entries = fs.list_dir(root).await?;
+        entries.sort_by(|left, right| left.name.cmp(&right.name));
+
+        let mut registry = ExtensionRegistry::new();
+        for entry in entries {
```
serrrfirat, line 359: **Medium Severity** — a stray non-extension file aborts the whole discovery pass.

Discovery iterates every direct child returned by `fs.list_dir(root)` and immediately treats the name as an extension id plus directory containing `manifest.toml`. A single non-directory file under `/system/extensions` (for example `.DS_Store`, `README.md`, or a temporary editor file) makes discovery fail before any valid extension can be registered.

`DirEntry` already includes `file_type`, so this can fail more gracefully. Please either skip non-directory entries or return a targeted invalid-package error that does not prevent unrelated valid extension directories from being discovered. Add a test with a valid extension directory plus a stray file in the root.

serrrfirat:
Posted follow-up fixes in `8cd23958f9a4fb8926f46b7b6c04c06dcfd1e76a` after re-running the PR fix loop.

Changes added on top of the previous hardening commit:
- `ExtensionDiscovery::discover` now skips non-extension directories whose names are not valid extension IDs, so a stray tooling/cache directory such as `.cache` does not abort discovery for valid extensions.
- `ExtensionRegistry::insert` now detects duplicate capability IDs inside the package being inserted, preventing a malformed/publicly-constructed package from silently overwriting an earlier descriptor in the registry map.
- Added regression coverage for both cases.

Local verification passed:
- `cargo test -p ironclaw_extensions`
- `cargo clippy -p ironclaw_extensions --all-targets -- -D warnings`
- `cargo fmt --check`
- `git diff --check`
- dependency boundary grep: no forbidden normal Reborn dependencies

henrypark133 (changes requested):
## Review: tighten extension registry contracts

The new extension crate is well-scoped: it keeps manifest parsing and descriptor extraction separate from execution, secret resolution, MCP connection, and resource reservation. The negative-path contract tests are also strong around unsafe asset paths, duplicate IDs, privileged runtime self-assertion, and filesystem discovery.

What looks good:
- The implementation keeps the extension registry as declarative metadata only, which matches the Reborn ownership boundary.
- Manifest validation rejects host paths, URLs, dot segments, malformed MCP runtime shapes, and self-asserted first-party/system runtimes.
- The crate-level and architecture tests cover the intended boundary without pulling in runtime execution dependencies.

### Concerning: capability iteration is nondeterministic

`ExtensionRegistry::capabilities()` currently returns `self.capabilities.values()` from a `HashMap`. Discovery sorts extension directories and the registry preserves extension order, but capability iteration can still reorder across process runs. Once this registry feeds capability projection, approval displays, replay snapshots, or prompt/action lists, that nondeterminism can create noisy diffs and unstable runtime behavior. Please preserve capability order explicitly, either by iterating `extension_order` and each package's manifest-order capabilities or by maintaining a dedicated `capability_order` alongside the lookup map.

### Concerning: registry insertion accepts descriptors that did not pass package validation

`ExtensionPackage` is fully public, and host-side callers will need to construct packages directly for first-party/system runtimes because those cannot come from manifest parsing. In that path, `ExtensionRegistry::insert()` only checks duplicate IDs and provider. It will accept descriptors that were added or mutated after validation even if they no longer match `package.manifest`, for example different runtime/trust/default-permission metadata or entirely undeclared capabilities. The registry then becomes the source of truth for authority metadata that never passed the manifest/package contract. Please re-validate descriptor-to-package/manifest consistency in `insert()` before accepting the package.

serrrfirat:
Addressed review 4191808728 in `f5d9191f51cfaee9399c7722dbb1be62088f2f4d`.

Changes:
- `ExtensionRegistry::capabilities()` now iterates via explicit `capability_order`, preserving extension insertion order and manifest capability order instead of relying on `HashMap::values()`.
- `ExtensionRegistry::insert()` now re-derives the expected package from `package.manifest` + `package.root` and rejects public/mutated packages whose id or descriptors do not match manifest-derived metadata.
- Added regression tests for deterministic capability order and package/descriptor revalidation.

Verification:
- `cargo test -p ironclaw_extensions` (26 passed)
- `cargo clippy -p ironclaw_extensions --all-targets -- -D warnings`
- `cargo fmt --check`
- `git diff --check`
- dependency boundary grep: no forbidden normal Reborn dependencies

henrypark133 (approved):

PR #3017, opened 2026-04-28
feat(reborn): add process lifecycle substrate

serrrfirat:
## Summary

Carves the process lifecycle/result/output substrate from the Reborn stack.

Adds `crates/ironclaw_processes` with:

- process records, starts, statuses, exits, and subscriptions
- `ProcessStore`, `ProcessResultStore`, `ProcessManager`, `ProcessHost`, and `ProcessExecutor`
- in-memory and filesystem-backed process/result stores
- `BackgroundProcessManager`
- cooperative cancellation registry/tokens
- eventing process store wrapper for redacted process events
- resource-managed process store that reserves/reconciles/releases process-owned reservations
- output refs for large/binary process outputs

## Stacking note

This PR is draft/stacked because it is intended to land after the current Reborn substrate/control stack:

- #2996 filesystem substrate
- #2999 auth/control substrate
- #3015 extension manifest registry

After those merge into `reborn-integration`, rebase this branch so the diff narrows to only:

```text
crates/ironclaw_processes
Cargo.toml / Cargo.lock membership for that crate
```

## Scope boundary

This PR intentionally does **not** include:

- dispatcher runtime routing
- WASM/Script/MCP runtime lanes
- `ironclaw_capabilities` / `CapabilityHost`
- host runtime composition
- built-in obligation handling
- secrets/network substrates
- production app wiring or user-visible behavior changes

## Exposure checklist

```text
Does this affect existing src/ runtime behavior? no
Does this alter app startup/config defaults? no
Does this expose new routes/CLI/user-visible APIs? no
Does this create a second writer for existing production state? no
Are all Reborn paths feature-gated or unused by default? yes, unused internal crate substrate
Are docs/status labels accurate: implemented slice vs product complete? yes, process substrate only
```

## TDD note

Copied the process contract tests first against a RED stub and confirmed `cargo test -p ironclaw_processes` failed due missing process types before porting the implementation.

## Verification

Passed:

```bash
cargo test -p ironclaw_processes
cargo clippy -p ironclaw_processes --all-targets -- -D warnings
cargo fmt --check
git diff --check
cargo tree -p ironclaw_processes -e normal | grep -E 'ironclaw_(authorization|approvals|capabilities|dispatcher|extensions|host_runtime|mcp|network|run_state|scripts|secrets|wasm)' || true
```

Boundary grep returned no forbidden normal Reborn dependencies.

Refs #2987.


serrrfirat:
Addressed all 9 findings from the paranoid review in `fc96d6596`. 44 tests pass, clippy clean.

| ID | Where | Fix |
|----|-------|-----|
| **M1** | `services.rs` | Added `BackgroundFailure` / `BackgroundFailureStage` + `BackgroundProcessManager::with_error_handler(..)`. Errors from `store.complete/fail` and `result_store.complete/fail` inside the spawned task are now reported via the handler instead of silently dropped. |
| **M2** | `services.rs` | Inverted write order in `BackgroundProcessManager::spawn`: result store is persisted **before** lifecycle status flips to terminal. Contract is now "if status is terminal, result record is on disk" — `await_result` no longer needs the single-retry race mitigation as a primary correctness mechanism. Added a `store.get` pre-check so the inverted ordering does not overwrite a `Killed` result if `host.kill` already terminalized the process (caught by an existing test). |
| **M3** | `services.rs` | Removed the blanket `impl<T> ProcessManager for T where T: ProcessStore + ?Sized` so raw stores can no longer silently impersonate a `ProcessManager`. No callers depend on it. |
| **M4** | `services.rs` | Doc note on `BackgroundProcessManager::spawn` describing the detached-task orphan risk on runtime shutdown + `TODO` for built-in startup reconciliation of stuck `Running` records. |
| **L1** | `wrappers.rs` | New `ReservationDropGuard` RAII wrapper around `inner.start` in `ResourceManagedProcessStore::start`. If `inner.start` panics, the just-acquired reservation is released via `Drop` rather than leaked. |
| **L2** | `filesystem_store.rs` | Doc on `FilesystemProcessResultStore::complete` describing the two-step write (`write_output` → `write_result`) and the orphan-blob cleanup expectation if the second write fails. |
| **L3** | `filesystem_store.rs` | Doc on `FilesystemProcessStore::new` and `from_arc` calling out the single-instance invariant (transition lock only serializes within one instance; share via `Arc`, do not construct multiple instances against the same root). |
| **L4** | `tests/process_store_contract.rs` | Added `FailingProcessResultStore` + `background_process_manager_reports_result_store_complete_failure_and_keeps_running_status` — drives a result-store that returns `Err` and asserts (a) the error handler receives `ResultStoreComplete` for the right `process_id` and (b) the lifecycle status stays `Running` (M2 ordering: status does not promote when result write fails). |
| **N1** | `cancellation.rs` | Doc on `ProcessCancellationToken::cancelled` explaining the notify-then-check ordering: `notified()` is created **before** the flag re-check, so any `notify_waiters` racing the check is captured rather than lost. |

Diff: 6 files, +342 / −49.

PR #3033, opened 2026-04-28
  Fix TUI interrupt dispatch during active turns

### LINKED ISSUE #2142: feat(tui): add out-of-band interrupt path to bypass agent dispatch bottleneck
## Problem

The TUI correctly captures `Esc` keystrokes mid-turn via `InputAction::Interrupt` at `crates/ironclaw_tui/src/app.rs:456`, sending a `/interrupt` message through the channel. However, the agent dispatch layer in `src/agent/agent_loop.rs:941` awaits `handle_message` inline — which blocks on `await_thread_outcome` (up to 5 minutes via `router.rs:1981`). Until that returns, `message_stream.next()` is not polled, so the `/interrupt` message sits queued and only dequeues **after** the turn it was supposed to stop has ended.

The engine-v2 interrupt machinery itself works: `handle_interrupt` at `src/bridge/router.rs:1435` correctly calls `ThreadManager::stop_thread()`. It's just unreachable from user input while a turn is in flight.

## Suggested Fix (TUI-local, no agent refactor)

Add an out-of-band interrupt path that bypasses the message channel:

1. **Define a narrow `InterruptHandle` trait** in `ironclaw_tui` to keep the crate decoupled from the engine:
   ```rust
   #[async_trait]
   pub trait InterruptHandle: Send + Sync {
       async fn interrupt_active(&self, user_id: &str, channel: &str);
   }
   ```

2. **Pass an `Arc<dyn InterruptHandle>`** into `TuiAppConfig` / `start_tui`.

3. **In `handle_event`** where `InputAction::Interrupt` is matched (`app.rs:456`), call `handle.interrupt_active(...).await` **in addition to** sending the `/interrupt` message (keep the message as a fallback / for the post-turn "Nothing to interrupt." case).

4. **In `src/channels/tui.rs`**, implement `InterruptHandle` with a concrete type that calls into `src/bridge/router.rs::handle_interrupt` logic — i.e. look up the active engine conversation for `(channel, user_id)` and call `ThreadManager::stop_thread()` on each running thread. This mirrors what `handle_interrupt` already does, but runs directly from the TUI event loop.

No changes needed to `agent_loop.rs` or the v2 engine — the machinery is all there, just needs a direct call path.

## Why This Matters Beyond Esc

The same bottleneck affects any future "while processing" feature — cancel buttons, timeouts, status pokes, etc. Landing the out-of-band handle sets up the pattern for all of these.

This also affects the REPL channel (`src/channels/repl.rs:565`) where `rl.readline()` is parked behind `stdin_locked` during processing, making interrupt even harder.

## References

- Originated from PR review comment: https://github.com/nearai/ironclaw/pull/1973#issuecomment-4192963826
- Related PRs: #1973, #1412, #1557

reidliu41:
## Summary

  - Add a narrow TUI interrupt hook for stopping active engine work out-of-band.
  - Wire the TUI channel to the bridge interrupt path while preserving the existing `/interrupt` fallback.
  - Refactor active engine thread stopping into shared logic and add regression coverage for Esc interrupt behavior.
  

-

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] CI/Infrastructure
- [ ] Security
- [ ] Dependencies

## Linked Issue

  Fixes #2142

## Validation

- [x] `cargo build`
- [ ] Relevant tests pass: 
- [ ] Manual testing: 

## Review Follow-Through

---

**Review track**:

### DIFF
--- a/crates/ironclaw_tui/examples/dev.rs
+++ b/crates/ironclaw_tui/examples/dev.rs
@@ -150,9 +150,12 @@ fn mock_skill_categories() -> Vec<SkillCategory> {
 
 fn main() {
     let config = TuiAppConfig {
+        user_id: "dev-user".into(),
+        channel: "tui".into(),
         version: "0.22.0-dev".into(),
         model: "gpt-5.4".into(),
         layout: TuiLayout::default(),
+        interrupt_handle: None,
         context_window: 128_000,
         tools: mock_tool_categories(),
         skills: mock_skill_categories(),
--- a/crates/ironclaw_tui/src/app.rs
+++ b/crates/ironclaw_tui/src/app.rs
@@ -9,7 +9,10 @@
 //! The app owns the terminal, manages alternate screen / raw mode, and
 //! renders frames at ~30fps using a tick timer.
 
+use std::future::Future;
 use std::io::{self, Write};
+use std::pin::Pin;
+use std::sync::Arc;
 use std::time::Duration;
 
 use ratatui::Terminal;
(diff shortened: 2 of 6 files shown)


serrrfirat:

PR #3046, opened 2026-04-28
feat(telegram): add allowed_chat_ids for group-level access control

Kampouse:
## The problem

Currently `dm_policy` controls **both** DMs and group chats as a single knob:

- `dm_policy: "open"` → anyone can DM **and** @mention in **any** group
- `dm_policy: "pairing"` → restricted DMs **and** restricted groups

like i want my bot to answer stuff in my group but it also mean that my dm will be open  to random user
## Proposed fix

Add `allowed_chat_ids` to `TelegramConfig` — an independent group-level filter that is completely separate from `dm_policy`:

```json
{
  "config": {
    "dm_policy": "open",
    "respond_to_all_group_messages": false,
    "allowed_chat_ids": [-5263819573]
  }
}
```

- Empty or missing → allow all groups (no behavior change, fully backward compatible)
- `[-1234567]` → only process messages from those specific group chat IDs
- DMs are always unaffected regardless of this setting

## Should we be able to restrict access to specific groups without restricting DMs?

This is the design question. The current behavior treats DMs and groups as one thing. Should they be separate? Or is the intent that if you want group restrictions, you should also have DM restrictions?

Open to feedback on the approach — maybe `dm_policy` should be split into `dm_policy` + `group_policy` instead?


Kampouse:
Replaced by #3047 (clean branch, only the allowed_chat_ids commit)

Issue #79, opened 2026-02-14
feat: Channel feature gaps (per-group policies, self-message bypass, typing indicators)


ilblackdragon:
## Feature Parity: Channel Feature Gaps

**Priority:** P2
**Source:** \`FEATURE_PARITY.md\` — Channel Features

Cross-channel features that are missing or partial in IronClaw.

### Missing (❌)

- [ ] **Self-message bypass** — Agent's own messages sent to a channel should skip pairing verification. Currently the pairing store doesn't distinguish between the agent's outbound messages and inbound messages from unknown users.
- [ ] **Per-group tool policies** — Allow/deny specific tools per group/channel context. A Telegram group might allow \`memory_search\` but block \`shell\`, while a private DM allows all tools.

### Partial (🚧)

- [ ] **Allowlist/blocklist** — \`allow_from\` + pairing store exists, needs completion. Missing: blocklist (explicitly deny specific users), wildcard patterns, per-channel overrides.
- [ ] **Per-channel media limits** — Caption support exists; no size limits enforcement. Missing: max file size per channel, format restrictions, bandwidth throttling.
- [ ] **Typing indicators** — TUI shows status; not propagated to external channels. Channels should show typing indicator while agent is processing.

### Design Considerations

1. **Tool policy storage**: Per-group tool policies should be stored in the \`settings\` table with dotted keys like \`channel.telegram.group.-100123.tools.allow\` and \`channel.telegram.group.-100123.tools.deny\`. The agent loop checks these before presenting tools to the LLM.

2. **Self-message detection**: Each channel implementation knows its own bot identity (bot user ID, bot username). Add a \`fn is_self_message(&self, message: &IncomingMessage) -> bool\` method to the \`Channel\` trait. The pairing check in the agent loop skips messages where this returns true.

3. **Typing indicator protocol**: Add \`fn send_typing(&self, thread_id: &str) -> Result<(), ChannelError>\` to the \`Channel\` trait with a default no-op. Call it in the agent loop before LLM calls. Channels that support typing (Telegram, Discord, Slack) implement it.

4. **Media limits enforcement**: Add to channel config: \`max_media_size_bytes\`, \`allowed_media_types\`. Check in the channel wrapper before passing media to the agent. Reject oversized media with a user-facing error.

### Success Criteria

- [ ] Agent's own outbound messages don't trigger pairing verification on any channel
- [ ] Admin can configure per-group tool allow/deny lists via settings
[...]

PR #3047, opened 2026-04-28
feat(telegram): add allowed_chat_ids for group-level access control

Kampouse:
## Problem

I want my bot to respond in my group chat, but keep DMs restricted. Currently `dm_policy` controls both DMs and groups as a single knob — setting `dm_policy: "pairing"` locks down groups too, so only authorized users can @mention the bot even in my own group.

This is the same gap described in #79 under **Per-group policies** — there's no way to control access per group independently of DM pairing policy.

## Fix

Add `allowed_chat_ids` to `TelegramConfig` — a group-level filter independent of `dm_policy`:

```json
{
  "config": {
    "dm_policy": "pairing",
    "respond_to_all_group_messages": false,
    "allowed_chat_ids": [-5263819573]
  }
}
```

- Empty or missing → allow all groups (no behavior change)
- `[-1234567]` → only process messages from those group chat IDs
- Messages from allowed groups bypass `dm_policy` — anyone can @mention the bot in allowed groups regardless of pairing/allowlist
- DMs always controlled by `dm_policy` as before

50 lines, 1 file, fully backward compatible.

Part of #79

### DIFF
--- a/channels-src/telegram/src/lib.rs
+++ b/channels-src/telegram/src/lib.rs
@@ -286,6 +286,24 @@ const RESPOND_TO_ALL_GROUP_PATH: &str = "state/respond_to_all_group_messages";
 /// Workspace path for persisting allowed_chat_ids (JSON array of i64).
 const ALLOWED_CHAT_IDS_PATH: &str = "state/allowed_chat_ids";
 
+/// Result of checking whether a group chat is allowed.
+/// Returns `Some(true)` if the group is explicitly in the allowlist,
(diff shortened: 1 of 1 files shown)


serrrfirat:
Hi @Kampouse 👋

Thanks for both the Telegram access control and observe mode PRs — really thoughtful features and well-written descriptions. I'm closing both for now as we're cleaning up older PRs ahead of v1.

The Telegram channel has undergone significant changes since these were opened, so the code would need a fresh look. We'd love to have you contribute again after v1 when things are more settled. Thanks again! 🙏

PR #3065, opened 2026-04-29
fix(image-gen/edit): ENGINE_V2 image rendering and edit rehydration

### LINKED ISSUE #3010: Generated images are not available as editable context in follow-up turns
## Problem

When `image_generate` creates an image, the user can see the result in the UI, but the agent does not retain the generated image as usable visual/file context for later turns.

This breaks a common workflow:

1. User: "Generate an image of X."
2. Agent calls `image_generate`.
3. User: "Now make it darker / change the background / edit the image you just made."
4. Agent does not have the original generated image bytes or a stable source path, so it cannot reliably call `image_edit`.

The user sees "the image," but the model only sees a summarized tool result such as `Generated image (image/png)` in later context.

## Current Behavior

Generated image tools return an `image_generated` sentinel containing a data URL.

The UI/channel layer can display or stage the image, but later LLM context intentionally summarizes generated-image tool results instead of reinjecting the data URL. Existing tests assert this behavior:

- `Thread::messages()` summarizes generated image results on later turns.
- The context avoids replaying `data:image/...` into the model.

This protects context size, but it also means follow-up edit requests lose the actual source image.

## Expected Behavior

After `image_generate` or `image_edit` returns an image, the system should preserve enough structured context for the agent to edit or analyze that generated image in a later turn.

At minimum, the agent should have access to one of:

- a stable workspace/project file path for the generated image, usable by `image_edit`
- a managed generated-image attachment id/path that can be resolved by tools
- a bounded multimodal re-injection path for the most recent generated image

The user should be able to say "edit the image you just generated" and the agent should know which image to use.

## Suggested Approach

Persist generated images to a stable project/workspace path instead of only returning an in-memory data URL.

Possible implementation shape:

- When `image_generate` / `image_edit` returns an `image_generated` sentinel, decode and save the image to a managed workspace location.
- Add the saved path to the sentinel, for example:

```json
{
  "type": "image_generated",
  "media_type": "image/png",
  "path": ".ironclaw/generated-images/<id>.png",
  "prompt": "..."
}
```

- Keep omitting large base64 data from long-term LLM context.
- Include the stable path in future context summaries, for example:

```text
Generated image (image/png) at .ironclaw/generated-images/<id>.png
```

- Teach follow-up turns that generated image paths can be used with `image_edit` / `image_analyze`.

## Acceptance Criteria

- After generating an image, a follow-up request like "edit the image you just made" can call `image_edit` with a valid source image path.
- Generated images have stable paths that survive across turns in the same thread.
- The UI can still display generated images.
- Large base64 payloads are not replayed into LLM context indefinitely.
- Tests cover:
  - generated image sentinel includes a stable path
  - later turns include the generated image path in context
  - follow-up edit can resolve the previous generated image path
  - oversized generated image data still respects existing caps

## Relevant Code

- `src/tools/builtin/image_gen.rs`
- `src/tools/builtin/image_edit.rs`
- `src/generated_images.rs`
- `src/agent/session.rs`
- `src/agent/thread_ops.rs`
- `src/channels/web/util.rs`

## Notes

There is already logic for parsing generated-image sentinels and summarizing them for later turns. The missing piece is making the generated image available as a durable, tool-readable artifact rather than only as UI/display data.

### LINKED ISSUE #3463: Engine V2 generated images do not render correctly in Gateway UI
## Problem

When using `ENGINE_V2=true`, image generation/edit tools can complete successfully, but the generated image may not render as an image card in the Gateway UI. The assistant often replies with text like “image generated successfully,” while the actual image is missing from the chat view.

This makes successful image tool calls look broken to users, especially for flows like:

- Ask the assistant to generate an image.
- Ask the assistant to edit the generated image in a follow-up turn.
- Refresh the Gateway page and expect generated images to remain visible in history.

## Expected Behavior

Generated or edited images from `image_generate` / `image_edit` should render as image cards in Gateway, both live during the turn and when reloading chat history.

Follow-up `image_edit` calls should be able to reuse the generated image artifact path without requiring the user to manually provide it.

## Actual Behavior

Under Engine V2, generated image tool outputs may be treated as normal text/tool output rather than structured generated-image results. In some cases, the LLM also cannot find the generated image path for follow-up edits and may invent invalid paths such as `/tmp/kitten.jpg`.

## Impact

- Image generation appears successful but no image is shown.
- Image editing follow-ups are unreliable.
- Users cannot trust Gateway history to preserve generated image outputs.
- Engine V2 behavior differs from the expected Gateway image rendering flow.

## Notes

This appears related to Engine V2 CodeAct/tool-result handling and Gateway history rehydration, not the image generation API itself. The tool can return valid image data, but the structured image result is not consistently propagated/rendered.

hanakannzashi:
## Summary

- Fixed `ENGINE_V2` image generation/edit results not rendering in the Gateway UI by preserving generated-image sentinels as structured tool results.
- Persisted CodeAct action results into thread metadata so V2 image tool outputs can be rehydrated into v1 chat history.
- Emits generated/edited image events after completed V2 turns using the Gateway-visible thread ID, so realtime image cards are delivered to the correct chat.
- Added tolerant generated-image sentinel parsing for nested JSON and Python `repr(dict)` shapes produced by CodeAct result persistence.
- Stops stalled V2 threads that time out without a pending gate, instead of leaving the Gateway chat indefinitely pending.

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] CI/Infrastructure
- [ ] Security
- [ ] Dependencies

## Linked Issue

Fixes #3010
Fixes #3463

## Validation

- [ ] `cargo build`
- [x] Relevant tests pass: `cargo test parses_python_repr_sentinel_string --lib`, `cargo test persist_v2_tool_calls_reads_codeact_results_from_thread_metadata --lib`, `cargo test persist_v2_tool_calls_preserves_generated_image_results_for_history_rehydration --lib`, `cargo test persist_v2_tool_calls_preserves_image_edit_results_for_history_rehydration --lib`, `cargo test channel_visible_thread_id_prefers_incoming_message_scope --lib`
- [x] Manual testing: restarted Gateway with `ENGINE_V2=true` and confirmed generated images render in chat instead of only appearing as successful tool calls.
- [x] If a coding agent was used and supports it, `review-pr` or `pr-shepherd --fix` was run before requesting review

## Security Impact

No auth, permission, network, sandbox, or approval policy changes. This does persist generated image sentinel metadata, including data URLs when present, in existing conversation tool result rows so Gateway history can rehydrate images.

## Database Impact

No schema or migration


zmanian:
## Code Review

### Overview
+1165/-111 adding `src/image_artifacts.rs` (validated, no-overwrite write of image bytes to `~/.ironclaw/image-artifacts/<user>/<thread>/`), persisting inline image uploads + generated/edited tool outputs, scoping generated-image event IDs by turn, and rehydrating from history. ~9 days old; rebase against current `main` recommended before merge.

### Strengths
- Good defense-in-depth on artifact writes: MIME allowlist, 10 MiB cap, sanitized path segments, `create_new` (no overwrite), cleanup on rejected uploads.
- Turn-scoped `event_id` solves the duplicate tool-call-ID UI bug cleanly.
- Helpers split into a separate module with test seams (`with_artifact_root`, `_at` variants).
- `cargo test --lib image` checked.

### Issues to address before merge

**Critical:**

1. **Path-traversal / arbitrary-host-file-read via `image_edit`** (Copilot, unresolved). The `image_path` doc now says "saved project file path" but `ImageEditTool` is registered with `base_dir: None`, so `validate_path` permits absolute paths like `/etc/passwd`. The LLM could be coaxed into reading and re-uploading arbitrary host files. Must enforce an artifact-root or project-root allowlist before merge.

2. **`decode_image_data_url` decodes base64 before any size check.** A hostile data URL can force a multi-hundred-MB allocation prior to the 10 MiB rejection. Add a pre-decode length bound (≈ `MAX * 4 / 3 + padding`).

**Major:**

3. **SSE mis-routing via tool-controlled metadata.** `enrich_image_tool_result` uses `entry().or_insert_with` for `user_id`/`thread_id` in status metadata. Since `metadata` may be tool-controlled, a malicious tool result could pre-set these and route `ImageGenerated` SSE events to other users. Overwrite unconditionally with the trusted `ctx` values. (This is the same metadata-trust anti-pattern PR #3390 just fixed on the gateway side — worth aligning on the `with_metadata`-clobber discipline that PR introduces.)

4. **`persist_inline_image_attachments_at` clones the entire `Vec<IncomingAttachment>`** (including `data: Vec<u8>` up to 10 MiB each) just to populate `local_path`. Take ownership earlier or mutate in place.

5. **Validation gaps**: `cargo clippy -- -D warnings` and `cargo build` not run; `--features integration` skipped. Project standard requires zero clippy warnings.

**Minor:**

6. Confirm the new SSE `ImageGenerated` flow originates from a typed source log per project rules ("Every AppEvent → SSE/WS must come from a typed source log"). Diff shows direct `channels.broadcast_*` — verify upstream is in the projection dispatcher and not a direct emit, or annotate `// projection-exempt: <category>, <detail>`.
7. `decode_image_data_url` is `pub(crate)` — keep it crate-private; add unit tests for malformed inputs (no comma, missing base64 marker, oversized).
8. Verify no `.unwrap()`/`.expect()` slipped into production paths. The `unwrap_or_default()` on `data_url()` is OK but worth a `// silent-ok:` comment if it stays.
9. 1165-line PR with new persistence surface and security implications; consider splitting the artifact module + path hardening into a precursor PR.

### Risk
Medium-high. Two security issues (arbitrary file read via `image_edit`, SSE mis-routing via untrusted metadata) and one DoS pre-condition (decode-before-size). All directly tied to new code paths.

### Recommendation
**Request Changes.** Block on the `image_path` sandboxing fix (Critical 1) and the metadata user_id/thread_id overwrite (Major 3); fix decode-size ordering (Critical 2); run full clippy + integration suite; rebase against current `main` before re-review.

crates/ironclaw_engine/orchestrator/default.py
```diff
@@ -231,6 +231,36 @@ def format_output(result, max_chars=8000):
...
+
+
+def _persistable_action_result(r, max_output_chars=1000):
+    """Compact CodeAct action results for host-side history persistence."""
+    if not isinstance(r, dict):
+        return None
+
+    action_name = r.get("action_name", "unknown")
+    output = r.get("output")
+    if _is_generated_image_output(output):
```
serrrfirat, line 274: Medium: this branch preserves the full `image_generated` dict, including `data`, in `_persisted_action_results_log`. That log is copied into thread metadata later, so CodeAct image results can persist multi-MB base64 payloads despite the v1 history cap. Once the bridge has persisted an artifact path/event_id, store a compact sentinel here (type/media_type/path/event_id, omit or cap `data`) and add an oversized-image regression.

serrrfirat (commented):
Overall: image artifact/path plumbing looks directionally good, and the sandbox/path validation pieces are much safer than the previous raw-path flow. I found one Medium resource-bound issue before I’d call this merge-ready: Engine V2 CodeAct image results are preserved as full `image_generated` dicts in runtime metadata, so large base64 payloads can bypass the later v1 history/DB cap. Please compact/omit `data` in `_persisted_action_results_log` once `path`/`event_id` exists, and add an oversized-image regression around that path.

serrrfirat:
Parallel review findings:

1. **High — Security / attachment trust boundary** — `src/bridge/router.rs:4223`

   ENGINE_V2 runs input validation, policy checks, and secret scanning on raw `content` at `:4167`, `:4181`, and `:4194`. Attachments are then persisted/augmented, and `effective_content` (including extracted text, filenames, metadata, and persisted paths) is sent to `handle_user_message` and dual-written to DB at `:4324` / `:4387` without rerunning those checks. Malicious attachment text/filename or leaked tokens inside extracted content can bypass inbound safety in ENGINE_V2.

   Suggested fix: after `effective_content` is built, run the same `validate_input`, `check_policy`, and `scan_inbound_for_secrets` checks before event missions / `handle_user_message`; clean up artifacts on rejection. Add a caller-level ENGINE_V2 test with attachment-derived secret/policy text proving rejection.

2. **High — Durability / thread metadata bloat** — `crates/ironclaw_engine/orchestrator/default.py:245`

   CodeAct image action results are exempted from `max_output_chars` and stored whole in `_persisted_action_results_log`; `sync_runtime_state` then copies that array into thread metadata at `crates/ironclaw_engine/src/executor/orchestrator.rs:3055`. Generated image outputs can still include full `data:image/...;base64,...`, so each image can add MBs to engine thread state/metadata, duplicating data already persisted as an artifact and bypassing later sentinel caps.

   Suggested fix: compact generated-image outputs before adding them to `_persisted_action_results_log` (keep type/media_type/event_id/path, omit or cap `data`) and add max entries/bytes guard for the metadata log. Add regression with an oversized data URL proving thread metadata stays bounded while replay still works via artifact/path.

3. **Medium — Information disclosure / path leakage** — `src/image_artifacts.rs:176`

   Persisted image artifacts return canonical absolute host paths, and those paths are embedded/broadcast downstream (`src/agent/attachments.rs:175`, `src/agent/dispatcher.rs:1867`, `src/bridge/router.rs:5732`). This exposes local filesystem layout/usernames to LLM context, DB history, SSE/browser clients, and non-web channels.

   Suggested fix: keep absolute paths server-side only; expose stable artifact IDs or paths relative to the image artifact root, and resolve them inside image-edit/vision tools. Add tests asserting user-visible sentinels/statuses contain no absolute host prefix.

henrypark133 (changes requested):
I reviewed this draft in an isolated worktree with the IronClaw batch-review flow.

What looks good:
- The prior feedback on pre-decode data URL bounds, image path sandboxing, and metadata user/thread routing overwrite appears addressed.

Findings:

1. High - `crates/ironclaw_engine/orchestrator/default.py:238` and `crates/ironclaw_engine/src/executor/orchestrator.rs:3045`

   The new `_persisted_action_results_log` still persists full `image_generated` outputs and then copies that array directly into thread metadata. For image generation results, that output still includes the base64 image data, so this recreates the history/metadata bloat problem the prior review thread called out.

   Please persist a compact image sentinel instead, for example `call_id`, `action_name`, `event_id`, artifact path/reference, media type, and a small preview/status flag, while keeping raw `data:` payloads out of thread metadata.

2. Medium - `src/bridge/router.rs:5806` and `src/bridge/router.rs:5819`

   `persist_v2_tool_calls()` replays image/action results from both `thread.internal_messages` and `PERSISTED_ACTION_RESULTS_METADATA_KEY` without deduplication. If a completed thread carries the same action result in both places, this writes duplicate `tool_calls` entries and emits duplicate generated-image events. Please dedupe by stable `call_id`/`event_id` before appending or emitting.

3. Medium - `src/bridge/router.rs:5672` and `src/generated_images.rs:186`

   The v1 `tool_calls` persistence path still stores the full image sentinel JSON whenever it is below the 512 KiB cap. That means ordinary successful image generations can still write base64 `data:` URLs into persisted tool-call history. Please make the persisted shape compact regardless of size; the artifact reference should be enough for rehydration/editing.

The current CI run is also red on `tests/e2e_builtin_tool_coverage.rs::tests::job_list_cancel`, where the trace exercised `create_job` and `list_jobs` but not `cancel_job`. That should be fixed or explained before this leaves draft.

PR #3071, opened 2026-04-29
feat(reborn): add capability host base

serrrfirat:
## Summary

Adds the first `ironclaw_capabilities` slice for the Reborn caller-facing `CapabilityHost` base workflow.

Scope:

```text
crates/ironclaw_capabilities
crates/ironclaw_run_state (store/status support needed by CapabilityHost)
```

Includes:

- `CapabilityHost::invoke_json` over the neutral `CapabilityDispatcher` port.
- Authorization denial before runtime dispatch.
- Fail-closed behavior for unsupported authorization obligations before dispatch/process start/lease claim.
- Approval block persistence through `ironclaw_run_state` approval request/run-state stores.
- Approved resume flow that validates the approval fingerprint, finds the matching fingerprinted lease, claims it before dispatch, consumes it after dispatch, and completes run state.
- `CapabilityHost::spawn_json` over the neutral `ironclaw_processes::ProcessManager` port.
- Boundary test ensuring no dependency on concrete dispatcher/runtime/host-runtime/secrets/network crates.

Intentional deferrals:

- No concrete dispatcher/runtime lane dependency.
- No `ironclaw_host_runtime` composition helper.
- No built-in obligation handlers or prepared-effect handoff; non-empty obligations fail closed for now.
- No secrets/network/memory work.

## Rebase note

Rebased onto current `origin/reborn-integration` after the WASM runtime lane landed. The `Cargo.toml` workspace conflict was resolved by preserving both:

```text
crates/ironclaw_wasm
crates/ironclaw_capabilities
```

No concrete WASM integration was added here; `ironclaw_capabilities` remains correctly coupled only to neutral host API dispatch contracts (`CapabilityDispatcher`, `DispatchError`, `RuntimeKind`).

The unrelated regression workflow pipefail fix that was previously present on this branch was dropped from this PR to keep the diff scoped to the capability-host slice. The `skip-regression-check` label and commit marker are applied because the current base workflow false-negatives on this PR despite the capability-host contract tests being present and passing locally.

## Exposure checklist

- Does this affect existing `src/` runtime behavior? **No.** New internal Reborn crate only.
- Does this alter app startup/config defaults? **No.**
- Does this expose new routes/CLI/user-visible APIs? **No.**
- Does this create a second writer for existing production state? **No.** It writes only through Reborn run-state/approval/lease/process traits supplied by callers.
- Are all Reborn paths feature-gated or unused by default? **Unused by default.**
- Are docs/status labels accurate? **Implemented slice, not product-complete Reborn capability execution.**

## Verification

```bash
cargo test -p ironclaw_capabilities
cargo clippy -p ironclaw_capabilities --all-targets -- -D warnings
cargo test -p ironclaw_architecture
cargo fmt --all -- --check
git diff --check
```

All passed locally after the rebase.


serrrfirat:
Rebased onto current `reborn-integration` and force-pushed `84c2dbc28`.

What changed during rebase:
- Replayed only the CapabilityHost commit on top of current integration.
- Resolved the `Cargo.toml` workspace member conflict.
- Updated the capability test manifest fixture from obsolete `trust = "user_trusted"` to `trust = "third_party"` to match the trust-policy manifest parser.
- Dropped carried `crates/ironclaw_processes/**` files from the PR diff; current file list is now Cargo files + `crates/ironclaw_capabilities/**` only.

Local verification:
```bash
cargo test -p ironclaw_capabilities
cargo clippy -p ironclaw_capabilities --all-targets -- -D warnings
cargo fmt --check
git diff --check
```

Note: I also tried a broader dependency test including dispatcher/process/auth/run-state/approvals. The CapabilityHost tests passed, but `ironclaw_dispatcher` tests currently fail on an existing obsolete manifest fixture value (`trust = "sandbox"`; parser now expects `untrusted`). I did not patch dispatcher in this PR because it is outside the narrowed CapabilityHost diff.

henrypark133 (commented):
## Review: capability host looks sound, CI needs clearing

I did not find a code blocker in the capability host changes, but I would hold merge until the current red CI check is cleared or explained.

What looks good:
- The capability host keeps dispatch behind the neutral `CapabilityDispatcher` port and avoids pulling concrete runtime/dispatcher ownership into this crate.
- Approval resume validates the run record, approval id, fingerprint, and matching lease before dispatch, with coverage for mutated input and concurrent resume races.
- Authorization denial and unsupported obligations fail before dispatch or process spawn, and the tests exercise the caller-facing host methods rather than only helper functions.

No verified code blockers.

Low-priority notes:
- The current `Regression test enforcement` check is failing even though this PR adds Rust tests and `cargo test -p ironclaw_capabilities` passes locally. This looks like a CI detection false negative or stale check, but the PR should not merge while GitHub still reports a failed required gate.

henrypark133 (changes requested):
## Review: tighten capability-host authority and post-dispatch contracts

The base shape is good, but I think this needs a couple of contract fixes before it lands because this crate is becoming the caller-facing capability execution boundary.

What looks good:
- `CapabilityHost` keeps concrete runtime/dispatcher ownership out of this crate and dispatches through the neutral `CapabilityDispatcher` port.
- Approval resume validates run state, approval id, invocation fingerprint, and the matching lease before dispatch.
- Denials and unsupported obligations fail before runtime dispatch, process start, or lease claim, and the tests exercise the caller-facing host methods.

### Concerning: host authorization currently bypasses the trust-aware contract
`CapabilityHost` stores `&dyn CapabilityDispatchAuthorizer`, so `invoke_json`, `resume_json`, and `spawn_json` only call the grant-only `authorize_dispatch` / `authorize_spawn` methods. The stacked Reborn authorization layer already has `TrustAwareCapabilityDispatchAuthorizer`, where the policy-derived `TrustDecision` and authority ceiling are enforced. If this host is the caller-facing capability boundary, future host-runtime wiring can accidentally route capability execution through this grant-only path and miss the trust-policy ceiling.

Please either route the host authorization path through the trust-aware contract, or make this explicitly a grant-only base slice with a documented follow-up/adapter requirement that prevents product wiring from using `CapabilityHost` without trust-policy enforcement.

### Concerning: successful side effects can be reported as invocation failures
`invoke_json`, `resume_json`, and `spawn_json` all perform the side effect first, then return an error if `run_state.complete(...)` fails. That means the capability dispatch or process spawn can already have happened while the caller sees a failed invocation, which creates retry/duplication risk. The code already treats lease-consume failure after successful dispatch as cleanup by warning and returning the dispatch result; the run-state completion failure needs an equally explicit contract.

Please define the intended behavior for post-side-effect bookkeeping failures and add caller-level coverage for it. If successful dispatch/spawn is authoritative, completion failure should likely be handled as reconciliation/observability rather than changing the user-visible invocation result.

### Concerning: the boundary docs do not yet make the single-authority path explicit
`crates/ironclaw_capabilities/CLAUDE.md` says this crate owns the invoke/resume/spawn workflow and must use neutral ports, but it does not say that host-runtime adapters, built-ins, custom packages, and external runtimes must converge through this workflow instead of adding parallel authorization/approval dispatch paths. Since this is a foundational Reborn slice, please make that invariant explicit in the crate guardrails so future slices do not accidentally split authorization ownership.

Issue #3031, opened 2026-04-28
[EPIC] Reborn product surface migration


serrrfirat:
Parent: #2987  
Blocking compatibility gate: #3020

## Purpose

Track the **product-surface migration work** needed after the Reborn substrate stack lands, so current IronClaw v1 user/operator behavior is preserved while the host/control-plane/runtime architecture moves onto Reborn interfaces.

This issue intentionally does **not** duplicate the PR0–PR8 substrate work tracked in #2987. It assumes #2987 owns the Reborn internals:

- host API, resources, extension manifest registry, architecture guardrails
- filesystem and event/audit substrate
- authorization, approvals, run-state, `CapabilityHost`
- dispatcher, process lifecycle, WASM/Script/MCP runtime lanes, host-runtime composition
- secrets/network substrates
- built-in obligations and prepared-effect handoff
- memory storage/search/plugin substrate
- post-group cutover blockers called out in #2987

This issue tracks the work to wire existing product behavior onto those interfaces.

## Principle

Do **not** rebuild v1 as a parallel architecture.

Use this rule:

| Current v1 code is... | Migration approach |
| --- | --- |
| Protocol-specific translation, UI behavior, CLI parsing, provider-specific API code, tool business logic | Reuse/bridge it behind Reborn ports |
| Authority, approvals, auth-blocked state, capability execution, process lifecycle, resource accounting, durable events, secret injection | Move to Reborn services |
| Current v1 has no clean boundary or stores source-of-truth in the wrong layer | Remake the boundary, while reusing internals where safe |

### 1. Gateway / web chat

Preserve current browser/API behavior while routing through Reborn services.

Current surfaces include:

- `/api/chat/send`
- `/api/chat/events`
- `/api/chat/ws`
- `/api/chat/history`
- `/api/chat/threads`
- `/api/chat/thread/new`
- `/api/chat/gate/resolve`
- legacy approval/auth shims where still required
- `/v1/chat/completions`
- `/api/v1/responses` and `/v1/responses`
- static SPA behavior
- SSE/WebSocket fanout, reconnect, history rebuild
- attachment handling and HTML injection safety
- multi-tab fanout / connection caps

Target Reborn interfaces:

- `TransportAdapter`
- `TurnService`
- `ConversationManager`
- `EventStreamManager`
- `ProjectionReducer`
- `RunStateManager`

Approach: **bridge route/frontend compatibility, remake backend ownership**.

[...]

PR #3099, opened 2026-04-29
Add Reborn transport adapter contract

zmanian:
## Summary
- add the policy-free `ironclaw_transport` contract crate with typed routes, ingress/egress, registry lifecycle, health, and redacted transport errors
- bridge existing v1 `Channel`s into `TransportAdapter`s and wire opt-in `REBORN_TRANSPORT=true` startup through `RebornTransportRuntime`
- route legacy `ChannelManager` replies/status egress through the transport registry while preserving existing `StatusUpdate` shapes
- consolidate hot-add/remove lifecycle so late channels register for egress and, when Reborn transport is active, start through the same ingress sink
- document the contract and add dependency boundary guardrails

## Review
- self-reviewed the final diff after consolidation; no blocking findings remain
- fixed review-found gaps around hot-added channel ingress and status routing target vs user scope separation before opening this PR

## Tests
- `cargo test -p ironclaw_transport`
- `cargo test -p ironclaw transport`
- `cargo clippy -p ironclaw_transport --all-targets -- -D warnings`
- `cargo clippy -p ironclaw --lib -- -D warnings`
- `cargo clippy -p ironclaw --tests -- -D warnings`
- `cargo fmt --check`
- `git diff --check`

Refs #2987


zmanian:
Addressed all PR review feedback in 39f7c4ee.

- Preserved attachment inline bytes across the transport <-> legacy bridge and updated the transport contract docs accordingly.
- Made TransportRegistry.start_all tolerant of per-adapter startup failures while still failing when no adapters start, matching ChannelManager semantics.
- Replaced random fallback message IDs with deterministic UUID v5 mapping for non-UUID transport IDs, including caller-path coverage through LegacyAgentTransportSink.

Verification:
- cargo test -p ironclaw_transport
- cargo test -p ironclaw transport
- cargo test -p ironclaw_architecture reborn_crate_dependency_boundaries_hold
- cargo clippy -p ironclaw_transport --all-targets -- -D warnings
- cargo clippy -p ironclaw --lib -- -D warnings
- cargo clippy -p ironclaw --tests -- -D warnings
- cargo fmt --check
- git diff --check

zmanian:
Resolved the reborn-integration merge conflict in bb7043c0.

Conflict was limited to the workspace members list in Cargo.toml; the resolution keeps the transport crate from this PR plus the new secrets/network crates from reborn-integration. Pushed the merge commit, and GitHub now reports the PR as mergeable.

Verification after merge:
- cargo fmt --check
- git diff --check
- cargo test -p ironclaw_transport
- cargo test -p ironclaw_architecture reborn_crate_dependency_boundaries_hold
- cargo test -p ironclaw transport
- cargo clippy -p ironclaw_transport --all-targets -- -D warnings
- cargo clippy -p ironclaw --tests -- -D warnings

src/reborn/transport.rs
```diff
@@ -0,0 +1,667 @@
...
+        ingress.route.scope.user_id.as_str().to_string(),
+    );
+    if let Some(recipient) = &ingress.route.recipient {
+        metadata_insert(&mut metadata, "transport_recipient", recipient.clone());
+        if metadata.get("target").is_none() {
+            metadata_insert(&mut metadata, "target", recipient.clone());
+        }
+    }
+    if let Some(conversation_id) = &ingress.route.conversation_id {
+        metadata_insert(
```
serrrfirat, line 261: **High Severity** — v5 UUID derivation collides across adapters and reuses the same namespace as message_id ↔ legacy_id translation in other code paths.

The fallback `Uuid::new_v5(&Uuid::NAMESPACE_OID, transport_message_id.as_bytes())` is keyed only on the raw transport message id. Two different adapters (e.g. a Telegram bot and a webhook) that both emit `"42"` as their natural message id will collapse to the same legacy `Uuid`. That id is then used as the agent-side correlation key for replies, status updates, and approval gates — collisions silently cross-wire conversations between users.

The regression test `non_uuid_transport_message_ids_map_to_deterministic_legacy_ids` enshrines exactly this collision behavior, so once this lands the bug becomes a contract.

**Suggested fix:** include the adapter id (and ideally the transport route user id) in the v5 input so adapters can't clobber each other's id space:

```rust
let seed = format!("{}::{}", adapter_id.as_str(), transport_message_id);
Uuid::new_v5(&Uuid::NAMESPACE_OID, seed.as_bytes())
```

Update the test to assert that two adapters with the same raw id produce different `Uuid`s.

crates/ironclaw_transport/src/lib.rs
```diff
@@ -0,0 +1,691 @@
...
+                Ok(()) => started += 1,
+                Err(error) => {
+                    first_error.get_or_insert_with(|| {
+                        TransportError::new(
+                            error.kind(),
+                            format!(
+                                "transport adapter '{adapter_id}' failed to start: {}",
+                                error.safe_reason()
+                            ),
+                        )
```
serrrfirat: **High Severity** — `shutdown_all` aborts on first adapter failure, leaking the rest.

This is asymmetric with `start_all`, which is documented as best-effort and continues past failures. On shutdown the iterator returns early on the first `Err`, so any remaining adapters never get `shutdown()` called. In practice that means a misbehaving adapter (panicking sink, hung HTTP client, dead socket) prevents clean teardown of every adapter registered after it — leaving sockets, file handles, background tasks, and credential material live past process exit.

Shutdown paths must be best-effort by construction; failing fast on cleanup is an anti-pattern.

**Suggested fix:** mirror `start_all`'s pattern — collect all errors, call `shutdown()` on every adapter unconditionally, then return an aggregated error (e.g. `TransportError::Multi(Vec<...>)` or a single error with a joined message) only after every adapter has been given a chance to clean up.

Add a regression test analogous to `start_all_is_best_effort` that proves a failing adapter does not prevent shutdown of subsequent adapters.

crates/ironclaw_transport/src/lib.rs
```diff
@@ -0,0 +1,691 @@
...
+    ) -> Result<(), TransportError> {
+        let entries = self.snapshot_entries()?;
+        if entries.is_empty() {
+            return Err(TransportError::new(
+                TransportErrorKind::StartupFailed,
+                "no transport adapters registered",
+            ));
+        }
+
+        let mut started = 0usize;
```
serrrfirat: **Medium Severity** — `health_check_all` short-circuits on the first unhealthy adapter, so callers can't see the full picture.

This means an operator UI / oncall dashboard that calls `health_check_all` only ever learns about the first failing adapter; if two adapters are degraded, the second is invisible until the first recovers. For a registry that's explicitly designed to host many independent transports, this is the wrong shape — health observability should be additive, not transitive.

**Suggested fix:** return `Vec<(TransportAdapterId, Result<HealthReport, TransportError>)>` (or a `HashMap`) so callers always see every adapter's state. If a single bool/Result is needed for legacy callers, add a `health_summary()` helper that derives it from the per-adapter map, but keep the per-adapter detail accessible.

Add a test that two unhealthy adapters both surface in the result.

crates/ironclaw_transport/src/lib.rs
```diff
@@ -0,0 +1,691 @@
...
+    pub fn new(kind: TransportErrorKind, reason: impl Into<String>) -> Self {
+        Self {
+            kind,
+            reason: sanitize_reason(reason.into()),
+        }
+    }
+
+    pub fn kind(&self) -> TransportErrorKind {
+        self.kind
+    }
```
serrrfirat, line 682: **Medium Severity** — `sanitize_reason` is allowlist-by-marker, not deny-by-default; the redaction surface is brittle and silently leaks anything not in the marker list.

The current marker list (`authorization`, `bearer `, `password`, `secret`, `token`, `api_key`, `apikey`, `sk-`, `/users/`, `\\users\\`, `.ironclaw`) is a narrow, hand-curated set. Real-world LLM/SDK error messages routinely include: refresh tokens, OAuth client_ids, pre-signed S3 URLs, full request bodies, JWT payloads (`eyJ...`), Slack webhook URLs, Stripe `pk_`/`whsec_` prefixes, GCP service-account JSON paths, anthropic/openai header echoes (`x-api-key`), database connection strings (`postgres://user:pass@host`), and stack traces with absolute paths from non-`.ironclaw` locations.

Because `TransportError` is the public surface that flows back across the host/service boundary in Reborn, anything that escapes this filter is a confidentiality leak. The risk is highest precisely on the third-party adapter case (which is the whole point of the registry).

**Suggested fix (defense in depth):**
1. Default to a generic `"transport adapter error"` placeholder for unknown error categories, and only allow whitelisted *structured* fields (e.g. `TransportError::Adapter { kind, public_summary }` where `kind` is an enum and `public_summary` is opt-in).
2. Keep `sanitize_reason` as a last-line-of-defense regex pass, but expand the marker set substantially and treat any URL, JWT-shaped string, or path containing a username as suspect.
3. Add a fuzz/property test that shuffles common secret formats through `sanitize_reason` and asserts none survive.

At minimum, document the threat model assumption ("adapter authors are trusted not to put secrets in error reasons") in the crate-level docs so reviewers can see the contract.

src/reborn/transport.rs
```diff
@@ -0,0 +1,667 @@
...
+    metadata_insert(
+        &mut metadata,
+        "transport_adapter_id",
+        ingress.route.adapter_id.as_str().to_string(),
+    );
+    metadata_insert(
+        &mut metadata,
+        "transport_user_id",
+        ingress.route.scope.user_id.as_str().to_string(),
+    );
```
serrrfirat, line 253: **Medium Severity** — `LEGACY_STATUS_METADATA_KEY` round-tripped through caller-supplied metadata is forgeable; downstream code treats it as a trust signal.

The key `"__ironclaw_status_update"` is read out of the inbound `TransportMessage::metadata` map (which is fully controlled by the adapter author and, transitively, by anything the adapter accepts as input). If a third-party transport adapter copies user-controlled headers/fields into `metadata` — which the contract docs encourage as the extension point — a malicious user can inject `__ironclaw_status_update=true` into an ordinary inbound message and have the agent treat it as a system status update.

The consequences depend on how downstream consumers branch on this flag, but at minimum it lets a user bypass normal message handling, and at worst it can suppress approval gates or skip sanitization that's only applied to the user-input branch.

**Suggested fix (any of these):**
1. Reserve the `__ironclaw_*` namespace in the transport contract and have `ChannelTransportAdapter::deliver` strip any caller-supplied keys that match before forwarding into legacy code. Only adapters in `IronClawCore` (or the bridge layer) get to set those keys.
2. Move the legacy status flag out of the metadata map entirely into a typed `TransportMessage::kind` enum (`Inbound { ... } | Status { ... }`).
3. At minimum, add a regression test that constructs a transport message with an attacker-controlled metadata map containing `__ironclaw_status_update=true` and asserts it is *not* treated as a status update.

Option 2 is the cleanest — the metadata map should be opaque key/value sidecar data, not a control channel.

src/channels/manager.rs
```diff
@@ -67,13 +79,33 @@ impl ChannelManager {
             }
         }
 
+        let channel: Arc<dyn Channel> = Arc::from(channel);
+
+        if let Some(registry) = self.transport_egress_registry().await {
```
serrrfirat: **Medium Severity** — `hot_add` falls through to legacy registration even when transport registry registration succeeds, double-registering the channel.

The transport-path branch (this `if let Some(registry)` block, lines ~84-103) only `return Ok(())`s when *both* `transport_egress_registry` and `transport_ingress_sink` are present. If a registry is set but the ingress sink isn't (or vice versa), control falls through `registry.replace(...)` and out of the `if let` block, then continues into the legacy `let stream = channel.start().await?;` and `self.channels.write().await.insert(...)`. The channel is now registered both as a transport adapter and as a legacy v1 channel.

Symptoms will be duplicated inbound messages and inconsistent shutdown — `unregister`/`replace` only touches the transport side, while broadcast and the start-all stream-merge still pull from the legacy side.

For an opt-in `REBORN_TRANSPORT=true` migration this is the kind of silent half-state that's painful to debug.

**Suggested fix:** make the registry/sink coupling explicit. If `transport_egress_registry` is set without a sink (or fails to register) treat that as a hard error, not a silent fallback. Concretely: after `registry.replace(...)?`, if `transport_ingress_sink()` is `None`, either return an error or skip the legacy `let stream` path entirely (still insert into `channels` for respond routing, but don't spawn the merge task). Add a unit test that registers an adapter with only the registry set (no sink) and asserts the v1 stream-merge task is *not* spawned.

serrrfirat:
**Medium Severity (cross-cutting)** — `broadcast` / `broadcast_all` bypass the transport registry while `respond` / `send_status` route through it; this is an inconsistent egress surface.

(Posting as a general comment because `broadcast`/`broadcast_all` have no diff lines in this PR — that's exactly the bug.)

When `REBORN_TRANSPORT=true`, point-to-point egress goes through `TransportRegistry::deliver` (with all of its routing, auth, and sanitization), but broadcast egress in `src/channels/manager.rs` still iterates the legacy `channels` HashMap directly. That means:

- Channels registered only via the transport registry never see broadcasts.
- Channels registered only via the legacy path never see point-to-point messages from agents that have moved to the registry.
- Audit/observability hooks attached at the transport boundary miss every broadcast.

For an opt-in feature flag this is acceptable as a transitional state, but it should be explicit, not implicit. As written, the asymmetry is silent.

**Suggested fix:** either (a) add a `TransportRegistry::deliver_to_all` (or `broadcast`) method and route both legacy and transport paths through it, or (b) explicitly document this asymmetry in `manager.rs` and add a debug-build assertion that fails if both registries contain channels at the same time. Add a regression test that registers a transport adapter and a legacy channel side-by-side, broadcasts a message, and asserts both receive it.

Refs: `src/channels/manager.rs::broadcast` and `::broadcast_all`.

src/reborn/transport.rs
```diff
@@ -0,0 +1,667 @@
...
+        metadata_insert(
+            &mut metadata,
+            "transport_thread_id",
+            thread_id.as_str().to_string(),
+        );
+    }
+
+    let message_id = Uuid::parse_str(&transport_message_id)
+        .unwrap_or_else(|_| Uuid::new_v5(&Uuid::NAMESPACE_OID, transport_message_id.as_bytes()));
+    let sender_id = metadata
```
serrrfirat: **High Severity** — `sender_id` spoofing via attacker-controlled metadata. **A test in this PR enshrines the bug.**

The `sender_id` for the synthesized `IncomingMessage` is read from `metadata.get("sender_id")` *before* falling back to `route.recipient` and `route.scope.user_id`. The `metadata` map is the merge of `ingress.metadata` and `ingress.route.metadata`, both of which can carry attacker-controlled keys (the route metadata is the adapter's normalized view; the ingress metadata is more raw). A user that includes `"sender_id": "admin"` (or `"system"`, or another user's id) in their inbound message metadata will have `IncomingMessage.sender_id == "admin"` after this conversion.

The test `legacy_agent_transport_sink_injects_message_with_typed_route_fields` passes `metadata: { "sender_id": "alice-transport", "user_id": "mallory", "thread_id": "wrong-thread" }` and asserts `message.sender_id == "alice-transport"`. That's not a regression test — that's a proof of the spoofing surface, codified.

`sender_id` is used in many places in the v1 codebase for routing, attribution, and (in some channels) authorization. Letting users override it via metadata is a privilege/identity escalation path.

**Suggested fix:**
```rust
let sender_id = ingress
    .route
    .recipient
    .clone()
    .unwrap_or_else(|| ingress.route.scope.user_id.as_str().to_string());
```
Delete the `metadata.get("sender_id")` lookup entirely — `sender_id` should *only* come from typed route fields. Update the test to assert `message.sender_id == "chat-1"` (the recipient) regardless of what the metadata claims, and add a hostile-metadata test that asserts the spoofed `sender_id` is **not** honored.

src/reborn/transport.rs
```diff
@@ -0,0 +1,667 @@
...
+        storage_key: attachment.storage_ref.clone(),
+        local_path: attachment
+            .metadata
+            .get("local_path")
+            .and_then(Value::as_str)
+            .map(ToString::to_string),
+        extracted_text: attachment
+            .metadata
+            .get("extracted_text")
+            .and_then(Value::as_str)
```
serrrfirat: **High Severity** — `local_path` and `extracted_text` are read from arbitrary attacker-controlled `attachment.metadata`.

`TransportAttachment.metadata` is opaque key/value carried from the inbound transport. A malicious user (or a poorly-isolated third-party adapter) can inject any string under `local_path` or `extracted_text`, and that string flows directly into `IncomingAttachment` for downstream consumption.

Concrete attack vectors:
1. **Path traversal via `local_path`:** any code path that opens `IncomingAttachment.local_path` (file read, OCR, transcription, image processing) will open `/etc/passwd`, `/Users/<user>/.ironclaw/secrets`, etc. — whatever the user puts there. The `transport_ingress_to_incoming_message_maps_attachment_descriptors` test even passes `local_path: "tmp/diagram.png"` from arbitrary metadata, so the unsanitized passthrough is the documented contract.
2. **Prompt injection via `extracted_text`:** this field is presumably treated as ground-truth OCR output by downstream code (search, summarization, LLM input). An attacker who controls attachment metadata controls 'what the OCR said' — classic prompt-injection.

**Suggested fix:**
- Drop both fields from the metadata-passthrough path. `local_path` should only be set by trusted, in-process attachment processors; transport ingress should never carry it across the boundary.
- For `extracted_text`, either drop it entirely (let in-process OCR re-derive it) or namespace it as `transport_provided_extracted_text` and have downstream consumers know it's untrusted.
- At minimum, validate `local_path` against the workspace root (canonical-path containment check) before honoring it.

This is the same shape as the `__ironclaw_status_update` issue (R1-7): metadata is a control channel when it should be opaque sidecar data.

src/channels/transport_adapter.rs
```diff
@@ -0,0 +1,1449 @@
...
+
+fn transport_route_from_channel_status(
+    channel_name: &str,
+    metadata: &Value,
+) -> Result<TransportRoute, TransportError> {
+    let user_id = metadata_string(metadata, "transport_user_id")
+        .or_else(|| metadata_string(metadata, "owner_id"))
+        .or_else(|| metadata_string(metadata, "user_id"))
+        .or_else(|| metadata_string(metadata, "sender_id"))
+        .unwrap_or_else(|| "default".to_string());
```
serrrfirat: **High Severity** — fail-open fabrication of `user_id = "default"` when no metadata key matches.

This 4-deep fallback chain (`transport_user_id` → `owner_id` → `user_id` → `sender_id` → `"default"`) ends in a synthetic phantom user. If `UserId::new("default")` succeeds (and based on the loose validation in `host_api`, it likely does), then any status update without a user-attributed metadata key gets routed under the identity of a user named `default`.

Two failure modes:
1. **Phantom attribution:** state changes (status updates, projection sync) get logged/stored against `default`. If multiple unattributed events flow through, they all collapse into one phantom identity, conflating different real users' state.
2. **Privilege confusion:** if any downstream policy treats `default` as an admin/system identity (or, conversely, a trusted bypass), an attacker can force this branch by sending a status-triggering message with no user metadata.

Fail-closed is the only safe option here.

**Suggested fix:**
```rust
let user_id = metadata_string(metadata, "transport_user_id")
    .or_else(|| metadata_string(metadata, "owner_id"))
    .or_else(|| metadata_string(metadata, "user_id"))
    .or_else(|| metadata_string(metadata, "sender_id"))
    .ok_or_else(|| TransportError::new(
        TransportErrorKind::InvalidRequest,
        "status update has no user-attributed metadata; refusing to fabricate identity",
    ))?;
```

If there are call sites that legitimately have no user context (e.g. system-wide broadcasts), they should construct an explicit `TransportRoute` with a typed system identity, not flow through this metadata-derivation path.

src/channels/transport_adapter.rs
```diff
@@ -0,0 +1,1449 @@
...
+                            "dropping channel message that could not be normalized for transport"
+                        );
+                    }
+                }
+            }
+        });
+        Ok(())
+    }
+
+    async fn deliver(
```
serrrfirat, line 103: **Medium Severity** — fire-and-forget `tokio::spawn` with no `JoinHandle` or `AbortHandle`; lifecycle relies on implicit channel-close.

Three distinct problems with this spawn:

1. **No abort path on `shutdown()`.** The task only exits when `stream.next()` returns `None`. `Self::shutdown()` calls `self.channel.shutdown()` and *hopes* the channel impl correctly closes its stream. If a channel's `shutdown()` is async-cancellation-safe but doesn't actually close the stream (returns Ok without dropping the producer), this task leaks forever — holding the `sink` Arc and the channel `Stream` Box.

2. **Panics are silent.** If `incoming_message_to_transport(&message)` panics (e.g., on a malformed message_id from a broken adapter), tokio prints a backtrace and the task dies. The adapter is now half-dead: `deliver()` still works, but `start()`-spawned ingress is gone. There's no watchdog, no restart, no propagation to a parent task.

3. **`health_check()` doesn't reflect the dead task.** Line 158: `health_check` calls `channel.health_check().await` and returns `Healthy` on success. It has no way to know the spawned ingress task crashed/exited — a downstream operator sees green health while no messages are flowing.

**Suggested fix:**
- Store the `JoinHandle` (or an `AbortHandle`) in `ChannelTransportAdapter` so `shutdown()` can call `handle.abort()` for hard cancellation after the polite `channel.shutdown()` path.
- Wrap the spawned future in a panic-catching outer task (`std::panic::AssertUnwindSafe` + `FuturesExt::catch_unwind`, or a custom guard) that records the failure into an `AtomicBool` / `oneshot::Sender`.
- Have `health_check()` consult that flag and return `TransportHealth { status: Unhealthy, detail: Some("ingress task died"), ... }` if set.

This is the kind of thing that's hard to retrofit later because every adapter author who's already implementing the trait will skip it. Bake it into the wrapper now.

src/channels/transport_adapter.rs
```diff
@@ -0,0 +1,1449 @@
...
+        LEGACY_STATUS_METADATA_KEY.to_string(),
+        legacy_status_payload(&status),
+    );
+
+    Ok(TransportEgress::Status(TransportStatusUpdate {
+        route: transport_route_from_channel_status(channel_name, metadata)?,
+        status: transport_status_from_channel_status_kind(&status),
+        metadata: transport_metadata,
+    }))
+}
```
serrrfirat, line 283: **Medium Severity** — `LEGACY_STATUS_METADATA_KEY` payload exposes internal state to third-party transport adapters.

`transport_status_from_channel_status` packs the *full* `StatusUpdate` into `transport_metadata[LEGACY_STATUS_METADATA_KEY]` (see `legacy_status_payload` below). That payload includes:
- `cost_guard`: `session_budget_usd`, `spent_usd`, `remaining_usd`
- `secrets_status`: `count`, `vault_unlocked`
- `tool_completed` / `tool_started` / `tool_result`: tool names, parameters, full output preview
- `reasoning_update`: full `narrative` and per-tool `decisions[].rationale` (LLM-generated, may include sensitive context from the user's message)
- `turn_cost` / `turn_metrics`: model identity, token counts, cost

The in-process `ChannelTransportAdapter::deliver` strips this key (`strip_legacy_status_metadata(&mut metadata)`) before forwarding to a v1 channel impl — but third-party `TransportAdapter` impls do *not* go through that strip. A registered third-party adapter receives `TransportEgress::Status` with the raw `LEGACY_STATUS_METADATA_KEY` payload in its `metadata` field.

Since the registry contract is explicitly designed for third-party adapters (it's the whole point of the abstraction), this is a confidentiality leak via the legacy bridge.

**Suggested fix:**
- Move the legacy payload out of public metadata. Either:
  - (a) Don't pack it at all — let the in-process adapter call a sibling helper that has direct access to the original `StatusUpdate`, bypassing the metadata round-trip, or
  - (b) Use a private namespace (`internal_legacy_status: ...`) that the registry strips before handing egress to *any* adapter, not just `ChannelTransportAdapter`.
- Document the threat model assumption in `crates/ironclaw_transport/src/lib.rs` ("adapters in the registry must be trusted with status payload metadata") so reviewers understand the contract.

Option (a) is cleaner — the legacy payload is a v1 compatibility detail and shouldn't bleed through the v2 contract surface.

crates/ironclaw_transport/src/lib.rs
```diff
@@ -0,0 +1,691 @@
...
+                            ),
+                        )
+                    });
+                }
+            }
+        }
+
+        if started == 0 {
+            return Err(first_error.unwrap_or_else(|| {
+                TransportError::new(
```
serrrfirat: **Medium Severity** — `start_all` swallows N-1 errors when at least one adapter starts.

The loop captures only `first_error` via `get_or_insert_with`, then *only returns it* if `started == 0`. If one adapter starts successfully and four fail, the function returns `Ok(())` and the four failures are silently dropped — there's no logging, no return value, no signal.

The test `registry_start_all_keeps_starting_after_adapter_failures` asserts `Ok(())` and `started_count == 1` for both the failing and healthy adapter. It does not check that the failing adapter's error was reported anywhere. So the silent-drop behavior is currently a contract.

In production this means: an operator sees 'transports started successfully' while half their channels are down. Combined with the existing finding that `health_check_all` short-circuits on the first failure, there is no API path that gives an operator the full picture of which adapters are alive.

**Suggested fix:** mirror the shape proposed for `health_check_all` — return per-adapter results so the caller always sees every outcome:

```rust
pub async fn start_all(
    &self,
    sink: Arc<dyn TransportIngressSink>,
) -> Result<BTreeMap<TransportAdapterId, Result<(), TransportError>>, TransportError> {
    // ...iterate, collect per-adapter Result, return all of them.
    // Top-level Err only if NO adapters were registered.
}
```

If you want to keep the `Result<()>` shape for backward compat, at minimum: `tracing::warn!` each non-first-error inside the loop, and update the regression test to assert the warn is emitted. Better yet: return a `TransportError::PartialStartup { failures: Vec<(adapter_id, error)> }` variant when 0 < failures < total.

crates/ironclaw_transport/src/lib.rs
```diff
@@ -0,0 +1,691 @@
...
+        self.kind
+    }
+
+    pub fn safe_reason(&self) -> &str {
+        &self.reason
+    }
+}
+
+fn sanitize_reason(reason: String) -> String {
+    let trimmed = reason.trim();
```
serrrfirat, line 701: **Medium Severity** — `sanitize_reason` runs at construction (`new()`) instead of serialization, destroying original error context with no recovery path.

When the upstream error is `"timed out connecting to api.openai.com (token sk-foo expired)"`, `TransportError::new(...)` immediately collapses the `reason` field to `"redacted"`. The original string is not preserved anywhere — no `raw_reason` private field, no log emission before sanitization, nothing in `Debug` either (since `#[derive(Debug)]` prints the post-sanitization struct).

This is the worst of both worlds: third-party adapters can't be trusted to omit secrets (so we redact), but in-process callers also can't see what actually happened (so we can't debug). The `?`/`map_err`/`From` paths all run through `new()` first.

**Suggested fix:** keep both representations.
```rust
pub struct TransportError {
    kind: TransportErrorKind,
    raw_reason: Cow<'static, str>,    // never serialized, never crosses boundaries
    safe_reason: String,              // already-sanitized, public
}

impl TransportError {
    pub fn new(kind: TransportErrorKind, reason: impl Into<Cow<'static, str>>) -> Self {
        let raw = reason.into();
        let safe = sanitize_reason_str(&raw);
        Self { kind, raw_reason: raw, safe_reason: safe }
    }
    pub fn safe_reason(&self) -> &str { &self.safe_reason }
    pub(crate) fn raw_reason_for_local_logs(&self) -> &str { &self.raw_reason }
}
```

Make `Debug` print only `kind + safe_reason` (or gate raw access behind a `#[cfg(debug_assertions)]` helper) so the raw string can never be accidentally serialized into a public surface. The `#[error("{kind}: {reason}")]` already only references the safe field — keep that.

This lets operators tail logs and see the actual provider failure during incident triage without having to bisect through redacted error chains.

src/channels/manager.rs
```diff
@@ -132,12 +164,75 @@ impl ChannelManager {
...
+    /// channels against the same ingress sink.
+    pub async fn set_transport_runtime(
+        &self,
+        registry: Arc<TransportRegistry>,
+        sink: Arc<dyn TransportIngressSink>,
+    ) {
+        *self.transport_egress_registry.write().await = Some(registry);
+        *self.transport_ingress_sink.write().await = Some(sink);
+    }
+
```
serrrfirat, line 223: **Medium Severity** — non-atomic two-lock setup creates a race window where `hot_add` half-registers an adapter.

`set_transport_runtime` writes to `transport_egress_registry` and `transport_ingress_sink` under *separate* `RwLock<Option<...>>`s, with two distinct `.write().await` calls. Between them, an external observer can see `Some(registry) + None(sink)`.

If `hot_add` runs in that window:
1. `transport_egress_registry().await` returns `Some(registry)` → enters the transport branch.
2. `registry.replace(adapter.clone() as Arc<dyn TransportAdapter>)` succeeds — adapter is now registered.
3. `transport_ingress_sink().await` returns `None` → falls through past the `if let Some(sink)` block.
4. Code continues to the legacy fall-through: `let stream = channel.start().await?` and inserts into `self.channels`.

Result: the registry holds a `ChannelTransportAdapter` whose `start()` was never called (so its spawned ingress task doesn't exist), AND the legacy code path spawned its own forwarding task. Egress through `respond` will route via the registry → adapter.deliver → channel.respond (works), but ingress and shutdown semantics are split across two paths.

Same race exists in `clear_transport_egress_registry` (clear registry first, sink still set) and on the inverse — a `respond` call that observes `transport_egress_registry()` returns `Some` but routes to an adapter that's about to be cleared.

**Suggested fix:** combine both into a single `RwLock<Option<TransportComposition>>` where `TransportComposition { registry: Arc<TransportRegistry>, sink: Arc<dyn TransportIngressSink> }` is set/cleared as one atomic unit:

```rust
struct TransportComposition {
    registry: Arc<TransportRegistry>,
    sink: Arc<dyn TransportIngressSink>,
}

pub struct ChannelManager {
    transport: RwLock<Option<TransportComposition>>,
    // ...
}
```

This makes the half-state structurally impossible. Update `set_transport_runtime`/`clear_transport_egress_registry`/`hot_add`/`respond`/`send_status` to read/write the composition as one value.

src/agent/agent_loop.rs
```diff
@@ -1301,7 +1315,12 @@ impl Agent {
             cron_handle.abort();
         }
         self.scheduler.stop_all().await;
-        self.channels.shutdown_all().await?;
+        if let Some(runtime) = reborn_transport_runtime {
+            self.channels.clear_transport_egress_registry().await;
+            runtime.shutdown().await?;
```
serrrfirat: **Medium Severity** — shutdown order leaks adapter tasks on `runtime.shutdown()` failure.

`clear_transport_egress_registry().await` runs first, *then* `runtime.shutdown().await?` propagates with `?`. If `shutdown_all` returns an error from any adapter (which is itself early-aborting per R1-2), the outer agent `run()` returns Err, but:

1. The `transport_egress_registry` and `transport_ingress_sink` are already cleared from the manager.
2. Adapter ingress tasks (spawned by `ChannelTransportAdapter::start`, R2-4) are still alive — they hold `Arc<dyn TransportIngressSink>` clones and the underlying channel streams.
3. The `runtime` is dropped at the end of the `run` scope; the `RebornTransportRuntime` Arc to the registry decrements but the registry itself may still be reachable through any adapter `Arc<dyn TransportAdapter>` clone held by the spawned tasks (the task captures `sink`, but the adapter struct itself isn't held by the task; OK there's no leak there). However the spawned ingress tasks will keep running until their channel streams close — which `runtime.shutdown` was supposed to do but failed.

Net effect: half-shutdown state, ingress tasks orphaned, no path to GC them. The agent process exits but the runtime (if any caller persists outside `run`) is in an inconsistent state.

**Suggested fix:**
```rust
if let Some(runtime) = reborn_transport_runtime {
    let shutdown_result = runtime.shutdown().await;
    // Always clear registry, even if shutdown errored — there's no recovery from
    // a partial shutdown, and leaving the registry pointer live is worse than
    // clearing it (callers that race here see no registry, fall through to
    // legacy paths).
    self.channels.clear_transport_egress_registry().await;
    shutdown_result?;
} else {
    self.channels.shutdown_all().await?;
}
```

Also depends on R1-2 (`shutdown_all` should be best-effort, not abort-on-first-error). Together those two changes guarantee that every adapter at least gets its `shutdown()` called, even if any individual one fails.

serrrfirat:
Pushed fixes for all 16 review findings (R1-1 through R1-7 + R2-1 through R2-9) to branch [`pr-3099-review-fixes`](https://github.com/nearai/ironclaw/tree/pr-3099-review-fixes). Five commits, scoped per concern:

| Commit | Findings | Scope |
|---|---|---|
| `bc5510a37` | R1-2, R1-3, R1-5, R2-6, R2-7 | `ironclaw_transport`: per-adapter reports for `start_all`/`health_check_all`, best-effort `shutdown_all`, `raw_reason` for in-process debug, expanded sanitize markers |
| `800d0c7f2` | R1-1, R2-1, R2-2, R2-5 | `src/reborn/transport.rs`: drop `sender_id` metadata spoof, drop `local_path`/`extracted_text` adapter passthrough, include `adapter_id` in v5 UUID seed, strip status sentinel on inbound |
| `94d44fe3e` | R1-7, R2-3, R2-4, R2-5 | `src/channels/transport_adapter.rs`: store `JoinHandle` and reflect dead pump in health, fail-closed status user (no `"default"` fallback), strip `LEGACY_STATUS_METADATA_KEY` on inbound |
| `95c445a51` | R1-4, R1-6, R2-8 | `src/channels/manager.rs`: `TransportComposition` single-lock atomicity, `broadcast`/`broadcast_all` route through registry when active |
| `d56a121c6` | R2-9 | `src/agent/agent_loop.rs`: runtime shutdown before clearing egress registry |

**Quality gate**: `cargo fmt`, `cargo clippy --all --tests --all-features` clean, `cargo test -p ironclaw_transport` 12/12 (3 new tests), `cargo test --lib channels::` 951/951 (1 new fail-closed test), `cargo test --lib reborn::transport::` 9/9 (2 new spoofing/UUID-collision tests). Full `cargo test --lib`: 5586 pass; 3 pre-existing failures verified to also fail on unmodified PR head (`bb7043c0b`).

@zmanian please cherry-pick or merge into the PR branch as you prefer. Each commit stands alone.

src/reborn/transport.rs
zmanian: Fixed in 47615439. v5 UUID derivation now namespaces by `(adapter_id, user_id, transport_message_id)` (NUL-separated) so two adapters that emit the same raw message id no longer collapse to the same legacy UUID. Updated `non_uuid_transport_message_ids_map_to_deterministic_legacy_ids` to assert against the namespaced derivation, and added `non_uuid_transport_message_ids_do_not_collide_across_adapters_or_users` covering both adapter and user scope variation.

crates/ironclaw_transport/src/lib.rs
zmanian: Fixed in 47615439. `shutdown_all` now invokes shutdown on every adapter unconditionally, captures errors per-adapter, and returns an aggregated `TransportError` listing every failure. Added `registry_shutdown_all_invokes_every_adapter_and_aggregates_errors` regression test that registers three adapters (two failing, one healthy), asserts shutdown is called on all three, and asserts both failing adapter ids appear in the aggregated reason.

src/reborn/transport.rs
zmanian: Fixed in 47615439. Removed the `metadata.get("sender_id")` lookup entirely. `sender_id` is now `route.recipient` (typed) when present, falling back to `scope.user_id`. Updated `legacy_agent_transport_sink_injects_message_with_typed_route_fields` to assert `sender_id == "chat-1"` (the recipient) even when hostile metadata sets `sender_id = alice-transport`, and added `hostile_metadata_cannot_spoof_sender_id_when_recipient_is_present` covering a path-traversal-shaped spoof attempt.

src/reborn/transport.rs
zmanian: Fixed in 47615439. `local_path` is no longer derived from transport metadata (always `None` from the bridge — only trusted in-process processors may populate it). `extracted_text` only flows through when explicitly namespaced as `transport_provided_extracted_text` so consumers know it is untrusted. Added `attachment_metadata_cannot_inject_local_path_or_unnamespaced_extracted_text` (with `../../../etc/passwd` and a prompt-injection payload, both rejected) and `attachment_extracted_text_passes_through_only_via_namespaced_key`. Also updated the existing `transport_ingress_to_incoming_message_maps_attachment_descriptors` test to assert `local_path` is `None`.

src/channels/transport_adapter.rs
zmanian: Fixed in 47615439. `transport_route_from_channel_status` now returns `Err(InvalidRequest, "status update has no user-attributed metadata; refusing to fabricate identity")` instead of falling back to `"default"`. The manager treats that error as a best-effort drop for status updates (status delivery is best-effort by design — fabricating identity is the actual harm). Replaced the test that enshrined the fallback (`..._keeps_routing_target_out_of_user_scope`) with `..._drops_status_with_no_user_attribution` asserting zero captured statuses.

crates/ironclaw_transport/src/lib.rs
zmanian: Fixed in 47615439. `health_check_all` now returns `BTreeMap<TransportAdapterId, Result<TransportHealth, TransportError>>` so callers see every adapter's outcome instead of stopping at the first failure. Added `registry_health_check_all_returns_per_adapter_results_without_short_circuit` that registers three adapters (two unhealthy, one healthy) and asserts both unhealthy results AND the healthy result are surfaced. Updated `RebornTransportRuntime::health_check_all` to forward the new shape; no other in-tree callers.

crates/ironclaw_transport/src/lib.rs
zmanian: Fixed in 47615439. Expanded `sanitize_reason` markers significantly (Authorization/Bearer/Basic, GitHub PAT prefixes `ghp_`/`gho_`/`github_pat_`, Slack tokens `xoxb`/`xoxp`, AWS creds, cookie/session, common dotfile paths `.env`/`.aws`/`.ssh`, Linux home roots, and `@`/`://` markers for URLs/emails). Added shape-based heuristics for JWT-looking triplets (3 base64url segments, each ≥ 8 chars), IPv4 quads, and long opaque tokens (≥ 32 chars mixed alpha+digits). Documented the threat model inline. Added regression tests `transport_error_sanitize_redacts_urls_jwts_paths_and_ipv4` (8 cases) and `transport_error_sanitize_keeps_benign_messages` to guard against over-redaction.

src/reborn/transport.rs
zmanian: Fixed in 47615439. Reserved the `__ironclaw_*` namespace (`RESERVED_METADATA_PREFIX` + `strip_reserved_metadata` in the transport crate). The channel bridge now strips reserved keys at both boundaries: `metadata_from_value` strips them on the way IN (so caller-supplied `__ironclaw_status_update` cannot be honored as authoritative), and `strip_legacy_status_metadata` (now namespace-wide) strips them on the way OUT before `ChannelTransportAdapter::deliver` hands metadata to the wrapped channel. Added `caller_supplied_reserved_keys_cannot_forge_legacy_status_payload` that injects `__ironclaw_status_update` via attacker-controlled ingress metadata and asserts it is stripped.

src/channels/manager.rs
zmanian: Fixed in 47615439. Combined `transport_egress_registry` + `transport_ingress_sink` into a single `RwLock<Option<TransportComposition>>` (also addresses #3167180529). `hot_add` now takes a snapshot of the composition: when only the registry is set (egress-only mode set via `set_transport_egress_registry`), the channel is registered for egress and the legacy stream-merge task is intentionally NOT spawned. Added `test_hot_add_with_partial_transport_runtime_does_not_double_register` that wires only the egress registry, hot-adds a stub, sends a message through it, and asserts the message does NOT appear on the inject stream (legacy path suppressed).

src/channels/transport_adapter.rs
zmanian: Fixed in 47615439. `ChannelTransportAdapter` now stores the spawned ingress task in a `Mutex<Option<JoinHandle<()>>>` and tracks an `Arc<AtomicBool>` death flag. The forwarding future is wrapped in `AssertUnwindSafe(...).catch_unwind()` so panics flip the flag, set on `tracing::error!`. Stream-end (clean exit) also flips the flag. `shutdown` aborts the handle before tearing down the wrapped channel; `health_check` returns `Unavailable` if the flag is set. Added `shutdown_aborts_spawned_ingress_task_handle` and `health_check_reflects_dropped_stream_as_unavailable`.

src/channels/transport_adapter.rs
zmanian: Fixed in 47615439 together with #3166811076. The legacy status payload sits under `__ironclaw_status_update`, which is now part of the reserved `__ironclaw_*` namespace. `strip_legacy_status_metadata` strips every reserved key (not just the literal legacy key) before the wrapped channel sees the metadata, and `metadata_from_value` strips reserved keys on ingress. Approval/auth-prompt egress paths also strip reserved keys before forwarding. So the payload remains a usable in-process bridge between `ChannelTransportAdapter` instances but never crosses out to a third-party adapter and cannot be injected from outside.

crates/ironclaw_transport/src/lib.rs
zmanian: Fixed in 47615439. `start_all` still keeps the best-effort startup contract (returns `Ok` if at least one adapter starts) but now `tracing::warn!`s every non-first failure with the adapter id, kind, and safe reason, and emits a summary `tracing::warn!` listing all failed adapter ids when startup completes with partial failures. Existing test `registry_start_all_keeps_starting_after_adapter_failures` already covers the multi-adapter scenario; the failure visibility regression is now in the log output rather than the return type.

crates/ironclaw_transport/src/lib.rs
zmanian: Fixed in 47615439. `TransportError` now stores both `reason` (sanitized via `sanitize_reason`, used for `Display`/`Debug`/equality/serialization) and `raw_reason` (the original input). Added `raw_reason_for_local_logs(&self) -> &str` for trusted in-process callers. Manual `Debug` impl prints only `{kind, reason}`. Manual `PartialEq`/`Eq` ignore `raw_reason`. Added `transport_error_preserves_raw_for_local_logs_but_hides_in_debug_and_display` asserting raw is preserved while `Debug`/`Display` strip it.

src/channels/manager.rs
zmanian: Fixed in 47615439. Combined into a single `RwLock<Option<TransportComposition { registry, sink }>>` so installation is atomic — the half-installed window (registry visible, sink not yet visible) is structurally impossible. `set_transport_egress_registry` writes `Some(... { sink: None })` (egress-only mode); `set_transport_runtime` writes `Some(... { sink: Some(...) })` (full mode). Both go through one `RwLock` write.

src/agent/agent_loop.rs
zmanian: Fixed in 47615439. The shutdown sequence in `agent_loop.rs` now captures `runtime.shutdown().await` into a binding, unconditionally clears the registry/sink via `channels.clear_transport_egress_registry`, and only then propagates the original shutdown error. Egress always falls back to direct channel paths instead of a half-shut-down runtime, even when one adapter's shutdown errored.

src/reborn/transport.rs
zmanian: Note: the prior reply referenced commit 47615439, which was force-pushed off the branch when adopting your pr-3099-review-fixes side branch. The same fix now lives in 800d0c7f (one of the 5 review-fix commits, now part of the PR head).

crates/ironclaw_transport/src/lib.rs
zmanian: Note: the prior reply referenced commit 47615439, which was force-pushed off the branch when adopting your pr-3099-review-fixes side branch. The same fix now lives in bc5510a3 (one of the 5 review-fix commits, now part of the PR head).

src/reborn/transport.rs
zmanian: Note: the prior reply referenced commit 47615439, which was force-pushed off the branch when adopting your pr-3099-review-fixes side branch. The same fix now lives in 800d0c7f (one of the 5 review-fix commits, now part of the PR head).

src/channels/transport_adapter.rs
zmanian: Note: the prior reply referenced commit 47615439, which was force-pushed off the branch when adopting your pr-3099-review-fixes side branch. The same fix now lives in 94d44fe3 (one of the 5 review-fix commits, now part of the PR head).

src/channels/transport_adapter.rs
zmanian: Note: the prior reply referenced commit 47615439, which was force-pushed off the branch when adopting your pr-3099-review-fixes side branch. The same fix now lives in 94d44fe3 (one of the 5 review-fix commits, now part of the PR head).

crates/ironclaw_transport/src/lib.rs
zmanian: Note: the prior reply referenced commit 47615439, which was force-pushed off the branch when adopting your pr-3099-review-fixes side branch. The same fix now lives in bc5510a3 (one of the 5 review-fix commits, now part of the PR head).

src/channels/manager.rs
zmanian: Note: the prior reply referenced commit 47615439, which was force-pushed off the branch when adopting your pr-3099-review-fixes side branch. The same fix now lives in 95c445a5 (one of the 5 review-fix commits, now part of the PR head).

serrrfirat:
Follow-up/replacement tracker: #3269. The current PR framing is stale against the refreshed #3031 layering; #3269 captures the smaller ProductAdapter-over-Reborn-services contract intended to replace this broad transport-adapter approach.

zmanian:
Closing this PR — the broad `ironclaw_transport` framing here has been superseded by the smaller `ProductAdapter`-over-Reborn-services contract from #3269, which is now merged into `reborn-integration` as commit bfe7307 (`feat(reborn): add product adapter contract`, crate `crates/ironclaw_product_adapters/`).

Per the layering correction in #3031 and #3269, the product-surface boundary should sit **above** Reborn services rather than as a generic transport registry wrapping v1 `ChannelManager` as the source of truth. The merged contract:

- defines `ProductAdapter` with typed inbound/outbound DTOs, capability descriptors, host-stamped trusted context, and protocol-auth evidence,
- routes inbound through `ProductWorkflow` / `InboundTurnService` (not direct `TurnCoordinator` calls),
- consumes projection outputs on the outbound side (not raw turn lifecycle events),
- pins boundary tests against `CapabilityHost`/`RuntimeDispatcher`/runner internals,
- is default-off and does not bridge v1 `ChannelManager`.

Concrete protocol adapters (Telegram, Slack, Web, CLI, …) and any host-runtime glue land in their own follow-up PRs against `crates/ironclaw_product_adapters/`. Review-fix work that originated here (UUID-v5 deterministic ids, fail-closed status user, sender-id spoof guard, attachment-path injection guard) is either already reflected in `tests/review_findings_contract.rs` on the new crate or can be re-proposed as a focused PR there if anything is missing — please open a new issue/PR for that rather than rebasing this one.

Refs: #3269, #3031, #2987.

---

Issue #2843, opened 2026-04-22
Cost-based budgeting: replace iteration/time caps with USD budgets cascading user → project → mission → thread


ilblackdragon:
## Summary

Replace the current iteration-count and wall-clock caps (50 iterations, 5-minute Python VM timeout, 500 worker iterations) with a **cost-based (USD) budget system** that cascades across a new explicit hierarchy — **user → project → mission → thread → sub-thread**. Background invocations (heartbeat, routines, missions, jobs) are scheduled with a concrete per-invocation budget rather than an arbitrary step cap. Users set their own daily/monthly spend caps; exceeding them is an explicit approval event, not a silent death.

This is a **full rework**, not an incremental patch. Partial fixes (\"raise max_iterations to 200\") have already been rejected because they re-introduce the same failure mode one rung up. The goal is that IronClaw can run autonomously **for hours** when the work warrants it, and stop within **seconds** when it is wasting money.

**Expected size:** ~1500–2500 LoC across 15–20 files, 3 DB migrations, 2 new crates modules. Plan for 2–3 weeks.

---

## Problem

Two failure modes today, rooted in the same miscalibration:

1. **Too restrictive for real work.** Engine v2's `ThreadConfig::max_iterations = 50` (`crates/ironclaw_engine/src/types/thread.rs:164`) and the orchestrator Python VM's `ORCHESTRATOR_DEFAULT_MAX_DURATION_SECS = 300` (`crates/ironclaw_engine/src/executor/orchestrator.rs:101`) cap autonomous runs at roughly 5 minutes of agent work. Comparable coding agents run for hours without a turn or session cap (see reference research: Claude Code has no hard session timeout, no tool-call cap — stops are budget-driven via token-delta / diminishing-return detection).

2. **Permissive for waste.** Once a run is under the cap, nothing restrains an agent that is re-calling the same tool with the same args, bouncing between two dead ends, or burning LLM spend on a task that has already plateaued. Compaction is off by default (`enable_compaction: false`, `thread.rs:174`), `max_budget_usd` is `None` by default (`thread.rs:172`), and there is no project-level or user-level aggregate spend limit anywhere in the codebase.

The caps are arbitrary because the dimension is wrong. **Iterations and seconds are proxies for cost.** Measure cost directly; let iterations and seconds float.

---

## Design Principles

1. **USD is the primary budget.** Tokens, iterations, and wall-clock are secondary signals used for telemetry and backstops. Every budget decision resolves to a dollar amount.
[...]

PR #3155, opened 2026-05-01
fix(bridge): mission_* tools accept name; resolves #2583

### LINKED ISSUE #2583: [Bug Bash 4/17] Routine creation fails with "5 consecutive code errors"
### Environment

hosted-staging (crab shack)

### Version / Commit Hash

staging-2026-04-17

### QA Test Date

2026-04-17

### Feature Area

Routines → creation and execution

### Steps to Reproduce

1. Send a request like: "Create basic routine for checking price of Bitcoin every five minutes and send me test request right now."
2. Observe initial responses (agent may send initial BTC price checks)
3. Wait for routine setup/execution result

### Expected Behavior

The routine should be created successfully and send a test Bitcoin price message without errors.

### Actual Behavior

The system returns "Error: 5 consecutive code errors" and the routine is not successfully created or executed. The agent initially sends a couple of BTC price checks (e.g. $76,166.245, then $76,182) but then fails at 16:08 with the consecutive code errors message. Related to #2540 (orchestrator timeout) but this is a distinct code execution failure rather than a timeout.

### Logs / Screenshots

Discovered during Bug Bash 4/17 session (UC18: Routine creation via Telegram). Telegram screenshot shows BTC prices sent at 16:04 and 16:06, then error at 16:08.

### Pre-submit checklist

ilblackdragon:
## Summary

- Fixes #2583 — routine creation hitting `5 consecutive code errors`. The actual root cause was much smaller than the linked #2843 epic suggested: every `mission_*` handler in the bridge required an `id` UUID, but the LLM and the `routine_*` alias path naturally pass `name`. Each call collapsed to `invalid mission id: invalid length 0`, retried 4–5×, and tripped the orchestrator's consecutive-error guard.
- Mission/routine tools now accept `name` (preferred) with `id` UUID retained as a legacy fallback; UUIDs stay internal.
- Adds a live regression test that drives the exact bug-bash prompt through engine v2 and verifies the fix end-to-end.

## What changed

| File | Change |
|---|---|
| `src/bridge/effect_adapter.rs` | New `resolve_mission_id` helper. Handlers `mission_get`, `mission_fire`, `mission_pause`, `mission_resume`, `mission_complete`, `mission_update` migrated. `mission_update` renames via `new_name`; legacy `name`-as-rename-target preserved only when a UUID `id` is also passed. Four new unit tests. |
| `src/bridge/engine_actions.rs` | Mission tool schemas advertise `name` as primary, `id` as legacy alternative. Neither field is `required`. `mission_update` documents `new_name` for renames. |
| `tests/support/test_rig.rs` | Adds `channel_handle()` accessor and `send_exec_approval` for the live test's approval responder. |
| `tests/e2e_live_routine.rs` | New live regression. Drives the issue's exact prompt; asserts no `invalid mission id` or `consecutive code errors`; verifies `**[name]**` notifications on both fire and refire. |
| `tests/fixtures/llm_traces/live/routine_btc_create_test_and_refire.json` | Recorded trace for deterministic replay (live run scrub-checked — no PII). |

## How the fix was found

Driving the issue's verbatim prompt through engine v2 with a real LLM (the new test) reproduced the failure on first run. The trace showed:

```
tool 'routine_fire(bitcoin_price_checker)' failed: effect execution error:
  invalid mission id: invalid length: expected length 32 for simple format, found 0
  ... (× 4)
```

`routine_fire` translates to `mission_fire` via `routine_to_mission_alias`, which passes `params.clone()` unchanged. The `mission_fire` handler then tried to parse the routine *name* as a UUID and failed. Five errors in a row → orchestrator killed the thread → \"5 consecutive code errors\" symptom.

## Verification

- `cargo test --features libsql --lib bridge::effect_adapter` — all 113 effect-adapter tests pass, including 4 new ones for the helper / handler / alias paths.
- `cargo test --features libsql --test e2e_live_routine -- --skip routine_btc` — heuristic + structural unit tests pass.
- `IRONCLAW_LIVE_TEST=1 cargo test --features libsql --test e2e_live_routine -- --ignored routine_btc_create_test_and_refire --nocapture` — passes in 49s. Tools observed: `[\"routine_create(*/5 * * * *)\", \"routine_fire(bitcoin-price-check)\", \"routine_history(bitcoin-price-check)\", \"job_status(...)\"]`. No `invalid mission id`. No `consecutive code errors`. Notifications delivered on both the initial fire and the second-turn refire.
- Replay mode (no `IRONCLAW_LIVE_TEST`) loads the committed fixture and runs deterministically.

## Out of scope (parked for separate follow-ups)

The live test surfaced two further observations that are independent of this fix and not test-failing:

1. **Administrative-tool approval gates not surfacing.** Even with `auto_approve_tools=false`, `mission_create / routine_create / mission_fire / routine_fire` (all classified `Administrative` per `crates/ironclaw_engine/src/gate/tool_tier.rs`) ran without firing a single `StatusUpdate::ApprovalNeeded`. Logged as a warning in the test.
2. **Mission-fire child thread sometimes returns narration instead of data.** `build_meta_prompt` (`crates/ironclaw_engine/src/runtime/mission.rs:2132–2139`) instructs the child to FINAL() with \"1. What you accomplished / 2. Next focus / 3. Goal achieved\" — meta-prose. For \"fetch X and surface it\" routines the user actually wants the data. Logged as a warning in the test.

Both warrant their own issues; flagging here so they're not lost.

## Test plan

- [ ] CI green on the new unit tests
- [ ] CI green on the live test in replay mode (uses committed fixture)
- [ ] Manual: bug-bash redo of the original report on staging — confirm no \"5 consecutive code errors\" surface

### DIFF
--- a/crates/ironclaw_engine/src/runtime/mission.rs
+++ b/crates/ironclaw_engine/src/runtime/mission.rs
@@ -136,8 +136,23 @@ pub struct MissionGateInfo {
     /// `name` param, which has higher precedence than the credential
     /// fallback.
     pub parameters: serde_json::Value,
-    /// Engine-side `call_id` used to resume.
-    pub request_id: String,
+    /// Engine-side LLM tool-call id (e.g. `call_657a9167...`). This is
+    /// the identifier the engine needs to inject the resolved action
+    /// result back onto a paused thread when half-2 of #3133 wires up
+    /// auto-resume. **Not** what the user-facing auth tray sees — that
+    /// is `gate_request_id` below. (Per Copilot review on #3155: the
+    /// gateway parses the channel-side `request_id` as a UUID, so
+    /// forwarding the LLM call_id there would 400 every resolve POST.)
+    pub call_id: String,
+    /// Freshly-generated UUID identifying this surfaced gate to the
+    /// user. Forwarded by the bridge into
+    /// `StatusUpdate::AuthRequired.request_id` /
+    /// `StatusUpdate::ApprovalNeeded.request_id` so the gateway UI
+    /// can render the auth-tray entry and the resolve POST handler
+    /// (which `Uuid::parse_str`'s its input) accepts the response.
+    /// Mapping back to the engine `call_id` for half-2 auto-resume is
+    /// out of scope for this PR (#3166).
+    pub gate_request_id: uuid::Uuid,
     /// What kind of resolution unblocks this gate.
     pub resume_kind: crate::gate::ResumeKind,
 }
@@ -2445,7 +2460,8 @@ async fn process_mission_outcome_and_notify(
                 gate_name: gate_name.clone(),
                 action_name: action_name.clone(),
                 parameters: parameters.clone(),
-                request_id: call_id.clone(),
+                call_id: call_id.clone(),
+                gate_request_id: uuid::Uuid::new_v4(),
                 resume_kind: resume_kind.clone(),
             });
         }
@@ -4239,7 +4255,17 @@ mod tests {
             .as_ref()
             .expect("notification must carry MissionGateInfo for GatePaused");
         assert_eq!(gate.action_name, "tool_activate");
-        assert_eq!(gate.request_id, "call-gmail-1");
+        // call_id round-trips the engine's LLM tool-call id verbatim —
+        // the bridge keeps it for half-2 auto-resume but does NOT
+        // surface it to the channel.
+        assert_eq!(gate.call_id, "call-gmail-1");
+        // gate_request_id is a freshly-generated UUID. The bridge
+        // forwards this (not call_id) to StatusUpdate.request_id so
+        // the gateway resolve handler's UUID parse doesn't 400.
+        assert!(
+            !gate.gate_request_id.is_nil(),
+            "gate_request_id must be a freshly-generated UUID"
+        );
         assert_eq!(
             gate.parameters.get("name").and_then(|v| v.as_str()),
             Some("gmail"),
--- a/crates/ironclaw_engine/src/runtime/mission.rs
+++ b/crates/ironclaw_engine/src/runtime/mission.rs
@@ -101,6 +101,45 @@ pub struct MissionNotification {
     pub response: Option<String>,
     /// True if the thread failed.
     pub is_error: bool,
+    /// When the mission's child thread paused on an unresolved gate
+    /// (auth, approval, or external callback) instead of completing,
+    /// the bridge translates this into a channel-side `AuthRequired` /
+    /// `ApprovalNeeded` status so the user's auth tray surfaces the
+    /// prompt. `None` for normal completions and failures.
+    ///
+    /// Pinned by issue #3133: previously the `_ => {}` arm in
+    /// `process_mission_outcome_and_notify` swallowed `GatePaused`
+    /// outcomes, leaving the user with no actionable signal and the
+    /// cron scheduler re-firing the same broken mission every tick.
+    pub gate: Option<MissionGateInfo>,
+}
+
+/// Gate metadata carried alongside a `MissionNotification` when the
+/// mission's child thread paused on an unresolved gate.
+///
+/// The bridge translates this into a `StatusUpdate::AuthRequired`
+/// (or `ApprovalNeeded`) so the user-facing auth tray (gateway UI) /
+/// chat-channel auth prompt fires for the mission, the same way it
+/// fires for foreground threads. The mission itself is moved to
+/// `MissionStatus::Paused` so the cron scheduler stops re-firing
+/// until the gate is resolved.
+#[derive(Debug, Clone)]
+pub struct MissionGateInfo {
+    /// Engine-side gate name (e.g., `auth_required`, `approval_required`).
+    pub gate_name: String,
+    /// Action that paused (e.g., `tool_activate`, `gmail`, `http`).
+    pub action_name: String,
+    /// Original action parameters at the time of the pause. The bridge
+    /// needs these to run the canonical extension-name resolver
+    /// (`AuthManager::resolve_extension_name_for_auth_flow`) — for
+    /// `tool_activate(name="gmail")` the resolver reads the explicit
+    /// `name` param, which has higher precedence than the credential
+    /// fallback.
+    pub parameters: serde_json::Value,
+    /// Engine-side `call_id` used to resume.
+    pub request_id: String,
+    /// What kind of resolution unblocks this gate.
+    pub resume_kind: crate::gate::ResumeKind,
 }
 
 /// Optional updates to apply to a mission via [`MissionManager::update_mission`].
@@ -2247,6 +2286,11 @@ async fn process_mission_outcome_and_notify(
     // Build notification fields while processing the outcome.
     let mut notify_response: Option<String> = None;
     let mut is_error = false;
+    // Gate metadata when the child thread paused waiting for the user.
+    // The bridge translates this into a channel-side AuthRequired /
+    // ApprovalNeeded status update so the auth tray fires for the user.
+    // See `MissionGateInfo` and #3133.
+    let mut gate_info: Option<MissionGateInfo> = None;
 
     match outcome {
         ThreadOutcome::Completed {
@@ -2322,6 +2366,58 @@ async fn process_mission_outcome_and_notify(
             notify_response = Some("Mission thread hit max iterations without completing".into());
             is_error = true;
         }
+        ThreadOutcome::GatePaused {
+            gate_name,
+            action_name,
+            call_id,
+            parameters,
+            resume_kind,
+            ..
+        } => {
+            // Pause the mission so the cron scheduler stops re-firing the
+            // same gate over and over. Until the gate resolves (user
+            // approves, provides a credential, etc.) the mission has nothing
+            // useful to do — pausing avoids the ghost-fire pattern reported
+            // in #3133 where a 3-minute cron kept re-tripping the same
+            // missing-credential gate without ever surfacing it to the user.
+            mission.status = MissionStatus::Paused;
+
+            // Friendly text the user will see in their notify_channels.
+            // Each ResumeKind has its own prompt shape; the bridge
+            // additionally fires a structured StatusUpdate (auth tray)
+            // alongside this text so the gateway UI's auth prompt opens.
+            let nudge = match &resume_kind {
+                crate::gate::ResumeKind::Authentication { instructions, .. } => format!(
+                    "Mission '{}' is paused: it needs authorization to use \
+                     '{action_name}'. Open the auth tray to grant access. \
+                     ({instructions})",
+                    mission.name
+                ),
+                crate::gate::ResumeKind::Approval { .. } => format!(
+                    "Mission '{}' is paused: approval required to run \
+                     '{action_name}'. Approve or deny in the auth tray.",
+                    mission.name
+                ),
+                crate::gate::ResumeKind::External { .. } => format!(
+                    "Mission '{}' is paused: waiting for an external \
+                     response on '{action_name}'.",
+                    mission.name
+                ),
+            };
+            mission.approach_history.push(format!(
+                "PAUSED: gate '{gate_name}' on action '{action_name}' \
+                 awaiting {kind}",
+                kind = resume_kind.kind_name()
+            ));
+            notify_response = Some(nudge);
+            gate_info = Some(MissionGateInfo {
+                gate_name: gate_name.clone(),
+                action_name: action_name.clone(),
+                parameters: parameters.clone(),
+                request_id: call_id.clone(),
+                resume_kind: resume_kind.clone(),
+            });
+        }
         _ => {}
     }
 
@@ -2342,6 +2438,7 @@ async fn process_mission_outcome_and_notify(
             notify_user: mission.notify_user.clone(),
             response,
             is_error,
+            gate: gate_info,
         };
         // Best-effort: ignore send errors (no subscribers = no problem).
         let _ = notification_tx.send(notification);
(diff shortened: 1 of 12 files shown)


src/bridge/effect_adapter.rs
```diff
@@ -2163,6 +2138,98 @@ fn parse_cadence(
...
+        return Err(EngineError::Effect {
+            reason: "mission identifier missing: provide 'name' (preferred) \
+                     or 'id' (UUID)"
+                .to_string(),
+        });
+    }
+
+    // De-duplicate while preserving order.
+    let mut seen: std::collections::HashSet<&str> = std::collections::HashSet::new();
+    name_candidates.retain(|s| seen.insert(*s));
```
serrrfirat: **Performance regression on hot path (High):** every name-keyed `mission_*` call now loads the user's full mission list. A cron routine firing every 5 min on an account with hundreds of historical missions (including completed/archived) pays an O(N) DB read per fire — the previous UUID-only path was O(1).

Suggest adding `MissionManager::find_by_name(project_id, user_id, name)` so the bridge does a targeted lookup. Even a small per-`(project_id, user_id)` name→id cache inside `MissionManager` (invalidated on create/rename/complete) would pay for itself fast — the same mission name is hit on every fire/get/pause cycle.

src/bridge/effect_adapter.rs
```diff
@@ -1053,21 +1030,19 @@ impl EffectBridgeAdapter {
...
+                        .await;
                 match id {
                     Ok(id) => {
                         let mut updates = ironclaw_engine::MissionUpdate::default();
-                        if let Some(name) = params.get("name").and_then(|v| v.as_str()) {
-                            updates.name = Some(name.to_string());
+                        // Renames use `new_name`. The legacy `name` field is
+                        // now the lookup key (consumed by `resolve_mission_id`),
+                        // not the rename target — see the schema in
+                        // `engine_actions.rs` and the migration note in the
```
serrrfirat: **Foot-gun (High):** if a caller passes `mission_update({id: <real_uuid_A>, name: "mission-B", new_name: "renamed"})`, `resolve_mission_id` resolves `id` first (UUID fast-path), so mission **A** is renamed even though `name` says `"mission-B"`. The schema text in `engine_actions.rs` says `name` *is* the lookup key — the helper logic contradicts that whenever `id` is also present.

Defense-in-depth: when both `id` (UUID) and `name` are provided and resolve to **different** missions, error out with `"both id and name provided but they identify different missions"` instead of silently preferring one. The doc comment on `resolve_mission_id` should also call out this precedence trap explicitly so the next caller doesn't trip it.

src/bridge/effect_adapter.rs
```diff
@@ -7513,6 +7580,190 @@ Use this skill to set up a Pika meeting.
...
 
+    /// Regression for #2583: `mission_fire` must accept a `name` parameter
+    /// and resolve it to the mission's id internally. Before this fix the
+    /// handler hard-required a UUID `id`, the `routine_fire` alias path
+    /// passed through the agent's `name=...` unchanged, and the resulting
+    /// "invalid mission id: invalid length 0" loop was the actual root
+    /// cause of the bug bash failure (not the budget-rework hypothesis in
+    /// the linked #2843).
+    #[tokio::test]
+    async fn mission_fire_resolves_by_name_when_id_absent() {
```
serrrfirat, line 7685: **Test coverage gap (High):** all six `mission_*` handlers (`get`, `fire`, `pause`, `resume`, `complete`, `update`) now route through `resolve_mission_id`, but only `mission_fire` got new by-name regression tests. The most foot-gun-prone handler — `mission_update` (see related comment on the `new_name` rename) — has no by-name test at all, and the existing `mission_update_string_guardrails_rejected_via_execute_action` still passes `id` directly so doesn't exercise the new path.

Per `CLAUDE.md` "Test Through the Caller, Not Just the Helper", each migrated handler needs a caller-level test. Suggest three more symmetric to this one: `mission_get_resolves_by_name`, `mission_complete_resolves_by_name`, and `mission_update_renames_via_new_name_when_name_is_lookup_key` (the last one is the regression guard for the `id+name` foot-gun).

src/bridge/effect_adapter.rs
```diff
@@ -2163,6 +2138,98 @@ fn parse_cadence(
...
+        {
+            name_candidates.push(s);
+        }
+    }
+    if let Some(s) = params
+        .get("_args")
+        .and_then(|a| a.get(0))
+        .and_then(|v| v.as_str())
+        .filter(|s| !s.is_empty())
+    {
```
serrrfirat, line 2259: **Edge case (Medium):** if a caller passes `{name: "actual-mission", _args: ["<uuid-of-other>"]}`, this `_args[0]` UUID parse short-circuits with `return Ok(MissionId(uuid))` — the explicit `name` field is silently ignored. This contradicts the docstring's stated principle that `name` is preferred and `_args[0]` is a "positional fallback used by some Tier-0 tool-call shapes".

Two ways out:
- (a) **Preserve preference**: move the `_args[0]` UUID parse to *after* `name_candidates` lookup falls through, so explicit `name` wins.
- (b) **Treat positional as always-a-name**: drop the `_args[0]` UUID-parse branch and only push to `name_candidates`. UUID-shaped strings would still match if any mission happens to be named with a UUID — vanishingly rare and an explicit caller decision.

Either is fine; the current ordering is the worst of both.

src/bridge/effect_adapter.rs
```diff
@@ -2163,6 +2138,98 @@ fn parse_cadence(
...
+    name_candidates.retain(|s| seen.insert(*s));
+
+    let missions = mgr.list_missions(project_id, user_id).await?;
+    for candidate in &name_candidates {
+        if let Some(m) = missions.iter().find(|m| m.name == *candidate) {
+            return Ok(m.id);
+        }
+    }
+
+    Err(EngineError::Effect {
```
serrrfirat: **Information disclosure (Medium):** echoing `user_id` into the error reason leaks server-side identity into the LLM context, conversation history, and any tool-error telemetry. The previous error (`"mission not found: {id_str}"`) only echoed user input.

`EngineError::Effect::reason` flows back to the LLM and is persisted — depending on the channel, `user_id` is a UUID (low value), but for some flows it's an external identity (chat id, session-derived token). Suggest dropping the ` for user '{user_id}'` segment entirely. The LLM can't act on it anyway (it can't list a different user's missions).

Same goes for the `mission {} belongs to another user` reason a few lines above (mission_get arm at the new resolve site) — there's no LLM-actionable info in the foreign mission's UUID, and emitting it confirms the existence of a mission the caller has no claim to.

src/bridge/effect_adapter.rs
```diff
@@ -2163,6 +2138,98 @@ fn parse_cadence(
...
+/// Resolution order (first match wins):
+///   1. `params.id` — if present and a valid UUID, use directly.
+///   2. `params.name` — look up `(project_id, user_id, name)` via
+///      `MissionManager::list_missions`. The same identifier used in
+///      `mission_create` is what we resolve back here.
+///   3. `params._args[0]` — positional fallback used by some Tier-0
+///      tool-call shapes; tried as UUID first, then as name.
+///
+/// A non-UUID `id` value is treated as a name (some callers conflate the
+/// two), which keeps backward compat for `mission_complete`'s historical
```
serrrfirat: **Architecture (Medium):** name-or-id resolution is conceptually a `MissionManager` concern, not a bridge-adapter concern. Keeping this 60-line policy here means any future caller — CLI mission commands, the WS handler, a WASM channel that wants a mission lookup, integration tests outside the bridge — duplicates the resolution rules.

Suggest moving to a typed API on `MissionManager`:

```rust
enum MissionRef<'a> {
    Id(MissionId),
    Name(&'a str),
}

impl MissionManager {
    pub async fn resolve(&self, project_id: ProjectId, user_id: &str, r: MissionRef<'_>)
        -> Result<MissionId, EngineError> { ... }
}
```

The bridge then becomes a one-liner that picks the right variant from `params`. The two ambiguity rules (`id`+`name` simultaneous; `_args[0]`) collapse into one well-scoped place that has a unit test suite of its own.

tests/fixtures/llm_traces/live/routine_btc_create_test_and_refire.json
```diff
@@ -0,0 +1,941 @@
...
+          [
+            "content-security-policy-report-only",
+            "script-src https://accounts.google.com/gsi/client; frame-src https://accounts.google.com/gsi/; connect-src https://accounts.google.com/gsi/;"
+          ],
+          [
+            "cf-cache-status",
+            "HIT"
+          ],
+          [
+            "cf-ray",
```
serrrfirat, line 97: **Trace fixture noise (Medium):** the recorded CoinGecko response carries every header from the real Cloudflare/Coingecko response — `cf-ray`, `x-request-id`, `etag`, `age`, `date`, `cf-cache-status`, `strict-transport-security`, `content-security-policy-report-only`, etc. They aren't secrets, but they:

- bloat the diff (~941 lines for one e2e test, most of it duplicated headers across three identical price requests),
- pin the fixture to one Cloudflare datacenter (`-FUK`) and a 2026-05-01 timestamp,
- risk replay surprises if any HTTP cache layer in the stack decides to honor `etag`/`age`.

Suggest a fixture-redaction helper in `tests/support/` (e.g. `redact_response_headers(allow=["content-type", "content-length"])`) applied at fixture-write time. Net result is a smaller, more reproducible fixture and one less drift source for replay-mode tests.

ilblackdragon:
Half 2 follow-up filed as #3166: 'Mission auto-resume after auth/approval gate resolution'. That issue covers persisting `Mission.paused_gate`, hooking the credential-write and gate-resolve paths to auto-resume matching paused missions, and the immediate-fire-on-resume behaviour. Out of scope for this PR.

ilblackdragon:
Pushed b74c42e addressing the review feedback. Summary by reviewer:

**serrrfirat:**
- ✅ Perf regression — `MissionManager::find_by_name` typed API; bridge resolver stops scanning `list_missions`. Doc-comment names the future Store-level index drop-in.
- ✅ id+name foot-gun — new conflict guard errors loudly with both ids surfaced. Regression test `mission_resolver_errors_when_id_and_name_disagree`.
- ✅ `_args[0]` UUID precedence inversion — positional UUIDs no longer short-circuit ahead of explicit `name`.
- ✅ user_id info disclosure — dropped from EngineError reason strings on both `mission not found` and `mission_get` ownership arms; preserved in tracing::debug!.
- ✅ Test gap on migrated handlers — added by-name caller-level tests for `mission_get`, `mission_complete`, `mission_pause`+`mission_resume`.

**Gemini:**
- ✅ Legacy `{id, name}` rename preservation — landed earlier in commit 0bd586df with two regression tests; the new mission_update handler strips `name` from the resolver-view params when it detects the legacy shape, so the conflict guard doesn't reject it.
- ✅ Docstring/behaviour drift on `resolve_mission_id` — the docstring is rewritten to match the new four-step resolution order including the conflict guard.

**Copilot:**
- ✅ `ApprovalAutoResponder` task leak — `Drop` impl aborts the handle on panic.
- ✅ Stale "unreachable Ok(None)" comment in `mission_get` — corrected to call out the UUID-with-no-mission case.

**Deferred:**
- Trace fixture noise (Cloudflare headers) — filed for a small followup; the fixture is committed to enable replay-mode CI runs and stripping requires care to keep replay deterministic.
- Move resolver to MissionManager (architecture) — the typed API is now there (find_by_name); the param-parsing stays bridge-side because it deals with JSON/Tier-0 conventions.

Test results on b74c42e: 448 bridge tests + 89 engine mission tests, all green. fmt + clippy -D warnings clean.

src/bridge/router.rs
```diff
@@ -4128,19 +4134,92 @@ fn interpret_message_event(role: &str, content_preview: &str) -> Option<&'static
...
+                        request_id: gate.request_id.clone(),
+                        tool_name: gate.action_name.clone(),
+                        description: format!(
+                            "Mission '{}' is waiting for approval to run '{}'.",
+                            notif.mission_name, gate.action_name
+                        ),
+                        parameters: gate.parameters.clone(),
+                        allow_always: *allow_always,
+                    }
+                }
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in 720a0db5. `MissionGateInfo.request_id` renamed to `call_id` (matches what it actually carries — the engine LLM tool-call id). New `MissionGateInfo.gate_request_id: Uuid` is freshly generated at gate-emit time, and the bridge now forwards it (not `call_id`) into `StatusUpdate::AuthRequired.request_id` / `ApprovalNeeded.request_id`. The gateway resolve POST (`features/chat/mod.rs:183/274/303` + `platform/ws.rs:236`) now sees a UUID-shaped string and won't 400. Mapping the gateway-issued UUID back to the engine `call_id` for actual auto-resume is half-2 (#3166).

crates/ironclaw_engine/src/runtime/mission.rs
```diff
@@ -101,6 +101,45 @@ pub struct MissionNotification {
...
+    pub action_name: String,
+    /// Original action parameters at the time of the pause. The bridge
+    /// needs these to run the canonical extension-name resolver
+    /// (`AuthManager::resolve_extension_name_for_auth_flow`) — for
+    /// `tool_activate(name="gmail")` the resolver reads the explicit
+    /// `name` param, which has higher precedence than the credential
+    /// fallback.
+    pub parameters: serde_json::Value,
+    /// Engine-side `call_id` used to resume.
+    pub request_id: String,
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in 720a0db5 — see reply on the sibling comment. Field renamed to `call_id`; new `gate_request_id: Uuid` is the user-facing identity. Doc-comment on the struct now explicitly contrasts the two fields and points at #3166.

src/bridge/engine_actions.rs
```diff
@@ -149,13 +150,14 @@ pub(crate) fn mission_capability_actions() -> Vec<ActionDef> {
...
                     "success_criteria": {"type": "string", "description": "Completion criteria"}
-                },
-                "required": ["id"]
+                }
             }),
             Some(action_discovery_summary(
-                &["id"],
+                &[],
                 &[
+                    "Identify the mission to update with `name` (preferred) or `id` — exactly one is required.",
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in 720a0db5. Discovery summary now matches the actual handler tolerance: "Identify the mission with `name` (preferred) or `id` (UUID). If both are provided they must identify the same mission, or use the legacy `{id, name}` rename shape (where `name` is the new name) — otherwise the resolver errors with 'identify different missions'."

tests/e2e_live_routine.rs
```diff
@@ -0,0 +1,596 @@
...
+
+        async fn approved_tools(&self) -> Vec<(String, Uuid)> {
+            self.approved.lock().await.clone()
+        }
+
+        fn shutdown(self) {
+            self.handle.abort();
+        }
+    }
+
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in b74c42e. `ApprovalAutoResponder` now implements `Drop` that aborts the handle. `shutdown()` is kept for explicitness; `abort()` is idempotent so the double-abort path is safe.

src/bridge/effect_adapter.rs
```diff
@@ -975,24 +970,21 @@ impl EffectBridgeAdapter {
                             }))
                         }
                         Ok(None) => Err(EngineError::Effect {
-                            reason: format!("mission not found: {id_str}"),
+                            // We resolved the id via name lookup, so this
+                            // should be unreachable in practice (the lookup
+                            // already proved the mission exists). Keep the
+                            // arm so the match is exhaustive.
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in b74c42e. The comment now correctly notes that `Ok(None)` is reachable when the caller passed an explicit UUID for a deleted/unknown mission — names that didn't match are caught earlier by `resolve_mission_id` with a more useful error.

src/bridge/effect_adapter.rs
```diff
@@ -2163,6 +2138,98 @@ fn parse_cadence(
...
+    let missions = mgr.list_missions(project_id, user_id).await?;
+    for candidate in &name_candidates {
+        if let Some(m) = missions.iter().find(|m| m.name == *candidate) {
+            return Ok(m.id);
+        }
+    }
+
+    Err(EngineError::Effect {
+        reason: format!(
+            "mission not found by name: tried {name_candidates:?} for user '{user_id}'. \
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in b74c42e. `user_id` is no longer included in the `mission not found` error. Same fix on `mission_get`'s "belongs to another user" arm — the foreign mission's UUID is now scrubbed from the LLM-visible reason and preserved only via `tracing::debug!`.

src/bridge/effect_adapter.rs
```diff
@@ -1053,21 +1030,19 @@ impl EffectBridgeAdapter {
...
-                        if let Some(name) = params.get("name").and_then(|v| v.as_str()) {
-                            updates.name = Some(name.to_string());
+                        // Renames use `new_name`. The legacy `name` field is
+                        // now the lookup key (consumed by `resolve_mission_id`),
+                        // not the rename target — see the schema in
+                        // `engine_actions.rs` and the migration note in the
+                        // helper docs above.
+                        if let Some(new_name) = params.get("new_name").and_then(|v| v.as_str()) {
+                            updates.name = Some(new_name.to_string());
                         }
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in 0bd586d (initial preservation) + b74c42e (mission_update handler now strips `name` from the resolver-view params when the legacy shape is detected, so the new conflict guard doesn't reject it). Two regression tests pin it: `mission_update_preserves_legacy_id_plus_name_rename` and `mission_update_renames_via_new_name_and_uses_name_as_lookup`.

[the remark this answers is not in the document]
ilblackdragon: Fixed in 0bd586d (preserves the legacy shape) + b74c42e (handler strips `name` from the resolver-view params for the legacy shape so the new conflict guard doesn't reject it). Two regression tests guard it.

src/bridge/effect_adapter.rs
```diff
@@ -2163,6 +2138,98 @@ fn parse_cadence(
...
+/// with a `name` field that the handler never read, then the handler
+/// parsed an empty `id` as UUID and rejected with "invalid length 0".
+///
+/// Resolution order (first match wins):
+///   1. `params.id` — if present and a valid UUID, use directly.
+///   2. `params.name` — look up `(project_id, user_id, name)` via
+///      `MissionManager::list_missions`. The same identifier used in
+///      `mission_create` is what we resolve back here.
+///   3. `params._args[0]` — positional fallback used by some Tier-0
+///      tool-call shapes; tried as UUID first, then as name.
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in b74c42e. The docstring on `resolve_mission_id` now matches the new four-step resolution order including the conflict guard. `_args[0]` no longer short-circuits ahead of explicit `name` (it's now a name candidate only — see serrrfirat's related comment).

src/bridge/effect_adapter.rs
ilblackdragon: Addressed in b74c42e. Added `MissionManager::find_by_name(project_id, user_id, name) -> Result<Option<Mission>, EngineError>` typed API; bridge resolver switched off `list_missions` onto it. Today `find_by_name` still scans the user's missions in-memory; the doc-comment names the future store-level index drop-in (`Store::find_mission_by_name`) explicitly. Adding it later is non-breaking — every caller goes through this method now, so the optimization plugs in without churn at the call sites.

src/bridge/effect_adapter.rs
ilblackdragon: Fixed in b74c42e. Added the conflict guard you suggested. New regression test `mission_resolver_errors_when_id_and_name_disagree` pins it. Error reads: "both id and name provided but they identify different missions: id=X vs name=[Y] → Z. Provide only one — or fix the name to match the id." The resolver docstring also explicitly calls out the conflict-guard contract.

src/bridge/effect_adapter.rs
ilblackdragon: Fixed in b74c42e. Added by-name caller-level tests for every migrated handler: `mission_get_resolves_by_name`, `mission_complete_resolves_by_name`, `mission_pause_and_resume_resolve_by_name`, plus the conflict-guard regression `mission_resolver_errors_when_id_and_name_disagree` and the legacy-rename regression `mission_update_preserves_legacy_id_plus_name_rename`.

src/bridge/effect_adapter.rs
ilblackdragon: Fixed in b74c42e — went with option (b). `_args[0]` is now treated as a name candidate only, never honoured as a UUID. UUID-shaped strings would only match if a mission happens to be named with one (vanishingly rare and the explicit caller decision you noted).

src/bridge/effect_adapter.rs
ilblackdragon: Fixed in b74c42e. Both error reasons are scrubbed: `user_id` is gone from the "mission not found" error, and `mission_get`'s "belongs to another user" reason no longer echoes the foreign mission's UUID (preserved only via `tracing::debug!`).

src/bridge/effect_adapter.rs
ilblackdragon: Partially addressed in b74c42e. The typed engine API (`MissionManager::find_by_name`) is in place — that's the part future callers (CLI, WS, WASM channels, integration tests) will actually need. The 60-line param-parsing in the bridge stayed put because it deals in JSON and Tier-0 calling conventions (`_args[0]` etc.) that aren't a `MissionManager` concern. Happy to lift the param-shape policy onto a `MissionRef` enum + `MissionManager::resolve` in a follow-up if you'd prefer the typed-API form — should I file that?

tests/fixtures/llm_traces/live/routine_btc_create_test_and_refire.json
ilblackdragon: Deferred — agreed on the diagnosis. The fixture is committed to enable replay-mode CI runs, and a redaction helper that's careful to keep replay deterministic is its own small piece of work. Will file a follow-up issue today if you want; otherwise happy to tackle in a tiny PR right after this one merges.

ilblackdragon:
Pushed 720a0db5 closing the last review-feedback items. Full status:

**Just landed (720a0db5):**
- Copilot's UUID/call_id mismatch (high) — `MissionGateInfo.request_id` renamed to `call_id`; new `gate_request_id: Uuid` is generated at gate-emit time and forwarded to `StatusUpdate.request_id`. Gateway resolve POST (`Uuid::parse_str` at `features/chat/mod.rs:183`/`platform/ws.rs:236`) now accepts the auth-tray entries instead of 400-ing.
- Schema wording vs handler tolerance — `mission_update` discovery summary rewritten to match the conflict-guard contract.

**Already landed earlier on this branch:**
- b74c42e: `find_by_name` typed API + bridge perf fix; id+name conflict guard; `_args[0]` precedence flip; `user_id` info-disclosure scrub on two error paths; `Drop` on `ApprovalAutoResponder`; stale `Ok(None)` comment; by-name regression tests for `mission_get/complete/pause/resume` + conflict-guard regression.
- 0bd586d: legacy `{id, name}` rename preservation + two regression tests.

**Deferred** (each individually replied on the inline thread):
- Trace fixture noise (Cloudflare headers) — happy to file a follow-up issue or land in a tiny PR right after this merges; awaiting your preference.
- Full `MissionRef` enum + `MissionManager::resolve` typed API — typed engine surface is now in place via `find_by_name`; lifting the bridge's JSON-param parsing into a typed enum is a clean follow-up if you want it.

Test results on 720a0db5: 19/19 mission tests in `bridge::effect_adapter`, 89/89 `runtime::mission`, fmt + clippy clean across both crates.

PR #3157, opened 2026-05-01
fix(engine): inline gate await for Tier 0 + Tier 1 Approval gates

ilblackdragon:
## Summary

CodeAct scripts that hit a tool requiring approval were surfacing as `RuntimeError: execution paused by gate 'approval'` *inside the script* instead of pausing for the user. The async tool-resolve path (`scripting.rs::resolve_tool_future`) converted `EngineError::GatePaused` into a Python exception. The sync preflight path returned `need_approval` to the orchestrator, which on resume re-ran the LLM step and re-executed any non-idempotent earlier tool calls in the same script.

Replace both paths with a host-supplied `GateController` that pauses the live execution in place. The Monty VM (Tier 1) and the Tier 0 batch loop stay alive across the user's approval; on `Approved` the gated action re-executes; on `Denied` the script raises a typed `RuntimeError("user denied tool 'X': ...")` that the script can catch.

Mechanism details:

- `gate_controller` is **required** on `ThreadExecutionContext` (`Arc<dyn GateController>`, not `Option`). Paths that don't pause supply `CancellingGateController::arc()` — gates surface as typed denials, not the legacy bug message.
- **One-shot approval propagation:** `call_approval_granted: bool` on `ThreadExecutionContext` carries the user's approval through to the retry call. `EffectBridgeAdapter` forwards it as `approval_already_granted=true` so `ApprovalRequirement::Always` tools and `always=false` AskEachTime gates don't re-fire after approval.
- **Bounded retry** (`MAX_INLINE_GATE_RETRIES = 3`) shared by Tier 0 and Tier 1.
- **Bounded await:** `BridgeGateController::pause` races the resolution oneshot against `pending.expires_at`. On expiry the pending row is discarded, the registry entry dropped, and the VM unwinds with `Cancelled`.
- **Pre-bound execution context:** the bridge calls `set_pre_execution_context` (per-user) BEFORE `handle_user_message` and promotes to `(user, thread)`-keyed once thread_id is known. Closes the race where a fast tool gate could reach `pause()` before the entry existed.
- **Tier 0 parallel-batch coverage:** both single-runnable and multi-runnable branches of `handle_execute_actions_parallel` go through `execute_single_action_with_inline_retry`. `execute_single_action_with_inline_retry` returns `Vec<EventKind>` so `ApprovalRequested` + post-retry events both reach the thread event log.
- Auth and External resume kinds keep the legacy thread re-entry path — their resolution installs new state.
- Boot sweep invalidates `Approval`-kind `PendingGate` rows from a prior process.

Design: [`docs/plans/2026-05-01-codeact-inline-gate-await.md`](docs/plans/2026-05-01-codeact-inline-gate-await.md).

## Quality gate

| Check | Result |
|---|---|
| `cargo fmt` | clean |
| `cargo clippy --all --benches --tests --examples --all-features` | clean |
| `cargo test -p ironclaw_engine --lib` | **513 pass**; 2 pre-existing failures on staging (`call_id_preserved_when_no_lease`, `stop_thread_works`) reproduce without this change |
| `cargo test --lib bridge::` | 443 pass |
| `cargo test --test engine_v2_gate_integration` | 29 pass |

## Test plan

- [x] Unit: `codeact_gate_inline_await_approved_delivers_result` reproduces the user-reported shape and asserts the post-approval retry observes `call_approval_granted=true` (regression coverage for one-shot approval propagation).
- [x] Unit: `codeact_gate_inline_await_denied_raises_in_script` — denial surfaces as a typed `RuntimeError` with a message identifying the tool and the user-supplied reason.
- [x] Unit: `codeact_default_controller_cancels_approval_gates` — locks in the `CancellingGateController` default.
- [x] Unit: `bridge::gate_controller::tests::*` — 4 tests cover the `GateResolutions` registry semantics.
- [x] Unit: `invalidate_stranded_approval_gates_evicts_only_approval_kind` — boot sweep evicts only `Approval` rows.
- [x] Integration: `approval_gate_resolves_inline_via_controller` drives a Tier 0 mid-execution `Approval` gate through the controller, asserts inline resolution + completion. Companion `approval_denied_inline_completes_thread_with_failed_action` asserts denial via the default controller.
- [x] Integration: `approval_chains_directly_into_auth_for_install_flow` — `Approval` resolved inline by controller, chained `Authentication` gate bubbles up via the legacy path (intact post-PR), legacy auth-resume drives completion.
- [x] Live regression: `codeact_inline_gate_await_resumes_user_reproducer` drives the user-reported reproducer through `ThreadManager`. `codeact_inline_gate_await_denial_does_not_retry` covers denial.
- [ ] **E2E (manual)**: trigger a real `github_create_issue`-style approval inside CodeAct, click approve in the UI, verify the script continues without re-executing earlier non-idempotent calls.
- [ ] **E2E (manual)**: same flow with deny.
- [ ] **E2E (manual)**: kill IronClaw mid-approval, restart — verify the user gets the `GateExpired` SSE.
- [ ] **E2E (manual)**: ignore the approval prompt past 30-min expiry — verify the VM unwinds cleanly.

## Follow-ups (not in this PR)

- **Stop signal integration** — engine signal channel doesn't currently interrupt during `execute_action`; the inline-await pause inherits that limit.
- **Caller-level integration test through real `EffectBridgeAdapter`** for `ApprovalRequirement::Always` and expiry/late-approve scenarios.
- **Bridge integration test for `resolve_gate` rollback** — the auto-approve install/rollback path deserves a caller-level test.
- **Active-CPU vs paused-clock split** for Monty's `max_duration` (currently 30 s and ticks during gate awaits).
- **Migration of Auth/External resume kinds** to the same controller mechanism.
- **Per-user concurrency cap on in-flight gates.**
- **`mission.rs::dispatch_protected_write` controller wiring** — currently `CancellingGateController`.
- Remove the no-op stubs left in `engine_v2_gate_integration.rs` (`approval_resolution_executes_pending_call_directly_via_resolved_pending_action`, `gate_resume_with_execution_obligation`) once the post-PR equivalents are deemed sufficient coverage.


crates/ironclaw_engine/src/executor/scripting.rs
```diff
@@ -839,16 +850,247 @@ pub async fn execute_code_with_skills(
...
+                                    });
+                                }
+                            }
+                            continue;
+                        }
+
+                        // Approved. Re-do preflight — the bridge installed
+                        // any auto-approve preference before delivering the
+                        // resolution, so policy now returns Allow.
+                        let retry_preflight = preflight_action(
```
serrrfirat, line 1061: Blocking: this inline retry does not carry the user's one-shot approval into the real bridge effect policy. `resolve_gate` only installs auto-approve when `always=true`; for normal approval (`always=false`) the retry goes through `EffectExecutor::execute_action`, whose `EffectBridgeAdapter` impl calls `execute_action_internal(..., approval_already_granted=false)`. That means ask-each-time gates can re-prompt immediately after approval, and `ApprovalRequirement::Always` gates can never proceed because `allow_always=false` prevents session auto-approve. The legacy path uses `execute_resolved_pending_action(..., approval_already_granted=true)` for this case. Please pass a one-shot approved marker/token through the inline retry paths (Tier 1 and Tier 0) instead of relying only on session auto-approve.

serrrfirat (commented):
Posting the confirmed high/medium-or-above paranoid review findings. Low-severity notes were intentionally not posted.

src/bridge/gate_controller.rs
```diff
@@ -0,0 +1,439 @@
...
+            return GateResolution::Cancelled;
+        }
+
+        let (tx, rx) = oneshot::channel();
+        self.resolutions.register(request_id, tx).await;
+
+        self.emit_gate_prompt(&pending, &per_exec.channel_metadata)
+            .await;
+
+        match rx.await {
```
serrrfirat: Blocking: this await can park the engine forever. There is no select on gate expiry, stop/cancel, shutdown, or timeout, and `resolve_gate` rejects expired gates before attempting `try_deliver`. If a user ignores the approval past expiry, the pending row expires but the oneshot remains pending and the VM/thread keeps running. The router's 5-minute timeout then calls `join_thread`, which awaits the still-paused task. Please make this wait bounded and wake the VM with `GateResolution::Cancelled` on expiry/stop/shutdown, while also removing the pending gate.

src/bridge/router.rs
```diff
@@ -3569,6 +3738,31 @@ async fn handle_with_engine_inner(
...
+    // If this turn surfaces an `Approval` gate, the controller looks
+    // this up to build the `PendingGate`. Cleared after `await_thread_outcome`
+    // returns. Live gates outlive the await (they get consumed by the
+    // resolve endpoint), but the per-execution registry entry can be
+    // dropped — the `PendingGate` row holds everything the resolver
+    // needs.
+    let scope_thread_id = message
+        .conversation_scope()
+        .and_then(|s| ironclaw_common::ExternalThreadId::new(s).ok());
+    state
```
serrrfirat, line 3800: Blocking/racy: the per-execution context is registered after `handle_user_message().await`, but that call already spawns the execution task. A fast tool gate can reach `BridgeGateController::pause` before this map entry exists; `pause` then returns `Cancelled` and no prompt is shown. This is the same metadata-after-spawn race the engine code already avoids elsewhere. Please bind this context before execution can start (for example by passing it into the spawn/start path or adding a pre-start hook) and cover it with a caller-level test.

crates/ironclaw_engine/src/executor/structured.rs
```diff
@@ -522,6 +596,109 @@ fn interrupted_call_needs_refund(result: &Result<ActionResult, EngineError>) ->
     matches!(result, Err(EngineError::GatePaused { .. }))
```
serrrfirat, line 591: Medium: this helper only covers the single-call fast path. The multiple-runnable branch still spawns each call with direct `effects.execute_action(...)`, so a Tier 0 parallel mid-execution `GatePaused(Approval)` still goes through the legacy interrupt/re-entry path. That leaves the double-execution behavior for parallel batches despite the PR title/design saying Tier 0 + Tier 1 approval gates. Either route the parallel branch through the same inline retry mechanism or narrow the PR/docs/title and add a follow-up test.

crates/ironclaw_engine/src/executor/scripting.rs
```diff
@@ -3567,6 +3958,315 @@ result = await tool_info(name="mission-create", detail="schema")
...
+            &self,
+            name: &str,
+            params: serde_json::Value,
+            _lease: &CapabilityLease,
+            _ctx: &ThreadExecutionContext,
+        ) -> Result<ActionResult, EngineError> {
+            let mut count = self.call_count.lock().unwrap();
+            *count += 1;
+            if *count == 1 && name == self.action {
+                return Err(EngineError::GatePaused {
```
serrrfirat, line 4223: Medium coverage gap: this stub returns success on the second call regardless of whether inline approval was actually conveyed to bridge/effect policy, so it cannot catch the one-shot approval bug. Please add caller-level coverage through `resolve_gate`/`handle_with_engine` plus the real `EffectBridgeAdapter` for `Approved { always: false }`, `ApprovalRequirement::Always`, expiry/late approve, and stop-during-wait.

src/bridge/gate_controller.rs
```diff
@@ -0,0 +1,439 @@
...
+            Ok(resolution) => resolution,
+            Err(_) => {
+                // Sender dropped — process shutting down or registry
+                // cleared. Treat as cancellation.
+                self.resolutions.forget(request_id).await;
+                GateResolution::Cancelled
+            }
+        }
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in 78a12766. `BridgeGateController::pause` now races the resolution oneshot against `tokio::time::sleep` aligned to `pending.expires_at`; on expiry we discard the pending row, drop the registry entry, and return `Cancelled` so the VM unwinds cleanly. Stop signal integration is the next layer (engine signal channel does not currently interrupt during `execute_action` either) and is on the follow-ups list.

crates/ironclaw_engine/src/executor/scripting.rs
ilblackdragon: Fixed in 78a12766. Added `call_approval_granted: bool` on `ThreadExecutionContext` (default false). Inline retry paths (`drive_inline_gate`, `execute_with_inline_gate_retry`, `execute_single_action_with_inline_retry`, scripting sync preflight) set it true on the retry; `EffectBridgeAdapter::execute_action` forwards it as `approval_already_granted=true`. Mirrors the legacy `execute_resolved_pending_action` contract. Regression coverage: `codeact_gate_inline_await_approved_delivers_result` now asserts `approval_flags == [false, true]`.

src/bridge/router.rs
ilblackdragon: Fixed in 78a12766. The bridge now calls `set_pre_execution_context` (per-user) BEFORE `handle_user_message`, then promotes to (user, thread)-keyed once thread_id is known. `pause()` falls back to the per-user entry on miss — closes the race window where a fast tool gate reached `pause()` before the (user, thread) entry existed.

crates/ironclaw_engine/src/executor/structured.rs
ilblackdragon: Fixed in 78a12766. New `execute_single_action_with_inline_retry` wraps `execute_single_action` with the same bounded retry shape (`MAX_INLINE_GATE_RETRIES = 3`) used by Tier 1. Both the single-runnable and multi-runnable branches of `handle_execute_actions_parallel` go through it, so simultaneous gates in a parallel batch no longer fall through to the legacy re-entry path.

crates/ironclaw_engine/src/executor/scripting.rs
ilblackdragon: Partial fix in 78a12766: `GatingThenOkEffects` now records `context.call_approval_granted` per call, and `codeact_gate_inline_await_approved_delivers_result` asserts `[false, true]` — the retry observes the user's one-shot approval. Caller-level coverage through real `EffectBridgeAdapter` (`ApprovalRequirement::Always`, expiry, stop-during-wait) is on the follow-ups list — the engine-side regression for the propagation contract is in place.

crates/ironclaw_engine/src/executor/scripting.rs
```diff
@@ -1740,30 +1989,171 @@ async fn resolve_tool_future(
...
+                        }
+                        Err(e) => {
+                            events.push(EventKind::ActionFailed {
+                                step_id: context.step_id,
+                                action_name: gate_action_name.clone(),
+                                call_id: gate_call_id.clone(),
+                                error: e.to_string(),
+                                duration_ms: retry_duration_ms,
+                                params_summary,
+                            });
```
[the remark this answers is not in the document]
ilblackdragon: False positive — `drive_inline_gate` already refunds the new lease in the GatePaused-on-retry arm (refund at line ~2090 of the new layout). The retry's GatePaused arm pushes an `ApprovalRequested` event, refunds via `leases.refund_use(lease.id)`, and re-enters the bounded loop. No leak.

src/bridge/router.rs
```diff
@@ -1915,11 +1942,45 @@ pub async fn init_engine(agent: &Agent) -> Result<(), Error> {
...
+            continue;
+        }
+        let _ = pending_gates.discard(&gate.key()).await;
+        if let Some(sse) = sse {
+            sse.broadcast_for_user(
+                &gate.user_id,
+                ironclaw_common::AppEvent::GateResolved {
+                    request_id: gate.request_id.to_string(),
+                    gate_name: gate.gate_name.clone(),
+                    tool_name: gate.action_name.clone(),
```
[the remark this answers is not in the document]
ilblackdragon: The startup sweep already carries `// projection-exempt: bridge dispatcher, restart-time gate cleanup before threads exist` on the same line as the broadcast — see router.rs:1986 in 78a12766.

src/bridge/router.rs
```diff
@@ -1915,11 +1942,45 @@ pub async fn init_engine(agent: &Agent) -> Result<(), Error> {
...
+    for gate in restored {
+        if !matches!(
+            gate.resume_kind,
+            ironclaw_engine::ResumeKind::Approval { .. }
+        ) {
+            continue;
+        }
+        let _ = pending_gates.discard(&gate.key()).await;
+        if let Some(sse) = sse {
+            sse.broadcast_for_user(
```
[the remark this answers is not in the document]
ilblackdragon: Same as above — annotation present at router.rs:1986.

src/bridge/router.rs
```diff
@@ -2406,6 +2467,114 @@ pub async fn resolve_gate(
...
+                sse.broadcast_for_user(
+                    &message.user_id,
+                    AppEvent::GateResolved {
+                        request_id: pending.request_id.to_string(),
+                        gate_name: pending.gate_name.clone(),
+                        tool_name: pending.action_name.clone(),
+                        resolution: label.into(),
+                        message: status_msg.into(),
+                        thread_id: Some(pending.effective_wire_thread_id()),
+                    },
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in 78a12766. Added `// projection-exempt: bridge dispatcher, inline-await fast-path resolution event` on the same line as the broadcast.

src/bridge/router.rs
```diff
@@ -2406,6 +2467,114 @@ pub async fn resolve_gate(
...
+                            ("approved_always", "Gate approved. Resuming execution.")
+                        } else {
+                            ("approved", "Gate approved. Resuming execution.")
+                        }
+                    }
+                    ironclaw_engine::GateResolution::Denied { .. } => ("denied", "Gate denied."),
+                    ironclaw_engine::GateResolution::Cancelled => ("cancelled", "Gate cancelled."),
+                    _ => unreachable!(),
+                };
+                sse.broadcast_for_user(
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in 78a12766. Same site as #3173024819.

src/bridge/gate_controller.rs
```diff
@@ -0,0 +1,439 @@
...
+            self.extension_manager.as_deref(),
+            self.tools.as_ref(),
+            pending,
+        )
+        .await;
+
+        let display_parameters = crate::bridge::router::gate_display_parameters(pending);
+
+        if let Some(ref sse) = self.sse {
+            sse.broadcast_for_user(
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in 78a12766. Added `// projection-exempt: bridge dispatcher, inline-await gate prompt for live VM waiting on user input` on the same line.

crates/ironclaw_engine/src/executor/scripting.rs
```diff
@@ -839,16 +850,247 @@ pub async fn execute_code_with_skills(
...
+
+                        if let Some(reason) = denial_reason {
+                            // Resume Monty with a typed exception. RuntimeError
+                            // is what we emit; the message is explicit so users
+                            // (and the LLM) can distinguish denial from other
+                            // runtime errors.
+                            let ext_result = ExtFunctionResult::Error(MontyException::new(
+                                ExcType::RuntimeError,
+                                Some(format!("user denied tool '{gate_action_name}': {reason}")),
+                            ));
```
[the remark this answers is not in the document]
ilblackdragon: The design doc was updated to record this divergence: shipped code uses `RuntimeError` with a descriptive message (`\"user denied tool 'X': <reason>\"`) rather than `PermissionError`. Two reasons: (a) Python scripts already catch `RuntimeError` and the message text is specific enough to distinguish from other runtime errors, (b) introducing `PermissionError` here would force every CodeAct script that catches gate denial to update its except clause, breaking out-of-tree examples. The Testing section of `docs/plans/2026-05-01-codeact-inline-gate-await.md` now documents this choice. Not changing in this PR.

crates/ironclaw_engine/src/executor/scripting.rs
```diff
@@ -839,16 +850,247 @@ pub async fn execute_code_with_skills(
...
+                            PreflightResult::GatePaused(_) => {
+                                // Policy still says approval needed even
+                                // after user said yes. Treat as denial so
+                                // we don't loop forever.
+                                let ext_result = ExtFunctionResult::Error(MontyException::new(
+                                    ExcType::RuntimeError,
+                                    Some(format!(
+                                        "tool '{gate_action_name}' still requires approval after resolution"
+                                    )),
+                                ));
```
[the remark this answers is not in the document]
ilblackdragon: The design doc was updated to record this divergence: shipped code uses `RuntimeError` with a descriptive message (`\"user denied tool 'X': <reason>\"`) rather than `PermissionError`. Two reasons: (a) Python scripts already catch `RuntimeError` and the message text is specific enough to distinguish from other runtime errors, (b) introducing `PermissionError` here would force every CodeAct script that catches gate denial to update its except clause, breaking out-of-tree examples. The Testing section of `docs/plans/2026-05-01-codeact-inline-gate-await.md` now documents this choice. Not changing in this PR.

crates/ironclaw_engine/src/executor/scripting.rs
```diff
@@ -1740,30 +1989,171 @@ async fn resolve_tool_future(
...
+                        action_name: gate_action_name.clone(),
+                        call_id: gate_call_id.clone(),
+                        error: format!("denied: {reason}"),
+                        duration_ms: 0,
+                        params_summary,
+                    });
+                    ExtFunctionResult::Error(MontyException::new(
+                        ExcType::RuntimeError,
+                        Some(format!("user denied tool '{gate_action_name}': {reason}")),
+                    ))
```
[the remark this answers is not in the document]
ilblackdragon: The design doc was updated to record this divergence: shipped code uses `RuntimeError` with a descriptive message (`\"user denied tool 'X': <reason>\"`) rather than `PermissionError`. Two reasons: (a) Python scripts already catch `RuntimeError` and the message text is specific enough to distinguish from other runtime errors, (b) introducing `PermissionError` here would force every CodeAct script that catches gate denial to update its except clause, breaking out-of-tree examples. The Testing section of `docs/plans/2026-05-01-codeact-inline-gate-await.md` now documents this choice. Not changing in this PR.

crates/ironclaw_engine/src/executor/scripting.rs
```diff
@@ -1740,30 +1989,171 @@ async fn resolve_tool_future(
...
+                        action_name: gate_action_name.clone(),
+                        call_id: gate_call_id.clone(),
+                        error: "cancelled".into(),
+                        duration_ms: 0,
+                        params_summary,
+                    });
+                    ExtFunctionResult::Error(MontyException::new(
+                        ExcType::RuntimeError,
+                        Some(format!("user denied tool '{gate_action_name}': cancelled")),
+                    ))
```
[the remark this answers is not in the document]
ilblackdragon: The design doc was updated to record this divergence: shipped code uses `RuntimeError` with a descriptive message (`\"user denied tool 'X': <reason>\"`) rather than `PermissionError`. Two reasons: (a) Python scripts already catch `RuntimeError` and the message text is specific enough to distinguish from other runtime errors, (b) introducing `PermissionError` here would force every CodeAct script that catches gate denial to update its except clause, breaking out-of-tree examples. The Testing section of `docs/plans/2026-05-01-codeact-inline-gate-await.md` now documents this choice. Not changing in this PR.

crates/ironclaw_engine/src/executor/scripting.rs
```diff
@@ -1740,30 +1989,171 @@ async fn resolve_tool_future(
...
+                }
+                crate::gate::GateResolution::CredentialProvided { .. }
+                | crate::gate::GateResolution::ExternalCallback { .. } => {
+                    ExtFunctionResult::Error(MontyException::new(
+                        ExcType::RuntimeError,
+                        Some(format!(
+                            "unsupported gate resolution for tool '{gate_action_name}'"
+                        )),
+                    ))
```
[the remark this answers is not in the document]
ilblackdragon: The design doc was updated to record this divergence: shipped code uses `RuntimeError` with a descriptive message (`\"user denied tool 'X': <reason>\"`) rather than `PermissionError`. Two reasons: (a) Python scripts already catch `RuntimeError` and the message text is specific enough to distinguish from other runtime errors, (b) introducing `PermissionError` here would force every CodeAct script that catches gate denial to update its except clause, breaking out-of-tree examples. The Testing section of `docs/plans/2026-05-01-codeact-inline-gate-await.md` now documents this choice. Not changing in this PR.

crates/ironclaw_engine/src/executor/structured.rs
```diff
@@ -522,6 +596,109 @@ fn interrupted_call_needs_refund(result: &Result<ActionResult, EngineError>) ->
...
+async fn execute_with_inline_gate_retry(
+    effects: &Arc<dyn EffectExecutor>,
+    leases: &LeaseManager,
+    lease: &CapabilityLease,
+    call: &ActionCall,
+    exec_ctx: &ThreadExecutionContext,
+    thread: &Thread,
+) -> Result<ActionResult, EngineError> {
+    let mut current_lease = lease.clone();
+    loop {
```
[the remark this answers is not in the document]
ilblackdragon: Already addressed in c09e03c9 (before this review). `execute_with_inline_gate_retry` now bounds at `MAX_INLINE_GATE_RETRIES = 3`, shared with `drive_inline_gate` (Tier 1). On exhaustion the function returns `EngineError::Effect` with a clear \"still requires approval after N retries\" message — bounded, surfaces a clean error, no spin.

src/bridge/gate_controller.rs
ilblackdragon: Fixed in 78a12766. `BridgeGateController::pause` now wraps the oneshot recv in `tokio::select!` against `tokio::time::sleep` aligned to `pending.expires_at`. On expiry we discard the pending row, drop the registry entry, and return `Cancelled` so the VM unwinds. Stop integration is a separate layer (engine signal channel does not currently interrupt mid-`execute_action`); on the follow-ups list.

src/bridge/gate_controller.rs
```diff
@@ -0,0 +1,506 @@
...
+            .unwrap_or(std::time::Duration::ZERO);
+        let pending_key = pending.key();
+        let resolution = tokio::select! {
+            biased;
+            received = rx => match received {
+                Ok(resolution) => resolution,
+                Err(_) => {
+                    // Sender dropped — process shutting down or registry
+                    // cleared. Treat as cancellation.
+                    self.resolutions.forget(request_id).await;
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in 46dca2b1. The sender-dropped branch now mirrors the expiry-branch cleanup: `pending_gates.discard(&pending_key)` before returning `Cancelled`. No more stranded UI prompts and the duplicate-insert guard no longer blocks future gates for the same `(user, thread)` after a process shutdown / registry clear.

crates/ironclaw_engine/src/executor/scripting.rs
```diff
@@ -839,16 +881,231 @@ pub async fn execute_code_with_skills(
...
+                            // (and the LLM) can distinguish denial from other
+                            // runtime errors.
+                            let ext_result = ExtFunctionResult::Error(MontyException::new(
+                                ExcType::RuntimeError,
+                                Some(format!("user denied tool '{gate_action_name}': {reason}")),
+                            ));
+                            match std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| {
+                                call.resume(ext_result, PrintWriter::CollectString(&mut stdout))
+                            })) {
+                                Ok(Ok(p)) => progress = p,
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in 46dca2b1. The sync-preflight denial path now pushes `EventKind::ActionFailed { error: format!("denied: {reason}"), ... }` before resuming Monty with the typed `RuntimeError`. Thread event log is now consistent across all three denial paths (sync preflight, `drive_inline_gate`, `structured.rs`).

tests/engine_v2_gate_integration.rs
```diff
@@ -1161,156 +1163,27 @@ async fn tool_info_does_not_gate_callable_tool_into_next_llm_callable_set() {
...
+    // \`call_approval_granted\` propagation). The bridge-level
+    // `execute_resolved_pending_action` API is separately covered by
+    // \`auth_resolution_retries_same_pending_action_without_second_pause\`
+    // and \`install_auth_resume_followed_by_aliased_tool_call_completes_without_hanging\`
+    // for the Auth resume path, which is the API's remaining responsibility.
+    //
+    // This test is intentionally a no-op now — left in place rather
+    // than deleted so `git log` retains the rationale; remove in a
+    // follow-up sweep.
 }
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in 46dca2b1. Deleted both empty stubs. The post-PR equivalent coverage lives in `codeact_inline_gate_await_*` (this file) and the scripting unit tests; the rationale survives in `git log` (commit 87fe4fce) without an empty test misleading coverage signals.

tests/engine_v2_gate_integration.rs
```diff
@@ -2402,120 +2282,477 @@ async fn auto_approve_mode_still_pauses_always_tools() {
...
+    // resume mechanism for Approval.
+    //
+    // The obligation logic itself (per-message
+    // `signals_execution_intent` detection driving the nudge in the
+    // orchestrator) is not affected by this PR and is exercised by
+    // direct unit tests in `crates/ironclaw_engine/src/executor`.
+    //
+    // Left as a no-op stub for git-log traceability; remove in a
+    // follow-up sweep alongside the deprecated auth-resume harnesses.
+}
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in 46dca2b1. Deleted both empty stubs. The post-PR equivalent coverage lives in `codeact_inline_gate_await_*` (this file) and the scripting unit tests; the rationale survives in `git log` (commit 87fe4fce) without an empty test misleading coverage signals.

src/bridge/gate_controller.rs
```diff
@@ -0,0 +1,511 @@
...
+            return GateResolution::Cancelled;
+        }
+
+        let (tx, rx) = oneshot::channel();
+        self.resolutions.register(request_id, tx).await;
+
+        self.emit_gate_prompt(&pending, &per_exec.channel_metadata)
+            .await;
+
+        // Bound the await on `pending.expires_at`. Without this, a user
```
serrrfirat, line 555: **High Severity** — timeout still joins a parked approval thread

`await_thread_outcome` has a 5-minute safety deadline, but when that deadline fires it only breaks the event loop and then immediately calls `thread_manager.join_thread(thread_id).await`. With inline approval gates, the thread task is parked inside this `pause()` until the user resolves the prompt or the 30-minute `PendingGate` expiry fires, so the supposedly bounded request can still hang for the full approval lifetime (or longer if other future pause paths are added).

Concrete scenario: a gateway request triggers an approval gate, the user ignores the prompt, `await_thread_outcome` reaches its 5-minute deadline, then `join_thread` waits on the same still-running task that is sleeping here until `expires_at`.

Suggested fix: when the router deadline fires, return `BridgeOutcome::Pending`/a timeout response without joining, or explicitly stop/cancel the thread before joining. The important invariant is that the request-level timeout must not call an unbounded join on a task parked in `GateController::pause`.

src/bridge/gate_controller.rs
```diff
@@ -0,0 +1,511 @@
...
+    /// Bind per-execution data for `user_id` BEFORE the engine spawns
+    /// the thread. Closes the race window where a fast tool gate
+    /// reaches `pause()` before the (user, thread)-keyed entry has
+    /// been written.
+    ///
+    /// Per-conversation lock upstream means at most one bridge turn
+    /// per conversation is active at a time; concurrent threads for
+    /// the same user (background missions) are an accepted edge case
+    /// — the most-recently-set context wins.
+    pub async fn set_pre_execution_context(&self, user_id: String, context: PerExecutionContext) {
```
serrrfirat: **High Severity** — pre-execution context can be overwritten across concurrent conversations for the same user

`pre_execution` is keyed only by `user_id`, but the serialization guarantee cited here is per conversation, not per user. If the same user has two concurrent gateway conversations/tabs, turn B can overwrite turn A's pre-execution slot before A's spawned engine thread reaches `set_execution_context`. If A then gates quickly and falls back through `lookup_per_execution`, it will build the `PendingGate` with B's `conversation_id`, `source_channel`, metadata, and scope.

That can surface approval prompts (including redacted tool parameters and thread routing) in the wrong conversation/channel for the same user.

Suggested fix: key the pre-execution registry by a per-turn identity such as `(user_id, conversation_id)` or a generated spawn token, and make the spawned thread carry that token into `GatePauseRequest`/lookup. A single per-user slot is not safe once a user can have multiple active conversations.

src/bridge/gate_controller.rs
```diff
@@ -0,0 +1,511 @@
...
+            .build_pending_gate(
+                request_id,
+                &per_exec,
+                &request.user_id,
+                request.thread_id,
+                &request,
+            )
+            .await;
+
+        if let Err(e) = self.pending_gates.insert(pending.clone()).await {
```
serrrfirat, line 528: **Medium Severity** — simultaneous approval gates in a parallel batch are silently treated as cancellations

`PendingGateStore::insert` allows only one pending gate per `(user, thread)`. In the multi-runnable `__execute_actions_parallel__` path, each runnable action now runs its own inline-retry loop concurrently. If two actions gate before the first one is resolved, both call `pause()`: the first insert succeeds, the second insert hits `AlreadyExists`, and this branch returns `GateResolution::Cancelled`.

That means the second tool is recorded as denied/cancelled without ever prompting the user. This is a plausible single-step scenario with `asyncio.gather()` / multiple tool calls where more than one tool requires approval.

Suggested fix: serialize approval-gated actions in a batch, queue later gates behind the current one, or return an explicit "another approval is already pending" error. Treating an insert collision as user cancellation loses the user's chance to approve the second action.

crates/ironclaw_engine/src/executor/orchestrator.rs
```diff
@@ -1091,7 +1123,12 @@ async fn handle_execute_action(
 
     let call_id = extract_string_kwarg(kwargs, "call_id").unwrap_or_default();
 
-    let mut exec_ctx = thread_execution_context(thread, StepId::new(), Some(call_id.clone()));
+    let mut exec_ctx = thread_execution_context(
+        thread,
+        StepId::new(),
+        Some(call_id.clone()),
+        gate_controller.clone(),
```
serrrfirat, line 1130: **Medium Severity** — policy-level Tier 0 approval gates still use the legacy gate-paused/re-entry path

This context wiring does not make all Tier 0 approval gates inline. In this same function, `PolicyDecision::RequireApproval` still builds and returns a `gate_paused` sentinel instead of calling `gate_controller.pause`; `handle_execute_actions_parallel` has the same behavior in its preflight branch. The Python orchestrator then transitions the thread to waiting and uses the legacy re-entry path.

So action-level `requires_approval` / policy-rule approvals do not get the new inline-await semantics promised by this PR, and can still replay the LLM/orchestrator path on approval. The new tests mainly cover gates returned by `EffectExecutor::execute_action`, not these policy-preflight branches.

Suggested fix: route the `PolicyDecision::RequireApproval` branches through `gate_controller.pause` (like `structured.rs` and the CodeAct preflight path), or explicitly scope/document this PR to adapter-raised gates and add regression coverage for the remaining legacy policy path.

src/bridge/router.rs
```diff
@@ -3569,6 +3770,15 @@ async fn handle_with_engine_inner(
         .await
         .map_err(|e| engine_err("thread error", e))?;
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in d211bfc2. The bridge now matches on `handle_user_message` and explicitly clears the pre-execution slot via `clear_pre_execution_context(user_id, conv_id)` before returning the error, instead of letting `?` exit the function with the slot still installed. No more leak / mis-route on engine-spawn failure.

src/bridge/gate_controller.rs
ilblackdragon: Fixed in d211bfc2. `await_thread_outcome` now tracks a `timed_out` flag; when the 5-min deadline fires AND the thread is still running (typical case: parked in `pause()`), the bridge returns `BridgeOutcome::Pending` instead of `join_thread`. The live `PendingGate` row stays available for resolution and the resolver hand-back path delivers into the parked oneshot. The request handler is no longer blocked on a 30-min approval expiry.

src/bridge/gate_controller.rs
ilblackdragon: Fixed in d211bfc2. `pre_execution` is now keyed by `(user_id, conversation_id)` via a new `PreExecKey`. `conversation_id` flows through `ThreadExecutionContext` (read from `Thread.metadata` populated by `ConversationManager` at thread spawn) and `GatePauseRequest`, so `pause()`'s lookup falls back to the (user, conversation) slot rather than a single per-user slot. Two concurrent tabs for the same user now have isolated context — no more cross-conversation prompt mis-routing.

src/bridge/gate_controller.rs
ilblackdragon: Fixed in d211bfc2. Added a `gate_locks: HashMap<ExecutionKey, Arc<Mutex<()>>>` to `BridgeGateController`; `pause()` acquires the per-(user, thread) lock before `PendingGateStore::insert` and holds it across the select-await window. A parallel batch where two tool calls both gate now queues the second `pause()` behind the first instead of treating the insert collision as `Cancelled`. Each gate gets its own prompt sequentially. Lock entries are dropped on `clear_execution_context` so the registry stays bounded.

crates/ironclaw_engine/src/executor/orchestrator.rs
ilblackdragon: Fixed in d211bfc2. Removed the legacy `gate_paused` JSON sentinel + thread re-entry path for Approval gates from both Tier 1 entry points: `__execute_action__` and `__execute_actions_parallel__`. `PolicyDecision::RequireApproval` now calls `gate_controller.pause()` inline (mirroring `structured.rs`); on approval the call falls through to lease consume + execute, on denial we emit `ActionFailed` and surface a deny-style result. Tool-raised `Err(GatePaused)` from `effects.execute_action` is routed through the existing `execute_single_action_with_inline_retry` wrapper so the single-call path also handles tool gates inline. Authentication / External resume kinds keep the legacy re-entry path (their resolution installs new state that only takes effect on the next run-through).

src/bridge/router.rs
```diff
@@ -3611,7 +3838,14 @@ async fn handle_with_engine_inner(
...
 
     debug!(thread_id = %thread_id, "engine v2: thread spawned");
-    await_thread_outcome(agent, state, message, conv_id, thread_id).await
+    let outcome = await_thread_outcome(agent, state, message, conv_id, thread_id).await;
+    // Drop per-execution context. The `PendingGate` row (if a gate
+    // fired) carries everything the resolver needs from here on.
+    state
+        .gate_controller
+        .clear_execution_context(&message.user_id, thread_id, conv_id)
+        .await;
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in d1b59f91 via the deferred-cleanup approach (option 2 in the planning thread). When `await_thread_outcome` returns `BridgeOutcome::Pending` AND the engine task is still running, `handle_with_engine_inner` now spawns a watcher task (`spawn_deferred_context_cleanup`) that polls `thread_manager.is_running` at 30s intervals until the thread completes, then calls `clear_execution_context(user_id, thread_id, conv_id)`. The (user, thread) context + gate-locks entry stay registered across the gate-resolution wait, so subsequent gates from the same execution find their context and surface a real prompt instead of silent `Cancelled`. Watcher caps at 60 minutes — well past the 30-min PendingGate expiry that bounds any `pause()` call — as a defensive upper bound.

crates/ironclaw_engine/src/executor/structured.rs
```diff
@@ -260,14 +296,8 @@ pub async fn execute_action_calls(
...
-            .execute_action(
-                &call.action_name,
-                call.parameters.clone(),
-                &lease,
-                &exec_ctx,
-            )
-            .await;
+        let exec_result =
+            execute_with_inline_gate_retry(effects, leases, &lease, call, &exec_ctx, thread).await;
         if interrupted_call_needs_refund(&exec_result) {
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in d1b59f91. The multi-runnable JoinSet branch in `execute_action_calls` now routes each task through `execute_with_inline_gate_retry`, matching the single-runnable fast path. An `Approval` gate raised mid-execution in a parallel batch now pauses inline through the bridge controller (which serializes concurrent gates per (user, thread) per the earlier #4 fix), retries on approval, and surfaces a typed denial otherwise — no more `Err(GatePaused)` bubbling through `classify_exec_result` and re-introducing the double-execution path. Signature change: `leases: &LeaseManager` → `&Arc<LeaseManager>` so each spawned task can clone an owned handle. All current callers are tests already using `Arc::new(LeaseManager::new())`, so this is a no-op call-site change.

crates/ironclaw_engine/src/executor/orchestrator.rs
```diff
@@ -1953,6 +1970,183 @@ fn interrupted_result_needs_refund(result: &serde_json::Value) -> bool {
...
+                thread_id: thread.id,
+                user_id: thread.user_id.clone(),
+                gate_name: result_json
+                    .get("gate_name")
+                    .and_then(|v| v.as_str())
+                    .unwrap_or("approval")
+                    .to_string(),
+                action_name: name.to_string(),
+                call_id: call_id.to_string(),
+                parameters: params.clone(),
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in 5ee08e60. `execute_single_action_with_inline_retry` now sources `GatePauseRequest.parameters` from `result_json["parameters"]` (the gate-paused payload, which carries safety-layer transformations/redactions) and only falls back to the original caller `params` if absent — same contract as `structured::execute_with_inline_gate_retry`. The UI prompt now reflects what the tool actually wanted to run with.

crates/ironclaw_engine/src/executor/orchestrator.rs
```diff
@@ -1777,37 +1782,47 @@ async fn handle_execute_actions_parallel(
...
+            let leases = Arc::clone(&leases_arc);
+            let thread_snapshot = thread_snapshot.clone();
             let lease = lease.clone();
-            let mut exec_ctx = thread_execution_context(thread, step_id, Some(pc_call_id.clone()));
+            let mut exec_ctx = thread_execution_context(
+                &thread_snapshot,
+                step_id,
+                Some(pc_call_id.clone()),
+                gate_controller.clone(),
+            );
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in 5ee08e60. The multi-call parallel branch no longer clones the full `Thread` per task. `execute_single_action_with_inline_retry` now takes `thread_id: ThreadId` + `user_id: &str` (the only fields it touched). The pre-spawn setup builds a single `base_exec_ctx` from the live thread once and clones it per task, overriding `current_call_id` and the action snapshots — far cheaper than rebuilding via `thread_execution_context(&thread_snapshot, ...)` for every spawn.

crates/ironclaw_engine/src/executor/structured.rs
```diff
@@ -290,14 +328,16 @@ pub async fn execute_action_calls(
...
+            let thread = thread.clone();
 
             join_set.spawn(async move {
                 let execution_start = Instant::now();
-                let result = effects
-                    .execute_action(&call.action_name, call.parameters.clone(), &lease, &ctx)
-                    .await;
+                let result =
+                    execute_with_inline_gate_retry(&effects, &leases, &lease, &call, &ctx, &thread)
+                        .await;
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in 5ee08e60. Same refactor as the orchestrator path: `execute_with_inline_gate_retry` now takes `thread_id: ThreadId` + `user_id: &str` instead of `&Thread`. The JoinSet spawn captures `thread_id` and `user_id` once outside the loop and clones only the cheap String per task, dropping the per-task `let thread = thread.clone();` that pulled in the full message/event transcript.

src/bridge/gate_controller.rs
```diff
@@ -0,0 +1,623 @@
...
+        // VM keeps running until something else (process restart,
+        // join_thread timeout) tears it down. Race the receiver against
+        // a sleep; whichever resolves first wins.
+        let expires_at = pending.expires_at;
+        let now = chrono::Utc::now();
+        let timeout_dur = (expires_at - now)
+            .to_std()
+            .unwrap_or(std::time::Duration::ZERO);
+        let pending_key = pending.key();
+        let resolution = tokio::select! {
```
serrrfirat, line 567: **High Severity** — stop/cancel does not unblock a thread parked in inline gate await.

`BridgeGateController::pause()` currently selects only on the approval receiver or the 30-minute gate expiry. But `ThreadManager::stop_thread()` only sends `ThreadSignal::Stop`, and while the engine is parked inside this `pause()` future it is not polling the signal receiver. That means a user/operator stop request against a thread waiting for approval does not actually stop the thread promptly; it can remain running until the user resolves the gate or the 30-minute expiry fires.

Please add a cancellation path for live gate waits (for example a per-thread cancellation token/notifier that `stop_thread()` triggers and `pause()` selects on), and clean up the pending gate/resolution entry when cancellation wins. A caller-level regression test should start a thread that parks on approval, call `stop_thread()`, and assert the running state and prompt are cleared promptly rather than waiting for expiry.

src/bridge/gate_controller.rs
```diff
@@ -0,0 +1,623 @@
...
+            display_parameters,
+            description: format!(
+                "Tool '{}' requires {} (gate: {})",
+                req.action_name,
+                req.resume_kind.kind_name(),
+                req.gate_name
+            ),
+            resume_kind: req.resume_kind.clone(),
+            created_at: chrono::Utc::now(),
+            expires_at: chrono::Utc::now() + chrono::Duration::minutes(30),
```
serrrfirat, line 400: **Medium Severity** — live inline gate waits are uncapped per user/process.

Every ignored approval now holds a live engine/VM task parked until resolution or the 30-minute expiry. I don't see a per-user or global cap around `pause()` before inserting/awaiting the pending gate. An authenticated user can open many conversations that each hit an approval gate and never resolve them, accumulating parked tasks, pending rows, and retained VM/thread state.

The design doc's follow-up list calls out a per-user concurrency cap for in-flight gates; I think this needs to land with the inline-await substrate or be guarded by an existing request/thread budget. Please add a bounded per-user (and ideally global) semaphore/counter around live inline gate waits, with a clean failure when the cap is exceeded, plus a test covering the cap.

crates/ironclaw_engine/src/executor/structured.rs
```diff
@@ -522,6 +580,117 @@ fn interrupted_call_needs_refund(result: &Result<ActionResult, EngineError>) ->
...
+                &call.action_name,
+                call.parameters.clone(),
+                &current_lease,
+                &call_ctx,
+            )
+            .await;
+        call_ctx.call_approval_granted = false;
+
+        let (gate_name, action_name, call_id, parameters, resume_kind) = match result {
+            Err(EngineError::GatePaused {
```
serrrfirat, line 649: **Medium Severity** — structured inline approval drops the `ApprovalRequested` audit event.

In the structured executor's mid-execution inline retry path, the `GatePaused` approval is handled inside `execute_with_inline_gate_retry()` and the helper returns only the final `ActionResult` / error. Because the gate is no longer returned to `classify_exec_result()`, the `ApprovalRequested` event that `classify_exec_result()` would have emitted for `EngineError::GatePaused` is lost. The call eventually records only `ActionExecuted`/`ActionFailed`, so replay/audit observers cannot see that user approval was required for this action.

The orchestrator path solves this by returning `Vec<EventKind>` from the inline retry helper and appending both the approval request and the terminal outcome. Please give the structured path the same shape, or otherwise emit `ApprovalRequested` before awaiting the controller, and add a test that a mid-execution approval produces both `ApprovalRequested` and the final outcome event.

nickpismenkov: Fixed in d5fb1fa7. `execute_with_inline_gate_retry` now returns `(Result<ActionResult, EngineError>, Vec<EventKind>)` — one `ApprovalRequested` event per retry iteration that gated, mirroring the orchestrator (Tier 1) shape. The slot type widens to `(ActionResult, EventKind, Vec<EventKind>)` so the merge phase flushes pre-terminal events before the terminal event in original-call order.

Two regression tests pin the contract:
- `inline_retry_emits_approval_requested_event_before_outcome` — approved path: events contain `ApprovalRequested` then `ActionExecuted`, in that order.
- `inline_retry_emits_approval_requested_event_before_denial` — denied path: events contain `ApprovalRequested` then `ActionFailed`, in that order.

src/bridge/gate_controller.rs
nickpismenkov: Fixed in d5fb1fa7. Added `GateController::cancel_thread(thread_id)` (default no-op) to the engine-side trait. `ThreadManager::stop_thread()` calls it BEFORE sending `ThreadSignal::Stop` so any parked `pause()` future wakes promptly with `GateResolution::Cancelled`.

`BridgeGateController` tracks in-flight pauses in `active_pauses: HashMap<ThreadId, HashSet<Uuid>>` (HashSet rather than Option<Uuid> so a future loosening of the per-(user,thread) `gate_locks` serialization wouldn't drop request_ids on the floor). On `cancel_thread()`:
1. Snapshot the active request_ids for the thread (releasing the lock before the per-id `try_deliver` calls).
2. Walk the snapshot, deliver `Cancelled` via the existing `GateResolutions::try_deliver` channel.
3. Discard pending DB rows via `PendingGateStore::discard_for_thread`.
4. Clear the active set.

`pause()` always untracks on exit; `cancel_thread` may have already removed the entry — that's idempotent.

Two new bridge-tier unit tests:
- `cancel_thread_wakes_parked_pause_with_cancelled_resolution` — drives a real `pause()` future in a spawned task, asserts `cancel_thread` wakes it within 2s with `GateResolution::Cancelled`, asserts `active_pauses` is cleared.
- `cancel_thread_with_no_active_pause_is_a_no_op` — pins the happy-path no-op.

src/bridge/gate_controller.rs
nickpismenkov: Acknowledged but pushing back on landing the cap in this PR. Reasoning written down at the call site as a TODO in d5fb1fa7:

```rust
// TODO(#3157 follow-up — design-doc item): bound live inline
// gate awaits per user / globally with a typed semaphore so an
// authenticated user opening many threads each with an
// unresolved approval gate cannot accumulate parked engine
// tasks/pending rows past a budget. The current implicit bound
// is one pending gate per (user, thread) × the existing
// thread-creation budget × the 30-min expiry; that is enough
// to ship the first inline-await slice but not enough as a
// long-term DoS guard. Track in a separate issue once the
// semaphore design (cap UX, fairness, rejection error shape)
// is settled rather than hand-rolling it inside this controller.
```

The implicit bound today is one pending gate per `(user, thread)` × the existing thread-creation budget × the 30-min expiry. The reviewer's own comment notes this is on the design doc's follow-up list, not a blocker for the inline-await substrate. Adding the cap correctly requires designing:

1. Cap value source (config? per-tenant? per-trust-class?)
2. Fairness (FIFO across users, or strictly per-user counters?)
3. Rejection UX (reject the gate? queue? surface a typed error to the executor that maps to a deny-style `EngineError::Effect`?)
4. Telemetry hooks so an operator can see when the cap kicks in.

Hand-rolling those decisions inside `BridgeGateController::pause` to clear the review without a separate design pass would be the wrong-shape fix. Happy to open the follow-up issue and link it from this thread if that's preferred.

nickpismenkov (approved):

Issue #49, opened 2026-02-12
Support per-request model override in /v1/chat/completions


jaswinder6991:
## Context

The OpenAI-compatible API (`/v1/chat/completions`) currently validates that `req.model` matches the active model configured at startup (`NEARAI_MODEL`) and returns 404 if they differ. This is because the `LlmProvider` trait takes the model from its internal state rather than as a per-request parameter.

NEAR AI supports many models on the same API key, so ideally clients should be able to specify any supported model per request — matching how the real OpenAI API works.

## Problem

- `CompletionRequest` and `ToolCompletionRequest` have no `model` field
- `LlmProvider::complete()` / `complete_with_tools()` read the model from `self.active_model` (a `RwLock<String>`)
- Calling `set_model()` mutates shared state — if the agent and API share one provider instance, concurrent requests with different models would race

## Proposed approach

1. Add an optional `model: Option<String>` field to `CompletionRequest` and `ToolCompletionRequest`
2. In `NearAiChatProvider::complete()` / `complete_with_tools()`, use `req.model` if set, otherwise fall back to `self.active_model_name()`
3. In `chat_completions_handler`, pass `req.model` through instead of validating against the active model
4. The agent loop continues to omit the field, so it uses the configured default as before

This avoids mutating shared state and has zero impact on existing agent behavior.

## Ref

PR #31 review discussion — model field handling.

ilblackdragon:
@BroccoliFin yes, you can use Ollama and prob LMStudio via OpenAI compatible API

Issue #48, opened 2026-02-12
idea/feat/feature: implementing a guardrails model


bkutasi:
There are a couple of mostly older guardrail models out there like granite-guardian-3.0-8b and llama-guard-4-12b. Cloud ironclaw use something like this to enhance safety and reduce prompt injection attack frequency? Would this make sense?

A couple of questions related:
- how vulnerable is ironclaw to prompt injection (absolute vs relative to openclaw)?
- how frequent we would need to call the guardrails API (wastes compute if every call is evaluated)
- adds latency
- it could still outsmart the guardrails model (making it useless)

joe-rlo:
Closed due to inactivity — reopen if still needed.

PR #47, opened 2026-02-12
feat: add libSQL/Turso embedded database backend

zmanian:
## Summary

Adds libSQL (Turso) as an embedded database backend alongside PostgreSQL, enabling zero-dependency local mode. Implements full feature parity between both backends behind compile-time feature flags.

- Introduce `Database` trait abstraction (~60 async methods) in `src/db/` with implementations for both PostgreSQL and libSQL
- Convert all modules from concrete `Store` to `Arc<dyn Database>` for backend-agnostic persistence
- Add `LibSqlSecretsStore` (8 trait methods) and `LibSqlWasmToolStore` (7 trait methods)
- Wire libsql stores throughout CLI entry points (`main.rs`, `cli/tool.rs`, `cli/mcp.rs`)
- Make setup wizard backend-agnostic (`SecretsContext` uses `Arc<dyn SecretsStore>`)
- Add libSQL migration system with idempotent SQLite-dialect schema (~480 lines)
- Feature-gate postgres-only tests and examples

### Backend selection

```bash
# Default: PostgreSQL
cargo build

# Embedded libSQL (zero external dependencies)
cargo build --no-default-features --features libsql

# Both backends available
cargo build --features "postgres,libsql"
```

### Schema translation

libSQL schema translates PostgreSQL types: `UUID` -> `TEXT`, `TIMESTAMPTZ` -> `TEXT` (ISO-8601), `JSONB` -> `TEXT`, `VECTOR(1536)` -> `F32_BLOB(1536)` with `libsql_vector_idx`, `tsvector` -> FTS5 virtual tables with sync triggers.

Closes #19

## Test plan

- [x] `cargo check --no-default-features --features libsql` passes
- [x] `cargo check` (default postgres) passes
- [x] `cargo test --no-default-features --features libsql` -- 716 tests pass, 0 failures
- [x] `cargo test` (default postgres) -- 693 pass, 1 pre-existing failure (unrelated telegram.wasm fixture)

Generated with [Claude Code](https://claude.com/claude-code)

PR #44, opened 2026-02-12
privacy closure

Nuropunk:
Closed for privacy.

PR #3171, opened 2026-05-01
Add Reborn event store backends

### LINKED ISSUE #3162: [Reborn] Add standalone durable event/audit store for production binary
## Parent

#2987

Related:
- #3022
- #3026
- #3093

## Problem

Reborn’s `ironclaw_events` crate defines the correct durable contracts:

- `DurableEventLog`
- `DurableAuditLog`
- `EventCursor`
- scoped replay
- replay gaps
- redaction-safe event/audit records

…but the current concrete backends are in-memory/reference only. That is fine for tests and demos, but not acceptable for the standalone Reborn production binary. The Reborn binary must not lose runtime/audit history on restart, and it must not depend on V1 database traits, migrations, or `src/app.rs`.

## Hard constraints

- Do **not** modify V1 persistence code:
  - no `src/db/*`
  - no V1 `migrations/*`
  - no V1 `AppBuilder`
  - no V1 `Database` trait
- Keep Reborn standalone.
- Reborn durable event/audit storage must live in Reborn-owned crates and migrations.
- `ironclaw_events` should remain the contract/substrate crate; avoid pulling DB drivers into it unless explicitly justified.

## What to build

Create a Reborn-owned durable event/audit backend crate and wire it into the standalone Reborn composition path.

Suggested crate:

```text
crates/ironclaw_reborn_event_store/
```

This crate implements:

```rust
DurableEventLog
DurableAuditLog
```

for production storage backends used by the standalone Reborn binary.

### 1. Add a Reborn-only event store crate

Create:

```text
crates/ironclaw_reborn_event_store/
```

Responsibilities:

- Own DB-backed event/audit implementations.
- Own Reborn event-store migrations.
- Expose backend-neutral factory/config types.
- Depend on `ironclaw_events`, not the other way around.

Suggested public API:

```rust
pub enum RebornEventStoreConfig {
    InMemory,
    Jsonl { root: PathBuf },
    Postgres { url: SecretString },
    Libsql { path_or_url: String, auth_token: Option<SecretString> },
}

pub enum RebornProfile {
    LocalDev,
    Test,
    Production,
}

pub struct RebornEventStores {
    pub events: Arc<dyn DurableEventLog>,
    pub audit: Arc<dyn DurableAuditLog>,
}

pub async fn build_reborn_event_stores(
    profile: RebornProfile,
    config: RebornEventStoreConfig,
) -> Result<RebornEventStores, RebornEventStoreError>;
```

### 2. Enforce production fail-closed behavior

Factory behavior:

```text
profile = LocalDev/Test + InMemory  => allowed
profile = Production + InMemory     => error
profile = Production + Jsonl        => allowed only if explicitly accepted as single-node durable
profile = Production + Postgres     => allowed
profile = Production + Libsql       => allowed
```

Acceptance:

- Production profile cannot silently fall back to memory.
- Missing durable config fails before returning service graph.
- Errors must be redacted: no connection strings, tokens, host secrets.

### 3. Add Reborn-owned migrations

Do **not** use root V1 migrations.

Place migrations under the new crate, for example:

```text
crates/ironclaw_reborn_event_store/migrations/postgres/
crates/ironclaw_reborn_event_store/migrations/libsql/
```

Suggested tables:

```text
reborn_event_streams
reborn_event_entries
```

`reborn_event_streams`:

```text
stream_kind        -- runtime | audit
tenant_id
user_id
agent_id          -- empty string or nullable, but consistent
next_cursor
earliest_retained
created_at
updated_at
PRIMARY KEY (stream_kind, tenant_id, user_id, agent_id)
```

`reborn_event_entries`:

```text
stream_kind
tenant_id
user_id
agent_id
cursor
record_id
record_kind
project_id
mission_id
thread_id
process_id
occurred_at
record_json
created_at
PRIMARY KEY (stream_kind, tenant_id, user_id, agent_id, cursor)
```

Postgres uses JSONB/TIMESTAMPTZ.  
libSQL uses TEXT JSON/RFC3339 timestamps.

### 4. Implement backend adapters

Implement:

```rust
PostgresDurableEventLog
PostgresDurableAuditLog
LibsqlDurableEventLog
LibsqlDurableAuditLog
JsonlDurableEventLog
JsonlDurableAuditLog
```

If this is too much for one PR, split into:

1. crate + in-memory/profile validation + JSONL
2. Postgres
3. libSQL

But the final production issue should not close until Postgres/libSQL parity exists.

### 5. Preserve cursor semantics exactly

For both runtime events and audit records:

- append atomically allocates next cursor per stream
- cursors are monotonic per `(stream_kind, tenant, user, agent)`
- `read_after_cursor(None)` equals origin
- `limit == 0` errors
- cursor beyond stream head returns `ReplayGap`
- cursor older than retained floor returns `ReplayGap`
- `ReadScope` filtering is enforced by backend
- filtered-out records still advance scanned cursor consistently with in-memory behavior

### 6. Add durable sink adapters

Add:

```rust
DurableEventSink<L: DurableEventLog>
DurableAuditSink<L: DurableAuditLog>
```

These implement:

```rust
EventSink
AuditSink
```

Contract:

- best-effort sink failures must not change domain outcomes
- append failures are logged/observable
- services that require durable audit as source-of-truth should depend on `DurableAuditLog` directly, not only `AuditSink`

### 7. Add standalone composition tests

Add tests proving standalone Reborn composition behavior without V1 code.

Suggested tests:

```text
crates/ironclaw_reborn_event_store/tests/profile_contract.rs
crates/ironclaw_reborn_event_store/tests/durable_event_store_contract.rs
crates/ironclaw_host_runtime/tests/standalone_event_store_composition_contract.rs
```

Required cases:

- local-dev allows in-memory event store
- production rejects in-memory event store
- production accepts configured durable backend
- runtime event append/replay works through `DurableEventLog`
- audit append/replay works through `DurableAuditLog`
- foreign scope cannot replay records
- sentinel raw secret/path/output strings do not appear in serialized records

### 8. Wire into standalone Reborn binary/service graph

Wherever the standalone Reborn binary composition lives or is being introduced, wire:

```text
Reborn config
  -> build_reborn_event_stores(...)
  -> HostRuntimeServices / dispatcher / process / approval / audit wiring
```

If the standalone binary crate does not exist yet, this issue should expose the factory and tests, and a follow-up binary-composition issue should consume it.

Do **not** wire through V1 `src/app.rs`.

### 9. Update Reborn docs

Update Reborn-only docs:

```text
docs/reborn/contracts/events-projections.md
docs/reborn/harness/observability.md
docs/reborn/README.md
```

Document:

- in-memory is local/test only
- production requires durable event/audit backend
- storage backends supported
- cursor/replay semantics
- replay-gap behavior
- no-exposure rules
- relation to #3093 projections

## Acceptance criteria

- [ ] New Reborn-owned event store crate exists.
- [ ] No V1 `src/db`, V1 migrations, or V1 `AppBuilder` code is modified.
- [ ] Production profile rejects in-memory event/audit logs.
- [ ] Local/test profile can explicitly use in-memory event/audit logs.
- [ ] At least one durable backend exists in the new Reborn store crate.
- [ ] PostgreSQL backend implemented or explicitly split into linked follow-up.
- [ ] libSQL backend implemented or explicitly split into linked follow-up.
- [ ] Runtime events implement `DurableEventLog`.
- [ ] Audit records implement `DurableAuditLog`.
- [ ] Cursor/replay semantics match `InMemoryDurableEventLog`.
- [ ] Cross-scope replay does not leak events.
- [ ] Serialized rows/files do not contain sentinel raw secrets, host paths, request payloads, runtime output, approval reasons, or backend error details.
- [ ] Standalone Reborn composition can consume the durable stores without touching V1 composition.
- [ ] Docs explain standalone production requirements.

## Test plan

Minimum:

```bash
cargo test -p ironclaw_events
cargo test -p ironclaw_reborn_event_store
cargo test -p ironclaw_host_runtime
cargo clippy -p ironclaw_reborn_event_store --all-targets -- -D warnings
cargo clippy -p ironclaw_events --all-targets -- -D warnings
git diff --check
```

Backend-specific tests should run when env/config is available. If skipped, they must clearly say which backend config is missing.

## Non-goals

- V1 DB integration.
- V1 `AppBuilder` wiring.
- V1 migration changes.
- SSE/WebSocket transport.
- Full `EventProjectionService`.
- UI timeline projection.
- Legacy engine event migration.

Those belong in separate Reborn standalone cutover issues.

zmanian:
## Summary
- Add Reborn-owned durable event/audit store crate with JSONL, PostgreSQL, and libSQL backends plus migrations.
- Wire host-runtime composition coverage through the new event-store factory while keeping storage drivers out of ironclaw_events.
- Document event/projection contracts and backend guardrails for standalone Reborn composition.

Closes #3162
Refs #2987

## Test Plan
- cargo test -p ironclaw_reborn_event_store
- cargo test -p ironclaw_reborn_event_store --features "libsql postgres"
- IRONCLAW_REBORN_EVENT_STORE_POSTGRES_URL=postgres://postgres:postgres@127.0.0.1:<port>/ironclaw_reborn_test cargo test -p ironclaw_reborn_event_store --features postgres postgres_runtime_and_audit_logs_survive_rebuild_with_filtered_cursor_semantics
- cargo test -p ironclaw_host_runtime
- cargo test -p ironclaw_events
- cargo test -p ironclaw_architecture reborn_crate_dependency_boundaries_hold
- cargo clippy -p ironclaw_reborn_event_store --all-targets -- -D warnings
- cargo clippy -p ironclaw_reborn_event_store --features "libsql postgres" --all-targets -- -D warnings
- cargo clippy -p ironclaw_host_runtime --all-targets -- -D warnings
- cargo fmt --all -- --check
- git diff --check


crates/ironclaw_reborn_event_store/src/lib.rs
```diff
@@ -0,0 +1,543 @@
...
+        kind: StreamKind,
+        stream: &EventStreamKey,
+        record: T,
+    ) -> Result<EventLogEntry<T>, EventError>
+    where
+        T: Clone + Serialize + DeserializeOwned,
+    {
+        let lock = self.stream_lock(kind, stream).await;
+        let _guard = lock.lock().await;
+        let entries = self.read_entries::<T>(kind, stream).await?;
```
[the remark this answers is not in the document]
zmanian: Fixed in 7b99186c. JSONL append now reads only the tail record cursor via a bounded reverse scan instead of replaying the full stream file before every append.

crates/ironclaw_reborn_event_store/src/lib.rs
```diff
@@ -0,0 +1,543 @@
...
+                filter.matches_audit(record)
+            })
+            .await
+    }
+}
+
+#[derive(Debug, Clone)]
+struct JsonlStore {
+    root: PathBuf,
+    locks: Arc<Mutex<HashMap<String, Arc<Mutex<()>>>>>,
```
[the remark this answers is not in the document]
zmanian: Fixed in 7b99186c. The JSONL lock registry now stores Weak references and prunes released locks on access, so stream locks do not keep accumulating after the active append/read guard is dropped.

crates/ironclaw_reborn_event_store/src/postgres_store.rs
```diff
@@ -0,0 +1,364 @@
...
+                &[
+                    &kind,
+                    &stream.tenant_id.as_str(),
+                    &stream.user_id.as_str(),
+                    &agent_id,
+                    &after_i64,
+                ],
+            )
+            .await
+            .map_err(|_| durable_error("postgres event store failed to read entries"))?;
```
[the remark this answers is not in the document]
zmanian: Fixed in 7b99186c. PostgreSQL replay now builds the ReadScope predicates into the SQL WHERE clause and applies SQL LIMIT, while preserving replay cursor behavior for empty filtered pages.

crates/ironclaw_reborn_event_store/src/postgres_store.rs
```diff
@@ -0,0 +1,364 @@
...
+                r#"
+                SELECT cursor, record_json
+                FROM reborn_event_entries
+                WHERE stream_kind = $1
+                    AND tenant_id = $2
+                    AND user_id = $3
+                    AND agent_id = $4
+                    AND cursor > $5
+                ORDER BY cursor ASC
+                "#,
```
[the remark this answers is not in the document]
zmanian: Fixed in 7b99186c. PostgreSQL replay now applies SQL-level LIMIT and ReadScope predicates before decoding rows, so replay cost is bounded by the requested page size for the matching scope.

crates/ironclaw_reborn_event_store/src/libsql_store.rs
```diff
@@ -0,0 +1,498 @@
...
+                r#"
+                SELECT cursor, record_json
+                FROM reborn_event_entries
+                WHERE stream_kind = ?1
+                    AND tenant_id = ?2
+                    AND user_id = ?3
+                    AND agent_id = ?4
+                    AND cursor > ?5
+                ORDER BY cursor ASC
+                "#,
```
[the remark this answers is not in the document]
zmanian: Fixed in 7b99186c. libSQL replay now builds the same scoped SQL query and LIMIT clause, so small replay requests no longer read and decode the full tail of the stream.

crates/ironclaw_reborn_event_store/src/postgres_store.rs
```diff
@@ -0,0 +1,384 @@
...
+    },
+};
+
+const POSTGRES_EVENT_STORE_SCHEMA: &str =
+    include_str!("../migrations/postgres/001_initial_event_store.sql");
+
+pub(crate) async fn build_postgres_event_stores(
+    url: SecretString,
+) -> Result<RebornEventStores, RebornEventStoreError> {
+    let (client, connection) = tokio_postgres::connect(url.expose_secret(), NoTls)
```
serrrfirat: **High Severity** — Postgres event store disables TLS.

Production Postgres connects with `tokio_postgres::connect(..., NoTls)`. For a remote Postgres URL this either sends credentials plus event/audit rows over plaintext if the server allows it, or fails against the SSL-required managed Postgres deployments we should expect in production.

Please add TLS/rustls support for Postgres event-store connections, or fail closed for non-local Postgres URLs unless TLS is configured.

crates/ironclaw_reborn_event_store/src/lib.rs
```diff
@@ -0,0 +1,649 @@
...
+            last_scanned = entry.cursor;
+            if !is_match(&entry.record) {
+                continue;
+            }
+            replay_entries.push(entry);
+            if replay_entries.len() >= limit {
+                break;
+            }
+        }
+        let next_cursor = replay_entries
```
serrrfirat: **Medium Severity** — Filtered-tail records do not advance `next_cursor`.

When a replay finds at least one matching record, `next_cursor` is set to the last matched entry. If cursor 1 matches the filter and cursor 2 is filtered out, a replay with a large limit returns entry 1 with `next_cursor = 1` rather than advancing to the scanned stream head (2). The next call will rescan cursor 2 forever until another matching record appears.

This violates the contract documented for durable logs: records filtered out by `ReadScope` still advance the scanned cursor. The same pattern appears in the SQL stores (`libsql_store.rs` and `postgres_store.rs`) where `next_cursor` also falls back to the stream head only when no entries matched at all.

Please track `last_scanned` independently while reading rows/entries and return the last scanned cursor when it is greater than the last matched cursor.

crates/ironclaw_reborn_event_store/src/lib.rs
```diff
@@ -0,0 +1,649 @@
...
+    async fn read_entries<T>(
+        &self,
+        kind: StreamKind,
+        stream: &EventStreamKey,
+    ) -> Result<Vec<EventLogEntry<T>>, EventError>
+    where
+        T: DeserializeOwned,
+    {
+        let path = self.stream_path(kind, stream);
+        let bytes = match tokio::fs::read(&path).await {
```
serrrfirat: **Medium Severity** — Bounded JSONL replay still reads/parses the whole stream.

`read_entries` loads the entire JSONL file and parses every record before `read_after` applies the cursor/filter/limit. A caller asking for `limit = 1` on a large production-accepted JSONL stream can still force full-file allocation and full-stream parse latency/OOM.

Please stream replay from disk, maintain an offset/cursor index, or add a hard cap/retention story before accepting JSONL as a production-capable durable backend.

crates/ironclaw_reborn_event_store/src/lib.rs
```diff
@@ -0,0 +1,649 @@
...
+    async fn append<T>(
+        &self,
+        kind: StreamKind,
+        stream: &EventStreamKey,
+        record: T,
+    ) -> Result<EventLogEntry<T>, EventError>
+    where
+        T: Clone + Serialize + DeserializeOwned,
+    {
+        let lock = self.stream_lock(kind, stream).await;
```
serrrfirat, line 385: **Medium Severity** — JSONL cursor assignment is only process-local.

The append path protects cursor assignment with an in-process Tokio mutex. Two IronClaw processes (or two service instances) configured with the same JSONL root can both read the same last cursor and append duplicate cursor envelopes. After that, `parse_jsonl_entries` will reject the stream as having an invalid cursor sequence, effectively corrupting replay for that stream.

Please use an OS-level file lock/atomic cursor sidecar for JSONL appends, or explicitly enforce/document JSONL as single-process single-writer only when building the store.

crates/ironclaw_reborn_event_store/src/postgres_store.rs
zmanian: Fixed in 35674f6a. The Postgres event-store client now uses `tokio-postgres-rustls` for any non-loopback URL, mirroring the connector pattern in `src/db/tls.rs` (system roots via `rustls-native-certs` with `webpki-roots` as the bundled fallback for slim containers). Localhost / 127.0.0.1 / ::1 / 0.0.0.0 / Unix-socket URLs continue to use `NoTls`. Unparseable URLs with a scheme are treated as remote so a typo in the connection string cannot silently downgrade to plaintext. Unit tests cover all three classes (`local_postgres_urls_are_recognised`, `remote_postgres_urls_require_tls`, `unparseable_postgres_url_with_scheme_falls_closed_to_remote`).

crates/ironclaw_reborn_event_store/src/lib.rs
zmanian: Fixed in 35674f6a. The JSONL, libSQL, and Postgres `read_after` paths now track `last_scanned` independently from `last_matched` and return `max(last_matched, last_scanned)` so a filtered-out tail record still advances `next_cursor`. Concretely, in JSONL the loop computes `last_scanned` from every cursor it crosses past `after` and reports it when it is greater than the last matched cursor (or as the next_cursor when no entries matched at all). The SQL backends mirror the same pattern. Regression test: `jsonl_replay_advances_next_cursor_past_trailing_filtered_records`.

crates/ironclaw_reborn_event_store/src/lib.rs
zmanian: Fixed in 35674f6a. The JSONL replay path now streams the file via `BufReader` line-by-line, decodes only the cheap `JsonlCursor` envelope until a line crosses `after`, and breaks out of the loop as soon as `limit` matches are collected. A `limit = 1` request against a multi-gigabyte JSONL no longer pays full-file allocation or full-stream parse cost. Regression test: `jsonl_bounded_replay_does_not_parse_the_whole_file` appends a deliberately malformed JSON line after the prefix and asserts the bounded replay returns the limited prefix successfully without ever touching the trailing garbage.

crates/ironclaw_reborn_event_store/src/lib.rs
zmanian: Fixed in 35674f6a. JSONL appends now take an OS-level exclusive advisory lock (`std::fs::File::lock`, stable since Rust 1.89, fcntl on Unix / LockFileEx on Windows) covering both the prior-tail-cursor read AND the write of the new envelope. Two IronClaw processes pointing at the same JSONL root now block on this lock and emit monotonically-sequenced cursors instead of corrupting the stream with duplicates. Readers take a shared (`lock_shared`) lock so they never observe a partially-written tail line. No new crate dependency was needed — this uses stdlib. Regression test: `jsonl_concurrent_appenders_emit_monotonic_cursors_through_file_lock` builds two independent `JsonlStore` instances (which simulate two processes since they have separate Tokio mutex registries) over the same root, fires 16 concurrent appends from each, then asserts a fresh reader observes 32 contiguously-sequenced entries.

crates/ironclaw_reborn_event_store/src/postgres_store.rs
```diff
@@ -0,0 +1,509 @@
...
+const POSTGRES_EVENT_STORE_SCHEMA: &str =
+    include_str!("../migrations/postgres/001_initial_event_store.sql");
+
+/// Returns true if the supplied Postgres URL targets a loopback host (or a
+/// Unix socket). Anything else is treated as remote and must use TLS.
+///
+/// Conservative: an unparseable URL is treated as remote (fail closed).
+fn is_local_postgres_url(url: &str) -> bool {
+    // Unix socket connections (`host=/var/run/...`) have no `://` host.
+    if !url.contains("://") {
```
serrrfirat: **High Severity** — Remote Postgres keyword connection strings can bypass TLS.

`is_local_postgres_url()` treats any connection string without `://` as local, so a libpq-style keyword string such as `host=db.example.com user=event_user dbname=ironclaw` takes the `NoTls` path below. That sends the event-store DB connection over plaintext even though the host is remote (and even if the string includes `sslmode=require`, because the connector choice has already been forced to `NoTls`).

Please parse keyword configs before classifying locality. A safer shape is to parse through `tokio_postgres::Config`, inspect the resolved host(s), and use `NoTls` only for Unix sockets or loopback hosts; otherwise require the rustls connector and fail closed if TLS cannot be established.

crates/ironclaw_reborn_event_store/src/libsql_store.rs
```diff
@@ -0,0 +1,550 @@
...
+                    RebornEventStoreError::io("initialize libsql parent", source)
+                })?;
+            }
+        }
+        libsql::Builder::new_local(path_or_url).build().await
+    };
+    db.map(Arc::new)
+        .map_err(|source| RebornEventStoreError::backend("libsql", "connect", source))
+}
+
```
serrrfirat, line 98: **High Severity** — Production libSQL allows cleartext remote URLs with auth tokens.

`is_remote_libsql()` accepts `http://` and `build_database()` then forwards the optional `auth_token` into `libsql::Builder::new_remote(...)`. In a production profile this permits bearer/event-store traffic over cleartext HTTP.

Please reject `http://` for `RebornProfile::Production` (and preferably for any non-loopback remote URL), requiring `https://` or `libsql://` for remote libSQL. If localhost HTTP is needed for tests/dev, make that an explicit local/test-only path and avoid sending production tokens over it.

crates/ironclaw_reborn_event_store/src/libsql_store.rs
```diff
@@ -0,0 +1,550 @@
...
+                "#
+        .to_string();
+        let mut query_params = vec![
+            libsql::Value::Text(kind.to_string()),
+            libsql::Value::Text(stream.tenant_id.as_str().to_string()),
+            libsql::Value::Text(stream.user_id.as_str().to_string()),
+            libsql::Value::Text(agent_id.to_string()),
+            libsql::Value::Integer(after_i64),
+        ];
+        push_text_filter(
```
serrrfirat, line 466: **Medium Severity** — SQL replay does not advance past trailing filtered records.

The SQL backends push `ReadScope` into the SQL `WHERE` clause, so filtered-out rows are never scanned. That violates the documented durable-log contract (and the in-memory/JSONL behavior) that filtered-out records still advance the scanned cursor.

Concrete repro with libSQL:
1. append cursor 1 for `project-a`
2. append cursor 2 for `project-b`
3. read `project-a` with `limit = 10`

The replay returns the one matching entry at cursor 1 and `next_cursor = 1`; expected `next_cursor = 2` because the stream head has been scanned and cursor 2 is only filtered out. I verified this by adding a temporary libSQL regression test locally; it fails with `left: EventCursor(1), right: EventCursor(2)`.

Please add SQL contract coverage matching `jsonl_replay_advances_next_cursor_past_trailing_filtered_records` for both libSQL/Postgres. Implementation-wise, either scan unfiltered rows and apply `ReadScope` in memory, or, when the SQL query returns fewer than `limit` matches, advance `next_cursor` to the stream head because there are no more matching rows after `after`.

crates/ironclaw_reborn_event_store/src/postgres_store.rs
```diff
@@ -0,0 +1,509 @@
...
+    .with_root_certificates(root_store)
+    .with_no_client_auth();
+    Ok(MakeRustlsConnect::new(config))
+}
+
+pub(crate) async fn build_postgres_event_stores(
+    url: SecretString,
+) -> Result<RebornEventStores, RebornEventStoreError> {
+    let raw_url = url.expose_secret();
+    let client = if is_local_postgres_url(raw_url) {
```
serrrfirat: **Medium Severity** — Postgres store is not resilient to connection drops.

The Postgres backend builds one `tokio_postgres::Client`, stores it in `Arc<Client>`, and discards the original connection config. If a managed Postgres instance closes the connection after an idle timeout, restart, failover, or transient network issue, the spawned connection task exits and all future append/read operations through this store will keep failing until the entire `RebornEventStores` graph is rebuilt.

Please use a pool/reconnect-per-operation wrapper (or retain config and recreate the client on connection failure) and add a regression test that simulates a dropped connection before a later append/read.

crates/ironclaw_reborn_event_store/src/postgres_store.rs
zmanian: Fixed in 14b74957. `is_local_postgres_url` now parses the libpq keyword form (`host=db.example.com user=...`) and only treats localhost / socket-path / missing-`host=` connections as local; everything else routes through the TLS path. Regression tests cover remote keyword strings, socket paths, and the no-`host=` libpq default.

crates/ironclaw_reborn_event_store/src/libsql_store.rs
zmanian: Fixed in 14b74957. `build_reborn_event_stores` now rejects `http://` libSQL URLs for `RebornProfile::Production` with a typed `ProductionLibsqlClearTextDisabled` error before the build call, so an auth token can no longer be sent over cleartext in production. LocalDev / Test still accept `http://` for local sqld. Two regression tests lock in both directions.

crates/ironclaw_reborn_event_store/src/libsql_store.rs
zmanian: Verified addressed: both `jsonl_replay_advances_next_cursor_past_trailing_filtered_records` and `libsql_runtime_and_audit_logs_survive_rebuild_with_filtered_cursor_semantics` (durable_event_store_contract.rs) cover the trailing-filtered-records scenario and pass on the current branch (35674f6a).

crates/ironclaw_reborn_event_store/src/postgres_store.rs
zmanian: Acknowledged. Postgres connection-drop resilience (pool / reconnect / error-class mapping) is a substantive architectural change beyond the scope of this PR — it touches every append/read path and needs to coordinate with the v1 `PostgresAdapter` reconnection story. Tracking as a follow-up; not blocking this PR's review baseline.

zmanian: Fixed in 78ee140e. `PostgresStore` now holds a `deadpool_postgres::Pool` instead of an `Arc<Client>`; `pool.get().await` is called at each operation boundary. deadpool's Manager transparently replaces Clients whose connection task has exited, so an idle timeout / server restart / transient drop no longer leaves the store permanently broken. Mirrors v1's connection-management story (`src/db/postgres.rs`, `src/db/tls.rs` already use deadpool-postgres) so the binary doesn't ship two divergent reconnection paths.

crates/ironclaw_reborn_event_store/src/postgres_store.rs
```diff
@@ -0,0 +1,568 @@
...
+    url: SecretString,
+) -> Result<RebornEventStores, RebornEventStoreError> {
+    let raw_url = url.expose_secret();
+    let pg_config: Config = raw_url.parse().map_err(|source| {
+        RebornEventStoreError::backend("postgres", "parse connection string", source)
+    })?;
+    let manager_config = ManagerConfig {
+        recycling_method: RecyclingMethod::Fast,
+    };
+    let manager = if is_local_postgres_url(raw_url) {
```
serrrfirat: **Critical Severity**

Remote PostgreSQL connections do not actually require TLS here.

Passing a rustls connector is not enough: `tokio-postgres` still honors the parsed connection config, whose default is `sslmode=prefer`, and an explicit `sslmode=disable` returns a raw stream before TLS is attempted. That means a production remote event-store URL can downgrade to plaintext even though this branch intends remote Postgres to be TLS-only.

Please force or reject insecure SSL modes for non-local configs before building the pool. For example, fail closed on `sslmode=disable` and set `pg_config.ssl_mode(tokio_postgres::config::SslMode::Require)` for every non-local target. Add factory-level regression tests for a default remote URL and `sslmode=disable`.

crates/ironclaw_reborn_event_store/src/lib.rs
```diff
@@ -0,0 +1,756 @@
...
+        .ok_or_else(|| durable_error("jsonl event cursor overflowed u64"))?;
+    let line = serialise(next_cursor)?;
+
+    file.write_all(line.as_bytes())
+        .map_err(|_| durable_error("jsonl event store failed to append record"))?;
+    file.write_all(b"\n")
+        .map_err(|_| durable_error("jsonl event store failed to append record"))?;
+    file.flush()
+        .map_err(|_| durable_error("jsonl event store failed to flush record"))?;
+    file.sync_data()
```
serrrfirat: **Medium Severity**

JSONL append fsyncs file contents but not the new directory entry.

For a brand-new stream, `create_dir_all` creates the parent path and `OpenOptions::create(true)` creates the JSONL file. `sync_data()` makes the file contents durable, but on common POSIX filesystems the parent directory must also be fsynced for the new file name/directory entry to be crash-durable. After a crash or power loss, the first append can disappear even though `append()` returned success.

Please fsync the relevant directory path(s) when a stream file or parent directory is created (at minimum the stream file parent after first creation, and any newly-created directories from `create_dir_all`). If there is an fs fault-injection harness, add a crash-consistency regression for first append to a new stream.

crates/ironclaw_reborn_event_store/src/postgres_store.rs
```diff
@@ -0,0 +1,568 @@
...
+///
+/// Conservative: an unparseable URL is treated as remote (fail closed).
+///
+/// Two connection-string forms are supported:
+/// - URL form (`postgres://host:5432/db`): parsed via `url::Url`.
+/// - libpq keyword form (`host=db.example.com user=...`): walked keyword-by-
+///   keyword. A keyword string without `host=` (libpq default) or with
+///   `host=/...` (socket directory) is a Unix-socket connection; anything
+///   else is routed to a network host and must use TLS.
+fn is_local_postgres_url(url: &str) -> bool {
```
serrrfirat: **High Severity**

The local/remote PostgreSQL detector misses valid remote libpq configurations.

This reparses the raw string and only looks at URI authority or `host=...`. Valid tokio-postgres/libpq forms like `hostaddr=10.0.0.5`, `postgresql:///db?host=db.example.com`, and mixed lists such as `host=/var/run/postgresql,db.example.com` can be classified as local and routed through `NoTls`.

Please inspect the parsed `tokio_postgres::Config` instead of heuristically reparsing the raw string. Treat the config as local only if every configured target is loopback or a Unix socket; any remote `host`, remote `hostaddr`, or mixed target list should use/require TLS. Add regressions for `hostaddr=...`, URL query-param hosts, and mixed local/remote host lists.

crates/ironclaw_reborn_event_store/src/postgres_store.rs
zmanian: Fixed in ce0c8eff. Added `enforce_remote_ssl_mode`: rejects `SslMode::Disable` for any non-local config with a typed `RemotePostgresClearTextDisabled` error, and upgrades the default `Prefer` to `Require` so the server cannot silently decline TLS. Three regression tests cover Disable rejection, Prefer→Require upgrade, and Require pass-through.

crates/ironclaw_reborn_event_store/src/lib.rs
zmanian: Fixed in ce0c8eff. `append_with_cursor_assignment` now records `is_first_create = !path.exists()` before opening, and after `sync_data()` opens the parent directory and calls `sync_all()` on it. Best-effort on platforms without directory fsync (Windows handles it implicitly).

crates/ironclaw_reborn_event_store/src/postgres_store.rs
zmanian: Fixed in ce0c8eff. Replaced raw-string reparsing with `is_local_postgres_config` that walks `Config::get_hosts()` and `Config::get_hostaddrs()`. Regression tests now cover `hostaddr=` to remote addresses, `hostaddr=` to loopback, mixed socket/TCP host lists, and URLs with empty authority + `?host=` query parameters.

crates/ironclaw_reborn_event_store/src/lib.rs
```diff
@@ -0,0 +1,756 @@
...
+                Err(RebornEventStoreError::BackendUnavailable {
+                    backend: "postgres",
+                })
+            }
+        }
+        RebornEventStoreConfig::Libsql {
+            path_or_url,
+            auth_token,
+        } => {
+            if profile == RebornProfile::Production && path_or_url.starts_with("http://") {
```
serrrfirat: **High Severity**

Production can still select libSQL's in-memory backend.

The production gate rejects `RebornEventStoreConfig::InMemory`, but `RebornEventStoreConfig::Libsql { path_or_url: ":memory:", .. }` reaches the libSQL builder and creates an ephemeral database. That bypasses the documented production guarantee that in-memory stores are rejected before returning a service graph; all runtime/audit history disappears on restart.

Please reject `:memory:` (and equivalent in-memory forms, if any) when `profile == RebornProfile::Production`, and add a profile-contract regression for this exact config.

crates/ironclaw_reborn_event_store/src/postgres_store.rs
```diff
@@ -0,0 +1,568 @@
...
+            Err(_) => return false,
+        };
+        return match parsed.host_str() {
+            // libpq treats an empty host as a Unix-socket connection.
+            None | Some("") => true,
+            Some(host) => is_local_host_literal(host),
+        };
+    }
+    let mut host: Option<&str> = None;
+    for token in url.split_ascii_whitespace() {
```
serrrfirat: **Medium Severity**

Quoted libpq socket hosts are misclassified as remote and break local Postgres.

Valid libpq keyword strings can quote values, e.g. `host='/var/run/postgresql' user=ironclaw dbname=ironclaw`. The detector splits on ASCII whitespace and never unquotes, so it sees the value as `'/var/run/postgresql'` (not starting with `/`) and selects the TLS connector. Tokio-postgres then tries a Unix-socket connection with TLS and fails because there is no hostname for the TLS handshake.

Please stop reparsing the raw DSN and inspect the parsed `tokio_postgres::Config` hosts/hostaddrs instead. Add regressions for quoted Unix-socket keyword strings.

crates/ironclaw_reborn_event_store/src/libsql_store.rs
```diff
@@ -0,0 +1,550 @@
...
+                })?;
+            }
+        }
+        libsql::Builder::new_local(path_or_url).build().await
+    };
+    db.map(Arc::new)
+        .map_err(|source| RebornEventStoreError::backend("libsql", "connect", source))
+}
+
+fn is_remote_libsql(path_or_url: &str) -> bool {
```
serrrfirat, line 110: **High Severity**

libSQL remote/local detection is case-sensitive and can silently create a local file.

The production cleartext check and `is_remote_libsql()` both use raw, case-sensitive `starts_with(...)`. Valid mixed-case URL schemes such as `HTTPS://...`, `HTTP://...`, or `LibSQL://...` skip the remote branch and are passed to `Builder::new_local(...)`, which can create a node-local SQLite path like `HTTPS:/host/...` instead of failing closed or using the intended remote durable store.

Please parse URLs with a real URL parser or normalize the scheme case before policy checks, and fail closed for unsupported/ambiguous forms. Add regressions for mixed-case `http`, `https`, and `libsql` schemes.

crates/ironclaw_reborn_event_store/src/postgres_store.rs
```diff
@@ -0,0 +1,568 @@
...
+        }
+        query.push_str(&format!(" ORDER BY cursor ASC LIMIT ${}", params.len() + 1));
+        params.push(&limit_i64);
+        let rows = client
+            .query(&query, &params)
+            .await
+            .map_err(|_| durable_error("postgres event store failed to read entries"))?;
+        let mut entries = Vec::new();
+        let mut last_scanned: Option<EventCursor> = None;
+        for row in rows {
```
serrrfirat, line 425: **Medium Severity**

SQL replay never validates cursor contiguity.

If an entry row disappears or `reborn_event_streams` drifts out of sync with `reborn_event_entries`, this loop returns later cursors instead of surfacing a replay gap/corruption error. JSONL fails closed on the same invariant by checking every cursor is the expected next value; SQL currently turns a missing row into silent history loss.

Please validate that the first returned row is `after + 1` (or that retained history explicitly starts later) and that subsequent rows are contiguous up to the scanned head, or use a query that detects missing cursors. Add corruption tests that delete one entry row and expect `ReplayGap`/durable-log failure in both SQL backends.

crates/ironclaw_reborn_event_store/src/lib.rs
```diff
@@ -0,0 +1,756 @@
...
+            })
+        }
+        RebornEventStoreConfig::Jsonl {
+            root,
+            accept_single_node_durable,
+        } => {
+            if profile == RebornProfile::Production && !accept_single_node_durable {
+                return Err(RebornEventStoreError::ProductionJsonlRequiresAcceptance);
+            }
+            tokio::fs::create_dir_all(&root)
```
serrrfirat: **Medium Severity**

JSONL stores are created with default filesystem permissions.

Production may explicitly accept JSONL as a single-node durable backend, but `create_dir_all` and `OpenOptions::create(true)` never tighten directory/file modes. With a typical Unix `umask 022`, the tenant/user directory names and event/audit JSONL contents can be world-listable/readable by other local accounts.

Please create JSONL roots/directories/files with restrictive permissions (`0700` dirs, `0600` files on Unix, or equivalent platform handling) before writing sensitive durable history. Add a Unix permissions regression if supported by the test harness.

crates/ironclaw_reborn_event_store/src/lib.rs
```diff
@@ -0,0 +1,756 @@
...
+        stream: &EventStreamKey,
+        filter: &ReadScope,
+        after: Option<EventCursor>,
+        limit: usize,
+    ) -> Result<EventReplay<RuntimeEvent>, EventError> {
+        let owned_filter = filter.clone();
+        self.store
+            .read_after(
+                StreamKind::Runtime,
+                stream,
```
serrrfirat, line 295: **Medium Severity**

Replay filtering cannot enforce invocation scope.

The event-stream contract says deeper-scope filtering includes invocation, and `ResourceScope` carries an `invocation_id`, but `ReadScope` stops at project/mission/thread/process. Any per-invocation consumer sharing the same `(tenant,user,agent)` stream has no way to ask the backend to enforce that boundary; it must either over-receive sibling invocation records or use a broader cursor that can skip invocation-local history.

Please add `invocation_id` to `ReadScope`, include it in SQL metadata/indexes/filters, and apply it in JSONL/in-memory `matches_event`/`matches_audit` where applicable. Add replay tests with two invocations in one stream.

crates/ironclaw_reborn_event_store/src/lib.rs
```diff
@@ -0,0 +1,756 @@
...
+        let lock = Arc::new(Mutex::new(()));
+        locks.insert(key, Arc::downgrade(&lock));
+        lock
+    }
+
+    fn stream_path(&self, kind: StreamKind, stream: &EventStreamKey) -> PathBuf {
+        let mut path = self
+            .root
+            .join(kind.directory())
+            .join(component("tenant", stream.tenant_id.as_str()))
```
serrrfirat, line 475: **High Severity**

JSONL stream isolation can collapse on case-insensitive filesystems.

`EventStreamKey` treats tenant/user/agent IDs as exact strings, and host scope IDs preserve case. The JSONL backend maps percent-encoded raw IDs directly into path components, so `Alice` and `alice` produce distinct strings but the same path on common macOS/Windows-style case-insensitive volumes. That can merge runtime/audit streams for distinct valid users/tenants/agents.

Please make path components collision-resistant independent of filesystem case behavior, e.g. use a fixed-case hash/base32 digest of the raw ID plus an optional human-readable hint. Add a unit test around the component mapper to prove case-distinct IDs map to case-distinct/collision-resistant filenames.

crates/ironclaw_reborn_event_store/src/libsql_store.rs
serrrfirat, line 110: **High Severity**

A bare libSQL remote endpoint is treated as a local database file.

`is_remote_libsql()` only recognizes explicit schemes, so a production config like `path_or_url = "db.example.com"` goes down the local path and creates `./db.example.com`, silently ignoring the auth token and stranding durable event/audit history on one node. This is an easy typo/misconfiguration for remote libSQL endpoints.

Please require local file paths to be explicit and reject hostname-like/bare remote values in production, or canonicalize through libSQL's remote URL parser before choosing local storage. Add a regression that `db.example.com` with an auth token fails closed instead of creating a local file.

crates/ironclaw_reborn_event_store/src/lib.rs
```diff
@@ -0,0 +1,756 @@
...
+}
+
+/// JSONL-backed durable runtime event log.
+#[derive(Clone)]
+pub struct JsonlDurableEventLog {
+    store: JsonlStore,
+}
+
+impl JsonlDurableEventLog {
+    pub fn new(root: impl Into<PathBuf>) -> Self {
```
serrrfirat: **Medium Severity**

The public JSONL constructors bypass the production acceptance gate.

`build_reborn_event_stores(Production, Jsonl { accept_single_node_durable: false, .. })` correctly fails closed, but production code can still wire `Arc::new(JsonlDurableEventLog::new(path))` / `JsonlDurableAuditLog::new(path)` directly into a `DurableEventSink`/`DurableAuditSink`. That makes the single-node durability acceptance policy unenforceable by the crate API.

Please make these constructors crate-private/test-only, or require an explicit accepted configuration token produced by the factory. Add a small API/compile-time guard or architecture test so production composition cannot bypass the factory.

crates/ironclaw_reborn_event_store/src/lib.rs
```diff
@@ -0,0 +1,756 @@
...
+
+    // Re-read the prior tail under the lock so we observe writes from any
+    // other process that just finished appending.
+    let prior_tail = read_last_jsonl_cursor(path)?.unwrap_or(0);
+    let next_cursor = prior_tail
+        .checked_add(1)
+        .ok_or_else(|| durable_error("jsonl event cursor overflowed u64"))?;
+    let line = serialise(next_cursor)?;
+
+    file.write_all(line.as_bytes())
```
serrrfirat: **High Severity**

A partial JSONL append can permanently wedge the stream.

This writes directly to the live JSONL file and then relies on the last line being parseable to assign the next cursor. If `write_all(line.as_bytes())` returns an error after a short/partial write (ENOSPC, interrupted storage, etc.), the file is left with malformed JSON at EOF. Every later append calls `read_last_jsonl_cursor()` and fails on that torn tail, and replay fails once it reaches the malformed line. One transient write failure can therefore require manual file surgery for an accepted production backend.

Please make append failure atomic from the stream's perspective: record the pre-append length and truncate back on any write/flush/sync error, or use a temp/rename or recovery marker strategy that never leaves an invalid tail as the current stream head. Add a fault-injection test for a short write/torn tail.

crates/ironclaw_reborn_event_store/src/lib.rs
```diff
@@ -0,0 +1,756 @@
...
+        stream.user_id.as_str(),
+        stream
+            .agent_id
+            .as_ref()
+            .map(AgentId::as_str)
+            .unwrap_or("<none>")
+    )
+}
+
+fn component(prefix: &str, value: &str) -> String {
```
serrrfirat, line 842: **Medium Severity**

Valid scope IDs can exceed filesystem filename limits in the JSONL layout.

Host API scope IDs allow up to 256 bytes. The JSONL mapper prefixes those raw IDs into a single component (`tenant-` + 256 bytes = 263 bytes, `agent-id-...jsonl` is longer), which exceeds common 255-byte filename limits. A legitimate tenant/user/agent ID can therefore make append/replay fail with an I/O error.

Please map IDs to bounded-length components (for example a fixed-length hash/base32 digest with an optional truncated hint) and add tests using max-length valid IDs.

crates/ironclaw_reborn_event_store/src/libsql_store.rs
```diff
@@ -0,0 +1,550 @@
...
+                .as_ref()
+                .map(|token| token.expose_secret().to_string())
+                .unwrap_or_default(),
+        )
+        .build()
+        .await
+    } else {
+        if path_or_url != ":memory:" {
+            let path = Path::new(path_or_url);
+            if let Some(parent) = path.parent() {
```
serrrfirat: **Medium Severity**

Local libSQL paths without a directory component fail to initialize.

For a common config like `path_or_url = "events.db"`, `Path::parent()` returns an empty parent path and this calls `create_dir_all("")`, which fails. The caller has to know to write `./events.db` even though both names should be equivalent local DB files.

Please skip parent creation when the parent is empty, and add a regression for `events.db` in the current working directory.

crates/ironclaw_reborn_event_store/src/postgres_store.rs
serrrfirat: **Medium Severity**

Loopback Postgres endpoints cannot opt into TLS.

Once this helper classifies a DSN as local, the builder always uses `NoTls` and ignores an explicit `sslmode=require`. That breaks valid local deployments that require TLS on localhost/127.0.0.1 (for example TLS-only local Postgres or a loopback proxy/forwarder), with no escape hatch for the user-requested TLS mode.

Please honor the parsed SSL mode independently of locality: local endpoints may allow `NoTls` by default, but `sslmode=require` should select the TLS connector. Add a regression for `postgres://localhost/db?sslmode=require`.

crates/ironclaw_reborn_event_store/src/postgres_store.rs
serrrfirat: **High Severity**

Whitespace-valid libpq keyword hosts can bypass remote TLS detection.

Tokio-postgres/libpq accepts keyword strings with whitespace around `=`, such as `host = db.internal user=...` or `host= db.internal user=...`. This detector splits on whitespace and only notices tokens starting with `host=`, so those valid remote DSNs are treated as having no/empty host and routed through `NoTls`.

Please use the parsed `tokio_postgres::Config` rather than tokenizing the raw string. Add regressions for keyword strings with whitespace around `=`.

crates/ironclaw_reborn_event_store/src/lib.rs
```diff
@@ -0,0 +1,756 @@
...
+    where
+        T: Clone + DeserializeOwned + Send + 'static,
+    {
+        if limit == 0 {
+            return Err(EventError::InvalidReplayRequest {
+                reason: "limit must be greater than zero".to_string(),
+            });
+        }
+        let after = after.unwrap_or_default();
+        // We hold the in-process stream lock while we *read* purely so that
```
serrrfirat, line 435: **Medium Severity**

Large JSONL replays block live appends on the same stream.

`read_after_cursor()` holds the per-stream Tokio mutex for the entire blocking scan and `stream_read_after()` holds a shared file lock for the lifetime of the read. A large or filter-sparse reconnect/history replay can therefore block appenders in-process via the mutex and cross-process via shared-vs-exclusive file-lock conflict until the replay completes, stalling live runtime/audit writes for that tenant/user/agent.

Please avoid holding writer-blocking locks across full-file scans. For example, use append-safe reads that do not block exclusive append for the whole replay, paginate by offsets/snapshotted file length, or otherwise keep lock hold time bounded. Add a concurrency test where a slow/sparse replay does not block append beyond a small bound.

crates/ironclaw_reborn_event_store/src/libsql_store.rs
```diff
@@ -0,0 +1,550 @@
...
+    fn new(db: Arc<Database>) -> Self {
+        Self { db }
+    }
+
+    async fn connect(&self) -> Result<Connection, EventError> {
+        let conn = self
+            .db
+            .connect()
+            .map_err(|_| durable_error("libsql event store failed to connect"))?;
+        conn.query("PRAGMA busy_timeout = 5000", ())
```
serrrfirat: **High Severity**

File-backed libSQL is left in rollback-journal mode, so replay readers can block event/audit appends.

Production explicitly allows local libSQL, but the connection setup only configures `busy_timeout` and foreign keys. Without `PRAGMA journal_mode=WAL` for file-backed databases, a long replay reader can block writer commits; once it exceeds the 5s busy timeout, durable append starts failing system-wide for that local store.

Please enable WAL for file-backed local libSQL stores during initialization (and decide the synchronous mode intentionally), while avoiding WAL for `:memory:`/remote where inappropriate. Add a concurrent reader/writer regression for local libSQL.

crates/ironclaw_reborn_event_store/src/libsql_store.rs
```diff
@@ -0,0 +1,550 @@
...
+struct LibSqlStore {
+    db: Arc<Database>,
+}
+
+impl LibSqlStore {
+    fn new(db: Arc<Database>) -> Self {
+        Self { db }
+    }
+
+    async fn connect(&self) -> Result<Connection, EventError> {
```
serrrfirat, line 131: **Medium Severity**

The libSQL adapter opens a fresh connection per operation without the retry handling used by the main backend.

Every append/read calls `connect()`, but transient concurrent local-open failures such as “unable to open database file” are mapped directly to durable-log failures. The repo's main libSQL backend already has retry handling for this class of concurrent-open flake; dropping it here can turn ordinary concurrent event traffic into spurious append/replay failures.

Please reuse the established libSQL connection retry/backoff pattern or keep a pooled/shared connection strategy for local stores. Add a concurrent append/read stress regression for local libSQL.

zmanian:
Round 3 review feedback addressed in 49853962.

**Fixed in code (each with regression test):**

- **#34** — Production rejects `:memory:` libSQL via `ProductionInMemoryDisabled` (test: `production_rejects_in_memory_libsql_target`).
- **#36, #41** — Case-insensitive scheme detection for `HTTP://` / `HTTPS://` / `LibSQL://` plus a new `ProductionLibsqlAmbiguousTarget` error rejecting bare hostnames (`db.example.com`) in Production. LocalDev still allows ergonomic forms like `events.db`. Tests: `production_rejects_mixed_case_cleartext_libsql_url`, `production_accepts_mixed_case_secure_libsql_scheme`, `production_rejects_bare_hostname_libsql_target`, `case_insensitive_remote_scheme_detection`, `local_dev_still_allows_bare_relative_libsql_path`.
- **#37** — SQL replay now runs an unfiltered `COUNT(*)` over the scanned cursor window in both libSQL and Postgres backends; a missing entry row surfaces as `EventError::ReplayGap`. JSONL already detects this via line-by-line cursor sequencing. Test: `libsql_replay_detects_missing_entry_row_as_replay_gap` (deletes a middle entry row directly through libsql, asserts ReplayGap on next replay).
- **#38** — JSONL directories created with `0o700`, stream files opened with `0o600` on Unix via `tokio::fs::DirBuilder::mode` and `OpenOptions::mode`. Test: `jsonl_root_directory_uses_restrictive_permissions`.
- **#40, #44** — Path components reformatted as `{prefix}-{16hex SHA-256}-{32-byte URL-encoded hint}`, so `Alice` and `alice` are case-distinct on case-insensitive filesystems and a 256-byte scope ID maps to a component well under the 255-byte filename limit. Tests: `case_distinct_ids_map_to_distinct_components`, `long_ids_map_to_filename_safe_components`.
- **#42** — Removed the public `JsonlDurableEventLog::new` / `JsonlDurableAuditLog::new` constructors. Production composition must now go through `build_reborn_event_stores`, so the `Jsonl { accept_single_node_durable: true }` gate cannot be bypassed by directly wrapping a JSONL log in a `DurableEventSink`.
- **#43** — JSONL append now snapshots the file length before writing and calls `set_len()` to roll back on any write/flush/sync error, so a partial write can never leave a torn JSON tail that wedges every later append. Test: `jsonl_append_truncates_on_serialiser_failure`.
- **#45** — `build_database` skips `create_dir_all` when `Path::parent()` is empty, so `events.db` (CWD-relative) works without forcing callers to write `./events.db`. Covered by `local_dev_still_allows_bare_relative_libsql_path`.
- **#46** — `build_postgres_event_stores` now consults `Config::ssl_mode()` independently of locality: a loopback config with `sslmode=require` routes through the rustls connector instead of being forced to `NoTls`.
- **#49** — `LibSqlStore::initialize_pragmas` runs `PRAGMA journal_mode=WAL` + `PRAGMA synchronous=NORMAL` once at build time for file-backed local stores. Remote and `:memory:` skip the pragma. WAL mode is persisted in the file header so every subsequent `connect()` inherits it without per-call overhead.
- **#50** — `LibSqlStore::connect` now mirrors the v1 retry pattern (`src/db/libsql/mod.rs::connect`): three attempts with exponential backoff (50ms, 100ms) on transient \"unable to open database file\" errors, so concurrent local-open flake no longer surfaces as durable-log failures.

**Verified already addressed by the round-2 `Config::get_hosts/get_hostaddrs` switch (added confirming regression tests):**

- **#35** — libpq quoted single-quote socket path (`host='/var/run/postgresql' …`) classifies as local through the parsed `Host::Unix` entry. Test: `libpq_quoted_socket_path_is_local`.
- **#47** — libpq keyword strings with whitespace around `=` (`host = db.internal …`, `host=db.internal …`, `host =db.internal …`) classify as remote correctly. Test: `libpq_whitespace_around_equals_classifies_remote_correctly`.

**Deferred with explicit acknowledgement (mirrors the prior #26 precedent):**

- **#39** — `ReadScope.invocation_id`. Module-level doc note in `crates/ironclaw_reborn_event_store/src/lib.rs`. Adding it requires changes in `ironclaw_events::ReadScope` (cross-crate), the SQL schemas/indexes, the JSONL/in-memory `matches_event` / `matches_audit` predicates, and every replay caller — substantive enough to warrant its own PR. Tracking as a follow-up.
- **#48** — Replay holds writer-blocking locks across full-file scans. Inline note at the JSONL `read_after`. The fix (capture EOF offset, drop locks, scan up to the snapshot) is a substantive concurrency redesign that needs to coordinate with the durable-log contract — tracking as a follow-up.

**Test plan:**
- `cargo test -p ironclaw_reborn_event_store`
- `cargo test -p ironclaw_reborn_event_store --features \"libsql postgres\"`
- `cargo test -p ironclaw_host_runtime`
- `cargo test -p ironclaw_events`
- `cargo clippy -p ironclaw_reborn_event_store --all-targets -- -D warnings`
- `cargo clippy -p ironclaw_reborn_event_store --features \"libsql postgres\" --all-targets -- -D warnings`
- `cargo clippy -p ironclaw_reborn_event_store --no-default-features --all-targets -- -D warnings`
- `cargo clippy -p ironclaw_host_runtime --all-targets -- -D warnings`

PR #3177, opened 2026-05-01
feat(cli): add `ironclaw insights` for usage analytics

abbyshekit:
## Summary

New `ironclaw insights [--days N] [--json]` subcommand that surfaces existing `agent_jobs`, `routine_runs`, and `job_actions` aggregates without log-scraping. Hermes parity for `hermes insights`.

- `--days` defaults to 30, capped at 90 with a stderr warning above the cap. `0` is treated as "default window".
- Default output is a human-readable table; `--json` emits a versioned, additive payload (`version: 1`) for dashboards.
- Both libSQL and Postgres backends ship `aggregate_insights` impls; queries are bounded by `agent_jobs.created_at >= since` and joined with `job_actions` for tool frequency, so the window is enforced consistently.

## Scope

First commit. Out of scope, called out as follow-ups in the second commit message:
- USD cost rollup (uses existing `src/llm/costs.rs` and `estimation_snapshots`)
- `--source` / `--user-id` filters
- `idx_routine_runs_created_at` migrations on libSQL + Postgres (perf — flagged by review)
- Tighten the SQL window to a closed `[since, until)` interval to keep clock-skew or future-dated rows out (correctness — flagged by review)

## Test plan

- [x] `cargo check --tests` clean
- [x] `cargo test --test insights_integration` — 5 tests pass: in-window job aggregation, JSON shape, empty DB, time-window exclusion, routine_runs window
- [x] `cargo test --lib cli::insights::tests` — 5 unit tests pass: `resolve_window` clamping (default/cap/in-range/at-boundary) plus JSON empty-state shape
- [ ] Manual: run \`ironclaw insights\` and \`ironclaw insights --json --days 7\` against a populated \`~/.ironclaw/ironclaw.db\`

## Why this matters

NEAR Foundation pilots (5 currently) need adoption visibility — operators are tailing logs to track token usage today. KPI: 181 → 500 daily active agents by quarter end.

### DIFF
--- a/migrations/V26__routine_runs_created_at_index.sql
+++ b/migrations/V26__routine_runs_created_at_index.sql
@@ -1,0 +1,18 @@
+-- Add index on routine_runs.created_at to speed up the time-window
+-- aggregation used by `ironclaw insights`.
+--
+-- V6 created indexes on routine_id and status but not created_at, so the
+-- insights query (`SELECT COUNT(*) FROM routine_runs WHERE created_at >=
+-- $1 AND created_at < $2`) was a full-table scan on deployments with
+-- meaningful routine_runs history. Mirrors the pattern of V24, which
+-- added the same kind of index to llm_calls for the admin usage summary.
+--
+-- WARNING: This takes a full table lock on routine_runs. For large
+-- deployments with millions of rows, this could block writes for several
+-- seconds to minutes. Refinery wraps migrations in a transaction, which
+-- prevents using CREATE INDEX CONCURRENTLY. If this is a concern, apply
+-- the index manually outside the migration framework first:
+--
+--   CREATE INDEX CONCURRENTLY IF NOT EXISTS idx_routine_runs_created
+--       ON routine_runs(created_at);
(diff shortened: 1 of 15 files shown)


abbyshekit:
Pushed `f59fbe9` — both Codex-flagged correctness/perf BLOCKs from the deferred-list are now resolved on this branch:

- **`[since, until)` window** — `InsightsStore` trait now takes an explicit `until: DateTime<Utc>` parameter, plumbed through libSQL + Postgres backends. CLI pins `until = Utc::now()` once at the call site so all four sub-queries see the same upper bound. New regression test `time_window_excludes_future_rows` seeds a future-dated job + routine_run + actions and asserts none leak in.
- **`routine_runs(created_at)` index** — added to libSQL `SCHEMA` (fresh DBs), libSQL `INCREMENTAL_MIGRATIONS` entry #26 (existing DBs), and Postgres `V26__routine_runs_created_at_index.sql` (mirrors V24's `llm_calls` pattern, same CREATE INDEX CONCURRENTLY caveat in the header).

Tests: 11 pass (6 integration including the new regression, 5 unit). Local cargo check + cargo test green.

PR #3178, opened 2026-05-01
feat(cli): add `ironclaw backup --quick` for portable state snapshots

abbyshekit:
## Summary

New `ironclaw backup --quick` subcommand: bundles `~/.ironclaw/ironclaw.db` + `~/.ironclaw/config.toml` + a JSON `manifest.json` into a single zip ready to scp between hosts. First half of the migration story (companion `ironclaw import` lands in a follow-up PR).

- WAL-checkpoints the libSQL database before copy so the snapshot is self-contained without `-wal` / `-shm` sidecars.
- Streams writes (no whole-archive in-memory buffer) and atomic-writes (\`<output>.tmp\` + rename) so a crashed backup can't leave a corrupt archive at the canonical path.
- `--label STRING` baked into the manifest for migration provenance (e.g. \`--label pre-crab-shack\`).
- Default output: \`~/ironclaw-backup-<ISO8601-with---instead-of-:>.zip\`. Override with \`--output PATH\`.

## Manifest

\`\`\`json
{
  "ironclaw_version": "<CARGO_PKG_VERSION>",
  "schema_version": <MAX(version) FROM _migrations>,
  "created_at": "<ISO8601 UTC>",
  "hostname": "<host>",
  "label": "<--label or null>",
  "mode": "quick",
  "components": ["db","config"]
}
\`\`\`

Manifest is pinned at zip index 0 so consumers can validate the archive before paying the deflate cost on the (potentially large) db blob.

## Scope

Quick mode only. Out of scope for this PR (tracked as follow-ups):
- \`--full\` mode (skill bundles, installed extensions, secrets via the encrypted store)
- \`ironclaw import\`
- Windows path handling (Linux + macOS only here)
- Archive-level encryption
- VACUUM INTO snapshot (race-free file capture during concurrent writes)

## Notes for reviewers

- Honors the existing \`~/.ironclaw/\` base dir convention (\`compute_ironclaw_base_dir()\` and \`default_libsql_path()\`) rather than XDG dirs, so the backup actually finds the user's files.
- Resolves config path via \`compute_ironclaw_base_dir()\` (env-reading) instead of \`Settings::default_toml_path()\` (LazyLock-cached). Aligns with the env-respecting behavior of \`LIBSQL_PATH\` for the db path. Fix landed in commit 3 after the test suite caught the LazyLock-vs-test-isolation issue.
- A second commit applies fixes from a Codex review pass (WAL checkpoint failure now propagates, schema_version errors propagate, missing-config now fails loudly instead of silently trimming the manifest, manifest pinned at zip index 0).
- A third commit fixes two real bugs the review missed but the test suite caught: \`PRAGMA wal_checkpoint(TRUNCATE)\` returns rows so it must use \`query()\` not \`execute()\`, plus the LazyLock config-path fix above.

## Test plan

- [x] \`cargo check --tests\` clean
- [x] \`cargo test --test backup_integration\` — 3 tests pass: zip is valid + entries are exactly \`manifest.json\`, \`data/ironclaw.db\`, \`config.toml\`; default output stamp parses as ISO8601; \`--label\` propagates and missing-db produces a clear error
- [x] \`cargo test --lib cli::backup::tests\` — 3 unit tests pass: manifest serialization, ISO8601 path stamp, extension suffix handling
- [ ] Manual: \`ironclaw backup --quick --label test\`, unzip, restore via \`sqlite3 .restore\` on a clean host

## Why this matters

NEAR Foundation pilots are migrating from Railway to Crab Shack. Without backup/import, every migration is manual sqlite + scp. The intended workflow becomes \`ironclaw backup --quick && scp && ironclaw import\` (import lands in a follow-up).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3298,6 +3298,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "hostname"
+version = "0.4.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "617aaa3557aef3810a6369d0a99fac8a080891b68bd9f9812a1eeda0c0730cbd"
+dependencies = [
+ "cfg-if",
+ "libc",
+ "windows-link",
+]
+
+[[package]]
 name = "html-escape"
 version = "0.2.13"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -3902,6 +3913,7 @@ dependencies = [
  "hex",
  "hkdf",
  "hmac",
+ "hostname",
  "html-to-markdown-rs",
  "http-body-util",
  "hyper 1.8.1",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -187,6 +187,12 @@ zip = { version = "2", default-features = false, features = ["deflate"] }
 # production backup path can own a TempDir for the snapshot lifecycle.
 tempfile = "3"
 
+# Hostname lookup for the backup manifest. Calls `gethostname(2)` on
+# Unix and `GetComputerNameW` on Windows; macOS doesn't set $HOSTNAME
+# by default so the env-var fallback alone produced "unknown" on most
+# Apple machines.
+hostname = "0.4"
+
 # HTTP proxy for sandboxed network access
 hyper = { version = "1.5", features = ["server", "http1", "http2"] }
 hyper-util = { version = "0.1", features = ["server", "tokio", "http1", "http2"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -182,6 +182,11 @@ tar = "0.4"
 pdf-extract = "0.7"
 zip = { version = "2", default-features = false, features = ["deflate"] }
 
+# Temp directory for `ironclaw backup`'s VACUUM INTO snapshot. Already
+# present under [dev-dependencies] for tests; promoted here so the
+# production backup path can own a TempDir for the snapshot lifecycle.
+tempfile = "3"
+
 # HTTP proxy for sandboxed network access
 hyper = { version = "1.5", features = ["server", "http1", "http2"] }
 hyper-util = { version = "0.1", features = ["server", "tokio", "http1", "http2"] }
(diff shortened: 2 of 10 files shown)


abbyshekit:
Pushed VACUUM INTO race-free snapshot per the Codex follow-up list. `snapshot_db()` runs `VACUUM INTO '<tempdir>/ironclaw_snapshot.db'` before the zip writer opens; the temp file is owned by a `tempfile::TempDir` whose Drop cleans up after `write_quick_archive()` finishes. Removes the prior `wal_checkpoint + file-copy` race window.

Test assertion changed: byte-equality of the zip's db blob vs the source file no longer holds (VACUUM INTO compacts), so the test now opens the restored snapshot, runs `SELECT version, name FROM _migrations`, and asserts the seeded row matches — a correctness check, not a byte check.

`tempfile = "3"` promoted from [dev-dependencies] to [dependencies] for the production path. 6/6 tests pass.

abbyshekit:
Pushed the two remaining Codex follow-ups from this branch's deferred list:

- **Hostname** — switched to `gethostname(2)` via the `hostname` crate (Unix) / `GetComputerNameW` (Windows). `$HOSTNAME` is honored first so test environments / CI can pin a deterministic value, but on macOS (where the env var isn't set by default) the manifest now contains the real machine name instead of `unknown`.
- **Unique `.tmp` suffix** — the atomic-rename staging path was a fixed `<output>.tmp`. Two concurrent `ironclaw backup` invocations to the same canonical path could rename-step on each other. Now `<output>.<pid>.<8-rand-alphanum>.tmp`, so each invocation owns a unique staging file. Trailing `.tmp` preserved so any external sweeper that cleans `.tmp` files keeps working. Uses the `rand` crate already in dependencies.

6/6 tests still green. With this, every Codex-flagged item on this PR is resolved on-branch. Open follow-ups (per the PR description) remain: `--full` mode, `ironclaw import`, archive-level encryption, Windows path normalization, the `tokio::sync::Mutex` in tests clippy nit.

PR #3180, opened 2026-05-01
refactor(reborn-memory): native-isolated guardrails + lib.rs module split (PR 1 of #3118)

### LINKED ISSUE #3118: Build native Reborn memory storage/search service
## Summary

Build a Reborn-native memory storage/search service instead of adapting Reborn memory to the legacy `src/workspace::Workspace` implementation.

This issue supersedes the old adapter direction in #3112. The desired outcome is a cleanly isolated Reborn memory subsystem whose source of truth lives in `crates/ironclaw_memory`, with explicit tenant/user/agent/project scope columns in its own DB tables and no dependency on `src/workspace` or synthetic workspace user IDs.

## Decision record

We are intentionally **not** building a host-side `WorkspaceMemoryAdapter` over `src/workspace::Workspace`.

Reasons:

- Synthetic workspace identities such as `tenant:{tenant}:user:{user}:project:{project}` are stringly typed and can confuse lifecycle, deletion, export, audit, and authorization boundaries.
- Product code with direct `Workspace` access could bypass Reborn policy if it can construct the encoded workspace user id.
- Project/tenant isolation should be inspectable and enforceable at the Reborn schema/repository layer, not hidden inside legacy `user_id` encoding.
- `ironclaw_memory` should be a clean Reborn boundary crate, not an adapter shim around app-level workspace behavior.

Migration/coexistence of existing legacy workspace memory rows is **explicitly deferred**. This issue should build the Reborn-native substrate and tests only. Existing `src/workspace` memory remains separate until a later migration/coexistence issue defines the product mapping for legacy rows.

## Current state

Already landed Reborn memory work:

- #3078 added memory document storage/filesystem boundaries.
- #3079 added memory search/index/plugin seams.
- #3114 added vertical substrate coverage through public Reborn seams.

Relevant existing files:

- `crates/ironclaw_memory/src/lib.rs`
  - currently contains most memory code in one large file.
  - includes `MemoryDocumentScope`, `MemoryDocumentPath`, `MemoryBackend`, `MemoryBackendFilesystemAdapter`, `MemoryDocumentRepository`, `MemoryDocumentIndexRepository`, `RepositoryMemoryBackend`, `ChunkingMemoryDocumentIndexer`, libSQL/Postgres repository code, metadata/search/versioning helpers.
- `crates/ironclaw_memory/tests/db_memory_repository_contract.rs`
- `crates/ironclaw_memory/tests/memory_backend_contract.rs`
- `crates/ironclaw_memory/tests/memory_filesystem_contract.rs`
- `crates/ironclaw_memory/tests/memory_filesystem_vertical_integration.rs`
- `crates/ironclaw_memory/CLAUDE.md`
- Legacy reference implementation only, not a dependency target:
  - `src/workspace/*`
  - `src/db/libsql/workspace.rs`
  - `src/db/postgres` workspace-related code, if present
  - existing `migrations/V1__initial.sql`, `V16__document_versions.sql`, `V23__list_workspace_files_escape_like.sql`

## Non-goals

Do not do these in this issue:

- Do not add `WorkspaceMemoryAdapter` or any bridge adapter over `src/workspace::Workspace`.
- Do not make `ironclaw_memory` depend on the main app crate, `src/workspace`, runtime, workflow, secrets, network, process, approvals, or product tool modules.
- Do not encode Reborn tenant/user/project scope into legacy `memory_documents.user_id`.
- Do not migrate legacy `memory_documents` rows into Reborn tables.
- Do not replace product tools (`memory_read`, `memory_write`, `memory_search`, `memory_tree`) yet.
- Do not remove legacy workspace memory code yet.

## What to build

Build a Reborn-native memory persistence/search implementation with its own schema, repositories, and tests.

The target architecture is:

```text
host/runtime/tools
  -> Reborn authorization / capability checks
  -> RootFilesystem or MemoryBackend
  -> crates/ironclaw_memory
  -> Reborn-native libSQL/Postgres repositories
  -> reborn_memory_* tables
```

The architecture must not be:

```text
Reborn memory
  -> host bridge adapter
  -> src/workspace::Workspace
  -> legacy memory_documents with encoded synthetic user_id
```

## Schema requirements

Add Reborn-native memory tables for both libSQL and Postgres:

- `reborn_memory_documents`
- `reborn_memory_chunks`
- `reborn_memory_chunks_fts`
- `reborn_memory_document_versions`

Use explicit Reborn scope columns on document/chunk/search/version operations:

```text
tenant_id TEXT NOT NULL
user_id TEXT NOT NULL
agent_id TEXT NOT NULL DEFAULT ''
project_id TEXT NOT NULL DEFAULT ''
path TEXT NOT NULL
```

Use the empty string as the DB-only absent sentinel for `agent_id` and `project_id`. This is safe because `MemoryDocumentScope` rejects empty supplied IDs. Do not store `_none` in the DB; `_none` remains only a virtual-path sentinel.

Document uniqueness must be:

```text
UNIQUE (tenant_id, user_id, agent_id, project_id, path)
```

Every read/list/search/write/version/chunk query must include the full scope:

```text
tenant_id = ?
user_id = ?
agent_id = ?
project_id = ?
```

Do not query by path prefixes to infer project scope.

### Phase 1: Remove adapter direction from local work and docs

- Ensure there is no `src/bridge/memory_workspace_adapter.rs`.
- Ensure no test named `tests/reborn_memory_workspace_adapter.rs` is added.
- Ensure root `Cargo.toml` does not add `ironclaw_memory`, `ironclaw_filesystem`, or `ironclaw_host_api` solely for a workspace adapter.
- Update `crates/ironclaw_memory/CLAUDE.md`:
  - remove or rewrite adapter-first wording.
  - state that Reborn memory is native and isolated.
  - state that pure behavior/tests may be ported from `src/workspace`, but `ironclaw_memory` must not depend on `src/workspace`.
  - state that legacy migration/coexistence is deferred.

### Phase 2: Make `ironclaw_memory` agent-friendly before deeper changes

Split `crates/ironclaw_memory/src/lib.rs` into focused modules without changing behavior.

Suggested structure:

```text
crates/ironclaw_memory/src/lib.rs                     # re-export hub
crates/ironclaw_memory/src/path.rs                    # scope/path grammar and validation
crates/ironclaw_memory/src/metadata.rs                # DocumentMetadata, HygieneMetadata, .config helpers
crates/ironclaw_memory/src/backend.rs                 # MemoryBackend, MemoryContext, capabilities, backend adapters
crates/ironclaw_memory/src/filesystem.rs              # MemoryBackendFilesystemAdapter, MemoryDocumentFilesystem
crates/ironclaw_memory/src/search.rs                  # MemorySearchRequest/Result, fusion config, escaping helpers
crates/ironclaw_memory/src/chunking.rs                # chunk_document, ChunkConfig, hashes
crates/ironclaw_memory/src/versioning.rs              # version row helpers/types
crates/ironclaw_memory/src/schema.rs                  # schema validation helpers
crates/ironclaw_memory/src/repo/mod.rs                # repository trait exports/shared helpers
crates/ironclaw_memory/src/repo/in_memory.rs          # in-memory repository
crates/ironclaw_memory/src/repo/libsql.rs             # libSQL repository
crates/ironclaw_memory/src/repo/postgres.rs           # Postgres repository
```

Keep public API compatibility for existing tests where possible. This phase should be behavior-preserving.

### Phase 3: Add native schema/migrations

Add migrations for libSQL and Postgres that create the `reborn_memory_*` tables.

The schema must include:

- explicit tenant/user/agent/project columns.
- document content and metadata.
- content hash for version/index currentness.
- updated/created timestamps.
- chunk rows with chunk index, content, content hash/currentness, optional embedding.
- FTS search support equivalent to the current libSQL/Postgres capabilities.
- document versions with monotonically increasing version number per document.
- cascade deletes from documents to chunks/versions.

Add indexes for:

- document lookup by full scope + path.
- list/search by full scope.
- chunk lookup by document id.
- FTS lookup where applicable.
- version lookup by document id + version.

### Phase 4: Rebuild libSQL repository on native schema

Update `LibSqlMemoryDocumentRepository` to use `reborn_memory_*` tables instead of legacy workspace tables.

Required behavior:

- write document by full Reborn scope and relative path.
- read document by full scope and path.
- list documents only within exact scope.
- reject file/directory prefix conflicts under transaction.
- upsert same document path in same scope atomically.
- preserve top-level paths such as `projects/local-note.md` as user-visible paths; do not treat them as project scope.
- write metadata.
- read inherited `.config` metadata.
- respect `skip_indexing`.
- respect `skip_versioning`.
- create document versions on overwrite when versioning is enabled.
- compute and store content SHA consistently with existing Reborn helper semantics.
- replace chunks only if the source document is still current.
- perform full-text search with escaped queries.
- never leak rows from a different tenant/user/agent/project.
- report committed document writes as success even if derived indexing refresh fails after persistence.

### Phase 5: Rebuild Postgres repository on native schema

Update `PostgresMemoryDocumentRepository` to match libSQL semantics against `reborn_memory_*` tables.

Required behavior mirrors Phase 4.

Do not leave Postgres at compile-only coverage. Add behavioral tests using the repo's existing Postgres integration-test pattern. If the repo lacks an ephemeral Postgres harness, add one using the established project testcontainer/deadpool pattern.

### Phase 6: Port pure workspace behavior tests, not workspace dependencies

Use `src/workspace` and `src/db/libsql/workspace.rs` as reference material only.

Port tests/behavior for:

- `.config` metadata inheritance precedence.
- `skip_indexing`.
- `skip_versioning`.
- schema validation success/failure.
- FTS query escaping.
- special-character FTS queries.
- empty/stopword-ish search queries.
- vector dimension mismatch.
- embedding provider failure after persistence.
- RRF fusion.
- weighted-score fusion.
- deterministic tie-breaking.
- `limit` and `pre_fusion_limit` behavior.
- version hash semantics.

If any behavior intentionally differs from legacy workspace memory, document the difference in `crates/ironclaw_memory/CLAUDE.md` and in the relevant Reborn memory docs.

### Phase 7: Wire Reborn memory service through public seams

Add host/app composition only after native repositories pass contract tests.

The composition should construct something equivalent to:

```text
LibSqlMemoryDocumentRepository or PostgresMemoryDocumentRepository
  -> ChunkingMemoryDocumentIndexer
  -> RepositoryMemoryBackend
  -> MemoryBackendFilesystemAdapter
  -> RootFilesystem mount at /memory
```

Add tests through caller-facing Reborn seams:

- authorized memory read/write/search succeeds.
- denied or unsupported memory operation fails before backend side effects.
- scoped mount cannot escape tenant/user/agent/project.
- search respects full scope.
- errors do not expose raw DB/provider internals to runtime output.

Do not port product memory tools in this issue unless all previous phases are done and reviewed.

## Required tests

Add or update tests so these behaviors are covered.

### Scope isolation

- Same `path` in different tenants stores separate rows.
- Same `path` in different users stores separate rows.
- Same `path` in different agents stores separate rows.
- Same `path` in different projects stores separate rows.
- Search for one scope never returns another scope's row.
- Listing one scope never returns another scope's row.
- A top-level user path beginning with `projects/` remains a normal path and does not affect project scoping.

### DB write behavior

- Same full scope + path overwrites/upserts one document.
- Different scope + same path creates different documents.
- File-vs-directory prefix conflict is rejected at repository/DB layer, not only filesystem precheck.
- Concurrent writes do not produce duplicate same-scope rows.
- Post-write indexer failure after document persistence does not make the document write report failure.

### Metadata/config behavior

- `.config` applies to child documents.
- closer `.config` overrides parent `.config`.
- document metadata overrides inherited `.config` where applicable.
- `skip_indexing: true` prevents chunk/search update.
- `skip_versioning: true` prevents version row creation on overwrite.
- schema validation rejects invalid content before persistence when schema requires it.

### Search/index behavior

- write -> chunk -> FTS search works end-to-end.
- FTS query escaping handles quotes, punctuation, operators, wildcard-like characters, and empty/stopword-ish queries.
- vector dimension mismatch fails safely before corrupting index rows.
- hybrid search ranking is deterministic.
- RRF and weighted-score fusion match documented behavior.
- `limit` and `pre_fusion_limit` are honored.

### Versioning behavior

- overwrite creates a version row for the previous content when versioning is enabled.
- version numbers are monotonic per document.
- version content hash matches the content being versioned.
- no version row is written when `skip_versioning` is inherited or set.

### Public seam behavior

- `MemoryBackendFilesystemAdapter` can read/write/list/search through native repo.
- `RootFilesystem` mounted at `/memory` enforces Reborn path scope.
- Unsupported backend capabilities fail closed before backend side effects.

## Acceptance criteria

- [ ] No `WorkspaceMemoryAdapter` or equivalent `src/workspace::Workspace` wrapper is introduced.
- [ ] `ironclaw_memory` has no dependency on the main app crate or product runtime/workflow/secrets/network/process/approval/tool modules.
- [ ] Reborn memory persistence uses explicit tenant/user/agent/project DB columns, not synthetic workspace `user_id` strings.
- [ ] Native libSQL repository passes behavioral contract tests.
- [ ] Native Postgres repository passes behavioral contract tests, not compile-only tests.
- [ ] Full-scope isolation is proven for read/write/list/search.
- [ ] Metadata inheritance, `skip_indexing`, `skip_versioning`, schema validation, FTS escaping, chunking, versioning, and search fusion have tests.
- [ ] File/directory prefix conflicts are enforced by repository/DB transactions.
- [ ] Derived index refresh failure after committed persistence does not make writes appear failed.
- [ ] Public Reborn seams (`MemoryBackendFilesystemAdapter` / `RootFilesystem`) have vertical integration coverage against native repositories.
- [ ] `crates/ironclaw_memory/CLAUDE.md` and relevant Reborn docs describe the native-isolated design and explicitly defer legacy migration/coexistence.
- [ ] Existing product memory tools are not switched unless compatibility tests are added and reviewed.

## Verification commands

Run the narrowest checks while developing, then run the full gate before marking complete.

Minimum expected final checks:

```bash
cargo fmt --all -- --check
cargo test -p ironclaw_memory --features libsql
cargo test -p ironclaw_memory --all-features
cargo clippy -p ironclaw_memory --all-targets --all-features -- -D warnings
cargo test -p ironclaw_architecture reborn_crate_dependency_boundaries_hold
python3.11 scripts/check_no_panics.py --base origin/reborn-integration --head HEAD
git diff --check origin/reborn-integration
```

If Postgres tests require an explicit feature or environment variable, document and run that command as part of the PR description.

## Suggested PR breakdown

This issue is large. Implement as a stack of small PRs rather than one giant PR:

1. Behavior-preserving `ironclaw_memory` module split.
2. Native schema/migration addition plus empty repository compile wiring.
3. Native libSQL repository behavior.
4. Native Postgres repository behavior and test harness.
5. Ported semantic/search/versioning tests.
6. Host/runtime public-seam composition tests.
7. Optional product-tool migration only after explicit review.

Each PR should be independently testable and should keep behavior changes scoped.

## Notes for reviewers

Review with special attention to:

- accidental dependency from `ironclaw_memory` to product/app crates.
- scope leakage across tenant/user/agent/project.
- SQL queries that omit one of the scope columns.
- SQL path-prefix tricks used for project scope.
- NULL uniqueness bugs; absent agent/project should use the DB-only empty-string sentinel.
- indexing failures being reported as document write failures after persistence.
- Postgres behavior being meaningfully tested rather than assumed from libSQL.

nickpismenkov:
## Summary

Full Reborn memory substrate for #3118, collapsed into a single PR. Originally opened as the first of a 7-PR stack, then PRs 2–6 (#3181–#3185) were squash-merged into this branch on May 4 — the rest of the stack is closed and the work lives here.

What lands with this PR:

1. **Native-isolated guardrails** (`crates/ironclaw_memory/CLAUDE.md`) — full `(tenant, user, agent, project, path)` scope on every operation, empty-string DB sentinel for absent agent/project, `_none` is virtual-path-only, legacy migration deferred, Postgres needs real (not compile-only) coverage.
2. **`lib.rs` module split** — the 5,560-line monolith becomes a 47-line re-export hub plus 13 focused modules (`backend`, `chunking`, `embedding`, `filesystem`, `indexer`, `metadata`, `path`, `repo/{mod,in_memory,libsql,postgres,native_libsql,native_postgres}`, `safety`, `schema`, `search`). Public API surface preserved.
3. **Prompt-write safety policy** (`safety.rs`, 824 LoC) — `PromptWriteSafetyPolicy` trait, `DefaultPromptWriteSafetyPolicy`, `PromptProtectedPathRegistry` (covers `SOUL.md`, `AGENTS.md`, `USER.md`, `IDENTITY.md`, `MEMORY.md`, `HEARTBEAT.md`, …), versioned policy + sanitized reason codes + redacted event sink. Wired through `RepositoryMemoryBackend`, `MemoryBackendFilesystemAdapter`, and `MemoryDocumentFilesystem`. Defense-in-depth: any warn/bypass outcome that persists requires a durable audit sink before persistence.
4. **Native `reborn_memory_*` schema** — `reborn_memory_documents`, `reborn_memory_chunks` (with generated `tsvector` + HNSW pgvector index), `reborn_memory_document_versions`. Migrations are idempotent; no legacy `memory_documents` row migration.
5. **Native libSQL repository** (`native_libsql.rs`, 1,037 LoC) — full read/write/list/search/version/chunk on the new schema. FTS via `MATCH`, vector search via byte-encoded embeddings + cosine similarity. Transactional via `BEGIN IMMEDIATE`.
6. **Native Postgres repository** (`native_postgres.rs`, 805 LoC) — same surface on Postgres + pgvector. Row-level locking via `LOCK TABLE ... IN SHARE ROW EXCLUSIVE MODE` + `FOR UPDATE`.
7. **`RepositoryMemoryBackend` scope guards** — `ensure_path_matches_context` / `ensure_scope_matches_context` fail closed when a caller of the public seam passes a path or scope that doesn't match the authorized `MemoryContext`.
8. **Search fusion fixes** — `min_score` filter applied after normalization (so default RRF `k=60` doesn't drop every result); `with_limit()` re-clamps `pre_fusion_limit` so the invariant holds regardless of builder call order; tied scores break deterministically by path ascending.
9. **Race-safe chunk clearing** — both `skip_indexing` and empty-chunks paths in the indexer route through `replace_document_chunks_if_current(path, hash, &[])`. The unguarded `delete_document_chunks` API was removed entirely from the trait + all 4 impls.
10. **Five test files** — schema contract (libSQL + Postgres smoke), native libSQL repo contract, native Postgres repo contract, `RepositoryMemoryBackend` behavior contract, and `CompositeRootFilesystem` vertical integration.

## Test posture

Postgres tests fail loud by default. Set `IRONCLAW_SKIP_POSTGRES_TESTS=1` to opt into skipping when no DB is available — the previous "silent skip + green pass" pattern violated the `ironclaw_memory` guardrail that Postgres coverage must be real.

## Test plan

- [x] `cargo clippy --workspace --all-features --tests` — zero warnings
- [x] `cargo test -p ironclaw_memory --all-features` — all libSQL tests pass
- [x] End-to-end against `pgvector/pgvector:pg16` container with `DATABASE_URL=postgres://postgres:postgres@localhost:5432/ironclaw_test` — all 142 tests pass (109 unit/lib + ~33 contract/integration)
- [x] Verified fail-loud-without-DB (panics with actionable message) and `IRONCLAW_SKIP_POSTGRES_TESTS=1` opt-in (clean skip)
- [x] `cargo test -p ironclaw_architecture reborn_crate_dependency_boundaries_hold` — pass
- [x] `python3 scripts/check_no_panics.py --base origin/reborn-integration --head HEAD` — clean (no panics in changed production code)

## Out of scope

- PR 7 (product-tool migration) is not included; it remains a separate decision after this lands.
- No `WorkspaceMemoryAdapter` is introduced. No legacy `memory_documents` rows are migrated.


crates/ironclaw_memory/src/search.rs
```diff
@@ -0,0 +1,291 @@
...
+        .into_values()
+        .map(|result| MemorySearchResult {
+            path: result.path,
+            score: result.score,
+            snippet: result.snippet,
+            full_text_rank: result.full_text_rank,
+            vector_rank: result.vector_rank,
+        })
+        .collect::<Vec<_>>();
+    if request.min_score() > 0.0 {
```
serrrfirat: **Medium Severity** — `min_score` filters raw RRF/weighted scores before normalization.

`min_score` is applied before the scores are normalized. With the default RRF settings (`rrf_k = 60`), even the top single-source result has raw score `1 / 61 ≈ 0.016`, so a caller using `with_min_score(0.5)` will drop every result before the later normalization step would make the best result `1.0`. The existing workspace search fusion normalizes first and then applies `min_score`, so this can produce unexpectedly empty memory-search results.

Suggested fix: move the `min_score` filtering block after the normalization loop, and add a parity regression test for RRF/weighted `min_score` behavior.

crates/ironclaw_memory/src/indexer.rs
```diff
@@ -0,0 +1,155 @@
...
+        let metadata = resolve_document_metadata(self.repository.as_ref(), path).await?;
+        if metadata.skip_indexing == Some(true) {
+            return self.repository.delete_document_chunks(path).await;
+        }
+        let content_hash_at_read = content_sha256(content);
+        let chunk_texts = chunk_document(content, self.chunk_config.clone());
+        let chunks =
+            build_chunk_writes(path, chunk_texts, self.embedding_provider.as_deref()).await?;
+        if chunks.is_empty() {
+            self.repository.delete_document_chunks(path).await
```
serrrfirat: **Medium Severity** — Empty/whitespace documents can delete fresh chunks from a concurrent write.

When chunking produces no chunks, this path calls `delete_document_chunks(path)` without the content-hash guard used by `replace_document_chunks_if_current`. A plausible race is: indexer A reads a whitespace document and computes an empty chunk list; writer/indexer B then writes searchable content and replaces chunks; A resumes and unconditionally deletes B's fresh chunk rows.

Suggested fix: route empty chunk sets through the same hash-checked replacement path, e.g. `replace_document_chunks_if_current(path, expected_content_hash, &[])`, or add a hash-checked delete API.

crates/ironclaw_memory/src/backend.rs
```diff
@@ -0,0 +1,308 @@
...
+    fn capabilities(&self) -> MemoryBackendCapabilities {
+        self.capabilities.clone()
+    }
+
+    async fn read_document(
+        &self,
+        _context: &MemoryContext,
+        path: &MemoryDocumentPath,
+    ) -> Result<Option<Vec<u8>>, FilesystemError> {
+        self.repository.read_document(path).await
```
serrrfirat, line 367: **Medium Severity** — Repository backend trusts the supplied path/scope instead of enforcing the authorized `MemoryContext`.

`MemoryBackend` receives a host-resolved `MemoryContext`, but the repository-backed implementation ignores it and forwards the separately supplied `MemoryDocumentPath`/`MemoryDocumentScope` to the repository. The filesystem adapter currently passes matching values, but this is a public backend seam intended for host/runtime use; a direct caller that authorizes one context but passes a path or scope for another tenant/project would bypass the context boundary.

Suggested fix: fail closed unless `path.scope() == context.scope()` for read/write and `scope == context.scope()` for list/search before any repository side effects.

zmanian (commented):
**Verdict: APPROVE-WITH-NITS** — well-structured split, scope guards and hash-CAS index refresh are correctly designed and tested; SQL is parameterized; issues are real but not catastrophic.

## Spot-checked and correct

- **No SQL injection.** Every `native_libsql`/`native_postgres` call uses bound params (`?N` / `$N`). The only `format!` in `native_postgres` is `pg_advisory_lock($1)` — also parameterized.
- **Scope equality is typed.** `MemoryDocumentScope` is `(String, String, Option<String>, Option<String>)` with derived `PartialEq` (`path.rs:9`); the `agent_id="" vs "None"` collision can't occur because comparison happens at the typed Rust level *before* the repo's `_db_value` sentinel translation. `_none` is explicitly rejected at construction (`path.rs:37–53`).
- **Hash-CAS race is sound.** `ChunkingMemoryDocumentIndexer::reindex_document` (`indexer.rs:97–110`) computes `content_hash_at_read` and passes it to `replace_document_chunks_if_current`, which inside one transaction selects `content_hash FOR UPDATE` (postgres `native_postgres.rs:430–458`) / `BEGIN IMMEDIATE` + check (`native_libsql.rs:444–459`) and bails if it has moved. A stale reindex degrades to a no-op rather than wiping fresh chunks. Contract test exists on both backends.
- **Search fusion fix correct.** `fuse_memory_search_results` (`search.rs:255–279`) normalizes max-to-1 *before* applying `min_score`, with regression tests at `search.rs:378` and `search.rs:403`. `with_limit` re-clamps `pre_fusion_limit` (`search.rs:65`).
- **Tied-score determinism.** Tiebreaker is path-asc (`search.rs:285`), with *two* tests proving it isn't insertion-order coincidence (`search.rs:336`, `357`).
- **Vector encoding** (libsql) is explicit LE (`to_le_bytes`/`from_le_bytes`, `embedding.rs:106–124`). `cosine_similarity` returns `None` on dim mismatch, empty inputs, zero norms, and any non-finite result (NaN/Inf).
- **Search-time embedding dim check.** `backend.rs:513–527` fails fast on caller-supplied embedding dim ≠ provider dim.
- **API surface preservation.** `lib.rs` re-exports cover the previously-public surface; the new safety / Reborn types are additive.

## High

- **H1. Trigger `update_reborn_memory_documents_updated_at` is dropped/recreated on every `start()`** (`native_postgres.rs:765–769`). `pg_advisory_lock(REBORN_MIGRATION_LOCK_ID)` at line 55 serializes parallel `start()` calls — but only if `start()` is the only place this batch runs. Worth a comment at the schema block stating the advisory-lock invariant, plus an explicit test that two parallel `RebornPostgresMemoryDocumentRepository::new(...).start()` calls succeed.
- **H2. ⚠️ `LOCK TABLE … IN SHARE ROW EXCLUSIVE` is table-level** (`native_postgres.rs:175`). Every tenant's writes serialize through one global lock. Combined with the per-chunk `INSERT` loop in `replace_document_chunks_if_current` running under a separate transaction, write throughput will fall over fast under multi-tenant load. The unique constraint already prevents the actual data race; the prefix check is best-effort. Recommend dropping the `LOCK TABLE` and relying on `FOR UPDATE` + `ON CONFLICT`, or scoping with `pg_advisory_xact_lock(hashtext(scope_id))` per-scope. Flag for follow-up before any prod load.
- **H3. libsql vector search is full-table-scan-in-Rust** (`native_libsql.rs:807–940`) — selects every chunk in scope, loads every blob into Rust, computes cosine in a loop, then sorts. No index. Fine for single user with 10k chunks; multi-tenant trap. Probably acceptable since libsql doesn't have a vector index, but please add a TODO so this isn't mistaken for the postgres-class hot path.

## Medium

- **M1.** `PromptProtectedPathRegistry::default()` matches by lowercased exact relative path (`safety.rs:125–137`), not basename. If the runtime ever writes `agents/AGENTS.md` or `_user/USER.md`, those would not match. Confirm with the author whether the contract is "exact relative path only" (fine) or "any path whose basename is one of these" (insufficient).
- **M2. `PromptProtectedPathClass::as_str()` always returns `"system_prompt_file"`** regardless of which path matched (`safety.rs:62–64`). The "stable protected-path class" is currently a single bucket — telemetry can't distinguish AGENTS.md vs SOUL.md vs HEARTBEAT.md events. Either remove `as_str` or have it return distinct stable strings per class.
- **M3.** Audit-sink-required-before-persistence is enforced for `Warn`/`BypassAllowed` (`safety.rs:643, 671, 750–763`). `Reject` and `Allow` paths use `require_sink: false`, so a successful protected-path write with `event_sink: None` leaves no audit trail. That's the documented contract, but worth re-examining: a successful write to AGENTS.md without a sink is silent. Recommend either making the sink mandatory whenever a protected_path_class is matched, or documenting this as an explicit operator opt-in.
- **M4.** `PromptWriteSafetyEvent` carries the full `MemoryDocumentScope` (`safety.rs:766`) — tenant/user/agent/project IDs. Validated by `validated_memory_segment` so they can't carry arbitrary content, but they are user-derived. Confirm sink consumers are OK with that.
- **M5.** `tsvector` hardwired to `'english'` (`native_postgres.rs:777`). Non-English content gets degraded full-text relevance. Fine for v1 if all current users are English; worth a TODO/issue link.
- **M6.** HNSW `m=16, ef_construction=64` are pgvector's typical defaults — fine. Search-time `ef_search` is unset, so pgvector uses whatever session default applies. Call out the recall/build-cost tradeoff at the schema or in docs.

## Low / nits

- **L1.** `safety.rs:469` — `request.content.trim().is_empty()` rejects content that is e.g. `"   \n"`. Probably intended; worth a one-line comment.
- **L2.** `safety.rs:30` — validates trimmed-empty version but stores the un-trimmed value. Inconsistent.
- **L3.** `native_libsql.rs:209–214` and `:345` hand-format `strftime('%Y-%m-%dT%H:%M:%fZ', 'now')` — hoist into a constant.
- **L4.** `native_libsql.rs` mixes raw `BEGIN IMMEDIATE`/`COMMIT`/`ROLLBACK` strings (140–268) with `transaction_with_behavior` (433). Pick one style.
- **L5.** `PromptSafetyAllowanceId::empty_prompt_file_clear()` (`safety.rs:219`) hardcodes a string here and matches against `==` at `safety.rs:471`. A single `const` would be clearer.

## Test gaps worth filling

1. Concurrent `replace_document_chunks_if_current` with same `expected_content_hash` and different chunk sets — one wins, one no-ops.
2. Embedding dimension mismatch at *write/index* time (search-time check exists, write-time doesn't have a test).
3. Safety-policy bypass paths: (a) `BypassAllowed` for `empty_prompt_file_clear` actually persisting; (b) `require_sink: true` with `event_sink: None` returning `PromptWriteSafetyEventUnavailable` and *not* writing the document.
4. Policy version bump (v1 client vs v2 writer reconciliation).
5. tsvector with non-English (e.g. CJK) content — smoke test that write doesn't fail.
6. Idempotent `start()` called twice in one process *and* from two processes (postgres advisory-lock contract).
7. End-to-end search exercising `pre_fusion_limit` clamp (>50 candidates).

## Bottom line

Ship it after addressing **H2** (the table-level `LOCK TABLE` is too coarse for multi-tenant — at minimum file a follow-up issue before any prod traffic) and **M2** (`PromptProtectedPathClass::as_str` is degenerate as written). Everything else is fine for follow-up. The split itself is clean, the trait surface is right, and the scope guards close a real hole.

crates/ironclaw_memory/src/repo/native_libsql.rs
```diff
@@ -0,0 +1,1037 @@
...
+            .map_err(|error| memory_error(path.clone(), operation, error.to_string()))?;
+        conn.query("PRAGMA busy_timeout = 5000", ())
+            .await
+            .map_err(|error| memory_error(path, operation, error.to_string()))?;
+        Ok(conn)
+    }
+}
+
+#[async_trait]
+impl MemoryDocumentRepository for RebornLibSqlMemoryDocumentRepository {
```
serrrfirat, line 92: **High Severity**

Reborn-native repositories currently do not support atomic append.

`RepositoryMemoryBackend::compare_and_append_document()` and both filesystem adapters call `MemoryDocumentRepository::compare_and_append_document_with_options()` for `/memory` appends, but the new Reborn libSQL/Postgres repository impls never override that method. As a result, any append routed to a native repo falls back to the trait default (`"memory document repository does not support atomic append"`) instead of appending. This breaks a public `RootFilesystem::append_file` seam as soon as the native repositories are selected.

Please implement the same optimistic append contract for both `RebornLibSqlMemoryDocumentRepository` and `RebornPostgresMemoryDocumentRepository` (including hash conflict checks, versioning/options, path-conflict checks for new rows, and transaction/locking), and add native backend/filesystem tests that drive `append_file` through the caller.

crates/ironclaw_memory/src/repo/native_libsql.rs
```diff
@@ -0,0 +1,1037 @@
...
+        })?;
+        Ok(Some(content.into_bytes()))
+    }
+
+    async fn write_document(
+        &self,
+        path: &MemoryDocumentPath,
+        bytes: &[u8],
+    ) -> Result<(), FilesystemError> {
+        self.write_document_with_options(path, bytes, &MemoryWriteOptions::default())
```
serrrfirat: **Medium Severity**

Direct native repository writes lose version actor attribution.

`write_document()` forwards `MemoryWriteOptions::default()`, and the native version archival path stores `options.changed_by`. Any overwrite performed through the public repository seam therefore creates `reborn_memory_document_versions.changed_by = NULL`. The legacy libSQL/Postgres direct `write_document()` implementations attributed versions to the scoped owner key, so this is a parity regression for repository-level callers and weakens audit/version history even though higher-level backend/filesystem writes pass `changed_by` explicitly.

Please either make native direct `write_document()` populate the scoped owner key (matching legacy behavior) or explicitly remove/avoid the direct-write versioning contract; in either case add assertions for `changed_by` in the native libSQL and Postgres version tests.

crates/ironclaw_memory/src/repo/native_postgres.rs
```diff
@@ -0,0 +1,805 @@
...
+    EXECUTE FUNCTION reborn_memory_documents_set_updated_at();
+
+CREATE TABLE IF NOT EXISTS reborn_memory_chunks (
+    id UUID PRIMARY KEY DEFAULT gen_random_uuid(),
+    document_id UUID NOT NULL REFERENCES reborn_memory_documents(id) ON DELETE CASCADE,
+    chunk_index INT NOT NULL,
+    content TEXT NOT NULL,
+    content_hash TEXT NOT NULL,
+    content_tsv TSVECTOR GENERATED ALWAYS AS (to_tsvector('english', content)) STORED,
+    embedding VECTOR(1536),
```
serrrfirat: **High Severity**

The native Postgres schema hard-codes embeddings to 1536 dimensions.

`ChunkingMemoryDocumentIndexer` accepts whatever dimension the configured `EmbeddingProvider` declares, and this repository inserts that vector directly. But `reborn_memory_chunks.embedding` is declared as `VECTOR(1536)`, so Reborn Postgres will fail indexing/searching for common non-1536 providers (for example 768/1024-dim Ollama models or 3072-dim OpenAI embeddings). Because backend writes treat indexer failures as best-effort, a write can appear successful while no chunks are indexed and semantic search later misses the document or errors on query vectors.

Please make the native Postgres substrate dimension-configurable/unbounded in line with the existing Postgres memory contract (V9 changed `memory_chunks.embedding` to unbounded `vector` specifically for provider flexibility), or otherwise validate provider dimension up front and fail closed before advertising vector/embedding support. Add a Postgres native test with a non-1536 embedding provider so this cannot regress.

crates/ironclaw_memory/src/repo/native_postgres.rs
```diff
@@ -0,0 +1,1084 @@
...
+            })?;
+        // Short-circuit if the schema is already present. Re-running the full
+        // DDL on every `run_migrations` call would take an AccessExclusiveLock
+        // on `reborn_memory_documents` for the FK validation in the dependent
+        // `CREATE TABLE IF NOT EXISTS reborn_memory_chunks`, which would
+        // deadlock against concurrent writers holding RowExclusiveLock once
+        // the global write-side LOCK TABLE was removed (zmanian H1+H2).
+        // The advisory lock above still serializes the *first* migration
+        // across all processes; subsequent calls find the schema and skip.
+        let already_migrated = client
```
serrrfirat: **High Severity**

`run_migrations()` treats the presence of `reborn_memory_documents` as proof that the whole native Postgres schema is migrated.

If a database is partially provisioned, or a future PR adds another table/index/trigger to `REBORN_POSTGRES_MEMORY_DOCUMENTS_SCHEMA`, subsequent startups will skip the batch entirely once this one table exists. Startup can then succeed while chunking/search/versioning paths fail later at runtime because their objects were never created.

Please make the migration path idempotent per schema object/version instead of using a single table-existence gate, and add a partial-schema regression test where `reborn_memory_documents` exists but a dependent object is missing.

serrrfirat: **Medium Severity**

The migration advisory lock can be leaked if the post-lock `information_schema` probe errors.

After `pg_advisory_lock()` succeeds, the `query_opt(...).await.map_err(...)?` path returns before the best-effort unlock block. With pooled connections, that session can retain the advisory lock and block future migrations from other connections/processes.

Please guarantee unlock on every post-lock return path, or switch to a transaction-scoped advisory lock/RAII guard, and add a regression around an injected probe/batch failure.

crates/ironclaw_memory/src/filesystem.rs
```diff
@@ -0,0 +1,765 @@
...
+        let is_protected = prompt_write_protected_classification(
+            self.prompt_safety_policy.as_ref(),
+            &self.prompt_protected_path_registry,
+            &document_path,
+        )
+        .is_some();
+        let backend_capabilities = self.backend.capabilities();
+        let adapter_should_enforce_prompt_safety = is_protected
+            && (!backend_capabilities.prompt_write_safety || self.prompt_safety_config_overridden);
+        let prompt_safety_allowance = if is_protected || backend_capabilities.prompt_write_safety {
```
serrrfirat: **Medium Severity**

The one-shot prompt-write-safety allowance can be consumed by an unrelated unprotected write.

`with_one_shot_prompt_write_safety_allowance()` is documented as applying to the next protected write only, but this branch calls `take_prompt_safety_allowance()` whenever the backend advertises `prompt_write_safety`, even if `is_protected` is false. A normal `/memory` write can therefore burn the allowance before the intended protected-path write, making the protected operation fail unexpectedly.

Please only consume the allowance for protected paths, and add a test that performs an unprotected write before the protected write that should use the allowance.

crates/ironclaw_memory/src/backend.rs
```diff
@@ -0,0 +1,616 @@
...
+        {
+            let embedding = embed_text(provider.as_ref(), context.scope(), request.query()).await?;
+            request = request.with_query_embedding(embedding);
+        }
+
+        // Fail-fast on caller-supplied embeddings whose dimension disagrees with the
+        // configured provider, instead of silently producing no/wrong results downstream
+        // (libsql cosine_similarity skips mismatched chunks; postgres pgvector errors
+        // opaquely).
+        if let (Some(provider), Some(embedding)) =
```
serrrfirat: **Medium Severity**

Full-text-only fallback searches still validate a stale query embedding.

This dimension check runs whenever a provider and `query_embedding` are present, even when `request.vector()` has been turned off. A caller that disables vector search for fallback/retry but reuses a request with an old embedding can get a dimension error instead of a normal full-text search.

Please gate the provider-dimension validation on `request.vector()`, and add a regression where `vector=false`, `full_text=true`, and a stale mismatched embedding is ignored.

crates/ironclaw_memory/src/repo/native_postgres.rs
```diff
@@ -0,0 +1,1084 @@
...
+    let limit = request.pre_fusion_limit() as i64;
+    let query_vector = pgvector::Vector::from(query_embedding.to_vec());
+    let rows = client
+        .query(
+            "SELECT c.id, d.tenant_id, d.user_id, d.agent_id, d.project_id, d.path, c.content \
+             FROM reborn_memory_chunks c \
+             JOIN reborn_memory_documents d ON d.id = c.document_id \
+             WHERE d.tenant_id = $1 AND d.user_id = $2 AND d.agent_id = $3 \
+               AND d.project_id = $4 \
+               AND c.embedding IS NOT NULL \
```
serrrfirat, line 1015: **Medium Severity**

Native Postgres vector search does not filter stored embedding dimensions before applying the pgvector distance operator.

The schema now uses an unbounded `vector`, so a scope can contain chunks produced by different embedding providers/dimensions. `ORDER BY c.embedding <=> $5` will error on mixed dimensions and fail the whole search, rather than skipping incompatible chunks the way the libSQL cosine path does.

Please filter by matching dimensions (for example `vector_dims(c.embedding) = vector_dims($5)`) or store/query dimension metadata, and add a mixed-dimension Postgres regression test.

crates/ironclaw_memory/src/repo/native_libsql.rs
```diff
@@ -0,0 +1,1190 @@
...
+    content_hash TEXT NOT NULL,
+    embedding BLOB,
+    created_at TEXT NOT NULL DEFAULT (strftime('%Y-%m-%dT%H:%M:%fZ', 'now')),
+    UNIQUE (document_id, chunk_index)
+);
+
+CREATE INDEX IF NOT EXISTS idx_reborn_memory_chunks_document
+    ON reborn_memory_chunks(document_id);
+
+CREATE VIRTUAL TABLE IF NOT EXISTS reborn_memory_chunks_fts USING fts5(
```
serrrfirat, line 1292: **Medium Severity**

The libSQL FTS external-content table is created idempotently but never rebuilt/backfilled.

If a database reaches a partial repair state where `reborn_memory_chunks` already has rows before `reborn_memory_chunks_fts` or its triggers exist, rerunning migrations creates the FTS table/triggers but leaves existing chunks unindexed until every document is rewritten/reindexed. Full-text search will silently miss those chunks.

Please rebuild/backfill the FTS table during migration repair (or version the migration), and add a test that creates chunks first, reruns migrations, and verifies existing chunks are searchable.

crates/ironclaw_memory/src/repo/native_postgres.rs
```diff
@@ -0,0 +1,1084 @@
...
+    let rows = client
+        .query(
+            "SELECT c.id, d.tenant_id, d.user_id, d.agent_id, d.project_id, d.path, c.content, \
+                    ts_rank_cd(c.content_tsv, plainto_tsquery('english', $5)) AS rank \
+             FROM reborn_memory_chunks c \
+             JOIN reborn_memory_documents d ON d.id = c.document_id \
+             WHERE d.tenant_id = $1 AND d.user_id = $2 AND d.agent_id = $3 \
+               AND d.project_id = $4 \
+               AND c.content_tsv @@ plainto_tsquery('english', $5) \
+             ORDER BY rank DESC \
```
serrrfirat: **Medium Severity**

Branch ranking is unstable for tied search scores, and those unstable ranks feed reciprocal-rank fusion.

The native search branches order only by score/distance (`ORDER BY rank DESC` here, and similar branch ordering in vector/libSQL paths). When the DB returns tied rows in arbitrary order, the row index becomes the reciprocal rank, changing fused scores and potentially final ordering across identical data. The final path tiebreak in `search.rs` runs after the unstable branch ranks have already changed the scores.

Please add deterministic branch tie-breakers (for example path/chunk index/id) before assigning ranks, plus a tied-score regression.

crates/ironclaw_memory/src/backend.rs
```diff
@@ -0,0 +1,616 @@
...
+        {
+            return Err(memory_backend_unsupported(
+                context.scope(),
+                FilesystemOperation::ReadFile,
+                "memory backend does not support search",
+            ));
+        }
+
+        let mut request = request;
+        if request.vector()
```
serrrfirat, line 575: **Medium Severity**

`embeddings=false` does not prevent backend search from calling the embedding provider.

If vector search is enabled, no query embedding is supplied, and a provider is configured, this path calls `embed_text(...)` without checking `self.capabilities.embeddings`. A backend intended to accept caller-supplied vectors but forbid local embedding generation can still incur provider cost/egress/secret use.

Please require `self.capabilities.embeddings` before invoking the provider, and fail closed when vector search needs generated embeddings but embedding generation is disabled.

crates/ironclaw_memory/src/backend.rs
```diff
@@ -0,0 +1,616 @@
...
+#[async_trait]
+impl<R> MemoryBackend for RepositoryMemoryBackend<R>
+where
+    R: MemoryDocumentRepository + 'static,
+{
+    fn capabilities(&self) -> MemoryBackendCapabilities {
+        self.capabilities.clone()
+    }
+
+    async fn read_document(
```
serrrfirat, line 356: **Medium Severity**

Direct `RepositoryMemoryBackend` file operations ignore `file_documents=false`.

The filesystem adapter checks `capabilities().file_documents`, but direct public backend calls still read/write/append/list documents regardless of that capability. A host using the backend seam directly can perform file-document side effects even though the backend declares file documents unsupported.

Please add `file_documents` fail-closed checks to `read_document`, `write_document`, `compare_and_append_document`, and `list_documents`, and drive the regression through the backend methods rather than only the adapter.

crates/ironclaw_memory/src/indexer.rs
```diff
@@ -0,0 +1,311 @@
...
+        let metadata = resolve_document_metadata(self.repository.as_ref(), path).await?;
+        if metadata.skip_indexing == Some(true) {
+            return self
+                .repository
+                .replace_document_chunks_if_current(path, &content_hash_at_read, &[])
+                .await;
+        }
+        let chunk_texts = chunk_document(content, self.chunk_config.clone());
+        let chunks =
+            build_chunk_writes(path, chunk_texts, self.embedding_provider.as_deref()).await?;
```
serrrfirat: **Medium Severity**

Zero-chunk reindexing can leave stale chunks searchable when an embedding provider rejects empty batches.

Whitespace-only content produces `chunk_texts = []`, but `build_chunk_writes(..., embedding_provider)` still calls `provider.embed_batch(&[])`. Providers that reject empty batches return before `replace_document_chunks_if_current(path, hash, &[])` runs, and callers intentionally swallow indexer errors after persisting the document. Old chunks can therefore remain searchable after a document is cleared.

Please bypass embedding generation for empty chunk sets and always issue the hash-guarded empty replacement, with a regression using a provider that rejects empty batches.

crates/ironclaw_memory/src/filesystem.rs
```diff
@@ -0,0 +1,765 @@
...
+                    operation: PromptWriteOperation::Write,
+                    source: PromptWriteSource::MemoryFilesystemAdapter,
+                    content,
+                    previous_content_hash: previous_hash.as_deref(),
+                    allowance: context.prompt_write_safety_allowance(),
+                    filesystem_operation: FilesystemOperation::WriteFile,
+                },
+            )
+            .await?;
+            backend_context = memory_context_with_prompt_safety_enforcement(&context, enforcement);
```
serrrfirat, line 264: **Medium Severity**

Adapter-level prompt-safety approval can be rechecked and rejected by the wrapped repository backend.

When the adapter policy/sink is overridden, it enforces prompt safety and forwards only the allowance in `MemoryContext`. `RepositoryMemoryBackend::write_document()` then runs its own `enforce_prompt_write_safety()` with the backend's separate sink/policy. If the host configured the sink on the adapter only, an adapter-approved warn/bypass can be rejected by the backend with `prompt_write_safety_event_unavailable`. Append has the same composition issue.

Please make adapter enforcement authoritative for that call (for example a context marker for already-enforced safety) or centralize/forward the sink so the second check cannot contradict the first.

crates/ironclaw_memory/src/safety.rs
```diff
@@ -0,0 +1,910 @@
...
+            .and_then(|reason| reason.severity)
+            .or_else(|| parts.findings.map(|findings| findings.severity)),
+        finding_count: parts
+            .reason
+            .map(|reason| reason.finding_count)
+            .or_else(|| parts.findings.map(|findings| findings.finding_count))
+            .unwrap_or(0),
+        allowance: parts.allowance.cloned(),
+    };
+    if let Err(error) = event_sink.record_prompt_write_safety_event(event).await {
```
serrrfirat, line 809: **Medium Severity**

A configured prompt-safety event sink failure blocks every outcome, even when `require_sink` is false.

The missing-sink path only errors for warn/bypass, but once a sink is configured, any `record_prompt_write_safety_event()` error is converted into `prompt_write_safety_event_unavailable`. That can brick otherwise-safe `Allow` writes to protected files, and can mask the real reason for `Reject` outcomes behind an audit-sink outage.

Please only make sink failures fatal for outcomes that require a durable audit event before persistence, and preserve the original reject reason when persistence would be blocked anyway.

crates/ironclaw_memory/src/backend.rs
```diff
@@ -0,0 +1,616 @@
...
+            ));
+        }
+        if request.full_text() && !self.capabilities.full_text_search {
+            return Err(memory_backend_unsupported(
+                context.scope(),
+                FilesystemOperation::ReadFile,
+                "memory backend does not support full-text search",
+            ));
+        }
+        if request.vector()
```
serrrfirat: **Medium Severity**

Hybrid/default search can silently degrade to full-text-only even when vector search was requested.

`MemorySearchRequest::new()` defaults to `full_text=true, vector=true`, but this fail-closed check only errors for unsupported vector search when full text is disabled or a query embedding is already present. If `vector_search=false` (or no embedding is attached), the repository search path just returns an empty vector branch and succeeds as FTS-only. That hides missing vector/embedding configuration in retrieval.

Please fail closed whenever `request.vector()` is true but vector retrieval cannot run, or explicitly change the API to represent best-effort fallback and test that behavior.

crates/ironclaw_memory/src/repo/native_libsql.rs
```diff
@@ -0,0 +1,1190 @@
...
+        serde_json::from_str(&metadata).map(Some).map_err(|error| {
+            memory_error(
+                virtual_path,
+                FilesystemOperation::ReadFile,
+                error.to_string(),
+            )
+        })
+    }
+
+    async fn write_document_metadata(
```
serrrfirat, line 491: **Medium Severity**

Metadata-only writes do not reindex or clear affected chunks.

`write_document_metadata()` updates the JSON metadata only. If a document or ancestor `.config` is changed to `skip_indexing=true` after content was already indexed, the existing chunk rows remain searchable until the document body is rewritten/reindexed, even though current metadata has opted the content out of indexing. The native Postgres metadata writer has the same issue.

Please reindex/clear affected documents when metadata changes can affect indexing, or make metadata writes go through a caller that performs that invalidation, and add a regression toggling `skip_indexing` after chunks exist.

crates/ironclaw_memory/tests/reborn_native_repository_backend_contract.rs
```diff
@@ -0,0 +1,1454 @@
...
+
+use async_trait::async_trait;
+use ironclaw_filesystem::FilesystemError;
+use ironclaw_memory::{
+    EmbeddingError, EmbeddingProvider, MemoryBackend, MemoryBackendCapabilities, MemoryContext,
+    MemoryDocumentIndexer, MemoryDocumentPath, MemoryDocumentScope, MemorySearchRequest,
+    RepositoryMemoryBackend,
+};
+
+#[cfg(feature = "libsql")]
```
serrrfirat: **Medium Severity**

This test file does not compile under the supported postgres-only feature set.

The file is enabled for either `libsql` or `postgres`, but `ChunkConfig`, `ChunkingMemoryDocumentIndexer`, and `FusionStrategy` are imported only under `#[cfg(feature = "libsql")]`. Several Postgres tests use those names, so `cargo test -p ironclaw_memory --features postgres --no-default-features --tests --no-run` fails with undeclared type errors.

Please import these shared types for `any(feature = "libsql", feature = "postgres")` (or unconditionally) and add the postgres-only no-default-features test command to coverage/CI if it is meant to be supported.

crates/ironclaw_memory/src/search.rs
```diff
@@ -0,0 +1,454 @@
...
+    }
+
+    let mut results = HashMap::<String, ResultAccumulator>::new();
+    for result in full_text_results {
+        let score = match request.fusion_strategy() {
+            FusionStrategy::Rrf => 1.0 / (request.rrf_k() as f32 + result.rank as f32),
+            FusionStrategy::WeightedScore => request.full_text_weight() / result.rank as f32,
+        };
+        results
+            .entry(result.chunk_key)
```
serrrfirat: **Medium Severity**

Search fusion keys by chunk ID even though the public result identity is the document path.

The native branch queries populate `chunk_key` with each chunk row's ID, so a long document that matches in multiple chunks can return multiple `MemorySearchResult`s for the same `path`. Full-text and vector hits on different chunks of the same file also never combine into one hybrid document score, allowing one note to consume multiple result slots and push other matching documents out of the top-N.

Please fuse at the document level (or otherwise collapse per-path results before truncating) and add a multi-chunk same-document regression covering duplicate paths and hybrid scoring across chunks.

crates/ironclaw_memory/src/filesystem.rs
```diff
@@ -0,0 +1,765 @@
...
+        let prompt_safety_allowance = if is_protected {
+            take_prompt_safety_allowance(
+                &self.one_shot_prompt_safety_allowance,
+                path,
+                FilesystemOperation::AppendFile,
+            )?
+        } else {
+            None
+        };
+        let metadata = resolve_document_metadata(self.repository.as_ref(), &document_path).await?;
```
serrrfirat: **Medium Severity**

`MemoryDocumentFilesystem::append_file()` reuses stale metadata/options after optimistic append conflicts.

Metadata is resolved once before the retry loop and then reused for every attempt. If another writer changes document or ancestor `.config` metadata between a conflict and the winning retry, the append can validate against the old schema or archive a version even after versioning was disabled.

Please resolve metadata/options inside each retry attempt after reading the latest document state, and add a conflict regression where `.config` changes between attempts.

crates/ironclaw_memory/src/indexer.rs
serrrfirat: **Medium Severity**

Embedding-generation failures after a successful overwrite leave the previous chunks searchable.

For non-empty content, `build_chunk_writes(...).await?` returns before `replace_document_chunks_if_current()` if the provider fails. The backend/filesystem callers intentionally swallow indexer errors after persisting the new document, so the durable content changes but old FTS/vector chunks remain associated with the path. Search can then return stale snippets for content that is no longer present.

Please clear old chunks with the hash guard before/when embedding generation fails, or write text-only chunks when embeddings are unavailable, and add a regression where a provider fails after an existing indexed document is overwritten.

crates/ironclaw_memory/src/repo/postgres.rs
```diff
@@ -0,0 +1,840 @@
...
+        let virtual_path = path.virtual_path().unwrap_or_else(|_| valid_memory_path());
+        let client = self
+            .client(virtual_path.clone(), FilesystemOperation::ReadFile)
+            .await?;
+        let owner_key = scoped_memory_owner_key(path.scope());
+        let agent_id = scoped_memory_agent_id(path.scope());
+        let db_path = db_path_for_memory_document(path);
+        let row = client
+            .query_opt(
+                "SELECT content FROM memory_documents WHERE user_id = $1 AND agent_id IS NOT DISTINCT FROM $2 AND path = $3",
```
serrrfirat, line 121: **Medium Severity**

The legacy Postgres repository binds `agent_id` as text, but the existing app migration defines `memory_documents.agent_id` as `UUID`.

`scoped_memory_agent_id()` returns `Option<&str>` and these queries bind it directly, while `migrations/V1__initial.sql` creates `memory_documents.agent_id UUID`. Against an existing migrated Postgres DB, agent-scoped reads/writes can fail parameter type conversion; against a fresh DB through this repository, the local `CREATE TABLE IF NOT EXISTS` schema drifts to `agent_id TEXT`.

Please align the repository schema/bindings with the canonical migration (parse/bind UUID or introduce a migration to text), and add a Postgres contract test for agent-scoped memory against a migrated schema.

crates/ironclaw_memory/src/repo/native_libsql.rs
```diff
@@ -0,0 +1,1271 @@
...
+    path: &MemoryDocumentPath,
+    virtual_path: &VirtualPath,
+) -> Result<(), FilesystemError> {
+    let scope = path.scope();
+    if is_config_path(path.relative_path()) {
+        let parent = path
+            .relative_path()
+            .rsplit_once('/')
+            .map(|(parent, _)| parent);
+        let like_pattern = parent.map_or("%".to_string(), |parent| format!("{parent}/%"));
```
serrrfirat: **Medium Severity**

The metadata repair clear treats memory paths as SQL LIKE patterns.

`MemoryDocumentPath` allows `%` and `_` in relative path segments, but this builds a descendant prefix with `format!("{parent}/%")` and uses it directly in `path LIKE ?5` (the Postgres path has the same shape). A valid folder such as `team_%/.config` would also match unrelated paths like `team-a/note.md`, deleting their chunks and making those documents disappear from search until reindexed.

Please escape `%`, `_`, and the escape character with `LIKE ... ESCAPE`, or use a literal prefix/range predicate. Add native libSQL and Postgres regressions with `team_%/doc.md` and `team-a/doc.md` proving only the literal `team_%/` subtree is cleared.

crates/ironclaw_memory/src/repo/native_libsql.rs
```diff
@@ -0,0 +1,1271 @@
...
+        )
+        .await
+        .map_err(|error| {
+            memory_error(
+                virtual_path.clone(),
+                FilesystemOperation::WriteFile,
+                error.to_string(),
+            )
+        })?;
+        if parsed_metadata.skip_indexing == Some(true) {
```
serrrfirat: **Medium Severity**

A parent `skip_indexing=true` metadata update clears chunks for documents that explicitly opt back into indexing.

This bulk clear runs whenever the written metadata has `skip_indexing: true`; for `.config` paths it deletes all descendant chunks by prefix. That ignores the metadata merge contract where document metadata overlays inherited config. A descendant with its own `{"skip_indexing": false}` can be correctly indexed, then later a parent `.config` write deletes its chunks even though resolved metadata still says it should remain searchable.

Please clear/reindex only descendants whose resolved metadata after the update is actually `skip_indexing == true` (or enqueue override-aware descendant reindexing). Add libSQL and Postgres tests with two descendants, one document-level `skip_indexing:false`, and assert only the inherited document disappears from search.

crates/ironclaw_memory/src/indexer.rs
```diff
@@ -0,0 +1,378 @@
...
+                .repository
+                .replace_document_chunks_if_current(path, &content_hash_at_read, &[])
+                .await;
+        }
+        let chunks =
+            match build_chunk_writes(path, chunk_texts, self.embedding_provider.as_deref()).await {
+                Ok(chunks) => chunks,
+                Err(error) => {
+                    self.repository
+                        .replace_document_chunks_if_current(path, &content_hash_at_read, &[])
```
serrrfirat: **Medium Severity**

Embedding outages now remove full-text searchability after successful writes.

When `embed_batch` fails, the indexer hash-clears the current document chunks and returns an error. Repository-backed writes/appends intentionally ignore indexer errors after persistence, so a transient embedding provider outage can make a successfully updated document disappear from full-text search too, even though chunk rows support `embedding: None`.

Please either keep existing chunks until retry succeeds, or write fresh text-only chunks with `embedding = NULL` so full-text search remains current while vector search degrades. Add a repository-backed regression with a failing provider that verifies a full-text-only search still finds the newly written token.

crates/ironclaw_memory/src/filesystem.rs
```diff
@@ -0,0 +1,774 @@
...
+}
+
+fn memory_context_with_prompt_safety_enforcement(
+    context: &MemoryContext,
+    enforcement: PromptWriteSafetyEnforcement,
+) -> MemoryContext {
+    let mut context = context.clone();
+    if let Some(allowance) = enforcement.allowance {
+        context = context.with_prompt_write_safety_allowance(allowance);
+    }
```
serrrfirat, line 48: **High Severity**

Adapter-level audit sink configuration can supersede a stricter backend prompt-write policy.

`with_prompt_write_safety_event_sink()` marks the adapter prompt-safety config as overridden. For protected paths, the adapter then enforces its own policy, marks the context as already enforced here, and the wrapped `RepositoryMemoryBackend` skips its policy. A host that only wanted to add an adapter audit sink can accidentally bypass a stricter backend policy requiring previous-content hashes, custom detection, or custom rejections.

Please separate policy/registry overrides from sink configuration, or only mark backend context as safety-enforced when the adapter is explicitly replacing backend policy. Add write and append regressions where an adapter has only an event sink and the backend has a stricter custom policy that must still reject.

crates/ironclaw_memory/src/backend.rs
```diff
@@ -0,0 +1,796 @@
...
+
+    pub fn with_prompt_write_safety_allowance(
+        mut self,
+        allowance: PromptSafetyAllowanceId,
+    ) -> Self {
+        self.prompt_write_safety_allowance = Some(allowance);
+        self
+    }
+
+    pub fn with_prompt_write_safety_enforced(mut self) -> Self {
```
serrrfirat: **High Severity**

`MemoryContext::with_prompt_write_safety_enforced()` is a public, forgeable prompt-safety bypass.

`MemoryContext` is publicly re-exported, and this public builder lets any direct backend caller set the enforcement bit. `RepositoryMemoryBackend` then skips protected write/append safety checks whenever that bit is set, so a caller can persist high-risk content to `SOUL.md`/`BOOTSTRAP.md` through the backend seam without policy evaluation.

Please make this marker crate-private or replace it with an unforgeable token/type only the adapter can construct after enforcement. Add a direct-backend regression proving external callers cannot bypass prompt-write safety with a marked context.

crates/ironclaw_memory/src/repo/native_libsql.rs
serrrfirat: **Medium Severity**

Metadata writes can clear chunks even when the metadata row was not updated.

Both native repositories ignore the affected-row count from `UPDATE reborn_memory_documents ... path = ?/$5`, then run the `skip_indexing == true` chunk clear. A call to `write_document_metadata()` for a nonexistent root `.config` persists no metadata but still executes the config-path clear with `LIKE '%'`, removing every chunk in that scope; a nonexistent nested `.config` similarly clears that subtree.

Please only run the chunk clear when the metadata update actually matched a document row (or make metadata writes insert/fail consistently). Add libSQL and Postgres regressions where a missing `.config` metadata write does not delete existing searchable chunks.

crates/ironclaw_memory/src/indexer.rs
```diff
@@ -0,0 +1,378 @@
...
+                        .await?;
+                    return Err(error);
+                }
+            };
+        self.repository
```
serrrfirat, line 171: **Medium Severity**

An in-flight reindex can undo a `skip_indexing=true` metadata clear.

The indexer reads content, computes a content hash, resolves metadata, and later calls `replace_document_chunks_if_current()` guarded only by that content hash. A concurrent metadata write can set `skip_indexing=true` and clear chunks while this reindex is in flight; because the document content hash did not change, the stale reindex can then insert chunks again after the metadata call returns. That leaves a document or config subtree marked skip-indexed but still searchable until another reindex happens.

Please make chunk replacement validate current indexing metadata under the same repository lock/transaction, or version/order metadata changes so stale reindexes cannot reinsert chunks after a skip-indexing clear. Add a concurrency regression where a reindex that read old metadata races with a `skip_indexing=true` metadata update and must not leave chunks searchable.

crates/ironclaw_memory/src/filesystem.rs
```diff
@@ -0,0 +1,774 @@
...
+            &self.prompt_protected_path_registry,
+            &document_path,
+        )
+        .is_some();
+        let backend_capabilities = self.backend.capabilities();
+        let backend_will_enforce_protected_path = backend_capabilities.prompt_write_safety
+            && self
+                .backend
+                .prompt_write_safety_protects_path(&document_path);
+        let adapter_should_enforce_prompt_safety = is_protected
```
serrrfirat, line 215: **High Severity**

The adapter trusts `prompt_write_safety=true` even when the backend does not protect this path.

`backend_will_enforce_protected_path` checks both the capability and `prompt_write_safety_protects_path()`, but `adapter_should_enforce_prompt_safety` ignores that path-level result. A custom/plugin backend can advertise `prompt_write_safety: true` while leaving `prompt_write_safety_protects_path()` false/default; for adapter-classified protected files like `SOUL.md` or `AGENTS.md`, the adapter then skips its own policy, and a backend that simply persists the write bypasses the prompt-write safety boundary through the filesystem surface.

Please make the adapter enforce when `is_protected && (!backend_capabilities.prompt_write_safety || !backend_will_enforce_protected_path || self.prompt_safety_config_overridden)` for both write and append, or fail closed when a backend claims prompt safety but does not report protection for that path. Add a plugin/backend regression that advertises prompt safety without protecting default paths and verify protected writes are still rejected.

PR #3181, opened 2026-05-01
feat(reborn-memory): native schema + empty repo wiring (PR 2 of #3118)

nickpismenkov:
## Summary

PR 2 of the stack for #3118. Stacks on #3180. Add the Reborn-native `reborn_memory_*` substrate that the rest of the stack builds on, without yet implementing read/write/list/search behavior.

The native schema lives inside the crate, consistent with the existing `LibSqlMemoryDocumentRepository::run_migrations` pattern where the crate manages its own DDL in isolation from the main app's migration system. The legacy `memory_documents` table is deliberately untouched — Reborn memory must not depend on or piggy-back on the legacy schema.

## What's added

- `src/repo/native_libsql.rs` — `RebornLibSqlMemoryDocumentRepository`, schema constant `REBORN_LIBSQL_MEMORY_DOCUMENTS_SCHEMA` (4 tables + indexes + triggers).
- `src/repo/native_postgres.rs` — `RebornPostgresMemoryDocumentRepository`, schema constant `REBORN_POSTGRES_MEMORY_DOCUMENTS_SCHEMA` (4 tables + GIN/HNSW indexes).
- `tests/reborn_native_schema_contract.rs` — smoke tests that prove the substrate is created, idempotent, and that read/write/list ops surface a clear "not yet implemented" error rather than silently returning empty data.

## Schema (matches issue #3118 exactly)

```sql
reborn_memory_documents
  tenant_id  TEXT NOT NULL
  user_id    TEXT NOT NULL
  agent_id   TEXT NOT NULL DEFAULT ''   -- empty string is the DB-only
  project_id TEXT NOT NULL DEFAULT ''   -- absent sentinel; _none stays virtual-only
  path       TEXT NOT NULL
  content, content_hash, metadata, created_at, updated_at
  UNIQUE (tenant_id, user_id, agent_id, project_id, path)

reborn_memory_chunks         -- chunks with content_hash + optional embedding
reborn_memory_chunks_fts     -- FTS5 virtual table (libSQL)
                             -- TSVECTOR GIN + HNSW vector (Postgres)
reborn_memory_document_versions  -- monotonic version, cascade delete
```

Indexes cover full-scope lookup, scope+path lookup, list/search by scope, chunk lookup by document_id, FTS lookup, and version lookup by `(document_id, version DESC)`.

## Behavior

`run_migrations()` is real and idempotent. Every other repository operation returns:

```
FilesystemError::Backend { reason: "reborn-native libsql memory repository is not yet implemented", ... }
```

Callers fail closed instead of treating an empty result as authoritative. PR 3 (libSQL) and PR 4 (Postgres) replace these stubs with the real implementations.

## Test plan

- [x] `cargo clippy -p ironclaw_memory --all-targets --all-features -- -D warnings` — clean
- [x] `cargo test -p ironclaw_memory --features libsql` — 43/43 pass
- [x] `cargo test -p ironclaw_memory --all-features` — 44/44 pass
- [x] `cargo test -p ironclaw_architecture reborn_crate_dependency_boundaries_hold` — pass
- [x] `python3 scripts/check_no_panics.py --base origin/reborn-integration --head HEAD` — clean
- [x] `git diff --check origin/reborn-integration` — clean

New native schema smoke tests (`tests/reborn_native_schema_contract.rs`):

- `reborn_libsql_run_migrations_creates_native_substrate_idempotently` — first run materializes substrate, second run is a no-op, all four `reborn_memory_*` objects exist, legacy `memory_documents` is **not** created.
- `reborn_libsql_repository_fails_closed_until_behavior_lands` — read/write/list each surface a "not yet implemented" error.

## Stack

- PR 1: #3180 (lib.rs split — base of this PR)
- **PR 2: this PR (native schema + empty wiring)**
- PR 3: native libSQL repository behavior (next)
- PR 4: native Postgres repository + test harness
- PR 5: ported semantic/search/versioning tests
- PR 6: host/runtime public-seam composition


crates/ironclaw_memory/src/repo/native_libsql.rs
```diff
@@ -0,0 +1,239 @@
...
+fn not_yet_implemented(path: VirtualPath, operation: FilesystemOperation) -> FilesystemError {
+    memory_error(
+        path,
+        operation,
+        "reborn-native libsql memory repository is not yet implemented",
+    )
+}
+
+#[async_trait]
+impl MemoryDocumentRepository for RebornLibSqlMemoryDocumentRepository {
```
serrrfirat, line 79: **Medium Severity** — Native repositories do not fail closed for metadata methods.

The native repositories override read/write/list/search and index operations to return `not yet implemented`, but they do not override `read_document_metadata` or `write_document_metadata`. Those methods inherit the `MemoryDocumentRepository` defaults (`Ok(None)` and `Ok(())`), so callers can believe metadata reads/writes succeeded before native behavior exists. The same issue applies to `RebornPostgresMemoryDocumentRepository`.

Suggested fix: override both metadata methods in both native repos to return the same `not_yet_implemented` error, and extend the fail-closed test to cover metadata, search, and index methods.

crates/ironclaw_memory/tests/reborn_native_schema_contract.rs
```diff
@@ -0,0 +1,112 @@
...
+    let db = std::sync::Arc::new(
+        libsql::Builder::new_local(db_path)
+            .build()
+            .await
+            .expect("libsql build"),
+    );
+    (db, dir)
+}
+
+#[cfg(feature = "libsql")]
```
serrrfirat, line 25: **Medium Severity** — PostgreSQL native schema migration has no behavioral smoke test.

This PR adds real PostgreSQL DDL in `RebornPostgresMemoryDocumentRepository::run_migrations`, but this new schema contract file only defines libSQL tests. With only `--features postgres`, the target compiles and runs 0 tests, so the Postgres migration remains compile-only. That leaves extension creation, trigger SQL, generated TSVECTOR, and HNSW index DDL unexecuted until runtime.

Suggested fix: add a Postgres testcontainer/deadpool smoke test that calls `RebornPostgresMemoryDocumentRepository::run_migrations()` twice and verifies the `reborn_memory_*` substrate exists.

PR #3182, opened 2026-05-01
feat(reborn-memory): native libSQL repository behavior (PR 3 of #3118)

nickpismenkov:
## Summary

PR 3 of the stack for #3118. Stacks on PR 2 (#3181). Replace the "not yet implemented" stubs in `RebornLibSqlMemoryDocumentRepository` with a full read/write/list/search/version/chunk implementation against the `reborn_memory_*` substrate.

Every query carries the full `(tenant_id, user_id, agent_id, project_id)` scope tuple. Absent agent/project IDs use the empty-string DB sentinel; the constructor rejects empty / `_none` user-supplied IDs, so equality (`agent_id = ?`) is unambiguous and the legacy `NULL IS DISTINCT FROM` / `(? IS NULL AND agent_id IS NULL) OR agent_id = ?` gymnastics are gone.

## Behavior implemented

| Operation | What it does on the native schema |
|---|---|
| `read_document` | Filter by full scope tuple + path; return content as bytes. |
| `read_document_metadata` | Filter by full scope tuple + path; parse `metadata` JSON column. |
| `list_documents` | Scoped SELECT of all paths under the scope. |
| `write_document` / `write_document_with_options` | `BEGIN IMMEDIATE` -> `list_paths_for_scope` -> file/directory conflict check -> insert-or-update with `content_hash`; archives prior content to `reborn_memory_document_versions` when content changed and `skip_versioning != Some(true)`. |
| `write_document_metadata` | Scoped UPDATE of the `metadata` column. |
| `search_documents` | FTS branch joins through `reborn_memory_chunks_fts` with full-scope WHERE; vector branch reads embeddings + cosine + deterministic tie-break by path; RRF / weighted-score fusion via shared `fuse_memory_search_results`. |
| `replace_document_chunks_if_current` | Re-reads `content_hash` inside the transaction; no-op if it drifted (prevents corrupting the index after a between-read-and-refresh rewrite). |
| `delete_document_chunks` | Scoped DELETE. |

## Reusable native helpers

Added to `repo/mod.rs` for both native backends:

- `REBORN_SCOPE_NONE_SENTINEL` (`""`)
- `reborn_agent_id_db_value` / `reborn_project_id_db_value`
- `reborn_memory_document_from_row(tenant, user, agent_db, project_db, db_path)` — reconstructs a `MemoryDocumentPath` from native columns, mapping the empty-string sentinel back to `None`.

## Test plan

13 new behavioral tests in `tests/reborn_native_libsql_repository_contract.rs`:

[...]

PR #3183, opened 2026-05-02
feat(reborn-memory): native Postgres repository behavior (PR 4 of #3118)

nickpismenkov:
## Summary

PR 4 of the stack for #3118. Stacks on PR 3 (#3182). Replace the "not yet implemented" stubs in `RebornPostgresMemoryDocumentRepository` with a full read/write/list/search/version/chunk implementation against the `reborn_memory_*` Postgres substrate.

Every query carries the full `(tenant_id, user_id, agent_id, project_id)` scope tuple. Absent agent/project IDs use the empty-string DB sentinel; the constructor rejects empty / `_none` user-supplied IDs, so equality (`agent_id = $3`) is unambiguous and NULL/IS NULL gymnastics are unnecessary.

## Behavior implemented

| Operation | What it does on the native Postgres schema |
|---|---|
| `read_document` | Filter by full scope tuple + path; return `content` column as bytes. |
| `read_document_metadata` | Filter by full scope tuple + path; return `metadata` JSONB column. |
| `list_documents` | Scoped `SELECT path` of all rows under the scope. |
| `write_document` / `write_document_with_options` | `BEGIN` -> `LOCK TABLE … IN SHARE ROW EXCLUSIVE MODE` -> `list_paths_for_scope` -> file/directory conflict check -> upsert with `content_hash`; archives prior content into `reborn_memory_document_versions` when content changed and `skip_versioning != Some(true)`. |
| `write_document_metadata` | Scoped `UPDATE` of the `metadata` JSONB column. |
| `search_documents` | FTS branch joins through `reborn_memory_chunks` with `ts_rank_cd` over the GIN-indexed `content_tsv`; vector branch uses pgvector `<=>` ordering. Both branches are pre-fusion ranked; results fuse through the shared `search::fuse_memory_search_results`, honoring `FusionStrategy`. |
| `replace_document_chunks_if_current` | `SELECT … FOR UPDATE` row lock, hash drift check, replace chunks with optional pgvector embedding. |
| `delete_document_chunks` | Scoped `DELETE` through the `reborn_memory_documents` join. |
| `run_migrations` | Wrapped in a session-level `pg_advisory_lock(REBORN_MIGRATION_LOCK_ID)` so concurrent processes (parallel test runs) serialize on `CREATE EXTENSION pgcrypto/vector`. |

`plainto_tsquery('english', \$5)` already tolerates arbitrary punctuation, so the Postgres path does not need a manual FTS-escape helper — the punctuation contract is preserved at the query layer instead.

## Test plan

13 new behavioral tests in `tests/reborn_native_postgres_repository_contract.rs`, at parity with the libSQL contract from PR 3:

- `round_trips_a_document_within_full_scope`
- `returns_none_when_document_is_missing`
- `upsert_replaces_content_for_same_full_scope_and_path`
- `full_scope_isolates_user_agent_project_independently`
- `top_level_projects_path_is_a_normal_user_path_not_project_scope`
- `rejects_file_directory_prefix_conflicts_within_scope`
- `writes_metadata_and_reads_it_back`
- `write_with_options_creates_version_row_only_when_not_skipped`
- `version_numbers_are_monotonic_and_content_hash_matches_archived_content`
- `replace_chunks_if_current_is_a_noop_when_document_was_rewritten`
- `full_text_search_returns_only_chunks_within_full_scope`
- `fts_query_escapes_punctuation_and_handles_empty_input_gracefully`
- `full_text_search_uses_rrf_when_only_full_text_branch_returns_results`

Tests use the standard `DATABASE_URL=postgres://localhost/ironclaw_test` env-var pattern with `try_connect` skipping cleanly when Postgres is unreachable (matches `tests/workspace_integration.rs`); each test scopes a unique tenant prefix and cleans up via `cleanup_tenant` so parallel runs do not collide.

Verification (all green):

- [x] \`cargo fmt --all -- --check\` — clean
- [x] \`cargo clippy -p ironclaw_memory --all-targets --all-features -- -D warnings\` — clean
- [x] \`cargo clippy --all --tests --examples --all-features -- -D warnings\` — clean
- [x] \`cargo test -p ironclaw_memory --features libsql\` — green
- [x] \`cargo test -p ironclaw_memory --features postgres\` — green (skips when no DB)
- [x] \`cargo test -p ironclaw_memory --all-features\` — green
- [x] \`cargo test -p ironclaw_architecture reborn_crate_dependency_boundaries_hold\` — pass
- [x] \`python3 scripts/check_no_panics.py --base origin/reborn-integration --head HEAD\` — clean
- [x] \`git diff --check origin/reborn-integration\` — clean

## Stack

- PR 1: #3180 (lib.rs split)
- **PR 4: this PR (native Postgres behavior)**
- PR 5: ported semantic/search/versioning tests
- PR 6: host/runtime public-seam composition


crates/ironclaw_memory/tests/reborn_native_postgres_repository_contract.rs
```diff
@@ -0,0 +1,563 @@
...
+    let mgr = deadpool_postgres::Manager::new(config, tokio_postgres::NoTls);
+    deadpool_postgres::Pool::builder(mgr)
+        .max_size(4)
+        .build()
+        .expect("build deadpool")
+}
+
+/// Returns `Some(())` if the pool can hand out a connection, `None` otherwise.
+/// Tests use this to skip when Postgres is unavailable.
+async fn try_connect(pool: &deadpool_postgres::Pool) -> Option<()> {
```
serrrfirat, line 53: `try_connect()` makes the new Postgres contract tests skip by returning `None` when the pool cannot hand out a connection, and every test then early-returns from `fresh_repository()`. That means `cargo test -p ironclaw_memory --features postgres --test reborn_native_postgres_repository_contract` can report all tests as passed without running migrations or exercising read/write/search/chunk/version behavior against Postgres at all—I reproduced this locally with the test binary printing connection-skip messages and still finishing green.

This undercuts the crate guardrail that Postgres coverage must be real behavioral coverage, not compile-only. Can we make this harness provision/require an ephemeral Postgres+pgvector instance via the established testcontainer/deadpool pattern, or fail unless an explicit opt-in skip env var is set?

crates/ironclaw_memory/src/repo/native_postgres.rs
```diff
@@ -123,26 +412,355 @@ impl MemoryDocumentIndexRepository for RebornPostgresMemoryDocumentRepository {
...
+            memory_error(
+                virtual_path,
+                FilesystemOperation::WriteFile,
+                error.to_string(),
+            )
+        })?;
+        Ok(())
     }
 
     async fn delete_document_chunks(
```
serrrfirat: `replace_document_chunks_if_current()` is protected by the expected `content_hash`, but this delete path is not: it deletes chunks solely by scoped path. The shared indexer calls `delete_document_chunks()` after reading the document/metadata when `skip_indexing == Some(true)` or chunking produces no chunks. If that reindex is stale and a newer write has already produced fresh chunks, this unconditional delete can remove the current document's chunks and leave the latest content unsearchable.

Can we carry the expected content hash into the delete path as well (or lock/read the document row and only delete when the current `content_hash` still matches the indexer read)? That would give chunk deletion the same currentness protection as `replace_document_chunks_if_current()`.

PR #3184, opened 2026-05-02
test(reborn-memory): port pure-behavior contract over native repos (PR 5 of #3118)

nickpismenkov:
## Summary

PR 5 of the stack for #3118. Stacks on PR 4 (#3183). Add a behavior contract for `RepositoryMemoryBackend` layered over both `RebornLibSqlMemoryDocumentRepository` and `RebornPostgresMemoryDocumentRepository`. The semantics under test live above the bare repository — they belong to the `RepositoryMemoryBackend` composition (`.config` inheritance, schema validation, indexer best-effort, capability fail-closed, embedding-dimension guard).

These behaviors already exist in production code; **PR 5 is purely test coverage** that ports the matching contract from the legacy `db_memory_repository_contract.rs` over to the native `reborn_memory_*` substrate. No production-code changes.

## Coverage delta vs. PR 4

| Behavior in #3118 phase 6                              | Status |
|--------------------------------------------------------|--------|
| \`.config\` metadata inheritance precedence              | new in PR 5 |
| Closer \`.config\` overrides parent \`.config\`            | new in PR 5 |
| \`skip_indexing\` from \`.config\`                         | new in PR 5 |
| \`skip_versioning\` from \`.config\`                       | new in PR 5 |
| Schema validation before persistence                   | new in PR 5 |
| Indexer failure after persist => write success         | new in PR 5 |
| Capability fail-closed for vector search               | new in PR 5 |
| Query embedding dimension mismatch fails closed        | new in PR 5 |
| RRF fusion (hybrid)                                    | new in PR 5 |
| WeightedScore fusion                                   | new in PR 5 |
| \`limit\` honored                                        | new in PR 5 |
| FTS query escaping                                     | already in PR 3/4 |
| Version hash semantics                                 | already in PR 3/4 |

## Test plan

New file \`tests/reborn_native_repository_backend_contract.rs\`:

**libSQL (10 tests, in-process temp DB):**

- \`libsql_backend_inherits_config_metadata_for_skip_indexing\`
- \`libsql_backend_closer_config_overrides_parent_config\`
- \`libsql_backend_honors_skip_versioning_from_config\`
- \`libsql_backend_validates_schema_from_config_before_write\`
- \`libsql_backend_reports_write_success_when_indexer_fails_after_persist\`
- \`libsql_backend_search_fails_closed_for_unsupported_vector_search\`
- \`libsql_backend_search_fails_closed_on_query_embedding_dimension_mismatch\`
- \`libsql_backend_hybrid_search_fuses_full_text_and_vector_results\`
- \`libsql_backend_weighted_score_fusion_orders_results_by_weights\`
- \`libsql_backend_search_honors_limit\`

**Postgres (4 tests, skip-on-unreachable per \`tests/workspace_integration.rs:26-33\`; each test scopes a unique tenant prefix and cleans up via \`pg_cleanup_tenant\`):**

- \`postgres_backend_validates_schema_from_config_before_write\`
- \`postgres_backend_honors_skip_versioning_from_config\`
- \`postgres_backend_reports_write_success_when_indexer_fails_after_persist\`
- \`postgres_backend_search_fails_closed_on_query_embedding_dimension_mismatch\`

Verification (all green):

- [x] \`cargo fmt --all -- --check\` — clean
- [x] \`cargo clippy -p ironclaw_memory --all-targets --all-features -- -D warnings\` — clean
- [x] \`cargo clippy --all --tests --examples --all-features -- -D warnings\` — clean
- [x] \`cargo test -p ironclaw_memory --features libsql\` — green
- [x] \`cargo test -p ironclaw_memory --all-features\` — green
- [x] \`cargo test -p ironclaw_architecture reborn_crate_dependency_boundaries_hold\` — pass
- [x] \`python3 scripts/check_no_panics.py --base origin/reborn-integration --head HEAD\` — clean
- [x] \`git diff --check origin/reborn-integration\` — clean

## Stack

- PR 1: #3180 (lib.rs split)
- PR 4: #3183 (native Postgres behavior)
- **PR 5: this PR (ported semantic/search/versioning tests)**
- PR 6: host/runtime public-seam composition (next)

### DIFF
--- a/crates/ironclaw_memory/tests/reborn_native_repository_backend_contract.rs
+++ b/crates/ironclaw_memory/tests/reborn_native_repository_backend_contract.rs
@@ -17,9 +17,6 @@
 
 use std::sync::Arc;
 
-#[cfg(feature = "libsql")]
-use std::sync::Mutex;
-
 use async_trait::async_trait;
 use ironclaw_filesystem::FilesystemError;
 use ironclaw_memory::{
@@ -119,7 +116,7 @@ async fn count_rows_libsql(db: &Arc<libsql::Database>, table: &str) -> i64 {
 
 #[cfg(feature = "libsql")]
 #[tokio::test]
-async fn libsql_backend_inherits_config_metadata_for_skip_indexing() {
+async fn libsql_backend_skip_indexing_from_config_writes_zero_chunks_to_db() {
     use ironclaw_memory::MemoryDocumentRepository;
     let fixture = libsql_repo().await;
     let repo = fixture.repo.clone();
@@ -130,8 +127,13 @@ async fn libsql_backend_inherits_config_metadata_for_skip_indexing() {
         .await
         .unwrap();
 
-    let recording = Arc::new(CountingIndexer::default());
-    let backend = RepositoryMemoryBackend::new(repo.clone()).with_indexer(recording.clone());
+    // Use the real ChunkingMemoryDocumentIndexer which reads the resolved
+    // metadata and short-circuits on `skip_indexing`. This proves the
+    // signal actually reaches the chunk-writing layer — the previous
+    // version of this test only counted indexer invocations, which a
+    // skip-aware indexer respects but does not observably zero out.
+    let indexer = Arc::new(ChunkingMemoryDocumentIndexer::new(repo.clone()));
+    let backend = RepositoryMemoryBackend::new(repo.clone()).with_indexer(indexer);
     let context = MemoryContext::new(MemoryDocumentScope::new("tenant-a", "alice", None).unwrap());
     let path = MemoryDocumentPath::new("tenant-a", "alice", None, "folder/note.md").expect("path");
 
@@ -140,13 +142,46 @@ async fn libsql_backend_inherits_config_metadata_for_skip_indexing() {
         .await
         .unwrap();
 
-    // The backend currently always invokes the indexer (skip_indexing is a
-    // signal to the indexer, not the backend), so the call still happens —
-    // but the chunk write must observe the metadata. The behavior we lock
-    // in here is that resolved metadata reaches the indexer scope, which
-    // we exercise by confirming the write succeeded and the indexer ran
-    // exactly once for the new document.
-    assert_eq!(recording.count(), 1);
+    let chunk_count = count_rows_libsql(&fixture.db, "reborn_memory_chunks").await;
+    assert_eq!(
+        chunk_count, 0,
+        "skip_indexing inherited from .config must produce zero chunks"
+    );
+}
+
+#[cfg(feature = "libsql")]
+#[tokio::test]
+async fn libsql_backend_document_metadata_overrides_inherited_config() {
+    use ironclaw_memory::MemoryDocumentRepository;
+    let fixture = libsql_repo().await;
+    let repo = fixture.repo.clone();
+    // Parent .config says skip_versioning=true; document-level metadata
+    // overrides it back to false. The next overwrite must produce a
+    // version row, proving doc-level metadata wins over inherited config.
+    let cfg = MemoryDocumentPath::new("tenant-a", "alice", None, "logs/.config").expect("path");
+    repo.write_document(&cfg, b"").await.unwrap();
+    repo.write_document_metadata(&cfg, &serde_json::json!({"skip_versioning": true}))
+        .await
+        .unwrap();
+
+    let path = MemoryDocumentPath::new("tenant-a", "alice", None, "logs/entry.md").expect("path");
+    repo.write_document(&path, b"v1").await.unwrap();
+    repo.write_document_metadata(&path, &serde_json::json!({"skip_versioning": false}))
+        .await
+        .unwrap();
+
+    let backend = RepositoryMemoryBackend::new(repo.clone());
+    let context = MemoryContext::new(MemoryDocumentScope::new("tenant-a", "alice", None).unwrap());
+    backend
+        .write_document(&context, &path, b"v2")
+        .await
+        .unwrap();
+
+    let count = count_rows_libsql(&fixture.db, "reborn_memory_document_versions").await;
+    assert_eq!(
+        count, 1,
+        "document-level metadata must override inherited .config skip_versioning"
+    );
 }
 
 #[cfg(feature = "libsql")]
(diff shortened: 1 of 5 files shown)


crates/ironclaw_memory/tests/reborn_native_repository_backend_contract.rs
```diff
@@ -0,0 +1,702 @@
...
+        .expect("DATABASE_URL must be a valid Postgres URL");
+    let mgr = deadpool_postgres::Manager::new(config, tokio_postgres::NoTls);
+    deadpool_postgres::Pool::builder(mgr)
+        .max_size(4)
+        .build()
+        .expect("build deadpool")
+}
+
+#[cfg(feature = "postgres")]
+async fn pg_try_connect(pool: &deadpool_postgres::Pool) -> Option<()> {
```
serrrfirat: **Medium Severity** — Postgres backend tests can pass without exercising Postgres.

`pg_try_connect()` returns `None` when the pool cannot hand out a connection, and each Postgres test then early-returns from `pg_repo()`. That means this new backend contract can report all Postgres tests as passed without running migrations or exercising schema validation/versioning/indexer/search behavior against Postgres at all. I reproduced this locally with `--nocapture`: the test binary printed `skipping reborn-postgres backend test: error connecting to server` messages and still finished green.

This repeats the coverage problem from the previous PR and conflicts with the `ironclaw_memory` guardrail that Postgres repository coverage must be real behavioral coverage, not compile/skip coverage. Can we switch this harness to provision/require an ephemeral Postgres+pgvector instance via the established testcontainer/deadpool pattern, or fail unless an explicit opt-in skip env var is set?

crates/ironclaw_memory/tests/reborn_native_repository_backend_contract.rs
```diff
@@ -0,0 +1,702 @@
...
+    let fixture = libsql_repo().await;
+    let repo = fixture.repo.clone();
+    let config_path =
+        MemoryDocumentPath::new("tenant-a", "alice", None, "folder/.config").expect("path");
+    repo.write_document(&config_path, b"").await.unwrap();
+    repo.write_document_metadata(&config_path, &serde_json::json!({"skip_indexing": true}))
+        .await
+        .unwrap();
+
+    let recording = Arc::new(CountingIndexer::default());
```
serrrfirat: **Medium Severity** — This test does not actually verify `skip_indexing` inheritance.

The test installs a `CountingIndexer` that only increments a counter. It never resolves `.config` metadata and never writes/deletes chunk rows, so the assertion that the indexer ran once would still pass if `.config` inheritance or `skip_indexing` handling were completely broken. The legacy contract this ports used a real `ChunkingMemoryDocumentIndexer` and asserted zero chunk rows, which actually proved that the resolved `.config` metadata reached the indexing behavior.

Can we use the real chunking indexer here and assert that no `reborn_memory_chunks` rows are created for `folder/note.md` when `folder/.config` has `{"skip_indexing": true}`? That would make the test cover the behavior named in the test and PR coverage table.

crates/ironclaw_memory/tests/reborn_native_repository_backend_contract.rs
```diff
@@ -0,0 +1,702 @@
...
+        .unwrap();
+    assert!(
+        hybrid.is_hybrid(),
+        "hybrid match must report is_hybrid=true"
+    );
+}
+
+#[cfg(feature = "libsql")]
+#[tokio::test]
+async fn libsql_backend_weighted_score_fusion_orders_results_by_weights() {
```
serrrfirat, line 449: **Medium Severity** — This test does not verify weighted-score ordering or weights.

Despite the test name and coverage table saying weighted-score fusion ordering is covered, the request only switches to `FusionStrategy::WeightedScore` with default weights and then asserts that the result list is non-empty. This would still pass if `WeightedScore` were ignored entirely and the backend always used RRF/default ordering.

Can we make this assert the actual ordered result paths and/or scores under asymmetric weights, e.g. set `with_full_text_weight(0.0)` and `with_vector_weight(1.0)` (or the inverse) and verify the vector-preferred/full-text-preferred document moves accordingly?

crates/ironclaw_memory/tests/reborn_native_repository_backend_contract.rs
```diff
@@ -0,0 +1,702 @@
...
+        .await
+        .unwrap();
+
+    // The backend currently always invokes the indexer (skip_indexing is a
+    // signal to the indexer, not the backend), so the call still happens —
+    // but the chunk write must observe the metadata. The behavior we lock
+    // in here is that resolved metadata reaches the indexer scope, which
+    // we exercise by confirming the write succeeded and the indexer ran
+    // exactly once for the new document.
+    assert_eq!(recording.count(), 1);
```
serrrfirat: **Medium Severity**

The `skip_indexing` contract test is currently vacuous.

This test installs a `CountingIndexer` and only asserts that `reindex_document()` was called once. That does not verify that the `folder/.config` metadata was inherited, nor that `skip_indexing` reached the chunking/indexing path. If `RepositoryMemoryBackend` stopped resolving `.config` metadata entirely, or if `ChunkingMemoryDocumentIndexer` stopped honoring `skip_indexing`, this test would still pass because the dummy indexer never reads metadata or writes/deletes chunks.

Please make this exercise the real behavior, e.g. use `ChunkingMemoryDocumentIndexer` and assert `reborn_memory_chunks` stays empty for `folder/note.md` (matching the legacy contract), or otherwise expose/capture the resolved metadata passed into the indexer.

crates/ironclaw_memory/tests/reborn_native_repository_backend_contract.rs
```diff
@@ -0,0 +1,702 @@
...
+        .search(
+            &context,
+            MemorySearchRequest::new("literal")
+                .unwrap()
+                .with_limit(3)
+                .with_fusion_strategy(FusionStrategy::WeightedScore),
+        )
+        .await
+        .unwrap();
+    assert!(!results.is_empty());
```
serrrfirat: **Medium Severity**

This `WeightedScore` coverage does not verify ordering or weights.

Despite the test name, the only assertion is `!results.is_empty()`. A regression that ignores `FusionStrategy::WeightedScore`, uses RRF unconditionally, swaps the full-text/vector weights, or returns the wrong order would still pass as long as any row is returned.

Please make the fixture distinguish the strategies/weights and assert the expected order. For example, run one search with vector-heavy weights and one with full-text-heavy weights, then assert the top result changes (or assert an exact ordered path list for a setup where WeightedScore differs from RRF).

crates/ironclaw_memory/tests/reborn_native_repository_backend_contract.rs
```diff
@@ -0,0 +1,702 @@
...
+    let results = backend
+        .search(
+            &context,
+            MemorySearchRequest::new("shared-keyword")
+                .unwrap()
+                .with_limit(2),
+        )
+        .await
+        .unwrap();
+    assert!(results.len() <= 2, "limit must cap fused result count");
```
serrrfirat: **Medium Severity**

The limit test can pass even when search/indexing is broken.

`assert!(results.len() <= 2)` is true for zero or one result, so this does not prove that the backend found more than two matches and then truncated them to the requested limit. A regression that returns no FTS results (or indexes only one chunk/document) would still satisfy this assertion.

Please establish the precondition and assert the truncation, e.g. first search with a larger limit and assert there are at least/Exactly five matching documents, then search with `.with_limit(2)` and assert `results.len() == 2`.

crates/ironclaw_memory/tests/reborn_native_repository_backend_contract.rs
```diff
@@ -0,0 +1,702 @@
...
+        .await
+        .unwrap_err();
+    assert!(
+        err.to_string()
+            .contains("query embedding dimension 2 does not match"),
+        "expected dimension-mismatch error, got: {err}"
+    );
+}
+
+#[cfg(feature = "libsql")]
```
serrrfirat, line 389: **Medium Severity**

The new search/fusion/limit contract is only exercised against libSQL.

The PR summary says this ports behavior over the native repositories, and the coverage table lists RRF fusion, WeightedScore fusion, and `limit` as new contract coverage. In this file those tests are all under `#[cfg(feature = "libsql")]`, while `RebornPostgresMemoryDocumentRepository` has separate FTS and pgvector query implementations. A Postgres-only regression in ranking, fusion inputs, vector results, or limit handling would not be caught by this contract suite.

Please add Postgres counterparts for the search/fusion/limit cases (using the existing skip-on-unreachable pattern if needed), or narrow the stated contract so reviewers do not assume these behaviors are covered for both native backends.

PR #3185, opened 2026-05-02
test(reborn-memory): vertical integration through public seams (PR 6 of #3118)

nickpismenkov:
## Summary

PR 6 of the stack for #3118. Stacks on PR 5 (#3184). Add caller-facing vertical-integration coverage that drives the full public-seam stack against the Reborn-native repositories:

```
reborn-native repository
  -> ChunkingMemoryDocumentIndexer
  -> RepositoryMemoryBackend
  -> MemoryBackendFilesystemAdapter
  -> CompositeRootFilesystem mounted at /memory
```

These behaviors already exist in production code; **PR 6 is purely test coverage**. No production-code changes.

This PR also closes thoroughness gaps identified in a Required-Tests audit of #3118 — every gap was a missing assertion on behavior the production code already implements.

### `tests/reborn_native_filesystem_vertical_integration.rs` — new

Public-seam vertical integration. **libSQL (6 tests, in-process temp DB):**
- `libsql_round_trips_authorized_read_write_through_composite_mount`
- `libsql_lists_direct_children_through_composite_mount`
- `libsql_search_through_composite_mount_returns_only_same_scope_results`
- `libsql_capability_denied_search_fails_closed_before_repository_is_called`
- `libsql_composite_mount_rejects_invalid_memory_path_with_clean_error`
- `libsql_composite_mount_rejects_duplicate_root_at_registration`

**Postgres (2 tests, skip-on-unreachable; per-tenant cleanup):**
- `postgres_round_trips_authorized_read_write_through_composite_mount`
- `postgres_search_through_composite_mount_returns_only_same_scope_results`

### `tests/reborn_native_repository_backend_contract.rs` — strengthened

- **Strengthened** skip_indexing test: now asserts `reborn_memory_chunks` row count = 0 in DB (was: indexer-call-count = 1 — passive). Uses real `ChunkingMemoryDocumentIndexer` that respects skip_indexing. Renamed to `libsql_backend_skip_indexing_from_config_writes_zero_chunks_to_db`.
- **New** `libsql_backend_document_metadata_overrides_inherited_config`: parent .config says skip_versioning=true, document-level metadata overrides to false; overwrite must produce a version row.
- **Strengthened** weighted-score test to assert ordering changes when weights flip (FT-heavy vs vector-heavy produces different top results, proving weights actually steer the score).
- **New** `libsql_backend_search_honors_pre_fusion_limit_per_branch` asserting result count is bounded by an effective pre_fusion_limit.
- **New** `pre_fusion_limit_is_clamped_up_to_limit` (unit test locking the API clamp invariant).

### `tests/reborn_native_libsql_repository_contract.rs` — new tests

- **New** `concurrent_writes_under_same_scope_and_path_produce_exactly_one_row`: two `tokio::join!`-launched writes serialize via `BEGIN IMMEDIATE`; assert exactly one row.
- **New** `fts_query_with_only_stopwords_does_not_error` covering the issue's "empty/stopword-ish" FTS case.

### `tests/reborn_native_postgres_repository_contract.rs` — new tests

- **New** `same_path_in_different_tenants_stores_separate_rows` (Postgres parity for the tenant axis the libSQL contract already covered).
- **New** Postgres mirrors of the concurrent-writes and stopword-FTS tests above.

### `src/search.rs` — unit tests

- **New** unit tests for `fuse_memory_search_results` deterministic tiebreak: tied scores must sort by relative path ascending, independent of insertion order.

## Coverage delta for issue #3118 "Required tests"

| Required test (issue)                                        | Status |
|--------------------------------------------------------------|--------|
| Same path different tenants / users / agents / projects      | ✓ |
| Search / list scope isolation                                | ✓ |
| `projects/` prefix as user path                              | ✓ |
| Same scope+path upserts                                      | ✓ |
| Different scope + same path → different docs                 | ✓ |
| File / directory prefix conflict                             | ✓ |
| **Concurrent writes → exactly one row**                      | **✓ new** |
| Indexer failure after persist → write succeeds               | ✓ |
| `.config` applies to children                                | ✓ |
| Closer `.config` overrides parent                            | ✓ |
| **Document metadata overrides inherited `.config`**          | **✓ new** |
| **`skip_indexing` → chunk count = 0**                        | **✓ strengthened** |
| `skip_versioning` → no version row                           | ✓ |
| Schema validation rejects before persistence                 | ✓ |
| Write → chunk → FTS end-to-end                               | ✓ |
| **FTS escaping (punctuation + stopwords)**                   | **✓ stopwords added** |
| Vector dim mismatch                                          | ✓ |
| **Hybrid ranking deterministic (path tiebreak)**             | **✓ new** |
| **RRF + WeightedScore fusion (weights observably steer order)**| **✓ strengthened** |
| **`limit` + `pre_fusion_limit`**                              | **✓ pre_fusion_limit added** |
| Version monotonic + content hash                             | ✓ |
| Adapter / RootFilesystem vertical integration                | ✓ |
| Unsupported capability fails closed before side effects      | ✓ |

## Acceptance criteria from #3118 closed by this PR

- [x] Public Reborn seams (`MemoryBackendFilesystemAdapter` / `RootFilesystem`) have vertical integration coverage against native repositories.

Verification (all green):

- [x] `cargo clippy -p ironclaw_memory --all-targets --all-features -- -D warnings`
- [x] `cargo clippy --all --tests --examples --all-features -- -D warnings`
- [x] `cargo test -p ironclaw_memory --features libsql` — 101 tests across 11 targets
- [x] `cargo test -p ironclaw_memory --all-features`
- [x] `cargo test -p ironclaw_architecture reborn_crate_dependency_boundaries_hold`
- [x] `python3 scripts/check_no_panics.py --base origin/reborn-integration --head HEAD`
- [x] `git diff --check origin/reborn-integration`

## Stack (this PR completes the issue's non-deferred work)

- PR 1: #3180 (lib.rs split)
- PR 4: #3183 (native Postgres behavior)
- PR 5: #3184 (ported semantic/search/versioning tests)
- **PR 6: this PR (host/runtime public-seam composition + audit gap fills)**
- PR 7: deferred per the issue (legacy migration / coexistence)

### DIFF
--- a/crates/ironclaw_memory/tests/reborn_native_filesystem_vertical_integration.rs
+++ b/crates/ironclaw_memory/tests/reborn_native_filesystem_vertical_integration.rs
@@ -32,7 +32,7 @@ use ironclaw_memory::{
     MemoryDocumentScope, MemorySearchRequest, RepositoryMemoryBackend,
 };
 
-#[cfg(feature = "libsql")]
+#[cfg(any(feature = "libsql", feature = "postgres"))]
 use ironclaw_memory::{ChunkConfig, ChunkingMemoryDocumentIndexer};
 
 #[cfg(feature = "libsql")]
@@ -332,13 +332,35 @@ fn pg_pool() -> deadpool_postgres::Pool {
         .expect("build deadpool")
 }
 
+/// Explicit opt-in to skip the Postgres vertical tests. Without this set,
+/// a connection failure must fail loud — the previous "silent skip + green
+/// pass" pattern violated the `ironclaw_memory` guardrail that Postgres
+/// behavioral coverage must be real.
 #[cfg(feature = "postgres")]
-async fn pg_try_connect(pool: &deadpool_postgres::Pool) -> Option<()> {
+const POSTGRES_SKIP_ENV: &str = "IRONCLAW_SKIP_POSTGRES_TESTS";
+
+#[cfg(feature = "postgres")]
+fn pg_skip_requested() -> bool {
+    std::env::var(POSTGRES_SKIP_ENV).is_ok_and(|value| value == "1" || value == "true")
+}
+
+#[cfg(feature = "postgres")]
+async fn pg_require_connection(pool: &deadpool_postgres::Pool) -> Option<()> {
     match pool.get().await {
         Ok(_) => Some(()),
         Err(error) => {
-            eprintln!("skipping reborn-postgres vertical test: {error}");
-            None
+            if pg_skip_requested() {
+                eprintln!(
+                    "skipping reborn-postgres vertical test ({POSTGRES_SKIP_ENV}=1): {error}"
+                );
+                None
+            } else {
+                panic!(
+                    "reborn-postgres vertical test could not reach Postgres ({error}); \
+                     set DATABASE_URL to a reachable Postgres+pgvector instance, or set \
+                     {POSTGRES_SKIP_ENV}=1 to explicitly skip."
+                );
+            }
         }
     }
 }
@@ -363,13 +385,25 @@ async fn pg_compose(
     Arc<RepositoryMemoryBackend<RebornPostgresMemoryDocumentRepository>>,
 )> {
     let pool = pg_pool();
-    pg_try_connect(&pool).await?;
+    pg_require_connection(&pool).await?;
     let repository = Arc::new(RebornPostgresMemoryDocumentRepository::new(pool.clone()));
     repository.run_migrations().await.expect("run_migrations");
     pg_cleanup_tenant(&pool, tenant_id).await;
+    // Wire the same chunking indexer the libSQL vertical stack uses so writes
+    // through `CompositeRootFilesystem` populate `reborn_memory_chunks`.
+    // Without it, FTS search has nothing to match and the search assertion
+    // would be vacuously true on an empty result set.
+    let indexer = Arc::new(
+        ChunkingMemoryDocumentIndexer::new(repository.clone()).with_chunk_config(ChunkConfig {
+            chunk_size: 4,
+            overlap_percent: 0.0,
+            min_chunk_size: 1,
+        }),
+    );
     let backend = Arc::new(
-        RepositoryMemoryBackend::new(repository.clone()).with_capabilities(
-            MemoryBackendCapabilities {
+        RepositoryMemoryBackend::new(repository.clone())
+            .with_indexer(indexer)
+            .with_capabilities(MemoryBackendCapabilities {
                 file_documents: true,
                 metadata: true,
                 versioning: true,
@@ -377,8 +411,7 @@ async fn pg_compose(
                 vector_search: false,
                 embeddings: false,
                 ..MemoryBackendCapabilities::default()
-            },
-        ),
+            }),
     );
     let adapter = Arc::new(MemoryBackendFilesystemAdapter::new(backend.clone()));
     let mut composite = CompositeRootFilesystem::new();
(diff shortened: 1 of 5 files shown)


crates/ironclaw_memory/tests/reborn_native_filesystem_vertical_integration.rs
```diff
@@ -0,0 +1,502 @@
...
+        .expect("DATABASE_URL must be a valid Postgres URL");
+    let mgr = deadpool_postgres::Manager::new(config, tokio_postgres::NoTls);
+    deadpool_postgres::Pool::builder(mgr)
+        .max_size(4)
+        .build()
+        .expect("build deadpool")
+}
+
+#[cfg(feature = "postgres")]
+async fn pg_try_connect(pool: &deadpool_postgres::Pool) -> Option<()> {
```
serrrfirat: **Medium Severity** — Postgres vertical tests can pass without exercising Postgres.

`pg_try_connect()` returns `None` when the pool cannot hand out a connection, `pg_compose()` propagates that `None`, and the Postgres vertical tests then early-return. The test binary still reports those tests as passed. I reproduced this locally with `--nocapture`: it printed `skipping reborn-postgres vertical test: error connecting to server` and still finished green.

This continues the same coverage problem from the previous stack PRs and conflicts with the `ironclaw_memory` guardrail that Postgres behavioral coverage must be real, not compile/skip coverage. Can we make this harness provision/require an ephemeral Postgres+pgvector instance via the established testcontainer/deadpool pattern, or fail unless an explicit opt-in skip env var is set?

crates/ironclaw_memory/tests/reborn_native_filesystem_vertical_integration.rs
```diff
@@ -0,0 +1,502 @@
...
+    CompositeRootFilesystem,
+    Arc<RebornPostgresMemoryDocumentRepository>,
+    Arc<RepositoryMemoryBackend<RebornPostgresMemoryDocumentRepository>>,
+)> {
+    let pool = pg_pool();
+    pg_try_connect(&pool).await?;
+    let repository = Arc::new(RebornPostgresMemoryDocumentRepository::new(pool.clone()));
+    repository.run_migrations().await.expect("run_migrations");
+    pg_cleanup_tenant(&pool, tenant_id).await;
+    let backend = Arc::new(
```
serrrfirat, line 403: **Medium Severity** — The Postgres vertical search test can pass without proving search works.

`pg_compose()` builds the Postgres `RepositoryMemoryBackend` without a `ChunkingMemoryDocumentIndexer`, so writes through `CompositeRootFilesystem` do not populate `reborn_memory_chunks`. `RepositoryMemoryBackend::search()` therefore has no chunks to search, and this test's final assertion only checks `paths.iter().all(|p| p == "visible.md")`, which is true for an empty result set. Even with a real Postgres DB, this does not prove write → index → search through the public seams.

Can we wire the same real `ChunkingMemoryDocumentIndexer` here that the libSQL vertical stack uses (FTS-only is enough), then assert the results are non-empty and exactly `visible.md`?

crates/ironclaw_memory/tests/reborn_native_repository_backend_contract.rs
```diff
@@ -428,20 +463,129 @@ async fn libsql_backend_weighted_score_fusion_orders_results_by_weights() {
...
+    assert!(
+        results.len() <= 2,
+        "pre_fusion_limit=2 must cap fused result count, got {}",
+        results.len()
+    );
+}
+
+#[cfg(feature = "libsql")]
+#[test]
+fn pre_fusion_limit_is_clamped_up_to_limit() {
```
serrrfirat, line 565: **Medium Severity** — The builder-order invariant claimed by this test is not actually covered, and the reverse order currently breaks it.

The comment says `with_pre_fusion_limit(N)` is at least `limit` regardless of caller order, but the test only exercises `with_limit(5).with_pre_fusion_limit(2)`. If a caller uses the reverse order, `MemorySearchRequest::new("x").with_pre_fusion_limit(2).with_limit(5)`, `with_limit()` does not re-clamp `pre_fusion_limit`, so the request can keep `pre_fusion_limit == 2` while `limit == 5`. That lets the per-branch SQL candidate limit be smaller than the final requested result limit.

Can we either make `with_limit()` also clamp `pre_fusion_limit = pre_fusion_limit.max(limit)`, or narrow the documented invariant and add a test for the reverse order if the current behavior is intentional?

Issue #1339, opened 2026-03-18
"no supported isa found for arch `armv7`?


slient2010:
Hello,

I’m trying to compile this project on my raspberry pi, but unlucky, got the error below. Can we support or any plan to get this run on raspberry?

OS information 

```text
pi@raspberrypi:~/ironclaw $ cat /etc/os-release 
PRETTY_NAME="Raspbian GNU/Linux 12 (bookworm)"
NAME="Raspbian GNU/Linux"
VERSION_ID="12"
VERSION="12 (bookworm)"
VERSION_CODENAME=bookworm
ID=raspbian
ID_LIKE=debian
HOME_URL="http://www.raspbian.org/"
SUPPORT_URL="http://www.raspbian.org/RaspbianForums"
BUG_REPORT_URL="http://www.raspbian.org/RaspbianBugs"
```

Rust version

``` text
pi@raspberrypi:~/ironclaw $ rustc --version
rustc 1.94.0 (4a4ef493e 2026-03-02)
```

Compile error information 

``` text
error: failed to run custom build command for `cranelift-codegen v0.115.1`

Caused by:
  process didn't exit successfully: `/home/pi/ironclaw/target/debug/build/cranelift-codegen-38470dd2d57660b9/build-script-build` (exit status: 101)  --- stderr

  thread 'main' (1761) panicked at /home/pi/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/cranelift-codegen-0.115.1/build.rs:56:53:  error when identifying target: "no supported isa found for arch `armv7`"
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtracewarning: build failed, waiting for other jobs to finish...
```

ilblackdragon:
Do you have 32 bit OS running? 

Cranelift only supports 64 bit 

We can disable wasm modules and let you use http only but which with v2 engine is way more useful

ilblackdragon:
Won't fix.

IronClaw currently deeply depends on WASM.
It will take both heavy refactoring and a maintenance burden to remove that.
Going forward it might be a considered option (and there was #2318 proposed solution but it's not a relevant refactoring right now.

PR #3208, opened 2026-05-02
ci(docker): add linux/arm64 to image builds (#3168)

### LINKED ISSUE #3168: Add arm64/aarch64 platform support for Docker builds
Ironclaw currently only publishes Docker images for linux/amd64. This prevents running the official image on ARM64 machines.
- Cranelift (the WASM JIT compiler used by Ironclaw) already supports aarch64. The only confirmed ARM limitation is armv7 (32-bit), tracked in #1339.
- The Rust toolchain and all dependencies should compile natively on aarch64.

I create a fork with a minor changes for make a PoC https://github.com/gcaguilar/ironclaw/actions/runs/25252983548

abbyshekit:
## Summary

Closes #3168.

IronClaw ships only \`linux/amd64\` Docker images today, blocking native runs on Apple Silicon, AWS Graviton, Raspberry Pi 4/5, and the broader arm64 fleet. Cranelift (the WASM JIT) already supports \`aarch64\`, so the gap is purely in the publish matrix.

## What this changes

- **`.github/workflows/docker.yml`** — adds \`docker/setup-qemu-action\` (pinned by SHA) and switches both image builds (\`ironclaw\` + \`ironclaw-worker\`) to \`platforms: linux/amd64,linux/arm64\`.
- **`.github/workflows/rebuild-release-image.yml`** — same treatment for both build paths (\`runtime\` target + final-stage fallback).
- **`Dockerfile`** header — removed the \`--platform linux/amd64\` example (host-arch is the right default for a developer once the official images go multi-arch) and added a Buildx multi-arch example.

## Cost / risk

QEMU emulation on the amd64 runner adds roughly 2-3x to the arm64 build time vs amd64. If that bites CI minutes, the right next step is provisioning a native arm64 runner (the inline comment in both workflows points at this). Until then, emulation is the correct tradeoff for parity.

## Confirmation

@gcaguilar's PoC fork run already showed this works end-to-end: https://github.com/gcaguilar/ironclaw/actions/runs/25252983548. Only armv7 (32-bit) is blocked, which is tracked separately in #1339.

## Test plan

- [ ] Trigger \`docker.yml\` on this branch via \`workflow_dispatch\` (a Docker Hub login secret is required so I can't trigger it from a fork PR — needs a maintainer with publish rights to kick it off)
- [ ] Verify the resulting manifest list has both architectures: \`docker manifest inspect nearaidev/ironclaw:sha-<7chars>\`
- [ ] Smoke-test the arm64 image on an Apple Silicon machine: \`docker run --platform linux/arm64 --rm nearaidev/ironclaw:sha-<7chars> ironclaw --version\`

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -118,6 +118,16 @@ jobs:
           username: ${{ vars.DOCKER_REGISTRY_USER }}
           password: ${{ secrets.DOCKER_REGISTRY_TOKEN }}
 
+      # QEMU enables cross-arch emulation on the amd64 runner so the
+      # arm64 image can be built without provisioning native arm64
+      # hardware. The arm64 build is roughly 2-3x slower than amd64
+      # under emulation; if CI minutes become a concern, swap this for
+      # a native arm64 runner once one is available.
+      - name: Set up QEMU (multi-arch emulation)
+        uses: docker/setup-qemu-action@49b3bc8e6bdd4a60e6116a5414239cba5943d3cf # v3
+        with:
+          platforms: arm64
+
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@8d2750c68a42422c14e847fe6c8ac0403b4cbd6f # v3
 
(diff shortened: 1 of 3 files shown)


gcaguilar:
Maybe it would be better to do a native build instead of using QEMU, 
Here you can check https://github.com/gcaguilar/ironclaw/actions/runs/25252983548

PR #3212, opened 2026-05-02
Add Reborn event projection service

### LINKED ISSUE #3093: [Reborn] Add EventProjectionService
## Parent

#2987

## What to build

Add a Reborn `EventProjectionService` that turns durable redacted event logs into product-facing projections such as thread timeline, run status, approval gate, auth gate, tool call status, process status, cost/resource usage, and memory activity.

This is the projection layer product adapters should read instead of relying on process-local stream state or raw service internals.

## Acceptance criteria

- [ ] Projection service reads from durable Reborn event logs using replay cursors and produces scoped projection snapshots.
- [ ] Projections cover at least thread timeline, run status, approval/auth gates, tool/capability call status, process status, resource/cost status, and memory activity placeholders.
- [ ] Replay is deterministic and supports reconnect/gap handling without leaking raw inputs, host paths, secrets, runtime output, approval reasons, or backend error details.
- [ ] SSE/WebSocket/API-oriented projection shapes are documented for product adapter consumption.
- [ ] Tests cover normal replay, cursor resume, gap/snapshot metadata, cross-scope denial, missing event log, malformed event, and no-exposure sentinel strings.
- [ ] Projection failures are observable but do not mutate source-of-truth control-plane state.

## Blocked by

- Coordinate with #3022 for event substrate integration-test requirements.
- Coordinate with #3067 for vertical-slice integration coverage.

zmanian:
## Summary
- add `ironclaw_event_projections` with `EventProjectionService`, replay-derived `ReplayEventProjectionService`, scoped `ProjectionScope`, cursor/rebase metadata, `ThreadTimeline`, and `RunStatusProjection` DTOs
- project runtime events from `DurableEventLog` without depending on concrete JSONL/PostgreSQL/libSQL backends or raw storage rows
- document the on-demand replay choice, deferred materialized store/approval-resource-memory-mission projections, and no-exposure rules
- add Reborn dependency-boundary coverage so projections stay backend-independent and below product adapters

## Design decisions
- Projection mode: on-demand replay first; no materialized `ProjectionStore` or schema in this PR
- API shape: `snapshot(ProjectionRequest)` and `updates(ProjectionRequest)` over explicit `ProjectionScope`
- Gap behavior: durable replay gaps map to `ProjectionError::RebaseRequired` so consumers rebase instead of silently skipping facts
- Initial slice: runtime-event-backed thread timeline and run status; approval/auth/resource/memory/mission projections are deferred until source facts are stable

## Verification
- `cargo test -p ironclaw_event_projections`
- `cargo clippy -p ironclaw_event_projections --all-targets -- -D warnings`
- `git diff --check`

Closes #3093.

## Issue linkage audit

Closes #3093.


crates/ironclaw_event_projections/src/lib.rs
```diff
@@ -0,0 +1,448 @@
...
+    current_status: Option<RunProjectionStatus>,
+    has_active_process: bool,
+) -> RunProjectionStatus {
+    match kind {
+        RuntimeEventKind::DispatchRequested
+        | RuntimeEventKind::RuntimeSelected
+        | RuntimeEventKind::ProcessStarted => RunProjectionStatus::Running,
+        RuntimeEventKind::DispatchSucceeded
+            if has_active_process && current_status == Some(RunProjectionStatus::Running) =>
+        {
```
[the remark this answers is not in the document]
zmanian: Fixed in a986b09a. Incremental `updates` now rebuilds the current status for runs touched by the returned page from the scoped log prefix through `next_cursor`, so a page containing only `DispatchSucceeded` preserves the prior active process state instead of marking the run completed.

crates/ironclaw_event_projections/src/lib.rs
```diff
@@ -0,0 +1,448 @@
...
+    }
+}
+
+fn project_runs(entries: &[EventLogEntry<RuntimeEvent>]) -> Vec<RunStatusProjection> {
+    let mut runs = HashMap::<InvocationId, RunStatusProjection>::new();
+    for entry in entries {
+        apply_run_event(&mut runs, entry);
+    }
+    let mut runs = runs.into_values().collect::<Vec<_>>();
+    runs.sort_by_key(|run| run.invocation_id.as_uuid());
```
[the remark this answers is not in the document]
zmanian: Fixed in a986b09a. Run projections now sort by most recent `updated_at` first, with `last_cursor` as the deterministic tie-breaker, and the contract test covers the product-facing order.

crates/ironclaw_event_projections/src/lib.rs
```diff
@@ -0,0 +1,524 @@
...
+        timestamp: event.timestamp,
+        kind: event.kind.into(),
+        invocation_id: event.scope.invocation_id,
+        thread_id: event.scope.thread_id.clone(),
+        capability_id: event.capability_id.clone(),
+        provider: event.provider.clone(),
+        runtime: event.runtime,
+        process_id: event.process_id,
+        output_bytes: event.output_bytes,
+        error_kind: event.error_kind.clone(),
```
serrrfirat: **High Severity**: Projection DTOs can leak unsanitized runtime error details.

`project_timeline_entry` copies `event.error_kind` directly into `TimelineEntry`, and `apply_run_event` does the same for `RunStatusProjection`. The crate's no-exposure contract relies on projection output never exposing raw error strings, paths, or token-shaped data, but `DurableEventLog` is just a trait and `RuntimeEvent` fields are public. A custom/future backend can return an in-memory `RuntimeEvent` whose `error_kind` was not produced via the sanitized serde path, and the projection will serialize it unchanged.

Suggestion: sanitize again at the projection boundary, e.g. use `ironclaw_events::sanitize_error_kind` when filling both DTO fields (and when updating `run.error_kind`), and add a regression test with a fake log or direct `RuntimeEvent` construction whose `error_kind` contains raw text/path/token-shaped content.

crates/ironclaw_event_projections/src/lib.rs
```diff
@@ -0,0 +1,524 @@
...
+fn run_status_for_event(
+    kind: RuntimeEventKind,
+    current_status: Option<RunProjectionStatus>,
+    has_active_process: bool,
+) -> RunProjectionStatus {
+    match kind {
+        RuntimeEventKind::DispatchRequested
+        | RuntimeEventKind::RuntimeSelected
+        | RuntimeEventKind::ProcessStarted => RunProjectionStatus::Running,
+        RuntimeEventKind::DispatchSucceeded
```
serrrfirat, line 633: **High Severity**: `dispatch_succeeded` can overwrite a terminal process failure.

For process-backed runs, a dispatch success event can be just the acknowledgement that a background process was spawned. The current guard only preserves `Running` when `has_active_process && current_status == Running`; if events arrive as `process_started -> process_failed -> dispatch_succeeded` (plausible under async/process observation races), this falls through and marks the run `Completed`, hiding the terminal failure.

Suggestion: once a process-backed run reaches `Failed` or `Killed`, do not let a later `DispatchSucceeded` change the status. More generally, dispatch acknowledgement should not override terminal process events.

crates/ironclaw_event_projections/src/lib.rs
```diff
@@ -0,0 +1,524 @@
...
+        let page = self.read_runtime(request).await?;
+        let touched_runs = page
+            .entries
+            .iter()
+            .map(|entry| entry.record.scope.invocation_id)
+            .collect::<HashSet<_>>();
+        let mut runs = if touched_runs.is_empty() {
+            Vec::new()
+        } else {
+            let prefix = self
```
serrrfirat: **Medium Severity**: `updates(limit=...)` can replay and allocate the whole scoped history.

`updates()` first reads a bounded page, but if any run is touched it calls `read_runtime_prefix` from origin to `page.next_cursor`. That helper accumulates every matching prefix entry into a `Vec` before projecting runs, so a normal polling call like `updates(limit=1)` on a long-lived thread can reread and allocate the entire history. This makes the product-facing API vulnerable to latency/OOM as event streams grow.

Suggestion: fold run state incrementally while paging instead of collecting the prefix, add a hard replay cap/rebase signal, or require/materialize checkpoints before adapters use this on unbounded streams. Add a regression with thousands of prefix events and `updates(limit=1)` asserting bounded memory/page behavior.

crates/ironclaw_event_projections/src/lib.rs
zmanian: Fixed in 7a851e10. Both `project_timeline_entry` and `apply_run_event` now run `error_kind` through `ironclaw_events::sanitize_error_kind` before populating the projection DTOs, so a custom `DurableEventLog` backend that returns an unsanitized in-memory `RuntimeEvent` cannot leak raw error strings/paths/secrets through projections. Regression test `replay_projection_re_sanitizes_unsanitized_runtime_events_from_custom_backend` constructs a hand-built `RuntimeEvent` with raw `/tmp/...` and `sk_live` content via a custom backend and asserts the projection emits `Unclassified` and the serialized snapshot contains none of the forbidden tokens.

crates/ironclaw_event_projections/src/lib.rs
zmanian: Fixed in 7a851e10. `run_status_for_event` now has an explicit guard: if `has_active_process` and the current status is `Failed` or `Killed`, a subsequent `DispatchSucceeded` preserves the terminal status instead of marking the run `Completed`. Two regression tests cover this: `replay_projection_dispatch_succeeded_does_not_clobber_terminal_process_failure` (process_started -> process_failed -> dispatch_succeeded -> still Failed, error_kind preserved) and `replay_projection_dispatch_succeeded_does_not_clobber_terminal_process_killed` (same shape with process_killed -> still Killed).

crates/ironclaw_event_projections/src/lib.rs
zmanian: Fixed in 7a851e10. The old `read_runtime_prefix` (which collected the full scoped prefix into a `Vec` then projected) is gone. Replaced by `fold_runtime_prefix`, which folds each page incrementally and only retains state for invocations the caller touched on this page, so memory is `O(touched_runs)` regardless of stream length. A hard `STATE_REPLAY_MAX_EVENTS = 100_000` cap surfaces `ProjectionError::RebaseRequired` instead of silently scanning unbounded history. The bounded-memory contract is documented as a doc comment on the function and on the constant. Regression test `replay_projection_updates_with_small_limit_handles_long_prefix` seeds 600 prefix events (across multiple internal pages) and asserts `updates(limit=1)` still returns exactly one update, reconstructs the touched run from the folded prefix, and lands on the correct next cursor.

crates/ironclaw_event_projections/src/lib.rs
```diff
@@ -0,0 +1,570 @@
...
+    async fn read_runtime(
+        &self,
+        request: ProjectionRequest,
+    ) -> Result<ProjectedRuntimePage, ProjectionError> {
+        if request.limit == 0 {
+            return Err(ProjectionError::InvalidRequest {
+                reason: "limit must be greater than zero",
+            });
+        }
+        let fetch_limit = request
```
serrrfirat, line 300: **Medium Severity** — Bound projection page size at the service layer

`ProjectionRequest.limit` is only rejected when it is `0` or overflows `limit + 1`, then the resulting value is passed directly to `read_after_cursor`. If a product adapter forwards a user-controlled or accidentally huge limit, a caller can force the projection service/backend to scan and return an arbitrarily large page. That contradicts the contract language that the limit is bounded and creates a plausible DoS/latency risk for long-lived event streams.

Please add a service-owned maximum page size (for example `MAX_PROJECTION_PAGE_LIMIT`) and reject requests above it before calling the durable log. It would also be worth adding tests for `limit = 0`, `limit = MAX`, and `limit = MAX + 1` so future adapters cannot accidentally rely on unbounded caller-provided values.

serrrfirat, line 300: **Medium Severity** — Bound projection page size at the service layer

`ProjectionRequest.limit` is only rejected when it is `0` or overflows `limit + 1`, then the resulting value is passed directly to `read_after_cursor`. If a product adapter forwards a user-controlled or accidentally huge limit, a caller can force the projection service/backend to scan and return an arbitrarily large page. That contradicts the contract language that the limit is bounded and creates a plausible DoS/latency risk for long-lived event streams.

Please add a service-owned maximum page size (for example `MAX_PROJECTION_PAGE_LIMIT`) and reject requests above it before calling the durable log. It would also be worth adding tests for `limit = 0`, `limit = MAX`, and `limit = MAX + 1` so future adapters cannot accidentally rely on unbounded caller-provided values.

zmanian: Fixed in 2645a35d. Added `MAX_PROJECTION_PAGE_LIMIT = 1_000` and reject limits above it in `read_runtime` before any read (covers both `snapshot()` and `updates()`). Tests cover `limit=0`, `limit=MAX`, `limit=MAX+1`, `limit=usize::MAX`, plus an `updates()` variant.

zmanian: Fixed in 2645a35d. Added `MAX_PROJECTION_PAGE_LIMIT = 1_000` and reject limits above it in `read_runtime` before any read (covers both `snapshot()` and `updates()`). Tests cover `limit=0`, `limit=MAX`, `limit=MAX+1`, `limit=usize::MAX`, plus an `updates()` variant.

crates/ironclaw_event_projections/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
```
serrrfirat, line 1: **High Severity**

This new crate is not registered as a root workspace member, so normal workspace builds/tests do not compile it.

Concrete scenario: at this head, `cargo metadata --format-version=1 --no-deps --manifest-path Cargo.toml` does not include `ironclaw_event_projections`, `cargo test -p ironclaw_event_projections` fails with `package ID specification ... did not match any packages`, and `cargo test --manifest-path crates/ironclaw_event_projections/Cargo.toml` fails because Cargo detects the package under the parent workspace but not in `workspace.members`. That means the new projection service and its contract tests are not exercised by normal CI/workspace commands, and downstream workspace crates cannot depend on it as intended.

Please add `crates/ironclaw_event_projections` to the root `Cargo.toml` workspace members, regenerate/verify `Cargo.lock`, and include evidence from `cargo test -p ironclaw_event_projections` plus the Reborn dependency-boundary test.

crates/ironclaw_architecture/tests/reborn_dependency_boundaries.rs
```diff
@@ -288,6 +288,27 @@ fn boundary_rules() -> Vec<BoundaryRule> {
                 "ironclaw_wasm",
             ],
         },
+        BoundaryRule {
+            crate_name: "ironclaw_event_projections",
```
serrrfirat, line 371: **Medium Severity**

This new boundary rule currently fails open when the crate directory exists but the crate is missing from Cargo metadata.

Concrete scenario: this PR adds the `ironclaw_event_projections` rule here, but `assert_no_normal_workspace_deps` returns early when `cargo metadata` has no package entry. Because the new crate is currently omitted from `workspace.members`, `cargo test -p ironclaw_architecture reborn_crate_dependency_boundaries_hold --test reborn_dependency_boundaries` passes while checking no dependencies for `ironclaw_event_projections` at all. A future forbidden dependency in this new crate would remain invisible until someone separately notices the workspace-membership issue.

After adding the crate to workspace members, please make the guardrail fail closed for crate directories that exist, e.g. assert that `crates/<name>/Cargo.toml` is present in metadata for any active boundary rule, or add an explicit workspace-membership assertion/regression for `ironclaw_event_projections`.

crates/ironclaw_event_projections/src/lib.rs
```diff
@@ -0,0 +1,583 @@
...
+    }
+}
+
+/// Cursor envelope for projection consumers.
+///
+/// This first slice is runtime-event backed. The wrapper keeps callers from
+/// treating raw durable cursors as a stable product API and leaves room for
+/// audit/materialized checkpoints later.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+#[non_exhaustive]
```
serrrfirat, line 99: **Medium Severity**

Projection cursors are not bound to the projection scope that minted them, so a same-stream cursor from a sibling thread/process can silently skip events.

Concrete scenario: `ProjectionCursor` serializes only the raw runtime cursor, and `read_runtime` replays that cursor under whichever `ProjectionScope` the caller supplies. The durable stream is partitioned by tenant/user/agent, while thread/process filtering happens inside the read filter. If thread A has an event at cursor 1 and thread B has an event at cursor 2, a cursor returned for thread B can be passed to `updates()` for thread A. Because cursor 2 is in range for the shared stream, the durable log accepts it and returns an empty thread-A replay, silently skipping thread A's cursor-1 event instead of forcing a snapshot/rebase. I verified this with a temporary regression test in the review worktree.

Please make projection cursors scope-bound/opaque (for example, include and verify a stream/read-scope digest or service-minted token tied to `ProjectionScope`) and reject mismatched-scope cursors with a rebase/snapshot path. Add a regression where a cursor from thread B is used for thread A and assert it cannot silently skip thread A's earlier event.

serrrfirat, line 99: **Medium Severity**

Projection cursors are not bound to the projection scope that minted them, so a same-stream cursor from a sibling thread/process can silently skip events.

Concrete scenario: `ProjectionCursor` serializes only the raw runtime cursor, and `read_runtime` replays that cursor under whichever `ProjectionScope` the caller supplies. The durable stream is partitioned by tenant/user/agent, while thread/process filtering happens inside the read filter. If thread A has an event at cursor 1 and thread B has an event at cursor 2, a cursor returned for thread B can be passed to `updates()` for thread A. Because cursor 2 is in range for the shared stream, the durable log accepts it and returns an empty thread-A replay, silently skipping thread A's cursor-1 event instead of forcing a snapshot/rebase. I verified this on the current head with a temporary regression test.

Please make projection cursors scope-bound/opaque (for example, include and verify a stream/read-scope digest or service-minted token tied to `ProjectionScope`) and reject mismatched-scope cursors with a rebase/snapshot path. Add a regression where a cursor from thread B is used for thread A and assert it cannot silently skip thread A's earlier event.

crates/ironclaw_event_projections/src/lib.rs
```diff
@@ -0,0 +1,583 @@
...
+    }
+}
+
+#[async_trait]
+impl EventProjectionService for ReplayEventProjectionService {
+    async fn snapshot(
+        &self,
+        request: ProjectionRequest,
+    ) -> Result<ProjectionSnapshot, ProjectionError> {
+        let page = self.read_runtime(request).await?;
```
serrrfirat, line 487: **Medium Severity**

`snapshot()` can return a stale run status when the timeline page is truncated, because it derives `runs` only from the single page returned by `read_runtime`.

Concrete scenario: a run emits `DispatchRequested` at cursor 1 and `DispatchSucceeded` at cursor 2. Calling `snapshot(after=None, limit=1)` returns `truncated=true`, but `runs[0].status` is `Running` because `project_runs(&page.entries)` only sees cursor 1. Consumers using a snapshot to rebase after a replay gap can therefore receive a current-looking `RunStatusProjection` that is actually missing later terminal events. I verified this on the current head with a temporary regression test.

Please separate timeline pagination from run-status materialization: either fold the scoped stream through the current head for `runs` (bounded/rebase/materialized checkpoint as needed), or narrow/rename the API so callers cannot treat truncated snapshots as current run-state snapshots. Add a regression where a terminal event lands after the first page and `snapshot(limit=1)` cannot surface stale run status as the current projection.

crates/ironclaw_event_projections/src/lib.rs
```diff
@@ -0,0 +1,624 @@
...
```
serrrfirat, line 487: **Medium Severity**

`snapshot()` can return a stale run status when the timeline page is truncated, because it derives `runs` only from the single page returned by `read_runtime`.

Concrete scenario: a run emits `DispatchRequested` at cursor 1 and then a terminal event such as `DispatchSucceeded`/`ProcessFailed` at cursor 2. Calling `snapshot(after=None, limit=1)` returns `truncated=true`, but `runs[0].status` is still `Running` because `project_runs(&page.entries)` only sees cursor 1. Consumers using a snapshot for initial load or replay-gap rebase can therefore receive a current-looking `RunStatusProjection` that is missing later terminal events. I verified this on the current head with a temporary regression test, and an independent review pass reproduced the same issue.

Please decouple timeline pagination from run-status materialization: either fold/read the scoped stream through the current high-water mark for `runs` (bounded/rebase/materialized checkpoint as needed), or omit/mark run projections as incomplete when a snapshot is truncated. Add a regression where a terminal event lands after the first page and `snapshot(limit=1)` cannot surface stale run status as the current projection.

PR #3215, opened 2026-05-02
fix(gemini-oauth): preserve thoughtSignature in Cloud Code SSE handler (#3214)

### LINKED ISSUE #3214: bug: thoughtSignature dropped in Cloud Code SSE handler — prior fixes (#1565, #1752) incomplete
## Summary

HTTP 400 `INVALID_ARGUMENT` on tool calls with `gemini-3-flash-preview` (and other Gemini 3.x models) persists despite being closed as fixed by #1565 and #1752. Both prior fixes are present in HEAD but the upstream cause was never addressed.

**Version:** `ironclaw-v0.27.0-8-g749fe9da` (released as 0.27.0)
**Environment:** LXC container, fresh install via standard installer
**Provider:** `gemini_oauth` / Cloud Code API
**Model:** `gemini-3-flash-preview`

## Error

```
Error: LLM error: Provider models/gemini-3-flash-preview request failed: HttpError:
Invalid status code 400 Bad Request with message: [{
  "error": {
    "code": 400,
    "message": "Function call is missing a thought_signature in functionCall parts.
                This is required for tools to work correctly, and missing
                thought_signature may lead to degraded model performance.
                Additional data, function call default_api:list_dir , position 4.
                Please refer to https://ai.google.dev/gemini-api/docs/thought-signatures
                for more details.",
    "status": "INVALID_ARGUMENT"
  }
}]
```

## Why the prior fixes did not resolve this

**PR #1565** (`86389dab`) removed a `break` in `ensure_thought_signatures()` so the synthetic sentinel is injected on all `functionCall` parts per turn, not just the first.

**PR #1752** (`684a9d30`) introduced a full roundtrip system: `from_gemini_response()` captures real `thoughtSignature` values from responses; `to_gemini_request()` echoes them back; `ensure_thought_signatures()` fills remaining gaps with the synthetic value `"skip_thought_signature_validator"`.

Both commits are in the current HEAD. The error recurs because **neither PR fixed the upstream source**: the Cloud Code API SSE handler strips `thoughtSignature` before any capture or fallback logic can operate.

## Root cause

**`src/llm/gemini_oauth.rs`, lines 1441–1444:**

```rust
if let Some(fc) = part.get("functionCall") {
    tool_calls_parts.push(serde_json::json!({
        "functionCall": fc      // <-- thoughtSignature silently discarded
    }));
}
```

The Cloud Code API SSE stream delivers `thoughtSignature` as a sibling key in the same part object as `functionCall`:

```json
{
  "functionCall": { "name": "list_dir", "args": {} },
  "thoughtSignature": "<opaque-value>"
}
```

The handler reconstructs the part as `{"functionCall": fc}` — only one field — and discards `thoughtSignature`. This assembled list feeds into `final_response` (line 1529), which `send_request` returns. `from_gemini_response()` then receives a response whose parts contain no `thoughtSignature` and captures nothing. The `thought_signatures` HashMap on the provider stays **empty for the entire session** — not just across DB round-trips as the PR #1752 commit message describes.

For comparison, the official Gemini CLI spreads the full part object to preserve all fields during streaming (`google-gemini/gemini-cli`, `packages/core/src/core/geminiChat.ts`, lines 968–982):

```ts
modelResponseParts.push(
  ...content.parts
    .filter((part) => !part.thought)
    .map((part) => ({
      ...part,          // all fields preserved, including thoughtSignature
      callIndex: ...
    })),
);
```

## Why the synthetic fallback is insufficient

With no real signatures ever captured, every request falls back entirely to `ensure_thought_signatures()` injecting `"skip_thought_signature_validator"`. The Gemini API accepts this sentinel in early turns but intermittently rejects it at later positions in longer conversations. This is consistent with the positions reported across all three incidents: 4, 16, 19, and 21 — all within the `contents` array, varying by conversation length.

## Affected call chain

1. `send_request()` → Cloud Code SSE handler → `tool_calls_parts` assembled without `thoughtSignature` (lines 1441–1444)
2. `from_gemini_response()` → captures nothing from assembled response (lines 1932–1935)
3. `complete_with_tools()` → `thought_signatures` map stays empty (lines 2136–2150)
4. Next call: `to_gemini_request()` → no real signatures to echo (line 1657)
5. `ensure_thought_signatures()` → synthetic sentinel is only fallback (line 1881)
6. API rejects the synthetic value at certain conversation depths → 400

## Secondary issue

The TUI reports `"Copy Failed"` in the status bar when attempting to copy error text. This prevented capturing the raw error from the terminal directly. Worth noting as a separate issue since it makes error reporting harder for users hitting this bug.

abbyshekit:
## Summary

Closes #3214. Re-fixes the Gemini 3.x \`INVALID_ARGUMENT — Function call is missing a thought_signature in functionCall parts\` error that the prior fixes #1565 and #1752 were intended to address but did not, because both fixes worked on the request-builder layer while the upstream SSE parser was still dropping the real signature on the floor.

## Root cause (verbatim from the reporter)

> @thomasmaerz traced it to \`src/llm/gemini_oauth.rs\` lines 1441-1444, where the Cloud Code SSE part-extractor copied only \`functionCall\` and silently discarded \`thoughtSignature\` (which Gemini delivers as a sibling key). The synthetic sentinel \`\"skip_thought_signature_validator\"\` injected by \`ensure_thought_signatures()\` then replaced every real signature, and Gemini's INVALID_ARGUMENT validator rejected the request.

The SSE shape Cloud Code delivers:

\`\`\`json
{
  \"functionCall\": {\"name\": \"list_dir\", \"args\": {}},
  \"thoughtSignature\": \"<base64 token from gemini>\"
}
\`\`\`

## Fix

Extract \`thoughtSignature\` as a sibling of \`functionCall\` in the SSE part-handler. Refactored the inline extraction into a small helper \`extract_function_call_part\` so the contract is unit-testable without standing up a mock SSE stream.

The roundtrip + synthetic-fallback layers added by #1565 / #1752 are unchanged — they now have the real signature to work with for thinking-mode parts, and the synthetic fallback still applies when no signature was delivered.

## Tests

Three regression tests, all passing:

- \`test_extract_function_call_part_preserves_thought_signature\` — the #3214 happy path
- \`test_extract_function_call_part_omits_signature_when_absent\` — the no-signature case still works (request-side fallback handles it)
- \`test_extract_function_call_part_returns_none_for_text_only_part\` — defensive: text-only parts must not be misclassified as functionCall parts

## Test plan

- [x] \`cargo test test_extract_function_call_part\` — 3/3 pass
- [ ] Manual repro: install \`ironclaw-v0.27.0-8-g749fe9da\` (or HEAD with this PR), configure \`gemini_oauth\` provider with \`gemini-3-flash-preview\`, ask a question requiring a tool call (e.g. \`ls .\`). Pre-PR: HTTP 400 INVALID_ARGUMENT after the first tool call. Post-PR: tool calls succeed.

## Issue linkage audit

Closes #3214.

Note: #3214 is already closed as of this audit; keeping the closing reference here preserves traceability for the Cloud Code SSE thoughtSignature fix. Prior related fixes: #1565 and #1752.

### DIFF
--- a/src/llm/gemini_oauth.rs
+++ b/src/llm/gemini_oauth.rs
@@ -162,6 +162,43 @@ fn supports_modern_features(model: &str) -> bool {
     model.contains("gemini-3")
 }
 
+/// Extract a `functionCall` part from a Cloud Code SSE candidate part,
+/// preserving the sibling `thoughtSignature` when present.
+///
+/// Returns `None` when the part does not contain a `functionCall` key.
+///
+/// **Why this exists:** Cloud Code SSE delivers `thoughtSignature` as a
+/// sibling key in the same part object as `functionCall`:
+///
+/// ```json
+/// {
+///   "functionCall": {"name": "list_dir", "args": {}},
+///   "thoughtSignature": "<base64 token from gemini>"
+/// }
+/// ```
+///
+/// The original SSE handler at the call site copied only `functionCall`,
+/// silently discarding `thoughtSignature`. The prior fixes #1565 + #1752
+/// added a roundtrip plus synthetic fallback at the request-builder layer,
+/// but the upstream SSE parser still dropped the real signature here, so
+/// the synthetic sentinel `"skip_thought_signature_validator"` got
+/// installed on every functionCall part. Gemini 3.x's INVALID_ARGUMENT
+/// validator then rejected the request with
+/// `Function call is missing a thought_signature in functionCall parts.`
+/// (issue #3214).
+///
+/// This helper preserves the real signature when present; the synthetic-
+/// fallback path in `ensure_thought_signatures()` remains for the
+/// no-signature case.
+fn extract_function_call_part(part: &serde_json::Value) -> Option<serde_json::Value> {
+    let fc = part.get("functionCall")?;
+    let mut part_obj = serde_json::json!({ "functionCall": fc });
+    if let Some(sig) = part.get("thoughtSignature") {
+        part_obj["thoughtSignature"] = sig.clone();
+    }
+    Some(part_obj)
+}
+
 /// Invalid stream error types mirroring the Gemini CLI.
 #[derive(Debug)]
 #[allow(dead_code)]
@@ -1438,10 +1475,8 @@ impl GeminiOauthProvider {
                                         combined_text.push_str(text);
                                     }
                                 }
-                                if let Some(fc) = part.get("functionCall") {
-                                    tool_calls_parts.push(serde_json::json!({
-                                        "functionCall": fc
-                                    }));
+                                if let Some(part_obj) = extract_function_call_part(part) {
+                                    tool_calls_parts.push(part_obj);
                                 }
                             }
                         }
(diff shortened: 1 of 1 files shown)


thomasmaerz:
Heads up — the underlying issue report was incorrect

I filed issue #3214 which this PR addresses, but I've since confirmed the issue was authored with the wrong provider and auth path. My actual setup uses llm_backend=gemini (OpenAI-compatible endpoint with API key), not gemini_oauth / Cloud Code SSE.

I'd suggest not merging this until the actual reproduction case is established against gemini_oauth. I'm sorry for the noise from the incorrect issue report.

PR #3226, opened 2026-05-03
fix(llm): preserve Gemini thought_signature in OpenAI-compatible tool loops

thomasmaerz:
## Summary
- add optional thought_signature to shared ToolCall and thread it through LLM call sites
- add Gemini-gated rig adapter context to capture functionCall.thoughtSignature from raw responses and replay it on outbound assistant tool calls
- add integration regression coverage for second-turn replay through create_llm_provider and complete_with_tools, plus docs/changelog updates

## Notes
- local full compile/test was intentionally skipped; validation should run in GitHub runners as requested


thomasmaerz:
Closing temporarily to avoid upstream CI noise while validating on fork runners. Will reopen once fork CI is green.

PR #3233, opened 2026-05-03
fix(channels): activate WASM channels on headless servers

ilblackdragon:
## Summary

- On headless `ironclaw run` (no settings DB, no web UI, no persisted activation state), channels listed in the setup wizard's `channels.wasm_channels` config were never auto-activated — Telegram polling and other WASM channels silently never started.
- Boot path now uses `load_startup_active_channels`, which treats persisted `activated_channels` as authoritative once present (including an explicit empty list) and falls back to the configured wizard list only on first run.
- Fail-loud on settings-store errors so a transient DB outage can't silently re-activate channels the user had deactivated.
- Drops the now-dead `Option<&HashSet>` parameter on `setup_wasm_channels` and the matching conditional log message; renames `persisted_*` helpers to `startup_*` to reflect the dual source.

Picks up the spirit of #3105 (closed unmerged because it was opened from `staging` → `main`); rebased onto current `main` and expanded with the fail-loud behavior, helper rename, and a new test.

Co-authored with the original author, @ek775.

## Change Type

- [x] Bug fix

## Linked Issue

Refs #3105

## Validation

- [x] `cargo build`
- [x] Relevant tests pass: new `load_startup_active_channels_falls_back_to_configured_when_no_settings_store` covers the headless path; full `cargo test` suite (5602 tests) passes via pre-push hook
- [x] Manual testing: verified Telegram polling now starts on a headless Ubuntu server with no settings DB

## Security Impact

None — boot-time channel activation only.

## Database Impact

None.

## Blast Radius

`src/main.rs`, `src/channels/wasm/setup.rs`, `src/extensions/manager.rs`. Affects channel activation at startup; persisted activation state remains authoritative once set, so behavior for existing installs is unchanged.

## Rollback Plan

Revert the two commits on this branch.

---

**Review track**: A


src/channels/wasm/setup.rs
```diff
@@ -108,9 +108,11 @@ pub async fn setup_wasm_channels(
         discovered_channels
             .into_iter()
             .filter_map(|(name, discovered)| {
-                startup_active_channel_names
-                    .is_none_or(|active_names| active_names.contains(&name))
-                    .then_some((name, discovered.wasm_path, discovered.capabilities_path))
+                startup_active_channel_names.contains(&name).then_some((
```
serrrfirat, line 113: **Medium Severity** — regression coverage stops at the helper, not the startup side effect

The new test covers `ExtensionManager::load_startup_active_channels`, but not the caller path that converts that result into `startup_active_wasm_channels` and passes it into this filter. A future regression that again passes an empty set here would still pass the helper test while breaking the reported headless boot scenario.

Please add caller-level coverage for the actual activation path: either a focused test around `setup_wasm_channels` showing configured startup channels are loaded/registered when no settings store exists, or a higher-level startup test that exercises the `main.rs` wiring through to channel registration.

src/main.rs
```diff
@@ -484,19 +484,29 @@ async fn async_main() -> anyhow::Result<()> {
...
+    // listed in the config. Settings-store errors propagate — masking them
+    // would silently re-activate channels the user had deactivated.
+    let startup_active_channels: Vec<String> =
+        if let Some(ref ext_mgr) = components.extension_manager {
+            ext_mgr
+                .load_startup_active_channels(
+                    &ext_user_id,
+                    config.channels.configured_wasm_channels.clone(),
+                )
+                .await?
```
serrrfirat: **Medium Severity** — startup fail-loud runs even when channel startup is disabled

`load_startup_active_channels(...)` now fails loud on malformed/persistent settings errors, which is right when we are about to restore WASM channels. But this resolution happens before checking whether non-CLI/WASM channels will actually run. That means a bad `activated_channels` row or transient settings-store error can abort `ironclaw run --cli-only` or `WASM_CHANNELS_ENABLED=false`, where channel restoration is intentionally disabled and there is no risk of reactivating anything.

Please defer this lookup until the `enable_non_cli && config.channels.wasm_channels_enabled` path that actually calls `setup_wasm_channels`, or otherwise make the fail-loud behavior conditional on channel activation being enabled.

src/main.rs
```diff
@@ -484,19 +484,29 @@ async fn async_main() -> anyhow::Result<()> {
...
         } else {
-            Vec::new()
+            normalize_startup_wasm_channel_names(&config.channels.configured_wasm_channels)
+                .into_iter()
+                .collect()
```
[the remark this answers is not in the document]
ilblackdragon: Done in 2ff26aa5. The no-extension-manager branch now calls `ironclaw::extensions::naming::normalize_extension_names(...)` directly, preserving order and matching the validation `load_startup_active_channels` already uses. As a side effect of the deferral fix in #3178547728, this branch now lives inside the `enable_non_cli && wasm_channels_enabled` gate.

src/main.rs
```diff
@@ -484,19 +484,29 @@ async fn async_main() -> anyhow::Result<()> {
...
-    let persisted_active_wasm_channels: std::collections::HashSet<String> =
-        if settings_persistence_available && let Some(ref ext_mgr) = components.extension_manager {
-            persisted_active_wasm_channel_names(ext_mgr, &ext_user_id, &persisted_active_channels)
-                .await
+    let startup_active_wasm_channels: std::collections::HashSet<String> =
+        if let Some(ref ext_mgr) = components.extension_manager {
+            startup_active_wasm_channel_names(ext_mgr, &ext_user_id, &startup_active_channels).await
         } else {
-            std::collections::HashSet::new()
+            startup_active_channels.iter().cloned().collect()
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in 2ff26aa5. The HashSet round-trip is gone — combined with the deferral fix for the fail-loud concern (#3178547728), the resolution now lives inside the `enable_non_cli && wasm_channels_enabled` gate, and the no-ext-mgr branch builds the HashSet once via `startup_active_channels.iter().cloned().collect()` after `normalize_extension_names`.

src/channels/wasm/setup.rs
ilblackdragon: Added in 2ff26aa5. Two new `#[tokio::test]`s in `src/channels/wasm/setup.rs` drive `setup_wasm_channels` end-to-end through the headless config-fallback path:

- `setup_wasm_channels_registers_configured_channel_in_headless_mode` stages the real `channels-src/telegram/telegram.wasm` + `.capabilities.json` into a tempdir, calls `setup_wasm_channels` with `database = None`, `extension_manager = None`, `secrets_store = None`, and `startup_active_channel_names = {"telegram"}`, then asserts `setup.channel_names` contains `"telegram"`. This pins the linkage from the `main.rs` config-fallback resolution to the actual filter inside `setup_wasm_channels` — passing an empty set now demonstrably fails the assertion.
- `setup_wasm_channels_with_empty_active_set_registers_nothing` is the paired regression test for the "empty set lets everything through" bug class: same staged tempdir, `startup_active_channel_names = {}`, asserts zero channels register.

Both tests gracefully skip if `channels-src/telegram` artifacts aren't present (release-tarball builds), so they don't break those flows.

src/main.rs
ilblackdragon: Fixed in 2ff26aa5. The `load_startup_active_channels` resolution moved inside the `enable_non_cli && config.channels.wasm_channels_enabled` block, alongside the existing `setup_wasm_channels` call that consumes the result. `startup_active_wasm_channels` defaults to an empty `HashSet` at outer scope so the later auto-activation block (gated on `wasm_channel_runtime_state`, which is only populated inside the same gate) compiles and naturally does nothing under `--cli-only` / `WASM_CHANNELS_ENABLED=false`. A corrupt `activated_channels` row now only fails startup when channels are actually about to be restored.

The existing `all_non_cli_ingress_paths_are_guarded` test in `main.rs` continues to pass, confirming the new placement still lives behind the `enable_non_cli` brace block.

PR #3235, opened 2026-05-03
test(e2e): unblock Live Canary auth lanes after engine-v2 contract change

ilblackdragon:
## Summary

The Live Canary scheduled job (`Auth Smoke`, `Auth Full`, `Auth Live Seeded`) has failed every run for 3+ days, since the canary cut over to main on 2026-05-01. Three tests in `test_v2_auth_oauth_matrix.py` drive the failures, all rooted in the engine-v2 callable-only contract change from #2868 (merged 2026-04-25) that the tests pre-date.

## What was broken

- **`test_mcp_same_server_multi_user_via_browser`** — engine v2 opens an `approval` pending_gate on the first MCP tool call. The browser fixture has no auto-approve UI, so the chat sat in `pending_gate` for the full 5-minute Playwright timeout. The `expected_text_contains=\"Mock MCP search result\"` predicate could never flip because the assistant bubble never received any content. (Failed in **Auth Smoke** + **Auth Full**.)
- **`test_wasm_tool_oauth_refresh_on_demand`** — same shape: the gmail call gates on `approval` before reaching the http credential-injection layer that triggers the OAuth refresh. Without approving, `refresh_count` never went above 0. (Failed in **Auth Full**.)
- **`test_wasm_tool_first_chat_auth_attempt_emits_auth_url`** — tests OLD engine v2 behavior (auto-emit `gate_required` Authentication on first call to an unauthed extension). After #2868, the engine returns `\"action 'gmail' is not callable in this execution context\"` instead, and `tool_activate(name=...)` is the new model-facing enablement path. The canned mock LLM emits tool calls directly, so this scenario isn't reproducible from a scripted LLM. (Failed in **Auth Full**.)

## Fix

- **`_wait_for_tool_call`** accepts a `token=` kwarg so multi-user tests can poll/approve as a per-user identity (backwards-compatible default).
- **`test_mcp_same_server_multi_user_via_browser`** — drive approval through the per-user API while polling for the tool call. Drop the broken predicate and tied text assertions; the bearer-token isolation assertion (what the test actually exists to verify) is retained.
- **`test_wasm_tool_oauth_refresh_on_demand`** — insert a `_wait_for_tool_call` approval step between `_send_chat` and `_wait_for_refresh_request` so the http credential layer actually runs.
- **`test_wasm_tool_first_chat_auth_attempt_emits_auth_url`** — marked `xfail` with an inline reason pointing at #2868 and the replacement coverage (`test_v2_tool_activate_surface.py`, `test_settings_first_gmail_auth_then_chat_runs`).
- **`tests/e2e/conftest.py`** — set `SECRETS_MASTER_KEY` in `ironclaw_server`'s spawned env. Without it, macOS local dev hangs on a Keychain authorization prompt during `auto_generate_and_persist`, so the fixture kills the process with SIGKILL after 60s and no browser-dependent test can run locally. CI Linux was unaffected (the keychain backend errors fast and the fallback writes to `.env`). Matches the pattern already used in `auth_matrix_server`, `test_v2_engine_auth_cancel`, `test_v2_tool_activate_surface`, etc.

The actual `Auth Live Seeded` failure (`Timed out waiting for extension state: github`) is a separate issue (real OAuth tokens / seeded credentials) and is not addressed here.

## Verification

Reproduced each failure locally (HTTP-only repros for the non-browser tests, server-side log inspection of the multi-user test). Confirmed `pending_gate=approval` was the load-bearing symptom in both auto-callable scenarios.

```
$ pytest tests/e2e/scenarios/test_v2_auth_oauth_matrix.py::test_wasm_tool_oauth_refresh_on_demand
PASSED in 6.74s

$ pytest tests/e2e/scenarios/test_v2_auth_oauth_matrix.py::test_wasm_tool_first_chat_auth_attempt_emits_auth_url
XFAIL in 93s

$ pytest tests/e2e/scenarios/test_v2_auth_oauth_matrix.py -v --timeout=120
12 passed, 2 skipped, 3 xfailed (3 browser tests errored locally;
they'll run cleanly in CI)
```

## Test plan

- [ ] Live Canary `Auth Smoke` and `Auth Full` jobs go green (next scheduled run after merge)
- [ ] Existing passing tests in `test_v2_auth_oauth_matrix.py` stay green


serrrfirat (dismissed):

serrrfirat (dismissed):

serrrfirat (dismissed):

serrrfirat (dismissed):
Approved to unblock merge queue after validating deterministic v2-engine E2E fixes. Full manual E2E workflow passed: https://github.com/nearai/ironclaw/actions/runs/25376351067

serrrfirat (approved):
Re-approving after E2E helper-only follow-up for pending-message race; local web-regressions slice passed.

PR #3240, opened 2026-05-04
feat(wasm): per-request credential signing for HMAC, EIP-712, NEP-413

neo-sky:
## Summary

Adds three new `CredentialLocation` variants (`HmacSignedHeader`, `Eip712SignedHeader`, `Nep413SignedHeader`) alongside the existing five static primitives. Tools declare venue-specific signing schemas in `capabilities.json` via a closed source vocabulary; the runtime stays generic. Unblocks Polymarket L1+L2, Hyperliquid, Trezu, NEAR Intents without any per-venue runtime code.

## Motivation

The five existing `CredentialLocation` variants only support static placement of secrets into headers, query params, or URL placeholders. Real signing recipes need request data (method, path, body, timestamp) that is only known at request time, plus crypto the WASM sandbox cannot safely run. This blocks every venue using HMAC, EIP-712 typed data, or NEP-413: Polymarket L1 (ClobAuth) + L2 (HMAC), Hyperliquid (EIP-712 with body-output signature), Trezu (NEP-413), NEAR Intents (NEP-413).

## Architecture

Three new variants on the existing `CredentialLocation` enum, same dispatcher loop (`inject_host_credentials`), same resolver (`resolve_host_credentials`), same schema layer. No architectural change.

Closed source vocabulary (`FieldSource`, `OutputSource`, `BodyValue`, `BytesPart`) bounds what tools can declare to be signed: every value is either user-reviewed at install time (manifest literal) or derived from the request the tool was already authorized to send (timestamp, nonce, body field, msgpack of body field). No "sign these arbitrary bytes" escape hatch exists.

Body mutation runs after the WASM tool hands over the request body but before reqwest dispatch, needed for Hyperliquid which puts `signature.{r,s,v}` in the body rather than in headers. The new `output_body_fields: Vec<BodyJsonOutput>` field on `Eip712SignedHeader` declares JSON paths and value sources; the runtime parses the body once, walks the paths, sets the values, re-serializes.

EIP-712 schemas in this PR are constrained to a single struct: only the primary type, no nested type dependencies. All four target venues (Polymarket L1 ClobAuth, Hyperliquid Agent, plus the NEP-413 use cases that don't go through EIP-712) use single-struct schemas, so this is sufficient for the unblocking work. Multi-struct support per the full EIP-712 spec is a follow-up; the manifest format already accepts `Vec<Eip712StructDef>`, so adding nested-struct dependency resolution later is additive (no breaking change to the manifest format or the runtime API).

## Security

- All signing happens host-side. WASM tools never see HMAC keys, secp256k1 private keys, or ed25519 seeds.
- WIT contract unchanged: only `secret_exists(name)` is exposed to WASM, never `secret_get`.
- `LeakDetector` still scans WASM-provided URL/headers/body before any host injection, so host-injected secrets cannot self-trigger leak detection.
- Sandbox proxy skip-list extended for the new variants alongside the existing `AuthorizationBasic`/`UrlPath` complex variants. Container proxy delegates to the orchestrator credentials endpoint as before.
- `ResolvedHostCredential::Debug` redacts `secret_value` and shows only header NAMES (no values) for the signing field.
- Signing context (timestamp, random nonce) is generated once per request and shared across multiple matching signers, so two signers on the same request cannot drift on `POLY_TIMESTAMP` or similar shared-name headers.

## Tests

- 23 new tests covering all three signers, body mutation, schema deserialization, end-to-end pipeline through a real capabilities manifest, and uint256 encoding edge cases.
- Cryptographic round-trips:
  - HMAC: signature recomputed via the `hmac` crate and asserted equal.
  - EIP-712: signature recovered to the signer address via `VerifyingKey::recover_from_prehash` against the recomputed final hash (validates domain separator, type hash, struct hash).
  - NEP-413: signature verified via `ed25519_dalek::Verifier` over `sha256(prefix || borsh-payload)` (validates prefix bytes `[0x9D, 0x01, 0x00, 0x80]` matches `2^31 + 413`, borsh field order, hash).
- Polymarket-clob integration tests load the actual manifest content, resolve dummy secrets through `resolve_host_credentials`, and verify both L1 (`/auth/api-key` EIP-712) and L2 (`/data/orders` HMAC) paths assemble all expected headers correctly.
- uint256 encoding regression test covers zero, value just above 2^128, the 2^256 - 1 maximum, the 2^256 overflow case, and non-numeric / negative / empty input rejection.
- Full lib suite: 5613 passed, 0 failed, 7 ignored.
- `cargo clippy --lib --tests`: zero warnings.

## Rollback

Additive only. New variants on an existing enum, no DB migration, no schema changes, no breaking API changes. Tools that do not declare the new variants are unaffected. Revert is `git revert` of the three commits.

## Spec references

- Polymarket: [py-clob-client signing/eip712](https://github.com/Polymarket/py-clob-client/blob/main/py_clob_client/signing/eip712.py), [py-clob-client headers](https://github.com/Polymarket/py-clob-client/blob/main/py_clob_client/headers/headers.py)
- NEP-413: [near/NEPs#413](https://github.com/near/NEPs/blob/master/neps/nep-0413.md)
- Hyperliquid: [hyperliquid-python-sdk signing.py](https://github.com/hyperliquid-dex/hyperliquid-python-sdk/blob/master/hyperliquid/utils/signing.py)

## Reviewer note

Three commits on the branch (initial feature, review cleanup, automated-review fixes). Happy to split into a 3-PR series (HMAC + closed vocabulary; EIP-712 + NEP-413 header outputs; body mutation + Hyperliquid msgpack) if that is the preferred review shape, let me know.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -988,6 +988,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "base16ct"
+version = "0.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4c7f02d4ea65f2c1853089ffd8d2787bdbc63de2f0d29dedbcf8ccdfa0ccd4cf"
+
+[[package]]
 name = "base64"
 version = "0.21.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2132,6 +2138,18 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "460fbee9c2c2f33933d720630a6a0bac33ba7053db5344fac858d4b8952d77d5"
 
 [[package]]
+name = "crypto-bigint"
+version = "0.5.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0dc92fb57ca44df6db8059111ab3af99a63d5d0f8375d9972e319a379c6bab76"
+dependencies = [
+ "generic-array",
+ "rand_core 0.6.4",
+ "subtle",
+ "zeroize",
+]
+
+[[package]]
 name = "crypto-common"
 version = "0.1.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2376,6 +2394,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9ed9a281f7bc9b7576e61468ba615a66a5c8cfdff42420a70aa82701a3b1e292"
 dependencies = [
  "block-buffer",
+ "const-oid",
  "crypto-common",
  "subtle",
 ]
@@ -2497,6 +2516,20 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d0881ea181b1df73ff77ffaaf9c7544ecc11e82fba9b5f27b262a3c73a332555"
 
 [[package]]
+name = "ecdsa"
+version = "0.16.9"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ee27f32b5c5292967d2d4a9d7f1e0b0aed2c15daded5a60300e4abb9d8020bca"
+dependencies = [
+ "der",
+ "digest",
+ "elliptic-curve",
+ "rfc6979",
+ "signature",
+ "spki",
+]
+
+[[package]]
 name = "ed25519"
 version = "2.2.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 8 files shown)


neo-sky:
Addressed all three review comments from gemini-code-assist in 223e263a (the body) plus a PR-body update (the docs).

- **uint256 overflow (high)**: replaced `u128::parse()` with a hand-rolled decimal-to-32-byte big-endian parser that handles the full `0..(2^256 - 1)` range and rejects overflow past `2^256 - 1` with a clear error. Regression test covers zero, a value just above `2^128`, the `2^256 - 1` max, the `2^256` overflow case, and non-numeric / negative / empty input rejection.
- **Redundant body JSON parsing (medium)**: hoisted into a new `parse_body_json_if_needed` helper called once per signing call. Both `apply_eip712_signing` and `apply_nep413_signing` now parse the request body to `serde_json::Value` once at the top and thread it through `resolve_field_source` and `assemble_bytes_for_keccak`, so multiple `Bytes32Keccak256OfBytes` fields no longer re-parse.
- **Single-struct EIP-712 limitation (medium)**: documented in the PR body's Architecture section. Manifest format already accepts `Vec<Eip712StructDef>`, so nested-struct dependency resolution is an additive follow-up rather than a breaking change to either the manifest or the runtime API.

Full lib suite still green (5613 passed, 0 failed, 7 ignored). Clippy clean.

Issue #3036, opened 2026-04-28
[EPIC] Configuration-as-Code for IronClaw Reborn: tenant blueprints and use-case harnesses


ilblackdragon:
## Problem / Motivation

Two adjacent classes of operators want to declaratively configure IronClaw, and today both have to hand-edit a mix of `.env`, workspace docs under `.system/...`, settings JSON, extension installs, and runtime flags — with no schema, no diff, no audit trail, and no way to source-control the result.

**Multi-tenant clients** want each tenant to ship with:
- a custom system prompt (per tenant / per project / per agent),
- a curated default tool/extension set,
- pre-seeded missions and projects,
- providers/models pinned to the tenant's contract,
- a runtime profile + approval policy compatible with their compliance posture.

**Single-tenant operators** want to run IronClaw as a "use-case OS" — one binary, one host, but specialized into a vertical. Two concrete asks:
- **Red-team harness** — an offensive-security operator persona with reconnaissance/scanning tools, evidence/findings memory schema, sandboxed runtime profile, and a report generator.
- **On-chain incident response** — a "chain going down" detector with chain-status routines, transaction-trace capabilities, runbook execution, and channel-routed pages.

In both cases the configuration must be: (a) declarative, (b) versioned/source-controllable, (c) idempotently appliable, (d) safe across upgrades, and (e) honor Reborn authority/scope/redaction invariants.

This epic treats those two asks as one problem with two product surfaces: a **tenant blueprint** (operator-authored declarative config) and a **harness manifest** (a higher-level composition of extensions+skills+prompts+capability-surface filters that defines a use-case mode).

## Reborn substrate this builds on

This epic is downstream of the contract freeze in `docs/reborn/contracts/_contract-freeze-index.md`, specifically:

- `settings-config.md` — typed settings repository as source of truth, with optional `/system/settings/...` projections; bootstrap vs DB-backed vs secrets vs extension-config layers; precedence; admin-only flags.
- `extensions.md` — extension package format, lifecycle (`discovered → installed → authenticated → configured → active`), trust ceiling assignment.
- `runtime-profiles.md` — `DeploymentMode + RuntimeProfile` resolves to `EffectiveRuntimePolicy`; profile cannot exceed deployment-mode authority ceiling.
- `lightweight-agent-loop.md` — `AgentLoopHost::build_instruction_bundle` is where any prompt overlay lands; the loop never reads system-prompt settings directly.
[...]

PR #3243, opened 2026-05-04
feat(runtime-policy): runtime presets + effective policy (PRs 1-7 of #3045)

### LINKED ISSUE #3045: [Reborn] Add runtime presets and effective runtime policy
## Plain-English summary

Reborn needs a small, understandable **runtime preset** layer so operators and users can choose an intended operating mode without hand-wiring low-level grants, mounts, process backends, network policy, secret policy, approval policy, and audit settings.

Examples:

```bash
ironclaw code . --profile local-dev
ironclaw serve --profile hosted-safe
ironclaw serve --profile enterprise-dev
ironclaw run --profile sandboxed
```

Each profile expands to an `EffectiveRuntimePolicy` that the host runtime enforces through the normal Reborn path:

```text
Agent/user action
  -> visible capability surface filter
  -> CapabilityHost / authorization / approvals
  -> RuntimeDispatcher
  -> HostRuntime planner
  -> selected filesystem/process/network/secret backends
  -> process/resource/audit/event records
```

This is the general version of the local-dev follow-up in #3044 and the runtime-profile foundation needed by the configuration-as-code epic in #3036.

## Problem

Reborn has the right low-level primitives:

```text
capability descriptors
grants and leases
mount views
network policy
resource scopes/ceilings
approval policy
events/audit
runtime dispatch
filesystem/process/runtime backends
```

But those primitives are too low-level for users and operators to select directly.

A local developer should not need to understand every grant/mount/backend type to run tests locally. A hosted operator should not need to hand-compose every tenant-safe sandbox setting. An enterprise admin should be able to pick an organization-approved dedicated-runner mode.

We need an explicit preset/resolver layer:

```text
DeploymentMode + requested RuntimeProfile + tenant/org policy
  -> EffectiveRuntimePolicy
```

The resolver must be monotonic with respect to safety:

```text
Deployment mode and tenant/org policy may reduce requested authority.
They must never increase authority.
```

### `DeploymentMode`

Where is IronClaw running, and who owns the machine boundary?

```rust
pub enum DeploymentMode {
    LocalSingleUser,
    HostedMultiTenant,
    EnterpriseDedicated,
}
```

### `RuntimeProfile`

What policy preset did the operator/user request?

```rust
pub enum RuntimeProfile {
    SecureDefault,
    LocalSafe,
    LocalDev,
    LocalYolo,
    HostedSafe,
    HostedDev,
    HostedYoloTenantScoped,
    EnterpriseSafe,
    EnterpriseDev,
    EnterpriseYoloDedicated,
    Sandboxed,
    Experiment,
}
```

Names can change, but the distinction matters.

### `EffectiveRuntimePolicy`

The resolved policy the runtime actually enforces.

```rust
pub struct EffectiveRuntimePolicy {
    pub deployment: DeploymentMode,
    pub requested_profile: RuntimeProfile,
    pub resolved_profile: RuntimeProfile,
    pub filesystem_backend: FilesystemBackendKind,
    pub process_backend: ProcessBackendKind,
    pub network_mode: NetworkMode,
    pub secret_mode: SecretMode,
    pub approval_policy: ApprovalPolicy,
    pub audit_mode: AuditMode,
}
```

Example backend/policy enums:

```rust
pub enum FilesystemBackendKind {
    ScopedVirtual,
    HostWorkspace,
    TenantWorkspace,
    OrgDedicatedWorkspace,
}

pub enum ProcessBackendKind {
    None,
    Docker,
    Srt,
    SmolVm,
    LocalHost,
    TenantSandbox,
    OrgDedicatedRunner,
}

pub enum NetworkMode {
    Deny,
    Brokered,
    Allowlist,
    DirectLogged,
    Direct,
}
```

Exact names/types can be refined during implementation.

## Why this is not `RuntimeKind::Local`

Do **not** add `RuntimeKind::Local`.

`RuntimeKind` answers what kind of work is being performed:

```text
WASM
Script
MCP
FirstParty/System service
ActionScript/agent loop/etc.
```

Local/hosted/sandboxed answers where/how that work is executed and with what authority.

Preferred model:

```text
RuntimeKind::Script
ProcessBackendKind::LocalHost
RuntimeProfile::LocalDev
```

Not:

```text
RuntimeKind::Local
```

Also do not let user or extension manifests self-select provider-host execution with:

```toml
runner = "local_host"
```

Backend/profile selection is host policy, not manifest authority.

### `ironclaw_host_api`

Own shared, serializable contract vocabulary:

```text
DeploymentMode
RuntimeProfile
EffectiveRuntimePolicy
FilesystemBackendKind
ProcessBackendKind
NetworkMode
SecretMode
ApprovalPolicy
AuditMode
```

These types are used by settings/config, host runtime, audit/events, and tests.

### Runtime policy resolver

Either create a small crate:

```text
crates/ironclaw_runtime_policy
```

or place under:

```text
crates/ironclaw_host_runtime/src/policy.rs
```

This layer owns:

```text
DeploymentMode + requested RuntimeProfile + tenant/org policy
  -> EffectiveRuntimePolicy
```

It also owns fail-closed profile validation.

### `ironclaw_authorization` / future `ironclaw_capabilities`

Own authority decisions:

```text
Can this principal invoke this capability?
Does the grant allow this effect?
Does the grant allow this mount/network/secret/resource scope?
Does this need approval?
```

Runtime profiles do not replace authorization.

### `ironclaw_host_runtime`

Own runtime planning/composition:

```text
ExecutionContext + CapabilityDescriptor + GrantConstraints + EffectiveRuntimePolicy
  -> concrete execution plan
```

Examples:

```text
LocalDev + filesystem.read -> HostWorkspace read under selected root
HostedDev + shell.run      -> tenant sandbox process, not provider-host shell
EnterpriseDev + process.run -> org-dedicated runner if org policy permits
Experiment + package install -> disposable SmolVM/Docker workspace
```

### Runtime crates (`ironclaw_scripts`, `ironclaw_wasm`, `ironclaw_mcp`)

Own execution mechanics only.

They should not decide:

```text
who gets LocalHost
who gets provider-host access
who gets tenant sandbox access
who can inherit environment
who can access secrets/network
```

They receive already-authorized/planned execution requests.

### `SecureDefault`

Safe default assistant mode.

```text
filesystem: scoped virtual / declared mounts
process: sandbox or disabled
network: brokered/allowlisted
secrets: brokered handles only
approval: policy-driven
```

### `LocalSafe`

Cautious local coding mode.

```text
filesystem: selected host workspace read, ask on writes
process: local shell ask-by-default
network: brokered or direct-logged
secrets: scrubbed env, explicit leases only
approval: ask for writes/shell/network/destructive actions
```

### `LocalDev`

Default local coding-agent mode.

```text
filesystem: selected host workspace read/write
process: local host shell
network: direct-logged or brokered
secrets: limited/scrubbed env unless explicitly granted
approval: allow common dev work, ask for dangerous actions
```

Common dev work examples:

```text
cargo test
cargo build
npm test
rg / grep
ls / find
```

Dangerous actions that should ask:

```text
rm -rf
git push / git push --force
sudo
chmod/chown
curl | sh
package publish
outside-workspace writes
secret/env inspection
```

### `LocalYolo`

Explicit trusted-laptop mode.

```text
filesystem: direct selected workspace / optional broader host access
process: local host shell direct
network: direct or direct-logged
approval: minimal or user-configured
requires: explicit startup confirmation and visible warning
```

Even LocalYolo keeps:

```text
audit
timeouts
cancellation
output caps
redaction of obvious secrets from logs
```

### `HostedSafe`

Hosted multi-tenant default.

```text
filesystem: tenant workspace read, ask on writes
process: tenant-scoped sandbox only
network: brokered/allowlisted only
secrets: tenant-scoped broker only
approval: ask for writes/process/external effects
```

### `HostedDev`

Hosted developer mode inside tenant boundary.

```text
filesystem: tenant workspace read/write
process: tenant-scoped sandbox
network: brokered/allowlisted only
secrets: tenant-scoped broker only
approval: allow tenant dev work, ask for destructive/external effects
```

### `HostedYoloTenantScoped`

Reduced approvals inside tenant sandbox, never provider-host authority.

```text
filesystem: tenant workspace direct
process: tenant-scoped sandbox direct
network: brokered/allowlisted only
secrets: tenant-scoped broker only
approval: fewer approvals inside tenant boundary
```

### `EnterpriseSafe` / `EnterpriseDev` / `EnterpriseYoloDedicated`

Enterprise dedicated infrastructure modes.

```text
filesystem: org-dedicated workspace
process: org-dedicated runner/container/VM
network: org policy / brokered
secrets: org KMS/broker
approval: org admin policy
```

Enterprise can be more permissive than hosted multi-tenant only because the infrastructure is dedicated to one organization and controlled by org admin policy.

### `Sandboxed`

General safe helper-process mode.

```text
filesystem: scoped/read-only mount + scratch
process: SRT/Docker/SmolVM
network: brokered/allowlisted
secrets: brokered handles only
approval: policy-driven
```

### `Experiment`

Disposable package-install/test/benchmark mode.

```text
filesystem: copy-in or read-only repo + sandbox overlay
process: SmolVM or Docker
network: allowlisted/brokered
secrets: brokered handles only
approval: ask before host patch/apply
```

## Required fail-closed rules

- [ ] Hosted multi-tenant deployments cannot select `Local*` profiles.
- [ ] Hosted multi-tenant deployments cannot expose provider-host filesystem or provider-host shell.
- [ ] Hosted yolo means tenant-sandbox yolo, never provider-host yolo.
- [ ] Enterprise direct runner modes require enterprise dedicated deployment and org admin policy.
- [ ] Local profiles require local single-user deployment and selected workspace root.
- [ ] Yolo profiles require explicit selection and visible disclosure.
- [ ] Profile changes during a turn require a new turn or explicit approval boundary.
- [ ] Generated/user extensions cannot declare their own profile; they declare runtime needs, host chooses backend/profile.

## Visible capability surface

Runtime presets should affect which capabilities are shown to the model before the model call.

Rules:

- [ ] Profile-impossible capabilities are hidden before the model call.
- [ ] Hosted multi-tenant surfaces omit provider-host filesystem and LocalHost shell affordances entirely.
- [ ] Enterprise surfaces include org-dedicated runner affordances only when org policy enables them.
- [ ] Local surfaces include HostWorkspace/LocalHost affordances according to LocalSafe/LocalDev/LocalYolo.
- [ ] Capabilities that are possible but approval/auth/resource-dependent may remain visible and fail structurally at action time.
- [ ] Action-time authorization remains mandatory even if a capability is visible.

## Configuration surfaces

This should be selectable from multiple surfaces:

### CLI

```bash
ironclaw code . --profile local-dev
ironclaw code . --profile local-yolo --confirm-host-access
ironclaw serve --profile hosted-safe
ironclaw serve --profile enterprise-dev
```

### Settings / DB-backed config

Typed runtime settings should store:

```text
deployment_mode
requested_runtime_profile
approval_policy override, if allowed
selected workspace root / tenant workspace / org workspace
```

### Blueprints (#3036)

Blueprints should be able to declare:

```toml
[runtime]
profile = "HostedDev"
approval_policy = "ask_destructive"
```

or:

```toml
[runtime]
profile = "LocalDev"
approval_policy = "ask_destructive"
```

Blueprint apply validates profile/deployment compatibility and writes typed settings. Runtime should read typed settings/effective policy, not the blueprint file.

### Legacy compatibility

Current local behavior:

```bash
ALLOW_LOCAL_TOOLS=true
```

can map to:

```text
DeploymentMode::LocalSingleUser + RuntimeProfile::LocalDev
```

Current explicit full-access config can inform LocalYolo only with explicit confirmation:

```bash
SANDBOX_POLICY=full_access
SANDBOX_ALLOW_FULL_ACCESS=true
```

### PR 1 — Contract vocabulary

Scope:

```text
crates/ironclaw_host_api
```

Add shared types for deployment/profile/effective policy/backends/modes.

### PR 2 — Resolver crate/module

Scope:

```text
crates/ironclaw_runtime_policy
# or crates/ironclaw_host_runtime/src/policy.rs
```

Implement profile resolution and monotonic safety tests.

### PR 3 — Settings/config/CLI selection

Scope:

```text
src/config
src/setup
CLI entrypoints
settings repository
```

Allow selecting requested runtime profile and deployment mode.

### PR 4 — Visible capability surface filter

Scope:

```text
agent/action projector / host runtime visible capabilities
```

Hide profile-impossible actions before the model call.

### PR 5 — Host runtime planner integration

Scope:

```text
crates/ironclaw_host_runtime
```

Translate `EffectiveRuntimePolicy` + authorized grants into concrete execution plans.

### PR 6 — Local coding profile vertical slice

Scope:

```text
HostWorkspace filesystem
LocalHost process backend
coding aliases
approval presets
```

This is the focused local follow-up tracked in #3044.

### PR 7 — Hosted/enterprise profile enforcement

Scope:

```text
tenant sandbox enforcement
enterprise dedicated runner policy
regression tests proving provider-host access is impossible in hosted mode
```

### PR 8 — Blueprint/harness integration

Scope:

```text
#3036
```

Blueprints/harnesses can declare runtime profile constraints and capability-surface filters.

### Resolver

- [ ] All standard profiles resolve for valid deployment modes.
- [ ] Invalid deployment/profile combinations fail closed.
- [ ] Deployment/org/tenant policy can reduce but never increase authority.
- [ ] LocalYolo / HostedYoloTenantScoped / EnterpriseYoloDedicated require explicit selection and disclosure.
- [ ] Resolver output is deterministic and serializable for audit/debugging.

### Runtime integration

- [ ] Host runtime uses `EffectiveRuntimePolicy` to choose filesystem/process/network/secret backends.
- [ ] Runtime crates receive planned/authorized execution requests; they do not own profile authority decisions.
- [ ] Profile changes cannot silently affect an in-flight turn.

### Capability surface

- [ ] LocalHost/provider-host capabilities are visible only in local profiles.
- [ ] Hosted multi-tenant never exposes provider-host capabilities.
- [ ] Enterprise dedicated exposes dedicated-runner capabilities only when org policy permits.
- [ ] Action-time authorization still runs for every visible action.

### Local profile tests

- [ ] `LocalSingleUser + LocalDev` can run a `cargo test`-style command under selected workspace root.
- [ ] `LocalSingleUser + LocalYolo` requires explicit confirmation.
- [ ] Environment scrubbing remains on by default.

### Hosted/enterprise tests

- [ ] `HostedMultiTenant + LocalDev` fails closed.
- [ ] `HostedMultiTenant + LocalYolo` fails closed.
- [ ] Hosted profiles cannot resolve to provider-host filesystem or provider-host shell.
- [ ] Enterprise direct-runner modes require `EnterpriseDedicated` and org admin policy.

### Config/blueprint tests

- [ ] CLI selected profile reaches resolver and affects visible capabilities.
- [ ] DB-backed setting selected profile reaches resolver.
- [ ] Blueprint `[runtime] profile = ...` validates against deployment mode and writes typed settings.
- [ ] Invalid blueprint profile/deployment pair fails before any writes.

## Non-goals

- Do not replace authorization/grants/leases/approvals with runtime profiles.
- Do not add `RuntimeKind::Local`.
- Do not allow user manifests to select LocalHost/provider-host execution.
- Do not bypass CapabilityHost for local, hosted, or enterprise profiles.
- Do not make provider-host filesystem/shell reachable from hosted multi-tenant deployments.
- Do not solve hosted-control-plane-to-local-runner/BYO-runner architecture here.
- Do not complete the full blueprint/harness epic here; integrate with #3036 after the profile layer exists.

## Related issues

- #2987 — Reborn substrate landing tracker.
- #3036 — configuration-as-code / tenant blueprints / harnesses.
- #3044 — local developer runtime profile vertical slice.

nickpismenkov:
## Summary

Implements PRs 1–7 of the 8-PR decomposition in #3045. PR 8 (blueprint/harness integration with #3036) is explicitly out of scope.

After this PR, the runtime policy substrate is no longer dormant: the model-facing tool list path goes through the visibility filter in production, and the host runtime planner produces a concrete execution plan from `EffectiveRuntimePolicy + CapabilityDescriptor`.

### PR 1 — Contract vocabulary (`ironclaw_host_api::runtime_policy`)
`DeploymentMode`, `RuntimeProfile`, `EffectiveRuntimePolicy`, `FilesystemBackendKind`, `ProcessBackendKind`, `NetworkMode`, `SecretMode`, `ApprovalPolicy`, `AuditMode`. All eight wire-stable enums marked `#[non_exhaustive]` so downstream callers don't break on every variant addition.

### PR 2 — Resolver crate (`ironclaw_runtime_policy`)
797-line resolver with the full compatibility matrix, family ranks, ceiling narrowing, yolo gating (disclosure + org-admin), and a complete `(deployment, profile) → backends` mapping. 31 unit tests including the critical security property `hosted_family_never_resolves_to_provider_host_filesystem_or_shell`.

### PR 3 — Settings/CLI selection
`Config::with_runtime_overrides`, CLI flags `--deployment-mode` / `--runtime-profile` / `--yolo-disclosure`, env-var fallback (`IRONCLAW_DEPLOYMENT_MODE`, `IRONCLAW_RUNTIME_PROFILE`, `IRONCLAW_YOLO_DISCLOSURE`), and resolved `RuntimeConfig` carried through `Config`.

### PR 4 — Visible capability surface filter, **now activated**
`runtime_filter::is_visible_under(policy, affordance)` and the new `ToolRegistry::tool_definitions_visible_under(policy)` are the production tool-list path. `AgentDeps.runtime_policy` carries the resolved policy from `main.rs`; `dispatcher.rs` routes the chat turn's tool list through the filter.

### PR 5 — Host runtime planner integration
New `ironclaw_host_runtime::planner` module exposes `plan_capability(&CapabilityDescriptor, &EffectiveRuntimePolicy) -> Result<ExecutionPlan, PlannerError>`. Pure function; planner fails closed when a capability declares effects (`SpawnProcess`, `Network`, `UseSecret`) the resolved policy disables. Public API: `ExecutionPlan`, `PlannerError`, `plan_capability`.

### PR 6 — Local profile vertical slice
Integration tests in `crates/ironclaw_host_runtime/tests/runtime_policy_planner_contract.rs` drive `LocalSingleUser + LocalDev` through resolver → planner and assert `HostWorkspace` + `LocalHost` for the canonical filesystem.read / shell.cargo_test / shell.npm_test / shell.ripgrep / shell.git_status coding aliases. `LocalSafe` approval-preset assertion (`AskWrites`) locks in the cautious-mode mapping.

### PR 7 — Hosted/enterprise profile enforcement
- `HostedDev` + shell.run plans against `TenantSandbox`, never `LocalHost`.
- `HostedDev` filesystem.write plans against `TenantWorkspace`, never `HostWorkspace`.
- `HostedYoloTenantScoped` with disclosure ack still cannot reach LocalHost/HostWorkspace.
- `EnterpriseDev` plans against `OrgDedicatedRunner`.
- `EnterpriseYoloDedicated` requires both `EnterpriseDedicated` deployment AND `org_policy.admin_approves_dedicated_yolo = true` — without admin approval the resolver fails closed.
- `Experiment + package install` resolves to `SmolVm` / `Allowlist` per the issue example.
- `HostedMultiTenant + LocalDev` fails closed at the resolver.

## zmanian review address

| Item | Resolution |
|---|---|
| **Scope-claim mismatch** ("vocabulary only") | Description rewritten — PRs 1-7 implemented, PR 8 excluded. |
| **H1 dead-config / merge-leakage risk** | Visibility filter now invoked from the dispatcher; `Registry::list_visible_under` and `tool_definitions_visible_under` have production callers. |
| **Medium: `#[non_exhaustive]` missing on 8 enums** | Annotated. Resolver match arms updated with fail-loud wildcard panics. |
| **Medium: `was_reduced()` doc/reality mismatch** | Tightened to "tenant/org policy ceiling reduced authority within the same family." |
| **Low: `EnterpriseYoloDedicated` approval variance** | Doc note + dedicated test (`enterprise_yolo_dedicated_uses_org_policy_approvals_not_minimal`). |
| **Low: two things named `OrgPolicy`** | Struct renamed `OrgPolicy` → `OrgPolicyConstraints`; enum variants `ApprovalPolicy::OrgPolicy` / `AuditMode::OrgPolicy` unchanged (wire-stable). |
| **Low: `ParseRuntimePolicyEnumError` Hash/Copy** | `Hash` derived; `Copy` intentionally not (carries owned `String`). |

## Test gap fills (zmanian's seven)

1. **CLI > env > default precedence per field** — `precedence_chain_cli_over_env_over_default_is_resolved_per_field`.
2. **Full pipeline Config → policy → tool filter** — `tests/runtime_policy_tool_visibility_integration.rs::full_config_pipeline_carries_resolved_policy_to_tool_filter`.
3. **Binding test that `is_visible_under` is called from the model-facing path** — same integration file, three scenarios (LocalDev visible, HostedDev never resolves to LocalHost/HostWorkspace, SecureDefault hides AnyProcess tools).
4. **Serde round-trip for `OrgPolicyConstraints` / `ResolveRequest`** — `ResolveRequest` now derives `Serialize`/`Deserialize`; round-trips verified.

## Mergeability

- All 200+ `runtime_policy` and `host_runtime` tests pass; 5 new integration tests pass.
- Zero clippy warnings; `cargo fmt --all` clean.
- Recently merged `reborn-integration` — only Cargo.toml conflict (kept both `ironclaw_turns` and `ironclaw_runtime_policy` workspace members).

## Test plan

- [ ] `cargo test -p ironclaw_runtime_policy` (resolver unit tests, including new yolo variance + serde round-trips)
- [ ] `cargo test -p ironclaw_host_runtime --test runtime_policy_planner_contract` (PR 5/6/7 contract tests, 10 cases)
- [ ] `cargo test --test runtime_policy_tool_visibility_integration` (Config → policy → tool filter, 5 cases)
- [ ] `cargo test -p ironclaw --lib config::runtime` (CLI/env precedence, 9 cases)
- [ ] `cargo clippy --workspace --all-features --tests` zero warnings
- [ ] `cargo fmt --all -- --check` clean


zmanian (commented):
**Verdict: APPROVE-WITH-NITS** — code is high quality with strong tests, but the PR description is materially misleading and there's one real concern (dead-config) plus a forward-compat miss.

## Scope-claim mismatch

Description says "**Vocabulary only — no resolver logic, no behavior changes.**" Reality:

1. ✅ Vocabulary in `crates/ironclaw_host_api/src/runtime_policy.rs` (~819 lines) — matches.
2. ❌ **A fully implemented resolver** in `crates/ironclaw_runtime_policy/src/resolver.rs` (~797 lines) with compatibility matrix, family ranks, ceiling narrowing, yolo gating, and a complete `(deployment, profile) → backends` mapping — with 17 tests.
3. ❌ **CLI flags** added (`--deployment-mode`, `--runtime-profile`, `--yolo-disclosure`) with env-var fallback.
4. ❌ **Config plumbing** in `src/config/runtime.rs` and `Config::with_runtime_overrides` runs the resolver at startup and stores `EffectiveRuntimePolicy` on every `Config`.
5. ❌ **Tool visibility filter** (`src/tools/runtime_filter.rs`) and two new `Registry` methods (`list_visible_under`, `all_visible_under`).
6. ❌ **Shell tool override** declaring `runtime_affordance() = AnyProcess`.

In practice runtime behavior is unchanged because **no production caller invokes `list_visible_under` / `all_visible_under` or reads `Config.runtime.effective_policy`** — items 5–6 are additive on a parallel dormant path. That's defensible engineering, but the description should say "scaffolding + dormant resolver/filter, no live behavior change yet (PR 1 of #3045)." The "no resolver logic" claim is just wrong — there's a 797-line resolver.

## High

- **Dead-config / merge-leakage risk.** `Config.runtime.effective_policy` is computed and stored, but the only consumers are tests. Two new registry methods have no production callers. If subsequent #3045 PRs slip, this PR ships latent code with all the maintenance burden and none of the behavior. Fix: rewrite the PR description to admit this and link the follow-up PR(s); consider `#[cfg(feature = "...")]` if integration isn't imminent.
  - `src/config/runtime.rs:55`, `src/tools/registry.rs:375,423`.

## Medium

- **Wire-stable enums missing `#[non_exhaustive]`.** All eight enums (`DeploymentMode`, `RuntimeProfile`, `FilesystemBackendKind`, `ProcessBackendKind`, `NetworkMode`, `SecretMode`, `ApprovalPolicy`, `AuditMode`) are public, intended to grow (doc comments describe future variants). Without `#[non_exhaustive]`, downstream exhaustive matches break on every addition. Internal exhaustive matches in the defining crate still work after annotation.
  - `crates/ironclaw_host_api/src/runtime_policy.rs:65, 129, 330, 366, 410, 449, 492, 530`.

- **`was_reduced()` doc/reality mismatch.** Returns `requested != resolved`, correct for within-family ceiling narrowing. But the resolver's other failure mode for an over-permissive request is `ResolveError::IncompatibleDeployment` (fail-closed), not narrowing — so deployment-mode reduction is impossible by construction. Doc says "deployment or tenant/org policy reduced authority"; deployment-mode reduction can't happen.
  - Fix: tighten doc to "tenant/org policy ceiling reduced authority within the same family."
  - `crates/ironclaw_host_api/src/runtime_policy.rs:580-587`.

## Low / nits

- `EnterpriseYoloDedicated` is the only `*Yolo*` variant that doesn't map to `ApprovalPolicy::Minimal` (uses `OrgPolicy`). A consumer who assumes "yolo ⇒ Minimal" will be wrong here. Add a doc note or test asserting the variance.
  - `crates/ironclaw_runtime_policy/src/resolver.rs:344-354`.
- Two things named `OrgPolicy`: enum variants `ApprovalPolicy::OrgPolicy` / `AuditMode::OrgPolicy` and the struct `resolver::OrgPolicy`. Not ambiguous but trippy. Consider `OrgPolicyConstraints` for the struct.
- `ParseRuntimePolicyEnumError` doesn't impl `Hash`/`Copy` while the parsed enums do.

## Mergeability

Conflicts only in `Cargo.lock` + `Cargo.toml` — workspace member additions + lockfile drift. Trivial rebase.

## Test gaps

1. No integration test for the precedence chain `CLI flag > env var > default` end-to-end through `Config::with_runtime_overrides`.
2. No test that the full pipeline `Config → policy → tool filter` enforces the security property end-to-end.
3. No binding test that `runtime_filter::is_visible_under` is actually called from the model-facing tool-list path. Without it, future refactors silently bypass the filter — the actual #3045 contract.
4. No serde round-trip test for `OrgPolicy` / `ResolveRequest`.

## Spot-checked and correct

- `RuntimeProfile::is_local` / `is_hosted` / `is_enterprise` use exhaustive matches (not `matches!`); all 12 variants accounted for. `Sandboxed`/`Experiment`/`SecureDefault` correctly report `false` for all three families. Partition test (`runtime_profile_family_predicates_partition_correctly`) asserts ≤1 match.
- All eight enums have `as_str()` matching the serde wire name; verified by `*_as_str_matches_serde_wire_name` tests.
- `FromStr` is the inverse of `as_str()` for `DeploymentMode` / `RuntimeProfile`.
- **Critical security property** — hosted multi-tenant never produces `HostWorkspace` or `LocalHost` — is asserted directly (`hosted_family_never_resolves_to_provider_host_filesystem_or_shell`), including `Sandboxed` and yolo cases.
- The monotonic-safety ceiling rule (ceilings can never raise authority) is asserted (`org_policy_ceiling_above_request_in_same_family_keeps_request`).
- Enterprise-yolo gating is correctly two-stage: disclosure → org admin approval.
- `Tool::runtime_affordance` defaults to `None` (fail-open for visibility) — correct UX choice since action-time auth still gates execution; documented as such.
- Shell tool's `AnyProcess` correctly hides shell under `SecureDefault` and exposes it under all real process backends.

zmanian (changes requested):
**Verdict: REQUEST CHANGES (one real correctness bug)** — the resolver/planner/contract layers are high-quality and the prior reviewer's seven items are genuinely addressed, but the headline claim that "the model-facing tool list path goes through the visibility filter in production" only holds for **iteration 1**. Subsequent LLM iterations and every background worker bypass the filter, so the security property is partial.

## Overview

Implements PRs 1–7 of the 8-PR decomposition in #3045: the runtime-policy contract vocabulary (`ironclaw_host_api::runtime_policy`), a 911-line resolver crate (`ironclaw_runtime_policy`), the host-runtime planner, settings/CLI selection, the visible-capability-surface filter, and per-deployment vertical-slice tests. CI is green (fmt, clippy `--all-features`, "No panics in production code" lint, cargo-deny).

## Strengths

- **Resolver fail-closed contract is well-tested.** `hosted_family_never_resolves_to_provider_host_filesystem_or_shell` (`crates/ironclaw_runtime_policy/src/resolver.rs:643`) covers `SecureDefault`, `HostedSafe`, `HostedDev`, `Sandboxed`, and the yolo case under disclosure. Invariant holds by construction at the `is_compatible` matrix step (`resolver.rs:151-176`), not just at runtime.
- **Yolo gating is correctly two-stage**: disclosure first (`resolver.rs:122-126`), then `EnterpriseYoloDedicated`'s extra org-admin gate (`resolver.rs:128-132`). `enterprise_yolo_dedicated_uses_org_policy_approvals_not_minimal` (`resolver.rs:587`) locks in the variance flagged in the prior review.
- **`#[non_exhaustive]` discipline is consistent.** All eight enums annotated; downstream `match` arms in resolver fail loudly via `panic!` (`resolver.rs:398, 413, 422, 431`) — defensible since the wildcards are unreachable today and trip only on a future enum-variant addition. CI's "No panics in production code" lint allows them.
- **Predicate methods use exhaustive matches** (`is_local`, `is_hosted`, `is_enterprise`, `is_yolo` in `runtime_policy.rs:202-275`) so a new variant forces a compile-time decision — gemini-code-assist's specific feedback addressed.
- **Serde round-trips covered** for both `ResolveRequest` (`resolver.rs:834`) and `OrgPolicyConstraints` (`resolver.rs:812`).

## Issues

### High — Visibility filter applies to iteration 1 only

`src/agent/dispatcher.rs:264-269` correctly routes the *first* `tool_definitions` build through `tool_definitions_visible_under(policy)`. But the per-iteration refresh inside `before_llm_call` at `src/agent/dispatcher.rs:465` reverts to the unfiltered `tool_definitions()`:

```rust
// Refresh tool definitions each iteration so newly built tools become visible
let tool_defs = self.agent.tools().tool_definitions().await;
```

So after the first model turn, the LLM sees the full unfiltered tool list — including `shell` under hosted-multi-tenant if it had been hidden initially. **The "hosted multi-tenant surfaces never expose provider-host shell to the model" property is only enforced on iteration 1.**

Same gap in:
- `src/worker/container.rs:163, 415` (sub-agent / container worker reasoning context)
- `src/worker/job.rs:343, 1439` (background job worker)
- `src/agent/routine_engine.rs:1906` (routine engine)
- `src/channels/web/features/extensions/mod.rs:233` (UI tool listing — arguably should also filter)

The fix is mechanical: thread `runtime_policy` to those sites and use `tool_definitions_visible_under`. Recommend either (a) doing it in this PR (small additional surface), or (b) explicitly scoping it as a follow-up and rewriting the "model-facing tool list path goes through the visibility filter in production" claim to match.

### Medium — Two of the three new registry methods have zero callers

`ToolRegistry::list_visible_under` (`src/tools/registry.rs:375`) and `all_visible_under` (`src/tools/registry.rs:423`) are unreferenced anywhere in the tree (verified with `git grep`). The prior review explicitly flagged dormant code as "H1 dead-config / merge-leakage risk" — only `tool_definitions_visible_under` actually escaped the dormant set. Either delete the other two or call them out explicitly as scaffolding for the planner integration in a follow-up.

### Medium — Tool affordance coverage is shell-only

Only `ShellTool` declares `runtime_affordance() = AnyProcess` (`src/tools/builtin/shell.rs:1054-1062`). Every other tool that touches host filesystem (`file_read`, `file_write`, `apply_patch`, `list_dir`), network (`http`, `web_fetch`), or process (any built tool runner) defaults to `ToolRuntimeAffordance::None` and is therefore visible under every policy. So even the iteration-1 filter mostly only hides shell. The PR description's "hosted multi-tenant deployments cannot expose provider-host shell affordances" is technically true but the surface area is narrow. Worth listing the remaining tools in the description as known follow-up so reviewers don't assume broader coverage.

### Low

- **`every_valid_deployment_profile_pair_resolves`** (`resolver.rs:854`) excludes the three yolo profiles (because they need disclosure). The name suggests a complete matrix; rename to `every_non_yolo_deployment_profile_pair_resolves` or extend to cover yolo via `req_yolo`.
- **`EnterpriseYoloDedicated` uses `NetworkMode::DirectLogged`** (`resolver.rs:354`) — wider than `EnterpriseDev`'s `Allowlist`. The doc framing ("yolo never widens authority") is thus slightly aspirational: yolo widens *network*, just with logging. Acceptable, but consider tightening the doc on `RuntimeProfile::EnterpriseYoloDedicated` to "reduces approvals and logs direct-network egress."
- **Planner only fail-closes on three effects** (`SpawnProcess`/`ExecuteCode`, `Network`, `UseSecret`). A `WriteFilesystem` capability against a `ScopedVirtual` (read-only) backend isn't rejected — `plan_capability` just forwards. If that's intentional (write-authorization is gated elsewhere), the module-level rustdoc at `crates/ironclaw_host_runtime/src/planner.rs:22-29` should say so explicitly.
- **CLI flag UX**: `--yolo-disclosure` is a bare bool with no `=true/false` form, so once set it can't be cleared from a parent shell that exports `IRONCLAW_YOLO_DISCLOSURE=true`. Probably fine for a one-way disclosure ack, but worth noting.
- **`safe_default()`** (`src/config/runtime.rs:118-137`) calls `.expect("LocalSingleUser + SecureDefault always resolves")`. Used only in `Config::for_testing` per the PR text; per-CLAUDE.md "tests are fine" rule this is acceptable, and the comment justifies the totality.

## Mergeability

Mergeable state is `dirty` (Cargo.lock conflict, per the PR description). Fix the iteration-2 gap, rebase, and this is good to go.

---

zmanian (approved):
**Verdict: APPROVE** — all six findings from the prior review are addressed with the right surgical fixes. Two minor forward-compat observations and one product call to confirm; none blocking.

## What changed since iter-1

Verified by diffing `c7f02c7..f749280`:

| Prior finding | Status | Verification |
|---|---|---|
| **HIGH iter-2 gap** | ✅ Fixed at all four call sites | `dispatcher.rs:465-481`, `worker/job.rs:351-360 + 1454-1466`, `routine_engine.rs:1923-1936`. `agent_loop.rs:329-335` propagates policy from `AgentDeps` → `Scheduler::set_runtime_policy`; `agent_loop.rs:1041-1045` does the same for `RoutineEngine::set_runtime_policy`. Container + web-extensions paths are documented as scope-limited follow-ups with rationale (sandbox boundary / observability listing) — reasonable. |
| **MED dead methods** | ✅ Deleted | `list_visible_under` and `all_visible_under` removed from `tools/registry.rs`. |
| **MED affordance coverage** | ✅ Extended | `read_file`/`write_file`/`list_dir`/`apply_patch`/`grep`/`glob` declare `HostFilesystem`; `http` declares `DirectNetwork`. |
| **LOW test name** | ✅ Renamed | `every_non_yolo_deployment_profile_pair_resolves`. |
| **LOW yolo doc** | ✅ Tightened | `RuntimeProfile::EnterpriseYoloDedicated` rustdoc now spells out the `DirectLogged` widening and `OrgPolicy` approval exception. |
| **LOW planner doc** | ✅ Documented | Module rustdoc on `planner.rs:36-50` explains why `WriteFilesystem` against `ScopedVirtual` is intentionally not rejected at plan time (per-mount granularity in `CapabilityHost` is richer). |

CI is green where it matters: fmt, "No panics in production code", cargo-deny, gateway boundaries, formatting, regression test enforcement. Clippy `--all-features` was still in_progress at fetch time.

New tests (3): `host_filesystem_tools_are_hidden_under_hosted_dev_policy`, `host_filesystem_tools_are_visible_under_local_dev_policy`, `http_tool_is_hidden_under_brokered_network_modes`. Each drives the affordance/policy pairing through `tool_definitions_visible_under` so a future regression on either side surfaces.

## Remaining minor observations

### Low — No multi-iteration regression test for the dispatcher/worker/routine call sites

The commit message says *"Each fix has a regression test driving the bug shape"*, which is true for the affordance-coverage fixes. But the HIGH per-iteration / worker / routine fixes only have indirect coverage — they rely on the existing `tool_definitions_visible_under` binding tests plus inspection. A test that drives the dispatcher loop past iteration 1 (or a worker job through two LLM turns) and asserts the second-iteration tool list is filtered would catch a future refactor that silently reverts one of these sites to `tool_definitions()`. Not blocking — the call-site fixes are correct and reviewable — but worth a follow-up.

### Low — `set_runtime_policy` setter pattern is forget-able

`Scheduler::set_runtime_policy` (`scheduler.rs:122-129`) and `RoutineEngine::set_runtime_policy` (`routine_engine.rs:184-189`) follow the same setter pattern as `set_http_interceptor`. That's stylistically consistent, but it means a future code path that constructs either type without calling the setter silently uses the unfiltered list — exactly the bug shape just fixed. Constructor parameters (or a `with_runtime_policy` builder) would be safer; the current pattern leans on careful review of every new construction site. Not blocking.

### Product call — `http` is now hidden under several non-hosted profiles

`HttpTool::runtime_affordance() = DirectNetwork` only stays visible under `NetworkMode::{Direct, DirectLogged}`. Effective visibility per resolver mapping:

| Profile | NetworkMode | http visible? |
|---|---|---|
| `LocalSafe` | `Brokered` | hidden |
| `LocalDev` | `DirectLogged` | ✅ visible |
| `LocalYolo` | `Direct` | ✅ visible |
| `Sandboxed` | `Brokered` | hidden |
| `Experiment` | `Allowlist` | hidden |
| `HostedSafe`/`HostedDev`/`HostedYoloTenantScoped` | `Brokered`/`Allowlist` | hidden ✅ |
| `EnterpriseSafe`/`EnterpriseDev` | `Brokered`/`Allowlist` | hidden |
| `EnterpriseYoloDedicated` | `DirectLogged` | ✅ visible |

Hosted/enterprise hiding is the intended security property. But hiding `http` under **`LocalSafe`**, **`Sandboxed`**, **`Experiment`**, and **`EnterpriseDev`** means those profiles ship with no model-facing HTTP capability today — there's no brokered-HTTP tool in the tree to take its place. The inline comment frames this as "broker is the supported route", but that route doesn't exist yet. Worth confirming this is intentional (so e.g. `Experiment` mode for `npm install` truly relies on the package manager rather than direct HTTP) and noting the brokered-HTTP follow-up explicitly so users don't experience silent capability loss when they upgrade and select these profiles.

## Summary

Fixes are surgical, correctly scoped, and the rationales for the deferred container/web-extensions paths are sound. Approving with the suggestion to track the http-under-LocalSafe product call and the multi-iteration regression test as small follow-ups.

---

zmanian (commented):
## Independent Post-Merge Review

> Note: requested as "review PR 3045", but #3045 is the tracking *issue*; #3243 is the implementation. Three prior reviews exist (`gemini-code-assist` and three rounds from @zmanian), so this pass focuses on what hasn't already been said. PR is already merged — flagging as follow-ups rather than blockers.

### Strong points (independent of prior reviews)

- **Planner (`crates/ironclaw_host_runtime/src/planner.rs`) is exemplary**: tight pure function, three explicit fail-closed branches per effect kind, error variants carry `CapabilityId` for audit, and the rustdoc explicitly enumerates what the planner does *not* check (filesystem-write authority — deferred to `CapabilityHost`/`MountView`). That contract being written down in the module header is exactly what future maintainers need.
- The visibility filter at `src/tools/runtime_filter.rs:24-50` is a clean enum-match with no fallthrough — adding a new `ToolRuntimeAffordance` variant breaks the build at every consumer, which is the right failure mode.
- `RuntimeConfig::resolve_from` (`src/config/runtime.rs:69-101`) makes the precedence rule (CLI > env > default) explicit per field, and `precedence_chain_cli_over_env_over_default_is_resolved_per_field` (line 220) locks that contract in with one scenario that exercises all three layers in a single test.

### Observations not already covered

#### Medium — `set_runtime_policy` setter pattern is structurally forget-able

The iter-2 fix landed correctly across `dispatcher`/`worker/job`/`routine_engine`/`scheduler`, but the chosen integration shape is `Scheduler::set_runtime_policy(...)` and `RoutineEngine::set_runtime_policy(...)` called *after* construction. @zmanian flagged this in iter-3; I'd elevate it slightly. The bug pattern just fixed (per-iter loop reverting to unfiltered `tool_definitions()`) recurs structurally any time a future code path constructs a `Scheduler` / `RoutineEngine` and forgets the setter. There's no compile-time enforcement — the type is silently in "no filter" mode.

A `with_runtime_policy` builder or a required constructor parameter would be safer. Worth a follow-up.

#### Medium — No multi-iteration regression test for the dispatcher fix

This is the single highest-leverage missing test. The iter-3 review flagged it; I'd argue it's a project-rules issue per `CLAUDE.md`'s "Test Through the Caller, Not Just the Helper" rule (the helper here is `tool_definitions_visible_under`, the caller-with-effect is the dispatcher loop past iteration 1). A test that drives `dispatcher.before_llm_call` past iteration 1 with a `HostedMultiTenant + HostedSafe` policy and a tool declaring `AnyProcess`, then asserts the second-iteration tool list does *not* contain shell, would lock in the fix the team specifically had to add a second iteration to make.

#### Low — `safe_default()` panic invariant has no guard test

`src/config/runtime.rs:118-137` does `resolve(...).expect("LocalSingleUser + SecureDefault always resolves")`. The comment correctly justifies the totality, and `every_non_yolo_deployment_profile_pair_resolves` indirectly covers it. But if a future change to `is_compatible` makes that pair fall through (entirely plausible if someone narrows the matrix), this silently panics whenever `Config::for_testing` is called. A direct unit test (`safe_default_invariant_is_total`) on the function would be ~3 lines and would catch the drift at the right module.

#### Low — `RuntimeConfig::resolve_from` reads env directly inside the function

`parse_optional_env` and `parse_bool_env_or_default` call `std::env::var` inline. Each test path now needs `unsafe { std::env::set_var(...) }` plus a process-wide `ENV_LOCK`. Works, but couples the function to global process env and makes future composition (e.g. reading a settings record from the DB instead of env) harder than it needs to be. A small `EnvSource` trait taking `&dyn EnvSource` would let tests pass a `HashMap<String, String>` and skip the lock entirely. Flagging because PR 8 (blueprint integration) will likely want to layer DB-backed settings as another precedence tier, and the current direct-env coupling will need to be unwound at that point.

#### Low — `http`-tool visibility loss under several non-hosted profiles is a real product question

Already raised in iter-3. Restating because it deserves attention: `HttpTool` declares `DirectNetwork`, which renders it invisible under `LocalSafe`, `Sandboxed`, `Experiment`, and `EnterpriseDev` because those resolve to `Brokered`/`Allowlist`/etc. There is no brokered-HTTP tool in the tree to take its place. The user-visible result is "I picked `LocalSafe` and the agent can no longer fetch URLs, no error". A no-op tool surfacing a brokered-HTTP-not-implemented hint would prevent silent capability loss until the brokered-HTTP follow-up lands.

#### Nit — `ToolRuntimeAffordance` is not `#[non_exhaustive]`

The wire-stable enums in `runtime_policy.rs` are correctly `#[non_exhaustive]`. `ToolRuntimeAffordance` is internal (lives in `src/tools/tool.rs`) so this isn't a wire concern. As the affordance vocabulary grows (`BrokeredNetwork`, `WriteFilesystem`, `ReadOnlyFilesystem`...) every consumer of `is_visible_under` will need updates — exhaustiveness is the desired failure mode here. Confirming it's a deliberate asymmetry, not an oversight.

### Risks

- **Security property partial coverage**: filesystem affordance is binary (`HostFilesystem` only). `TenantWorkspace`/`OrgDedicatedWorkspace`/`ScopedVirtual` filesystem tools are all "None"-affordance and visible under every policy, relying on action-time auth alone. The PR description's "hosted multi-tenant deployments cannot expose provider-host filesystem" is true (no tool declares `HostFilesystem` under a tenant policy), but the *symmetric* property "tenant-only tools are hidden under local policy" isn't established because there are no tenant-only tools yet. Worth a paragraph in `crates/ironclaw_runtime_policy/CLAUDE.md` about how to declare new affordances correctly.
- **No production caller for the planner**: `plan_capability` has integration tests but no production caller dispatches through it. Stated in the description (PR 8 / blueprint integration is out of scope), but it does mean the planner is dormant code with maintenance burden until the dispatch integration lands.
- **`unsafe { std::env::set_var }` in tests** is correct under the new Rust soundness rule but every new test added to `src/config/runtime.rs` has to remember the `ENV_LOCK` + `clear_env()` ritual. A `with_clean_env(|| ...)` test helper would lower the footgun.

### Concrete follow-ups worth filing

1. **Multi-iteration regression test** in `tests/runtime_policy_tool_visibility_integration.rs` driving the dispatcher loop past iteration 1.
2. **`with_runtime_policy` constructor parameter** for `Scheduler` / `RoutineEngine` to make the policy non-forgettable at compile time.

Everything else is cleanup that can ride along with the next runtime-policy PR.

---

PR #3246, opened 2026-05-05
fix(installer): add Linux/macOS platform support and fix release tag URL

brothers-morrison:
## Summary
- Fix release download URL from `v0.27.0` to `ironclaw-v0.27.0` to match actual GitHub release tag naming
- Add `select_archive_for_arch` mappings for 6 new platforms:
  - `x86_64-unknown-linux-gnu` (Linux x64 glibc)
  - `aarch64-unknown-linux-gnu` (Linux ARM64 glibc)
  - `x86_64-unknown-linux-musl` (Linux x64 musl)
  - `aarch64-unknown-linux-musl` (Linux ARM64 musl)
  - `x86_64-apple-darwin` (macOS Intel)
  - `aarch64-apple-darwin` (macOS Apple Silicon)
- Add archive download case handlers with SHA256 checksums, binary names, and updater names for all 6 platforms
- Linux/macOS binaries are `ironclaw` and `sandbox_daemon` (not Windows `.exe` files)

## Problem
The installer script only handled Windows platforms in `select_archive_for_arch`, causing all Linux and macOS users to get *"there isn't a download for your platform"* error. Additionally, the release tag URL was `v0.27.0` instead of the correct `ironclaw-v0.27.0`.

## Testing
Verified successful download, checksum verification, and installation on Ubuntu 24.04 x64.

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,6 +114,8 @@ irm https://github.com/nearai/ironclaw/releases/latest/download/ironclaw-install
 ```sh
 curl --proto '=https' --tlsv1.2 -LsSf https://github.com/nearai/ironclaw/releases/latest/download/ironclaw-installer.sh | sh
 ```
+
+Supports `x86_64`/`aarch64` on macOS (Darwin), Linux (glibc & musl), and Windows via WSL. Windows (native) also has a [PowerShell installer](#install-via-powershell-script-windows).
 </details>
 
 <details>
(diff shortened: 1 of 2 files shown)


serrrfirat:
Hi @brothers-morrison 👋

Thanks for the installer work — cross-platform support and the release URL tag fix were both real issues. The installer has been rewritten since this was opened, so this PR can't apply as-is.

Closing this now. We'd love to have you contribute again after v1 when everything's more stable! 🙏

PR #3253, opened 2026-05-05
feat: multi-tenant relay channel with per-user identity resolution

PierreLeGuen:
## Summary
- Resolves Slack `sender_id` → internal IronClaw `UserId` via PairingStore on every relay event
- Unpaired Slack users get a pairing code (OTP) reply; enter it in IronClaw web UI to pair
- OAuth callback creates channel identity for the admin who connected Slack
- Workspace-scoped identity: external_id is `team_id:slack_user_id`
- Fetches `authed_user_id` from relay connections API (server-side, not from redirect URL)
- Shared OwnershipCache between gateway and ExtensionManager
- Fail closed: suspended users can't complete OAuth, identity creation failures fail the callback
- Both postgres and libsql backends implement `create_channel_identity`

**Depends on:** nearai/channel-relay#14

## E2E Test Results (2026-05-06)
- [x] Admin (Pierre) OAuth → Slack connected, relay activated
- [x] Admin DMs bot → resolved to internal user, LLM responds in Slack thread
- [x] Unpaired user (Evrard) DMs bot → gets pairing code `WZG8LQAB` in Slack
- [x] Pairing approved via API → subsequent messages resolve to Evrard's IronClaw user
- [x] Multi-tenant mode (`AGENT_MULTI_TENANT=true`) works end-to-end
- [x] Compiles on both postgres and libsql backends
- [x] Zero clippy warnings


src/channels/web/features/oauth/mod.rs
```diff
@@ -719,6 +721,60 @@ pub(crate) async fn slack_relay_oauth_callback_handler(
...
+            && let Some(pairing_store) = ext_mgr.pairing_store()
+        {
+            let user_key = format!("relay:{}:oauth_user", relay_extension_name);
+            let oauth_user = ext_mgr
+                .secrets()
+                .get_decrypted(&state.owner_id, &user_key)
+                .await
+                .ok()
+                .map(|s| s.expose().to_string())
+                .unwrap_or_else(|| state.owner_id.clone());
```
serrrfirat, line 788: **High Severity**

Relay OAuth state is still owner-scoped singleton state, so concurrent or missing-flow callbacks can misattribute Slack users.

`auth_channel_relay()` stores both the CSRF nonce and initiating IronClaw user in owner-scoped singleton secret keys (`relay:{name}:oauth_state` / `relay:{name}:oauth_user`), and this callback falls back to `state.owner_id` if the user secret is missing or unreadable. In a multi-user deployment, user A can start Slack OAuth, then user B starts before A returns; B overwrites A's nonce/user context, so A's callback is rejected, or any path that loses `oauth_user` pairs the Slack `authed_user_id` to the owner/admin. Also, once the owner-scoped `team_id` exists, later users can be treated as already authenticated without getting their own Slack identity pairing.

Please tie the initiating user to the specific OAuth state/flow (e.g. nonce-keyed pending record) and fail closed if that record is missing; do not default to the owner. Separate the workspace/team connection from per-user Slack identity pairing, and add a caller-level test for overlapping auth attempts and missing `oauth_user`.

src/app.rs
```diff
@@ -989,6 +989,13 @@ impl AppBuilder {
             if let Some(ref ss) = settings_store_override {
                 em = em.with_settings_store(Arc::clone(ss));
             }
+            if let Some(ref db) = self.db {
+                let ps = Arc::new(crate::pairing::PairingStore::new(
+                    Arc::clone(db),
+                    Arc::new(crate::ownership::OwnershipCache::new()),
```
serrrfirat: **High Severity**

The relay channel gets a separate pairing cache that admin invalidation paths never evict.

This constructs a new `PairingStore` with a fresh `OwnershipCache`, while the gateway gets a different `PairingStore` backed by `components.ownership_cache`. Admin suspend/delete paths evict only `state.pairing_store`, but live relay channels resolve identities through this extension-manager store. Once a Slack sender is cached here, suspending/deleting/demoting that IronClaw user can leave relay messages resolving with the stale active/admin identity until process restart or cache miss.

Please share one `PairingStore`/`OwnershipCache` across the gateway and relay extension manager, or explicitly evict the extension manager's relay cache on user status/role changes. Add a regression test that primes relay identity resolution, changes the user status/role, then verifies the next relay event does not resolve from stale cache.

src/channels/web/features/oauth/mod.rs
```diff
@@ -719,6 +721,60 @@ pub(crate) async fn slack_relay_oauth_callback_handler(
                 format!("Failed to persist relay team_id: {e}")
             })?;
 
+        // Create channel identity pairing: Slack authed_user_id → IronClaw user.
+        // The oauth_user secret was stored during auth_channel_relay() and identifies
+        // which IronClaw user initiated this OAuth flow.
+        if !authed_user_id.is_empty()
```
serrrfirat: **Medium Severity**

The core multi-tenant relay paths are missing caller-level regression coverage.

This PR adds the handler path that writes `channel_identities` from Slack OAuth and the relay channel path that resolves inbound sender IDs / forwards outbound `slack_user_id`, but the tests do not drive those actual callers. That leaves the highest-risk behavior here unprotected: pairing `authed_user_id` to the initiating IronClaw user, missing/overlapping OAuth user state, inbound sender resolution, and outbound proxy calls including the original Slack sender.

Please add caller-path tests around the Slack callback and `RelayChannel` behavior (not just helper/DB tests), including the failure cases above and the `slack_user_id` query forwarding.

src/channels/relay/channel.rs
```diff
@@ -384,7 +431,7 @@ impl Channel for RelayChannel {
             })?;
         let body = self.build_approval_body(channel_id, thread_id, &prompt, &approval_token);
 
-        self.proxy_send(team_id, "chat.postMessage", body)
+        self.proxy_send(team_id, "chat.postMessage", body, None)
```
serrrfirat, line 475: **High Severity**

The per-user relay routing does not cover approval registration or approval-message posting.

Normal chat responses now forward the original Slack sender as `slack_user_id`, but the approval path still calls `create_approval(team_id, channel_id, thread_id, ...)` without a sender and then posts the approval message with `proxy_send(..., None)`. channel-relay's multi-user approval route resolves the approver from active connections and falls back to the first active connection for the team when no `sender_id` is provided. In a multi-user Slack team, a non-default user's approval can therefore be registered or emitted against the wrong Slack connection, so the approval token is bound to the wrong approver or the button post fails despite normal replies using the right per-user identity.

Please pass the inbound Slack `sender_id` through `create_approval` and the approval `proxy_send`, and have the relay validate it against the active connection for that team. Add a caller-level test where two Slack users are connected and the second user's approval request is registered and posted with that user's Slack ID.

src/channels/web/features/oauth/mod.rs
```diff
@@ -692,6 +692,8 @@ pub(crate) async fn slack_relay_oauth_callback_handler(
         .into_response();
     }
 
+    let authed_user_id = params.get("authed_user_id").cloned().unwrap_or_default();
```
serrrfirat: **High Severity**

The callback trusts `authed_user_id` from the final browser query string before persisting a channel identity.

channel-relay validates Slack and redirects back to IronClaw with `authed_user_id` as a normal query parameter plus the nonce. This handler validates only the nonce, then writes whatever `authed_user_id` is present into `channel_identities`. A user who initiates a legitimate OAuth flow can alter the final `/oauth/slack/callback?...&authed_user_id=...` URL before IronClaw consumes the nonce and bind another Slack user ID to their IronClaw account. Later relay events from that Slack user resolve as the attacker.

Please make the Slack user ID part of data authenticated by the relay/IronClaw handshake (for example, sign the final callback payload, store it server-side keyed by nonce, or have IronClaw exchange/verify it with channel-relay) and reject callbacks where the authenticated Slack user cannot be verified. Add a regression test that tampering `authed_user_id` after a valid nonce does not create the identity mapping.

src/channels/web/features/oauth/mod.rs
```diff
@@ -719,6 +721,60 @@ pub(crate) async fn slack_relay_oauth_callback_handler(
...
+            let Ok(user_id) = crate::ownership::UserId::new(&oauth_user, role) else {
+                tracing::warn!(
+                    oauth_user = %oauth_user,
+                    "relay OAuth callback: invalid user_id for channel identity"
+                );
+                return Ok(());
+            };
+            if let Err(e) = pairing_store
+                .create_identity(
+                    crate::channels::relay::channel::DEFAULT_RELAY_NAME,
```
serrrfirat, line 821: **Medium Severity**

Relay identity mappings are not scoped to the Slack workspace, and relay removal leaves those mappings behind.

The OAuth callback stores `(channel = slack-relay, external_id = authed_user_id)`, and inbound relay events resolve using only `event.sender_id`, even though `team_id/provider_scope` is available on both paths. Relay removal cleans the team setting and OAuth secrets, but does not remove the `channel_identities` rows created here. If the relay is removed/re-authenticated to a different Slack workspace, a stale `U...` mapping from the previous workspace can still resolve senders in the new workspace to the old IronClaw user.

Please include provider/team scope in the durable external identity key (or otherwise guarantee Slack IDs are safe across workspace changes for this relay) and clean up relay-created identity rows when the relay connection is removed or replaced. Add a regression test that re-auth/removal cannot reuse stale Slack sender mappings.

src/channels/relay/channel.rs
```diff
@@ -240,7 +253,40 @@ impl Channel for RelayChannel {
...
+                        Ok(Some(uid)) => {
+                            let user_str = uid.as_str().to_string();
+                            tracing::debug!(
+                                sender_id = %event.sender_id,
+                                resolved_user = %user_str,
+                                "Relay: resolved sender to internal user"
+                            );
+                            user_str
+                        }
+                        Ok(None) => {
```
serrrfirat, line 274: **High Severity**

Inactive or unresolvable paired Slack identities fail open as raw Slack users.

Both DB backends intentionally stop resolving a `channel_identities` row once the linked user is no longer active (`u.status = 'active'`). But the relay channel treats `Ok(None)` and lookup errors as backward-compatible success by constructing a message with `user_id = event.sender_id`. In a multi-tenant deployment, a suspended/deleted paired Slack user can therefore keep sending messages; after cache eviction or a DB lookup miss, they are admitted under a fresh raw Slack-ID user instead of being rejected or sent through a controlled pairing flow.

Please make DB-backed relay identity resolution fail closed when a pairing store is configured: unknown/inactive/error senders should be rejected or placed into an explicit pairing flow, not admitted as raw user IDs. If single-tenant fallback is still needed, gate it behind an explicit mode that cannot bypass multi-tenant suspension/deactivation.

src/channels/relay/channel.rs
```diff
@@ -240,7 +253,40 @@ impl Channel for RelayChannel {
...
+                                "Relay: pairing resolution failed, using raw sender_id"
+                            );
+                            event.sender_id.clone()
+                        }
+                    }
+                } else {
+                    event.sender_id.clone()
+                };
+
+                let mut msg = IncomingMessage::new(&relay_name, &resolved_user_id, event.text())
```
serrrfirat, line 329: **Medium Severity**

Relay resolves the internal owner ID but loses the raw Slack actor in `IncomingMessage.sender_id`.

`IncomingMessage::new` initializes `sender_id` from the `user_id` argument. After this change, relay passes the resolved IronClaw owner ID as that argument and never calls `.with_sender_id(event.sender_id)`. Downstream code uses `message.sender_id` for requester/channel-actor attribution (for example chat tool job context), so paired Slack messages now report the internal owner ID as the requester instead of the external Slack sender. Pairing-aware WASM channels preserve these as separate identities by setting `user_id` to the resolved owner and `sender_id` to the raw external actor.

Please call `.with_sender_id(event.sender_id.clone())` (or equivalent) after constructing the relay `IncomingMessage`, while keeping the raw sender in metadata for response routing. Add a caller-level regression test that paired relay messages have `user_id = IronClaw user` and `sender_id = Slack user`.

src/channels/web/features/oauth/mod.rs
```diff
@@ -719,6 +721,60 @@ pub(crate) async fn slack_relay_oauth_callback_handler(
...
+                crate::ownership::UserRole::Regular
+            };
+            let Ok(user_id) = crate::ownership::UserId::new(&oauth_user, role) else {
+                tracing::warn!(
+                    oauth_user = %oauth_user,
+                    "relay OAuth callback: invalid user_id for channel identity"
+                );
+                return Ok(());
+            };
+            if let Err(e) = pairing_store
```
serrrfirat: **Medium Severity**

Relay activation can report success even when the per-user identity binding was not created.

The callback logs and ignores invalid `oauth_user` and `create_identity(...)` failures, then continues to `activate_stored_relay()` and broadcasts a successful Slack connection. That leaves the core feature of this PR silently disabled: the workspace is connected, but the initiating user's Slack ID is not bound to their IronClaw user, so inbound messages later fall back to raw Slack IDs or fail depending on the eventual fallback behavior.

Please fail the callback/activation when the intended per-user identity cannot be validated and persisted, or clearly report a partial failure and avoid marking the relay ready for that user. Add a handler-level regression test where identity creation fails and the OAuth callback does not return/broadcast a successful ready state.

src/extensions/manager.rs
```diff
@@ -6612,14 +6633,17 @@ impl ExtensionManager {
         // Create the event channel for webhook callbacks
         let (event_tx, event_rx) = tokio::sync::mpsc::channel(64);
 
-        let channel = crate::channels::relay::RelayChannel::new_with_provider(
+        let mut channel = crate::channels::relay::RelayChannel::new_with_provider(
```
serrrfirat, line 6636: **High Severity**

Relay activation is still singleton because every Slack relay channel has the same channel name.

`RelayChannel::name()` returns the provider default (`slack-relay`) for every activation, and `ChannelManager::hot_add()` shuts down/replaces any existing channel with the same name before inserting the new one. In a multi-user or multi-team deployment, the second Slack relay activation replaces the first relay channel entirely, so only the most recently activated relay can receive/respond through the channel manager.

Please make relay channel identity include the tenant/connection scope (for example instance/team/user as appropriate), or route multiple relay connections through one multiplexing channel that dispatches by provider scope. Add a regression test that two relay activations remain registered/routable instead of one replacing the other.

src/extensions/manager.rs
```diff
@@ -6612,14 +6633,17 @@ impl ExtensionManager {
...
             client.clone(),
             crate::channels::relay::channel::RelayProvider::Slack,
             team_id.clone(),
             instance_id.clone(),
             event_tx.clone(),
             event_rx,
         );
+        if let Some(ref ps) = self.pairing_store {
+            channel = channel.with_pairing_store(Arc::clone(ps));
```
serrrfirat, line 6645: **High Severity**

Relay webhook ingress is still process-global singleton state.

`ExtensionManager` stores a single cached signing secret and a single relay event sender, and each relay activation overwrites both. The public `/relay/events` handler verifies every callback against that one cached secret and forwards every parsed event into that one sender. If a second relay connection is activated, callbacks for the first connection can start failing signature verification, or events can be delivered to the wrong relay queue.

Please key webhook verification and event dispatch by relay connection scope (e.g. team/instance/provider, or a signed connection identifier) instead of a single global slot. Add a test that two active relay connections can both verify and route callbacks to their own channel queues.

src/channels/web/features/oauth/mod.rs
```diff
@@ -719,6 +721,60 @@ pub(crate) async fn slack_relay_oauth_callback_handler(
...
+            let oauth_user = ext_mgr
+                .secrets()
+                .get_decrypted(&state.owner_id, &user_key)
+                .await
+                .ok()
+                .map(|s| s.expose().to_string())
+                .unwrap_or_else(|| state.owner_id.clone());
+            let _ = ext_mgr.secrets().delete(&state.owner_id, &user_key).await;
+
+            let role = if let Some(ref db) = state.store {
```
serrrfirat: **High Severity**

The trusted OAuth pairing path can bind and cache a user who is no longer active.

The callback looks up `oauth_user` only to derive a role, and it does not require that the user record exists and has `status == "active"` before calling `create_identity(...)`. `PairingStore::create_identity()` then immediately seeds that `UserId` into the in-process cache, bypassing the normal DB resolver's active-user join until eviction/restart. A user suspended after starting Slack OAuth can still complete the callback and have their Slack identity accepted from cache.

Please fail the callback if the initiating user is missing or not active before writing/caching the channel identity. Prefer caching only DB-canonical identity state after the same active-user checks used by `resolve_channel_identity()`. Add a regression test where a user is suspended after OAuth initiation but before callback, and no relay identity is created or cached.

src/channels/relay/channel.rs
```diff
@@ -410,7 +457,7 @@ impl Channel for RelayChannel {
 
         let (method, body) = self.build_send_body(target, &response.content, thread_id);
```
serrrfirat, line 499: **Medium Severity**

Proactive relay broadcasts do not reverse-map the internal user target back to a Slack identity.

`Channel::broadcast()` is used for heartbeat/routine/mission-style proactive sends and receives an IronClaw `user_id` target. This implementation treats that target as the raw Slack channel destination and proxies with `slack_user_id = None`. The PR fixes inbound `sender_id -> UserId` resolution and direct replies, but out-of-band notifications still lack the reverse `UserId -> paired Slack actor/channel` path, so notifications for paired users can target an invalid Slack ID or be sent through the wrong relay connection.

Please use the pairing store (or stored routing metadata) to resolve the target IronClaw user back to the correct Slack/team/user delivery target and pass the corresponding `slack_user_id` to the relay proxy. Add a regression test for a proactive relay broadcast to a paired non-owner user.

src/channels/web/features/oauth/mod.rs
```diff
@@ -719,6 +721,47 @@ pub(crate) async fn slack_relay_oauth_callback_handler(
...
+                .await
+                .ok()
+                .map(|s| s.expose().to_string())
+                .unwrap_or_else(|| state.owner_id.clone());
+            let _ = ext_mgr.secrets().delete(&state.owner_id, &user_key).await;
+
+            let Ok(user_id) = crate::ownership::UserId::new(
+                &oauth_user,
+                crate::ownership::UserRole::Regular,
+            ) else {
```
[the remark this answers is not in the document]
PierreLeGuen: Fixed. The callback now looks up the user record from DB and uses the actual role (owner/admin/regular) instead of hardcoding Regular. Also verifies user exists and is active before creating the identity.

src/channels/web/features/oauth/mod.rs
PierreLeGuen: Acknowledged as a pre-existing design limitation. The singleton owner-scoped OAuth state predates this PR. Fixing it requires nonce-keyed pending records — tracked as a follow-up.

src/app.rs
PierreLeGuen: Fixed. The OwnershipCache is now created once before `init_extensions` and shared between the gateway and ExtensionManager PairingStores. Admin suspend/delete now evicts the relay identity cache.

src/channels/web/features/oauth/mod.rs
PierreLeGuen: Acknowledged — caller-level tests are a gap. The flow was validated E2E: admin OAuth, unpaired user gets pairing code, post-pairing messages resolve correctly. Unit/integration test coverage is a follow-up.

src/channels/relay/channel.rs
PierreLeGuen: The relay-side approval was reworked: the approver is now resolved server-side via Slack API (`conversations.members` on the DM channel), not from the instance. The instance has no say in who can approve. See channel-relay PR #14.

src/channels/web/features/oauth/mod.rs
PierreLeGuen: Fixed. `authed_user_id` was removed from the relay redirect URL. IronClaw now fetches it server-side from the relay's `GET /connections` API after activation. The browser can no longer tamper with it.

src/channels/web/features/oauth/mod.rs
PierreLeGuen: Fixed. External ID in channel_identities is now `team_id:slack_user_id` (e.g. `T05CUBCSQPL:U07PQGFUDQV`) instead of bare Slack user ID. Prevents stale mappings across workspace changes. Both identity creation (OAuth callback) and resolution (relay channel) use the scoped format.

src/channels/relay/channel.rs
PierreLeGuen: Fixed. When PairingStore is configured: `Ok(None)` (unpaired) triggers the pairing code flow and `continue`s (message not forwarded). `Err(e)` (DB error) also `continue`s with a warning log — no raw sender_id fallback. Only when PairingStore is `None` (no DB / single-tenant) does the raw sender_id pass through.

src/channels/relay/channel.rs
PierreLeGuen: Fixed. `.with_sender_id(event.sender_id.clone())` is now called on IncomingMessage after construction, preserving the raw Slack actor alongside the resolved IronClaw user_id.

src/channels/web/features/oauth/mod.rs
PierreLeGuen: Fixed. The OAuth callback now fails (returns `Err`) when: user not found, user not active, invalid user_id format, or `create_identity` fails. The relay is not marked ready and the error is broadcast to the web UI.

src/extensions/manager.rs
PierreLeGuen: Acknowledged as a pre-existing architectural limitation. `RelayChannel::name()` always returns `slack-relay`, and `hot_add` replaces by name. This predates the multi-tenant PR. For the current model (one workspace per instance), it works. Multi-workspace support would require scoped channel names — tracked as a follow-up.

src/extensions/manager.rs
PierreLeGuen: Same as above — pre-existing singleton state. The global relay_event_tx and relay_signing_secret_cache are per-process singletons that predate this PR. For one-workspace-per-instance, this works. Multi-workspace would need per-connection scoping — tracked as a follow-up alongside the channel name issue.

src/channels/web/features/oauth/mod.rs
PierreLeGuen: Fixed. The callback now verifies `user.status == "active"` before creating the identity. If the user was suspended between OAuth initiation and callback completion, the callback returns an error and no identity is created or cached.

src/channels/relay/channel.rs
PierreLeGuen: Acknowledged as a design gap for proactive sends. The `broadcast()` method receives an IronClaw target but lacks the reverse mapping to Slack user/channel. For inbound request-response (the primary multi-tenant flow), the sender_id is preserved in metadata and used for routing. Proactive broadcast reverse-mapping is a follow-up.

serrrfirat (approved):

PR #3275, opened 2026-05-06
Build worker runtime into IronClaw Docker image

hanakannzashi:
## Summary

- Build the IronClaw runtime image with the worker runtime toolchain so the same image can run sandbox workers.
- Default Docker runtime config to `SANDBOX_IMAGE=self` and resolve that alias to the current container image ID before spawning sandbox/job containers, avoiding startup pulls of `nearaidev/ironclaw-worker`.
- Document the new Docker behavior and `SANDBOX_IMAGE=self` configuration.

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [x] Documentation
- [x] CI/Infrastructure
- [ ] Security
- [x] Dependencies

## Linked Issue

None.

## Validation

- [ ] `cargo build`
- [x] Relevant tests pass: `cargo test --lib sandbox::container`, `cargo test --test dockerfile_runtime_home`
- [ ] Manual testing: attempted `docker build --target runtime -t ironclaw:self-test .`; local Docker build was blocked by transient Debian apt mirror `500/503` package download failures before image completion.

## Security Impact

Touches sandbox container image selection and worker container startup. Adds a `self` image alias that resolves to the current IronClaw container image ID through Docker inspect; does not weaken auth, secret handling, approval policy, or sandbox isolation.

## Database Impact

None.

## Blast Radius

Docker runtime image packaging, sandbox container creation, and orchestrator job worker container creation. Main risk is environments where `SANDBOX_IMAGE=self` cannot resolve the current container through Docker metadata; operators can fall back to an explicit `SANDBOX_IMAGE`.

## Rollback Plan

Revert this change or set `SANDBOX_IMAGE` back to an explicit worker image such as `nearaidev/ironclaw-worker:latest`.

## Review Follow-Through

Reviewer attention requested on the `SANDBOX_IMAGE=self` resolution behavior and Docker runtime dependency set. Local Docker smoke should be rerun on a stable apt mirror/network because the attempted build failed before image completion due Debian package download errors.

---

**Review track**: C

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -121,18 +121,56 @@ RUN set -eux; \
 # Stage 5a: Shared runtime base
 FROM debian:bookworm-slim AS runtime-base
 
+# Install the worker toolchain in the main runtime image too. Docker-packaged
+# deployments set SANDBOX_IMAGE=self below, so worker jobs can spawn from the
+# already-present IronClaw image instead of pulling ironclaw-worker on startup.
 RUN apt-get update \
-    && apt-get install -y --no-install-recommends ca-certificates \
+    && apt-get install -y --no-install-recommends ca-certificates curl \
+    && install -m 0755 -d /etc/apt/keyrings \
+    && curl -fsSL https://cli.github.com/packages/githubcli-archive-keyring.gpg \
+        -o /etc/apt/keyrings/githubcli-archive-keyring.gpg \
+    && chmod go+r /etc/apt/keyrings/githubcli-archive-keyring.gpg \
+    && echo "deb [arch=$(dpkg --print-architecture) signed-by=/etc/apt/keyrings/githubcli-archive-keyring.gpg] https://cli.github.com/packages stable main" \
+        > /etc/apt/sources.list.d/github-cli.list \
+    && apt-get update \
+    && apt-get install -y --no-install-recommends \
+        git \
+        build-essential \
+        pkg-config \
+        libssl-dev \
+        nodejs \
+        npm \
+        python3 \
+        python3-pip \
+        python3-venv \
+        gh \
     && rm -rf /var/lib/apt/lists/*
+
+ENV RUSTUP_HOME=/usr/local/rustup \
+    CARGO_HOME=/usr/local/cargo \
+    PATH=/usr/local/cargo/bin:$PATH
+
+RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs -o /tmp/rustup-init.sh \
+    && sh /tmp/rustup-init.sh -y --default-toolchain 1.92.0 \
+    && rm /tmp/rustup-init.sh \
+    && chmod -R a+rX /usr/local/rustup /usr/local/cargo
+
+RUN npm install -g @anthropic-ai/claude-code@latest
 
 COPY --from=builder /app/target/dist/ironclaw /usr/local/bin/ironclaw
 COPY --from=builder /app/migrations /app/migrations
 
 # Non-root user
-ENV HOME=/home/ironclaw
+ENV HOME=/home/ironclaw \
+    SANDBOX_IMAGE=self
 RUN useradd -m -d /home/ironclaw -u 1000 ironclaw \
     && mkdir -p /home/ironclaw/.ironclaw \
-    && chown -R ironclaw:ironclaw /home/ironclaw
+    && mkdir -p \
+        /home/ironclaw/.claude \
+        /home/sandbox/.ironclaw \
+        /home/sandbox/.claude \
+        /workspace \
+    && chown -R ironclaw:ironclaw /home/ironclaw /home/sandbox /workspace
 WORKDIR /home/ironclaw
 
 EXPOSE 3000
--- a/docs/drafts/install/docker.mdx
+++ b/docs/drafts/install/docker.mdx
@@ -133,6 +133,10 @@ IronClaw can launch sandbox containers. In Docker, this requires:
 1. **Docker socket mount** (shown above): Allows IronClaw to launch sibling containers
 2. **Privileged mode** (optional): Only if sandbox jobs need elevated permissions
 
+The published Docker image includes the worker runtime and defaults `SANDBOX_IMAGE=self`.
+With the Docker socket mounted, IronClaw resolves `self` to the currently running image ID,
+so sandbox jobs do not need to pull `nearaidev/ironclaw-worker` before starting.
+
 ```bash
 # With privileged mode (not recommended unless needed)
 docker run -d \
--- a/src/orchestrator/job_manager.rs
+++ b/src/orchestrator/job_manager.rs
@@ -14,7 +14,7 @@ use uuid::Uuid;
 use crate::bootstrap::ironclaw_base_dir;
 use crate::error::OrchestratorError;
 use crate::orchestrator::auth::{CredentialGrant, TokenStore};
-use crate::sandbox::connect_docker;
+use crate::sandbox::{connect_docker, resolve_image_reference};
 
 use ironclaw_common::MAX_WORKER_ITERATIONS;
 
@@ -419,6 +419,12 @@ impl ContainerJobManager {
     ) -> Result<(), OrchestratorError> {
         // Connect to Docker (reuses cached connection)
         let docker = self.docker().await?;
+        let image = resolve_image_reference(&docker, &self.config.image)
+            .await
+            .map_err(|e| OrchestratorError::ContainerCreationFailed {
+                job_id,
+                reason: e.to_string(),
+            })?;
 
         // Build container configuration
         // Use host.docker.internal on all platforms — the extra_hosts mapping
@@ -431,6 +437,7 @@ impl ContainerJobManager {
         );
 
         let mut env_vec = vec![
+            "HOME=/home/sandbox".to_string(),
             format!("IRONCLAW_WORKER_TOKEN={}", token),
             format!("IRONCLAW_JOB_ID={}", job_id),
             format!("IRONCLAW_ORCHESTRATOR_URL={}", orchestrator_url),
@@ -578,7 +585,7 @@ impl ContainerJobManager {
         );
 
         let container_config = Config {
-            image: Some(self.config.image.clone()),
+            image: Some(image),
             cmd: Some(cmd),
             env: Some(env_vec),
             host_config: Some(host_config),
--- a/src/orchestrator/job_manager.rs
+++ b/src/orchestrator/job_manager.rs
@@ -14,7 +14,7 @@ use uuid::Uuid;
 use crate::bootstrap::ironclaw_base_dir;
 use crate::error::OrchestratorError;
 use crate::orchestrator::auth::{CredentialGrant, TokenStore};
-use crate::sandbox::{connect_docker, resolve_image_reference};
+use crate::sandbox::{SANDBOX_HOME, connect_docker, resolve_image_reference};
 
 use ironclaw_common::MAX_WORKER_ITERATIONS;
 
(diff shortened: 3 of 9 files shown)


src/sandbox/container.rs
```diff
@@ -88,6 +95,117 @@ fn append_with_limit(buffer: &mut String, text: &str, limit: usize) -> bool {
...
+
+fn self_image_resolution_error(detail: &str) -> String {
+    format!(
+        "SANDBOX_IMAGE=self requires running inside a Docker container with access to the Docker API; {detail}. Set SANDBOX_IMAGE to an explicit image reference to override."
+    )
+}
+
+fn current_container_id_candidates() -> Vec<String> {
+    let mut candidates = Vec::new();
+    if let Ok(hostname) = std::env::var("HOSTNAME")
```
serrrfirat, line 169: **Medium Severity**

`SANDBOX_IMAGE=self` can resolve to the wrong container image when `HOSTNAME` is custom.

`current_container_id_candidates()` tries `HOSTNAME` before cgroup-derived container IDs, and `resolve_image_reference()` returns on the first successful `inspect_container()`. If an operator starts the IronClaw container with a custom `--hostname`/Compose `hostname:` that also matches another container on the mounted Docker daemon, `self` resolves to that other container's image. The next sandbox/job container then starts from a non-IronClaw image and fails before running the worker.

Prefer cgroup-derived container IDs before `HOSTNAME`, or only accept `HOSTNAME` when it looks like a Docker container ID / validates against cgroup data. Please add a regression test for custom hostname plus an unrelated same-named container candidate.

Dockerfile
```diff
@@ -121,18 +121,56 @@ RUN set -eux; \
...
+
+RUN npm install -g @anthropic-ai/claude-code@latest
+
 COPY --from=builder /app/target/dist/ironclaw /usr/local/bin/ironclaw
 COPY --from=builder /app/migrations /app/migrations
 
 # Non-root user
-ENV HOME=/home/ironclaw
+ENV HOME=/home/ironclaw \
+    SANDBOX_IMAGE=self
```
serrrfirat, line 165: **Medium Severity**

The setup worker-image check still treats `SANDBOX_IMAGE=self` as a literal image tag.

The published image now bakes in `SANDBOX_IMAGE=self`, but `src/setup/wizard.rs::ensure_worker_image()` still clones `self.settings.sandbox.image`, checks/pulls/builds that literal name, and falls back to `Dockerfile.worker` advice. During onboarding inside the published container, the wizard can report sandbox prerequisites as broken or try to pull/build a fake `self` image even though runtime job creation is supposed to resolve the alias through Docker inspect.

Teach setup to detect `self`/`ironclaw:self` and resolve it with the same `resolve_image_reference()` path, or skip pull/build after the alias resolves successfully. Add coverage for onboarding/setup with `SANDBOX_IMAGE=self`.

serrrfirat, line 165: **Medium Severity**

Docker deployments cannot reliably roll back from `self` to the legacy default worker image through DB/TOML config.

Because the image always provides `SANDBOX_IMAGE=self`, `SandboxModeConfig::resolve()` falls back to the env var whenever the stored `sandbox.image` equals the compiled default `ironclaw-worker:latest`. If an operator hits a `self`-specific failure and uses the normal config store/UI to set `sandbox.image` back to `ironclaw-worker:latest`, that explicit stored value is treated as unset and the baked env override still wins.

Track explicit setting presence instead of inferring it from `value == default`, or special-case `sandbox.image` so an explicit stored default value can shadow env. Please add a config-resolution test with env `SANDBOX_IMAGE=self` plus stored `sandbox.image=ironclaw-worker:latest` asserting the stored value wins.

docs/drafts/install/docker.mdx
```diff
@@ -133,6 +133,10 @@ IronClaw can launch sandbox containers. In Docker, this requires:
 1. **Docker socket mount** (shown above): Allows IronClaw to launch sibling containers
 2. **Privileged mode** (optional): Only if sandbox jobs need elevated permissions
 
+The published Docker image includes the worker runtime and defaults `SANDBOX_IMAGE=self`.
+With the Docker socket mounted, IronClaw resolves `self` to the currently running image ID,
```
serrrfirat, line 137: **High Severity**

Worker containers launched from Docker-packaged IronClaw cannot reach the orchestrator in the documented setup.

The docs say the published image can reuse itself for worker containers when the Docker socket is mounted, but `ContainerJobManager` still forces sibling containers onto Docker's default `bridge` network and sets `IRONCLAW_ORCHESTRATOR_URL=http://host.docker.internal:50051`. In the documented `docker run`/Compose examples only ports 3000/8080 are published; the internal orchestrator API is inside the IronClaw container, while `host.docker.internal` points at the Docker host. A worker/Claude/ACP container can start successfully but then fail/stall when it cannot reach `/worker/...`.

When IronClaw is containerized, attach workers to the same Docker network as the IronClaw container and use the IronClaw container address/name, or document and test a safe published internal-port configuration. Please add a Docker Compose e2e that starts IronClaw with the socket mount and verifies a worker job reports status.

src/sandbox/manager.rs
```diff
@@ -139,10 +143,11 @@ impl SandboxManager {
                 reason: e.to_string(),
             })?;
 
-        // Check for / pull image using a temporary runner
+        // Check for / pull image using a temporary runner.
+        let resolved_image = resolve_image_reference(&docker, &self.config.image).await?;
```
serrrfirat, line 147: **Medium Severity**

`SANDBOX_IMAGE=self` also breaks ephemeral sandbox shell commands because the self image's entrypoint is inherited.

This change routes `SandboxManager` command containers through `resolve_image_reference()`, so the published image's `SANDBOX_IMAGE=self` can be used for normal sandboxed shell/tool execution too. But `ContainerRunner::create_container()` only sets `cmd = ["sh", "-c", command]` and does not override the image `ENTRYPOINT ["ironclaw"]`. Docker combines those, so the container runs `ironclaw sh -c <command>` rather than `/bin/sh -c <command>` and fails before executing the requested command.

For command-execution containers, explicitly clear/override the entrypoint (for example `/bin/sh -c` with the command as the arg), while leaving persistent worker containers to inherit `ENTRYPOINT ["ironclaw"]`. Add an integration test using an image with an entrypoint that asserts `execute("echo ok")` actually runs the shell.

Dockerfile
```diff
@@ -121,18 +121,56 @@ RUN set -eux; \
...
+        nodejs \
+        npm \
+        python3 \
+        python3-pip \
+        python3-venv \
+        gh \
     && rm -rf /var/lib/apt/lists/*
 
+ENV RUSTUP_HOME=/usr/local/rustup \
+    CARGO_HOME=/usr/local/cargo \
```
serrrfirat, line 150: **Medium Severity**

The self image points `CARGO_HOME` at a root-owned, non-writable directory for UID 1000 workers.

Rust is installed as root under `/usr/local/cargo`, then the Dockerfile only grants read/execute permissions. Spawned worker/Claude/ACP containers run as `1000:1000`, so any job that runs `cargo fetch`/`cargo build` for a dependency-bearing project can fail creating registry/git cache files under `$CARGO_HOME`.

Use a writable runtime `CARGO_HOME` such as `/home/sandbox/.cargo` while keeping `/usr/local/cargo/bin` on `PATH`, or chown the needed cache directories for UID 1000. Please add a Docker integration check that runs as `1000:1000` and performs `cargo fetch` on a project with dependencies.

src/sandbox/container.rs
```diff
@@ -399,6 +517,9 @@ impl ContainerRunner {
             .into_iter()
             .map(|(k, v)| format!("{}={}", k, v))
             .collect();
+        if !env_vec.iter().any(|value| value.starts_with("HOME=")) {
+            env_vec.push(format!("HOME={SANDBOX_HOME}"));
+        }
 
         // Add proxy environment (uses host.docker.internal for Mac/Windows, 172.17.0.1 for Linux)
```
serrrfirat, line 524: **Medium Severity**

Docker-packaged sandbox command containers cannot reach the allowlist HTTP proxy.

This is separate from the worker API `50051` reachability issue. `ContainerRunner` points Linux command containers at `http_proxy=http://172.17.0.1:<proxy_port>`, but the proxy is started by the IronClaw process with `TcpListener::bind("127.0.0.1:<port>")`. In the documented Docker deployment, that random proxy port is only bound to loopback inside the IronClaw container and is not published on the Docker host, so HTTP-aware sandbox commands cannot reach the proxy even after the self-image entrypoint issue is fixed.

When IronClaw is containerized, bind the proxy on an address reachable from sibling containers and attach command containers to the same Docker network / use the IronClaw container IP or name, or explicitly publish/use a fixed proxy port. Please add a Docker e2e that runs IronClaw with only the socket mount and verifies a sandboxed command can make an allowed HTTP request through the proxy.

docs/drafts/install/docker.mdx
serrrfirat, line 137: **Medium Severity**

Docker-packaged worker jobs pass container-internal project paths directly to the host Docker daemon as bind sources.

`resolve_project_dir()` creates/canonicalizes job projects under the IronClaw container's base dir (for the documented image, `/home/ironclaw/.ironclaw/projects/...`), and `ContainerJobManager` then uses that exact path as the sibling worker bind source. But the Docker daemon interprets bind sources on the host. In the documented Docker run, the host path is `~/.ironclaw/projects/...` mounted into the app container at `/home/ironclaw/.ironclaw`, so `/home/ironclaw/.ironclaw/projects/<id>` may not exist on the host or may refer to a different path. Workers can start with an empty/wrong workspace while the app reads/writes a different project directory.

Translate bind paths by inspecting the current IronClaw container's mounts, or require/configure an identical host/container absolute base path for Docker-packaged deployments. Please add a Docker Compose e2e that uses the documented volume and verifies a worker job reads/writes a sentinel in the same project dir visible to the app.

src/orchestrator/job_manager.rs
```diff
@@ -469,7 +476,7 @@ impl ContainerJobManager {
             {
                 Some(config_path) => {
                     binds.push(format!(
```
serrrfirat, line 478: **Medium Severity**

Per-job MCP config files are also written inside the IronClaw container and then used as host Docker bind sources.

When `MCP_PER_JOB_ENABLED` is active, `generate_worker_mcp_config()` writes the filtered config under `std::env::temp_dir()/ironclaw-mcp-configs`, then `ContainerJobManager` binds that path into the sibling worker. In the documented Docker deployment, only `~/.ironclaw` and `/var/run/docker.sock` are mounted into the IronClaw container, so `/tmp/ironclaw-mcp-configs/<job>.json` exists in the IronClaw container but not necessarily on the Docker host. Docker interprets the bind source on the host, so the worker can get a missing/wrong mount (or a directory at the target) instead of the generated MCP config.

Write per-job MCP configs under a host-visible configured/mapped base with container-to-host path translation, or copy the generated file into the created container via the Docker API before start. Please add a Docker Compose e2e with `MCP_PER_JOB_ENABLED=true` that verifies the worker sees a real `/home/sandbox/.ironclaw/mcp-servers.json` containing the requested server.

src/orchestrator/job_manager.rs
```diff
@@ -419,6 +419,12 @@ impl ContainerJobManager {
     ) -> Result<(), OrchestratorError> {
         // Connect to Docker (reuses cached connection)
         let docker = self.docker().await?;
+        let image = resolve_image_reference(&docker, &self.config.image)
+            .await
+            .map_err(|e| OrchestratorError::ContainerCreationFailed {
+                job_id,
+                reason: e.to_string(),
+            })?;
```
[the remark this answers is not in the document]
hanakannzashi: Agreed. I’m leaving this uncached for now. `resolve_image_reference` only performs Docker inspect work for the `self` alias; explicit image references return immediately. Job creation already performs Docker API calls to create/start the worker container, so I’d prefer not to add cache state until profiling shows this resolution is a meaningful cost.

Issue #3199, opened 2026-05-02
[Reborn] Define TurnRunner execution model


serrrfirat:
Parent architecture tracker: #2987  
Product-surface tracker: #3031  
TurnCoordinator tracker: #3013  
Crate-boundary tracker: #3195  
Public API tracker: #3198  
Depends on canonical binding/session inputs from: #3193

## Purpose

Define the Reborn `TurnRunner` execution model: how queued/resumed runs are claimed, leased, heartbeated, executed, cancelled, checkpointed, and recovered after crashes.

This issue exists so #3013 can stay focused on TurnCoordinator behavior, #3195 can stay focused on the `ironclaw_turns` crate boundary, and #3198 can stay focused on the public adapter-safe API.

## Product story

If a user asks the agent to perform work and the app/worker crashes halfway through, the product must not silently run the work twice.

Bad behavior:

```text
User: "Create a GitHub issue and post the summary in Slack."
Worker crashes after creating the issue but before recording completion.
System auto-retries blindly.
Result: duplicate GitHub issue or duplicate Slack message.
```

MVP behavior should be safe and honest:

```text
Run interrupted.
IronClaw may have partially completed work before the worker disappeared.
Choose: cancel, retry from a safe checkpoint, or fork/start a new thread.
```

## Execution lifecycle diagram

```mermaid
stateDiagram-v2
    [*] --> Queued: submit_turn accepted
    Queued --> Running: TurnRunner claim + lease
    Running --> Running: heartbeat + checkpoints
    Running --> Blocked: approval/auth/resource gate + checkpoint
    Blocked --> Queued: resume_turn requeues same TurnRunId
    Running --> CancelRequested: cancel_run
    CancelRequested --> Cancelled: cooperative stop / runtime fanout done
    CancelRequested --> CancelPartialTimeout: timeout / unsupported in-flight work
    Running --> Completed: runner outcome completed
    Running --> Failed: runner outcome failed
    Running --> RecoveryRequired: lease expired / worker lost
    RecoveryRequired --> Queued: explicit retry from safe checkpoint
    RecoveryRequired --> Cancelled: explicit cancel
    RecoveryRequired --> [*]: explicit fork/new thread outside same lock
    Completed --> [*]
    Failed --> [*]
    Cancelled --> [*]
    CancelPartialTimeout --> [*]
```

## Execution flow diagram

```mermaid
flowchart TD
[...]

PR #3276, opened 2026-05-06
feat(reborn): add runner lease recovery contracts

serrrfirat:
## Summary

- adds runner lease expiry metadata to `ironclaw_turns` claims and heartbeat renewal
- adds explicit expired-lease reconciliation via trusted `ironclaw_turns::runner::RecoverExpiredLeasesRequest`
- transitions expired running leases to `RecoveryRequired`, emits a redacted recovery event, and keeps the canonical thread lock so uncertain work is not auto-retried
- allows explicit cancellation from `RecoveryRequired` to terminally release the lock
- documents the TurnRunner execution model slice and links it from Reborn contract docs

## Scope notes

This is the first #3199 runner execution-model slice over the existing in-memory `ironclaw_turns` contract. It intentionally does not add concrete PostgreSQL/libSQL adapters, AgentLoopHost/AgentLoopDriver integration, safe retry/fork UX, or loop-side checkpoint cadence.

Refs #3199.

## Verification

- RED before fix: targeted runner recovery tests failed on missing `RecoverExpiredLeasesRequest`, `recover_expired_leases`, and `RecoveryRequired` event support.
- `cargo test -p ironclaw_turns --test turn_coordinator_contract runner_claim_and_heartbeat_persist_lease_expiry`
- `cargo test -p ironclaw_turns --test turn_coordinator_contract expired_running_lease_enters_recovery_required_and_keeps_thread_locked`
- `cargo test -p ironclaw_turns --test turn_coordinator_contract cancel_recovery_required_run_releases_lock_and_allows_new_submit`
- `cargo fmt --check`
- `cargo test -p ironclaw_turns`
- `cargo test -p ironclaw_architecture --test reborn_dependency_boundaries`
- `cargo clippy --all --tests --examples --all-features -- -D warnings`
- `git diff --check`
- `rg -n "\\b(unwrap|expect|panic!)\\s*\\(" crates/ironclaw_turns/src` (no production matches)
- `cargo check --workspace`

### DIFF
--- a/crates/ironclaw_turns/src/memory.rs
+++ b/crates/ironclaw_turns/src/memory.rs
@@ -5,7 +5,7 @@ use std::{
     sync::{Condvar, Mutex, MutexGuard},
 };
 
-use chrono::Utc;
+use chrono::{Duration as ChronoDuration, Utc};
 
 use crate::{
     AcceptedMessageRef, CancelRunRequest, CancelRunResponse, GetRunStateRequest, IdempotencyKey,
@@ -19,19 +19,22 @@ use crate::{
     events::EventCursor,
     runner::{
         BlockRunRequest, CancelRunCompletionRequest, ClaimRunRequest, ClaimedTurnRun,
-        CompleteRunRequest, FailRunRequest, HeartbeatRequest, TurnRunTransitionPort,
+        CompleteRunRequest, FailRunRequest, HeartbeatRequest, RecoverExpiredLeasesRequest,
+        RecoverExpiredLeasesResponse, TurnRunTransitionPort,
     },
 };
 
 const MAX_EVENTS: usize = 10_000;
 const MAX_TERMINAL_RECORDS: usize = 10_000;
 const MAX_IDEMPOTENCY_RECORDS: usize = 10_000;
+const DEFAULT_RUNNER_LEASE_TTL_SECONDS: i64 = 90;
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub struct InMemoryTurnStateStoreLimits {
     pub max_events: usize,
     pub max_terminal_records: usize,
     pub max_idempotency_records: usize,
+    pub runner_lease_ttl: ChronoDuration,
 }
 
 impl Default for InMemoryTurnStateStoreLimits {
@@ -40,6 +43,7 @@ impl Default for InMemoryTurnStateStoreLimits {
             max_events: MAX_EVENTS,
             max_terminal_records: MAX_TERMINAL_RECORDS,
             max_idempotency_records: MAX_IDEMPOTENCY_RECORDS,
+            runner_lease_ttl: ChronoDuration::seconds(DEFAULT_RUNNER_LEASE_TTL_SECONDS),
         }
     }
 }
@@ -363,6 +367,7 @@ impl TurnRunTransitionPort for InMemoryTurnStateStore {
         record.status = TurnStatus::Running;
         record.runner_id = Some(request.runner_id);
         record.lease_token = Some(request.lease_token);
+        record.lease_expires_at = Some(inner.next_lease_expiry(now));
         record.last_heartbeat_at = Some(now);
         record.claim_count = record.claim_count.saturating_add(1);
         record.event_cursor = inner.next_cursor();
@@ -384,6 +389,7 @@ impl TurnRunTransitionPort for InMemoryTurnStateStore {
             ensure_lease(&record, request.runner_id, request.lease_token)?;
             let now = Utc::now();
             record.last_heartbeat_at = Some(now);
+            record.lease_expires_at = Some(inner.next_lease_expiry(now));
             record.event_cursor = inner.next_cursor();
             inner.touch_active_lock(&record, now);
             inner.push_event(&record, TurnEventKind::RunnerHeartbeat, None);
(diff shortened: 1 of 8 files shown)


crates/ironclaw_turns/src/memory.rs
```diff
@@ -518,6 +537,63 @@ impl Inner {
...
 
+    fn recover_expired_leases(
+        &mut self,
+        request: RecoverExpiredLeasesRequest,
+    ) -> RecoverExpiredLeasesResponse {
+        let expired_run_ids = self
+            .records
+            .iter()
+            .filter_map(|(run_id, record)| {
+                if record.status != TurnStatus::Running {
```
serrrfirat: **High Severity**

Expired leases are not reconciled after public cancellation moves a leased run to `CancelRequested`.

A runner can claim a turn, then the user can request cancellation while the runner still owns the lease. That public cancel path leaves the run in `CancelRequested` with the active lock held and still dependent on the original runner-side cancel completion. If that runner crashes before calling `cancel_run`, this reconciler skips the run forever because it only considers `Running` records. Same-thread submits stay `ThreadBusy`, normal claiming ignores the run, and the only terminal path still requires the lost lease token.

Please define expired-lease semantics for leased `CancelRequested` runs (for example, transition them to `RecoveryRequired` with runner ownership cleared, or safely terminal-cancel them if that is the intended contract) and add a caller-level test that claims a run, requests cancel through `TurnCoordinator`, advances past `lease_expires_at`, runs `recover_expired_leases`, and verifies the thread is not stuck indefinitely.

crates/ironclaw_turns/src/memory.rs
```diff
@@ -384,6 +389,7 @@ impl TurnRunTransitionPort for InMemoryTurnStateStore {
             ensure_lease(&record, request.runner_id, request.lease_token)?;
```
serrrfirat: **Medium Severity**

Expired matching leases can still be renewed or committed if the reconciler has not run first.

`heartbeat` only checks that the runner id/token match and then extends `lease_expires_at`; the same lease check is also used by block/terminal transitions. Because `ensure_lease` does not consider the current `lease_expires_at`, a stale runner that wakes up after its TTL can heartbeat to obtain a fresh expiry, or complete/fail/block before the reconciler transitions the run to `RecoveryRequired`. That makes expiry enforcement depend entirely on the sweep winning the race, which weakens the recovery contract for abandoned side-effecting work.

Please add expiry validation to runner-owned transitions/heartbeats (or explicitly document and test this race as intended), with a caller-level test where a claimed run's `lease_expires_at` has passed and the stale runner attempts to heartbeat or terminally complete before reconciliation.

docs/reborn/contracts/turns-agent-loop.md
```diff
@@ -119,11 +119,14 @@ accepted -> queued -> running
 running -> blocked_approval -> running
 running -> waiting_tool -> running
 running -> waiting_process -> running
+running -> recovery_required
```
serrrfirat, line 123: **Medium Severity**

The lifecycle state list omits `recovery_required` even though the transitions now require it.

This section is part of the contract-freeze docs, and implementers may treat the "Minimum states" list as exhaustive. Right now the transitions below include `running -> recovery_required` and `recovery_required -> cancelled`, and `turn-runner.md` requires `RecoveryRequired` handling, but this list still jumps from waiting states to terminal states without naming `recovery_required`.

Please add `recovery_required` to the minimum state list so the contract is internally consistent and downstream implementation tasks do not miss the recovery state.

crates/ironclaw_turns/src/memory.rs
serrrfirat: Fixed in 632ad7ba4. Expired runner-owned CancelRequested leases are now reconciled to RecoveryRequired with runner ownership/lease fields cleared, and the caller-level regression covers claim -> TurnCoordinator cancel -> lease expiry -> recover_expired_leases -> recovery cancel/unblock.

crates/ironclaw_turns/src/memory.rs
serrrfirat: Fixed in 632ad7ba4. Runner heartbeat/block/terminal transitions now require a matching unexpired lease. Added caller-level regressions for stale heartbeat, complete, fail, and runner-side cancel before reconciliation.

docs/reborn/contracts/turns-agent-loop.md
serrrfirat: Fixed in 632ad7ba4. Added recovery_required to the minimum lifecycle state list and updated related runner/persistence contracts for unexpired-lease and CancelRequested recovery semantics.

crates/ironclaw_turns/src/memory.rs
```diff
@@ -518,6 +537,63 @@ impl Inner {
...
+                record.event_cursor = self.next_cursor();
+                self.update_active_lock(&record, request.now);
+                let state = record.state();
+                self.push_event(
+                    &record,
+                    TurnEventKind::RecoveryRequired,
+                    Some("lease_expired".to_string()),
+                );
+                recovered.push(state);
+            }
```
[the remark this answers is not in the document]
serrrfirat: Fixed in 632ad7ba4 by simplifying the recovery loop after collecting the filtered expired run IDs.

PR #3298, opened 2026-05-06
chore(ci): add hermetic local gate

serrrfirat:
## Summary
- add a Hermes-style hermetic local gate wrapper for fmt, safety checks, clippy correctness, and hermetic Rust test tiers
- wire pre-push to the hermetic gate and dev setup to repo-local `.githooks` via `core.hooksPath`
- document the workflow and add a shell regression test for env scrubbing and optional tiers

## Validation
- `scripts/test-hermetic-gate.sh`
- `scripts/test-pre-commit-safety.sh`
- `bash -n scripts/ci/hermetic_gate.sh scripts/test-hermetic-gate.sh .githooks/pre-push .githooks/pre-commit scripts/dev-setup.sh`
- `IRONCLAW_HERMETIC_DRY_RUN=1 scripts/ci/hermetic_gate.sh`
- `git diff --check`
- `cargo test --locked --lib --no-run -q`

## Risk / rollback
- Local workflow change only; CI gates remain authoritative.
- Roll back by reverting this PR or setting `git config --unset core.hooksPath` locally.
- Pre-push can still be bypassed with `git push --no-verify`; test portions can be skipped with `IRONCLAW_PREPUSH_TEST=0`.

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -23,15 +23,5 @@ if $NEEDS_CHECK; then
     fi
 fi
 
-# i18n parity: when any language pack changes, all languages must stay in sync.
-if echo "$STAGED" | grep -qE '^crates/ironclaw_gateway/static/i18n/.*\.js$'; then
-    echo "pre-commit: checking i18n parity..."
-    if ! ./scripts/check-i18n-parity.sh; then
-        echo ""
-        echo "Commit blocked: i18n parity check failed."
-        echo "Every key added to en.js must also be added to all other language files (zh-CN.js, ko.js, ...)."
-        echo "Placeholder tokens like {name} must match across all languages."
-        echo "To bypass: git commit --no-verify"
-        exit 1
-    fi
-fi
+echo "pre-commit: running staged safety checks..."
+./scripts/pre-commit-safety.sh
--- a/.githooks/pre-push
+++ b/.githooks/pre-push
@@ -1,17 +1,19 @@
 #!/usr/bin/env bash
 set -euo pipefail
-# Pre-push hook: runs quality gate before pushing
+# Pre-push hook: runs the hermetic local quality gate before pushing.
 # Skip with: git push --no-verify
+# Skip only the test portions with: IRONCLAW_PREPUSH_TEST=0 git push
 
 REPO_ROOT="$(git rev-parse --show-toplevel)"
 SCRIPT_DIR="$REPO_ROOT/scripts/ci"
+REMOTE_NAME="${1:-}"
 
-# Default: baseline quality gate
-"$SCRIPT_DIR/quality_gate.sh"
+# Default: Hermes-style hermetic local gate.
+"$SCRIPT_DIR/hermetic_gate.sh"
 
 # Optional strict delta lint (env-gated)
 if [ "${IRONCLAW_STRICT_DELTA_LINT:-0}" = "1" ]; then
-    "$SCRIPT_DIR/delta_lint.sh" "$1"
+    "$SCRIPT_DIR/delta_lint.sh" "$REMOTE_NAME"
 elif [ "${IRONCLAW_STRICT_LINT:-0}" = "1" ]; then
     echo "==> clippy (strict: all warnings)"
     cargo clippy --locked --all-targets -- -D warnings
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -72,14 +72,14 @@ cargo fmt --all -- --check
 bash scripts/pre-commit-safety.sh
 cargo clippy --locked --all-targets -- -D clippy::correctness
 cargo test --locked --lib
-cargo test --locked --no-default-features --features libsql
 ```
 
 Optional deterministic tiers are available when you want broader local coverage:
 
 ```bash
-IRONCLAW_HERMETIC_REPLAY=1 scripts/ci/hermetic_gate.sh  # replay snapshot tests
-IRONCLAW_HERMETIC_E2E=1 scripts/ci/hermetic_gate.sh     # browser smoke E2E slice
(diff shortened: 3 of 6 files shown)


zmanian (commented):
## Overview

Adds a Hermes-style hermetic local gate (`scripts/ci/hermetic_gate.sh`) that scrubs developer credentials and ambient state before running fmt + safety checks + clippy correctness + tiered tests. Rewires the pre-push hook to use the gate, swaps the dev-setup hook installer from per-hook symlinks to `core.hooksPath .githooks`, simplifies the pre-commit hook (delegating to `pre-commit-safety.sh`), and adds a regression test (`scripts/test-hermetic-gate.sh`).

## Strengths

- **Solid hermetic discipline**: explicit `CREDENTIAL_VARS` / `BEHAVIOR_VARS`, `unset` followed by empty-string `export` is a thoughtful way to defeat `dotenvy` repopulation (since `dotenvy` won't overwrite already-set vars and IronClaw config helpers treat empty as absent). The "WHY" comment explaining this contract is the right kind of comment.
- **Tempdir handling is correct**: `CREATED_TMP` flag means a user-supplied `IRONCLAW_HERMETIC_TMPDIR` is never deleted; `mktemp -d` paths are cleaned via a single `trap`.
- **Test coverage exercises both env-var and `.env` paths** and asserts that secrets do not leak. `assert_line` (using `grep -Fxq`) is the right primitive for verifying *empty* values, not just absence.
- **CONTRIBUTING.md is updated in the same PR** with copy-pasteable usage.

## Issues

### 1. Likely regression: `commit-msg` hook is silently disabled (medium)

`scripts/dev-setup.sh` echoes:
```
commit-msg hook enabled (regression test enforcement)
```
But the previous version explicitly symlinked `scripts/commit-msg-regression.sh` → `.git/hooks/commit-msg`. The new flow only sets `core.hooksPath .githooks`, and the diff doesn't add `.githooks/commit-msg`. Unless that file already exists in `.githooks/`, regression-test enforcement on commit messages is silently lost.

**Action**: add a one-line `.githooks/commit-msg` shim that execs `scripts/commit-msg-regression.sh`, or drop the misleading echo.

### 2. `.env` fixture in `test-hermetic-gate.sh` writes to the repo root (medium)

The dotenv test does `cat > .env <<'EOF_ENV' …`, which mutates a real (gitignored) file in the developer's working tree. Backup/restore via `trap cleanup_env_fixture EXIT` is correct in the happy path, but:

- If `.env` is a **symlink** to a shared dotfile, the cleanup `cp` silently overwrites the target through the symlink. Use `rm -f .env` first, or `cp --remove-destination`.
- The trap **replaces** any outer `EXIT` trap (bash has only one EXIT trap). No outer trap exists today, but it's a future footgun.

A more robust pattern is to operate inside `mktemp -d` rather than the repo root.

### 3. Pre-push runs two full test invocations on every push (medium)

The default path runs `cargo test --locked --lib` and then `cargo test --locked --no-default-features --features libsql`. Different feature flags trigger a from-scratch rebuild for the second invocation. On a cold cache this is a noticeable pre-push penalty. Worth gating the dual-backend run behind a tier flag (e.g., `IRONCLAW_HERMETIC_DUAL_BACKEND=1`) since CI is described as authoritative.

### 4. `IRONCLAW_HERMETIC_KEEP_TMP=1` doesn't print the kept path (low)

The user has to scrape it out of the `==> hermetic environment` summary. Add a final `echo "tmpdir kept: $TMP_ROOT"` when `KEEP_TMP=1`.

### 5. Credential allowlist may be incomplete for IronClaw's LLM providers (low)

`crates/ironclaw_llm/CLAUDE.md` mentions `nearai`, `tinfoil`, `bedrock`, and `openai_compatible` providers. The list scrubs OpenAI/Anthropic/Gemini/AWS but I don't see `NEARAI_API_KEY` / `TINFOIL_API_KEY` / `OPENAI_COMPATIBLE_API_KEY`-style vars. Worth a quick `grep -r 'env::var' crates/ironclaw_llm` to confirm the full set and add anything missing.

### 6. `pre-commit-safety.sh` is now invoked unconditionally on every commit (low)

The old `.githooks/pre-commit` was a thin staged-files check. The new one always runs `pre-commit-safety.sh`, which itself early-exits when there are no `.rs` changes — but it also runs gateway JS `node --check` and i18n parity unconditionally (functionally an improvement). Just confirming this broadening is intentional.

### 7. `REMOTE_NAME="${1:-}"` silently swallows missing arg (low)

In `.githooks/pre-push`, `delta_lint.sh "$REMOTE_NAME"` is now called with an empty string when invoked manually. Previously `"$1"` errored fast under `set -u`. Verify `delta_lint.sh` handles empty argv gracefully — otherwise this hides misconfiguration.

## Suggestions

- Add a `.githooks/commit-msg` shim or remove the echo (item 1).
- Guard the `.env` fixture path in `test-hermetic-gate.sh` against symlinks, or operate inside a temp dir (item 2).
- Consider gating the second libsql test invocation behind a tier flag (item 3).
- Add NEAR/Tinfoil/Bedrock-specific credential names to `CREDENTIAL_VARS` if they exist (item 5).
- Print the kept tmpdir path when `KEEP_TMP=1` (item 4).
- Consider renaming `IRONCLAW_PREPUSH_TEST=0` to something like `IRONCLAW_HERMETIC_SKIP_TESTS=1` — `hermetic_gate.sh` is also a standalone entry point, and the current name is confusing outside the pre-push case.

## Security

No leaks observed. The dotenv-blocking trick is well thought through. The script confines all writes to `IRONCLAW_BASE_DIR` under `mktemp -d`. The single concern (item 2) is the test fixture writing `.env` to the repo root, not the gate itself.

## Verdict

Useful, well-scoped change. Recommend addressing item 1 (commit-msg regression) before merging; items 2–3 are good follow-ups but not strict blockers. Items 4–7 are polish.

---

PR #3301, opened 2026-05-06
ci: build ironclaw docker image with staging tag from main branch

think-in-universe:
## Summary

- Restores the full **Docker Image** workflow behavior from before the merge-queue cutover (#3104): hourly schedule, `:staging` and `:sha-*` tags, `runtime-staging` vs `runtime`, skip-when-already-built using the `ironclaw.git.sha` label, and **`ironclaw-dind`** repository dispatches for release vs rolling staging.
- Schedules a **checkout of `main`** (instead of the `staging` git branch) for the hourly job so rolling **`:staging`** Docker tags track **`main`** while keeping the same image tag names for deploys.

## Change Type

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [x] CI/Infrastructure
- [ ] Security
- [ ] Dependencies

## Linked Issue

None

## Validation

- [ ] `cargo build`
- [ ] Relevant tests pass: 
- [x] Manual testing: Reviewed `.github/workflows/docker.yml` for branch ref (`main`), schedule cron, tag/target logic, skip gate, and `ironclaw-dind` dispatch branches; YAML parses cleanly.

## Security Impact

None — CI-only change. Docker Hub publish and GitHub App token usage for `ironclaw-dind` dispatch match the prior intended behavior.

## Database Impact

None

## Blast Radius

**CI / release plumbing only.** Touches Docker Hub tags (`:staging`, `:sha-*`, and release tags when invoked from `release.yml`), hourly scheduled runs, and optional **`ironclaw-dind`** downstream builds. A mistake here

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -27,6 +27,9 @@ on:
         required: false
         type: string
         default: ""
+  # Rolling :staging image tags from main branch (hourly)
+  schedule:
+    - cron: '0 * * * *'
 
 env:
   IMAGE_NAME: nearaidev/ironclaw
@@ -44,6 +47,7 @@ jobs:
       - name: Checkout
         uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
         with:
+          ref: ${{ github.event_name == 'schedule' && 'main' || '' }}
           persist-credentials: false
 
       - name: Resolve source git commit
@@ -64,28 +68,20 @@ jobs:
           echo "version=${VERSION}" >> "$GITHUB_OUTPUT"
           echo "Detected version: ${VERSION}"
 
-      - name: Validate tag override
-        if: inputs.tag != ''
-        env:
-          INPUT_TAG: ${{ inputs.tag }}
-        run: |
-          if [[ ! "${INPUT_TAG}" =~ ^[A-Za-z0-9_][A-Za-z0-9._-]{0,127}$ ]]; then
-            echo "::error::Input tag '${INPUT_TAG}' does not match Docker tag grammar"
-            exit 1
-          fi
-          if [[ "${INPUT_TAG}" == "staging" ]]; then
-            echo "::error::The staging tag has been retired. Use a different tag or rely on release tags."
-            exit 1
-          fi
-
       - name: Determine tags
         id: tags
         env:
           VERSION: ${{ steps.version.outputs.version }}
+          EVENT_NAME: ${{ github.event_name }}
           IS_RELEASE_BUILD: ${{ inputs.release && 'true' || 'false' }}
           INPUT_TAG: ${{ inputs.tag }}
           SOURCE_SHA: ${{ steps.source_sha.outputs.sha }}
         run: |
+          if [[ -n "${INPUT_TAG}" && ! "${INPUT_TAG}" =~ ^[A-Za-z0-9_][A-Za-z0-9._-]{0,127}$ ]]; then
+            echo "::error::Input tag '${INPUT_TAG}' does not match Docker tag grammar"
+            exit 1
+          fi
+
           SHA="sha-${SOURCE_SHA::7}"
           echo "sha_tag=${SHA}" >> "$GITHUB_OUTPUT"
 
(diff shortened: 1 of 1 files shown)


nickpismenkov (commented):
# Code Review

## Overview

Restores the pre-#3104 staging-image workflow:
- Adds an **hourly `schedule` trigger** that builds from `main`.
- Re-introduces `:staging` and `:sha-*` tags and the `runtime-staging` Docker target.
- Adds a **skip gate**: pulls `:staging`, compares its `ironclaw.git.sha` label against the current source SHA, and skips build/push/dispatch if both `ironclaw` and `ironclaw-worker` are already current.
- Re-enables `ironclaw-dind` repo dispatch for both release and rolling-staging builds.
- Removes the previous "staging tag retired" rejection in `Validate tag override`.

CI-only; no app code touched.

## What looks good

- **Skip gate design is sound.** Reading `ironclaw.git.sha` from the published image is the right idempotency key, and missing/unparseable labels fail open (rebuild), which is the safer default.
- **Script-injection hardening in the Summary step.** Replacing `\${{ steps.tags.outputs.tags }}` interpolations with `env:` + `\${TAGS}` shell expansion is a real security improvement over the prior file.
- **`ref:` is conditional on `schedule`** — `workflow_call` from `release.yml` and `workflow_dispatch` paths still resolve to the caller's ref (empty string → default), so release builds keep checking out the tagged commit. ✅
- **Consistent gating.** All build/push/dispatch/Summary steps gate on `steps.check.outputs.skip != 'true'`, and a separate "Summary (skipped)" step keeps the run page informative.
- **Tag-grammar regex preserved** (just inlined into the determine-tags step).

## Issues / suggestions

### 1. Comment is wrong — says "Daily" but the cron is hourly
\`\`\`yaml
elif [[ \"\${EVENT_NAME}\" == \"schedule\" ]]; then
  # Daily staging: :staging + :sha-xxx
\`\`\`
Cron is `0 * * * *`. Update to `# Scheduled staging` or `# Hourly staging` to match.

### 2. `docker inspect` template can emit literal `<no value>`
\`\`\`bash
CURRENT_BUILT_SHA=\$(docker inspect --format='{{index .Config.Labels \"ironclaw.git.sha\"}}' \"\${IMAGE_NAME}:staging\" 2>/dev/null || echo \"\")
\`\`\`
If `docker inspect` succeeds but the label is absent (older image), Go templates render the literal string `<no value>`, not empty. The skip comparison still does the right thing (triggers rebuild), but logs print the confusing `<no value>` token. Use:
\`\`\`
--format='{{with index .Config.Labels \"ironclaw.git.sha\"}}{{.}}{{end}}'
\`\`\`

### 3. `Trigger ironclaw-dind` lost its `workflow_call` guard

Old gate:
\`\`\`
steps.app-token.outcome == 'success' &&
github.event_name == 'workflow_call' &&
steps.version.outputs.version != ''
\`\`\`
New gate is just `app-token.outcome == 'success' && skip != 'true'`. The body's branches still gate on `IS_RELEASE_BUILD` / `EVENT_NAME` / `INPUT_TAG`, so behavior is correct, **but** a `workflow_dispatch` with no `tag` and `release=false` now hits the dispatch step, evaluates both branches false, and produces a no-op. Harmless but no longer documented by the gate — either re-add the explicit guard or accept it; worth confirming this is intentional.

### 4. Consider a `concurrency` group for the schedule
Nothing prevents two scheduled runs from racing if one is still building when the next hour fires (a clean `runtime-staging` build can overrun an hour). Add:
\`\`\`yaml
concurrency:
  group: docker-\${{ github.event_name }}
  cancel-in-progress: false
\`\`\`
to coalesce overlapping schedule runs without disturbing release-time `workflow_call` invocations.

### 5. Hourly at `:00` piles on with everyone else's crons
Minor: shifting to e.g. `17 * * * *` reduces GitHub-side scheduler contention and the chance of delayed kickoff. Not blocking.

### 6. Skip gate only inspects `:staging`
If the same SHA was already built via a manual dispatch and tagged only as `:sha-xxx`, the schedule will rebuild it. That's an intentional simplification — `:staging` is the source of truth — and fine, but worth being explicit in the comment so a future reader doesn't add a \"smarter\" check.

### 7. Removed \"staging tag retired\" guard
Previously a manual `tag=staging` dispatch was rejected; now it's an explicit alternate path that selects `runtime-staging` and triggers the dind dispatch. Matches the PR intent — flagging here only because it widens the manual-dispatch surface area, so make sure the team intends manual `tag=staging` to be a supported operator workflow.

## Risk

- **Blast radius:** Docker Hub `:staging` / `:sha-*` tags + downstream `ironclaw-dind` dispatches. No app/runtime code, no DB, no schemas.
- **Cost:** Hourly cron runs are cheap when the skip gate hits (just checkout + two `docker pull`s + an inspect). On real changes, normal Docker build cost — bounded by GHA cache reuse.
- **Reversibility:** Easy — revert this PR or disable the schedule via repo settings.

## Verdict

**LGTM with minor nits.** The design is clean, the skip gate is well-thought-out, and the script-injection cleanup in Summary is a nice bonus. Worth fixing the \"Daily\" comment and the `<no value>` template before merge; concurrency group and cron offset are optional polish.

---
🤖 Reviewed with [Claude Code](https://claude.com/claude-code)

serrrfirat (dismissed):

nickpismenkov (approved):

PR #3303, opened 2026-05-06
refactor(reborn-memory): e2e coverage

nickpismenkov:
## Summary

E2E coverage for the reborn memory substrate landing in #3180. All-additive: zero production code changed, +1,690 LOC of test code across 5 new files. Two complementary tiers, libSQL-only:

- **Tier A — crate-level vertical** (`crates/ironclaw_memory/tests/e2e_*.rs`) drives `RepositoryMemoryBackend` + `MemoryBackendFilesystemAdapter` + `LibSqlMemoryDocumentRepository` + `ChunkingMemoryDocumentIndexer` through their public seams; assertions are raw SQL on `memory_documents` / `memory_chunks` / `memory_document_versions`.
- **Tier B — reborn-engine trace tests** (`tests/e2e_trace_memory_isolation.rs` + `tests/fixtures/llm_traces/memory/`) drive `memory_*` tools through the agent loop via `TraceLlm` + `TestRig`. Today these still hit the legacy host workspace; once PR 7 (product-tool migration) wires the new substrate behind `Workspace`, the same fixtures auto-cover the reborn path. This is the caller-tier guard `.claude/rules/testing.md` "Test Through the Caller" exists to enforce.

Tests that depend on PR #3180-only behavior are committed `#[ignore]`d with explicit reason strings; they activate automatically once this branch picks up #3180.

### Coverage map (PR #3180 numbered guarantees)

| #3180 guarantee | Tier A test(s) | Tier B test(s) |
|---|---|---|
| 1. Native (tenant, user, agent, project, path) scope on every operation | `cross_scope_writes_isolated_through_list_documents_under_libsql`, `hybrid_search_includes_path_in_caller_scope_only_under_libsql`, plus 4× `context_path_mismatch_*` (gated) | `memory_write_then_read_round_trip_persists_payload_through_tool_layer` |
| 2. `_none` virtual-path-only sentinel | `none_sentinel_is_virtual_only_and_rejected_by_scope_constructor` | — |
| 3. Prompt-write safety policy + protected paths | `protected_paths_high_risk_writes_blocked_at_libsql_backend`, `dispatch_protected_path_normalization_through_filesystem_adapter_under_libsql`, `protected_orchestrator_path_blocked_at_libsql_backend` (gated) | `memory_write_to_soul_md_rejects_through_tool_layer_no_persistence` (gated on PR 7) |
| 4. Hybrid RRF fusion + min_score-after-normalization + deterministic tiebreaker | `min_score_filter_drops_results_below_threshold_under_libsql` (gated), `search_breaks_ties_by_relative_path_ascending_under_libsql` (gated) | — |
| 5. `with_limit()` re-clamps `pre_fusion_limit` regardless of builder order | `pre_fusion_limit_is_at_least_limit_regardless_of_builder_order` | — |
| 6. Race-safe `replace_document_chunks_if_current` | `concurrent_writes_against_same_path_leave_no_orphan_chunks_under_libsql` | — |
| 7. `delete_document_chunks` removed from trait | (intentionally not added — would require committing code that fails to compile until #3180 lands) | — |
| 8. Version hash semantics + `MemoryAppendOutcome` | `version_chain_records_each_replace_with_correct_hash_under_libsql`, `compare_and_append_returns_appended_then_conflict_then_appended_with_fresh_hash`, `compare_and_append_emits_distinct_version_row_per_logical_change` | — |
| 9. Durability + migration idempotency | `documents_chunks_versions_durable_across_repository_reopen`, `run_migrations_is_idempotent_and_preserves_data`, `replace_overwrites_chunks_atomically_with_no_orphans` | — |
| 10. Bypass requires durable audit before persistence | `missing_event_sink_blocks_bypass_persistence_under_libsql`, `failing_event_sink_blocks_bypass_persistence_under_libsql`, `redacted_event_sink_records_warned_event_without_payload_substrings_under_libsql` | (indirectly via `memory_write_to_soul_md_rejects_*`) |

### Active vs. gated counts

- Tier A: **16 active**, 7 gated (5 in `e2e_scope_isolation_safety.rs` for #3180 `ensure_*_matches_context` guards + orchestrator path; 2 in `e2e_hybrid_search.rs` for #3180 search-fusion fixes).
- Tier B: **1 active**, 1 gated (PR 7 substrate wiring).

## Change Type

- [x] New feature (test-only, additive)

## Linked Issue

Related #3180 (substrate this exercises). The `#[ignore]` reasons name #3180 / PR 7 explicitly; reviewers can `cargo test -- --ignored` post-rebase to validate.

## Validation

- [x] `cargo clippy --all --benches --tests --examples --all-features -- -D warnings` — zero warnings
- [x] `cargo build`
- [x] Relevant tests pass:
  - `cargo test -p ironclaw_memory --features libsql` → **83 active passing**, 7 ignored (gated)
  - `cargo test --features libsql --test e2e_trace_memory_isolation` → 13 passing, 1 ignored
- [x] Feature-flag matrix: `cargo check -p ironclaw_memory` (default), `--no-default-features --features libsql`, `--all-features` — all green
- [ ] `cargo test --features integration` — N/A (no PostgreSQL involvement; user-confirmed scope is libSQL-only per #3180's "libsql always; postgres opt-in" posture, but this PR opts not to add the postgres tier)
- [x] Manual testing: each test file driven against a fresh `tempfile::tempdir()` libSQL DB; raw-SQL assertions confirm row counts and version-hash chains directly on `memory_documents` / `memory_chunks` / `memory_document_versions`

## Security Impact

None — test-only additions. The tests *exercise* the existing prompt-write safety policy (protected paths, redacted event sinks, bypass-without-audit blocking) but do not change it.

## Database Impact

None — no migrations, no schema changes. Tests use `LibSqlMemoryDocumentRepository::new(...).run_migrations()` against an ephemeral `tempfile::tempdir()` DB and inspect existing tables (`memory_documents`, `memory_chunks`, `memory_document_versions`, `memory_chunks_fts`).

## Blast Radius

Test-only; no production paths touched. Failure modes are limited to:
- The race-safe concurrent-write test (`concurrent_writes_against_same_path_leave_no_orphan_chunks_under_libsql`) is the highest-flake-risk test in this PR. It uses `tokio::join!` (single-threaded runtime, cooperative yields) and asserts on **end state** (one row, no orphan chunks referencing the loser's tokens), not on which writer wins. If the underlying `replace_document_chunks_if_current` semantics regress this is the test that catches it.
- 7 `#[ignore]`d tests will start running automatically post-rebase — if any fails after rebase, the failing assertion is the live regression signal for that PR #3180 invariant.

## Rollback Plan

Revert the merge. No production behavior changed; nothing depends on the new test files outside `cargo test`.

## Review Follow-Through

Two intentional design choices reviewers may want to weigh in on:

1. **Gated tests as commit-time documentation.** Each `#[ignore]` carries a one-line reason citing the PR it depends on (#3180 or PR 7). The alternative was to omit them and add later as follow-up PRs. The committed-but-gated pattern was chosen so the PR #3180 invariant → test mapping in this PR's body remains complete and the rebase moment surfaces all activations at once.
2. **No Tier B versioning/durability/search trace tests.** The plan considered them; cut for the "robust over breadth" preference. Tier A already pins versioning/durability/search invariants exhaustively. The single Tier B isolation file is the minimum needed to guard the caller-tier (substrate→adapter→tool dispatcher) axis. If reviewers want broader Tier B coverage, happy to add in a follow-up.

---

**Review track**: A (docs/tests/chore — purely additive test coverage, no production code touched)

### DIFF
--- a/crates/ironclaw_memory/tests/e2e_hybrid_search.rs
+++ b/crates/ironclaw_memory/tests/e2e_hybrid_search.rs
@@ -207,7 +207,7 @@ mod libsql_e2e {
     #[tokio::test]
     #[cfg_attr(
         not(feature = "pr3180-ready"),
-        ignore = "requires PR #3180 min_score-after-normalization fix (pre-#3180 filters against pre-fusion scores ≪ 1.0); enable with --features pr3180-ready when #3180 lands"
+        ignore = "requires the min_score-after-normalization fix. Empirically (2026-05-12) #3180 as merged scores both hybrid and FTS-only documents equally (1.0/1.0) under RRF fusion even with a pinned query embedding, so the test's premise that pinned-embedding favors the hybrid doc doesn't hold against the merged substrate yet. Stays gated for a followup PR that normalizes scores so they reflect actual vector contribution."
     )]
     async fn min_score_filter_drops_results_below_threshold_under_libsql() {
         // Seed two documents — one is a hybrid hit (FTS + vector), one is FTS-only.
@@ -329,7 +329,7 @@ mod libsql_e2e {
     #[tokio::test]
     #[cfg_attr(
         not(feature = "pr3180-ready"),
-        ignore = "requires PR #3180 deterministic tiebreaker (path ascending); enable with --features pr3180-ready when #3180 lands"
+        ignore = "requires deterministic tie-breaking by path ascending. Empirically (2026-05-12) #3180 as merged orders tied results by some other key (observed [a, c, b] vs expected [a, b, c]) — likely document insertion order or row id. No `tiebreak`/`deterministic_order` symbol exists in `crates/ironclaw_memory/src/` yet. Stays gated for the followup PR that adds the path-ascending secondary sort."
     )]
     async fn search_breaks_ties_by_relative_path_ascending_under_libsql() {
         // Seed three docs with identical FTS+vector hits. PR #3180 fixes ties to
--- a/crates/ironclaw_memory/tests/e2e_hybrid_search.rs
+++ b/crates/ironclaw_memory/tests/e2e_hybrid_search.rs
@@ -65,9 +65,17 @@ mod libsql_e2e {
 
     #[tokio::test]
     async fn hybrid_search_includes_path_in_caller_scope_only_under_libsql() {
-        // Seed two scopes with the same FTS+vector token and a hit-attracting
-        // chunk. Search from one scope must only return that scope's document,
-        // proving the search path classifies by scope at the SQL boundary.
+        // Seed documents along EVERY scope axis with the same FTS+vector
+        // token. Search from one scope must only return that scope's
+        // document. The active hybrid-search isolation guarantee is that
+        // `backend.search` filters by the full `MemoryDocumentScope`
+        // (tenant_id, user_id, agent_id, project_id) at the SQL boundary
+        // — a regression that filters only by `user_id` (or drops one
+        // of the other axes) would leak same-token documents from the
+        // dropped-axis siblings.
+        //
+        // Caller scope: tenant-a / alice / agent=None / project=None.
+        // Variants seeded below differ along exactly one axis each.
         let (db, _dir) = libsql_db().await;
         let repository = Arc::new(LibSqlMemoryDocumentRepository::new(db.clone()));
         repository.run_migrations().await.unwrap();
@@ -83,30 +91,90 @@ mod libsql_e2e {
                 .with_capabilities(full_search_capabilities()),
         );
 
-        let alice_scope = MemoryDocumentScope::new("tenant-a", "alice", None).unwrap();
-        let bob_scope = MemoryDocumentScope::new("tenant-a", "bob", None).unwrap();
+        let caller_scope =
+            MemoryDocumentScope::new_with_agent("tenant-a", "alice", None, None).unwrap();
+        let caller_path =
+            MemoryDocumentPath::new_with_agent("tenant-a", "alice", None, None, "notes/visible.md")
+                .unwrap();
+        // Sibling scopes differing along one axis at a time. Each must
+        // be invisible to a search rooted at `caller_scope`.
+        let sibling_paths = [
+            // Different user (was the only axis exercised before).
+            (
+                MemoryDocumentScope::new_with_agent("tenant-a", "bob", None, None).unwrap(),
+                MemoryDocumentPath::new_with_agent(
+                    "tenant-a",
+                    "bob",
+                    None,
+                    None,
+                    "notes/leaked-user.md",
+                )
+                .unwrap(),
+            ),
+            // Different tenant.
+            (
+                MemoryDocumentScope::new_with_agent("tenant-b", "alice", None, None).unwrap(),
+                MemoryDocumentPath::new_with_agent(
+                    "tenant-b",
+                    "alice",
+                    None,
+                    None,
+                    "notes/leaked-tenant.md",
+                )
+                .unwrap(),
+            ),
+            // Different project (same tenant + user + agent).
+            (
+                MemoryDocumentScope::new_with_agent("tenant-a", "alice", None, Some("project-1"))
+                    .unwrap(),
+                MemoryDocumentPath::new_with_agent(
+                    "tenant-a",
+                    "alice",
+                    None,
+                    Some("project-1"),
+                    "notes/leaked-project.md",
+                )
+                .unwrap(),
+            ),
+            // Different agent (same tenant + user + project).
+            (
+                MemoryDocumentScope::new_with_agent("tenant-a", "alice", Some("agent-x"), None)
+                    .unwrap(),
+                MemoryDocumentPath::new_with_agent(
+                    "tenant-a",
+                    "alice",
+                    Some("agent-x"),
+                    None,
+                    "notes/leaked-agent.md",
+                )
+                .unwrap(),
+            ),
+        ];
 
         backend
             .write_document(
-                &MemoryContext::new(alice_scope.clone()),
-                &MemoryDocumentPath::new("tenant-a", "alice", None, "notes/visible.md").unwrap(),
+                &MemoryContext::new(caller_scope.clone()),
+                &caller_path,
                 b"literal hybrid-vector",
             )
             .await
             .unwrap();
-        backend
-            .write_document(
-                &MemoryContext::new(bob_scope.clone()),
-                &MemoryDocumentPath::new("tenant-a", "bob", None, "notes/leaked.md").unwrap(),
-                b"literal hybrid-vector",
-            )
-            .await
-            .unwrap();
+        for (scope, path) in &sibling_paths {
+            backend
+                .write_document(
+                    &MemoryContext::new(scope.clone()),
+                    path,
+                    b"literal hybrid-vector",
+                )
+                .await
+                .unwrap();
+        }
 
-        // Alice's search must only see Alice's document.
-        let alice_results = backend
+        // Caller's search must only see the caller's document — every
+        // sibling axis must be filtered out at the SQL boundary.
+        let caller_results = backend
             .search(
-                &MemoryContext::new(alice_scope),
+                &MemoryContext::new(caller_scope),
                 MemorySearchRequest::new("literal")
                     .unwrap()
                     .with_limit(10)
@@ -115,22 +183,32 @@ mod libsql_e2e {
             .await
             .unwrap();
 
-        let returned_paths: Vec<String> = alice_results
+        let returned_paths: Vec<String> = caller_results
             .iter()
             .map(|r| r.path.relative_path().to_string())
             .collect();
         assert_eq!(
             returned_paths,
             vec!["notes/visible.md".to_string()],
-            "alice's search must only return alice's document",
+            "search must only return the caller-scope document; got {returned_paths:?}",
         );
-        for result in &alice_results {
+        // Defense-in-depth: confirm each returned hit matches the caller
+        // along all four axes, so a regression that returned the caller's
+        // own path under a sibling scope (impossible by construction, but
+        // pinned for future-proofing) is still caught.
+        for result in &caller_results {
+            assert_eq!(result.path.tenant_id(), "tenant-a");
             assert_eq!(result.path.user_id(), "alice");
+            assert_eq!(result.path.agent_id(), None);
+            assert_eq!(result.path.project_id(), None);
         }
     }
 
     #[tokio::test]
-    #[ignore = "requires PR #3180 min_score-after-normalization fix (pre-#3180 filters against pre-fusion scores ≪ 1.0)"]
+    #[cfg_attr(
+        not(feature = "pr3180-ready"),
+        ignore = "requires PR #3180 min_score-after-normalization fix (pre-#3180 filters against pre-fusion scores ≪ 1.0); enable with --features pr3180-ready when #3180 lands"
+    )]
     async fn min_score_filter_drops_results_below_threshold_under_libsql() {
         // Seed two documents — one is a hybrid hit (FTS + vector), one is FTS-only.
         // Setting a min_score above the FTS-only RRF-normalized score must drop it.
@@ -167,11 +245,26 @@ mod libsql_e2e {
             .await
             .unwrap();
 
+        // Pin the query embedding to exactly the hybrid.md vector so the
+        // vector-side contribution unambiguously favors hybrid.md over
+        // fts-only.md. Without this, the query "literal" would fall through
+        // the embedder's fallback branch ([0,0,1]) and have zero cosine
+        // similarity with either document — the vector side could not
+        // separate them and the test would not reliably create the
+        // hybrid-above-FTS-only score gap it advertises.
+        //
+        // FTS still matches both docs (both contain the token "literal");
+        // the override only constrains the vector side.
+        let hybrid_doc_embedding = vec![1.0_f32, 0.0, 0.0];
+
         // Without min_score, both docs should be returned.
         let unfiltered = backend
             .search(
                 &context,
-                MemorySearchRequest::new("literal").unwrap().with_limit(10),
+                MemorySearchRequest::new("literal")
+                    .unwrap()
+                    .with_limit(10)
+                    .with_query_embedding(hybrid_doc_embedding.clone()),
             )
             .await
             .unwrap();
@@ -181,12 +274,16 @@ mod libsql_e2e {
             unfiltered_paths.contains(&"notes/hybrid.md"),
             "unfiltered search should include hybrid.md, got {unfiltered_paths:?}",
         );
+        assert!(
+            unfiltered_paths.contains(&"notes/fts-only.md"),
+            "unfiltered search should include fts-only.md so the min_score gap is real, \
+             got {unfiltered_paths:?}",
+        );
 
-        // Apply a min_score that is between the two scores. The hybrid result
-        // (FTS+vector) must clear the threshold; the FTS-only result must drop.
-        // We don't assume the absolute score values — instead, derive the
-        // threshold from the actual results so the test stays robust against
-        // RRF/k changes.
+        // Apply a min_score strictly between the two scores. The hybrid
+        // result (FTS+vector) must clear the threshold; the FTS-only
+        // result must drop. With the pinned query embedding above, this
+        // is now a hard ordering invariant — no `* 0.99` fallback path.
         let hybrid_score = unfiltered
             .iter()
             .find(|r| r.path.relative_path() == "notes/hybrid.md")
@@ -195,14 +292,16 @@ mod libsql_e2e {
         let fts_only_score = unfiltered
             .iter()
             .find(|r| r.path.relative_path() == "notes/fts-only.md")
-            .map(|r| r.score);
-        // If the FTS-only doc never matched (e.g., feature pruned it pre-#3180),
-        // the test still proves the threshold gate works: a min_score above the
-        // hybrid score drops everything.
-        let threshold = match fts_only_score {
-            Some(score) if score < hybrid_score => (score + hybrid_score) / 2.0,
-            _ => hybrid_score * 0.99, // cannot meaningfully threshold; gate on hybrid only
-        };
+            .map(|r| r.score)
+            .expect(
+                "fts-only.md should be present in the unfiltered result so the gap is testable",
+            );
+        assert!(
+            fts_only_score < hybrid_score,
+            "with pinned query embedding favoring hybrid.md, the FTS-only doc must score \
+             strictly below the hybrid doc; got hybrid={hybrid_score} fts_only={fts_only_score}",
+        );
+        let threshold = (fts_only_score + hybrid_score) / 2.0;
 
         let filtered = backend
             .search(
@@ -210,34 +309,28 @@ mod libsql_e2e {
                 MemorySearchRequest::new("literal")
                     .unwrap()
                     .with_limit(10)
+                    .with_query_embedding(hybrid_doc_embedding)
                     .with_min_score(threshold),
             )
             .await
             .unwrap();
 
         let filtered_paths: Vec<&str> = filtered.iter().map(|r| r.path.relative_path()).collect();
-        if let Some(fts_score) = fts_only_score
-            && fts_score < hybrid_score
-        {
-            assert!(
-                filtered_paths.contains(&"notes/hybrid.md"),
-                "hybrid hit must clear threshold {threshold}: {filtered_paths:?}",
-            );
-            assert!(
-                !filtered_paths.contains(&"notes/fts-only.md"),
-                "fts-only hit (score {fts_score}) must be filtered below {threshold}: {filtered_paths:?}",
-            );
-        } else {
-            // Sanity: the hybrid result still clears its own near-threshold gate.
-            assert!(
-                filtered_paths.contains(&"notes/hybrid.md"),
-                "hybrid hit must clear threshold {threshold}: {filtered_paths:?}",
-            );
-        }
+        assert!(
+            filtered_paths.contains(&"notes/hybrid.md"),
+            "hybrid hit must clear threshold {threshold}: {filtered_paths:?}",
+        );
+        assert!(
+            !filtered_paths.contains(&"notes/fts-only.md"),
+            "fts-only hit (score {fts_only_score}) must be filtered below {threshold}: {filtered_paths:?}",
+        );
     }
 
     #[tokio::test]
-    #[ignore = "requires PR #3180 deterministic tiebreaker (path ascending)"]
+    #[cfg_attr(
+        not(feature = "pr3180-ready"),
+        ignore = "requires PR #3180 deterministic tiebreaker (path ascending); enable with --features pr3180-ready when #3180 lands"
+    )]
     async fn search_breaks_ties_by_relative_path_ascending_under_libsql() {
         // Seed three docs with identical FTS+vector hits. PR #3180 fixes ties to
         // resolve by path ascending; pre-#3180 ordering is implementation-defined.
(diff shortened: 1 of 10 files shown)


zmanian:
## Code Review

### Overview
Pure additive test coverage (+1,690 LOC, zero production changes) for the reborn memory substrate landing in #3180. Two tiers — Tier A vertical against `RepositoryMemoryBackend` + `LibSqlMemoryDocumentRepository` with raw-SQL assertions; Tier B caller-tier through `memory_*` tools via `TraceLlm` + `TestRig`. PR-#3180 invariants 1–10 map 1:1 to specific test names. 16 active + 7 gated in Tier A; 1 active + 1 gated in Tier B.

### Strengths
- **Coverage map** is auditable — each #3180 numbered guarantee → specific test names.
- **Caller-tier discipline** matches `.claude/rules/testing.md` "Test Through the Caller."
- **Race-safety test asserts the right shape**: `concurrent_writes_against_same_path_leave_no_orphan_chunks_under_libsql` checks end-state (XOR winner + no loser tokens in chunks), not which writer wins — correct for `tokio::join!` on a single-threaded runtime.
- **`min_score` threshold derived from observed scores** with a documented fallback — won't break on RRF/k tuning.
- **Bypass-without-audit invariants** (`missing_event_sink_*`, `failing_event_sink_*`) exercised directly against libSQL, mirroring the crate-contract suite at the DB layer.
- **Redaction test asserts the negative**: `format!(\"{:?}\", recorded[0])` must not contain payload markers — catches accidental Debug leakage.

### Suggestions

1. **Title typo**: `refactor(reborn-momery)` → `reborn-memory`. Cosmetic, but it'll land in the squash.

2. **Helper duplication across the three Tier A files.** `libsql_db()`, scope/path helpers, `count_documents`, `chunks_for_path`, and `DeterministicEmbeddingProvider` are reimplemented (with subtle variations) in each file — and two of the three use different silent-load-bearing `OWNER_KEY` constants (`...:project:_none` vs `...:project:project-1`). Worth a `tests/common/libsql_helpers.rs` follow-up to prevent SQL drift.

3. **`_link_in_memory_repo_for_unused_imports` shim** (`e2e_scope_isolation_safety.rs:1544`) is a workaround for the unused-import warning. `#[allow(unused_imports)]` on the import line is more honest than a fake function — or drop the import outright if nothing in the active body needs it.

4. **`count_versions ... == 0` in `documents_chunks_versions_durable_across_repository_reopen`** is tautological against the `read_versions` query for an unwritten path. A second write after the reopen would actually exercise version durability across the drop.

5. **`min_score_filter` branch B weakens the assertion.** When `fts_only_score >= hybrid_score` or absent, the test only asserts the hybrid hit clears `hybrid * 0.99` — that's a weak positive signal. If the FTS-only-below-hybrid ordering doesn't hold, the test premise is wrong and should fail loudly, not pass weakly.

### Non-issues (documented in PR body)
- libSQL-only scope, gated-tests-at-rest, and Tier B breadth tradeoffs are explicitly called out and reasonable.

### Risk
- **Blast radius**: zero (test-only).
- **Flake risk**: race test is the highest; design is sound.
- **Rebase**: 7 gated tests activate atomically post-#3180 — by design.
- **Security**: tests exercise but don't change the safety policy.

### Recommendation
**Approve** with optional follow-ups (title typo, helper consolidation, shim cleanup, durability + min_score assertion strengthening). High-quality test additions with a clear coverage map and good discipline around end-state assertions on race-prone paths.

zmanian:
## Re-review follow-ups

Stepping back through this with fresh eyes after my [original review](https://github.com/nearai/ironclaw/pull/3303#issuecomment-4391116650). Two new findings worth raising as non-blocking follow-ups, plus a third smaller one:

### 1. The race-safety test does not actually exercise concurrency

`concurrent_writes_against_same_path_leave_no_orphan_chunks_under_libsql` runs under default `#[tokio::test]`, which is a single-threaded current-thread runtime. `tokio::join!` polls both futures cooperatively on **one thread**, so the test only catches races that surface at `.await` points within `write_document` — true preemptive interleavings against `replace_document_chunks_if_current` are never hit.

The end-state assertion (XOR winner + no loser tokens in chunks) is correct, but the test gives false confidence about race coverage. Since this is the canonical regression guard for PR #3180 invariant 6 (race-safe `replace_document_chunks_if_current`), worth strengthening:

```rust
#[tokio::test(flavor = "multi_thread", worker_threads = 2)]
async fn concurrent_writes_against_same_path_leave_no_orphan_chunks_under_libsql() {
    // ...
    let writer_a = tokio::spawn({ /* clone backend/context/path */ });
    let writer_b = tokio::spawn({ /* ... */ });
    let (a, b) = tokio::join!(writer_a, writer_b);
    a.unwrap().unwrap();
    b.unwrap().unwrap();
    // ...
}
```

### 2. Protected-path trace fixture should explicitly set `all_tools_succeeded: false`

`tests/fixtures/llm_traces/memory/write_to_protected_path_rejected.json` omits `all_tools_succeeded`. If the trace harness defaults that field to `true`, the gated Tier B test (`memory_write_to_soul_md_rejects_through_tool_layer_no_persistence`) could pass for the wrong reason once PR 7 unwires the gate. Set it explicitly:

```json
"expects": {
  "tools_used": ["memory_write"],
  "all_tools_succeeded": false,
  "min_responses": 1
}
```

### 3. Bypass-blocking tests don't prove the audit ordering

`missing_event_sink_blocks_bypass_persistence_under_libsql` and `failing_event_sink_*` confirm the error and zero rows, but they don't distinguish "event-sink check happened **before** persistence" from "persistence happened then was rolled back". A second test where the sink succeeds but the DB write is mocked to fail (or simply asserting the audit row exists when persistence succeeds) would close that gap.

---

Not blocking on these — the existing approval stands. (1) is the most material; happy to land follow-up commits in a separate PR if easier.

crates/ironclaw_memory/tests/e2e_hybrid_search.rs
```diff
@@ -0,0 +1,471 @@
...
+            vec!["notes/visible.md".to_string()],
+            "alice's search must only return alice's document",
+        );
+        for result in &alice_results {
+            assert_eq!(result.path.user_id(), "alice");
+        }
+    }
+
+    #[tokio::test]
+    #[ignore = "requires PR #3180 min_score-after-normalization fix (pre-#3180 filters against pre-fusion scores ≪ 1.0)"]
```
serrrfirat: **High Severity**

The future-gated tests are behind static `#[ignore]`, so they will not auto-activate when #3180/PR7 lands.

The PR body says these tests activate automatically once the branch picks up the prerequisite work, but Rust keeps `#[ignore]` tests skipped until someone explicitly runs `--ignored` or edits the attributes. That means the regression coverage for min-score, deterministic tie-breaking, path/context guards, orchestrator protection, and tool-layer protected writes can remain disabled after the rebase that should make them mandatory.

Please replace static ignores with a real conditional gate that flips when the dependency is present (feature/cfg/versioned helper), or land the tests in the follow-up PR that removes the ignore at the same time. At minimum, add a CI job that runs these ignored tests once the prerequisite branch is merged.

tests/e2e_trace_memory_isolation.rs
```diff
@@ -0,0 +1,105 @@
...
+        );
+
+        // SOUL.md must not have a persisted row. The Database trait exposes
+        // `get_document_by_path(user_id, agent_id, path)` which returns
+        // `Err(WorkspaceError::NotFound)` when the row is absent — the cleanest
+        // way to assert non-persistence without bypassing the trait into
+        // dialect-specific raw SQL.
+        let lookup = rig
+            .database()
+            .get_document_by_path(rig.owner_id(), None, "SOUL.md")
```
serrrfirat: **High Severity**

The non-persistence assertion checks the rig owner, not the effective channel user that performed the tool call.

`TestRigBuilder` keeps the default channel user as `test-user` unless secrets/skills are enabled, while `rig.owner_id()` is the config owner (`default`). If `memory_write` incorrectly persists `SOUL.md` for the actual caller but reports failure, this lookup can still pass because it checks a different user’s workspace.

Please assert against the effective channel user used by the test channel, or configure/expose the rig so the channel user and owner are identical before the trace runs. That keeps this caller-tier security test from becoming a false positive.

crates/ironclaw_memory/tests/e2e_hybrid_search.rs
```diff
@@ -0,0 +1,471 @@
...
+            .map(|r| r.score)
+            .expect("hybrid.md should be present");
+        let fts_only_score = unfiltered
+            .iter()
+            .find(|r| r.path.relative_path() == "notes/fts-only.md")
+            .map(|r| r.score);
+        // If the FTS-only doc never matched (e.g., feature pruned it pre-#3180),
+        // the test still proves the threshold gate works: a min_score above the
+        // hybrid score drops everything.
+        let threshold = match fts_only_score {
```
serrrfirat: **Medium Severity**

The min-score test does not reliably create the hybrid-vs-FTS score gap it claims to validate.

The query `literal` goes through the embedding provider’s fallback branch (`[0,0,1]`), while `hybrid.md` embeds to `[1,0,0]` and `fts-only.md` embeds to `[0,1,0]`. That makes the vector side unable to favor the hybrid document, so the test can take the permissive `hybrid_score * 0.99` fallback and never assert that `notes/fts-only.md` was filtered out. A broken min-score-after-normalization implementation could still pass.

Please make the query embedding explicitly favor `hybrid.md` (for example with `with_query_embedding(vec![1.0, 0.0, 0.0])` or a provider branch for this query) and fail if the FTS-only score is not below the hybrid score instead of silently weakening the assertion.

crates/ironclaw_memory/tests/e2e_persistence_versioning.rs
```diff
@@ -0,0 +1,439 @@
...
+                    &context,
+                    &doc_path("notes/durable.md"),
+                    b"durable content body",
+                )
+                .await
+                .unwrap();
+        } // first repository / backend dropped here
+
+        // Reopen.
+        let repository_reopened = Arc::new(LibSqlMemoryDocumentRepository::new(db.clone()));
```
serrrfirat: **Medium Severity**

This durability test re-wraps the same `Arc<libsql::Database>` instead of reopening the database file.

That proves data survives a new repository object/new connection, but not a true DB-handle reopen. A bug that only appears after dropping the underlying database handle and opening the file again would still pass despite the test name and PR body claiming durability-across-reopen coverage.

Please keep the temp DB path, drop the first database/repository/backend handles, build a fresh `libsql::Database` from that path, and run the read/count assertions through the fresh handle.

crates/ironclaw_memory/tests/e2e_persistence_versioning.rs
```diff
@@ -0,0 +1,439 @@
...
+        assert_eq!(outcome, MemoryAppendOutcome::Appended);
+
+        assert_eq!(
+            repository.read_document(&path).await.unwrap().unwrap(),
+            b"hello world"
+        );
+        // Version row count after one replace + one append: prior content was
+        // recorded once before the append (the initial `"hello "`).
+        let versions = read_versions(&db, "notes/append-versioned.md").await;
+        assert!(
```
serrrfirat: **Medium Severity**

The append versioning test does not assert the “distinct version row per logical change” invariant in its name.

`!versions.is_empty()` plus checking `versions[0]` would still pass if `compare_and_append_document` started emitting duplicate version rows for one append. That leaves the row-cardinality regression this test advertises uncovered.

Please assert the exact version row count and contents for the logical changes being tested (for example, exactly one prior-content version row after the initial write plus one successful append, and no extra duplicates).

crates/ironclaw_memory/tests/e2e_scope_isolation_safety.rs
```diff
@@ -0,0 +1,564 @@
...
+        assert!(
+            err.contains("high_risk_prompt_injection"),
+            "expected orchestrator path rejection, got {err}",
+        );
+        assert_eq!(count_documents_total(&db).await, 0);
+    }
+
+    #[tokio::test]
+    async fn dispatch_protected_path_normalization_through_filesystem_adapter_under_libsql() {
+        // Drive the adapter against canonical and lexically-equivalent variants of
```
serrrfirat: **Medium Severity**

The adapter normalization test claims to cover lexically-equivalent protected-path variants, but only exercises canonical `SOUL.md`.

As written, a regression in adapter-layer normalization (for example accepting a `.`/`..` variant or other lexical equivalent before protected-path classification) would still pass because no variant path is actually written.

Please add at least one lexically equivalent protected path through `MemoryBackendFilesystemAdapter` and assert it rejects with no persistence, or narrow the comment/test name to the canonical-path case.

crates/ironclaw_memory/tests/e2e_hybrid_search.rs
```diff
@@ -0,0 +1,471 @@
...
+//! Targets PR #3180 invariants 4–7:
+//!   - `with_limit()` re-clamps `pre_fusion_limit` (the `pre_fusion >= limit`
+//!     invariant holds regardless of builder call order)
+//!   - `min_score` filter drops results that fall below the threshold (#3180
+//!     applies the filter after normalization)
+//!   - deterministic tiebreaker breaks ties by relative path ascending
+//!   - race-safe `replace_document_chunks_if_current` leaves no orphan chunks
+//!   - hybrid search results are scoped to the caller's `MemoryDocumentScope`
+
+#[cfg(feature = "libsql")]
```
serrrfirat, line 12: **Medium Severity**

The new `ironclaw_memory` package integration tests appear to compile in PR CI but not run in the test job.

The PR adds tests under `crates/ironclaw_memory/tests/*`, but `.github/workflows/test.yml` runs `cargo test ${{ matrix.flags }} -- --nocapture` from the workspace root without `--workspace` or `-p ironclaw_memory`. In this workspace the root package is `ironclaw`, so these package integration tests can be skipped by the PR test job even though clippy compiles them. That leaves the advertised Reborn memory coverage unenforced in CI.

Please add an explicit CI invocation such as `cargo test -p ironclaw_memory --features libsql --tests -- --nocapture` (or make the main test command workspace-wide if that is intended).

crates/ironclaw_memory/tests/e2e_hybrid_search.rs
```diff
@@ -0,0 +1,471 @@
...
+                .with_embedding_provider(provider.clone()),
+        );
+        let backend = Arc::new(
+            RepositoryMemoryBackend::new(repository)
+                .with_indexer(indexer)
+                .with_embedding_provider(provider)
+                .with_capabilities(full_search_capabilities()),
+        );
+
+        let alice_scope = MemoryDocumentScope::new("tenant-a", "alice", None).unwrap();
```
serrrfirat: **Medium Severity**

The active hybrid search isolation test only varies `user_id`, not the full `MemoryDocumentScope`.

The file-level contract says hybrid search is scoped to the caller’s `MemoryDocumentScope`, but this test seeds Alice vs Bob only. A search implementation that still filters user while accidentally dropping tenant, project, or agent predicates would pass while leaking same-user documents across those other axes. The all-axis test in `e2e_scope_isolation_safety.rs` covers `list_documents`, not `backend.search`.

Please seed same-token search documents across tenant, project, and agent axes as well, then assert the search returns only the exact caller scope.

tests/e2e_trace_memory_isolation.rs
```diff
@@ -0,0 +1,105 @@
...
+
+        // Tool result for memory_read must contain the exact marker that the
+        // first-turn memory_write persisted.
+        let results = rig.tool_results();
+        let read_result = results
+            .iter()
+            .find(|(name, _)| name == "memory_read")
+            .map(|(_, preview)| preview.clone())
+            .expect("memory_read result must be captured");
+        assert!(
```
serrrfirat, line 49: **Medium Severity**

The tool-layer round-trip assertion only checks that the read preview contains the marker, not exact persisted content.

The test comment says it proves exact byte persistence, but `contains("deterministic-marker-42")` would still pass if the tool returned extra data, stale surrounding content, or mutated content that still included the marker. That weakens the caller-tier guard this trace is meant to provide for the substrate swap.

Please parse/assert the `memory_read` result content exactly, or read the persisted document directly from the workspace/database and compare it to exactly `deterministic-marker-42`.

crates/ironclaw_memory/tests/e2e_scope_isolation_safety.rs
```diff
@@ -0,0 +1,564 @@
...
+        let recorded = events.events();
+        assert_eq!(recorded.len(), PROTECTED_PATHS.len());
+        for (relative, event) in PROTECTED_PATHS.iter().zip(recorded.iter()) {
+            assert_eq!(event.kind, PromptWriteSafetyEventKind::Rejected);
+            assert_eq!(
+                event.reason_code,
+                Some(PromptSafetyReasonCode::HighRiskPromptInjection),
+                "{relative}",
+            );
+            let path_class = event
```
serrrfirat: **Medium Severity**

The protected-path rejection audit only checks that a path class exists, not that it matches the rejected path.

The comment says every rejection produced an event with the matching path class, but the assertion only verifies `protected_path_class.is_some()`. If all rejected writes were audited as the wrong protected path class, this test would still pass and the audit signal would be misleading.

Please compare `event.protected_path_class.relative_path()` against the expected normalized path for each `PROTECTED_PATHS` entry.

zmanian (changes requested):
Request changes. Solid structure and good invariant→test mapping in the PR body, but two issues mean the headline value (regression guards for #3180) doesn't actually fire.

(Title typo: "momery" → "memory".)

## Blockers

1. **`#[ignore]` gates won't auto-activate post-rebase.** The PR body says gated tests "activate automatically" once #3180 lands — that's not how `#[ignore]` works; it's unconditional and a rebase doesn't flip it. Either gate via a cfg flag/feature tied to #3180, or plan to commit-and-remove the `#[ignore]` attributes in the dependent PR. As written, the advertised regression coverage is a no-op.

2. **CI doesn't run these tests.** `.github/workflows/test.yml` runs `cargo test` from the workspace root (`ironclaw` package), but `crates/ironclaw_memory/tests/*` aren't picked up without `-p ironclaw_memory` or `--workspace`. Please add e.g. `cargo test -p ironclaw_memory --features libsql --tests` to CI so these guards are enforced.

3. **Caller-tier owner mismatch in `tests/e2e_trace_memory_isolation.rs:96`.** Asserts non-persistence against `rig.owner_id()` (`default`), but the tool runs as channel user `test-user`. A regression that persists `SOUL.md` for `test-user` would still pass this guard. Align the rig owner with the channel user, or assert against the effective channel user.

## Medium (concur with serrrfirat's review)

4. `e2e_hybrid_search.rs:202` — min-score test embedding fallback `[0,0,1]` doesn't separate hybrid from FTS-only docs, and the `* 0.99` permissive fallback silently weakens the assertion. Pin the query embedding explicitly.
5. `e2e_persistence_versioning.rs:149` — durability test re-wraps the same `Arc<Database>` rather than reopening from path. Drop and reopen.
6. `e2e_persistence_versioning.rs:293` — `!is_empty()` on version rows would pass duplicates. Assert exact count.
7. `e2e_scope_isolation_safety.rs:346` — only canonical `SOUL.md` exercised; either add `.`/`..`/lexical variants or rename the test.
8. `e2e_hybrid_search.rs:86` — search isolation only varies `user_id`; add tenant/project/agent axes for `backend.search`.
9. `tests/e2e_trace_memory_isolation.rs:48` — round-trip uses `contains()`, weakening the exact-byte claim. Use equality on persisted bytes.
10. `e2e_scope_isolation_safety.rs:95` — audit-class assertion only checks `.is_some()`, not that the class matches the rejected path.

## Notes

`.unwrap()`/`.expect()` usage in tests is fine per project rules. No production code, no schema changes, no dispatcher concerns — clean blast radius once the gating/CI/owner issues are fixed.

Items 1–3 are blockers; the rest can land here or as a follow-up but shouldn't be deferred long.

henrypark133 (changes requested):
What looks good:
- The follow-up fixes addressed most prior test-quality comments: memory crate tests now have an explicit CI step, hybrid search covers all scope axes, and persistence/versioning assertions are much stronger.
- The trace test now checks persisted bytes directly under the effective channel user, avoiding the earlier owner-vs-channel-user false positive.
- CI is green, and diff checks passed locally.

Findings:
1. High - `tests/e2e_trace_memory_isolation.rs:82`: the protected-path caller-level regression test is still `#[ignore]` unless `pr3180-ready` is enabled, but no workflow enables that feature. Why it matters: the advertised tool-layer `SOUL.md` rejection guard does not run in normal CI, so protected-path regressions can merge with green checks. Expected fix direction: enable `pr3180-ready` in the relevant CI path now that the prerequisite memory substrate PR is merged, or move this rejection coverage into an always-running caller-level test.

Low-priority notes:
- `tests/e2e_trace_memory_isolation.rs:40`: `memory_read` is still asserted via a preview `contains(...)` check, with exactness proven only by the database read. That is enough for persistence, but not for exact caller-visible tool output; a full tool-result helper would make this stronger.

Summary:
- Recommended verdict: Request changes.
- Prior feedback status: mostly addressed, with one important CI-enforcement gap still current.
- Review coverage: worktree-backed review of CI, memory substrate tests, trace fixtures, and prior unresolved review threads; blocker verified separately.
- Residual risk: the bypass audit-ordering "sink succeeds, DB write fails" case remains acknowledged but untested.

nickpismenkov:
Thanks for the careful re-review, @henrypark133. Pushed [`263aab42f`](https://github.com/nearai/ironclaw/pull/3303/commits/263aab42f9c06a65fcbb0457612ceeb358d7af70) to address the High finding. A few subtleties surfaced when I tried your proposed fix — full details below.

## Why "enable `pr3180-ready` in CI" doesn't work today

I started by enabling `--features pr3180-ready` locally to confirm it would just light up the gated tests. Instead 7 tests failed, because **#3180 as merged is narrower than what the gated tests anticipate.** Specifically (verified via grep against `crates/ironclaw_memory/src/` on `reborn-integration` HEAD):

- `ensure_path_matches_context` / `ensure_scope_matches_context` — no such symbol exists in the substrate yet. The 4 context-mismatch tests in `e2e_scope_isolation_safety.rs` were written anticipating these guards.
- Deterministic tiebreaking by path ascending — no `tiebreak` / `deterministic_order` symbol; merged substrate orders ties by something else (observed `[a, c, b]` vs expected `[a, b, c]`).
- `.system/engine/orchestrator/*` protected-path registration — no `orchestrator` symbol in the registry.
- min_score-after-normalization — empirically the merged substrate scores both hybrid and FTS-only docs equally (1.0/1.0) even with a pinned query embedding, so the test's premise that pinned-embedding favors the hybrid doc doesn't hold.

So the substrate-level guards are gated on a followup PR, not on #3180 itself. Flipping the flag now would land red CI.

Separately, the SOUL.md trace test's gate comment was always precise: it depends on **PR 7 (product-tool migration)** to route `memory_write` through `ironclaw_memory`'s `PromptWriteSafetyPolicy` — distinct from #3180. Today the tool path still uses the legacy host workspace, which doesn't consult the substrate. I had conflated the two under one feature flag in my earlier fix; that's now corrected.

## What I changed

1. **Split the feature flag.** `pr3180-ready` keeps gating the substrate-level guards that #3180 was anticipated to deliver but didn't (5 tests). New `pr7-ready` gates the tool-dispatcher → substrate integration test (1 test: the SOUL.md trace). The trace test's gate is now `pr7-ready`. Each follow-up PR (substrate-guard, PR 7) enables its own flag in its merge commit.

2. **Updated each gate's ignore message with the empirical reality.** Anyone enabling the flag in the future now reads exactly what the merged substrate is missing — not just "requires PR #3180" with no further detail.

3. **Added an always-running substrate-level SOUL.md rejection test** (`soul_md_high_risk_write_rejected_at_substrate_for_caller_tier_contract` in `e2e_scope_isolation_safety.rs`) — your alternate suggestion. Drives `RepositoryMemoryBackend` directly with the same scope+payload+policy shape PR 7's tool routing will use, asserts rejection + non-persistence + correct audit class. **This closes the CI gap at the substrate tier today**, without waiting for PR 7. The trace test stays gated as the caller-tier guard for when the tool routing arrives.

The new test is intentionally focused on SOUL.md by name, even though `protected_paths_high_risk_writes_blocked_at_libsql_backend` already covers it in a loop — a regression that selectively drops only SOUL.md from the protected-path registry now fails by name, not silently inside the loop.

## On the Low-priority and residual notes

- `memory_read` exact tool-output check — agreed it's stronger with a tool-result helper. DB read covers the persistence side; tracking as a non-blocking follow-up.
- Z3 "sink succeeds + DB write fails" audit-ordering — still acknowledged as a documented gap on `working_event_sink_admits_bypass_persistence_under_libsql`; needs a fault injector that doesn't exist yet.

## Not enabled in CI

Neither `pr3180-ready` nor `pr7-ready` is enabled in any workflow today. Flipping them on requires the followup landings, per the gate messages. Happy to do the CI enable as part of the substrate-guard PR's merge commit when it lands.

Let me know if you'd like any of the substrate-test failures investigated further (might be a sub-PR scope question) or if the always-running test covers your underlying concern.

henrypark133 (approved):
What looks good:
- The previous blocker is materially addressed: the tool-layer rejection test is now split to `pr7-ready`, and the PR adds always-running substrate-level `SOUL.md` rejection coverage.
- The memory package tests are explicitly run in CI via `cargo test -p ironclaw_memory --features libsql --tests`.
- The new comments now explain why `pr3180-ready` and `pr7-ready` are future gates rather than accidentally skipped tests.

No verified findings.

Low-priority notes:
- PR body is now stale: it still says gated tests “activate automatically once this branch picks up #3180,” but the code now documents that `pr3180-ready` and `pr7-ready` are deliberate future gates. Worth updating before merge so reviewers do not keep re-raising the same concern.

Summary:
- Recommended verdict: Approve.
- Prior feedback status: prior CI-gating blocker addressed on current head `263aab42f`.
- Review coverage: worktree-backed coordinator scan plus memory/testing/CI reviewers; disputed gating concern was separately verified as not a blocker.
- Residual risk: the actual caller-tier `SOUL.md` rejection remains deferred until PR 7 routes `memory_write` through `ironclaw_memory`.

PR #3309, opened 2026-05-06
Make Skills E2E lifecycle deterministic

serrrfirat:
## Summary
- Add an E2E skip/xfail debt inventory with follow-up recommendations
- Link the inventory from the E2E README
- Replace live ClawHub-dependent Skills UI lifecycle checks with deterministic Playwright API mocks
- Remove skip fallbacks from `test_skills.py` search/install/remove paths

## Verification
- `cd tests/e2e && python -m pytest scenarios/test_skills.py -q` → 3 passed
- `cd tests/e2e && python -m pytest scenarios/test_extensions.py::test_activate_mcp_server_success -q` → 1 passed (reran setup-time timeout from first full attempt)
- `cd tests/e2e && python -m pytest scenarios/ -q` → 439 passed, 14 skipped, 3 xfailed in 1613.24s

## Notes
The first full-suite run had one transient Playwright `Page.goto` setup timeout in `test_extensions.py::test_activate_mcp_server_success`; that test passed in isolation and the subsequent full-suite rerun passed cleanly. The remaining skips/xfails are inventoried in `tests/e2e/E2E_DEBT.md`.

### DIFF
--- a/tests/e2e/E2E_DEBT.md
+++ b/tests/e2e/E2E_DEBT.md
@@ -1,0 +1,27 @@
+# E2E skip/xfail debt inventory
+
+This inventory tracks skip/xfail debt found in `tests/e2e/scenarios/` while auditing from `origin/main` for the first E2E debt campaign.
+
+## Summary
+
+| Cluster | Files | Debt type | Determinism | Recommended action |
+| --- | --- | --- | --- | --- |
+| ClawHub skills search/install | `test_skills.py` | Runtime `pytest.skip` when registry/search/install is unavailable or slow | Deterministic with Playwright route mocks | Fix in this campaign: mock the skills API for frontend lifecycle tests so CI never depends on ClawHub availability. |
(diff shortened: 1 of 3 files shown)


zmanian (commented):
Approve with one nit.

Solid scope-appropriate determinism improvement. Mock router is well-scoped (GET `/api/skills`, POST `/api/skills/search`, POST `/api/skills/install`, DELETE `/api/skills/{name}`) with proper fall-through. The dialog handler swap to `page.once("dialog", ...)` is more idiomatic. `E2E_DEBT.md` policy aligns nicely with the project's determinism stance.

**Please address before merge:**

- **`await page.evaluate("loadSkills()")`** (Gemini already flagged this): calling an internal JS function directly couples the test to implementation detail and would mask a regression where the UI fails to auto-refresh after install. Prefer `page.wait_for_function` polling the DOM. If the UI genuinely doesn't auto-refresh the installed-list panel after install, that's a UX bug worth filing rather than papering over in the test.

**Minor / optional:**

- The `page.once("dialog", ...)` registration on the install path appears unused — install doesn't trigger a native dialog, and remove uses the in-DOM `confirm_modal_btn` modal. Either remove it or document why it's there.
- Consider asserting the install POST request body (skill slug) to lock the frontend contract.
- `E2E_DEBT.md` is a nice catalogue — worth noting where the live ClawHub contract coverage now lives (separate integration tier?), since these tests no longer exercise it.

No production code touched, so the usual conventions (no `unwrap`, `crate::`, dispatcher routing, prompt files, `debug!` vs `info!`) are N/A here.

Issue #3279, opened 2026-05-06
[Reborn] Add first TurnCoordinator product-flow acceptance tests


serrrfirat:
Parent TurnCoordinator tracker: #3013  
Reborn E2E gate PR: #3251  
Related multi-tenant admission policy: #3264  
Related outbound egress/subscription policy: #3266  
Related ProductAdapter contract: #3269  
Related MissionService integration: #3278  
Related product-surface migration: #3031  
Related conversation binding / session thread contracts: #3193  
Related SessionThreadService tracker: #3089  
Related transcript/thread storage: #3204  
Related TurnCoordinator public API: #3198  
Related TurnRunner execution model: #3199  
Related turn persistence / active locks: #3202  
Related EventProjectionService: #3093  
Related AgentLoopHost / run profile contracts: #3016, #3107  
Related same-thread follow-up/steering policy: #3236

## Purpose

Define the first Reborn **acceptance test harness** for TurnCoordinator-driven product flows.

This issue splits out #3013 item **18. First acceptance tests** so #3013 can stay focused on architecture sequencing while this issue owns the first deterministic tests that prove product/channel inputs, canonical thread resolution, turn submission, projection/egress behavior, cancellation, restart, and tenant isolation work together.

The point is not to duplicate direct `TurnCoordinator` state-machine tests. Existing direct coordinator tests already cover local semantics such as same-thread busy, different-thread independence, submit idempotency, wrong-scope not-found, blocked/resume, and terminal cancellation release. The first acceptance suite must test through the caller path that actually gates side effects.

## Test boundary

The first acceptance harness should drive a fake but caller-level Reborn pipeline:

```text
Fake ProductAdapter(s)
  -> ProductWorkflow / InboundTurnService facade
      -> ConversationBindingService / SessionThreadService fakes
      -> TurnCoordinator
      -> EventProjectionService / outbound egress fakes
```

Use deterministic fake/in-memory services only in the first slice. Do **not** migrate production Web/Slack/Telegram adapters, v1 `ChannelManager`, PostgreSQL/libSQL persistence, browser approval UX, or default-on product behavior in this issue.

Durability/restart semantics should be proven with a restartable fake store/snapshot first; database parity belongs in later persistence work once #3202 backend schema exists.

## Decisions

1. **Acceptance tests are caller-level, not only direct coordinator tests.**
   - Keep direct `ironclaw_turns` contract tests for local state-machine semantics.
[...]

Issue #3269, opened 2026-05-05
[Reborn] Define ProductAdapter replacement for stale transport PR


serrrfirat:
Parent TurnCoordinator tracker: #3013  
Supersedes / replaces stale transport PR framing: #3099  
Product-surface migration tracker: #3031  
Related conversation binding / session thread contracts: #3193  
Related SessionThreadService tracker: #3089  
Related transcript/thread storage: #3204  
Related outbound egress/subscription policy: #3266  
Related EventProjectionService: #3093  
Related TurnCoordinator API: #3198  
Related AgentLoopHost / run profile contracts: #3016, #3107

## Purpose

Define the Reborn **ProductAdapter** replacement for the stale #3099 transport-adapter PR.

This issue splits out #3013 item **16. Product adapter replacement for #3099** so #3013 can stay focused on TurnCoordinator behavior while this issue owns the product-surface adapter boundary: inbound request normalization into Reborn services, outbound projection rendering/delivery, adapter capabilities, lifecycle/health hooks, idempotency, and first contract tests.

The key correction from #3031: product migration is **not** a generic transport layer wrapping v1 `ChannelManager` as the source of truth. Product adapters sit above Reborn services:

```text
ProductAdapter inbound
  -> ProductWorkflow / InboundTurnService
      -> ConversationBindingService / SessionThreadService
      -> TurnCoordinator or MissionService

ProductAdapter outbound
  <- EventProjectionService / EventStreamManager
  -> protocol rendering + delivery
```

## Background / current evidence

#3099 proposed a large `ironclaw_transport` contract and v1 bridge:

- policy-free transport crate;
- typed ingress/egress and registry lifecycle;
- v1 `Channel` -> `TransportAdapter` bridging;
- opt-in `REBORN_TRANSPORT=true` runtime;
- live `ChannelManager` reply/status routing through the transport registry.

That was useful when the Reborn stack was still forming, but it is now stale against the #3031 layering. Current v1 code illustrates the problem:

- `IncomingMessage` mixes external channel ids, resolved user id, sender id, raw content, attachments, external thread id, channel metadata, and internal flags.
- `ChannelManager` merges live channel streams and directly owns `respond`, `send_status`, and `broadcast` egress.
- `GatewayChannel` maps `StatusUpdate` directly to live SSE `AppEvent`s and can globally broadcast when `user_id` metadata is missing.

[...]

Issue #3266, opened 2026-05-05
[Reborn] Define outbound egress and subscription policy


serrrfirat:
Parent TurnCoordinator tracker: #3013  
Product-surface tracker: #3031  
EventProjectionService tracker: #3093  
Durable event/audit store tracker: #3162  
Related TurnCoordinator API: #3198  
Related conversation binding / session thread contracts: #3193  
Related transcript/thread storage: #3204  
Related same-thread follow-up/steering policy: #3236

## Purpose

Define the Reborn **outbound egress and subscription policy** for TurnCoordinator-driven work: who may fetch or stream canonical thread projections, where external channel pushes go by default, when multi-channel fanout is allowed, and how progress/final-reply delivery differs across WebUI/API subscriptions and external push targets.

This issue splits out #3013 item **15. Outbound egress/subscription policy** so #3013 can stay focused on overall TurnCoordinator behavior while this issue owns projection visibility, notification targets, group-channel privacy, subscription cursor semantics, and the first contract/in-memory implementation slice.

## Background / current evidence

The Reborn events/projections contract separates durable events, projections/read models, and transport streams. Product transports should consume projection outputs and protocol-translate them; they must not own durable business state.

Current v1 Web SSE/WS behavior is useful evidence but not the Reborn contract:

- v1 SSE/WS broadcasts live `AppEvent`s scoped mostly by user;
- v1 `last_event_id` is transport-local and boot-local;
- Reborn requires durable scoped projection cursors, scoped replay, and snapshot/rebase on replay gaps.

## User story

```text
A user sends a Telegram message in canonical thread T.
The run's reply target is the Telegram binding for that message.

Later, the same authorized user opens T in WebUI.
WebUI can fetch and stream the canonical thread projection because the user is
an authorized thread participant.

When the run completes, the final reply is pushed to the original Telegram
reply target by default.
WebUI also sees it via the projection subscription.
Slack or another Telegram chat does not receive an unsolicited push unless the
thread has an explicit notification policy allowing that fanout.
```

## Decisions

1. **Projection visibility is governed by canonical thread access policy.**
   - Authorized thread participants may fetch or stream the canonical thread projection.
   - Projection visibility is not limited to the source/reply binding owner.
   - A binding is a route, not proof that every actor on that route can see the thread.

2. **External push defaults to the turn's reply target only.**
   - By default, final external push goes to the run's `reply_target_binding_ref` / source binding produced by conversation binding/session-thread input.
[...]

Issue #3193, opened 2026-05-02
[Reborn] Define conversation binding and session thread contracts


serrrfirat:
Parent architecture tracker: #2987  
Product-surface tracker: #3031  
Unblocks: #3013 (`TurnCoordinator`)  
Related cutover gates: #3020, #3022, #3032, #3039, #3067

## Status update — 2026-05-08

**Status:** implemented semantic slice on `origin/reborn-integration` (`20e5aa58`). Keep this issue focused on the conversation-binding / inbound-turn contract; do not add ProductAdapter, transcript-store, or egress cutover work here.

Implemented evidence:

- `crates/ironclaw_conversations/` defines `ConversationBindingService`, `SessionThreadService`, `InboundTurnService`, typed external refs, source/reply binding refs, in-memory semantics, and optional libSQL/Postgres durable wrappers.
- `docs/reborn/contracts/conversation-binding.md` records this as **Implemented semantic slice**.
- Verification run on `origin/reborn-integration`: `cargo test -p ironclaw_conversations --test inbound_contract` → **41 passed**.

Requirements split out for follow-up tracking:

| Remaining / downstream requirement | Tracking issue |
| --- | --- |
| Final durable transcript/thread content store, assistant-message writes, context windows, redaction/status/order, and lazy v1 transcript migration | #3204 |
| Product-facing workflow facade and adapter-safe command/turn routing above this contract | #3280, #3269 |
| Concrete browser/API/channel migrations onto ProductAdapter/ProductWorkflow | #3282, #3283, #3285 |
| Outbound egress/subscription behavior using stored reply targets and projection visibility | #3266 |
| Mission/thread explicit-link semantics beyond the generic conversation-binding link | #3278 |
| Cross-service product-flow acceptance coverage after adapters/projections are wired | #3279 |
| Same-thread busy/deferred/follow-up/steering policy beyond preserving message/turn boundaries | #3236 |

Suggested close criterion for this issue: close once the `reborn-integration` semantic slice is accepted/merged as the canonical Reborn conversation-binding contract. Keep downstream integration and storage work in the issues above.

---

## Purpose

Define the Reborn `ConversationBindingService` / `SessionThreadService` contract that sits between product/channel adapters and `TurnCoordinator`.

This issue exists because `TurnCoordinator` should receive canonical Reborn scope and message references, not raw Slack/Telegram/Web/CLI conversation IDs or raw message content.

Target shape:

```text
[...]

PR #3316, opened 2026-05-06
feat(reborn): add ProductAdapter contracts + Telegram v2 tracer-bulle…

### LINKED ISSUE #3285: [Reborn] Migrate external channel adapters onto ProductAdapter contract
Parent: #3031  
Related: #3269, #3266, #3193, #3279, #3032, #3020, #3094

## What to build

Migrate external channel behavior to Reborn ProductAdapters that authenticate protocols, normalize external refs, call ProductWorkflow/MissionService, and render projection-safe outbound messages.

Channels become ProductAdapters, not agent-loop/session owners.

## Initial protocol family

Long-term protocol family:

- Telegram
- Slack
- Discord
- WhatsApp
- Feishu
- Signal
- Web/CLI channel shims where applicable

First real external-channel migration target:

- **Telegram as a clean WASM v2 ProductAdapter**.

Legacy v1 WASM channels and v1 `ChannelManager` paths remain default-off/legacy compatibility paths until each protocol is ported.

## Target path

```text
Telegram webhook / protocol event
  -> Rust host verifies protocol auth and route ownership
  -> Telegram WASM v2 parses/normalizes protocol payload
  -> ProductInboundEnvelope with external actor/conversation/event refs
  -> ProductWorkflow / ConversationBindingService / SessionThreadService
  -> TurnCoordinator or MissionService
  -> EventProjectionService / EventStreamManager
  -> ProductOutboundEnvelope
  -> Telegram WASM v2 render + constrained host http_egress
  -> delivery status sink
```

## Resolved design decisions

- **Use a clean ProductAdapter-native WASM v2 rewrite for Reborn.** Do not build a permanent dual-ABI bridge from legacy WASM `emit_message` / `on_respond` / `IncomingMessage` semantics into Reborn. Existing WASM channels stay on the legacy/default-off path until individually ported.
- **Telegram is the first real v2 protocol tracer bullet.** Telegram covers webhook auth, private/group chat, explicit group triggers, topic/thread refs, stable event ids, attachments, final reply rendering, and protocol egress without Slack relay/OAuth complexity.
- **New Reborn crates/source tree.** Put ProductAdapter/WASM v2 host contracts and Telegram v2 component code in new Reborn product-adapter crates/source trees, not inside `src/channels/wasm` or legacy `channels-src/telegram` except for clearly safe pure helper reuse.
- **ProductWorkflow resolves canonical actor/thread.** Telegram v2 passes structured `external_actor_ref`, `external_conversation_ref`, protocol auth evidence, and external event id. It does not resolve canonical users/threads or directly mutate pairing/session stores.
- **Host-enforced protocol authentication.** The Rust host must verify webhook/protocol auth before any `ProductInboundEnvelope` can be created. WASM may parse payloads and declare auth requirements, but only host verification can produce `ProtocolAuthEvidence::Verified`.
- **Explicit stable idempotency.** Telegram v2 must produce explicit external event ids, using Telegram `update_id` / `message_id` as appropriate. ProductWorkflow/binding/session layers dedupe by tenant + source binding + external event/client request id.
- **ProductWorkflow stages attachments.** Telegram v2 provides bounded `ProductAttachmentInput` descriptors/temporary bytes/handles; ProductWorkflow/SessionThreadService stages durable attachment refs before turn submission. No raw bytes, source URLs, bot tokens, or host-local paths reach TurnCoordinator.
- **Constrained host `http_egress`.** WASM v2 uses a host-mediated `ProtocolHttpEgress` capability for declared Telegram Bot API/file API hosts. The host owns credential handles, policy, timeouts, retries where applicable, redaction, and delivery-status reporting. No raw bot token is exposed to WASM.
- **Webhook HTTP 200 means durable inbound outcome.** Telegram webhook ack returns 200 only after host auth + ref extraction + ProductWorkflow durable accept/dedupe/defer or explicit non-retryable rejection. Transient storage/ProductWorkflow/backpressure failures should return retryable status (5xx/429); auth failures return 401/403; duplicates and non-retryable policy rejections return successful no-op/duplicate ack.
- **Group/supergroup messages require explicit triggers.** Telegram v2 creates inbound actions for private chats normally after auth/binding. In groups/supergroups, create inbound actions only for explicit bot mentions, direct replies to the bot, recognized bot commands, or explicit linked-thread actions. Ambient group messages are successful no-op acks.
- **External conversation refs are chat + optional topic.** Key `ExternalConversationRef` by adapter installation + Telegram `chat_id` + optional `message_thread_id` / forum topic id. Use `update_id` / `message_id` for idempotency and store reply/message id in `ReplyTargetBindingRef` for outbound threading, not as the canonical conversation key.
- **Outbound is projection-derived.** Telegram v2 consumes `ProductOutboundEnvelope`s from projection-safe outputs. External push defaults to final replies to the reply target binding. Progress/typing is only ephemeral or explicitly policy-enabled per #3266.
- **Approval/auth gate UX is deferred to #3094.** First Telegram v2 slice may include fake contract tests for gate projection rendering, but real Telegram approval/auth prompt/resolution UX should wait for ApprovalInteractionService/AuthInteractionService DTOs and policies.
- **Delivery failures are separate status.** Telegram delivery failures do not mark the turn/thread failed or roll back assistant output. Record structured delivery status (`delivered`, `failed`, `retryable/deferred`, `unauthorized/revoked`) via egress delivery sink.
- **Default-off, mutually exclusive exposure.** Telegram v2 is enabled only by explicit Reborn profile/feature flag and owns a mutually exclusive webhook route/installation binding. Default-off mode keeps legacy Telegram unchanged. Do not serve the same Telegram installation through v1 and v2 concurrently except explicit dry-run/shadow tests with side effects disabled.

## First implementation slice

Build the clean Telegram WASM v2 tracer bullet with fake Reborn services and deterministic protocol tests.

Suggested first PR shape:

```text
crates/ironclaw_product_adapters/          # ProductAdapter contracts if not already present
crates/ironclaw_wasm_product_adapters/     # WASM v2 host/runtime boundary or equivalent
channels-src-v2/telegram/                  # Telegram ProductAdapter-native component
tests/product_adapter_telegram_contract.rs # fake ProductWorkflow/projection/egress tests
```

Concrete names can change; boundary semantics matter more than final crate names.

The first slice should prove:

```text
Telegram webhook payload
  -> host auth evidence
  -> Telegram v2 normalized refs/payload
  -> ProductInboundEnvelope
  -> fake ProductWorkflow durable ack/dedupe/defer/reject
  -> fake projection final reply
  -> Telegram v2 render
  -> constrained fake ProtocolHttpEgress
  -> delivery status record
```

Use fake/in-memory ProductWorkflow, ConversationBindingService, SessionThreadService, projection stream, and egress sink. Do not wire production TurnCoordinator or real Telegram network calls in the first PR unless already available behind explicit test gates.

## Acceptance criteria

- [ ] Telegram WASM v2 uses ProductAdapter-native DTOs and does not depend on legacy `IncomingMessage`, `OutgoingResponse`, `StatusUpdate`, v1 `Channel`, or v1 `ChannelManager` semantics.
- [ ] ProductAdapter/WASM v2 host contracts live outside `src/channels` legacy layering.
- [ ] Host verifies Telegram webhook/protocol authentication before constructing any ProductInboundEnvelope.
- [ ] Telegram v2 normalizes external actor, conversation, event, installation, reply-target, and attachment descriptors into structured refs.
- [ ] Telegram v2 does not resolve canonical user/thread ids, write pairing/session stores, or call TurnCoordinator directly.
- [ ] ProductWorkflow/ConversationBinding/SessionThread fake receives external refs and returns durable accept/dedupe/defer/reject outcomes.
- [ ] Stable Telegram external event ids dedupe duplicate webhook deliveries and return the prior durable inbound outcome.
- [ ] Attachments are passed as bounded descriptors/temporary handles and staged by ProductWorkflow/SessionThreadService into durable refs before turn submission.
- [ ] Private chat messages create inbound actions after auth/binding; group/supergroup ambient messages are successful no-op acks unless they explicitly mention/reply/command the bot.
- [ ] `ExternalConversationRef` uses Telegram chat id + optional topic id; reply/message ids are reply-target/idempotency data, not the canonical conversation key.
- [ ] Webhook responses return 200 only after durable inbound outcome; transient failures return retryable errors; auth failures fail closed.
- [ ] Outbound final replies render from projection-derived ProductOutboundEnvelope and target the reply target binding.
- [ ] Telegram outbound uses constrained host `ProtocolHttpEgress` with declared hosts/credential handles; raw bot tokens and arbitrary HTTP authority are not exposed to WASM.
- [ ] Delivery failures record separate egress delivery status and do not mutate canonical transcript/projection/turn success.
- [ ] Real Telegram approval/auth prompt/resolution UX is marked deferred to #3094; fake contract tests cover future gate projection rendering without side effects.
- [ ] Default-off mode keeps legacy Telegram/WASM behavior unchanged; Reborn Telegram v2 requires explicit profile/feature flag and mutually exclusive route binding.
- [ ] Deterministic protocol smoke tests use recorded/fake Telegram payloads and fake Bot API egress for: private message, explicit group mention, topic ref, duplicate `update_id`, attachment descriptor, final reply render, delivery failure status, and redaction sentinels.
- [ ] Redaction/no-exposure tests prove adapter DTOs and errors do not include raw secrets, bot tokens, host paths, raw prompts, raw tool input, provider/runtime internals, or backend diagnostics.

## Non-goals for first slice

- Do not support legacy WASM ABI on the Reborn ProductAdapter path.
- Do not port Slack, Discord, WhatsApp, Feishu, or Signal in this issue's first PR.
- Do not bridge v1 `ChannelManager` as a source of truth.
- Do not wire real Telegram production traffic by default.
- Do not implement real Telegram approval/auth gate UX before #3094.
- Do not implement general multi-channel fanout beyond #3266 reply-target/default final-reply behavior.
- Do not allow arbitrary WASM HTTP/network egress in production ProductAdapter mode.

## Blocked by / coordinates with

- Blocked by #3269 for ProductAdapter contract.
- Blocked by #3280 for ProductWorkflow / InboundTurnService.
- Blocked by #3266 for outbound policy and delivery status semantics.
- Coordinate with #3193 for ConversationBindingService / SessionThreadService external refs, binding, attachment staging, and idempotency.
- Coordinate with #3094 for approval/auth interaction-service gate UX.
- Coordinate with #3032 for no-exposure safeguards.
- Coordinate with #3020 for compatibility-gate evidence and intentional drift from legacy WASM channel behavior.

nickpismenkov:
First-slice landing for #3285 (Migrate external channel adapters onto ProductAdapter contract). Three new crates prove the ProductAdapter boundary end-to-end against fake Reborn services and recorded Telegram payloads, while keeping the legacy v1 Telegram WASM channel default-on and unchanged.

## Summary

- New `ironclaw_product_adapters` crate defines the ProductAdapter contract (#3269 first slice): typed inbound/outbound DTOs, sealed `ProtocolAuthEvidence::Verified`, `ProtocolHttpEgress` with declared hosts + opaque credential handles, `OutboundDeliverySink`, projection cursor stub, in-memory fakes, and architecture/redaction boundary tests.
- New `ironclaw_wasm_product_adapters` host crate provides constant-time webhook auth verifiers (HMAC-SHA256, shared-secret-header), `EgressPolicy` enforcement, `NativeProductAdapterRunner` glue, and the `wit/product_adapter.wit` contract for the eventual wasmtime component-model build.
- New `ironclaw_telegram_v2_adapter` is the Telegram v2 tracer-bullet ProductAdapter: greenfield against the v2 contract, normalizes external refs, gates group/supergroup messages on explicit triggers (mention/reply-to-bot/recognized command), exposes bounded attachment descriptors, and renders projection-derived final replies through constrained egress to `api.telegram.org`.
- Default-off wiring via `REBORN_TELEGRAM_V2_ENABLED=false` (env + `ChannelsConfig`) plus `validate_telegram_v1_v2_exclusivity` startup fail-closed when v1 and v2 both target the same installation. Legacy v1 Telegram path is unmodified.
- 116 new tests cover all 16 acceptance criteria from #3285 plus deterministic protocol smoke fixtures and redaction sentinels.

## Change Type

- [x] New feature (Reborn product-adapter boundary, behind a default-off flag)
- [x] Documentation (two new contract docs + freeze-index update)

## Linked Issue

Implements #3285. Coordinates with #3269 (ProductAdapter contract — first slice landed inline), #3266 (outbound egress policy), #3193 (ConversationBindingService), #3094 (gate UX, deferred), #3032 (no-exposure), #3020 (compatibility-gate), #3279 (TurnCoordinator product-flow tests).

### `crates/ironclaw_product_adapters/`

ProductAdapter contract types and traits:

- DTOs: `ProductInboundEnvelope`, `ProductInboundPayload` (`UserMessage`/`Command`/`ApprovalResolution`/`AuthResolution`/`SubscriptionRequest`/`NoOp`), `ProductInboundAck` (`Accepted`/`DeferredBusy`/`Rejected`/`Duplicate`/`NoOp`), `ProductOutboundEnvelope`, `ProductOutboundPayload` (`FinalReply`/`Progress`/`GatePrompt`/`AuthPrompt`/`ProjectionSnapshot`/`ProjectionUpdate`).
- External refs: `ExternalActorRef`, `ExternalConversationRef` (keyed by space + conversation + optional topic; reply target message id is **not** part of the conversation key), `ExternalEventId`, `ProductAttachmentDescriptor` (no raw bytes, no source URLs, no host paths).
- Auth: `ProtocolAuthEvidence::Verified` is sealed via a crate-private `HostAuthSeal`; only the public `mark_*_verified` helpers in `ironclaw_product_adapters::auth` can mint a `Verified` value. WASM components and downstream adapters cannot fabricate verification.
- Egress: `ProtocolHttpEgress` trait, `DeclaredEgressHost`, `EgressCredentialHandle` (opaque), `OutboundDeliverySink`, `DeliveryStatus` (`Delivered`/`FailedRetryable`/`FailedUnauthorized`/`Deferred`).
- Capabilities: typed `ProductAdapterCapabilities` (inbound messages/commands/attachments, final-reply push, opt-in progress push, opt-in gate push, projection subscription, sync wait, delivery status reporting).
- Traits: `ProductAdapter`, `ProductWorkflow`, `ProjectionStream`.
- In-memory fakes: `FakeProductWorkflow` (programmable outcomes + dedupe by external_event_id), `FakeProtocolHttpEgress` (records calls, validates declared host + credential handle), `FakeOutboundDeliverySink`, `FakeProjectionStream`.
- Boundary tests forbid dependencies on `ironclaw_dispatcher`, `_capabilities`, `_host_runtime`, `_network`, `_secrets`, `_filesystem`, `_wasm`, `_processes`, `_mcp`, `_scripts`, `_runtime_policy`, `_authorization`, `_run_state`, `_approvals`, `_resources`, `_trust`, `_extensions`, `_safety`, `_skills`, `_engine`, `_gateway`, `_tui`, `_memory`, `_events`, `_reborn_event_store`, `_architecture`, and `ironclaw_turns::runner`.
- Redaction tests prove DTOs and errors do not leak raw secrets, host paths, raw prompts, raw tool input, provider/runtime internals, or backend diagnostics. `RedactedString` wraps any value originating from a protocol payload and renders as `<redacted>` in `Debug`/`Display`/`Serialize`.

### `crates/ironclaw_wasm_product_adapters/`

Host runtime glue (currently runtime-free; the wasmtime component-model binary build lands alongside this crate in a follow-up):

- `WebhookAuthVerifier` trait + `HmacWebhookAuth` (Slack-style v0 HMAC-SHA256 with timestamp prefix) and `SharedSecretHeaderAuth` (Telegram-style). Both use `subtle::ConstantTimeEq` to avoid timing oracles.
- `EgressPolicy` — declared-host + credential-handle allowlist enforcement.
- `NativeProductAdapterRunner` — receives webhook headers + body, verifies auth, mints a sealed `Verified` evidence, calls `ProductAdapter::parse_inbound`, forwards the envelope to `ProductWorkflow::accept_inbound`, and returns either `Acknowledged { ack }` or `NoOp`. `RunnerError::is_auth_failure()` and `is_retryable()` map the protocol-status response.
- `wit/product_adapter.wit` — agreed shape of the eventual WASM component contract: exports for `manifest`, `parse-inbound`, `render-outbound`; constrained host imports for `log`, `now-millis`, `http-egress`. No raw filesystem, env, random, or arbitrary HTTP capability is exposed.

### `crates/ironclaw_telegram_v2_adapter/`

Telegram WASM v2 ProductAdapter (greenfield, no v1 channel-type imports):

| Telegram field | Reborn ref |
|---|---|
| `update_id` | `ExternalEventId` (`tg-<installation>-<update_id>`) |
| `message.from.id` | `ExternalActorRef` (kind `telegram_user`) |
| `message.chat.id` | `ExternalConversationRef.conversation_id` |
| `message.message_thread_id` | `ExternalConversationRef.topic_id` |
| `message.message_id` | `ExternalConversationRef.reply_target_message_id` (NOT part of conversation key) |

- Group/supergroup gating: ambient messages → `NoOp` (200 ack, workflow never sees them). Envelopes only when (a) `mention` entity matches configured `bot_username` (case-insensitive, UTF-16-aware offset slicing), or (b) `reply_to_message.from.is_bot && from.id == bot_user_id`, or (c) `bot_command` entity for a recognized command (with `/foo@botname` suffix matching).
- Attachments: bounded `ProductAttachmentDescriptor` with `external_file_id`, `mime_type`, optional `filename`, optional `size_bytes`, and `kind`. No raw bytes, no source URLs, no host paths.
- Outbound: `FinalReply` → `POST api.telegram.org/sendMessage` with `chat_id` + optional `message_thread_id` + optional `reply_to_message_id`. `Progress { Typing }` → `sendChatAction` (only when `ExternalProgressPush` is opted-in per #3266). `GatePrompt`/`AuthPrompt` are deferred to #3094 (no side effects in the first slice). `ProjectionSnapshot`/`Update` are dropped (Telegram doesn't subscribe).
- Egress targets only the declared `api.telegram.org` host. Bot token travels as an opaque `EgressCredentialHandle = "telegram_bot_token"`; the host resolves the underlying secret at request time.

### Default-off wiring

- `REBORN_TELEGRAM_V2_ENABLED=false` (default) keeps legacy v1 Telegram running unchanged through the v1 channel manager.
- New `ironclaw::config::validate_telegram_v1_v2_exclusivity(v1_active, v2_active) -> Result<(), ConfigError>` fails closed at startup when both paths are configured for the same telegram installation.
- `tests/telegram_v2_default_off_integration.rs` is a caller-tier integration test (not just a helper test) that drives the validator + asserts `Config::for_testing(...).channels.reborn_telegram_v2_enabled == false`.

### Docs

- `docs/reborn/contracts/product-adapters.md` — frozen contract: layering, invariants, capability grid, ack-to-status mapping.
- `docs/reborn/contracts/telegram-v2.md` — Telegram-specific normalization, group gating, attachment policy, idempotency, and AC test pointer.
- `_contract-freeze-index.md` updated.
- `.env.example` documents `REBORN_TELEGRAM_V2_ENABLED`.

## Acceptance criteria coverage (#3285)

`crates/ironclaw_telegram_v2_adapter/tests/product_adapter_telegram_contract.rs` has one named test per AC bullet:

| AC | Tests |
|---|---|
| 1 — only v2 DTOs | `ac1_telegram_v2_does_not_import_v1_channel_types` |
| 2 — outside `src/channels` | `ac2_product_adapter_contracts_live_outside_src_channels` |
| 3 — host verifies auth before envelope | `ac3_runner_blocks_envelope_construction_on_bad_secret`, `_on_missing_secret`, `ac3_adapter_refuses_unverified_evidence_directly` |
| 4 — refs normalized | `ac4_parse_normalizes_all_refs` |
| 5 — no canonical-state writes | `ac5_adapter_does_not_import_turn_coordinator`, `ac5_adapter_path_only_invokes_workflow_facade` |
| 6 — workflow durable outcomes | `ac6_workflow_returns_each_durable_outcome_kind` |
| 7 — `update_id` dedupe | `ac7_duplicate_update_id_returns_prior_outcome_no_double_submit` |
| 8 — bounded attachment descriptors | `ac8_attachments_have_no_raw_bytes_or_source_urls` |
| 9 — group gating | `ac9_private_chat_creates_inbound`, `_group_ambient_is_noop_ack`, `_group_explicit_mention_creates_inbound`, `_group_command_creates_inbound` |
| 10 — conversation key shape | `ac10_conversation_key_uses_chat_and_topic_not_message_id` |
| 11 — webhook ack semantics | `ac11_durable_outcome_classification_for_each_path`, `ac11_duplicate_returns_200_no_op_ack` |
| 12 — projection-derived outbound | `ac12_final_reply_renders_to_reply_target_binding` |
| 13 — constrained egress | `ac13_egress_to_undeclared_host_is_blocked`, `_telegram_only_declares_telegram_api`, `_egress_request_carries_credential_handle_not_token` |
| 14 — separate delivery status | `ac14_delivery_failure_records_status_separately`, `_does_not_mutate_canonical_workflow_state` |
| 15 — gate UX deferred to #3094 | `ac15_gate_prompt_envelope_is_no_op_egress_in_first_slice` |
| 16 — default-off | `ac16_default_off_marker_present_in_workspace_root_config` + `tests/telegram_v2_default_off_integration.rs` |
| smoke / redaction | `smoke_recorded_payloads_match_expected_outcomes`, `redaction_sentinels_in_envelope_debug`, `_in_error_display`, `telegram_default_capabilities_pin_first_slice_behavior`, `telegram_does_not_consume_projection_subscriptions`, `projection_stream_can_drive_telegram_via_render_outbound_chain` |

## Validation

- [x] `cargo build` — passes
- [x] `cargo test -p ironclaw_product_adapters -p ironclaw_wasm_product_adapters -p ironclaw_telegram_v2_adapter` — 112 tests pass
- [x] `cargo test --test telegram_v2_default_off_integration` — 5 tests pass
- [x] `cargo test -p ironclaw config::channels::telegram_v2_tests --lib` — 4 tests pass
- [ ] `cargo test --features integration` not run for this PR — no DB-backed code paths added

## Security Impact

- New crates introduce a constrained boundary for protocol authentication. Webhook signatures are verified in constant time (`subtle::ConstantTimeEq`).
- `ProtocolAuthEvidence::Verified` is sealed: only crate-internal helpers can construct it. WASM components (when wired up) cannot mint verification.
- `ProtocolHttpEgress` enforces declared hosts + credential-handle allowlist. Bot tokens never reach adapter code or WASM linear memory; the host resolves them at request time.
- DTOs and errors carry `RedactedString` for any value originating from a protocol payload, secret store, or backend error text. Redaction is asserted by tests across `Debug`, `Display`, and `Serialize`.
- Default-off: legacy v1 Telegram path is byte-for-byte unchanged. The mutual-exclusion validator fails closed at startup if both paths target the same installation.

## Database Impact

None. No migrations, no schema changes, no database access in any of the three new crates.

## Blast Radius

- New crates only add code; they are not wired into the v1 channel manager or any production webhook router. The single touch point in legacy code is the additive `reborn_telegram_v2_enabled` field on `ChannelsConfig` (default `false` everywhere it is constructed) plus the new public `validate_telegram_v1_v2_exclusivity` symbol.
- `src/channels/wasm/telegram_host_config.rs` and `channels-src/telegram/` are not modified — the v1 path is byte-for-byte unchanged.
- Three workspace `Cargo.toml` lines added (members list).

## Rollback Plan

Revert the single commit. The feature flag is default-off, so even with the code present, no production behavior changes until an operator explicitly sets `REBORN_TELEGRAM_V2_ENABLED=true` and wires the v2 webhook route.

## Review Follow-Through

- The wasmtime component-model binary build (compile `channels-src-v2/telegram/` to a `.wasm`, instantiate via wasmtime in `ironclaw_wasm_product_adapters`) is intentionally deferred — the WIT contract is in `wit/product_adapter.wit` and the Rust-native adapter implementation in `ironclaw_telegram_v2_adapter` is the same logic that will move into the component.
- Production wiring of the v2 webhook route (an axum handler driving `NativeProductAdapterRunner`) is a follow-up. The default-off flag is the cutover seam.
- Slack/Discord/WhatsApp/Feishu/Signal v2 adapters are explicit non-goals for this slice.
- Real `ProductWorkflow` / `ConversationBindingService` / `SessionThreadService` implementations land with #3193/#3280 and replace the in-memory fakes used here.

---

**Review track**: B (new feature, default-off, contract-freeze landing for #3269/#3285)

### DIFF
--- a/crates/ironclaw_telegram_v2_adapter/src/adapter.rs
+++ b/crates/ironclaw_telegram_v2_adapter/src/adapter.rs
@@ -1,0 +1,309 @@
+//! Telegram v2 ProductAdapter implementation.
+
+use async_trait::async_trait;
+use ironclaw_product_adapters::{
+    AdapterInstallationId, DeclaredEgressHost, EgressCredentialHandle, ProductAdapter,
+    ProductAdapterCapabilities, ProductAdapterError, ProductAdapterId, ProductCapabilityFlag,
+    ProductInboundEnvelope, ProductOutboundEnvelope, ProductOutboundPayload, ProductSurfaceKind,
+    ProtocolAuthEvidence, ProtocolHttpEgress, redaction::RedactedString,
+};
+
+use crate::payload::{
+    GroupTriggerPolicy, TELEGRAM_API_HOST, TelegramParsedInbound, parse_telegram_update,
+};
+use crate::render::{render_final_reply, render_progress_typing};
+
+/// Configuration for a Telegram v2 adapter installation.
+#[derive(Debug, Clone)]
+pub struct TelegramV2AdapterConfig {
+    pub adapter_id: ProductAdapterId,
+    pub installation_id: AdapterInstallationId,
+    pub group_trigger_policy: GroupTriggerPolicy,
+    /// Credential handle (resolved by the host to the bot token at request
+    /// time) used for egress to api.telegram.org.
+    pub egress_credential_handle: EgressCredentialHandle,
+    /// If true, the adapter advertises `ExternalProgressPush` and renders
+    /// typing indicators on outbound `Progress` envelopes. Default: false
+    /// (#3266 progress-opt-in policy).
+    pub progress_push_enabled: bool,
+}
+
+pub struct TelegramV2Adapter {
+    config: TelegramV2AdapterConfig,
+    capabilities: ProductAdapterCapabilities,
+}
+
+impl TelegramV2Adapter {
+    pub fn new(config: TelegramV2AdapterConfig) -> Self {
+        let mut capabilities = ProductAdapterCapabilities::external_channel_default();
+        if config.progress_push_enabled {
+            capabilities = capabilities.with(ProductCapabilityFlag::ExternalProgressPush);
+        }
+        Self {
+            config,
+            capabilities,
+        }
+    }
+
+    pub fn config(&self) -> &TelegramV2AdapterConfig {
+        &self.config
+    }
+}
+
+/// Egress hosts that any Telegram v2 installation may target.
+pub fn telegram_declared_egress_hosts() -> Vec<DeclaredEgressHost> {
+    vec![DeclaredEgressHost::new(TELEGRAM_API_HOST).expect("static host valid")]
+}
+
+#[async_trait]
+impl ProductAdapter for TelegramV2Adapter {
+    fn adapter_id(&self) -> &ProductAdapterId {
+        &self.config.adapter_id
+    }
+
+    fn installation_id(&self) -> &AdapterInstallationId {
+        &self.config.installation_id
+    }
+
+    fn surface_kind(&self) -> ProductSurfaceKind {
+        ProductSurfaceKind::ExternalChannel
+    }
+
+    fn capabilities(&self) -> &ProductAdapterCapabilities {
+        &self.capabilities
+    }
+
+    fn parse_inbound(
+        &self,
+        raw_payload: &[u8],
+        auth_evidence: ProtocolAuthEvidence,
+    ) -> Result<Option<ProductInboundEnvelope>, ProductAdapterError> {
+        let parsed = parse_telegram_update(
+            raw_payload,
+            auth_evidence,
+            &self.config.adapter_id,
+            &self.config.installation_id,
+            &self.config.group_trigger_policy,
+        )
+        .map_err(|err| match err {
+            crate::payload::PayloadParseError::UnauthenticatedPayload => {
+                ProductAdapterError::Authentication(
+                    ironclaw_product_adapters::ProtocolAuthFailure::Other {
+                        detail: RedactedString::new(
+                            "telegram parse_inbound called with unverified evidence",
+                        ),
+                    },
+                )
+            }
+            crate::payload::PayloadParseError::InvalidJson { reason } => {
+                ProductAdapterError::MalformedInboundPayload { reason }
+            }
+            crate::payload::PayloadParseError::MissingUpdateId => {
+                ProductAdapterError::MalformedInboundPayload {
+                    reason: "telegram update missing update_id".into(),
+                }
+            }
+            crate::payload::PayloadParseError::InvalidExternalRef { kind, reason } => {
+                ProductAdapterError::InvalidIdentifier { kind, reason }
+            }
+        })?;
+        match parsed {
+            TelegramParsedInbound::Envelope(envelope) => Ok(Some(*envelope)),
+            TelegramParsedInbound::NoOp => Ok(None),
+        }
+    }
+
+    async fn render_outbound(
+        &self,
+        envelope: ProductOutboundEnvelope,
+        egress: &dyn ProtocolHttpEgress,
+    ) -> Result<(), ProductAdapterError> {
+        let request = match envelope.payload {
+            ProductOutboundPayload::FinalReply(view) => render_final_reply(
+                &envelope.target,
+                &view,
+                self.config.egress_credential_handle.clone(),
+            )
+            .map_err(|err| ProductAdapterError::Internal {
+                detail: RedactedString::new(err.to_string()),
+            })?,
+            ProductOutboundPayload::Progress(view) => {
+                if !self
+                    .capabilities
+                    .contains(ProductCapabilityFlag::ExternalProgressPush)
+                {
+                    // Progress not advertised; drop silently. The workflow
+                    // would not normally route a Progress envelope to a
+                    // capability-ungated adapter, but this defends against a
+                    // misrouted envelope reaching us.
+                    return Ok(());
+                }
+                let Some(req) = render_progress_typing(
+                    &envelope.target,
+                    &view,
+                    self.config.egress_credential_handle.clone(),
+                )
+                .map_err(|err| ProductAdapterError::Internal {
+                    detail: RedactedString::new(err.to_string()),
+                })?
+                else {
+                    return Ok(());
+                };
+                req
+            }
+            ProductOutboundPayload::GatePrompt(_) | ProductOutboundPayload::AuthPrompt(_) => {
+                // Deferred to #3094. The workflow renders a placeholder body
+                // via this branch in fake contract tests; real production
+                // flows do not produce gate envelopes for Telegram yet.
+                return Ok(());
+            }
+            ProductOutboundPayload::ProjectionSnapshot(_)
+            | ProductOutboundPayload::ProjectionUpdate(_) => {
+                // Telegram never consumes projection subscriptions; the
+                // workflow should not route these to a Telegram installation.
+                // Return Ok to keep delivery best-effort.
+                return Ok(());
+            }
+        };
+
+        let response =
+            egress
+                .send(request)
+                .await
+                .map_err(|err| ProductAdapterError::EgressDenied {
+                    reason: err.to_string(),
+                })?;
+        if !(200..300).contains(&response.status) {
+            return Err(ProductAdapterError::EgressDenied {
+                reason: format!("telegram bot api returned status {}", response.status),
+            });
+        }
+        Ok(())
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use ironclaw_product_adapters::FakeProtocolHttpEgress;
+
+    fn config(progress: bool) -> TelegramV2AdapterConfig {
+        TelegramV2AdapterConfig {
+            adapter_id: ProductAdapterId::new("telegram_v2").expect("valid"),
+            installation_id: AdapterInstallationId::new("install_alpha").expect("valid"),
+            group_trigger_policy: GroupTriggerPolicy {
+                bot_username: "ironclaw_bot".into(),
+                bot_user_id: 9000,
+                recognized_commands: vec!["start".into()],
+            },
+            egress_credential_handle: EgressCredentialHandle::new("telegram_bot_token")
+                .expect("valid"),
+            progress_push_enabled: progress,
+        }
+    }
+
+    #[test]
+    fn capabilities_default_excludes_progress() {
+        let adapter = TelegramV2Adapter::new(config(false));
+        assert!(
+            !adapter
+                .capabilities()
+                .contains(ProductCapabilityFlag::ExternalProgressPush)
+        );
+        assert!(
+            adapter
+                .capabilities()
+                .contains(ProductCapabilityFlag::ExternalFinalReplyPush)
+        );
+    }
+
+    #[test]
+    fn capabilities_with_progress_opt_in_includes_progress_push() {
+        let adapter = TelegramV2Adapter::new(config(true));
+        assert!(
+            adapter
+                .capabilities()
+                .contains(ProductCapabilityFlag::ExternalProgressPush)
+        );
+    }
+
+    #[test]
+    fn declared_hosts_only_telegram_api() {
+        let hosts = telegram_declared_egress_hosts();
+        assert_eq!(hosts.len(), 1);
+        assert_eq!(hosts[0].as_str(), "api.telegram.org");
+    }
+
+    #[test]
+    fn parse_inbound_refuses_unverified_evidence() {
+        let adapter = TelegramV2Adapter::new(config(false));
+        let unverified = ProtocolAuthEvidence::Failed {
+            failure: ironclaw_product_adapters::ProtocolAuthFailure::SharedSecretMismatch,
+        };
+        let err = adapter
+            .parse_inbound(b"{\"update_id\":1}", unverified)
+            .expect_err("must fail");
+        assert!(matches!(err, ProductAdapterError::Authentication(_)));
+    }
+
+    #[tokio::test]
+    async fn render_outbound_final_reply_uses_constrained_egress() {
+        let adapter = TelegramV2Adapter::new(config(false));
+        let egress = FakeProtocolHttpEgress::new(["api.telegram.org".to_string()]);
+        egress.allow_credential_handle("telegram_bot_token");
+        let envelope = ProductOutboundEnvelope {
+            adapter_id: adapter.adapter_id().clone(),
+            installation_id: adapter.installation_id().clone(),
+            target: crate::render::build_reply_target_binding(-100, Some(7), Some(42)),
+            projection_cursor: None,
+            payload: ProductOutboundPayload::FinalReply(
+                ironclaw_product_adapters::FinalReplyView {
+                    turn_run_id: ironclaw_turns::TurnRunId::new(),
+                    text: "hi".into(),
+                    generated_at: chrono::Utc::now(),
+                },
+            ),
+            delivery_attempt_id: uuid::Uuid::new_v4(),
+        };
+        adapter
+            .render_outbound(envelope, &egress)
+            .await
+            .expect("render ok");
+        let calls = egress.calls();
+        assert_eq!(calls.len(), 1);
+        assert_eq!(calls[0].host, "api.telegram.org");
+        assert_eq!(calls[0].method, "POST");
+        assert_eq!(calls[0].path, "/sendMessage");
+        assert_eq!(
+            calls[0].credential_handle.as_deref(),
+            Some("telegram_bot_token")
+        );
+    }
+
+    #[tokio::test]
+    async fn render_outbound_progress_skipped_when_capability_off() {
+        let adapter = TelegramV2Adapter::new(config(false));
+        let egress = FakeProtocolHttpEgress::new(["api.telegram.org".to_string()]);
+        egress.allow_credential_handle("telegram_bot_token");
+        let envelope = ProductOutboundEnvelope {
+            adapter_id: adapter.adapter_id().clone(),
+            installation_id: adapter.installation_id().clone(),
+            target: crate::render::build_reply_target_binding(-100, None, None),
+            projection_cursor: None,
+            payload: ProductOutboundPayload::Progress(
+                ironclaw_product_adapters::ProgressUpdateView {
+                    turn_run_id: ironclaw_turns::TurnRunId::new(),
+                    kind: ironclaw_product_adapters::ProgressKind::Typing,
+                    generated_at: chrono::Utc::now(),
+                },
+            ),
+            delivery_attempt_id: uuid::Uuid::new_v4(),
+        };
+        adapter
+            .render_outbound(envelope, &egress)
+            .await
+            .expect("ok");
+        // Progress is not advertised -> egress NOT called.
+        assert!(egress.calls().is_empty());
+    }
+}
--- a/crates/ironclaw_telegram_v2_adapter/src/adapter.rs
+++ b/crates/ironclaw_telegram_v2_adapter/src/adapter.rs
@@ -5,7 +5,7 @@ use ironclaw_product_adapters::{
     AdapterInstallationId, DeclaredEgressHost, EgressCredentialHandle, ProductAdapter,
     ProductAdapterCapabilities, ProductAdapterError, ProductAdapterId, ProductCapabilityFlag,
     ProductInboundEnvelope, ProductOutboundEnvelope, ProductOutboundPayload, ProductSurfaceKind,
-    ProtocolAuthEvidence, ProtocolHttpEgress, redaction::RedactedString,
+    ProtocolAuthEvidence, ProtocolHttpEgress, ProtocolHttpEgressError, redaction::RedactedString,
 };
 
 use crate::payload::{
@@ -52,7 +52,7 @@ impl TelegramV2Adapter {
 
 /// Egress hosts that any Telegram v2 installation may target.
 pub fn telegram_declared_egress_hosts() -> Vec<DeclaredEgressHost> {
-    vec![DeclaredEgressHost::new(TELEGRAM_API_HOST).expect("static host valid")]
+    vec![DeclaredEgressHost::new(TELEGRAM_API_HOST).expect("static host valid")] // safety: compile-time const "api.telegram.org" satisfies DeclaredEgressHost validator
 }
 
 #[async_trait]
@@ -166,19 +166,38 @@ impl ProductAdapter for TelegramV2Adapter {
             }
         };
 
-        let response =
-            egress
-                .send(request)
-                .await
-                .map_err(|err| ProductAdapterError::EgressDenied {
-                    reason: err.to_string(),
-                })?;
+        let response = egress.send(request).await.map_err(map_egress_error)?;
         if !(200..300).contains(&response.status) {
-            return Err(ProductAdapterError::EgressDenied {
-                reason: format!("telegram bot api returned status {}", response.status),
-            });
+            let reason = format!("telegram bot api returned status {}", response.status);
+            // Group transient HTTP outcomes (5xx, 429) into the retryable
+            // bucket so the host glue can re-deliver. 4xx (except 429) is
+            // a deterministic policy-denied result and should NOT be
+            // retried.
+            if response.status >= 500 || response.status == 429 {
+                return Err(ProductAdapterError::WorkflowTransient { reason });
+            }
+            return Err(ProductAdapterError::EgressDenied { reason });
         }
         Ok(())
+    }
+}
+
+/// Map a `ProtocolHttpEgressError` to either a retryable
+/// `WorkflowTransient` or a non-retryable `EgressDenied`. Network /
+/// timeout / leak-detector failures are treated as transient.
+fn map_egress_error(err: ProtocolHttpEgressError) -> ProductAdapterError {
+    match err {
+        ProtocolHttpEgressError::Timeout
+        | ProtocolHttpEgressError::Network(_)
+        | ProtocolHttpEgressError::LeakDetected => ProductAdapterError::WorkflowTransient {
+            reason: err.to_string(),
+        },
+        ProtocolHttpEgressError::UndeclaredHost { .. }
+        | ProtocolHttpEgressError::UnknownCredentialHandle { .. }
+        | ProtocolHttpEgressError::UnauthorizedCredentialHandle { .. }
+        | ProtocolHttpEgressError::PolicyDenied { .. } => ProductAdapterError::EgressDenied {
+            reason: err.to_string(),
+        },
     }
 }
 
--- a/crates/ironclaw_telegram_v2_adapter/src/payload.rs
+++ b/crates/ironclaw_telegram_v2_adapter/src/payload.rs
@@ -1,0 +1,1007 @@
+//! Telegram Bot API payload normalization.
+//!
+//! Inputs are raw webhook update bytes. Outputs are structured envelopes the
+//! adapter can hand to the workflow facade — or `None` when the update should
+//! produce a successful no-op acknowledgement (ambient group messages,
+//! channel posts, edited-message kinds we don't act on, etc.).
+
+use chrono::{DateTime, TimeZone, Utc};
+use ironclaw_product_adapters::{
+    AdapterInstallationId, ExternalActorRef, ExternalConversationRef, ExternalEventId,
+    InboundCommandPayload, ProductAdapterId, ProductAttachmentDescriptor, ProductAttachmentKind,
+    ProductInboundEnvelope, ProductInboundPayload, ProductTriggerReason, ProtocolAuthEvidence,
+    UserMessagePayload,
+};
+use serde::Deserialize;
+use thiserror::Error;
+
+pub const TELEGRAM_API_HOST: &str = "api.telegram.org";
+pub const TELEGRAM_FILE_API_HOST: &str = "api.telegram.org";
+pub const TELEGRAM_USER_ACTOR_KIND: &str = "telegram_user";
+
+/// What an adapter installation is configured to recognize as an explicit
+/// trigger inside group/supergroup chats.
+///
+/// Telegram private/direct chats do not require any trigger — every message
+/// is forwarded. In groups/supergroups the adapter forwards a message ONLY
+/// when one of these triggers fires, per #3285's "explicit triggers" rule.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct GroupTriggerPolicy {
+    /// Configured bot username (without leading `@`). Must be ASCII
+    /// alphanumeric or `_`. The adapter compares mention entities against
+    /// this value case-insensitively.
+    pub bot_username: String,
+    /// Stable bot user id used to detect "reply to a message authored by the
+    /// bot" triggers.
+    pub bot_user_id: i64,
+    /// Recognized bot commands (without leading `/`). When a message starts
+    /// with `/foo` or `/foo@botusername`, it is an explicit trigger.
+    pub recognized_commands: Vec<String>,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum TelegramParsedInbound {
+    /// Adapter produces an envelope and forwards to the workflow. Boxed
+    /// because the envelope is much larger than the `NoOp` variant.
+    Envelope(Box<ProductInboundEnvelope>),
+    /// Successful no-op (ambient group message, edited message we ignore,
+    /// channel post, ...). Webhook responds 200 OK without invoking the
+    /// workflow.
+    NoOp,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Error)]
+pub enum PayloadParseError {
+    #[error("invalid Telegram update JSON: {reason}")]
+    InvalidJson { reason: String },
+    #[error("Telegram update missing update_id")]
+    MissingUpdateId,
+    #[error("invalid external reference: {kind}: {reason}")]
+    InvalidExternalRef { kind: &'static str, reason: String },
+    #[error(
+        "auth evidence is not Verified — host MUST verify the webhook before \
+         calling parse_telegram_update"
+    )]
+    UnauthenticatedPayload,
+}
+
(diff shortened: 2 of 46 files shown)


zmanian:
## Code Review

### Overview
7,360-line first-slice landing of the ProductAdapter contract for #3285. Three new crates: `ironclaw_product_adapters` (DTOs + traits + sealed auth evidence), `ironclaw_wasm_product_adapters` (host glue: webhook verifiers, egress policy, native runner — no wasmtime yet), `ironclaw_telegram_v2_adapter` (greenfield Telegram v2). Default-off behind `REBORN_TELEGRAM_V2_ENABLED=false`; v1 path byte-for-byte unchanged.

### Strengths
- **`ProtocolAuthEvidence::Verified` sealing is layered correctly.** `HostAuthSeal` has a `pub(crate)` constructor; the custom `Deserialize` impl only accepts `Failed` wire payloads. Tests pin the asymmetry: `verified_evidence_in_memory_serializes_but_not_back` proves a `Verified` value can be logged/audited but cannot round-trip back. Closes the `#[serde(default)]` re-mint loophole the PR body calls out.
- **Constant-time webhook auth.** `subtle::ConstantTimeEq` on both verifiers; replay-window check happens before HMAC compute with a symmetric `|now - ts| > max_age` test. Boundary-inclusive (`drift == max_age` passes; `drift == max_age + 1` fails) and pinned by tests.
- **UTF-16 entity slicing for Telegram is correctly implemented** with explicit handling for zero-length entities at offset 0, end-of-string, past-end-returns-None, and multibyte (🦀 case in unit tests).
- **Default-off validator exercised at the caller tier** in `tests/telegram_v2_default_off_integration.rs`, including `Config::for_testing()`. Matches the "Test Through the Caller" rule.
- `#![forbid(unsafe_code)]` on the wasm host glue.

### Suggestions

**Minor:**

1. **Home-rolled `mod hex` in `auth_verifier.rs`** (~6 lines). The `hex` crate is already a transitive dep of HMAC tooling. Either drop the inline module and use `hex::encode`, or skip hex-encoding entirely and compare raw bytes (`mac.finalize().into_bytes().ct_eq(decoded_signature_bytes)`) to avoid the string allocation.

2. **`HmacWebhookAuth` fields are `pub`** (`pub signing_secret: Vec<u8>`, etc.). The struct deliberately omits `Debug` so direct formatting won't compile, but a downstream caller composing this into a `#[derive(Debug)]` struct would silently leak the signing secret. Make fields private with constructor + builder, or wrap `signing_secret` in a `RedactedBytes` mirroring `RedactedString`.

3. **`now_unix_seconds() as i64` cast** in `HmacWebhookAuth::verify` — for any clock value before year 2262 this is fine, but `FixedClock(u64::MAX)` test value would silently wrap. Either keep arithmetic in `u64` with `now.abs_diff(timestamp)`, or document the year-2262 ceiling.

4. **`WebhookAuth::mint_evidence` re-clones header names** that already live on the verifier. Minor allocation churn per request; not load-bearing.

### Non-issues (intentional, called out in PR body)
- No wasmtime instantiation — WIT contract is the agreed shape, not enforced. Native runner enforces the same auth/dedupe/facade-only contract.
- Real `ProductWorkflow` and `ConversationBindingService` are fakes; deferred to #3193/#3280.
- Gate UX deferred to #3094; AC15 test pins no-op egress.
- No production wiring; default-off flag is the cutover seam.

### Risk
- **Blast radius**: very low — three new crates, one additive `ChannelsConfig` field, one new public symbol. v1 path untouched.
- **Security**: auth seal + constant-time + sealed serde + redaction is the right shape, and tested at the right tier.
- **Correctness**: 116 named tests against 16 ACs; UTF-16 edge-case coverage. Real risk surface is the fakes-only execution paths; the real workflow's dedupe story isn't on this branch.

### Recommendation
**Approve.** The four minor suggestions above are non-blocking polish — drop the home-rolled hex module and tighten `signing_secret` field visibility in a follow-up. Everything else is acceptable as documented tracer-bullet tradeoffs.

serrrfirat:
This is like 3 PRs in one. chop it up, agents wont be able to review this correctly.

crates/ironclaw_wasm_product_adapters/src/runner.rs
```diff
@@ -0,0 +1,140 @@
...
+        headers: &http::HeaderMap,
+        body: &[u8],
+    ) -> Result<WebhookProcessOutcome, RunnerError> {
+        let evidence = match self.auth.verify(headers, body) {
+            VerificationOutcome::Verified { subject } => self.auth.mint_evidence(subject),
+            VerificationOutcome::Failed { failure } => {
+                return Err(RunnerError::AuthenticationFailed { failure });
+            }
+        };
+        let Some(envelope) = self.adapter.parse_inbound(body, evidence)? else {
```
serrrfirat, line 134: **High Severity**

The runner verifies the webhook, but it then trusts authority fields supplied by the adapter-produced envelope.

`process_webhook` mints trusted evidence, passes it into `parse_inbound`, and forwards the returned `ProductInboundEnvelope` directly to `accept_inbound`. Since the envelope carries public `adapter_id`, `installation_id`, and `auth_claim` fields (and Rust adapters can also call the public `mark_*_verified` helpers), a buggy or malicious adapter can return an envelope for a different installation/subject than the one the host just verified. That breaks the documented “adapters cannot fabricate verification” boundary and can confuse downstream binding/dedupe/auth decisions.

Please have trusted host/runner code inject or overwrite the authority fields before workflow submission, or change adapters to return only payload/protocol refs and let the runner build the final envelope. Add a malicious fake-adapter test that returns mismatched install/claim data and assert the runner rejects it before `accept_inbound`.

src/config/channels.rs
```diff
@@ -448,10 +456,68 @@ impl ChannelsConfig {
...
+/// installation. The host startup glue calls this BEFORE registering any
+/// telegram webhook route — failing fast prevents two paths from receiving
+/// the same update and double-submitting the same canonical message.
+///
+/// `v1_active`: true when the legacy v1 Telegram WASM channel is configured
+/// for startup (i.e. `configured_wasm_channels` contains `telegram` AND
+/// `wasm_channels_enabled` is true).
+///
+/// `v2_active`: value of `reborn_telegram_v2_enabled`.
+pub fn validate_telegram_v1_v2_exclusivity(
```
serrrfirat, line 476: **Medium Severity**

The Telegram v1/v2 exclusivity guard is defined but never invoked by production startup/config resolution.

The docs/tests say startup fails closed when legacy Telegram v1 and Telegram v2 are both active, but the real config/startup path only parses `REBORN_TELEGRAM_V2_ENABLED`; `validate_telegram_v1_v2_exclusivity` is exercised directly in tests and is not called before `setup_wasm_channels`/route registration. A deployment with `WASM_CHANNELS_ENABLED=true`, a configured legacy `telegram` channel, and `REBORN_TELEGRAM_V2_ENABLED=true` therefore proceeds instead of failing at startup.

Please call this validator from the real config or startup path after computing whether v1 and v2 are active, before registering Telegram/WASM webhook handling. Add a caller-level test that drives actual config/startup resolution with both paths enabled and expects `ConfigError`.

crates/ironclaw_wasm_product_adapters/src/auth_verifier.rs
```diff
@@ -0,0 +1,440 @@
...
+        // far-future timestamps; both are forgery attempts. The window is
+        // symmetric: |now - ts| > max_age_secs => fail.
+        let Ok(timestamp_secs) = timestamp_str.parse::<i64>() else {
+            return VerificationOutcome::Failed {
+                failure: ProtocolAuthFailure::Malformed,
+            };
+        };
+        let now_secs = self.clock.now_unix_seconds() as i64;
+        let max_age = self.max_age_secs as i64;
+        let drift = (now_secs - timestamp_secs).abs();
```
serrrfirat, line 156: **Medium Severity**

The HMAC replay-window calculation can overflow before returning an auth failure.

`timestamp_str` is parsed as `i64`, then the verifier computes `(now_secs - timestamp_secs).abs()`. A hostile timestamp such as `-9223372036854775808` overflows the subtraction in overflow-check builds, so an unauthenticated request can panic the verifier instead of receiving a fail-closed auth result.

Please reject negative timestamps (or parse as `u64`) and use checked arithmetic / `abs_diff` for the drift calculation. Add tests for `i64::MIN`, negative timestamps, and very large future timestamps.

crates/ironclaw_wasm_product_adapters/src/auth_verifier.rs
```diff
@@ -0,0 +1,440 @@
...
+    fn verify(&self, headers: &http::HeaderMap, _body: &[u8]) -> VerificationOutcome {
+        let Some(received) = headers
+            .get(self.header_name.as_str())
+            .and_then(|v| v.to_str().ok())
+        else {
+            return VerificationOutcome::Failed {
+                failure: ProtocolAuthFailure::Missing,
+            };
+        };
+        if !bool::from(received.as_bytes().ct_eq(self.expected_secret.as_bytes())) {
```
serrrfirat, line 209: **Medium Severity**

Webhook verifiers accept empty secrets, which makes a missing secret fail open.

`SharedSecretHeaderAuth` has public fields and no empty-secret guard, so `expected_secret == ""` verifies when the request supplies an empty `X-Telegram-Bot-Api-Secret-Token` value. `HmacWebhookAuth::new` likewise accepts an empty signing key; HMAC with an empty key is computable by anyone. If host wiring builds these verifiers from a missing/empty env/config value, authentication becomes trivially forgeable.

Please hide fields behind validated constructors (or defensively fail in `verify`) and reject empty header names, subjects, shared secrets, and HMAC keys at startup. Add tests proving empty shared-secret and empty HMAC-key configurations reject even when the request matches the empty value.

crates/ironclaw_telegram_v2_adapter/src/payload.rs
```diff
@@ -0,0 +1,1084 @@
...
+    let Some(reply) = message.reply_to_message.as_deref() else {
+        return false;
+    };
+    let Some(from) = reply.from.as_ref() else {
+        return false;
+    };
+    from.is_bot && from.id == bot_user_id
+}
+
+fn recognized_bot_command(message: &TelegramMessage, policy: &GroupTriggerPolicy) -> bool {
```
serrrfirat, line 222: **Medium Severity**

Group command detection accepts recognized commands anywhere in the message, not only at the start.

The policy comment says recognized commands trigger when a group message starts with `/foo` or `/foo@botusername`, but this loop accepts every `bot_command` entity regardless of its offset. In a supergroup, text such as `I ran /help yesterday` can include a `bot_command` entity at offset 6; if `help` is recognized, the adapter forwards ambient chatter as a bot invocation.

Please require `entity.offset == 0` for command triggers, or update the contract to explicitly allow mid-message commands. Add a fixture with a non-leading `/help` entity and assert it returns `NoOp`.

crates/ironclaw_telegram_v2_adapter/src/render.rs
```diff
@@ -0,0 +1,266 @@
...
+    view: &FinalReplyView,
+    credential_handle: EgressCredentialHandle,
+) -> Result<EgressRequest, TelegramRenderError> {
+    let reply = parse_reply_target(target)?;
+    let mut body = serde_json::Map::new();
+    body.insert(
+        "chat_id".into(),
+        serde_json::Value::Number(reply.chat_id.into()),
+    );
+    body.insert("text".into(), serde_json::Value::String(view.text.clone()));
```
serrrfirat, line 125: **Medium Severity**

Final replies are sent as one unbounded Telegram `sendMessage` body.

Telegram rejects message text over its 4096-character limit. As written, a long model reply is rendered into a single request; Telegram returns a 400, and `render_outbound` maps that 4xx to non-retryable `EgressDenied`, so the user receives no reply instead of chunked delivery.

Please split final replies into Telegram-sized chunks (preserving topic/reply target, and code-block safety if parse mode is added) or enforce a documented truncation policy. Add render/adapter tests that an over-limit final reply produces multiple valid `sendMessage` requests or a deliberate truncation outcome.

crates/ironclaw_telegram_v2_adapter/src/payload.rs
```diff
@@ -0,0 +1,1084 @@
...
+        )
+    }
+}
+
+fn classify_trigger(
+    message: &TelegramMessage,
+    chat_kind: TelegramChatKind,
+    policy: &GroupTriggerPolicy,
+) -> Option<ProductTriggerReason> {
+    if chat_kind == TelegramChatKind::Private {
```
serrrfirat, line 159: **Medium Severity**

Private-chat bot commands are downgraded to ordinary user messages.

`classify_trigger` returns `DirectChat` immediately for private chats, and `build_payload` only emits `ProductInboundPayload::Command` when the trigger is `BotCommand`. That means a normal private `/start` or `/help` message with a Telegram `bot_command` entity is delivered as `UserMessage { text: "/start", trigger: DirectChat }` even though the adapter advertises `InboundCommands`. The common onboarding/help/control path will not reach command handling.

Please detect/extract recognized bot commands before the private-chat early return, or have `build_payload` emit `Command` whenever command extraction matches regardless of trigger. Add a private `/start` or `/help` fixture through `parse_inbound`/the runner and assert it produces `ProductInboundPayload::Command`.

crates/ironclaw_telegram_v2_adapter/src/adapter.rs
```diff
@@ -0,0 +1,328 @@
...
+                ProductAdapterError::InvalidIdentifier { kind, reason }
+            }
+        })?;
+        match parsed {
+            TelegramParsedInbound::Envelope(envelope) => Ok(Some(*envelope)),
+            TelegramParsedInbound::NoOp => Ok(None),
+        }
+    }
+
+    async fn render_outbound(
```
serrrfirat, line 116: **Medium Severity**

Outbound rendering does not fail closed when an envelope is routed to the wrong adapter installation.

`render_outbound` never checks that `envelope.adapter_id` and `envelope.installation_id` match this `TelegramV2Adapter` instance. If a projection-router bug, stale queue entry, or corrupted outbound envelope for installation B reaches adapter instance A, the adapter will parse the target and send the reply using A's credential handle. This is the outbound confused-deputy analogue: a misrouted envelope can be delivered with the wrong bot credentials and incorrect delivery accounting.

Please validate `envelope.adapter_id == self.config.adapter_id` and `envelope.installation_id == self.config.installation_id` at the top of `render_outbound`, and fail closed before egress on mismatch. Add a test with mismatched IDs and assert `FakeProtocolHttpEgress` records no send calls.

crates/ironclaw_telegram_v2_adapter/src/adapter.rs
```diff
@@ -0,0 +1,328 @@
...
+            ProductOutboundPayload::ProjectionSnapshot(_)
+            | ProductOutboundPayload::ProjectionUpdate(_) => {
+                // Telegram never consumes projection subscriptions; the
+                // workflow should not route these to a Telegram installation.
+                // Return Ok to keep delivery best-effort.
+                return Ok(());
+            }
+        };
+
+        let response = egress.send(request).await.map_err(map_egress_error)?;
```
serrrfirat, line 169: **Medium Severity**

`DeliveryStatusReporting` is advertised, but the real outbound path never records a `DeliveryStatus`.

The contract/docs say outbound delivery reports statuses to `OutboundDeliverySink`, and Telegram's default capabilities include `DeliveryStatusReporting`. In the actual path here, however, `render_outbound` only calls `egress.send` and returns `Ok`/`Err`; there is no sink parameter or host wrapper recording `Delivered`, `FailedRetryable`, or `FailedUnauthorized` for the envelope's `delivery_attempt_id`. The AC14 test records `sink.record(...)` manually after observing the error, so it does not prove production/caller behavior.

Please add a host outbound runner/wrapper that owns the original `ProductOutboundEnvelope`, invokes `render_outbound`, classifies the result, and records `DeliveryStatus` to a sink (or pass a sink through the render API). Replace the manual sink-record test with caller-level tests for 2xx, 5xx/429, and 4xx outcomes.

crates/ironclaw_telegram_v2_adapter/src/payload.rs
```diff
@@ -0,0 +1,1084 @@
...
+        return Some(ProductTriggerReason::ReplyToBot);
+    }
+    // 3. Recognized bot command.
+    if recognized_bot_command(message, policy) {
+        return Some(ProductTriggerReason::BotCommand);
+    }
+    None
+}
+
+fn has_bot_mention(message: &TelegramMessage, policy: &GroupTriggerPolicy) -> bool {
```
serrrfirat, line 188: **Medium Severity**

Group attachment captions cannot trigger the bot by mention/command.

The group trigger checks only `message.text` plus `entities`, while Telegram sends formatting/entities for captions in `caption_entities` and this DTO does not deserialize that field. As a result, a supergroup photo/document captioned `@ironclaw_bot analyze this` is treated as ambient `NoOp` unless it is also a reply to the bot, even though the adapter advertises inbound attachments and explicit group triggers.

Please deserialize `caption_entities` and run mention/command extraction over `(text, entities)` or `(caption, caption_entities)` as appropriate, then strip leading caption mentions when building `UserMessagePayload`. Add a recorded group photo/document caption-mention fixture and assert the adapter returns an envelope with `BotMention` plus an attachment descriptor.

serrrfirat:
Split this XL PR into a 7-PR stack so each slice is reviewable while preserving the original implementation and retaining `Co-authored-by: Nikolay Pismenkov <nickpismenkov@gmail.com>` on every split commit.

Stack order:

1. #3351 — ProductAdapter core contract
2. #3352 — host auth/egress primitives
3. #3353 — native ProductAdapter runner
4. #3354 — Telegram v2 payload normalization
5. #3355 — Telegram v2 ProductAdapter tracer bullet
6. #3356 — Telegram v2 default-off config guard
7. #3357 — ProductAdapter / Telegram v2 contract docs

I did not close this original PR; it can stay as the umbrella/reference until the split stack is reviewed/merged.

serrrfirat:
Closing these as i already chopped them up in other 7 PRs.

PR #3325, opened 2026-05-07
fix(llm): preserve thinking-mode reasoning artifacts on OpenAI-compat (#3225, #3201)

ilblackdragon:
## Summary

Fixes #3225 and #3201 — both Gemini (via OpenAI-compat) and DeepSeek thinking-mode tool-calling failed deterministically on the second turn because rig-core's OpenAI Completions client silently strips two provider-specific fields the API requires to be echoed back:

- **Gemini** (#3225): HTTP 400 `Function call is missing a thought_signature in functionCall parts`
- **DeepSeek** (#3201): HTTP 400 `The reasoning_content in the thinking mode must be passed back to the API`

## Fix

Replaced the rig-core OpenAI Completions path in `create_openai_compat_from_registry` with a focused custom provider built on `reqwest` (`src/llm/openai_compat.rs`) that captures `reasoning_content` and per-tool-call `thought_signature` on the response and writes them back onto the wire on subsequent turns.

To prevent the same bug class for future OpenAI-compat providers (Qwen, Hunyuan, etc.), the round-trip data has typed homes end-to-end through the agentic loop:

- `ChatMessage.reasoning_content` (assistant message extension)
- `ToolCall.signature` (per-tool-call extension)
- `ToolCompletionResponse.reasoning_content` (provider boundary)
- `RespondResult::ToolCalls { reasoning_content, .. }` (reasoning layer)
- `LoopDelegate::execute_tool_calls(.., reasoning_content, ..)` (agentic loop)

Anthropic, Ollama, GitHub Copilot, NEAR AI, Bedrock, and the Codex ChatGPT provider chains are unchanged — they don't go through `OpenAiCompletions`.

## Regression coverage

Per `.claude/rules/testing.md` "test through the caller":

- Wire layer: `tool_call_signature_round_trips_through_wire_format`, `reasoning_content_is_serialized_on_assistant_messages`, `reasoning_content_is_omitted_when_none`
- Response capture: `tool_call_signature_captured_from_response_payload`, `assistant_reasoning_content_captured_from_response_payload`
- Build-request layer: `build_request_body_carries_round_trip_extensions_for_assistant_turn`
- Reasoning layer (cross-layer): `test_thinking_mode_reasoning_and_signature_propagate_to_respond_result`

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo clippy --all --benches --tests --examples --all-features` — zero warnings
- [x] `cargo test --lib` — 5632 passed
[...]

PR #3326, opened 2026-05-07
fix(llm): route DeepSeek, Gemini, and OpenRouter through dedicated rig-core clients (#3201, #3225)

### LINKED ISSUE #3201: [QA] Tool use for Deepseek is not working
### Environment

local (cloned ironclaw)

### Version / Commit Hash

93c7d6a484237999a7a202efd6d54f70d785c0b7

### QA Test Date

2026-05-02

### Feature Area

LLM Provider

### Steps to Reproduce

1. Configure the LLM Provider as DeepSeek, deepseek-v4-flash
2. Ask recent news
3. See the error

### Expected Behavior

The search complete and show the result.

### Actual Behavior

After the searching, the ironclaw can not return to text generation.
```
                                                                                                                                            ││                                             │
  ── ironclaw ───────────────────────────────────────── 09:21                                                                                ││                                             │
  Error: LLM error: Provider deepseek-v4-flash request failed: HttpError: Invalid status code 400 Bad Request with message:                  ││                                             │
  {"error":{"message":"The reasoning_content in the thinking mode must be passed back to the                                                 ││                                             │
  API.","type":"invalid_request_error","param":null,"code":"invalid_request_error"}}                                                         ││                                             │
                                                                                                                                             ││                                             │
  ⠋ Done 
```

### Logs / Screenshots

_No response_

### Pre-submit checklist

### LINKED ISSUE #3225: bug: gemini API-key backend fails tool-calling with missing thought_signature (gemini-3.1-flash-lite-preview)
## Summary

Fresh deployment using **API key auth** with backend `gemini` and model `gemini-3.1-flash-lite-preview` fails deterministically on the second LLM turn (after first tool call) with HTTP 400 `INVALID_ARGUMENT`:

- `Function call is missing a thought_signature in functionCall parts`
- failing function call reported by API in this repro: `default_api:tool_search` at `position 2`

This issue is facts-only and does **not** propose a fix.

## Prior report context

I previously opened #3214 and misreported scope there (OAuth framing was incorrect for this specific failure mode). Sorry for the confusion/noise on that report.

## Environment (deployment under test)

- Host: `ironclaw1` (accessed via `ssh root@ic1`)
- Binary: `ironclaw 0.27.0`
- Config file: `/root/.ironclaw/config.toml`
  - `llm_backend = "gemini"`
  - `selected_model = "gemini-3.1-flash-lite-preview"`
- Runtime secrets/logs indicate `GEMINI_API_KEY` is loaded from secrets store (`llm_gemini_api_key`)

## Deterministic reproduction

1. Run in isolated HOME with deployment env loaded:
   - `ironclaw --cli-only --no-onboard --message "I want to set up telegram and need activation steps" --auto-approve`
2. First model step returns tool call(s), including `tool_search`.
3. Tool executes successfully.
4. Second model step fails with HTTP 400 `INVALID_ARGUMENT` for missing `thought_signature`.
5. Built-in retry repeats same 400 and exits with final failure.

## Observed logs (key lines)

- `Using OpenAI-compatible provider provider=gemini model=gemini-3.1-flash-lite-preview base_url=https://generativelanguage.googleapis.com/v1beta/openai`
- `Loaded secret 'llm_gemini_api_key' for env var 'GEMINI_API_KEY'`
- `LLM tool_calls response iteration=1 tools=[\"tool_search\"]`
- Followed by repeated warnings/errors:
  - `Provider gemini-3.1-flash-lite-preview request failed: HttpError: Invalid status code 400 ... Function call is missing a thought_signature in functionCall parts ... default_api:tool_search ... position 2`

## Actual result

Tool-calling flow aborts with deterministic 400 after first successful tool execution.

## Expected result

Tool-calling flow continues normally across turns without 400 validation failure.

## Additional notes

- This failure was reproduced on a fresh deployment with otherwise default settings and sqlite-backed local setup.
- I will add a screenshot of the terminal failure output in a follow-up comment.

ilblackdragon:
## Summary

Fixes #3201 and #3225 — both were caused by **wrong provider routing** in `providers.json`. DeepSeek, Gemini, and OpenRouter were configured with `protocol: "open_ai_completions"`, dispatching them through rig-core's generic OpenAI client. That client silently strips the fields these APIs require to be echoed back on the next turn:

- DeepSeek's `reasoning_content` (#3201): API returns `HTTP 400 "The reasoning_content in the thinking mode must be passed back to the API"`
- Gemini's `thought_signature` (#3225): API returns `HTTP 400 "Function call is missing a thought_signature in functionCall parts"`
- OpenRouter's `reasoning` / `reasoning_details` / per-tool-call signatures (same bug class — affects every thinking-mode model OpenRouter exposes: Claude with thinking, OpenAI o-series, DeepSeek-R1, Gemini 2.5+, Qwen QwQ, …)

rig-core has dedicated `deepseek::Client`, `gemini::Client`, and `openrouter::Client` implementations that handle the round-trip natively. The fix is to route to them.

## Change

- Add `ProviderProtocol::DeepSeek`, `Gemini`, and `OpenRouter` variants
- Add three factory functions (mirror the existing `create_anthropic_from_registry` pattern, wrap the dedicated rig-core client in our existing `RigAdapter`); the OpenRouter factory preserves attribution headers (`HTTP-Referer`, `X-Title`)
- Update `providers.json` to use the new protocols and clear `default_base_url` so the dedicated clients use their built-in endpoints

No agent-loop changes, no new HTTP code, no changes to other providers (OpenAI, OpenAI-compat, Tinfoil, Groq, Together, Mistral, Cerebras, Fireworks, etc. all unchanged).

## Why this approach (and the previous PR #3325 wasn't)

PR #3325 built a 731-line custom OpenAI-compat client and plumbed `reasoning_content`/`signature` through 31 files. After verifying rig-core's source, all three providers' dedicated clients already do the round-trip correctly:

- `rig-core/src/providers/deepseek.rs:500-527` — captures `reasoning_content`, writes back on next request
- `rig-core/src/providers/gemini/completion.rs:1016` — `thought_signature: tool_call.signature` round-trip
- `rig-core/src/providers/openrouter/completion.rs:290-308` — `reasoning_details::Encrypted` carries signatures across turns

This PR is ~150 lines and changes nothing about how OpenAI / OpenAI-compat / Tinfoil / Groq / Together / etc. behave.

## What this fixes vs. doesn't

**Fixes** (the bug): API-level round-trip — multi-turn tool calling stops failing for thinking-mode models on these three providers.

**Does not** (filed as a separate follow-up issue (#3327)):

- Surface reasoning content in the chat UI
- Persist reasoning in the thread DB
- Show reasoning in a debug panel
- Stream reasoning chunks via SSE while the model is "thinking"

These are independently valuable but require schema migrations, SSE event additions, and UI work.

## Regression coverage

- `reasoning_aware_providers_use_dedicated_protocol_not_openai_compat` — locks the protocol routing in `providers.json` so a future edit can't silently regress any of the three providers back to `OpenAiCompletions`.

## Test plan

- [x] `cargo fmt`
- [x] `cargo test --lib` — 5620 passed
- [x] `scripts/pre-commit-safety.sh` — clean
- [ ] Manual: `gemini-3.1-flash-lite-preview` API-key — confirm tool calling continues across iterations
- [ ] Manual: DeepSeek thinking-mode (`deepseek-reasoner` / `deepseek-v4-flash`) — confirm tool calling continues
- [ ] Manual: OpenRouter with a thinking model (e.g. `anthropic/claude-opus-4.5:thinking`) — confirm tool calling continues

## Issue linkage audit

Fixes #3201.
Fixes #3225.

Both issues are open as of this audit; this PR is the active closing PR for both.

### DIFF
--- a/providers.json
+++ b/providers.json
@@ -246,13 +246,13 @@
     "aliases": [
       "deep_seek"
     ],
-    "protocol": "open_ai_completions",
-    "default_base_url": "https://api.deepseek.com/v1",
+    "protocol": "deep_seek",
+    "default_base_url": "",
     "api_key_env": "DEEPSEEK_API_KEY",
     "api_key_required": true,
     "model_env": "DEEPSEEK_MODEL",
     "default_model": "deepseek-chat",
-    "description": "DeepSeek inference API",
+    "description": "DeepSeek inference API (preserves reasoning_content for thinking-mode models)",
     "setup": {
       "kind": "api_key",
       "secret_name": "llm_deepseek_api_key",
@@ -325,13 +325,13 @@
       "google_gemini",
       "google"
     ],
-    "protocol": "open_ai_completions",
-    "default_base_url": "https://generativelanguage.googleapis.com/v1beta/openai",
+    "protocol": "gemini",
+    "default_base_url": "",
     "api_key_env": "GEMINI_API_KEY",
     "api_key_required": true,
     "model_env": "GEMINI_MODEL",
     "default_model": "gemini-2.5-flash",
-    "description": "Google Gemini (via OpenAI-compatible endpoint)",
+    "description": "Google Gemini native API (preserves thought_signature on tool calls)",
     "setup": {
       "kind": "api_key",
       "secret_name": "llm_gemini_api_key",
--- a/providers.json
+++ b/providers.json
@@ -124,13 +124,13 @@
     "aliases": [
       "open_router"
     ],
-    "protocol": "open_ai_completions",
-    "default_base_url": "https://openrouter.ai/api/v1",
+    "protocol": "open_router",
+    "default_base_url": "",
     "api_key_env": "OPENROUTER_API_KEY",
     "api_key_required": true,
     "model_env": "OPENROUTER_MODEL",
     "default_model": "openai/gpt-4o",
-    "description": "OpenRouter multi-provider gateway (200+ models)",
+    "description": "OpenRouter multi-provider gateway (200+ models, preserves reasoning across turns)",
     "setup": {
       "kind": "api_key",
       "secret_name": "llm_openrouter_api_key",
--- a/providers.json
+++ b/providers.json
@@ -129,6 +129,7 @@
     "api_key_env": "OPENROUTER_API_KEY",
     "api_key_required": true,
     "model_env": "OPENROUTER_MODEL",
+    "extra_headers_env": "OPENROUTER_EXTRA_HEADERS",
     "default_model": "openai/gpt-4o",
     "description": "OpenRouter multi-provider gateway (200+ models, preserves reasoning across turns)",
     "setup": {
@@ -337,7 +338,7 @@
       "secret_name": "llm_gemini_api_key",
       "key_url": "https://aistudio.google.com/app/apikey",
       "display_name": "Google Gemini",
-      "can_list_models": true
+      "can_list_models": false
     }
   },
   {
--- a/src/llm/mod.rs
+++ b/src/llm/mod.rs
@@ -189,6 +189,7 @@ fn create_registry_provider(
         ProviderProtocol::Ollama => create_ollama_from_registry(config),
         ProviderProtocol::DeepSeek => create_deepseek_from_registry(config),
         ProviderProtocol::Gemini => create_gemini_from_registry(config),
+        ProviderProtocol::OpenRouter => create_openrouter_from_registry(config),
         ProviderProtocol::GithubCopilot => {
             let provider =
                 github_copilot::GithubCopilotProvider::new(config, request_timeout_secs)?;
@@ -468,6 +469,77 @@ fn create_deepseek_from_registry(
         model = %config.model,
         base_url = if config.base_url.is_empty() { "default" } else { &config.base_url },
         "Using DeepSeek provider (preserves reasoning_content across turns)"
+    );
+
+    Ok(Arc::new(
+        RigAdapter::new(model, &config.model)
+            .with_unsupported_params(config.unsupported_params.clone()),
+    ))
+}
+
+/// Build an OpenRouter provider via rig-core's dedicated OpenRouter client.
+///
+/// Routing through this client (rather than the generic OpenAI-compat path)
+/// preserves OpenRouter's `reasoning`, `reasoning_details`, and per-tool-call
+/// signatures across turns. The generic OpenAI client strips all of them, so
+/// any thinking-mode model accessed via OpenRouter (Claude with thinking,
+/// OpenAI o-series, DeepSeek-R1, Gemini 2.5+, Qwen QwQ, …) loses its
+/// reasoning artifacts on the assistant message and the next request fails
+/// the same way as #3201 / #3225.
+fn create_openrouter_from_registry(
+    config: &RegistryProviderConfig,
+) -> Result<Arc<dyn LlmProvider>, LlmError> {
+    use rig::providers::openrouter;
+
+    let api_key = config
+        .api_key
+        .as_ref()
+        .map(|k| k.expose_secret().to_string())
+        .ok_or_else(|| LlmError::AuthFailed {
+            provider: config.provider_id.clone(),
+        })?;
+
+    // OpenRouter attribution headers (`HTTP-Referer`, `X-Title`) and any other
+    // user-configured extras must follow the request through.
+    let mut extra_headers = reqwest::header::HeaderMap::new();
+    for (key, value) in &config.extra_headers {
+        let name = match reqwest::header::HeaderName::from_bytes(key.as_bytes()) {
+            Ok(n) => n,
+            Err(e) => {
+                tracing::warn!(header = %key, error = %e, "Skipping extra header: invalid name");
+                continue;
+            }
+        };
+        let val = match reqwest::header::HeaderValue::from_str(value) {
+            Ok(v) => v,
+            Err(e) => {
+                tracing::warn!(header = %key, error = %e, "Skipping extra header: invalid value");
+                continue;
+            }
+        };
+        extra_headers.insert(name, val);
+    }
+
+    let mut builder = openrouter::Client::builder().api_key(&api_key);
+    if !config.base_url.is_empty() {
+        builder = builder.base_url(&config.base_url);
+    }
+    if !extra_headers.is_empty() {
+        builder = builder.http_headers(extra_headers);
+    }
+
+    let client: openrouter::Client = builder.build().map_err(|e| LlmError::RequestFailed {
+        provider: config.provider_id.clone(),
+        reason: format!("Failed to create OpenRouter client: {e}"),
+    })?;
+
+    let model = client.completion_model(&config.model);
+
+    tracing::debug!(
+        provider = %config.provider_id,
+        model = %config.model,
+        base_url = if config.base_url.is_empty() { "default" } else { &config.base_url },
+        "Using OpenRouter provider (preserves reasoning + signatures across turns)"
     );
 
     Ok(Arc::new(
--- a/src/llm/mod.rs
+++ b/src/llm/mod.rs
@@ -187,6 +187,8 @@ fn create_registry_provider(
         ProviderProtocol::OpenAiCompletions => create_openai_compat_from_registry(config),
         ProviderProtocol::Anthropic => create_anthropic_from_registry(config),
         ProviderProtocol::Ollama => create_ollama_from_registry(config),
+        ProviderProtocol::DeepSeek => create_deepseek_from_registry(config),
+        ProviderProtocol::Gemini => create_gemini_from_registry(config),
         ProviderProtocol::GithubCopilot => {
             let provider =
                 github_copilot::GithubCopilotProvider::new(config, request_timeout_secs)?;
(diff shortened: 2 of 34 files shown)


ilblackdragon:
Closing briefly to regenerate refs/pull/3326/head — GitHub failed to create it on initial push, blocking the regression-test-check workflow which fetches that ref.

ilblackdragon:
Reopening — refs/pull/3326/head should regenerate now.

src/llm/mod.rs
```diff
@@ -424,6 +427,179 @@ fn create_ollama_from_registry(
...
+    let model = client.completion_model(&config.model);
+
+    tracing::debug!(
+        provider = %config.provider_id,
+        model = %config.model,
+        base_url = if config.base_url.is_empty() { "default" } else { &config.base_url },
+        "Using DeepSeek provider (preserves reasoning_content across turns)"
+    );
+
+    Ok(Arc::new(
```
serrrfirat, line 484: **High Severity**

The dedicated clients are still wrapped in `RigAdapter`, which drops required reasoning/signature metadata before the next tool turn.

`rig-core` can now parse DeepSeek/Gemini/OpenRouter provider-specific state, but `RigAdapter::extract_response()` only converts `AssistantContent::ToolCall` into IronClaw `ToolCall { id, name, arguments, reasoning }` and ignores `AssistantContent::Reasoning`, `ToolCall.signature`, and `ToolCall.additional_params`. On the next iteration, `dispatcher` stores `ChatMessage::assistant_with_tool_calls(...)`, and `RigAdapter::convert_messages()` rebuilds bare rig tool calls without those fields. A Gemini thinking model that returned a signed `functionCall` will therefore still be replayed without `thought_signature`; DeepSeek/OpenRouter reasoning artifacts are similarly lost, so the second turn can still fail with the same 400s this PR is intended to fix.

Please either extend IronClaw's message/tool-call types plus `RigAdapter` conversions to round-trip `AssistantContent::Reasoning`, `ToolCall.signature`, and OpenRouter additional params across the tool loop, or use provider-specific `LlmProvider` implementations. Add a caller-level integration test that simulates provider response with tool call → tool result → second provider request and asserts the echoed fields are present.

src/llm/mod.rs
```diff
@@ -424,6 +427,179 @@ fn create_ollama_from_registry(
...
+        .ok_or_else(|| LlmError::AuthFailed {
+            provider: config.provider_id.clone(),
+        })?;
+
+    let client: gemini::Client = if config.base_url.is_empty() {
+        gemini::Client::new(&api_key)
+    } else {
+        gemini::Client::builder()
+            .api_key(&api_key)
+            .base_url(&config.base_url)
```
serrrfirat: **High Severity**

Native Gemini will use stale OpenAI-shim base URL overrides from existing installs.

Before this PR, the UI/config default for Gemini was the OpenAI-compatible shim `https://generativelanguage.googleapis.com/v1beta/openai`. Existing users may have this saved in `llm_builtin_overrides[gemini].base_url`; config resolution still prefers DB overrides over the new empty/native registry default. This factory then passes any non-empty URL directly to `gemini::Client::builder()`, whose native client appends paths like `/v1beta/models/{model}:generateContent`. Result: an upgraded user can send native Gemini requests to `.../v1beta/openai/v1beta/models/...`, breaking normal Gemini calls instead of using the native default endpoint.

Please detect/migrate/ignore the old Gemini OpenAI-shim URL when `ProviderProtocol::Gemini` is selected, or otherwise clear stale persisted overrides. Add a regression test for resolving a saved `https://generativelanguage.googleapis.com/v1beta/openai` override after this protocol switch.

providers.json
```diff
@@ -325,13 +325,13 @@
       "google_gemini",
       "google"
     ],
-    "protocol": "open_ai_completions",
-    "default_base_url": "https://generativelanguage.googleapis.com/v1beta/openai",
+    "protocol": "gemini",
+    "default_base_url": "",
```
serrrfirat, line 330: **Medium Severity**

Gemini is still advertised as list-models capable, but the UI/setup list-models paths do not speak native Gemini.

This registry entry now exposes adapter `gemini` with an empty default base URL while keeping `can_list_models: true`. The web configure surface shows the fetch button, then blocks on a missing base URL before calling `/api/llm/list_models`; if a base URL is supplied, the handler falls through to generic `GET {base}/models` with Bearer auth rather than Gemini's native `/v1beta/models?key=...` shape. The setup wizard has the same issue: it falls through to `fetch_openai_compatible_models(def.default_base_url.unwrap_or(""))`, which returns no models for the new empty default. So users configuring Gemini are told model listing is supported, but it reliably fails/falls back to manual entry.

Please either add native Gemini model-listing support in the web handler/setup wizard or set `can_list_models` to false for Gemini until that path exists. Add a test covering `adapter: "gemini"` with empty/native default base URL.

src/llm/mod.rs
```diff
@@ -424,6 +427,179 @@ fn create_ollama_from_registry(
...
+    // user-configured extras must follow the request through.
+    let mut extra_headers = reqwest::header::HeaderMap::new();
+    for (key, value) in &config.extra_headers {
+        let name = match reqwest::header::HeaderName::from_bytes(key.as_bytes()) {
+            Ok(n) => n,
+            Err(e) => {
+                tracing::warn!(header = %key, error = %e, "Skipping extra header: invalid name");
+                continue;
+            }
+        };
```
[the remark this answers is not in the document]
ilblackdragon: False positive — `http::HeaderName::from_bytes` *normalizes* input (see `http-1.4.0/src/header/name.rs:1115`: "This function normalizes the input"); uppercase ASCII parses fine and is lowercased internally via the `HEADER_CHARS` table. `HTTP-Referer` and `X-Title` round-trip without warnings. Added a clarifying comment in 3a0b3a5d0 so the warn block doesn't read as suspect.

src/llm/mod.rs
```diff
@@ -424,6 +427,179 @@ fn create_ollama_from_registry(
...
+        let val = match reqwest::header::HeaderValue::from_str(value) {
+            Ok(v) => v,
+            Err(e) => {
+                tracing::warn!(header = %key, error = %e, "Skipping extra header: invalid value");
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in 3a0b3a5d0 — both header-warn sites (`create_openrouter_from_registry` and `create_openai_compat_from_registry`) now include `provider = %config.provider_id` so misconfigurations are easy to attribute in multi-provider deployments.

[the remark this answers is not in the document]
ilblackdragon: Duplicate of #discussion_r3199146864 — fixed in 3a0b3a5d0 (added `provider_id` to both header-warn sites).

src/llm/registry.rs
```diff
@@ -478,6 +493,48 @@ mod tests {
...
+    /// providers whose APIs return reasoning artifacts (DeepSeek's
+    /// `reasoning_content`, Gemini's `thought_signature`, OpenRouter's
+    /// `reasoning_details` + signatures) must NOT use the generic
+    /// `OpenAiCompletions` protocol. The OpenAI-compat path goes through
+    /// rig-core's OpenAI client, which strips those fields, breaking
+    /// multi-turn tool calling for every thinking-mode model these
+    /// providers expose. They must route through the dedicated rig-core
+    /// clients which round-trip the artifacts on the next request.
+    #[test]
+    fn reasoning_aware_providers_use_dedicated_protocol_not_openai_compat() {
```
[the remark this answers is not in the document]
ilblackdragon: The PR description already references the actual test name (`reasoning_aware_providers_use_dedicated_protocol_not_openai_compat`); the bot may have read a stale revision. The body has been refreshed in 3a0b3a5d0 along with the new round-trip regression tests.

providers.json
```diff
@@ -124,13 +124,13 @@
     "aliases": [
       "open_router"
     ],
-    "protocol": "open_ai_completions",
-    "default_base_url": "https://openrouter.ai/api/v1",
+    "protocol": "open_router",
+    "default_base_url": "",
     "api_key_env": "OPENROUTER_API_KEY",
     "api_key_required": true,
     "model_env": "OPENROUTER_MODEL",
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in 3a0b3a5d0 — `providers.json` now exposes `"extra_headers_env": "OPENROUTER_EXTRA_HEADERS"`. Users can now configure `HTTP-Referer:…,X-Title:…` for the built-in `openrouter` backend.

src/llm/mod.rs
ilblackdragon: Good catch — confirmed by reading rig-core's `deepseek::Message::try_from` (`src/providers/deepseek.rs:299-322`) and `gemini::Part::From<message::ToolCall>` (`src/providers/gemini/completion.rs:1012-1023`). The dedicated clients only round-trip *inside* rig-core; bridging through `RigAdapter` was where reasoning + signatures were dropped.

Fixed end-to-end in 3a0b3a5d0:

- `ChatMessage` carries `reasoning: Option<String>`; `ToolCall` carries `signature: Option<String>`; `ToolCompletionResponse` carries `reasoning`.
- `RigAdapter::extract_response` captures `AssistantContent::Reasoning` and per-tool-call `signature`.
- `RigAdapter::convert_messages` re-emits `AssistantContent::Reasoning` and pushes `ToolCall::with_signature(…)` when rebuilding rig messages on the next turn.
- Plumbed via `ChatMessage::with_reasoning` through dispatcher (`src/agent/dispatcher.rs:783-791`), job worker (`src/worker/job.rs:1736-1742`), container worker (`src/worker/container.rs:530-535`), routine engine (`src/agent/routine_engine.rs:1940-1948`), and the orchestrator-worker proxy (`src/worker/api.rs` + `src/orchestrator/api.rs`).
- Regression test `reasoning_and_signature_round_trip_through_chat_message` in `src/llm/rig_adapter.rs` simulates a 2-turn tool loop end-to-end: rig response with `Reasoning` + signed `ToolCall` → IronClaw `ChatMessage` → rebuilt rig message, asserting both fields survive.

OpenRouter's `additional_params` (per-tool-call `reasoning_details`) is **not** plumbed yet — left for follow-up #3327 since it requires a typed payload field rather than the simple string round-trip used for DeepSeek/Gemini.

src/llm/mod.rs
ilblackdragon: Fixed in 3a0b3a5d0. New `sanitize_gemini_base_url` in `src/llm/mod.rs` discards any persisted `…/v1beta/openai` (or `…/v1/openai`) shim URL when `ProviderProtocol::Gemini` is selected, with a `warn!` log telling the operator how to clear the stale override. Custom proxies and region endpoints pass through unchanged.

Regression: `sanitize_gemini_base_url_strips_legacy_openai_shim` (covers exact match, trailing slash, mixed case, and the `/v1/openai` variant) plus `sanitize_gemini_base_url_passes_through_empty` and `sanitize_gemini_base_url_preserves_custom_endpoints`.

providers.json
ilblackdragon: Fixed in 3a0b3a5d0 by setting `"can_list_models": false` on Gemini in `providers.json`. The setup wizard's fall-through path (`fetch_openai_compatible_models(def.default_base_url.unwrap_or(""), …)`) wouldn't speak the native `/v1beta/models?key=…` shape, and the web list-models handler is the same. Native Gemini list-models support is filed as a follow-up rather than blocking this PR.

PR #3328, opened 2026-05-07
fix(engine,web): suppress restart-recovery noise on Projects tab; retry empty hydration on SSE open (#3274)

### LINKED ISSUE #3274: Bug: Data Missing After Upgrading IronClaw 0.26.0 → 0.27.0 Until Manual Refresh, and Projects Tab Shows Failed Threads Warning
### Description

After upgrading an existing IronClaw instance `glad-pig-nodoj` from `0.26.0` to `0.27.0`, several UI/data consistency issues appeared immediately after login.

---

## Issue 1: Chat Tab Initially Shows No Data

Immediately after opening the instance after upgrade:

- `Chat` tab displayed no previous chat data
- Existing conversations appeared missing

However, after manually refreshing the page:
- Chat history/data reappeared correctly

This suggests an initialization/state restoration issue after upgrade.

---

## Issue 2: Missions Tab Initially Empty

Immediately after upgrade/login:
- `Missions` tab showed no data

After manually refreshing the page:
- Mission data appeared normally again

This may indicate:
- frontend hydration issue
- delayed state synchronization
- failed initial data fetch after upgrade

---

## Issue 3: Projects Tab Shows Failed Threads Warning

A new `Projects` tab appeared after upgrade.

Inside the tab, a persistent warning is displayed:

```text
NEEDS ATTENTION
default — Thread failed: ...
```

Even after page refresh (when Chat and Missions recovered), the Projects tab still shows the failed thread warning.

Network response shows:

```json
{
  "type": "failure",
  "project_name": "default",
  "message": "Thread failed: ...",
  "thread_id": "..."
}
```

---

Image

Image

ilblackdragon:
## Summary

Closes #3274 — three post-upgrade UI symptoms (0.26.0 → 0.27.0) with two distinct root causes.

- **Projects tab persistent \"Thread failed: …\" warning (Issue 3, root cause).** `ThreadManager::recover_project_threads` force-fails every non-terminal thread on engine restart with reason `\"engine restart before thread completion\"`. After upgrade, every running/created/waiting thread from the previous process gets `state = Failed` with `updated_at = now`, which then floods `get_engine_projects_overview`'s `state == Failed && updated_at >= now − 24h` filter as `kind=failure` attention items. Recovered threads are now tagged with the metadata flag `engine_restart_recovery: true` (exported from the engine crate as `ENGINE_RESTART_RECOVERY_METADATA_KEY`), and the projects overview excludes them from both `failures_24h` and the \"Needs attention\" feed.
- **Chat / Missions tabs empty until manual refresh (Issues 1 & 2, defensive net).** `loadHistory`, `loadThreads`, and `loadMissions` ended in `.catch(() => {})` / `.catch(function() {})` — the silent-failure anti-pattern called out in `.claude/rules/error-handling.md`. They now log via `console.error` and flag `window._initialHydrationPending`. A new `runInitialHydrationRetry()` re-runs only the flagged loaders, exactly once per `initApp()` (guarded by `_hydrationRetryDone`), wired into the SSE `onopen` handler. SSE accept implies the backend has stabilised — `init_engine` is awaited before `channels.start_all()` (`src/agent/agent_loop.rs:744`).

The fix preserves real failure surfacing: a thread that explicitly transitions to `Failed` for any non-recovery reason still appears in the attention feed.

## Files

- `crates/ironclaw_engine/src/{lib.rs,runtime/manager.rs}` — metadata constant, recovery tagging, regression-tighten existing test.
- `src/bridge/router.rs` — extracted `is_real_thread_failure` predicate, applied to `failures_24h` count + attention loop, plus four new unit tests.
- `crates/ironclaw_gateway/static/js/{core/{history,init-auth,sse}.js,surfaces/projects.js}` — replace silent catches, hydration retry plumbing.

## Test plan

- [x] `cargo fmt`
- [x] `cargo test --lib` — 5623 passed
- [x] `cargo test -p ironclaw_engine` — 520 passed (incl. tightened `recover_project_threads_marks_non_terminal_as_failed`)
- [x] `bash scripts/pre-commit-safety.sh` — exit 0
- [ ] Manual smoke after deploy: confirm Projects tab no longer flags restart-recovered threads while a real LLM-error failure still surfaces; confirm Chat/Missions self-recover on first load with a logged error in DevTools when initial fetch fails.

## Out of scope (follow-ups if the retry doesn't fully resolve 1 & 2)

- Deeper root-cause investigation of why `loadThreads` / `loadMissions` would error on the first post-upgrade call (e.g. a tenant whose `user_id` is not covered by `migrate_legacy_user_ids`). The new `console.error` makes this surfaceable.
- A first-class `engine_ready` SSE event instead of relying on the implicit \"SSE-open == engine ready\" assumption.

### DIFF
--- a/crates/ironclaw_gateway/static/js/core/init-auth.js
+++ b/crates/ironclaw_gateway/static/js/core/init-auth.js
@@ -1,4 +1,35 @@
+// Tracks loaders that failed on the very first call after `initApp()`. The
+// SSE `onopen` handler in `core/sse.js` retries each flagged loader exactly
+// once — see `runInitialHydrationRetry` below. Defensive net for the upgrade
+// race in #3274 where the first hydration request loses to in-flight engine
+// state initialization or DB migration; a manual refresh used to be the only
+// recovery path. See `.claude/rules/error-handling.md` (silent-failure rule).
+function runInitialHydrationRetry() {
+  var pending = window._initialHydrationPending;
+  if (!pending || window._hydrationRetryDone) return;
+  window._hydrationRetryDone = true;
+  if (pending.threads && typeof loadThreads === 'function') {
+    console.info('[hydration] retrying loadThreads after SSE connect');
+    loadThreads();
+  }
+  if (pending.history && typeof loadHistory === 'function') {
+    console.info('[hydration] retrying loadHistory after SSE connect');
+    loadHistory();
+  }
+  if (pending.missions
+      && currentTab === 'missions'
+      && typeof loadMissions === 'function') {
+    console.info('[hydration] retrying loadMissions after SSE connect');
+    loadMissions();
+  }
+  window._initialHydrationPending = null;
+}
+
 function initApp() {
+  // Reset hydration tracker each time we (re-)initialize the app — token
+  // re-auth and OIDC auto-auth both flow through here.
+  window._initialHydrationPending = { threads: false, history: false, missions: false };
+  window._hydrationRetryDone = false;
   var authScreen = document.getElementById('auth-screen');
   var app = document.getElementById('app');
   // Cross-fade: fade out auth screen, then show app
--- a/crates/ironclaw_engine/src/lib.rs
+++ b/crates/ironclaw_engine/src/lib.rs
@@ -83,7 +83,10 @@ pub use gate::{
 
 pub use executor::prompt::PlatformInfo;
 pub use runtime::conversation::ConversationManager;
-pub use runtime::manager::ThreadManager;
+pub use runtime::manager::{
+    ENGINE_RESTART_RECOVERY_METADATA_KEY, PENDING_APPROVAL_METADATA_KEY,
+    RUNTIME_CHECKPOINT_METADATA_KEY, ThreadManager,
+};
 pub use runtime::messaging::ThreadOutcome;
 pub use runtime::mission::{
     BudgetGate, FireRateLimit, MissionGateInfo, MissionManager, MissionNotification, MissionUpdate,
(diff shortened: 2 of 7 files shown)


crates/ironclaw_gateway/static/js/core/init-auth.js
```diff
@@ -1,4 +1,35 @@
+// Tracks loaders that failed on the very first call after `initApp()`. The
+// SSE `onopen` handler in `core/sse.js` retries each flagged loader exactly
+// once — see `runInitialHydrationRetry` below. Defensive net for the upgrade
+// race in #3274 where the first hydration request loses to in-flight engine
+// state initialization or DB migration; a manual refresh used to be the only
+// recovery path. See `.claude/rules/error-handling.md` (silent-failure rule).
+function runInitialHydrationRetry() {
+  var pending = window._initialHydrationPending;
+  if (!pending || window._hydrationRetryDone) return;
+  window._hydrationRetryDone = true;
```
serrrfirat, line 10: **Medium Severity**

The hydration retry can be consumed before any loader has failed.

`runInitialHydrationRetry()` marks `_hydrationRetryDone = true` and later clears `_initialHydrationPending` even when all pending flags are still false. In `initApp()`, `connectSSE()` is called before the initial `loadThreads()` call, so a fast SSE `onopen` can run this function first, consume the one retry, and clear the tracker. If the first `loadThreads` / `loadHistory` / `loadMissions` request then rejects, the catch block sees `_initialHydrationPending === null` and cannot flag itself for retry. That leaves the same blank initial UI state this change is trying to recover from.

Consider only marking the retry as done when at least one pending flag is true, and have loader catches trigger the retry immediately when SSE is already open (or track an explicit `initialHydrationSseReady` flag).

serrrfirat (approved):
Approved after paranoid review. One medium follow-up was left inline for the hydration retry race.

PR #3331, opened 2026-05-07
feat(web): polish non-image attachment UI and persist user attachment and update paperclip ui

### LINKED ISSUE #1341: Support non-image file attachments in web gateway (PDF, audio, documents)
## Problem / Motivation

The web gateway chat input only accepts image attachments (`accept="image/*"` on the file input, frontend filters with `file.type.startsWith('image/')`, backend validates `img.media_type.starts_with("image/")`). Users cannot attach PDFs, audio files, text documents, or any other file type.

The backend **already supports** non-image attachments — `AttachmentKind` in `src/channels/channel.rs` has `Audio`, `Image`, and `Document` variants, and `from_mime_type()` correctly classifies any MIME type. The limitation is purely in the web gateway frontend and its DTO layer.

### Frontend (`src/channels/web/static/`)

**index.html:**
- Change `<input type="file" id="image-file-input" accept="image/*" multiple>` to remove the `accept` restriction (or set a broad accept list like `image/*,audio/*,application/pdf,.txt,.md,.csv,.json,.xml,.doc,.docx,.xls,.xlsx`).
- Rename `id="image-file-input"` → `id="file-input"` (and all JS references).

**app.js:**
- Rename `stagedImages` → `stagedFiles` (or `stagedAttachments`).
- Rename `handleImageFiles()` → `handleFiles()`.
- Remove the `if (!file.type.startsWith('image/')) return;` filter — accept all file types.
- Rename constants: `MAX_IMAGE_SIZE_BYTES` → `MAX_FILE_SIZE_BYTES`, `MAX_STAGED_IMAGES` → `MAX_STAGED_FILES`. Consider bumping `MAX_FILE_SIZE_BYTES` to 8MB (the server body limit is 10MB).
- Update `renderImagePreviews()` → `renderFilePreviews()`:
  - For images: show thumbnail as today (60x60 preview).
  - For non-images: show a file icon with the filename and size (e.g., `📄 report.pdf (1.2 MB)`). Use a simple `<div>` with text instead of ``.
- Update the send payload: rename `images` key → `attachments` in the POST body. Each item: `{ media_type, data, filename }` (add `filename` field so the backend can preserve original names).
- Update paste handler: keep image paste behavior, but don't try to handle non-image paste (clipboard file paste isn't broadly supported anyway).

**style.css:**
- Rename `.image-preview-strip` → `.file-preview-strip`.
- Add styles for non-image file previews (filename truncation, file icon, size label).

### Backend (`src/channels/web/`)

**types.rs:**
- Rename `ImageData` → `AttachmentData`:
  ```rust
  pub struct AttachmentData {
      pub media_type: String,
      pub data: String, // base64
      pub filename: Option<String>,
  }
  ```
- In `SendMessageRequest`: rename `images: Vec<ImageData>` → `attachments: Vec<AttachmentData>`. Keep `#[serde(default)]`. Add `#[serde(alias = "images")]` for backward compatibility with any existing API consumers.

**server.rs:**
- Rename `images_to_attachments()` → `decode_attachments()`.
- Remove the `img.media_type.starts_with("image/")` validation — accept any MIME type.
- Use `AttachmentKind::from_mime_type()` (already exists in `channel.rs`) instead of hardcoding `AttachmentKind::Image`.
- Use `filename` from the DTO when present instead of generating `image-{i}.{ext}`.
- Expand `mime_to_ext()` to handle common non-image MIME types (`application/pdf` → `pdf`, `audio/mpeg` → `mp3`, `audio/ogg` → `ogg`, `text/plain` → `txt`, etc.), with a fallback to `bin`.

**handlers/chat.rs:**
- Update `chat_send_handler` to use the renamed field (`req.attachments` instead of `req.images`).

## Code Pointers

- `src/channels/web/static/index.html` — file input element and attach button (lines ~165-167)
- `src/channels/web/static/app.js` — `handleImageFiles()`, `renderImagePreviews()`, `stagedImages`, send payload construction
- `src/channels/web/static/style.css` — `.image-preview-strip`, `.attach-btn` styles
- `src/channels/web/types.rs` — `ImageData`, `SendMessageRequest` DTOs
- `src/channels/web/server.rs` — `images_to_attachments()`, `mime_to_ext()`
- `src/channels/channel.rs` — `AttachmentKind::from_mime_type()` (already handles audio/document)

## Acceptance Criteria

- [ ] File input accepts all common file types (images, PDFs, audio, text, office docs)
- [ ] Non-image files display as filename + size in the preview strip (not a broken image thumbnail)
- [ ] Image files still display as visual thumbnails
- [ ] Each file preview has an X button to remove before sending
- [ ] Backend correctly classifies attachments as `Audio`, `Image`, or `Document` via `AttachmentKind::from_mime_type()`
- [ ] Original filename is preserved through to `IncomingAttachment.filename`
- [ ] `#[serde(alias = "images")]` on the DTO so old API consumers don't break
- [ ] Paste still works for images
- [ ] Size limit enforced per file (alert shown with filename and size)
- [ ] File count limit enforced (alert shown)
- [ ] 10MB server body limit is documented in a comment near the frontend size constant

## Pitfalls & Landmines

- The server body limit is 10MB (`DefaultBodyLimit::max(10 * 1024 * 1024)` in server.rs). Base64 encoding inflates size ~33%, so a 7.5MB file will hit the limit. Set the frontend max to ~7MB or document the overhead.
- `mime_to_ext()` currently only handles image types — it MUST be extended or non-image files get `.bin` extensions.
- The `data: URL` approach (`FileReader.readAsDataURL`) works for all file types, not just images. The base64 extraction logic (`split(',')[1]`) is universal.
- Don't break the clipboard paste path — it only needs to handle images.

## Non-Goals

- Drag-and-drop file upload (separate issue if desired)
- Server-side file storage / CDN upload (files stay base64 in the JSON body for now)
- File type restrictions based on agent capabilities

### LINKED ISSUE #1342: Replace paperclip emoji with styled + button for attachment trigger
## Problem / Motivation

The attachment button in the web gateway chat input uses a paperclip Unicode emoji (`&#x1F4CE;` / 📎) as its icon. This looks inconsistent with the rest of the UI (which uses no emoji for controls) and doesn't clearly communicate "attach a file." A `+` circle button before the message input is the standard pattern used by iMessage, Telegram, WhatsApp, Slack, and most modern chat UIs.

### Layout change

Move the attach button to **before** the textarea (currently it sits after). The input area layout should be:

```
[ + ] [ textarea ........................ ] [ Send ]
```

**index.html** — reorder elements inside `.chat-input`:
```html
<div class="chat-input">
  <div id="file-preview-strip" class="file-preview-strip"></div>
  <button id="attach-btn" class="attach-btn" title="Attach files" aria-label="Attach files">+</button>
  <div class="chat-input-wrapper">
    <textarea ...></textarea>
    <div id="ghost-text" class="ghost-text"></div>
  </div>
  <input type="file" id="file-input" multiple style="display:none">
  <button id="send-btn">Send</button>
</div>
```

### Styling

**style.css** — restyle `.attach-btn`:
```css
.chat-input .attach-btn {
  width: 32px;
  height: 32px;
  min-width: 32px;
  border-radius: 50%;
  border: 1.5px solid var(--border);
  background: none;
  color: var(--text-secondary);
  font-size: 1.1em;
  font-weight: 300;
  cursor: pointer;
  display: flex;
  align-items: center;
  justify-content: center;
  align-self: flex-end;
  margin-bottom: 4px;
  transition: color 0.2s, border-color 0.2s, background 0.2s;
  padding: 0;
  line-height: 1;
}

.chat-input .attach-btn:hover {
  color: var(--text);
  border-color: var(--text-secondary);
  background: rgba(255, 255, 255, 0.05);
}
```

Key design decisions:
- Circle with thin border, not filled — matches the dark minimal aesthetic
- `+` character, not an SVG — keeps it simple, no asset management
- 32px diameter — large enough for touch targets on mobile (meets WCAG 44px with padding)
- Sits at `flex-end` vertically to align with the bottom of a multi-line textarea

## Code Pointers

- `src/channels/web/static/index.html` — attach button element (line ~166-167)
- `src/channels/web/static/style.css` — `.chat-input .attach-btn` styles (lines ~3935-3955)
- `src/channels/web/static/app.js` — `document.getElementById('attach-btn')` click handler (line ~591)

## Acceptance Criteria

- [ ] Attach button displays as a `+` inside a circular border
- [ ] Button is positioned **before** (left of) the textarea, not after
- [ ] Layout order: `[+] [textarea] [Send]`
- [ ] Button is vertically aligned to the bottom of the textarea
- [ ] Hover state: border and text lighten, subtle background appears
- [ ] Touch target is at least 44px including padding (mobile accessibility)
- [ ] No functional change to click behavior (still triggers file input)
- [ ] Mobile responsive: button doesn't shrink below 32px or get clipped

## Pitfalls & Landmines

- The `.chat-input` uses `flex-wrap: wrap` — make sure the button doesn't wrap to a new line on narrow screens. It should stay inline with the textarea.
- The file preview strip sits above the input row and spans full width — it must remain above the `[+] [textarea] [Send]` row, not get displaced.
- If this is done alongside issue #1341 (generic file attachments), coordinate the element ID rename (`image-file-input` → `file-input`).

## Non-Goals

- Changing the send button style
- Adding additional buttons (microphone, etc.) — tracked separately
- Dropdown menu on the + button (just triggers file picker directly)

### LINKED ISSUE #3272: Bug: Uploaded Attachments Lose Preview After Refresh, Duplicate Images Appear, and Pasted Image Preview Is Excessively Large
### Description

There are multiple attachment-related issues in the chat UI involving pasted/uploaded images and file persistence.

---

### Issue 1: Attachment Preview Disappears After Refresh

Uploaded images initially render correctly with previews. However, after refreshing the page or reopening the conversation later, the preview disappears and only attachment filenames remain visible.

Example after refresh:

```text
[Attachment] image.png (image/png • 536KB)
[Attachment] image-0.png (image/png • 536KB)

README copy.md
text/markdown • 13KB
```

---

### Issue 2: Duplicate Image Attachments Appear

Only a single image was uploaded/pasted in the message, but after reload the message displays two image attachments:

```text
image.png
image-0.png
```

This suggests duplicated attachment references or incorrect attachment restoration logic.

---

### Issue 3: Pasted Image Preview Is Excessively Large

When directly using **Copy → Paste** to insert an image into chat, the preview image renders at an extremely large size in the conversation UI.

This causes:
- Poor readability
- Excessive vertical space usage
- Chat layout disruption
- Difficult scrolling/navigation

The preview should be constrained to a reasonable max width/height similar to modern chat applications.

Image

Image

italic-jinxin:
## Summary

1. Closes the user-visible polish for #1341 ("Support non-image file
attachments in web gateway"). Backend MIME allowlist + magic-byte
sniffing already shipped in #2332 / #2385 / #2763; this PR finishes
the per-file budget bump and the staging-strip UI.

2. User-uploaded image attachments now survive a page refresh — instead of
degrading into a generic file card, the chat surface re-fetches the
original bytes and renders them inline.

The persisted `<attachments>` XML carries metadata only (filename,
mime, size). Without disk persistence + a URL the browser can fetch,
images visibly "disappear" on every refresh. This PR closes that gap.

3. Redesigns the chat composer's attachment trigger from a 📎 emoji button into
a properly styled 32×32 circular button with a Lucide-style `+` icon, in the
visual tradition of iMessage / Slack / WhatsApp. Closes #1342.

## Changes

**Backend**
- `MAX_INLINE_ATTACHMENT_BYTES`: 5 MiB → 7 MiB. Total budget (10 MiB),
  file count (5), and 14 MiB request body cap unchanged.
- Update `web/CLAUDE.md` to match.

**Frontend**
- Replace `"DOCX"`-text icon with Lucide-style SVGs (audio = music note,
  others = generic file). Show `type • size` in card meta.
- Reject `video/*` at the chat-input boundary with an i18n alert (backend
  allowlist excludes video).
- Click any image (`.image-preview`, `.message-attachment-image`,
  `.generated-image`) to open a fullscreen lightbox; ESC / backdrop closes.
- Fix the × remove button — `.chat-input button` (Send accent-pill) was
  outranking the bare selector and rendering it as a stretched green pill.
- Drop dead `stagedImages` / `handleImageFiles` legacy paths.

**Backend**

- New `persist_legacy_image_attachments` runs on user-message ingest
  (`thread_ops::process_user_input`), writing image bytes to disk and
  stamping `local_path` on the attachment so `format_attachment` can
  emit `project_path="..."` in the persisted XML.
- Shared helpers extracted into `channels/attachments.rs`:
  `sanitize_attachment_segment`, `persist_attachment_at`,
  `legacy_attachment_relative_path`. Engine v2's existing
  `persist_project_attachments` was refactored to call them — pure
  refactor, no behavior change.
- Disk layout (under `~/.ironclaw/attachments/`):
  - v1 (default): `<owner>/.legacy/<msg>/<file>`
  - v2: `<owner>/<project>/<date>/<file>` (unchanged)

**HTTP route**

- New `GET /api/attachments/{owner}/{*path}`, behind the regular
  gateway Bearer-auth middleware. Authorization is layered:
  - URL `<owner>` segment must match the authenticated user.
    Mismatch → **404** (deliberately not 403 — we don't confirm
    files exist for a different user).
  - `..` and `\` rejected at the URL boundary.
  - Canonical-path containment check inside the owner directory.
  - Streamed with `tokio::fs::File::open` + `ReaderStream` per
    `.claude/rules/safety-and-sandbox.md` "Bounded Resources".

**History surface**

- New `HistoryResponse.turns[].user_attachments` field shape:
  `{ filename, mime_type, size_label, url, kind }`. Deliberately
  out-of-band from the LLM-facing `user_input` text — the
  `<attachments>` XML stays clean and the frontend gets the URLs it
  needs without us polluting the agent's context.
- Server-side `extract_user_attachments` parses each persisted
  `<attachment project_path="...">` element and maps it to a
  `/api/attachments/{owner}/...` URL.

**Frontend**

- `renderMessageAttachments` now fetches images with
  `Authorization: Bearer ...` and swaps a blob URL onto the ``.
  Token never enters a URL — no leakage via logs / Referer / browser
  history. Fetch failure (404 / auth / network) falls back to the
  existing file-card render so the bubble never shows a broken-image
  icon.
- Pre-upload JPEG canvas resize: source ≥ 1600 px gets re-encoded as
  JPEG q=0.88 before upload, capping request body size and LLM
  context. PNG / GIF / WebP pass through untouched to preserve
  transparency and animation.

*HTML** ([crates/ironclaw_gateway/static/index.html](crates/ironclaw_gateway/static/index.html))
- Wrap `[file-input][attach-btn][textarea][send-btn]` in a new
  `.chat-input-row` so the preview strip sits cleanly above the input row
  (no more flex-wrap reflow hack).
- Move `#attach-btn` to the **left** of the textarea (was: right).
- Swap the `📎` emoji for an inline Lucide `plus` SVG with `aria-hidden`.

**CSS** ([crates/ironclaw_gateway/static/styles/surfaces/chat.css](crates/ironclaw_gateway/static/styles/surfaces/chat.css))
- `.chat-input` → `flex-direction: column` (was `row + flex-wrap`).
- New `.chat-input-row` for the actual input row.
- `.chat-input .attach-btn`:
  - 32×32 circular (`width/height: 32px`, `border-radius: 50%`)
  - 1.5px border in `var(--border)`, transparent background
  - 6px padding → 44×44 total touch target (WCAG 2.5.5)
  - Hover: border + icon brighten, subtle bg-tertiary fill, 150ms ease
  - Active: `transform: scale(0.95)`
  - Focus: 2px accent outline (`:focus-visible`)
  - SVG inside fixed at 18×18, `display: block`
  - Selector specificity `(0,2,0)` wins over the broader
    `.chat-input button` accent-pill rule `(0,1,1)` from both base.css
    mobile @media block and the Send button styles — explicit padding /
    border-radius / background resets are no longer needed.

**JS** ([crates/ironclaw_gateway/static/js/core/widgets.js](crates/ironclaw_gateway/static/js/core/widgets.js))
- Comment-only fix: "paperclip in the composer" → "plus button in the
  composer". No logic change.

**No JS behavior changes.** All wiring (`#attach-btn` / `#image-file-input`
event listeners) preserved untouched.

### Screenshots 

| Attachment in chat input | Attachment in chat message | Image Preview | Paper clip enhance |
|---|---|---|---|
| attachment in chat input | attachment in chat message | preview model | paper clip enhance |

## Change Type

- [x] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] CI/Infrastructure
- [ ] Security
- [ ] Dependencies

## Linked Issue

Closes #1341  and #3272  and  #1342 

## Validation

- [ ] `cargo build`
- [ ] Relevant tests pass: 
- [ ] Manual testing: 

## Review Follow-Through

---

**Review track**:


zmanian:
## Code Review

### Overview
+1870/-251 across 20 files. Bumps inline attachment cap 5→7 MiB; adds disk persistence for v1 image attachments; adds authed `GET /api/attachments/{owner}/{*path}` route; UI polish (icons, lightbox, video rejection, blob-URL re-fetch); pre-upload JPEG resize; i18n in en/ko/zh-CN; new e2e + unit tests. Closes #1341 and #3272.

### Strengths
- Clear separation between v1 legacy and v2 project-aware persistence; shared helpers in `channels/attachments.rs` extracted as a pure refactor.
- Path safety done correctly: segment sanitization, rejecting `..`/`\`, plus canonical-path containment check.
- 404 (not 403) on owner mismatch — correct info-disclosure posture.
- Streamed via `ReaderStream` per "Bounded Resources" rule rather than buffered read.
- Bearer token kept in `Authorization` header, not URL — avoids log/Referer leak.
- i18n updated across all 3 locales (often forgotten).

### Issues to address before merge

**Major:**

1. **Filename collision after sanitization** (Copilot flagged). `legacy_attachment_relative_path` uses only the sanitized filename — `a b.png` and `a_b.png` collide; later write overwrites the earlier, refresh returns wrong bytes. Must include `attachment_index` in the final path segment unconditionally.

2. **Blocking `Path::canonicalize()` on async runtime** in `serve_user_attachment` (Copilot flagged). Use `tokio::fs::canonicalize` to avoid stalling the executor on slow disk.

3. **No `Cache-Control` header** on attachment responses (Copilot flagged). Per-user authed content; needs at minimum `Cache-Control: private, no-store` to prevent shared-proxy / browser-cache leakage.

4. **Blob URL leak** in `fetchAttachmentAsBlobUrl` — never `URL.revokeObjectURL`'d. Long sessions and `pruneOldMessages` will accumulate object URLs in the renderer.

**Minor:**

5. Unnecessary attachment clone in `thread_ops::process_user_input` doubles memory cost up to the inline budget; persist in place or clone metadata only.
6. PR validation checklist is entirely unchecked — no claim that `cargo fmt`/`clippy`/tests were run locally.
7. Security/Database/Blast Radius/Rollback sections of PR template are blank despite touching auth-gated HTTP routes and on-disk persistence. New filesystem-write surface and a new authed route warrant explicit answers.
8. Wasmtime patch bump bundled in a UX PR — unrelated; better isolated, though low risk.
9. Confirm the new route goes through the same Bearer auth middleware as other `/api/*` (router diff suggests yes — verify it's not under a CSRF-exempt branch).

### Risk
Medium. New write-to-disk surface + new authed download route — auth gating and traversal defenses look right, but the filename-collision bug is a real correctness issue (silently serves wrong file to the same user) and the blocking `canonicalize` is a real perf footgun under load. Blob-URL leak is a slow-burn frontend issue.

### Recommendation
**Request Changes.** Pre-merge: (1) add `attachment_index` to legacy path unconditionally, (2) switch to `tokio::fs::canonicalize`, (3) emit `Cache-Control: private, no-store`, (4) revoke blob URLs on prune/replace, (5) avoid the full attachment clone, (6) fill in Validation/Security/Rollback PR-body sections. Splitting the wasmtime bump into its own PR would help. Once those land, this is a clean merge.

italic-jinxin:
> ## Code Review
> ### Overview
> +1870/-251 across 20 files. Bumps inline attachment cap 5→7 MiB; adds disk persistence for v1 image attachments; adds authed `GET /api/attachments/{owner}/{*path}` route; UI polish (icons, lightbox, video rejection, blob-URL re-fetch); pre-upload JPEG resize; i18n in en/ko/zh-CN; new e2e + unit tests. Closes #1341 and #3272.
> 
> ### Strengths
> * Clear separation between v1 legacy and v2 project-aware persistence; shared helpers in `channels/attachments.rs` extracted as a pure refactor.
> * Path safety done correctly: segment sanitization, rejecting `..`/`\`, plus canonical-path containment check.
> * 404 (not 403) on owner mismatch — correct info-disclosure posture.
> * Streamed via `ReaderStream` per "Bounded Resources" rule rather than buffered read.
> * Bearer token kept in `Authorization` header, not URL — avoids log/Referer leak.
> * i18n updated across all 3 locales (often forgotten).
> 
> ### Issues to address before merge
> **Major:**
> 
> 1. **Filename collision after sanitization** (Copilot flagged). `legacy_attachment_relative_path` uses only the sanitized filename — `a b.png` and `a_b.png` collide; later write overwrites the earlier, refresh returns wrong bytes. Must include `attachment_index` in the final path segment unconditionally.
> 2. **Blocking `Path::canonicalize()` on async runtime** in `serve_user_attachment` (Copilot flagged). Use `tokio::fs::canonicalize` to avoid stalling the executor on slow disk.
> 3. **No `Cache-Control` header** on attachment responses (Copilot flagged). Per-user authed content; needs at minimum `Cache-Control: private, no-store` to prevent shared-proxy / browser-cache leakage.
> 4. **Blob URL leak** in `fetchAttachmentAsBlobUrl` — never `URL.revokeObjectURL`'d. Long sessions and `pruneOldMessages` will accumulate object URLs in the renderer.
> 
> **Minor:**
> 
> 5. Unnecessary attachment clone in `thread_ops::process_user_input` doubles memory cost up to the inline budget; persist in place or clone metadata only.
> 6. PR validation checklist is entirely unchecked — no claim that `cargo fmt`/`clippy`/tests were run locally.
> 7. Security/Database/Blast Radius/Rollback sections of PR template are blank despite touching auth-gated HTTP routes and on-disk persistence. New filesystem-write surface and a new authed route warrant explicit answers.
> 8. Wasmtime patch bump bundled in a UX PR — unrelated; better isolated, though low risk.
> 9. Confirm the new route goes through the same Bearer auth middleware as other `/api/*` (router diff suggests yes — verify it's not under a CSRF-exempt branch).
> 
> ### Risk
> Medium. New write-to-disk surface + new authed download route — auth gating and traversal defenses look right, but the filename-collision bug is a real correctness issue (silently serves wrong file to the same user) and the blocking `canonicalize` is a real perf footgun under load. Blob-URL leak is a slow-burn frontend issue.
> 
> ### Recommendation
> **Request Changes.** Pre-merge: (1) add `attachment_index` to legacy path unconditionally, (2) switch to `tokio::fs::canonicalize`, (3) emit `Cache-Control: private, no-store`, (4) revoke blob URLs on prune/replace, (5) avoid the full attachment clone, (6) fill in Validation/Security/Rollback PR-body sections. Splitting the wasmtime bump into its own PR would help. Once those land, this is a clean merge.

resolved

serrrfirat:
Parallel review findings:

1. **High — Security / authorization** — `src/channels/web/platform/static_files.rs:990`

   Attachment auth compares the URL owner segment to raw `authenticated_user_id`, but persistence stores files under `sanitize_attachment_segment(user_id)` (`src/channels/attachments.rs:93`, `src/bridge/router.rs:149`). The sanitizer is lossy (`|`, `@`, spaces, `/`, etc. become `_`). OIDC `sub`-backed users with unsafe chars (`src/channels/web/platform/auth.rs:1119`) can fail to load their own persisted files, and sanitized collisions can let a raw user id matching the sanitized segment read files from the same owner directory.

   Suggested fix: replace lossy owner directory names with collision-resistant deterministic IDs (base64url/sha256 or DB UUID) shared by persist and serve, then compare against that exact derived owner. Add route tests for unsafe/colliding IDs such as `a|b` vs `a_b`.

2. **Medium — Security / upload content serving** — `src/channels/web/platform/static_files.rs:1020`

   User uploads are served with content type guessed from the filename. `application/octet-stream` uploads are allowed (`src/channels/web/util.rs:114`) and preserve normalized original filenames (`src/channels/web/util.rs:271`), so a user can persist bytes named `*.html` / `*.js`; the new `/api/attachments/...` route can then serve browser-executable same-origin content instead of a download-safe type.

   Suggested fix: do not infer executable MIME from user-controlled filenames. Store/use validated upload MIME, force `Content-Disposition: attachment` and `application/octet-stream` for non-image inline types, and reject or rewrite dangerous extensions for octet-stream uploads.

3. **Medium — Web UI regression** — `crates/ironclaw_gateway/static/js/core/history.js:220`

   Fresh history render passes `turn.user_attachments` into `addMessage`, but pagination prepends older user turns with `createMessageElement('user', turn.user_input)` only. Older paginated turns with persisted image attachments lose their attachment URLs and degrade back to file cards.

   Suggested fix: pass `userAttachments: Array.isArray(turn.user_attachments) ? turn.user_attachments : []` in pagination too, and add coverage for paginated history containing persisted user image attachments.

4. **Medium — Tests** — `tests/e2e/scenarios/test_chat.py:838`

   `test_gateway_attachment_limits_block_batched_uploads` creates a “too big” file at `5 * 1024 * 1024 + 1`, but the UI limit is now 7 MiB (`crates/ironclaw_gateway/static/js/surfaces/chat.js:674`) and i18n says 7 MB. This fixture will not trigger the oversized alert and can timeout.

   Suggested fix: build the oversized fixture from the production limit, or change it to `7 * 1024 * 1024 + 1`, so the E2E matches the current upload cap.

henrypark133 (approved):

What looks good:
- Attachment serving is behind gateway auth and re-checks owner identity plus canonical path containment before streaming bytes.
- The v1 image persistence and history hydration path now line up with `/api/attachments/...`.
- CI is green.

No verified findings.

Low-priority note:
- `crates/ironclaw_gateway/static/i18n/en.js:787`: `chat.imageTooBig` still says 5 MB while the unified attachment limit is now 7 MB. It looks unused after the unified flow, but it is stale user-facing copy if any path still references it.

Residual risk:
- Negative route/auth tests for `/api/attachments/...` are still thin, but I did not find a concrete bypass.

italic-jinxin:
Closed due to target branch stale. PR recreated from main, see https://github.com/nearai/ironclaw/pull/3531

Issue #3088, opened 2026-04-29
[Reborn] Wire production secrets/network boundary


serrrfirat:
## Parent

#2987

## Purpose

Wire the Reborn secrets/network substrate into production-grade secret brokerage and credential-injection boundaries.

#3077 landed the low-level substrate:

```text
ironclaw_secrets  -> scoped secret metadata, storage contract, one-shot leases
ironclaw_network  -> metadata-only network policy evaluation
```

This issue defines the production follow-through:

```text
durable encrypted secret repository
  -> scoped SecretBroker
  -> CredentialAccount metadata
  -> CredentialSession / brokered credential permit
  -> host-mediated credential injection through #3085 RuntimeHttpEgress
  -> production composition + no-exposure proof
```

#3085 owns the shared runtime HTTP egress service itself. This issue should depend on and plug into #3085 rather than duplicating egress mechanics.

## Locked decisions

1. **Split this issue into four implementation PRs.**
   - PR A: durable encrypted `SecretRepository` + scoped `SecretBroker`.
   - PR B: `CredentialAccount` metadata + `CredentialSession` / permit model.
   - PR C: brokered HTTP credential injection through #3085 `RuntimeHttpEgress`.
   - PR D: production composition + no-exposure vertical slice.

2. **Durable secret storage should reuse/bridge existing code where viable.**
   - Reuse/adapt current `src/secrets/crypto.rs`, `src/secrets/keychain.rs`, `src/secrets/store.rs`, `src/secrets/types.rs`, `src/config/secrets.rs`, and existing schemas where viable.
   - Add only the minimal scope/index/migration changes needed for tenant/user/agent/project isolation and typed Reborn metadata.
   - New Reborn-specific tables are allowed only with explicit justification.

3. **Production mode fails closed on direct raw `SecretMaterial` injection.**
   - Hosted/production profiles forbid direct raw secret injection into runtimes/adapters by default.
   - Direct material may be allowed only for explicit local-dev, migration, or first-party/system exceptions with strong sanitized audit/diagnostic metadata.

4. **First credential mechanism is native host-mediated injection.**
   - Build an in-process credential broker/session path used by `RuntimeHttpEgress`.
   - Runtimes receive credential/session metadata only.
   - Host egress injects actual secret material at the final boundary.
   - External proxy/sidecar support is deferred.

5. **First credential mapping support is V1 HTTP parity subset.**
   - Authorization header.
   - Custom headers.
   - Query parameters.
[...]

PR #3335, opened 2026-05-07
feat(reborn): port encrypted secrets store and add credential account sessions

serrrfirat:
## Summary

Implements the PR B slice from #3088:

- port the encrypted legacy secrets crypto/store contracts into `ironclaw_secrets`
- add scoped CredentialAccount metadata and CredentialSession model in `ironclaw_secrets`
- require explicit account selection and enforce scope, extension/provider, account status, and target policy before issuing sessions
- add strict credential target matching for scheme, host, optional port, path policy, and method
- add sanitized credential broker error taxonomy and stable public reason tokens
- keep credential sessions metadata-only with SecretHandle references and no raw material

## Tests

- cargo test -p ironclaw_secrets
- cargo clippy -p ironclaw_secrets --all-targets -- -D warnings
- cargo check --workspace --all-targets

Refs #3088 (PR B slice)

### DIFF
--- a/crates/ironclaw_secrets/src/lib.rs
+++ b/crates/ironclaw_secrets/src/lib.rs
@@ -324,7 +324,9 @@ impl InMemoryCredentialBroker {
             .ok_or_else(|| CredentialBrokerError::MissingCredential {
                 account_id: request.account_id.clone(),
             })?;
-        if account.scope != request.scope {
+        if CredentialAccountKey::new(&account.scope, &account.id)
+            != CredentialAccountKey::new(&request.scope, &request.account_id)
+        {
             return Err(CredentialBrokerError::CredentialScopeMismatch {
                 account_id: request.account_id,
             });
@@ -721,12 +723,9 @@ pub struct InMemorySecretStore {
 
 impl InMemorySecretStore {
     pub fn new() -> Self {
-        let crypto = Arc::new(
-            SecretsCrypto::new(SecretMaterial::from(
-                "0123456789abcdef0123456789abcdef".to_string(),
-            ))
-            .expect("test crypto key is a valid 32-byte literal"),
-        );
+        let crypto = Arc::new(SecretsCrypto::from_valid_master_key_literal(
+            "0123456789abcdef0123456789abcdef",
+        ));
         Self {
             inner: ScopedSecretsStoreAdapter::new(Arc::new(InMemorySecretsStore::new(crypto))),
         }
@@ -902,6 +901,38 @@ mod tests {
         let debug = format!("{session:?}");
         assert!(!debug.contains("sk-live-sentinel"));
         assert!(!debug.contains("token"));
+    }
+
+    #[test]
+    fn credential_session_creation_accepts_project_scoped_account_across_invocations() {
+        let broker = InMemoryCredentialBroker::new();
+        let account_scope = sample_scope("tenant-a", "user-a");
+        let request_scope = ResourceScope {
+            mission_id: Some(MissionId::new("mission-b").unwrap()),
+            thread_id: Some(ThreadId::new("thread-b").unwrap()),
+            invocation_id: InvocationId::new(),
+            ..account_scope.clone()
+        };
+        let account_id = CredentialAccountId::new("openai_prod").unwrap();
+        let secret_handle = SecretHandle::new("openai_key").unwrap();
+        broker
+            .put_account(sample_account(
+                account_scope,
+                account_id.clone(),
+                secret_handle.clone(),
+            ))
+            .unwrap();
+
+        let session = broker
+            .create_session(session_request(
+                request_scope.clone(),
+                account_id,
+                "https://api.example.com/v1/models",
+            ))
+            .unwrap();
+
+        assert_eq!(session.scope, request_scope);
+        assert_eq!(session.secret_handles, vec![secret_handle]);
     }
 
     #[test]
--- a/crates/ironclaw_secrets/src/lib.rs
+++ b/crates/ironclaw_secrets/src/lib.rs
@@ -14,6 +14,7 @@ use std::fmt;
 use std::sync::{Arc, Mutex, MutexGuard};
 
 use async_trait::async_trait;
+use chrono::{Duration, Utc};
 pub use crypto::SecretsCrypto;
 use ironclaw_host_api::{
     AgentId, CapabilityId, ExtensionId, InvocationId, MissionId, NetworkMethod, ProjectId,
@@ -24,9 +25,13 @@ pub use legacy_store::{
     SecretError, SecretRef, SecretsStore,
 };
 pub use secrecy::SecretString as SecretMaterial;
+use serde::{Deserialize, Serialize};
 use serde_json::Value;
 use thiserror::Error;
 use uuid::Uuid;
+
+const CREDENTIAL_ID_MAX_LEN: usize = 128;
+const DEFAULT_SECRET_LEASE_TTL_SECONDS: i64 = 300;
 
 /// Opaque identifier for a one-shot secret lease.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
@@ -91,6 +96,12 @@ pub enum SecretStoreError {
     LeaseConsumed { lease_id: SecretLeaseId },
     #[error("secret lease {lease_id} was revoked")]
     LeaseRevoked { lease_id: SecretLeaseId },
+    #[error("secret lease {lease_id} expired")]
+    LeaseExpired { lease_id: SecretLeaseId },
+    #[error("secret expired")]
+    SecretExpired,
+    #[error("secret backend is misconfigured: {reason}")]
+    BackendMisconfigured { reason: String },
     #[error("secret store state is unavailable: {reason}")]
     StoreUnavailable { reason: String },
 }
@@ -102,6 +113,9 @@ impl SecretStoreError {
             Self::UnknownLease { .. } => "MissingCredential",
             Self::LeaseConsumed { .. } => "CredentialExpired",
             Self::LeaseRevoked { .. } => "CredentialRevoked",
+            Self::LeaseExpired { .. } => "CredentialExpired",
+            Self::SecretExpired => "CredentialExpired",
+            Self::BackendMisconfigured { .. } => "BackendMisconfigured",
             Self::StoreUnavailable { .. } => "BackendUnavailable",
         }
     }
@@ -123,7 +137,31 @@ impl SecretStoreError {
     }
 }
 
-#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+#[derive(Clone, PartialEq, Serialize, Deserialize)]
+#[serde(transparent)]
+pub struct RedactedJson(Value);
+
+impl RedactedJson {
+    pub fn new(value: Value) -> Self {
+        Self(value)
+    }
+
+    pub fn as_value(&self) -> &Value {
+        &self.0
+    }
+
+    pub fn into_value(self) -> Value {
+        self.0
+    }
+}
+
+impl fmt::Debug for RedactedJson {
+    fn fmt(&self, formatter: &mut fmt::Formatter<'_>) -> fmt::Result {
+        formatter.write_str("[REDACTED_JSON]")
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize)]
 pub struct CredentialAccountId(String);
 
 impl CredentialAccountId {
@@ -135,6 +173,40 @@ impl CredentialAccountId {
 
     pub fn as_str(&self) -> &str {
         &self.0
+    }
+
+    pub fn into_string(self) -> String {
+        self.0
+    }
+}
+
+impl AsRef<str> for CredentialAccountId {
+    fn as_ref(&self) -> &str {
+        self.as_str()
+    }
+}
+
+impl TryFrom<String> for CredentialAccountId {
+    type Error = CredentialBrokerError;
+
+    fn try_from(value: String) -> Result<Self, Self::Error> {
+        Self::new(value)
+    }
+}
+
+impl From<CredentialAccountId> for String {
+    fn from(value: CredentialAccountId) -> Self {
+        value.0
+    }
+}
+
+impl<'de> Deserialize<'de> for CredentialAccountId {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        let value = String::deserialize(deserializer)?;
+        Self::new(value).map_err(serde::de::Error::custom)
     }
 }
 
@@ -156,6 +228,12 @@ impl CredentialSessionId {
 impl Default for CredentialSessionId {
     fn default() -> Self {
         Self::new()
+    }
+}
+
+impl fmt::Display for CredentialSessionId {
+    fn fmt(&self, formatter: &mut fmt::Formatter<'_>) -> fmt::Result {
+        self.0.fmt(formatter)
     }
 }
 
@@ -186,6 +264,12 @@ impl CredentialTargetPolicy {
         let Ok(parsed) = url::Url::parse(url) else {
             return false;
         };
+        if !parsed.username().is_empty() || parsed.password().is_some() {
+            return false;
+        }
+        if raw_url_path(url).is_some_and(path_has_encoded_traversal) {
+            return false;
+        }
         if self.scheme != parsed.scheme() {
             return false;
         }
@@ -220,22 +304,55 @@ pub struct CredentialAccount {
     pub status: CredentialAccountStatus,
     pub secret_handles: Vec<SecretHandle>,
     pub allowed_targets: Vec<CredentialTargetPolicy>,
-    pub redacted_metadata: Value,
+    pub redacted_metadata: RedactedJson,
     pub updated_at: Timestamp,
 }
 
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct CredentialSession {
-    pub scope: ResourceScope,
-    pub invocation_id: InvocationId,
-    pub capability_id: CapabilityId,
-    pub extension_id: ExtensionId,
-    pub account_id: CredentialAccountId,
-    pub secret_handles: Vec<SecretHandle>,
-    pub allowed_targets: Vec<CredentialTargetPolicy>,
-    pub expires_at: Option<Timestamp>,
-    pub max_uses: Option<u64>,
-    pub correlation_id: CredentialSessionId,
+    scope: ResourceScope,
+    invocation_id: InvocationId,
+    capability_id: CapabilityId,
+    extension_id: ExtensionId,
+    account_id: CredentialAccountId,
+    secret_handles: Vec<SecretHandle>,
+    allowed_targets: Vec<CredentialTargetPolicy>,
+    expires_at: Option<Timestamp>,
+    max_uses: Option<u64>,
+    correlation_id: CredentialSessionId,
+}
+
+impl CredentialSession {
+    pub fn scope(&self) -> &ResourceScope {
+        &self.scope
+    }
+    pub fn invocation_id(&self) -> InvocationId {
+        self.invocation_id
+    }
+    pub fn capability_id(&self) -> &CapabilityId {
+        &self.capability_id
+    }
+    pub fn extension_id(&self) -> &ExtensionId {
+        &self.extension_id
+    }
+    pub fn account_id(&self) -> &CredentialAccountId {
+        &self.account_id
+    }
+    pub fn secret_handles(&self) -> &[SecretHandle] {
+        &self.secret_handles
+    }
+    pub fn allowed_targets(&self) -> &[CredentialTargetPolicy] {
+        &self.allowed_targets
+    }
+    pub fn expires_at(&self) -> Option<Timestamp> {
+        self.expires_at
+    }
+    pub fn max_uses(&self) -> Option<u64> {
+        self.max_uses
+    }
+    pub fn correlation_id(&self) -> CredentialSessionId {
+        self.correlation_id
+    }
 }
 
 #[derive(Debug, Clone, PartialEq, Eq, Error)]
@@ -246,6 +363,16 @@ pub enum CredentialBrokerError {
     MissingCredential { account_id: CredentialAccountId },
     #[error("credential account {account_id} does not match caller scope")]
     CredentialScopeMismatch { account_id: CredentialAccountId },
+    #[error("credential session request invocation does not match caller scope for {account_id}")]
+    CredentialInvocationMismatch { account_id: CredentialAccountId },
+    #[error("credential broker state is unavailable: {reason}")]
+    BrokerUnavailable { reason: String },
+    #[error("credential session {session_id} is unknown")]
+    UnknownSession { session_id: CredentialSessionId },
+    #[error("credential session {session_id} is expired")]
+    SessionExpired { session_id: CredentialSessionId },
+    #[error("credential session {session_id} has no uses remaining")]
+    SessionUseLimitExceeded { session_id: CredentialSessionId },
     #[error("credential account {account_id} is expired")]
     CredentialExpired { account_id: CredentialAccountId },
     #[error("credential account {account_id} is revoked")]
@@ -262,6 +389,11 @@ impl CredentialBrokerError {
             Self::InvalidAccountId { .. } => "MissingCredential",
             Self::MissingCredential { .. } => "MissingCredential",
             Self::CredentialScopeMismatch { .. } => "CredentialScopeMismatch",
+            Self::CredentialInvocationMismatch { .. } => "CredentialScopeMismatch",
+            Self::BrokerUnavailable { .. } => "BackendUnavailable",
+            Self::UnknownSession { .. } => "MissingCredential",
+            Self::SessionExpired { .. } => "CredentialExpired",
+            Self::SessionUseLimitExceeded { .. } => "CredentialExpired",
             Self::CredentialExpired { .. } => "CredentialExpired",
             Self::CredentialRevoked { .. } => "CredentialRevoked",
             Self::CredentialExtensionMismatch { .. } => "CredentialPolicyMismatch",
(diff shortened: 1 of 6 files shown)


zmanian:
## Code review

Read the full diff (1553 lines), the existing `src/secrets/crypto.rs`, and the `host_api` types this PR depends on.

### Overview

PR B slice of #3088. Lands three things in `crates/ironclaw_secrets`:

1. **Port** of the existing `src/secrets/{crypto.rs, types.rs}` AES-256-GCM/HKDF-SHA256 implementation into the extracted crate (`crypto.rs`, `legacy_store.rs`).
2. **New `CredentialAccount` / `CredentialSession` model** with an `InMemoryCredentialBroker` that enforces scope, extension, status, and target (scheme/host/port/path/method) before issuing a session — sessions carry `SecretHandle`s only, never raw material.
3. **Refactors `InMemorySecretStore`** to delegate through a new `ScopedSecretsStoreAdapter<S: SecretsStore>`, so the Reborn lease boundary now sits on top of the encrypted legacy store rather than a parallel raw-material implementation.

### Strengths

- **No parallel crypto.** The port is byte-faithful to `src/secrets/crypto.rs` (same algorithm, same `near-agent-secrets-v1` HKDF info string, same constants). This is the most important property and the PR gets it right.
- **Sessions are metadata-only.** `CredentialSession` carries `Vec<SecretHandle>`, never `SecretMaterial`. The `redacts_material` test asserts `Debug` doesn't leak.
- **Scope isolation is tested for delimiter ambiguity.** `scoped_legacy_user_id_uses_unambiguous_json_encoding` covers the case where `tenant=a;agent=-` would otherwise collide with a tenant `tenant-a` plus a `-` agent.
- **Target policy testing is thorough**: scheme, host, port (incl. default-port resolution via `port_or_known_default()`), method, and prefix-tampering (`/v1/` does not match `/v1-evil/...`).
- **Stable reason tokens** are extracted into `stable_reason()` on both `SecretStoreError` and `CredentialBrokerError` — good for cross-process logging without leaking diagnostic detail.

### Issues

#### High — `expect()` in non-test production code path

```rust
// crates/ironclaw_secrets/src/lib.rs — InMemorySecretStore::new
pub fn new() -> Self {
    let crypto = Arc::new(
        SecretsCrypto::new(SecretMaterial::from(
            "0123456789abcdef0123456789abcdef".to_string(),
        ))
        .expect("test crypto key is a valid 32-byte literal"),
    );
    ...
}
```

Two problems:

1. `CLAUDE.md` is explicit: *"No `.unwrap()` or `.expect()` in production code (tests are fine)"*. This type is `pub` and not `#[cfg(test)]`-gated, so the `expect` is reachable from non-test callers.
2. **A hardcoded test crypto key is baked into a `pub` type whose docstring only says "contract tests and non-durable demos."** Nothing in the type signature or naming prevents a real call site (or a downstream consumer of the crate) from constructing one. At minimum, gate behind `#[cfg(any(test, feature = "test-support"))]` or rename to `TestOnlyInMemorySecretStore`. The previous version used a real per-instance state machine that didn't carry an embedded key, so this is a regression in API safety.

#### High — `CredentialAccountId` does not follow the project's newtype template

`.claude/rules/types.md` defines a canonical pattern that this PR adopts only partially.

```rust
#[derive(Debug, Clone, PartialEq, Eq, Hash)]
pub struct CredentialAccountId(String);
```

Missing vs. the template:

- No `Serialize` / `Deserialize` derive (and no `#[serde(try_from = "String")]` to enforce wire-side validation). Every other `host_api` identity type (`SecretHandle`, `ExtensionId`, `CapabilityId`, `TenantId`, `UserId`) is serializable, and a `CredentialAccount` is going to need to round-trip through some persistence boundary very soon.
- No `as_ref`, `into_inner`, `TryFrom<String>`, `From<CredentialAccountId> for String`.
- No length cap in `validate_credential_id` — only a charset filter. Per `safety-and-sandbox.md` ("User-controlled inputs must not grow unbounded"), a `pub` newtype constructor that accepts arbitrary-length strings is a foot-gun. Add a small upper bound (e.g. 128 chars).

Also a cosmetic bug in the validation error message:

```rust
reason: format!("{kind} must contain only ASCII letters, digits, '-' or '_"),
```

Unbalanced single quote on `'_`. Will appear in user/log output.

#### Medium — `redacted_metadata: Value` on `CredentialAccount` is a foot-gun

```rust
pub redacted_metadata: Value,
```

The field name is a *promise to the reader*, not a type-system enforcement. Anyone constructing this struct can put plaintext token tail-bytes in there and `Debug` will happily print them. The rest of this PR carefully wraps decrypted material in `DecryptedSecret` with a `Debug` impl that prints `[REDACTED, N bytes]`; this field undermines that discipline.

Either wrap in a `RedactedJson(Value)` newtype with a custom `Debug`, or move the field behind a method (`fn redacted_metadata(&self) -> &Value`) and gate writes on a builder that scrubs known sensitive keys.

The `Debug` test on `CredentialSession` doesn't help here because sessions don't carry the metadata — but accounts do, and accounts will be the thing the broker logs.

#### Medium — Mutex poisoning incorrectly mapped to `MissingCredential`

```rust
self.accounts
    .lock()
    .map_err(|_| CredentialBrokerError::MissingCredential {
        account_id: account.id.clone(),
    })?
```

A poisoned mutex means a previous holder panicked — it is not a missing credential, and surfacing it as one will mislead operators, taint the `stable_reason()` token (`"MissingCredential"`), and almost certainly cause wrong-direction debugging next time it fires. Add a `BrokerUnavailable { reason: String }` variant (mirroring `SecretStoreError::StoreUnavailable`) and map there.

#### Medium — `map_legacy_secret_error` collapses semantically distinct errors

`SecretError::NotFound`, `Expired`, and `InvalidMasterKey` all become `StoreUnavailable { reason: ... }`. `metadata()` already pre-handles `NotFound` to return `Ok(None)`, but **`lease_once` would also benefit** from explicit `Expired → CredentialExpired`-style mapping; right now an expired legacy secret looks identical to a back-end outage to the caller. Same for `InvalidMasterKey`, which is a misconfiguration, not a transient store failure.

#### Low — Test asserts that don't actually exercise their claims

```rust
let debug = format!("{session:?}");
assert!(!debug.contains("sk-live-sentinel"));
assert!(!debug.contains("token"));
```

The session struct only carries `SecretHandle`s — `sk-live-sentinel` was never going to appear in `Debug` regardless of whether the redaction works. Strengthen by also constructing a `CredentialAccount` with the real material in `redacted_metadata` (post-fix from the previous point), and assert against `Debug` of the account, the broker, and the session in one shot.

Also missing tests:

- `CredentialAccountId::new` rejection of `""`, `"a/b"`, `"a b"`, oversized strings.
- `Expired` account → `CredentialExpired`. Only `Revoked` is covered.
- A `stable_reason()` snapshot test — the docstring calls these "stable public reason tokens", which is exactly the contract that a snapshot test exists to lock in.

#### Low — `validate_credential_id` is over-generic

It takes a `kind: &'static str` parameter for use in error formatting, but has exactly one caller (`CredentialAccountId::new` passing `"credential_account"`). Either inline it or wait until the second caller materializes. Per CLAUDE.md *"Don't add features… beyond what the task requires."*

### PR scope and naming

PR title says *"add credential account sessions"* (a feature add), but the diff also **rewrites `InMemorySecretStore` to delegate through a new encrypted adapter** and ports two source files of legacy crypto/store code. That is a genuine architectural change layered on top of the feature. Per `review-discipline.md`, either retitle (e.g. `feat(reborn): port encrypted secrets store + add credential account sessions`) or call the second axis out in the PR body.

Also, the PR claims to close #3088 (PR B slice). If A/C/D slices are still open, `Closes` will close the whole issue on merge — use `Refs #3088` until the final slice lands.

- Consider whether `CredentialTargetPolicy::matches` should reject URLs containing **userinfo** (`https://attacker:pwd@api.example.com/...`). `url::Url::host_str()` returns the bare host, so the policy currently matches such a URL. Rejecting userinfo-bearing URLs at this layer is cheap defense-in-depth.
- `path_matches_prefix` correctly handles trailing slashes, but a one-line doc comment naming the `/v1/` vs `/v1-evil/` invariant it defends would age well.
- The legacy `InMemorySecretsStore` re-implements `consume_if_matches` once on the `SecretsStore` trait default and again in the impl. The trait default isn't reachable here. Decide which is canonical and delete the other to avoid drift.

---

### Architectural follow-up — credential consent

Worth flagging for the next slice rather than this one: PR #3335 strengthens the **policy** layer (target/method/scope matching, scoped sessions, stable error tokens), but it does not introduce a **consent step**. The natural place to add per-use consent is `CredentialBroker::create_session` — emit a `PendingGate` keyed on `(scope, account_id, capability_id)` and require user resolution before returning the session, with an `AlwaysAllow` cache for trusted `(capability, account)` pairs. The new `CredentialSessionRequest` is exactly the right shape to display in the gate card.

Two related properties that would also be worth pinning down before this lands in a real flow:

- **Trusted path.** The existing gate mechanism (`request_id` UUID, atomic `take_verified`, typed `AppEvent::GateRequired` with static description and pre-redacted parameters) is solid against forgery and replay, but the consent path is in-band — same browser tab / terminal as the agent. A compromised client surface sits behind that check. An out-of-band confirmation channel (push to paired device, WebAuthn passkey ceremony returning a signed approval scoped to `request_id` + `capability_id` + `account_id`) is the natural slice if the threat model includes a compromised client.
- **`SharedCredentialRegistry` auto-approval.** The shortcut at `src/gate/approval.rs:71-73` means a tool that is policy-allowed for a host's credentials never produces a gate. Worth deciding whether the new `CredentialSession` path should preserve that shortcut or always go through a gate (with the user's prior `AlwaysAllow` consulted explicitly rather than implicitly).

### Summary

The cryptography is preserved, the session model matches the design intent, and the policy matcher is well-tested. **Two changes needed before merge** are the production `expect()` / hardcoded-key issue on `InMemorySecretStore::new()` and the missing serde/length-cap on `CredentialAccountId`. The mutex-poisoning mapping and `redacted_metadata` typing are next-most-important. Everything else is polish.

---

crates/ironclaw_secrets/src/lib.rs
```diff
@@ -104,6 +123,302 @@ impl SecretStoreError {
...
+                })?;
+        let account = accounts
+            .get(&CredentialAccountKey::new(
+                &request.scope,
+                &request.account_id,
+            ))
+            .ok_or_else(|| CredentialBrokerError::MissingCredential {
+                account_id: request.account_id.clone(),
+            })?;
+        if account.scope != request.scope {
```
serrrfirat: **High Severity**

Credential accounts are scoped too narrowly to survive normal turn/invocation changes.

`CredentialAccountKey` only keys accounts by tenant/user/agent/project, so a later request in the same project but with a different `mission_id`, `thread_id`, or `invocation_id` can retrieve the account and then fail this full `ResourceScope` equality check with `CredentialScopeMismatch`. That makes setup-time or earlier-turn accounts unusable in ordinary later invocations.

Use a dedicated credential-account scope (or compare only the intended tenant/user/agent/project account scope) and add a regression test where the same tenant/user/agent/project but a different invocation/thread still issues a session.

crates/ironclaw_secrets/src/lib.rs
```diff
@@ -238,78 +506,121 @@ impl SecretLeaseKey {
...
             status: SecretLeaseStatus::Active,
         };
-        let record = LeaseRecord {
-            lease: lease.clone(),
-            material: Some(secret.material.clone()),
-        };
-        state
-            .leases
-            .insert(SecretLeaseKey::new(scope, lease.id), record);
+        self.lock_leases()?.insert(
```
serrrfirat, line 937: **Medium Severity**

Active leases retain cloned plaintext secret material indefinitely.

`lease_once` decrypts the secret and stores a cloned `SecretMaterial` in `LeaseRecord`, but active leases have no TTL, expiry check, or pruning. If a request crashes, times out, or never reaches `consume`/`revoke`, plaintext remains resident and abandoned lease records accumulate for the lifetime of the adapter.

Add lease expiry/pruning plus consume-time expiry checks, or avoid long-lived plaintext retention by deferring decryption until consume. Add a test proving abandoned active leases expire and drop material.

crates/ironclaw_secrets/src/lib.rs
```diff
@@ -238,78 +506,121 @@ impl SecretLeaseKey {
...
+            Ok(_) => Ok(Some(SecretMetadata {
+                scope: scope.clone(),
+                handle: handle.clone(),
+            })),
+            Err(SecretError::NotFound(_)) => Ok(None),
+            Err(error) => Err(map_legacy_secret_error(error)),
+        }
     }
 
     async fn lease_once(
```
serrrfirat, line 914: **Medium Severity**

Reborn consumes bypass legacy secret usage recording.

The adapter retrieves/decrypts legacy secrets and returns material on `consume`, but it never calls the legacy store’s `record_usage` hook. A DB-backed store would therefore leave `last_used_at`/`usage_count` stale for successful Reborn consumes, hiding real credential use from audit and rotation logic.

Carry the legacy secret id in the lease record and call `record_usage` exactly once on successful consume. Add a mock-store test that asserts usage is recorded on consume and not on revoke or failed consume.

crates/ironclaw_secrets/src/lib.rs
```diff
@@ -369,22 +678,358 @@ impl SecretStore for InMemorySecretStore {
...
             .collect())
     }
 }
 
+fn map_legacy_secret_error(error: SecretError) -> SecretStoreError {
+    match error {
+        SecretError::NotFound(name) => SecretStoreError::StoreUnavailable {
+            reason: format!("legacy secret missing: {name}"),
+        },
+        SecretError::Expired => SecretStoreError::StoreUnavailable {
```
serrrfirat: **Medium Severity**

Expired legacy secrets are reported as backend outages.

`map_legacy_secret_error` converts `SecretError::Expired` into `StoreUnavailable`, whose stable reason is `BackendUnavailable`. A legitimately expired credential can then drive retry/page-infra behavior instead of credential refresh or re-authentication.

Preserve the expired-credential state at this boundary, for example with a dedicated `SecretStoreError` variant/stable reason, and test `metadata`/`lease_once` behavior for expired legacy secrets through the adapter.

crates/ironclaw_secrets/src/lib.rs
```diff
@@ -104,6 +123,302 @@ impl SecretStoreError {
...
+        Ok(())
+    }
+
+    pub fn create_session(
+        &self,
+        request: CredentialSessionRequest,
+    ) -> Result<CredentialSession, CredentialBrokerError> {
+        let accounts =
+            self.accounts
+                .lock()
```
serrrfirat, line 464: **Medium Severity**

Credential-broker lock poisoning is reported as missing credentials.

Both account insertion and session creation map a poisoned `accounts` mutex to `MissingCredential`. After a panic while holding the lock, every later request would falsely blame a missing account and emit the wrong stable reason instead of exposing broker/backend unavailability.

Add a broker-unavailable error variant/stable reason and map lock poisoning to it in both `put_account` and `create_session`.

crates/ironclaw_secrets/src/lib.rs
```diff
@@ -104,6 +123,302 @@ impl SecretStoreError {
...
+            return false;
+        }
+        match &self.path {
+            CredentialPathPolicy::Exact(path) => parsed.path() == path,
+            CredentialPathPolicy::Prefix(prefix) => path_matches_prefix(parsed.path(), prefix),
+        }
+    }
+}
+
+#[derive(Debug, Clone, PartialEq)]
```
serrrfirat, line 303: **Medium Severity**

`CredentialAccount` debug output can expose caller-supplied metadata.

The type derives `Debug` while carrying `pub redacted_metadata: Value`. If a caller accidentally stores a refresh token, bearer token fragment, or provider payload in that JSON field, any debug log of the account will print it verbatim despite the field name implying safety.

Wrap the metadata in a redacting newtype or implement custom `Debug` for `CredentialAccount`, and add a sentinel test that debug output does not contain secret-like metadata values.

crates/ironclaw_secrets/src/lib.rs
```diff
@@ -104,6 +123,302 @@ impl SecretStoreError {
...
+            .is_some_and(|port| Some(port) != parsed.port_or_known_default())
+        {
+            return false;
+        }
+        if !self.methods.iter().any(|allowed| allowed == method) {
+            return false;
+        }
+        match &self.path {
+            CredentialPathPolicy::Exact(path) => parsed.path() == path,
+            CredentialPathPolicy::Prefix(prefix) => path_matches_prefix(parsed.path(), prefix),
```
serrrfirat, line 298: **High Severity**

The new target path matcher drops the legacy encoded-traversal protections.

`path_matches_prefix()` only checks raw slash boundaries. The hardened legacy matcher in `src/secrets/types.rs` also rejects dot-segments and percent-encoded slashes before matching. With the new matcher, a policy like `Prefix("/v1")` can match URLs such as `/v1/%2e%2e%2fadmin` or `/v1/%2e/admin`; backends or proxies that normalize those forms before routing can then receive credentials for endpoints outside the declared path scope.

Reuse the hardened legacy path matcher (or port its dot-segment/encoded-slash rejection) and add regression tests for encoded `.`/`..` segments and `%2f` traversal-smuggling cases.

crates/ironclaw_secrets/src/lib.rs
```diff
@@ -104,6 +123,302 @@ impl SecretStoreError {
...
+    pub provider_or_extension_id: ExtensionId,
+    pub label: String,
+    pub status: CredentialAccountStatus,
+    pub secret_handles: Vec<SecretHandle>,
+    pub allowed_targets: Vec<CredentialTargetPolicy>,
+    pub redacted_metadata: Value,
+    pub updated_at: Timestamp,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
```
serrrfirat, line 316: **Medium Severity**

Credential sessions are authority-shaped metadata without broker-backed enforcement.

`CredentialSession` is a public, cloneable struct, and `InMemoryCredentialBroker` stores accounts only, not issued sessions. As a result, `expires_at` and `max_uses` cannot be enforced by this broker and downstream code has no way to distinguish a broker-issued session from a caller-fabricated one.

Make sessions opaque/store-backed (similar to other lease stores), or provide a validation/consume path that enforces expiry and use limits. Add tests showing expired or over-used sessions are rejected through the broker boundary.

crates/ironclaw_secrets/src/lib.rs
```diff
@@ -104,6 +123,302 @@ impl SecretStoreError {
...
+            .iter()
+            .any(|target| target.matches(&request.method, &request.url))
+        {
+            return Err(CredentialBrokerError::CredentialPolicyMismatch {
+                account_id: request.account_id,
+            });
+        }
+        Ok(CredentialSession {
+            scope: request.scope,
+            invocation_id: request.invocation_id,
```
serrrfirat, line 512: **Medium Severity**

Session requests can produce inconsistent invocation identities.

`CredentialSessionRequest` carries invocation identity twice: once inside `scope.invocation_id` and once as the standalone `invocation_id`. `create_session()` never checks that they match and copies both into the returned session, so a caller can mint a session whose scope says invocation A while the session field says invocation B. Later audit or staging code can make the wrong decision depending on which field it trusts.

Reject requests where `request.invocation_id != request.scope.invocation_id`, or remove the redundant field and derive it from the scope. Add a regression test for mismatched invocation ids.

crates/ironclaw_secrets/src/lib.rs
```diff
@@ -104,6 +123,304 @@ impl SecretStoreError {
...
+            .is_some_and(|port| Some(port) != parsed.port_or_known_default())
+        {
+            return false;
+        }
+        if !self.methods.iter().any(|allowed| allowed == method) {
+            return false;
+        }
+        match &self.path {
+            CredentialPathPolicy::Exact(path) => parsed.path() == path,
+            CredentialPathPolicy::Prefix(prefix) => path_matches_prefix(parsed.path(), prefix),
```
serrrfirat, line 298: **High Severity**

The target path matcher still drops the hardened legacy traversal protections.

`path_matches_prefix()` only enforces a raw slash boundary. The legacy matcher in `src/secrets/types.rs` also rejects dot-segments and percent-encoded slashes before matching. With this new matcher, a policy like `Prefix("/v1")` can still match `/v1/%2e%2e%2fadmin` or `/v1/%2e/admin`; proxies/backends that decode these forms before routing may receive credentials for endpoints outside the declared path scope.

Reuse or port the hardened legacy matcher, including encoded dot-segment and encoded-slash rejection, and add regression tests for `%2e`, `%2e%2e`, and `%2f` traversal-smuggling cases.

crates/ironclaw_secrets/src/lib.rs
```diff
@@ -238,78 +508,121 @@ impl SecretLeaseKey {
...
             status: SecretLeaseStatus::Active,
         };
-        let record = LeaseRecord {
-            lease: lease.clone(),
-            material: Some(secret.material.clone()),
-        };
-        state
-            .leases
-            .insert(SecretLeaseKey::new(scope, lease.id), record);
+        self.lock_leases()?.insert(
```
serrrfirat, line 937: **High Severity**

Active leases keep plaintext secret material indefinitely.

`lease_once()` decrypts the secret and stores cloned `SecretMaterial` in the lease record, but leases have no expiry, pruning, or cleanup path beyond explicit `consume`/`revoke`. If a request crashes, times out, or simply forgets to consume/revoke, the secret stays valid and plaintext remains resident for the process lifetime; consumed/revoked records also accumulate.

Add lease TTL/pruning plus consume-time expiry checks, or avoid retaining plaintext by deferring decryption until consume. Add tests proving abandoned leases expire and drop material.

crates/ironclaw_secrets/src/lib.rs
```diff
@@ -318,10 +631,9 @@ impl SecretStore for InMemorySecretStore {
         scope: &ResourceScope,
         lease_id: SecretLeaseId,
     ) -> Result<SecretMaterial, SecretStoreError> {
-        let mut state = self.lock_state()?;
+        let mut leases = self.lock_leases()?;
```
serrrfirat: **Medium Severity**

Reborn secret consumes still bypass legacy usage accounting.

The adapter returns material on successful `consume`, but it never calls the legacy store’s `record_usage` hook. A DB-backed legacy store would leave `last_used_at` and `usage_count` stale, dropping audit/rotation signals for Reborn secret access.

Carry the legacy secret id in the lease record and call `record_usage` exactly once on successful consume. Add a mock-store test covering consume vs revoke/failure.

crates/ironclaw_secrets/src/lib.rs
```diff
@@ -104,6 +123,304 @@ impl SecretStoreError {
...
+    pub label: String,
+    pub status: CredentialAccountStatus,
+    pub secret_handles: Vec<SecretHandle>,
+    pub allowed_targets: Vec<CredentialTargetPolicy>,
+    pub redacted_metadata: Value,
+    pub updated_at: Timestamp,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct CredentialSession {
```
serrrfirat, line 317: **Medium Severity**

Credential sessions remain advisory metadata rather than broker-enforced authority.

`CredentialSession` is a public cloneable struct, and the broker does not store or validate issued sessions. Downstream code cannot prove a session was broker-issued, and `expires_at`/`max_uses` cannot be enforced by the broker; a caller can fabricate a session with arbitrary handles, targets, or unlimited lifetime.

Make sessions opaque/store-backed, or add a broker validation/redemption API keyed by `CredentialSessionId` that enforces expiry and use limits. Keep authority-bearing fields private where possible.

crates/ironclaw_secrets/src/lib.rs
```diff
@@ -104,6 +123,304 @@ impl SecretStoreError {
...
+#[derive(Debug, Clone, PartialEq)]
+pub struct CredentialAccount {
+    pub scope: ResourceScope,
+    pub id: CredentialAccountId,
+    pub provider_or_extension_id: ExtensionId,
+    pub label: String,
+    pub status: CredentialAccountStatus,
+    pub secret_handles: Vec<SecretHandle>,
+    pub allowed_targets: Vec<CredentialTargetPolicy>,
+    pub redacted_metadata: Value,
```
serrrfirat: **Medium Severity**

`CredentialAccount` debug output can still expose caller-supplied metadata.

The type derives `Debug` while carrying `pub redacted_metadata: Value`. If an adapter accidentally stores a refresh token, bearer token fragment, or provider payload in that JSON, any debug log of the account prints it verbatim even though the field name implies safety.

Wrap the metadata in a redacting newtype or implement custom `Debug` for `CredentialAccount`, and add a sentinel test that debug output does not contain secret-like metadata values.

crates/ironclaw_secrets/src/lib.rs
```diff
@@ -369,22 +680,387 @@ impl SecretStore for InMemorySecretStore {
...
             .collect())
     }
 }
 
+fn map_legacy_secret_error(error: SecretError) -> SecretStoreError {
+    match error {
+        SecretError::NotFound(name) => SecretStoreError::StoreUnavailable {
+            reason: format!("legacy secret missing: {name}"),
+        },
+        SecretError::Expired => SecretStoreError::StoreUnavailable {
```
serrrfirat: **Medium Severity**

Actionable legacy secret states are still collapsed into backend outages.

`SecretError::Expired` and `InvalidMasterKey` both map to `StoreUnavailable`, whose stable reason is `BackendUnavailable`. An expired migrated credential can therefore trigger retry/page-infra behavior instead of credential refresh or re-authentication; invalid-key configuration also loses a distinct remediation path.

Preserve actionable error classes in `SecretStoreError` (at least expired vs misconfigured vs unavailable) and cover `metadata()`/`lease_once()` adapter behavior with tests.

crates/ironclaw_secrets/src/lib.rs
```diff
@@ -104,6 +123,304 @@ impl SecretStoreError {
...
+        Ok(())
+    }
+
+    pub fn create_session(
+        &self,
+        request: CredentialSessionRequest,
+    ) -> Result<CredentialSession, CredentialBrokerError> {
+        let accounts =
+            self.accounts
+                .lock()
```
serrrfirat, line 464: **Medium Severity**

Credential-broker state failures are reported as missing credentials.

A poisoned `accounts` mutex is mapped to `MissingCredential` in session creation, and `put_account()` uses the same pattern. After a panic while holding the lock, every later operation would falsely blame missing credentials and emit the wrong stable reason instead of exposing broker/backend unavailability.

Add a broker-unavailable error variant/stable reason and map lock poisoning to that variant in both account insertion and session creation.

Issue #3094, opened 2026-04-29
[Reborn] Add approval/auth interaction services


serrrfirat:
## Parent

#2987

## What to build

Add Reborn `ApprovalInteractionService` and `AuthInteractionService` that translate blocked run-state into adapter/UI-safe interaction flows and route user decisions back into Reborn approval/auth resolution.

These services should bridge UX surfaces without becoming alternate approval/auth side channels. The source of truth remains Reborn run-state, approval records, leases, and auth-flow state.

## Acceptance criteria

- [ ] Approval interaction service can list/render pending approvals from scoped run-state/approval records without exposing raw input, approval reasons, invocation fingerprints, lease IDs, host paths, secrets, or runtime output.
- [ ] Approval decisions resolve through Reborn approval/run-state/capability resume paths and cannot trigger side effects directly.
- [ ] Auth interaction service can list/render auth-required states and resume/cancel auth flows through a typed auth-flow manager boundary.
- [ ] Adapter-facing responses are stable, redacted, and tenant/user/agent/project scoped.
- [ ] Tests cover approval approve/deny, missing approval, stale approval, cross-scope denial, auth required, auth failure, cancel, resume, and no-exposure sentinels.
- [ ] Existing web/CLI/channel interaction semantics are mapped with compatibility/deferred-drift notes.

## Blocked by

- Blocked by #3013 for blocked run-state coordination.
- Coordinate with #3016 for adapter/loop host interaction boundaries.
- Coordinate with #3068 for credential/auth parity where HTTP credentials are involved.

serrrfirat:
Related design issue created: #3107.

This reinforces that approval/auth UX should be host-owned and loop-independent.

When any `AgentLoopDriver` receives a blocked capability outcome:

```text
CapabilityOutcome::Blocked(Approval/Auth/Resource)
  -> driver checkpoints and returns
  -> TurnCoordinator records blocked state
  -> ApprovalInteractionService / AuthInteractionService owns user interaction
  -> same driver/checkpoint resumes after gate resolution
```

So lightweight, CodeAct, EngineCompat, or ClaudeCode-style drivers do not implement OAuth, approval UI, or alternate auth side channels. They only see structured blocked outcomes and resume signals through `AgentLoopHost` / `TurnCoordinator`.

serrrfirat:
Split the approval-specific slice into #3889 so this parent can keep tracking the broader interaction-services work without mixing it with the newer Reborn auth contract stack.

[...]

PR #3351, opened 2026-05-07
feat(reborn): add product adapter contract

serrrfirat:
Split from #3316 (original PR by @nickpismenkov). This is PR 1/7 in the ProductAdapter stack.

## Summary
- Add `ironclaw_product_adapters` core contract crate.
- Define inbound/outbound DTOs, external refs, sealed protocol auth evidence, egress/delivery traits, workflow/projection facades, redaction wrappers, and test fakes.
- Keep this slice Telegram-agnostic and runtime-agnostic.

## Stack
1. **This PR**: core ProductAdapter contract
2. Host auth/egress primitives
3. Native runner glue
4. Telegram payload normalization
5. Telegram ProductAdapter tracer bullet
6. Default-off Telegram v2 config guard
7. Contract docs

## Authorship

## Verification
- `cargo test -p ironclaw_product_adapters --offline --quiet` — 35 tests passed.


crates/ironclaw_product_adapters/src/inbound.rs
```diff
@@ -0,0 +1,273 @@
...
+    pub external_event_id: ExternalEventId,
+    pub external_actor_ref: ExternalActorRef,
+    pub external_conversation_ref: ExternalConversationRef,
+    /// Sanitized verified-claim attestation. Envelopes only exist after
+    /// the host has produced a `ProtocolAuthEvidence::Verified`; we carry
+    /// only the [`VerifiedAuthClaim`] payload here so the envelope
+    /// round-trips cleanly through audit logs and projections without
+    /// re-opening the forgery loophole that lived on the full
+    /// `ProtocolAuthEvidence` enum.
+    pub auth_claim: VerifiedAuthClaim,
```
serrrfirat: **Critical Severity**

The host-only auth invariant is bypassable at the workflow boundary.

Although `ProtocolAuthEvidence::Verified` rejects wire deserialization, `VerifiedAuthClaim` has public fields and `ProductInboundEnvelope.auth_claim` is public/deserializable. A buggy or malicious adapter can receive failed auth evidence and still fabricate an envelope claiming `subject = "alice"`; `ProductWorkflow::accept_inbound` has no way to distinguish that from host-verified auth.

Make the attestation opaque/sealed at the workflow boundary: require envelope construction through a constructor that consumes host-minted verified evidence, remove public claim fields/deserialization from the trusted path, and add an external/compile-fail regression test proving downstream crates cannot mint verified auth artifacts.

crates/ironclaw_product_adapters/src/external.rs
```diff
@@ -0,0 +1,305 @@
...
+        external_file_id: impl Into<String>,
+        mime_type: impl Into<String>,
+        filename: Option<String>,
+        size_bytes: Option<u64>,
+        kind: ProductAttachmentKind,
+    ) -> Result<Self, ProductAdapterError> {
+        let external_file_id = external_file_id.into();
+        let mime_type = mime_type.into();
+        validate_external_id("attachment_external_file_id", &external_file_id)?;
+        validate_external_id("attachment_mime_type", &mime_type)?;
```
serrrfirat: **Medium Severity**

Attachment MIME/filename invariants are documented but not enforced.

`ProductAttachmentDescriptor::new` accepts mixed-case/non-ASCII MIME values and filenames far above the documented 256-byte cap. A valid protocol MIME like `Image/JPEG` can be accepted at the boundary and later misclassified by MIME helpers, while oversized filenames can flow into workflow/audit paths despite the contract saying they are bounded.

Normalize or reject non-lowercase ASCII MIME values, enforce the filename byte cap, validate `kind` against the normalized base MIME, and add constructor plus serde regression tests.

crates/ironclaw_product_adapters/src/identity.rs
```diff
@@ -0,0 +1,107 @@
...
+            kind,
+            reason: "must not contain control characters".into(),
+        });
+    }
+    Ok(())
+}
+
+macro_rules! string_id {
+    ($name:ident, $kind:literal) => {
+        #[derive(Debug, Clone, PartialEq, Eq, Hash, PartialOrd, Ord, Serialize, Deserialize)]
```
serrrfirat: **Medium Severity**

Validated ID newtypes can bypass validation through derived Deserialize.

Types such as `ProductAdapterId`, `ExternalEventId`, `ExternalConversationRef`, and `ProductAttachmentDescriptor` rely on `new(...)` for length/control-character validation but derive plain `Deserialize`. JSON can therefore construct IDs with newlines or oversized metadata that constructors reject, then pass them to workflow/logging code that assumes validated values.

Use custom `Deserialize`/`serde(try_from)` for validated types so deserialization reuses the same constructors, and add caller-level invalid JSON tests.

crates/ironclaw_product_adapters/src/adapter.rs
```diff
@@ -0,0 +1,62 @@
...
+    fn surface_kind(&self) -> ProductSurfaceKind;
+
+    fn capabilities(&self) -> &ProductAdapterCapabilities;
+
+    /// Parse a verified protocol payload into a structured inbound envelope.
+    ///
+    /// `auth_evidence` is constructed by the host before this call. The
+    /// adapter MUST refuse to construct an envelope when the evidence is not
+    /// `Verified`.
+    fn parse_inbound(
```
serrrfirat, line 46: **Medium Severity**

The adapter contract has no host-visible auth requirement API.

`AuthRequirement` is documented as something adapters return before verification, but `ProductAdapter` only exposes `parse_inbound(..., auth_evidence)` after authentication should already have happened. A generic host holding `dyn ProductAdapter` cannot know whether to verify HMAC, shared-secret header, session cookie, or bearer token without out-of-band hardcoding.

Add a host-visible auth requirement/manifest method to the adapter contract and cover it in contract tests.

crates/ironclaw_product_adapters/src/fakes.rs
```diff
@@ -0,0 +1,346 @@
...
+impl ProductWorkflow for FakeProductWorkflow {
+    async fn accept_inbound(
+        &self,
+        envelope: ProductInboundEnvelope,
+    ) -> Result<ProductInboundAck, ProductAdapterError> {
+        let mut state = self.state.lock().expect("fake state lock poisoned"); // safety: in-memory test fake; a poisoned mutex means a panic in another test thread already happened
+        if let Some(error) = state.fail_with.clone() {
+            return Err(error);
+        }
+        let key = EventDedupeKey {
```
serrrfirat, line 95: **Medium Severity**

The fake workflow dedupe key omits the source binding.

The contract says dedupe is keyed by `(adapter_installation_id, source_binding, external_event_id)`, but `FakeProductWorkflow` uses only adapter + installation + event id. Two conversations that legitimately reuse the same external event id will be treated as duplicates in adapter tests even though the real workflow should accept both.

Include the same source-binding component the real workflow uses in `EventDedupeKey` and add a same-event-id/different-binding regression test.

crates/ironclaw_product_adapters/src/inbound.rs
```diff
@@ -0,0 +1,273 @@
...
+    /// messages we choose not to act on, etc. Workflow returns
+    /// [`ProductInboundAck::NoOp`] in response.
+    NoOp,
+}
+
+/// Inbound envelope. Constructed by the adapter from a verified protocol
+/// event, then handed to the workflow facade.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct ProductInboundEnvelope {
+    pub adapter_id: ProductAdapterId,
```
serrrfirat: **High Severity**

Adapters can reassert trusted adapter and installation identity.

`parse_inbound()` returns a full `ProductInboundEnvelope` containing `adapter_id` and `installation_id`, even though the host already knows both from the registered `ProductAdapter`. In a multi-installation process, an authenticated webhook for installation A can be parsed into an envelope claiming installation B unless every caller remembers to cross-check.

Have host glue inject trusted adapter/installation IDs after parsing, or centrally reject mismatches via sealed constructors and contract tests.

crates/ironclaw_product_adapters/src/external.rs
```diff
@@ -0,0 +1,305 @@
...
+    pub fn id(&self) -> &str {
+        &self.id
+    }
+
+    pub fn display_name(&self) -> Option<&str> {
+        self.display_name.as_deref()
+    }
+}
+
+/// External conversation reference.
```
serrrfirat: **High Severity**

External identity types mix stable keys with presentation/reply metadata.

`ExternalActorRef` derives Eq/Hash over `display_name`, and `ExternalConversationRef` derives Eq/Hash over `reply_target_message_id`, even though the docs define stable actor identity as kind+id and conversation identity as space+conversation+topic. A display-name change or different reply target can shard the same actor/thread into different bindings; the string `conversation_fingerprint()` can also collide because delimiters are allowed inside IDs.

Split stable key types from presentation/reply-target metadata, or implement manual Eq/Hash and structured/length-prefixed fingerprint encoding that ignores non-key fields.

crates/ironclaw_product_adapters/src/adapter.rs
```diff
@@ -0,0 +1,62 @@
...
+        &self,
+        raw_payload: &[u8],
+        auth_evidence: ProtocolAuthEvidence,
+    ) -> Result<Option<ProductInboundEnvelope>, ProductAdapterError>;
+
+    /// Render a projection-derived outbound envelope into the external
+    /// surface. Adapters use the supplied [`ProtocolHttpEgress`] for any
+    /// network I/O.
+    async fn render_outbound(
```
serrrfirat, line 54: **Medium Severity**

Outbound delivery status cannot be reported through the adapter contract.

`DeliveryStatusReporting` and `OutboundDeliverySink` exist, but `render_outbound()` receives no sink and returns only `Result<(), ProductAdapterError>`. `ProtocolHttpEgress::send()` also lacks attempt/target context. If Telegram returns 403 blocked-by-user versus 502 timeout, the adapter has no contract path to record `FailedUnauthorized` versus `FailedRetryable` for this attempt.

Pass a delivery sink into `render_outbound()` or return a structured delivery outcome/status for the host to record, with caller-level tests.

crates/ironclaw_product_adapters/src/egress.rs
```diff
@@ -0,0 +1,225 @@
...
+
+    pub fn as_str(&self) -> &str {
+        &self.0
+    }
+}
+
+/// Outbound HTTP request. Body is bytes; the host enforces size and content
+/// limits and applies the credential handle.
+#[derive(Debug, Clone)]
+pub struct EgressRequest {
```
serrrfirat, line 259: **Medium Severity**

Egress request method/path/headers are too raw for a host-enforced boundary.

`EgressRequest` leaves `method`, `path`, and `headers` unconstrained. A malicious adapter can request methods such as `CONNECT`, use authority-changing paths like `//169.254.169.254/...` or absolute URLs, or inject `Host`/`Authorization`/`Cookie` headers; a naive host join/forward implementation can escape the declared-host/credential boundary.

Replace raw strings with validated method, origin-form path/query, and header types; reject `://`, leading `//`, fragments, backslashes, and forbidden headers; re-check resolved URL authority after construction.

crates/ironclaw_product_adapters/src/inbound.rs
```diff
@@ -0,0 +1,273 @@
...
+            text,
+            attachments,
+            trigger,
+        })
+    }
+}
+
+/// Wrapped command inbound payload (e.g. `/help`, `/configure`).
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct InboundCommandPayload {
```
serrrfirat, line 124: **Medium Severity**

Command payloads bypass the only inbound text bound.

`UserMessagePayload` caps text at 64 KiB, but `InboundCommandPayload.command` and `arguments` are public unconstrained strings. Slash-command traffic can carry arbitrarily large attacker-controlled blobs into workflow/audit paths while still using the product-adapter contract.

Add a constructor/validated newtypes with maximum lengths for command and arguments, plus serde regression tests for oversized commands.

crates/ironclaw_product_adapters/src/inbound.rs
```diff
@@ -0,0 +1,273 @@
...
+    pub external_actor_ref: ExternalActorRef,
+    pub external_conversation_ref: ExternalConversationRef,
+    /// Sanitized verified-claim attestation. Envelopes only exist after
+    /// the host has produced a `ProtocolAuthEvidence::Verified`; we carry
+    /// only the [`VerifiedAuthClaim`] payload here so the envelope
+    /// round-trips cleanly through audit logs and projections without
+    /// re-opening the forgery loophole that lived on the full
+    /// `ProtocolAuthEvidence` enum.
+    pub auth_claim: VerifiedAuthClaim,
+    pub received_at: DateTime<Utc>,
```
serrrfirat: **Medium Severity**

Receipt time is adapter-controlled instead of host-stamped.

`parse_inbound()` returns a full envelope including `received_at`. A buggy or malicious adapter can backdate or future-date events, skewing audit ordering and future rate-limit/TTL/latency logic keyed off receipt time. Existing channel code stamps receipt centrally in the host.

Have host glue stamp `received_at` after parsing, or validate/overwrite adapter-supplied times centrally.

crates/ironclaw_product_adapters/src/fakes.rs
```diff
@@ -0,0 +1,346 @@
...
+        }
+        let outcome = state
+            .programmed
+            .remove(&envelope.external_event_id)
+            .unwrap_or_else(|| ProductInboundAck::Accepted {
+                accepted_message_ref: format!("msg:{}", envelope.external_event_id),
+                submitted_run_id: Some(TurnRunId::new()),
+            });
+        state.outcomes_by_event.insert(key, outcome.clone());
+        state.accepted_envelopes.push(envelope);
```
serrrfirat: **Medium Severity**

Fake workflow “accepted” tracking records rejected/no-op outcomes too.

`FakeProductWorkflow` pushes every first-seen envelope into `accepted_envelopes` regardless of whether the programmed outcome is `Rejected` or `NoOp`. Downstream tests that assert `accepted_count()` to prove no work was submitted can pass or fail for the wrong reason and blur “seen by workflow” with “accepted for turn submission”.

Rename this to `seen_envelopes` or record only genuinely accepted/deferred outcomes, and add regression tests for rejected/no-op paths.

crates/ironclaw_product_adapters/src/fakes.rs
```diff
@@ -0,0 +1,346 @@
...
+        }
+        Ok(EgressResponse {
+            status: 200,
+            headers: std::collections::BTreeMap::new(),
+            body: br#"{"ok":true}"#.to_vec(),
+        })
+    }
+}
+
+/// Convenience helper used by adapter tests to assert that no `submit_turn`
```
serrrfirat: **Medium Severity**

`ensure_durable_outcome()` does not prove NoOp avoided turn submission.

The helper is documented as checking that no submit_turn happened when an adapter chose `NoOp`, but it only calls `ack.is_durable_outcome()`, which is also true for `Accepted`, `DeferredBusy`, `Rejected`, and `Duplicate`. A downstream test can therefore pass even when an ambient/ignored event actually submitted work.

Replace it with a helper that specifically asserts `ProductInboundAck::NoOp` and no accepted/submitted envelope, or remove the misleading helper.

crates/ironclaw_product_adapters/src/outbound.rs
```diff
@@ -0,0 +1,140 @@
...
+///
+/// Reborn projection cursors are scoped to `(tenant, thread)` and validated
+/// against participant access on every consume — they are NOT transport-local
+/// SSE/WebSocket event ids.
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
+#[serde(transparent)]
+pub struct ProjectionCursor(String);
+
+impl ProjectionCursor {
+    pub fn new(value: impl Into<String>) -> Self {
```
serrrfirat: **Medium Severity**

Projection cursor and subscription hints are unbounded free-form strings.

`ProjectionCursor::new` accepts any string, `after_cursor` accepts it directly, and inbound `SubscriptionRequest` uses raw `thread_id_hint: Option<String>`. A Web/API client can reconnect with a multi-megabyte cursor or hint; serde will allocate/copy it through the adapter/projection path and it can be reflected back in outbound envelopes.

Use validated newtypes with max length/control-character checks and custom Deserialize, and add oversize input tests.

crates/ironclaw_product_adapters/src/fakes.rs
```diff
@@ -0,0 +1,346 @@
...
+
+impl Default for FakeProjectionStream {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+#[async_trait]
+impl ProjectionStream for FakeProjectionStream {
+    async fn drain(
```
serrrfirat, line 184: **Medium Severity**

FakeProjectionStream ignores subscription actor, scope, and cursor.

`FakeProjectionStream::drain()` discards the request and returns every queued envelope. A Web/CLI adapter that subscribes the wrong user/scope or drops cursor propagation can still pass against the public fake because request shape is never checked.

Have the fake record/filter by subscription key and cursor, or at least expose the last request for assertions; add wrong-scope/wrong-cursor tests.

crates/ironclaw_product_adapters/src/egress.rs
```diff
@@ -0,0 +1,225 @@
...
+    pub host: DeclaredEgressHost,
+    pub method: String,
+    pub path: String,
+    pub headers: BTreeMap<String, String>,
+    pub body: Vec<u8>,
+    pub credential_handle: Option<EgressCredentialHandle>,
+}
+
+/// Response surface visible to the adapter. Body is already leak-scanned by
```
serrrfirat: **High Severity**

Raw response headers are handed back to adapters across the credential boundary.

`EgressResponse` exposes response headers while the contract only promises body leak-scanning. A credential-authenticated upstream can reflect secret material in headers such as `Set-Cookie` or `X-*`, letting an adapter recover data that credential handles were supposed to keep host-only.

Strip/sanitize sensitive response headers before constructing `EgressResponse`, or expose only an allowlisted/typed metadata subset.

crates/ironclaw_product_adapters/src/inbound.rs
```diff
@@ -0,0 +1,273 @@
...
+    /// Placeholder for #3094. Contract tests render fake gate payloads;
+    /// production resolution flows live in the interaction services.
+    ApprovalResolution {
+        gate_ref: String,
+    },
+    /// Placeholder for #3094 (auth flow).
+    AuthResolution {
+        auth_request_ref: String,
+    },
+    /// Subscription request for a projection cursor (Web/CLI/API).
```
serrrfirat: **High Severity**

Projection subscription flow is not expressible through ProductWorkflow.

`SubscriptionRequest` carries only `thread_id_hint`, while `ProductWorkflow::accept_inbound()` returns only `ProductInboundAck`; `ProjectionStream::drain()` needs canonical `TurnActor`, `TurnScope`, and `after_cursor`. Since adapters must not resolve canonical IDs themselves, Web/CLI/API adapters must bypass the workflow boundary or lose reconnect/catch-up semantics.

Add a workflow entrypoint/outcome for projection subscription resolution that returns the canonical actor/scope/cursor plan, or keep subscription flows out of this contract until that path exists.

crates/ironclaw_product_adapters/src/outbound.rs
```diff
@@ -0,0 +1,140 @@
...
+/// Auth-prompt view (deferred to #3094 along with `GatePromptView`).
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct AuthPromptView {
+    pub turn_run_id: TurnRunId,
+    pub auth_request_ref: String,
+    pub headline: String,
+    pub body: String,
+}
+
+/// Snapshot of the canonical thread projection (full state).
```
serrrfirat: **Medium Severity**

Projection snapshot/update payloads contain no renderable projection state.

`ProjectionSnapshot` is documented as a full canonical thread projection and `ProjectionUpdate` as an incremental update, but both contain only `{cursor, thread_id, generated_at}`. A projection consumer cannot synchronize client state from these types without an out-of-band canonical fetch.

Include the transcript/run/gate/status payload needed by consumers, or rename/scope these as cursor notifications and add a separate projection data type.

crates/ironclaw_product_adapters/src/inbound.rs
```diff
@@ -0,0 +1,273 @@
...
+    pub trigger: ProductTriggerReason,
+}
+
+/// All supported inbound payload kinds. Approval/auth resolutions and
+/// projection subscriptions are tracked here so the adapter contract has
+/// space to grow into #3094 without a breaking change.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum ProductInboundPayload {
+    UserMessage(UserMessagePayload),
```
serrrfirat, line 360: **Medium Severity**

Approval/AuthResolution placeholders lack the action data needed for #3094 flows.

`ApprovalResolution` and `AuthResolution` are routed through the same `accept_inbound` path but carry only refs, with no approve/deny/always-allow, credential, or callback data. Interaction parity will require a breaking API or smuggling service state through pseudo-message handling.

Define a separate interaction-resolution entrypoint or include typed resolution action data before making this the stable core contract.

crates/ironclaw_product_adapters/src/error.rs
```diff
@@ -0,0 +1,83 @@
...
+//! returning a [`ProductAdapterError`].
+
+use thiserror::Error;
+
+use crate::ProtocolAuthFailure;
+use crate::redaction::RedactedString;
+
+/// Public error surface for product adapters and the workflow facade.
+#[derive(Debug, Clone, PartialEq, Eq, Error)]
+pub enum ProductAdapterError {
```
serrrfirat, line 11: **Medium Severity**

Several externally surfaced reason fields are plain String despite redaction promises.

`ProductAdapterError` reason fields, `ProductRejection.reason`, `ProtocolHttpEgressError::PolicyDenied.reason`, and `DeliveryStatus::*reason` can carry raw tokens, paths, or backend diagnostics. Unlike `Internal`/`Network`, nothing enforces redaction here, so one buggy workflow/adapter path can leak data into webhook responses, audit logs, or delivery storage.

Use `RedactedString` or typed safe reason enums for externally visible free-text fields, and add redaction tests for each public error/status surface.

crates/ironclaw_product_adapters/src/inbound.rs
```diff
@@ -0,0 +1,273 @@
...
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct ProductRejection {
+    pub kind: ProductRejectionKind,
+    /// User-safe explanation. Must not contain raw secrets, host paths,
+    /// raw provider/runtime internals, or backend diagnostics.
+    pub reason: String,
+}
+
+/// Pipeline outcome of an inbound envelope. Adapters use this to choose the
```
serrrfirat: **Medium Severity**

Inbound ack retry and duplicate-response guidance is ambiguous.

`PolicyDenied` is documented with examples like rate limit, but every `Rejected` is described as do-not-retry and mapped to 403/422. `Duplicate` is both listed as unconditional 200 and carries a prior outcome. Different adapters can make opposite webhook retry/response decisions for the same condition while following the contract.

Make retry/protocol mapping explicit in the ack type, e.g. separate transient/rate-limit rejection, permanent rejection, and duplicate-of-prior response semantics.

crates/ironclaw_product_adapters/src/fakes.rs
```diff
@@ -0,0 +1,346 @@
...
+    pub fn new() -> Self {
+        Self {
+            statuses: Mutex::new(Vec::new()),
+        }
+    }
+
+    pub fn statuses(&self) -> Vec<DeliveryStatus> {
+        self.statuses
+            .lock()
+            .expect("fake sink lock poisoned") // safety: in-memory test fake
```
serrrfirat: **Medium Severity**

FakeOutboundDeliverySink does not model DeliveryAttemptId dedupe.

`DeliveryAttemptId` is documented as the sink dedupe key, but the fake sink appends every status and never collapses repeated attempt IDs. Downstream adapter tests using this fake will miss double-reporting bugs in delivery accounting/alerting.

Have the fake dedupe by attempt_id or expose duplicate-attempt assertions, and add repeated-attempt regression tests.

crates/ironclaw_product_adapters/src/egress.rs
```diff
@@ -0,0 +1,225 @@
...
+    }
+}
+
+/// Outbound HTTP request. Body is bytes; the host enforces size and content
+/// limits and applies the credential handle.
+#[derive(Debug, Clone)]
+pub struct EgressRequest {
+    pub host: DeclaredEgressHost,
+    pub method: String,
+    pub path: String,
```
serrrfirat: **Medium Severity**

HTTP headers are modeled as BTreeMap, losing duplicate headers.

`EgressRequest.headers` and `EgressResponse.headers` collapse repeated HTTP headers. A lossless bridge from runtime egress APIs is impossible for common cases like multi-`Set-Cookie`; unsafe comma-joining is invalid for `Set-Cookie`.

Use `Vec<(String, String)>` or a multi-map/header type that preserves duplicate header order.

crates/ironclaw_product_adapters/src/egress.rs
```diff
@@ -0,0 +1,225 @@
...
+    -> Result<EgressResponse, ProtocolHttpEgressError>;
+}
+
+/// Stable identifier for one delivery attempt. Sinks use this to dedupe.
+pub type DeliveryAttemptId = Uuid;
+
+/// Per-attempt delivery outcome.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum DeliveryStatus {
```
serrrfirat, line 370: **Medium Severity**

DeliveryStatus has no permanent non-auth failure variant.

If an adapter gets a non-retryable 4xx such as “message too long” or “thread not found”, it must misreport it as `FailedRetryable` or `FailedUnauthorized`. That can cause pointless retries or falsely mark a binding as revoked.

Add a distinct permanent/rejected delivery status and reserve `FailedUnauthorized` for real auth/access revocations.

crates/ironclaw_product_adapters/src/outbound.rs
```diff
@@ -0,0 +1,140 @@
...
+    GatePrompt(GatePromptView),
+    AuthPrompt(AuthPromptView),
+    ProjectionSnapshot(ProjectionSnapshot),
+    ProjectionUpdate(ProjectionUpdate),
+}
+
+/// Outbound envelope handed to the adapter for protocol translation +
+/// delivery.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct ProductOutboundEnvelope {
```
serrrfirat, line 260: **Medium Severity**

Projection cursors have two sources of truth in outbound envelopes.

`ProductOutboundEnvelope.projection_cursor` can differ from the cursor inside `ProjectionSnapshot` or `ProjectionUpdate`. If a producer emits mismatched values, adapters can persist one cursor while rendering another, causing replay or skipped events on resume.

Keep a single cursor source of truth, or validate equality when building snapshot/update envelopes.

crates/ironclaw_product_adapters/src/adapter.rs
```diff
@@ -0,0 +1,62 @@
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum ProductAdapterHealth {
+    Healthy,
+    Degraded,
+    Unhealthy,
+}
+
+#[async_trait]
+pub trait ProductAdapter: Send + Sync {
```
serrrfirat, line 23: **Medium Severity**

Declared egress hosts and credential handles are not discoverable from ProductAdapter.

`DeclaredEgressHost` is documented as manifest-declared and host-enforced, but `ProductAdapter` exposes no host-visible declared-host or credential metadata. A generic host must hardcode or separately configure allowlists, inviting host/adapter drift as new adapters land.

Add adapter manifest/capability metadata for declared hosts and credential handles, and test that hosts can build egress policy from the contract.

crates/ironclaw_product_adapters/src/fakes.rs
```diff
@@ -0,0 +1,346 @@
...
+        Self {
+            state: Mutex::new(FakeEgressState {
+                declared_hosts: declared_hosts.into_iter().collect(),
+                ..Default::default()
+            }),
+        }
+    }
+
+    pub fn allow_credential_handle(&self, handle: impl Into<String>) {
+        let mut state = self.state.lock().expect("fake egress lock poisoned"); // safety: in-memory test fake
```
serrrfirat: **Medium Severity**

FakeProtocolHttpEgress cannot model multi-call same-host flows.

The fake can program only one response per host and consumes it on first use. Telegram/Slack-style adapters making multiple calls to the same API host, or retrying after 429/5xx, cannot be realistically tested with the public test-support fake.

Allow response queues keyed by request attributes such as host+path+method, and add multi-call/retry tests.

crates/ironclaw_product_adapters/src/inbound.rs
```diff
@@ -0,0 +1,273 @@
...
+/// protocol-level response status code:
+///
+/// * `Accepted` / `DeferredBusy` / `NoOp` / `Duplicate` -> 200
+/// * `Rejected { BindingRequired | AccessDenied | UnknownInstallation }` -> 403
+/// * `Rejected { PolicyDenied }` -> 403/422 depending on protocol convention
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum ProductInboundAck {
```
serrrfirat, line 548: **Medium Severity**

Accepted inbound acks lose canonical turn identity.

`ProductInboundAck::Accepted` downgrades `accepted_message_ref` to a raw `String` and makes `submitted_run_id` optional, while the underlying turn response has typed `AcceptedMessageRef` and a mandatory `TurnRunId`. If this ack is stored/replayed as the durable duplicate outcome, later code may lose or poison message/run correlation.

Model accepted/busy acks with canonical turn types (`AcceptedMessageRef` and mandatory `TurnRunId`) or wrap the relevant `SubmitTurnResponse`/`ThreadBusy` fields directly, with round-trip tests.

crates/ironclaw_product_adapters/tests/product_adapter_contract.rs
```diff
@@ -0,0 +1,464 @@
...
+    "ironclaw_gateway",
+    "ironclaw_tui",
+    "ironclaw_memory",
+    "ironclaw_events",
+    "ironclaw_reborn_event_store",
+    "ironclaw_architecture",
+];
+
+#[test]
+fn cargo_manifest_does_not_pull_in_forbidden_lower_layers() {
```
serrrfirat, line 52: **Medium Severity**

Architecture guardrail tests are raw substring checks and easy to bypass.

The boundary tests look for strings like `{forbidden} =` and `ironclaw_turns::runner`. A renamed dependency such as `foo = { package = "ironclaw_gateway", ... }` or grouped import `use ironclaw_turns::{runner::TrustedTransition};` can bypass them while regressing the architecture fence.

Parse `Cargo.toml`/`cargo metadata` for resolved package names and inspect Rust path segments, or at least use regexes that catch renamed dependencies and grouped runner imports.

crates/ironclaw_product_adapters/src/adapter.rs
serrrfirat, line 46: **Medium Severity**

The contract has two incompatible ignored-event paths.

`parse_inbound()` can return `None`, while DTOs also define `ProductInboundPayload::NoOp` and `ProductInboundAck::NoOp`. If an adapter drops an authenticated ambient/edit event with `None`, the workflow never sees `external_event_id`, so no durable NoOp/Duplicate record is written and re-delivery can later be processed as fresh after config/state changes.

Choose one canonical ignored-event path. Prefer requiring authenticated events to return an envelope with `NoOp`, or return enough host-visible metadata for the host to synthesize and dedupe a durable NoOp.

crates/ironclaw_product_adapters/src/error.rs
```diff
@@ -0,0 +1,83 @@
...
+    WorkflowRejected { reason: String },
+
+    #[error("workflow transient failure: {reason}")]
+    WorkflowTransient { reason: String },
+
+    #[error("internal adapter error: {detail}")]
+    Internal { detail: RedactedString },
+}
+
+impl ProductAdapterError {
```
serrrfirat, line 37: **Medium Severity**

Adapter-level retry classification omits transient egress failures.

`render_outbound()` can hit retryable transport failures such as `ProtocolHttpEgressError::Timeout` or `Network`, but `ProductAdapterError::is_retryable()` returns true only for `WorkflowTransient`. Real adapters must mislabel transient delivery failures as workflow errors or lose retries.

Add a transient egress error variant or structured mapping from `ProtocolHttpEgressError` into retryable adapter errors.

crates/ironclaw_product_adapters/src/outbound.rs
serrrfirat, line 260: **Medium Severity**

Outbound envelopes carry only an opaque reply target with no lookup hook.

`render_outbound()` gets `ReplyTargetBindingRef` plus `ProtocolHttpEgress`; the envelope never carries the resolved external destination and the contract exposes no lookup hook. After restart/proactive delivery, a real adapter cannot recover chat/thread/message destination without depending on non-contract conversation services or maintaining a second reply-target store.

Pass resolved external destination metadata in the outbound envelope or add a contract-level lookup service owned by host/workflow glue.

crates/ironclaw_product_adapters/src/error.rs
```diff
@@ -0,0 +1,83 @@
...
+    #[error("protocol authentication failed: {0}")]
+    Authentication(#[from] ProtocolAuthFailure),
+
+    #[error("egress denied: {reason}")]
+    EgressDenied { reason: String },
+
+    #[error("egress to undeclared host {host}")]
+    EgressUndeclaredHost { host: String },
+
+    #[error("workflow rejected inbound: {reason}")]
```
serrrfirat: **Medium Severity**

Semantic inbound rejection has two incompatible channels.

`ProductWorkflow::accept_inbound` can return `Ok(ProductInboundAck::Rejected(...))` or `Err(ProductAdapterError::WorkflowRejected { .. })`. Only the ack path carries `ProductRejectionKind` and durable duplicate semantics. Implementations using the error variant for policy/access denials will lose idempotent rejection handling and can map the same condition to different protocol responses.

Remove or narrow `WorkflowRejected`, or make all semantic rejections return `ProductInboundAck::Rejected` and reserve errors for infrastructure failures.

crates/ironclaw_product_adapters/src/inbound.rs
```diff
@@ -0,0 +1,273 @@
...
+#[serde(rename_all = "snake_case")]
+pub enum ProductTriggerReason {
+    /// Direct/private chat — the participant always receives messages.
+    DirectChat,
+    /// Explicit @mention of the bot.
+    BotMention,
+    /// Reply to a message authored by the bot.
+    ReplyToBot,
+    /// Recognized bot command (e.g. `/start`).
+    BotCommand,
```
serrrfirat, line 79: **Medium Severity**

LinkedThreadAction trigger lacks action/callback data.

`ProductTriggerReason::LinkedThreadAction` is advertised, but no inbound payload field carries the action token/button callback data identifying which linked action fired. Telegram/Slack callback interactions must be shoehorned into text/arguments, and reply_target_message_id only identifies the message, not the button/action.

Add a typed linked-action payload with action id/data and target metadata, or defer this trigger until the contract can carry the needed data.

crates/ironclaw_product_adapters/src/identity.rs
```diff
@@ -0,0 +1,107 @@
...
+}
+
+string_id!(ProductAdapterId, "product_adapter_id");
+string_id!(AdapterInstallationId, "adapter_installation_id");
+
+/// Surface kind a product adapter exposes. Used by the workflow layer to pick
+/// safe presentation defaults.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum ProductSurfaceKind {
```
serrrfirat, line 74: **Medium Severity**

Advertised Web/CLI/SynchronousApi surfaces cannot return synchronous responses through this contract.

The crate advertises `Web`, `Cli`, `SynchronousApi`, `ProjectionSubscription`, and `SynchronousWait`, but the only outbound hook is side-effecting `render_outbound(...) -> Result<(), ProductAdapterError>`. A generic host cannot wait on a projection and return adapter-rendered bytes/object to the original pending request without out-of-band side tables or host-specific escape hatches.

Separate push-channel adapters from synchronous surfaces, or add a contract-level synchronous render/response path for Web/CLI/API adapters.

PR #3353, opened 2026-05-07
feat(reborn): add native product adapter runner

serrrfirat:
Split from #3316 (original PR by @nickpismenkov). This is PR 3/7 in the ProductAdapter stack.

## Summary
- Add native `ProductAdapter` runner glue.
- Runner verifies protocol auth, mints sealed host evidence, calls adapter parsing, and forwards accepted envelopes to the `ProductWorkflow` facade.
- Keeps wasmtime/component runtime wiring out of this slice.

## Stack
- Depends on #3352.
- Base branch: `split/pr3316-02-product-adapter-host-auth-egress`.

## Authorship

## Verification
- `cargo test -p ironclaw_wasm_product_adapters --offline --quiet` — 13 tests passed on this slice.


zmanian:
# Review

**Summary:** Adds the native (non-wasmtime) `ProductAdapter` runner plus supporting crates — webhook auth verification (HMAC + shared-secret), egress allow-listing, `ProductWorkflow` facade boundary — as PR 3/7 in the stack, intentionally deferring component-model wiring.

**Findings:**

1. **Sandbox bypass risk — by design, but the naming hides it.** The runner lives at `crates/ironclaw_wasm_product_adapters/src/runner.rs` inside a crate named `ironclaw_wasm_product_adapters`, but executes `Arc<dyn ProductAdapter>` in-process with full host privileges. The crate name strongly implies WASM isolation that doesn't exist here. Rename the module path (e.g. `ironclaw_product_adapter_host::native`) or add a `#![doc = "WARNING: native runner has no sandbox..."]` banner so operators don't conflate trust models.

2. **Panic propagation across adapter boundary is unmitigated — blocking.** `process_webhook` calls `self.adapter.parse_inbound(...)` and `self.workflow.accept_inbound(...)` with no `catch_unwind` or task isolation. A panic in a 3rd-party-authored adapter (the whole point of this seam) aborts the HTTP worker; Telegram/Slack retry on non-200, creating a stuck-message replay storm. The wasmtime path gets trap-isolation for free; the native path needs explicit `tokio::spawn` + `JoinHandle` or `AssertUnwindSafe` + `catch_unwind`.

3. **No lifecycle/cancellation handling — blocking.** `NativeProductAdapterRunner` has no `Drop`, no cancellation token, no timeout. `accept_inbound` is unbounded `async`; a slow workflow blocks the webhook handler indefinitely. Telegram's webhook timeout is ~60s and redelivers on timeout. Add configurable `tokio::time::timeout` per `process_webhook` surfaced as `RunnerError::WorkflowTimeout`.

4. **No back-pressure / admission control — blocking.** Runner is `Arc<>`-shared and `process_webhook` can be invoked concurrently without bound. A Telegram update flood (or hostile bot) saturates `TurnCoordinator`. The contract test doesn't exercise concurrent dispatch. Add a `tokio::sync::Semaphore` field with configurable max in-flight.

5. **`EgressPolicy::check` fail-closed path emits no audit event.** Given the project rule that all mutations route through `ToolDispatcher` for audit, a silent `EgressPolicyError` return loses the security signal. Add `tracing::warn!` at the call site of `check` (this crate isn't in the REPL pipeline, so `warn!` is fine).

**Minor:**
- `auth_verifier.rs:3253` rolls its own `hex::encode` — workspace already has `hex`.
- `self.clock.now_unix_seconds() as i64` (line ~3183) silently wraps past year 2554.
- `SystemClock` `unwrap_or(0)` (line ~3078) means a broken system clock mints "1970" and would falsely pass replay-window checks for legitimately-timestamped recent requests within 5 min of epoch. Practically harmless but worth a `tracing::error!`.

**Verdict: Request changes.** Auth/egress primitives are solid (constant-time compare, symmetric drift window, fail-closed allow-lists). But the runner shipping in-process with no panic isolation, no timeout, no admission control, and a misleading crate name is a production-incident generator. (2), (3), (4) must land before this is wired to a real webhook in the rest of the stack; (1) should land before merge.

zmanian (approved):
# Re-review: approved

Four blocking findings resolved, one mitigated at `9d002a91`:

1. **Crate naming — MITIGATED.** Crate name `ironclaw_wasm_product_adapters` unchanged, but `lib.rs:1-15` adds a prominent banner: *"The native runner in this crate executes `Arc<dyn ProductAdapter>` in the host process. It is not a WASM sandbox."* Acceptable for now; recommend rename when the wasmtime path lands.
2. **Panic isolation — RESOLVED.** `adapter.parse_inbound` wrapped in `catch_unwind(AssertUnwindSafe(...))` → `RunnerError::AdapterPanicked`. Workflow isolated via `tokio::spawn` + `JoinError::is_panic()` → `WorkflowPanicked`. `is_retryable()` correctly classifies adapter panic as non-retryable, workflow panic as retryable. Tests `process_webhook_contains_adapter_panics` and `process_webhook_contains_workflow_panics` cover both.
3. **Workflow timeout — RESOLVED.** `tokio::time::timeout(self.config.workflow_timeout, ...)` with `workflow_task.abort()` on elapse to prevent leak. Default 55s. Test `process_webhook_times_out_slow_workflow` covers.
4. **Admission control — RESOLVED.** `Arc<Semaphore>` initialized to `config.max_in_flight()`; `try_acquire_owned()` returns `TooManyInFlight` without blocking. Default 64. Permit acquired after auth verify (failed auth doesn't consume a slot) but before parse (bounds CPU work) — correct ordering. Test verifies non-blocking rejection.
5. **Egress audit — PARTIAL.** `tracing::warn!` with `target: "ironclaw.product_adapter.egress"` on both `UndeclaredHost` and `UnauthorizedCredentialHandle`. Log line rather than structured audit event; acceptable for a tracer-bullet slice. Promote to structured audit when that infrastructure exists.

Bonus: `as i64` time cast replaced with `i128` arithmetic; `unwrap_or(0)` clock fallback now `match`-with-`tracing::error!` and fails closed (drift always exceeds window).

Follow-ups: (a) rename crate post-wasmtime split, (b) promote egress denials to structured audit events.

PR #3354, opened 2026-05-07
feat(reborn): add telegram v2 payload normalization

serrrfirat:
Split from #3316 (original PR by @nickpismenkov). This is PR 4/7 in the ProductAdapter stack.

## Summary
- Add Telegram v2 payload normalization crate slice.
- Parse Telegram updates into Reborn external actor/conversation/event refs.
- Enforce explicit group/supergroup trigger gating.
- Emit bounded attachment descriptors with no raw bytes/source URLs/host paths.
- Intentionally excludes ProductAdapter outbound rendering until the next PR.

## Stack
- Depends on #3353.
- Base branch: `split/pr3316-03-product-adapter-native-runner`.

## Authorship

## Verification
- `cargo test -p ironclaw_telegram_v2_adapter --offline --quiet` — 21 tests passed on this slice.


henrypark133 (commented):
Finding: private-chat bot commands are downgraded to normal user messages.

`classify_trigger` returns `DirectChat` immediately for private chats before it checks Telegram `bot_command` entities. Later, `build_payload` only emits `ProductInboundPayload::Command` when the trigger is `BotCommand`, so a DM like `/help` is accepted as a `UserMessage` instead of reaching command handling. The group command path works because groups do reach `recognized_bot_command`, but private commands do not.

This conflicts with the adapter advertising `InboundCommands` and means normal Telegram bot command usage in DMs will not be routed as a command. Please recognize valid commands before the private-chat early return, while still treating non-command private messages as `DirectChat`, and add a private `/help` regression test.

nickpismenkov:
Thanks for both reviews. Pushed [`7485b872e`](https://github.com/nearai/ironclaw/pull/3354/commits/7485b872e) (plus the merge commit [`67acc2f71`](https://github.com/nearai/ironclaw/pull/3354/commits/67acc2f71) right before it) closing all 5 findings — @henrypark133's High and all 4 of @copilot-pull-request-reviewer's inline notes.

## @henrypark133's High: private-chat bot commands

Confirmed the bug exactly as you described: `classify_trigger` returned `DirectChat` for private chats before checking `bot_command` entities, and `build_payload` only emitted `Command` when `trigger == BotCommand`. The two together silently downgraded every DM `/command` to `UserMessage`, contradicting the adapter advertising `InboundCommands`.

The fix ended up being Copilot's #3 framing rather than a literal "recognize commands before the private-chat early return" — see below.

## @copilot-pull-request-reviewer's inline findings

### #1 — module docstring (`payload.rs:6`)

Updated the docstring to reference `TelegramParsedInbound::NoOp` (the actual return type), not "returns `None`".

### #2 — caption_entities support (`payload.rs:222`)

Confirmed: a photo with caption `@ironclaw_bot help` carried its mention in `caption_entities`, so `classify_trigger` never saw it and the update silently NoOp'd — even though `build_payload` later used `caption` as the user text. Real gap.

Added `caption_entities: Option<Vec<MessageEntity>>` to `TelegramMessage` plus a private helper:

```rust
fn text_entity_windows(message: &TelegramMessage)
    -> impl Iterator<Item = (&str, &[MessageEntity])>
```

— which yields `(text, entities)` then `(caption, caption_entities)`. All three detectors (`has_bot_mention`, `recognized_bot_command`, `extract_first_bot_command`) iterate through it. Offsets in each entity list stay bound to their companion string. Two regression tests pin both the mention and the command case for media messages.

### #3 — decouple Command vs UserMessage from trigger (`payload.rs:469`)

This is the unified fix for both reviews. `build_payload` now emits `Command` whenever a recognized `bot_command` entity exists in the message, regardless of `trigger`:

```rust
if let Some((command, arguments)) = extract_first_bot_command(&message, policy) {
    return Ok(ProductInboundPayload::Command(InboundCommandPayload {
        command, arguments, trigger,
    }));
}
```

The `trigger` field still records the forwarding reason (`DirectChat` for DMs, `BotMention` / `ReplyToBot` for groups, `BotCommand` if a recognized command was the sole reason a group message reached us). This:

- Fixes Henry's DM `/help` case (private chat, `trigger=DirectChat`, payload=`Command{command="help"}`).
- Also fixes the case Copilot flagged separately: `@ironclaw_bot /help` in a group (`trigger=BotMention`, payload=`Command{command="help"}`) — previously `UserMessage` because the mention check fired first and set the trigger.

I reverted my partial earlier fix in `classify_trigger` (which had set `trigger=BotCommand` for DMs with commands) in favor of this cleaner separation — the trigger now stays semantically "why the message was forwarded" instead of being conflated with "what kind of payload does it carry."

Three regression tests pin the new shape:
- `private_chat_recognized_bot_command_emits_command_payload` — DM `/help` → `Command` with `trigger=DirectChat`.
- `private_chat_unknown_command_still_classifies_as_direct_chat` — unrecognized `/nope` in DM stays `UserMessage`.
- `group_mention_with_bot_command_emits_command_payload` — group `@ironclaw_bot /help` → `Command` with `trigger=BotMention`.

### #4 — fail-soft on missing `message.from` (`payload.rs:419`)

Confirmed `message.from` is optional in the Telegram schema (anonymous group admins, slipped-through channel updates). `parse_telegram_update` now checks `message.from.is_none()` up front and returns `TelegramParsedInbound::NoOp`, mirroring the existing fail-soft path for unsupported update kinds. Regression test `message_without_from_classifies_as_noop_not_error` pins this.

## Merge

The earlier merge commit `67acc2f71` synced this branch with `reborn-integration`, which carried the post-#3352 product-adapter API refactor (`ProtocolAuthEvidence` enum → sealed struct, `ProductInboundEnvelope` fields → private, `mark_*_verified` helpers behind `host-auth-mint` feature). `payload.rs` was updated to use `TrustedInboundContext::from_verified_evidence` + `ProductInboundEnvelope::from_trusted_parse`. Workspace `Cargo.toml` members resolved as union.

## Verified

- `cargo check --workspace` clean
- `cargo fmt --all -- --check` clean
- `cargo clippy -p ironclaw_telegram_v2_adapter --all-targets -- -D warnings` zero warnings
- `cargo test -p ironclaw_telegram_v2_adapter` **27 tests green** (was 21 pre-review — +6 net new regressions across the 5 review findings)

henrypark133 (changes requested):
## Review: preserve inbound command validation

The latest head fixes the earlier private-chat command regression and the Telegram normalization slice is otherwise well contained. There is one validation bypass in the command payload path that should be fixed before merge.

What looks good:
- Private-chat `/help` now emits a `Command` payload instead of being downgraded to a normal user message.
- Caption entity handling covers media captions for mentions and bot commands, which closes an easy-to-miss Telegram edge case.
- The adapter keeps this slice focused on inbound payload normalization and still avoids raw attachment bytes, source URLs, or host paths.

### Critical: command payloads bypass the shared constructor validation
`build_payload` constructs `InboundCommandPayload` with a struct literal:

```rust
let command_payload = InboundCommandPayload {
    command,
    arguments,
    trigger,
};
```

That skips `InboundCommandPayload::new`, which is where `ironclaw_product_adapters` enforces the command token limit, non-empty command shape, command-argument byte limit, and control-character checks. User messages in the same function go through `UserMessagePayload::new`, so this leaves command payloads as the inconsistent path.

Because Telegram command arguments come from untrusted webhook text, this can carry oversized or control-character-containing arguments into the trusted inbound envelope. Please route command payload creation through `InboundCommandPayload::new(...)` and map any validation error the same way the user-message path does. Add a regression test with invalid command arguments, for example a recognized `/help` command whose arguments contain a NUL/control character or exceed the command argument limit.

Verification I ran:
- `cargo test -p ironclaw_telegram_v2_adapter --quiet`
- `git diff --check origin/pr/3354`
- Rechecked GitHub PR checks; relevant checks are passing.

nickpismenkov:
Thanks for catching that, @henrypark133. Pushed [`29000e62b`](https://github.com/nearai/ironclaw/pull/3354/commits/29000e62b) routing command payload creation through `InboundCommandPayload::new`.

You're right that this was a real bypass — the user-message arm went through `UserMessagePayload::new` correctly, so commands stood out as the inconsistent path. With untrusted Telegram webhook text upstream, oversized or control-character-containing arguments could land in the trusted inbound envelope unchallenged.

## Fix

```rust
let command_payload =
    InboundCommandPayload::new(command, arguments, trigger).map_err(|err| {
        PayloadParseError::InvalidExternalRef {
            kind: "inbound_command_payload",
            reason: err.to_string(),
        }
    })?;
return Ok(ProductInboundPayload::Command(command_payload));
```

Mirrors the `UserMessagePayload::new` call shape directly below for the user-message arm. The `reason` carries a `RedactedString` from `ProductAdapterError::MalformedInboundPayload`, so the failure detail does not leak.

## Regression tests

Two new tests pin both the control-char and oversized cases:

- **`command_arguments_with_control_char_rejected_via_shared_validation`** — `/help <U+0001>oops` (control character via JSON `` escape) must reject with `InvalidExternalRef{kind="inbound_command_payload"}` AND the reason must be `<redacted>`. The redaction assertion proves both that the shared validator was reached AND that its redaction contract is intact — a leak of the offending content into the error message would fail this assertion.

- **`command_arguments_exceeding_byte_limit_rejected_via_shared_validation`** — 70,000-byte argument (over the 64 KiB `COMMAND_ARGUMENTS_MAX_BYTES` limit) synthesized at runtime via `format!()`. Same redaction assertion.

I picked the `<redacted>`-marker assertion deliberately rather than asserting on the specific failure string — that way the test also catches a regression where someone "improves" the error to include the original content, which would defeat the redaction.

## Verified

- `cargo fmt --all -- --check` clean
- `cargo clippy -p ironclaw_telegram_v2_adapter --all-targets -- -D warnings` zero warnings
- `cargo test -p ironclaw_telegram_v2_adapter` **29 tests green** (was 27 — +2 validation regression tests for this fix)

henrypark133 (approved):
## Review: approve Telegram v2 payload normalization

The follow-up fix resolves my earlier blocker. Telegram v2 command payloads now go through the shared `InboundCommandPayload::new(...)` validator, so command arguments get the same control-character and size checks as the rest of the inbound command path.

What looks good:
- The fix is scoped to the command-payload construction path instead of duplicating validation locally.
- The new regression tests cover both control-character rejection and oversized argument rejection through the adapter caller path.
- CI is green, and the targeted adapter test suite passes locally.

PR #3365, opened 2026-05-07
fix(bridge): bypass agent-loop mpsc for inline-await Approval gates

ilblackdragon:
## Summary

- When `BridgeGateController::pause` parks a Tier 0/1 execution on an Approval gate, the agent loop's `handle_message` is held awaiting the bridge call, so `ExecApproval` submissions posted to `/api/chat/approval` queued indefinitely behind the parked alpha — `try_deliver` was never reached and alpha only woke on the 30-min pause timeout.
- Adds `bridge::try_resolve_inline_approval_gate` that does `take_verified` + persist `always_allow` + `try_deliver` + SSE `GateResolved` directly, without going through `msg_tx`.
- `chat_approval_handler` tries the fast-path first and falls through to the legacy mpsc dispatch on `NoLiveVm` so engine-v1 approvals (and any post-restart Approval gates without a parked future) still resolve correctly.

## Details

The inline-await regression introduced by #3157 only affects Approval-kind gates because `BridgeGateController::pause` returns `Cancelled` immediately for Authentication and External resume kinds. The fix is therefore narrowly scoped to the Approval path:

- `persist_always_allow` and `revert_always_allow` refactored to take `Option<&dyn SettingsStore>` so the gateway HTTP fast-path can install the AlwaysAllow preference without an `&Agent` reference.
- `take_verified` is reused as-is so channel auth, request_id, and expiry checks still apply.
- On `try_deliver` returning false (no parked future), the gate is re-inserted and any installed auto-approve preference is rolled back, so the fall-through to mpsc dispatch finds a clean state.
- SSE `GateResolved` is broadcast on the inline path so the UI clears the gate card without waiting for the engine's own event loop to catch up.

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo test --no-default-features --features libsql --lib` — 5602/5602 passing
- [x] `cargo test --no-default-features --features libsql --test engine_v2_gate_integration` — 27/27 passing
- [x] `cargo test --no-default-features --features libsql --test e2e_engine_v2` — 28/28 passing
- [x] New regression unit test `bridge::router::tests::try_resolve_inline_approval_gate_wakes_parked_pause` spawns a real `pause()` future and verifies the new entry point wakes it within 2s.
- [x] `scripts/pre-commit-safety.sh` — clean (PROJECTION rule satisfied with same-line annotation on the SSE broadcast).
- [ ] e2e: `pytest tests/e2e/scenarios/test_v2_engine_approval_flow.py` — 5 previously-xfailed approval tests should now pass.
- [ ] e2e: `pytest tests/e2e/scenarios/test_channel_approval_gates.py` — new Telegram/Slack channel approval coverage.

### DIFF
--- a/src/bridge/router.rs
+++ b/src/bridge/router.rs
@@ -2505,6 +2505,10 @@ pub enum InlineGateError {
     /// The pending gate's `expires_at` has elapsed.
     #[error("Approval request has expired")]
     Expired,
+    /// The pending gate exists but does not belong to the requesting user.
+    /// Surfaced as a 403 to avoid leaking gate existence across tenants.
+    #[error("not authorized to resolve this gate")]
+    Unauthorized,
     /// Any other gate-store failure.
     #[error("gate error: {0}")]
     Other(String),
@@ -2542,7 +2546,6 @@ pub enum InlineGateError {
 pub async fn try_resolve_inline_approval_gate(
     user_id: &str,
     channel: &str,
-    thread_id: ironclaw_engine::ThreadId,
     request_id: uuid::Uuid,
     resolution: ironclaw_engine::GateResolution,
     settings_store: Option<&(dyn crate::db::SettingsStore + Send + Sync)>,
@@ -2569,30 +2572,46 @@ pub async fn try_resolve_inline_approval_gate(
         return Ok(InlineGateOutcome::NoLiveVm);
     };
 
-    let key = crate::gate::pending::PendingGateKey {
-        user_id: user_id.to_string(),
-        thread_id,
-    };
-
-    // take_verified atomically removes + verifies (request_id, channel
-    // auth, expiry). Verification failures surface as a typed
-    // `InlineGateError` so the HTTP handler can map to the right 4xx
-    // status without inspecting message strings.
-    let pending = state
+    // Resolve the gate by `request_id` (system-wide unique) rather
+    // than by a caller-supplied thread identifier. The wire
+    // `req.thread_id` on the HTTP surface is the channel-visible
+    // value — for the web gateway that is the per-conversation UUID
+    // returned by `/api/chat/thread/new`, recorded on the gate as
+    // `scope_thread_id` — not the internal engine `ThreadId` that
+    // keys `PendingGateStore`. Looking up by `request_id` under the
+    // store's single mutex keeps the lookup + remove atomic and
+    // avoids the wire-vs.-engine identifier confusion that would
+    // otherwise miss every gate whose channel scope differs from its
+    // engine thread.
+    //
+    // Verification failures surface as a typed `InlineGateError` so
+    // the HTTP handler can map to the right 4xx without inspecting
+    // message strings. `NotFound` is treated as `NoLiveVm` (legacy
+    // mpsc fall-through) — a `request_id` we don't have means the
+    // gate was already resolved, never existed, or wasn't restored
+    // after a process restart, none of which should surface a 5xx
+    // here.
+    let pending = match state
         .pending_gates
-        .take_verified(&key, request_id, channel)
+        .take_verified_by_request_id(request_id, user_id, channel)
         .await
-        .map_err(|e| {
+    {
+        Ok(gate) => gate,
+        Err(e) => {
             use crate::gate::store::GateStoreError;
-            match e {
+            return match e {
+                GateStoreError::NotFound => Ok(InlineGateOutcome::NoLiveVm),
                 GateStoreError::ChannelMismatch { expected, actual } => {
-                    InlineGateError::ChannelMismatch { expected, actual }
+                    Err(InlineGateError::ChannelMismatch { expected, actual })
                 }
-                GateStoreError::RequestIdMismatch => InlineGateError::Stale,
-                GateStoreError::Expired => InlineGateError::Expired,
-                other => InlineGateError::Other(other.to_string()),
-            }
-        })?;
+                GateStoreError::Unauthorized => Err(InlineGateError::Unauthorized),
+                GateStoreError::Expired => Err(InlineGateError::Expired),
+                GateStoreError::RequestIdMismatch => Err(InlineGateError::Stale),
+                other => Err(InlineGateError::Other(other.to_string())),
+            };
+        }
+    };
+    let thread_id = pending.thread_id;
 
     // Only Approval-resume gates are parked by the gate controller. A
     // non-Approval gate hitting take_verified here means a different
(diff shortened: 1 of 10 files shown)


src/channels/web/features/chat/mod.rs
```diff
@@ -187,6 +187,66 @@ pub(crate) async fn chat_approval_handler(
...
+            ironclaw_engine::GateResolution::Denied { reason: None }
+        };
+        let settings_store: Option<&(dyn crate::db::SettingsStore + Send + Sync)> = state
+            .settings_cache
+            .as_deref()
+            .map(|s| s as &(dyn crate::db::SettingsStore + Send + Sync));
+        match crate::bridge::try_resolve_inline_approval_gate(
+            &user.user_id,
+            "gateway",
+            ironclaw_engine::ThreadId(thread_uuid),
```
serrrfirat: **High Severity** — Inline approval fast-path uses the wrong thread identity

`req.thread_id` here is the browser/channel-visible thread id from the pending gate/history contract, not necessarily the internal engine `ThreadId` that keys `PendingGateStore`. `PendingGateView` is built from `effective_wire_thread_id()` (external `scope_thread_id` when present), while `take_verified` looks up by internal `PendingGateKey { user_id, thread_id }`.

In the normal web flow (`/api/chat/thread/new` → `/api/chat/send` with that UUID → pending gate → `/api/chat/approval` with the returned `pending_gate.thread_id`), the fast-path constructs `ironclaw_engine::ThreadId(thread_uuid)` from the web conversation UUID. That misses the pending row, maps `GateStoreError::NotFound` to `InlineGateError::Other`, and returns 500 instead of falling through to the legacy request-id scan. This breaks the primary advertised approval path whenever the wire thread id is a scoped/external id.

Please resolve the pending gate using the existing scope-aware matching (`scope_thread_id`, internal engine `thread_id`, or `conversation_id`) before calling `take_verified`, or change the inline API to accept request id + optional wire scope and locate the actual pending gate internally.

src/bridge/router.rs
```diff
@@ -3928,6 +4199,335 @@ fn spawn_deferred_context_cleanup(
...
+                                    tool_name: pending.action_name.clone(),
+                                    description: pending.description.clone(),
+                                    parameters: pending
+                                        .display_parameters
+                                        .clone()
+                                        .unwrap_or_else(|| pending.parameters.clone()),
+                                    allow_always: *allow_always,
+                                }
+                            }
+                            _ => StatusUpdate::AuthRequired {
```
serrrfirat: **Medium Severity** — Follow-up gates lose their real resume metadata in the post-park continuation

This continuation re-implements pending-gate notification instead of using the existing gate-status helpers. In the non-Approval branch it converts every follow-up gate into `StatusUpdate::AuthRequired` with `instructions: pending.description` and `auth_url: None`.

For an actual `ResumeKind::Authentication { instructions, auth_url, .. }`, that drops the real auth instructions and OAuth URL, so the user can get an unactionable prompt after the first inline approval resumes into an auth gate. For `ResumeKind::External`, it also misclassifies the gate as an auth prompt even though the normal `send_pending_gate_status` path ignores External gates.

Please reuse the existing `notify_pending_gate` / `send_pending_gate_status` path here (or match `ResumeKind::Authentication` explicitly and carry `instructions`/`auth_url` through), and avoid sending a prompt if `pending_gates.insert` failed because the resulting request id cannot be resolved.

src/channels/web/features/chat/mod.rs
```diff
@@ -187,6 +187,66 @@ pub(crate) async fn chat_approval_handler(
...
+    {
+        let resolution = if approved {
+            ironclaw_engine::GateResolution::Approved { always }
+        } else {
+            ironclaw_engine::GateResolution::Denied { reason: None }
+        };
+        let settings_store: Option<&(dyn crate::db::SettingsStore + Send + Sync)> = state
+            .settings_cache
+            .as_deref()
+            .map(|s| s as &(dyn crate::db::SettingsStore + Send + Sync));
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in 5b147264. The inline fast path now mirrors the legacy mpsc path's settings precedence by calling `resolve_settings_store(&state).ok()` (cache first, falling back to raw `state.store`). A gateway with a DB but no `CachedSettingsStore` will now correctly persist `tool_permissions.<tool>=always_allow` on `action="always"`.

src/bridge/router.rs
```diff
@@ -11278,4 +11940,725 @@ mod tests {
...
+            .expect_err("cross-channel resolve must error");
+            assert!(
+                matches!(err, super::InlineGateError::ChannelMismatch { .. }),
+                "channel-mismatch must surface as InlineGateError::ChannelMismatch; got: {err:?}"
+            );
+
+            // Parked alpha must NOT have woken — the rejection happens
+            // inside take_verified, before any try_deliver call.
+            let still_parked =
+                tokio::time::timeout(Duration::from_millis(200), &mut { pause_task })
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in 5b147264. Bound `pause_task` as `mut` and pass `&mut pause_task` directly, matching the pattern in `try_resolve_inline_approval_gate_parallel_threads_same_user_independent`. (Note: dropping a tokio `JoinHandle` only detaches — it doesn't abort — so the previous `&mut { pause_task }` was confusing rather than incorrect, but the cleaner form keeps the intent obvious and the explicit `controller.cancel_thread` cleanup still load-bearing.)

src/channels/web/features/chat/mod.rs
ilblackdragon: Fixed in 5b147264. Good catch — this was the dominant failure mode for the gateway approval path whenever the engine `ThreadId` differed from the wire/scope id.

Took the second approach you suggested: dropped `thread_id` from `try_resolve_inline_approval_gate` entirely and resolve the gate by `request_id` (system-wide unique). To keep that lookup atomic with `take_verified`'s remove, added `PendingGateStore::take_verified_by_request_id` which performs lookup → user-ownership check → channel/expiry verification → remove under a single mutex. A new `GateStoreError::Unauthorized` (mapped to `InlineGateError::Unauthorized` → HTTP 403) is used for cross-tenant request-id probes so a foreign user can't observe gate existence.

Regression coverage: `try_resolve_inline_approval_gate_resolves_when_scope_id_differs_from_thread_id` parks a pause whose `scope_thread_id` is a fresh UUID different from the engine `ThreadId`, asserts the wire-effective id surfaces as the scope id (the exact shape that broke before), then verifies the inline path resolves it via `request_id` only. Plus three store-level tests for `take_verified_by_request_id` covering success, cross-user 403, and channel-mismatch leaving the gate intact.

src/bridge/router.rs
ilblackdragon: Fixed in 5b147264. The post-park continuation now matches `ResumeKind` variants explicitly rather than falling everything through the catch-all `AuthRequired` arm:

- `Approval { allow_always }` → `StatusUpdate::ApprovalNeeded` (unchanged).
- `Authentication { instructions, auth_url, .. }` → `StatusUpdate::AuthRequired` carrying the **real** `instructions` and `auth_url` from `resume_kind` instead of `pending.description` and `auth_url: None`. So a follow-up auth gate's OAuth URL no longer gets dropped.
- `External { .. }` → no status emitted, matching `send_pending_gate_status`'s deliberate no-op for external callbacks.

Also tracks the `pending_gates.insert` outcome and skips the entire status emit on insert failure — without a row backing the `request_id`, the user has nothing to resolve against, so emitting a card would just dead-end the click.

I considered routing through `notify_pending_gate` directly (`send_pending_gate_status` takes `&Agent` and `agent.channels`, while this continuation has `Arc<ChannelManager>` cloned out for spawn), and went with the inline match to keep the change scoped — same semantics, no helper-signature churn. Happy to extract a `ChannelManager`-only helper as a follow-up if you'd prefer the unified call site; it'd be the natural next step alongside the existing `notify_pending_gate` / `send_pending_gate_status` cluster.

PR #3374, opened 2026-05-07
fix: multi-tenant workspace memory isolation

serrrfirat:
## Summary
- Rebind private workspace memory layers to the authenticated/session user instead of preserving startup owner scopes.
- Centralize per-user workspace scoping for agent prompt, skill marker, extracted document, and compaction paths.
- Add regressions proving default private layers do not read owner memory, explicit shared layers remain shared, and conversation-derived writes land in the sender/session workspace.

## Tests
- `cargo fmt --check`
- `cargo check --quiet`
- `cargo test --test workspace_scoped_rebind --no-default-features --features libsql --quiet`
- `cargo test --lib --no-default-features --features libsql channels::web::tests::multi_tenant::workspace_pool --quiet`
- `cargo test --lib --no-default-features --features libsql process_compact_writes_summary_to_session_user_workspace --quiet`
- `cargo test --lib --no-default-features --features libsql store_extracted_documents_writes_to_message_user_workspace --quiet`
- `cargo clippy --no-default-features --features libsql --lib -- -D warnings`

Note: `cargo clippy --no-default-features --features libsql --lib --tests -- -D warnings` is currently blocked by a pre-existing `clippy::duplicated_attributes` warning in `tests/support/live_mission_helpers.rs` / `tests/support/mod.rs`, unrelated to this change.


src/workspace/mod.rs
```diff
@@ -751,18 +751,26 @@ impl Workspace {
     pub fn scoped_to_user(&self, user_id: impl Into<String>) -> Self {
         let user_id = user_id.into();
 
+        let private_source_scopes: Vec<String> = self
+            .memory_layers
+            .iter()
+            .filter(|layer| layer.sensitivity == crate::workspace::layer::LayerSensitivity::Private)
+            .map(|layer| layer.scope.clone())
+            .collect();
```
zetyquickly, line 759: `private_source_scopes` filters `read_user_ids` by scope-string equality. If a shared layer ever shared a scope name with a private layer, it'd get dropped on rebind. Doesn't affect current callers but worth a comment noting the assumption, or matching on layer name + sensitivity instead.

src/agent/agent_loop.rs
```diff
@@ -2422,6 +2425,71 @@ mod tests {
         )
     }
 
+    #[cfg(feature = "libsql")]
+    async fn make_agent_loop_test_db() -> (Arc<dyn crate::db::Database>, tempfile::TempDir) {
```
zetyquickly: nit: `make_agent_loop_test_db` and `make_thread_ops_test_db` (in `thread_ops.rs`) are identical — could move to a shared test helper if either file gets touched again.

tests/workspace_scoped_rebind.rs
```diff
@@ -3,7 +3,7 @@
 //! Issue: <https://github.com/nearai/ironclaw/issues/1652>
 //!
 //! `scoped_to_user()` rebinds a workspace to a different primary user while
-//! preserving shared read scopes. These 7 tests verify the rebinding contracts
+//! preserving shared read scopes. These tests verify the rebinding contracts
```
zetyquickly: nit: the comment now says "These tests" — the new test brings the count to 8 if you want to keep it accurate.

zetyquickly (approved):
LGTM. Correct fix at the right layers — pool rebinds private scopes for fresh workspaces, `scoped_to_user` handles the case where a workspace is re-projected from a non-matching primary, and `workspace_for_user` centralizes the four message-driven call sites. Tests are caller-level and assert the negative (owner scope must not receive the per-user write), which is the right shape for this bug class.

One follow-up question and a couple of nits inline — none blocking.

Follow-up (not in diff): `process_heartbeat` in `src/agent/commands.rs` still calls `self.workspace()` directly. On a multi-tenant gateway, would a non-owner running `/heartbeat` hit the owner's `HEARTBEAT.md`? Either intentional (system-scoped heartbeat) or the same bug shape this PR fixes — worth a follow-up issue if not in scope here.

zetyquickly (dismissed):

src/workspace/mod.rs
serrrfirat: Fixed in 4a25f0185: `scoped_to_user` now keeps non-private layer scopes even when their scope string collides with a private source scope, with regression coverage in `workspace_scoped_rebind`.

src/agent/agent_loop.rs
serrrfirat: Fixed in 4a25f0185: moved the duplicated libSQL test DB setup into `agent::test_support::make_libsql_test_db` and updated the three callers.

tests/workspace_scoped_rebind.rs
serrrfirat: Fixed in 4a25f0185: changed the header wording to count-neutral so it stays accurate as coverage grows.

src/agent/thread_ops.rs
```diff
@@ -2873,6 +2876,70 @@ mod tests {
...
+            .read(&path)
+            .await
+            .expect("compaction summary should be written to the session user's workspace");
+        assert!(alice_doc.content.contains("Context Summary"));
+
+        let owner_ws = crate::workspace::Workspace::new_with_db("owner-scope", Arc::clone(&db));
+        assert!(
+            owner_ws.read(&path).await.is_err(),
+            "compaction summary must not be written to the startup owner workspace"
+        );
```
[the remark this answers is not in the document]
serrrfirat: Fixed in f1d3c94e0: the compaction test now captures UTC dates before and after the write and probes both candidate daily paths before asserting owner isolation.

src/agent/agent_loop.rs
```diff
@@ -2422,6 +2425,71 @@ mod tests {
...
+                duration_secs: None,
+            }]);
+
+        agent.store_extracted_documents(&message).await;
+
+        let date = chrono::Utc::now().format("%Y-%m-%d");
+        let path = format!("documents/{date}/conversation-notes.txt");
+        let alice_ws = crate::workspace::Workspace::new_with_db("alice", Arc::clone(&db));
+        let alice_doc = alice_ws
+            .read(&path)
```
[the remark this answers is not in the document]
serrrfirat: Fixed in f1d3c94e0: the extracted-document test now captures UTC dates before and after storage and probes both candidate document paths before asserting owner isolation.

think-in-universe (dismissed):
LGTM. Let's test more before ship.

think-in-universe (dismissed):

think-in-universe (dismissed):

think-in-universe (approved):

PR #3396, opened 2026-05-08
feat: add pairing_approve tool for Slack binding via chat

PierreLeGuen:
## Summary
Users can now paste their Slack pairing code in the IronClaw chat and the LLM will call `pairing_approve` to bind their accounts automatically.

When a user receives a pairing code in Slack (e.g. `WZG8LQAB`), they say something like "bind my Slack with code WZG8LQAB" in the IronClaw web UI, and the agent calls the tool to complete the pairing.

## Changes
- New `pairing_approve` tool in `src/tools/builtin/pairing.rs`
- Registered in `app.rs` when a PairingStore is available

## Test plan
- [ ] User receives pairing code in Slack DM
- [ ] User pastes code in IronClaw chat
- [ ] Agent calls pairing_approve tool
- [ ] Subsequent Slack messages resolve to the correct user

### DIFF
--- a/src/tools/builtin/pairing.rs
+++ b/src/tools/builtin/pairing.rs
@@ -5,6 +5,8 @@ use crate::context::JobContext;
 use crate::pairing::PairingStore;
 use crate::tools::tool::{ApprovalRequirement, Tool, ToolError, ToolOutput, require_str};
 
+const CHANNEL: &str = "slack-relay";
+
 pub struct PairingApproveTool {
     store: Arc<PairingStore>,
 }
@@ -22,7 +24,7 @@ impl Tool for PairingApproveTool {
     }
 
     fn description(&self) -> &str {
-        "Approve a channel pairing code to bind an external account (e.g. Slack) to the current IronClaw user. The user receives the code in their messaging app and provides it here."
+        "Approve a Slack pairing code to bind the user's Slack account to their IronClaw user. The user receives the code in Slack and provides it here."
     }
 
     fn parameters_schema(&self) -> serde_json::Value {
@@ -31,12 +33,7 @@ impl Tool for PairingApproveTool {
             "properties": {
                 "code": {
                     "type": "string",
-                    "description": "The pairing code received in the messaging app (e.g. WZG8LQAB)"
-                },
-                "channel": {
-                    "type": "string",
-                    "description": "The channel to pair with (default: slack-relay)",
-                    "default": "slack-relay"
+                    "description": "The pairing code received in Slack (e.g. WZG8LQAB)"
                 }
             },
             "required": ["code"]
@@ -51,10 +48,7 @@ impl Tool for PairingApproveTool {
         let start = std::time::Instant::now();
 
         let code = require_str(&params, "code")?;
-        let channel = params
-            .get("channel")
-            .and_then(|v| v.as_str())
-            .unwrap_or("slack-relay");
+        let channel = CHANNEL;
 
         let user_id =
             crate::ownership::UserId::new(&ctx.user_id, crate::ownership::UserRole::Regular)
@@ -82,25 +76,43 @@ impl Tool for PairingApproveTool {
     }
 
     fn requires_approval(&self, _params: &serde_json::Value) -> ApprovalRequirement {
-        ApprovalRequirement::UnlessAutoApproved
+        ApprovalRequirement::Always
     }
 }
 
 #[cfg(test)]
 mod tests {
     use super::*;
+    use crate::tools::registry::is_protected_tool_name;
 
     #[test]
     fn tool_metadata() {
         let store = Arc::new(PairingStore::new_noop());
         let tool = PairingApproveTool::new(store);
         assert_eq!(tool.name(), "pairing_approve");
-        assert!(tool.description().contains("pairing code"));
+        assert!(tool.description().contains("Slack"));
         let schema = tool.parameters_schema();
         assert!(schema["properties"]["code"].is_object());
+        assert!(schema["properties"]["channel"].is_null());
+    }
+
+    #[test]
+    fn always_requires_approval() {
+        let store = Arc::new(PairingStore::new_noop());
+        let tool = PairingApproveTool::new(store);
         assert_eq!(
             tool.requires_approval(&serde_json::json!({})),
-            ApprovalRequirement::UnlessAutoApproved
+            ApprovalRequirement::Always
         );
+    }
+
+    #[test]
+    fn is_protected_builtin() {
+        assert!(is_protected_tool_name("pairing_approve"));
+    }
+
+    #[test]
+    fn channel_is_slack_relay() {
+        assert_eq!(CHANNEL, "slack-relay");
     }
 }
--- a/src/tools/builtin/pairing.rs
+++ b/src/tools/builtin/pairing.rs
@@ -85,3 +85,22 @@ impl Tool for PairingApproveTool {
         ApprovalRequirement::UnlessAutoApproved
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn tool_metadata() {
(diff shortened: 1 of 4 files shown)


src/tools/builtin/pairing.rs
```diff
@@ -0,0 +1,83 @@
...
+            .and_then(|v| v.as_str())
+            .unwrap_or("slack-relay");
+
+        let user_id = crate::ownership::UserId::new(
+            &ctx.user_id,
+            crate::ownership::UserRole::Regular,
+        )
+        .map_err(|e| ToolError::ExecutionFailed(format!("invalid user_id: {e}")))?;
+
+        match self.store.approve(channel, code, &user_id).await {
```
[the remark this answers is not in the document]
PierreLeGuen: Fixed. Added `requires_approval` returning `ApprovalRequirement::UnlessAutoApproved` — the tool now requires explicit user confirmation before binding the account.

serrrfirat:
Paranoid strategy audit for PR #3396 (`feat: add pairing_approve tool for Slack binding via chat`) at head `2254c5996b0c05b64c78dc6df8d11b1503021480`.

I am **not 100% factually confident this is safe to merge as-is**. Recommended strategy: fix the issues below, update/rebase onto current `main`, rerun CI + targeted tests, then rerun paranoid review on the new head.

### Findings / risks

1. **Medium — `pairing_approve` can bypass explicit human confirmation in auto/autonomous modes**
   - `PairingApproveTool::requires_approval` returns `ApprovalRequirement::UnlessAutoApproved` (`src/tools/builtin/pairing.rs`).
   - `UnlessAutoApproved` is explicitly bypassable by auto-approve/session settings and autonomous contexts. Account/channel identity binding is security-sensitive enough that it should require fresh explicit approval.
   - **Fix:** make the tool require non-bypassable explicit approval, likely `ApprovalRequirement::Always`, and add tests proving global/session auto-approve and autonomous execution cannot silently approve a pairing.

2. **Medium — tool path bypasses existing web pairing orchestration / rollback**
   - New tool only calls `PairingStore::approve(...)` and returns success.
   - Existing web handler also calls `complete_pairing_approval(...)`, reverts durable approval if runtime propagation fails, broadcasts state, and dispatches follow-up callbacks.
   - For non-relay/WASM channels, chat approval can leave DB state approved while live channel runtime remains unbound until restart.
   - **Fix:** share the web approval orchestration/service from the tool, including runtime propagation and revert-on-failure. Add a test covering propagation failure rollback.

3. **Medium — generic `channel` parameter overclaims Slack-only safety**
   - PR body describes Slack pairing, but tool schema exposes arbitrary `channel` with default `slack-relay`.
   - If this is intended to be Slack-only, arbitrary channel approval is unnecessary risk. If generic, it needs the full orchestration path above.
   - **Fix:** either lock/allowlist to `slack-relay`, or implement generic channel approval correctly via shared orchestration.

4. **Medium — new built-in is missing from protected built-in tool list**
   - `src/tools/registry.rs` says new built-ins must be added to `PROTECTED_TOOL_NAMES`.
   - `pairing_approve` is not in that list, so dynamic registrations/self-repair protections may not treat it as protected.
   - **Fix:** add `"pairing_approve"` to `PROTECTED_TOOL_NAMES` and add a regression test.

5. **Verification gaps**
   - New test only checks tool metadata / approval enum.
   - Missing coverage: DB-backed execution, protected-name registration, approval-gate bypass prevention, runtime propagation rollback, Slack relay flow from code issuance to later sender resolution.
   - PR base is older than current `main`; update/rebase and rerun checks.

### Suggested next step

Do not merge this head. Fix above, add targeted tests, update onto current `main`, then rerun CI and another paranoid review on the new exact head.

PierreLeGuen:
All five findings addressed in the latest push:

1. **Always approval** — Changed from `UnlessAutoApproved` to `ApprovalRequirement::Always`. Cannot be bypassed by auto-approve or autonomous modes. Test added.

2. **Orchestration gap** — Acknowledged. For the relay channel specifically, `PairingStore::approve` is sufficient because relay identity resolution uses `PairingStore::resolve_identity` directly (no WASM runtime propagation needed). The web handler's `complete_pairing_approval` path is for WASM channels that need runtime restart. This tool is Slack-only and doesn't need that path. If we add generic channel support later, we'll share the orchestration.

3. **Channel locked to slack-relay** — Removed the `channel` parameter entirely. Hardcoded to `CHANNEL = "slack-relay"` constant.

4. **Added to PROTECTED_TOOL_NAMES** — `pairing_approve` is now in the protected list. Test added.

5. **Tests added** — 4 tests: metadata, always-approval, protected-name, channel constant. Rebased onto current main.

src/tools/builtin/pairing.rs
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    async fn execute(
+        &self,
+        params: serde_json::Value,
+        ctx: &JobContext,
+    ) -> Result<ToolOutput, ToolError> {
+        let start = std::time::Instant::now();
+
+        let code = require_str(&params, "code")?;
```
serrrfirat, line 50: **Medium Severity** — Pairing code should be redacted in observability paths

`code` is a pairing secret, but `PairingApproveTool` does not override `sensitive_params()`. Tool parameters can flow through logs/action records/hooks/approval UI/session state before execution, so a pasted Slack pairing code may be persisted or displayed in places that should not retain secrets.

Suggestion: mark the parameter sensitive:

```rust
fn sensitive_params(&self) -> &[&str] {
    &["code"]
}
```

Please add a regression test asserting `code` is treated as sensitive.

src/tools/builtin/pairing.rs
```diff
@@ -0,0 +1,118 @@
...
+
+        match self.store.approve(channel, code, &user_id).await {
+            Ok(approval) => {
+                let msg = format!(
+                    "Pairing approved! Your {} account (external ID: {}) is now linked to your IronClaw user.",
+                    approval.channel, approval.external_id
+                );
+                Ok(ToolOutput::text(&msg, start.elapsed()))
+            }
+            Err(e) => {
```
serrrfirat, line 65: **Medium Severity** — Internal approval failures are returned as successful tool output

This branch converts every `PairingStore::approve` error into `Ok(ToolOutput::text(...))`. That makes a DB pool/query failure look like an invalid or expired code, and the dispatcher records the tool call as successful even though pairing could not be attempted reliably.

Suggestion: keep `NotFound` user-facing as an invalid/expired code, but propagate infrastructure errors as `ToolError::ExecutionFailed`, e.g. match `DatabaseError::NotFound { .. }` separately and return `Err(...)` for other variants. Add tests for invalid-code vs DB/internal failure behavior.

src/tools/builtin/pairing.rs
```diff
@@ -0,0 +1,118 @@
...
+    fn requires_sanitization(&self) -> bool {
+        false
+    }
+
+    fn requires_approval(&self, _params: &serde_json::Value) -> ApprovalRequirement {
+        ApprovalRequirement::Always
+    }
+}
+
+#[cfg(test)]
```
serrrfirat, line 83: **Medium Severity** — Security-sensitive tool lacks execution-path tests

Current tests cover metadata, approval requirement, protected-name registration, and channel constant, but not DB-backed execution of the tool itself. For an auth/account-binding tool, metadata-only coverage misses the important behavior: success binds the authenticated `ctx.user_id`, invalid codes are handled safely, DB/internal errors are surfaced correctly, and secret params are redacted.

Suggestion: add tool-level tests that execute `pairing_approve` against a real test pairing store for success + invalid code, plus a failure-path test (or mock/faulty store if needed) for internal DB errors, and a redaction test for `sensitive_params()`.

serrrfirat (approved):
Approved. Findings posted as non-blocking follow-ups: redact pairing code parameter, distinguish invalid-code vs internal DB errors, and add execution-path tests. Prior blocker fixes are present on current head (Always approval, Slack-only channel, protected tool name).

PR #3399, opened 2026-05-08
test(bridge): contract test for deny-message vs execution-intent classifier

### LINKED ISSUE #2787: Add contract test: deny message must not trigger user_signals_execution_intent
## Context

PR #2744 reworded the denied-action message in `src/bridge/router.rs:2571` from:

> "Do not execute it; choose an alternative approach."

to:

> "Do not retry; choose a different approach."

Reason: the old wording contained `"execute it"`, which `user_signals_execution_intent` (in `src/llm/reasoning.rs`) classified as fresh user execution intent. When this message was fed back into the engine on resume, it re-armed `require_action_attempt` and nudged the LLM to issue another tool call — exactly the opposite of a denial.

## Problem

The reword is correct, but the invariant it relies on ("deny message must not match `EXEC_PHRASES`") is not pinned anywhere. Any future rephrasing — by a human or an LLM — can silently re-introduce the bug. The current guard is "we happen to have tested the e2e flow after writing this particular string."

## Ask

Add a Rust unit test that:

1. Imports the actual deny-message template used by `resolve_gate` in `src/bridge/router.rs`.
2. Formats it with a representative `action_name` and `reason` (including the empty-reason branch).
3. Feeds the formatted string through `user_signals_execution_intent` and asserts `false`.

This locks the contract regardless of future wording changes — rephrase the deny message and the test fails at CI, not in staging e2e.

## Acceptance

- Unit test lives next to the deny-message construction site (or in `src/llm/reasoning.rs` tests if the template can be extracted as a `const`).
- Test is red against the pre-PR-#2744 wording and green against the post-PR wording (sanity check during review).
- Extracted deny-message template, if introduced, is the single source of truth read by both production code and the test.

## References

- PR #2744 (source of the reword)
- `src/bridge/router.rs:2571` (current deny-message site)
- `src/llm/reasoning.rs` (`user_signals_execution_intent`, `EXEC_PHRASES`)

nlok5923:
## Summary

- Extracts the denied-action message format into `format_deny_message` in `src/bridge/router.rs` so production and tests read from the same template.
- Adds a contract test that asserts the formatted deny-message does not trigger `ironclaw_llm::user_signals_execution_intent`, locking the invariant established by PR #2744.

Closes #2787.

## Why

The denied-action message recorded on the thread when an approval gate is denied must not match any phrase in `ironclaw_llm::user_signals_execution_intent`'s `EXEC_PHRASES`. If it does, the resume handler interprets the denial as fresh user execution intent, re-arms `require_action_attempt`, and the LLM issues another tool call — exactly the opposite of a denial. PR #2744 reworded the message from `"Do not execute it; choose an alternative approach."` to `"Do not retry; choose a different approach."` to fix this, but the invariant was not pinned anywhere — any future rephrasing (human or LLM) could silently re-introduce the bug.

## Test plan

- [x] `cargo test --lib bridge::router::tests::deny_message_does_not_signal_execution_intent` passes.
- [x] All 97 existing `bridge::router::tests` still pass.
- [x] `cargo fmt -- --check` clean.
- [x] `cargo clippy --lib --tests -- -D warnings` clean.
- [x] Manual sanity check: feeding the pre-PR-#2744 wording (`"Do not execute it; choose an alternative approach."`) through `EXEC_PHRASES` matches `"execute it"` — the test would fail against the old wording, confirming it actually fences the contract.
- [x] Test covers both empty-reason and with-reason branches, plus action names that themselves contain `EXEC_PHRASES` substrings (`execute_query`, `run_the_pipeline`, `send_the_message`) so a future template change cannot bypass the check via interpolation.

### DIFF
--- a/src/bridge/router.rs
+++ b/src/bridge/router.rs
@@ -2404,6 +2404,23 @@ fn gate_view_is_authentication(gate: &crate::gate::pending::PendingGateView) -> 
     )
 }
 
+/// Format the user-visible message recorded on the thread when an approval
+/// gate is denied.
+///
+/// Wording is deliberately constrained to avoid every phrase in
+/// `ironclaw_llm::user_signals_execution_intent`'s `EXEC_PHRASES` list. If
+/// the deny-message text matches a phrase from that list the resume handler
+/// re-arms the `require_action_attempt` obligation and the LLM is nudged
+/// into another tool call — exactly the opposite of what a denial means.
+/// `deny_message_does_not_signal_execution_intent` in this module's test
+/// suite locks that contract in.
+fn format_deny_message(action_name: &str, reason: Option<&str>) -> String {
+    let trailer = reason.map(|r| format!(" Reason: {r}")).unwrap_or_default();
+    format!(
+        "User denied action '{action_name}'. Do not retry; choose a different approach.{trailer}"
+    )
+}
+
 async fn hinted_pending_gate_thread_id(
     user_id: &str,
     conversation_scope: Option<&str>,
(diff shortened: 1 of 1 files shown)


serrrfirat:

This test protects message formatting in the retired v1 bridge/router and its execution-intent classifier. Reborn does not route approval denial text back through that bridge; typed gate resolution and turn transitions own this behavior.

The invariant is still worthwhile. A Reborn test should drive a denial through ProductWorkflow and the approval interaction service, then prove it cannot be reinterpreted as a new execution request or user instruction.

serrrfirat:

This protects formatting in the retired v1 bridge/router. Reborn uses typed approval interaction and turn transitions instead of routing denial text through that classifier.

A useful Reborn test would drive denial through ProductWorkflow and prove it cannot be reinterpreted as a new execution request.

serrrfirat:

This tests formatting in the retired v1 bridge/router. Reborn uses typed approval interaction and turn transitions instead of routing denial text through that classifier.

A useful Reborn test would drive denial through ProductWorkflow and prove it cannot become a new execution request.

PR #3418, opened 2026-05-09
fix(mission): route outcome notifications to originating chat thread

### LINKED ISSUE #3415: Bug: Mission results are posted to the wrong conversation instead of the mission-creation conversation
## Description
In Production, a daily NYC weather mission is posting its results to a different conversation than the one where the mission was created.

For `tall-elk-vuvim` (`v0.27.0`), the weather task was created from one conversation, but the daily weather results are being pushed into another conversation that appears to be my first/default conversation instead of the mission’s original conversation.

I have also observed the same issue on another agent, `gold-ram-pofoh` (`v0.26.0`).

Image

Image

italic-jinxin:
## Summary

- Mission outcome notifications now route back to the chat thread where the user fired the mission, instead of dumping into the user's assistant conversation.
- Adds `parent_thread_id` to `Mission` (captured at create time) and `source_conversation_thread_id` to `ThreadExecutionContext` (plumbed via thread metadata) so `handle_mission_notification` can resolve the right v1 conversation row across all four output paths (channel broadcast, SSE `Response`, v2 `record_external_agent_message`, v1 `add_conversation_message`).
- Cron / learning / API-imported missions still fall back to the assistant conversation when there is no originating chat thread.

## Bug

Symptom: user fires a mission in chat thread A → mission completes → `[mission name] result` card appears in the user's assistant conversation (visible only when they happen to open it), never in thread A. Persists across page refresh because the v1 DB write is what populates `/api/chat/history`.

## Root cause (two layers)

**Mission data model lost the parent.** [`Mission`](crates/ironclaw_engine/src/types/mission.rs#L103) had no field referencing the originating conversation. `MissionNotification` carried only the *internal* execution thread ID (the freshly-spawned thread the mission goal runs in), so [`handle_mission_notification`](src/bridge/router.rs#L4131) had nothing to route to except that internal thread — not visible to the user anywhere.

**The first attempted fix used the wrong thread ID.** Setting `parent_thread_id = Some(context.thread_id)` from `effect_adapter::mission_create` looked correct but failed silently: `context.thread_id` is the *v2 ephemeral engine thread* (`Thread::new() → ThreadId::new()` per turn), not the chat URL parameter (v1 `conversations.id`). The v1 write `add_conversation_message(parent.0, ...)` then violated the `conversation_messages.conversation_id → conversations(id)` FK constraint. The defensive fallback dumped output to the assistant conversation, exactly reproducing the original symptom.

The bridge router already encodes the chat thread into the v2 `channel_key` as `<channel>:<scope_uuid>` ([`router.rs`](src/bridge/router.rs#L3527)). The fix splits that scope back out in [`ConversationManager::handle_user_message`](crates/ironclaw_engine/src/runtime/conversation.rs#L295), stores it as `source_conversation_thread_id` in thread metadata, exposes it on `ThreadExecutionContext`, and consumes it in `effect_adapter::mission_create`.

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] CI/Infrastructure
- [ ] Security
- [ ] Dependencies

## Linked Issue

Fixes #3415 

## Validation

- [ ] `cargo build`
- [ ] Relevant tests pass: 
- [ ] Manual testing: 

## Review Follow-Through

---

**Review track**:


serrrfirat:
Parallel review findings:

1. **High — Security / cross-chat leakage** — `src/bridge/router.rs:4207`

   `handle_mission_notification` trusts `notif.parent_thread_id` and writes mission output directly with `db.add_conversation_message(parent.0, ...)` before any owner/channel check. That parent is derived from the user-facing `conversation_scope` embedded into the engine channel key (`src/bridge/router.rs:3532` -> `crates/ironclaw_engine/src/runtime/conversation.rs:300`), and web send forwards `req.thread_id` into `IncomingMessage` at `src/channels/web/features/chat/mod.rs:101` without the ownership check used by history reads. If a caller supplies another user’s existing conversation UUID, the mission output can be inserted into the victim chat.

   Suggested fix: validate `parent_thread_id` against the persisted conversation owner/channel before using it for SSE/thread routing or v1 DB writes; otherwise skip parent routing and fall back to the assistant conversation. Add a caller-level regression from `chat_send_handler` / mission creation proving an unowned `thread_id` cannot receive mission output.

2. **Medium — Correctness / thread-scoped follow-up context** — `src/bridge/router.rs:4233`

   v2 conversation recording still uses `get_or_create_conversation(channel_name, &notif.user_id)` for mission output. Web turns with a chat thread use scoped conversation keys (`format!("{}:{}", message.channel, tid)` at `src/bridge/router.rs:3532-3541`), and the new parent metadata is extracted from that scoped key. Recording the notification into plain `gateway` means the originating `gateway:<parent_thread_id>` conversation does not get the `record_external_agent_message` entry, so the next reply in that chat can still spawn without mission result history.

   Suggested fix: when `notif.parent_thread_id` is present, resolve/record into the same scoped conversation key used by the caller (for gateway: `format!("{channel_name}:{}", parent.0)`) after ownership validation; otherwise keep the unscoped fallback. Add a caller-level follow-up test asserting spawned thread history includes the mission output.

henrypark133 (approved):

What looks good:
- The old `context.thread_id` fallback issue is fixed: `src/bridge/effect_adapter.rs:832` now passes only `source_conversation_thread_id`.
- Mission notifications prefer `parent_thread_id` and fall back cleanly when no originating chat thread exists.
- CI is green.

No verified findings.

Residual risk:
- Older/pre-existing engine threads without the new metadata still fall back to the assistant conversation. I am treating that as a compatibility gap rather than a new regression in this PR.

PR #3421, opened 2026-05-09
Add Reborn shared storage substrate

### LINKED ISSUE #3419: Define shared Reborn storage substrate and secret-store boundary
## Problem

Reborn persistence has grown several domain-local SQL adapters (`db.rs`, `libsql.rs`, `postgres.rs`) across crates such as turns, threads, run-state, authorization, outbound, conversations, memory, secrets, and event stores.

That split was useful for moving quickly and preserving domain invariants, but it creates architecture pressure:

- repeated SQL/backend plumbing across crates;
- repeated libSQL/Postgres parity work;
- unclear boundary between filesystem, structured stores, and secrets;
- harder production composition/migration story;
- reviewers reasonably expect a clearer storage abstraction than “SQL files everywhere”.

We also want to explore whether storage can feel more filesystem-like without flattening all domain semantics into generic file reads/writes.

## Direction

Adopt a layered storage architecture:

```text
Domain store traits
  TurnStore / SessionThreadService / SecretStore / OutboundStateStore / etc.
        |
        v
Shared persistence substrate
  records + blobs + append logs + locks + transactions + migrations
        |
        v
Backends
  Postgres / libSQL / filesystem / object store / memory-for-tests
        |
        v
Optional filesystem-like virtual views
  diagnostics / import-export / admin projection / file-shaped state
```

The goal is one shared persistence substrate and optional virtual file views, **not** one universal filesystem trait replacing typed domain APIs.

## Architecture principle

Separate these concepts:

```text
Domain ownership
  -> owns schema meaning, invariants, validation, and operations

Persistence substrate
  -> owns common mechanics: records, blobs, append logs, locks, transactions, migrations, backend identity, redacted errors

Filesystem abstraction
  -> owns file-shaped content, virtual path authority, mount views, and optional projections
```

A filesystem-like interface can exist as a projection/import-export/admin layer over domain stores. It should not become the only mutation/read authority for structured control-plane state.

## Candidate substrate traits

Instead of a single giant `Persistence` trait, use a small family of primitive traits:

```rust
trait BlobStore {
    // file-shaped or object-shaped bytes; can back artifacts and encrypted blobs
}

trait RecordStore {
    // keyed structured records with optional compare-and-swap/versioning
}

trait AppendLog {
    // ordered event/audit/projection streams with replay cursors
}

trait LockStore {
    // leases, heartbeats, ownership, fencing tokens
}

trait TransactionalStore {
    // backend transaction boundary for multi-record operations
}
```

Domain stores compose these primitives where useful, but domain operations remain typed:

```rust
TurnStateStore::claim_next_run(...)
SessionThreadService::accept_inbound_message(...)
OutboundStateStore::advance_subscription_cursor(...)
SecretStore::get_secret(...)
```

## Filesystem-like views

Filesystem-like views are useful and should be supported where they help operators and agents:

```text
/reborn/threads/{thread}/messages/{message}.json
/reborn/turns/{turn}/runs/{run}.json
/reborn/outbound/subscriptions/{subscription}.json
/reborn/events/{stream}/{cursor}.json
/secrets/... redacted metadata projection only
```

Rules for these views:

- may be read-only diagnostics by default;
- mutation, if allowed, validates and calls typed domain APIs;
- must preserve tenant/user/project/agent/thread scope checks;
- must not expose raw host paths, raw runtime payloads, provider/backend error details, or secrets;
- must not cause runtime/control-plane state to be indexed as semantic memory by accident.

## Secrets boundary

Secrets need a dedicated semantic boundary above raw persistence:

```text
SecretStore / Credential*Store
  -> authorization assumptions, expiry/use-count semantics, encryption/decryption, rotation, audit
  -> plaintext only crosses this API after policy checks

RawEncryptedBlobStore / RawSecretRecordStore
  -> ciphertext + metadata only
  -> can be backed by filesystem, SQL, or object store
  -> no plaintext API, no secret semantics
```

`ironclaw_filesystem` may implement raw encrypted blob reads/writes, but must not own secret semantics or expose secret material through generic file listing/read APIs.

## Non-goals

- Do not make `ironclaw_filesystem` the universal persistence layer.
- Do not replace typed domain store APIs with generic `read(path)` / `write(path)` calls.
- Do not move turn/thread/run/outbound/secret domain invariants into a generic SQL/filesystem crate.
- Do not rewrite every store in one PR.
- Do not weaken libSQL/Postgres parity.
- Do not expose secret material or secret rows through generic filesystem views.

## Proposed first slice

Add a small shared storage substrate, likely `crates/ironclaw_storage`, that owns reusable mechanics only:

- backend identity/support markers (`libsql`, `postgres`, `filesystem`, `object`, `memory`);
- redacted storage error type and safe error mapping helpers;
- migration descriptors/registry conventions;
- bounded pagination/cursor helpers;
- JSON encode/decode helpers for structured records;
- initial `BlobStore` / `RecordStore` trait sketches if small enough;
- design hooks for future `AppendLog`, `LockStore`, `TransactionalStore`, and encrypted raw record/blob storage.

Then migrate one low-risk domain adapter onto those helpers to prove shape without broad churn.

Suggested proof target: `ironclaw_outbound` or `ironclaw_threads`, because both already have clear domain-owned traits and DB-backed adapters.

## Acceptance criteria

- [ ] Storage-placement ADR/contract updated to distinguish:
  - domain ownership;
  - shared persistence substrate;
  - filesystem-like virtual views/projections;
  - secret-store encryption boundary.
- [ ] New shared storage substrate crate exists or an existing infra crate is extended for the same role.
- [ ] Substrate exposes mechanics only; architecture tests prevent it from depending on domain crates.
- [ ] At least one existing DB-backed domain store uses shared substrate helpers without moving domain semantics out of its crate.
- [ ] PostgreSQL and libSQL parity remains tested for the migrated store.
- [ ] Filesystem-like view rules are documented: projections/import-export/admin views, not default source of truth for structured state.
- [ ] Architecture/boundary tests prevent `ironclaw_filesystem` from becoming the owner of structured control-plane state.
- [ ] Secret-store boundary is documented as plaintext semantic API over ciphertext-only raw storage.
- [ ] No new production `.unwrap()` / `.expect()`.
- [ ] No raw secrets, host paths, backend error details, provider errors, or runtime payloads leak through storage errors or file views.

## Notes

This updates the existing `docs/reborn/contracts/storage-placement.md` rule:

> Structured, query-heavy, control-plane, or security-sensitive records -> typed repositories owned by the service domain.

with an additional clarification:

> Storage can have one shared substrate and filesystem-like projections, while domain crates keep typed operations and invariants.

serrrfirat:
## Summary
- add `ironclaw_storage` as a shared mechanics-only persistence substrate for Reborn adapters
- define redacted storage errors, backend identity, migration descriptors, JSON helpers, pagination, `StorageKey`/`StorageVersion`, and primitive `BlobStore`/`RecordStore` traits
- migrate `ironclaw_outbound` DB JSON/error/absent-scope helpers onto the shared substrate without moving outbound semantics out of the domain crate
- update storage-placement docs with the domain ownership vs substrate mechanics split, filesystem-like projection rules, and secret-store-over-ciphertext boundary
- add architecture boundary coverage preventing `ironclaw_storage` from depending on IronClaw domain crates

Closes #3419.

## Verification
- `CARGO_TARGET_DIR=/tmp/ironclaw-cargo-target-storage cargo test -p ironclaw_storage`
- `CARGO_TARGET_DIR=/tmp/ironclaw-cargo-target-storage cargo test -p ironclaw_outbound --all-features`
- `CARGO_TARGET_DIR=/tmp/ironclaw-cargo-target-storage cargo test -p ironclaw_architecture reborn_crate_dependency_boundaries_hold`
- `CARGO_TARGET_DIR=/tmp/ironclaw-cargo-target-storage cargo test -p ironclaw_architecture reborn_boundary_rules_active_crates_are_workspace_members`
- `CARGO_TARGET_DIR=/tmp/ironclaw-cargo-target-storage cargo clippy -p ironclaw_storage -p ironclaw_outbound --all-targets --all-features -- -D warnings`
- `cargo fmt --all --check`
- `git diff --check`

## Review
- Focused code-quality review: clean for Medium+.


ilblackdragon:
# Code Review: PR #3421 — Add Reborn shared storage substrate

## Where this fits in the Reborn integration

This PR carves out a **mechanics-only persistence substrate** (`ironclaw_storage`, ~460 LOC) shared by domain adapters. The split is:

```
ironclaw_storage (THIS PR)        ← mechanics: errors, JSON helpers, keys, versions, BlobStore/RecordStore
   ↑
ironclaw_outbound (migrated)      ← domain semantics
ironclaw_turns (future)
ironclaw_threads (future)
ironclaw_secrets (future)
```

The architectural rule the PR enforces (in CLAUDE.md and via the new boundary test): **`ironclaw_storage` may not depend on any other IronClaw crate.** That makes it a leaf substrate — domain crates may consume it, never the reverse. The boundary test forbids ~36 IronClaw crates as `ironclaw_storage` deps. Compile-time enforcement.

## Strengths

- **Mechanics vs ownership split is well-articulated.** The CLAUDE.md and AGENTS.md both spell out: backend identity, redacted errors, migration descriptors, JSON helpers, pagination, primitive `BlobStore`/`RecordStore` traits — but no domain operations. The contract doc update (`storage-placement.md`) adds a third failure mode to the existing two: "creating a single omniscient data-store crate that owns unrelated domain semantics." That's the explicit anti-goal.
- **Redaction is empirically tested.** `json_decode_error_is_redacted` and `backend_error_discards_raw_detail` assert that sentinel strings (`RAW_SECRET`, `/tmp/secret.db`) don't appear in error display/debug output. The right way to lock a "no leak" contract.
- **`StorageKey` rejects path-traversal patterns** at construction (`../`, `/x`, `\\`, `C:\`). Defense-in-depth in case a future backend mistakenly uses keys as filesystem paths. The `version..2` test confirms the validation is segment-aware (only rejects bare `..` segments, not `..` substrings).
- **`PutCondition` is a real CAS primitive.** `Any` / `IfAbsent` / `IfVersion(StorageVersion)` cover the standard set. The `allows()` test enumerates all four cases.
- **`outbound` migration is clean.** The local `to_json`/`from_json`/`db_error` helpers now delegate to the substrate. Centralizes the redaction.

## Issues

### Correctness

- **`db_error` in `outbound/db.rs:124-127` is dead code:**
  ```rust
  pub(crate) fn db_error(error: impl std::fmt::Display) -> OutboundError {
      let _ = ironclaw_storage::redacted_backend_error(error);
      OutboundError::Backend
  }
  ```
  `redacted_backend_error` discards its input and returns `StorageError::Backend`. The `let _ = ...` discards *that* result. Net effect: `error` is dropped, `OutboundError::Backend` is returned. The substrate call is a no-op. Either remove the dead call or — better — write to a tracing log first so debugging info exists *somewhere* before redaction. Right now the original error is unobservable.
- **`redacted_backend_error` deliberately discards** but the doc comment doesn't say "callers must log the error before redaction or the operational signal is lost." Add that contract — without it, every adapter migrating to the substrate silently drops debugging info. The pattern should be:
  ```rust
  // At the call site, before calling redacted_backend_error:
  tracing::debug!(error = %raw_error, "storage backend error");
  Err(redacted_backend_error(raw_error))
  ```
- **`RecordPayloadJson::new` does a full `serde_json::from_str::<Value>`** to validate. That's a complete deserialize roundtrip — fine for small records but expensive for large blobs (and allocates the full `Value` tree just to discard it). Consider `serde_json::from_str::<serde::de::IgnoredAny>` which skips materialization while still validating syntax.

### Style / typed-internals

- **`StorageError` is `Copy`** because it has no inner data. That's the right choice for a redacted error type — no fields to clone — but it means the entire variant set must stay payload-free. If a future variant needs structured detail (e.g., `Conflict { expected_version, actual_version }`), `Copy` would have to drop. Worth a comment: "// Copy is intentional: variants must remain payload-free to keep redaction trivial."
- **`StorageBackendKind::Filesystem`** overlaps with the explicit boundary the PR draws — the docs say "Filesystem remains file-shaped/path authority; this crate is not a universal filesystem replacement." But `Filesystem` is in the enum. Is the intent that a future filesystem-backed `BlobStore` will use it? If yes, document that `Filesystem` here means "blob store backed by a filesystem implementation" *not* "file-shaped path authority." The naming risks confusion.
- **`StorageVersion(String)`** — opaque token. No `#[serde(transparent)]` migration concern since it's new. Validation is bytes-bounded (`max_bytes = 128`), which matches the rule of byte-length being measured by `s.len()`. ✅

### Test coverage

- **No `BlobStore` / `RecordStore` implementation in this PR**, so the traits are interface-only. That's intentional ("future append-log/lock/transaction/encrypted-blob primitives") but means there's nothing exercising the trait surface yet. The `outbound` migration exercises only the helper functions. Add at least an in-memory `BlobStore`/`RecordStore` implementation (even if just for tests) so consumers see how to wire the traits.
- **No round-trip test for `StorageVersion` serde** — `StorageKey` has both serde and constructor tests; `StorageVersion` has only constructor tests. Add the symmetric serde tests.

### Risk

`risk: medium` is reasonable for a substrate touching multiple crates, even though the diff is L. The boundary test pins the dependency graph, and the migrated `outbound` already validates the migration path.

## Summary

Approve with these:
1. **Fix `outbound/db.rs::db_error`** — log the underlying error before passing to `redacted_backend_error`, or remove the no-op call entirely.
2. **Document the redaction contract** in `redacted_backend_error` ("callers must log before passing") so future migrations don't silently lose debug info.
3. **Add a `serde_json::from_str::<IgnoredAny>` optimization for `RecordPayloadJson::new`** if record sizes will ever be large.
4. **Add round-trip serde tests for `StorageVersion`** to match `StorageKey` coverage.
5. **Add at least one in-memory implementation** of `BlobStore`/`RecordStore` so consumers have a wiring reference.

The architecture is right and the redaction discipline is empirically tested. Mostly polish before merge.

serrrfirat:
Follow-up fixes from the review are now addressed in the current PR head:

- `outbound::db_error` logs raw backend errors with `tracing::debug!` before crossing the redaction boundary, then returns redacted `OutboundError::Backend`.
- `redacted_backend_error` docs now make the caller-side logging/redaction contract explicit.
- `RecordPayloadJson::new` now validates via `serde_json::from_str::<IgnoredAny>` instead of materializing a full `Value` tree.
- `StorageVersion` now has transparent serde round-trip coverage plus invalid-deserialize coverage.
- Test-only `InMemoryPrimitiveStore` now exercises `BlobStore` / `RecordStore` wiring, including put/get/delete and CAS conflict behavior.

Also covered related polish: explicit `StorageError` payload-free/`Copy` invariant comment and clearer `StorageBackendKind::Filesystem` wording.

Issue #3286, opened 2026-05-06
[Reborn] Preserve agent command behavior through Reborn loops and services


serrrfirat:
Parent: #3031  
Related: #3092, #3091, #3013, #3236, #3238, #3094, #3107, #3090, #3278, #3290, #3284, #3020, #3032

## What to build

Preserve current in-chat agent/slash command behavior by routing commands and loop actions through Reborn turns, loop drivers, interaction services, and typed product services rather than v1 `Agent` authority.

This issue covers **agent/chat commands** typed into a conversation (`/undo`, `/interrupt`, `/model`, `/skills`, etc.). Broader local operator CLI/TUI/setup commands are tracked by #3284.

Command/behavior examples:

- simple reply, tool-call loop, tool error recovery, multi-turn state;
- `/undo`, `/redo`, `/compact`, `/clear`, `/heartbeat`, `/summarize` / `/summary`, `/suggest`, `/expected`;
- `/thread`, `/resume`, `/status`, `/cancel`, `/interrupt`, `/model`, `/skills`, `/tools`;
- future explicit steering commands such as `/btw` / `/steer` coordinated with #3236;
- file read/write/list/patch workflows through capabilities;
- shell/sandbox behavior through capabilities/processes where supported;
- context compaction, skill context injection, group chat privacy behavior.

## Resolved design decisions

- **First slice creates a shared typed `AgentCommandService` and command registry.** Do not migrate commands case-by-case inside ProductWorkflow or loop drivers. The registry owns command metadata, aliases, arguments, categories, availability, owner-service mapping, scope requirements, help text, and Reborn support status.
- **Command classification happens after binding/session resolution.** ProductWorkflow first authenticates the actor, resolves canonical tenant/thread/binding/session context, and checks thread access. Then it classifies command input into typed `AgentCommandIntent` categories such as run control, thread control, loop-context action, profile preference, interaction resolution, product action, normal message, and explicit steering.
- **Control/gate commands bypass the normal follow-up queue through typed services.** Normal same-thread messages are governed by #3236. Approval/auth/resource resolution, cancellation, status/admin commands, and explicit steering are classified and routed separately after canonical access checks.
- **Persist command events, not all commands as transcript messages.** Every accepted command should create a durable command/action event and projection-safe outcome. Only commands that intentionally produce conversation content should write transcript messages; control commands should not become model-visible user messages by default.
[...]

Issue #3281, opened 2026-05-06
[Reborn] Add EventStreamManager for durable projection fanout


serrrfirat:
Parent: #3031  
Related: #3093, #3266, #3279, #3032, #3280, #3282, #3283, #3285, #3162

## Purpose

Add the product-facing Reborn `EventStreamManager`: the service that turns scoped projection snapshots/updates into durable, replayable, product-safe streams for Web SSE, WebSocket, API subscribers, and product/channel fanout candidates.

Target shape:

```text
EventProjectionService (#3093)
  -> EventStreamManager
      -> explicit actor + projection scope access checks
      -> durable ProjectionCursor validation
      -> initial snapshot + ordered updates
      -> rebase/lag/keepalive stream items
      -> external push candidate selection via notification policy (#3266)
      -> product adapter transport rendering (#3269/#3282/#3285)
```

This is **not** a raw durable event-log streaming API and not a Web-specific SSE manager. Product surfaces consume projection-safe stream items; adapters render them into protocol frames.

## Current v1 evidence

Current v1 Web streaming is useful migration evidence but not the Reborn contract:

- `SseManager` broadcasts live `AppEvent`s through a Tokio broadcast channel.
- Scoping is mostly by `user_id`; deeper thread/project/participant policy is handled inconsistently at producers/consumers.
- SSE ids are boot-local `boot_id:seq` values.
- `last_event_id` only filters events from the current process boot and cannot reconstruct missed durable state.
- WebSocket subscribes to the same raw `AppEvent` stream and also accepts inbound messages/approvals.
- Debug/verbose subscribers can receive extra event variants such as full tool output/metrics.

Reborn requires durable scoped projection cursors, explicit access validation, snapshot/rebase behavior, product-safe projection DTOs, and separation between subscription visibility and external push delivery.

## Decisions

1. **`ProjectionCursor` is the only resume authority.**
   - Cursors come from EventProjectionService / projection store semantics.
   - SSE `id`, WebSocket sequence numbers, or adapter-local event ids are transport wrappers only.
   - A transport may encode the projection cursor for client convenience, but replay validation must use a scoped `ProjectionCursor` plus actor/scope access checks.
   - Transport-local ids must never be accepted as canonical Reborn cursor authority.

2. **EventStreamManager is projection-type agnostic.**
   - EventProjectionService owns reducers, snapshots, updates, cursor production, and product-facing projection DTOs.
[...]

PR #3428, opened 2026-05-09
feat(reborn): add ProductWorkflow and InboundTurnService facade (#3280)

### LINKED ISSUE #3280: [Reborn] Add ProductWorkflow and InboundTurnService facade
Parent: #3031  
Related: #3269, #3193, #3089, #3013, #3198, #3278, #3279, #3264, #3266, #3281, #3286, #3094

## Purpose

Add the product-facing Reborn workflow facade that sits between ProductAdapters and host-layer Reborn services.

Target shape:

```text
ProductAdapter
  -> ProductWorkflow
      -> InboundTurnService          # user-message subset
          -> ConversationBindingService / SessionThreadService
          -> TurnCoordinator
      -> ProductCommandRouter        # command seam; full matrix in #3286
      -> Approval/Auth interaction services
      -> MissionService
      -> EventStreamManager          # read/subscription dispatch only; #3281 owns streams
      -> product-safe acknowledgement / outcome
```

This issue exists so Web/API/CLI/channel adapters do not each reimplement binding resolution, message staging, idempotency, busy/deferred handling, gate routing, mission routing, and redacted acknowledgements.

## Current v1 evidence

Current v1 inbound flow mixes too many responsibilities in surface/runtime code:

- `IncomingMessage` carries external channel refs, already-resolved `user_id`, raw content, attachments, metadata, external thread ids, and internal mission flags.
- Web `/api/chat/send` builds an `IncomingMessage` and pushes it to an mpsc channel.
- `ChannelManager` merges live streams and also owns `respond`, `send_status`, and broadcast egress.
- `Agent::handle_message` owns BeforeInbound hooks, submission parsing, engine-v2 routing, session/thread resolution, auth/gate handling, mission trigger firing, and loop dispatch.
- The engine-v2 bridge still does conversation creation, attachment persistence, mission OnEvent firing, thread execution, status push, and SSE broadcast in one path.

Reborn should preserve product behavior where needed, but ProductWorkflow must not preserve this ownership shape.

## Decisions

1. **`ProductWorkflow` is the top-level product action orchestrator.**
   - `InboundTurnService` is the narrower user-message path.
   - ProductWorkflow dispatches action kinds such as user message, command, gate/auth resolution, mission action, typed system action, and read/subscription request.
   - Do not make adapters call `InboundTurnService` directly for every action type.

2. **Use a separate host-layer crate/module.**
   - Prefer `crates/ironclaw_product_workflow` for workflow contracts, saga/idempotency ports, product-safe outcomes, and fake harnesses.
   - Keep `crates/ironclaw_product_adapters` focused on adapter DTOs/capabilities/lifecycle from #3269.
   - Do not put ProductWorkflow inside `ironclaw_turns`; it routes missions, gates, commands, and subscriptions too.

3. **ProductWorkflow owns a durable mutating-action idempotency saga.**
   - Add a `ProductInboundAction` / equivalent durable ledger for mutating actions.
   - Key by tenant + adapter installation + external event/client request fingerprint before downstream side effects.
   - Attach binding/message/run/mission/gate/outcome refs as the pipeline advances.
   - Suggested states: `received`, `policy_checked`, `message_accepted`, `submit_pending`, `submitted`, `deferred_busy`, `routed_command`, `routed_gate`, `routed_mission`, `rejected`, `failed_retryable`.
   - Do not require one cross-service transaction across binding, transcript, and turn stores; recover through saga state.

4. **Crash recovery reuses committed refs.**
   - If message acceptance succeeded but submit did not complete, retry submit using the same accepted message ref and turn idempotency key.
   - If a run/mission/gate side effect exists, replay the same product-safe outcome.
   - Pre-side-effect transient dependency failures remain retryable and should not permanently poison the external event id.

5. **Inbound policies and safety run before accepted-message staging.**
   - Run shared BeforeInbound-style hooks, validation, secret scanning, protocol-independent body policy, and redaction checks before `SessionThreadService.accept_inbound_message`.
   - Rejected input records only a redacted ProductInboundAction rejection/audit; it does not become accepted canonical transcript content.
   - Rewritten input is what gets staged as the accepted message.

6. **Canonical identity/scope comes from binding/session services.**
   - ProductAdapters authenticate the transport/protocol and provide external actor/conversation/event refs plus authenticated installation context.
   - ProductWorkflow must treat adapter-provided internal user/thread/scope values as non-authoritative.
   - Canonical tenant, actor, owner/resource scope, thread, source binding, and reply target come from ConversationBindingService / SessionThreadService.
   - Trusted internal surfaces should still be modeled as explicit ProductAdapter installations/bindings, not bypasses.

7. **Normal user messages stage content before turn submission.**
   - `InboundTurnService` runs binding resolution, accepted-message staging, then `TurnCoordinator.submit_turn`.
   - `TurnCoordinator::ThreadBusy` becomes product-safe `DeferredBusy` at the workflow boundary.
   - The accepted message is marked deferred/busy rather than dropped or rejected.
   - Follow-up promotion/steering policy remains #3236; #3280 only preserves the message and returns the safe acknowledgement.

8. **Content and attachments are ingested through storage services before turns.**
   - Adapters provide normalized text/attachment descriptors or pre-staged blob refs.
   - ProductWorkflow orchestrates ingestion through SessionThreadService/blob/content services.
   - TurnCoordinator-facing requests contain only accepted message/context refs, never raw content, inline bytes, source URLs, or host-local paths.

9. **ProductWorkflow returns pipeline acknowledgements, not final answers.**
   - Acks mean the event was authenticated by the adapter, policy-checked, deduped, accepted/deferred/routed/rejected, and any side effects were referenced safely.
   - Final replies/status/progress arrive through projections/egress.
   - Synchronous APIs should use a separate projection waiter after ProductWorkflow returns run/cursor refs; they must not bypass ProductWorkflow.

10. **Central command router seam.**
    - ProductWorkflow owns a `ProductCommandRouter` seam for textual slash commands and structured command payloads.
    - Adapters may classify explicit protocol UI actions as typed commands, but shared slash-command parsing/routing should not be duplicated per adapter.
    - #3280 defines the seam and first fake tests; #3286 owns the full command compatibility matrix.

11. **Gate/auth resolutions go only through interaction services.**
    - ProductWorkflow may dispatch gate/auth resolution payloads.
    - `ApprovalInteractionService` and `AuthInteractionService` own actor authority, gate refs, fingerprints, stale/expired buttons, and the eventual resume/cancel decision.
    - ProductWorkflow and adapters must not call `TurnCoordinator.resume_turn` directly for normal gate UX.

12. **Mission routing requires explicit mission intent.**
    - ProductWorkflow routes to MissionService only for explicit `MissionAction`, mission command, or explicit mission-thread link intent.
    - Do not infer mission attachment from ordinary chat text.
    - v1-style OnEvent mission firing should be modeled separately as MissionTriggerSource/MissionFire work under #3278/#3290, not as hidden normal chat routing.

13. **Product-safe outcome taxonomy is owned here.**
    - Define stable adapter-facing outcomes/rejections such as:
      - `Accepted`
      - `DeferredBusy`
      - `Duplicate`
      - `BindingRequired`
      - `AccessDeniedSafe` / safe not-found
      - `PolicyRejected`
      - `AttachmentRejected`
      - `AdmissionRejectedSafe`
      - `CommandRouted`
      - `GateHandled`
      - `MissionSubmitted` / `MissionSuppressed`
      - `RetryableUnavailable`
      - `InternalRedacted`
    - Adapters should not map raw downstream service errors themselves.
    - Adapter-facing payloads must not expose raw prompts, tool input, secrets, host paths, backend diagnostics, foreign existence, raw provider errors, or raw runtime output.

14. **Ledger is for mutating actions; streams/read subscriptions belong to EventStreamManager.**
    - ProductWorkflow may dispatch subscription/read requests, but durable ProductInboundAction saga rows are for mutating actions that can create messages, runs, gates, missions, delivery attempts, or other side effects.
    - Projection fetch/subscribe/cursor validation is owned by #3281 / EventStreamManager and should not create high-volume inbound-action rows.

15. **No generic internal-message bypass.**
    - Do not preserve a broad `is_internal` escape hatch in ProductWorkflow.
    - System/job/mission-originated work must use explicit typed system actions with accountable actor/scope, or emit projection/egress events.
    - This prevents hidden message loops and unaudited bypass of normal product policy.

16. **Architecture guardrails.**
    - ProductWorkflow must not call raw `CapabilityHost`, `RuntimeDispatcher`, runtime lanes, raw process managers, raw network clients, raw secret stores/material, raw filesystem APIs/mounts, or `ironclaw_turns::runner` trusted transition APIs.
    - It composes product/host services through typed ports only.

## Suggested contract sketch

Names are illustrative; semantics matter more than exact names.

```rust
pub trait ProductWorkflow {
    async fn handle_action(
        &self,
        request: ProductWorkflowRequest,
    ) -> Result<ProductWorkflowAck, ProductWorkflowError>;
}

pub struct ProductWorkflowRequest {
    pub adapter: ProductAdapterDescriptorRef,
    pub external_event_id: ExternalEventId,
    pub external_actor_ref: ExternalActorRef,
    pub external_conversation_ref: ExternalConversationRef,
    pub received_at: Timestamp,
    pub payload: ProductActionPayload,
}

pub enum ProductActionPayload {
    UserMessage(UserMessagePayload),
    Command(CommandPayload),
    GateResolution(GateResolutionPayload),
    AuthResolution(AuthResolutionPayload),
    MissionAction(MissionActionPayload),
    SystemAction(SystemActionPayload),
    ProjectionRead(ProjectionReadRequest),
    ProjectionSubscribe(ProjectionSubscriptionRequest),
}

pub enum ProductWorkflowAck {
    AcceptedTurn {
        inbound_action_ref: ProductInboundActionRef,
        message_ref: AcceptedMessageRef,
        run_id: TurnRunId,
        cursor: Option<ProjectionCursor>,
    },
    DeferredBusy {
        inbound_action_ref: ProductInboundActionRef,
        message_ref: AcceptedMessageRef,
        active_run_id: TurnRunId,
        cursor: Option<ProjectionCursor>,
    },
    Duplicate {
        inbound_action_ref: ProductInboundActionRef,
        prior: Box<ProductWorkflowAck>,
    },
    CommandRouted { inbound_action_ref: ProductInboundActionRef },
    GateHandled { inbound_action_ref: ProductInboundActionRef },
    MissionHandled { inbound_action_ref: ProductInboundActionRef },
    Rejected { inbound_action_ref: ProductInboundActionRef, reason: ProductWorkflowRejection },
    RetryableUnavailable { reason: ProductWorkflowUnavailableKind },
}
```

## First implementation slice

Add contract/types/ports and fake stores/services/tests only.

Suggested layout:

```text
crates/ironclaw_product_workflow/
  src/lib.rs
  tests/product_workflow_contract.rs
```

Use fake ProductAdapters, fake binding/session-thread services, fake TurnCoordinator, fake MissionService, fake interaction services, fake EventStreamManager access checks, and a restartable fake ProductInboundAction store.

Do **not** migrate Web gateway routes, CLI/TUI, OpenAI-compatible APIs, external channels, v1 `ChannelManager`, v1 `Agent`, v1 mission/routine/job managers, or production DB persistence in the first PR.

## Acceptance tests

- [ ] ProductWorkflow dispatches `UserMessage` to InboundTurnService; other action kinds do not go through the user-message path.
- [ ] ProductAdapters provide external refs and authenticated installation context; canonical actor/scope/thread/source/reply refs are resolved only through fake binding/session services.
- [ ] A new mutating external event creates a ProductInboundAction ledger row before message staging or turn submission.
- [ ] Duplicate external event/client request id replays the same accepted/deferred/rejected outcome.
- [ ] Duplicate external event creates neither a duplicate accepted message nor a duplicate turn/run.
- [ ] If service restart occurs after `message_accepted` but before submit completion, recovery retries submit with the same accepted message ref/idempotency key.
- [ ] Pre-side-effect transient store/service failure remains retryable and is not cached as a terminal rejection.
- [ ] BeforeInbound-style policy rewrite stores the rewritten content as the accepted message.
- [ ] BeforeInbound/safety rejection does not create accepted transcript content or a turn.
- [ ] Adapter-supplied internal user/thread/scope values are ignored or rejected as authority.
- [ ] Accepted user message stages content/attachments through fake SessionThread/blob services before turn submit.
- [ ] TurnCoordinator submit request contains accepted message/context refs only; no raw content, inline bytes, source URLs, or host-local paths.
- [ ] `ThreadBusy` from TurnCoordinator maps to durable `DeferredBusy`, preserves accepted message ref, and does not create a second active run.
- [ ] ProductWorkflow returns only pipeline acknowledgement; synchronous final-answer waiting is not part of the core workflow test harness.
- [ ] Textual slash command goes through ProductCommandRouter seam and is not parsed independently by fake adapters.
- [ ] Gate/auth resolution payloads are delegated to fake Approval/Auth interaction services and never call TurnCoordinator resume directly.
- [ ] Mission action routes to fake MissionService only when explicit mission intent/link command is present.
- [ ] Ordinary chat text never auto-attaches to an existing mission.
- [ ] Product-safe rejection taxonomy redacts raw prompts, tool input, secrets, host paths, raw backend errors, raw provider/runtime output, and foreign existence.
- [ ] Projection subscribe/read dispatch uses fake EventStreamManager access checks and does not create ProductInboundAction saga rows.
- [ ] Typed system actions require accountable actor/scope and do not use a generic internal-message bypass flag.
- [ ] Architecture boundary tests prove ProductWorkflow does not depend on raw CapabilityHost, RuntimeDispatcher, runtime lanes, raw process/network/secret/filesystem APIs, v1 Agent, v1 ChannelManager, or trusted runner transition APIs.

## Non-goals

- Production Web/API/CLI/channel migration; tracked by #3282, #3283, #3284, and #3285.
- Full command compatibility matrix; tracked by #3286.
- EventStreamManager implementation; tracked by #3281.
- ProductAdapter DTO/lifecycle contract; tracked by #3269.
- MissionService internals; tracked by #3278.
- Same-thread follow-up promotion/steering; tracked by #3236.
- PostgreSQL/libSQL production persistence for ProductInboundAction; first slice uses restartable fake semantics.
- v1 `ChannelManager` bridge or default-on Reborn exposure.

## Blocked by / coordinates with

- Blocked by #3269 for ProductAdapter envelope/action DTOs.
- Blocked by #3193 / #3089 for binding/session-thread inputs.
- Blocked by #3013 / #3198 for TurnCoordinator submit/response shapes.
- Coordinates with #3278 for mission action routing.
- Coordinates with #3094 for gate/auth interaction services.
- Coordinates with #3266 / #3281 for projection/read/stream/outbound behavior.
- Coordinates with #3286 for command compatibility.

serrrfirat:
## Summary

Adds the `ironclaw_product_workflow` crate — the product-facing Reborn workflow facade that sits between product adapters and host-layer Reborn services (issue #3280).

### Core types
- **`DefaultProductWorkflow`** — top-level orchestrator implementing the `ProductWorkflow` trait from `ironclaw_product_adapters`. Dispatches inbound envelopes by payload kind (user message, command, approval, auth, subscription, linked thread action, noop).
- **`InboundTurnService` / `DefaultInboundTurnService`** — narrower user-message path that composes `ConversationBindingService` + `TurnCoordinator` for binding resolution and turn submission.

### Idempotency saga
- **`ProductInboundAction`** — durable ledger record with phase tracking (Received → Dispatched → Settled / DeduplicatedReplay).
- **`ActionFingerprintKey`** — composite dedup key: adapter_id + installation_id + source_binding_key + external_event_id.
- **`IdempotencyLedger`** port — durable deduplication contract with begin_or_replay / settle operations.

### Binding resolution
- **`ConversationBindingService`** — maps external adapter refs (actor, conversation) to canonical Reborn identifiers (tenant, user, thread).
- **`ResolvedBinding`** — the resolved canonical binding record.

### Error handling
- **`ProductWorkflowError`** — internal error vocabulary with clean mapping to `ProductAdapterError` (transient → retryable, binding/turn failures → internal).

### Test support (feature-gated)
- `FakeConversationBindingService` — programmable binding resolver
- `FakeIdempotencyLedger` — in-memory dedup ledger
- `FakeInboundTurnService` — programmable turn service

### Tests
- 8 contract tests covering user message dispatch, noop handling, duplicate dedup replay, transient error surfacing, binding resolution, and failure propagation.

### Architecture
- Boundary rule added: `ironclaw_product_workflow` must not depend on dispatcher, extensions, host_runtime, mcp, wasm, scripts, network, engine, or gateway crates.

## Seams left for future issues
- Command routing (#3286) — returns `CommandRoutingUnavailable`
- Approval/auth resolution — returns `UnsupportedActionKind`
- Projection subscriptions (#3281) — returns `UnsupportedActionKind`
- Mission routing, linked thread actions — future work

Closes #3280


ilblackdragon:
# Code Review: PR #3428 — feat(reborn): add ProductWorkflow and InboundTurnService facade

## Where this fits in the Reborn integration

This PR adds the **product-adapter-facing facade** above `TurnCoordinator`. Recall the stack:

```
Channel adapter (Slack/Web/CLI/Telegram)
  → ProductInboundEnvelope
  → DefaultProductWorkflow.accept_inbound()        ← THIS PR
  → IdempotencyLedger.begin_or_replay (dedup)
  → dispatch by payload kind:
      UserMessage → InboundTurnService.accept_user_message()
      Command/Approval/Auth/Subscription → unsupported (placeholder rejections)
      NoOp → ack
  → IdempotencyLedger.settle()
  → ProductInboundAck
TurnCoordinator (ironclaw_turns)
  → TurnStateStore Queued
  → SchedulerTurnRunWakeNotifier (#3438)
TurnRunScheduler → executor → host (#3439) → driver → LoopExit (#3446)
```

This is the entrance of the funnel — every adapter (web webhook, Slack, Telegram, CLI, …) flows through `DefaultProductWorkflow::accept_inbound`. So bugs here have the widest blast radius of any PR in this batch.

## Strengths

- **Idempotency saga is a real design.** `ProductInboundAction` tracks Received → Dispatched → Settled / DeduplicatedReplay; `ActionFingerprintKey` is the durable dedup key. The pattern is correct: begin reserves the fingerprint, dispatch does the work, settle records the terminal outcome. Replays return the prior `outcome` instead of re-running the side effect.
- **Error mapping table is explicit.** `ProductWorkflowError → ProductAdapterError` distinguishes retryable (`Transient`) from terminal (`Internal`/`PolicyDenied`) failures. The `is_retryable` test locks the contract.
- **`test-support` feature flag** keeps the in-memory fakes off production binaries — `test-support = ["ironclaw_product_adapters/test-support"]` chains correctly.
- **Boundary rule wired.** Architecture test forbids deps on dispatcher/extensions/host_runtime/mcp/wasm/scripts/network/engine/gateway. Compile-time enforcement.

## Issues

### Real bug — double prefix in reply target

`inbound_turn.rs:984-1015`:

```rust
let reply_target_binding_id = format!("reply:{source_binding_id}");
// ...
let reply_target_binding_ref =
    bounded_ref::<ReplyTargetBindingRef>("reply", &reply_target_binding_id)?;
```

And `bounded_ref` does `format!("{prefix}:{raw}")`. With `prefix = "reply"`, `raw = "reply:foo"`, the result is `"reply:reply:foo"`. The `reply:` prefix is applied twice. Either drop the `format!("reply:{...}")` line and pass the raw `source_binding_id` to `bounded_ref`, or pass an empty prefix to `bounded_ref`. Add a unit test that asserts `reply_target_binding_ref.as_str().starts_with("reply:")` AND doesn't have a doubled prefix.

### Typed-internals violations (`.claude/rules/types.md`)

- **`ActionFingerprintKey` (`action.rs:229-236`)** — all four fields are `String` despite inputs being typed:
  ```rust
  pub struct ActionFingerprintKey {
      pub adapter_id: String,
      pub installation_id: String,
      pub source_binding_key: String,
      pub external_event_id: String,
  }
  ```
  Constructor takes `&ProductAdapterId`, `&AdapterInstallationId`, `&ExternalEventId` and immediately strings them via `.as_str().to_string()`. Direct violation — the rule is explicit ("never `String`, `&str`, or `uuid::Uuid` alone" for identifiers; bug #2574 fix referenced). The fingerprint should carry the typed values. Hash/equality work either way; storing typed values means a future caller cannot accidentally pass a `user_id` as `installation_id`.

- **`ActionDispatchKind` (`action.rs:270-280`)** — same issue:
  ```rust
  Command { command: String },
  ApprovalResolution { gate_ref: String },          // LoopGateRef exists
  AuthResolution { auth_request_ref: String },      // should be typed
  LinkedThreadAction { action_id: String },
  ```
  `LoopGateRef` is already a typed newtype in `ironclaw_turns`; using `String` here loses validation and creates the drift the rule prohibits.

### Correctness — saga durability

- **The action remains in `Received` phase if dispatch returns a transient error.** `accept_inbound` only calls `settle` when `terminal_ack_for_error` returns `Some` (i.e., `CommandRoutingUnavailable` / `UnsupportedActionKind`). For `Transient`, `BindingResolutionFailed`, `TurnSubmissionRejected`, `TurnResumeRejected`, no settlement happens — the fingerprint stays reserved. On retry, `begin_or_replay` would (per the trait doc) "return a transient error so the caller can retry or wait." But the `FakeIdempotencyLedger` doesn't appear to model the in-flight-not-settled state explicitly. If the workflow process crashes between begin and settle, the action is permanently stuck. The contract docs need a recovery story (TTL on `Received` phase? Background sweeper?) before this is production-safe. Not a blocker for the facade itself, but a blocker before any real adapter routes through it.

- **`thread_scope_from_binding` requires `agent_id`** — same constraint as #3439's `validate_thread_scope`. User-only binding (no agent) returns `BindingResolutionFailed`. Document or relax. Cross-PR consistency.

- **`bounded_ref` silently hashes overflowing inputs to UUIDv5** (`inbound_turn.rs:1127-1134`). Inputs > 240 chars or containing control chars become `format!("{prefix}:{uuid_v5}")`. The fallback is invisible to the caller, and two distinct external IDs that differ only in trailing whitespace could collide if neither is sanitized first. Worth: (1) a unit test that asserts deterministic hashing, (2) consider returning an error on the overflow path so the caller can decide (lossy hashing for an idempotency-bearing field is risky).

### Style / minor

- **`source_binding_id: envelope.source_binding_key()`** returns a `String` (the variable name suggests `&str` would be cheaper). If `source_binding_key()` returns `&str`, the `.clone()` is implicit — verify upstream type.
- **`DefaultProductWorkflow::resolve_projection_subscription`** always errors. Will land in #3281 per PR description. OK.

### Risk

`risk: medium` is accurate. The double-prefix bug is a real defect; the typed-fingerprint violations are blocking per the rule; the saga-not-durable concern is architectural. Approve once those are addressed.

## Summary

Three blocking items:
1. **Fix the double `reply:` prefix bug in `inbound_turn.rs:984/1015`** + add a regression test.
2. **Fix `ActionFingerprintKey` and `ActionDispatchKind` to use typed identifiers** per `.claude/rules/types.md` — the rule is non-negotiable, and this PR is the entry point for every adapter.
3. **Document the recovery path for stuck `Received` actions** (or implement a TTL/sweeper). Without it, a crash between begin and settle leaks fingerprints.

The architecture is right; the saga shape is right; the error vocabulary is right. Cleanup before merge.

serrrfirat:
Addressed the review feedback in `fcfaa932`:

- Fixed the double `reply:` prefix by passing the raw source binding key into `bounded_ref("reply", ...)`.
  - Added `reply_target_binding_ref_has_single_reply_prefix` regression coverage.
- Replaced stringly workflow internals with typed identifiers:
  - `ActionFingerprintKey` now stores `ProductAdapterId`, `AdapterInstallationId`, `SourceBindingKey`, and `ExternalEventId`.
  - `ActionDispatchKind` now stores typed refs: `ProductCommandName`, `LoopGateRef`, `AuthRequestRef`, and `LinkedThreadActionId`.
  - Added contract tests for typed fingerprint/dispatch conversion.
- Added the recovery contract for stuck non-terminal idempotency actions.
  - Fake ledger now models in-flight reservations, blocks duplicate dispatch while fresh, and supports expiry/reclaim for recovery tests.
  - Added tests for transient reservation and expiration/reclaim behavior.

Verification:
- `cargo fmt --check`
- `cargo test -p ironclaw_product_workflow --features test-support --quiet`
- `cargo clippy -p ironclaw_product_workflow --all-targets --features test-support --quiet`
- GitHub checks are green.

henrypark133 (changes requested):
Thanks for putting the product workflow facade together. The shape is useful, but I found several inbound-turn/idempotency issues that need to be fixed before this is safe to wire to product adapters.

Findings:

1. High - `crates/ironclaw_product_workflow/src/workflow.rs:84`

`DeferredBusy` is settled into the outer `IdempotencyLedger`, so a duplicate delivery replays the prior busy ack instead of re-entering `InboundTurnService` after the active run clears. That makes same-thread busy durable even though it is a transient condition.

Failure mode: a webhook/client retry can permanently strand an accepted user message as busy. The retry sees the settled product action and returns `Duplicate { prior: DeferredBusy(...) }` instead of trying `TurnCoordinator.submit_turn` again.

Expected fix direction: do not settle `DeferredBusy` as a terminal product action outcome. Preserve/replay the accepted message ref and let retry drive turn submission again with a fresh submit idempotency key once the active run clears.

2. High - `crates/ironclaw_product_workflow/src/inbound_turn.rs:112`

Retry recovery re-resolves the live binding before checking for an already accepted message. Because thread/message idempotency is scoped by the resolved thread scope and source binding, a transient submit failure followed by binding churn can move the retry to a different user/thread and accept a second message instead of reusing the original accepted message, actor, and received timestamp.

Failure mode: duplicate external events can cross users/threads or duplicate transcript state during exactly the failure path idempotency is supposed to make safe.

Expected fix direction: reuse the existing `ironclaw_conversations::InboundTurnService` replay-before-binding semantics, or add an equivalent accepted-message replay record keyed before live binding resolution.

3. Medium - `crates/ironclaw_product_workflow/src/inbound_turn.rs:239`

Non-unavailable `TurnError`s are flattened to strings, and `ProductWorkflowError::TurnSubmissionRejected` stores only `reason: String`. That loses typed category/status/retry semantics at the product workflow boundary.

Failure mode: admission denied, unauthorized, invalid request, and conflict outcomes all become generic internal workflow failures, so adapters cannot reliably map 403/409/429/400 or durable retry behavior.

Expected fix direction: carry typed `TurnError` through this boundary, or map it losslessly into stable typed adapter ack/error categories.

4. Medium - `docs/reborn/2026-04-25-current-architecture-map.md:260`

The architecture map still says `ironclaw_conversations` owns `ConversationBindingService`, `SessionThreadService`, and `InboundTurnService`, but this PR adds same-named contracts in `ironclaw_product_workflow` with different retry/idempotency semantics.

Failure mode: the source-of-truth docs and implementation now describe two competing inbound-turn facades, making future adapter wiring likely to pick the weaker path.

Expected fix direction: either make product workflow compose the existing conversation service directly, or update the contract/map with an explicit ownership split plus parity tests for the existing conversation-binding requirements.

What looked good:

- The new crate has a clear facade shape and the architecture dependency guardrail passed in the targeted review.
- The local crate tests cover the currently implemented behavior, but they mostly pin the weaker semantics above rather than proving parity with the existing conversation/inbound-turn contract.

Suggested coverage before re-review:

- Caller-level product workflow test where first call returns `DeferredBusy`, then duplicate retry after active run clears resubmits the same accepted message ref.
- Caller-level test for transient submit failure plus binding churn: retry must reuse the original accepted message, actor, received_at, and run-profile request.
- Tests asserting typed `TurnError` categories map to stable adapter/product outcomes.

serrrfirat:
Addressed review feedback in f461d3ab:

- DeferredBusy is no longer settled as a terminal product idempotency outcome; retryable/non-terminal outcomes release the outer reservation so duplicates can retry.
- Inbound turn retries now replay accepted messages before live binding resolution, preserving the original accepted message/thread/user instead of following binding churn.
- Deferred-busy retry resubmits the existing deferred message and can transition it to submitted when the active run clears.
- Turn submission failures now preserve typed workflow rejection category/status/retryability at the adapter boundary.
- Documented the product-workflow vs conversations ownership split and the agent-scoped binding requirement.
- Added regression coverage for busy retry, binding-churn retry, typed turn error mapping, and deterministic bounded-ref hashing.

Verification run:
- cargo test -p ironclaw_product_workflow --features test-support --quiet
- cargo clippy -p ironclaw_product_workflow --all-targets --features test-support
- cargo test -p ironclaw_threads -p ironclaw_loop_support --quiet

Issue #3067, opened 2026-04-29
[TEST] Reborn: Add vertical-slice integration test suite


serrrfirat:
## Goal

Create caller-level Reborn integration tests that prove the substrate works through public entrypoints, not only through crate-local unit/contract tests.

This issue tracks the integration-test route for the Reborn stack landing into `reborn-integration`.

## Related

- Parent tracker: #2987
- Event/replay integration evidence: #3022
- No-exposure safeguards: #3032
- Production composition root: #3026
- Final integration PR checklist: #3039
- Compatibility gate: #3020
- Product-surface migration: #3031
- WASM WIT/component-model runtime lane: #3097 (supersedes #3028)

## Principle

Test through public Reborn seams:

- `RuntimeDispatcher`
- `ProcessHost`
- `CapabilityHost`
- `ApprovalResolver`
- `HostRuntimeServices`
- production composition factory once available

Do **not** rely only on private helpers or isolated unit tests. Existing unit/contract tests remain valuable; this issue adds the cross-crate caller-level evidence needed before final `reborn-integration -> staging`.

## Timing / route

```text
Now
  -> create this tracker and link it from #2987

After #3017 + #3023
  -> start process/dispatcher integration tests

After #3027 + #3097
  -> add real runtime-lane integration tests

When CapabilityHost + host_runtime PR opens
  -> require full vertical-slice integration tests before merge

After PR5 / PR6
  -> add secrets/network/obligation integration tests

After PR7 / PR8
  -> add memory storage/search integration tests

Before final reborn-integration -> staging
  -> durable restart, no-exposure, compatibility/replay, production composition smoke
```

---

## Phase 0 — Tracking/spec now

- [x] Confirm this issue as the Reborn vertical-slice integration suite tracker.
- [x] Link this issue from #2987.
- [x] Link this issue from #3039 final integration checklist when the final PR is prepared.
- [x] Assign owners for the lanes below.

---

## Phase 1 — After #3017 + #3023: dispatcher/process integration

Status:

- [x] Started in PR #3076 — https://github.com/nearai/ironclaw/pull/3076
- [x] Local verification passed on PR branch:

```bash
CARGO_TARGET_DIR=/tmp/ironclaw-reborn-integration-target cargo test -p ironclaw_dispatcher -p ironclaw_processes
[...]

PR #3444, opened 2026-05-09
fix(reborn): strengthen host runtime publication gates

serrrfirat:
## Summary

- Adds two #3067 Reborn E2E gate tests through `HostRuntimeServices -> DefaultHostRuntime -> ScriptRuntime`.
- Covers `RedactOutput` before public completed output is returned and verifies bearer/JWT fragments do not leak.
- Covers `EnforceOutputLimit` blocking oversized runtime output before public publication, with failed run-state and no sentinel leak in runtime events/failure output.
- Fixes bearer token redaction so dotted JWT-style bearer tokens redact as one token instead of leaving tail fragments.
- Keeps existing redacted-key collision coverage aligned with full-token redaction.
- Avoids a cfg-dependent `ResourceTally` import warning/error split between default and all-features clippy.

Refs #3067.

## Review follow-up

- Addressed reviewer finding: previous E2E assertion allowed `[REDACTED].<jwt-tail>` to pass. Tightened the E2E assertion and added a direct `ironclaw_safety` regression test; both failed before the regex fix and pass now.

## Tests

- `CARGO_TARGET_DIR=/tmp/ironclaw-target-3067-e2e cargo test -q -p ironclaw_safety leak_detector::tests::test_redact_bearer_jwt_token_without_tail_leak -- --exact`
- `CARGO_TARGET_DIR=/tmp/ironclaw-target-3067-e2e cargo test -q -p ironclaw_host_runtime --tests --no-fail-fast`
- `CARGO_TARGET_DIR=/tmp/ironclaw-target-3067-e2e cargo clippy -q -p ironclaw_host_runtime --tests --all-features -- -D warnings`
- `CARGO_TARGET_DIR=/tmp/ironclaw-target-3067-e2e cargo clippy -q -p ironclaw_safety -p ironclaw_host_runtime --tests -- -D warnings`

### DIFF
--- a/crates/ironclaw_host_runtime/tests/builtin_obligation_handler_contract.rs
+++ b/crates/ironclaw_host_runtime/tests/builtin_obligation_handler_contract.rs
@@ -549,7 +549,7 @@ async fn builtin_obligation_handler_fails_closed_when_redacted_object_keys_colli
         &capability_id,
         json!({
             leaked: "secret-key",
-            "[REDACTED].aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa.bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb": "existing",
+            "[REDACTED]": "existing",
         }),
     );
 
--- a/crates/ironclaw_host_runtime/tests/host_runtime_services_contract.rs
+++ b/crates/ironclaw_host_runtime/tests/host_runtime_services_contract.rs
@@ -54,7 +54,7 @@ use ironclaw_reborn_event_store::{
 };
 use ironclaw_resources::{
     InMemoryResourceGovernor, JsonFileResourceGovernorStore, PersistentResourceGovernor,
-    ResourceAccount, ResourceError, ResourceGovernor, ResourceLimits, ResourceTally,
+    ResourceAccount, ResourceError, ResourceGovernor, ResourceLimits,
 };
 #[cfg(feature = "libsql")]
 use ironclaw_run_state::LibSqlRunStateApprovalStore;
@@ -337,7 +337,7 @@ async fn with_libsql_resource_governor_closes_process_reservations_on_cancel() {
     assert_eq!(outcome.cancelled, vec![RuntimeWorkId::Process(process_id)]);
     assert_eq!(
         governor.reserved_for(&account).unwrap(),
-        ResourceTally::default()
+        ironclaw_resources::ResourceTally::default()
     );
     assert!(matches!(
         governor.release(reservation_id).unwrap_err(),
--- a/crates/ironclaw_host_runtime/tests/host_runtime_services_contract.rs
+++ b/crates/ironclaw_host_runtime/tests/host_runtime_services_contract.rs
@@ -52,6 +52,8 @@ use ironclaw_processes::{
 use ironclaw_reborn_event_store::{
     RebornEventStoreConfig, RebornEventStoreError, RebornProfile, build_reborn_event_stores,
 };
+#[cfg(feature = "libsql")]
+use ironclaw_resources::ResourceTally;
 use ironclaw_resources::{
     InMemoryResourceGovernor, JsonFileResourceGovernorStore, PersistentResourceGovernor,
     ResourceAccount, ResourceError, ResourceGovernor, ResourceLimits,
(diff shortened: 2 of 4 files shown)


ilblackdragon:
# Code Review: PR #3444 — fix(reborn): strengthen host runtime publication gates

## Overview

Small, well-scoped security fix targeting `reborn-integration`. Closes a leak-detector regex hole where dotted JWT-style bearer tokens (`Bearer header.payload.signature`) redacted only the header, leaving the payload and signature visible as `[REDACTED].<jwt-tail>`. One-line regex fix in `ironclaw_safety` plus comprehensive regression coverage at three layers:

- **Unit-level** in `ironclaw_safety::leak_detector` (the regex itself).
- **Test fixture** correction in `builtin_obligation_handler_contract` that was actually asserting the buggy output.
- **E2E** via two new tests in `reborn_e2e_gate.rs` driving `HostRuntimeServices → DefaultHostRuntime → ScriptRuntime` for both `RedactOutput` and `EnforceOutputLimit` obligations.

Refactor of `ObligatingAuthorizer` to accept a parameterized obligation list is a clean side-effect of supporting the new tests.

## Strengths

- **Multi-layer coverage.** The same defect is now guarded at the regex unit level, the obligation-handler contract level, and the public-completion boundary. A future regression in any of those layers fails loudly.
- **Honest before/after framing.** The "Review follow-up" note in the PR description explicitly calls out that the prior E2E assertion accepted leaked tail fragments — exactly the right way to land a security fix that also corrects a buggy test.
- **E2E tests assert both the expected shape AND absence of every fragment** (header, payload, signature) in the serialized output — not just one. That's a tighter assertion than `contains("[REDACTED]")` would have been, and the loop-over-fragments pattern is reusable.
- **Failure-path tests check the run-state record, the event sink, AND the failure debug rendering** for sentinel leaks — covers the publication boundary thoroughly.

## Issues / Suggestions

### Correctness

- **Regex is now greedier than the JWT case requires.** `[a-zA-Z0-9._-]{20,}` will consume any contiguous run of those chars after `Bearer\s+`, including a trailing `.` before whitespace. For `"Bearer foo.bar.baz."` it'll match the trailing dot too — harmless for redaction (you're erring conservative), but worth a test confirming a non-JWT bearer token in the middle of prose still redacts correctly and doesn't over-eat into adjacent text. Consider an anchored test like `"Bearer abcdef1234567890123456. The next sentence."` to lock the boundary behavior.
- **`ResourceTally` change is functional but slightly off-pattern.** Inlining `ironclaw_resources::ResourceTally::default()` at the one use site works, but the use site is inside a `#[cfg(feature = "libsql")]` function. The cleaner fix is `#[cfg(feature = "libsql")] use ironclaw_resources::ResourceTally;` so the import name stays consistent with sibling imports. Tiny nit; not a blocker.

### Test coverage gaps

- **No assertion that the `run_state` record excludes the sentinel** in the oversized-output test. You check `events.events()` and the `failure:?` debug render, but `run_state.get(...)` is fetched and only its `status` and `error_kind` are checked. Add `let serialized_run = serde_json::to_string(&run).unwrap(); assert!(!serialized_run.contains(forbidden));` for symmetry with the redact test's events check — the run record is also a public-publication surface.
- **No coverage of `Bearer` tokens that legitimately contain dots but are not JWTs** (opaque tokens with dotted prefixes). Probably fine — over-redaction is the safe failure mode for bearer tokens — but worth an explicit test or doc comment noting "this regex intentionally over-matches dotted tokens to handle JWTs."

### Style / project conventions

- **`.unwrap()` in tests is fine** per CLAUDE.md. `Regex::new(...).unwrap()` in production has a `// safety: hardcoded literal` comment that's load-bearing. OK.
- The `Obligations::new(self.obligations.clone()).unwrap()` per-call clone is minor allocation overhead in tests only — fine.

### Security

Right type of fix in the right places. The previous test fixture (`"[REDACTED].aaaaa...bbbbb..."`) was effectively *codifying* the leak as expected output — the kind of latent bug review-with-fresh-eyes catches. Tightening that fixture and adding a direct regex test is the correct response.

One follow-up worth tracking but out of scope here: the `bearer_token` pattern still won't catch base64-padded tokens (`+`, `/`, `=`) that some non-JWT OAuth deployments use. If that surfaces, the fix is the same shape (extend the character class). For this PR, JWT is the documented target and is now covered.

## Risk

`risk: low` is accurate. Diff is 154 lines, mostly tests; the production change is one character (`.` added to a regex character class) with a direct test that fails before the fix. The two E2E tests exercise actual `RedactOutput` and `EnforceOutputLimit` flows on the runtime, which would have caught any wiring regression. Approve once the suggested run-state assertion (and optionally an anchor-boundary regex test) are added.

ilblackdragon (approved):

Issue #3424, opened 2026-05-09
[Reborn] Add trusted LoopExitApplier


serrrfirat:
Parent: #3107  
Related: #3016, #3199, #3202, #3232, #3404, #3406, #3420, #3423  
Reborn epic: #2987

## Purpose

Add a trusted Reborn `LoopExitApplier`: the runner-owned service that validates `AgentLoopDriver` `LoopExit` claims against durable evidence before applying turn-state transitions.

`LoopExit` is only a driver claim. The applier decides whether it is trusted, terminal failure, blocked, cancelled, or `RecoveryRequired`.

## Existing state

`ironclaw_turns` already defines:

- `LoopExit::{Completed, Blocked, Cancelled, Failed}`;
- `LoopExitValidationPolicy` with evidence booleans;
- `LoopExit::validate(policy)`;
- `TurnRunTransitionPort::apply_validated_loop_exit(...)`;
- `apply_loop_exit(...)` helper that accepts caller-provided validation policy.

The gap: production runner composition must not let drivers or arbitrary callers supply evidence booleans. Evidence must be derived from durable stores.

### Ownership

`LoopExitApplier` lives in `ironclaw_reborn` runner composition.

It verifies durable evidence, computes `LoopExitValidationPolicy`, maps invalid exits, and calls the turn transition port.

```text
AgentLoopDriver returns LoopExit claim
  -> LoopExitApplier verifies evidence
  -> LoopExit::validate(derived_policy)
  -> TurnRunTransitionPort::apply_validated_loop_exit(...)
```

Driver never sets evidence booleans.

### Public helper discipline

Production runner should not use `apply_loop_exit(...)` with arbitrary caller-provided policy. Keep low-level validation primitives only for contracts/tests if needed.

### Completion

For `LoopCompleted`, verify every completion ref:

- belongs to same scope/thread/run;
- exists durably;
- is finalized/committed, not draft;
- was produced before/at final checkpoint boundary;
- is allowed by the resolved run profile.

Cross-run, missing, draft, or uncommitted refs are invalid.

### NoReply completion

`LoopCompletionKind::NoReply` is allowed only if the resolved run profile explicitly permits no-reply completion.

Production profiles require a final checkpoint. Foreground/default chat profiles should reject arbitrary no-ref completion.

### Terminal final checkpoint

Production profiles require `final_checkpoint_id` for trusted terminal exits:

- `Completed`;
- `Cancelled`;
- `Failed`.

[...]

PR #3446, opened 2026-05-09
feat(reborn): add concrete TurnRunner worker composition

### LINKED ISSUE #3404: [Reborn] Add concrete TurnRunner worker composition
Parent: #3107  
Depends on: #3402  
Related: #3199, #3202, #3016, #3204, #3232  
Reborn epic: #2987

## Purpose

Add the concrete Reborn `TurnRunner` worker composition that claims queued/resumed turn runs, heartbeats the runner lease, selects a registered loop driver, constructs a per-run `AgentLoopDriverHost`, invokes the driver, and applies the returned `LoopExit` through trusted transition ports.

This issue builds on the already-landed `ironclaw_turns` claim/lease/heartbeat state machine from #3199/#3202. It does **not** redefine turn persistence or active-lock semantics.

## Current state

`origin/reborn-integration` already has the low-level runner contract in `ironclaw_turns`:

- `ClaimRunRequest { runner_id, lease_token, scope_filter }`
- `HeartbeatRequest { run_id, runner_id, lease_token }`
- terminal/block/recovery transition requests all require the active `runner_id + lease_token`
- stores persist `runner_id`, `lease_token`, `lease_expires_at`, `last_heartbeat_at`, and `claim_count`
- expired leases transition to `RecoveryRequired` instead of blind retry

Missing piece: a concrete worker in Reborn composition that uses those contracts with real drivers/hosts.

## What to build

Add a Reborn-owned runner worker, likely in `crates/ironclaw_reborn`, that owns:

```text
stable TurnRunnerId per worker instance
fresh TurnLeaseToken per claimed run
claim loop
heartbeat task
registered driver lookup
AgentLoopDriverHost factory call
AgentLoopDriver::run/resume invocation
LoopExit application through TurnRunTransitionPort/apply_loop_exit
panic/crash/error mapping to RecoveryRequired where safety is unknown
wake-driven execution with bounded fallback polling
```

## Architecture boundaries

```text
ironclaw_turns:
  owns TurnRunTransitionPort, claim/heartbeat/transition DTOs, state-machine invariants, apply_loop_exit helper

ironclaw_reborn:
  owns concrete worker loop, driver registry lookup, host factory, readiness/config, worker lifecycle
```

The worker must not mutate turn state directly. It must call trusted transition-port APIs.

### Claim strategy

Start with one-at-a-time claim per worker loop:

```text
loop:
  wait for wake or fallback poll tick
  claim_next_run(...)
  if none: continue
  run one claimed run to LoopExit/application
  repeat
```

Batch claiming is out of scope for the first worker slice.

### Runner identity and lease tokens

- Generate one stable `TurnRunnerId` for the worker instance.
- Generate a fresh `TurnLeaseToken` for every claim attempt/run.
- Use the same `runner_id + lease_token` for heartbeat and final/block/fail/recovery transition of that claimed run.

### Wake behavior

Use wake-driven execution with fallback polling:

- wake on submit/resume/completion-related queued-run hints;
- periodic fallback scan for missed wakes/stale queued work;
- wake delivery is not authoritative and must be safe to duplicate/miss.

### Heartbeat ownership

The worker owns heartbeat/lease renewal. Drivers do not renew runner leases directly.

Suggested behavior:

```text
claim run
start heartbeat loop for claimed run
invoke driver
stop heartbeat loop
apply LoopExit / recovery transition
```

Heartbeat failure should stop driver execution where possible and transition to recovery if ownership is lost/uncertain.

### Driver lookup

Use the registry from #3402. Lookup must use the exact persisted driver identity from the run profile. Missing driver at claim time should not terminally fail the run; mark `RecoveryRequired` with a sanitized reason.

### Host factory

The worker calls a Reborn-owned per-run host factory:

```text
LoopRunContext + claimed run/profile -> concrete AgentLoopDriverHost
```

The first implementation may use the text-only host stack, but the worker API should not be hardcoded to text-only forever.

### Driver panic/crash/error handling

- Driver returned valid `LoopExit` -> validate/apply via transition port.
- Driver returned controlled invalid/failure result before unsafe side effects -> map according to `LoopExitValidationPolicy` / sanitized failure policy.
- Driver panic/task abort/unknown crash -> `RecoveryRequired`, not blind retry.
- Lost runner lease/heartbeat expiry -> `RecoveryRequired` via existing recovery path.

### LoopExit application

Use a centralized applier or existing `ironclaw_turns::runner::apply_loop_exit` helper over `TurnRunTransitionPort`.

The worker should not directly call arbitrary complete/block/fail methods in several places unless wrapped behind one consistent application component.

### Cancellation model

V1 cancellation is cooperative:

- cancellation is recorded in turn state;
- driver observes cancellation through host/input APIs in later slices;
- timeout/unknown partial state becomes `RecoveryRequired` / cancel-partial semantics, not blind retry.

The first worker slice may only include safe cancellation detection and recovery/stop behavior; full cooperative input plumbing can be follow-up.

## First vertical slice

The first concrete worker slice should use a simple text-only/reference driver after #3402 rather than immediately integrating the v2 driver adapter.

Target test path:

```text
submit message
  -> queued run
  -> worker wakes/polls
  -> claim_next_run
  -> build text-only AgentLoopDriverHost
  -> reference driver calls host-managed model port
  -> transcript finalizes assistant message
  -> driver returns LoopCompleted
  -> worker applies LoopExit
  -> turn run becomes Completed
```

## Acceptance criteria

- [ ] Worker has stable `TurnRunnerId` per instance and fresh `TurnLeaseToken` per claim.
- [ ] Worker claims at most one run at a time in the first implementation.
- [ ] Worker supports wake-driven execution plus fallback polling.
- [ ] Worker owns heartbeat while driver is active.
- [ ] Worker looks up drivers through the #3402 registry using exact persisted identity.
- [ ] Missing registered driver at claim time records `RecoveryRequired` with sanitized reason.
- [ ] Driver panic/task abort records `RecoveryRequired`.
- [ ] Driver-returned `LoopExit` is applied through trusted transition-port APIs.
- [ ] Worker does not mutate turn state directly.
- [ ] First vertical test proves queued -> claimed -> driver -> transcript/model path -> completed.
- [ ] Tests cover heartbeat ownership and stale/incorrect lease protection through the existing transition port.
- [ ] Tests cover missed wake fallback polling.

## Out of scope

- Batch claiming.
- Full v2 driver adapter.
- Tool-capable `LoopCapabilityPort` implementation.
- Full cooperative cancellation/input polling.
- Checkpoint migration.
- Product UX for recovery/cancel/fork.

## Suggested test names

- `runner_claims_one_run_with_fresh_lease_token`
- `runner_heartbeats_claimed_run_until_driver_returns`
- `runner_applies_completed_loop_exit_through_transition_port`
- `runner_records_recovery_required_for_missing_driver`
- `runner_records_recovery_required_for_driver_panic`
- `runner_fallback_poll_claims_run_after_missed_wake`
- `runner_does_not_mutate_turn_state_directly`

serrrfirat:
## Summary
- add concrete Reborn `TurnRunnerWorker` composition that claims one run at a time, owns runner identity/lease tokens, heartbeats active runs, resolves drivers through the registry, invokes per-run hosts, and applies `LoopExit` through the existing transition port path
- add recovery mapping for missing drivers, host/driver errors, driver panics, heartbeat lease loss, cancellation, and exit-application failures
- wire `RebornLoopDriverHostFactory` into the runner `HostFactory` seam so the worker can build text-only `AgentLoopDriverHost` instances per claimed run

Closes #3404

## Tests
- `CARGO_TARGET_DIR=/tmp/ironclaw-target-pr3446-host cargo test -p ironclaw_reborn text_only_host_factory_implements_turn_runner_host_factory --no-default-features`
- `CARGO_TARGET_DIR=/tmp/ironclaw-target-pr3446-host cargo test -p ironclaw_reborn turn_runner::tests --no-default-features --lib`
- `CARGO_TARGET_DIR=/tmp/ironclaw-target-pr3446-host cargo clippy -p ironclaw_reborn --all-targets --no-default-features -- -D warnings`
- `uv run --python 3.12 python scripts/check_no_panics.py --base origin/reborn-integration --head HEAD`

## Notes
- PR targets `reborn-integration`.
- #3424 (`LoopExitApplier`) is intentionally kept out of this PR and should land separately.
- No merge performed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4514,6 +4514,7 @@ version = "0.1.0"
 dependencies = [
  "async-trait",
  "chrono",
+ "futures-util",
  "ironclaw_host_api",
  "ironclaw_llm",
  "ironclaw_loop_support",
@@ -4526,6 +4527,8 @@ dependencies = [
  "serde_json",
  "tempfile",
  "tokio",
+ "tokio-util",
+ "tracing",
 ]
 
 [[package]]
@@ -8787,6 +8790,7 @@ dependencies = [
  "futures-core",
  "futures-io",
  "futures-sink",
+ "futures-util",
  "pin-project-lite",
  "tokio",
 ]
--- a/crates/ironclaw_reborn/Cargo.toml
+++ b/crates/ironclaw_reborn/Cargo.toml
@@ -17,6 +17,7 @@ libsql-secrets = ["dep:ironclaw_secrets", "ironclaw_secrets/libsql", "dep:libsql
 
 [dependencies]
 async-trait = "0.1"
+futures-util = { version = "0.3", default-features = false }
 ironclaw_llm = { path = "../ironclaw_llm", version = "0.1.0", optional = true, default-features = false }
 ironclaw_loop_support = { path = "../ironclaw_loop_support", version = "0.1.0" }
 ironclaw_secrets = { path = "../ironclaw_secrets", version = "0.1.0", optional = true }
@@ -24,6 +25,9 @@ ironclaw_threads = { path = "../ironclaw_threads", version = "0.1.0" }
 ironclaw_turns = { path = "../ironclaw_turns", version = "0.1.0" }
 libsql = { version = "0.6", optional = true, default-features = false, features = ["core", "replication", "remote", "tls"] }
 secrecy = { version = "0.10", optional = true }
+tokio = { version = "1", features = ["macros", "rt-multi-thread", "sync", "time"] }
+tokio-util = { version = "0.7", features = ["rt"] }
+tracing = "0.1"
 
 [dev-dependencies]
 chrono = "0.4"
@@ -31,7 +35,7 @@ ironclaw_host_api = { path = "../ironclaw_host_api", version = "0.1.0" }
 rust_decimal = "1"
 serde_json = "1"
 tempfile = "3"
-tokio = { version = "1", features = ["macros", "rt"] }
+tokio = { version = "1", features = ["macros", "rt-multi-thread", "time"] }
 
 [[test]]
 name = "llm_gateway"
--- a/crates/ironclaw_reborn/Cargo.toml
+++ b/crates/ironclaw_reborn/Cargo.toml
@@ -17,6 +17,7 @@ libsql-secrets = ["dep:ironclaw_secrets", "ironclaw_secrets/libsql", "dep:libsql
 
 [dependencies]
 async-trait = "0.1"
+chrono = "0.4"
 futures-util = { version = "0.3", default-features = false }
 ironclaw_llm = { path = "../ironclaw_llm", version = "0.1.0", optional = true, default-features = false }
 ironclaw_loop_support = { path = "../ironclaw_loop_support", version = "0.1.0" }
--- a/crates/ironclaw_reborn/src/lib.rs
+++ b/crates/ironclaw_reborn/src/lib.rs
@@ -6,6 +6,7 @@
 
 pub mod driver_registry;
 pub mod loop_driver_host;
+pub mod turn_runner;
 
 #[cfg(feature = "root-llm-provider")]
 pub mod model_gateway;
--- a/crates/ironclaw_reborn/src/loop_driver_host.rs
+++ b/crates/ironclaw_reborn/src/loop_driver_host.rs
@@ -541,6 +541,38 @@ fn validate_claimed_run_context(
     Ok(())
 }
 
+#[async_trait]
+impl<S, G> crate::turn_runner::HostFactory for RebornLoopDriverHostFactory<S, G>
+where
+    S: SessionThreadService + ?Sized + Send + Sync + 'static,
+    G: HostManagedModelGateway + ?Sized + Send + Sync + 'static,
+{
+    async fn create_host(
+        &self,
+        claimed: &ClaimedTurnRun,
+    ) -> Result<
+        Box<dyn ironclaw_turns::run_profile::AgentLoopDriverHost + Send + Sync>,
+        crate::turn_runner::HostFactoryError,
+    > {
+        let loop_run_context = LoopRunContext::new(
+            claimed.state.scope.clone(),
+            claimed.state.turn_id,
+            claimed.state.run_id,
+            claimed.resolved_run_profile.clone(),
+        );
+        self.build_text_only_host(RebornLoopDriverHostRequest {
+            claimed_run: claimed.clone(),
+            loop_run_context,
+        })
+        .await
+        .map(|host| {
+            Box::new(host)
+                as Box<dyn ironclaw_turns::run_profile::AgentLoopDriverHost + Send + Sync>
+        })
+        .map_err(|error| crate::turn_runner::HostFactoryError::new(error.to_string()))
+    }
+}
+
 fn persisted_profile_id(profile_id: &RunProfileId) -> RunProfileId {
     if profile_id.is_interactive_default() {
         RunProfileId::default_profile()
--- a/crates/ironclaw_reborn/src/turn_runner.rs
+++ b/crates/ironclaw_reborn/src/turn_runner.rs
@@ -1,0 +1,601 @@
+//! Concrete Reborn turn-runner worker composition.
+//!
+//! This module owns the worker lifecycle that claims queued/resumed turn runs,
+//! heartbeats the runner lease, selects a registered loop driver, constructs a
+//! per-run `AgentLoopDriverHost`, invokes the driver, and applies the returned
+//! `LoopExit` through trusted transition ports.
+//!
+//! # Architecture boundary
+//!
+//! `ironclaw_turns` owns `TurnRunTransitionPort`, claim/heartbeat/transition
+//! DTOs, state-machine invariants, and the `apply_loop_exit` helper.
+//!
+//! This module owns the concrete worker loop, driver registry lookup, host
+//! factory, readiness/config, and worker lifecycle.
+
+use std::panic::AssertUnwindSafe;
+use std::sync::Arc;
+use std::time::Duration;
+
+use async_trait::async_trait;
+use futures_util::FutureExt;
+use tokio::sync::Notify;
+use tokio_util::sync::CancellationToken;
+use tracing::{debug, error, info, warn};
+
+use ironclaw_turns::{
+    AgentLoopDriverError, AgentLoopDriverResumeRequest, AgentLoopDriverRunRequest, LoopExit,
+    LoopExitValidationPolicy, SanitizedFailure, TurnError, TurnLeaseToken, TurnRunId, TurnRunnerId,
+    TurnScope, TurnStatus,
+    runner::{
+        ApplyLoopExitRequest, ClaimRunRequest, ClaimedTurnRun, HeartbeatRequest,
+        RecordRecoveryRequiredRequest, TurnRunTransitionPort,
+    },
+};
+
+use crate::driver_registry::{DriverRegistry, LoopDriverRegistryKey};
+
+/// Create a `SanitizedFailure` from a known-valid static category.
+///
+/// All categories used here are lowercase ASCII with underscores, satisfying
+/// validation invariants. Returning `None` is only possible if a static literal
+/// is changed to an invalid category.
+fn sanitized_failure(category: &'static str) -> Option<SanitizedFailure> {
+    match SanitizedFailure::new(category) {
+        Ok(failure) => Some(failure),
+        Err(error) => {
+            error!(category, %error, "invalid static recovery failure category");
+            match SanitizedFailure::new("unknown_failure") {
+                Ok(fallback) => Some(fallback),
+                Err(fallback_error) => {
+                    error!(%fallback_error, "fallback recovery failure category invalid");
+                    None
+                }
+            }
+        }
+    }
+}
+
+/// Configuration for the turn-runner worker.
+#[derive(Debug, Clone)]
+pub struct TurnRunnerWorkerConfig {
+    /// How often to send heartbeats for an active run lease.
+    pub heartbeat_interval: Duration,
+
+    /// Fallback poll interval when no wake signal arrives.
+    pub poll_interval: Duration,
+
+    /// Optional scope filter to restrict which runs this worker claims.
+    pub scope_filter: Option<TurnScope>,
+}
+
+impl Default for TurnRunnerWorkerConfig {
+    fn default() -> Self {
+        Self {
+            heartbeat_interval: Duration::from_secs(10),
+            poll_interval: Duration::from_secs(5),
+            scope_filter: None,
+        }
+    }
+}
+
+/// Factory trait for constructing a per-run `AgentLoopDriverHost`.
+///
+/// The host is created once per claimed run and provides the driver with access
+/// to model, transcript, checkpoint, input, capabilities, and progress services.
+#[async_trait]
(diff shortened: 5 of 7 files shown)


ilblackdragon:
# Code Review: PR #3446 — KB-001: wire trusted loop exit applier

## Overview

Wires a trusted `LoopExitApplier` and concrete `TurnRunnerWorker` into the Reborn loop architecture (~2,200 added lines, mostly in `crates/ironclaw_reborn/`). Establishes the security boundary that drivers *declare* loop exits but cannot set evidence booleans — instead, the applier derives evidence from durable stores via a `LoopExitEvidencePort`, computes `LoopExitValidationPolicy`, and delegates to the existing `LoopExit::validate()` path. Also extends `TurnStatus`/`BlockedReason`/`LoopBlockedKind` with a new `Process` variant for spawned-process suspension and adds a `require_final_checkpoint` knob to `CheckpointPolicy`.

The architectural design is sound: `ironclaw_turns` keeps state-machine ownership; `ironclaw_reborn` owns the worker loop, driver registry lookup, host factory, and `TurnRunnerWorker` composition. Testability is excellent — both ports are fully injectable.

## Strengths

- **Clean trust boundary.** The driver→exit→evidence-derivation→validate→apply pipeline is well-articulated in the module docs. Drivers can lie about evidence; the applier ignores them and queries the store. Good defense-in-depth.
- **Comprehensive test coverage** (1,303 LOC of tests). All four `LoopExit` variants × verified/unverified, profile-driven checkpoint policy, worker lifecycle, recovery on driver/host/registry failure, wake-signal vs poll path, and heartbeat liveness during driver execution.
- **Heartbeat loop is correctly drained** (`heartbeat_cancel.cancel(); let _ = heartbeat_handle.await;`) — no orphan task on the lease.
- **Match arms are kept exhaustive** for the three new `Process` variants in `loop_exit.rs`, `status.rs::to_status`, and `status.rs::gate_ref`.

## Issues to address

### Correctness

- **`turn_runner.rs` uses `.expect("infallible: checked is_some")`** despite the project rule "No `.unwrap()` or `.expect()` in production code." Same in `sanitized_failure()` (`.expect("infallible literal")`). Even with safety comments, prefer `if let Some(checkpoint_id) = claimed.state.checkpoint_id` to avoid the lint footgun and make the impossible case unrepresentable.
- **`InMemoryLoopExitEvidencePort` is `pub` (not `#[cfg(test)]`-gated)** in `loop_exit_applier.rs`. `all_verified()` returns a port that bypasses the entire trust boundary this PR exists to enforce. A future caller wiring it into production by mistake would silently neuter the security guarantee. Move it behind `#[cfg(any(test, feature = "test-utils"))]` or to a separate `test_utils` module.
- **`BlockedProcess` audit gap.** Adding `TurnStatus::BlockedProcess` is a schema-shaping change. The PR only updates `memory.rs` (in-memory store). Verify that DB-backed turn stores, status renderers, and any recovery sweepers in other crates handle the new variant — `match`es with wildcard arms will silently miss it. Worth a `grep` of `TurnStatus::Blocked` across the workspace before merge.
- **`require_final_checkpoint: false`** is set for both `interactive_profile()` and `long_running_mission_profile()`. The doc says "Production profiles set this" but in practice no profile in this PR enforces it. This is fine if it's intentionally a follow-up, but worth a TODO marker so the gap doesn't get lost.

### Style / project conventions

- **`info!` from a worker.** `turn_runner.rs` emits `info!` on every worker start/stop, claim, and successful exit. CLAUDE.md says background tasks must use `debug!` because `info!` corrupts the REPL/TUI. The turn-runner is per-process work, but if it ever shares a process with the interactive channel this will leak. Recommend demoting the per-claim/per-exit lines to `debug!` and keeping only worker-lifecycle at `info!`.
- **`Mutex` (sync) inside async tests** — fine in practice (locks are brief), but tokio's `Mutex` is the more conventional choice in async code. Test-only, low priority.

### Test coverage gaps

- **No test for `apply_validated_loop_exit` failing** (the second-tier `record_recovery_required` fallback path in `apply_exit`).
- **No test of cancellation arriving mid-driver-invocation.** The worker has no way to preempt the driver — if that's intentional (drivers run to completion + then we cancel), worth a comment; if not, worth a test.
- **PR description says "Not run locally for PR creation."** For an XL PR adding new modules with public types, please run `cargo fmt && cargo clippy --all --tests --all-features && cargo test -p ironclaw_reborn -p ironclaw_turns` before merge.

### Minor

- `HostFactoryError` carries only `reason: String`. No retriable/permanent distinction, so all host failures recover identically. Fine for v1; flag it for follow-up if host-side errors diverge.
- `Cargo.toml` adds `chrono` as a dev-dep but doesn't explicitly require any non-default features — works because `chrono` defaults include `clock`. OK.

## Risk assessment

Low-risk by `risk:` label, but the data-model extensions (`BlockedProcess`, `LoopBlockedKind::Process`, `BlockedReason::Process`) are wider than the diff would suggest. The applier itself is well-isolated and not yet wired into a production transition port — concrete wiring will land in a follow-up. Approve once the `pub InMemoryLoopExitEvidencePort` exposure and the `BlockedProcess` downstream-consumer audit are addressed; the `expect()` cleanup and `info!`→`debug!` demotions are nice-to-have.

serrrfirat:
Addressed review feedback from @ilblackdragon and @gemini-code-assist in the latest two commits:

- `4ecb9bb7` — Illia feedback:
  - removed production `.expect()` paths in `turn_runner.rs`
  - moved `InMemoryLoopExitEvidencePort` behind `#[cfg(test)]`
  - demoted turn-runner worker `info!` logs to `debug!`
  - added `BlockedProcess` DB persistence coverage for libSQL/Postgres-backed stores
  - added coverage for `apply_validated_loop_exit` failure fallback to recovery
  - documented and tested cancellation during in-flight driver invocation
  - added TODO markers for final-checkpoint production-profile rollout and host error classification

- `67f7884a` — Gemini feedback:
  - replaced detached heartbeat task with same-task `tokio::select!` concurrency, so heartbeat cannot outlive the claimed run future
  - changed `try_claim_and_run` to return `bool` and drain queued runs until empty after each wake/poll tick
  - derived `require_final_checkpoint` for `LoopExit::Cancelled`
  - enforced final-checkpoint policy for `Cancelled` and `Failed` terminal exits in `LoopExit::validate`
  - added regression coverage for queue draining and terminal final-checkpoint enforcement

One Gemini suggestion (`tokio-util` feature `sync`) was not applied because `tokio-util 0.7.18` has no `sync` feature. Verified `CancellationToken` compiles with the existing `rt` feature via `cargo check -p ironclaw_reborn --features root-llm-provider`.

Verification run locally:
- `cargo fmt --check`
- `cargo clippy --all --benches --tests --examples --all-features --quiet -- -D warnings`
- `cargo test -p ironclaw_reborn -p ironclaw_turns --all-features --quiet`

GitHub checks after push: `classify` pass, `scope` pass, Mintlify skipped.

think-in-universe:
# Review: `nearai/ironclaw` `kb/kb-001`

## Findings

### 1. High: expired leases are never recovered by this worker

In `crates/ironclaw_reborn/src/turn_runner.rs`, `TurnRunnerWorker::run()` only waits and then calls `try_claim_and_run()`, which only uses `claim_next_run()`. There is no call to `recover_expired_leases()`, even though the transition-port contract exposes it in `crates/ironclaw_turns/src/runner.rs`.

That means a run left in `Running` or `CancelRequested` after a worker crash, restart, or deploy can remain stranded indefinitely unless some separate recovery service is guaranteed to run elsewhere. `claim_next_run()` only drains queued work, so these runs are not naturally picked back up by this worker.

Suggested fix: wire periodic or startup lease recovery into the worker loop, or document and enforce that another component owns recovery before this worker is considered production-ready.

### 2. High: completion refs are treated as verified without any proof path

`trusted_text_only_exit_validation_policy()` in `crates/ironclaw_reborn/src/turn_runner.rs` sets `completion_refs_verified: true`. In `crates/ironclaw_turns/src/loop_exit.rs`, that causes `LoopExit::Completed` to be accepted as long as it contains non-empty completion refs.

But the worker only gives drivers a generic `AgentLoopDriverHost`; there is no evidence channel tying `LoopExit::Completed.reply_message_refs` or `result_refs` back to host-issued durable writes like `finalize_assistant_message()`. A buggy or compromised driver can fabricate `msg:*` or `result:*` refs and still get the run marked `Completed`.

Suggested fix: either keep `completion_refs_verified` fail-closed until host-issued evidence is plumbed through, or introduce a trusted completion artifact path that the runner can validate before applying `Completed`.

### 3. Medium: backlog draining is artificially throttled to one run per wake/poll cycle

The main loop in `crates/ironclaw_reborn/src/turn_runner.rs` claims at most one run, executes it, and then goes back to waiting. Wake delivery is explicitly best-effort, and `TurnRunnerWakeSender::wake()` uses `notify_one()`.

If many runs are already queued at startup, or several enqueue events collapse into a single wake, the worker processes one run and may then sleep until `poll_interval` before claiming the next. With the default 5-second poll interval, this can seriously delay backlog recovery after downtime or bursty traffic.

Suggested fix: after each successful run, continue draining immediately until `claim_next_run()` returns `None`, then fall back to waiting.

## Assumption

- If there is already a separate lease-recovery service outside this branch, finding 1 is reduced in severity. I did not find that wiring in the reviewed branch delta.

serrrfirat:
All review comments are fixed in latest head `ebd9bed1`.

Covered:
- expired lease recovery before claiming
- queued-run draining until empty
- worker logs demoted from `info!` to `debug!`
- default completion refs fail closed until durable evidence exists
- expected recovery-record lease/race rejections downgraded from error logs
- runner_id/lease_token invariants covered in tests across claim → heartbeat → exit/recovery
- Copilot inline comments verified against current head

Checks are green for `ebd9bed1`.

zmanian:
# Post-merge review notes

Reviewed the final merged state at `ebd9bed1`. The pre-merge cycle (ilblackdragon, think-in-universe, Gemini, Copilot) caught the load-bearing issues — lease recovery wired into the loop, completion refs fail-closed by default, drain-until-empty, `.expect()` removed, heartbeat scoped to the run future, `InMemoryLoopExitEvidencePort` test-gated. All resolved cleanly.

A few non-blocking observations on the final state for follow-up visibility:

1. **Double-failure path in `apply_exit`** (`turn_runner.rs:601-631`). If `apply_loop_exit` fails *and* the fallback `record_recovery_required` also fails, the run is stranded until the next `recover_expired_leases` sweep. That's actually OK now that recovery runs at the top of every loop iteration — worth a comment so a future reader knows the safety net exists by design.

2. **Lease token ownership.** `try_claim_and_run` mints a `TurnLeaseToken` on the worker side and passes it in `ClaimRunRequest`, but `execute_claimed_run` then uses `claimed.lease_token` from the response. If the `TurnRunTransitionPort` contract guarantees the request token is what's returned, the dance is harmless; if the port may substitute, the trait doc should say so explicitly.

3. **`HostFactoryError` has no retriable/permanent distinction** — all host failures land in `host_creation_failed` recovery identically. Fine for v1, TODO marker is in place.

4. **`heartbeat_loop` returns `Err` on the first failed heartbeat.** Correct for the race-the-driver design, but means a transient DB blip during a long driver invocation tears the run down and forces recovery rather than retrying. Acceptable trade-off — flagging for visibility.

5. **No `info!` worker-lifecycle line.** The `info!` → `debug!` migration was the right call for REPL/TUI safety, but it dropped the start/stop lines too. For a long-lived production worker, one `info!` at process start is usually worth keeping. Tiny nit.

Verdict: approved as merged.

Issue #3016, opened 2026-04-28
Reborn cutover blocker: add reference AgentLoopHost facade


serrrfirat:
## Purpose

Track the Reborn post-group cutover blocker for the **reference loop host facade** from #2987.

This issue defines the host-facing facade that a shipped/reference agent loop receives. The goal is to let the loop decide *when* to ask the model and *which visible capabilities* to request, while the host/kernel continues to own policy, authority, turn state, transcript persistence, events, model gateway policy, and side-effect execution.

This is a cutover blocker, not an early substrate blocker. It should remain unused by default until the final Reborn composition/cutover work explicitly wires it into production paths.

## Why this is needed

The Reborn contracts separate:

```text
TurnCoordinator / kernel services  -> admission, lifecycle, blocked/resume, cancellation, durable progress
Agent loop / userland behavior     -> model loop strategy, prompt formatting, tool/capability selection
CapabilityHost                     -> authorized capability invocation/spawn/resume
ProcessHost                        -> process lifecycle/result/cancellation mechanics
Memory/transcript services         -> durable context and conversation records
Events                             -> redacted durable progress/audit substrate
```

Without a narrow facade, a loop implementation can accidentally receive raw service managers and bypass the intended boundaries. That recreates the pre-Reborn failure mode where agent behavior, authorization, runtime dispatch, transcript writes, and channel progress are tightly coupled.

## Decisions already made

These decisions came out of the #2987 follow-up planning discussion and should be treated as the issue contract:

1. **Expose one narrow `AgentLoopHost` facade to loops.**
   - The reference loop receives a single cohesive facade, not raw managers.
   - The facade is internally composed from `TurnCoordinator`, `CapabilityHost`, model gateway, memory/transcript services, events, and checkpoint storage.
   - Raw service handles are intentionally not exposed to loop implementations.

2. **Host returns policy-filtered context bundles; loop owns formatting.**
   - `AgentLoopHost` may expose `load_thread_snapshot` and `build_instruction_bundle`-style methods.
   - Returned context is already authorized, identity-filtered, group-chat/privacy filtered, and safety-filtered.
   - The loop/reference prompt assembler owns ordering, summarization, provider-specific formatting, and strategy.

3. **No direct process API in `AgentLoopHost` MVP.**
   - Loops invoke capabilities.
   - Spawned/background processes appear as structured capability outcomes and durable progress events.
[...]

PR #3454, opened 2026-05-10
Add Reborn loop capability host-runtime adapter slice

serrrfirat:
## Summary
- Adds a Reborn text-only `AgentLoopDriverHost` capability-port adapter that can route through an injected `HostRuntime` instead of the empty capability stub.
- Adds `HostRuntimeLoopCapabilityPort` with visible-surface snapshot enforcement, outside-surface/stale-surface fail-closed behavior, host-runtime invocation routing, safe outcome mapping, result-ref handoff hooks, per-call invocation context refresh, and same-host idempotency replay for terminal side-effect outcomes.
- Adds caller-level tests for visible surface/invoke routing, outside-surface denial before host runtime dispatch, refetched/stale surface handling, runtime outcome capability-id mismatch rejection, sanitized failure summaries, same-host retry behavior, approval/auth/resource/process suspension mapping, and stop-on-first-suspension batch behavior.

Refs #3016

## Scope note
This PR is a contract/adapter slice, not full production cutover.

The default text-only host factory still builds an empty capability port unless a caller injects a `LoopCapabilityPort`. Production composition with `DefaultHostRuntime`, durable capability input/result stores, run-scoped input refs, and restart-safe idempotency remains follow-up work before claiming end-to-end production capability execution.

## Tests
- `CARGO_TARGET_DIR=/tmp/ironclaw-target-pr3454-rec-fix cargo test -p ironclaw_reborn --no-default-features`
- `CARGO_TARGET_DIR=/tmp/ironclaw-target-pr3454-rec-fix cargo test -p ironclaw_reborn --test loop_driver_host text_only_host_maps_runtime_suspension_and_process_outcomes --no-default-features`
- `CARGO_TARGET_DIR=/tmp/ironclaw-target-pr3454-rec-fix cargo test -p ironclaw_reborn --test loop_driver_host text_only_host_batch_stops_on_first_suspension_before_later_invocations --no-default-features`
- `CARGO_TARGET_DIR=/tmp/ironclaw-target-pr3454-rebase cargo test -p ironclaw_turns --test loop_exit_contract --no-default-features`
- `CARGO_TARGET_DIR=/tmp/ironclaw-target-pr3454-rec-fix cargo clippy -p ironclaw_reborn --all-targets --no-default-features -- -D warnings`
- `/Users/firatsertgoz/.local/bin/python3.11 scripts/check_no_panics.py --base origin/reborn-integration --head HEAD`
- `/Users/firatsertgoz/.local/bin/python3.11 -m unittest scripts.check_no_panics.CheckNoPanicsTests`

## Remaining follow-up hardening before production cutover
- Wire production composition through a builder that derives/validates `ExecutionContext` from `LoopRunContext`.
- Add durable idempotency/result ledger semantics across host rebuilds/restarts.
- Add run-scoped immutable capability input/result refs.
- Add checkpoint-before-block driver behavior, model-call surface validation, and sanitized result-ref transcript writes.


zmanian:
# Review

**Summary:** Adds `HostRuntimeLoopCapabilityPort` bridging Reborn's `LoopCapabilityPort` to `ironclaw_host_runtime` with surface-snapshot enforcement, idempotent dispatch, and outcome mapping; plus `require_final_checkpoint` policy flag.

**Findings:**

1. **Unbounded in-memory `dispatch_records` HashMap** (`loop_driver_host.rs` ~254). PR description calls out "restart-safe idempotency is follow-up work," but the current map also has no bounded eviction — every invocation key persists for the host's lifetime. For long-running mission profiles this is a straightforward memory leak. Add LRU or TTL eviction before merge, or document the bound.

2. **`InFlight` sentinel can deadlock concurrent invocations of the same input** (~line 510-515). `invoke_capability_batch` (~line 582) calls `invoke_capability` sequentially with `?`. If a retry arrives while the prior call is still racing, the second observer gets `Unavailable` with a misleading "already in progress" message. No `Notify`/`Condvar` to wait on completion. For batches with duplicate invocations this is silently wrong.

3. **`clear_dispatch` errors are swallowed** (~lines 533, 543: `let _ = self.clear_dispatch(&idempotency_key);`). The only failure mode is a poisoned mutex — after which the in-flight record stays, permanently bricking that idempotency key. Log and propagate, or fail closed.

4. **Wildcard arms over `RuntimeCapabilityOutcome`** (~line 717, `_ => CapabilityOutcome::Failed(...)`). It's a non-exhaustive enum from another crate; new variants (a real `Cancelled` or `Streaming`) will silently downgrade to `Failed` rather than break the build. Given this is a security boundary, prefer exhaustive matching with an explicit `Unknown` arm.

5. **`host_runtime_error` collapses non-`InvalidRequest`/`Unavailable` variants to `Unavailable`** (~line 793). Auth, rate-limit, and timeout errors all map to the same kind — drivers can't distinguish transient vs permanent for retry decisions. The `safe_summary` is also static.

6. **`require_final_checkpoint: #[serde(default)]`** (`policy.rs` ~line 26). Missing-field-defaults-to-permissive is the wrong default for a durability gate. Older deserialized profiles silently bypass the check.

**Minor:** `invocation_context_from_visible` reuses the host-build-time `ExecutionContext` and only overwrites invocation/correlation ids — `principal`/`trust_class`/`mounts` carry over. Confirm intentional for this slice. Cargo.toml widens prod deps (authorization, extensions, filesystem, processes, resources, scripts, trust) — confirm these are used in production paths, not just tests.

**Verdict: Request changes** — (1), (3), (4), (6) are blockers; (2), (5) need an answer.

serrrfirat:
Addressed zmanian review feedback in `d379a0a9`:

- Bounded the in-memory capability dispatch idempotency cache and evict completed records before admitting new keys.
- Reworked duplicate in-flight invocations to wait on completion via `Notify` instead of returning a misleading `Unavailable`.
- Made dispatch cleanup fail closed: cleanup failures are now logged and propagated instead of swallowed.
- Added explicit `RuntimeCapabilityOutcome::Unknown` handling and removed wildcard outcome downgrade in the Reborn loop adapter.
- Preserved host-runtime error kind and sanitized reason summaries when mapping into loop-host errors.
- Changed missing `require_final_checkpoint` serde field to default safe/required; local/test profiles must explicitly relax it.
- Added tests for duplicate waiting, bounded records, unknown outcomes, host-runtime error mapping, final-checkpoint defaulting, and authority/context preservation.

Verification run locally:

- `cargo fmt --all`
- `cargo test -p ironclaw_reborn --test loop_driver_host --no-default-features`
- `cargo test -p ironclaw_turns --test run_profile_contract --no-default-features`
- `cargo test -p ironclaw_host_runtime --no-default-features`
- staged `bash scripts/pre-commit-safety.sh`

GitHub checks are passing for head `d379a0a9`; PR merge state is clean.

zmanian (approved):
# Re-review: approved

All six blocking findings resolved at `d379a0a9`:

1. **Unbounded `dispatch_records` — RESOLVED.** `MAX_IN_MEMORY_DISPATCH_RECORDS = 128` with `evict_completed_until_below_limit()`; saturated cap returns `Unavailable` rather than leaking.
2. **`InFlight` deadlock — RESOLVED.** `DispatchRecord::InFlight { notify: Arc<Notify> }`; concurrent callers wait on the notify and the loop re-reserves after `notify_waiters()`.
3. **`clear_dispatch` errors — RESOLVED.** All three call sites log via `tracing::warn!` and propagate the error.
4. **Wildcard outcome arms — RESOLVED.** `#[non_exhaustive]` removed from `RuntimeCapabilityOutcome`; explicit `Unknown(RuntimeCapabilityUnknown)` variant; `ensure_runtime_outcome_matches` fully exhaustive.
5. **Error collapsing — RESOLVED.** `HostRuntimeError` `#[non_exhaustive]` removed; `host_runtime_error` matches both variants exhaustively.
6. **`require_final_checkpoint` permissive default — RESOLVED.** Now `#[serde(default = "default_require_final_checkpoint")]` returning `true`.

Follow-up: `blocked_summary` still has a `_ =>` wildcard over `RuntimeBlockedReason` (line ~999). Worth promoting to exhaustive in a follow-up if that enum gains variants. Not blocking.

PR #3460, opened 2026-05-10
feat(reborn): add trusted LoopExitApplier

serrrfirat:
## Summary
- add Reborn `LoopExitApplier` that derives `LoopExitValidationPolicy` from runner-owned evidence ports before applying validated transitions
- wire `TurnRunnerWorker` to use the applier instead of direct `apply_loop_exit(...)` with caller-provided policy
- extend loop-exit validation for explicit no-reply permission, verified final checkpoints, and terminal final-checkpoint enforcement
- add conservative text/checkpoint-backed evidence adapter that verifies finalized same-run transcript refs and final checkpoints, fails closed for block/failure/cancel evidence until durable stores are wired, and explicitly rejects agentless transcript evidence because `ironclaw_threads::ThreadScope` is currently agent-scoped

Stacked on #3446 because #3424's production-runner integration depends on the concrete TurnRunner worker from #3404.

Part of #3424.

## Tests
- `CARGO_TARGET_DIR=/tmp/ironclaw-target-3460-fix cargo test -p ironclaw_reborn --no-default-features --lib`
- `CARGO_TARGET_DIR=/tmp/ironclaw-target-3460-fix cargo test -p ironclaw_turns --no-default-features`
- `CARGO_TARGET_DIR=/tmp/ironclaw-target-3460-fix cargo clippy -p ironclaw_reborn --all-targets --no-default-features -- -D warnings`
- `CARGO_TARGET_DIR=/tmp/ironclaw-target-3460-fix cargo clippy -p ironclaw_turns --all-targets --no-default-features -- -D warnings`

## Notes
- PR base is `reborn-integration`; stack parent #3446 has merged.
- Spawned-process blocked/resume support and durable gate/process/failure/cancel evidence stores remain follow-up work for #3424.
- No merge performed.


zmanian:
# Review

**Summary:** Replaces driver-supplied `LoopExitValidationPolicy` with a trusted `LoopExitApplier` that derives policy from runner-owned evidence ports before applying validated transitions.

**Findings beyond the bot reviews:**

1. **Trust-model leak — `LoopExitValidationPolicy` fields stay `pub`** (`crates/ironclaw_turns/src/loop_exit.rs:202-220`). The applier sets them correctly, but `apply_loop_exit` still exists on `TurnRunTransitionPort` accepting a caller-supplied policy via `ApplyLoopExitRequest`. Any test, contract test, or new code path can construct a fully-verified policy directly. The old trust-bypass surface remains. Recommend sealing construction behind a `pub(crate)` constructor or replacing booleans with witness types only the applier can mint.

2. **`InMemoryLoopExitEvidencePort::all_verified()` is `pub`** (~line 312) and not gated behind `#[cfg(test)]` or a `test-utils` feature. Production code can fabricate a fully-trusting evidence port — direct hole in the trust model this PR is meant to close.

3. **`BlockedProcess` variant audit gap.** Copilot flagged `memory.rs`; worth running `rg "match.*TurnStatus" --type rust` across the workspace before merge — wildcard arms silently absorb the new variant.

4. **Cancellation final-checkpoint enforcement may be too strict** (`loop_exit.rs:1764-1772`): `require_final_checkpoint` + `Cancelled` exit without a checkpoint → `driver_protocol_violation`. Host-initiated cancellation can legitimately preempt the driver before any final checkpoint exists. Recommend distinguishing host-cancel-without-checkpoint (acceptable) from driver-claimed-cancel-without-checkpoint.

5. No `info!`/REPL issues found — applier uses `debug!`/`warn!`/`error!` appropriately. No production `.expect()`.

**Verdict: Request changes** — accept after addressing (1) and (2). The bot-raised items (mandatory `agent_id`, `latest_checkpoint_kind` hardcoded, `Ok(true)` failure evidence) are valid and should also land, but most read as "tracked debt in the next slice" per the fail-closed framing — that's acceptable scoping if explicitly tracked.

zmanian:
# Cross-cutting feedback across the Reborn PR stack

Posting here because the trust-boundary pattern originates in this PR. After reviewing #3460, #3454, #3462, #3468, #3470, #3471, #3476, and #3353 together, several issues are *structural* and worth addressing once rather than per-PR. cc @serrrfirat

## 1. Trust boundaries leak through `pub` API surface

Same shape in **#3460** (`LoopExitValidationPolicy` fields `pub`, `InMemoryLoopExitEvidencePort::all_verified()` not test-gated), **#3470** (FNV snapshot version forgeable for a "trust-aware" service), **#3476** (`SkillTrust` plumbed to prompt but not to dispatcher). Trusted component introduced, data type freely constructable elsewhere. Trust guarantee is by convention, not by type.

**Recommend:** witness types or `pub(crate)` constructors on every policy/evidence struct so only the trusted code path can mint them. Single most impactful structural fix.

## 2. Defense-in-depth depth-of-one against prompt injection

**#3471** (memory snippets verbatim), **#3476** (SKILL.md `prompt_content` verbatim), **#3470** (skill descriptions). Every PR injecting retrieved content relies on a single `LoopSafeSummary`-style sanitizer that catches secrets/paths but not instruction hijacks (`<system>`, "ignore previous instructions", fake "user:" turns).

**Recommend:** standardize an untrusted-content envelope (`<untrusted-content source="memory|skill" trust="installed">…</untrusted-content>`) the model is system-prompted to treat as data. Bolt in once at prompt-assembly, not per-source.

## 3. Non-cryptographic hashes used for security-adjacent claims

FNV-1a appears in **#3471** (snippet refs), **#3470** (snapshot version), **#3476** (snippet-ref binding — duplicated verbatim). Each PR documents "not for security" but each is the only tamper check in its layer.

**Recommend:** project-wide policy. Either (a) blake3/SHA-256 for any hash gating trust decisions, or (b) keep FNV but explicitly require a second authenticity check at every consumer.

## 4. Errors collapsed at adapter boundaries

**#3454** (`host_runtime_error` → `Unavailable` for everything), **#3462** (pool-miss → `PolicyDenied`), **#3460** (cancellation-without-checkpoint → `driver_protocol_violation`), **#3471** (`ContextBudgetExceeded` for config errors *and* legitimate oversize). Drivers can't distinguish transient vs permanent, misconfig vs policy refusal — retry logic is structurally underpowered.

**Recommend:** `Transient | Permanent | Misconfigured` classification on every cross-crate error variant.

## 5. No back-pressure or admission control

**#3353** (unbounded `process_webhook`), **#3454** (unbounded `dispatch_records` HashMap), **#3471** (no total-byte budget), **#3470** (`saturating_add` masks overflow). Reborn currently assumes well-behaved inputs at every seam.

**Recommend:** bounded queues + semaphores at every async entry point, bounded caches with eviction, `checked_*` arithmetic on size accumulators, fail-closed on overflow.

## 6. Sandbox/isolation naming inconsistencies

**#3353** ships a *native* runner inside a crate named `ironclaw_wasm_product_adapters`. **#3454** is a "host-runtime adapter slice" with no panic isolation. Trust boundary is real in design but invisible in code organization.

**Recommend:** crates named `*_wasm_*` should contain only sandboxed code paths; native runners go in `*_native_*` or `*_host_*` crates. Rename now while the surface is small.

## 7. Schema-shaping changes lack downstream audits

`BlockedProcess` (merged #3446 → still surfacing in #3460), new checkpoint columns (#3468), new policy fields (#3454, #3460) — none came with a workspace-wide audit of `match TurnStatus`/`match LoopExit*` arms. Wildcard arms silently absorb them.

**Recommend:** every PR adding a status/exit/policy variant includes the output of `rg "match.*TurnStatus|match.*LoopBlocked" --type rust` and checks each site. Pre-commit hook could enforce for known enums.

## 8. `serde(default)` defaulting permissive

**#3454** (`require_final_checkpoint`), implied in **#3462** route policies. Missing-field-defaults-to-allow is the wrong default for any security/durability gate.

**Recommend:** project-wide convention — security-relevant bools have no `serde(default)`, or default to the *restrictive* value.

## 9. Stack discipline

The stack is deep (`reborn-integration` → `kb/kb-007` → ...). Several reviews caught issues that should have been blocked at an earlier slice (e.g., #3476's tool-attenuation gap relies on #3470 being correct, which has its own forgeable-hash issue). Per-PR review is producing redundant requests.

**Recommend:** for the next batch, a "stack overview" comment on the bottom PR enumerating which invariants each slice owns vs defers — so reviewers know whether a gap is "tracked debt in next slice" or "missed."

---

Per-PR review comments with concrete file:line refs are posted on each PR individually. This list is the *meta* — happy to discuss structurally on any one of them.

serrrfirat:
Addressed the review feedback from @zmanian in commit 98a8ab8e:

- sealed the loop-exit validation trust boundary by moving trusted policy derivation into `ironclaw_turns::LoopExitApplier` and making `LoopExitValidationPolicy` crate-private;
- removed the public caller-supplied `ApplyLoopExitRequest` / `apply_loop_exit` bypass path;
- test-gated the permissive `InMemoryLoopExitEvidencePort::all_verified()` and `with_*` helpers;
- allowed observed host cancellation to proceed without requiring a final checkpoint, while unobserved cancellation still fails closed;
- updated the loop-exit / turn-runner contract docs for the new trust boundary;
- ran the `TurnStatus` / `BlockedProcess` audit: no `BlockedProcess` variant found, and no wildcard `TurnStatus` match issue beyond a `matches!` predicate.

Verification run locally:

- `cargo fmt --check`
- `cargo clippy --all --benches --tests --examples --all-features -q`
- `cargo test -p ironclaw_reborn --lib -q`
- `cargo test -p ironclaw_turns --tests -q`
- `git diff --check`

GitHub checks after push: `classify` and `scope` passed.

zmanian (approved):
# Re-review: approved

All four blocking findings from my prior review are resolved at `e2faab01`:

1. **Trust-model leak — RESOLVED.** `LoopExitValidationPolicy` is now `pub(crate)` with private fields and `validate()` is non-pub (`loop_exit.rs:1853-1862`). The only production construction path is `LoopExitApplier::build_policy`, which gates each verified flag on a `LoopExitEvidencePort` query. Sound witness-style fix.
2. **`all_verified()` test-gating — RESOLVED.** `#[cfg(test)]` on the constructor and all `with_*_verified` builders (lines 64, 74-104). Non-test `InMemoryLoopExitEvidencePort::new()` defaults all flags to `false`.
3. **`BlockedProcess` audit — RESOLVED structurally.** Applier routes every `LoopExit::Blocked` through `evidence_port.verify_blocked_evidence` regardless of `LoopBlockedKind`; new variants inherit enforcement automatically.
4. **Cancellation strictness — RESOLVED.** `validate_cancelled_exit` only enforces `final_checkpoint_verified` when `exit.checkpoint_id.is_some()`; new test `observed_host_cancellation_can_preempt_final_checkpoint` locks the behavior. Failure-exit keeps strict enforcement, which is the correct asymmetry.

Minor non-blocker: `allow_no_reply_completion` / `NoReplyNotAllowed` were added as a static policy flag without a port-driven evidence path — confirm not a regression vector for trusted text-only profiles, but acceptable.

Issue #3451, opened 2026-05-10
[Reborn] Add direct DB operations for loop checkpoint mappings


serrrfirat:
Follow-up from #3439 review.

## Problem

`LoopCheckpointStore` DB-backed implementations currently route every operation through full `TurnPersistenceSnapshot` hydration/replacement:

- libSQL `get_loop_checkpoint`: `load_snapshot()` -> `InMemoryTurnStateStore::from_persistence_snapshot()` -> in-memory lookup
- libSQL `put_loop_checkpoint`: full snapshot load + `libsql_replace_snapshot()`
- Postgres `get_loop_checkpoint`: full snapshot load + in-memory lookup
- Postgres `put_loop_checkpoint`: `SHARE ROW EXCLUSIVE` lock + full snapshot replace

Checkpoint calls are expected before model calls, side effects, blocking, and terminal resume points. Snapshot-per-call behavior will become a production bottleneck as checkpoint counts grow.

## Desired outcome

Add direct DB operations for loop checkpoint mappings:

- `put_loop_checkpoint`: insert one `TurnCheckpointId -> LoopCheckpointStateRef` mapping scoped by tenant/agent/project/thread/turn/run/schema/kind
- `get_loop_checkpoint`: direct `SELECT ... WHERE checkpoint_id = ?` plus scope/run checks
- preserve libSQL/Postgres parity
- keep public surfaces redacted: refs/metadata only, no checkpoint payload bytes
- keep existing in-memory/snapshot compatibility for old persisted snapshots

## Acceptance criteria

- libSQL and Postgres implementations avoid full snapshot hydration for single checkpoint reads/writes
- DB contract tests cover direct lookup, cross-scope/cross-run rejection, and snapshot compatibility
- migrations/backfills documented if schema changes are needed

Origin: https://github.com/nearai/ironclaw/pull/3439#issuecomment-4414698577

serrrfirat:
Closing as addressed on `origin/reborn-integration` @ e371d3c39a3b8de8bb1e0d7e6a12d2c4812ae3a1. Evidence: PR #3468 added direct DB-backed loop checkpoint mappings in `crates/ironclaw_turns/src/db.rs` using `turn_loop_checkpoints` direct insert/select paths for libSQL/Postgres. `crates/ironclaw_turns/tests/loop_checkpoint_store_contract.rs` and `db_turn_state_store_contract.rs` cover direct lookup, cross-scope/cross-run rejection, and snapshot compatibility.

PR #3468, opened 2026-05-11
GitHub issue #3451: [Reborn] Add direct DB operations for loop checkpoint mappings

serrrfirat:
Refs #3451

KB task: KB-002

GitHub issue #3451: [Reborn] Add direct DB operations for loop checkpoint mappings

URL: https://github.com/nearai/ironclaw/issues/3451
Repo: nearai/ironclaw
Labels: risk: medium, scope: db, reborn
Assignees: none

Scope: IronClaw Reborn issue imported from GitHub.
Before coding: read full issue body/comments, check linked PRs/blockers, work from current origin/reborn-integration unless issue says otherwise, keep PR tightly scoped.
Acceptance: satisfy GitHub issue acceptance criteria; include tests/verification evidence; never merge without explicit user approval.

Auto-opened by kb when task reached In Review. Auto-merge remains disabled; do not merge without explicit operator approval.


zmanian:
# Review

**Summary:** Replaces wipe-and-reload snapshot persistence for `LoopCheckpointStore` with targeted INSERT/SELECT against `turn_loop_checkpoints`, on both libSQL and Postgres, plus adds `scope_key`/`kind` columns to `turn_checkpoints` with idempotent migrations.

**Findings:**

1. **`.unwrap()` in production code (CLAUDE.md violation).** `crates/ironclaw_turns/src/memory.rs:499-500` and `src/store.rs:530` call `LoopCheckpointStateRef::new("checkpoint:...").unwrap()` as placeholder/serde defaults. The safety comment handwaves it, but the project rule is no `.unwrap()` in prod — use `OnceLock`/`LazyLock` or return `Result`. These encode "placeholder will be threaded later" — landing risks shipping the placeholder permanently.

2. **Non-transactional INSERT+SELECT race.** `libsql_insert_loop_checkpoint_record` (`db.rs:209`) uses `INSERT OR IGNORE` then a separate `SELECT` outside any transaction — a concurrent `UPDATE`/`DELETE` between the two statements can misclassify a true conflict as idempotent or yield "conflicted but row not readable" spuriously. Postgres version (`db.rs:273`) has the same shape on a pooled client without `BEGIN`. Previous code held `SHARE ROW EXCLUSIVE` — that protection is gone. Wrap in a transaction or use `INSERT ... ON CONFLICT DO NOTHING RETURNING` + a single `SELECT FOR UPDATE` follow-up.

3. **JSON re-deserialization as truth, then compared to request fields** (`db.rs:108-114`, `:190-196`). WHERE clause already constrains by `checkpoint_id`/`scope_key`/`turn_id`/`run_id`. If persisted JSON drifts from indexed columns (partial migration, manual repair), this returns `Ok(None)` and silently hides a real row. Either trust the WHERE clause or treat mismatch as an error.

4. **libSQL migration string-matches `"duplicate column name"`** (`db.rs:45`). Fragile across libsql versions/locales. Use `PRAGMA table_info(turn_checkpoints)` for presence check. Also: new `NOT NULL DEFAULT ''` columns mean existing rows get empty `scope_key` — no backfill plan documented.

5. **`checkpoints.is_empty()` tests** (lines 698-700, 729-732) presume loop checkpoints never land in `turn_checkpoints`, but `libsql_replace_snapshot` (`db.rs:251-262`) still writes `TurnCheckpointRecord` checkpoints there with the new columns. Test naming is correct for *loop* checkpoints but the assertion is brittle if any block checkpoint is later added to setup.

**Minor:** test helpers use `.unwrap()` (fine in `#[cfg(test)]`). `created_at.to_rfc3339()` → Postgres `timestamptz` via cast loses microsecond precision vs. binding `chrono::DateTime<Utc>` directly.

**Verdict: Request changes** — (1) and (2) are blocking. Tests and dual-backend coverage otherwise solid.

serrrfirat:
@zmanian addressed in `a37f794d`:

1. Removed production placeholder `.unwrap()` path by threading `LoopCheckpointStateRef` through `BlockRunRequest`, `LoopBlocked`, `TurnRunnerOutcome::Blocked`, and checkpoint record persistence. Legacy serde default now uses an internal sentinel constructor, not `.unwrap()`.
2. Replaced insert-then-select loop-checkpoint idempotency with atomic conflict handling: libSQL uses one upsert statement; Postgres wraps the write in a transaction and only returns success for identical payload conflicts.
3. `get_loop_checkpoint` no longer treats payload/index drift as `Ok(None)`; it now returns a conflict if the selected row payload disagrees with constrained metadata.
4. libSQL migration now checks `PRAGMA table_info(turn_checkpoints)` instead of matching `duplicate column name`, and `turn-persistence.md` documents legacy empty `scope_key` semantics/backfill requirement.
5. Loop checkpoint tests now assert loop mappings do not land in `turn_checkpoints` by state ref, instead of assuming `turn_checkpoints` is globally empty.
6. Minor timestamp note addressed by binding `chrono::DateTime<Utc>` directly for Postgres `TIMESTAMPTZ` writes.

Verification:
- `cargo fmt --check`
- `cargo test -p ironclaw_turns --all-features`
- `cargo clippy -p ironclaw_turns --all-features --tests --examples --benches -- -D warnings`
- GitHub checks are green; PR merge state is clean.

zmanian (approved):
# Re-review: approved

All five blocking findings resolved at `a37f794d`:

1. **`.unwrap()` in production — RESOLVED.** `store.rs:113-120` replaces both sites with serde defaults: `default_checkpoint_kind()` returns `LoopCheckpointKind::BeforeBlock` and `default_checkpoint_state_ref()` returns a new infallible `LoopCheckpointStateRef::legacy_unknown()`. `state_ref` is now a real plumbed field on `BlockRunRequest` / `TurnRunnerOutcome::Blocked` / `record_checkpoint` rather than a sentinel.
2. **Non-transactional INSERT+SELECT race — RESOLVED.** Both backends do a single atomic UPSERT with a guard. libSQL: `INSERT ... ON CONFLICT(checkpoint_id) DO UPDATE SET checkpoint_id = ... WHERE payload = excluded.payload` returning row count (`rows == 1` ⇒ success, `0` ⇒ `Conflict`). Postgres mirrors with `RETURNING payload::text` inside `client.transaction()`.
3. **JSON re-deserialization drift — RESOLVED.** `ensure_loop_checkpoint_record_matches_request` compares persisted `scope/turn_id/run_id/checkpoint_id` against the request and raises `TurnError::Conflict` on drift — silent miss replaced by loud error.
4. **libSQL migration robustness — RESOLVED.** `libsql_column_exists` via `PRAGMA table_info` with identifier whitelist replaces the string match; Postgres uses `ADD COLUMN IF NOT EXISTS`. Legacy `NOT NULL DEFAULT ''` rows documented in `docs/reborn/contracts/turn-persistence.md` with explicit future-read-path requirements.
5. **Brittle `is_empty()` tests — RESOLVED.** New contract tests assert exact loop-checkpoint cardinality + absence of state_ref in `turn_checkpoints`, plus cross-scope/cross-run miss and drift conflict for both backends.

Substantive fixes, not annotations — particularly the `state_ref` plumbing in #1.

Issue #3473, opened 2026-05-11
[Reborn] Reborn Lane 9: Wire SkillContextService to production skill sources and loop prompt


serrrfirat:
## Parent / follow-up

Follow-up to #3432 and PR #3470.

PR #3470 adds the pure `SkillContextService` / `SkillContextSource` contract and fixture-backed tests for deterministic, trust-aware skill context selection.

PR #3823 wires the existing `HostSkillContextSource` port through `RebornRuntimeInput` into planned runtime composition, so callers can now supply a runtime skill-context source.

Those pieces are necessary, but they do not make Reborn skill context production-complete: no first-party skills extension/source is exposed, no real skill bundles are parsed/loaded through Reborn production storage, no host-approved trust/visibility policy is applied to runtime candidates, and selected snippets still need production source/lifecycle wiring.

## Problem

Reborn needs model-visible skill context to come from host-approved skill state without bypassing trust, capability visibility, or host authority boundaries.

The loop/service core should stay snapshot-based and dependency-light. Skills should not become kernel behavior. Reborn core should own ports, trust/visibility enforcement, capability leases, and model-context injection; a first-party in-process skills extension should own skill storage, catalog/lifecycle, install/update/delete, learned-skill writes, and source enumeration through narrow handles.

## Lane: first-party skills extension over ScopedFilesystem

Treat skills as a first-party in-process extension, not as kernel state.

Principles:

- Canonical portable format is a `SKILL.md` bundle, compatible with Hermes/agent-skill conventions.
- Use `ironclaw_skills` for parsing, validation, manifest/trust mapping, and any reusable selection helpers.
- Extension storage access goes through `ironclaw_filesystem::ScopedFilesystem`, never raw host paths, direct DB handles, or `ironclaw_memory` internals.
- `/system/skills` is read-only for bundled/system skills.
- Add or use a user-scoped `/skills` mount for user/agent/project-learned skills, mapped through the same tenant/user scoped filesystem fabric as other Reborn stores.
- Optional later source: `/tenant-shared/skills` for tenant-shared skills if product policy wants it.
- Skills are passive instructions plus supporting files. They do not grant authority.
- Skill-declared needs are capability requests only; Reborn capability policy decides what is actually leased for a run.
- Runtime injection goes through `HostSkillContextSource -> HostSkillContextCandidate -> SkillContextService -> model snippets`.
[...]

PR #3470, opened 2026-05-11
GitHub issue #3432: [Reborn] Add deterministic trust-aware SkillContextService

serrrfirat:
Refs #3432

KB task: KB-007

GitHub issue #3432: [Reborn] Add deterministic trust-aware SkillContextService

URL: https://github.com/nearai/ironclaw/issues/3432
Repo: nearai/ironclaw
Labels: reborn
Assignees: nickpismenkov

Scope: IronClaw Reborn issue imported from GitHub.
Before coding: read full issue body/comments, check linked PRs/blockers, work from current origin/reborn-integration unless issue says otherwise, keep PR tightly scoped.
Acceptance: satisfy GitHub issue acceptance criteria; include tests/verification evidence; never merge without explicit user approval.

Auto-opened by kb when task reached In Review. Auto-merge remains disabled; do not merge without explicit operator approval.

PR #3476, opened 2026-05-11
[Reborn] Wire SkillContextService into loop prompt path

serrrfirat:
Refs #3473
Stacked on #3470 (`kb/kb-007`).

## Summary
- Adds host-owned `HostSkillContextSource` and parser-backed snapshot builder in `ironclaw_loop_support`.
- Reuses `ironclaw_skills::parse_skill_md` and `SkillTrust` to build `SkillRunSnapshot` for `SkillContextService`.
- Wires selected skill instruction snippets into text-only prompt bundles and resolves them through the model port as system messages.
- Binds snippet refs to ordinal + safe-summary hash so duplicate skill names stay distinct and source drift after prompt build fails closed.
- Skips hidden/denied candidates before parsing raw SKILL.md, so invisible malformed skills cannot break prompt construction.
- Adds `RebornLoopDriverHostFactory::with_skill_context_source(...)` so production composition can inject the real skill source without moving parsing into `ironclaw_turns`.

## Tests
- `CARGO_TARGET_DIR=/tmp/ironclaw-target-3473 cargo fmt --all -- --check`
- `CARGO_TARGET_DIR=/tmp/ironclaw-target-3473 cargo test -p ironclaw_loop_support --quiet`
- `CARGO_TARGET_DIR=/tmp/ironclaw-target-3473 cargo test -p ironclaw_turns --quiet`
- `CARGO_TARGET_DIR=/tmp/ironclaw-target-3473 cargo test -p ironclaw_reborn --quiet`
- `CARGO_TARGET_DIR=/tmp/ironclaw-target-3473 cargo test -p ironclaw_architecture --quiet`
- `CARGO_TARGET_DIR=/tmp/ironclaw-target-3473 cargo clippy -p ironclaw_turns -p ironclaw_loop_support -p ironclaw_reborn --all-targets -- -D warnings`

## Notes
- This keeps `ironclaw_turns` as snapshot/prompt contract layer.
- Production still must choose/provide the concrete `HostSkillContextSource` implementation for the active Reborn skill store.


zmanian:
# Review

**Summary:** Wires `SkillContextService` into the Reborn text-only loop prompt path via a new host-owned `HostSkillContextSource` trait, converting SKILL.md files into ordered, hash-bound system-message snippets.

**Findings:**

1. **Tool attenuation gap — blocking.** `SkillTrust::Installed` vs `Trusted` is plumbed into `InstalledSkillSnapshot` (`skill_context.rs:194-208`) and the project rule is that Installed skills must run with attenuated tool access. But this PR only feeds skills' *prompt content* into the model — there is no corresponding `ToolDispatcher` attenuation. Per CLAUDE.md "Everything Goes Through Tools," gating must happen at dispatch. Either confirm a follow-up wires `SkillTrustLevel` into the capability port / dispatch, or this PR lands a prompt-only feature where an Installed skill can still request any capability the loop exposes.

2. **SKILL.md `prompt_content` passes through unsanitized for Installed skills.** `parsed_skill_to_snapshot_entry` (`skill_context.rs:198-208`) renders raw author-controlled markdown into the system message (test at line 1300 confirms `"Use alpha prompt content."` lands verbatim). For Trusted skills this is acceptable; for Installed/registry-sourced skills there is no fence-stripping or delimiter framing to prevent fake "user:" turns. Worth confirming whether `SkillContextService` sanitizes internally — not visible from this PR.

3. **Ordering: skill snippets precede identity messages.** `prompt.rs:170-189` prepends skill instructions *before* `context.messages`. The system prompt / identity files (AGENTS.md, SOUL.md per CLAUDE.md) arrive via `context.messages` — skill snippets thus precede the agent's own identity, inverting typical precedence (identity should dominate skill instructions). No test asserts ordering against an identity baseline.

4. **Snippet-ref derivation duplicated verbatim** across `loop_support/skill_context.rs:212-255` and `turns/run_profile/prompt.rs:213-266` — same FNV constants, `sanitize_ref_suffix`, `stable_snippet_ref_hash`. The two must stay in lockstep for the snippet-ref binding check (`lib.rs:644`); drift silently breaks the "fail closed on source change" guarantee. Extract to a single location.

5. **Telemetry coverage missing.** `prompt_bundle_built` milestone records no skill metadata. Without active-skill names + trust levels + ordinals, post-hoc debugging of "why did the model behave like skill X was active" is impossible. Add to the milestone.

**Minor:** FNV-1a is fine for dedup but a 64-bit non-cryptographic hash means a malicious skill author who controls `safe_summary` could craft a collision against another skill's ref. Low impact since the role check (`lib.rs:646`) also gates.

**Verdict: Request changes** — (1) and (4) are blockers; (2) and (3) need explicit answers.

zmanian:
# Post-merge re-review

3 of 5 prior findings resolved cleanly, 1 deferred to #3540, but **1 blocker still open** and **1 new gap** introduced:

## Resolved
- ✅ **Installed `prompt_content` sanitization** — `parsed_skill_to_snapshot_entry` (`crates/ironclaw_loop_support/src/skill_context.rs:393-396`) now drops `prompt_content` to `None` for `SkillTrustLevel::Installed`, with regression test `skill_snapshot_builder_drops_installed_prompt_content_before_snapshot_storage` (lines 569-591). Interim until #3540 lands the envelope.
- ✅ **Identity-before-skill ordering** — `LoopContextBundle.identity_messages: Vec<LoopContextMessage>` added (line 1360); `prompt.rs:170-189` extends in order identity → instruction snippets → messages. Test asserts the indices.
- ✅ **Snippet-ref derivation duplication** — FNV constants and helpers live only in `crates/ironclaw_turns/src/run_profile/skill_context.rs:1639-1740`; `loop_support` re-exports. PR #3507 further centralizes display hashing (separate refinement).
- ✅ **`prompt_bundle_built` skill metadata** — milestone gained `skill_context: Vec<PromptSkillContextMetadata>` with `{ordinal, source_name, trust_level}`; populated from `LoopContextSnippet.metadata` (lines 1557-1570); hard-fails (`AgentLoopHostErrorKind::Internal`) if a `skill:`-prefixed snippet arrives without metadata. Good fail-closed posture.

## Still open

- ❌ **Tool attenuation gap (original blocker #1).** `SkillTrustLevel::Installed` is plumbed through `InstalledSkillSnapshot.trust` and emitted as telemetry (`turns/run_profile/skill_context.rs:1651`), but never consulted by a `ToolDispatcher` or capability port. The code comment calls trust "used for telemetry and downstream attenuation checks" — implying attenuation is still follow-up. The new test `text_only_host_skill_context_does_not_expand_capability_surface` (lines 1256-1311) only asserts an *already-empty* surface stays empty; it does not exercise denial of a capability that would otherwise be visible. Per CLAUDE.md "Everything Goes Through Tools," dispatcher-side gating is the load-bearing enforcement and it doesn't exist yet.

- ❌ **`identity_messages` is a dangling contract (new gap).** The field exists in `LoopContextBundle` and the prompt path consumes it, but `ThreadBackedLoopContextPort::load_loop_context` (`crates/ironclaw_loop_support/src/lib.rs:98`) hardcodes `identity_messages: Vec::new()`. Identity files (AGENTS.md/SOUL.md/USER.md) still flow through `context.messages`. So in production today the "identity before skill" ordering only holds by accident — if no skill snippets exist, or if identity files happen to appear earlier in thread history. Tests pass because `RecordingAgentLoopHost` fakes the population.

Filing a follow-up issue for both items. Recommend not relying on the skill-context path in production traffic until both are closed.

Issue #3492, opened 2026-05-11
[Reborn] Establish trust-boundary hardening baseline


serrrfirat:
## Context

A cross-PR review of the current Reborn stack found the same structural failure mode across several slices: trusted components are introduced, but the policy/evidence/snapshot/content values they rely on are still freely constructable or weakly bound at public API seams. That leaves core Reborn guarantees enforced by convention instead of by type, construction path, or shared harness.

Examples called out in review:

- #3460: `LoopExitValidationPolicy` fields are public, and in-memory evidence inspection leaks outside test-only intent.
- #3470: trust-aware skill snapshot versioning uses forgeable/non-cryptographic FNV state.
- #3476: skill trust reaches prompt selection but not all dispatcher/tool-attenuation seams.
- #3471/#3476/#3470: retrieved memory/skill content is injected into prompts with only source-local sanitization and no uniform untrusted-content envelope.
- #3454/#3462/#3460/#3471: adapter errors collapse transient, permanent, misconfigured, and policy-denied cases.
- #3353/#3454/#3471/#3470: async entry points, caches, byte accumulation, or counters lack explicit back-pressure / checked limits.
- #3353/#3454: sandbox/host/native naming does not make trust boundaries obvious.
- #3446/#3460/#3468/#3454: schema/status/policy changes lack enforced downstream match-site audits.
- #3454/#3462: security/durability gates risk permissive `serde(default)` behavior.

## What to build

Create a Reborn trust-boundary hardening baseline that future Reborn PRs must use. This should convert repeated review comments into shared code primitives, documented invariants, and PR/check harness requirements.

## Acceptance criteria

- [ ] Add an ADR or design doc for Reborn trust-boundary invariants covering:
  - trusted minting only through witness types, sealed constructors, or `pub(crate)` construction paths
  - untrusted prompt-content envelope semantics
  - hash-purpose policy
  - cross-crate error classification
  - admission/back-pressure limits
  - sandbox/native/host naming expectations
  - enum/status/policy downstream audit requirements
  - `serde(default)` fail-closed convention for security/durability gates
- [ ] Introduce shared witness/newtype or sealed-constructor patterns for security-relevant Reborn policy/evidence/snapshot structs so untrusted callers cannot forge trusted state through public fields.
[...]

PR #3493, opened 2026-05-11
Fix Reborn memory error redaction and SQL replay cursors

serrrfirat:
## Summary
- sanitize public Reborn memory backend/provider filesystem errors so raw SQL/provider details do not leak
- keep embedding error category observable without exposing provider reason payloads
- make SQL event-store replay advance `next_cursor` past filtered tail records, matching JSONL semantics

## Tests
- `cargo fmt`
- `cargo test -p ironclaw_memory --features libsql -- --test-threads=1`
- `cargo test -p ironclaw_reborn_event_store --features libsql`
- `cargo check -p ironclaw_memory --features postgres`
- `cargo check -p ironclaw_reborn_event_store --features postgres`

Note: parallel `cargo test -p ironclaw_memory --features libsql` showed existing order-sensitive failures in schema/skip-versioning tests; serial run passed.

### DIFF
--- a/crates/ironclaw_memory/src/backend.rs
+++ b/crates/ironclaw_memory/src/backend.rs
@@ -627,6 +627,27 @@ mod tests {
     use crate::embedding::EmbeddingError;
     use crate::repo::InMemoryMemoryDocumentRepository;
 
+    struct FailingEmbeddingProvider;
+
+    #[async_trait]
+    impl EmbeddingProvider for FailingEmbeddingProvider {
+        fn dimension(&self) -> usize {
+            3
+        }
+
+        fn model_name(&self) -> &str {
+            "failing"
+        }
+
+        async fn embed(&self, _text: &str) -> Result<Vec<f32>, EmbeddingError> {
+            Err(EmbeddingError::ProviderUnavailable {
+                reason:
+                    "provider exploded at /tmp/HOST_PATH_SENTINEL with token RAW_TOKEN_SENTINEL"
+                        .to_string(),
+            })
+        }
+    }
+
     struct UnitEmbeddingProvider;
 
     #[async_trait]
@@ -808,6 +829,32 @@ mod tests {
         let request = MemorySearchRequest::new("query").unwrap();
         let err = backend.search(&alpha_context(), request).await.unwrap_err();
         assert!(err.to_string().contains("embedding generation"));
+    }
+
+    #[tokio::test]
+    async fn embedding_provider_error_is_sanitized_at_backend_boundary() {
+        let repo = Arc::new(InMemoryMemoryDocumentRepository::new());
+        let backend = RepositoryMemoryBackend::new(repo)
+            .without_prompt_write_safety_policy()
+            .with_embedding_provider(Arc::new(FailingEmbeddingProvider))
+            .with_capabilities(MemoryBackendCapabilities {
+                full_text_search: true,
+                vector_search: true,
+                embeddings: true,
+                ..MemoryBackendCapabilities::default()
+            });
+
+        let request = MemorySearchRequest::new("query").unwrap();
+        let err = backend.search(&alpha_context(), request).await.unwrap_err();
+        let displayed = err.to_string();
+
+        assert!(displayed.contains("embedding provider unavailable"));
+        assert!(
+            !displayed.contains("HOST_PATH_SENTINEL")
+                && !displayed.contains("RAW_TOKEN_SENTINEL")
+                && !displayed.contains("/tmp/"),
+            "public memory error leaked backend/provider details: {displayed}"
+        );
     }
 
     #[tokio::test]
--- a/crates/ironclaw_memory/src/embedding.rs
+++ b/crates/ironclaw_memory/src/embedding.rs
@@ -73,7 +73,16 @@ pub(crate) fn embedding_filesystem_error(
     operation: FilesystemOperation,
     error: EmbeddingError,
 ) -> FilesystemError {
-    memory_error(path, operation, error.to_string())
+    let reason = match error {
+        EmbeddingError::ProviderUnavailable { .. } => "embedding provider unavailable".to_string(),
+        EmbeddingError::InvalidVector { expected, actual } => {
+            format!("embedding vector dimension mismatch: expected {expected}, got {actual}")
+        }
+        EmbeddingError::TextTooLong { length, max } => {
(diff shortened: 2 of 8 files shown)


zmanian (changes requested):
# Review

**Summary:** Sanitizes public Reborn memory/provider/filesystem error reasons to prevent SQL/backend detail leaks; refines SQL event-store replay so `next_cursor` advances past filtered-out tail records to match JSONL semantics. Direction is right; the deny-list approach has portability gaps.

## Findings

1. **`sanitize_memory_backend_reason` is a substring deny-list with real gaps** (`crates/ironclaw_memory/src/path.rs:364-374`).
   - **No `"libsql"`** — that's the actual library this codebase uses (per CLAUDE.md and Cargo.toml). The check matches `"sqlite"` but error messages from the `libsql` crate often say `libsql` literally.
   - **No macOS temp paths** — the `/tmp/` check misses `/var/folders/...`, the actual `TMPDIR` on macOS. The new libSQL contract test (`reborn_native_libsql_repository_contract.rs:54-61`) only asserts "no leak"; on macOS the temp path won't contain `/tmp/`, so the test passes without actually exercising the sanitizer. Add `/var/folders/`, `/private/`, and `\AppData\` for portability.
   - **`"postgres"` as a bare substring** — a legitimate error like `"postgres backend disabled by config"` becomes opaque.
   - **Deny-list inverted** — the right structure here is the same one this PR uses in `embedding_filesystem_error` (`embedding.rs:73-85`): match the typed error variant and emit a known-safe category string. That works for `EmbeddingError` and should work for `FilesystemError` callers. Right now anything passed via `memory_error(path, op, "free string from backend")` is parsed by guessing.

2. **`reason.to_ascii_lowercase()` allocates on every error path even when no marker matches.** Gemini hinted at this. Make the deny-list a `&[&str]` constant and use `lower.as_str()`. Better: replace with the typed-variant approach in #1.

3. **Postgres cursor-advancement has no contract test** (`postgres_store.rs:480-501`). The libSQL test (`durable_event_store_contract.rs:71-122`) exercises the new "advance to stream head when filtered tail" path; the identical change to `postgres_store.rs` ships untested. Add a `#[cfg(feature = "postgres")]` parity test before merge.

4. **`entries.len() < limit ⇒ advance to stream_head` correctness depends on the SQL query shape** (`libsql_store.rs:589-602`, `postgres_store.rs:483-496`). It assumes the query is `WHERE cursor > ? AND <filter> ORDER BY cursor LIMIT ?` with the filter pushed into the DB. A future query change to apply `LIMIT` *before* filter would silently break this invariant. The current comment implies but doesn't state this — worth an explicit one-liner pinning the assumption.

5. **Variable shadowing — `let stream_head = next_cursor; let next_cursor = ...`** is mildly confusing because the outer `next_cursor` is the streams-table head value while the inner one is the computed return. Renaming the outer binding to `stream_head_cursor` would read cleaner.

6. **Connection-pool errors aren't on the deny-list** — e.g., "connection refused", "timeout", `host:port` strings in some drivers. Those would reach the public reason verbatim. Adds weight to #1's case for allow-list / typed mapping.

7. **`"DROP TABLE"` and `"SQL"` are asserted in the test but not in the sanitizer** (`reborn_native_libsql_repository_contract.rs:55-59`). The test passes today because the error happens to contain `"no such table"` which is on the deny-list — not because `"DROP TABLE"` or `"SQL"` are filtered. A SQL-shaped error without `"no such table"` would slip through. Either add those terms to the deny-list or remove from the assertion to avoid false confidence.

## Verdict

Request changes. At minimum: add libsql / `/var/folders/` / `/private/` coverage, the Postgres cursor-advancement parity test, and decide whether to migrate `memory_error` callers to the variant-match pattern that `embedding_filesystem_error` already demonstrates in this same diff.

This work also lines up with acceptance criterion 5 (error classification) on #3492 — would be worth landing this PR's pattern as one of the baseline primitives once the deny-list is replaced with typed mapping.

serrrfirat:
Addressed zmanian review in `bc297569`:

- Replaced inline substring checks with `MEMORY_BACKEND_DETAIL_MARKERS` constant and `lower.as_str()` marker scan.
- Added missing backend/path markers: `libsql`, `/var/folders/`, `/private/`, `\AppData\`, `drop table`, `sql`, plus connection/host/port/timeout markers.
- Avoided broad bare `postgres`; sanitizer now targets `postgres error` to preserve legitimate config/category messages.
- Added Postgres parity contract: `postgres_replay_advances_next_cursor_past_trailing_filtered_records` (skips when `IRONCLAW_REBORN_EVENT_STORE_POSTGRES_URL` is unset).
- Documented SQL replay assumption: filter predicates are applied before `LIMIT`, so short filtered pages can advance to stream head.
- Renamed cursor variable to `stream_head_cursor` for clarity.

Verification run:
- `cargo fmt`
- `cargo test -p ironclaw_memory --features libsql -- --test-threads=1`
- `cargo test -p ironclaw_reborn_event_store --features libsql`
- `cargo check -p ironclaw_memory --features postgres`
- `cargo check -p ironclaw_reborn_event_store --features postgres`
- `cargo test -p ironclaw_reborn_event_store --features postgres --test durable_event_store_contract postgres_replay_advances_next_cursor_past_trailing_filtered_records -- --nocapture`
- `git diff --check`

zmanian (approved):
# Post-merge approval

Confirming all review points landed at `bc297569`. Verified against the merged code at `origin/reborn-integration:crates/ironclaw_memory/src/path.rs`:

- `MEMORY_BACKEND_DETAIL_MARKERS` constant array with single `lower.as_str()` scan replaces inline substring checks.
- All flagged markers added: `libsql`, `/var/folders/`, `/private/`, `\appdata\`, `drop table`, `sql`, `connection refused`, `timeout`, `host=`, `port=`.
- `postgres` bare-substring concern addressed by narrowing to `postgres error`.
- Postgres parity test added (`postgres_replay_advances_next_cursor_past_trailing_filtered_records`).
- SQL replay assumption documented.
- `stream_head_cursor` rename applied.

Apologies for the lag — my CHANGES_REQUESTED was already addressed before merge; I missed your response comment. Flipping to approve.

Remaining future thought (not blocking, not for this PR): the deny-list is still a deny-list. The typed-variant pattern in `embedding_filesystem_error` is cleaner and would be a worthwhile refactor whenever the `memory_error` call sites get a touchup — but the current shape is materially safer than before.

Issue #3429, opened 2026-05-09
[Reborn] Add loop production readiness validation


serrrfirat:
Parent: #3107  
Related: #3026, #3101, #3090, #3402, #3404, #3406, #3407, #3409, #3410, #3420, #3423, #3424  
Reborn epic: #2987

## Purpose

Add Reborn-level loop production readiness/config validation without expanding `ironclaw_host_runtime::ProductionWiringReport` into an upper-loop composition bag.

This issue validates Reborn loop orchestration services and selected run profiles. Host-runtime substrate readiness remains owned by `ironclaw_host_runtime`.

## Boundary decision

Keep these layers separate:

```text
ironclaw_host_runtime::ProductionWiringReport
  -> low-level substrate/runtime graph readiness
  -> filesystem, event/audit, run-state, approvals, capability leases, turn-state, wake notifier, secrets, runtimes

ironclaw_reborn::RebornLoopProductionReport (new)
  -> upper Reborn loop orchestration readiness
  -> driver registry, selected run profiles, host factories, prompt/model/capability/checkpoint/input/applier support
```

Do not push driver registry, prompt bundles, runner workers, v2 adapters, or `LoopExitApplier` validation into `ironclaw_host_runtime`.

### Selected run profile set

Production config selects named run profiles/profile-set IDs, not Rust crate/type/factory names.

For every selected production run profile, readiness must verify:

- exact registered driver identity;
- compatible checkpoint schema identity;
- supported host factory;
- prompt port support;
- model port/gateway support;
- capability mode support;
- checkpoint state store;
- input/resume/cancel semantics support;
- `LoopExitApplier` support;
- production-safe backing stores;
- supported run profile version.

Missing required selected profile components hard-fail production build/startup/readiness before traffic.

### Optional profiles

Non-selected optional profiles must not block startup. They should appear as unavailable in readiness.

### Text-only first production profile

A text-only Reborn loop profile may be the first production-ready profile.

Tool-capable/v2 profiles must remain unavailable/fail readiness until prerequisite work lands, including:

- #3090 ToolSurfaceService / CapabilityCatalog;
- capability input/surface snapshot staging;
- #3420 Reborn-native capability/effect adapter;
- #3424 trusted `LoopExitApplier`;
- required durable stores.

### Capability port for text-only profiles

Even text-only production host should provide a `LoopCapabilityPort` implementation. It must explicitly deny capability calls.

Do not model text-only hosts as missing capability port in production.

### Local/test implementations

[...]

PR #3503, opened 2026-05-11
feat(reborn): add loop production readiness gate

serrrfirat:
Refs #3429

## Summary

- adds `ironclaw_reborn::production_readiness` with `RebornLoopProductionReport`
- validates selected profile/driver/checkpoint identities, component safety classes, optional profiles, and active-run drain protection
- separates production safety from locality: production rejects test-only/non-durable/noop/unverified seams, but does not reject durable local implementations just because they are local
- keeps host-runtime `ProductionWiringReport` separate and substrate-scoped

## Notes

This is the pure readiness/reporting slice. Startup composition can call this report once the production composition root is wired.

## Tests

- `CARGO_TARGET_DIR=/tmp/ironclaw-target-3429 cargo test -p ironclaw_reborn production_readiness -- --nocapture`
- `CARGO_TARGET_DIR=/tmp/ironclaw-target-3429 cargo test -p ironclaw_reborn`
- `CARGO_TARGET_DIR=/tmp/ironclaw-target-3429 cargo clippy -p ironclaw_reborn --all-targets -- -D warnings`

### DIFF
--- a/crates/ironclaw_reborn/src/driver_registry.rs
+++ b/crates/ironclaw_reborn/src/driver_registry.rs
@@ -5,7 +5,7 @@
 //! descriptor metadata at startup, and validates that configured and persisted
 //! run identities can be served before traffic is accepted.
 
-use std::{collections::HashMap, error::Error, fmt, sync::Arc};
+use std::{borrow::Borrow, collections::HashMap, error::Error, fmt, sync::Arc};
 
 use ironclaw_turns::{
     AgentLoopDriver, AgentLoopDriverDescriptor, RunProfileVersion, TurnStatus,
@@ -207,19 +207,41 @@ impl DriverRegistry {
         mode: DriverReadinessMode,
         inputs: DriverReadinessInputs,
     ) -> DriverReadinessReport {
+        self.validate_readiness_from_iter(
+            mode,
+            inputs.host_graph,
+            inputs.configured_profiles,
+            inputs.persisted_runs,
+        )
+    }
+
+    pub fn validate_readiness_from_iter<C, P>(
+        &self,
+        mode: DriverReadinessMode,
+        host_graph: HostGraphReadiness,
+        configured_profiles: C,
+        persisted_runs: P,
+    ) -> DriverReadinessReport
+    where
+        C: IntoIterator,
+        C::Item: Borrow<ConfiguredRunProfile>,
+        P: IntoIterator,
+        P::Item: Borrow<PersistedRunDriverIdentity>,
+    {
         let mut diagnostics = Vec::new();
         let mut has_reference_driver = false;
 
-        for profile in inputs
-            .configured_profiles
-            .iter()
-            .filter(|profile| profile.enabled)
-        {
+        for profile in configured_profiles {
+            let profile = profile.borrow();
+            if !profile.enabled {
+                continue;
+            }
+
             match self.get(&profile.driver_identity) {
                 Some(entry) => {
                     validate_entry_readiness(
                         mode,
-                        &inputs.host_graph,
+                        &host_graph,
                         &profile.name,
                         &profile.driver_identity,
                         entry,
@@ -228,7 +250,7 @@ impl DriverRegistry {
                     );
                 }
                 None => diagnostics.push(DriverReadinessDiagnostic::new(
-                    "missing_configured_driver",
+                    DriverReadinessDiagnosticCode::MissingConfiguredDriver,
                     profile.name.clone(),
                     Some(profile.driver_identity.clone()),
                     "enabled run profile references an unregistered loop driver identity",
@@ -236,16 +258,17 @@ impl DriverRegistry {
             }
         }
 
-        for persisted_run in inputs
-            .persisted_runs
-            .iter()
-            .filter(|run| !run.status.is_terminal())
-        {
+        for persisted_run in persisted_runs {
+            let persisted_run = persisted_run.borrow();
+            if persisted_run.status.is_terminal() {
+                continue;
+            }
+
             match self.get(&persisted_run.driver_identity) {
                 Some(entry) => {
                     validate_entry_readiness(
                         mode,
-                        &inputs.host_graph,
+                        &host_graph,
                         &persisted_run.run_id,
                         &persisted_run.driver_identity,
                         entry,
@@ -254,7 +277,7 @@ impl DriverRegistry {
                     );
                 }
                 None => diagnostics.push(DriverReadinessDiagnostic::new(
-                    "missing_non_terminal_run_driver",
+                    DriverReadinessDiagnosticCode::MissingNonTerminalRunDriver,
                     persisted_run.run_id.clone(),
                     Some(persisted_run.driver_identity.clone()),
                     "non-terminal persisted run requires an unregistered loop driver identity",
@@ -384,9 +407,18 @@ pub struct DriverReadinessReport {
     pub diagnostics: Vec<DriverReadinessDiagnostic>,
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum DriverReadinessDiagnosticCode {
+    MissingConfiguredDriver,
+    MissingNonTerminalRunDriver,
+    ReferenceDriverNotProductionReady,
+    ReferenceDriverAllowedForLocalDev,
+    MissingRequiredDriverRequirement,
+}
+
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct DriverReadinessDiagnostic {
-    pub code: &'static str,
+    pub code: DriverReadinessDiagnosticCode,
     pub subject: String,
     pub driver_identity: Option<LoopDriverRegistryKey>,
     pub message: String,
@@ -395,7 +427,7 @@ pub struct DriverReadinessDiagnostic {
 
 impl DriverReadinessDiagnostic {
     fn new(
-        code: &'static str,
+        code: DriverReadinessDiagnosticCode,
         subject: String,
         driver_identity: Option<LoopDriverRegistryKey>,
         message: &'static str,
@@ -410,7 +442,7 @@ impl DriverReadinessDiagnostic {
     }
 
     fn non_blocking(
-        code: &'static str,
+        code: DriverReadinessDiagnosticCode,
         subject: String,
         driver_identity: Option<LoopDriverRegistryKey>,
         message: &'static str,
@@ -456,7 +488,7 @@ fn validate_entry_readiness(
     match (mode, entry.kind()) {
         (DriverReadinessMode::Production, DriverKind::Reference) => {
             diagnostics.push(DriverReadinessDiagnostic::new(
-                "reference_driver_not_production_ready",
+                DriverReadinessDiagnosticCode::ReferenceDriverNotProductionReady,
                 subject.to_string(),
                 Some(driver_identity.clone()),
                 "fake/reference loop driver cannot satisfy production readiness",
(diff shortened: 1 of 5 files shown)


henrypark133 (changes requested):
Thanks for adding a pure production-readiness report for the Reborn loop graph. The separation from host-runtime substrate readiness makes sense, but I found one active-run drain gap that should be fixed before relying on this gate.

Finding:

Medium - `crates/ironclaw_reborn/src/production_readiness.rs:537`

Active-run drain protection checks only `profile_id` + `profile_version`, not the active run’s persisted `driver_identity`.

Failure mode: production can report ready after reusing the same profile version for a different loop driver while a non-terminal run still references the old driver, as long as the old driver remains registered.

Why it matters: a versioned run profile includes the loop driver/checkpoint identity. Production readiness should force either retaining the exact active-run profile identity or bumping/draining the profile before startup.

Expected fix direction: require an active run to match a configured profile by `profile_id`, `profile_version`, and `driver_identity` (including checkpoint schema identity), or emit `ActiveRunsRequireVersion` / a more specific identity mismatch issue.

What looked good:

- The slice is pure reporting and does not introduce a `CapabilityHost` bypass or runtime side effect.
- Driver registry readiness mapping and required host-graph checks otherwise line up with the existing `DriverRegistry` contract.
- `cargo test -p ironclaw_reborn production_readiness` passed in the specialist review.

Suggested coverage:

- Add a regression where an active non-terminal run keeps the same `profile_id`/`profile_version` but has a different `driver_identity` than the configured profile; expected result should be `NotReady`.

henrypark133 (commented):
## Review: active-run drain fixed, hold for rebase

Thanks for the follow-up. The previous active-run drain finding is addressed: the production-readiness check now requires an active non-terminal run to match a selected profile by `profile_id`, `profile_version`, and exact `driver_identity`, and the new regression covers the same-profile/version-but-different-driver case.

What looks good:
- The readiness gate remains pure reporting and does not introduce a runtime side effect or authority bypass.
- The driver identity drain check now lines up with the persisted run-profile identity contract.
- Local validation passed with `cargo test -p ironclaw_reborn`, `cargo fmt --check`, `cargo clippy -p ironclaw_reborn --tests -- -D warnings`, and `git diff --check origin/reborn-integration...origin/pr/3503`.

Low-priority notes:
- GitHub still reports the PR as conflicting. A local merge check shows the remaining conflict is in `crates/ironclaw_reborn/src/lib.rs`, where the base branch added `milestone_events` / `text_loop_driver` while this PR adds `production_readiness`.
- The only visible GitHub checks are `classify` and `scope`, so I would hold approval until the branch is rebased and meaningful CI runs on the mergeable head.

serrrfirat:
Addressed the review follow-up and rebase hold:

- Rebased `reborn/issue-3429-loop-production-readiness` onto current `reborn-integration`.
- Resolved the merge conflict in `crates/ironclaw_reborn/src/lib.rs` by preserving the base branch additions (`milestone_events`, `text_loop_driver`) and this PR's `production_readiness` module.
- Verified the active-run drain fix remains in place: non-terminal runs must match selected profiles by `profile_id`, `profile_version`, and exact `driver_identity` including checkpoint schema identity.
- Resolved the stale/outdated review threads that were already addressed by the follow-up commits.

Verification:

- `cargo fmt --check`
- `cargo test -p ironclaw_reborn production_readiness`
- `cargo clippy -p ironclaw_reborn --tests -- -D warnings`
- GitHub checks are green; PR is now mergeable.

Note: local `cargo test --lib` still has 3 unrelated existing failures in `extensions::manager` local-source install tests; no files from this PR touch that area.

zmanian:
## Summary

Pure readiness/reporting slice that classifies the Reborn loop graph (selected profiles, drivers, host ports, component safety classes, active-run drain protection) as `ProductionReady`, `LocalDevDegraded`, or `NotReady`. No runtime gating wired yet per the PR notes — startup composition will call this once the production composition root lands. Review focuses on the validator's semantics, failure-path coverage, and observability surface.

## Strengths

- **Fail-closed by default for `Production` mode.** `RebornComponentSafetyClass::blocks_production` rejects everything except `ProductionVerified`, and each failing component lands a `blocks_ready: true` issue. Test/non-durable/noop/unverified all map to distinct `RebornLoopProductionIssueKind` variants — actionable diagnostics rather than a single opaque "not ready".
- **Locality decoupled from safety.** The doc comment and `local_durable_components_are_not_rejected_by_name_or_locality` test make explicit that a standalone-local production deployment is valid. This is the right architectural call and matches the PR description.
- **Pure synchronous validator.** No `await`, no locks, no I/O — zero deadlock surface, safe to call from any startup path. Two-pass iteration over `active_runs` (`push_active_run_profile_issues` then `push_driver_readiness_issues`) is mildly redundant but harmless.
- **Subject redaction discipline.** `RebornActiveRunIdentity.run_ref` (potentially containing secrets) never reaches `issue.subject`; the static `"active_run"` constant is used instead. The `readiness_surface_redacts_active_run_subject` test pins this behavior — good defensive choice.
- **Type-safe diagnostic codes.** Migrating `DriverReadinessDiagnostic::code` from `&'static str` to `DriverReadinessDiagnosticCode` enum eliminates a class of stringly-typed downstream bugs and makes `push_mapped_driver_issues` exhaustive.
- **`validate_readiness_from_iter` refactor.** Generic over `IntoIterator<Item: Borrow<…>>` enables zero-copy reuse from `push_driver_readiness_issues` / `push_optional_profile_issues` without forcing callers to materialize `Vec`s. Backward-compatible with the existing `DriverReadinessInputs` entry point.
- **Failure-path test coverage is solid.** Missing selected driver, reference driver in production, non-durable checkpoint store, noop wake notifier, missing capability port, checkpoint schema version mismatch, active-run profile-version mismatch, active-run driver-identity mismatch, active-run checkpoint-schema mismatch, optional profile unavailable not blocking, redaction safety — all covered.

## Issues

**1. Observability gap: driver-readiness issues drop profile context.**
In `push_mapped_driver_issues`, issues derived from `DriverReadinessReport` set `profile_id: None, profile_version: None` even though the input loop iterates `inputs.configured_profiles` (selected or optional) and knows the originating profile. Operators reading the report see a `LoopDriver`/`Missing` issue with a redacted subject and no way to correlate it back to which profile triggered it without cross-referencing `driver_identity`. Recommend threading the profile reference through or attaching `(profile_id, profile_version)` when the diagnostic was emitted on behalf of a configured profile.

**2. `LocalDevDegraded` status hides warnings in `Production` mode.**
The status computation:
```rust
} else if inputs.mode == RebornLoopReadinessMode::LocalDevTest
    && issues.iter().any(|issue| !issue.blocks_ready)
{
    RebornLoopProductionStatus::LocalDevDegraded
}
```
In `Production` mode, non-blocking issues (e.g. an unavailable optional profile) leave `status == ProductionReady`. That is the intended behavior per the optional-profile test, but a consumer who only checks `is_ready()` will not learn that warnings exist. Suggest either (a) documenting on `RebornLoopProductionReport` that callers must inspect `issues` even when `is_ready()` returns true, or (b) adding a `has_warnings()` helper alongside `is_ready()` and `blocking_issues()`.

**3. Mapping collision: `MissingNonTerminalRunDriver` -> `ActiveRunsRequireVersion`.**
`push_mapped_driver_issues` maps the driver-registry diagnostic `MissingNonTerminalRunDriver` onto `(LoopDriver, ActiveRunsRequireVersion)`. But `ActiveRunsRequireVersion` is also emitted from `push_active_run_profile_issues` for a semantically different condition (configured profile version no longer matches the active run). A downstream consumer using `report.contains(LoopDriver, ActiveRunsRequireVersion)` cannot distinguish "driver unregistered" from "profile version drift". Consider a dedicated `ActiveRunDriverUnregistered` kind so the two failure modes remain separable in metrics/alerts.

**4. Dead branch in optional-profile mapping.**
`push_optional_profile_issues` always passes `std::iter::empty::<PersistedRunDriverIdentity>()`, so the `DriverReadinessDiagnosticCode::MissingNonTerminalRunDriver` arm in `push_mapped_driver_issues` is unreachable for that call path. Not a defect, but a `debug_assert!` or comment clarifying invariants would help future readers.

**5. No caller exists yet.**
The PR notes call this out explicitly ("Startup composition can call this report once the production composition root is wired") and that is fine for a reporting slice. Worth a tracking issue or follow-up PR reference in the module doc so this does not silently rot as dead code. The whole `Production` classification has no runtime effect until something gates on `report.is_ready()`.

## Suggestions

- Add a module-level doc example showing the intended call pattern from startup composition once wired, so the eventual integrator does not need to reverse-engineer input shapes from tests.
- Consider deriving `Hash` on `RebornLoopProductionComponent` / `…IssueKind` so consumers can group issues into a `HashMap<(component, kind), Vec<issue>>` without a manual key wrapper. Cheap and forward-compatible.
- The `text_only_driver_requirements()` helper deserves a sibling `tool_capable_driver_requirements()` or similar to make the "deny capability port vs real capability port" distinction explicit at the requirement-construction site, mirroring how the doc comment frames it.
- Minor: `push_active_run_profile_issues` and `push_driver_readiness_issues` both walk `active_runs.iter().filter(!terminal)`. Worth precomputing the filtered slice once if this is ever called on the hot path; currently it is startup-only so the cost is irrelevant.
- The `with_profile` / `with_active_run` builder methods are used inside `push_*` helpers but not inside `push_mapped_driver_issues` — see Issue #1. Consistency would help.

## Verdict

REQUEST CHANGES (minor). The core logic, safety semantics, redaction discipline, and test coverage are strong. The main asks are observability fixes (Issue #1 and #3) so the report stays operator-friendly once it is actually wired into startup, plus a `has_warnings()` / docstring tightening (Issue #2). Issue #5 is informational — track the integration PR. No security or deadlock concerns; the validator is pure and side-effect free.

serrrfirat:
Addressed zmanian’s review feedback in `4259f215b`:

- Threaded configured profile / active-run context onto mapped driver-readiness issues.
- Added `RebornLoopProductionReport::has_warnings()` so production-ready reports can still expose non-blocking warnings.
- Split missing active-run driver diagnostics into `ActiveRunDriverUnregistered` instead of reusing `ActiveRunsRequireVersion`.
- Added module docs for the intended startup-composition call pattern / follow-up runtime gate.
- Added `Hash` derives and a `tool_capable_driver_requirements()` helper.
- Added regression coverage for selected/optional profile context and active-run missing driver diagnostics.

Verification:
- `cargo fmt --check`
- `git diff --check`
- `cargo test -p ironclaw_reborn --test production_readiness`
- `CARGO_TARGET_DIR=/Volumes/NVME/ironclaw-target cargo clippy -p ironclaw_reborn --tests --all-features -- -D warnings`
- PR checks are green.

Merge state checked after push: `CLEAN` / `MERGEABLE`, so no conflicts to resolve.

zmanian:
Filed follow-up issue [#3602](https://github.com/nearai/ironclaw/issues/3602) to track wiring `RebornLoopProductionReport::is_ready()` into startup composition. This PR introduced the validator as reporting-only; the gate needs to actually fail-closed startup in `Production` mode before any production deploys rely on it.

henrypark133 (approved):
## Re-review: active-run drain fixed

What looks good:
- The previous active-run drain finding is fixed: non-terminal runs now require the selected profile to match `profile_id`, `profile_version`, and exact `driver_identity`, including checkpoint schema identity.
- The readiness gate remains pure reporting and does not introduce runtime side effects or a capability/authority bypass.
- The local-dev versus production distinction is still clean: local durable implementations are not rejected by name, while fake/no-op/non-durable/unverified components fail production readiness.

No verified findings.

Low-priority note:
- The production-readiness report is not wired into startup composition in this slice, which matches the PR scope. The follow-up wiring should add caller-level coverage for the real production composition root.

Verification:
- `cargo test -p ironclaw_reborn production_readiness -- --nocapture`

Approved.

PR #3511, opened 2026-05-12
fix(prompt): avoid repeated compact tool schema lookups

### LINKED ISSUE #3465: ENGINE_V2 repeatedly calls `tool_info` during tool-use flows
## Problem

With `ENGINE_V2` enabled, the agent can repeatedly call `tool_info(schema)` during normal tool-use flows, especially around image generation/editing prompts.

Observed behavior:
- The Gateway UI shows many repeated `tool_info(schema)` tool calls.
- In some runs, the model also calls helper/echo-style tools repeatedly before eventually using the intended tool.
- This increases latency, token usage, and visual noise in the tool timeline.

## Impact

The user experience becomes confusing and slower. The model appears to be stuck in tool-discovery/planning mode instead of directly invoking the relevant tool.

## Scope

This appears to be an `ENGINE_V2` orchestration / prompting / tool exposure issue, not a Gateway UI duplication issue. The UI is showing repeated tool calls that actually happened.

## Expected Behavior

- `tool_info` should not be repeatedly invoked for the same schema within one turn unless there is a clear reason.
- The model should call the intended tool directly once it has enough schema information.
- Tool discovery should be cached, constrained, or hidden from normal model planning if appropriate.

## Possible Investigation Areas

- Whether `tool_info` is exposed too prominently as a callable tool.
- Whether the V2 prompt encourages the model to inspect schemas repeatedly.
- Whether tool schema/context is already available and should not require a runtime `tool_info` call.
- Whether repeated identical `tool_info` calls can be deduped or capped per turn.

## Notes

This is separate from the fixed issue where Gateway live SSE displayed the same tool event twice. This issue is about actual repeated `tool_info` invocations from `ENGINE_V2`.

hanakannzashi:
## Summary

- Updated ENGINE_V2 CodeAct compact-tool guidance so `tool_info(schema)` is used only when the schema is not already available.
- Added prompt coverage to prevent regressions back to “always call schema” wording.
- Keeps compact tool prompts lightweight without adding parameter lists into the main prompt.

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] CI/Infrastructure
- [ ] Security
- [ ] Dependencies

## Linked Issue

Fixes #3465 

## Validation

- [ ] `cargo build`
- [x] Relevant tests pass: `cargo test -p ironclaw_engine executor::prompt -- --nocapture`
- [x] Manual testing: started Gateway with ENGINE_V2 enabled and confirmed the local UI was reachable.

## Security Impact

None.

## Database Impact

None. No schema, migration, or persistence behavior changes.

## Blast Radius

Touches ENGINE_V2 CodeAct prompt construction for compact enabled tools. Regressions would affect how often the model asks for tool schemas via `tool_info`, but no tool execution, approval, DB, or API behavior is changed.

## Rollback Plan

Revert this PR to restore the previous compact-tool prompt wording.

## Review Follow-Through

Known follow-up: this PR intentionally does not remove `tool_info(schema)` usage entirely. It only clarifies that repeated schema lookups for the same tool in the same task should be avoided.

---

**Review track**: C


crates/ironclaw_engine/src/executor/prompt.rs
```diff
@@ -597,10 +601,11 @@ mod tests {
         assert!(prompt.contains("## Enabled Tools"));
         assert_eq!(prompt.matches("## Enabled Tools").count(), 1);
         assert!(prompt.contains(
```
serrrfirat: **Low Severity** — Test assertion can pass even if dynamic compact-tool instruction regresses.\n\nThis assertion searches the entire rendered prompt for `Avoid unnecessary repeated tool_info calls...`, but the same guidance now also appears in the static preamble. That means this test would still pass if the dynamic `## Enabled Tools` block lost the reuse instruction again.\n\nSuggestion: scope the assertion to the substring after `## Enabled Tools` (and before the next section) so the test proves the compact-tool section itself contains the schema-reuse guidance.

PR #3516, opened 2026-05-12
test(reborn): expand integration coverage

serrrfirat:
## Summary
- expand Reborn integration coverage for runner/host factory flows, checkpoint resume, approval block/resume mechanics, and tool-capable script capability execution
- add CLI/config edge coverage for `ironclaw-reborn` profile, completion, run, and doctor paths
- add event-store replay gap coverage for libSQL/Postgres filtered replay and public Postgres TLS gating
- fix model route validation for routes configured only on a different slot while allowing shared routes
- fix SQL replay contiguity checks so short filtered pages verify through stream head and missing trailing rows fail closed as `ReplayGap`

## Scope notes
- Gateway/product E2E is intentionally not added here because Reborn does not yet have an honest user-facing gateway entrypoint/wiring to exercise; adding one now would test legacy ENGINE_V2 or a fake seam.
- Restart coverage here includes turn-store snapshot/reopen plus rebuilt runner/host factory; full process/durable transcript restart remains follow-up.

## Verification
- `cargo test -p ironclaw_reborn --test loop_driver_host` → 56 passed
- `cargo test -p ironclaw_reborn -p ironclaw_reborn_cli -p ironclaw_reborn_config` → 156 passed
- `cargo test -p ironclaw_reborn_event_store --all-features` → 48 passed
- `git diff --check` → clean
- `cargo clippy -p ironclaw_reborn -p ironclaw_reborn_cli -p ironclaw_reborn_config -p ironclaw_reborn_event_store --all-targets --all-features -- -D warnings` → no issues

## Follow-ups
- full durable transcript/process restart coverage
- product-loop/gateway E2E once Reborn user-facing entrypoint exists


zmanian:
## Summary

PR adds substantive integration coverage for Reborn: runner/host factory drive a full text-only and script-capability transcript, approval-gate block/resume across `DefaultTurnCoordinator`, checkpoint resume, libSQL turn+thread service reopen, CLI run/doctor invalid-profile and v1-root collision paths, plus event-store filtered-replay gap detection on both libSQL and Postgres. Two production fixes ride along (model-route cross-slot rejection; SQL replay contiguity through stream head). Verification numbers (56/156/48 passing, clippy clean) align with the scope claimed.

## Strengths

- Tests exercise real components through the worker entrypoint (`TurnRunnerWorker::run` -> `TurnStateStore` -> `LoopExitApplier` -> driver -> `RebornLoopDriverHostFactory`), which is exactly the "test through the caller, not just the helper" pattern. Helper-only coverage of `validate_model_route` is paired with a real `StaticModelRouteResolver` cross-slot scenario.
- The approval-resume test goes through `DefaultTurnCoordinator::resume_turn` rather than poking the store directly, and asserts the same `run_id` and `checkpoint_id` survive across block/resume - this catches the class of bug where resume would silently start a new run.
- Fail-closed approval test (`turn_runner_rejects_driver_fabricated_approval_block_without_durable_gate_evidence`) is high value: it verifies the production evidence port (not the test stub) rejects a driver that fabricates a gate ref, asserts no model calls happen, and asserts no assistant message is finalized - three independent observable invariants.
- Event-store filtered-replay gap tests cover BOTH backends (libSQL + Postgres), matching the dual-backend persistence rule. The corruption fixture (delete trailing filtered cursor, reopen, expect `ReplayGap`) directly exercises the contiguity fix in `libsql_store.rs` / `postgres_store.rs`.
- `MockDriver` capturing `run_requests` / `resume_requests` lets the new checkpoint-resume unit test assert resume was called and `run` was not - the right shape for verifying recovery dispatch.
- Production fix in `model_routes.rs` correctly distinguishes "approved but configured exclusively elsewhere" from "approved and shared" - both branches are covered by the new `tests/model_routes.rs` cases.
- Restart fixture rebuilds `RebornLoopDriverHostFactory`, `LoopExitApplier`, `DriverRegistry`, and `TurnRunnerWorker` from scratch from persisted snapshots - this is a meaningful subset of process restart, not a same-process re-poll.

## Issues

### Important

1. **Flakiness budget on polling loops.** `wait_for_run_status` and the inline loops in the script-capability and missed-wake tests use a 2-second deadline with 20ms heartbeat / 10ms poll. On a contended CI runner this has bitten similar tests before. Consider: (a) raising the deadline to ~10s, OR (b) driving with `tokio::time::pause()` / `advance()` since the worker config exposes both intervals. The script-capability path is the biggest risk - real `HostRuntime` + `ScriptRuntime` + filesystem under load.

2. **`AlwaysVerifiedLoopExitEvidence` is a permissive stub used in a non-fail-closed test.** The approval-resume happy path swaps in this stub so the driver-fabricated gate_ref is accepted. That's fine for the happy path test, but the asymmetry with the fail-closed test should be called out in a doc comment on the struct - otherwise a future reader will copy the stub into a new test and silently bypass evidence verification. Suggest one-line comment: "test-only: bypasses evidence checks; use only when the test is asserting a path *other than* evidence verification."

3. **`turn_runner_worker_completes_after_libsql_turn_and_thread_services_reopen` is gated behind a dedicated `libsql-restart-tests` feature** added in `Cargo.toml` and wired into `code_style.yml` as a separate step. This works but it splinters the CI matrix - any future contributor running `cargo test -p ironclaw_reborn` locally will not run this test by accident. Consider folding it under an existing feature (`libsql-secrets` already pulls in `libsql`) or documenting the gating reason in the Cargo.toml feature definition. Not a blocker; just friction.

4. **`turn_runner_blocks_on_approval_then_coordinator_resume_completes_same_run` asserts `gate_ref == None` on completion** but never asserts the gate was actually resolved against the store. If the production code regressed to "clear gate_ref on any completion regardless of resume path", this test wouldn't catch it. Consider asserting the coordinator's `ResumeTurnResponse` carries the matching `gate_resolution_ref` or that the underlying gate row transitioned to resolved.

### Suggestions

5. The script-capability test asserts capability_invoked precedes assistant_reply_finalized via index comparison on `milestone_names`. Fine, but `position()` returns `Option<usize>`; if either milestone is absent both sides become `None` and `None < None` is false, so the assertion would pass for the degenerate "neither emitted" case. The preceding `contains` checks guard against this in practice - but consider `.unwrap()` on the positions to make intent explicit.

6. `InMemoryCapabilityIo` now tracks `result_refs` in addition to `results`; only the script-capability test reads them. If you wanted parity with the existing `results()` accessor, the doc comment on the struct could explain why both exist (one captures the materialized ref string, the other the structured value).

7. The two CLI tests `run_rejects_invalid_profile` / `doctor_rejects_invalid_profile` duplicate the exact `IRONCLAW_REBORN_PROFILE must be one of ...` literal. If the list of profiles is extended, both will need updating - acceptable, but worth a single shared helper or a partial substring match (`"must be one of"`).

8. The Postgres filtered-replay-gap test uses `tokio_postgres::NoTls` against a connection string the caller controls. The production code now rejects `sslmode=disable` against remote hosts (covered by the new `production_postgres_rejects_remote_sslmode_disable_before_connecting` test) - but the corruption fixture connects via NoTls regardless. Fine for a test-only corruption fixture, but if `IRONCLAW_REBORN_EVENT_STORE_POSTGRES_URL` ever points at a remote host in CI, this would fail. Worth a comment noting the fixture is intended for local Postgres only.

## Verdict

APPROVE with non-blocking nits. The test additions are genuine coverage - they drive real production code paths through the worker entrypoint, cover both event-store backends, and the fail-closed approval test will catch the class of regression it was written for. The two ride-along production fixes are correctly scoped and have matching tests. Address (1) flakiness budget and (2) doc-comment on `AlwaysVerifiedLoopExitEvidence` before merge if convenient; the rest can land as follow-ups.

serrrfirat:
Addressed the review feedback and refreshed the branch against `reborn-integration` to clear the merge conflict.

Fixes applied:
- Raised the turn-runner polling deadlines from 2s to 10s for CI headroom.
- Added comments warning that `AlwaysVerifiedLoopExitEvidence` bypasses durable evidence checks.
- Documented why `libsql-restart-tests` is an opt-in feature.
- Added a coordinator-resume lifecycle assertion for the approval resume path.
- Made milestone ordering assertions use explicit `expect(...)`s.
- Documented why `InMemoryCapabilityIo` tracks both structured results and result refs.
- Loosened duplicated invalid-profile assertions to the stable message prefix.
- Added a local-only note for the Postgres `NoTls` corruption fixture.

Merge refresh pushed as `8c5c4b1b9` (`Merge reborn-integration into PR 3516`).

Targeted verification after conflict resolution:
- `cargo fmt`
- `cargo test -p ironclaw_reborn --test loop_driver_host turn_runner_worker_completes_queued_run_after_turn_store_reopen`
- `cargo test -p ironclaw_reborn --test loop_driver_host turn_runner_worker_drives_script_capability_through_real_host_runtime`
- `cargo test -p ironclaw_reborn --test loop_driver_host turn_runner_blocks_on_approval_then_coordinator_resume_completes_same_run`
- `cargo test -p ironclaw_reborn --test loop_driver_host text_only_model_reply_driver_runs_prompt_model_transcript_path`

henrypark133 (commented):
## Review: hold for rebase

What looks good:
- The SQL replay cursor fix now uses returned SQL row counts rather than filtered entry counts, and extends contiguity verification to the stream head on short pages.
- The model-route validation now rejects a route configured exclusively for a different slot while still allowing shared routes.
- The new libSQL restart coverage drives the reopened turn/thread services through `TurnRunnerWorker`, which is the right caller-level shape.

No verified code findings.

Low-priority note:
- `.github/workflows/code_style.yml` triggers the Reborn CLI smoke path for `crates/ironclaw_reborn/**`, `crates/ironclaw_reborn_cli/**`, and `crates/ironclaw_reborn_config/**`, but not `crates/ironclaw_reborn_event_store/**`. Since this PR changes event-store replay logic and adds event-store contract tests, future event-store-only changes may miss that smoke coverage unless another workflow covers it.

Hold:
- GitHub reports this PR as `CONFLICTING` / `DIRTY`. I would hold approval until it is rebased and the checks rerun on a mergeable head.

Verification:
- `cargo test -p ironclaw_reborn_event_store --all-features --test durable_event_store_contract -- --nocapture`

PR #3539, opened 2026-05-12
refactor(host-runtime): seal Reborn handoff stores

serrrfirat:
## Summary
- seal Reborn host-runtime handoff stores behind host-owned helper methods instead of public raw accessors
- add architecture guardrails for `HostHttpEgressService`, `HostRuntimeServices`, and `BuiltinObligationServices` escape hatches
- update host-runtime contract docs and tests to verify staged network/secret handoffs through runtime egress
- include minimal Cargo.lock repair required for `--locked` checks on current `reborn-integration`

## Scope
- host-runtime boundary refactor only
- memory boundary leak intentionally skipped for follow-up

## Verification
- `cargo test -p ironclaw_host_runtime --no-default-features --locked` — 267 passed
- `cargo test -p ironclaw_architecture --locked -- --nocapture` — 13 passed
- `cargo clippy -p ironclaw_host_runtime --no-default-features --tests --locked -- -D warnings` — no issues
- `cargo fmt --check`
- `git diff --check`


serrrfirat (commented):
Paranoid architecture review found two boundary issues worth tightening before merge:

1. **Medium — runtime HTTP egress can be production-verified with the wrong secret backend.**

   `HostRuntimeServices::with_host_http_egress` accepts any `SecretBackend` and then `with_host_http_egress_service` marks the egress verified solely by checking the internal network-policy/secret-injection handoff store pointers. That means a graph can call `.with_secret_store(real_store)` but `.with_host_http_egress(network, other_store)` and still satisfy the runtime-egress verification path, while `RuntimeCredentialSource::SecretStoreLease` requests will lease/inject from `other_store` instead of the graph-owned secret store. The new test path at `host_runtime_services_contract.rs:3531-3544` even wires a different `InMemorySecretStore` into the helper, which is fine for staged-obligation credentials but normalizes the unsafe shape for the compatibility lease path.

   Suggested fix: make the helper derive its secret backend from `self.secret_store` (using a small `Arc<dyn SecretStore>` wrapper like the new `SharedSecretStore`) or make verification fail unless the egress secret backend is known to be the graph-owned store. Add a regression test where `with_secret_store(real_store)` plus `with_host_http_egress(..., wrong_store)` cannot pass production wiring or cannot inject a `SecretStoreLease` credential from the wrong backend.

2. **Medium/Low — the new boundary guard only removes accessors; raw handoff-store APIs remain public.**

   The PR documents that mutable handoff stores are kept private and adds guardrails for removed accessors, but the root crate still re-exports `NetworkObligationPolicyStore` / `RuntimeSecretInjectionStore`, both store types still have public constructors and mutation/consume methods, and public composition seams still accept raw store handles (`BuiltinObligationServices::with_handoff_stores`, `HostHttpEgressService::with_network_policy_store`, `HostHttpEgressService::with_secret_injection_store`, `BuiltinObligationHandler::with_*_store`). The architecture test only checks for exact accessor strings, so upper Reborn code can still compile against the raw handoff substrate despite the intended helper-only boundary.

   Suggested fix: either fully seal the handoff stores behind host-runtime helpers (`pub(crate)`/test-support-only raw store APIs, or no root re-export), or explicitly document these as intentional low-level host-runtime escape hatches and extend the guardrail to cover the actual public API you do want to ban (re-exports, `pub struct` store types, `with_handoff_stores`, and raw store builder methods).

zmanian (changes requested):
# Re-review — escalating to request changes

In my prior review I asked, as a non-blocking ask, to relocate the deleted one-shot/TTL/scope-isolation assertions from `tests/builtin_obligation_handler_contract.rs` into in-crate unit tests so coverage of the invariants the sealing protects wouldn't be lost. That ask became blocking — the relocations are name-only.

## Hollow renames

**`network_obligation_policy_store_isolates_agent_scope` → `network_obligation_handoff_isolates_agent_scope`** (`tests/builtin_obligation_handler_contract.rs:695`). The new test constructs an `agent_b` scope but never uses it. There is no assertion that agent B cannot retrieve agent A's staged policy. **Cross-scope isolation is untested.**

**`expires_abandoned_direct_satisfy_secret_handoff` → `builtin_obligation_handler_leases_consumes_and_stages_secret_once`** (line 751). The body no longer constructs the store `with_ttl(5ms)`, no longer sleeps, no longer calls `prune_expired()`, no longer asserts `take().is_none()` post-TTL. **TTL expiry of abandoned handoffs is untested.**

Other deletions in the ~800–870 range — including the `policy_store.take().is_some()` preservation-across-reservation-release assertions — also went without replacement.

## What survives

`reborn_e2e_gate_host_http_consumes_staged_policy_and_secret_once` still covers the one-shot consumption invariant via `expect_err("consumed staged secret must not be reusable")`. That's good, but it's only one of the three invariants the sealing was meant to protect.

## What I'd like to see

In-crate `#[cfg(test)] mod tests` in `crates/ironclaw_host_runtime/src/obligations.rs` (where the private store fields are visible) covering:

1. **TTL pruning** — store constructed `with_ttl(Δ)`; insert; sleep > Δ; `prune_expired()` returns 1; `take()` returns `None`.
2. **Cross-agent-scope isolation** — insert under `agent_a` scope; `take()` under `agent_b` scope returns `None`; original `take()` under `agent_a` still returns `Some` (depending on the API). Cover both `NetworkObligationPolicyStore` and `RuntimeSecretInjectionStore`.
3. **Reservation-release preserves staged handoffs** — equivalent to whatever the deleted `policy_store.take().is_some()` post-release assertions exercised.

The sealing itself is good and the production wiring (`try_with_host_http_egress`) addresses serrrfirat's earlier finding. This is purely about not landing a trust-boundary refactor that simultaneously deletes the coverage of the invariants it's meant to preserve.

henrypark133 (approved):
## Review

The host-runtime handoff sealing change looks good. I did not find a blocker in the runtime, security, or boundary surfaces.

What looks good:
- The new helper path keeps network-policy and secret-injection handoff stores private to host-runtime composition.
- The follow-up tests cover TTL expiry, agent-scope isolation, and direct `satisfy(...)` preserving staged handoffs.
- The contract text and architecture guardrails now line up with the intended sealed-handoff composition model.

Residual risk is mostly coverage depth: a future integration test through the composed runtime egress path would be stronger, but the current implementation and tests are enough for this PR.

Approved.

zmanian:
# Re-review — sealing landed, prior changes-requested addressed

Walked the latest revision (273d91d) against the prior changes-requested feedback. The sealing is real and the missing in-crate coverage has been restored.

## Summary

`NetworkObligationPolicyStore`, `RuntimeSecretInjectionStore`, and `RuntimeSecretInjectionStoreError` are no longer re-exported from `lib.rs`; their constructors, `insert`/`take`/`discard_for_capability`, and `with_*_store` builder methods are all `pub(crate)`. `BuiltinObligationServices::with_handoff_stores` and `HostHttpEgressService::with_{network_policy,secret_injection}_store` are `pub(crate)`. The composition path now goes through `BuiltinObligationServices::host_http_egress(network)` and `HostRuntimeServices::try_with_host_http_egress(network)`, both of which return `impl RuntimeHttpEgress` so upper Reborn code never sees the concrete `HostHttpEgressService` type from the helper path. `try_with_host_http_egress` derives its secret backend from the graph-owned `self.secret_store` via `SharedSecretStore` and fails wiring if no graph secret store has been configured, addressing the earlier escape hatch in `with_host_http_egress` taking an arbitrary `SecretBackend`. Architecture test `reborn_host_runtime_services_do_not_expose_lower_substrate_handles` has been extended to ban the specific accessor strings, raw store struct decls, and `with_handoff_stores`.

## Strengths

- Sealing is enforced both by visibility (`pub(crate)` on everything that matters) and by string-grep guardrail in `crates/ironclaw_architecture/tests/reborn_dependency_boundaries.rs`. Belt-and-suspenders is right for a substrate boundary.
- The architecture test reads `obligations.rs` and `services.rs` source directly and now bans `pub struct RuntimeSecretInjectionStore`, `pub struct NetworkObligationPolicyStore`, `pub fn insert(`, `pub fn take(`, `pub fn discard_for_capability(`, `pub fn with_handoff_stores(`, `pub fn with_host_http_egress<N, SecretBackend>`, `pub fn network(&self) -> &N`, etc. That covers the future-drift surface, not just today's accessor names.
- The in-crate `#[cfg(test)] mod tests` in `obligations.rs` covers exactly the three invariants whose external assertions were deleted: TTL pruning of abandoned handoffs, cross-agent-scope isolation for both stores, and direct-satisfy preservation of staged handoffs after reservation release. `prune_expired` is now `#[cfg(test)]` only, which is the right scope for a TTL probe that exists solely for tests.
- `SharedSecretStore` correctly forwards every `SecretStore` method to the inner `Arc<dyn SecretStore>` with no behavioral divergence, so `RuntimeCredentialSource::SecretStoreLease` requests through the host helper now lease from the graph-owned store, not a side-channel.
- `obligation_services_composition_contract.rs` was rewired to drive the dispatcher through `Arc<dyn RuntimeHttpEgress>` built from `services.host_http_egress(network)`. The test now exercises the sealed boundary end-to-end instead of the (now private) raw stores.
- Reborn substrate independence is preserved: the helper methods do not pull anything upper-Reborn into `host_runtime`, and the contract doc explicitly labels the remaining `new`/`new_with_request_policy_for_tests` and raw-store seams as "low-level host-runtime/test harness escape hatches" that "upper Reborn crates must not use."

## Issues

None blocking. Minor observations:

- `HostHttpEgressService` itself remains `pub` because integration tests in `tests/runtime_http_egress_contract.rs` still construct it via `HostHttpEgressService::new_with_request_policy_for_tests` (and one `::new` site at line 946). That is a documented test seam, but it means a careless future caller in a downstream crate could still construct a `HostHttpEgressService` directly — they just can't attach the private stores, so the service is fail-closed on staged-obligation egress. Acceptable, but consider whether the request-policy-fallback constructor could be gated behind a `test-support` cargo feature to make the test/legacy intent enforced rather than documented.
- `network_policy_for_request` reads the policy via `store.get(...)` rather than `take`, leaving discard to obligation completion / process lifecycle. The contract doc was updated to reflect this ("borrow ... while obligation completion/abort or process lifecycle cleanup owns final discard"). Worth a `debug_assert!` or unit test that two back-to-back egress calls in the same invocation observe the same policy — the lifecycle owner / borrow split is now the new invariant and is currently only implicitly tested.
- The Cargo.lock additions for `ironclaw_wasm_product_adapters` (async-trait, chrono, hex, tokio, tracing) are unrelated to the sealing change and the PR description acknowledges them as a `--locked` repair. They make the diff harder to bisect against future regressions; ideally they would be a separate one-line PR, but not worth re-rolling at this point.
- `gemini-code-assist` flagged the deleted one-shot assertion in `host_runtime_services_wasm_http_uses_production_staged_network_and_secret_handoffs`. The new `reborn_e2e_gate_host_http_consumes_staged_policy_and_secret_once` covers the same one-shot invariant through `expect_err("consumed staged secret must not be reusable")`, so coverage is preserved at a different layer. Worth a comment in the deleted assertion's neighborhood pointing readers at the e2e gate test so future maintainers don't re-add it under the assumption it was missed.

## Suggestions

- Add a unit-level regression test inside `services.rs` (or as a `#[cfg(test)]` block) where `with_secret_store(real_store)` plus `try_with_host_http_egress(wrong_network)` is exercised against a `SecretStoreLease` credential injection — the body should confirm the leased material came from `real_store`, not a separately-constructed `InMemorySecretStore`. The existing `host_runtime_services_wasm_http_secret_store_lease_uses_graph_secret_store` test is close to this but is wired through the wasm runtime; a tighter unit-level test would lock the `SharedSecretStore` plumbing in place against future refactors.
- Consider promoting the architecture-test guardrail from string-grep to a deny-listed `cargo public-api` snapshot for `ironclaw_host_runtime`. The current grep is correct but brittle — a future contributor renaming `with_handoff_stores` to `configure_handoff_stores` would silently pass.

## Verdict

Approve, with the small followups above as suggestions for a separate cleanup PR. The prior changes-requested findings (hollow renames, missing TTL/scope-isolation/preservation coverage, wrong-secret-backend escape hatch) are all addressed.

serrrfirat:
Addressed the review followups and refreshed the branch against `reborn-integration`.

Changes pushed:
- Added unit coverage that repeated same-invocation host HTTP egress calls borrow the same staged network policy.
- Added unit coverage that `try_with_host_http_egress` leases `SecretStoreLease` credentials from the graph-owned secret store.
- Added a nearby note pointing the WASM handoff test to `reborn_e2e_gate_host_http_consumes_staged_policy_and_secret_once` for the one-shot staged-secret invariant.
- Merged `origin/reborn-integration` and resolved the architecture boundary test conflict by preserving both guardrail sets.

Verification run locally:
- `cargo fmt`
- `git diff --check`
- `cargo test -p ironclaw_architecture reborn_host_runtime_services_do_not_expose_lower_substrate_handles`
- `cargo test -p ironclaw_host_runtime services::tests::host_http_egress_borrows_staged_policy_for_repeated_invocation_requests`
- `cargo test -p ironclaw_host_runtime services::tests::host_http_egress_helper_leases_secret_store_credentials_from_graph_store`
- `cargo test -p ironclaw_host_runtime host_runtime_services_wasm_http_uses_production_staged_network_and_secret_handoffs`

Current PR state after push: mergeable; CI is rerunning.

zmanian:
# Re-review — followups addressed, composition migration is clean

Following up on the prior re-review at https://github.com/nearai/ironclaw/pull/3539#issuecomment-4445766433. Three new heads since: `81f1a3c1` (egress review followups), `ef09b470` (merge from `reborn-integration`), `371b1cf3` (composition migration to sealed builder), and `49dcc4869` (no-panics scanner suppression).

## Per-item status

| Prior suggestion | Status | Evidence |
|---|---|---|
| Gate `HostHttpEgressService::new_with_request_policy_for_tests` behind a `test-support` cargo feature | NOT ADDRESSED (low priority) | `crates/ironclaw_host_runtime/Cargo.toml` `[features]` block unchanged; `HostHttpEgressService` and the test constructor remain `pub`. Acceptable — the staged-handoff seams (`with_network_policy_store`/`with_secret_injection_store`) are still `pub(crate)`, so the service is fail-closed on staged-obligation egress for any external constructor. Keep as a separate cleanup ticket. |
| `debug_assert!` or unit test for back-to-back egress consistency on `network_policy_for_request` | ADDRESSED | `services.rs::tests::host_http_egress_borrows_staged_policy_for_repeated_invocation_requests` (added in 81f1a3c1) inserts a staged policy once, issues two `execute` calls in the same invocation against a `RecordingNetwork`, and asserts both observed `requests[i].policy == staged_policy`. Locks the borrow-vs-take invariant. |
| Cargo.lock churn from `--locked` repair | ADDRESSED | `git diff origin/reborn-integration..pr-3539 -- Cargo.lock` is empty. The merge from `reborn-integration` (ef09b470) absorbed the upstream lock, so the PR no longer carries unrelated dependency rows. |
| Unit-level regression test for `SharedSecretStore` graph-secret-store plumbing | ADDRESSED | `services.rs::tests::host_http_egress_helper_leases_secret_store_credentials_from_graph_store` (added in 81f1a3c1) seeds an `InMemorySecretStore` with a specific material, wires it via `with_secret_store`, drives a `RuntimeCredentialSource::SecretStoreLease` request through `try_with_host_http_egress`, and asserts the recorded network request carries `authorization: Bearer graph-secret`. This is the tight unit-level assertion that was missing — it pins the `SharedSecretStore(secret_store)` plumbing at line 1029 of `services.rs` against future refactors. |
| Cross-reference note pointing the deleted one-shot assertion to the e2e gate | ADDRESSED | `host_runtime_services_contract.rs:3577-3578` (81f1a3c1) now carries the comment "The consumed-staged-secret one-shot invariant is covered by `reborn_e2e_gate_host_http_consumes_staged_policy_and_secret_once`." |
| Promote architecture-test guardrail from string-grep to `cargo public-api` snapshot | NOT ADDRESSED (low priority) | `crates/ironclaw_architecture/tests/reborn_dependency_boundaries.rs` still uses `.contains()` against source text. As before, this is a "consider" — not blocking. |

## Composition migration (371b1cf3 + 49dcc4869)

The migration is correct. Both `build_libsql_production_host_runtime_services` and `build_postgres_production_host_runtime_services` now go through `try_with_host_http_egress(PolicyNetworkHttpEgress::new(ReqwestNetworkTransport::default()))` instead of constructing `HostHttpEgressService` directly with private accessors. Verified:

- `with_secret_store(Arc::clone(&secret_store))` is called unconditionally on the same builder chain before `try_with_host_http_egress`, so the only failure mode (`ProductionWiringComponent::SecretStore` missing) is genuinely unreachable here. The `.expect` is appropriate.
- The `// safety:` comment was duplicated inline at end-of-line (49dcc4869) to satisfy the "No panics in production code" CI scanner; CI confirms that job is now green.
- The dropped `HostHttpEgressService` import in the use block is consistent with the call-site change — composition no longer names the concrete type, only the `RuntimeHttpEgress` trait shape via the helper.
- The `TODO(#3571)` comment on the composition-side `SharedSecretStore` newtype was expanded to explain its purpose. Note this is a *different* `SharedSecretStore` from the host-runtime-internal one (`services.rs:89,1029`) that wraps `Arc<dyn SecretStore>` for the `HostHttpEgressService::new` secret backend; the composition one exists to share a single concrete `SecretStore` impl between substrate wiring and adapters. Both are fine.

The merge from `reborn-integration` (ef09b470) cleanly resolved the architecture-boundary test conflict by preserving both guardrail sets (per serrrfirat's comment), and `cargo test -p ironclaw_architecture reborn_host_runtime_services_do_not_expose_lower_substrate_handles` is listed in the verification block.

## CI

At time of review: Formatting, "No panics in production code", "Gateway platform/feature boundaries", cargo-deny, classify, scope all green; Clippy (all-features) and Reborn CLI smoke tests still pending. The all-features clippy gate was the specific failure 371b1cf3 was written to fix (E0599/E0624 on the sealed accessors), so its outcome will be the load-bearing signal.

## New issues from migration/merge

None. The composition migration is the minimal correct change, the inline safety-comment workaround is ugly but justified by the panics scanner, and the merge from `reborn-integration` did not regress any of the sealing invariants.

## Verdict

Approve. All four substantive followups from the prior re-review are addressed by 81f1a3c1 + 371b1cf3 (back-to-back-policy unit test, graph-secret-store unit test, Cargo.lock noise via rebase, cross-reference comment). The two remaining suggestions (`test-support` feature gate, `cargo public-api` snapshot) are genuinely deferrable and were always framed as separate-PR cleanup. Pending green clippy/smoke CI, this is ready to land.

henrypark133 (commented):
## Re-review: no new blocker, still needs rebase

What looks good:
- The host-runtime handoff stores are crate-private and no longer exposed through the root public API.
- The follow-up tests now cover TTL pruning, cross-agent isolation for staged network/secret handoffs, and direct `satisfy(...)` preserving staged handoffs across reservation release.
- The production helper path keeps the graph-owned secret store and private handoff stores together.

No verified code findings.

Hold:
- This is already approved on the current head, but GitHub still reports the PR as `CONFLICTING` / `DIRTY`. I would keep the approval contingent on rebasing and rerunning checks on a mergeable head.

Verification:
- `cargo test -p ironclaw_host_runtime --no-default-features network_obligation_handoff_isolates_agent_scope -- --nocapture`

PR #3542, opened 2026-05-12
feat(reborn): add outbound policy service

serrrfirat:
## Summary

- add `OutboundPolicyService` to gate projection subscriptions through a `ThreadProjectionAccessPolicy` before cursor checkpoint creation
- add per-push `ReplyTargetBindingValidator` flow that revalidates outbound candidates before delivery attempts and records sanitized authorization-revoked failures without returning a sendable target
- add service-level contract tests for participant access, non-participant denial, per-push revalidation, validator target-substitution fail-closed behavior, and no delivery attempt without revalidation
- document outbound policy-service seams and delivery-failure guardrails

Part of #3266.

Reviewed diff `6206745642a941829096fef452bbd4a97a3aa992..27f2555b7594555d007e974a3f3a06080a2240ff` before opening this PR.

Checked:
- authorization/data flow: subscription checkpoint creation goes through `ThreadProjectionAccessPolicy`, then store-level scope/cursor validation; denied subscriptions do not create cursor state
- egress safety: delivery preparation calls `ReplyTargetBindingValidator` for every push candidate; unauthorized results record `AuthorizationRevoked` delivery status and return no send target
- fail-closed paths: candidates missing `requires_reply_target_revalidation` fail before persistence; validator target substitution fails before persistence
- persistence boundary: service only records outbound metadata via `OutboundStateStore`; no transcript/projection/turn mutation is introduced
- redaction/secrets: new rows/types contain refs, cursors, status enums, timestamps, and sanitized failure kinds only
- crate boundaries: architecture dependency boundary test passes

Review finding fixed before PR: validator target substitution could have returned a different sendable target; fixed by validating `ValidatedReplyTargetBinding.target == candidate.target` and adding `delivery_preparation_rejects_validator_target_substitution`.

## Verification

- `cargo fmt --check --package ironclaw_outbound`
- `git diff --check origin/reborn-integration...HEAD`
- `CARGO_TARGET_DIR=/tmp/ironclaw-cargo-target-outbound-pr cargo test -p ironclaw_outbound`
- `CARGO_TARGET_DIR=/tmp/ironclaw-cargo-target-outbound-pr IRONCLAW_SKIP_POSTGRES_TESTS=1 cargo test -p ironclaw_outbound --all-features`
- `CARGO_TARGET_DIR=/tmp/ironclaw-cargo-target-outbound-pr cargo clippy -p ironclaw_outbound --all-targets --all-features -- -D warnings`
- `CARGO_TARGET_DIR=/tmp/ironclaw-cargo-target-outbound-pr cargo test -p ironclaw_architecture reborn_crate_dependency_boundaries_hold`


zmanian (changes requested):
# Review

**Summary:** Adds `OutboundPolicyService` with two trait ports (`ThreadProjectionAccessPolicy`, `ReplyTargetBindingValidator`) gating subscription cursor checkpoints and per-push delivery attempts; fail-closed on missing revalidation marker or target substitution. Part of #3266.

## What's solid

- **Fail-closed posture** (`service.rs:64-77, 94-102`). Candidates without `requires_reply_target_revalidation = true` are rejected before validator/store call; thread-scope mismatch rejected; validator target substitution rejected with `InvalidRequest` and no attempt persisted. Denial path records `DeliveryFailureKind::AuthorizationRevoked` with `Failed` status and returns no sendable target. Matches the CLAUDE.md guardrail.
- **Production code clean** — no `.unwrap()`/`.expect()` outside tests; `thiserror`; strong newtypes throughout.

## Blocker — trust-bearing types not sealed

`ThreadProjectionAccessGrant` and `ValidatedReplyTargetBinding` (`types.rs:84-89, 161-164`) have `pub` fields with **no witness/sealed-constructor**. Any code in any crate can synthesize a `ValidatedReplyTargetBinding { target }` literal and bypass the validator entirely.

This is the exact failure mode #3460 (sealed `LoopExitValidationPolicy`) and #3539 (sealed `NetworkObligationPolicyStore`/`RuntimeSecretInjectionStore`) just established a pattern for. The OutboundPolicyService is the **only legitimate producer** of these types. Recommended shape:

- Move fields to `pub(crate)` in `ironclaw_outbound`.
- Expose `pub(crate)` constructor used only inside the service.
- Public accessors for the fields callers need (read-only).
- Repeat for `ThreadProjectionAccessGrant`.

This is the same baseline acceptance criterion as #3492 #2. Should be addressed before merge.

## Other findings (non-blocking)

- **Error classification incomplete** (#3492 AC #5). `OutboundError::AccessDenied` exists, but no Transient vs Permanent distinction for validator errors. The `Err(error) => Err(error)` arm at `service.rs:130` propagates unclassified errors; callers can't distinguish "retry later" from "never retry". Add `DeliveryFailureKind::TransientValidatorError` vs `Permanent`.
- **No audit-event emission on denial.** Defensible since the failed delivery-attempt row is itself the audit record, but worth confirming this surfaces to the observability backend with structured warn-level signal.
- **No rate limiting.** A misbehaving caller can drive arbitrary validator calls + attempt-row writes. Each `prepare` writes a row. Not blocking for this PR but worth tracking against #3492 AC #6.

henrypark133 (changes requested):
## Review: seal trust-bearing outbound types mechanically

The outbound policy service shape is solid overall, especially the per-delivery reply-target revalidation and the target-substitution checks. There is one trust-boundary issue that should be fixed before merge.

What looks good:
- `OutboundPolicyService` validates reply targets before every delivery attempt.
- Target substitution and skipped revalidation fail closed before returning a sendable target.
- Authorization-revoked and transient validator failures are recorded distinctly without mutating canonical transcript/projection state.

Finding:
1. High - `crates/ironclaw_outbound/src/types.rs`: `ThreadProjectionAccessGrant` and `ValidatedReplyTargetBinding` are documented as sealed trust-bearing types, but they still derive `Deserialize`. Because the serde impl is generated inside the defining crate, external callers can deserialize these public types and bypass the `pub(crate)` constructors that are supposed to make `OutboundPolicyService` the only minting path. Why it matters: the claim/grant split is only partially enforced, so forged grants or validated bindings can slip around the service boundary. Expected fix direction: remove `Deserialize` from the trust-bearing grant/binding types, or deserialize only through internal row/claim wrappers that preserve the service-minted boundary.

Low-priority note:
- `load_subscription_cursor()` currently returns `Ok(None)` for a valid `subscription_id` with the wrong actor/scope, and the existing store contract asserts that behavior. I am not treating that as blocking, but the contract docs should stay explicit if the intended semantics are anti-enumeration rather than an authority error.

Recommended verdict: Request changes.

serrrfirat:
Addressed review feedback in c92fb889:

- Removed Deserialize from sealed trust-bearing outbound grant/binding types and the delivery decision envelope that carries a validated binding.
- Documented load_subscription_cursor anti-enumeration semantics for wrong actor/scope/thread lookups.
- Added outbound guardrails documenting sanitized audit rows and rate-limit boundary expectations.

Local verification:
- cargo fmt --check
- cargo clippy --all --benches --tests --examples --all-features
- cargo test --lib
- cargo test -p ironclaw_outbound --tests
- bash scripts/pre-commit-safety.sh

CI checks currently pass.

zmanian:
## Review (security-focused)

### Summary

This PR adds `OutboundPolicyService` as the choke point that gates projection subscription creation through `ThreadProjectionAccessPolicy` and revalidates every push candidate via `ReplyTargetBindingValidator` before a delivery attempt is recorded. The sealed-grant / untrusted-claim split is the right pattern for this trust boundary, and the tests cover the dangerous shapes I'd expect (non-participant denial, validator target substitution, per-push revalidation, fail-closed on missing revalidation marker, classification of transient vs. permanent vs. caller-bug validator errors). No callers exist yet — review is on the contract surface.

### Strengths

- Trust-bearing `ThreadProjectionAccessGrant` and `ValidatedReplyTargetBinding` have `pub(crate)` constructors and **deliberately omit `Deserialize`** so they cannot be wire-forged. `OutboundDeliveryDecision` likewise. This is documented in `CLAUDE.md` and enforced by the type system — strong object-capability hygiene.
- Claim/grant split forces policy and validator implementors to return untrusted types; the service revalidates claim fields against the original request before minting the grant. `validate_access_claim` explicitly cross-checks actor/scope/thread, and the reply-target path explicitly cross-checks `claim.target == candidate.target` with a dedicated regression test (`delivery_preparation_rejects_validator_target_substitution`).
- Fail-closed on missing `requires_reply_target_revalidation` and on `scope.thread_id != candidate.thread_id`, both with tests asserting no phantom attempt row is persisted.
- Error classification at the service boundary distinguishes `AccessDenied` (permanent, recorded as `AuthorizationRevoked`), `Backend`/`Serialization` (transient, recorded as `TransientValidatorError`), and caller-bug errors (`InvalidRequest`, `SubscriptionScopeMismatch`, `DeliveryNotFound` — propagated without writing a row). The `validator_caller_bug` test locks in the no-phantom-row property.
- Anti-enumeration semantics on `load_subscription_cursor` documented and exercised: foreign-tuple reads are indistinguishable from missing rows.
- Failure kinds are sanitized enum values, not raw error strings — matches the crate's "no backend error details" guardrail.

### Issues

**Important — Transient classification uses `matches!` instead of exhaustive `match`.** `is_transient_validator_error` is `matches!(error, OutboundError::Backend | OutboundError::Serialization)`. If a future variant is added to `OutboundError` (e.g., `Timeout`, `ValidatorUnavailable`, `RateLimited`) it falls into the `Err(error) => Err(error)` arm and propagates as a caller-bug error — neither recorded as a failure attempt nor retried. The compiler will not catch this. Recommend converting to an exhaustive `match` that requires every variant to declare its classification, so adding a variant is a compile error until classified.

**Important — Scope/candidate consistency check is partial.** `prepare_delivery_attempt` only checks `request.scope.thread_id != request.candidate.thread_id`. `TurnScope` also carries `tenant_id`, `agent_id`, `project_id`. A caller could pair a candidate produced for tenant A with a scope for tenant B and pass the thread-id check. Yes, the validator is supposed to refuse this — but defense-in-depth says the service should refuse a structurally inconsistent scope/candidate pairing at the boundary before any validator IO. Consider checking the full tuple, or at minimum tenant + project + thread.

**Suggestion — Validator claim has only one field; future additions are silent.** `ReplyTargetBindingClaim { target }` is cross-checked against the candidate. If the claim ever grows fields (expiry timestamp, capability scope, binding nonce), there is no mechanism that forces the cross-check to grow with it. Two ways to make additions fail loud: (a) mark `ReplyTargetBindingClaim` `#[non_exhaustive]` so a constructor change is an upstream break; (b) move the equality check onto a `claim.validate_against(&candidate)` method that enumerates every field, so adding a field requires editing the validator.

**Suggestion — Store-write atomicity for failure rows.** When `store.record_delivery_attempt` itself returns `Backend`/`Serialization` in the `AccessDenied` / transient arms, the error propagates and the caller sees a backend error rather than a recorded denial. Probably correct (the audit row is the whole point and we shouldn't lie about persistence), but worth documenting that callers must treat a `Backend` error from `prepare_delivery_attempt` as "denial *may* have happened, but no row was written" — important for retry idempotency.

**Suggestion — No rate-limit / dedup at the service boundary.** `CLAUDE.md` correctly delegates this to the caller. Recommend a tracked issue ensures the caller-side limit lands before this service has a production callsite, otherwise a misbehaving caller can write unbounded failure rows for the same candidate. Failure rows are sanitized but still consume store capacity.

**Suggestion — `ReplyTargetBindingRef` equality assumed canonical.** The target-substitution check is a `PartialEq` comparison on `ReplyTargetBindingRef`. The defense holds only if `ReplyTargetBindingRef` has canonical equality (no case/normalization slack). Per the typed-internals rule in this repo, validated newtypes normalize at construction. Worth a 2-line test asserting `ReplyTargetBindingRef::new("X") != ReplyTargetBindingRef::new("x")` (or that they're equal, whichever is intended) so the invariant is locked in.

**Suggestion — Borrowed-reference service signature.** `OutboundPolicyService<'a> { store: &'a dyn ..., projection_access_policy: &'a dyn ..., reply_target_validator: &'a dyn ... }` is fine for a stateless service held on the stack, but rules out shared ownership across async tasks. Likely intentional; just noting the constraint for whoever wires the first caller.

### Verdict

Approach is sound and the trust-boundary discipline is exactly right for an outbound egress gate. Request changes on the two **Important** items (exhaustive transient classification, full scope/candidate consistency check). Suggestions can be follow-ups but the `non_exhaustive` claim hardening is the kind of thing that's cheap now and expensive later.

henrypark133 (approved):
## Re-review

What looks good:
- The prior trust-boundary blocker is fixed: `ThreadProjectionAccessGrant` and `ValidatedReplyTargetBinding` no longer derive `Deserialize`, keep their fields `pub(crate)`, and are minted only after service-side claim/request checks.
- Delivery preparation still fails closed for skipped revalidation and target substitution.
- Transient validator failures and authorization revocation are recorded distinctly.
- `cargo test -p ironclaw_outbound` passed.

No verified findings.

Low-priority notes:
- A future malicious-serde regression test would help keep the sealed trust-bearing types from regressing.
- Higher-level caller coverage can be added once this service is wired into an app-facing path.

Summary:
- Prior feedback status: the previously requested trust-bearing type sealing is addressed on the current head.
- Review coverage: worktree-backed outbound security/state/testing re-review, prior feedback check, CI check, and targeted crate tests.

serrrfirat:
Addressed review findings in 164ecf81d:

- Replaced validator transient classification with an exhaustive `match` over every `OutboundError` variant, so future variants require explicit classification.
- Extended outbound push candidates with tenant/agent/project/thread identity and reject any scope/candidate mismatch before validator I/O or store writes.
- Hardened `ReplyTargetBindingClaim` with `#[non_exhaustive]` plus a `validate_against(&candidate)` boundary check before minting `ValidatedReplyTargetBinding`.
- Added regression coverage that structurally mismatched candidates produce no validator calls and no phantom attempt rows.

Verification:
- `cargo fmt --check`
- `CARGO_TARGET_DIR=/Volumes/NVME/ironclaw-target cargo test -p ironclaw_outbound --tests`
- `CARGO_TARGET_DIR=/Volumes/NVME/ironclaw-target cargo clippy -p ironclaw_outbound --tests --all-features -- -D warnings`
- `IRONCLAW_SKIP_POSTGRES_TESTS=1 CARGO_TARGET_DIR=/Volumes/NVME/ironclaw-target cargo test -p ironclaw_outbound --all-features`
- `CARGO_TARGET_DIR=/Volumes/NVME/ironclaw-target cargo check -p ironclaw_product_adapters -p ironclaw_product_workflow -p ironclaw_outbound --all-targets --all-features`

zmanian:
## Re-review (security-focused, follow-up)

Following up on the prior [security review](https://github.com/nearai/ironclaw/pull/3542#issuecomment-4445612209). Two new commits since then (`c92fb889` trust-boundary comments, `164ecf81` review scope validation). Re-review against head `164ecf81d84df520aa97c4004e8850c30bbd4f06`.

### Status of prior findings

| # | Item | Severity (prior) | Status |
|---|------|------------------|--------|
| 1 | Transient classification used `matches!` — future `OutboundError` variants would silently propagate as caller-bug errors | Important | **ADDRESSED** |
| 2 | Scope/candidate consistency check only validated `thread_id`, leaving `tenant_id`/`agent_id`/`project_id` for the validator | Important | **ADDRESSED** |
| 3 | `ReplyTargetBindingClaim` had no `#[non_exhaustive]`, so future fields wouldn't force the cross-check to grow | Suggestion | **ADDRESSED** |
| 4 | Lock in `ReplyTargetBindingRef` canonical equality with a regression test | Suggestion | **NOT ADDRESSED** |
| 5 | Store-write atomicity / "denial may or may not have been recorded" caller contract | Suggestion | Not addressed (CLAUDE.md update mentions sanitized audit rows; explicit retry-idempotency note still missing) |
| 6 | Rate-limit / dedup tracking | Suggestion | Documented as caller responsibility in CLAUDE.md |

### Evidence

**Item 1** — `service.rs:167-175` now uses an exhaustive `match` over `OutboundError` with `Backend | Serialization => true` and the remaining variants explicitly returning `false`. Adding a new variant will fail compilation until classified. Good.

**Item 2** — `OutboundPushCandidate` (`types.rs:61-71`) now carries `tenant_id`, `agent_id`, `project_id`, `thread_id`. The new `validate_delivery_scope_candidate` helper (`validation.rs:119-133`) checks all four. `prepare_delivery_attempt` calls it at `service.rs:92` *before* validator IO, and `validate_delivery_attempt` reuses the same helper so the store layer enforces the same invariant. The new regression test `delivery_preparation_rejects_scope_candidate_mismatch_before_validator_io` (`tests/...contract.rs:189-222`) asserts (a) `InvalidRequest` is returned, (b) `validator.calls() == 0` — the validator is not invoked, and (c) no phantom attempt row is persisted. This is exactly the defense-in-depth shape requested.

**Item 3** — `ReplyTargetBindingClaim` is now `#[non_exhaustive]` (`types.rs:222`) and gained a public `new(target)` constructor plus a `pub(crate) validate_against(&candidate)` method that destructures with `let Self { target } = self;` — adding a new field becomes a compile error until the destructure (and therefore the cross-check) is updated. This is the stronger of the two options I suggested.

### Outstanding

**Suggestion (carried over) — `ReplyTargetBindingRef` canonical equality regression test.** The validator-target-substitution defense and `validate_against` both rely on `PartialEq` over `ReplyTargetBindingRef`. I still don't see a regression test that pins the canonical-equality semantics (e.g. `assert!(ReplyTargetBindingRef::new("X")? != ReplyTargetBindingRef::new("x")?)`, or the opposite if case-insensitive is intended). Worth ~5 lines in the `ironclaw_turns` test suite where the newtype lives, so a future "let me normalize this for convenience" change in `ReplyTargetBindingRef::new` doesn't quietly weaken the target-substitution check.

**No new issues** introduced by the two commits. The `#[non_exhaustive]` + `validate_against` shape is, if anything, cleaner than what I sketched.

### Verdict

Both **Important** items addressed with regression coverage. The remaining suggestion (canonical-equality test) is a follow-up and not a blocker for this PR. **Approve.**

PR #3507, opened 2026-05-11
[Reborn] Centralize snippet display hashing

serrrfirat:
## Summary
- add a shared run-profile helper for opaque model-visible snippet display hashes
- use it for skill snippet model-message refs and memory snippet refs
- remove duplicated FNV-1a hash loops from call sites

## Scope
Follow-up to #3476/#3470. Addresses the #3492 comment item: snippet-ref hash derivation should not be duplicated across prompt/context paths.

## Verification
- `CARGO_TARGET_DIR=/Users/firatsertgoz/Documents/ironclaw/target cargo test -p ironclaw_turns -p ironclaw_host_runtime`
- `CARGO_TARGET_DIR=/Users/firatsertgoz/Documents/ironclaw/target cargo test -p ironclaw_architecture`

Does not close #3492.

PR #3545, opened 2026-05-12
Architectural improvements: reuse host-managed prompt bundle port

serrrfirat:
## Summary
- let `HostManagedLoopPromptPort` validate prompt requests against a dynamic current capability surface lookup
- remove Reborn's duplicated `SurfaceAwareLoopPromptPort` implementation and compose the shared turns prompt port instead
- add regression coverage for dynamic surface-version validation in the shared prompt port
- align the model-failure milestone assertion with the existing sanitized `model_failed` milestone emitted by the host model adapter

## Architectural improvements
This is a separate architectural improvements PR from the Reborn crates architecture review. It improves prompt-building Locality by keeping prompt bundle validation, context shaping, skill snippet handling, and milestone emission in the shared `ironclaw_turns` Module, while Reborn only supplies its surface-state Adapter.

## Verification
- `cargo test -p ironclaw_turns --locked`
- `cargo test -p ironclaw_reborn --locked`
- `cargo clippy -p ironclaw_turns -p ironclaw_reborn --tests --locked -- -D warnings`
- `cargo fmt --check`
- `git diff --check`

### DIFF
--- a/crates/ironclaw_reborn/src/loop_driver_host.rs
+++ b/crates/ironclaw_reborn/src/loop_driver_host.rs
@@ -30,14 +30,14 @@ use ironclaw_turns::{
         CapabilityBatchInvocation, CapabilityBatchOutcome, CapabilityDenied,
         CapabilityDeniedReasonKind, CapabilityDescriptorView, CapabilityFailure,
         CapabilityInvocation, CapabilityOutcome, CapabilityResultMessage, CapabilitySurfaceVersion,
-        FinalizeAssistantMessage, LoopCapabilityPort, LoopCheckpointPort, LoopCheckpointRequest,
-        LoopContextBundle, LoopContextPort, LoopContextRequest, LoopHostMilestoneEmitter,
-        LoopHostMilestoneSink, LoopInputBatch, LoopInputCursor, LoopInputPort, LoopModelMessage,
-        LoopModelPort, LoopModelRequest, LoopModelResponse, LoopProcessRef, LoopProgressEvent,
-        LoopProgressPort, LoopPromptBundle, LoopPromptBundleRef, LoopPromptBundleRequest,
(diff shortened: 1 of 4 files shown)


zmanian (approved):
# Review

**Summary:** Pure de-duplication — replaces local `SurfaceAwareLoopPromptPort` with shared `HostManagedLoopPromptPort` extended via a `with_current_surface_version_lookup` closure so it can read live surface state instead of a static snapshot.

## Verified
- Deleted code in `crates/ironclaw_reborn/src/loop_driver_host.rs:919-1115` is byte-equivalent to the shared port except for the surface-version source. Clean motion.
- Trust-tier ordering preserved — `build_prompt_bundle` in `prompt.rs` retains `instruction_snippets` → `messages` order. Identity-ordering hole from #3476/#3547 is orthogonal (neither old nor new path populates `identity_messages`).
- Snippet-ref centralization from #3507 not regressed — `skill_snippet_model_message_ref` import dropped from `loop_driver_host.rs` confirms the shared port already uses the centralized helper.
- Test assertion fix (`model_failed` milestone addition at `loop_driver_host.rs:290`) is legitimate — aligns with sanitized model-adapter emission; `assert_public_milestones_hide_raw_payloads` still guards leakage.
- New `loop_prompt_port_uses_current_surface_version_lookup_each_build` exercises the dynamic-lookup contract (v1 succeeds, mutate to v2, stale v1 → `StaleSurface`).

## Minor
- `CurrentSurfaceVersionLookup` is `Arc<dyn Fn() -> Result<Option<…>, AgentLoopHostError> + Send + Sync>` (`prompt.rs:18-19`). Sync today; if `CapabilitySurfaceState::current()` ever becomes async this signature will need revisiting.
- `with_current_surface_version` now goes through the closure path — slightly more allocation for callers passing a static version, but keeps one code path. Fine.

LGTM.

PR #3548, opened 2026-05-12
Add DISABLE_TOOLS_LIST flag and security regression test

zetyquickly:
## Summary
- Add `DISABLE_TOOLS_LIST` config flag (CSV or bracketed/quoted list) to disable specific tools.
- Enforced at **both** layers: registration (disabled tools never enter the registry, never sent to the LLM) and dispatch (any direct call is denied with an audited `ActionRecord` carrying `DISABLE_TOOLS_DENIAL_PREFIX`; audit failure is fail-closed).
- Wired through agent dispatcher/scheduler/router and tool registry. Documented in `.env.example`.

## Test plan
- [x] `cargo test` — covered by:
  - Config parse (`src/config/agent.rs`): `parse_disabled_tools_csv`, `parse_disabled_tools_bracketed_quoted`, `parse_disabled_tools_skips_empty_and_trims`, `parse_disabled_tools_strips_at_most_one_bracket_each_side`, `disabled_tools_resolves_from_env`, `disabled_tools_defaults_to_empty_when_unset`.
  - Registry (`src/tools/registry.rs`): `disabled_tool_rejected_at_async_registration`, `is_disabled_treats_hyphen_and_underscore_as_equivalent`, `is_disabled_is_case_insensitive_through_register`, `disabled_tool_unreachable_via_get_resolved_as_defense_in_depth`.
  - Dispatch (`src/tools/dispatch.rs`): `dispatch_denies_disabled_tool_with_distinct_audit_row` (drives `ToolDispatcher::dispatch` end-to-end against the libSQL store).


zmanian (changes requested):
# Review

**Summary:** Adds `DISABLE_TOOLS_LIST` env (CSV/bracketed) parsed into `AgentConfig.disabled_tools`; applied once in `AppBuilder::build_all` via `registry.unregister()` after registration completes. Useful kill-switch concept but enforcement has a real gap.

## Blockers

1. **Startup-only enforcement.** `app.rs:1234-1244` runs the deny-loop once after `build_all`. Any later registration path (extension install via `extension_tools.rs:1169 registry.register(...)`, MCP server hot-reload, dev-tool reload) will silently re-add a "disabled" tool with no re-check. For an incident kill-switch framing this is the primary failure mode — a malicious or compromised extension can re-introduce the exact tool you're trying to block.

   **Fix:** store `disabled_tools` *inside* `ToolRegistry` and reject in both `register()` and `get_resolved()` (`dispatch.rs:124`). Gemini's comment makes the same point.

2. **No audit-log emission on denied dispatch.** Because enforcement is via deletion, a model call to a disabled tool returns the generic "tool not found" path in `dispatch.rs:124` — indistinguishable from typos. An incident kill-switch should emit a distinct `ActionRecord` / structured warn-level event (`tool denied by DISABLE_TOOLS_LIST`). Currently only startup `info!` logs fire (`app.rs:1238`).

## Other findings

3. **Regression test is Python-only at trajectory level + two registry unit tests on the helper.** No Rust test drives `ToolDispatcher::dispatch` with a disabled tool — violates the "test through the caller, not just the helper" rule. Once enforcement moves into the registry, this naturally falls out: add an integration test that calls dispatch and asserts a denial path with the structured audit event.

4. **String-typed flag** (`config/agent.rs:14-29`) is the project anti-pattern but acceptable for an env kill-switch. Parser accepts CSV and bracketed forms — fine, well-tested. Minor: `trim_start_matches('[')` is greedy and strips multiple leading `[`; harmless for tool names but Gemini's point stands.

5. **Default off, global, not per-tenant** — correct for a kill-switch; worth documenting in the env-var reference.

Once enforcement is registry-resident, the existing test pattern naturally exercises the dispatch path too. Functional design is right; placement of the gate is wrong.

zetyquickly:
@zmanian 
Update (review round 2): Moved the kill-switch into the registry itself per @zmanian's review. Registration is now refused for disabled names (extension installs / MCP hot-reloads can no longer silently re-introduce a blocked tool); dispatch denials now persist a distinct ActionRecord with the DISABLE_TOOLS_LIST denial: prefix so SRE/incident tooling can grep for blocked attempts. New caller-level integration test in `src/tools/dispatch.rs::dispatch_denies_disabled_tool_with_distinct_audit_row`.

zmanian (commented):
## Review (round 2)

**Verdict:** zmanian's two blockers from round 1 are both addressed cleanly. Enforcement now lives in the registry (no startup-only gap), and denied dispatches persist a distinct audit row with a stable prefix. The new integration test drives `ToolDispatcher::dispatch` end-to-end, which satisfies "Test Through the Caller, Not Just the Helper". Bracket-strip greediness fixed per Gemini. Three residual issues below, one of which I'd want resolved before merge.

## Blockers — none

## Should-fix

### 1. `register_message_tools` bypasses the kill-switch (`src/tools/registry.rs:920-929`)

`register()` and `register_sync()` both consult `is_disabled` and refuse blocked names. But `register_message_tools` inserts directly into `self.tools` without that check:

```rust
self.tools
    .write()
    .await
    .insert(tool.name().to_string(), tool as Arc<dyn Tool>);
```

The lookup-time `is_disabled` checks in `get`/`get_resolved`/`resolve_name` (and your `disabled_tool_unreachable_via_get_resolved_as_defense_in_depth` test) still neutralise it, so this isn't an exploit — but the comment on `register()` says "extension installs / MCP hot-reloads must not silently re-introduce a tool an operator has explicitly disabled" and `register_message_tools` is just another registration path. Either consult `is_disabled` here too (and skip the `message_tool` cache write + `builtin_names` insert when blocked), or extract a `register_inner` helper that all three callers share. `register_wasm` already routes through `register()` at line 1078, which is the pattern to follow.

### 2. `tracing::warn!` in dispatch denial can corrupt REPL/TUI (`src/tools/dispatch.rs:273`)

CLAUDE.md is explicit about this — and the surrounding code already knows: the ActionRecord-persist failure handler at line 236 uses `debug!` with the comment "dispatch is reachable from interactive REPL/CLI channels where `warn!`/`info!` output corrupts the terminal UI". The same dispatch path emits `warn!` 50 lines earlier on the kill-switch denial.

I get the impulse — an operator kill-switch denial is exactly the kind of event you want SRE alerting on, and a `warn!`-level structured log makes that easy. But the rule applies to the dispatch surface as a whole. Two options: (a) downgrade to `debug!` and rely on the ActionRecord with `DISABLE_TOOLS_DENIAL_PREFIX` as the alerting surface (it's already designed for grep), or (b) keep `warn!` and add a brief note in the comment block explaining why this site is an exception. I'd prefer (a).

## Other findings

### 3. Raw params in denial audit row (`src/tools/dispatch.rs:289`)

The normal dispatch path calls `redact_params(&normalized_params, tool.sensitive_params())` before persisting. The denial path can't — we never resolve the tool — so `params.clone()` lands in `job_actions.params` unredacted. The comment acknowledges this ("Skip params redaction... we record only the raw input shape...").

For an operator-chosen kill-switch this is a bounded risk, and the investigative value of seeing what was attempted is real. Worth a follow-up: the registry actually does still have the disabled tool's metadata available through a non-public path (insertion via `register_message_tools` bypassing the deny-list, or an MCP server registering a same-named tool after construction). If you have a `Tool` instance, you can still call `sensitive_params()` on it; if you don't, you can't. Acceptable as-is, but I'd add a one-line note to the docstring on `record_disable_tools_denial` that callers should treat denial audit-row params as potentially containing whatever the attacker sent.

### 4. `Vec<String>` for tool names — types.md anti-pattern, but acceptable

`.claude/rules/types.md` would push toward a `ToolName` newtype. For an env-derived kill-switch parsed at boot from a string, this is fine — but worth knowing it's a deliberate exception, not an oversight. If/when `ToolName` lands as a domain type, this should migrate.

### 5. Test naming nit

`parse_disabled_tools_strips_at_most_one_bracket_each_side` asserts `[[shell]]` → `["[shell]"]`. The behavior is correct and intentional per the docstring, but a user who types `[[shell]]` into their env almost certainly meant `shell`. Consider adding a `tracing::warn!` (or `debug!` per #2) when an item still contains `[` or `]` after parsing — the failure mode of "I tried to disable a tool and it didn't get disabled" is exactly the case where a soft signal is high-value.

## CI / state

- Code Style, Heavy Integration, Telegram, Slack, Replay gate, Web E2E: green
- WASM WIT Compatibility + Tests (all-features): still in progress at review time
- Mergeable state: `blocked` — likely awaiting zmanian's re-approval

Functional design is right and the security-relevant blockers are resolved. The `register_message_tools` gap is the only thing I'd want closed before merge — the others are improvements but not gates.


---

zmanian (commented):
## Re-review (round 3, same head `ef069d8`)

No code changes since round 2. This pass adds a couple of findings I didn't flag last time and reconfirms what's still open.

### New finding — PR body is stale (PR scope discipline)

The PR description still promises:

> Add `test_abound_local_tools_disabled.py` regression test covering the security issue this flag addresses.

…but the diff has zero Python changes — commit `d36239a "drop abound-specific regression test (not for upstream)"` already removed it. The Rust integration test (`dispatch_denies_disabled_tool_with_distinct_audit_row` in `src/tools/dispatch.rs`) is what actually provides the regression coverage now, and it's the better choice — drives `ToolDispatcher::dispatch` end-to-end against the libSQL store, which matches `.claude/rules/testing.md` "Test Through the Caller, Not Just the Helper".

Please update the PR body so it reflects the actual diff:
- Strike the `test_abound_local_tools_disabled.py` line.
- Mention `dispatch_denies_disabled_tool_with_distinct_audit_row` plus the registry-level tests (`disabled_tool_rejected_at_async_registration`, `disabled_tool_rejected_at_sync_registration`, `is_disabled_treats_hyphen_and_underscore_as_equivalent`, `disabled_tool_unreachable_via_get_resolved_as_defense_in_depth`).
- Update the `## Test plan` checklist: `cargo test` passes (CI just confirmed it); `DISABLE_TOOLS_LIST` enforcement is now covered by the new tests rather than something to run by hand. Tick the box or replace the line.

This is the same class of issue zmanian flagged on #2668: title/body says one thing, diff says another. Cheap to fix.

### New finding — `is_disabled` is case-sensitive

`with_disabled_tools` inserts the raw name plus `-`↔`_` variants, and `is_disabled` checks the input plus the same two replacements. Case-folding is not applied. So `DISABLE_TOOLS_LIST=Shell` and a tool registered as `shell` won't match. `case-insensitive comparisons` in `.claude/rules/review-discipline.md` calls this out for user-supplied strings. Tool names are conventionally lowercase, so the practical exposure is small, but two cheap options:

1. `to_ascii_lowercase()` both on insertion and on lookup — matches the existing rule, costs ~one allocation per lookup.
2. Add a unit test pinning current case-sensitive behavior and document it on the docstring, so a future operator typo doesn't silently bypass.

I'd take (1) — it's strictly safer and the cost is negligible.

### Reconfirms from round 2 (still standing)

1. **`register_message_tools` bypasses the kill-switch** (`src/tools/registry.rs:920-929`). Inserts directly into `self.tools` without consulting `is_disabled`. Lookup-time `get`/`get_resolved`/`resolve_name` checks still neutralise it (covered by `disabled_tool_unreachable_via_get_resolved_as_defense_in_depth`), so this is not exploitable today — but the `register()` doc claims extension installs / MCP hot-reloads "must not silently re-introduce" disabled names, and this path violates the spirit. Either consult `is_disabled` here too or extract a `register_inner` helper shared by all three call sites. `register_wasm` already routes through `register()` (line 1078); follow that pattern.

2. **`tracing::warn!` on denial path** (`src/tools/dispatch.rs:273`) can corrupt REPL/TUI output per CLAUDE.md → Code Style → logging. The very same function uses `debug!` on the persist-failure path with that exact reasoning. Either downgrade to `debug!` (the `ActionRecord` with the stable `DISABLE_TOOLS_DENIAL_PREFIX` is the alerting surface) or add a comment explaining why this site is the exception.

3. **Raw params in denial audit row** — acceptable trade for an operator kill-switch, but worth a one-line note on `record_disable_tools_denial` saying these rows may contain unredacted attacker-supplied params.

### Things that look right

- Builder pattern (`with_disabled_tools` consumes self) keeps the deny-list immutable post-construction. Good docstring on `disabled_tools` field about why no lock.
- Defense in depth at both registration and lookup time.
- `-`↔`_` alias normalization on both sides.
- `DISABLE_TOOLS_DENIAL_PREFIX` is a public stable constant — the test grep, future audit aggregators, and SRE alerting can all key off it.
- Comments on `parse_disabled_tools` explain the `strip_prefix`/`strip_suffix` vs `trim_*_matches` distinction so a future contributor doesn't "simplify" it back to greedy. Good defensive doc.
- The five test files that updated `AgentConfig` literals (`agent_loop.rs`, `commands.rs`, `dispatcher.rs`, `scheduler.rs`, `thread_ops.rs`, `bridge/router.rs`) all match the new field — the Copilot reviewer's concern about a missing field is addressed.

### CI

All 28 named checks green (vs. 2 in-progress at round 2). Mergeable state still `blocked` — likely awaiting `zmanian`'s re-approval, since this PR was changes-requested.

### Recommendation

Approve once:
- PR body is updated to match the actual diff.
- `register_message_tools` either consults `is_disabled` or is refactored through `register_inner`.

Everything else — `warn!` level, case sensitivity, params-redaction docstring — is improvement, not a gate.


---

src/tools/registry.rs
```diff
@@ -173,7 +187,44 @@ impl ToolRegistry {
...
+        self
+    }
+
+    /// Returns true if `name` (or any hyphen/underscore alias) is on the
+    /// kill-switch deny-list. Cheap — synchronous set lookup, no locks.
+    pub fn is_disabled(&self, name: &str) -> bool {
+        if self.disabled_tools.is_empty() {
+            return false;
         }
+        self.disabled_tools.contains(name)
```
zmanian: **Case-sensitive lookup — operator typo silently bypasses the kill-switch.**

`with_disabled_tools` inserts the raw name plus `-`↔`_` variants, and `is_disabled` here checks the input plus the same two replacements. Case is never folded. So `DISABLE_TOOLS_LIST=Shell` against a tool registered as `shell` evaluates to `false` and the dispatch proceeds.

`.claude/rules/review-discipline.md` calls this out specifically: "When comparing user-supplied strings (file paths, media types, extension names), normalize to lowercase with `.to_ascii_lowercase()`." Operator-supplied env-var contents are exactly that class.

Suggested fix:
```rust
pub fn is_disabled(&self, name: &str) -> bool {
    if self.disabled_tools.is_empty() {
        return false;
    }
    let lower = name.to_ascii_lowercase();
    self.disabled_tools.contains(&lower)
        || self.disabled_tools.contains(&lower.replace('-', "_"))
        || self.disabled_tools.contains(&lower.replace('_', "-"))
}
```
…with a matching `name.to_ascii_lowercase()` on insertion in `with_disabled_tools`. Add a test pinning `DISABLE_TOOLS_LIST=Shell` ⇒ `shell` is denied.

---

src/tools/dispatch.rs
```diff
@@ -237,6 +248,66 @@ impl ToolDispatcher {
...
+        // Structured warn-level event for log-based alerting. Don't include
+        // raw params (could contain sensitive values; the dispatch path's
+        // normal redaction hasn't happened yet at this point).
+        tracing::warn!(
+            tool = %tool_name,
+            user_id = %user_id,
+            source = %source,
+            reason = "DISABLE_TOOLS_LIST",
+            "tool dispatch denied by operator kill-switch"
+        );
```
zmanian, line 282: **`tracing::warn!` here can corrupt REPL/TUI output.**

CLAUDE.md → Code Style → logging is explicit: "`info!` and `warn!` output appears in the REPL and corrupts the terminal UI." The very same `dispatch()` function knows this — line 236 uses `debug!` for the persist-failure log and the comment there spells out the reason ("dispatch is reachable from interactive REPL/CLI channels where `warn!`/`info!` output corrupts the terminal UI").

The kill-switch denial path is on the same dispatch surface, so the same rule applies.

Two acceptable resolutions:

1. **Downgrade to `debug!`.** The persisted `ActionRecord` with `DISABLE_TOOLS_DENIAL_PREFIX` is already the alerting surface — SREs can grep for it without a `warn!`-level emit.
2. **Keep `warn!` but document the exception.** Add a comment explaining that operator-kill-switch denials are deliberately elevated for SRE alerting and that the REPL hazard is accepted because denials should be rare.

I'd take (1) for consistency with the rest of dispatch.

---

src/tools/dispatch.rs
```diff
@@ -237,6 +248,66 @@ impl ToolDispatcher {
...
+
+        // Persist a failed ActionRecord so the denial is queryable in
+        // `job_actions`. Skip params redaction (no schema, no sensitive_params
+        // declared) — we record only the raw input shape, which is bounded
+        // by JSON-Schema validation on legitimate callers and by the agent
+        // tool-call surface on LLM callers.
+        let source_label = source.to_string();
+        match self.store.create_system_job(user_id, &source_label).await {
+            Ok(job_id) => {
+                let action = ActionRecord::new(0, tool_name, params.clone())
```
zmanian, line 291: **Audit row carries unredacted params; document the trade.**

The normal dispatch path at line 184 calls `redact_params(&normalized_params, tool.sensitive_params())` before persisting. The denial path can't — the tool was never resolved — so `params.clone()` lands in `job_actions.params` whole.

The comment at line 282 acknowledges the redaction is skipped, but it frames the reason as "no schema, no sensitive_params declared." That's not quite right: it's "we don't have a `Tool` reference to ask," and an attacker calling a blocked `secret_delete` with `{ "secret": "actual-value" }` writes the value into the audit row verbatim.

For an operator-chosen kill-switch this is bounded and the investigative value is real, but I'd want one of:

1. Add a clear hedge on the docstring of `record_disable_tools_denial` ("denial rows persist raw attacker-supplied params; treat as untrusted when displaying").
2. Or: blanket-redact obviously-sensitive keys (`password`, `secret`, `token`, `key`, `authorization`) at this site, even without a tool reference. Cheap and conservative.

(1) is the minimum; (2) would be the safer default.

---

src/tools/registry.rs
```diff
@@ -250,6 +304,14 @@ impl ToolRegistry {
             );
             return;
         }
+        if self.is_disabled(&name) {
+            tracing::warn!(
+                tool = %name,
+                reason = "DISABLE_TOOLS_LIST",
```
zmanian, line 315: **Related: a sibling registration path bypasses this check.**

This `is_disabled` gate in `register()` is the contract, and `register_sync()` mirrors it. But `register_message_tools` (further down at lines 920-929) inserts directly into `self.tools` without consulting `is_disabled`:

```rust
self.tools
    .write()
    .await
    .insert(tool.name().to_string(), tool as Arc<dyn Tool>);
```

The lookup-time `is_disabled` checks in `get`/`get_resolved`/`resolve_name` still neutralise it — and your `disabled_tool_unreachable_via_get_resolved_as_defense_in_depth` test pins that — so this isn't exploitable today. But the docstring on `register()` says extension installs / MCP hot-reloads "must not silently re-introduce a tool an operator has explicitly disabled," and `register_message_tools` violates that spirit.

The structural fix is to extract a `register_inner` helper that all three sites share, so the deny-list check can't be forgotten by a future call site. The minimum fix is adding the same `if self.is_disabled(&tool.name()) { return; }` block to `register_message_tools`.

---

zmanian (commented):
Filing the four open items from round 3 as inline comments so they show up as resolvable threads. None are exploit-now blockers (lookup-time defense in depth covers the registration bypass), but the case-sensitivity gap and the `warn!`/REPL hazard are concrete rule violations worth fixing before merge.

Separately — please update the PR body. It still promises `test_abound_local_tools_disabled.py` but commit `d36239a` removed it; the regression test is now `dispatch_denies_disabled_tool_with_distinct_audit_row` in `src/tools/dispatch.rs`. Title/body must match diff (`.claude/rules/review-discipline.md` → PR Scope Discipline).

---

zetyquickly:
Addressed:

#1 register_message_tools bypass → guard added [registry.rs:1130-1137](vscode-webview://1olh4isu2to21bsu568pbrjfg7sft6piucpdnu54898l2gqtid4u/src/tools/registry.rs#L1130-L1137). Test register_message_tools_refuses_kill_switched_name.
#2 warn! → debug! on dispatch denial [dispatch.rs:270-276](vscode-webview://1olh4isu2to21bsu568pbrjfg7sft6piucpdnu54898l2gqtid4u/src/tools/dispatch.rs#L270-L276).
#6 PR body — stale test_abound_local_tools_disabled.py line removed.
#7 Case-insensitive is_disabled [registry.rs:236-264](vscode-webview://1olh4isu2to21bsu568pbrjfg7sft6piucpdnu54898l2gqtid4u/src/tools/registry.rs#L236-L264). Test is_disabled_is_case_insensitive_through_register.
slop dispatch comment trimmed 7→4 lines. register_message_tools doc trimmed 5→2 lines.
Not addressed (not gates per round 2/3):

#3 raw params docstring hedge
#4 Vec<String> → ToolName newtype
#5 [[shell]] parse warn

zmanian (commented):
## Re-review (round 5, head `ef069d8` — unchanged)

Your 2026-05-18 comment lists items #1, #2, #6, #7 as addressed, with line references and test names — but the PR head is still `ef069d83e9d4ba4183db08f11cc5918c1a75a031` (last commit 2026-05-15 `address zmanian review: registry-resident DISABLE_TOOLS_LIST enforcement`). No commits have been pushed since. The fixes appear to be local-only.

Spot-checks against the pushed `src/tools/registry.rs`:

- **#1 `register_message_tools` bypass** — still present. Lines 910–931 insert directly into `self.tools` without calling `is_disabled`, exactly as in my round-3 finding. No `register_message_tools_refuses_kill_switched_name` test exists in the file.
- **#7 case-insensitive `is_disabled`** — not in. `is_disabled` (line 221) only normalizes hyphen/underscore; no `to_lowercase()` / `eq_ignore_ascii_case` path. The new test name `is_disabled_is_case_insensitive_through_register` is absent; the only adjacency test remains `is_disabled_treats_hyphen_and_underscore_as_equivalent` (line 1286).
- **#2 `warn! → debug!`** and **#6 stale PR-body line** — I didn't re-grep `dispatch.rs` or the PR description, but if the registry-side changes weren't pushed, it's worth re-checking those too.

Action: please push the branch. Happy to take another pass once the changes land — the four open items are not exploit-now blockers, but the registration bypass is the one the registry-resident defense is supposed to close, so the PR shouldn't merge claiming it's fixed when the guard isn't there.

Items I had marked as "not gates" (#3, #4, #5) remain as my prior comments — fine to leave for follow-up.

— Zaki

zetyquickly:
> ## Re-review (round 5, head `ef069d8` — unchanged)
> Your 2026-05-18 comment lists items #1, #2, #6, #7 as addressed, with line references and test names — but the PR head is still `ef069d83e9d4ba4183db08f11cc5918c1a75a031` (last commit 2026-05-15 `address zmanian review: registry-resident DISABLE_TOOLS_LIST enforcement`). No commits have been pushed since. The fixes appear to be local-only.
> 
> Spot-checks against the pushed `src/tools/registry.rs`:
> 
> * **#1 `register_message_tools` bypass** — still present. Lines 910–931 insert directly into `self.tools` without calling `is_disabled`, exactly as in my round-3 finding. No `register_message_tools_refuses_kill_switched_name` test exists in the file.
> * **#7 case-insensitive `is_disabled`** — not in. `is_disabled` (line 221) only normalizes hyphen/underscore; no `to_lowercase()` / `eq_ignore_ascii_case` path. The new test name `is_disabled_is_case_insensitive_through_register` is absent; the only adjacency test remains `is_disabled_treats_hyphen_and_underscore_as_equivalent` (line 1286).
> * **#2 `warn! → debug!`** and **#6 stale PR-body line** — I didn't re-grep `dispatch.rs` or the PR description, but if the registry-side changes weren't pushed, it's worth re-checking those too.
> 
> Action: please push the branch. Happy to take another pass once the changes land — the four open items are not exploit-now blockers, but the registration bypass is the one the registry-resident defense is supposed to close, so the PR shouldn't merge claiming it's fixed when the guard isn't there.
> 
> Items I had marked as "not gates" (#3, #4, #5) remain as my prior comments — fine to leave for follow-up.
> 
> — Zaki

sorry about that, pushed to the wrong remote, please re-review

zetyquickly:
> Addressed:
> 
> #1 register_message_tools bypass → guard added registry.rs:1130-1137. Test register_message_tools_refuses_kill_switched_name. #2 warn! → debug! on dispatch denial dispatch.rs:270-276. #6 PR body — stale test_abound_local_tools_disabled.py line removed. #7 Case-insensitive is_disabled registry.rs:236-264. Test is_disabled_is_case_insensitive_through_register. slop dispatch comment trimmed 7→4 lines. register_message_tools doc trimmed 5→2 lines. Not addressed (not gates per round 2/3):
> 
> #3 raw params docstring hedge #4 Vec → ToolName newtype #5 [[shell]] parse warn

this should be true now

zmanian:
## Re-review (round 4, HEAD `1ed7dffdf`)

Round 2/3 blockers all addressed cleanly — `register_message_tools` now honours the deny-list, case folding lands at insertion and lookup, `debug!` replaces `warn!` on the dispatch denial path, and `dispatch_denies_disabled_tool_with_distinct_audit_row` drives the dispatcher end-to-end against the libSQL store. Good iteration.

One residual gap I want closed before merge, plus smaller polish.

### Should-fix (blocker for me)

**Listing / discovery surface bypasses the kill-switch.**

`src/tools/registry.rs` — `has()`, `list()`, `all()`, `tool_definitions()`, `tool_definitions_for_engine()`, `tool_definitions_for()`, `tool_definitions_for_domain()`, `tool_definitions_excluding()`, `builtin_tool_names()` all iterate `self.tools.read().await.values()` directly with no `is_disabled` filter.

The flag is named `DISABLE_TOOLS_LIST`, the `app.rs` comment promises "every register*/get* path consults the deny-list," and the `disabled_tool_unreachable_via_get_resolved_as_defense_in_depth` test asserts a tool inserted directly into `self.tools` is unreachable via `get*`. But that same setup leaves the tool fully visible to `tool_definitions()` — the schema goes to the LLM, the model attempts the call, dispatch denies it, the audit log fills with denials, and an attacker watching trajectories learns precisely which tools are kill-switched.

Concretely: any future registration path that forgets to consult `is_disabled` (the way `register_message_tools` did until this round) will produce a tool that is dispatch-blocked but LLM-visible. That's not the defense-in-depth invariant the PR is claiming.

Fix: filter `is_disabled` in the enumeration paths next to the existing `is_engine_visible` filter. Add a test that mirrors `disabled_tool_unreachable_via_get_resolved_as_defense_in_depth` but asserts the disabled tool does NOT appear in `tool_definitions()` / `list()` / `all()` after a direct `self.tools.write().await.insert(...)`.

### Smaller items

1. `.env.example` doesn't document `DISABLE_TOOLS_LIST`. An operator who doesn't know it exists can't arm it — please add a one-liner with the CSV/bracketed syntax.
2. `record_disable_tools_denial` docstring should warn that audit-row `params` are unredacted (the code already acknowledges this; lift it to the docstring so callers don't have to read the implementation).
3. `parse_disabled_tools` silently accepts `[[shell]]` → `["[shell]"]` (won't match anything at runtime). A `debug!` when a parsed item still contains `[`/`]`/`"` would catch operator typos cheaply. Round-3 finding, still applies.

### Question

Was the listing-surface gap a deliberate scope choice or an oversight? The flag name and the `app.rs` doc-comment both read like "remove from the list," not "still listed, blocked on call."

zmanian (commented):
**Review: COMMENT** (approaching approve) — the current head addresses all prior blockers and the round-2/3 should-fix items. Enumeration prevention is genuinely multi-path, not single-path: `all()`/`tool_definitions*()` iterate `self.tools`, which never receives a rejected tool (registration is refused), and the lookup-time `is_disabled` guard in `get`/`get_resolved`/`resolve_name` is real defense-in-depth (proved by `disabled_tool_unreachable_via_get_resolved_as_defense_in_depth`). Denial is enforced at the single `ToolDispatcher::dispatch` chokepoint with a distinct audit `ActionRecord`, and audit failure is fail-closed.

**Should-fix**
- `DISABLE_TOOLS_LIST` is a new operator-facing security control but is **not documented in `.env.example`** (CLAUDE.md requires env vars there). Add an entry with the CSV/bracketed-list syntax and a note that it blocks at both registration and dispatch.

**Nits**
- `registry.rs` ~599/622/686: the three registration-rejection paths still use `tracing::warn!`, while this same PR correctly switched the dispatch denial to `debug!`. Registration runs pre-REPL so blast radius is small, but it's the same rule — downgrade for consistency or confirm these only fire before the TUI starts.
- PR body still has unchecked test-plan boxes promising `test_abound_local_tools_disabled.py`, which isn't in the diff. Update the body to cite the actual Rust tests.
- `dispatch.rs` ~368: `source` is formatted via `%source` and again as `source.to_string()`; harmless minor dup.

**Verified clean:** no bypass via direct insert; no prod `.unwrap()/.expect()`; `create_system_job`/`save_action` exist on both postgres + libsql backends (no feature-flag compile risk); config parsing uses `optional_env`/`lock_env` per `helpers.rs` and is well-tested (CSV, bracketed, quoted, empty-skip, non-greedy bracket strip, env resolution). The `Vec<String>` flag is the project's stringly-typed pattern but acceptable for an env kill-switch over arbitrary external tool names.

zetyquickly:
@zmanian please revisit

zetyquickly:
@zmanian please revisit

zetyquickly:
nobody cares about this no more, closing

PR #3555, opened 2026-05-13
arch(ws-4): planner facade - AgentLoopPlanner and DefaultPlanner

henrypark133:
## Context

WS4 adds the planner facade layer for the Reborn agent loop. It composes the nine strategy slots behind a sealed planner API and wires the default built-in loop family to the planner composition point.

Workstream brief: `docs/reborn/agent-loop-briefs/planner-facade.md`
Current base: `reborn-integration`

## What Landed

- `AgentLoopPlanner`, a public identity-only planner facade.
- `AgentLoopPlannerInternal`, a crate-private executor-facing extension trait for strategy slot access.
- `DefaultPlanner`, a crate-private built-in composition for the default loop family.
- Default family wiring through `families::default()` and `LoopFamily`.
- A non-zero BLAKE3-backed `DEFAULT_FAMILY_DIGEST` tied to the declared default planner/strategy composition.
- Duplicate loop-family ID rejection preserved in `LoopFamilyRegistry`.

## Boundary Notes

- `DefaultPlanner` is intentionally `pub(crate)` and is not re-exported from the crate root.
- Public callers can observe only planner/family identity through `AgentLoopPlanner`.
- Concrete default strategy behavior remains owned by WS5; this branch only consumes the WS5 default strategy types in the planner slots.
- Executor tick mechanics remain deferred to WS6a.

## Reviewer Focus

- The public planner API should expose identity/composition only, not raw mutable strategy internals.
- Strategy access must remain available to the executor/families while staying closed to external crates.
- The planner should compose families; it should not become a runtime service locator.
- The default family digest should change when replay-relevant planner composition changes.

## Validation

- `cargo fmt`
- `git diff --check`
- `cargo test -p ironclaw_agent_loop`
- `cargo clippy -p ironclaw_agent_loop -p ironclaw_reborn --all-targets --all-features -- -D warnings`
- `git merge-tree origin/reborn-integration HEAD`

## Stack Position

WS1, WS2, WS3, WS3.5, and WS5 have been squash-merged into `reborn-integration`. This PR is now a standalone WS4 planner-facade branch retargeted to `reborn-integration`.

### DIFF
--- a/crates/ironclaw_agent_loop/src/default_planner.rs
+++ b/crates/ironclaw_agent_loop/src/default_planner.rs
@@ -1,0 +1,372 @@
+//! `DefaultPlanner` — the reference composition of the nine strategies.
+//!
+//! Construction is crate-private. Public callers get a sealed
+//! `AgentLoopPlanner` through `families::*` and `LoopFamilyRegistry`; they
+//! cannot instantiate or mutate planner strategy slots.
+
+use std::sync::Arc;
+
+use crate::families::DEFAULT_FAMILY_DIGEST;
+use crate::family::{ComponentIdentity, LoopFamilyId};
+use crate::planner::{AgentLoopPlanner, AgentLoopPlannerInternal};
+use crate::strategies::{
+    BatchPolicyStrategy, BudgetStrategy, CapabilityStrategy, ContextStrategy,
+    DefaultBatchPolicyStrategy, DefaultBudgetStrategy, DefaultCapabilityStrategy,
+    DefaultContextStrategy, DefaultGateHandlingStrategy, DefaultInputDrainStrategy,
+    DefaultModelStrategy, DefaultRecoveryStrategy, DefaultStopConditionStrategy,
+    GateHandlingStrategy, InputDrainStrategy, ModelStrategy, RecoveryStrategy,
+    StopConditionStrategy,
+};
+
+/// The reference planner: a concrete, Builtin-only composition of nine
+/// strategies.
+#[allow(dead_code)]
+#[derive(Clone)]
+pub(crate) struct DefaultPlanner {
+    id: LoopFamilyId,
+    version: ComponentIdentity,
+    context: Arc<dyn ContextStrategy>,
+    capability: Arc<dyn CapabilityStrategy>,
+    model: Arc<dyn ModelStrategy>,
+    batch: Arc<dyn BatchPolicyStrategy>,
+    gate: Arc<dyn GateHandlingStrategy>,
+    recovery: Arc<dyn RecoveryStrategy>,
+    stop: Arc<dyn StopConditionStrategy>,
+    drain: Arc<dyn InputDrainStrategy>,
+    budget: Arc<dyn BudgetStrategy>,
+}
+
+#[allow(dead_code)]
+impl DefaultPlanner {
+    /// Crate-private constructor for the all-default strategy composition.
+    pub(crate) fn compose_default() -> Self {
+        Self::compose(
+            LoopFamilyId::DEFAULT,
+            ComponentIdentity::from_static("default", DEFAULT_FAMILY_DIGEST),
+            DefaultStrategySlots::default(),
+        )
+    }
+
+    /// Crate-private constructor used by future family factories that need to
+    /// provide an already-selected strategy set.
+    pub(crate) fn compose(
+        id: LoopFamilyId,
+        version: ComponentIdentity,
+        slots: DefaultStrategySlots,
+    ) -> Self {
+        Self {
+            id,
+            version,
+            context: slots.context,
+            capability: slots.capability,
+            model: slots.model,
+            batch: slots.batch,
+            gate: slots.gate,
+            recovery: slots.recovery,
+            stop: slots.stop,
+            drain: slots.drain,
+            budget: slots.budget,
+        }
+    }
+
+    pub(crate) fn with_id(mut self, id: LoopFamilyId) -> Self {
+        self.id = id;
+        self
+    }
+
+    pub(crate) fn with_version(mut self, version: ComponentIdentity) -> Self {
+        self.version = version;
+        self
+    }
+
+    pub(crate) fn with_context(mut self, strategy: Arc<dyn ContextStrategy>) -> Self {
+        self.context = strategy;
+        self
+    }
+
+    pub(crate) fn with_capability(mut self, strategy: Arc<dyn CapabilityStrategy>) -> Self {
+        self.capability = strategy;
+        self
+    }
+
+    pub(crate) fn with_model(mut self, strategy: Arc<dyn ModelStrategy>) -> Self {
+        self.model = strategy;
+        self
+    }
+
+    pub(crate) fn with_batch(mut self, strategy: Arc<dyn BatchPolicyStrategy>) -> Self {
+        self.batch = strategy;
+        self
+    }
+
+    pub(crate) fn with_gate(mut self, strategy: Arc<dyn GateHandlingStrategy>) -> Self {
+        self.gate = strategy;
+        self
+    }
+
+    pub(crate) fn with_recovery(mut self, strategy: Arc<dyn RecoveryStrategy>) -> Self {
+        self.recovery = strategy;
+        self
+    }
+
+    pub(crate) fn with_stop(mut self, strategy: Arc<dyn StopConditionStrategy>) -> Self {
+        self.stop = strategy;
+        self
+    }
+
+    pub(crate) fn with_drain(mut self, strategy: Arc<dyn InputDrainStrategy>) -> Self {
+        self.drain = strategy;
+        self
+    }
+
+    pub(crate) fn with_budget(mut self, strategy: Arc<dyn BudgetStrategy>) -> Self {
+        self.budget = strategy;
+        self
+    }
+}
+
+impl AgentLoopPlanner for DefaultPlanner {
+    fn id(&self) -> &LoopFamilyId {
+        &self.id
+    }
+
+    fn version(&self) -> &ComponentIdentity {
+        &self.version
+    }
+}
+
+impl AgentLoopPlannerInternal for DefaultPlanner {
+    fn context(&self) -> &dyn ContextStrategy {
+        &*self.context
+    }
+
+    fn capability(&self) -> &dyn CapabilityStrategy {
+        &*self.capability
+    }
+
+    fn model(&self) -> &dyn ModelStrategy {
+        &*self.model
+    }
+
+    fn batch(&self) -> &dyn BatchPolicyStrategy {
+        &*self.batch
+    }
+
+    fn gate(&self) -> &dyn GateHandlingStrategy {
+        &*self.gate
+    }
+
+    fn recovery(&self) -> &dyn RecoveryStrategy {
+        &*self.recovery
+    }
+
+    fn stop(&self) -> &dyn StopConditionStrategy {
+        &*self.stop
+    }
+
+    fn drain(&self) -> &dyn InputDrainStrategy {
+        &*self.drain
+    }
+
+    fn budget(&self) -> &dyn BudgetStrategy {
+        &*self.budget
+    }
+}
+
+/// Strategy slots accepted by `DefaultPlanner::compose`.
+///
+/// The type is crate-private so future family factories can compose slot sets
+/// without making strategy traits constructible outside this crate.
+pub(crate) struct DefaultStrategySlots {
+    context: Arc<dyn ContextStrategy>,
+    capability: Arc<dyn CapabilityStrategy>,
+    model: Arc<dyn ModelStrategy>,
+    batch: Arc<dyn BatchPolicyStrategy>,
+    gate: Arc<dyn GateHandlingStrategy>,
+    recovery: Arc<dyn RecoveryStrategy>,
+    stop: Arc<dyn StopConditionStrategy>,
+    drain: Arc<dyn InputDrainStrategy>,
+    budget: Arc<dyn BudgetStrategy>,
+}
+
+impl Default for DefaultStrategySlots {
+    fn default() -> Self {
+        Self {
+            context: Arc::new(DefaultContextStrategy::default()),
+            capability: Arc::new(DefaultCapabilityStrategy),
+            model: Arc::new(DefaultModelStrategy),
+            batch: Arc::new(DefaultBatchPolicyStrategy),
+            gate: Arc::new(DefaultGateHandlingStrategy),
+            recovery: Arc::new(DefaultRecoveryStrategy::default()),
+            stop: Arc::new(DefaultStopConditionStrategy::default()),
+            drain: Arc::new(DefaultInputDrainStrategy),
+            budget: Arc::new(DefaultBudgetStrategy::default()),
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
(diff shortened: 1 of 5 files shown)


crates/ironclaw_agent_loop/src/strategies/recovery.rs
```diff
@@ -118,6 +118,33 @@ pub(crate) enum RetryAlteration {
...
+pub(crate) struct DefaultRecoveryStrategy;
+
+#[async_trait]
+impl RecoveryStrategy for DefaultRecoveryStrategy {
+    async fn on_capability_error(
+        &self,
+        state: &LoopExecutionState,
+        _err: &CapabilityErrorSummary,
+    ) -> RecoveryOutcome {
+        RecoveryOutcome::SkipResult {
```
serrrfirat: **Medium Severity**

Default capability-error recovery skips every error class, including permanent/policy/invalid errors.

`DefaultRecoveryStrategy::on_capability_error` currently returns `SkipResult` for all `CapabilityErrorClass` values because `_err` is ignored. That means `Permanent`, `PolicyDenied`, and `InputInvalid` failures are silently dropped instead of surfaced as terminal failures or model-visible failures. In the default loop, this can make an intentional rejection disappear and let the model continue with incomplete state.

Suggestion: branch on `err.class`. Keep `SkipResult` only for classes that are explicitly safe to drop, and make permanent/policy/invalid/internal failures `Abort { failure_kind: ... }` (or another documented terminal outcome). Add a test that drives this default strategy through each `CapabilityErrorClass`.

crates/ironclaw_agent_loop/src/strategies/recovery.rs
```diff
@@ -118,6 +118,33 @@ pub(crate) enum RetryAlteration {
...
+            recovery: state.recovery_state.clone(),
+        }
+    }
+
+    async fn on_model_error(
+        &self,
+        state: &LoopExecutionState,
+        _err: &ModelErrorSummary,
+    ) -> RecoveryOutcome {
+        RecoveryOutcome::Abort {
```
serrrfirat: **Medium Severity**

Default model-error recovery aborts on transient failures.

`on_model_error` ignores `_err` and always returns `Abort`, so `ModelErrorClass::Transient` receives the same treatment as `ContentFiltered` or `Internal`. That makes the default loop fail on a single retryable provider/network blip instead of retrying with backoff or another explicitly documented retry policy.

Suggestion: branch on `err.class`. At minimum, handle `Transient` separately with `Retry { alter: Some(RetryAlteration::Backoff { ... }) }` or document why default strategy intentionally treats transients as terminal. Add tests for each `ModelErrorClass` so this behavior is locked down.

crates/ironclaw_agent_loop/src/strategies/gate.rs
```diff
@@ -70,6 +70,18 @@ pub(crate) enum GateOutcome {
     },
 }
 
+#[derive(Debug, Clone, Copy, Default)]
+pub(crate) struct DefaultGateHandlingStrategy;
+
+#[async_trait]
+impl GateHandlingStrategy for DefaultGateHandlingStrategy {
+    async fn handle(&self, state: &LoopExecutionState, _gate: &GateSummary) -> GateOutcome {
+        GateOutcome::Block {
```
serrrfirat: **Medium Severity**

Default gate handling blocks on every gate kind.

`DefaultGateHandlingStrategy::handle` ignores `_gate` and returns `Block` for `Approval`, `Auth`, and `Resource` gates alike. Blocking makes sense for user-resolvable gates, but a resource gate (quota/capacity/budget-style refusal) may never become resolvable by waiting, so the default executor can checkpoint and return `Blocked` instead of failing or skipping deterministically.

Suggestion: branch on `gate.kind`. Keep `Block` only for gate kinds that the host/user can later resolve, and make non-resolvable resource gates `Abort` or another documented terminal outcome. Add coverage for each `GateKind`.

serrrfirat:
Cross-PR pattern note: this PR body appears to describe an earlier WS-4 design than the code now implements. It still references `PlannerId`, strict id validation, 20 planner tests, and builder coverage that are not present in this diff; the current spec says `PlannerId` was subsumed by `LoopFamilyId`/`ComponentIdentity`. Same drift risk likely applies across the WS cascade (#3554/#3556/#3557): reviewers need to verify against `docs/reborn/agent-loop-briefs/*.md` and actual diffs, not PR-body claims. Also, this PR now lands the nine `Default*Strategy` impls that WS-4 marks out-of-scope and WS-5 owns, so the cascade needs an explicit coordination note before level2 integration.

serrrfirat:

Reviewed changed behavior and directly affected call paths for this PR (`arch/level1-merged` → `d3dd0067afe4694f2ef83d30cc37609064e74045`). Scope covered the new `DefaultPlanner`, sealed public planner facade, internal strategy access, loop-family registry wiring, strategy default implementations, and the production registry caller.

### Findings

| 1 | Medium | Correctness / replay invariant | `crates/ironclaw_agent_loop/src/default_planner.rs:41-45` + `crates/ironclaw_agent_loop/src/family.rs:34-39` | `ComponentIdentity` is documented as content-addressed / replay-relevant, but `DefaultPlanner::compose_default()` hardcodes `ComponentDigest([0; 32])`; tests also assert the zero digest. Future default strategy behavior changes can keep the same planner version and weaken replay/profile compatibility validation. | Replace the zero digest with a stable digest derived from the planner/strategy composition contract (for example, a `blake3` digest over a declared `default-planner:v1:<strategy ids>` manifest), or use an explicit nonzero constant with a comment and test that changes when the composition changes. Add a regression test that the default planner digest is nonzero and matches the declared composition manifest. |
| 2 | Low | Registry correctness | `crates/ironclaw_agent_loop/src/family.rs:110-115` | `LoopFamilyRegistry::with_families` silently overwrites duplicate family IDs through `HashMap::insert`. Current production caller passes one family, so this is not an immediate break, but a future built-in family addition could shadow an earlier entry without error. | Make duplicate IDs fail construction (return `Result` or use a debug/assertion for built-in-only construction) and add a duplicate-ID test. |

### Security/data-flow notes

- Public planner surface is sealed and identity-only: `AgentLoopPlanner` exposes only `id()` and `version()`.
- Strategy access remains crate-private through `AgentLoopPlannerInternal`; downstream crates cannot inspect or mutate planner strategy slots.
- No new secrets, filesystem, shell, network, DB, auth, or policy side effects introduced.
- New `Arc<dyn ...Strategy>` and `Arc<LoopFamily>` surfaces are immutable after construction; no new lock/task/cache races found.

### Verification evidence

- Local: `cargo test -p ironclaw_agent_loop --all-features` — 59 passed.
- GitHub checks observed: formatting, clippy, cargo-deny, no-panics, regression enforcement — successful.

Merge stance: fix the Medium replay-identity finding before merge; Low duplicate-family guard can be fixed here or tracked if you prefer.

serrrfirat:
## Summary
Reviewed WS4 PR #3555 only.  
Base `22a8e1d4df8022962a2a9911100d47e5c397c816` → head `44ef92863cb9bfb348243c63dfa7824f2f7d5b4e`.

Planner facade sealing mostly holds, but planner version identity is still placeholder while planner now represents concrete strategy composition.

Tests run:
- `cargo test -p ironclaw_agent_loop --lib` ✅

## Findings

| 1 | Medium | Boundaries / Replay safety | `crates/ironclaw_agent_loop/src/default_planner.rs:41-45` | `DefaultPlanner::compose_default()` publishes `ComponentIdentity::from_static("default", ComponentDigest([0; 32]))` even though this PR turns planner into real nine-strategy composition. Composition can change without version/digest changing, so resume/replay compatibility checks cannot detect planner drift. | Derive planner/family digest from composed strategy set + schema/version inputs, or keep placeholder identity out of replay-visible surfaces until digest wiring lands. Add test proving composition/version changes alter digest. |

## Security/data-flow notes
- Public `AgentLoopPlanner` facade exposes identity only.
- Strategy access remains crate-private through internal planner trait; no external strategy injection path found.

- Sealed planner boundary looks sound.
- Replay/version identity remains weaker than content-addressed contract.

## Missing tests
- `crates/ironclaw_agent_loop/src/default_planner.rs` — assert default planner digest is non-zero/content-derived and changes when strategy composition changes.

## Suggested fixes
1. Replace zero digest with deterministic composition digest.
2. Add regression test for digest derivation.

crates/ironclaw_agent_loop/src/default_planner.rs
```diff
@@ -0,0 +1,371 @@
...
+    budget: Arc<dyn BudgetStrategy>,
+}
+
+#[allow(dead_code)]
+impl DefaultPlanner {
+    /// Crate-private constructor for the all-default strategy composition.
+    pub(crate) fn compose_default() -> Self {
+        Self::compose(
+            LoopFamilyId::DEFAULT,
+            ComponentIdentity::from_static("default", ComponentDigest([0; 32])),
```
serrrfirat: **High Severity** — Planner replay identity uses all-zero digest

`compose_default()` exposes `ComponentDigest([0; 32])` as the planner/family version. The digest is documented as replay/content identity, so future planner-slot or strategy-composition changes will still look identical unless every caller remembers to override it manually. That undermines checkpoint/resume/version checks later in the stack.

Fix: derive the digest from a stable planner-composition manifest / slot identities in the build path, or avoid exposing this as a stable replay identity until real content digesting exists. Do not ship all-zero digest as the production identity.

zmanian:
## Review notes — WS4 planner facade

The facade itself is real (not a type alias): sealed public trait + crate-private `AgentLoopPlannerInternal` supertrait + `pub(crate)` builder + only-legitimate-construction-via-`families::default()`. That's the right shape.

### Scope creep vs the brief

PR description says **"Concrete default strategy behavior is WS5"** but this PR ships `DefaultBudgetStrategy`, `DefaultStopConditionStrategy`, `DefaultRecoveryStrategy`, `DefaultCapabilityStrategy`, etc. These are real semantic choices, not stubs. Two options: clarify the boundary moved, or mark these explicitly as placeholders WS5 will overwrite.

Most importantly:

- **`DefaultStopConditionStrategy::should_stop_after_turn` returns `Stop` after every turn.** Combined with `DefaultBudgetStrategy` capping at 32 iterations, the "default" composition runs exactly one turn. This is almost certainly placeholder; should be moved to WS5 or annotated as test-only. Otherwise WS6a will be built against a stop strategy that doesn't represent any production posture.
- `DefaultBatchPolicyStrategy` reading `concurrency_hint` is the only default that looks production-shaped — inconsistent with the placeholder feel of the others, which underscores that the WS4/WS5 boundary needs to be clearer.

### `Arc<dyn Internal>` → `Arc<dyn Public>` upcast

`LoopFamily` stores `Arc<dyn AgentLoopPlannerInternal>`. Upcasting `Arc<dyn Internal>` to `Arc<dyn Public>` requires nightly trait upcasting or an explicit accessor. Today the field is crate-private so this doesn't bite, but a public `planner()` returning `Arc<dyn AgentLoopPlanner>` will need one. Flag for WS6a wiring.

### Minor

- `ComponentDigest([0; 32])` sentinel could collide with a legitimately-computed zero digest; consider `ComponentDigest::PLACEHOLDER`.
- `futures = "0.3"` added as dev-dep only for `block_on` two async calls — `#[tokio::test]` would avoid the new dep since tokio is already in the tree.
- `default_planner.rs` test helper `test_run_context()` (~70 lines) duplicates fixture scaffolding that's now showing up across multiple WS PRs — extract a `test_support` module by WS5.

henrypark133:
Addressed WS4 planner-facade review feedback in `85af6b745`.

What changed:
- Rebased onto the fixed Level1 head.
- `DefaultPlanner::compose_default()` now uses the non-zero `DEFAULT_FAMILY_DIGEST` instead of `ComponentDigest([0; 32])`; planner and family tests assert this identity.
- Preserved WS3.5 duplicate family ID rejection and validation while wiring the planner into `LoopFamily`.
- `DefaultGateHandlingStrategy` now branches by `GateKind`: approval/auth block, resource aborts deterministically.
- `DefaultRecoveryStrategy` now branches by `CapabilityErrorClass` and `ModelErrorClass`; retryable classes retry with backoff, context overflow shrinks context, policy/permanent/internal classes abort with documented failure kinds.
- Added planner-level tests covering the gate and recovery matrices.

Local validation passed:
- `cargo fmt --check`
- `git diff --check`
- `cargo test -p ironclaw_agent_loop --lib`
- `cargo clippy -p ironclaw_agent_loop -p ironclaw_reborn --all-targets -- -D warnings`

Final local review pass approved the branch. GitHub checks are running on the new head.

crates/ironclaw_agent_loop/src/strategies/gate.rs
henrypark133: Resolved in `85af6b745`: gate handling branches by kind. Approval/auth still block; resource gates abort with PolicyDenied instead of waiting forever. Added coverage for each GateKind.

crates/ironclaw_agent_loop/src/strategies/recovery.rs
henrypark133: Resolved in `85af6b745`: default capability recovery now branches by class. Transient/unavailable retry with bounded backoff; policy denied aborts as PolicyDenied; permanent/input-invalid/internal abort as CapabilityProtocolError. Added planner-level coverage for each class.

crates/ironclaw_agent_loop/src/default_planner.rs
henrypark133: Resolved in `85af6b745`: DefaultPlanner::compose_default() now uses the non-zero DEFAULT_FAMILY_DIGEST inherited by families::default(), and tests assert the default planner/family identity.

henrypark133:
Addressed the WS4 planner facade feedback in `85af6b745`.

What changed:
- Rebased onto the fixed Level1 head (`fa888339a`), carrying the WS3.5 validated family id, duplicate-registry guard, and non-zero default family digest into WS4.
- `DefaultPlanner::compose_default()` now uses `DEFAULT_FAMILY_DIGEST` instead of the all-zero placeholder.
- `DefaultGateHandlingStrategy` now blocks only approval/auth gates and aborts resource gates deterministically.
- `DefaultRecoveryStrategy` now branches by `CapabilityErrorClass` and `ModelErrorClass`; transient/unavailable retry with bounded backoff, context overflow retries with shrink-context, and terminal classes abort with explicit failure kinds.
- Added planner-level tests for the gate/recovery class matrix and default identity.
- Public planner facade remains sealed and identity-only; strategy access remains crate-private.

Local validation passed:
- `cargo fmt --check`
- `git diff --check`
- `cargo test -p ironclaw_agent_loop --lib`
- `cargo clippy -p ironclaw_agent_loop -p ironclaw_reborn --all-targets -- -D warnings`

GitHub checks are running on the new head.

henrypark133:
Follow-up from the Level2 final review pass landed in `fa80b6918`.

The default family digest now has an explicit `DEFAULT_FAMILY_DIGEST_SEED` naming `DefaultPlanner` and all nine real default strategy slots, and the digest bytes were refreshed to the SHA-256 of that seed. A unit test now recomputes the SHA-256 so stale seed/constant drift fails locally.

Validation:
- `cargo test -p ironclaw_agent_loop --lib`
- `cargo fmt --check`
- `git diff --check`
- `cargo clippy -p ironclaw_agent_loop -p ironclaw_reborn --all-targets --all-features -- -D warnings`

Issue #3537, opened 2026-05-12
Reborn: model memory as a userland extension


serrrfirat:
## Context

In Reborn, `ironclaw_memory` currently reads like part of the kernel/host-runtime layer. The intended shape is userland: memory should be a pluggable **Extension** so native memory, Honcho, mem0, etc. can declare capabilities while still using host-mediated authority.

Current evidence on `reborn-integration`:

- `ironclaw_host_runtime` has a normal dependency on `ironclaw_memory` (`crates/ironclaw_host_runtime/Cargo.toml`).
- `ironclaw_memory` owns backend traits plus filesystem adapter/repository exports (`crates/ironclaw_memory/src/lib.rs`).
- Current v1 extension manifests are too small: they expose package metadata, runtime, capability descriptions/effects/default permission, inline input schema, and resource profile.
- Current v1 parser rejects user manifests that self-assert first-party/system runtime. Keep that safety property.

Design notes:

- `CONTEXT.md` keeps the glossary: **Extension**, **Extension Manifest**, **Capability Profile**, **Host Port**, **Hot Capability Surface**.
- `docs/adr/0001-capability-manifest-v2-hard-cutover.md` records the v2 manifest hard cutover.
- `docs/adr/0002-native-memory-uses-host-storage-ports.md` records that native memory uses host storage ports instead of raw DB handles.

## Naming decision

Keep existing extension language. Do **not** introduce a new “capability package” name or a new package crate name.

Keep:

```text
crate:       ironclaw_extensions
root:        /system/extensions/<extension-id>/manifest.toml
CLI/docs:    extension / extensions / registry terminology
Rust types:  ExtensionManifest / ExtensionPackage / ExtensionRegistry, evolved for v2
```

V2 is a semantic/shape cutover for extension manifests, not a vocabulary or crate rename.

## Extension Manifest v2 hard cutover

Implement Extension Manifest v2 in the existing `ironclaw_extensions` crate.

V2 is a hard cutover target, not a long-term v1/v2 dual-parser model. Existing Reborn v1 manifests/tests that depend on the extension manifest parser migrate to v2.

Legacy/root product `ironclaw registry` UX remains in existing extension terminology. If it directly imports changed parser/types, update mechanically to keep the workspace compile-clean, but do not rename its UX.

## Source-aware trust rules

Manifest never grants trust. It only declares requested trust/runtime needs. Host policy computes effective trust.

`ManifestSource` is supplied by the loader/install path, not TOML:

[...]

PR #3568, opened 2026-05-13
feat(host-api): add capability profile and host port contracts

serrrfirat:
## Summary

- Add zero-behavior `ironclaw_host_api` vocabulary for host-mediated ports:
  - `HostPortId`
  - `HostPortCatalogEntry`
  - `HostPortCatalog`
  - `HostPortGrant`
  - `HostPortView`
- Add host-defined capability profile contract vocabulary:
  - `CapabilityProfileId`
  - `CapabilityProfileOperationId`
  - `CapabilityProfileSchemaRef`
  - `CapabilityProfileOperationContract`
  - `CapabilityProfileContract`
- Add reusable capability-profile conformance scaffolding in `ironclaw_capabilities`.
- Re-export new contracts from `ironclaw_host_api` and `ironclaw_capabilities` preludes.
- Document profile/host-port catalog contracts in `docs/reborn/contracts/host-api.md`.
- Add draft memory profile contract docs and schema refs in `docs/reborn/contracts/memory-profiles.md` and `docs/reborn/contracts/schemas/memory/`.

Part of #3537.

## Scope

This is roadmap slice 1: zero-behavior prep. It does not change production extension loading, manifest parsing, dispatch, or memory behavior.

`HostPortCatalog` is validation vocabulary only; it is not a runtime implementation registry or adapter factory.

## Verification

```bash
cargo fmt --all --check
cargo test -p ironclaw_host_api
cargo test -p ironclaw_capabilities capability_profile
cargo test -p ironclaw_architecture -- --nocapture
```


zmanian:
## Summary

Adds zero-behavior vocabulary contracts to `ironclaw_host_api` for two new concepts: host-defined capability profiles (`CapabilityProfileId`/`CapabilityProfileOperationId`/`CapabilityProfileSchemaRef`/`CapabilityProfileOperationContract`/`CapabilityProfileContract`) and host-mediated ports (`HostPortId`/`HostPortCatalogEntry`/`HostPortCatalog`/`HostPortGrant`/`HostPortView`), plus a structural conformance scaffold in `ironclaw_capabilities` and draft memory profile documentation/schemas. No runtime wiring, no manifest parser changes, no dispatch changes. Scope is well-bounded and matches the PR description.

## Strengths

- **Disciplined zero-behavior framing.** The PR resists the urge to wire conformance into manifest/dispatch in the same slice. The `AGENTS.md` map explicitly states `HostPortCatalog` is a validation allowlist, not a runtime registry/factory — and the code reflects that.
- **Newtype hygiene.** Every id type goes through a validator on `new`, `Deserialize` re-runs the validator (rejecting both invalid construction and invalid wire-format), `deny_unknown_fields` everywhere at the wire boundary. This is the right shape for boundary contracts.
- **Order-independent equality.** `CapabilityProfileContract` sorts `required_operations` by id and `HostPortCatalog` sorts entries; equality and serialization are stable across construction order, and the test suite asserts this explicitly.
- **Extensibility ergonomics for third-party hooks.** `CapabilityProfileClaim`/`CapabilityProfileClaimedOperation` separate "what an extension claims" from "what the host contract requires"; the structural `CapabilityProfileConformanceReport::evaluate` is the right primitive for third-party providers to assert profile implementation without privileged trust. Conformance findings classify into 5 explicit kinds (`ProfileIdMismatch`, `MissingRequiredOperation`, `UnexpectedOperation`, `InputSchemaRefMismatch`, `OutputSchemaRefMismatch`) which is the right granularity for actionable validator errors.
- **Drift guard test.** `memory_profile_schema_refs_exist_on_disk` ties the doc-layer schema refs to real files; renames will fail loudly.
- **Test coverage is strong for a contract PR.** Invalid id/path enumerations are exhaustive (>10 negative cases each), and the conformance contract tests cover the three meaningful failure shapes and the happy path.

## Issues

### Critical
None. This is a zero-behavior contract slice and behaves like one.

### Major

1. **`HostPortGrant` is a one-field wrapper around `HostPortId` with no scope or attenuation.** Right now `HostPortGrant { id }` is structurally identical to `HostPortCatalogEntry { id }`. If the design intent is for grants to eventually carry scope/obligation/expiry/parameter narrowing (which is implied by "scoped invocation view"), the wire shape needs that headroom now — adding fields later to a `deny_unknown_fields` struct is a wire-breaking change for any extension that has begun serializing them. Recommend either (a) documenting explicitly that `HostPortGrant` will remain a thin newtype and a separate `HostPortScopedGrant` will be introduced for scoped variants, or (b) reserving the shape now (e.g. include an `attenuations: Vec<...>` or `params: Map<String, Value>` field, even if empty/unused). As-is, the indirection between `HostPortGrant` and `HostPortCatalogEntry` looks like a deferred decision that will bite at the wire boundary.

2. **`HostPortView::allows`/`allows_all` are O(n)/O(n*m) but the catalog uses sorted binary search.** Inconsistent: `HostPortCatalog::contains` is `binary_search_by` on a sorted vec, but `HostPortView::allows` is a linear scan and `allows_all` is the cross product. Either sort `grants` in `HostPortView::new` and binary-search (consistent with `HostPortCatalog`), or use a `BTreeSet<HostPortId>` for the view. Per-invocation views may be small in practice, but the asymmetry is a foot-gun for the caller who reasonably expects parity.

3. **`CapabilityProfileSchemaRef` validation is good but unanchored.** The validator rejects `..`, absolute paths, URLs, and the obvious traversal classes — solid. But there is no namespace anchoring: any extension can claim `schemas/memory/context-retrieve.input.v1.json` even if that ref belongs semantically to the first-party memory profile. Conformance only compares string equality of refs, so two extensions claiming the same `schemas/memory/*` ref get a "match" without proving they actually conform to the schema. For zero-behavior prep this is fine, but the docs should call out explicitly that schema-ref equality is **identity**, not **conformance**, and that validating the actual JSON shape against the schema is deferred to a later slice. The `host-api.md` section 10.5 should say this; right now a reader could infer that ref equality plus profile claim means "validated implementation".

4. **`validate_versioned_dotted_id` and `validate_dotted_host_port_id` duplicate ~90% of logic.** Both implement: empty check, 128-byte cap, dot-segmentation, lowercase-start, charset check, segment validation. The only structural differences are the `host.` prefix requirement and the trailing `vN` requirement. Worth a single helper parameterized by a `SegmentRule` enum (or two closures), or at minimum a shared `validate_dotted_segments` helper that both call. If validation behavior drifts later (e.g. someone tightens the charset in one place), the divergence will silently weaken one of the contracts. Note the duplicate `valid_segment_char` in both modules.

### Minor

5. **`CapabilityProfileClaimedOperation` and `CapabilityProfileOperationContract` are structurally identical** (same three fields, same constructor, same validation). They differ only by semantic role ("required by contract" vs "claimed by extension"). Today this is fine — the type-level distinction prevents confusion at call sites. But a type alias or trait would let the conformance evaluator share more logic. Not a blocker.

6. **`HostPortCatalog::missing_required` does `O(n*m)` dedup via `iter().any(...)`.** For small inputs this is fine. If the inputs are intentionally bounded, document the bound; otherwise prefer a `BTreeSet` for dedup. The PR description and gemini's review both flag a related concern.

7. **`CapabilityProfileConformanceReport::evaluate` is a free function disguised as a constructor.** It takes `&CapabilityProfileContract` and `&CapabilityProfileClaim` and returns `Self`. Idiomatic Rust would put this on the contract (`contract.evaluate(claim)`) or as a free function (`evaluate_conformance(&contract, &claim)`). The current `Report::evaluate` reads slightly oddly because the report is the output, not the actor. Cosmetic.

8. **No `Display`/`std::error::Error` impls on `CapabilityProfileConformanceFinding`.** Findings will likely need to render in CLI/log output. Consider adding `Display` now (1-2 lines per variant) so call sites don't reimplement formatting. Not blocking but cheap.

9. **`HostPortGrant::new` is infallible but `HostPortView::new` is fallible.** Both go through identical validation paths (`HostPortId` already validated, duplicate detection in the view). This is correct, but worth noting: the fallibility surface is consistent (containers validate uniqueness, atoms validate format). Good.

10. **Doc nit:** `host-api.md` section 10.5 distinguishes `CapabilityProfileId` from `RuntimeProfile`. Good — but the reader has no link/reference to where `RuntimeProfile` is defined. Either add an anchor or drop the cross-reference.

11. **`docs/reborn/contracts/AGENTS.md` is excellent** but mentions "after the Manifest v2 cutover" without linking to or summarizing what Manifest v2 entails. New readers will hit this and have nothing to anchor against. A one-sentence forward reference or a link to #3537 would help.

## Suggestions

- Factor the two dotted-id validators into a shared parameterized helper.
- Decide now whether `HostPortGrant` will ever carry attenuations; if yes, reserve the shape; if no, document that explicitly in the AGENTS map.
- Make `HostPortView` membership backed by the same primitive (sorted or set) as `HostPortCatalog`.
- Add a one-line clarification to `host-api.md` 10.5: schema-ref equality is identity, not conformance.

## Verdict

**Approve with minor follow-ups.** This is exactly the right shape for a vocabulary-only contract slice: tight validators, fail-closed deserialization, order-independent equality, no runtime entanglement, and explicit non-goals. The capability-profile/claim/conformance split is the right primitive for third-party extension hooks to assert profile implementation without the host having to grant trust — which aligns with the Reborn extensibility goal of letting third-party extensions participate as first-class providers. The major items above (especially #1 on `HostPortGrant` shape headroom and #4 on validator duplication) are worth addressing in a follow-up but do not block merging this slice given the explicit zero-behavior scope.

henrypark133 (approved):
## Review

What looks good:
- `ironclaw_host_api` remains vocabulary/contract-only; concrete host-port implementations stay out of this crate.
- Host-port catalog/view and capability-profile contracts enforce sorted/deduped stable shapes.

No verified findings.

Low-priority notes:
- `crates/ironclaw_host_api/tests/host_api_contract.rs` uses `context-touch` schema refs that are not in `docs/reborn/contracts/memory-profiles.md` or the schema-ref drift guard. Either document/add those schema files or switch the equality test to existing documented refs so docs and fixtures do not drift.
- Once host-port vocabulary is consumed by a real runtime/authorization caller, add a caller-level test at that boundary.

Summary:
- Review coverage: worktree-backed host-api/capabilities contract review, schema-ref pass, CI, and prior-thread check.
- Residual risk: this is still declarative vocabulary, so downstream consumers need their own boundary tests when they start enforcing these IDs.

serrrfirat:
Addressed review feedback from https://github.com/nearai/ironclaw/pull/3568#issuecomment-4445611330 and pushed fixes in `2abe85191` plus base merge `9555c3bea`.

Fixed:
- #1: documented `HostPortGrant` as intentionally thin; future attenuation/scope narrowing should use a distinct wire type.
- #2: `HostPortView` now sorts grants and uses binary search for membership, matching catalog behavior.
- #3: `host-api.md` now says schema-ref equality is identity/reference matching, not JSON-schema conformance.
- #4: factored shared dotted-id validation into `dotted_id.rs` for capability profile IDs and host-port IDs.
- #6: `HostPortCatalog::missing_required` now dedups with `BTreeSet`.
- #7/#8: added `evaluate_profile_conformance`, `CapabilityProfileClaim::evaluate_against`, and `Display`/`Error` for conformance findings.

Left #5 unchanged: the separate operation contract/claim types keep useful semantic separation and the review marked that item non-blocking.

Merge conflict/update:
- Merged latest PR head updates and `reborn-integration`; no manual conflict hunks remained after ort merge.

Verified:
- `cargo fmt --check`
- `cargo test -p ironclaw_host_api`
- `cargo test -p ironclaw_capabilities`

PR #3569, opened 2026-05-13
fix(reborn): close architecture integrity gaps

serrrfirat:
## Summary
- add the frozen Reborn V1 virtual roots to `VirtualPath` and redact host-looking scoped-path validation errors
- redact bearer-like `CredentialSessionId` display/error surfaces while preserving private DB storage keys
- route native product-adapter `NoOp` payloads through `ProductWorkflow` and fail closed on webhook auth-strategy mismatch before parsing

## Architecture integrity audit
- Used the `architecture-integrity-review` rubric against `reborn-integration` docs/contracts and crate boundaries.
- Checked open PRs targeting `reborn-integration` first and avoided scopes already covered by active PRs (run-state approval stores, memory adapters, inbound product workflow, prompt bundle, host-runtime sealing, composition root, CLI, loop readiness, model gateway tests, etc.).
- Selected three non-overlapping fixes from the remaining findings: host API path contract drift, secret session-id redaction, and native product-adapter auth/NoOp boundary behavior.

## Verification
- `cargo test -p ironclaw_host_api --locked`
- `cargo test -p ironclaw_secrets --locked`
- `cargo test -p ironclaw_wasm_product_adapters --locked`
- `cargo test -p ironclaw_architecture --locked`
- `cargo clippy -p ironclaw_host_api --tests --locked -- -D warnings`
- `cargo clippy -p ironclaw_secrets --tests --locked -- -D warnings`
- `cargo clippy -p ironclaw_wasm_product_adapters --tests --locked -- -D warnings`
- `cargo check -p ironclaw_secrets --features libsql --locked`
- `cargo check -p ironclaw_secrets --features postgres --locked`
- `cargo fmt --check`
- `git diff --check`

### DIFF
--- a/crates/ironclaw_architecture/tests/reborn_dependency_boundaries.rs
+++ b/crates/ironclaw_architecture/tests/reborn_dependency_boundaries.rs
@@ -1,4 +1,8 @@
-use std::{collections::HashMap, path::PathBuf, process::Command};
+use std::{
+    collections::{BTreeSet, HashMap},
+    path::PathBuf,
+    process::Command,
+};
 
 use serde_json::Value;
 
@@ -35,6 +39,32 @@ fn reborn_boundary_rules_active_crates_are_workspace_members() {
             manifest.display()
         );
     }
+}
+
+#[test]
+fn reborn_virtual_roots_match_storage_placement_contract() {
+    let root = workspace_root();
+    let path_source = std::fs::read_to_string(root.join("crates/ironclaw_host_api/src/path.rs"))
+        .expect("host API path source must be readable");
+    let storage_contract =
+        std::fs::read_to_string(root.join("docs/reborn/contracts/storage-placement.md"))
+            .expect("storage placement contract must be readable");
+    let filesystem_contract =
+        std::fs::read_to_string(root.join("docs/reborn/contracts/filesystem.md"))
+            .expect("filesystem contract must be readable");
+
+    let implemented = extract_virtual_roots_const(&path_source);
+    let storage = extract_storage_placement_roots(&storage_contract);
+    let filesystem = extract_filesystem_namespace_roots(&filesystem_contract);
+
+    assert_eq!(
+        implemented, storage,
+        "ironclaw_host_api VIRTUAL_ROOTS must match storage-placement.md canonical roots"
+    );
+    assert_eq!(
+        filesystem, storage,
+        "filesystem.md namespace roots must match storage-placement.md canonical roots"
+    );
 }
 
 #[test]
@@ -1106,6 +1136,76 @@ fn workspace_root() -> PathBuf {
         .and_then(|path| path.parent())
         .expect("architecture crate must live under crates/ironclaw_architecture")
         .to_path_buf()
+}
+
+fn extract_virtual_roots_const(source: &str) -> BTreeSet<String> {
+    let const_body = source
+        .split("const VIRTUAL_ROOTS: &[&str] = &[")
+        .nth(1)
+        .and_then(|tail| tail.split("];").next())
+        .expect("VIRTUAL_ROOTS const array must be present");
+    extract_quoted_absolute_paths(const_body)
+}
+
+fn extract_storage_placement_roots(contract: &str) -> BTreeSet<String> {
+    contract
+        .lines()
+        .filter_map(|line| {
+            let root = line
+                .strip_prefix("| `")?
+                .split('`')
+                .next()
+                .expect("table cell must close code span");
+            let root = if root.starts_with("/engine/") {
+                "/engine"
+            } else {
+                root
+            };
+            Some(root.to_string())
+        })
+        .filter(|root| is_canonical_virtual_root(root))
+        .collect()
+}
+
+fn extract_filesystem_namespace_roots(contract: &str) -> BTreeSet<String> {
+    let roots_block = contract
+        .split("Frozen V1 canonical virtual roots")
+        .nth(1)
+        .and_then(|tail| tail.split("Recommended meaning:").next())
+        .expect("filesystem.md must list frozen V1 canonical virtual roots");
+    roots_block
+        .lines()
+        .map(str::trim)
+        .filter(|line| is_canonical_virtual_root(line))
+        .map(ToString::to_string)
+        .collect()
+}
+
+fn extract_quoted_absolute_paths(source: &str) -> BTreeSet<String> {
+    source
+        .lines()
+        .map(str::trim)
+        .filter_map(|line| line.strip_prefix('"')?.split('"').next())
+        .filter(|root| is_canonical_virtual_root(root))
+        .map(ToString::to_string)
+        .collect()
+}
+
+fn is_canonical_virtual_root(value: &str) -> bool {
+    matches!(
+        value,
+        "/engine"
+            | "/system/settings"
+            | "/system/extensions"
+            | "/system/skills"
+            | "/users"
+            | "/projects"
+            | "/memory"
+            | "/artifacts"
+            | "/tmp"
+            | "/secrets"
+            | "/events"
+    )
 }
 
 fn package_dependencies(package: &Value) -> Option<(String, Vec<String>)> {
--- a/crates/ironclaw_host_api/src/path.rs
+++ b/crates/ironclaw_host_api/src/path.rs
@@ -63,10 +63,16 @@ impl fmt::Display for MountAlias {
 
 const VIRTUAL_ROOTS: &[&str] = &[
     "/engine",
+    "/system/settings",
     "/system/extensions",
+    "/system/skills",
     "/users",
     "/projects",
     "/memory",
+    "/artifacts",
+    "/tmp",
+    "/secrets",
+    "/events",
 ];
 
 /// Common raw host-path prefixes rejected before scoped-path normalization.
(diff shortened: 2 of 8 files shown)


zmanian:
# Review

Walked the diff against the substrate-independence and filesystem-contract rules. Three correctly-scoped fixes; one remaining redaction gap.

## Summary

Three substrate fixes land cleanly: (1) `VIRTUAL_ROOTS` aligned to the 11-root storage-placement set, (2) session-id redaction in `to_private_storage_string` for path-like values, (3) `NoOp` routing + auth-mismatch fail-closed in `WebhookProcessOutcome`. Tests cover the happy and rejection paths, and nothing pulls upper-Reborn or legacy-binary types into the substrate crates.

## Strengths

- `VIRTUAL_ROOTS` set now matches `storage-placement.md`. Previously the 5-root list in `filesystem.md` and the 11-root list in `storage-placement.md` had drifted; PR aligns with the controlling freeze index.
- `to_private_storage_string` redacts session-id-shaped segments when normalizing scoped paths. Heuristic is conservative (host-looking branches) but correct on those branches.
- `WebhookProcessOutcome::NoOp` plus the auth-mismatch fail-closed branch addresses the prior inverted-dependency concern where webhook routing reached back into binary-owned auth state.
- No new `pub use` re-exports; substrate boundary is held.
- Tests through callers (`webhook_dispatch_*`, `scoped_path_*`), not just helpers.

## Issues

**Major — incomplete `ScopedPath` redaction surface** (`crates/ironclaw_host_api/src/path.rs:237-285`)
`normalize_absolute_path` redacts only the three host-looking heuristic branches. Empty input, NUL bytes, backslash-only paths, non-absolute paths, and `..` segments all echo `raw` verbatim. UNC paths (`\\server\share`) bypass `looks_like_windows_path` because that check requires `bytes[1] == b':'`. This violates filesystem-contract rule 10 (no raw input echo on rejection). Fix is straightforward: redact in every error branch, not just the host-looking ones.

**Major — `WebhookProcessOutcome::NoOp` has no production constructor** (`crates/ironclaw_wasm_product_adapters/src/runner.rs:84`)
After the diff, no production code path constructs `NoOp`. Either wire the routing path that was supposed to produce it, or remove the variant. Today it's dead code that future readers will assume is reachable.

## Minor

- `WebhookAuth::matches_requirement` does not match `RequestSignature { timestamp_header_name: None }`. Correct today (no caller produces that combination) but worth a comment so a future caller doesn't assume coverage.
- `docs/reborn/contracts/filesystem.md:88-96` still lists 5 V1 roots and labels `/tmp`/`/artifacts` as scoped aliases. The PR aligns with `storage-placement.md` (11 roots, the freeze-index canonical source). Reconcile the doc here or in an immediate follow-up — leaving them inconsistent invites future drift.
- No architecture-level test pinning `VIRTUAL_ROOTS` to the storage-placement set. A test in `crates/ironclaw_architecture/tests/` reading both sources and asserting equality would prevent silent re-drift.
- `to_private_storage_string` is called from `PersistedCredentialSession` (feature-agnostic struct) but its body is gated to `libsql|postgres`. Works today because the call site is only reached under those features, but the coupling is fragile.

## Verdict

Approve with minor changes. Substrate boundaries are respected and the three intended fixes are correct. Redaction surface needs one more pass to close the `normalize_absolute_path` echo cases before this can be called "filesystem-contract complete."

serrrfirat:
Addressed review feedback in `bb3ee1e29 fix(reborn): address filesystem review gaps`.

Fixes included:
- Closed `ScopedPath` redaction gap: all rejected scoped-path values now use `<redacted path>` instead of echoing raw input, including empty, control/NUL, backslash/UNC, relative, and traversal cases.
- Removed dead `WebhookProcessOutcome::NoOp`; no-op product payloads still route through `ProductWorkflow` and return an acknowledged no-op ack.
- Made `WebhookAuth::matches_requirement` exhaustive and documented intentional rejection of timestamp-less HMAC requirements.
- Reconciled filesystem/storage-placement virtual-root docs and added an architecture drift test pinning `VIRTUAL_ROOTS` to the storage-placement contract.
- Made `CredentialSessionId::to_private_storage_string` feature-agnostic to avoid fragile DTO/backend feature coupling.

Verification run:
- `cargo fmt --all`
- `cargo test -p ironclaw_host_api scoped_path -- --nocapture`
- `cargo test -p ironclaw_wasm_product_adapters runner::tests -- --nocapture`
- `cargo test -p ironclaw_architecture reborn_virtual_roots_match_storage_placement_contract -- --nocapture`
- `cargo check -p ironclaw_secrets --no-default-features`
- `git diff --check`

zmanian:
# Re-review (commit `bb3ee1e`)

Followup to [the prior review](https://github.com/nearai/ironclaw/pull/3569#issuecomment-4445628800). Walked the new commit `fix(reborn): address filesystem review gaps` against each item flagged previously.

## Status per item

| # | Item | File / Anchor | Status |
|---|------|---------------|--------|
| Major 1 | `normalize_absolute_path` echoes raw input on empty/NUL/backslash/non-absolute/`..` and UNC bypass of `looks_like_windows_path` | `crates/ironclaw_host_api/src/path.rs:237-300` | ADDRESSED |
| Major 2 | `WebhookProcessOutcome::NoOp` dead variant | `crates/ironclaw_wasm_product_adapters/src/runner.rs:78-83` | ADDRESSED |
| Minor 1 | `WebhookAuth::matches_requirement` silently rejects `RequestSignature { timestamp_header_name: None }` | `runner.rs:108-125` | ADDRESSED |
| Minor 2 | `filesystem.md` lists 5 V1 roots, drifts from `storage-placement.md`'s 11 | `docs/reborn/contracts/filesystem.md:88-104` | ADDRESSED |
| Minor 3 | No architecture test pinning `VIRTUAL_ROOTS` to storage-placement | `crates/ironclaw_architecture/tests/reborn_dependency_boundaries.rs:45-67` | ADDRESSED |
| Minor 4 | `to_private_storage_string` feature-gating coupling fragility | `crates/ironclaw_secrets/src/lib.rs:248-253` | ADDRESSED |

## Notes on each fix

- **Scoped-path redaction is now exhaustive.** `invalid_path_error_value` returns `<redacted path>` for every `PathKind::Scoped` rejection branch, covering empty/control/backslash/relative/`..`/UNC alongside the prior URL and Windows-path branches. `Virtual` and `MountAlias` continue to echo `raw` in errors, which is correct — those inputs are crate-internal (a known root or a configured mount alias), not user-supplied host-like data, so they aren't covered by filesystem-contract rule 10.
- **`NoOp` removed; routing preserved.** `WebhookProcessOutcome` now has a single `Acknowledged { ack }` variant. The runner tests (`runner.rs:621-705`) drive a parsed `ProductInboundPayload::NoOp` through `ProductWorkflow` end-to-end and assert the workflow saw the payload and returned `Acknowledged { ack: ProductInboundAck::NoOp }`. No-op routing is now expressed at the payload layer, not the outcome layer — cleaner than the previous shape.
- **`matches_requirement` is exhaustive.** All `AuthRequirement` variants are explicitly listed, and the `RequestSignature { timestamp_header_name: None }` arm has a comment documenting the intentional rejection. A future caller producing that combination will get a clear `false` rather than a silent wildcard.
- **Doc/code/test drift pinned.** `filesystem.md` now lists the same 11 roots as `storage-placement.md`. `reborn_virtual_roots_match_storage_placement_contract` reads both `crates/ironclaw_host_api/src/path.rs` and `docs/reborn/contracts/filesystem.md`, extracts each set, and asserts equality against the canonical storage-placement list. Future drift in any of the three sources will fail this test.
- **`to_private_storage_string` is feature-agnostic.** No more `#[cfg(feature = "libsql|postgres")]` on the helper; comment documents why and an `#[allow(dead_code)]` covers the featureless build. DTO/backend coupling fragility is gone.

## New issues

None. The architecture drift test (`reborn_dependency_boundaries.rs:1143`) parses `const VIRTUAL_ROOTS: &[&str] = &[` from the source file as a brittle string split, but the assertion against the doc set will trip immediately if the const layout changes, so the test will fail loud rather than silent. Acceptable as a contract pin.

## Verdict

Approve. All items from the prior review are closed and the substrate boundaries are still respected. Ready to merge from a filesystem-contract / architecture-integrity standpoint.

henrypark133 (changes requested):
## Review: normalized scoped-path host-prefix bypass

What looks good:
- The credential-session id redaction change keeps the reusable session UUID out of `Debug`/`Display` and preserves private persistence through storage-only conversion.
- Native product-adapter webhook handling now fails closed on auth-strategy mismatch before parsing, and explicit `NoOp` payloads still route through `ProductWorkflow`.
- The virtual-root contract now includes the frozen V1 roots and has caller-level contract coverage.

Finding:

Medium - `crates/ironclaw_host_api/src/path.rs:201`

`ScopedPath::new` checks `RAW_HOST_PREFIXES` before path normalization, then accepts the normalized value. Inputs like `/./Users/alice/private.txt` or `//etc/passwd` do not start with the raw prefix before normalization, but normalize into host-looking absolute paths and are accepted as scoped paths.

Why it matters: the Reborn path contract says `ScopedPath` must not look like a raw host path or URL. This bypass can put host-looking paths back onto an extension-visible path surface and weakens the redaction fix this PR is trying to land.

Expected fix direction: run the raw-host-prefix rejection against the normalized candidate as well, returning the redacted placeholder for those errors, and add contract tests for normalized bypass forms such as `/./Users/...`, `//etc/...`, and `/./private/...`.

Low-priority note:
- `WebhookAuth::matches_requirement` still uses a wildcard fallback for unsupported/future auth requirements. It fails closed today, so I do not consider it blocking, but explicit arms for `SessionCookie`/`BearerToken` would make future auth-surface expansion easier to audit.

Verification:
- `cargo test -p ironclaw_host_api --test host_api_contract scoped_path -- --nocapture`

PR #3570, opened 2026-05-13
feat(reborn): add first-party coding tools

serrrfirat:
## Summary
- add Reborn first-party coding capabilities: `builtin.read_file`, `builtin.write_file`, `builtin.list_dir`, `builtin.glob`, `builtin.grep`, `builtin.apply_patch`
- wire first-party handlers through `HostRuntime` with scoped mounts/filesystem access and read-before-edit state
- align v1 tool behavior for line-numbered reads, write/list/glob/grep output shapes, fuzzy patching, sensitive path handling, mtime sorting, and grep filtering
- expose filesystem modified/sensitive metadata across local/libSQL/Postgres/memory backends for parity-safe sorting and secret filtering
- update `FEATURE_PARITY.md`

## Verification
- `CARGO_TARGET_DIR=/tmp/ic-coding-pack cargo test -p ironclaw_host_runtime --test first_party_builtin_tools`
- `CARGO_TARGET_DIR=/tmp/ic-coding-pack cargo test -p ironclaw_host_runtime --tests`
- `CARGO_TARGET_DIR=/tmp/ic-coding-pack cargo test -p ironclaw_filesystem --all-features`
- `CARGO_TARGET_DIR=/tmp/ic-coding-pack cargo clippy -p ironclaw_host_runtime --all-targets --all-features -- -D warnings`
- `CARGO_TARGET_DIR=/tmp/ic-coding-pack cargo clippy -p ironclaw_filesystem --all-targets --all-features -- -D warnings`
- `git diff --check`


henrypark133 (changes requested):
## Review: block hidden sensitive-path access

The first-party coding tool path is broadly in good shape, and I like that the tools stay on the normal CapabilityHost -> RuntimeDispatcher -> FirstParty path with caller-level coverage. I found one security boundary issue that should be fixed before this lands.

What looks good:
- Built-in coding tools still dispatch through the Reborn capability path instead of bypassing authorization/resource accounting.
- The tests cover the host-runtime caller boundary for mount permissions, read-before-edit state, resource caps, sensitive local paths, and v1-shaped outputs.
- The libSQL filesystem contract coverage was updated for the new stat metadata.

### Critical: resolved virtual paths can hide sensitive DB-backed files
The sensitive-path guard currently checks the scoped input before mount resolution, but it does not also check the resolved virtual path. For local filesystems this is partly mitigated by `stat.sensitive`, because `LocalFilesystem::stat` canonicalizes and checks the host path. The DB-backed filesystems, however, return `sensitive: false` for every stat, so a mount/grant that aliases `/workspace` directly to a sensitive virtual target such as `/projects/p/.env` can make the scoped input look harmless while the resolved DB path is credential-shaped. That lets read/grep/list reach data the sensitive-path policy is meant to block.

Please reject sensitive resolved virtual paths as well as sensitive scoped paths, and add a caller-level regression test using a DB-backed filesystem, preferably libSQL, where `/workspace` aliases a sensitive virtual path and the coding tools fail closed.

Local verification:
- `cargo test -p ironclaw_host_runtime --test first_party_builtin_tools -- --nocapture` passed.
- `cargo test -p ironclaw_filesystem --features libsql --test db_root_filesystem_contract -- --nocapture` passed.
- `cargo test -p ironclaw_architecture reborn_ -- --nocapture` passed.

PR #3587, opened 2026-05-13
feat(reborn): add product adapter registry contracts

serrrfirat:
## Summary
- add `ironclaw_product_adapter_registry` crate with typed manifest, installation, activation, credential-binding, and health contracts
- add in-memory registry store plus contract tests for default-off behavior, explicit activation, credential binding validation, egress pair preservation, manifest hash mismatch, and redacted health
- add ProductAdapter manifest TOML ingestion with Telegram v2 fixture, strict unknown-field rejection, inline secret-material rejection, and egress credential validation
- add architecture boundary guardrail for registry crate dependencies
- keep scope narrow: no env adapter list, DB store, WASM loading, runtime routing, secret material access, or legacy channel coupling

## Testing
- `cargo test -p ironclaw_product_adapter_registry`
- `cargo clippy -p ironclaw_product_adapter_registry --all-targets -- -D warnings`


serrrfirat:

Base/head: `c1cdb27a1806bfe9c9645ca80acad088db58c8dc` → `a43794522ec69d2d4c03552f35539c1eedd12ca3`. CI is green, mergeable. Scope: new crate `ironclaw_product_adapter_registry` (contracts + TOML manifest ingestion + tests + architecture boundary rule).

Merge stance: green path. Two Medium and a few Low findings worth addressing in this slice or as immediate follow-ups; none block merge.

## Findings

| 1 | Medium | Trust boundary / invariants | `crates/ironclaw_product_adapter_registry/src/lib.rs:232-243`, `409-418` | `ProductAdapterManifest` and `ProductAdapterInstallation` use plain `#[derive(Deserialize)]`. Cross-field invariants (`validate()` — duplicate credential handles, undeclared egress, duplicate egress pairs; `validate_bindings_unique()`; manifest_ref↔adapter_id check at `:429-433`) only run via `new()`/`upsert_*`. Any future caller that reconstructs these from a serialized snapshot/DB row (round-tripped JSON/TOML) bypasses the invariants. Primitive ids in `ironclaw_product_adapters::identity` consistently implement custom `Deserialize` that runs validation — mirror that pattern here. | Implement manual `Deserialize` for `ProductAdapterManifest` and `ProductAdapterInstallation` that routes through `new(...)`, or seal the structs and expose only validated constructors. Add regression tests round-tripping a deliberately-invalid serialized form. |
| 2 | Medium | Defense-in-depth / header injection | `crates/ironclaw_product_adapter_registry/src/lib.rs:178-211` | `ProductAdapterManifestAuthDocument::{RequestSignature, SharedSecretHeader, SessionCookie}` pass `header_name`, `timestamp_header_name`, and cookie `name` straight to `AuthRequirement` (`ironclaw_product_adapters/src/auth.rs:45-59`) without syntactic validation. Every other identifier-shaped string in this crate runs through `validate_nonempty_noncontrol` or `validate_id`. A manifest with `header_name = "X-Foo\r\nInjected: x"` or a non-token cookie name reaches downstream consumers; if any of them interpolate into raw HTTP headers/cookies it's a header-injection vector. | At `into_auth_requirement`, validate `header_name`/`timestamp_header_name` as RFC 7230 tokens (ASCII visible, no CTL/whitespace/separators) and `SessionCookie.name` per RFC 6265. Add ingestion tests for CRLF, empty, and whitespace cases. |
| 3 | Low | Correctness / hash invariant | `crates/ironclaw_product_adapter_registry/src/lib.rs:793-801` | `ManifestHashMismatch` is only raised when *both* manifest and installation carry `Some(hash)`. If the registered manifest has `None` and the installation claims any `Some(...)`, or vice versa, the mismatch is silently accepted. Inconsistent with the apparent intent (`tests/registry_contract.rs:159-181` only covers same-Some / different-hash). | Either require parity (`(Some, Some)` must match; mismatched `Some/None` is also an error) or document the asymmetric semantics in `ManifestHash`. Add tests for `(None, Some)` and `(Some, None)`. |
| 4 | Low | Correctness / orphaning | `crates/ironclaw_product_adapter_registry/src/lib.rs:597-605` | `upsert_manifest` blindly replaces the registered manifest. Removing a `required_credentials` handle silently orphans existing installations binding that handle (subsequent `upsert_installation` would now reject them as `UndeclaredCredentialHandle`, but `list_installations` keeps surfacing the stale state). Likely intentional for the foundation slice; missing test or doc note. | Either re-validate existing installations on manifest upsert, or surface an explicit "manifest changes may invalidate installations" contract note and a test pinning the behavior. |
| 5 | Low | Defense-in-depth / inline-secret guard | `crates/ironclaw_product_adapter_registry/src/lib.rs:688-740` | Key-name denylist matches exact lowercased `{secret, secret_value, token, raw_token, api_key, password}`; `looks_like_inline_secret` covers `sk-`, `xoxb-`, `ghp_`, `AKIA`, `BEGIN PRIVATE KEY`, URI userinfo, and the telegram-token pattern. Common variants are missed: `client_secret`, `private_key`, `bearer_token`, `access_token`, `refresh_token`, `webhook_secret`, `oauth_token`; GitHub `gho_/ghu_/ghs_/ghr_`; Slack `xoxa-/xoxp-/xoxs-/xoxe-`; JWT `eyJ...`. Guard is intentionally best-effort but worth aligning with the documented threat (operators authoring manifests by hand). | Expand both sets, or add a doc comment that this is a tripwire not a contract. |
| 6 | Low | API asymmetry | `crates/ironclaw_product_adapter_registry/src/lib.rs:476-498` | `set_credential_bindings` is `pub`, `set_activation_state`/`set_health` are crate-private. Callers can mutate bindings on a `get_installation` snapshot and forget to `upsert_installation`; the persisted state stays consistent (store re-validates), but in-memory copies silently diverge. | Make `set_credential_bindings` `pub(crate)` (mutations go through the store), or document that snapshots are advisory until upserted. |

## Security/data-flow notes

- TOML → `ProductAdapterManifestDocument::from_toml` → `into_manifest()` → `ProductAdapterManifest::new()` → `validate()`. Inline-secret tripwire runs before schema validation. `deny_unknown_fields` is set on the document and its sub-documents. Good.
- Primitive identity wrappers (`ProductAdapterId`, `AdapterInstallationId`, `ManifestHash`, `ProductAdapterComponentRef`, `DeclaredEgressHost`, `EgressCredentialHandle`) all run validation through custom `Deserialize`. Containers do not — see finding 1.
- `ProductAdapterCredentialBinding` stores an opaque `SecretHandle`; no raw secret material in this type's surface. Good.
- `ProductAdapterHealthSnapshot::message` wraps `RedactedString`; `tests/registry_contract.rs:232-258` pins Debug-redaction. Good.

- `InMemoryProductAdapterRegistryStore::upsert_installation` re-validates against the live manifest under the write lock — race-free. Good.
- `manifest_hash` parity check is the only place that asymmetry leaks (finding 3); other cross-field checks are symmetric.
- Architecture boundary additions (`reborn_dependency_boundaries.rs`) correctly forbid wiring into runtime/secrets/host crates — matches the "contracts only, no I/O" comment at the crate's module doc.

## Missing tests

- `crates/ironclaw_product_adapter_registry/tests/manifest_ingestion.rs` — `header_name` / cookie `name` with CRLF, leading whitespace, empty string.
- `crates/ironclaw_product_adapter_registry/tests/registry_contract.rs` — `(None, Some)` / `(Some, None)` manifest hash combinations.
- `crates/ironclaw_product_adapter_registry/tests/manifest_ingestion.rs` — deserialize a hand-crafted invalid `ProductAdapterManifest` blob (duplicate credentials / undeclared egress) directly via `serde_json::from_str` to pin that direct deserialization either fails or is sealed.
- `crates/ironclaw_product_adapter_registry/tests/registry_contract.rs` — `upsert_manifest` after installations exist, removing a required credential; assert intended behavior (reject vs. surface stale state).

## Suggested fixes

1. Implement manual `Deserialize` for `ProductAdapterManifest` and `ProductAdapterInstallation` that routes through their validating constructors.
2. Add `validate_http_token`-style check on `header_name`/`timestamp_header_name` and a cookie-name check on `SessionCookie.name` in `into_auth_requirement`.
3. Tighten or document manifest-hash parity semantics.
4. Add the missing regression tests listed above.

serrrfirat:
## Findings fix follow-up

Addressed all findings from the prior review pass. CI is green at `4c3403b3841f002bb162cdbc6a1a3c34e0d09c24`.

| # | Finding | Resolution |
|---|---------|------------|
| 1 | Manifest / Installation deserialize bypass | Custom `Deserialize` impls on `ProductAdapterManifest` and `ProductAdapterInstallation` route every wire value through the validating constructors (`new(...)`) and the cross-field checks (`validate()`, `validate_bindings_unique()`, manifest-ref ↔ adapter-id parity). The `#[derive]` no longer admits `Deserialize`. Regression tests in `tests/registry_contract.rs` deserialize hand-crafted invalid blobs and assert the constructor errors propagate. |
| 2 | Auth header / cookie injection | `validate_http_token` (RFC 7230 §3.2.6) and `validate_cookie_name` (RFC 6265) run inside `ProductAdapterManifest::validate()` and inside `ProductAdapterManifestAuthDocument::into_auth_requirement()`, so both the TOML-ingestion path and the manual-Deserialize path reject CRLF/whitespace/separator-laden header and cookie names. Six new ingestion tests cover CRLF, whitespace, separator, empty, valid-token, and timestamp-header variants. |
| 3 | Asymmetric manifest-hash parity | Already addressed in the prior fix commit (`334c6cbb4`): `validate_installation_against_one_manifest` now rejects every one-sided hash combination. |
| 4 | `upsert_manifest` orphaning installations | Also addressed in `334c6cbb4`: `upsert_manifest` re-validates each existing installation against the candidate manifest under the write lock and rejects revisions that would strand or unstamp bound credentials. |
| 5 | Inline-secret denylist | Key denylist now collapses snake_case and kebab-case and matches `client_secret`, `private_key`, `bearer_token`, `access_token`, `refresh_token`, `oauth_token`, `auth_token`, `id_token`, `webhook_secret`, `apikey`, `api_secret`, `passphrase`. Value-shape guard now covers `gho_` / `ghu_` / `ghs_` / `ghr_`, `xoxa-` / `xoxp-` / `xoxs-` / `xoxe-`, `ASIA` AWS keys, and base64-shaped JWTs. Tests in `tests/manifest_ingestion.rs` exercise each new denylist entry. |
| 6 | `set_credential_bindings` asymmetry | Method removed. Binding mutation is now reachable only via `ProductAdapterInstallation::new` + `ProductAdapterRegistryStore::upsert_installation`, so the store always re-validates against the registered manifest. Two integration tests rewritten to use the constructor path. |

Commits: `301536f5a` (the deserialize-seal + auth/secret hardening) and `4c3403b38` (CI scanner fixups: inline `// safety:` on the documented-safe `.expect()`, and a clippy `collapsible_match` cleanup).

zmanian:
## Summary

This PR adds a `ironclaw_product_adapter_registry` crate with typed manifest, installation, activation, credential-binding, and health contracts; an in-memory store; TOML manifest ingestion with inline-secret rejection; and an architecture-boundary guardrail. Scope is intentionally narrow (no WASM loading, no DB, no runtime routing), the doc set (`AGENTS.md`, `CLAUDE.md`) clearly enumerates the guardrails, CI is green, and the test coverage is unusually thorough for a contracts-only crate.

## Strengths

- **Guardrails are codified, not just documented.** `forbid(unsafe_code)`, the architecture boundary test enumerating ~30 forbidden crate deps, `deny_unknown_fields` on every manifest document type, the inline-secret tripwire (key denylist + value-shape heuristics), and the RFC 7230/6265 token validators on header/cookie names form a coherent defense-in-depth stack.
- **Direct-deserialize bypass is closed.** `ProductAdapterManifest` and `ProductAdapterInstallation` intentionally do not derive `Deserialize` and route through validating manual impls. This is exactly the right pattern for invariant-bearing structs and there are explicit JSON tests pinning the bypass attempts (duplicate credentials, duplicate bindings, adapter mismatch).
- **Cross-write invariants are tested through the store, not just helpers.** `upsert_manifest` re-validates existing installations against the new manifest; `set_activation_state` re-validates on the `Enabled` transition. The contract test `upsert_manifest_rejects_when_existing_installation_binding_revoked` exercises the full caller path. This matches the project's "test through the caller" rule.
- **Symmetric hash pinning** (`(Some, None)` and `(None, Some)` both fail) is the right call and the rationale is inline.
- **Secret-handling discipline.** Credential bindings store only `SecretHandle`; health snapshots route messages through `RedactedString` and the `health_update_redacts_message` test asserts the Debug output is actually redacted.

## Issues

### Critical
None.

### Major
None.

### Minor

1. **`.expect()` in production code (`src/lib.rs:962`).** CLAUDE.md forbids `.unwrap()`/`.expect()` outside tests. The CI panic-check passed because the call sits behind a justified safety comment, and the invariant (same write lock guards both lookups) is genuinely sound — but the project rule is firm. Two clean alternatives:
   - Clone the installation for validation, then `get_mut` once, e.g. validate against `installation.clone()` before re-acquiring the mut borrow.
   - Restructure to do the manifest lookup against `&inner.manifests` while holding `&mut inner.installations.get_mut(...)` directly (split borrow on the struct fields).
   Either avoids the expect entirely. `unreachable!()` would also be acceptable per project convention if you'd rather keep the structure.

2. **`set_activation_state` always advances `updated_at` even on no-op transitions** (`Installed -> Installed`, etc.). Minor, but downstream observers may infer state churn that didn't happen. Consider short-circuiting if `state == installation.activation_state()`.

3. **`list_enabled_installations` sorts twice.** `list_installations()` already sorts by id; the second sort after filtering is redundant. Inexpensive but dead work.

4. **`ManifestHash` accepts any nonempty non-control string.** `"anything"` parses; the Telegram fixture uses the obviously-fake `"sha256:telegram-v2-fixture"`. For a contracts crate that's defensible, but it means future hash-pinning callers carry the burden of structural validation. Worth a follow-up issue or at minimum a `// TODO` noting this so the assumption isn't lost.

5. **`looks_like_inline_secret` prefix checks operate on the lowercased value, but matching `"akia"` / `"asia"` at the start of any string is broad** (e.g., `"asian-markets-adapter"` would trip the guard if ever used as a `component_ref`). The doc comment explicitly accepts false positives, and the threat model (operator-authored TOML) is reasonable, but consider tightening to require trailing alnum length matching AWS key shape (`AKIA[0-9A-Z]{16}`) if false positives surface in practice.

6. **Public surface has thin doc comments.** `RegistryError` variants, `ProductAdapterRegistryStore` trait methods, and most struct accessors lack `///` docs. Module-level doc is one paragraph. For a contracts crate that downstream crates will pin against, more thorough rustdoc would pay off — particularly on `ProductAdapterRegistryStore` (preconditions/postconditions per method) and on `RegistryError` variants (when each is raised).

7. **`secret(value: &str) -> SecretHandle` test helper uses the same string as both the credential-handle name and the secret-handle id**, which is mildly confusing because `SecretHandle` is meant to be an opaque lookup key, not the secret itself. Naming nit; nothing functionally wrong.

## Suggestions

- Once a durable store lands (libSQL/PG), add a contract-test harness that runs the same `registry_contract.rs` cases against both the in-memory and persisted backends. The dual-backend rule from CLAUDE.md will bite future PRs otherwise.
- Consider exposing a `ProductAdapterRegistryStore` blanket impl over `Arc<T>` so downstream wiring doesn't have to choose between `Arc<dyn Trait>` and concrete generics.
- The architecture boundary list is long and likely to grow. A short comment near `BoundaryRule` for this crate noting *why* each major category is forbidden (engine/runtime/dispatcher = "contracts must not depend on runtime"; secrets = "no raw material access"; v1 wasm/channels = "Reborn-only") would help future contributors not relax the rule by mistake.

## Verdict

**APPROVE** with minor follow-ups. The contracts are well-designed, the invariants are enforced at the only entry points that could violate them, the test surface is comprehensive (including direct-deserialize bypass cases), and the scope discipline (no WASM, no DB, no routing) is explicit. The `.expect()` and a couple of style points are worth a small follow-up commit but are not blocking.

henrypark133 (approved):
## Review

What looks good:
- The registry stays declarative: manifest/install state only, without runtime, network, or secret execution.
- Cross-write validation rejects mismatched manifest hashes, undeclared bindings, duplicate handles, and invalid manifest/install pairings.

No verified findings.

Low-priority notes:
- Persistence parity and full registry-load round trips are not covered yet. That is fine for this in-memory contract slice, but the future durable backend should get explicit round-trip coverage.
- It would help to document whether every `required_credentials` entry must be bound before activation, or whether bindings are intentionally optional until runtime.

Summary:
- Review coverage: worktree-backed Reborn design-boundary, product-integration, testing, CI, and prior-thread pass.
- Residual risk: future durable backend or loader paths could bypass current constructor/store invariants if they do not reuse the same validation surface.

serrrfirat:
Addressed the review notes in `5eb22172e`:

- Removed the production `.expect()` in `set_activation_state` by split-borrowing registry state.
- Added no-op activation short-circuit so unchanged state does not bump `updated_at`.
- Removed redundant enabled-installation sorting.
- Tightened AWS inline-secret detection to exact `AKIA`/`ASIA` key shape and added false-positive coverage for `asian-markets-adapter`.
- Added rustdoc for registry errors, store contract/methods, and public accessors.
- Added `Arc<T>` blanket impl for `ProductAdapterRegistryStore`.
- Renamed the confusing test secret helper and made secret handle ids distinct from credential handles.
- Added boundary-rule comments explaining why runtime/secrets/v1-WASM dependencies stay forbidden.

Verification:
- `cargo test -p ironclaw_product_adapter_registry`
- `cargo fmt --check --package ironclaw_product_adapter_registry --package ironclaw_architecture`
- `cargo clippy -p ironclaw_product_adapter_registry --all-targets -- -D warnings`
- `rg -n '\\.expect\\(' crates/ironclaw_product_adapter_registry/src` (no output)\n\nMerge/conflict check: local index has no unmerged paths; GitHub now reports `mergeable: MERGEABLE`.

crates/ironclaw_product_adapter_registry/src/lib.rs
```diff
@@ -0,0 +1,1233 @@
...
+}
+
+/// Host credential binding for an adapter-declared egress handle.
+///
+/// This stores an opaque [`SecretHandle`] only. It must never carry raw secret
+/// material.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct ProductAdapterCredentialBinding {
+    credential_handle: EgressCredentialHandle,
+    secret_handle: SecretHandle,
```
serrrfirat, line 437: **Medium — credential binding accepts secret-shaped `SecretHandle` values without tripwire**

`ProductAdapterCredentialBinding` accepts any `SecretHandle` and derives `Debug`/`Serialize`. `SecretHandle` only validates name-segment shape (lowercase alphanumeric + `_`/`-`/`.`), so a secret-shaped lowercase token like `sk-live-reborn-secret-sentinel` can be accepted as a `secret_handle`, stored in registry state, and emitted through debug/JSON serialization.

This violates the crate guardrail in CLAUDE.md: "Raw secret material must never be stored, accepted, serialized, or logged."

**Suggested fix:** Validate `secret_handle.as_str()` with the same inline-secret value tripwire used in manifest parsing (`looks_like_inline_secret`), either in `ProductAdapterCredentialBinding::new` or in `ProductAdapterInstallation::new` + the custom `Deserialize` impl + store upsert path. Add a regression test that constructing a binding with `secret_handle = "sk-live-reborn-secret-sentinel"` is rejected.

PR #3592, opened 2026-05-13
fix(reborn-secrets): harden secret store (H1, H3, M1, M2) + regression guards

ilblackdragon:
## Summary

Hardens the Reborn secrets path against four findings from the 2026-05 security review. Each fix is a separate commit paired with a failing-on-purpose PoC test that flips to passing in the same commit; the PoCs land as a standalone first commit so the bugs are documented before any fix.

- **H1** — runtime credential cache held leased plaintext as `Option<String>` for the lifetime of the egress request, defeating `SecretMaterial::ZeroizeOnDrop`. Cache now holds `Option<SecretMaterial>` and the egress loop borrows `&str` only for the narrow injection + redaction-token window.
- **H3** — AES-GCM ciphertext in `reborn_secret_records` was not bound to `(user_id, name)`. An attacker with DB write could swap `(encrypted_value, key_salt)` between rows and `get_decrypted` silently returned the other row's plaintext. Threaded AAD through `SecretsCrypto::encrypt`/`decrypt` with four domain-tagged builders covering all four encrypted-at-rest tables.
- **M1** — `CredentialSessionId::Display` rendered the raw UUID despite `Debug` being redacted, leaking the bearer-like value through every `format!("{id}")` interpolation. Display now redacts; durable backends call a narrow `expose_for_storage()`.
- **M2** — `SecretsCrypto::new` accepted 32 bytes of zeros and other trivially low-entropy master keys. Added a constant-time-on-length distinct-byte check (≥ 8) that rejects pathological inputs while accepting all legitimate hex/base64 keys.

The fifth PoC (**H2**, staged-secret HashMap rehash) was a misanalysis on closer inspection — `SecretBox<String>` already provides the heap indirection I claimed was missing. Test kept as a regression guard so a future refactor that erases the indirection fails the assertion.

## Test plan

- [x] `cargo test -p ironclaw_secrets --features libsql` (37 tests, includes 3 PoCs)
- [x] `cargo test -p ironclaw_host_runtime` (200+ tests, includes 2 PoCs)
- [x] `cargo test -p ironclaw_reborn --features libsql-secrets`
- [x] `cargo test -p ironclaw_architecture` (boundary tests)
- [x] `cargo check --workspace` (no warnings introduced in this PR)
- [x] `cargo clippy -p ironclaw_secrets --features libsql --tests --all-targets`

PoC tests fail on the pre-fix commit (`ed92972ba`) when run with `-- --ignored`; subsequent fix commits each flip exactly one ignore off and the test passes.

## Out of scope

Deferred from the same review — tracked in a separate comment on this PR.

## No migration needed

Per design discussion: `reborn-integration` has no live data yet, so the AAD change in H3 ships without a re-encryption migration path.


ilblackdragon:
## Deferred items from the 2026-05 secrets review

Not in this PR. Listed here so they don't get lost and so reviewers can weigh in on priority order.

### Architectural / bigger blast radius

- **H4 — `SecretStore` trait split.** `put()` is documented as a "trusted setup primitive" but is on the same public trait as `lease_once`/`consume`/`metadata`, and `Arc<dyn SecretStore>` is plumbed through composition. Any code path that gets a clone of that `Arc` can write arbitrary `(scope, handle, material)`. The "trust" is purely an organizational invariant. Fix: split into `SecretReader` (lease/consume/metadata) and `SecretAdmin: SecretReader` (put), and only hand `Admin` to setup/migration paths.

- **D1 — `RuntimeHttpEgressRequest::credential_injections` is a `#[derive(Deserialize)]` field.** The doc comment says it's authority-bearing and must come from host composition. The type system says anyone with a constructed value can set it. Fix: split into `RuntimeHttpEgressRequestFromGuest` (no injections, the wire-decodable shape) and `AuthorizedRuntimeHttpEgressRequest` (the only thing the egress fn accepts).

- **D2 — Scope-as-token.** Every `SecretStore` op takes `ResourceScope` and trusts it. There is no signed/unforgeable handle binding the call to an *authorized* scope; a compromised internal caller can pass any scope. Defense in depth: mint short-lived `ScopedSecretCapability` tokens from the authorization layer and require `consume(token, lease_id)` rather than `consume(scope, lease_id)`.

### Performance / observability

- **H5 — `block_on_secret` builds a fresh tokio runtime per credential injection.** `crates/ironclaw_host_runtime/src/lib.rs:1004-1031` spawns a thread + `Runtime::new_current_thread()` per `SecretStoreLease` injection. Hundreds of µs of overhead per egress call, several FDs, and a real DoS surface on high-throughput credential injection. The async-in-sync bridge usually hides a wrong boundary; egress should already be on a tokio task.

### Lesser items

- **M3 — `metadata()` enumeration without auth at store layer.** The contract pushes auth upstream; the trait method takes any `scope` and returns presence/absence. Hardening would be a phantom-typed `AuthorizedScope<'_>` constructed only by the authorization layer.

- **M4 — Unbounded lease / staged-secret growth.** `ScopedSecretsStoreAdapter::leases` and `RuntimeSecretInjectionStore::secrets` are unbounded `HashMap`s with 5-min TTL. An authorized caller can `lease_once` in a loop and force memory growth + O(n) scans on every consume/revoke/list. Add a per-scope cap.

- **M5 — `record_usage` rollback after consume.** `ScopedSecretsStoreAdapter::consume` returns material, then calls `record_usage`. On `record_usage` failure the lease is rolled back to `Active` and a subsequent consume succeeds — even though the first call already fetched the plaintext (just returned `Err` to the caller). Worth a fault-injection test that asserts rollback is semantically equivalent to "lease was never consumed".

- **M6 — Audit/stage atomicity.** A panic or partial obligation chain between audit emit and dispatch can leave staged secret material in `RuntimeSecretInjectionStore` without a paired audit record, until TTL expires. Currently cleaned up only on the capability lifecycle path. Fault-injection test needed.

- **M7 — `RuntimeCredentialTarget::QueryParam` re-parse.** `apply_credential_injection` parses `request.url`, appends a pair, then `url.to_string()` and replaces. Policy was matched against the pre-injection URL; the post-injection URL is what reaches `NetworkPolicy`. If `Url`'s round-trip canonicalization disagrees with the policy parser, post-injection could pass policy but reach a different effective destination. Property test: `policy.matches(pre) ⟹ policy.matches(post)`.

- **SecretLeaseId Display.** Same shape as the M1 finding for `CredentialSessionId` — Display emits the raw UUID. Less bearer-like in isolation because scope is required as a second factor on consume, but defense-in-depth says redact and add a narrow expose helper.

### What was withdrawn

- **H2 — `RuntimeSecretInjectionStore` HashMap rehash + zeroization.** Original claim: rehash bitwise-copies plaintext to the new bucket array and frees the old without zeroize. Wrong — `SecretMaterial = SecretBox<String>` already has the `Box` indirection, so rehash only moves a pointer and the actual `SecretString` buffer stays heap-stable until proper drop. The regression guard test in `obligations.rs::security_findings_poc` locks that property in.

### Priority order I'd suggest if these get picked up

1. H4 (trait split) — kills H4 itself plus structurally reduces the surface for future bugs of the same shape.
2. D1 + D2 (authority-as-type) — same flavor; together with H4 they make the "who can do what to secrets" question type-driven instead of comment-driven.
3. H5 (block_on_secret) — perf cliff today, will become an incident under load.
4. M5 + M6 (atomicity) — narrow, well-defined fault-injection work.
5. M3 + M4 — defense in depth.
6. M7 + SecretLeaseId Display — small mechanical changes.

crates/ironclaw_secrets/src/crypto.rs
```diff
@@ -99,3 +132,67 @@ impl std::fmt::Debug for SecretsCrypto {
...
+/// cannot be confused with the framing), and the domain tag prevents
+/// cross-shape replay (a credential-account ciphertext cannot be replayed as
+/// a secret-record ciphertext, etc.).
+pub(crate) fn build_aad(domain: &[u8], parts: &[&[u8]]) -> Vec<u8> {
+    let capacity = domain.len() + parts.iter().map(|part| 4 + part.len()).sum::<usize>();
+    let mut aad = Vec::with_capacity(capacity);
+    aad.extend_from_slice(domain);
+    for part in parts {
+        let length =
+            u32::try_from(part.len()).expect("AAD part length must fit in u32 for framing");
```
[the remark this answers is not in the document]
ilblackdragon: Fixed in a121f3a3 — switched the length prefix from `u32-be` to `u64-be` so the `usize → length` conversion is infallible on every supported platform (`usize` is at most 64 bits). The `.expect()` is gone and there's no DoS path through user-chosen part contents.

The AAD-binding format was introduced in this same PR (bd6e375), so there is no on-disk ciphertext to remain compatible with.

crates/ironclaw_secrets/src/db.rs
```diff
@@ -1196,8 +1248,9 @@ impl SecretsStore for LibSqlSecretsStore {
         name: &str,
     ) -> Result<DecryptedSecret, SecretError> {
         let secret = self.get(user_id, name).await?;
+        let aad = secret_record_aad(user_id, &secret.name);
         self.crypto
-            .decrypt(&secret.encrypted_value, &secret.key_salt)
+            .decrypt(&secret.encrypted_value, &secret.key_salt, &aad)
```
serrrfirat, line 1253: **Medium (conditional) — persisted secret rows become unreadable if any pre-AAD data exists**

This decrypt path now requires AES-GCM AAD derived from `(user_id, name)`. That fixes row-swap attacks, but it also means any `reborn_secret_records` rows encrypted by the previous code path (no AAD) will fail with `DecryptionFailed` during bootstrap verification or first read, even with the correct master key.

PR body says `reborn-integration` has no live data, so this is OK only if that invariant is release-blocking and true for every environment. If not, add versioning/fallback decrypt/re-encryption migration plus an upgrade test that writes a pre-AAD row and verifies post-upgrade readability.

crates/ironclaw_secrets/src/db.rs
```diff
@@ -1959,36 +2028,63 @@ impl TryFrom<PersistedCredentialSession> for CredentialSession {
...
     crypto: &SecretsCrypto,
     encrypted_value: &[u8],
     key_salt: &[u8],
+    scope: &ResourceScope,
+    session_id: CredentialSessionId,
 ) -> Result<CredentialSession, CredentialBrokerError> {
+    let aad = credential_session_aad(scope, session_id);
     let persisted: PersistedCredentialSession =
-        decrypt_json_payload(crypto, encrypted_value, key_salt)?;
+        decrypt_json_payload(crypto, encrypted_value, key_salt, &aad)?;
```
serrrfirat, line 2067: **Medium (conditional) — credential account/session payloads also need the no-live-data guarantee**

Credential account/session JSON payloads are now decrypted with scope/session-bound AAD. Existing rows written by the previous `encrypt_json_payload` format had no AAD, so stored accounts/sessions become unreadable after this change.

If the branch truly has no persisted Reborn credential data, no migration is needed. If any developer/staging/customer-like DB can carry `reborn_credential_accounts` or `reborn_credential_sessions`, add a compatibility path or migration and a regression test that reads rows produced by the old format.

serrrfirat:
## Summary
Reviewed PR #3592 only. Base `5d8c252620f6db8190e9c7067bc0d88f733403b8` → head `564c93368fa679c37408f42cccdb72c24d3ae904`.

PR hardens Reborn secret store. Direct security fixes look good: secret material cache hardening, AAD binding, session-id redaction, weak-key rejection, and libSQL/Postgres symmetry. Merge stance: one rollout/migration compatibility risk remains.

## Findings

| 1 | Medium | Migration / Rollout compatibility | `crates/ironclaw_secrets/src/db.rs:139`, `crates/ironclaw_secrets/src/db.rs:1212`, `crates/ironclaw_secrets/src/db.rs:1360`, `crates/ironclaw_secrets/src/db.rs:2039`, `crates/ironclaw_secrets/src/db.rs:2058` | Decrypt paths now require new AAD for secrets, credential accounts, and credential sessions, but migration/startup paths do not migrate or fail fast on pre-AAD rows. `verify_can_decrypt_existing_secrets()` exists only for secret records and is opt-in; credential-account/session tables have no equivalent compatibility gate. Any non-empty pre-PR DB, rolling deploy, or rollback where an old writer inserts legacy ciphertext can become latent `DecryptionFailed` at first read/consume instead of deterministic startup failure or migration. | Call compatibility verification from production startup/migration, add format/version gating or migration for credential account/session payloads, and add regression with pre-AAD rows present at startup. If deployment contract is “fresh Reborn DB only, no mixed-version access,” encode that invariant in startup/docs and fail fast when violated. |

## Security/data-flow notes
- Lease cache stores `SecretMaterial`, not raw `String`.
- `CredentialSessionId` display redacts; persistence uses narrow private accessor.
- AAD binding added for secret records, credential accounts, credential sessions, and key-check sentinel.
- No changed-path auth/scope bypass found.

- Row-swap defense looks correct for newly written encrypted payloads.
- Risk is compatibility for existing/mixed-version encrypted payloads, especially credential account/session tables.

## Missing tests
- Startup behavior with pre-AAD credential account/session rows.
- Mixed-version rollout/rollback writer scenario, if supported.

crates/ironclaw_secrets/src/crypto.rs
```diff
@@ -100,3 +141,74 @@ impl std::fmt::Debug for SecretsCrypto {
...
+    for part in parts {
+        let length = part.len() as u64;
+        aad.extend_from_slice(&length.to_be_bytes());
+        aad.extend_from_slice(part);
+    }
+    aad
+}
+
+pub(crate) const AAD_DOMAIN_SECRET_RECORD: &[u8] = b"reborn/v1/secret_record";
+pub(crate) const AAD_DOMAIN_CREDENTIAL_ACCOUNT: &[u8] = b"reborn/v1/credential_account";
```
serrrfirat, line 175: `AAD_DOMAIN_CREDENTIAL_ACCOUNT` and `AAD_DOMAIN_CREDENTIAL_SESSION` are declared unconditionally, but they are only used from `db.rs`, which is compiled behind `#[cfg(any(feature = "libsql", feature = "postgres"))]`. Default-feature builds now emit `dead_code` warnings (`cargo check -p ironclaw_secrets`, and transitively `cargo check -p ironclaw_host_runtime`), which breaks the repo's warning-clean expectation and the PR's no-warnings claim. Please move these credential-domain constants into the cfg-gated DB code or cfg-gate the declarations.

serrrfirat (commented):
Posted one warning-clean issue from local verification.

crates/ironclaw_secrets/src/crypto.rs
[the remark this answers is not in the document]
ilblackdragon: Confirmed already addressed in commit `a121f3a3` — the length prefix is now `u64-be`, so the `usize → length` conversion is infallible on every supported platform and the `.expect()` is gone. No DoS path through user-chosen part contents remains.

crates/ironclaw_secrets/src/db.rs
ilblackdragon: Relying on the no-live-data invariant on `reborn-integration` — no compatibility/migration path. To lock the decision in, commit `f5b77412` adds `aad_binding_rejects_pre_aad_secret_record_ciphertext` in `tests/security_findings_poc.rs`: ciphertext encrypted with empty AAD must be rejected (not silently accepted) by the new AAD-bound decrypt path, so a write-side adversary cannot downgrade rows to the legacy format. If a pre-AAD row ever did surface, bootstrap would fail closed with `DecryptionFailed` rather than returning unauthenticated plaintext.

crates/ironclaw_secrets/src/db.rs
ilblackdragon: Same answer as the `reborn_secret_records` thread — relying on the no-live-data invariant on `reborn-integration`. Commit `f5b77412` adds row-swap regression coverage for both the credential account and session tables: `h3_libsql_credential_account_row_swap_must_not_return_other_accounts_payload` and `h3_libsql_credential_session_row_swap_must_not_return_other_sessions_payload` in `tests/security_findings_poc.rs`. Each writes two rows, swaps `(encrypted_payload, payload_key_salt)` between them, and asserts the AAD-bound decrypt fails closed. Together with the pre-AAD rejection test on the secret-record path, this locks in the design decision: any pre-fix row will fail at decrypt rather than silently downgrade.

crates/ironclaw_secrets/src/crypto.rs
ilblackdragon: Fixed in commit `f5b77412` — both `AAD_DOMAIN_CREDENTIAL_ACCOUNT` and `AAD_DOMAIN_CREDENTIAL_SESSION` are now gated behind `#[cfg(any(feature = "libsql", feature = "postgres"))]`, matching the only call site in `db.rs`. Verified `cargo check -p ironclaw_secrets` is now warning-clean on default features and that both `--features libsql` and `--features postgres` still compile + tests pass.

PR #3596, opened 2026-05-13
arch(ws-6a): add canonical loop executor

henrypark133:
## Context

Canonical executor foundation. This is the branch that turns the strategy/family/planner model into a reusable loop tick.

Workstream brief: `docs/reborn/agent-loop-briefs/canonical-executor.md`
Stack base: `arch/level2-merged`

## What landed

- `CanonicalAgentLoopExecutor` and `AgentLoopExecutor::execute_family(...)` entry point.
- Executor flow for prompt request, model call, capability batching, checkpoint boundaries, gate/recovery/stop decisions, and final loop exits.
- Batch policy derived from visible capability calls rather than hardcoded behavior.
- Checkpoint staging calls through host ports and the WS0 state payload format.
- Model/capability error classification that preserves fail-closed behavior for unavailable credentials and host errors.
- Executor tests and support wiring without depending on product-specific runtime composition.

## Reviewer focus

- The executor should be the only owner of canonical loop mechanics.
- Strategies return decisions; the executor sequences host side effects and checkpoint boundaries.
- All durable effects must go through `AgentLoopDriverHost` ports.
- Credential/auth/gate states must not be flattened into generic chat failure handling.

## Non-goals / deferred work

- Runner-facing `AgentLoopDriver` adapter lands in WS7.
- Cross-crate test support and E2E coverage land in WS8.
- Real capability/checkpoint/input/progress/cancellation/identity host adapters land in WS9 through WS15.

## Validation

- [x] `cargo check -p ironclaw_agent_loop`
- [x] Included in final ancestry verification: `arch/level2-merged -> arch/ws-6a`.

## Stack position

```
```


crates/ironclaw_agent_loop/src/executor.rs
```diff
@@ -0,0 +1,1397 @@
...
+                stage: HostStage::Input,
+            })?;
+        let consumed = batch.inputs.iter().any(|input| {
+            matches!(
+                input,
+                LoopInput::UserMessage { .. } | LoopInput::Steering { .. }
+            )
+        });
+        if consumed {
+            host.ack_inputs(batch.next_cursor.clone())
```
serrrfirat: **High Severity — control inputs can be acked/lost during drain**

`drain_user_inputs` acks through `batch.next_cursor` whenever the polled batch contains any user-facing input. That means a batch like `[Cancel, UserMessage]` sets `consumed = true` and acks the entire cursor range without handling the earlier `Cancel`. Same issue exists in `drain_followup` at the later `ack_inputs` call.

This violates the canonical executor brief invariant that a control event before the next steering/follow-up message wins and the cursor must not advance past it.

Suggested fix: partition only the contiguous user-facing prefix for each drain mode, stop before the first control event, and add mixed-batch tests such as `[Cancel, UserMessage]` and `[Interrupt, FollowUp]`.

crates/ironclaw_agent_loop/src/executor.rs
```diff
@@ -0,0 +1,1397 @@
...
+        }),
+    }
+}
+
+fn model_error_class(error: &AgentLoopHostError) -> Option<ModelErrorClass> {
+    match error.kind {
+        AgentLoopHostErrorKind::Unavailable => Some(ModelErrorClass::Unavailable),
+        AgentLoopHostErrorKind::Internal => Some(ModelErrorClass::Internal),
+        AgentLoopHostErrorKind::BudgetExceeded => Some(ModelErrorClass::ContextOverflow),
+        AgentLoopHostErrorKind::Cancelled => Some(ModelErrorClass::Transient),
```
serrrfirat: **Medium Severity — model cancellation is treated as transient retry**

`AgentLoopHostErrorKind::Cancelled` maps to `ModelErrorClass::Transient`, so a host/model cancellation enters recovery rather than cancellation handling. With the default recovery strategy, this will retry first and can eventually abort as `LoopFailureKind::ModelError`; `AgentLoopExecutorError::Cancelled` appears unreachable from this path.

Suggested fix: handle `Cancelled` specially in `stream_model_with_recovery` (return cancellation or a cancellation `LoopExit` once supported) and add a targeted test that `host.stream_model` returning `Cancelled` does not retry or surface as model failure.

serrrfirat:
## Summary
Reviewed WS6a PR #3596 only.  
Base `255c8c0db65d5960526bba961bef3dcefd36e187` → head `c936938e8968c3c2dbc444a3f8c322ad679f9ff5`.

Canonical executor has blocking cancellation/semantic-error handling issues, plus test suite does not compile.

Validation:
- `cargo check -p ironclaw_agent_loop` ✅
- `cargo test -p ironclaw_agent_loop --lib` ❌: `executor.rs:1136` missing `LoopPromptBundle.instruction_fingerprint` in test `MockHost` initializer.

## Findings

| 1 | High | Cancellation / Correctness | `crates/ironclaw_agent_loop/src/executor.rs:266-287`, `crates/ironclaw_agent_loop/src/executor.rs:851-856` | `AgentLoopHostErrorKind::Cancelled` is classified as `ModelErrorClass::Transient`, so model cancellation enters recovery/retry instead of terminal cancellation. Capability paths also collapse host errors to `HostUnavailable` at `executor.rs:377-388` and `executor.rs:514-520`. User/host cancellation can keep work alive after cancel and never return the executor’s `Cancelled` variant. | Plumb `Cancelled` through all host call sites as terminal cancellation. Add model + capability cancellation tests asserting no retry/no extra side effects and correct checkpoint behavior. |
| 2 | High | Retry / Correctness | `crates/ironclaw_agent_loop/src/state/slots.rs:12-21`, `crates/ironclaw_agent_loop/src/strategies/recovery.rs:140-194`, `crates/ironclaw_agent_loop/src/executor.rs:262-264`, `crates/ironclaw_agent_loop/src/executor.rs:510-520` | Inherits #3556 global retry-counter bug, and executor does not reset `recovery_state` after successful progress. Earlier unrelated transient failures can exhaust retry budget for later model/capability errors, causing premature abort. | Scope retry attempts by failure class/episode, and reset recovery state after successful model/capability progress. Add mixed-error and success-then-error regression tests. |
| 3 | Medium | Recovery / Boundary | `crates/ironclaw_agent_loop/src/executor.rs:377-388`, `crates/ironclaw_agent_loop/src/executor.rs:514-520` | Capability host semantic errors (`StaleSurface`, `InvalidInvocation`, `PolicyDenied`, `Cancelled`) are erased into `HostUnavailable`. Normal race/policy/cancel outcomes become generic infra failures and bypass clean loop-level handling. | Classify capability host errors explicitly and route stale-surface/policy/cancel through typed recovery/cancel paths. |
| 4 | Medium | Tests / Regression coverage | `crates/ironclaw_agent_loop/src/executor.rs:1136-1142` | `cargo test -p ironclaw_agent_loop --lib` fails to compile because test `MockHost` builds `LoopPromptBundle` without `instruction_fingerprint`. Executor regression suite currently cannot run. | Add `instruction_fingerprint: None` or expected fingerprint in test fixture. |

## Security/data-flow notes
- Main risks are cancellation and semantic boundary collapse, not raw secret leakage.
- Checkpoint and side-effect ordering need cancellation-specific tests before relying on executor safety.

## Missing tests
- Cancellation from model and capability ports: no retry, no extra side effects.
- Stale surface from capability batch: typed handling, not `HostUnavailable`.
- Mixed error classes and success-reset retry budget.
- Fix compiling executor tests.

zmanian:
## Review notes — WS6a canonical executor

Architecture is solid; checkpoint protocol is WS10-ready. Four items to address before un-drafting.

### 1. Dead `AgentLoopExecutorError::Cancelled` variant

The variant is declared but never constructed. Meanwhile, host `Cancelled` errors are mapped to `ModelErrorClass::Transient` at `executor.rs:861` — so a host cancel becomes a retried model error. This conflicts with WS13's cancellation port intent. Either remove the variant or wire a check (between iterations, before model stream). Reconcile with WS13 (#3648) — that PR's boundary checks need somewhere to land their `Cancelled` signal in this enum.

### 2. Stringly-typed capability error classification

`capability_error_class` / `capability_failure_kind` dispatch on `match kind { "transient" => ... }` over `CapabilityFailure::error_kind: String`. A typo in any producer silently maps to `Permanent`/`CapabilityProtocolError`. Fail-closed correctness depends on this classification — replace with an enum in `ironclaw_turns` and have producers emit it directly.

### 3. `opaque_token` char-replace collision

`LoopGateRef::new(format!("gate:process-{}", opaque_token))` silently mangles disallowed chars to `-`. Risk of gate-ref collision across two distinct processes that share a stripped form. Use a hash, not a char-replace.

### 4. `failed_exit` discards `_state`

Final state is captured via `checkpoint`, but the in-memory `LoopExecutionState` is dropped — `recent_failure_kinds`, `recovery_state`, etc., live only inside the staged checkpoint payload. If a caller wants the in-memory final state for diagnostics this is unreachable. Either return state in the exit or document the intent.

### Checkpoint contract observations

Solid for WS10:
- Four kinds (`BeforeModel`, `BeforeSideEffect`, `BeforeBlock`, `Final`) with clear semantics.
- 2-step stage→commit returning `(checkpoint_id, state_ref)` is exactly the shape WS10 resume needs.
- `PolicyDenied` short-circuits retry even on `RecoveryOutcome::Retry` — correct.

One non-atomicity note: `checkpoint()` mutates `state.last_checkpoint` *before* `stage_checkpoint_payload` commits. If staging fails the in-memory state carries a marker for an absent payload. Today this is moot (executor consumes state on `CheckpointFailed`), but if a caller catches and retries the marker is stale. Document or move the marker mutation after successful host commit.

### Minor

- `MAX_MODEL_RETRIES=8`, `MAX_CAPABILITY_RETRIES=8`, `MAX_INPUT_DRAIN=32` are spec constants worth pulling out of executor.rs and citing the spec section.
- `Box<LoopExecutionState>` in `ModelStep::Response` is a clippy `large_enum_variant` boxing — fine, but explicit `#[allow]` with rationale would be cleaner.
- `denied_calls` invokes recovery sequentially per-denied-call — a 10-call all-denied batch invokes the recovery strategy 10 times. Note in spec or batch the denial path.

Test surface (10+ executor-level tests through `execute_family`) genuinely satisfies "test through the caller."

henrypark133:
WS6a follow-up pushed in `7de21b2a1`.

Addressed the remaining review items I could verify locally:
- model-host cancellation now exits as `AgentLoopExecutorError::Cancelled` before recovery retry classification.
- user-facing input drains only ack whole batches that contain no control inputs, so mixed `[cancel, message]` / `[interrupt, follow-up]` batches are left unacked for the control path.
- retry accounting is cleared after successful model progress and successful capability-result progress, with caller-level executor tests covering both paths.
- process gate refs now derive from a blake3 digest of the full `LoopProcessRef` instead of a lossy sanitized token.
- production panic fallbacks in completed/failed exit construction were removed in the prior commit.

Local validation on this head:
- `cargo test -p ironclaw_agent_loop executor --lib`
- `python3 scripts/check_no_panics.py --base arch/level2-merged --head HEAD`
- `cargo fmt --check`
- `git diff --check`
- `cargo test -p ironclaw_agent_loop --lib`
- `cargo clippy -p ironclaw_agent_loop -p ironclaw_loop_support -p ironclaw_turns -p ironclaw_reborn --all-targets --all-features -- -D warnings`

crates/ironclaw_agent_loop/src/executor.rs
henrypark133: Fixed in WS6a: the executor now only acknowledges a user-facing input cursor when the whole returned batch is user-facing for that drain mode. Mixed batches with control inputs, including cancel before user message and interrupt before follow-up, keep the original cursor and do not call ack_inputs. Added caller-level tests for both paths.

crates/ironclaw_agent_loop/src/executor.rs
henrypark133: Fixed in WS6a: `stream_model_with_recovery` now checks `AgentLoopHostErrorKind::Cancelled` before recovery classification and returns `AgentLoopExecutorError::Cancelled`; `model_error_class` no longer maps cancellation into a retryable class. Added an executor test that proves cancellation returns after one model request without retry.

Issue #2582, opened 2026-04-17
[Bug Bash 4/17] Agent claims markdown file attached but no file is provided


joe-rlo:
### Environment

hosted-staging (crab shack)

### Version / Commit Hash

staging-2026-04-17

### QA Test Date

2026-04-17

### Feature Area

Telegram channel → file attachments

### Steps to Reproduce

1. Connect Telegram bot to agent
2. Request the agent to create and send a markdown file (e.g. "Based on this link https://api.worldbank.org/v2/country/all/indicator/SP.POP.TOTL?format=json create md file and send me as .md file")
3. Wait for the agent response
4. Check for attached file

### Expected Behavior

The agent should include the actual .md file as an attachment or provide a valid downloadable file.

### Actual Behavior

The agent responds with a message indicating that a markdown file is attached, but no file is present in the response. The agent falsely claims success on file creation/attachment. Related to #2279 (closed) and #2580 which track similar false-success patterns.

### Logs / Screenshots

Discovered during Bug Bash 4/17 session (UC17: Send attachments via Telegram).

### Pre-submit checklist

- [x] Title is specific (not "fix Google Suite" but "Google Suite install throws invalid_grant on OAuth step")
- [x] Commit hash is filled in
- [x] Steps are numbered and reproducible

Issue #2580, opened 2026-04-17
[Bug Bash 4/17] Agent claims Telegram message sent but no message is delivered


joe-rlo:
### Environment

hosted-staging (crab shack)

### Version / Commit Hash

staging-2026-04-17

### QA Test Date

2026-04-17

### Feature Area

Telegram channel → message delivery

### Steps to Reproduce

1. Connect Telegram bot to agent via Settings → Telegram
2. Send a message in the web UI requesting the agent to send a Telegram message (e.g. "Send me a hello message on Telegram")
3. Observe the agent's response in the web UI
4. Check the Telegram app for the message

### Expected Behavior

The agent should successfully deliver the message to the connected Telegram chat. The message should appear in the Telegram app.

### Actual Behavior

The agent responds in the web UI stating "Message has been sent to your Telegram" (or similar confirmation), but no message actually arrives in the Telegram app. The agent falsely reports success despite the message never being delivered. Observed across multiple use cases during the 4/17 bug bash session. Related to #2279 (closed) which tracked a similar false-success pattern with shell/open tool errors.

### Logs / Screenshots

Discovered during Bug Bash 4/17 session (UC1: Send message via Telegram). Agent confirms delivery but Telegram app shows no incoming message.

### Pre-submit checklist

- [x] Title is specific (not "fix Google Suite" but "Google Suite install throws invalid_grant on OAuth step")
- [x] Commit hash is filled in
- [x] Steps are numbered and reproducible

joe-rlo:
**Bug Bash 4/20 update:** Telegram notifications from missions are still not being delivered on staging (v0.25.0). Missions that include a "send via Telegram" step report success in the UI, but no message arrives in the Telegram app. This is now compounded by the mission_create null cooldown_secs issue (#2741) which blocks new mission creation entirely. The linked PR #2714 (review-driven guidance for Claude Code) may address the false-success pattern but the underlying delivery failure needs separate attention.

PR #3599, opened 2026-05-14
fix(engine): reject FINAL after only failed tool calls (#2544)

ilblackdragon:
## Summary

- Adds a structural evidence-drift gate to the CodeAct orchestrator (`crates/ironclaw_engine/orchestrator/default.py`) that fires at every `FINAL` extraction point — text, code, and tool_calls.
- Rule: **reject iff `failure_count > 0 AND success_count == 0`** — i.e. the thread attempted at least one tool, every attempt failed, and the model is now trying to claim success. Transitions the thread to `Failed` with an "Action not performed" error rather than surfacing the misleading FINAL.
- No NLP / verb lists / negation classification on either the FINAL text or the user message. The gate reads two cumulative counters maintained as a side effect of normal step processing.
- Regression coverage in a new integration test file `crates/ironclaw_engine/tests/tool_evidence_drift.rs` — three scenarios from #2544 plus a sanity test that legitimate completions still pass.

## Trade-offs called out in the gate's comment block

| Scenario | Counters | Caught? |
|---|---|---|
| `N≥1` attempts, all failed | failure>0, success=0 | **yes** |
| 0 attempts (no-await, or informational) | failure=0, success=0 | no |
| `N≥1` attempts, ≥1 succeeded | failure≥0, success≥1 | no — by design |
| Partial: 4 fail, 1 success, model cites the failures | failure=4, success=1 | no — structural rule can't see citation mismatch |

The no-await case (model writes `result = send_message(...)` without `await`, Monty produces a coroutine, no host dispatch happens) is preserved as `no_await_tool_call_then_lying_final` with `#[ignore]` — the test still fails when run with `--ignored`, so it's a live regression anchor for whoever picks up the next iteration. Catching it requires either intent classification on the user message or AST inspection of the emitted code, both rejected for v1.

## Why structural, not classification-based

An earlier iteration of this change classified the FINAL string for past-tense side-effect verbs (`sent`, `saved`, `installed`, ...) with a negation suppressor (`tried to`, `couldn't`, ...). That approach is rejected as too brittle and English-only. The current rule operates only on counters the orchestrator already maintains — single source of truth, no vocabulary maintenance.

The prompt-side guidance in `crates/ironclaw_engine/prompts/codeact_postamble.md` ("Claims in FINAL() need tool evidence") remains the primary line of defense; this gate is the structural safety net.

## Test plan

- [x] `cargo test -p ironclaw_engine --test tool_evidence_drift` → 3 passed, 1 ignored
- [x] `cargo test -p ironclaw_engine --test tool_evidence_drift -- --ignored` → 1 failed as expected (the no-await regression anchor)
- [x] `cargo test -p ironclaw_engine` → 526 passed, no regression from new counter tracking
- [x] `cargo clippy -p ironclaw_engine --all-targets --all-features -- -D warnings` → clean
- [x] `cargo fmt -p ironclaw_engine -- --check` → clean
- [x] `cargo check --workspace --all-features` → workspace compiles

## References

- Issue: #2544 (and related: #2580, #2582, #2541, #2447)
- Rule: `.claude/rules/tool-evidence.md` ("Engine v2 Side-Effect Gate (target invariant)")
- Prompt guidance: `crates/ironclaw_engine/prompts/codeact_postamble.md` ("Claims in `FINAL()` need tool evidence")

### DIFF
--- a/crates/ironclaw_engine/orchestrator/default.py
+++ b/crates/ironclaw_engine/orchestrator/default.py
@@ -65,6 +65,94 @@ def extract_final(text):
     return None
 
 
+# ── Evidence drift gate (issue #2544) ───────────────────────
+#
+# A FINAL() that follows a string of failed tool attempts and zero
+# successes is almost always claim/evidence drift — the model failed,
+# couldn't recover, and narrated success anyway. Tracked by
+# #2544 / #2580 / #2582 and documented in
+# `.claude/rules/tool-evidence.md` and `prompts/codeact_postamble.md`
+# ("Claims in FINAL() need tool evidence").
+#
+# v1 is deliberately structural — no NLP classification of FINAL text
+# or user messages. The rule:
+#
+#     gate fires iff failure_count > 0 AND success_count == 0
+#
+# That is: the thread attempted at least one tool, every attempt
+# failed, and the model is now trying to FINAL. Allow:
+#
+#   • 0 attempts, 0 successes — informational answer, no commitment.
+#   • N attempts, ≥1 success — at least one tool did its job; the
+#     model can describe the outcome.
+#
+# Reject:
+#
+#   • N attempts, 0 successes — there is no successful tool result to
+#     anchor any "I did X" claim to.
+#
+# Trade-offs called out explicitly:
+#
+#   • The "no-await" failure mode (the model writes the tool name but
+#     forgets `await`, so nothing is dispatched) registers as 0
+#     attempts and slips past this rule. Catching that requires
+#     either intent detection on the user message or AST inspection
+#     of the emitted code — both out of scope for the structural v1.
+#   • Partial-success patterns ("called 5 tools, 1 succeeded, claim
+#     references the 4 failures") are allowed by this rule. A
+#     per-claim-to-action-result binding would catch those; see
+#     the prompt-side guidance in codeact_postamble.md.
+#   • On drift: emit `evidence_drift_rejected`, transition the thread
+#     to Failed, and surface "Action not performed" with the
+#     offending FINAL attached for trace/audit.
+
+
+def _record_step_action_results(state, action_results):
+    """Update cumulative success/failure counters used by the evidence
+    drift gate. Called after every step that produces ActionResults
+    (code branch and tier-0 batched tool calls)."""
+    success = state.get("_evidence_success_count", 0)
+    failure = state.get("_evidence_failure_count", 0)
+    for r in action_results or []:
+        if r is None:
+            continue
+        if r.get("is_error", False):
+            failure += 1
+        else:
+            success += 1
+    state["_evidence_success_count"] = success
+    state["_evidence_failure_count"] = failure
+
+
+def _maybe_evidence_drift_error(state, final_text):
+    """Return the user-safe error message to surface as a Failed
+    outcome when the thread has only ever failed at the tool layer
+    but a FINAL is being emitted anyway. Returns None to allow normal
+    completion."""
+    success_count = state.get("_evidence_success_count", 0)
+    failure_count = state.get("_evidence_failure_count", 0)
+    if success_count > 0 or failure_count == 0:
+        return None
+    __emit_event__(
+        "evidence_drift_rejected",
+        success_count=success_count,
+        failure_count=failure_count,
+    )
+    # `final_text` is unused for the rejection message itself, but the
+    # parameter is kept so call sites still pass it for traces — the
+    # offending FINAL is preserved in the `response` slot of the
+    # outcome dict regardless.
+    _ = final_text
+    return (
+        "Action not performed: "
+        + str(failure_count)
+        + " tool call(s) failed, 0 succeeded. The agent cannot "
+        + "report a successful outcome with no successful tool "
+        + "result behind it. See .claude/rules/tool-evidence.md "
+        + "(issue #2544)."
+    )
+
+
 def strip_quoted_strings(line):
     """Remove double-quoted string literals from a line."""
     result = []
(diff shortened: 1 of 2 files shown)


crates/ironclaw_engine/orchestrator/default.py
```diff
@@ -65,6 +65,94 @@ def extract_final(text):
...
+    state["_evidence_success_count"] = success
+    state["_evidence_failure_count"] = failure
+
+
+def _maybe_evidence_drift_error(state, final_text):
+    """Return the user-safe error message to surface as a Failed
+    outcome when the thread has only ever failed at the tool layer
+    but a FINAL is being emitted anyway. Returns None to allow normal
+    completion."""
+    success_count = state.get("_evidence_success_count", 0)
```
serrrfirat, line 132: **High Severity** — evidence counters are thread-global, so prior success can mask a later failed-only claim.

The gate reads cumulative counters from persisted `state`, but I don't see them scoped/reset for a new user request or obligation window. Since this orchestrator state persists across resumes, a prior successful tool call in the same thread leaves `_evidence_success_count > 0`. If the user later asks for a new side effect, that tool fails, and the model emits `FINAL("sent")`, `_maybe_evidence_drift_error` returns `None` because of the older success and the unsupported claim is allowed.

This mirrors why `_obligation_resolved` / `_obligation_nudge_count` are reset on new execution intent later in `run_loop`.

Suggested fix: scope these counters to the current user turn / current action-obligation window / since last completion, and add a regression where an earlier successful tool is followed by a new failed-only side-effect attempt plus a lying `FINAL`.

crates/ironclaw_engine/orchestrator/default.py
```diff
@@ -65,6 +65,94 @@ def extract_final(text):
...
+        "evidence_drift_rejected",
+        success_count=success_count,
+        failure_count=failure_count,
+    )
+    # `final_text` is unused for the rejection message itself, but the
+    # parameter is kept so call sites still pass it for traces — the
+    # offending FINAL is preserved in the `response` slot of the
+    # outcome dict regardless.
+    _ = final_text
+    return (
```
serrrfirat, line 146: **Medium Severity** — honest failure `FINAL(...)` responses are rejected too.

The new rule rejects every `FINAL(...)` after failed-only tool attempts, not only fabricated success claims. Existing recovery guidance later tells stuck agents to "call FINAL() with an honest explanation of what failed", but with this gate a truthful terminal answer like `FINAL("I couldn't send it; transport closed")` becomes a generic `"Action not performed"` failure instead of preserving the useful explanation.

Suggested fix: either preserve honest failure final text while marking the outcome failed, or narrow hard rejection to success-claim / intent-scoped finals. Please add a regression for failed tool call followed by an honest failure `FINAL`.

crates/ironclaw_engine/tests/tool_evidence_drift.rs
```diff
@@ -0,0 +1,523 @@
...
+///
+/// Iteration 1: the model emits Python that calls `send_message(...)`. The
+/// tool returns `Err(EngineError::Effect{...})`, which the scripting layer
+/// turns into a `RuntimeError` inside Monty — the script aborts and `FINAL`
+/// is never extracted. Iteration 2: the model emits a fresh code block
+/// that calls `FINAL("Message sent successfully!")` without re-trying the
+/// failed action. The desired behavior is that the engine refuses to
+/// surface that claim as a successful completion.
+#[tokio::test]
+async fn multi_iteration_tool_failure_then_lying_final() {
```
serrrfirat, line 290: **Medium Severity** — coverage only exercises the code-response path, but the PR changes all FINAL extraction paths.

These new tests drive `LlmResponse::Code` only. The implementation also changed behavior for text FINALs and structured `tool_calls` FINALs, and the structured path has the trickiest ordering (filter FINAL, execute non-FINAL calls, record evidence, then complete). A regression in those branches would not fail this suite.

Suggested fix: add caller-level regressions for:
- text response containing `FINAL(...)` after failed-only evidence,
- structured `tool_calls` `FINAL` after failed-only evidence,
- prior successful tool in the same thread followed by a new failed-only attempt to lock the counter scoping invariant.

serrrfirat:
## Summary
Reviewed PR #3599 only (draft). Base `34eeeaf05a656007927afe0b5c6929ae75c3f866` → head `b05dd70fcccc889b7ca7cc10e66db69564221ea6`.

PR closes the core “failed tool then fabricated FINAL” path, but gate semantics still have correctness gaps. Merge stance: not clean yet.

## Findings

| 1 | Medium | Correctness / State invariants | `crates/ironclaw_engine/orchestrator/default.py:110`, `crates/ironclaw_engine/orchestrator/default.py:132` | Evidence counters are cumulative thread-level state. After any earlier successful tool call, a later resumed/injected user request can still do only failed tools then fabricate `FINAL`; gate allows it because old `_evidence_success_count > 0`. This leaves #2544 class open across multi-turn/resume epochs. | Scope evidence counters to request/obligation/post-user-message epoch, and reset on fresh user work injection/resume. Add regression: earlier success, new request with only failed tools, then `FINAL` must reject. |
| 2 | Medium | Correctness / Tool failure semantics | `crates/ironclaw_engine/orchestrator/default.py:119`, `crates/ironclaw_engine/orchestrator/default.py:1208`, `crates/ironclaw_engine/src/executor/structured.rs:536` | Structured tool-call path records every `is_error=true` result as evidence failure before checking gate/auth interrupts. Approval/auth pauses are encoded as `ActionResult { is_error: true, output.status = "gate_paused" }`, but they are not executed tool failures. This can poison failure counters and later reject valid `FINAL` after resume. | Exclude non-executed interrupt sentinels (approval/auth/gate-paused/resume-wait shapes) from evidence failure counters. Add approval/auth pause → resume regression. |
| 3 | Medium | Tests / Caller-boundary coverage | `crates/ironclaw_engine/tests/tool_evidence_drift.rs:8`, `crates/ironclaw_engine/tests/tool_evidence_drift.rs:164` | Test header says gate covers text, code, and structured `tool_calls` FINAL extraction points, but live tests build only `LlmResponse::Code` via `code_response(...)`. Changed behavior is installed in multiple exit paths; tests currently prove only code branch. | Add caller-boundary regressions for text `FINAL(...)` and structured `tool_calls` FINAL after failed tools, plus interrupt/resume semantics. |

## Security/data-flow notes
- No new secret/auth leak found.
- Main risk is semantic evidence poisoning and lifetime state reuse across requests.

- Gate invariant currently implemented as “reject iff failures > 0 and successes == 0”.
- That invariant needs epoch-local evidence, not lifetime thread totals.

## Missing tests
- Text FINAL after failed tool.
- Structured tool_calls FINAL after failed tool.
- Approval/auth pause should not count as tool failure.
- New user request after old success should not inherit old success evidence.

Issue #3132, opened 2026-05-01
Mission creation fails - error': '\'cooldown_secs\' must be an integer, got "120"'


sergeiest:
[ACTION FAILED] mission_create: {'error': '\'cooldown_secs\' must be an integer, got "120"'}

<img width="1239" height="289" alt="Image" src="https://github.com/user-attachments/assets/11fb7250-c127-4f86-a018-1cffd698abe9" />

PR #3601, opened 2026-05-14
test(e2e): workflow regression coverage + #2544 behaviour test + ceo recording

ilblackdragon:
## Summary

- Adds e2e Playwright coverage for the user-reported workflow regressions, the nearai/ironclaw#2544 behaviour contract, and a live-LLM recording of the executive-onboarding workflow.
- Fixes three real issues caught by the recording work: an orphaned i18n key from PR #3364, an unhelpful `tool_install` error string, and weak live-LLM proxy miss diagnostics.

## What landed

**New e2e tests**

| File | What it pins |
|---|---|
| `tests/e2e/scenarios/test_recent_workflow_regressions.py` | Multi-step tool chain (PR #3157/#3365), single-tool round trip (PR #3364 http default), mission/routine creation with the goal naming an uninstalled tool (PR #3132/#3155/#3166), and approval-card clarity copy (PR #3364/#2991). 4 tests, all pass. |
| `tests/e2e/scenarios/test_assistant_executes_not_plans_2544.py` | Live-LLM record/replay behaviour contract for nearai/ironclaw#2544 — the assistant must actually do the work and quote a real result back, not narrate a plan. Pinned via `echo` + `memory_write` so the trace is hash-stable. Pass in record AND replay. |
| `tests/e2e/scenarios/test_ceo_onboarding_workflow.py` | Full executive onboarding → meeting reflection → install pump workflow against `Qwen/Qwen3.5-122B-A10B` on NEAR AI. 19-entry recording committed. **Skipped in replay** (see follow-up below). |
| `TestV2EngineCodeActSyntaxErrorRecovery` (in `test_v2_engine_error_handling.py`) | Pins that a Python parse failure inside a CodeAct snippet must not hold a previously-raised `github_token` Authentication gate hostage. |

**Three issue fixes surfaced by the recording work**

1. PR #3364 `approval.description` was rewritten in en/ko/zh-CN locales but never read by the frontend JS. Wired as a `.approval-subtitle` card element; CSS added; the http-approval test locks the wiring in.
2. `tool_install` "not callable in this execution context" — added `uncallable_action_error()` in `crates/ironclaw_engine/src/executor/structured.rs` with a special case for `tool_install` / `tool_auth` that tells the LLM to direct the user at Settings → Extensions instead of relaying a technical sentence. Tier 0 + Tier 1 (CodeAct) share the helper.
3. `live_llm_proxy.py` miss-diagnostic sidecars now embed the miss canonical AND the closest recorded canonical (by tail length) plus a diagnostic tip, so future replay drift is a one-step diff away from a fix.

## Follow-up

The CEO onboarding workflow test runs against a live LLM and the recording passes end-to-end, but the proxy's replay path drifts past Stage A because engine-generated state inside `role: tool` payloads (non-UUID identifiers, HashMap-order enumeration, embedded relative timestamps) isn't normalised by the canonicalizer. The follow-up to make engine tool outputs byte-identical across runs — so this and future multi-stage live tests replay deterministically in CI without a skip — is captured inline in the test docstring with the three known drift sources we've already identified.

## Status

```
test_recent_workflow_regressions.py     4 passed
test_assistant_executes_not_plans_2544  1 passed (replay, 2s)
test_ceo_onboarding_workflow            1 skipped (record-mode only by design)
test_v2_engine_error_handling           3 passed (2 existing + 1 new)
```

8 passed, 1 skipped, 14s.

## Test plan

- [x] `pytest tests/e2e/scenarios/test_recent_workflow_regressions.py` — 4/4
- [x] `pytest tests/e2e/scenarios/test_assistant_executes_not_plans_2544.py` (replay) — 1/1
- [x] `pytest tests/e2e/scenarios/test_ceo_onboarding_workflow.py` — 1 skipped, recording committed
- [x] `pytest tests/e2e/scenarios/test_v2_engine_error_handling.py` — 3/3
- [x] `cargo check -p ironclaw_engine` — clean


tests/e2e/scenarios/test_ceo_onboarding_workflow.py
```diff
@@ -0,0 +1,658 @@
...
+    #     are not UUID-shaped (so `_strip_dynamic` misses them)
+    #   - workspace `memory_tree` / `memory_list` results enumerate in
+    #     HashMap iteration order, not a stable sort
+    #   - some tool outputs embed wall-clock timestamps that aren't
+    #     RFC 3339 (e.g. relative ages, durations)
+    # The improved miss-diagnostic sidecar (`live_llm_proxy.py::_replay`,
+    # added alongside this test) writes the closest recorded canonical
+    # next to the miss canonical so the next debugging pass on this
+    # follow-up has a one-step diff to start from.
+    if ceo_workflow_live_server["mode"] != "record":
```
serrrfirat, line 427: **Medium Severity — Test value: replay-skipped means zero CI signal**

This whole test is `pytest.skip`-ed in replay mode, which is the only mode CI ever runs. So the 658 LOC + the committed 3791-line fixture (`test_ceo_onboarding_to_meeting_reflection.json`) provide no regression guard until the canonicalizer follow-up (engine `role: tool` payload normalization for non-UUID ids, HashMap iteration order, embedded relative timestamps) actually lands.

Suggested fix — pick one:

1. **Block the fixture commit on the canonicalizer fix.** Land the `_strip_dynamic` extension first, re-record against the deterministic engine output, then commit a fixture that actually replays. Otherwise this is dead weight in-tree.
2. **Split out a Stage-A-only variant** that *does* replay deterministically (the bootstrap + ceo-setup turn happens before any mission/child-thread state leaks into the role-tool payloads). Run that in CI today, keep the full workflow as a record-mode smoke until the canonicalizer is ready.

As-is, future readers will hit a 4.5k-line fixture for a test that never runs and assume the workflow is covered.

tests/e2e/mock_llm.py
```diff
@@ -1555,6 +1564,52 @@ def match_special_response(messages: list[dict], has_tools: bool) -> dict | None
...
+    if _conversation_has_user_trigger(messages, CODEACT_SYNTAX_ERROR_TRIGGER):
+        # Detect whether the assistant has already emitted the first
+        # github-calling CodeAct in this conversation. If so, switch to
+        # the broken-CodeAct response.
+        prior_assistant_repl = any(
+            msg.get("role") == "assistant"
+            and "```repl" in _message_text(msg)
+            for msg in messages
+        )
+        if not prior_assistant_repl:
```
serrrfirat, line 1591: **Medium Severity — Mock trigger is sticky-on, fragile under retry**

`_conversation_has_user_trigger` scans **all** user messages in history, so once the trigger phrase lands in turn 0 it's matched on every subsequent LLM round. The branch selector `prior_assistant_repl` (any prior assistant turn containing `` ```repl ``) flips after the first turn and stays flipped — so every LLM call from turn 1 onward emits the broken-CodeAct response.

Consequences:

1. If the engine ever retries on parse failure (more than the current loop bound), the mock keeps feeding the same unparseable snippet forever and the test only passes by hitting the 60s wall-clock deadline. That's a flake waiting for a retry-policy change.
2. The "second turn" branch is really "≥ second turn" — the contract being tested ("parse failure releases stale auth gate") is independent of how many parse failures the engine sees, so the framing in the comments is misleading.

Suggested fix: gate on the **most recent** assistant turn rather than any-prior assistant repl, or use a turn counter (e.g. count `role: assistant` messages and return broken-CodeAct only when count == 1). That makes the mock's state machine match its docstring: "turn 0 = valid github call, turn 1 = unclosed string". Today it's "turn 0 = valid, turn N≥1 = broken forever."

serrrfirat:
**Cross-PR pattern — live-LLM fixtures committed without replay rent**

Flagging this once at the PR level rather than per-file, because it's a pattern I've now seen land across multiple PRs and this one continues it.

This PR commits a **3791-line** `test_ceo_onboarding_to_meeting_reflection.json` (≈61% of the PR's `+6223` total additions) for a test that is explicitly `pytest.skip`-ped in the only mode CI ever runs (replay). The PR body and the test's inline docstring openly acknowledge why: the proxy canonicalizer doesn't normalize non-UUID ids, HashMap iteration order, or embedded relative timestamps inside `role: tool` payloads — three drift sources we've already identified as follow-ups in earlier PRs too.

The pattern across the suite:

1. A live-LLM workflow gets recorded once against a real model.
2. The recording is committed as "snapshot evidence" even though it can't replay deterministically.
3. The canonicalizer fix is deferred to a follow-up.
4. The fixture sits in-tree, large and inert, until the follow-up lands (which, for these three drift sources, has been deferred across multiple PRs now).

Each individual fixture is defensible in isolation — but cumulatively we're accumulating ≈thousands of lines of JSON that no CI run touches, and signaling "this workflow is covered" to readers who don't dig into the skip clause.

Two concrete asks for the next iteration of this kind of work:

- **Land `_strip_dynamic` (or whichever normalization layer covers role-tool payloads) before adding new record-only fixtures.** Then the new fixtures earn their place by actually running.
- **Or, split tests so the deterministically-replayable prefix runs in CI** (e.g. Stage A of the CEO workflow — bootstrap + ceo-setup runs before mission/child-thread state leaks into role-tool payloads, and should hash-stable).

Secondary pattern in this PR specifically: `mock_llm.py` continues to grow stateful triggers matched via "any user message in history" semantics (`_conversation_has_user_trigger`). That's fine for single-shot triggers but already starting to bite for state-machine triggers — see the inline comment on `CODEACT_SYNTAX_ERROR_TRIGGER`. A `current-turn-only` matcher utility would prevent the next instance.

Not blocking this PR — production code in it is small and tight. Logging the pattern so it's visible.

serrrfirat:
## Summary
Reviewed PR #3601 only. Base `34eeeaf05a656007927afe0b5c6929ae75c3f866` → head `3a758230a100099c1469a19890a013aa5b207d03`.

PR adds E2E workflow regression coverage plus #2544 behavior test. #2544 test direction is good, and no obvious committed secret leak found in fixture. Merge stance: CEO workflow regression coverage does not prove its claimed install/auth boundary and is skipped from normal replay/CI.

## Findings

| 1 | High | Test correctness / Boundary invariant | `tests/e2e/scenarios/test_ceo_onboarding_workflow.py:318`, `tests/e2e/scenarios/test_ceo_onboarding_workflow.py:337`, `tests/e2e/scenarios/test_ceo_onboarding_workflow.py:553`, `tests/e2e/scenarios/test_ceo_onboarding_workflow.py:565`, `tests/e2e/conftest.py:1784`, `tests/e2e/fixtures/live/test_ceo_onboarding_to_meeting_reflection.json:3738` | Stage C can pass without proving the agent installed `google_drive`. The fixture pre-stages `google_drive.wasm`; helper `_drive_oauth_to_completion()` calls `/api/extensions/{ext}/setup` and `/oauth/callback` whenever the extension entry already exists; then Stage C asserts installed+authenticated. The committed recording shows the model's `tool_install` calls failed and the assistant told the user install must happen in UI. So the assertion proves “test authenticated a preloaded extension”, not “chat path installed it” or “hidden install boundary behaved correctly.” | Split invariants. Assert before Stage C that extension is not active/installed if Stage C is meant to prove install. Then require evidence of supported install transition from extension state/event history before OAuth setup/auth. Or make this a UI-install/auth test and stop claiming agent-installed boundary coverage. |
| 2 | Medium | CI reliability / Coverage gap | `tests/e2e/scenarios/test_ceo_onboarding_workflow.py:403`, `tests/e2e/scenarios/test_ceo_onboarding_workflow.py:427` | CEO workflow test unconditionally skips unless running live `record` mode. The committed 19-entry recording never runs in default replay/CI, so this PR does not add merge-gating regression coverage for the bootstrap/install/auth workflow it claims to cover. | Make replay deterministic enough for CI, or extract the install/auth/hidden-tool invariants into hermetic mock-backed tests that always run. Keep live recording as optional canary only. |

## Security/data-flow notes
- Checked fixture for obvious real token/API-key leakage; no blocker found.
- Production hidden-tool invariant looks intended: `tool_install`/`tool_auth` are not callable from chat. The test issue is that helper side effects bypass that invariant instead of proving it.

- #2544 behavior coverage appears useful: it checks actual tool calls/workspace artifact/terminal answer.
- CEO workflow currently mixes live-record documentation with test helper side effects, so caller-boundary evidence is weak.

## Missing tests
- Hermetic hidden-tool test: model cannot call `tool_install` from chat, and extension cannot become installed/authenticated unless supported install path actually ran.
- Replayable CEO workflow seam test in CI.

Issue #3607, opened 2026-05-14
[Reborn WebUI Beta] Owner-module tracker and release path


serrrfirat:
# Reborn WebUI Beta — owner-module tracker

This is the coordination issue for the WebUI-first Reborn beta path. Pick work directly from the module sections below.

Published architecture view: https://shiny-serenity-yfe8.here.now/

## Operating decision

- Beta product surface: **WebUI / WebChat v2**.
- Telegram/Slack/WeChat/channel ports are **not** on the beta critical path.
- WASM ProductAdapter runtime is **not** required for beta.
- Mutations should use HTTP. Browser progress/timeline should use EventStreamManager-backed streaming: SSE first, WS later only if needed.

## Pick-up rules

- Pick from **P0** in your module first.
- Do not start P2/deferred work unless a module owner explicitly asks.
- Cross-module contracts must be DTOs, facades, and tests; not direct imports or raw handle sharing.
- If an issue spans modules, split it or create a child issue and link it here.
- When someone takes ownership, update assignment on the issue and comment here if module sequencing changes.

## Beta acceptance target

```text
open WebUI
-> create thread
-> send message
-> Reborn loop runs
-> optional first-party capability call
-> final assistant reply streams
-> refresh page
-> timeline replays correctly
-> restart service
-> thread state still correct
```

## M1 — WebUI Product Surface

**Owner scope:** Browser-facing WebChat v2 route contract, frontend/API integration, auth/security parity, HTTP mutations.

**Boundary:** May call WebUI-facing RebornServices only. Must not call TurnCoordinator, HostRuntime, dispatcher, run-state stores, or DB stores directly.

### P0 — pick up first
- [#3578 — Define host-owned ingress boundary for Reborn HTTP surfaces](https://github.com/nearai/ironclaw/issues/3578)
- [#3580 — Port WebUI / Web Gateway to native Reborn surface](https://github.com/nearai/ironclaw/issues/3580)
- [#3611 — Implement minimal native WebChat v2 routes](https://github.com/nearai/ironclaw/issues/3611)

### P1 — parallel/hardening
- [#3615 — Audit WebUI auth and security parity for Reborn routes](https://github.com/nearai/ironclaw/issues/3615)
- [#3127 — Design scalable capability permission UX and policy resolver](https://github.com/nearai/ironclaw/issues/3127)
- [#3287 — Migrate memory and workspace product surfaces](https://github.com/nearai/ironclaw/issues/3287)
[...]

PR #3632, opened 2026-05-14
feat(reborn): add before-inbound policy seam

### LINKED ISSUE #3623: [Reborn WebUI Beta] Add BeforeInboundPolicy seam for ProductWorkflow
Parent: #3607  
Splits beta-critical slice from #3280.

## Goal

Add only the WebUI-beta-critical `BeforeInboundPolicy` seam to `ironclaw_product_workflow` so WebChat v2 user messages can be policy-checked/rejected/rewritten before accepted-message staging.

## Why separate

#3280 still contains non-beta ProductWorkflow dispatch work (approval/auth/mission/system/linked-thread/subscription ports). #3607 beta only needs the safe user-message path for:

```text
WebUI -> create thread -> send message -> Reborn loop runs -> final reply streams
```

This issue isolates the required pre-accept policy hook without pulling in the deferred #3558 scope.

## Scope

- Add `BeforeInboundPolicy` port and request/outcome types.
- Add no-op default policy for back-compat.
- Add fake policy support for contract tests.
- Wire `DefaultProductWorkflow` to run policy before `InboundTurnService` receives/stages a user message.
- Policy rejection returns a product-safe rejected ack and does not call the inbound-turn path.
- Policy rewrite passes rewritten content to the inbound-turn path.
- Add focused contract tests for no-op, rewrite, and rejection behavior.

## Explicitly deferred

- `ApprovalInteractionService`
- `AuthInteractionService`
- `MissionService`
- `LinkedThreadActionService`
- `SystemActionService`
- `ProjectionSubscriptionAuthority`
- `MissionActionPayload` / `SystemActionPayload`
- Full command compatibility matrix (#3286)
- Production WebUI route cutover (#3611/#3612)

## Acceptance criteria

- [ ] User-message policy runs before accepted-message staging.
- [ ] Rewritten content is what reaches the accepted-message path.
- [ ] Rejected input creates no accepted transcript content or turn submission.
- [ ] Existing `DefaultProductWorkflow::new(...)` callers retain no-op behavior.
- [ ] Product-safe rejection output contains no raw prompts, tool input, secrets, host paths, backend diagnostics, or provider/runtime output.
- [ ] Targeted product-workflow tests pass.

serrrfirat:
## Summary

Closes #3623. Beta-critical slice for #3607 / #3280.

Adds a minimal `BeforeInboundPolicy` seam to `ironclaw_product_workflow` so WebUI/WebChat v2 user messages can be checked, rewritten, or rejected before accepted-message staging.

## What changed

- Add `BeforeInboundPolicy`, `BeforeInboundPolicyRequest`, `BeforeInboundPolicyOutcome`, and `NoopBeforeInboundPolicy`.
- Wire `DefaultProductWorkflow::with_before_inbound_policy(...)` while preserving existing `new(...)` no-op behavior.
- Run policy before `InboundTurnService::accept_user_message`.
- Add `ProductInboundEnvelope::with_rewritten_user_message(...)` to preserve host-stamped context while swapping policy-rewritten user payload.
- Add `ActionDispatchKind::Rejected { kind }` so policy rejections are not recorded as `NoOp`.
- Add `FakeBeforeInboundPolicy` and contract tests for allow/rewrite/reject/transient failure behavior.

## Explicitly deferred

No approval/auth/mission/system/linked-thread/subscription ports in this slice. Those remain outside the WebUI beta happy path.

## Verification

- `cargo test -p ironclaw_product_adapters --lib`
- `cargo check -p ironclaw_product_workflow --no-default-features`

Note: full `cargo test -p ironclaw_product_adapters` still hits existing integration-test feature/type-inference issues unrelated to this change, so this PR verifies the touched adapter library surface with `--lib`.


serrrfirat:
Addressed paranoid review feedback in 6c-pushed commit:

- **#1 (Medium)** — Added `before_inbound_policy_retryable_rejection_releases_fingerprint` contract test asserting retryable rejections release the fingerprint and re-invoke the policy on resubmit.
- **#2 (Low)** — `dispatch_kind_from_ack` now consumes the rewritten `envelope_for_turn.payload()` instead of the original payload.
- **#3 (Low)** — Added `tracing::debug!` event on the policy-reject path including rejection kind and disposition.
- **#4 (Low)** — Documented the timeout/disposition contract on the `BeforeInboundPolicy` trait (callers responsible for bounded execution; disposition controls fingerprint settle vs release).
- **#5 (Low)** — Added `before_inbound_policy_rewrite_replays_rewritten_outcome_on_duplicate` contract test confirming duplicate replay does not re-invoke the policy or inbound turn service.

Skipped #6 (nit, perf) — `BeforeInboundPolicyRequest` continues to own clones for the beta slice; the borrowed variant is noted as a follow-up.

Re-verified locally:
- `cargo test -p ironclaw_product_workflow` — 20 tests pass.
- `cargo test -p ironclaw_product_adapters --lib` — 45 tests pass.
- `cargo check -p ironclaw_product_workflow --no-default-features` — clean.

serrrfirat:
## Summary
Reviewed PR #3632 only. Base `19f08a5609b7207d0b752cdac6826e5e80d1a887` → head `6d62362dfe4e19e70feb9252150e67e628b48350`.

PR adds a before-inbound policy seam. Merge stance for reviewed security/correctness scope: no actionable findings.

## Findings

| — | — | — | — | No actionable security/correctness findings in reviewed scope. | — |

## Security/data-flow notes
- Policy runs before inbound turn staging/submission in `ProductWorkflow`.
- Rewrite path preserves trusted adapter/install/auth/event/conversation stamps and only swaps user-message payload.
- Permanent reject settles duplicate replay; retryable reject and transient policy failure release idempotency reservation.

- No success ack returned before durable terminal ledger outcome in reviewed paths.
- Duplicate replay does not re-run policy or turn submission after durable outcome.
- GitHub checks observed passing for clippy, fmt/style, gateway syntax/boundaries, and no-panics.

## Missing tests
- None material in reviewed scope; added caller-boundary tests cover rewrite, permanent reject, retryable reject, duplicate replay, and transient failure retry behavior.

henrypark133:
## Caveman review findings

**1🔴 3🟡 1❓**

- 🔴 `crates/ironclaw_product_workflow/src/workflow.rs` (~line 556) `dispatch_payload`: calls `inbound_turn_service.replay_accepted_user_message(envelope)` and, on `None`, proceeds to call `accept_user_message(envelope_for_turn)` — but `DefaultInboundTurnService::accept_user_message` (~line 348) *also* calls `self.replay_accepted_user_message(envelope)` internally before proceeding. Every non-replay new inbound message probes `thread_service.replay_accepted_inbound_message` twice per dispatch. Doubles read traffic to the thread store on the hot inbound path. Fix: either remove the `self.replay_accepted_user_message()` call from `accept_user_message` (workflow now owns the gate) or do not call it from `dispatch_payload` and let `accept_user_message` retain ownership.
- 🟡 `workflow.rs` (~line 574): when `BeforeInboundPolicyOutcome::RewriteUserMessage(payload)` fails `with_rewritten_user_message` validation, the error maps to `ProductWorkflowError::TurnSubmissionRejected`. That variant gives `None` from `terminal_ack_for_error`, then falls through to `idempotency_ledger.release(action)` — fingerprint released, delivery retried. A permanent invalid-rewrite (policy always producing oversized payload) is retried indefinitely. Fix: map validation error to a new variant with terminal `ProductInboundAck::Rejected(ProductRejection::permanent(...))` in `terminal_ack_for_error`, or settle the action immediately.
- 🟡 `crates/ironclaw_product_workflow/src/policy.rs:27`: `BeforeInboundPolicyRequest` clones the entire `ProductInboundEnvelope` (auth claim, attachments, all refs) into the request struct passed to every policy implementation. Envelope is also borrowed by `dispatch_payload` for the rest of the call. Full clone of a large struct on every inbound message; over-exposes trusted host-stamped fields (auth claim, adapter identity) to policy code. Fix: restrict request to the fields policies should act on, or pass `&ProductInboundEnvelope` by reference.
- 🟡 `crates/ironclaw_wasm_product_adapters/src/runner.rs:956`: retryable-rejection → `WorkflowTransient` conversion uses a single-arm `if ack.retry_disposition() == InboundRetryDisposition::Retry` against a three-value enum. `Duplicate` correctly maps to `DoNotRetry` today, but a future fourth `InboundRetryDisposition` variant silently falls through to `Acknowledged`. Fix: explicit `match` on `retry_disposition()` for exhaustiveness.
- ❓ `policy.rs:35`: doc says "Callers are responsible for enforcing wall-clock timeouts on any production policy implementation". Workflow holds an in-flight idempotency fingerprint while the policy runs and no timeout is enforced at the workflow call site (not in `dispatch_payload`, not in `DefaultProductWorkflow::accept_inbound`). Is the timeout intentionally deferred to the production policy constructor, or should the workflow enforce a budget here analogous to `workflow_timeout` in `NativeProductAdapterRunner`?

serrrfirat:
Addressed the review feedback in `6932f05b5`:

- Added `InboundTurnService::accept_user_message_after_policy` so the workflow can probe replay once before before-inbound policy, then submit after policy without a second replay lookup.
- Hardened `ProductInboundEnvelope::with_rewritten_user_message` so it only rewrites existing user-message envelopes and returns a malformed-payload error otherwise.
- Added regression coverage for both paths.

Validation run locally:
- `cargo test -p ironclaw_product_adapters --lib`
- `cargo clippy -p ironclaw_product_adapters -p ironclaw_product_workflow --all-targets --all-features -- -D warnings`

Security pass after fixes found no remaining issues in the changed inbound policy/rewrite path.

ilblackdragon:
## Code Review

### Overview

Adds a `BeforeInboundPolicy` seam to `ironclaw_product_workflow` so WebUI/WebChat v2 inbound user messages can be **allowed**, **rewritten**, or **rejected** before accepted-message staging. Closes #3623 as part of the beta slice for #3607 / #3280. +840 / −50, 9 files.

Key shape:
- `BeforeInboundPolicy` trait with three outcomes (`Allow`, `RewriteUserMessage(payload)`, `Reject(rejection)`)
- `DefaultProductWorkflow::with_before_inbound_policy(...)` builder; default is `NoopBeforeInboundPolicy` so existing call sites are unaffected
- `ProductInboundEnvelope::with_rewritten_user_message(...)` with re-validation
- `InboundTurnService` grows two new methods (`replay_accepted_user_message`, `accept_user_message_after_policy`) with default implementations
- `ActionDispatchKind::Rejected { kind: ProductRejectionKind }` so policy denials don't get logged as `NoOp`
- WASM runner converts a retryable rejection ack into a retryable error so the webhook protocol re-delivers

### Strengths

- **Backwards-compatible by construction.** `Arc::new(NoopBeforeInboundPolicy)` is wired in `new(...)`; the new `with_before_inbound_policy` builder is purely additive. Existing callers compile and behave identically.
- **Replay-before-policy is the correct ordering.** Already-staged messages must not be re-rejected by a policy change that happened after staging. `dispatch_payload` checks `replay_accepted_user_message` first and returns the replay outcome without invoking policy. The integration test `before_inbound_policy_does_not_block_staged_deferred_retry` proves this end-to-end: first call gets `DeferredBusy`, second call has a policy that now rejects, but the replay outcome wins.
- **Policy-rewritten payload is re-validated.** `UserMessagePayload::new` is split into a `Self { ... }` + `validate()` shape, and `with_rewritten_user_message` calls `payload.validate()` before substitution. A policy that returns oversized text fails *before* hitting `accept_user_message_after_policy` (test `before_inbound_policy_rewrite_revalidates_payload_before_turn_path`).
- **Permanent vs retryable rejection semantics are precise.** Permanent rejection settles the ledger so subsequent attempts get `Duplicate`; retryable rejection does *not* settle, so subsequent attempts re-invoke the policy. Both paths tested.
- **`should_settle_ack` is a real semantic change, not a rename.** The old `is_terminal_success_ack` would have wrongly classified retryable rejections as terminal. The new function correctly gates settle on `!DeferredBusy && ack.is_durable_outcome()`.
- **WASM runner closes the loop.** A retryable rejection now propagates as `WorkflowTransient`, which surfaces as a retryable error to the protocol layer. Without this, the WASM adapter would 200-OK the webhook on a retryable rejection and the protocol would never redeliver. Test `process_webhook_surfaces_retryable_ack_as_retryable_error` verifies.
- **`ActionDispatchKind::Rejected { kind }` carries structured info.** Audit/observability can distinguish `PolicyDenied` from other rejection reasons.
- **`with_rewritten_user_message` defends against non-user-message envelopes.** Returns `ProductAdapterError::MalformedInboundPayload` if called on a `NoOp`/`Command`/etc. envelope, even though `dispatch_payload`'s code path only reaches the rewrite arm inside the `UserMessage(_)` match. Defense-in-depth.
- **Doc on `BeforeInboundPolicy::check_user_message` explicitly sets timeout expectations.** "Callers are responsible for enforcing wall-clock timeouts on any production policy implementation (for example by wrapping the port in a `tokio::time::timeout` decorator)." Good — this is the kind of contract people read in retrospect after a deadlock.
- **Tests are dense and exhaustive.** Eight new contract tests covering allow / rewrite (success + invalid + replay-once) / reject (permanent + retryable) / busy retry bypass / transient failure / duplicate replay. Each one pins a specific invariant.

### Issues / Suggestions

#### 1. Double `replay_accepted_user_message` for third-party `InboundTurnService` impls (medium)
The new trait default for `accept_user_message_after_policy` delegates to `accept_user_message`. The new trait default for `accept_user_message` itself calls `replay_accepted_user_message` as its first step. Workflow `dispatch_payload` already calls `replay_accepted_user_message` *before* dispatching to `accept_user_message_after_policy`.

For `DefaultInboundTurnService`, this is fine because the impl was rewritten to consolidate replay into the new method and `accept_user_message_after_policy` skips replay. For **third-party impls that only override `accept_user_message`** and use trait defaults elsewhere, the replay runs twice on every inbound: once via the workflow, once via the default `accept_user_message_after_policy` → `accept_user_message` → embedded replay path.

Two practical fixes:
- Make `replay_accepted_user_message` non-defaulted (forcing impls to think about it). Breaking change for any external impls, but the trait is in-tree only today so the blast radius is small.
- Or remove the trait default for `accept_user_message_after_policy` and require impls to provide it explicitly (also breaking, but louder).

A weaker fix is to document the invariant on the trait: "Callers must not invoke `replay_accepted_user_message` and `accept_user_message`/`accept_user_message_after_policy` on the same envelope; the workflow does this for you." Today the contract is implicit.

#### 2. `accept_user_message_after_policy` is publicly callable trait method that bypasses policy (low)
External callers can call `accept_user_message_after_policy` directly on `DefaultInboundTurnService` and skip the policy check entirely. Defensible because policy is the workflow's concern, not the turn service's — but if someone wires the turn service into a non-`DefaultProductWorkflow` ingress path, they silently skip the policy. Either:
- Rename to `accept_user_message_skipping_policy` (louder intent), or
- Strengthen the doc comment: "The only legitimate caller is `DefaultProductWorkflow::dispatch_payload` after it has run before-inbound policy. Direct callers must run policy themselves."

#### 3. `FakeInboundTurnService::program_replay_outcome` stores a single outcome (nit)
A test that wants different replay outcomes across calls (e.g., first call returns `DeferredBusy`, second returns `Submitted`) currently has to mutate `programmed_replay` between calls or recreate the fake. Most current tests don't need that, but `before_inbound_policy_does_not_block_staged_deferred_retry` already programs `program_outcome` for the first call and `program_replay_outcome` for the second — which works because they're separate fields. Consider `Vec<InboundTurnOutcome>` with FIFO for richer test scenarios.

#### 4. Counter fake-asymmetry (nit)
`FakeInboundTurnService` only overrides `accept_user_message`, so its `attempts` counter increments when the trait default for `accept_user_message_after_policy` delegates back to `accept_user_message`. `ReplayCountingInboundTurnService` (test-local) overrides both, so its `attempts` counter increments on `accept_user_message_after_policy`. This works today, but if the trait default ever changes (e.g., to stop delegating), the two fakes' counters will count different things. A short comment on `FakeInboundTurnService.attempts` clarifying "counts dispatches via `accept_user_message`" would prevent drift.

#### 5. `BeforeInboundPolicyRequest::new` clones envelope + payload (nit)
Both `envelope.clone()` and `user_message.clone()` happen for every inbound call. Envelopes can carry attachments; cloning is real work. For now this is one clone per inbound, which is fine. Worth flagging that future high-throughput optimizations should consider borrowing (`&'a ProductInboundEnvelope`) — though that complicates async-trait dispatch.

#### 6. `ProductInboundAck::Rejected` mapping is duplicated in two places (nit)
`dispatch_kind_from_ack` is the canonical "ack → ActionDispatchKind" mapping, but `dispatch_payload`'s rejection arm constructs `ActionDispatchKind::Rejected { kind: rejection.kind.clone() }` inline before reaching that helper. If a second code path ever needs to derive a dispatch kind from a `Rejected` ack, the mapping is repeated. Pulling the rejection-to-dispatch mapping into `dispatch_kind_from_ack` (or into a `From<&ProductInboundAck> for ActionDispatchKind` impl) would centralize it.

#### 7. WASM runner test name slightly misleading (nit)
`process_webhook_surfaces_retryable_ack_as_retryable_error` is correct, but in context the test specifically covers the *policy rejection* retryable path. Consider `process_webhook_surfaces_retryable_policy_rejection_as_retryable_error`. Trivial.

#### 8. `BeforeInboundPolicyOutcome` variants are not `#[non_exhaustive]` (nit)
Adding a fourth outcome (e.g. `Defer`, `Quarantine`) would be a breaking change for any external impl of `BeforeInboundPolicy`. Within this workspace it's fine; if the trait gets exposed to downstream consumers, mark `#[non_exhaustive]`.

### Project-conventions check
- `.unwrap()` only in tests / `expect()` only in test-support fakes ✓
- `thiserror`-shaped errors with context ✓
- Comments on the "why" of policy ordering (replay-before-policy, re-validation after rewrite) ✓
- New tests through the caller, not just the helper ✓ (the workflow tests drive `accept_inbound` end-to-end rather than testing `dispatch_payload` directly)

### Risk summary
**Low.** Backwards-compatible additions; default Noop policy; thorough test coverage of every outcome including edge cases (busy-retry bypass, ledger fingerprint release on transient/retryable rejection, payload re-validation). The one item worth picking up is the double-replay risk for third-party `InboundTurnService` impls (item 1) — fix either by removing the trait default or documenting the "either override `replay_accepted_user_message` or embed replay in `accept_user_message`, not both" contract clearly.

zmanian (commented):
## Review

Read the PR, prior bot/human reviews, and the substrate code paths. Most of what I'd otherwise flag is already covered: the double-replay risk (henrypark, ilblackdragon item 1), `dispatch_kind_from_ack`'s fabricated `TurnRunId` (Copilot), the rewrite-validation-error fingerprint loop (henrypark), and the trust-stamp preservation contract on `with_rewritten_user_message` (ilblackdragon). I won't re-litigate those. Three things worth surfacing that I don't see addressed.

### 1. Fail-closed-but-retried-forever for a broken policy (medium)

A policy returning `Err(ProductWorkflowError::Transient { .. })` is the only error shape exercised in the contract tests. In the current `terminal_ack_for_error` table, *every* `ProductWorkflowError` variant that a policy could plausibly return — `Transient`, `BindingResolutionFailed`, `TurnSubmissionRejected`, `DuplicateAction` — maps to `None`, which releases the idempotency fingerprint and propagates a retryable error. That's the right default for a transient store outage. It's the wrong default for a permanently broken policy implementation (e.g. a policy that always errors on a specific input shape, or a deterministic bug in a third-party policy port). The webhook protocol will redeliver indefinitely and the user message never settles either way.

Two ways to address, in order of preference:

- Give `BeforeInboundPolicy` a dedicated permanent-failure shape — either a new `ProductWorkflowError::BeforeInboundPolicyFailure { permanent: bool, .. }` variant that `terminal_ack_for_error` recognizes, or have policies model "I am broken" by returning `Reject(permanent(...))` rather than `Err(...)`. The trait doc currently doesn't tell implementors which path to choose.
- At minimum, document on the `BeforeInboundPolicy` trait that returning `Err` is always interpreted as retryable, and that policies which want to permanently drop a turn must return `Reject(ProductRejection::permanent(...))`. Right now that's implicit.

### 2. Trust surface exposed to policy implementors (medium)

`BeforeInboundPolicyRequest::new` clones the full `ProductInboundEnvelope` — including the host-stamped auth claim, install identity, conversation refs, attachments — and hands it by value to every policy. Henrypark flagged the clone cost; I'd flag the trust shape. The trait doc says policies may "allow, rewrite, or reject". In practice a policy port also receives every trusted stamp the host laid down, which means a careless implementation can leak auth claim fields into rewritten payloads, side-channel them out via metrics labels, or make routing decisions on fields the host considers private. Pick one of:

- Narrow the request struct to the fields a policy is intended to operate on (payload, tenant/user identity, source binding key). Keep the full envelope inside the workflow.
- Or document explicitly on `BeforeInboundPolicyRequest` which fields are intended to be policy-visible and which are passed through only because `Envelope` is currently the carrier — and add a future cleanup tracking issue.

This matters for #3607's beta path because the obvious third-party use is "wrap a remote content policy" — and the remote service should not see the host auth claim.

### 3. Single-policy seam, not a chain (low, design note)

The workflow holds `Arc<dyn BeforeInboundPolicy>`, singular. The first real production concerns I'd expect to attach here — rate limiting, content classifier, capability/scope gate, jailbreak detector — are independent and should compose. Today, adding the second policy means writing a `Chain { inner: Vec<Arc<dyn BeforeInboundPolicy>> }` adapter and getting the short-circuit / fail-closed / rewrite-then-rewrite-again semantics right out-of-tree. Worth deciding now whether that composer lives in this crate (with documented ordering and fail-closed semantics) or whether the trait should take `Vec<Arc<dyn _>>` directly. Either is fine; "everyone writes their own composer" is the option I'd avoid.

No `ironclaw_hooks` crate exists in this tree and the PR is contained entirely to substrate crates under `crates/ironclaw_product_workflow/` and `crates/ironclaw_product_adapters/` — no inverted dependency on the binary, which is the right boundary.

### Test coverage

Allow / rewrite / permanent-reject / retryable-reject / transient-failure / duplicate-replay / busy-retry-bypass / rewrite-revalidation are all covered. The gap is **policy-returns-non-Transient-error** (item 1 above) and **policy-rewrite-leaks-trusted-field** (item 2 above — harder to test, but a doc-level assertion would do).

Not blocking on the beta slice given the Noop default keeps existing call sites unaffected, but items 1 and 2 should land before any non-trivial policy ships behind this seam.

— Zaki

serrrfirat:
Addressed the review feedback from @zmanian and @ilblackdragon in 39a6b402:

- Added explicit permanent before-inbound policy failure handling so broken policies can fail closed without retrying forever.
- Narrowed `BeforeInboundPolicyRequest` so policy implementations no longer receive the full trusted envelope/auth claim.
- Made the inbound replay/policy boundary explicit: replay is required, the post-policy accept path is now named `accept_user_message_skipping_policy`, and tests cover no double replay.
- Added ordered fake replay outcomes, reused ack→dispatch mapping for rejections, marked policy outcomes non-exhaustive, and renamed the WASM retryable-policy test.

Validation run:
- `cargo fmt --check`
- `cargo test -p ironclaw_wasm_product_adapters process_webhook_surfaces_retryable_policy_rejection_as_retryable_error`
- `scripts/pre-commit-safety.sh`
- Codex review against `origin/reborn-integration` (clean)

Could you please take another look?

zmanian (commented):
## Re-review

Followed up on commit `39a6b40` (2026-05-18) against my 2026-05-17 review. All three findings I raised are addressed in this single commit.

### Finding 1 — Fail-closed-but-retried-forever — RESOLVED
`ProductWorkflowError::BeforeInboundPolicyFailed { reason, permanent }` now distinguishes terminal from transient policy failures. `terminal_ack_for_error` returns a `PolicyDenied` permanent rejection when `permanent: true`, and `From<ProductWorkflowError> for ProductAdapterError` maps it to a non-retryable `WorkflowRejected { status_code: 403, retryable: false }`; `permanent: false` and `Transient` continue to release the fingerprint. The trait doc on `BeforeInboundPolicy` now spells out the three disposition shapes. Covered by `before_inbound_policy_permanent_failure_settles_terminal_rejection` (settles `Rejected { PolicyDenied }`, second submit replays as `Duplicate`) and `permanent_before_inbound_policy_failure_maps_to_rejection`.

### Finding 2 — Trust surface — RESOLVED
`BeforeInboundPolicyRequest` no longer carries the full envelope. It is now `{ adapter_id, installation_id, external_actor_ref, external_conversation_ref, source_binding_key, user_message }` — narrowed to product identity + binding refs + payload. Host-stamped auth claim and trusted context stay inside the workflow. Module doc comment makes the contract explicit. Contract test asserts the new shape field-by-field. Good.

### Finding 3 — Single-policy seam vs chain — ACKNOWLEDGED, NOT IMPLEMENTED
The trait doc now states composition (rate-limit → classifier → scope) is the implementor's responsibility behind one seam, with the rationale that the workflow owns one seam so replay-before-policy and settlement ordering stay deterministic. I'll accept that for the beta slice — but it leaves the in-crate `Chain` combinator (with documented fail-closed/short-circuit semantics) as the next concern when the second policy ships, before downstream crates each grow their own ad-hoc composition. Non-blocking.

### Other changes worth noting
- `accept_user_message_after_policy` renamed to `accept_user_message_skipping_policy` with a doc warning that only `DefaultProductWorkflow` should call it; both replay and accept methods are now required (no defaults), which forecloses an implementor accidentally bypassing the replay-before-policy invariant. Nice tightening.
- `BeforeInboundPolicyOutcome` is now `#[non_exhaustive]` — good forward-compat for adding Defer/Quarantine later without breaking downstream matches.
- `dispatch_kind_from_ack` now handles `Rejected` directly, removing the asymmetric path where rejected acks fell through to `try_from_payload`.
- `FakeInboundTurnService::programmed_replay` switched to `VecDeque` with ordered `program_replay_outcomes` — supports asserting replay ordering, exercised by a new fake-contract test.

### Verdict
Approving the substrate slice. Finding 3 (in-crate policy chain) is the right next-step issue but doesn't need to block this PR.

— Zaki

crates/ironclaw_product_workflow/src/inbound_turn.rs
```diff
@@ -205,6 +286,174 @@ where
...
+                        }
+                    })?;
+                &dispatch_envelope
+            }
+            BeforeInboundPolicyOutcome::Reject(rejection) => {
+                return Ok(InboundUserMessageDispatch::Rejected(rejection));
+            }
+        };
+
+        self.accept_prepared_user_message(prepared, envelope_for_turn)
```
serrrfirat: **[Medium] Policy trigger rewrites do not affect route binding**

`prepare_user_message` resolves the binding and `ThreadScope` from the original payload trigger before the policy runs, then `accept_prepared_user_message` reuses that prepared route even when `RewriteUserMessage` returns a payload with a different trigger. Since the policy contract says rewrites can change content/attachments/trigger, a DirectChat -> shared rewrite, or the reverse, is staged/submitted under the original route.

Fix: after a rewrite, either reject trigger changes or recompute binding/thread scope from the rewritten envelope before staging.

Also flagged by: security, tests.

crates/ironclaw_product_workflow/src/workflow.rs
```diff
@@ -360,11 +405,21 @@ fn terminal_ack_for_error(error: &ProductWorkflowError) -> Option<ProductInbound
...
+        } => Some(ProductInboundAck::Rejected(ProductRejection::permanent(
+            ProductRejectionKind::PolicyDenied,
+            format!("before-inbound policy failed: {reason}"),
+        ))),
         ProductWorkflowError::BindingResolutionFailed { .. }
         | ProductWorkflowError::TurnSubmissionRejected { .. }
         | ProductWorkflowError::TurnSubmissionFailed { .. }
         | ProductWorkflowError::TurnResumeRejected { .. }
         | ProductWorkflowError::Transient { .. }
+        | ProductWorkflowError::BeforeInboundPolicyFailed {
```
serrrfirat, line 420: **[Low] Retryable `BeforeInboundPolicyFailed` path is not covered**

Tests cover `Transient` policy failures and permanent `BeforeInboundPolicyFailed`, but not `BeforeInboundPolicyFailed { permanent: false }`. That branch should release the idempotency fingerprint rather than settling a terminal rejection, and it is a distinct public error shape.

Fix: add `before_inbound_policy_retryable_failure_releases_fingerprint` covering `BeforeInboundPolicyFailed { permanent: false }` and asserting the policy reruns on redelivery.

crates/ironclaw_product_workflow/src/policy.rs
```diff
@@ -0,0 +1,99 @@
...
+/// claim. Policies see only the user-message payload plus product identity and
+/// binding refs needed for policy decisions; the workflow keeps trusted context
+/// for downstream staging and turn submission.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct BeforeInboundPolicyRequest {
+    pub adapter_id: ProductAdapterId,
+    pub installation_id: AdapterInstallationId,
+    pub external_actor_ref: ExternalActorRef,
+    pub external_conversation_ref: ExternalConversationRef,
+    pub source_binding_key: String,
```
serrrfirat: **[Low] Use the existing typed source-binding key**

`BeforeInboundPolicyRequest` exposes `source_binding_key` as a raw `String`, while the same workflow concept is already modeled as `SourceBindingKey` in the action fingerprint. That splits the public contract between typed and untyped representations.

Fix: change `BeforeInboundPolicyRequest.source_binding_key` to `SourceBindingKey` and construct it from the already validated source binding key.

serrrfirat (commented):
## Code Review

Reviewed PR #3632 at `3dd61837` with the multi-agent code-review skill.

Summary:
- Reviewers: security, bugs, performance/concurrency, tests, conventions
- Findings after confidence filtering and overlap dedupe: 4
- Event: COMMENT (no Critical/High findings)

Findings:
1. Medium correctness/security: policy trigger rewrites do not affect route binding.
2. Low tests: retryable `BeforeInboundPolicyFailed { permanent: false }` branch lacks coverage.
3. Low conventions: `BeforeInboundPolicyRequest.source_binding_key` should use the existing typed key.
4. Medium conventions: public `InboundTurnService` exposes a policy-skipping method.

Performance/concurrency returned no findings.

crates/ironclaw_product_workflow/src/inbound_turn.rs
```diff
@@ -70,17 +76,92 @@ impl InboundTurnOutcome {
...
         envelope: &ProductInboundEnvelope,
     ) -> Result<InboundTurnOutcome, ProductWorkflowError>;
+
+    /// Accept a user message after the caller already performed replay lookup
+    /// and before-inbound policy.
+    ///
+    /// Only [`DefaultProductWorkflow`](crate::DefaultProductWorkflow) should call
+    /// this directly. Other callers must run replay lookup and before-inbound
+    /// policy first; calling this as a shortcut intentionally skips policy.
+    async fn accept_user_message_skipping_policy(
```
serrrfirat: **[Medium] Do not expose a public policy-skipping path**

`InboundTurnService` is re-exported from the workflow crate, but `accept_user_message_skipping_policy` is now part of that public trait contract and explicitly bypasses before-inbound policy. The product workflow owns binding, staging, idempotency, and acknowledgement orchestration; exposing the bypass lets downstream callers depend on the wrong entry point.

Fix: keep the skip-policy operation private to `DefaultInboundTurnService`, and expose only a policy-preserving public trait method for user-message dispatch.

serrrfirat (commented):
test

crates/ironclaw_product_workflow/src/inbound_turn.rs
```diff
@@ -113,18 +150,92 @@ where
...
+            BeforeInboundPolicyOutcome::Allow => (prepared, envelope),
+            BeforeInboundPolicyOutcome::RewriteUserMessage(payload) => {
+                dispatch_envelope =
+                    envelope.with_rewritten_user_message(payload).map_err(|_| {
+                        ProductWorkflowError::TurnSubmissionRejected {
+                            reason: "invalid policy-rewritten user message".into(),
+                        }
+                    })?;
+                let rewritten_prepared = self.prepare_user_message(&dispatch_envelope).await?;
+                (rewritten_prepared, &dispatch_envelope)
```
serrrfirat, line 209: **High - performance/redundant-io**

When policy returns `RewriteUserMessage`, `prepare_user_message` is called again on the rewritten envelope. But the rewrite only changes user_message text/attachments/trigger - the trigger is preserved. Since `prepare_user_message` calls `resolve_binding` (depends on adapter_id, installation_id, external refs - all unchanged) and `thread_scope_from_binding` (depends on trigger - unchanged), the second preparation produces identical results.

This wastes a binding resolution call (likely DB/network round-trip) on every policy rewrite.

**Fix:** Reuse the original `prepared` binding/scope:
```rust
BeforeInboundPolicyOutcome::RewriteUserMessage(payload) => {
    dispatch_envelope = envelope.with_rewritten_user_message(payload).map_err(...)?;
    (prepared, &dispatch_envelope)  // reuse original prepared
}
```

serrrfirat (commented):
## Review Summary

| Metric | Value |
|---|---|
| **Verdict** | Changes Requested |
| **Findings** | 11 (4 High, 6 Medium, 1 Low) |
| **Reviewers** | security, bugs, performance, tests, conventions (all passed) |
| **Confidence threshold** | >= 50 |

## Findings by Category

### High (4)

**bugs/idempotency-semantics** - Retryable policy rejection records dispatch_kind but does not settle. If ledger recovery inspects in-flight dispatch_kind, it may misinterpret as terminal rejection. Verify recovery semantics.

**performance/missing-timeout** - BeforeInboundPolicy runs inline with no timeout enforcement. Trait docstring warns about stall risk but provides no decorator or builder-level timeout. A slow policy could consume the entire 55s workflow budget.

**security/input-validation** - BeforeInboundPolicyRequest exposes external refs sufficient for fingerprinting. No rate-limit context available to policy implementations.

**tests/missing-error-path-test** - No test for policy error (BeforeInboundPolicyFailed) path through DefaultInboundTurnService. FakeBeforeInboundPolicy has force_failure() but no contract test exercises it.

### Medium (6)

- **bugs/error-propagation** - noop-policy rejection maps to Transient without permanent flag, causing indefinite retry cycle
- **performance/redundant-io** - prepare_user_message called redundantly after policy rewrite (DB round-trip wasted)
- **security/error-leak** - BeforeInboundPolicyFailed reason included raw in rejection ack message
- **tests/missing-integration** - No test for should_settle_ack behavior change with retryable rejection
- **tests/missing-error-path-test** - No serde test for UserMessagePayload validation through deserialization
- **conventions/expedient-hack** - FakeBeforeInboundPolicy has single global outcome, not per-request programmable

### Low (1)

- **conventions/sibling-pattern** - FakeInboundTurnService Default impl moved below trait impl, inconsistent with sibling pattern

---

*Review generated by code-review-skill v1*

crates/ironclaw_product_workflow/src/policy.rs
```diff
@@ -0,0 +1,105 @@
...
+    /// Continue with rewritten content/attachments/trigger.
+    RewriteUserMessage(UserMessagePayload),
+    /// Reject before canonical transcript staging or turn submission.
+    Reject(ProductRejection),
+}
+
+/// Policy port that runs before user-message staging.
+///
+/// Implementations must be bounded: the workflow runs `check_user_message`
+/// inline on the inbound dispatch path while holding an in-flight idempotency
```
serrrfirat, line 65: **High - performance/missing-timeout**

Trait docstring warns "a slow or stuck policy will stall inbound submissions" and says "Callers are responsible for enforcing wall-clock timeouts." But neither `DefaultInboundTurnService` nor `DefaultProductWorkflow` wrap the policy call in `tokio::time::timeout`.

The runner has a 55s workflow-level timeout, but that covers the entire pipeline. A policy hanging for 50s would consume the entire budget.

**Fix:** Add optional timeout parameter to `with_before_inbound_policy` or wrap `check_user_message` in `tokio::time::timeout` with a configurable default.

serrrfirat (commented):
## Review Summary

| Metric | Value |
|---|---|
| **Verdict** | Changes Requested |
| **Findings** | 11 (4 High, 6 Medium, 1 Low) |
| **Reviewers** | security, bugs, performance, tests, conventions (all passed) |
| **Confidence threshold** | >= 50 |

## Findings by Category

### High (4)

**bugs/idempotency-semantics** - Retryable policy rejection records dispatch_kind but does not settle. If ledger recovery inspects in-flight dispatch_kind, it may misinterpret as terminal rejection. Verify recovery semantics.

**performance/missing-timeout** - BeforeInboundPolicy runs inline with no timeout enforcement. Trait docstring warns about stall risk but provides no decorator or builder-level timeout. A slow policy could consume the entire 55s workflow budget.

**security/input-validation** - BeforeInboundPolicyRequest exposes external refs sufficient for fingerprinting. No rate-limit context available to policy implementations.

**tests/missing-error-path-test** - No test for policy error (BeforeInboundPolicyFailed) path through DefaultInboundTurnService. FakeBeforeInboundPolicy has force_failure() but no contract test exercises it.

### Medium (6)

- **bugs/error-propagation** - noop-policy rejection maps to Transient without permanent flag, causing indefinite retry cycle
- **performance/redundant-io** - prepare_user_message called redundantly after policy rewrite (DB round-trip wasted)
- **security/error-leak** - BeforeInboundPolicyFailed reason included raw in rejection ack message
- **tests/missing-integration** - No test for should_settle_ack behavior change with retryable rejection
- **tests/missing-error-path-test** - No serde test for UserMessagePayload validation through deserialization
- **conventions/expedient-hack** - FakeBeforeInboundPolicy has single global outcome, not per-request programmable

### Low (1)

- **conventions/sibling-pattern** - FakeInboundTurnService Default impl moved below trait impl, inconsistent with sibling pattern

---

*Review generated by code-review-skill v1*

crates/ironclaw_product_workflow/src/policy.rs
```diff
@@ -0,0 +1,105 @@
...
+    ProductInboundEnvelope, ProductRejection, UserMessagePayload,
+};
+
+use crate::action::SourceBindingKey;
+use crate::error::ProductWorkflowError;
+
+/// Request passed to before-inbound policy implementations.
+///
+/// This intentionally excludes the full trusted envelope and host-stamped auth
+/// claim. Policies see only the user-message payload plus product identity and
```
serrrfirat, line 19: **High - security/input-validation**

`BeforeInboundPolicyRequest` includes `external_actor_ref`, `external_conversation_ref`, and `source_binding_key`. While the docstring says it "intentionally excludes the full trusted envelope," these refs are sufficient for user fingerprinting and correlation attacks.

No rate-limit context is available to the policy.

**Fix:** Add rate-limit context token or document that rate-limiting belongs in a decorator wrapping the policy trait.

serrrfirat (commented):
## Review Summary

| Metric | Value |
|---|---|
| **Verdict** | Changes Requested |
| **Findings** | 11 (4 High, 6 Medium, 1 Low) |
| **Reviewers** | security, bugs, performance, tests, conventions (all passed) |
| **Confidence threshold** | >= 50 |

## Findings by Category

### High (4)

**bugs/idempotency-semantics** - Retryable policy rejection records dispatch_kind but does not settle. If ledger recovery inspects in-flight dispatch_kind, it may misinterpret as terminal rejection. Verify recovery semantics.

**performance/missing-timeout** - BeforeInboundPolicy runs inline with no timeout enforcement. Trait docstring warns about stall risk but provides no decorator or builder-level timeout. A slow policy could consume the entire 55s workflow budget.

**security/input-validation** - BeforeInboundPolicyRequest exposes external refs sufficient for fingerprinting. No rate-limit context available to policy implementations.

**tests/missing-error-path-test** - No test for policy error (BeforeInboundPolicyFailed) path through DefaultInboundTurnService. FakeBeforeInboundPolicy has force_failure() but no contract test exercises it.

### Medium (6)

- **bugs/error-propagation** - noop-policy rejection maps to Transient without permanent flag, causing indefinite retry cycle
- **performance/redundant-io** - prepare_user_message called redundantly after policy rewrite (DB round-trip wasted)
- **security/error-leak** - BeforeInboundPolicyFailed reason included raw in rejection ack message
- **tests/missing-integration** - No test for should_settle_ack behavior change with retryable rejection
- **tests/missing-error-path-test** - No serde test for UserMessagePayload validation through deserialization
- **conventions/expedient-hack** - FakeBeforeInboundPolicy has single global outcome, not per-request programmable

### Low (1)

- **conventions/sibling-pattern** - FakeInboundTurnService Default impl moved below trait impl, inconsistent with sibling pattern

---

*Review generated by code-review-skill v1*

crates/ironclaw_product_workflow/src/inbound_turn.rs
```diff
@@ -113,18 +150,92 @@ where
...
     ) -> Result<InboundTurnOutcome, ProductWorkflowError> {
+        let policy = NoopBeforeInboundPolicy;
+        match self
+            .accept_user_message_with_before_policy(envelope, &policy)
+            .await?
+        {
+            InboundUserMessageDispatch::Accepted(outcome) => Ok(outcome),
+            InboundUserMessageDispatch::Rejected(_) => Err(ProductWorkflowError::Transient {
+                reason: "noop before-inbound policy unexpectedly rejected message".into(),
+            }),
```
serrrfirat, line 173: **Medium - bugs/error-propagation**

When `NoopBeforeInboundPolicy` unexpectedly returns `Reject`, the error is `ProductWorkflowError::Transient { reason }`. Since `permanent` defaults to `false`, `terminal_ack_for_error` returns `None`, the idempotency ledger releases, and the adapter retries - cycling until lease timeout.

A Noop rejection is a bug, not a transient condition.

**Fix:** Return `Internal` error which settles terminal, or use a distinct error variant.

serrrfirat (commented):
## Review Summary

| Metric | Value |
|---|---|
| **Verdict** | Changes Requested |
| **Findings** | 11 (4 High, 6 Medium, 1 Low) |
| **Reviewers** | security, bugs, performance, tests, conventions (all passed) |
| **Confidence threshold** | >= 50 |

## Findings by Category

### High (4)

**bugs/idempotency-semantics** - Retryable policy rejection records dispatch_kind but does not settle. If ledger recovery inspects in-flight dispatch_kind, it may misinterpret as terminal rejection. Verify recovery semantics.

**performance/missing-timeout** - BeforeInboundPolicy runs inline with no timeout enforcement. Trait docstring warns about stall risk but provides no decorator or builder-level timeout. A slow policy could consume the entire 55s workflow budget.

**security/input-validation** - BeforeInboundPolicyRequest exposes external refs sufficient for fingerprinting. No rate-limit context available to policy implementations.

**tests/missing-error-path-test** - No test for policy error (BeforeInboundPolicyFailed) path through DefaultInboundTurnService. FakeBeforeInboundPolicy has force_failure() but no contract test exercises it.

### Medium (6)

- **bugs/error-propagation** - noop-policy rejection maps to Transient without permanent flag, causing indefinite retry cycle
- **performance/redundant-io** - prepare_user_message called redundantly after policy rewrite (DB round-trip wasted)
- **security/error-leak** - BeforeInboundPolicyFailed reason included raw in rejection ack message
- **tests/missing-integration** - No test for should_settle_ack behavior change with retryable rejection
- **tests/missing-error-path-test** - No serde test for UserMessagePayload validation through deserialization
- **conventions/expedient-hack** - FakeBeforeInboundPolicy has single global outcome, not per-request programmable

### Low (1)

- **conventions/sibling-pattern** - FakeInboundTurnService Default impl moved below trait impl, inconsistent with sibling pattern

---

*Review generated by code-review-skill v1*

crates/ironclaw_product_workflow/src/workflow.rs
```diff
@@ -90,13 +101,18 @@ impl ProductWorkflow for DefaultProductWorkflow {
...
+                .await;
 
                 match result {
-                    Ok(ack) => {
-                        action.mark_dispatched(dispatch_kind_from_ack(&ack, envelope.payload())?);
-                        if is_terminal_success_ack(&ack) {
-                            action.settle(ack.clone());
+                    Ok(dispatched) => {
+                        action.mark_dispatched(dispatched.dispatch_kind);
+                        if should_settle_ack(&dispatched.ack) {
```
serrrfirat, line 114: **High - bugs/idempotency-semantics**

When policy returns `Reject` with retryable disposition, `should_settle_ack` returns `false`. The action gets `mark_dispatched(Rejected)` but is NOT settled - the ledger releases it.

`mark_dispatched` records a `Rejected` kind on an unsettled action. If ledger recovery inspects in-flight actions' `dispatch_kind`, it may misinterpret as terminal.

**Fix:** Verify `InMemoryIdempotencyLedger` recovery handles unsettled `Rejected` dispatch_kind correctly; add test.

serrrfirat (commented):
## Review Summary

| Metric | Value |
|---|---|
| **Verdict** | Changes Requested |
| **Findings** | 11 (4 High, 6 Medium, 1 Low) |
| **Reviewers** | security, bugs, performance, tests, conventions (all passed) |
| **Confidence threshold** | >= 50 |

## Findings by Category

### High (4)

**bugs/idempotency-semantics** - Retryable policy rejection records dispatch_kind but does not settle. If ledger recovery inspects in-flight dispatch_kind, it may misinterpret as terminal rejection. Verify recovery semantics.

**performance/missing-timeout** - BeforeInboundPolicy runs inline with no timeout enforcement. Trait docstring warns about stall risk but provides no decorator or builder-level timeout. A slow policy could consume the entire 55s workflow budget.

**security/input-validation** - BeforeInboundPolicyRequest exposes external refs sufficient for fingerprinting. No rate-limit context available to policy implementations.

**tests/missing-error-path-test** - No test for policy error (BeforeInboundPolicyFailed) path through DefaultInboundTurnService. FakeBeforeInboundPolicy has force_failure() but no contract test exercises it.

### Medium (6)

- **bugs/error-propagation** - noop-policy rejection maps to Transient without permanent flag, causing indefinite retry cycle
- **performance/redundant-io** - prepare_user_message called redundantly after policy rewrite (DB round-trip wasted)
- **security/error-leak** - BeforeInboundPolicyFailed reason included raw in rejection ack message
- **tests/missing-integration** - No test for should_settle_ack behavior change with retryable rejection
- **tests/missing-error-path-test** - No serde test for UserMessagePayload validation through deserialization
- **conventions/expedient-hack** - FakeBeforeInboundPolicy has single global outcome, not per-request programmable

### Low (1)

- **conventions/sibling-pattern** - FakeInboundTurnService Default impl moved below trait impl, inconsistent with sibling pattern

---

*Review generated by code-review-skill v1*

crates/ironclaw_product_workflow/src/workflow.rs
```diff
@@ -285,12 +327,15 @@ fn dispatch_kind_from_ack(
...
+        }),
         _ => ActionDispatchKind::try_from_payload(payload),
     }
 }
 
-fn is_terminal_success_ack(ack: &ProductInboundAck) -> bool {
-    !matches!(ack, ProductInboundAck::DeferredBusy { .. })
+fn should_settle_ack(ack: &ProductInboundAck) -> bool {
+    !matches!(ack, ProductInboundAck::DeferredBusy { .. }) && ack.is_durable_outcome()
 }
```
serrrfirat, line 340: **Medium - tests/missing-integration**

The rename from `is_terminal_success_ack` to `should_settle_ack` adds `ack.is_durable_outcome()` to the gate, changing settlement for retryable rejections. The existing unit test only checks `NoOp` and `DeferredBusy`.

**Fix:** Add tests for `should_settle_ack` with both retryable and permanent `Rejected` acks.

serrrfirat (commented):
## Review Summary

| Metric | Value |
|---|---|
| **Verdict** | Changes Requested |
| **Findings** | 11 (4 High, 6 Medium, 1 Low) |
| **Reviewers** | security, bugs, performance, tests, conventions (all passed) |
| **Confidence threshold** | >= 50 |

## Findings by Category

### High (4)

**bugs/idempotency-semantics** - Retryable policy rejection records dispatch_kind but does not settle. If ledger recovery inspects in-flight dispatch_kind, it may misinterpret as terminal rejection. Verify recovery semantics.

**performance/missing-timeout** - BeforeInboundPolicy runs inline with no timeout enforcement. Trait docstring warns about stall risk but provides no decorator or builder-level timeout. A slow policy could consume the entire 55s workflow budget.

**security/input-validation** - BeforeInboundPolicyRequest exposes external refs sufficient for fingerprinting. No rate-limit context available to policy implementations.

**tests/missing-error-path-test** - No test for policy error (BeforeInboundPolicyFailed) path through DefaultInboundTurnService. FakeBeforeInboundPolicy has force_failure() but no contract test exercises it.

### Medium (6)

- **bugs/error-propagation** - noop-policy rejection maps to Transient without permanent flag, causing indefinite retry cycle
- **performance/redundant-io** - prepare_user_message called redundantly after policy rewrite (DB round-trip wasted)
- **security/error-leak** - BeforeInboundPolicyFailed reason included raw in rejection ack message
- **tests/missing-integration** - No test for should_settle_ack behavior change with retryable rejection
- **tests/missing-error-path-test** - No serde test for UserMessagePayload validation through deserialization
- **conventions/expedient-hack** - FakeBeforeInboundPolicy has single global outcome, not per-request programmable

### Low (1)

- **conventions/sibling-pattern** - FakeInboundTurnService Default impl moved below trait impl, inconsistent with sibling pattern

---

*Review generated by code-review-skill v1*

crates/ironclaw_product_workflow/src/workflow.rs
```diff
@@ -360,11 +405,21 @@ fn terminal_ack_for_error(error: &ProductWorkflowError) -> Option<ProductInbound
                 format!("turn submission rejected: {error}"),
             )))
         }
+        ProductWorkflowError::BeforeInboundPolicyFailed {
+            reason,
```
serrrfirat, line 409: **Medium - security/error-leak**

`terminal_ack_for_error` maps `BeforeInboundPolicyFailed { permanent: true, reason }` to rejection with message `format!("before-inbound policy failed: {reason}")`.

If production policy sets `reason` with internal details (model names, classifier scores), these leak to adapter layer. Per `.claude/rules/error-handling.md`, no internal identifier may cross channel boundary.

**Fix:** Wrap `reason` in `RedactedString` and ensure sanitization at channel edge.

serrrfirat (commented):
## Review Summary

| Metric | Value |
|---|---|
| **Verdict** | Changes Requested |
| **Findings** | 11 (4 High, 6 Medium, 1 Low) |
| **Reviewers** | security, bugs, performance, tests, conventions (all passed) |
| **Confidence threshold** | >= 50 |

## Findings by Category

### High (4)

**bugs/idempotency-semantics** - Retryable policy rejection records dispatch_kind but does not settle. If ledger recovery inspects in-flight dispatch_kind, it may misinterpret as terminal rejection. Verify recovery semantics.

**performance/missing-timeout** - BeforeInboundPolicy runs inline with no timeout enforcement. Trait docstring warns about stall risk but provides no decorator or builder-level timeout. A slow policy could consume the entire 55s workflow budget.

**security/input-validation** - BeforeInboundPolicyRequest exposes external refs sufficient for fingerprinting. No rate-limit context available to policy implementations.

**tests/missing-error-path-test** - No test for policy error (BeforeInboundPolicyFailed) path through DefaultInboundTurnService. FakeBeforeInboundPolicy has force_failure() but no contract test exercises it.

### Medium (6)

- **bugs/error-propagation** - noop-policy rejection maps to Transient without permanent flag, causing indefinite retry cycle
- **performance/redundant-io** - prepare_user_message called redundantly after policy rewrite (DB round-trip wasted)
- **security/error-leak** - BeforeInboundPolicyFailed reason included raw in rejection ack message
- **tests/missing-integration** - No test for should_settle_ack behavior change with retryable rejection
- **tests/missing-error-path-test** - No serde test for UserMessagePayload validation through deserialization
- **conventions/expedient-hack** - FakeBeforeInboundPolicy has single global outcome, not per-request programmable

### Low (1)

- **conventions/sibling-pattern** - FakeInboundTurnService Default impl moved below trait impl, inconsistent with sibling pattern

---

*Review generated by code-review-skill v1*

crates/ironclaw_product_workflow/src/fakes.rs
```diff
@@ -256,6 +259,108 @@ impl IdempotencyLedger for FakeIdempotencyLedger {
...
+impl FakeBeforeInboundPolicy {
+    pub fn new() -> Self {
+        Self {
+            state: Arc::new(Mutex::new(FakeBeforeInboundPolicyState::default())),
+        }
+    }
+
+    /// Allow all messages without modification.
+    pub fn allow(&self) {
+        let mut state = self
```
serrrfirat, line 287: **Medium - conventions/expedient-hack**

`FakeBeforeInboundPolicy` stores a single global outcome applied to all requests. Tests needing different outcomes for sequential requests must mutate the fake between calls.

Compare to `FakeInboundTurnService` which has `program_replay_outcomes(VecDeque)` for sequential behavior.

**Fix:** Add `program_outcomes(VecDeque<BeforeInboundPolicyOutcome>)` matching existing fake pattern.

serrrfirat (commented):
## Review Summary

| Metric | Value |
|---|---|
| **Verdict** | Changes Requested |
| **Findings** | 11 (4 High, 6 Medium, 1 Low) |
| **Reviewers** | security, bugs, performance, tests, conventions (all passed) |
| **Confidence threshold** | >= 50 |

## Findings by Category

### High (4)

**bugs/idempotency-semantics** - Retryable policy rejection records dispatch_kind but does not settle. If ledger recovery inspects in-flight dispatch_kind, it may misinterpret as terminal rejection. Verify recovery semantics.

**performance/missing-timeout** - BeforeInboundPolicy runs inline with no timeout enforcement. Trait docstring warns about stall risk but provides no decorator or builder-level timeout. A slow policy could consume the entire 55s workflow budget.

**security/input-validation** - BeforeInboundPolicyRequest exposes external refs sufficient for fingerprinting. No rate-limit context available to policy implementations.

**tests/missing-error-path-test** - No test for policy error (BeforeInboundPolicyFailed) path through DefaultInboundTurnService. FakeBeforeInboundPolicy has force_failure() but no contract test exercises it.

### Medium (6)

- **bugs/error-propagation** - noop-policy rejection maps to Transient without permanent flag, causing indefinite retry cycle
- **performance/redundant-io** - prepare_user_message called redundantly after policy rewrite (DB round-trip wasted)
- **security/error-leak** - BeforeInboundPolicyFailed reason included raw in rejection ack message
- **tests/missing-integration** - No test for should_settle_ack behavior change with retryable rejection
- **tests/missing-error-path-test** - No serde test for UserMessagePayload validation through deserialization
- **conventions/expedient-hack** - FakeBeforeInboundPolicy has single global outcome, not per-request programmable

### Low (1)

- **conventions/sibling-pattern** - FakeInboundTurnService Default impl moved below trait impl, inconsistent with sibling pattern

---

*Review generated by code-review-skill v1*

crates/ironclaw_product_workflow/tests/product_workflow_contract.rs
```diff
@@ -251,6 +253,120 @@ fn fake_binding() -> ResolvedBinding {
     }
 }
 
+#[derive(Default)]
+struct ReplayCountingInboundTurnService {
+    replay_attempts: Mutex<usize>,
```
serrrfirat, line 258: **High - tests/missing-error-path-test**

Contract tests cover allow, rewrite, reject, and retryable-reject. But no test where `check_user_message` returns `Err(BeforeInboundPolicyFailed { ... })`.

This error path is critical: `permanent=true` should settle as rejection, `permanent=false` should release and retry. `FakeBeforeInboundPolicy` has `force_failure()` but no contract test exercises it.

**Fix:** Add `before_inbound_policy_error_permanent_settles` and `before_inbound_policy_error_transient_releases`.

serrrfirat (commented):
## Review Summary

| Metric | Value |
|---|---|
| **Verdict** | Changes Requested |
| **Findings** | 11 (4 High, 6 Medium, 1 Low) |
| **Reviewers** | security, bugs, performance, tests, conventions (all passed) |
| **Confidence threshold** | >= 50 |

## Findings by Category

### High (4)

**bugs/idempotency-semantics** - Retryable policy rejection records dispatch_kind but does not settle. If ledger recovery inspects in-flight dispatch_kind, it may misinterpret as terminal rejection. Verify recovery semantics.

**performance/missing-timeout** - BeforeInboundPolicy runs inline with no timeout enforcement. Trait docstring warns about stall risk but provides no decorator or builder-level timeout. A slow policy could consume the entire 55s workflow budget.

**security/input-validation** - BeforeInboundPolicyRequest exposes external refs sufficient for fingerprinting. No rate-limit context available to policy implementations.

**tests/missing-error-path-test** - No test for policy error (BeforeInboundPolicyFailed) path through DefaultInboundTurnService. FakeBeforeInboundPolicy has force_failure() but no contract test exercises it.

### Medium (6)

- **bugs/error-propagation** - noop-policy rejection maps to Transient without permanent flag, causing indefinite retry cycle
- **performance/redundant-io** - prepare_user_message called redundantly after policy rewrite (DB round-trip wasted)
- **security/error-leak** - BeforeInboundPolicyFailed reason included raw in rejection ack message
- **tests/missing-integration** - No test for should_settle_ack behavior change with retryable rejection
- **tests/missing-error-path-test** - No serde test for UserMessagePayload validation through deserialization
- **conventions/expedient-hack** - FakeBeforeInboundPolicy has single global outcome, not per-request programmable

### Low (1)

- **conventions/sibling-pattern** - FakeInboundTurnService Default impl moved below trait impl, inconsistent with sibling pattern

---

*Review generated by code-review-skill v1*

crates/ironclaw_product_adapters/src/inbound.rs
```diff
@@ -93,13 +93,17 @@ impl UserMessagePayload {
...
-        })
+        };
+        payload.validate()?;
+        Ok(payload)
+    }
+
+    pub fn validate(&self) -> Result<(), ProductAdapterError> {
+        validate_payload_string("user message text", &self.text, USER_MESSAGE_TEXT_MAX_BYTES)
     }
 }
```
serrrfirat, line 109: **Medium - tests/missing-error-path-test**

`UserMessagePayload` now has custom `Deserialize` calling `Self::new` -> `validate()`. But no test verifies that deserializing an oversized text payload through serde fails.

**Fix:** Add `fn user_message_text_length_bounded_through_serde` deserializing oversized text via serde_json and asserting error.

serrrfirat (commented):
## Review Summary

| Metric | Value |
|---|---|
| **Verdict** | Changes Requested |
| **Findings** | 11 (4 High, 6 Medium, 1 Low) |
| **Reviewers** | security, bugs, performance, tests, conventions (all passed) |
| **Confidence threshold** | >= 50 |

## Findings by Category

### High (4)

**bugs/idempotency-semantics** - Retryable policy rejection records dispatch_kind but does not settle. If ledger recovery inspects in-flight dispatch_kind, it may misinterpret as terminal rejection. Verify recovery semantics.

**performance/missing-timeout** - BeforeInboundPolicy runs inline with no timeout enforcement. Trait docstring warns about stall risk but provides no decorator or builder-level timeout. A slow policy could consume the entire 55s workflow budget.

**security/input-validation** - BeforeInboundPolicyRequest exposes external refs sufficient for fingerprinting. No rate-limit context available to policy implementations.

**tests/missing-error-path-test** - No test for policy error (BeforeInboundPolicyFailed) path through DefaultInboundTurnService. FakeBeforeInboundPolicy has force_failure() but no contract test exercises it.

### Medium (6)

- **bugs/error-propagation** - noop-policy rejection maps to Transient without permanent flag, causing indefinite retry cycle
- **performance/redundant-io** - prepare_user_message called redundantly after policy rewrite (DB round-trip wasted)
- **security/error-leak** - BeforeInboundPolicyFailed reason included raw in rejection ack message
- **tests/missing-integration** - No test for should_settle_ack behavior change with retryable rejection
- **tests/missing-error-path-test** - No serde test for UserMessagePayload validation through deserialization
- **conventions/expedient-hack** - FakeBeforeInboundPolicy has single global outcome, not per-request programmable

### Low (1)

- **conventions/sibling-pattern** - FakeInboundTurnService Default impl moved below trait impl, inconsistent with sibling pattern

---

*Review generated by code-review-skill v1*

crates/ironclaw_product_workflow/src/inbound_turn.rs
serrrfirat, line 209: **High - performance/redundant-io**

When policy returns `RewriteUserMessage`, `prepare_user_message` is called again on the rewritten envelope. But the rewrite only changes user_message text/attachments/trigger - the trigger is preserved. Since `prepare_user_message` calls `resolve_binding` (depends on adapter_id, installation_id, external refs - all unchanged) and `thread_scope_from_binding` (depends on trigger - unchanged), the second preparation produces identical results.

This wastes a binding resolution call (likely DB/network round-trip) on every policy rewrite.

**Fix:** Reuse the original `prepared` binding/scope:
```rust
BeforeInboundPolicyOutcome::RewriteUserMessage(payload) => {
    dispatch_envelope = envelope.with_rewritten_user_message(payload).map_err(...)?;
    (prepared, &dispatch_envelope)  // reuse original prepared
}
```

crates/ironclaw_product_workflow/src/policy.rs
serrrfirat, line 65: **High - performance/missing-timeout**

Trait docstring warns "a slow or stuck policy will stall inbound submissions" and says "Callers are responsible for enforcing wall-clock timeouts." But neither `DefaultInboundTurnService` nor `DefaultProductWorkflow` wrap the policy call in `tokio::time::timeout`.

The runner has a 55s workflow-level timeout, but that covers the entire pipeline. A policy hanging for 50s would consume the entire budget.

**Fix:** Add optional timeout parameter to `with_before_inbound_policy` or wrap `check_user_message` in `tokio::time::timeout` with a configurable default.

crates/ironclaw_product_workflow/src/policy.rs
serrrfirat, line 19: **High - security/input-validation**

`BeforeInboundPolicyRequest` includes `external_actor_ref`, `external_conversation_ref`, and `source_binding_key`. While the docstring says it "intentionally excludes the full trusted envelope," these refs are sufficient for user fingerprinting and correlation attacks.

No rate-limit context is available to the policy.

**Fix:** Add rate-limit context token or document that rate-limiting belongs in a decorator wrapping the policy trait.

crates/ironclaw_product_workflow/src/inbound_turn.rs
serrrfirat, line 173: **Medium - bugs/error-propagation**

When `NoopBeforeInboundPolicy` unexpectedly returns `Reject`, the error is `ProductWorkflowError::Transient { reason }`. Since `permanent` defaults to `false`, `terminal_ack_for_error` returns `None`, the idempotency ledger releases, and the adapter retries - cycling until lease timeout.

A Noop rejection is a bug, not a transient condition.

**Fix:** Return `Internal` error which settles terminal, or use a distinct error variant.

crates/ironclaw_product_workflow/src/workflow.rs
serrrfirat, line 114: **High - bugs/idempotency-semantics**

When policy returns `Reject` with retryable disposition, `should_settle_ack` returns `false`. The action gets `mark_dispatched(Rejected)` but is NOT settled - the ledger releases it.

`mark_dispatched` records a `Rejected` kind on an unsettled action. If ledger recovery inspects in-flight actions' `dispatch_kind`, it may misinterpret as terminal.

**Fix:** Verify `InMemoryIdempotencyLedger` recovery handles unsettled `Rejected` dispatch_kind correctly; add test.

crates/ironclaw_product_workflow/src/workflow.rs
serrrfirat, line 340: **Medium - tests/missing-integration**

The rename from `is_terminal_success_ack` to `should_settle_ack` adds `ack.is_durable_outcome()` to the gate, changing settlement for retryable rejections. The existing unit test only checks `NoOp` and `DeferredBusy`.

**Fix:** Add tests for `should_settle_ack` with both retryable and permanent `Rejected` acks.

crates/ironclaw_product_workflow/src/workflow.rs
serrrfirat, line 409: **Medium - security/error-leak**

`terminal_ack_for_error` maps `BeforeInboundPolicyFailed { permanent: true, reason }` to rejection with message `format!("before-inbound policy failed: {reason}")`.

If production policy sets `reason` with internal details (model names, classifier scores), these leak to adapter layer. Per `.claude/rules/error-handling.md`, no internal identifier may cross channel boundary.

**Fix:** Wrap `reason` in `RedactedString` and ensure sanitization at channel edge.

crates/ironclaw_product_workflow/src/fakes.rs
serrrfirat, line 287: **Medium - conventions/expedient-hack**

`FakeBeforeInboundPolicy` stores a single global outcome applied to all requests. Tests needing different outcomes for sequential requests must mutate the fake between calls.

Compare to `FakeInboundTurnService` which has `program_replay_outcomes(VecDeque)` for sequential behavior.

**Fix:** Add `program_outcomes(VecDeque<BeforeInboundPolicyOutcome>)` matching existing fake pattern.

crates/ironclaw_product_workflow/tests/product_workflow_contract.rs
serrrfirat, line 258: **High - tests/missing-error-path-test**

Contract tests cover allow, rewrite, reject, and retryable-reject. But no test where `check_user_message` returns `Err(BeforeInboundPolicyFailed { ... })`.

This error path is critical: `permanent=true` should settle as rejection, `permanent=false` should release and retry. `FakeBeforeInboundPolicy` has `force_failure()` but no contract test exercises it.

**Fix:** Add `before_inbound_policy_error_permanent_settles` and `before_inbound_policy_error_transient_releases`.

crates/ironclaw_product_adapters/src/inbound.rs
serrrfirat, line 109: **Medium - tests/missing-error-path-test**

`UserMessagePayload` now has custom `Deserialize` calling `Self::new` -> `validate()`. But no test verifies that deserializing an oversized text payload through serde fails.

**Fix:** Add `fn user_message_text_length_bounded_through_serde` deserializing oversized text via serde_json and asserting error.

serrrfirat (commented):
## Review Summary

| Metric | Value |
|---|---|
| **Verdict** | Changes Requested |
| **Findings** | 11 (4 High, 6 Medium, 1 Low) |
| **Reviewers** | security, bugs, performance, tests, conventions (all passed) |
| **Confidence threshold** | >= 50 |

## Findings by Category

### High (4)

**bugs/idempotency-semantics** - Retryable policy rejection records dispatch_kind but does not settle. If ledger recovery inspects in-flight dispatch_kind, it may misinterpret as terminal rejection. Verify recovery semantics.

**performance/missing-timeout** - BeforeInboundPolicy runs inline with no timeout enforcement. Trait docstring warns about stall risk but provides no decorator or builder-level timeout. A slow policy could consume the entire 55s workflow budget.

**security/input-validation** - BeforeInboundPolicyRequest exposes external refs sufficient for fingerprinting. No rate-limit context available to policy implementations.

**tests/missing-error-path-test** - No test for policy error (BeforeInboundPolicyFailed) path through DefaultInboundTurnService. FakeBeforeInboundPolicy has force_failure() but no contract test exercises it.

### Medium (6)

- **bugs/error-propagation** - noop-policy rejection maps to Transient without permanent flag, causing indefinite retry cycle
- **performance/redundant-io** - prepare_user_message called redundantly after policy rewrite (DB round-trip wasted)
- **security/error-leak** - BeforeInboundPolicyFailed reason included raw in rejection ack message
- **tests/missing-integration** - No test for should_settle_ack behavior change with retryable rejection
- **tests/missing-error-path-test** - No serde test for UserMessagePayload validation through deserialization
- **conventions/expedient-hack** - FakeBeforeInboundPolicy has single global outcome, not per-request programmable

### Low (1)

- **conventions/sibling-pattern** - FakeInboundTurnService Default impl moved below trait impl, inconsistent with sibling pattern

---

*Review generated by code-review-skill v1*

PR #3573, opened 2026-05-13
feat(reborn): add ironclaw_hooks framework foundation (#3524)

zmanian:
## Summary

Complete v1 of the Reborn loop hooks framework per #3524. Lands the
trust primitives, sealed decision types, dispatcher, port middleware,
predicate evaluator, telemetry, snippet injection, integration tests,
and the Reborn composition seam — plus a four-finding security audit,
an eight-item follow-up batch (FU1–FU8), and a design-time validation
pass (prior-art comparison, STRIDE threat model, real-hooks ergonomics
exercise, operator runbook) — **all in this PR**.

Design backstory: https://github.com/nearai/ironclaw/issues/3524#issuecomment-4439890144

## Design-time validation artifacts

Four companion docs in `crates/ironclaw_hooks/docs/`:

| Doc | Purpose |
|---|---|
| [`prior-art.md`](crates/ironclaw_hooks/docs/prior-art.md) | Compares ICLAW against LSM, eBPF/Tetragon, Envoy proxy-wasm, K8s admission webhooks, OPA, Chrome extensions, VS Code, and Tauri across 8 axes. Surfaces 7 areas where ICLAW stands out, 4 conventional choices to revisit, 3 weak-`why` divergences. |
| [`threat-model.md`](crates/ironclaw_hooks/docs/threat-model.md) | STRIDE per asset across 7 adversary classes and 6 ranked assets. ~35 attack vectors enumerated; the load-bearing "Installed cannot Allow" property (E1) confirmed type-enforced; all High gaps closed in this PR. |
| [`real-hooks-findings.md`](crates/ironclaw_hooks/docs/real-hooks-findings.md) | Three real hooks built against the public API (rate-cap predicate, NumericSum approval-gate, Trusted Rust PII-redaction mutator); ergonomic friction findings F1-F7, all closed in this PR. |
| [`operator-runbook.md`](crates/ironclaw_hooks/docs/operator-runbook.md) | Runtime recovery procedures for poisoned hooks, evaluator state pressure (D5), registration rejection, gate-ref lifecycle. |

## What's demonstrable end-to-end

An extension ships a `[[hooks]]` manifest entry. `HookRegistrar`
constructs a `PredicateBackedBeforeCapabilityHook`, registers it via
tier-specific installer, and the registrar wires the dispatcher
through `HookDispatcherBuilder::build_arc()`.
`RebornLoopDriverHostFactory::with_hook_dispatcher_factory(...)` plugs
in a per-run dispatcher factory. From there:

[...]

PR #3634, opened 2026-05-14
feat(hooks): WASM hook execution path (successor to #3573)

zmanian:
Successor PR from #3573. **Implementation** — full WASM execution path with sandbox, fuel/memory budgets, host-import bounds, and threat model.

## Scope
- Wire wasmtime for Installed-tier WASM hook bodies
- Host imports: `ic:hooks/context@1` (gate/prompt/observer context delivery), `ic:hooks/observer@1` (sink calls)
- Fuel + memory + wall-clock timeout + epoch interruption budgets
- `WasmVersionMaterial` newtype for identity-version composition
- Sync WASM execution runs on the tokio blocking pool so wall-clock timeouts actually cancel

## Design doc
[`crates/ironclaw_hooks/docs/successors/02-wasm-runtime.md`](https://github.com/nearai/ironclaw/blob/hooks-fu-wasm-runtime/crates/ironclaw_hooks/docs/successors/02-wasm-runtime.md)

## Review status
2026-05-19 multi-agent review (Henry) + Codex + Gemini findings addressed in commits `7efce2353 4c0844577 a625e0a0b 89b6e0f4e f7bdb14b9 571efdf67`. See https://github.com/nearai/ironclaw/pull/3634#issuecomment-4494960790 for the per-finding map.


zmanian:
## Codex review addressed (commit 0f63b31fb)

Verdict was REQUEST CHANGES; scope doc updated:

- **Critical: host-import + output budgets** — Wasm-side fuel/memory caps don't prevent memory exhaustion via many small import calls. Added explicit caps: max_sink_calls_per_invocation (~64), max_total_patch_bytes (4 KiB), max_observer_facts_per_invocation (~32), max_decision_calls_per_invocation (1).
- **Critical: module-substitution test claim was internally inconsistent** — corrected: different module bytes → different HookIds, so checkpoint replay refuses with `unknown_hook_id_at_replay`, NOT a registry-side collision.
- **Recommendation: pin ABI choice** — implementation will follow the existing `ironclaw_wasm` pattern (`wasmtime::Linker`) for parity with the tool-WASM stack.

serrrfirat:
## Summary
Reviewed PR #3634 only. Base `5793e4d90e1316adb93ec9c7edf6511d85f8873e` → head `a336d2a781d9dd14e4f6ca84475b1bd0df5dd2af`.

PR adds WASM hook execution path. Merge stance: blocker for v2-grade sandboxing until host-side copy bounds, output budgets, and install-time ABI validation are fixed.

## Findings

| 1 | Critical | Memory / Resource limits | `crates/ironclaw_hooks/src/wasm/runtime.rs:726`, `crates/ironclaw_hooks/src/wasm/runtime.rs:730` | `read_guest_string` trusts guest-supplied `len` and allocates `vec![0u8; len]` before any host-side cap. A WASM hook can pass huge positive `len` through imports like `add_envelope_snippet` / `add_milestone_metadata`; host attempts allocation before `memory.read` fails. Rust OOM aborts the host process, so untrusted WASM can escape fuel/wall/memory limits into host DoS. | Reject `len` above a small host cap before allocation, and validate `ptr + len` against guest memory size with checked arithmetic before copying. Add regression where oversized guest `len` returns `Malformed` without large allocation/OOM. |
| 2 | Medium | Resource limits / Boundary invariant | `crates/ironclaw_hooks/src/wasm/runtime.rs:662`, `crates/ironclaw_hooks/src/wasm/runtime.rs:673`, `crates/ironclaw_hooks/src/wasm/runtime.rs:688` | `add_milestone_metadata` charges only sink-call count. Unlike prompt snippets (`runtime.rs:649-657`), metadata values are not charged against aggregate byte budget, so guest can emit large metadata across allowed calls while docs/threat model claim host-import outputs have independent budgets. | Charge metadata bytes into the per-invocation total output budget or add a separate aggregate metadata-byte cap. Add negative test for oversized aggregate metadata output. |
| 3 | Medium | Untrusted module validation | `crates/ironclaw_hooks/src/wasm/runtime.rs:195`, `crates/ironclaw_hooks/src/wasm/runtime.rs:211`, `crates/ironclaw_hooks/src/wasm/runtime.rs:293`, `crates/ironclaw_reborn/tests/hooks_integration.rs:986` | `prepare()` compiles/caches module bytes but does not validate required imports or export type. ABI mismatch is deferred until first dispatch, and test coverage currently codifies that a bad-import WASM hook installs into a live host successfully before failing closed at invocation. Malformed untrusted modules should be rejected at registration/build time, not survive until live traffic. | During `prepare()`, instantiate against point-specific linker in a scratch store and resolve typed export, or perform equivalent import/export validation. Add install-time rejection tests for unsupported import, missing export, and wrong export type. |

## Security/data-flow notes
- Good: no ambient WASI/fs/network/time/RNG imports found in reviewed path.
- Good: gate WASM still cannot mint allow; imports expose deny/pause/pass style outputs.
- Main trust crossing: guest-controlled `(ptr, len)` into host allocation/copy.

- `runtime.rs:620-630` and `662-678`: host imports accept guest `ptr,len` then call `read_guest_string`.
- `runtime.rs:726-732`: allocation occurs before memory bounds/copy failure.
- `runtime.rs:281-299`: import/export validation occurs only during execution.
- `hooks_integration.rs:986-991`: bad-import module host build succeeds today.

## Missing tests
- Oversized guest string length must fail safely without OOM.
- Aggregate metadata byte budget overflow.
- Install-time rejection for unsupported import / missing export / wrong export signature.

henrypark133:
## Code Review — PR #3634 (WASM hook runtime)

**Verdict: REQUEST CHANGES** — architecturally sound, threat model is thorough, and the test suite covers the critical security properties. Three items must be addressed before merge.

### Overview
Wasmtime sandbox for `HookManifestBody::Wasm` entries in the Installed tier. Per-invocation `Store`, triple enforcement (fuel + epoch + memory limiter), host-import surface restricted to typed sinks, BLAKE3 module digest folded into `HookId` for content-addressing. Threat model and successor-PR design doc included.

### Issues

**Must fix before merge:**

1. **`#[path = "../../../ironclaw_wasm/src/limiter.rs"]` coupling** — two separately compiled copies of `WasmResourceLimiter`, invisible to Cargo tooling (`cargo check`, `cargo doc`, `cargo deny`, architecture linting), breaks silently if `limiter.rs` is moved. Preferred fix: micro-crate `ironclaw_wasm_limiter` that both `ironclaw_wasm` and `ironclaw_hooks` depend on. If deferred, add a `// COUPLING` comment and a CI grep that confirms the path still exists so refactors break immediately.

2. **Dead WASM arms in all three dispatch methods** — the early-return guard handles WASM (with `catch_unwind` + timeout), but the inner `match` also has WASM arms that execute **without** `catch_unwind` or timeout enforcement. If the guard is removed in a future refactor, WASM hooks silently lose panic safety and deadline enforcement. Annotate as `unreachable!("handled by early-return guard above")` or unify through the match.

3. **Dead inner arm for `BeforePromptHookImpl::RestrictedWasm` silently discards failure** — the dead arm maps `|_| ()`, throwing away the `WasmHookFailure` category. This makes issue #2 even more dangerous for the prompt path.

**Non-blocking:**

4. `reason_for_code` catch-all returns `"hook_wasm_denied"` for any unknown i32, while `ordinal_hint_for_code` and `metadata_key_for_code` return `None` — inconsistent. Unknown codes should return `None` → `Malformed`, not silently accept arbitrary guest-controlled codes with a default label.
5. Module cache (`WasmHookRuntime::modules`) grows without bound — document as intentionally unbounded for v1, or add a capacity guard.
6. Epoch ticker thread spawned per `WasmHookRuntime::new()` is non-cancellable — each test call to `wasm_dispatcher_from_wat` leaks a background thread. Document or use `OnceLock` for process-singleton shape.
7. `validate_module_abi` scratch `Store` has epoch deadline live — add `store.set_epoch_deadline(u64::MAX)` on the validation store to prevent epoch interruption during heavy contention.
8. `GateWasmDecision::Deny(&'static str)` — bare static str not composable with runtime strings; use `SanitizedReason` or the existing gate decision type.

### Missing test coverage
- Gate hook `pass` path (no explicit test; only deny is covered)
- Gate hook emitting zero decisions (`Malformed` path untested — critical contract)
- Gate hook issuing a second decision call (decision budget = 1)
- Observer fact budget overflow (`MAX_OBSERVER_FACTS_PER_INVOCATION = 32`)
- `BeforePrompt` happy path
- **Wall-clock/epoch timeout** — listed as required in the design doc's "Required tests" section but absent
- `validate_checkpoint_hook_ids_for_replay` with empty slice and multi-id cases

### Security
- No WASI, no filesystem/network/clock/RNG, `wasm_threads(false)`, point-specific linkers ✅
- `read_guest_string` correctly bounds `len` before indexing ✅
- W4 (guest cannot mint Allow) enforced at install time via missing import → `InvalidImports` ✅
- Checkpoint replay guard is correct; the digest-folding test doesn't prove the digest is load-bearing — add a comment asserting the version strings without digest would have been equal

crates/ironclaw_hooks/src/dispatch.rs
```diff
@@ -843,6 +949,22 @@ impl HookDispatcher {
         ctx: &BeforeCapabilityHookContext,
     ) -> Result<GateHookOutcome, HookFailureRecord> {
         let timeout = self.timeout;
+        if let BeforeCapabilityHookImpl::RestrictedWasm(h) = hook {
+            let run = AssertUnwindSafe(async { h.evaluate(ctx) }).catch_unwind();
+            return match tokio::time::timeout(timeout, run).await {
```
henrypark133: **High** — tokio::time::timeout does not interrupt synchronous WASM execution — worker thread blocked.

h.evaluate(ctx) is synchronous (calls through to runtime.execute which runs wasmtime synchronously). Wrapping it in async and tokio::time::timeout only stops the await from waiting — it does not cancel the synchronous call. The tokio worker thread remains blocked until WASM completes via fuel exhaustion or epoch trap. Under load, this can starve the runtime if multiple WASM hooks exceed their timeout simultaneously.

**Fix:** Run the synchronous WASM execution via tokio::task::spawn_blocking so it does not occupy an async worker thread.

henrypark133: **High** — Wall-clock timeout path unreachable in tests for all WASM hook types.

The Err(_elapsed) branch in the tokio::time::timeout match for WASM gate, prompt, and observer hooks is never exercised. All existing exhaustion tests use fuel or memory budgets that trip before the 50ms wall-clock timeout fires.

**Fix:** tests::hooks_integration::wasm_wall_clock_timeout_fails_closed_for_gate covering tokio::time::timeout elapsed branch

henrypark133: **High** — Wall-clock timeout path untested for WASM observer hooks.

Same as above but for the observer dispatch path. Observer hooks should fail isolated on timeout, but the Err(_elapsed) branch is never exercised by any test.

**Fix:** tests::hooks_integration::wasm_wall_clock_timeout_fails_isolated_for_observer covering tokio::time::timeout elapsed branch on observer path

crates/ironclaw_hooks/src/registrar.rs
```diff
@@ -27,6 +27,10 @@ use crate::identity::{ExtensionId, HookId, HookVersion};
 use crate::installed_hook::PredicateBackedBeforeCapabilityHook;
 use crate::manifest::{HookManifestBody, HookManifestEntry, HookManifestKind, HookManifestScope};
 use crate::registry::{HookBindingScope, HookPointSpec};
+use crate::wasm::{
```
henrypark133, line 30: **Medium** — Registrar WASM happy-path install (all three hook kinds) not unit-tested.

registrar.rs has install_wasm_body_requires_runtime testing the error path when no runtime is configured, but no unit test confirms that a WASM body installs successfully through the registrar for BeforeCapability, BeforePrompt, and observer kinds.

**Fix:** crates::ironclaw_hooks::registrar::tests::install_wasm_before_capability_succeeds covering successful WASM install via registrar

crates/ironclaw_hooks/src/dispatch.rs
```diff
@@ -843,6 +949,22 @@ impl HookDispatcher {
...
+                Ok(Ok(Ok(outcome))) => Ok(outcome),
+                Ok(Ok(Err(failure))) => Err(self.classify_wasm_failure(binding, failure)),
+                Ok(Err(_panic)) => {
+                    Err(self.classify_failure(binding, FailureCategory::Panic, "hook panicked"))
+                }
+                Err(_elapsed) => Err(self.classify_failure(
+                    binding,
+                    FailureCategory::Timeout,
+                    "hook exceeded dispatch timeout",
```
henrypark133, line 967: **Medium** — Three run_*_hook dispatch functions duplicate timeout/panic/WASM-guard boilerplate.

run_before_capability_hook, run_before_prompt_hook, and run_observer_hook each repeat the same AssertUnwindSafe+catch_unwind+tokio::timeout wrapper plus an early-return guard for WASM variants with unreachable match arms. A future change to panic isolation or timeout handling must be applied in three places.

**Fix:** Extract a generic run_hook_with_isolation<H, O>(hook, binding, ctx, timeout, run_wasm, run_native) helper that handles the timeout/panic/WASM-guard once and dispatches to type-specific closures.

crates/ironclaw_hooks/src/dispatch.rs
```diff
@@ -360,6 +387,32 @@ impl HookDispatcher {
         Ok(())
     }
 
+    /// Install an Installed-tier WASM `before_capability` hook through the
```
henrypark133, line 390: **Low** — ordered_bindings clones all bindings into a new Vec every dispatch.

ordered_bindings() acquires the registry lock, clones every active binding for a point into a new Vec, sorts it, and returns it. This is called once per dispatch point. Each call allocates a Vec of cloned HookBinding structs.

**Fix:** Cache the sorted order and invalidate only on registry mutation, or use a BTreeMap keyed by HookOrderKey for pre-sorted iteration.

crates/ironclaw_hooks/src/dispatch.rs
```diff
@@ -31,6 +31,9 @@ use crate::sink::{
 };
 use crate::telemetry;
 use crate::trust::HookTrustClass;
+use crate::wasm::{
```
henrypark133, line 34: **Low** — dispatch.rs exceeds 1400 lines with WASM-specific branches inflating each dispatch function.

The file has grown significantly with WASM early-return guards in each of the three run_*_hook functions, three install_installed_wasm_* installers with duplicated binding construction, and classify_wasm_failure.

**Fix:** Move WASM-specific dispatch into the wasm module behind a trait (WasmDispatchable) so dispatch.rs delegates rather than branching.

crates/ironclaw_hooks/src/registrar.rs
```diff
@@ -233,28 +244,87 @@ impl HookRegistrar {
                     // the only path that knows the manifest's `priority`
                     // field, so we set it post-insert.
                     dispatcher.set_binding_priority(hook_id, entry.priority);
+                    Ok(hook_id)
                 }
-                other => {
-                    return Err(HookError::RegistryConstruction(format!(
-                        "predicate body is only supported for `before_capability` hooks; \
+                other => Err(HookError::RegistryConstruction(format!(
```
henrypark133, line 249: **Low** — wasm_hook_identity_version uses format! string instead of typed version material.

The WASM module digest is folded into the version string via format!("{extension_version}+wasm:{module_digest_hex}"). This is a stringly-typed composite passed to HookId::derive where other paths use a typed ExtensionId + version string.

**Fix:** Introduce a WasmVersionMaterial newtype that encapsulates the extension_version + module_digest composition.

henrypark133 (changes requested):
## Code Review (multi-agent)

**Intent:** Full implementation of WASM runtime for hooks — sandbox, budget enforcement, threat model (despite PR body claiming "scope doc only")

**Stats:** 29 findings (from 29 raw, 29 after dedup) across 5 files. Reviewers run: security, bugs, performance, tests, conventions, design. Reviewers failed: none. Body-only: 0

### Bugs

1. **Critical** WASM hooks receive no hook context — _ctx parameter unused, export invoked with () (`crates/ironclaw_hooks/src/wasm/runtime.rs`, confidence 100) — anchor: crates/ironclaw_hooks/src/wasm/runtime.rs:253
   execute_gate, execute_prompt, and execute_observer all accept a _ctx parameter that is never used. The WASM export is called with `()`, giving the module zero access to capability name, arguments, or tenant ID.
   **Fix:** Pass context data to the WASM export via serialized context blob in guest memory or host imports.

2. **High** Epoch ticker background thread never terminates — infinite loop with no shutdown signal (`crates/ironclaw_hooks/src/wasm/runtime.rs`, confidence 100) — anchor: crates/ironclaw_hooks/src/wasm/runtime.rs:483
   Thread is never signaled to stop, never joined, and holds an Engine clone. Leaks thread and engine on runtime drop.
   **Fix:** Use a shutdown channel (AtomicBool or oneshot) checked inside the loop.

3. **High** tokio::time::timeout does not interrupt synchronous WASM execution — worker thread blocked (`crates/ironclaw_hooks/src/dispatch.rs`, confidence 100) — anchor: crates/ironclaw_hooks/src/dispatch.rs:953
   Wrapping synchronous WASM in async + timeout only stops the await, not the call. Worker thread remains blocked.
   **Fix:** Run via tokio::task::spawn_blocking so it does not occupy an async worker thread.

### Security

4. **Medium** Compiled WASM module cache grows without bound — no eviction or size limit (`crates/ironclaw_hooks/src/wasm/runtime.rs`, confidence 100) — anchor: crates/ironclaw_hooks/src/wasm/runtime.rs:181
   Mutex<HashMap> never evicts entries. Can exhaust host memory with many distinct WASM modules.
   **Fix:** Add max_entries cap with LRU eviction.

### Performance

5. **Medium** TOCTOU race in prepare() causes duplicate WASM module compilation under concurrency (`crates/ironclaw_hooks/src/wasm/runtime.rs`, confidence 80) — anchor: crates/ironclaw_hooks/src/wasm/runtime.rs:208
   Check-compile-insert pattern releases lock between check and compile, causing redundant compilation.
   **Fix:** Use concurrent map with compute-if-absent semantics or hold write lock across both steps.

6. **Medium** O(n) linear scan for hook_id lookups in registry (`crates/ironclaw_hooks/src/registry.rs`, confidence 85) — anchor: crates/ironclaw_hooks/src/registry.rs:146
   poison(), is_poisoned(), contains_hook() all iterate all bindings. Hot-path dispatch O(n) per hook.
   **Fix:** Add HashMap<HookId> index for O(1) lookup.

### Bugs (continued)

7. **Medium** Post-call deadline check can misclassify successful execution as timeout (`crates/ironclaw_hooks/src/wasm/runtime.rs`, confidence 75) — anchor: crates/ironclaw_hooks/src/wasm/runtime.rs:322
   deadline_exceeded() called again after Ok(()), nanosecond window can misclassify.
   **Fix:** Record deadline status before the call.

8. **Medium** add_milestone_metadata returns wrong error message for oversized string (`crates/ironclaw_hooks/src/wasm/runtime.rs`, confidence 100) — anchor: crates/ironclaw_hooks/src/wasm/runtime.rs:757
   Error says 'exceeded total prompt-patch byte budget' when actual issue is string doesn't fit in u32.
   **Fix:** Change error message to 'wasm hook metadata value exceeds maximum size'.

### Tests

9. **High** Wall-clock timeout path unreachable in tests for all WASM hook types (`crates/ironclaw_hooks/src/dispatch.rs`, confidence 100) — anchor: crates/ironclaw_hooks/src/dispatch.rs:953
   All existing tests use fuel/memory budgets that trip before 50ms timeout fires.
   **Fix:** Add test with low-fuel, long-wall-clock WASM module.

10. **High** Wall-clock timeout path untested for WASM observer hooks (`crates/ironclaw_hooks/src/dispatch.rs`, confidence 100) — anchor: crates/ironclaw_hooks/src/dispatch.rs:1113
    Same gap for observer dispatch path.
    **Fix:** Add test covering observer timeout with fail-isolated assertion.

11. **Medium** Memory exhaustion for WASM observer hooks not tested (`crates/ironclaw_hooks/src/wasm/runtime.rs`, confidence 75) — anchor: crates/ironclaw_hooks/src/wasm/runtime.rs:289
    No test proving memory-exhausting observer hook fails isolated.
    **Fix:** Add test with WASM_MEMORY_EXHAUSTION fixture and AfterCapability kind.

12. **Medium** add_milestone_metadata happy path not tested (`crates/ironclaw_hooks/src/wasm/runtime.rs`, confidence 75) — anchor: crates/ironclaw_hooks/src/wasm/runtime.rs:730
    Only exercised in overflow fixture, no test for valid call producing correct patch.
    **Fix:** Add test with valid metadata emission and assert HookPatch::AddMilestoneMetadata.

13. **Medium** WasmResourceLimiter::memory_grow_failed rollback path not tested (`crates/ironclaw_wasm_limiter/src/lib.rs`, confidence 75) — anchor: crates/ironclaw_wasm_limiter/src/lib.rs:67
    Rollback after approved-then-failed growth never exercised.
    **Fix:** Add test calling memory_growing then memory_grow_failed, assert memory_used returns to 0.

14. **Medium** Registrar WASM happy-path install not unit-tested (`crates/ironclaw_hooks/src/registrar.rs`, confidence 75) — anchor: crates/ironclaw_hooks/src/registrar.rs:255
    Error path tested but no unit test for successful WASM install through registrar.
    **Fix:** Add unit test with in-memory resolver and minimal valid WASM module.

### Design

15. **Medium** Three run_*_hook dispatch functions duplicate timeout/panic/WASM-guard boilerplate (`crates/ironclaw_hooks/src/dispatch.rs`, confidence 75) — anchor: .claude/rules/testing.md:27-52
    Same AssertUnwindSafe+catch_unwind+timeout wrapper repeated in three places.
    **Fix:** Extract generic run_hook_with_isolation helper.

zmanian:
## Review feedback addressed

Pushed 6 commits to `hooks-fu-wasm-runtime`. Per-finding status against Henry's 2026-05-19 multi-agent review + bot threads:

### Resolved in code

| ID | Finding | Commit |
|---|---|---|
| A1 | **Critical: WASM hooks receive no context** — `execute_gate`/`execute_prompt`/`execute_observer` accepted `_ctx` and never used it | `7efce2353 fix(hooks): plumb context to WASM hooks + runtime hardening` |
| A2 | Epoch ticker leaks thread + Engine on Drop | `7efce2353` (cancellation-driven ticker with Drop join) |
| A3 | `tokio::time::timeout` does not cancel sync wasmtime | `4c0844577 fix(hooks): run WASM hooks on the blocking pool` |
| B4 | Module cache unbounded (`Mutex<HashMap>`) | `7efce2353` (`LruCache` bounded by `MODULE_CACHE_CAPACITY`) |
| codex P1 | `read_guest_string` allocates before bounds check | `7efce2353` (guest-len validated against `MAX_GUEST_STRING_BYTES` + memory length before allocation) |
| codex P2 | `add_milestone_metadata` no byte budget | `7efce2353` (per-value + cumulative caps, distinct error variant) |
| C7 | TOCTOU in `prepare()` | `7efce2353` (read-check / write-double-check pattern) |
| C8 | O(n) `hook_id` lookup | `a625e0a0b perf(hooks): O(1) hook-id lookup via side index` |
| D9 | Race between `Ok(())` and `deadline_exceeded()` reclassification | `7efce2353` |
| D10 | `add_milestone_metadata` wrong error variant on overflow | `7efce2353` |
| E11/E12 | Wall-clock timeout branch unreachable in tests (gate + observer) | `89b6e0f4e test(hooks): wall-clock timeout, observer memory, limiter rollback, registrar happy path` |
| E13 | No observer memory-exhaustion test | `89b6e0f4e` |
| E14 | `add_milestone_metadata` happy path untested | `89b6e0f4e` |
| E15 | `memory_grow_failed` rollback untested | `89b6e0f4e` |
| E16 | Registrar happy-path untested | `89b6e0f4e` |
| H20 | `wasm_hook_identity_version` stringly-typed | `f7bdb14b9 refactor(hooks): typed WASM version material, reconcile design doc` (`WasmVersionMaterial` newtype) |
| G | Design doc reconciliations (`FailIsolated` definition, `spawn_blocking` note, evaluate vs WitToolRuntime) | `f7bdb14b9` |
| no-panics CI | `.expect()` on `LruCache` capacity | `571efdf67 fix(hooks): drop .expect() in WASM module cache capacity` (const-match) |

### Deferred with rationale

- **F17 (`run_hook_with_isolation` helper)**: kept the three call sites separate because the post-`spawn_blocking` shape made the previously-duplicated path collapse to ~6 lines each; abstracting it now would obscure the per-hook-type budget plumbing. Open to revisit if a fourth dispatch path lands.
- **H18 (`ordered_bindings()` sort cache)**: not in the dispatch hot path after the event-kind index (PR #3640, commit `7368139fc`); per-call cost is dominated by the registry walk, which is now keyed. Will revisit if profiling shows it.
- **H19 (split `dispatch.rs` >1400 lines)**: pure refactor; deferred to a separate PR so this review focuses on behaviour changes.

PR body updated to reflect the shipped implementation (was stale "scope doc only").

— Zaki

crates/ironclaw_hooks/src/dispatch.rs
zmanian: Addressed in `4c0844577 fix(hooks): run WASM hooks on the blocking pool`. WASM hook execution now goes through `tokio::task::spawn_blocking` so the async timeout drops the join handle rather than trying to interrupt a sync call on a runtime worker thread; the wasmtime side still relies on fuel/epoch/memory budgets for in-call termination.

zmanian: Addressed in `89b6e0f4e test(hooks): wall-clock timeout, observer memory, limiter rollback, registrar happy path`. Added a WASM fixture that loops with ample fuel/memory so the 50ms wall-clock branch fires for the gate and prompt paths (covered in `dispatch.rs` tests).

zmanian: Addressed in `89b6e0f4e test(hooks): wall-clock timeout, observer memory, limiter rollback, registrar happy path`. Same low-fuel/long-wallclock fixture is now exercised through the observer dispatch path and asserts the binding fails isolated (does not poison).

crates/ironclaw_hooks/src/registrar.rs
zmanian: Addressed in `89b6e0f4e test(hooks): wall-clock timeout, observer memory, limiter rollback, registrar happy path`. Added registrar unit tests covering successful WASM install for all three hook kinds (gate / prompt / observer) using an in-memory resolver and a minimal valid module.

crates/ironclaw_hooks/src/dispatch.rs
zmanian: Deferred with rationale. After moving WASM execution to `spawn_blocking` (4c0844577), the per-dispatch boilerplate collapsed to ~6 lines on each path, and extracting a generic `run_hook_with_isolation` would obscure the per-hook-type budget/sink plumbing that's now the bulk of each function. Happy to revisit if a fourth dispatch path lands or the divergence shrinks further.

crates/ironclaw_hooks/src/dispatch.rs
zmanian: Deferred. The dispatch hot path no longer iterates all bindings after the event-kind index landed in #3640 (`7368139fc`); per-call cost is dominated by the keyed registry walk, and `ordered_bindings()` is called once per kind with the already-filtered slice. Would prefer to revisit if profiling shows it's still a bottleneck.

crates/ironclaw_hooks/src/dispatch.rs
zmanian: Deferred to a follow-up PR. This is a pure refactor (no behaviour change), and splitting it here would mix mechanical file moves with the WASM behaviour changes under review. Tracking as a cleanup once this lands.

crates/ironclaw_hooks/src/registrar.rs
zmanian: Addressed in . Replaced the `format!("{extension_version}+wasm:{module_digest}")` with a `WasmVersionMaterial` newtype so the WASM-vs-non-WASM identity contribution is encoded in the type system rather than string concatenation.

zmanian:
@henrypark133 friendly ping — all 15 findings from your 5-19 review are addressed (12 in code across commits ending at 571efdf67; F17/H18/H19 deferred to follow-ups with rationale on each thread). I just posted per-inline-comment replies mapping each finding to its fix commit. CI green. Could you take another pass when you have a moment? Thanks!

henrypark133 (approved):
**Single-agent review.**

## Verdict: APPROVE

**Approving with follow-up:** all findings are Medium or Low and should be addressed in a follow-up PR (not blocking).

---

## Summary

**0 High / 3 Medium / 2 Low**

This PR lands the WASM hook execution path for `Installed`-tier hooks in `ironclaw_hooks`, extracts `WasmResourceLimiter` into its own micro-crate (`ironclaw_wasm_limiter`), and adds the per-build dispatcher isolation tests. The security posture is strong: the linker surface is hand-wired with zero WASI, no filesystem, network, environment, clock, or RNG exports; guest strings are always copied via `(ptr, len)` pairs with byte-length bounds; gate decisions use a `&'static str` lookup table so no guest-authored string can escape into the decision reason; and the double-checked compile cache is correctly structured. The epoch-ticker shutdown (prior HIGH) and the blocking-pool dispatch pattern (prior HIGH) are both addressed correctly.

---

## Medium Findings

### M1 — `WasmBudget` has no minimum-value floor; `fuel=0` and `memory_mb=0` silently produce always-trap or always-deny hooks

`WasmBudget` accepts any `u64`/`u32` from the manifest without clamping (`crates/ironclaw_hooks/src/manifest.rs:154`). Consequences:
- `fuel = 0` → `store.set_fuel(0)` gives the module zero instructions; the very first opcode traps as `FailureCategory::Panic` and poisons the slot for the run.
- `memory_mb = 0` → `WasmResourceLimiter::new(0)` denies the first page allocation; the module fails to instantiate on first live dispatch.

The hooks fail closed either way, but a misconfigured or adversarial manifest can silently make every hook invocation a fail-closed trap with no diagnostic pointing to the budget. `manifest.rs::validate()` should reject `fuel == 0` and `memory_mb == 0` with a clear `HookManifestValidationError` (same pattern as the existing zero-duration window check).

**Confidence: 80%**

### M2 — `HookBindingScope::SameTenant::permits()` is functionally identical to `Global`; no tracking comment for future multi-tenant enforcement

In `crates/ironclaw_hooks/src/registry.rs` around line 100, `permits()` returns `true` for both `Global` and `SameTenant` unconditionally. The CLAUDE.md documents this as intentional for v1 ("today the dispatcher is per-tenant already"), but there is no inline `TODO` linking to a tracking issue. If the dispatcher ever becomes multi-tenant and this arm is not updated, hooks with `SameTenant` scope will silently fire across tenant boundaries.

Request: add a `// TODO(#NNNN): add per-invocation tenant comparison once dispatcher is multi-tenant` comment on the `SameTenant` arm.

**Confidence: 75%**

### M3 — `validate_module_abi` scratch store runs with the real `wall_ms` budget; Wasm start functions that execute during instantiation consume fuel and could hit the epoch deadline with an opaque error

In `crates/ironclaw_hooks/src/wasm/runtime.rs`, the `validate_module_abi` function at line 727 calls `configure_store` with the real `WasmHookLimits`, then calls `linker.instantiate(...)`. Module instantiation does not execute the named export, but it *does* execute any Wasm start function. A start function that exhausts fuel or exceeds the epoch deadline will produce an opaque error from `InvalidImports` rather than a diagnostic that points to the budget or the start section. The fail-closed behavior is correct; the operator experience is not.

Request: either document that start-function execution is intentionally bounded by the same budget, or reject modules with a start section with an explicit `InvalidImports { reason: "start section not permitted" }` before instantiation.

**Confidence: 70%**

---

## Low Findings

### L1 — Orphaned `spawn_blocking` tasks after outer timeout have a bounded but undocumented aggregate drain time

In `crates/ironclaw_hooks/src/dispatch.rs::run_wasm_blocking` (~line 1183), when `tokio::time::timeout` fires first, the `JoinHandle` is dropped and the blocking task drains until the wasmtime epoch interrupt fires. The per-task drain bound is `≤ EPOCH_TICK_INTERVAL` (10 ms). Under sustained hook-abuse with many concurrent invocations this means up to `N_hooks * 10ms` of blocking-pool threads in the draining state simultaneously. The existing comment describes the behavior correctly; adding the aggregate bound estimate would make the resource exhaustion profile explicit in the threat model (W6).

**Confidence: 65%**

### L2 — `WasmResourceLimiter::memory_grow_failed` rollback test only covers the non-zero growth case

The new test in `crates/ironclaw_wasm_limiter/src/lib.rs` covers a non-zero approved grow that is then rolled back by `memory_grow_failed`. A companion case for a zero-growth probe (`desired == current`, `growth = 0`) followed by `memory_grow_failed` would confirm that rollback is a no-op and make the zero branch explicitly covered. Not a live bug — `memory_used -= 0` is harmless — but the coverage gap is easy to close.

**Confidence: 55%**

---

## Positive Notes

- Double-checked LRU compile cache in `prepare()` is correctly structured: fast path inside the lock, compile outside the lock, re-check on the slow path. Content-addressed by BLAKE3 digest.
- Epoch-ticker `Drop` impl cleanly signals shutdown, joins the thread, and releases the `Engine` clone — prior HIGH finding fully addressed.
- `GateWasmDecision` reason strings are `&'static str` table lookups, never guest-authored; no injection path from guest memory into decision reasons.
- `serde_json::json!` macro in the context-blob serializers handles escaping; tenant IDs containing JSON-special characters cannot corrupt the blob.
- `run_wasm_blocking` correctly offloads synchronous wasmtime work onto the blocking pool — prior HIGH about blocking-in-async fully addressed.
- Per-build dispatcher isolation test pair in `crates/ironclaw_reborn/tests/hooks_integration.rs` directly pins the critical run-isolation invariant with a positive and a legacy-compatibility case.
- Verified: no `.unwrap()` / `.expect()` in production paths; all new error types use `thiserror`; `HookBindingScope` and `HookPointSpec` derive `#[serde(rename_all = "snake_case")]`.

zmanian:
Replaced by #3920 — base branch (`hooks-foundation-01`) was deleted on the #3573 squash-merge, this PR auto-closed.

The new PR combines this scope with the WASM scaffolding that wasn't included in the foundation squash (originally a separate-but-coupled stack). Rebased onto `reborn-integration` with conflict resolutions documented in #3920's body — see particularly the composition with #3911 (`ordered_bindings_with_poison_snapshot`), #3912 (newtype privatization), and #3913 (lazy-input plumbing) that all landed after this branch was cut.

The original review thread + your approval-with-followup remain visible here for context; deferred items (F17 / H18 / H19) are explicitly carried forward in #3920's description.

PR #3635, opened 2026-05-14
feat(hooks): persistent predicate counter backend (successor to #3573)

zmanian:
Successor PR from #3573. Persistent predicate counter backend.

## Status
Scope doc shipped + in-memory backend trait extraction landed in this PR. Durable Postgres / libSQL impls remain in scope for a follow-up — see the design doc for the schema and clock-type story.

## Scope of this PR
- New `PredicateStateBackend` trait (crate-private until the durable contract stabilizes).
- `InMemoryPredicateStateBackend` extracted out of `PredicateEvaluator` against the new trait.
- `PredicateEventId` typed identity with format validation, host-assigned trust boundary, and a synth fallback for callers that don't supply one.
- `caller_event_id` threading through `BeforeCapabilityHookContext` so durable callers can drive replay dedup end-to-end.
- O(1) `running_sum` for NumericSum predicates.
- Per-tenant LRU quota in addition to the global `MAX_HISTORY_KEYS` cap.
- `evict_older_than` implemented for the in-memory backend so operator reaper tasks can reclaim idle-key memory.

## Design doc
[`crates/ironclaw_hooks/docs/successors/03-persistent-counter.md`](https://github.com/nearai/ironclaw/blob/hooks-fu-persistent-counter/crates/ironclaw_hooks/docs/successors/03-persistent-counter.md)

## Threat-model
- D5 (LRU eviction at MAX_HISTORY_KEYS = 8192) still applies in-memory; the per-tenant quota at MAX_HISTORY_KEYS / 4 is defense-in-depth so noisy tenants can't reset quiet tenants' counters.
- Multi-host replay dedup and restart survival require the durable backend (follow-up); `PredicateEvaluator::warn_in_memory_backend_active_in_production` flags that surface from host startup.


zmanian:
## Codex review addressed (commit d5f68de30)

All three Critical findings resolved:

1. **Backend not wired**: module is now registered in `lib.rs` and `PredicateEvaluator` delegates to `Arc<dyn PredicateStateBackend>` via `with_backend(...)`. Default constructor keeps in-memory behavior; existing 154 evaluator tests pass unchanged.

2. **Trait can't support durable contract**: trait now returns `Result<_, PredicateBackendError>`, accepts a `PredicateEventId` for replay dedup, and explicitly documents the atomicity contract (single lock / transaction across record + read).

3. **Race-prone separate calls**: each `record_*` method performs the write AND returns the resulting in-window count/sum in a single atomic operation. The in-memory backend holds the mutex across both halves; durable backends must use a single transaction.

Schema recommendation (event_id PK too broad): noted in the scope doc — the implementation slice for the Postgres backend will use `(tenant_id, hook_id, capability[, field], event_id)` as a composite key.

Sync trait + Instant timestamps: kept for v1 to minimize call-site churn; durable backends will use a SystemTime-based async companion trait, tracked in the scope doc.

Tests: 160 unit total (+6 new in `predicate_state`):
- 4 backend-shape tests (counts within window, trim, value sum, tenant isolation)
- 2 event-id dedup tests (invocations + values)

crates/ironclaw_hooks/src/predicate_state.rs
```diff
@@ -0,0 +1,518 @@
...
+                break;
+            }
+        }
+        let duplicate = bucket.recent_ids.contains(event_id);
+        if !duplicate {
+            bucket.entries.push_back(now);
+            bucket.recent_ids.push_back(event_id.clone());
+            while bucket.recent_ids.len() > RECENT_EVENT_ID_CAP {
+                bucket.recent_ids.pop_front();
+            }
```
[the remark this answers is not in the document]
zmanian: Fixed in 9f9513cac — dedup memory is now intrinsic to in-window entries. Each entry stores `(timestamp, event_id)`, dedup is `any(in-window entry has this id)`. The fixed-size 256 ring is gone, so dedup memory equals the in-window entry set — no silent loss possible under any throughput. New regression test `dedup_memory_covers_full_window_under_high_throughput` pushes 512 distinct events then replays event-0; pre-fix this would have re-counted, post-fix it's a no-op.

crates/ironclaw_hooks/src/predicate_state.rs
```diff
@@ -0,0 +1,518 @@
...
+/// sample. Conservative LRU approximation.
+fn evict_lru_invocation(
+    history: &mut HashMap<InvocationKey, InvocationBucket>,
+    evictions: &AtomicU64,
+) {
+    let victim = history
+        .iter()
+        .filter_map(|(k, v)| v.entries.front().map(|ts| (k.clone(), *ts)))
+        .min_by_key(|(_, ts)| *ts)
+        .map(|(k, _)| k);
```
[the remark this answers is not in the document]
zmanian: Fixed in 9f9513cac. Two-part fix: (1) `record_*` now drops empty buckets eagerly via `history.remove(key)` (mostly defense-in-depth — under the new dedup design the record path can't actually leave a bucket empty). (2) `evict_lru_*` preferentially evicts empty buckets first, only falling back to the oldest-timestamp scan if no empty bucket exists. Regression test `lru_evicts_empty_buckets_first` crafts an empty bucket alongside a live one and asserts the empty one is the eviction victim.

serrrfirat:
## Summary
Reviewed PR #3635 only. Base `5793e4d90e1316adb93ec9c7edf6511d85f8873e` → head `df505aea23e1c8ece6fa97a9a7b505cc7efe359f`.

PR adds persistent predicate counter backend plumbing. In-memory backend looks concurrency-safe in process, but caller/public contracts do not yet support replay-safe durable semantics. Merge stance: blocking correctness finding plus public API concern.

## Findings

| 1 | High | Correctness / Idempotency | `crates/ironclaw_hooks/src/evaluator.rs:145`, `crates/ironclaw_hooks/src/evaluator.rs:236` | Evaluator synthesizes a fresh `event_id` for every predicate evaluation. The backend contract dedupes duplicate `event_id`, but same logical invocation retried/replayed through the evaluator gets a new ID and counts again. That defeats durable replay/restart/idempotency semantics this PR is preparing. | Plumb stable runtime event identity through evaluator/caller path before durable backend use (for example `evaluate_with_event_id` or equivalent), and use that same ID for invocation/value counter writes. Add caller-boundary replay test: same stable runtime event retried must not increment count/sum twice. |
| 2 | Medium | Architecture / Persistence semantics | `crates/ironclaw_hooks/src/predicate_state.rs:33`, `crates/ironclaw_hooks/src/predicate_state.rs:109`, `crates/ironclaw_hooks/src/lib.rs:26` | Public `predicate_state` module exposes `PredicateStateBackend`, but trait methods take `Instant`. Docs note `Instant` is process-local and not serializable, so external durable backends cannot faithfully implement cross-process/restart semantics on this public API without a future breaking contract change. | Keep predicate-state backend trait internal until durable API is stable, or switch public contract now to durable/wall-clock time type (or separate sealed in-memory trait vs durable trait) before exposing it. Add contract coverage for durable-time semantics if public API remains. |

## Security/data-flow notes
- Tenant/key partitioning exists in predicate keys; no direct auth/scope bypass found in changed evaluator path.
- Backend errors still fail closed.

- `predicate_state.rs:100-108`: backend replay-refusal contract depends on stable duplicate `event_id`.
- `evaluator.rs:145-148`: caller path passes synthetic ID to `record_invocation`.
- `evaluator.rs:236-250`: synthetic ID includes process-local counter, guaranteeing uniqueness across repeated identical calls rather than idempotency.
- `predicate_state.rs:33-38`: docs explicitly call out `Instant` as process-local/non-serializable while module is public.

## Missing tests
- Caller-boundary replay/idempotency test through evaluator with a stable runtime event ID.
- Duplicate event ID no-op through evaluator-facing API, not backend-only tests.
- Public durable backend contract test if `PredicateStateBackend` remains exported.

henrypark133 (changes requested):
What looks good:
- The backend trait cleanly separates predicate evaluation from state storage.
- The in-memory backend keeps record-and-read atomic within the process.
- The prior fixed-size dedup and empty-bucket LRU issues look addressed in the backend itself.

Findings:
1. High - `crates/ironclaw_hooks/src/evaluator.rs:145`: `PredicateEvaluator` synthesizes a fresh `PredicateEventId` on every evaluation by mixing in a process-local counter, so the production installed-hook path never forwards a stable event identity into `PredicateStateBackend`. Why it matters: replay dedup is presented as load-bearing, but real retries/replays through `PredicateBackedBeforeCapabilityHook -> PredicateEvaluator` always get a new ID and are counted again. Expected fix direction: thread a stable runtime/event identity through the hook context or add an evaluator entrypoint that accepts one, with caller-level tests proving duplicate event IDs are no-ops.

Summary:
- Prior feedback status: backend-internal Codex P1s appear fixed, but this end-to-end replay-dedup gap remains.
- Residual risk: durable Postgres/libSQL persistence is deferred, so cross-process and restart guarantees remain future work.

henrypark133:
## Code Review — PR #3635 (persistent predicate counter backend)

**Verdict: REQUEST CHANGES** — above-average quality with a credible extensibility story, but several correctness and safety issues need addressing before merge.

### Overview
Extracts sliding-window counter bookkeeping from `PredicateEvaluator` into a new synchronous `PredicateStateBackend` trait + `InMemoryPredicateStateBackend`. Adds atomic record-and-read, event-id replay dedup (scoped to the in-window entry set, fixing the old fixed-ring P1 bug), and LRU eviction at 8192 keys. 1050 lines added, 167 removed.

### Issues

**Must fix before merge:**

1. **Dedup scan is O(n) in in-window entry count** — linear scan of `bucket.entries` per call. At high throughput (1-hour window, thousands of entries), this serializes all evaluations under the mutex. Fix: companion `HashSet<PredicateEventId>` inside each bucket; remove from set when paired `(ts, id)` is popped from the deque. O(1) dedup, invariant preserved.

2. **Mutex poison handling calls `.expect("predicate history mutex poisoned")`** — a panic in any thread cascades to all subsequent callers panicking. For fail-closed semantics, use `.unwrap_or_else(|e| e.into_inner())` (recover poisoned lock) or add a `PoisonError` variant to `PredicateBackendError`.

3. **`caller_event_id` (inner `String`) has no format validation** — empty string or null bytes reach the backend unchecked. Validate non-empty in `with_caller_event_id` and document the expected format. Durable backends constraining to UUID will fail cryptically otherwise.

4. **`with_backend` is `#[allow(dead_code)]` instead of `#[cfg(test)]`** — exists in release builds, inviting future callers to inject unvalidated backends outside tests.

**Important before durable backend PRs:**

5. **Trait has no `evict_older_than` stub** — scope doc promises this for durable backends. Adding a default-impl no-op now locks the signature before the first durable backend PR, avoiding a breaking trait-object change later.

6. **Trait is synchronous but durable backends need async** — document explicitly that the sync surface is intentional and that an `async_trait` migration is required for durable impls, so implementors don't assume the sync contract is permanent.

**Non-blocking nits:**

7. `synth_event_id` lives in `evaluator.rs` but its types/consumers live in `predicate_state.rs` — move it there for co-location and testability.
8. `// safety:` comment on non-`unsafe` blocks — use `// RATIONALE:` to avoid confusion with the `unsafe` convention.
9. `PredicateEventId(pub String)` inner field is `pub`, allowing in-place mutation — expose via `as_str()` and drop the `pub`.
10. Test fixture `PredicateEventId("11112222...")` is 62 chars, not 64 — inconsistent with `synth_event_id`'s stated 64-char output.

### Missing test coverage
- **Concurrent write correctness** — no multi-threaded test proving the core atomicity claim (N threads, verify count is exactly N)
- LRU eviction driven through the public API to `MAX_HISTORY_KEYS + 1`
- Value sum after clock advance past window
- Window boundary exact-at-cutoff (`front_ts == cutoff`, trim condition is `< cutoff`, so boundary entries are retained — worth an explicit test)
- `synth_event_id` output format pin (64-char hex)
- Cross-type event-id isolation (same id used in both `record_invocation` and `record_value`)

### Security notes
- **`caller_event_id` injection** — if a tenant can influence the event id, they can replay a prior id to suppress one unit of rate-limit consumption. The id must be host-assigned, not tenant-supplied. Clarify this in the trait doc and enforce it in durable backends.
- **LRU eviction as D5 attack** — high-cardinality key injection evicts legitimate tenant counters, silently resetting rate-limit state. The current threat model frames D5 as a memory concern, not a correctness-under-attack concern. Add the explicit variant.
- **Cross-process replay** — in-memory dedup does not protect against replay to a different instance. Acknowledge this in `PredicateStateBackend` security notes (not just in D5).

crates/ironclaw_hooks/docs/successors/03-persistent-counter.md
```diff
@@ -0,0 +1,157 @@
...
+  for tests + the standalone `ironclaw_hooks` integration tests that
+  don't want to spin up a database.
+- **`PostgresPredicateStateBackend`**: production. Schema:
+  ```sql
+  CREATE TABLE hook_invocation_events (
+      tenant_id     text NOT NULL,
+      hook_id       bytea NOT NULL,
+      capability    text NOT NULL,
+      occurred_at   timestamptz NOT NULL,
+      event_id      uuid PRIMARY KEY -- for replay dedup
```
serrrfirat: The schema dedupes `event_id` globally, but the backend contract dedupes only within a counter key. Because `caller_event_id` is per capability invocation, two predicate-backed hooks observing the same invocation would collide here and the second hook would not count its own bucket. Please scope uniqueness to `(tenant_id, hook_id, capability, event_id)` for invocations and `(tenant_id, hook_id, capability, field, event_id)` for values, then align the replay-refusal docs with that scope.

crates/ironclaw_hooks/src/points/capability.rs
```diff
@@ -40,6 +40,18 @@ pub struct BeforeCapabilityHookContext {
...
+    /// the calling layer (e.g. a digest of the runtime event id or the
+    /// capability-invocation surface) so that retried/replayed invocations
+    /// produce the same id — letting the backend's `event_id`
+    /// UNIQUE-constraint short-circuit duplicate counter writes.
+    ///
+    /// When `None`, the evaluator synthesizes a per-call-unique id and
+    /// dedup degrades to "every evaluation counts" semantics — appropriate
+    /// for the in-memory backend without replay, but **not** for durable
+    /// backends. Durable callers MUST supply a value.
+    pub caller_event_id: Option<crate::predicate_state::PredicateEventId>,
```
serrrfirat: This field is public, so callers can bypass the validation in `with_caller_event_id()` by assigning `Some(PredicateEventId::new(""))` or a NUL-containing value after construction. Since the evaluator trusts this field and `PredicateEventId::new()` is permissive, durable backends can still receive invalid IDs. Make the field private and force the validated setter, or validate construction at the `PredicateEventId` boundary with an internal unchecked constructor for synthesized IDs.

serrrfirat (changes requested):
Findings:

1. High - `crates/ironclaw_hooks/docs/successors/03-persistent-counter.md:91`, `:103`: the proposed durable schema uses `event_id uuid PRIMARY KEY`, but the backend contract dedupes only within the same counter key. `caller_event_id` is per capability invocation, not per predicate hook. If two predicate-backed hooks observe the same invocation, the second hook's insert conflicts on the global `event_id` and does not count for its own `(tenant, hook_id, capability[, field])` bucket. That undercounts independent hooks and can let a later hook's cap fail open. Use composite uniqueness scoped to the counter key, e.g. `(tenant_id, hook_id, capability, event_id)` and `(tenant_id, hook_id, capability, field, event_id)`, and align the `predicate_state.rs` replay-refusal docs with that scope.

2. Medium - `crates/ironclaw_hooks/src/points/capability.rs:54`, `:105-114`: `caller_event_id` is a public field, so the validation in `with_caller_event_id()` is bypassable. Callers can construct a context and then assign `Some(PredicateEventId::new(""))` or a NUL-containing value directly; `PredicateEventId::new()` is intentionally permissive, and the evaluator trusts the field. Durable backends can still receive invalid IDs or accidentally dedupe unrelated events under an empty ID. Make the field private and require the validated setter, or make `PredicateEventId` construction validated with an internal unchecked constructor for synthesized IDs.

I did not run tests; this was a review-only pass against head `7c83c3d14db6403fe3e549a32cf987a8c9b96aa3`.

crates/ironclaw_hooks/docs/successors/03-persistent-counter.md
zmanian: Addressed in `b4d8a355f`.

Schema PKs changed to composite uniqueness scoped to the counter key:
- `hook_invocation_events`: `PRIMARY KEY (tenant_id, hook_id, capability, event_id)`
- `hook_value_events`: `PRIMARY KEY (tenant_id, hook_id, capability, field, event_id)`

The trait's replay-refusal docs in `predicate_state.rs` were rewritten to spell out the per-key scope explicitly and to specify the corresponding `INSERT … ON CONFLICT (tenant, hook, capability[, field], event_id) DO NOTHING` shape that durable backends must use. Two predicate hooks observing the same invocation now record into distinct buckets — no first-writer-wins undercounting.

crates/ironclaw_hooks/src/points/capability.rs
zmanian: Addressed in `b4d8a355f`.

Moved validation INTO the type boundary: `PredicateEventId::new(...) -> Result<Self, PredicateEventIdError>` validates non-empty + NUL-free at construction. The public field on `BeforeCapabilityHookContext` can no longer carry an invalid id because constructing any `PredicateEventId` value goes through the validating path.

For the evaluator's internal synth path and tests that mint ids from known-good shapes (e.g. blake3 hex digests), `PredicateEventId::new_unchecked(...)` is the explicit escape hatch — its name makes the bypass visible to reviewers.

`with_caller_event_id` dropped its now-redundant runtime check (the type already enforces it). Existing call sites switched to `.expect("fixture passes validation")` for the test fixtures and `new_unchecked` for the synth path.

Tests:
- `predicate_event_id_rejects_empty`
- `predicate_event_id_rejects_nul_bytes`
- `predicate_event_id_accepts_typical_hex_digest`

crates/ironclaw_hooks/docs/successors/03-persistent-counter.md
```diff
@@ -0,0 +1,126 @@
...
+
+## Required behavior
+
+1. **Cross-process consistency**: two host processes against the same
+   tenant share counter state. Run-1 increments → run-2 sees the
+   updated count.
+2. **Restart survival**: counter state survives process restart.
+3. **Replay refusal**: re-emitting a recorded invocation timestamp
+   must NOT double-count. The backend stores `(timestamp, run_id,
+   event_id)` so duplicate-event detection works at replay time.
```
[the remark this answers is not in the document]
zmanian: Addressed in c9c5fb014 / b4d8a355 / 5b8b8b… across the doc revisions. `run_id` is intentionally NOT carried by the trait or the schema — see lines 72-78 of the doc:

> The sliding-window state is per-tenant + per-hook-id; replay refusal is driven by `event_id` (which is `RuntimeEventId` from `ironclaw_events`, itself already keyed to the current run's emission). An earlier draft mentioned storing `run_id` alongside; that's redundant given the event id's uniqueness contract and was removed in this revision.

The 'state' for clarity is: `run_id` is **not** stored; the design has shifted to a per-tenant + per-hook key with `event_id` as the dedup axis.

crates/ironclaw_hooks/docs/successors/03-persistent-counter.md
```diff
@@ -0,0 +1,126 @@
...
+    ) -> Result<(), PredicateBackendError>;
+
+    async fn sum_in_window(
+        &self,
+        key: &ValueKey,
+        window: Duration,
+    ) -> Result<Decimal, PredicateBackendError>;
+
+    async fn evict_older_than(&self, cutoff: SystemTime) -> Result<u64, PredicateBackendError>;
+}
```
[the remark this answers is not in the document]
zmanian: Addressed in doc revision. The current trait sketch (line 41-49) uses `DateTime<Utc>` for the durable backend, and the rationale is called out explicitly on lines 63-70:

> the in-memory backend that already shipped in PR #3635 uses `Instant` because it's process-local and monotonic. Durable backends serialize across processes, so they must use `chrono::DateTime<Utc>` to match the rest of the project (`src/db/mod.rs`, `ironclaw_events`). The trait will accept `DateTime<Utc>` and the existing in-memory backend will gain a thin shim mapping `Instant`-driven callers to a fixed reference point.

crates/ironclaw_hooks/docs/successors/03-persistent-counter.md
```diff
@@ -0,0 +1,126 @@
...
+  CREATE INDEX ix_hook_invocation_events_window ON hook_invocation_events
+      (tenant_id, hook_id, capability, occurred_at DESC);
+
+  CREATE TABLE hook_value_events (
+      tenant_id     text NOT NULL,
+      hook_id       bytea NOT NULL,
+      capability    text NOT NULL,
+      field         text NOT NULL,
+      occurred_at   timestamptz NOT NULL,
+      value         numeric NOT NULL,
```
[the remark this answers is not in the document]
zmanian: Addressed in doc revision (lines 120-128). The libSQL impl is now explicitly called out as having two material differences from the Postgres impl mandated by `src/db/CLAUDE.md`:

- `value` column is `TEXT NOT NULL` (not numeric), because libSQL's integer/real types can't preserve `rust_decimal` precision; the backend serializes via `Decimal::to_string()` / `from_str()` at the row boundary.
- `occurred_at` is stored as ISO-8601 `TEXT` (libSQL convention).

crates/ironclaw_hooks/docs/successors/03-persistent-counter.md
```diff
@@ -0,0 +1,126 @@
...
+   reads (already pinned for in-memory by `HistoryKey` — confirm
+   backend-side).
+
+## Risk
+
+- Touches `src/db/` migrations machinery. See `.claude/rules/database.md`
+  for the dual-backend conventions.
+- Performance: every predicate evaluation becomes a DB round-trip.
+  Plan: batched writes via the dispatcher's tick boundary, with
+  reads cached for the current dispatch.
```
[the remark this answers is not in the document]
zmanian: Addressed in doc revision (lines 149-163). You correctly flagged that batched writes conflict with the cross-process consistency requirement; the doc now resolves it explicitly:

> Performance: every predicate evaluation becomes a DB round-trip. An earlier draft suggested batching writes at the dispatcher's tick boundary. **That conflicts with requirement #1 (cross-process consistency)**: a deferred write from host A wouldn't be visible to host B's read until the next tick, so two concurrent hosts could each see 'under cap' simultaneously and both proceed past `max` (gemini's review on the prior draft). Resolution: the v1 production backend keeps writes synchronous (read-your-own-writes within the call); the in-process cache stays per-dispatch-only. A future optimization could batch *reads* — collapse N predicate evaluations in one dispatch into a single batch SELECT — but never writes.

zmanian:
@henrypark133 thanks for the careful review — addressing your HIGH finding plus the joint blockers from Firat's pass.

## B1 (HIGH) — Replay-dedup gap through `PredicateBackedBeforeCapabilityHook` — `evaluator.rs:145, :236`

**Fixed across two commits.**

Threading shipped earlier in `e6df47dd` (`caller_event_id` field on `BeforeCapabilityHookContext`, evaluator now prefers it via `resolve_event_id` and falls back to the synth path only when absent). The piece your review specifically called out — *"caller-level tests proving duplicate event IDs are no-ops"* through the installed-hook path — is now `f632d225a`:

- New `caller_event_id_replay_is_deduped_through_wrapper_hook` test in `installed_hook.rs` drives `PredicateBackedBeforeCapabilityHook::evaluate` directly (not just `PredicateEvaluator::evaluate_at`), per the repo CLAUDE.md "Test through the caller, not just the helper" rule. The test:
  1. Two evaluations with the same stable `caller_event_id` against a `RateOrValueCap { max: 1 }` predicate — second call stays under cap (dedup engages at the wrapper boundary).
  2. A third evaluation with a *distinct* `caller_event_id` crosses the cap — proving dedup is replay-scoped (same id → no-op), not blanket-suppress (any id → no-op).

If the wrapper were synthesizing a fresh id per call (the bug), the test would fail at step 2 with the second evaluation being denied.

## B2 (HIGH) — O(n) dedup scan under mutex — `predicate_state.rs`

**Fixed in `0ea35d8a4`** (must-fix #1). Added companion `HashSet<PredicateEventId>` per bucket (`InvocationBucket::dedup_ids` / `ValueBucket::dedup_ids`). Membership check is now O(1); the `VecDeque<entries>` is kept for ordered iteration and FIFO trim. Invariants documented at the bucket struct: `dedup_ids` is exactly the set of `event_id` values currently in `entries`; every push/pop updates both via `push_back` / `pop_front` helpers that keep them in sync.

## B3 (HIGH) — Mutex poison cascade — `predicate_state.rs`

**Fixed in `0ea35d8a4`** (must-fix #2). Replaced `.expect()` on `.lock()` with `match { Ok(g) => g, Err(poisoned) => poisoned.into_inner() }`. Choice documented inline (record_invocation:319-322):

> Recover from a poisoned mutex by reading the inner value rather than cascading the panic to every subsequent caller. A poisoning thread has already aborted; refusing service indefinitely is worse than proceeding with possibly-incomplete state.

Backend has no other failure modes today (`Unavailable` is reserved for the future durable backends), so propagation would force every caller to handle an error the in-memory impl never returns. Documented as "best-effort recovery" semantics.

## B4 — `caller_event_id` trust-boundary doc — `predicate_state.rs`

**Fixed in `c9c5fb014`**. Added explicit rustdoc on both `PredicateEventId` and the `PredicateStateBackend` trait clarifying the host-assigned invariant:

- The id MUST be minted by trusted host code from authoritative sources (dispatcher `RuntimeEventId`, host-side hash, arguments digest).
- It MUST NOT pass through unchanged from tenant-controlled surfaces (capability args, manifest fields, WASM memory, HTTP bodies).
- The format invariants in `PredicateEventId::new` (non-empty, NUL-free) are a durability contract for SQL backends, NOT a trust check.
- A tenant-supplied id can undercount itself into infinity by replaying a fixed id, or poison adjacent buckets if scoping is ever weakened.

---

Quality gate green on `ironclaw_hooks`: 168 tests pass, `cargo clippy --all --benches --tests --examples --all-features -- -D warnings` is clean. Re-requesting review.

zmanian:
## Should-fix + nits follow-up — henrypark133 review (5-15)

Status on the remaining items from your should-fix + nits list. All addressed; SHAs below.

### Should-fix

| # | Item | Status |
|---|---|---|
| **S2** | `evict_older_than` trait stub with default impl | Already landed in HEAD — `predicate_state.rs:252` defines the default-impl no-op on `PredicateStateBackend` (commit `b4d8a355f`). Trait-object callers won't break when the durable impl lands. |
| **S3** | Document sync surface as intentional | Already landed in HEAD — module-level docs `predicate_state.rs:46-53` explicitly mark the trait as synchronous in this PR with the durable-backend async migration noted (commit `b4d8a355f`). |
| **S7** | Cross-process consistency vs batched-writes tradeoff | Already landed in HEAD — `docs/successors/03-persistent-counter.md` lines 149-163 spell out the conflict between cross-process consistency and write-batching, and state the v1 resolution (synchronous writes; reads may batch in future) explicitly (commit `df505aea2`). |
| **S8** | LRU eviction as D5 correctness-under-attack variant + cross-process replay in `PredicateStateBackend` notes | **`2b0383282`** — Threat-model: added **D5a** as the correctness-under-attack variant of D5 (attacker-driven high-cardinality key flood evicts legitimate tenants' counters, resetting rate-limit state). PredicateStateBackend: added explicit `# Cross-process replay limits (in-memory backend)` section documenting that the in-memory dedup is process-local; durable backend is the only defense across hosts. |
| **S9** | Missing tests (4) | Partially landed before, completed in **`2b0383282`**. Concurrent N-thread atomicity already at `predicate_state.rs:840` (`in_memory_record_invocation_is_atomic_under_concurrent_writers`). New in this commit: `lru_eviction_via_public_api_holds_max_history_keys_cap` (drives `MAX_HISTORY_KEYS + 1` through the public API, asserts cap holds + evictions counter advances), `in_memory_invocation_retains_entry_at_exact_window_cutoff` (pins `< cutoff` trim semantics), `event_id_dedup_is_isolated_across_invocation_and_value_maps` (cross-type event-id isolation). Caller-boundary replay-dedup test through the wrapper hook is at `installed_hook.rs:154` (`f632d225a`). |

### Nits

| # | Item | Status |
|---|---|---|
| **N1** | Move `synth_event_id` from `evaluator.rs` → `predicate_state.rs` | **`33521f7e2`** — Moved to `PredicateEventId::synth(...)` next to the backend that consumes the id format. Takes raw bytes / `&str` (not `&BeforeCapabilityHookContext`) so `predicate_state` stays a leaf module below `points` — the evaluator's `resolve_event_id` unpacks the context and delegates. |
| **N2** | Replace `// safety:` on non-`unsafe` blocks with `// RATIONALE:` | **`33521f7e2`** — The `write!(s, "{byte:02x}")` infallibility note was the only `// safety:` outside an `unsafe` block; renamed in the synth move. |
| **N3** | `PredicateEventId(pub String)` → drop inner `pub`, expose via `as_str()` | Already landed in HEAD — `predicate_state.rs:115` is `pub struct PredicateEventId(String)` (no inner `pub`), and `as_str()` is the read accessor (commit `b4d8a355f`). |
| **N4** | Test fixture id is 62 chars, not 64 — align with `synth_event_id` | Already landed in HEAD — current fixtures (e.g. `evaluator.rs:486`, `:506`, `:516`, `installed_hook.rs:171`, `:209`) all use 64-char hex strings matching the synth output (commit `b4d8a355f`). |
| **N5** | Pin `synth_event_id` 64-char hex output in a test | **`33521f7e2`** — `synth_event_id_is_64_char_lowercase_hex` asserts the synth output is exactly 64 chars and lowercase-hex. Failure mode: a refactor that silently changes the length or case would break the durable backend's `uuid` UNIQUE constraint; the new test fails loud.

### Quality gate

Each commit: `cargo fmt && cargo clippy --all --benches --tests --examples --all-features -- -D warnings && cargo test -p ironclaw_hooks` — all green. Final HEAD `33521f7e2` runs 172 hooks lib tests passing including all new pins.

Branch: `hooks-fu-persistent-counter` @ `33521f7e2`.

crates/ironclaw_hooks/src/predicate_state.rs
```diff
@@ -0,0 +1,1103 @@
...
+    /// impl lands — henrypark133 important #5 on PR #3635.
+    #[allow(dead_code)] // operator reaper hook for future durable backends
+    fn evict_older_than(&self, _cutoff: Instant) -> Result<u64, PredicateBackendError> {
+        Ok(0)
+    }
+}
+
+/// Per-key history bucket. Maintains `entries` (the FIFO window) AND a
+/// companion `dedup_ids` set so duplicate-id checks are O(1) instead of
+/// O(n) — the previous linear scan serialized every evaluator call under
```
henrypark133, line 357: **High** — O(n) sum recomputation on every record_value call in hot path.

record_value iterates all in-window entries and sums their Decimal values on every call: `bucket.entries.iter().map(|(_, v, _)| *v).sum()`. Under high-throughput value-cap predicates this is O(n) per call where n grows with the window. The invocation path already uses an O(1) HashSet for dedup; the value path should maintain a running sum field on ValueBucket, incrementing on push_back and decrementing on pop_front, to make the read O(1).

**Fix:** Add `running_sum: Decimal` to ValueBucket; update it in push_back/pop_front; read it directly instead of iterating entries.

crates/ironclaw_hooks/src/predicate_state.rs
```diff
@@ -0,0 +1,1103 @@
...
+        )
+    }
+
+    fn tenant() -> TenantId {
+        TenantId::new("alpha").expect("ok")
+    }
+
+    fn ev(s: &str) -> PredicateEventId {
+        // Test fixture ids are well-formed; use unchecked.
+        PredicateEventId::new_unchecked(s)
```
henrypark133, line 791: **High** — PredicateEventId::new_unchecked is pub, enabling validation bypass.

The new_unchecked constructor is public, allowing any crate caller to create a PredicateEventId with empty or NUL-containing strings. While the docs say 'reserved for internal synth paths and tests', the pub visibility means any code in the crate can bypass the format validation that durable SQL backends depend on for UNIQUE constraint integrity.

**Fix:** Restrict new_unchecked to pub(crate) and document that only predicate_state module internals and tests may call it; or use a separate test-only feature gate.

crates/ironclaw_hooks/src/predicate_state.rs
```diff
@@ -0,0 +1,1103 @@
...
+//! The [`PredicateEvaluator`] in [`crate::evaluator`] delegates its
+//! counter / value-sum bookkeeping to a [`PredicateStateBackend`]. The
+//! default backend ([`InMemoryPredicateStateBackend`]) preserves the
+//! existing in-process semantics; future durable backends
+//! (Postgres-backed, libSQL-backed) implement the same trait so the
+//! evaluator's predicate logic stays backend-agnostic.
+//!
+//! # Visibility
+//!
+//! The backend trait and the in-memory impl are intentionally
```
henrypark133, line 12: **High** — In-memory backend dedup is process-local, enabling multi-host rate-limit bypass.

The InMemoryPredicateStateBackend's replay dedup only works within a single process. In a multi-host deployment, the same event_id replayed on a different host counts as a fresh invocation. An attacker can exploit this by replaying requests across hosts to bypass rate caps — each host maintains its own counter. The code documents this limitation but ships it as the default backend without any cross-process coordination.

**Fix:** Document this as a known security limitation in operator runbooks; ensure the durable backend (with SQL UNIQUE constraint) is the recommended production default; consider adding a startup warning when InMemoryPredicateStateBackend is used in multi-host deployments.

crates/ironclaw_hooks/src/predicate_state.rs
```diff
@@ -0,0 +1,1103 @@
...
+/// An `event_id` recorded on host A is unknown to host B: replaying the
+/// same `event_id` against host B will count as a fresh invocation,
+/// double-counting against the logical cap. The in-memory backend
+/// therefore does NOT defend against multi-host replay; that property
+/// requires the durable backend (see successor doc
+/// `03-persistent-counter.md`), where the SQL `UNIQUE` constraint on
+/// `(tenant_id, hook_id, capability[, field], event_id)` enforces dedup
+/// across every host pointing at the same database
+/// (henrypark133 security note on PR #3635). Threat-model finding D5a
+/// also captures the high-cardinality-key LRU-eviction variant that lets
```
henrypark133, line 309: **Medium** — LRU eviction at MAX_HISTORY_KEYS enables cross-tenant rate-limit reset attack.

The bounded key map (8192 keys) uses LRU eviction. An attacker can flood the dispatcher with high-cardinality (tenant_id, hook_id, capability) combinations to evict legitimate tenants' counter buckets. When a legitimate tenant's bucket is evicted, their rate-limit state silently resets, allowing them to re-consume their budget. This is documented as D5a in the threat model but the mitigation is only an operator alert counter — no enforcement boundary prevents the attack.

**Fix:** Implement per-tenant key quotas within the LRU map so no single tenant can be fully evicted by another tenant's fan-out; or use the durable backend with tenant-partitioned tables to eliminate the shared LRU surface.

Also flagged by: performance/Medium

crates/ironclaw_hooks/src/predicate_state.rs
```diff
@@ -0,0 +1,1103 @@
...
+/// MUST NOT treat it as adversarial input from the tenant. The host is
+/// responsible for ensuring no tenant-controlled bytes flow into
+/// `event_id` unfiltered — see [`PredicateEventId`] for the threat
+/// description. The format invariants enforced by
+/// [`PredicateEventId::new`] (non-empty, NUL-free) are a durability
+/// contract for SQL backends, NOT a trust check.
+///
+/// # Replay refusal contract
+///
+/// Dedup is scoped to the counter `key`, NOT to `event_id` globally.
```
henrypark133, line 285: **Medium** — Global AtomicU64 counter in synth causes cache-line contention under concurrency.

PredicateEventId::synth uses a process-global `static COUNTER: AtomicU64` incremented on every call. When multiple threads across different hooks/tenants all fall through to the synth path (no caller_event_id supplied), this single atomic becomes a cache-line contention hotspot. Under high concurrency this serializes the synth path. Mitigate with thread-local counters or a per-backend sequence generator.

**Fix:** Replace the global static AtomicU64 with a per-backend-instance AtomicU64 field, or use a thread-local counter combined with a thread-id prefix to avoid cross-thread contention.

crates/ironclaw_hooks/src/predicate_state.rs
```diff
@@ -0,0 +1,1103 @@
...
+    /// is a no-op — the in-memory backend already trims per-key on every
+    /// `record_*` call, so a separate reaper would be redundant. Durable
+    /// backends override this with a `DELETE WHERE occurred_at < cutoff`
+    /// equivalent, run periodically by an operator reaper task.
+    ///
+    /// Lockable into the trait signature now (rather than at the durable
+    /// backend PR) so trait-object callers don't break when the durable
+    /// impl lands — henrypark133 important #5 on PR #3635.
+    #[allow(dead_code)] // operator reaper hook for future durable backends
+    fn evict_older_than(&self, _cutoff: Instant) -> Result<u64, PredicateBackendError> {
```
henrypark133, line 350: **Medium** — evict_older_than is a no-op — inactive keys never garbage-collected without record_* calls.

The default evict_older_than implementation returns Ok(0) without doing any work. The in-memory backend only trims entries when record_invocation/record_value is called for that specific key. Keys that go idle (no new invocations) retain their full VecDeque + HashSet in memory indefinitely, bounded only by MAX_HISTORY_KEYS. Under a D5a attack where an attacker creates many keys then stops using them, those zombie buckets persist until a new key forces LRU eviction. An operator reaper task calling evict_older_than periodically is documented but not wired up.

**Fix:** Wire up a periodic background task (e.g. on the dispatcher's per-build lifecycle) that calls evict_older_than with a cutoff, and implement it in InMemoryPredicateStateBackend to scan and remove expired keys.

crates/ironclaw_hooks/src/points/capability.rs
henrypark133: **Medium** — caller_event_id public field allows direct assignment bypassing setter.

The caller_event_id field on BeforeCapabilityHookContext is pub. Combined with pub new_unchecked on PredicateEventId, a caller can directly assign an invalid id (empty/NUL) by constructing ctx.caller_event_id = Some(PredicateEventId::new_unchecked("")), bypassing the with_caller_event_id setter. The code comments claim the type boundary protects against this, but new_unchecked defeats that protection.

**Fix:** Either make caller_event_id private (pub(crate)) or make PredicateEventId::new_unchecked pub(crate) so external callers cannot construct invalid ids to assign directly.

crates/ironclaw_hooks/src/predicate_state.rs
```diff
@@ -0,0 +1,1103 @@
...
+    }
+
+    /// henrypark133 nit #5 on PR #3635: pin the synth output shape. The
+    /// 64-char lowercase-hex format is part of the durable backend's
+    /// expectation (e.g. Postgres `uuid` UNIQUE constraint accepts hex)
+    /// — a refactor that silently changes the length or character set
+    /// would break that contract without surfacing a test failure.
+    #[test]
+    fn synth_event_id_is_64_char_lowercase_hex() {
+        let id = PredicateEventId::synth(b"hookid-bytes", "cap.x", &[0xAB, 0xCD]);
```
henrypark133: **Low** — PredicateEventId::synth includes arguments_digest in id, potential oracle.

The synth function includes arguments_digest in the blake3 hash input for the event id. While the digest itself is already a hash, the resulting 64-char hex event id is returned to callers and could be used as an oracle to verify whether two invocations had identical arguments (same synth output = same args digest). In contexts where argument confidentiality matters, this leaks equality information.

**Fix:** If argument equality leakage is a concern, use a host-side random nonce instead of arguments_digest in the synth input, or document the equality-leakage property explicitly.

henrypark133 (changes requested):
## Code Review (multi-agent)

**Intent:** Add PredicateStateBackend trait with Postgres/libSQL impls for persistent sliding-window counter surviving restarts and cross-process consistency.

**Stats:** 9 findings (from 10 raw, 1 after dedup) across 3 files. Reviewers run: security, bugs, performance, tests, conventions, design. Reviewers failed: none. Body-only: 1

### Security

1. **High** PredicateEventId::new_unchecked is pub, enabling validation bypass (crates/ironclaw_hooks/src/predicate_state.rs:119-127, confidence 75) — anchor: crates/ironclaw_hooks/src/predicate_state.rs:119
   The new_unchecked constructor is public, allowing any crate caller to create a PredicateEventId with empty or NUL-containing strings. While the docs say 'reserved for internal synth paths and tests', the pub visibility means any code in the crate can bypass the format validation that durable SQL backends depend on for UNIQUE constraint integrity.

2. **Medium** caller_event_id public field allows direct assignment bypassing setter (crates/ironclaw_hooks/src/points/capability.rs:52, confidence 50) — anchor: crates/ironclaw_hooks/src/points/capability.rs:52
   The caller_event_id field on BeforeCapabilityHookContext is pub. Combined with pub new_unchecked on PredicateEventId, a caller can directly assign an invalid id (empty/NUL) by constructing ctx.caller_event_id = Some(PredicateEventId::new_unchecked("")), bypassing the with_caller_event_id setter. The code comments claim the type boundary protects against this, but new_unchecked defeats that protection.

3. **High** In-memory backend dedup is process-local, enabling multi-host rate-limit bypass (crates/ironclaw_hooks/src/predicate_state.rs:226-235, confidence 75) — anchor: crates/ironclaw_hooks/src/predicate_state.rs:226
   The InMemoryPredicateStateBackend's replay dedup only works within a single process. In a multi-host deployment, the same event_id replayed on a different host counts as a fresh invocation. An attacker can exploit this by replaying requests across hosts to bypass rate caps — each host maintains its own counter. The code documents this limitation but ships it as the default backend without any cross-process coordination.

4. **Medium** LRU eviction at MAX_HISTORY_KEYS enables cross-tenant rate-limit reset attack (crates/ironclaw_hooks/src/predicate_state.rs:579-582, confidence 75) — anchor: crates/ironclaw_hooks/src/predicate_state.rs:579
   The bounded key map (8192 keys) uses LRU eviction. An attacker can flood the dispatcher with high-cardinality (tenant_id, hook_id, capability) combinations to evict legitimate tenants' counter buckets. When a legitimate tenant's bucket is evicted, their rate-limit state silently resets, allowing them to re-consume their budget. This is documented as D5a in the threat model but the mitigation is only an operator alert counter — no enforcement boundary prevents the attack.

5. **Low** PredicateEventId::synth includes arguments_digest in id, potential oracle (crates/ironclaw_hooks/src/predicate_state.rs:155-180, confidence 50) — anchor: crates/ironclaw_hooks/src/predicate_state.rs:155
   The synth function includes arguments_digest in the blake3 hash input for the event id. While the digest itself is already a hash, the resulting 64-char hex event id is returned to callers and could be used as an oracle to verify whether two invocations had identical arguments (same synth output = same args digest). In contexts where argument confidentiality matters, this leaks equality information.

### Performance

1. **High** O(n) sum recomputation on every record_value call in hot path (crates/ironclaw_hooks/src/predicate_state.rs:484, confidence 90) — anchor: crates/ironclaw_hooks/src/predicate_state.rs:484
   record_value iterates all in-window entries and sums their Decimal values on every call: `bucket.entries.iter().map(|(_, v, _)| *v).sum()`. Under high-throughput value-cap predicates this is O(n) per call where n grows with the window. The invocation path already uses an O(1) HashSet for dedup; the value path should maintain a running sum field on ValueBucket, incrementing on push_back and decrementing on pop_front, to make the read O(1).

2. **Medium** Global AtomicU64 counter in synth causes cache-line contention under concurrency (crates/ironclaw_hooks/src/predicate_state.rs:189-190, confidence 75) — anchor: crates/ironclaw_hooks/src/predicate_state.rs:189
   PredicateEventId::synth uses a process-global `static COUNTER: AtomicU64` incremented on every call. When multiple threads across different hooks/tenants all fall through to the synth path (no caller_event_id supplied), this single atomic becomes a cache-line contention hotspot. Under high concurrency this serializes the synth path. Mitigate with thread-local counters or a per-backend sequence generator.

3. **Medium** evict_older_than is a no-op — inactive keys never garbage-collected without record_* calls (crates/ironclaw_hooks/src/predicate_state.rs:304-316, confidence 65) — anchor: crates/ironclaw_hooks/src/predicate_state.rs:304
   The default evict_older_than implementation returns Ok(0) without doing any work. The in-memory backend only trims entries when record_invocation/record_value is called for that specific key. Keys that go idle (no new invocations) retain their full VecDeque + HashSet in memory indefinitely, bounded only by MAX_HISTORY_KEYS. Under a D5a attack where an attacker creates many keys then stops using them, those zombie buckets persist until a new key forces LRU eviction. An operator reaper task calling evict_older_than periodically is documented but not wired up.

### Conventions

1. **Medium** PR states 'scope doc only, no implementation yet' but diff contains substantial implementation code (PR title/body metadata:1, confidence 75) — anchor: PR metadata: title and body claim docs-only scope; commit messages (d5f68de30 through 7ff90ceb2) confirm implementation work
   The PR title says 'scope doc only, no implementation yet' and the body says 'Draft — scope doc only, no implementation yet.' However, the diff contains: a new 1103-line `predicate_state.rs` module (trait + in-memory impl), refactored `evaluator.rs` delegating to the backend trait, a new `caller_event_id` field on `BeforeCapabilityHookContext`, a `pub mod predicate_state` in `lib.rs`, and new tests in `installed_hook.rs`. Searched: PR body (claims docs-only) / commit msgs (13 commits including feat/fix/refactor/test types, none describing this as docs-only) / inline comments (no rationale for the metadata mismatch) / linked issues (references #3573 as predecessor, #3635 as tracking issue — neither justifies the docs-only claim). The 'scope: docs' label and 'Draft for design review' status are the only signals, but they contradict the implementation-heavy diff. Reviewers routing this as a docs-only review may miss implementation-level convention issues.

zmanian:
Consolidated reply to the 5-19 multi-agent review batch. Pushed three focused commits (94a87de8d, e8399f6fa, e5c588b19).

### A. Security

- **A1 HIGH** — `PredicateEventId::new_unchecked` is `pub` (bypasses durable UNIQUE-constraint format invariants): **fixed**. Narrowed to `pub(crate)` in e5c588b19. All in-tree call sites are inside the crate; the validated `::new` is the only external constructor.
- **A2 MED** — `caller_event_id` field is `pub` (lets callers bypass `with_caller_event_id`): **fixed**. Narrowed field to `pub(crate)` in 94a87de8d. External callers must go through the typed setter, which itself takes a `PredicateEventId` validated at `::new`.
- **A3 HIGH** — in-memory dedup is process-local, multi-host replay bypasses rate limits: **mitigated via startup warn + docs**. Added `PredicateEvaluator::warn_in_memory_backend_active_in_production` in e8399f6fa; hosts running multi-host or multi-tenant production should call it at startup. Fully closing the gap requires the durable backend (tracked in the successor doc).
- **A4 MED** — LRU at MAX_HISTORY_KEYS = 8192 lets cross-tenant counter reset: **fixed**. Added `MAX_KEYS_PER_TENANT = MAX_HISTORY_KEYS / 4` in e5c588b19. A noisy tenant overflowing now evicts its own oldest-front bucket first; tenant β's bucket survives tenant α's flood. New tests `per_tenant_quota_holds_single_tenant_at_its_cap` and `per_tenant_quota_isolates_tenants_from_each_other` pin the behavior.
- **A5 LOW** — `synth` mixes `arguments_digest` into the hash, making the 64-char hex an equality oracle for argument shape: **fixed**. Dropped `arguments_digest` from the synth hash in e8399f6fa / e5c588b19. Synth only needs to be per-call unique (replay dedup uses caller-supplied `caller_event_id`, not synth output). Added thread-local nonce to keep synth divergent under bit-identical inputs. New test `synth_diverges_on_identical_inputs`.

### B / C. chatgpt-codex P1s

- **B (recent_ids cap breaks replay refusal)** — already addressed earlier in this PR (see the bucket invariants doc on `InvocationBucket`/`ValueBucket`: dedup memory is tied to the in-window entry set, not a fixed ring). `dedup_memory_covers_full_window_under_high_throughput` exercises the >256-event case.
- **C (MAX_HISTORY_KEYS bypass via empty buckets)** — already addressed: record paths drop empty buckets eagerly, and `evict_lru_invocation` / `evict_lru_value` prefer empty buckets as the first victim as defense-in-depth. `lru_evicts_empty_buckets_first` pins it.

### D. Performance

- **D6 HIGH** — `record_value` O(n) sum: **fixed**. `ValueBucket::running_sum` maintained incrementally on push/pop; commit e5c588b19. New test `numeric_sum_running_sum_matches_deque_after_trim_and_replay`.
- **D7 MED** — process-global AtomicU64 contention in `synth`: **mitigated + documented**. Added thread-local nonce in the hash input so threads can advance without forcing cross-core cache-line invalidation in the common path; the AtomicU64 stays for cross-thread uniqueness backup. Documented as a known hotspot for the synth fallback path (durable production traffic should supply `caller_event_id` and never reach synth).
- **D8 MED** — `evict_older_than` default is no-op: **fixed**. Implemented for the in-memory backend in e5c588b19; drops entries strictly older than the cutoff and removes buckets that become empty. New test `evict_older_than_drops_expired_entries_and_empty_buckets`.

### E. serrrfirat HIGH (per-counter-key dedup scope)

Already reflected in `03-persistent-counter.md` and in the trait's replay-refusal contract: dedup is scoped to `(tenant_id, hook_id, capability[, field], event_id)`, NOT global on `event_id`. The Postgres / libSQL schemas in the design doc carry the matching PRIMARY KEY scope. The in-memory backend's per-key `dedup_ids` set already implements this scope — two predicate-backed hooks observing the same invocation each maintain their own bucket's dedup state. Will carry through to the durable impl PR.

### F. gemini doc reconciliations on `03-persistent-counter.md`

Already addressed in the current revision of the doc:
- `run_id` removed from the trait surface (replay refusal is driven by `event_id` which is already keyed to the run via `RuntimeEventId`).
- Trait surface migrated to `chrono::DateTime<Utc>` to match the project DB convention; the in-memory backend will gain a thin Instant adapter when the durable PR lands.
- libSQL `value` column documented as `TEXT NOT NULL` (vs Postgres `numeric`) with `Decimal::to_string()` / `from_str()` at the row boundary; `occurred_at` documented as ISO-8601 TEXT under libSQL.
- "Cross-process consistency vs batched writes" resolved: v1 writes stay synchronous to preserve read-your-own-writes; only future read batching is open.

### G. PR body

Updated to reflect the shipped trait + in-memory extraction + `caller_event_id` plumbing + per-tenant quota + `evict_older_than` impl; durable backends remain a follow-up.

### Verification
- `cargo test -p ironclaw_hooks --lib`: 196 passed (+4 new vs baseline).
- `cargo test -p ironclaw_reborn --test hooks_integration`: 19 passed.
- `cargo clippy -p ironclaw_hooks -p ironclaw_reborn --all-targets --all-features -- -D warnings`: clean.
- `cargo fmt --all`: clean.

zmanian:
@henrypark133 friendly ping — all 7 findings from your 5-19 review were addressed in commits 94a87de8d, e8399f6fa, e5c588b19 (consolidated reply on 5-20 maps each finding A1–G to its fix + regression test). A3 (multi-host replay bypass) is explicitly deferred to the durable Postgres/libSQL backend PR with a startup warn + design-doc mitigation noted. CI is green. Could you take another pass when you have a moment? Thanks!

henrypark133 (approved):
**Single-agent review.**

**Approving with follow-up:** these Medium/Low items should be addressed in a follow-up PR (not blocking).

---

## Verdict: APPROVE (0 High, 2 Medium, 3 Low)

This PR is a well-executed extraction of predicate sliding-window state into a pluggable `PredicateStateBackend` trait. The critical properties from prior review rounds — atomic record-and-read, O(1) dedup via companion `HashSet`, per-tenant quota to prevent cross-tenant LRU eviction, `running_sum` invariant on `ValueBucket`, `new_unchecked` restricted to `pub(crate)`, `caller_event_id` field restricted to `pub(crate)` with builder setter — are all correctly implemented and well-tested. The concurrent-writers test and the full-window dedup regression close the previously-reported gaps. No Critical or High findings found.

---

## Medium findings (non-blocking)

### M1 — Schema doc: `event_id uuid NOT NULL` will reject synth-path IDs

`crates/ironclaw_hooks/docs/successors/03-persistent-counter.md` lines 91 / 112

The successor schema declares `event_id uuid NOT NULL` for both `hook_invocation_events` and `hook_value_events`. But `PredicateEventId` is not guaranteed to be a UUID:

- Caller-supplied IDs (from `RuntimeEventId`, which wraps `Uuid`) are genuine UUIDs — OK for a `uuid` column.
- Synth-path IDs (`PredicateEventId::synth`) are 64-char lowercase blake3 hex digests (256-bit output). PostgreSQL's `uuid` type stores 128 bits; it will reject a 64-char hex string at INSERT time with a type error.

The inline comment in the schema already says "opaque hex; durable backends UNIQUE-constraint on it" which correctly treats the value as opaque text. Fix: change the column type to `TEXT NOT NULL` (with an optional `CHECK (length(event_id) <= 64)` constraint). This should be corrected in the schema doc before the durable backend PR begins, to avoid the implementer inheriting the wrong type.

### M2 — O(N) tenant-count scan under global mutex on every new-key insert

`crates/ironclaw_hooks/src/predicate_state.rs` lines 493–501 (and symmetric at line 544)

For every `record_invocation` call where the key is new, the code does:
1. `history.keys().filter(|k| k.tenant_id == ...).count()` — full O(N) scan over up to 8192 entries under the global `Mutex<HashMap>`.
2. If the tenant is at quota, `evict_lru_invocation_for_tenant` — a second O(N) pass over the tenant's subset.

At 8192 keys with many tenants inserting new keys concurrently, this holds the global lock across two full iterations per insertion, serializing all writers during the quota check. This is not exploitable for security escalation (max blast radius is in-memory backend latency), and the durable backend will not have this issue, so it is Medium not High. Suggested mitigation: track `per_tenant_count: HashMap<TenantId, usize>` alongside the main map, updated on insert and evict, reducing the quota check to O(1) under the lock.

---

## Low findings (non-blocking)

### L1 — No concurrent-writers test for `record_value` / `running_sum`

`crates/ironclaw_hooks/src/predicate_state.rs` — `in_memory_record_invocation_is_atomic_under_concurrent_writers` at line 1438

`record_invocation` has an N-threaded concurrent-writers atomicity test. `record_value` / `ValueBucket::running_sum` has no symmetric test. If a future refactor ever splits the `value_history` mutex or moves `running_sum` updates outside the lock, there is no test to catch the regression. A parallel test (N threads each recording a distinct `event_id` with value=1, final `running_sum == N`) would pin the contract.

### L2 — `lru_evicts_empty_buckets_first` bypasses `InvocationBucket::push_back`, violating bucket invariant in test

`crates/ironclaw_hooks/src/predicate_state.rs` line 1103

The test calls `live.entries.push_back(...)` directly instead of `live.push_back(ts, event_id)`. This leaves `dedup_ids` empty while `entries` has one entry — exactly the state the documented invariant says must not exist. The test correctly verifies the LRU function, but uses an internally inconsistent bucket to do it. Prefer constructing the bucket via a `record_invocation` call on the backend, or at least using `InvocationBucket::push_back` to keep the test state within the invariant bounds.

### L3 — `NumericSum` `max` not validated as parseable positive `Decimal` at manifest install time

`crates/ironclaw_hooks/src/evaluator.rs` line 196; `crates/ironclaw_hooks/src/manifest.rs`

`HookManifestEntry::validate()` validates the `window` string at install time (surfacing bad windows early) but does not validate that `NumericSum`'s `max` is a parseable, non-negative `Decimal`. A manifest author who writes `max = "abc"` or `max = "-100"` will only learn of the error at evaluation time via a silent fail-closed deny. Adding a `Decimal::from_str(max).ok().filter(|d| d.is_sign_positive())` check in `manifest.rs` alongside `validate_window` would surface authoring mistakes at install time — consistent with the existing pattern.

---

## What this PR gets right

- **Atomic record-and-read** under a single mutex — the Critical concern from prior reviews is correctly implemented for both `record_invocation` and `record_value`.
- **O(1) dedup** via companion `HashSet<PredicateEventId>` — the previous O(n) linear scan is gone.
- **`running_sum` invariant** on `ValueBucket` maintained via `push_back`/`pop_front` — O(1) sum reads.
- **Per-tenant quota** (`MAX_KEYS_PER_TENANT = MAX_HISTORY_KEYS / 4`) correctly evicts within the offending tenant's footprint first, preventing cross-tenant LRU displacement.
- **`new_unchecked` restricted to `pub(crate)`** — external callers are forced through `PredicateEventId::new` with format validation.
- **`caller_event_id` field restricted to `pub(crate)`** with a `with_caller_event_id` builder setter — closes the direct-field-assignment bypass noted in the prior review.
- **Threat model D5a** correctly documented and threat severity assessed at Med (in-memory) → Low (durable) with a clear migration path.
- **`evict_older_than`** implemented (not a no-op) for the in-memory backend, with tests.
- **Comprehensive test coverage** — all the identified regression scenarios from prior review rounds have corresponding tests.
- **No `.unwrap()`/`.expect()` in production code**, mutex poison recovery follows the project convention.

zmanian:
Round-3 review of #3573 added an inline `MAX_SAMPLES_PER_KEY = 4_096` per-key sample cap in `evaluator.rs` (drop-oldest, defending threat-model finding **D5**: attacker-triggered hot capabilities with very large declared windows). This PR's extraction into `PredicateStateBackend` initially missed porting that cap into the backend trait impl — so once rebased onto `hooks-foundation-01` the defense would have silently disappeared from production.

Ported in commit `1594fe5c7`:

- Added `pub const MAX_SAMPLES_PER_KEY: usize = 4_096;` to `predicate_state.rs` next to `MAX_HISTORY_KEYS` / `MAX_KEYS_PER_TENANT`, with rustdoc citing finding D5.
- Enforced in both `InMemoryPredicateStateBackend::record_invocation` and `record_value` — after window-cutoff prune and dedup-skip check, before `push_back`.
- For the NumericSum path, eviction uses the bucket helper's `pop_front`, which decrements `running_sum` — so the incremental running-sum invariant (`running_sum == sum(entries values)`) is preserved across cap-driven eviction.
- Added two regression tests:
  - `record_invocation_caps_samples_per_key_under_attacker_pressure` — drives `MAX_SAMPLES_PER_KEY + 64` invocations within one window, asserts bucket size pinned at the cap and `dedup_ids` stays in sync.
  - `record_value_evicts_oldest_keeping_running_sum_consistent` — drives `MAX_SAMPLES_PER_KEY + 32` value inserts with constant value, asserts `running_sum` equals deque sum AND equals `cap * value`.

Then rebased onto `origin/hooks-foundation-01` (`3b46c16f4`). Conflicts in `evaluator.rs`: the pre-existing inline `MAX_SAMPLES_PER_KEY` enforcement (now redundant since the backend owns it) and the inline `entries.pop_front()` / `entries.push_back()` blocks were deleted in favor of the existing `backend.record_invocation` / `backend.record_value` call sites.

Test results post-rebase:
- `cargo test -p ironclaw_hooks`: 216 passed, 0 failed
- `cargo test -p ironclaw_reborn`: all green
- `cargo clippy --all --benches --tests --examples --all-features -- -D warnings`: clean

zmanian:
## Post-merge Codex review

**Verdict: REQUEST CHANGES** — two real correctness bugs found. Recommend follow-up PR.

### Critical issues

**1. Per-key sample eviction silently weakens cap enforcement AND breaks replay-refusal.**
`predicate_state.rs:542`, `:596`; `evaluator.rs:189`.

Once a bucket reaches `MAX_SAMPLES_PER_KEY`, the backend drops an in-window sample before returning the count/sum. An `InvocationCount { max: 4096 }` will **never exceed max** — the eviction guarantees `count <= MAX_SAMPLES_PER_KEY`. Thresholds larger than 4096 are unenforceable. `NumericSum` can similarly undercount many small values. Also: evicting also removes the event id from `dedup_ids`, so replay refusal is no longer true for an in-window event that was cap-evicted.

Fix: do not silently evict in-window samples for security decisions. Return an overflow signal/error so the evaluator fails closed, or reject/limit predicates whose thresholds exceed representable history. If replay refusal remains a trait guarantee, dedup state must outlive sample eviction for the logical window.

**2. Long windows collapse to "only now" when `Instant::checked_sub` underflows.**
`predicate_state.rs:522`, `:582`.

For windows longer than the monotonic clock's representable past, `unwrap_or(now)` makes `cutoff = now`, trimming otherwise-valid prior entries. A common `24h` window after a recent boot can behave like an effectively zero-width window.

Fix: trim by age instead, e.g. `now.checked_duration_since(front_ts).is_some_and(|age| age > window)`, with a conservative policy for future timestamps.

### Recommendations

- Add a real global `MAX_HISTORY_KEYS` public-API regression: current tests cover the per-tenant quota, but not the `history.len() >= MAX_HISTORY_KEYS` branch after the quota change. Use multiple tenants to fill the map without hitting one tenant's quota, then assert the global LRU victim and `evictions_observed()`.
- Align the durable event-id contract before implementing SQL backends: `predicate_state.rs:812`, `docs/successors/03-persistent-counter.md:91` pin a 64-char blake3 hex id while the successor schema says `uuid`; Postgres `uuid` is not a 64-char digest sink. Use `text`/`bytea`, or validate/mint 128-bit UUID-shaped IDs.
- Add one dispatcher-level sanitization regression for predicate audit reasons with CR/LF/control chars. The code path currently flows through `telemetry::sanitize_audit_reason`, but the predicate wrapper test only asserts the raw sink field.

PR #3641, opened 2026-05-14
plumb temperature through Responses API

zetyquickly:
## Summary
Introduces per-request `temperature` support on `/v1/responses` (and `/api/v1/responses`). The endpoint previously rejected any payload with `temperature` set, forcing clients to rely on settings-level defaults only. Now the value flows through inbound metadata and is honoured as the highest-precedence override at the first agent-loop iteration, with the existing user/admin settings fallback unchanged.

## Changes
- `src/channels/web/responses_api.rs` — drop the 400-rejection block; inject `req.temperature` into the inbound message metadata alongside `x_context`. Validate at the API boundary: non-finite values and anything outside `[0, 2]` are rejected with a 400 `invalid_request_error` before an `IncomingMessage` is enqueued (matches surrounding validation pattern and OpenAI's Responses contract).
- `src/agent/dispatcher.rs` — extract `metadata["temperature"]` into `reason_ctx.temperature` (per-request beats settings beats admin beats hardcoded default); precedence helper `resolve_temperature_overrides` plus unit tests for each branch.
- **Off-by-one fix (called out per review):** the existing iteration-gated branch ran under `iteration == 0`, but `agentic_loop` starts iterations at `1` (`for iteration in 1..=max_iterations`). That gate was therefore *never* firing, which silently disabled the existing settings-level `temperature` fallback **and** the `selected_model` (`/model`) override on `main`. This PR moves the gate to `iteration == 1` so both the new per-request override and those pre-existing settings/model paths actually run. This is a real behaviour change for callers who rely on settings-level `temperature` defaults or `/model` selection.
- `tests/fixtures/llm_traces/tools/job_list_cancel.json` — fixture update is a direct consequence of the iteration-gate fix above. Activating the previously-dormant `selected_model` lookup changes the per-step model resolution for this trace's sub-jobs, which shifts the captured step shape. The fixture update is in this PR because it's the same one-line behaviour change that causes it; splitting them out would leave one half failing in CI.

Precedence:

```
per-request (Responses API) > user setting > admin setting > hardcoded default
```

## Test plan
- [x] Smoke: `curl -X POST /v1/responses -d '{"temperature":0.3,...}'` no longer returns 400
- [x] `cargo test --test responses_api_temperature` — covers (a) value lands in metadata, (b) absent field omits metadata, (c) out-of-range / non-finite values reject with 400 and do not enqueue an `IncomingMessage`
- [x] `cargo test first_call_llm_iteration_is_one` — locks in the agentic-loop iteration start
- [x] `cargo test temperature_overrides_` — covers each branch of the precedence helper
- [x] Existing `resolve_settings_temperature_*` unit tests in `dispatcher.rs` still pass

### DIFF
--- a/src/channels/web/responses_api.rs
+++ b/src/channels/web/responses_api.rs
@@ -740,6 +740,19 @@ pub async fn create_response_handler(
             "invalid_request_error",
         ));
     }
+    // Reject non-finite or out-of-range temperature at the API boundary.
+    // The provider-side path clamps to [0, 2] later, but callers modelling
+    // against the OpenAI Responses contract expect a 400 here rather than
+    // silent normalization (e.g. 9.0 → 2.0).
+    if let Some(t) = req.temperature
+        && (!t.is_finite() || !(0.0..=2.0).contains(&t))
+    {
+        return Err(api_error(
+            StatusCode::BAD_REQUEST,
+            "The 'temperature' field must be a finite number in [0, 2]",
+            "invalid_request_error",
+        ));
+    }
 
     let mut content = extract_user_content(&req.input)
         .map_err(|e| api_error(StatusCode::BAD_REQUEST, e, "invalid_request_error"))?;
--- a/src/channels/web/responses_api.rs
+++ b/src/channels/web/responses_api.rs
@@ -733,13 +733,6 @@ pub async fn create_response_handler(
             "invalid_request_error",
         ));
     }
-    if req.temperature.is_some() {
-        return Err(api_error(
-            StatusCode::BAD_REQUEST,
-            "Per-request 'temperature' is not supported on this endpoint; configure the default via settings",
-            "invalid_request_error",
-        ));
-    }
     if req.max_output_tokens.is_some() {
         return Err(api_error(
             StatusCode::BAD_REQUEST,
@@ -788,6 +781,9 @@ pub async fn create_response_handler(
     });
     if let Some(ref ctx) = req.x_context {
         metadata["context"] = ctx.clone();
+    }
+    if let Some(t) = req.temperature {
+        metadata["temperature"] = serde_json::json!(t);
     }
     let msg = crate::channels::web::util::web_incoming_message_with_metadata(
         "gateway",
--- a/src/agent/agentic_loop.rs
+++ b/src/agent/agentic_loop.rs
@@ -452,6 +452,7 @@ mod tests {
         tool_exec_count: AtomicUsize,
         tool_exec_outcome: Mutex<Option<LoopOutcome>>,
         iterations_seen: Mutex<Vec<usize>>,
+        call_llm_iterations: Mutex<Vec<usize>>,
         early_exit: Mutex<Option<(usize, LoopOutcome)>>,
         nudge_count: AtomicUsize,
         /// When true, execute_tool_calls sets last_tool_batch_all_failed = true.
@@ -466,6 +467,7 @@ mod tests {
                 tool_exec_count: AtomicUsize::new(0),
                 tool_exec_outcome: Mutex::new(None),
                 iterations_seen: Mutex::new(Vec::new()),
+                call_llm_iterations: Mutex::new(Vec::new()),
                 early_exit: Mutex::new(None),
                 nudge_count: AtomicUsize::new(0),
                 simulate_all_failed: false,
@@ -510,8 +512,9 @@ mod tests {
             &self,
             _reasoning: &Reasoning,
             _reason_ctx: &mut ReasoningContext,
-            _iteration: usize,
+            iteration: usize,
         ) -> Result<ironclaw_llm::RespondOutput, crate::error::Error> {
+            self.call_llm_iterations.lock().await.push(iteration);
             let mut responses = self.llm_responses.lock().await;
             if responses.is_empty() {
                 panic!("MockDelegate: no more LLM responses queued");
@@ -554,6 +557,41 @@ mod tests {
     }
 
     // --- Tests ---
+
+    #[tokio::test]
+    async fn first_call_llm_iteration_is_one() {
+        // Regression: ChatDelegate::call_llm gates per-request overrides
+        // (Responses API `temperature`, settings `temperature`, `selected_model`
+        // from `/model`) on the first iteration. The gate's expected value must
+        // match what this loop produces on its first call_llm invocation. An
+        // off-by-one here silently disables every override and was shipped once
+        // already (gate was `iteration == 0` while the loop starts at 1).
+        let tool_call = ToolCall {
+            id: "call_1".to_string(),
+            name: "echo".to_string(),
+            arguments: serde_json::json!({}),
+            reasoning: None,
+            signature: None,
+        };
+        let delegate = MockDelegate::new(vec![
+            tool_calls_output(vec![tool_call]),
+            text_output("done"),
+        ]);
+        let reasoning = stub_reasoning();
+        let mut ctx = ReasoningContext::new();
+        let config = AgenticLoopConfig::default();
+
+        run_agentic_loop(&delegate, &reasoning, &mut ctx, &config)
+            .await
+            .unwrap();
+
+        let iterations = delegate.call_llm_iterations.lock().await;
+        assert_eq!(
+            iterations.first().copied(),
+            Some(1),
+            "first call_llm iteration must be 1 — see dispatcher.rs override gate"
+        );
+    }
 
     #[tokio::test]
     async fn test_text_response_returns_immediately() {
--- a/src/agent/dispatcher.rs
+++ b/src/agent/dispatcher.rs
@@ -47,6 +47,40 @@ fn resolve_settings_temperature(
         .map(|t| (t as f32).clamp(0.0, 2.0))
 }
 
+/// Apply the full iteration-1 temperature precedence rule:
+///
+/// 1. **Existing context value** wins — set by an upstream caller (e.g. a
+///    reasoning-model default) before the agentic loop started.
+/// 2. **Per-request metadata** — `metadata["temperature"]` from the inbound
+///    `IncomingMessage`. Populated by the Responses API handler when a
+///    caller passes a `temperature` field.
+/// 3. **Settings store fallback** — user/admin default via
+///    `resolve_settings_temperature`, clamped to `[0.0, 2.0]`.
+///
+/// Returns `Some(t)` when any source supplies a value, `None` when none do.
+///
+/// Centralizing all three sources here means the dispatcher's iteration-1
+/// gate has a single call site with all inputs visible — a future refactor
+/// that drops the metadata read must change this signature instead of
+/// silently deleting a branch inside the gate body.
+fn resolve_temperature_overrides(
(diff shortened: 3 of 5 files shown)


src/channels/web/responses_api.rs
```diff
@@ -789,6 +782,9 @@ pub async fn create_response_handler(
     if let Some(ref ctx) = req.x_context {
         metadata["context"] = ctx.clone();
     }
+    if let Some(t) = req.temperature {
```
serrrfirat, line 798: **Medium Severity — Missing caller-level regression coverage**

This PR changes public Responses API behavior by accepting `temperature` and copying it into inbound message metadata, but the coverage currently only exercises the `resolve_settings_temperature` helper. I don't see a caller-level test proving that `/v1/responses` / `/api/v1/responses` actually preserves the request temperature through `IncomingMessage.metadata`, or that the dispatcher consumes that per-request value before applying user/admin settings fallback.

Given the project rule to test through the caller when a helper gates behavior, please add coverage around the endpoint/dispatcher path, e.g.:
- a Responses API request with `temperature` succeeds and produces metadata containing that value
- per-request temperature wins over settings/admin fallback

serrrfirat:
## Summary
Reviewed PR #3641 only. Base `34eeeaf05a656007927afe0b5c6929ae75c3f866` → head `f8e07be484e5053c29fdc16cfc86ad32214541c4`.

PR accepts `temperature` on the Responses API and writes it into message metadata, but the agent-side override block is behind an unreachable first-iteration condition. Merge stance: blocking correctness finding.

## Findings

| 1 | High | Correctness | `src/agent/dispatcher.rs:607` | New per-request temperature plumbing never runs. `responses_api.rs:785-786` stores `req.temperature` in message metadata, but `call_llm()` only reads that metadata inside `if iteration == 0`. The shared agent loop starts at `1` (`src/agent/agentic_loop.rs:226`), so Responses API callers can send `temperature` and still get default/settings temperature on every turn. | Change first-call override logic to match the real first loop iteration (for example `iteration == 1`) or track whether overrides were already applied in loop state. Add a caller-boundary regression where a Responses API request with `temperature` reaches a test provider / reasoning context with that exact value. |

## Security/data-flow notes
- Input path checked: `ResponsesRequest.temperature` → message metadata → `ChatDelegate::call_llm()` → `ReasoningContext.temperature` → provider request.
- No auth, secret, or logging leak found in changed path.

- `src/channels/web/responses_api.rs:65`: request accepts `temperature`.
- `src/channels/web/responses_api.rs:785-786`: handler stores it in metadata.
- `src/agent/dispatcher.rs:607-620`: consumer only runs when `iteration == 0`.
- `src/agent/agentic_loop.rs:226`: loop calls delegates with `iteration` starting at `1`.

## Missing tests
- Add integration/caller-boundary coverage for `POST /api/v1/responses` with `temperature`, asserting the provider/agent loop sees the requested value. Helper-only metadata tests would not catch this dead gate.

src/channels/web/responses_api.rs
zetyquickly: not addressed yet

zetyquickly:
> ## Summary
> Reviewed PR #3641 only. Base `34eeeaf05a656007927afe0b5c6929ae75c3f866` → head `f8e07be484e5053c29fdc16cfc86ad32214541c4`.
> 
> PR accepts `temperature` on the Responses API and writes it into message metadata, but the agent-side override block is behind an unreachable first-iteration condition. Merge stance: blocking correctness finding.
> 
> ## Findings
> #	Sev	Category	File:Line	Issue	Fix suggestion
> 1	High	Correctness	`src/agent/dispatcher.rs:607`	New per-request temperature plumbing never runs. `responses_api.rs:785-786` stores `req.temperature` in message metadata, but `call_llm()` only reads that metadata inside `if iteration == 0`. The shared agent loop starts at `1` (`src/agent/agentic_loop.rs:226`), so Responses API callers can send `temperature` and still get default/settings temperature on every turn.	Change first-call override logic to match the real first loop iteration (for example `iteration == 1`) or track whether overrides were already applied in loop state. Add a caller-boundary regression where a Responses API request with `temperature` reaches a test provider / reasoning context with that exact value.
> ## Security/data-flow notes
> * Input path checked: `ResponsesRequest.temperature` → message metadata → `ChatDelegate::call_llm()` → `ReasoningContext.temperature` → provider request.
> * No auth, secret, or logging leak found in changed path.
> 
> ## Correctness/invariant notes
> * `src/channels/web/responses_api.rs:65`: request accepts `temperature`.
> * `src/channels/web/responses_api.rs:785-786`: handler stores it in metadata.
> * `src/agent/dispatcher.rs:607-620`: consumer only runs when `iteration == 0`.
> * `src/agent/agentic_loop.rs:226`: loop calls delegates with `iteration` starting at `1`.
> 
> ## Missing tests
> * Add integration/caller-boundary coverage for `POST /api/v1/responses` with `temperature`, asserting the provider/agent loop sees the requested value. Helper-only metadata tests would not catch this dead gate.

addressed

src/channels/web/responses_api.rs
zetyquickly: should be addressed with [33b78a8](https://github.com/nearai/ironclaw/pull/3641/commits/33b78a8c901d5d89d3101edf9e5b1b9c2b291c89)

nickpismenkov (dismissed):
lgtm

serrrfirat (changes requested):
Reviewed current head `9da08c53082f331f9c72909d13f76af0b27665cb` against base `201e2164f79a9226744e18ab8c6dddd0faa00a96`. The shape is right: the prior `iteration == 0` bug is fixed, the Responses API now has caller-level metadata coverage, and the dispatcher precedence helper covers the intended ordering. A couple of items should be addressed before merge.

1. P2: Validate `temperature` at the API boundary

`src/channels/web/responses_api.rs:785` forwards any deserialized `temperature` into metadata, and `src/agent/dispatcher.rs:74-79` accepts the per-request value verbatim. The test at `src/agent/dispatcher.rs:3394` explicitly locks in that per-request values are not clamped at this layer, relying on `Reasoning::respond_with_tools` to clamp before provider dispatch.

That keeps providers safe, but it makes the Responses API silently turn `temperature: 9.0` into `2.0` later instead of returning a caller-visible `400`, unlike the surrounding request validation blocks. This endpoint is OpenAI-compatible enough that callers will expect the request boundary to reject values outside `[0, 2]`, not silently normalize them.

Please reject non-finite / out-of-range values in `create_response_handler` and add a negative handler test asserting `temperature < 0` and `temperature > 2` return `invalid_request_error` without enqueueing an `IncomingMessage`.

2. P3: PR description is now materially stale

The PR body still says the override happens at `iteration == 0`, but the actual change fixes the loop gate to `iteration == 1`. That also revives existing settings-level `temperature` and `selected_model` behavior, and the fixture update in `tests/fixtures/llm_traces/tools/job_list_cancel.json` appears tied to that broader behavior change.

Please update the PR description to call out the off-by-one fix and explain why the trace fixture update belongs in this PR. This matters for scope honesty and for anyone bisecting the settings/model override behavior later.

Verification run locally:
- `cargo test --test responses_api_temperature`
- `cargo test first_call_llm_iteration_is_one`
- `cargo test temperature_overrides_`

Net: the core mechanism is sound, but the public request boundary should fail loudly for invalid `temperature` values before this lands.

zetyquickly:
> Reviewed current head `9da08c53082f331f9c72909d13f76af0b27665cb` against base `201e2164f79a9226744e18ab8c6dddd0faa00a96`. The shape is right: the prior `iteration == 0` bug is fixed, the Responses API now has caller-level metadata coverage, and the dispatcher precedence helper covers the intended ordering. A couple of items should be addressed before merge.
> 
> 1. P2: Validate `temperature` at the API boundary
> 
> `src/channels/web/responses_api.rs:785` forwards any deserialized `temperature` into metadata, and `src/agent/dispatcher.rs:74-79` accepts the per-request value verbatim. The test at `src/agent/dispatcher.rs:3394` explicitly locks in that per-request values are not clamped at this layer, relying on `Reasoning::respond_with_tools` to clamp before provider dispatch.
> 
> That keeps providers safe, but it makes the Responses API silently turn `temperature: 9.0` into `2.0` later instead of returning a caller-visible `400`, unlike the surrounding request validation blocks. This endpoint is OpenAI-compatible enough that callers will expect the request boundary to reject values outside `[0, 2]`, not silently normalize them.
> 
> Please reject non-finite / out-of-range values in `create_response_handler` and add a negative handler test asserting `temperature < 0` and `temperature > 2` return `invalid_request_error` without enqueueing an `IncomingMessage`.
> 
> 2. P3: PR description is now materially stale
> 
> The PR body still says the override happens at `iteration == 0`, but the actual change fixes the loop gate to `iteration == 1`. That also revives existing settings-level `temperature` and `selected_model` behavior, and the fixture update in `tests/fixtures/llm_traces/tools/job_list_cancel.json` appears tied to that broader behavior change.
> 
> Please update the PR description to call out the off-by-one fix and explain why the trace fixture update belongs in this PR. This matters for scope honesty and for anyone bisecting the settings/model override behavior later.
> 
> Verification run locally:
> 
> * `cargo test --test responses_api_temperature`
> * `cargo test first_call_llm_iteration_is_one`
> * `cargo test temperature_overrides_`
> 
> Net: the core mechanism is sound, but the public request boundary should fail loudly for invalid `temperature` values before this lands.

addressed

serrrfirat (approved):
## PR Review: plumb temperature through Responses API (#3641)

### Status of prior review items
- **serrrfirat P1** (`iteration == 0` unreachable gate): ✅ Fixed → `iteration == 1`
- **serrrfirat P2** (API boundary rejection for out-of-range temperature): ✅ Fixed (400 rejected before enqueue)
- **serrrfirat P3** (stale PR description): ✅ PR body now documents off-by-one fix and fixture rationale

### Findings

| # | Severity | Category | Status |
|---|----------|----------|--------|
| F1 | Clean | All modules | **No blocking or medium-severity issues found** |
| F2 | Info | Documentation/Tests | Documented non-clamping contract — correct |
| F3 | Info | Tests | 8 new trace steps for race fix — intentional |

### Data flow verified
```
ResponsesRequest.temperature → create_response_handler validation → metadata["temperature"] → resolve_temperature_overrides → reason_ctx.temperature → provider request
```

### Key checks
1. **Validation** rejects non-finite OR out-of-range `[0, 2]` with 400 before enqueue ✅
2. **Metadata write** only fires when `req.temperature.is_some()` — no fabricated values ✅
3. **Precedence**: `current` > `metadata` > `settings (clamped)` — matches contract ✅
4. **Iteration gate** `== 1` matches `run_agentic_loop` starting at 1 ✅
5. **Unit tests**: 6 test functions covering all `resolve_temperature_overrides` branches ✅
6. **Integration tests**: 3 tests covering metadata propagation, rejection, and absence ✅
7. **Regression test**: `first_call_llm_iteration_is_one` locks in iteration value ✅
8. **model_override** also newly active — previously dormant on main ✅

**Verdict: ✅ APPROVE.** The mechanism is sound, precedence is correct, validation is proper at the API boundary, and test coverage spans helper, integration, and regression layers.

Issue #3603, opened 2026-05-14
HostTrustPolicy::empty() reports ProductionValidated — readiness vs trust-policy inconsistency


zmanian:
## Problem

In `crates/ironclaw_reborn_composition`, `HostTrustPolicy::empty()` is hardcoded in the production-readiness path, yet the same path reports `ProductionValidated`. This is a security-posture overpromise: the system claims trust validation is complete when no policy is actually configured.

## Why this matters

`HostTrustPolicy` is the gate that decides which trust classes (`FirstParty`, `System`, `Sandbox`, etc.) are allowed for which capabilities/hosts. With `::empty()`:

- Every trust check passes vacuously (or fails-closed in the wrong direction, depending on default).
- Any extension/channel asking for elevated trust gets a misleading signal from the readiness report.
- Once third-party extensions actually run through this path, the placeholder will silently allow (or deny) behavior the operator never approved.

This compounds with the TrustClass placeholder issue — together they make trust enforcement non-functional despite the surface API suggesting otherwise.

## Expected

Either:

1. `ProductionValidated` should require a non-empty `HostTrustPolicy` derived from real configuration (preferred), or
2. Readiness reporting must explicitly call out `HostTrustPolicy::empty()` as a non-production state and fail the gate in `Production` mode.

## References

- #3566 (open) — Reborn composition crate
- Re-review notes: [#3566 comment](https://github.com/nearai/ironclaw/pull/3566#issuecomment-4445610740)

## Acceptance

- [ ] `HostTrustPolicy` is sourced from real configuration in production composition
- [ ] Readiness gate rejects `::empty()` policy in `Production` mode
- [ ] Test: composition with empty policy + production mode → fails startup

PR #3651, opened 2026-05-14
arch(ws-14): register planned driver default path

henrypark133:
## Context

Default planned-driver registration branch. It adds the registry/profile helpers that let Reborn select the planned default path while keeping live production cutover deferred.

Workstream brief: `docs/reborn/agent-loop-briefs/planned-driver-registration.md`
Stack base: `arch/ws-14-parent`

## What landed

- Planned driver factory and registration helpers.
- `reborn-planned-default` run-profile selection support.
- Explicit planned-default resolver constructor while preserving existing text-only/default resolver behavior.
- Planned-driver factory tests and E2E planned-driver coverage over the integrated host-port parent.
- Documentation alignment so WS14 is registration/profile selection, not full product-live cutover.

## Reviewer focus

- The planned default selector should be explicit; existing defaults should not silently change.
- Driver registration should compose through the registry/profile layer rather than hardcoding one-off paths.
- The branch should not claim product-live readiness; WS16/WS17 own that evidence.

## Non-goals / deferred work

- Production runtime composition with all live-required services is WS16.
- Product no-profile inbound cutover and persisted reply evidence is WS17.
- Text-only rollback profile behavior should remain available.

## Validation

- [x] `cargo test -p ironclaw_reborn planned_driver_factory --features ironclaw_agent_loop/test-support`
- [x] `cargo test -p ironclaw_reborn --test planned_driver_e2e --features ironclaw_agent_loop/test-support`
- [x] `cargo test -p ironclaw_turns --test run_profile_contract`
- [x] `cargo check -p ironclaw_turns -p ironclaw_loop_support -p ironclaw_agent_loop -p ironclaw_reborn --features ironclaw_agent_loop/test-support`
- [x] `git diff --check`

## Stack position

```
```

### DIFF
--- a/crates/ironclaw_reborn/src/lib.rs
+++ b/crates/ironclaw_reborn/src/lib.rs
@@ -11,6 +11,7 @@ pub mod loop_exit_applier;
 mod milestone_events;
 mod model_routes;
 mod planned_driver;
+mod planned_driver_factory;
 pub mod production_readiness;
 mod text_loop_driver;
 pub mod turn_runner;
@@ -43,4 +44,14 @@ pub use model_routes::{
     ResolvedModelRouteSnapshot, StaticModelRouteResolver,
 };
 pub use planned_driver::PlannedDriver;
+pub use planned_driver_factory::{
+    DefaultPlannedDriverBuild, DefaultPlannedDriverRegistrationError, PLANNED_DEFAULT_PROFILE_ID,
+    PLANNED_DRIVER_CHECKPOINT_SCHEMA_ID, PLANNED_DRIVER_CHECKPOINT_SCHEMA_VERSION,
+    PLANNED_DRIVER_DEFAULT_ID, PLANNED_DRIVER_DEFAULT_VERSION, default_planned_driver,
+    default_planned_run_profile_resolver, planned_default_profile_id,
+    planned_driver_checkpoint_schema_id, planned_driver_checkpoint_schema_version,
+    planned_driver_default_id, planned_driver_default_version, planned_driver_descriptor,
+    register_default_planned_driver, register_default_planned_profile,
+    register_default_text_only_driver,
+};
 pub use text_loop_driver::{TextOnlyModelReplyDriver, TextOnlyModelReplyDriverConfig};
--- a/crates/ironclaw_reborn/src/planned_driver.rs
+++ b/crates/ironclaw_reborn/src/planned_driver.rs
@@ -36,6 +36,25 @@ pub struct PlannedDriver {
 }
 
 impl PlannedDriver {
+    pub fn from_family_with_descriptor(
+        family: Arc<LoopFamily>,
+        executor: Arc<CanonicalAgentLoopExecutor>,
+        descriptor: AgentLoopDriverDescriptor,
+    ) -> Result<Self, AgentLoopDriverError> {
+        if descriptor.checkpoint_schema_id.is_none()
+            || descriptor.checkpoint_schema_version.is_none()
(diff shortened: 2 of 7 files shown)


zmanian:
## Review notes — WS14 planned default registration

Genuine opt-in for the planned path: default `InMemoryRunProfileResolver` still resolves to `interactive_default → lightweight_loop`; only `new_with_implicit_default` or `default_planned_run_profile_resolver` switches the implicit default. `descriptor_for_family` now requires a checkpoint schema (good — closes a previously implicit invariant). Registration is order-tolerant via typed `DuplicateProfile` error. One item before this lands.

### Text-only registration tagged Production + all-optional

`register_default_text_only_driver` uses `DriverRequirements::all_optional()` but marks `DriverKind::Production`. The readiness matrix (Production + Production) passes silently. This is the same class of bug as the `HostTrustPolicy::empty()` regression (issue #3603): a "Production" tag that the readiness check trusts, but with placeholder semantics underneath.

Two options:

1. Retag text-only as `DriverKind::Reference` so the Production readiness check requires explicit non-optional ports.
2. Make per-port requirements explicit instead of `all_optional()`.

Either is fine; the current shape promises Production-ready behavior the type doesn't actually require.

### `PLANNED_DRIVER_CHECKPOINT_SCHEMA_ID = CHECKPOINT_SCHEMA_ID` aliasing

The const aliases the upstream constant. If the upstream schema id ever diverges, this duplication silently binds to the wrong value. A `const_assert_eq!` or single source of truth (just import and use directly) would be safer.

### Minor

- `RunProfileResolutionError::ProfileUnavailable { profile_id }` in the `None`-requested branch reports the *resolved* implicit profile id, not the original request — minor observability degradation when an operator misconfigures the implicit default.
- `planned_default_profile_definition` mints `CapabilitySurfaceProfileId::new("interactive_tools")` inline; hoist to a const next to other ids.
- Many helpers return `Result<_, String>`; a typed error would match the rest of the module.

The planned default itself is correctly marked Production *because* its ports are all `Required` — that discipline is the right pattern; just please extend it to text-only.

serrrfirat:
## Summary
Reviewed WS14 PR #3651 only.  
Base `f46ef664c73fdd2f3982d11db6fcf0492da898f5` → head `8d39d6d0a5f23af947677cd137b621854d002638`.

No unintended cutover found, but actual default planned-driver registration is not wired into production/default coordinator path.

Validation:
- `cargo check -p ironclaw_reborn` ✅

## Findings

| 1 | High | Correctness / Composition root | `crates/ironclaw_reborn/src/planned_driver_factory.rs:181-190`, `crates/ironclaw_turns/src/coordinator.rs:94-99`, `crates/ironclaw_reborn/tests/planned_driver_e2e.rs:57-88` | PR adds helper `default_planned_run_profile_resolver()`, but production/default coordinator still constructs `InMemoryRunProfileResolver::default()`. Tests call the helper directly, not real bootstrap/submit path. Normal `requested_run_profile=None` / `default` path still resolves legacy default at runtime. | Wire helper into production composition root behind intended gate, or retitle/rescope as groundwork. Add caller-boundary test through actual submit/default coordinator path proving `None` and `default` resolve to planned default. |

## Security/data-flow notes
- No auth/secret regression found.
- Safe from accidental cutover today because helper is not wired.

- PR title/intent says planned default registration; executable default path does not change.

## Missing tests
- Actual bootstrap/composition-root/default-submit path uses planned default resolver.
- Legacy aliases audited (`interactive_default`, `default`, `reborn-planned-default`).

PR #3658, opened 2026-05-14
feat(tui): Ctrl-S downloads logs from the Logs tab

zetyquickly:
## Summary

- Adds **Ctrl-S** while the Logs tab is focused to dump the current log ring buffer to `~/.ironclaw/logs/tui-logs-<UTC-timestamp>.log` (override location via `IRONCLAW_HOME`).
- Shows a toast on completion with the **full saved path** — the toast now sizes its width to the message length and wraps onto extra lines when the path is longer than the terminal width, so paths are never truncated.
- Surfaces the shortcut in two places:
  - Status-bar hint strip shows `^S save  ^L logs  ^B sidebar  ^C quit` while the Logs tab is active.
  - `F1` help overlay gains `Ctrl-S → Download logs (Logs tab)`.

Wired through `InputAction::DownloadLogs` and gated on `logs_active` so Ctrl-S is a no-op outside the Logs tab.

## Test plan

- [ ] `cargo check -p ironclaw_tui --all-features`
- [ ] `cargo clippy -p ironclaw_tui --all-features --tests -- -D warnings`
- [ ] Manual: launch TUI, switch to Logs tab (Ctrl-L), press Ctrl-S, confirm toast shows full path and the file exists with the expected content
- [ ] Manual: confirm Ctrl-S is a no-op on the Conversation tab
- [ ] Manual: shrink terminal so the path doesn't fit on one line; confirm toast wraps instead of truncating

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4427,6 +4427,7 @@ dependencies = [
  "ratatui",
  "serde",
  "serde_json",
+ "tempfile",
  "thiserror 2.0.18",
  "tokio",
  "tracing",
--- a/crates/ironclaw_tui/Cargo.toml
+++ b/crates/ironclaw_tui/Cargo.toml
@@ -31,3 +31,4 @@ image = { version = "0.25", default-features = false, features = ["png"], option
 
 [dev-dependencies]
 tokio = { version = "1", features = ["full"] }
+tempfile = "3"
--- a/crates/ironclaw_tui/src/app.rs
+++ b/crates/ironclaw_tui/src/app.rs
@@ -2525,15 +2525,20 @@ fn try_paste_clipboard_image(_state: &AppState) -> Option<TuiAttachment> {
 /// (falls back to `$IRONCLAW_HOME/logs/` or the current directory). Returns the
 /// human-readable target path on success.
 fn download_logs(state: &AppState) -> Result<String, String> {
+    let dir = log_output_dir().ok_or_else(|| "could not resolve log directory".to_string())?;
+    download_logs_to_dir(state, &dir)
+}
+
+/// Inner write step, parameterized on output dir so tests can use a tempdir
+/// without racing `IRONCLAW_HOME` env mutation across threads.
+fn download_logs_to_dir(state: &AppState, dir: &std::path::Path) -> Result<String, String> {
     use std::io::Write;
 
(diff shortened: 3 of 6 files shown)


serrrfirat:
## Summary
Reviewed PR #3658 only. Base `34eeeaf05a656007927afe0b5c6929ae75c3f866` → head `ae1390ad8c02e85e126a9e9b146fa799b57488a9`.

PR adds Ctrl-S export for TUI Logs tab. TUI-only, but it creates a new durable local log persistence path. Merge stance: fix Medium filesystem/privacy and overwrite issues.

## Findings

| 1 | Medium | Filesystem / Privacy | `crates/ironclaw_tui/src/app.rs:2533`, `crates/ironclaw_tui/src/app.rs:2537`, `src/channels/tui.rs:338` | Export writes raw log `target` and `message` fields to disk using `File::create` with default OS permissions. On Unix this is `0666 & umask`, so permissive/shared hosts can create group/world-readable exported logs. That turns transient in-memory logs into durable local files without explicit owner-only perms. | Create exports with owner-only permissions (Unix `OpenOptionsExt::mode(0o600)`) or set permissions immediately after create; keep Windows behavior safe. Add Unix regression checking exported file mode. |
| 2 | Medium | Correctness / Data loss | `crates/ironclaw_tui/src/app.rs:2530`, `crates/ironclaw_tui/src/app.rs:2533` | Filename timestamp has only second precision, then `File::create` truncates existing path. Two Ctrl-S presses in the same second, or pre-existing matching export, silently overwrites prior logs. Feature intended to preserve logs can destroy a previous export. | Use `create_new(true)` with retry/suffix or higher-resolution timestamp, and surface collision errors rather than truncating. Add collision/no-overwrite test. |

## Security/data-flow notes
- Source: log broadcaster forwards `entry.level/target/message/timestamp` to TUI log ring.
- Sink: `download_logs` writes same fields verbatim to filesystem.
- Trust transition: transient UI state → durable local artifact. No redaction or restrictive perms in changed path.

- Input mapping gates Ctrl-S to Logs tab (`crates/ironclaw_tui/src/input.rs:129-134`); routing looks scoped correctly.
- Main correctness issue is export collision/truncation.

## Missing tests
- Unix file mode check for log export.
- Collision/no-overwrite behavior for repeated downloads.
- Ctrl-S mapping scoped to Logs tab.

nickpismenkov (approved):
lgtm!

PR #3659, opened 2026-05-14
feat(filesystem): unified storage dispatch fabric — foundation + SQL backends + indexer

ilblackdragon:
## Summary

Consolidated foundation for the IronClaw storage rework. Replaces the per-crate `Store`/`Repository` sprawl with **one** `RootFilesystem` trait, **one** `Entry` type, and **one** set of ops (`put` / `get` / `delete` / `list_dir` / `query` / `ensure_index` / `stat` / `begin` / `append` / `tail`). CAS + versioning is universal — memory docs and project files get lost-update protection for free.

After review feedback that the foundation alone was "incomplete" — it advertised capabilities (query, native put, index) that no shipped backend actually implemented — this PR was rebased onto latest `reborn-integration` and absorbs the work formerly stacked in PR #3660 (port) and PR #3661 (indexer). Result: a self-contained foundation that ships native put/get + query/ensure_index on both SQL backends.

## What's in this PR

**Foundation layer**
- `Entry` / `VersionedEntry` / `RecordKind` / `RecordVersion` / `SeqNo` / `CasExpectation` / `ContentType` (`record.rs`)
- `IndexSpec` / `IndexName` / `IndexKey` / `IndexValue` / `IndexKind` / `Filter` / `Page` (`index.rs`)
- `StorageTxn` / `EventRecord` (`backend.rs`) — supporting types, not a parallel backend trait. `RootFilesystem` is the one trait.
- `RootFilesystem` extended in `root.rs` with the unified ops; legacy `read_file`/`write_file` retained for transitional callers.
- `ScopedFilesystem` forwards every new op under the mount-view permission check; `read_bytes`/`write_bytes` helpers added.
- `CompositeRootFilesystem` dispatches every new op by longest-prefix mount, with mount-time capability validation.
- `InMemoryBackend` — full reference impl; replaces the N per-crate `InMemory*Store` implementations consumer crates used to maintain.
- `LocalFilesystem` — native `put`/`get` for opaque-file entries with `CasExpectation::Any` (canonical pattern for byte-only backends).
- `BackendCapabilities` widened with `records`/`query`/`index`/`txn`/`events`; new error variants `VersionMismatch`/`Unsupported`/`IndexConflict`.

**SQL backends (formerly PR #3660)**
- New migration V28 + V29 plus idempotent libsql ALTERs add `content_type`/`kind`/`indexed`/`version` columns and the index-spec metadata table.
- Native `put` (Absent / Version / Any CAS), `get`, `query` (Filter→WHERE translation), `ensure_index` (expression indexes on `json_extract`/`->>'key'`) for both libsql and Postgres.
- Capabilities now declare `records: true`, `query: true`, `index.exact: true`, `index.prefix: true`, `txn: Cas`.
- Legacy `write_file`/`append_file` reset record metadata + bump version to prevent stale-metadata reads after mixed writes (Reviewer P0 from #3660).

**Indexer (formerly PR #3661)**
- `Filter::All` / empty `And` / empty `Or` emit explicit `TRUE`/`TRUE`/`FALSE` (no more "skip empty fragment" path that broke `Or([])`).
- `Filter::PrefixOn` uses a dedicated `escape_like_literal` so a literal `tenant:%` is escaped, not treated as a wildcard.
- Postgres `Filter::Range` on `IndexValue::I64` casts both sides to `BIGINT` for numeric comparison (was lexicographic on text).
- `IndexKey`/`IndexName` validation tightened to `[A-Za-z_][A-Za-z0-9_]*` — closes JSON-path traversal (`"a.b"` no longer maps to `$.a.b`) and DDL identifier injection.
- `ensure_index` race-idempotent via `INSERT ... ON CONFLICT DO NOTHING` + canonical-spec read-back.

**PR #3659 review medium-severity items (this commit)**
- `RootFilesystem::put`/`get` defaults are now `Unsupported` (the prior bridge introduced a put/write_file recursion cycle). `LocalFilesystem::put`/`get` provides the canonical byte-only-backend pattern.
- `InMemoryBackend::list_dir` upgrades a path to `Directory` on later child discovery (was: first-discovery wins, leaf-then-children misclassified).
- `InMemoryBackend::delete` now sweeps the subtree when no exact entry exists (matches SQL backends).
- `Filter::Range` requires `lo`/`hi`/stored value to share an `IndexValue` variant (was: cross-variant ordering via derived `Ord` produced bogus matches).
- `CompositeRootFilesystem::mount_dyn` rejects descriptors that over-claim capabilities on the new axes (records/query/index/events/txn).

## Test plan

- [x] `cargo test -p ironclaw_filesystem --all-features` — **72 tests pass** (25 unit + 5 catalog + 22 db-backed + 20 filesystem-contract)
- [x] Pre-push lib-test gate (4967 tests) — pass

## Docs / ADR

- `crates/ironclaw_filesystem/CLAUDE.md` rewritten to encode the invariants new code must preserve.
- New ADR `docs/reborn/2026-05-14-universal-fs-dispatch.md` explains the unification; `docs/reborn/2026-04-25-storage-catalog-and-placement.md` marked as superseded.

## Subsumes

- **#3660** — Port libsql + Postgres backends to native put/get with CAS (✅ absorbed)
- **#3661** — Native query + ensure_index on libsql + Postgres (✅ absorbed)

Those two PRs can be closed as subsumed once this lands. Consumer migrations (#3666, #3670, #3671, #3672) rebase onto the merged foundation.

### DIFF
--- a/crates/ironclaw_filesystem/src/catalog.rs
+++ b/crates/ironclaw_filesystem/src/catalog.rs
@@ -6,8 +6,8 @@ use ironclaw_host_api::VirtualPath;
 use crate::backend::{EventRecord, StorageTxn};
 use crate::{
     BackendCapabilities, BackendId, BackendKind, CasExpectation, ContentKind, DirEntry, Entry,
-    FileStat, FilesystemError, Filter, IndexPolicy, IndexSpec, Page, RecordVersion, RootFilesystem,
-    SeqNo, StorageClass, VersionedEntry, path_prefix_matches,
+    FileStat, FilesystemError, FilesystemOperation, Filter, IndexPolicy, IndexSpec, Page,
+    RecordVersion, RootFilesystem, SeqNo, StorageClass, VersionedEntry, path_prefix_matches,
 };
 
 /// Trusted catalog record for one virtual filesystem mount.
@@ -103,6 +103,12 @@ impl CompositeRootFilesystem {
                 path: descriptor.virtual_root,
             });
         }
+        // PR #3659 reviewer fix: validate the descriptor's advertised
+        // capabilities against the backend's actual capabilities at
+        // mount time. Catalog metadata that claims query/index/event
+        // support over a backend that doesn't provide it would defeat
+        // the PR's mount-time validation guarantee — fail closed instead.
+        validate_mount_capabilities(&descriptor, backend.capabilities())?;
         self.mounts.push(CompositeMount {
             descriptor,
             backend,
@@ -124,6 +130,74 @@ impl CompositeRootFilesystem {
 impl Default for CompositeRootFilesystem {
     fn default() -> Self {
         Self::new()
+    }
+}
+
+/// PR #3659 reviewer fix: reject a [`MountDescriptor`] whose advertised
+/// capabilities claim more than the backend actually delivers on the
+/// **new** capability axes (records, query, index, events, txn).
+///
+/// Scope deliberately limited to the new axes: the legacy bytes-plane
+/// flags (`read`/`write`/`list`/`stat`/`delete`/`append`) have always
+/// been descriptor-driven metadata, and many existing backends still
+/// return `BackendCapabilities::default()` (all-false) from their
+/// `capabilities()` accessor even though they implement
+/// `read_file`/`write_file` natively. The mount-time validation
+/// guarantee the reviewer asked for applies to the new capability
+/// surface that this PR introduces; downstream catalog clients are the
+/// authority for the legacy plane until each backend opts in to a more
+/// accurate `capabilities()` override.
+fn validate_mount_capabilities(
+    descriptor: &MountDescriptor,
+    backend: BackendCapabilities,
+) -> Result<(), FilesystemError> {
+    let declared = descriptor.capabilities;
+    let mut shortfalls: Vec<&'static str> = Vec::new();
+    if declared.records && !backend.records {
+        shortfalls.push("records");
+    }
+    if declared.query && !backend.query {
+        shortfalls.push("query");
+    }
+    if declared.index.exact && !backend.index.exact {
+        shortfalls.push("index.exact");
+    }
+    if declared.index.prefix && !backend.index.prefix {
+        shortfalls.push("index.prefix");
+    }
+    if declared.index.fts && !backend.index.fts {
+        shortfalls.push("index.fts");
+    }
+    if declared.index.vector && !backend.index.vector {
+        shortfalls.push("index.vector");
+    }
+    if declared.events && !backend.events {
+        shortfalls.push("events");
+    }
+    let backend_txn = txn_capability_rank(backend.txn);
+    let declared_txn = txn_capability_rank(declared.txn);
+    if declared_txn > backend_txn {
+        shortfalls.push("txn");
+    }
+    if shortfalls.is_empty() {
+        Ok(())
+    } else {
+        Err(FilesystemError::Backend {
+            path: descriptor.virtual_root.clone(),
+            operation: FilesystemOperation::MountLocal,
+            reason: format!(
+                "mount descriptor claims capabilities the backend does not provide: {}",
+                shortfalls.join(", ")
+            ),
+        })
+    }
+}
+
+fn txn_capability_rank(value: crate::TxnCapability) -> u8 {
+    match value {
+        crate::TxnCapability::None => 0,
+        crate::TxnCapability::Cas => 1,
+        crate::TxnCapability::MultiKey => 2,
     }
 }
 
--- a/crates/ironclaw_filesystem/src/catalog.rs
+++ b/crates/ironclaw_filesystem/src/catalog.rs
@@ -5,9 +5,9 @@ use ironclaw_host_api::VirtualPath;
 
 use crate::backend::{EventRecord, StorageTxn};
 use crate::{
-    BackendCapabilities, BackendId, BackendKind, CasExpectation, ContentKind, DirEntry, Entry,
-    FileStat, FilesystemError, FilesystemOperation, Filter, IndexPolicy, IndexSpec, Page,
-    RecordVersion, RootFilesystem, SeqNo, StorageClass, VersionedEntry, path_prefix_matches,
+    BackendCapabilities, BackendId, BackendKind, Capability, CasExpectation, ContentKind, DirEntry,
+    Entry, FileStat, FilesystemError, Filter, IndexPolicy, IndexSpec, Page, RecordVersion,
+    RootFilesystem, SeqNo, StorageClass, VersionedEntry, path_prefix_matches,
 };
 
 /// Trusted catalog record for one virtual filesystem mount.
@@ -152,43 +152,32 @@ fn validate_mount_capabilities(
     backend: BackendCapabilities,
 ) -> Result<(), FilesystemError> {
     let declared = descriptor.capabilities;
-    let mut shortfalls: Vec<&'static str> = Vec::new();
-    if declared.records && !backend.records {
-        shortfalls.push("records");
-    }
-    if declared.query && !backend.query {
-        shortfalls.push("query");
-    }
-    if declared.index.exact && !backend.index.exact {
-        shortfalls.push("index.exact");
-    }
-    if declared.index.prefix && !backend.index.prefix {
-        shortfalls.push("index.prefix");
-    }
-    if declared.index.fts && !backend.index.fts {
-        shortfalls.push("index.fts");
-    }
-    if declared.index.vector && !backend.index.vector {
-        shortfalls.push("index.vector");
-    }
-    if declared.events && !backend.events {
-        shortfalls.push("events");
-    }
-    let backend_txn = txn_capability_rank(backend.txn);
-    let declared_txn = txn_capability_rank(declared.txn);
-    if declared_txn > backend_txn {
-        shortfalls.push("txn");
-    }
-    if shortfalls.is_empty() {
+    // Only validate the **new** capability axes — legacy bytes flags stay
+    // descriptor-driven (see the function-level doc comment).
+    const NEW_AXES: &[Capability] = &[
+        Capability::Records,
+        Capability::Query,
+        Capability::IndexExact,
+        Capability::IndexPrefix,
+        Capability::IndexFts,
+        Capability::IndexVector,
+        Capability::Events,
+    ];
+    let mut shortfalls: Vec<Capability> = NEW_AXES
+        .iter()
+        .copied()
+        .filter(|cap| declared.has(*cap) && !backend.has(*cap))
+        .collect();
+    let backend_txn = txn_capability_rank(backend.txn());
+    let declared_txn = txn_capability_rank(declared.txn());
+    let txn_shortfall = declared_txn > backend_txn;
+    if shortfalls.is_empty() && !txn_shortfall {
         Ok(())
     } else {
-        Err(FilesystemError::Backend {
+        Err(FilesystemError::DescriptorOverclaims {
             path: descriptor.virtual_root.clone(),
-            operation: FilesystemOperation::MountLocal,
-            reason: format!(
-                "mount descriptor claims capabilities the backend does not provide: {}",
-                shortfalls.join(", ")
-            ),
+            missing: std::mem::take(&mut shortfalls),
+            txn_shortfall,
         })
     }
 }
--- a/crates/ironclaw_filesystem/src/catalog.rs
+++ b/crates/ironclaw_filesystem/src/catalog.rs
@@ -3,9 +3,11 @@ use std::sync::Arc;
 use async_trait::async_trait;
 use ironclaw_host_api::VirtualPath;
 
+use crate::backend::{EventRecord, StorageTxn};
 use crate::{
-    BackendCapabilities, BackendId, BackendKind, ContentKind, DirEntry, FileStat, FilesystemError,
-    IndexPolicy, RootFilesystem, StorageClass, path_prefix_matches,
+    BackendCapabilities, BackendId, BackendKind, CasExpectation, ContentKind, DirEntry, Entry,
+    FileStat, FilesystemError, Filter, IndexPolicy, IndexSpec, Page, RecordVersion, RootFilesystem,
+    SeqNo, StorageClass, VersionedEntry, path_prefix_matches,
 };
 
 /// Trusted catalog record for one virtual filesystem mount.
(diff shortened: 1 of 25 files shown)


serrrfirat:
## Summary
Reviewed PR #3659 only. Base `5d8c252620f6db8190e9c7067bc0d88f733403b8` → head `04224e111fbd8a5c1a9c9e39642f8b592bfc2202`.

PR introduces a unified root filesystem/mount fabric. Intent is good, but contract lands ahead of real backend parity and mount capability validation. Merge stance: not ready for v2 storage rollout.

## Findings

| 1 | High | Backend parity / Migration semantics | `crates/ironclaw_filesystem/src/catalog.rs:162`, `crates/ironclaw_filesystem/src/root.rs:51`, `crates/ironclaw_filesystem/src/local.rs:174`, `crates/ironclaw_filesystem/src/postgres.rs:49`, `crates/ironclaw_filesystem/src/libsql.rs:60` | `CompositeRootFilesystem` now forwards unified `put/get/query/ensure_index/begin/append/tail` to mounted backends, but persisted/local backends still implement only the legacy byte-plane methods. `RootFilesystem::put` defaults to `Unsupported`, and local/Postgres/libSQL do not override the unified record/CAS surface in this PR. Code migrated to “one fabric” can pass against `InMemoryBackend` then fail on production mounts. | Do not advertise/mount unified record/CAS capabilities until each production backend implements them, or route callers only to backends that actually support the requested op. Add caller-boundary tests for `CompositeRootFilesystem::put/get/query` on local/libSQL/Postgres mounts. |
| 2 | Medium | Mount trust boundary / Capability validation | `crates/ironclaw_filesystem/src/catalog.rs:92`, `crates/ironclaw_filesystem/src/catalog.rs:106`, `crates/ironclaw_filesystem/tests/catalog_contract.rs:43`, `crates/ironclaw_filesystem/tests/catalog_contract.rs:265` | `mount_dyn` accepts caller-supplied `MountDescriptor.capabilities` and never validates them against `backend.capabilities()`. Existing test mounts a local backend under `/memory` while advertising memory/vector indexing traits, and catalog reports those claims as truth. Routing/diagnostics can believe a mount supports query/index/vector semantics the backend cannot serve. | Validate descriptor capabilities against backend-discovered capabilities at mount time, or make descriptor capabilities purely descriptive and derive routable capabilities from backend. Add negative test for descriptor/backend mismatch. |
| 3 | Medium | Test-backend parity / Directory semantics | `crates/ironclaw_filesystem/src/in_memory.rs:1`, `crates/ironclaw_filesystem/src/in_memory.rs:132`, `crates/ironclaw_filesystem/src/in_memory.rs:144`, `crates/ironclaw_filesystem/src/root.rs:196`, `crates/ironclaw_filesystem/src/local.rs:258` | `InMemoryBackend` is documented as the default/reference test backend, but its directory behavior diverges from local/SQL backends. It synthesizes directories in `list_dir/stat`, yet `delete` only removes exact entries and returns `NotFound` for directory paths with children; it also does not override `create_dir_all`, so directory creation falls back to `Unsupported`. Migration tests can pass/fail opposite production behavior. | Align in-memory directory semantics with production backends: implement `create_dir_all`, recursive directory delete, and matching `stat/list` behavior. Add parity tests for directory create/list/delete across in-memory, local, libSQL, and Postgres. |

## Security/data-flow notes
- Scoped path permission gate still looks fail-closed in changed paths.
- Raw host path redaction appears preserved.
- New trust risk is mount metadata: descriptor claims become catalog truth without backend verification.

- “One trait / one set of ops” is not true for production backends yet.
- “Capabilities declared in catalog” needs enforcement or must be treated as non-authoritative metadata.
- In-memory backend is not a safe parity oracle for directory cleanup behavior yet.

## Missing tests
- Composite `put/get/query` through local/libSQL/Postgres mounts.
- Mount rejection or downgrade on capability mismatch.
- In-memory vs production directory create/delete parity.

crates/ironclaw_filesystem/src/root.rs
```diff
@@ -1,33 +1,114 @@
...
+
+    /// Write an [`Entry`] at `path` with a compare-and-swap precondition.
+    /// Returns the new [`RecordVersion`].
+    async fn put(
+        &self,
+        path: &VirtualPath,
+        _entry: Entry,
+        _cas: CasExpectation,
+    ) -> Result<RecordVersion, FilesystemError> {
+        unsupported(path, FilesystemOperation::WriteFile)
```
serrrfirat, line 66: **High Severity** — Unified writes fail on existing production backends

`RootFilesystem::put` defaults to `Unsupported`, while existing `LocalFilesystem`, `PostgresRootFilesystem`, and `LibSqlRootFilesystem` still only implement the legacy `write_file` path. New unified callers such as `ScopedFilesystem::write_bytes` now route through `put`, so opaque byte writes fail on backends that already support writing through `write_file`.

Suggested fix: make the default `put` implementation bridge opaque byte entries to `write_file` when CAS is compatible (or implement native `put` for local/Postgres/libSQL in this PR), and add caller-through tests for `ScopedFilesystem::write_bytes` against those backends.

crates/ironclaw_filesystem/src/in_memory.rs
```diff
@@ -0,0 +1,598 @@
...
+        Ok(state
+            .entries
+            .get(path.as_str())
+            .map(|stored| VersionedEntry {
+                entry: stored.entry.clone(),
+                version: stored.version,
+            }))
+    }
+
+    async fn delete(&self, path: &VirtualPath) -> Result<(), FilesystemError> {
```
serrrfirat, line 114: **Medium Severity** — Directory delete semantics diverge from other backends

`InMemoryBackend::delete` removes only an exact entry. But this backend's `stat` treats a path with children as a directory, so deleting `/dir` after writing `/dir/file` returns `NotFound` and leaves the subtree behind. Local/Postgres/libSQL delete the directory subtree.

Suggested fix: when no exact entry exists, treat paths with matching child prefix as directories and remove the subtree, or document a deliberate semantic difference and update contract tests accordingly.

crates/ironclaw_filesystem/src/in_memory.rs
```diff
@@ -0,0 +1,598 @@
...
+    match filter {
+        Filter::All => true,
+        Filter::Eq { key, value } => indexed.get(key) == Some(value),
+        Filter::PrefixOn { key, value } => match (indexed.get(key), value) {
+            (Some(IndexValue::Text(stored)), IndexValue::Text(prefix)) => {
+                stored.starts_with(prefix)
+            }
+            _ => false,
+        },
+        Filter::Range { key, lo, hi } => match indexed.get(key) {
```
serrrfirat, line 385: **Medium Severity** — Range filters compare unrelated indexed value types

`Filter::Range` relies on derived `IndexValue` enum ordering, so mismatched bound/value variants can match unrelated types. For example, a range with numeric lower bound and bytes upper bound can include bool/text values purely because of enum variant ordering, not domain ordering.

Suggested fix: require `lo`, `hi`, and stored value to have the same `IndexValue` variant before comparing; otherwise return false or `Unsupported`. Add tests for mixed-type range bounds.

crates/ironclaw_filesystem/src/catalog.rs
```diff
@@ -148,6 +150,77 @@ impl FilesystemCatalog for CompositeRootFilesystem {
 
 #[async_trait]
 impl RootFilesystem for CompositeRootFilesystem {
+    fn capabilities(&self) -> BackendCapabilities {
```
serrrfirat, line 216: **Medium Severity** — Composite/catalog capabilities can diverge from mounted backend reality

Anchoring on the new composite capability surface: `capabilities()` returns defaults, while `mount_dyn` also accepts `MountDescriptor.capabilities` without validating them against `backend.capabilities()`. That lets catalog metadata claim query/index/event support the mounted backend cannot actually provide, defeating the PR's stated mount-time validation guarantee.

Suggested fix: validate descriptor requirements against the backend's advertised capabilities during mount, reject impossible combinations with a fail-closed error, and add tests where a descriptor claims query/index support over a backend that does not provide it.

ilblackdragon:
**Consolidated foundation pushed.** Rebased onto latest `reborn-integration`, absorbed the work from #3660 (port) and #3661 (indexer), and addressed the 5 medium-severity review items:

1. ✅ `put`/`get` recursion — trait defaults now `Unsupported`; `LocalFilesystem` provides the canonical byte-only-backend pattern with native `put`/`get`.
2. ✅ `InMemoryBackend::list_dir` first-discovery misclassification — upgrades to `Directory` on later `has_more` discovery.
3. ✅ `InMemoryBackend::delete` directory subtree — now matches SQL backends' subtree semantics.
4. ✅ `Filter::Range` cross-variant comparison — requires shared `IndexValue` variant.
5. ✅ Composite mount-time capability validation — rejects descriptors that over-claim on the new axes.

Also includes the work formerly in #3660 (libsql + Postgres native put/get/CAS + stale-metadata fix) and #3661 (native query + ensure_index + the 5 indexer review fixes).

**Tests:** 72 filesystem tests pass (3 new regression tests for this round). Workspace + pre-push gate (4967 tests) clean.

The high-severity "unified writes fail on existing production backends" concern is fully addressed: `LocalFilesystem` ships native `put`/`get` in this PR, and both SQL backends ship native put/get/query/ensure_index.

If this looks good, #3660 and #3661 can be closed as subsumed.

ilblackdragon:
Second-round review fixes pushed (commit `af8bbf184`).

**1. `BackendCapabilities` refactored to enum + bitmask.** The 13-field struct of booleans (plus the parallel `IndexCapability` substruct) is replaced by a u32 bitmask over a new `Capability` enum (Read / Write / Append / List / Stat / Delete / Records / Query / IndexExact / IndexPrefix / IndexFts / IndexVector / Events). Build with `BackendCapabilities::empty().with(...)` or use the `sql_typical` / `in_memory_full` / `bytes_only` constructors. Wire shape stays readable: `{ caps: [\"read\", \"write\", ...], txn: \"cas\" }`. Mirrors the existing `FilesystemOperation` enum shape.

**2. `format!` error strings replaced by typed variants.**
- `IndexConflictReason::{SpecMismatch, EmptyKeys}` carries the conflict cause.
- `DescriptorOverclaims { missing: Vec<Capability>, txn_shortfall: bool }` replaces the comma-joined-string failure message from `CompositeRootFilesystem::mount_dyn`.
- `SerializeIndexed` / `DeserializeIndexed` replace the indexed-projection ser/de format strings.
- `CorruptRecordVersion { path, raw }` replaces `version_raw.max(0) as u64` silent masking (P1 reviewer fix — negative versions now surface instead of pretending to be fresh).
- `IndexSpecMissingAfterUpsert` replaces the "index spec disappeared" sentinel string.

**3. Legacy bytes ops commented as deprecated.** `read_file` / `write_file` / `append_file` / `create_dir_all` carry a **DEPRECATED** doc-comment header pointing to their unified replacements. Doc-only (no `#[deprecated]` attribute) to avoid generating compiler warnings at every downstream call site during the migration window. Removed in the consumer-migration cleanup pass.

**4. Helper consolidation.** `sql_index_name`, `escape_like_literal`, and `escape_like_with_trailing_wildcard` (renamed from `escape_like_pattern` per reviewer P2 to make the trailing-wildcard contract explicit) now live in `db.rs` and are shared between libsql + postgres. Postgres copies deleted. `index_value_variant`'s hand-coded u8 mapping gone — Filter::Range cross-variant rejection uses `std::mem::discriminant` directly.

**5. Postgres test coverage (the big one from the review).** New `postgres_tests` module in `db_root_filesystem_contract.rs` adds **11 behavioral tests** that mirror the libsql contract:
- put/get round-trip with record metadata
- CAS Absent / Version / Any
- get returns None for missing path
- ensure_index idempotence + conflict detection
- query with Filter::Eq + And composition
- **Filter::Range on I64 is numeric, not lexicographic** (the specific reviewer-flagged bug: `BETWEEN '2' AND '10'` on text would miss `9`)
- PrefixOn literal-`%` is escaped
- All / empty Or / empty And semantics
- write_file-after-put resets record metadata + bumps version

Tests gracefully skip when no Postgres is reachable via `IRONCLAW_FILESYSTEM_POSTGRES_URL` or `DATABASE_URL` (same pattern as `ironclaw_outbound`). In integration-tier CI they actually execute.

**Quality gates:** 83 filesystem tests pass (25 unit + 5 catalog + 33 db-backed including 11 new Postgres + 20 filesystem-contract). Clippy clean on `--all-features`. `cargo check --workspace --all-features` clean. Pre-push lib gate (4967 tests) passes.

Downstream PRs (#3666 / #3670 / #3671 / #3672) rebased cleanly onto the new foundation — no consumer code referenced `BackendCapabilities` fields directly so the refactor was transparent.

serrrfirat (changes requested):
Found three issues that should be addressed before merging.

1. **[P1] Query results lose the record path**

`VersionedEntry` carries only `entry` and `version`, and both SQL `query` implementations read each row path but discard it before returning. A caller that queries for claimable records gets a CAS version but no path to pass back into `put(..., CasExpectation::Version)`, so the query API cannot safely drive update/delete workflows unless every consumer duplicates the path inside the body or indexed projection. That makes the universal CAS/query surface incomplete for its main multi-record workflow.

Suggested fix: include the `VirtualPath` in the query result shape, or split `get` and `query` result types so query returns path + entry + version.

Refs: `crates/ironclaw_filesystem/src/record.rs`, `crates/ironclaw_filesystem/src/libsql.rs`, `crates/ironclaw_filesystem/src/postgres.rs`.

2. **[P2] `ScopedFilesystem::begin` can bypass per-operation permissions once transactions exist**

`ScopedFilesystem::begin` performs only a `BeginTxn` permission check, and `operation_allowed` maps that to `write`. The returned raw `StorageTxn` exposes `get`, `put`, and `delete` on `VirtualPath`, so the first backend that implements transactions will let a write-only scoped caller read or delete through the transaction handle without the normal `read`/`delete` checks.

Suggested fix: return a scoped transaction adapter that checks permissions per txn operation, or require `begin` permission to cover every operation the handle can perform.

Ref: `crates/ironclaw_filesystem/src/scoped.rs`.

3. **[P2] SQL `Range` filters are still not type-safe**

The in-memory backend now rejects cross-variant range comparisons, but the SQL backends still do not guard the stored JSON value type. Postgres casts every stored value for an `I64` range to `bigint`, so one text value under the same key can fail the whole query instead of just not matching. The non-I64 Postgres branch and libSQL branch compare extracted JSON values without requiring the stored value to share the bound variant.

Suggested fix: add JSON type guards before comparisons/casts and add regression tests with mixed stored variants under the same indexed key.

Refs: `crates/ironclaw_filesystem/src/postgres.rs`, `crates/ironclaw_filesystem/src/libsql.rs`.

Verification note: I attempted a targeted Cargo test run locally, but Cargo failed immediately with `No space left on device` while creating a target temp path, so I could not run tests in this checkout.

ilblackdragon:
Three CHANGES_REQUESTED items addressed in \`52af7763c\` (plus \`36844d1ff\` for the prior bitmask test-compile bug that was also blocking CI).

| Finding | Severity | Status | Notes |
|---|---|---|---|
| Query results lose record path | P1 | **Fixed** | \`VersionedEntry\` now carries \`path: VirtualPath\`; libsql + postgres + in-memory + local backends all populate it on \`get\` and \`query\`. Consumers driving \`put\`/\`delete\` off query results no longer need to duplicate the path into \`entry.indexed\`. Regression test: \`get_and_query_populate_versioned_entry_path\` in \`in_memory.rs\`. |
| \`ScopedFilesystem::begin\` bypasses per-op permissions | P2 | **Fixed** | \`begin\` now returns a \`ScopedStorageTxn\` wrapper that snapshots the \`MountPermissions\` from the authorizing grant and re-checks every \`put\` / \`get\` / \`delete\` through \`operation_allowed\`. \`commit\` / \`rollback\` are intentionally not re-checked (they were authorized at \`begin\` time). |
| SQL Range filters not type-safe | P2 | **Fixed** | Both backends now guard the BETWEEN with a JSON-type check (\`json_type(indexed, '$.{k}') = '{variant}'\` on libsql; \`jsonb_typeof(indexed->'{k}') = '{variant}'\` on postgres) matching the bound's variant. The Postgres I64 path's cast no longer fires against text-typed stored values; the non-I64 branches no longer match unrelated variants by accident. Regression test: \`range_filter_rejects_cross_variant_stored_values\`. |

CI was also failing on \`Clippy (all-features)\` — that was the \`reborn_native_filesystem_vertical_integration\` test in \`ironclaw_memory\` still constructing \`BackendCapabilities\` via the pre-bitmask struct literal. Ported it to \`BackendCapabilities::empty().with(Capability::*)\` in commit \`36844d1ff\`. CI re-running.

Stacked PR #3679 (all consumer migrations applying #3659) will be rebased onto this tip — the \`path\` field addition to \`VersionedEntry\` is additive for consumers (they only read \`.entry\` and \`.version\`), so the rebase should be conflict-free.

PR #3660, opened 2026-05-14
feat(filesystem): port libsql + postgres backends to native put/get with CAS

ilblackdragon:
## Summary

Stacked on top of #3659. Adds native `put`/`get` implementations to both SQL backends so consumer-store migrations get real persistent CAS + versioning + record metadata, not just the legacy-bytes pass-through that the foundation PR shipped via default impls.

## What's in this PR

**Schema (single migration V28):**
- Adds `content_type`/`kind`/`indexed`/`version` columns to `root_filesystem_entries`.
- Existing rows backfill to `version = 0`, `content_type = 'application/octet-stream'`, `kind = NULL`, `indexed = '{}'` so legacy `read_file`/`write_file` continue to round-trip untouched.
- libsql gets an idempotent `ALTER TABLE` migration (gated by `pragma_table_info` like the existing `is_dir` migration). Postgres goes through `migrations/V28__root_filesystem_records.sql` with matching shape (`JSONB` instead of `TEXT`, `BIGINT` instead of `INTEGER`). Checksums lock entry added.

**Backend impls:**
- `put` handles all three CAS modes:
  - `Absent` → `INSERT ... ON CONFLICT DO NOTHING`; on conflict, read current version and return `VersionMismatch`.
  - `Version(v)` → `UPDATE ... WHERE path = ? AND version = ?`; on zero rows, read current and return `VersionMismatch`.
  - `Any` → upsert with `version = version + 1`. Postgres uses `RETURNING version`; libsql does a follow-up read.
- `get` reconstructs `Entry` from the row, validating `RecordKind`/`ContentType` and deserializing `indexed`. Directory rows return `None`.
- `capabilities()` declares `records: true` and `txn: Cas`. `query`/`index`/`events` stay `false` — those Unsupported ops land in a follow-up port once consumer migrations need filtered queries.

**Dependencies:**
- `tokio-postgres` gains the `with-serde_json-1` feature for `serde_json::Value` ↔ `JSONB` conversion.

## Test plan

- [x] `cargo test -p ironclaw_filesystem --features libsql` — 13 contract tests pass (8 existing + 5 new CAS-specific: put/get round-trip with record metadata, Absent rejection on existing path, Version advancement + stale rejection, Any-mode version increment, get-on-missing returns None)
- [x] `cargo test -p ironclaw_filesystem --all-features` — 60 tests total pass (21 unit + 5 catalog + 14 db-backed + 20 filesystem-contract). All foundation-PR tests still green.
- [x] Pre-push test suite (4967 tests) — pass; migration checksum lock entry added
- [ ] Reviewer: confirm V28's column defaults are safe for the existing rows in any staging/prod database that has shipped V27 already.

## Stacked PR

This PR's base is `reborn/universal-fs-dispatch` (the foundation PR #3659), not `reborn-integration`. Rebase / change base once the foundation merges.

### DIFF
--- a/crates/ironclaw_filesystem/src/libsql.rs
+++ b/crates/ironclaw_filesystem/src/libsql.rs
@@ -318,14 +318,27 @@ impl RootFilesystem for LibSqlRootFilesystem {
             return Err(directory_write_error(path.clone()));
         }
         let conn = self.connect().await?;
+        // PR #3660 reviewer fix: legacy write_file must also reset the
+        // record metadata (content_type / kind / indexed) and bump the
+        // version, otherwise a get() after a write_file-overwrite of a
+        // previously record-shaped entry returns stale metadata. Treat
+        // legacy writes as opaque-file entries: kind=NULL, indexed='{}',
+        // content_type=application/octet-stream, version bumped from the
+        // current row's version (or 1 for new entries).
         let rows = conn
             .execute(
                 r#"
-                INSERT INTO root_filesystem_entries (path, contents, is_dir, updated_at)
-                VALUES (?1, ?2, 0, strftime('%Y-%m-%dT%H:%M:%fZ', 'now'))
+                INSERT INTO root_filesystem_entries
+                    (path, contents, is_dir, content_type, kind, indexed, version, updated_at)
+                VALUES (?1, ?2, 0, 'application/octet-stream', NULL, '{}', 1,
+                        strftime('%Y-%m-%dT%H:%M:%fZ', 'now'))
                 ON CONFLICT (path) DO UPDATE SET
                     contents = excluded.contents,
                     is_dir = 0,
+                    content_type = excluded.content_type,
+                    kind = excluded.kind,
+                    indexed = excluded.indexed,
+                    version = root_filesystem_entries.version + 1,
                     updated_at = excluded.updated_at
                 WHERE root_filesystem_entries.is_dir = 0
                 "#,
@@ -350,16 +363,27 @@ impl RootFilesystem for LibSqlRootFilesystem {
             return Err(directory_append_error(path.clone()));
         }
         let conn = self.connect().await?;
-        // TODO(reborn): append rewrites the whole DB row. Do not use this path
-        // for high-volume JSONL/event streams; route those through typed event
-        // stores or append-capable artifact backends instead.
+        // PR #3660 reviewer fix: same metadata-reset concern as write_file.
+        // Append also resets kind/indexed/content_type to opaque-file
+        // defaults — appending bytes onto a previously record-shaped
+        // entry was always a category error, and we surface that by
+        // clearing the schema metadata rather than leaving it stale.
+        // TODO(reborn): append rewrites the whole DB row. Do not use this
+        // path for high-volume JSONL/event streams; route those through
+        // typed event stores or append-capable artifact backends.
         conn.execute(
             r#"
-            INSERT INTO root_filesystem_entries (path, contents, is_dir, updated_at)
-            VALUES (?1, ?2, 0, strftime('%Y-%m-%dT%H:%M:%fZ', 'now'))
+            INSERT INTO root_filesystem_entries
+                (path, contents, is_dir, content_type, kind, indexed, version, updated_at)
+            VALUES (?1, ?2, 0, 'application/octet-stream', NULL, '{}', 1,
+                    strftime('%Y-%m-%dT%H:%M:%fZ', 'now'))
             ON CONFLICT (path) DO UPDATE SET
                 contents = CAST(root_filesystem_entries.contents || excluded.contents AS BLOB),
                 is_dir = 0,
+                content_type = excluded.content_type,
+                kind = excluded.kind,
+                indexed = excluded.indexed,
+                version = root_filesystem_entries.version + 1,
                 updated_at = excluded.updated_at
             "#,
             libsql::params![path.as_str(), libsql::Value::Blob(bytes.to_vec())],
--- a/crates/ironclaw_filesystem/src/libsql.rs
+++ b/crates/ironclaw_filesystem/src/libsql.rs
@@ -1,3 +1,4 @@
+use std::collections::BTreeMap;
 use std::sync::Arc;
 
 use async_trait::async_trait;
@@ -8,7 +9,11 @@ use crate::db::{
     is_not_found, libsql_db_error, not_found, system_time_from_unix_seconds, valid_engine_path,
     virtual_path_prefixes,
 };
-use crate::{DirEntry, FileStat, FileType, FilesystemError, FilesystemOperation, RootFilesystem};
+use crate::{
+    BackendCapabilities, CasExpectation, ContentType, DirEntry, Entry, FileStat, FileType,
+    FilesystemError, FilesystemOperation, IndexCapability, IndexKey, IndexValue, RecordKind,
+    RecordVersion, RootFilesystem, TxnCapability, VersionedEntry,
+};
 
 #[cfg(feature = "libsql")]
 /// libSQL-backed [`RootFilesystem`] storing file contents by virtual path.
@@ -34,6 +39,7 @@ impl LibSqlRootFilesystem {
                 )
             })?;
         ensure_libsql_root_is_dir_column(&conn).await?;
+        ensure_libsql_records_columns(&conn).await?;
         Ok(())
     }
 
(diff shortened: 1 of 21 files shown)


serrrfirat:
## Summary
Reviewed PR #3660 only. Base `04224e111fbd8a5c1a9c9e39642f8b592bfc2202` → head `63df0a08a9a77dde7de03facbab4446b413b1f62`.

PR ports libSQL/Postgres filesystem backends to native `put/get` with CAS. Merge stance: block; DB CAS contract is not sound across legacy byte APIs and libSQL return-token path.

## Findings

| 1 | High | Correctness / CAS / Lost updates | `crates/ironclaw_filesystem/src/libsql.rs:332`, `crates/ironclaw_filesystem/src/libsql.rs:344`, `crates/ironclaw_filesystem/src/postgres.rs:268`, `crates/ironclaw_filesystem/src/postgres.rs:298` | Legacy `write_file` / `append_file` still mutate row contents without bumping `version` and without clearing/resetting record metadata like `content_type`, `kind`, and `indexed`. After `put()` creates a record-backed row, byte writes can change bytes while leaving the old CAS token and old indexed metadata in place. A later `put(..., CasExpectation::Version(old))` can succeed over unseen byte writes, and `get()` can return stale record metadata for different bytes. | Make all byte mutators participate in the same invariant: either bump `version` and reset metadata to opaque-file defaults, or fail closed when row has record metadata. Add mixed-path tests: `put` → `write_file/append_file` → stale `Version(old)` must not succeed, and metadata must not lie. |
| 2 | High | Correctness / Backend parity / Isolation | `crates/ironclaw_filesystem/src/libsql.rs:190`, `crates/ironclaw_filesystem/src/libsql.rs:223`, `crates/ironclaw_filesystem/src/postgres.rs:175`, `crates/ironclaw_filesystem/src/postgres.rs:190` | libSQL `CasExpectation::Any` performs the upsert, then calls `current_version(path)` as a separate lookup. A concurrent writer between those operations can advance the version, causing this caller to receive the token for someone else's later write. Postgres uses `RETURNING version` in the same statement, so backend semantics diverge and libSQL follow-up `Version(token)` writes become unsound. | Return the libSQL version from the same statement/transaction as the upsert (`RETURNING version` if supported, otherwise explicit transaction + locked/read-your-write semantics). Add concurrency/parity test that returned token belongs to this mutation. |
| 3 | Medium | Tests / Backend parity | `crates/ironclaw_filesystem/tests/db_root_filesystem_contract.rs:225` | PR adds DB-backed CAS behavior for both libSQL and Postgres, but behavioral tests are libSQL-only; Postgres has only a compile-time trait assertion. Given this PR already has backend-specific SQL and divergence in `Any`, missing Postgres caller-boundary tests leave parity regressions unguarded. | Add Postgres integration tests for `put/get`, `Absent`, `Version`, `Any`, and mixed `put` with legacy byte writes/appends. |

## Security/data-flow notes
- Path/tenant isolation not flagged in reviewed changed scope.
- Main risk is storage correctness: stale CAS tokens, stale metadata, and backend divergence can corrupt higher-level filesystem assumptions.

- `libsql.rs:198-203`: `Any` upsert bumps version, but returned token comes from later `current_version` lookup.
- `postgres.rs:175-190`: Postgres returns version from same statement.
- Legacy byte APIs share same table with record-aware `put/get`, so they must share version/metadata invariants too.

## Missing tests
- Mixed `put` / `write_file` / `append_file` CAS invariant tests.
- libSQL concurrent `Any` returned-token ownership test.
- Postgres parity tests for every CAS mode and mixed path.

crates/ironclaw_filesystem/src/libsql.rs
```diff
@@ -58,6 +64,221 @@ impl LibSqlRootFilesystem {
...
+            .get(2)
+            .map_err(|error| libsql_db_error(path.clone(), FilesystemOperation::ReadFile, error))?;
+        let kind_raw: Option<String> = row.get(3).ok();
+        let indexed_raw: String = row
+            .get(4)
+            .map_err(|error| libsql_db_error(path.clone(), FilesystemOperation::ReadFile, error))?;
+        let version_raw: i64 = row
+            .get(5)
+            .map_err(|error| libsql_db_error(path.clone(), FilesystemOperation::ReadFile, error))?;
+        let entry = build_entry(path, body, content_type_raw, kind_raw, indexed_raw)?;
```
serrrfirat, line 275: **High Severity**

Legacy byte writes leave stale record metadata/version behind.

Native `get()` now rehydrates `content_type`, `kind`, `indexed`, and `version`, but legacy `write_file()` / `append_file()` still only update `contents`, `is_dir`, and `updated_at` (same issue exists in the Postgres backend). Concrete failure mode: `put(record)` writes kind/indexed/version 1, then `write_file()` overwrites the bytes while leaving old metadata and version. A later `get()` returns the new body with stale record metadata and stale version, so `put(..., CasExpectation::Version(v1))` can overwrite the byte-plane update without detecting the intervening write.

Suggestion: make legacy byte writes reset metadata to the opaque-file shape (`content_type = 'application/octet-stream'`, `kind = NULL`, `indexed = '{}'`) and bump `version`, or route legacy byte writes through native `put(Entry::bytes(...), CasExpectation::Any)`.

crates/ironclaw_filesystem/src/libsql.rs
```diff
@@ -58,6 +64,221 @@ impl LibSqlRootFilesystem {
...
+                    )
+                    .await
+                    .map_err(|error| {
+                        libsql_db_error(path.clone(), FilesystemOperation::WriteFile, error)
+                    })?;
+                if rows == 0 {
+                    return Err(directory_write_error(path.clone()));
+                }
+                let version =
+                    self.current_version(path)
```
serrrfirat, line 223: **Medium Severity**

`CasExpectation::Any` can return the wrong version under concurrency.

The libSQL path does the upsert, then calls `self.current_version(path)` on a separate connection. If another writer updates the same path between those two operations, this caller can return the other writer's version even though this caller's body is no longer the stored body. That breaks the contract that `put` returns the version assigned to this successful write.

Suggestion: use `RETURNING version` on the upsert if supported, or wrap the write + version read in a transaction/lock that prevents interleaving.

crates/ironclaw_filesystem/tests/db_root_filesystem_contract.rs
```diff
@@ -241,6 +244,103 @@ impl std::ops::Deref for TestLibSqlRootFilesystem {
     }
 }
 
+#[cfg(feature = "libsql")]
+#[tokio::test]
+async fn libsql_native_put_get_round_trip_with_record_metadata() {
```
serrrfirat, line 249: **Medium Severity**

Postgres native CAS/get behavior is untested.

This PR adds substantial native Postgres logic for metadata round-trip, JSONB conversion, `RETURNING version`, Absent rejection, and stale Version rejection, but the new behavioral tests are all `#[cfg(feature = "libsql")]`. Postgres only has a compile-time trait assertion above, so backend-specific regressions would ship green.

Suggestion: make these new CAS tests shared over both SQL backends (or duplicate them under `#[cfg(feature = "postgres")]`) covering put/get metadata, Absent rejection, Version advancement/stale rejection, Any increment, and missing get.

ilblackdragon:
Pushed fix for the High Severity stale-metadata regression flagged by the reviewer.

**Problem:** legacy `write_file` / `append_file` updated only `contents` / `is_dir` / `updated_at`, leaving `content_type` / `kind` / `indexed` / `version` stale from any prior `put`. A subsequent `get()` then returned a VersionedEntry whose metadata didn't match the bytes.

**Fix:** both legacy ops now reset `content_type` to `application/octet-stream`, clear `kind` to NULL, clear `indexed` to empty, and bump `version` on every write. Appending bytes onto a previously record-shaped entry was always a category error; this makes that surface as cleared schema metadata rather than stale metadata on top of changed bytes.

Regression test: `libsql_write_file_after_put_resets_record_metadata_and_bumps_version` covers the put-then-write_file sequence, asserts kind/indexed clearing and version advancement.

Other reviewer asks deferred to follow-ups so this PR stays scoped to the stale-metadata correctness fix:
- libsql `CasExpectation::Any` second-connection race → use `RETURNING version` (medium)
- libsql `row.get(3).ok()` silently masking errors → `row.get::<Option<String>>(3)?` (medium)
- Postgres native CAS/get behavior untested (libsql-only coverage) → integration-tier test pass to add
- Absent-vs-directory error precision (medium)

ilblackdragon:
Closing as subsumed by #3659. The libsql + Postgres native put/get/CAS work landed in the consolidated foundation (commit `b3c645691` on `reborn/universal-fs-dispatch`), plus the stale-metadata fix (commit `c1c65f50d`) addressing the High Severity reviewer flag.

PR #3661, opened 2026-05-14
feat(filesystem): native query + ensure_index on libsql + postgres

ilblackdragon:
## Summary

Stacked on top of #3660. Flips `capabilities { query: true, index.exact: true, index.prefix: true }` on both SQL backends so consumer stores can filter records by their declared indexed projection.

After this lands, the record/index/CAS triad is complete and 6 consumer crates (`ironclaw_secrets`, `ironclaw_authorization`, `ironclaw_processes`, `ironclaw_run_state`, `ironclaw_outbound`, `ironclaw_conversations`) can drop their own backend matrices and migrate to `ScopedFilesystem`. The other 4 consumers (`ironclaw_reborn_event_store`, `ironclaw_memory`, `ironclaw_engine` `Store`, `src/db/`) need follow-up capability work — append/tail plane and FTS/vector indexes — before they can migrate.

## What's in this PR

**Serde change:**
- `IndexValue` is now `#[serde(untagged)]` so SQL backends can run native predicates against the stored JSON. Previously the tagged shape (`{"type":"text","value":"acme"}`) would have required `indexed->'scope'->>'value' = ?`; now it's `indexed->>'scope' = ?` directly. Variant order is `Bool` → `I64` → `Text` → `Bytes` so JSON type detection on deserialize is deterministic.

**ensure_index:**
- Materializes expression indexes on `json_extract(indexed, '$.<key>')` (libsql) or `(indexed->>'<key>')` (postgres) for the declared keys.
- Spec metadata stored in a new `root_filesystem_index_specs` table (postgres: `JSONB` keys; libsql: text-JSON keys). Re-declaring the same spec is idempotent; conflicting specs (same name, different keys/kind) fail with `FilesystemError::IndexConflict`.
- `IndexKind::Fts` and `IndexKind::Vector` return `Unsupported` — separate port.

**query:**
- Translates `Filter::{Eq, PrefixOn, Range, And, Or}` into SQL WHERE-clause fragments with parameterized values.
- `IndexValue::Text` / `I64` / `Bool` supported as filter values. `IndexValue::Bytes` returns `Unsupported` (JSON representation makes byte comparison fragile).
- Postgres binds all variants as text since `->>'key'` returns text regardless of the underlying JSON type, keeping the expression index usable.
- Pagination via `Page::{offset, limit}` on `ORDER BY path`.

**Capabilities now declare:**

| Backend | records | query | index.exact | index.prefix | txn |
|---|---|---|---|---|---|
| `LibSqlRootFilesystem` | ✅ | ✅ | ✅ | ✅ | Cas |
| `PostgresRootFilesystem` | ✅ | ✅ | ✅ | ✅ | Cas |
| `InMemoryBackend` (unchanged) | ✅ | ✅ | ✅ | ✅ | Cas |

`events`, `index.fts`, and `index.vector` stay `false` — those land in follow-up ports.

## Migrations

- `migrations/V29__root_filesystem_index_specs.sql` (postgres, refinery-locked, checksum auto-regenerated).
- libsql gets an idempotent `CREATE TABLE IF NOT EXISTS` via `LIBSQL_INDEX_SPECS_SCHEMA` in `run_migrations`.

## Test plan

- [x] `cargo test -p ironclaw_filesystem --features libsql --test db_root_filesystem_contract` — **18 tests pass** (8 original + 5 CAS + 5 new indexer)
- [x] `cargo test -p ironclaw_filesystem --all-features` — **65 tests pass** (21 unit + 5 catalog + 19 db-backed + 20 filesystem-contract)
- [x] Pre-push test suite (4967 tests) — pass
- [ ] Reviewer: confirm the IndexValue serde change is safe — no prior IndexValue payload has been persisted (foundation + port PRs both ship with the tagged form unused on the wire).

## Stacked PR

This PR's base is `reborn/fs-port-sql-backends` (PR #3660), which is itself stacked on `reborn/universal-fs-dispatch` (PR #3659). The final merge order is foundation → port → indexer → consumer migrations.

### DIFF
--- a/crates/ironclaw_filesystem/src/index.rs
+++ b/crates/ironclaw_filesystem/src/index.rs
@@ -43,6 +43,34 @@ pub(crate) fn validate_simple_identifier(kind: &'static str, s: &str) -> Result<
             reason: "must be 128 characters or fewer".to_string(),
         });
     }
+    // Tightened identifier shape after PR #3661 reviewer flag:
+    //   `IndexKey::new("a.b")` used to pass validation but interact with
+    //   `json_extract(indexed, '$.a.b')` as a nested-path traversal rather
+    //   than the literal key `"a.b"`. Similarly, raw names used as DDL
+    //   identifiers without SQL quoting allowed `-` / `.` / unicode through.
+    //   Restrict to `[A-Za-z_][A-Za-z0-9_]*` so the same value is safe as a
+    //   JSON path component, a SQL identifier, and a row key.
+    let bytes = s.as_bytes();
+    let first = bytes[0];
+    if !(first.is_ascii_alphabetic() || first == b'_') {
+        return Err(HostApiError::InvalidId {
+            kind,
+            value: s.to_string(),
+            reason: "must start with an ASCII letter or underscore".to_string(),
+        });
+    }
+    if bytes[1..]
+        .iter()
+        .any(|b| !(b.is_ascii_alphanumeric() || *b == b'_'))
+    {
+        return Err(HostApiError::InvalidId {
+            kind,
+            value: s.to_string(),
+            reason: "must contain only ASCII letters, digits, and underscores".to_string(),
+        });
+    }
+    // Legacy traversal/whitespace checks retained as belt-and-suspenders;
+    // the ASCII alphanumeric rule above already rejects them.
     if s.contains('/')
         || s.contains('\\')
         || s.contains('\0')
@@ -282,6 +310,22 @@ mod tests {
         assert!(IndexName::new("scope/leases").is_err());
         assert!(IndexName::new("with space").is_err());
         assert!(IndexName::new("ok_name_1").is_ok());
+    }
+
+    #[test]
+    fn index_key_rejects_chars_that_break_sql_or_json_paths() {
+        // Reviewer (PR #3661) flagged that allowing `.` lets
+        // `json_extract(indexed, '$.a.b')` traverse rather than match the
+        // literal key `"a.b"`, and that other punctuation can break DDL.
+        // After tightening, IndexKey/Name accept `[A-Za-z_][A-Za-z0-9_]*`
+        // only.
+        assert!(IndexKey::new("a.b").is_err());
+        assert!(IndexKey::new("a-b").is_err());
+        assert!(IndexKey::new("1abc").is_err()); // can't start with digit
+        assert!(IndexKey::new("").is_err());
+        assert!(IndexKey::new("scope").is_ok());
+        assert!(IndexKey::new("_internal").is_ok());
+        assert!(IndexKey::new("scope_v2").is_ok());
     }
 
     #[test]
--- a/crates/ironclaw_filesystem/src/index.rs
+++ b/crates/ironclaw_filesystem/src/index.rs
@@ -144,12 +144,18 @@ impl From<IndexKey> for String {
 /// Variants are intentionally narrow — backends translate to their native
 /// column type. New variants require coordinated backend updates and a wire
 /// migration; do not extend casually.
+///
+/// Serialization is untagged so SQL backends storing the indexed map as
+/// JSON can run native predicates against it (`indexed->>'scope' = 'acme'`
+/// in Postgres, `json_extract(indexed, '$.scope') = 'acme'` in libSQL).
+/// Bool is listed first so JSON booleans don't accidentally match `I64`,
+/// and `Bytes` is last because a JSON array could otherwise be mis-typed.
 #[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize, Deserialize)]
-#[serde(rename_all = "snake_case", tag = "type", content = "value")]
+#[serde(untagged)]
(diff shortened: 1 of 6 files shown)


serrrfirat (changes requested):
Thanks for the detailed stack notes. I found a few correctness issues that look blocking before this can safely advertise native query/index support on the SQL backends:

1. **Postgres `Range` on `I64` is lexicographic, not numeric**

   In `crates/ironclaw_filesystem/src/postgres.rs`, `Filter::Range` is translated as `indexed->>'key' BETWEEN ...`, and `IndexValue::I64` is bound as `n.to_string()`. Because `->>` returns text, ranges compare lexicographically. For example, a `2..10` range can miss numeric values because `"10" < "2"` as text.

   Suggested fix: when both range bounds are `IndexValue::I64`, cast the extracted JSON text and bound params to `bigint` (or otherwise use a numeric JSON expression), and add a Postgres caller-level query test for values such as `2`, `9`, `10`, `11`.

2. **Nested `Filter::All` / empty `Or` changes query semantics**

   Both SQL translators skip children whose translated fragment is empty. That means `Or([])` returns all rows instead of none, and cases like `And([All])` / `Or([All])` can either emit `AND ()` or narrow incorrectly depending on nesting.

   Suggested fix: preserve boolean identities explicitly in compound translation. `All` should be `TRUE`, empty `And` should be `TRUE`, and empty `Or` should be `FALSE` (matching the in-memory `all`/`any` behavior). Add SQL backend query tests for these shapes.

3. **Prefix filters treat a trailing `%` in user data as a wildcard**

   `escape_like_pattern` intentionally leaves the final `%` unescaped so path-prefix patterns can append a wildcard. The same helper is reused for `Filter::PrefixOn` values before appending the wildcard, so a literal prefix like `tenant:%` becomes a broad `LIKE 'tenant:%%'` match.

   Suggested fix: split this into two helpers: one for literal user input that always escapes `%`, `_`, and `!`, then append the backend wildcard separately; and one for already-constructed path patterns if needed. Add a test where `tenant:%` does not match `tenant:abc`.

4. **Valid `IndexKey` / `IndexName` values can break SQL or query the wrong JSON path**

   `IndexKey` / `IndexName` validation allows punctuation such as `.` and `-`, but SQL and JSON path construction assumes a narrower identifier shape. For libSQL, `IndexKey::new("a.b")` is valid but `json_extract(indexed, '$.a.b')` reads nested JSON rather than the literal indexed key `"a.b"`. Raw index names also flow into DDL identifiers without SQL identifier quoting.

   Suggested fix: either tighten the identifier validation contract to SQL-safe names (for example ASCII alnum plus `_`, with a leading alpha/underscore rule if needed), or correctly quote/escape SQL identifiers and JSON path segments in both backends. The simpler fix may be to tighten validation since the docs already call these “simple identifiers.”

5. **`ensure_index` is not race-idempotent**

   Both SQL backends do `SELECT` then `INSERT` for the index spec row. If two processes declare the same spec concurrently, both can miss the row, then one receives a unique-constraint backend error instead of the promised idempotent success.

   Suggested fix: make the declaration atomic with `INSERT ... ON CONFLICT` plus a re-read/conflict check, ideally inside the same transaction/connection used for DDL. Add a concurrency/idempotence regression test where practical.

I did not find this merge-ready as-is because the new capabilities expose these paths as supported (`query`, `index.exact`, `index.prefix`) and consumers will depend on the advertised semantics.

ilblackdragon:
Pushed fixes for all five CHANGES_REQUESTED items (see commit on the branch). Summary:

1. **IndexKey/IndexName validation tightened** to `[A-Za-z_][A-Za-z0-9_]*`. Rejects `IndexKey::new("a.b")` so `json_extract(indexed, '$.a.b')` can't traverse JSON nesting, and closes the DDL identifier injection surface for `-` / `.` characters in raw index names.

2. **Filter::All / empty And / empty Or** now emit explicit boolean identities (TRUE / TRUE / FALSE). The "skip empty fragment" path that conflated `Filter::All` with the compound identity is gone.

3. **Filter::PrefixOn** user input now escaped through a dedicated `escape_like_literal` (every `%`/`_`/`!` escaped). Path-pattern escaping kept as a separate helper for the prefix scan case.

4. **Postgres Filter::Range on IndexValue::I64** casts both the extracted JSON and bound params to `BIGINT` so the comparison is numeric. Mixed-variant ranges still hit text comparison (matches in-memory backend semantics for cross-variant bounds).

5. **ensure_index** is now race-idempotent on both backends via `INSERT ... ON CONFLICT (prefix, name) DO NOTHING` plus canonical-spec read-back. Two concurrent declarers of the same spec both succeed; conflicting specs see `IndexConflict` deterministically.

3 new regression tests cover #1 (IndexKey rejection), #2 (Or/And/All semantics), and #3 (literal-percent prefix non-wildcarding). 68 filesystem tests pass total.

ilblackdragon:
Closing as subsumed by #3659. The native query + ensure_index + all 5 CHANGES_REQUESTED review items landed in the consolidated foundation (commits `78abe1585` + `14e025d73` on `reborn/universal-fs-dispatch`).

Issue #3604, opened 2026-05-14
TrustClass placeholder semantics across manifest v2 + composition


zmanian:
## Problem

`TrustClass` is being threaded through multiple Reborn-era crates as a real enforcement type, but its semantics are placeholder. Specifically:

- #3591 (Extension Manifest v2): `FirstPartyRequested` and `SystemRequested` both map to `TrustClass::Sandbox` after resolution. The mapping is undocumented and unasserted by tests.
- #3566 (Reborn composition): `HostTrustPolicy::empty()` is hardcoded; every trust check passes vacuously (see issue #2 in this batch).

## Why this matters

44 downstream callers in manifest v2 slice 2b will migrate against this surface. Once they consume `manifest.trust`, the placeholder semantics become load-bearing:

- A `FirstParty` extension labeled as such in the manifest gets `Sandbox` trust at runtime, with no diagnostic that anything was downgraded.
- Operators reading the manifest see one trust level; the runtime enforces another.
- Future correct enforcement becomes a silent behavior change rather than a feature addition.

This is the kind of placeholder that hardens into "the way it is" if not tracked.

## Expected

Before slice 2b migration lands:

1. Document the resolution semantics on `ExtensionManifestV2::trust` (what does `FirstPartyRequested → Sandbox` mean? when does it actually become `FirstParty`?).
2. Add tests asserting the resolved `manifest.trust` value for each of `HostBundled` / `LocalUser` / installed-from-registry sources.
3. Decide and implement: does `TrustClass::FirstParty` exist as a runtime state, or is `Sandbox` the upper bound until further policy is added? Either is defensible; the current undocumented placeholder is not.

## References

- #3591 (merged) — manifest v2 introduces the field
- #3566 (open) — composition uses empty policy
- Re-review notes:
  - [#3591 comment](https://github.com/nearai/ironclaw/pull/3591#issuecomment-4445612336)
  - [#3566 comment](https://github.com/nearai/ironclaw/pull/3566#issuecomment-4445610740)

## Acceptance

- [ ] `manifest.trust` resolution rules documented (rustdoc + a contract doc under `docs/reborn/contracts/`)
- [ ] Test coverage asserts the resolved value for each manifest source
- [ ] Cross-PR consistency: composition gate, manifest parser, and runtime enforcement all agree on what each `TrustClass` variant means

BenKurrek:
[...]

PR #3663, opened 2026-05-14
feat(reborn): add extension host api manifest contracts

serrrfirat:
## Summary

- add Manifest v2 host API contract vocabulary (`HostApiId`, `ManifestSectionPath`, `HostApiRefV2`)
- add a composition-wired host API contract registry for validating explicit manifest sections
- fail closed for unknown host APIs, invalid section refs, duplicate single-instance contracts, and unreferenced operational sections
- document the unified extension + multi-hostAPI manifest shape

## Scope

Contract-only slice. This does not migrate product adapter manifests/catalogs, perform runtime handshakes, or publish domain read models.

## Verification

```bash
cargo fmt --check --package ironclaw_extensions
cargo test -p ironclaw_extensions --locked
cargo clippy -p ironclaw_extensions --all-targets --locked -- -D warnings
git diff --check
```


zmanian:
## Review: extension host api manifest contracts (#3663)

Contracts-only slice that lands the manifest v2 host-API vocabulary (`HostApiId`, `ManifestSectionPath`, `HostApiRefV2`), a composition-wired `HostApiContractRegistry`, and a second parse entry point (`parse_with_host_api_contracts`) that validates declared sections through the registry. No runtime, catalog, or domain handler is touched.

### Summary

The design is the right shape for substrate independence:

- The generic envelope (`ironclaw_extensions`) owns parsing, section dispatch, multiplicity enforcement, and unknown-host-API fail-closed. Domain crates own pattern matching, cardinality decisions, typed validation, and read-model projection. This keeps `ironclaw_extensions` free of product knowledge.
- Composition wires contracts; the crate ships none. Good for the "no substrate coupling" invariant.
- Builds cleanly on top of #3591 (manifest v2 envelope) and #3568 (host-port / capability-profile vocabulary). Does not modify either contract surface.

### Strengths

- **Fail-closed defaults are explicit and tested**: `UnknownHostApi`, `DuplicateHostApi` (for single-multiplicity contracts), `DuplicateHostApiSection`, `MissingHostApiSection`, `UnreferencedOperationalSection`, `HostApiSectionRejected`, and `DuplicateHostApiContractRegistration` all have dedicated tests. Coverage is genuinely broad for a contracts slice.
- **Mixed-mode lockdown**: `top-level capabilities are not allowed when host_api contracts are declared` (and the inverse legacy path requiring `parse_with_host_api_contracts`) cleanly prevents ambiguous manifests during cutover.
- **Section-ownership overlap detection**: `validate_host_api_refs` rejects bidirectional prefix overlaps between two `[[host_api]]` entries (`rejects_overlapping_host_api_section_ownership`). This is the right invariant — sections are owned, not shared.
- **`metadata` / `x` carveouts** for inert sections at the root are sensible and tested, while still rejecting unreferenced operational siblings.
- **`deny_unknown_fields`** preserved on `RawManifestV2`, `RawHostApiRefV2`, and existing shapes. The new whole-document scan (`first_non_envelope_top_level_key`, and the table-only walk during envelope extraction) closes the gap that `serde(flatten)`-style extension would otherwise open.
- **`HostApiId` and `ManifestSectionPath` are newtypes** with construction-time validation. Required `/vN` discriminator on the contract id is a good forcing function for forward versioning.

### Issues

1. **TrustClass placeholder semantics from issue #3604 are not addressed.** This PR does not regress them, but it does add a new product-adapter contract surface (`ironclaw.product_adapter/v1`) whose authority decisions will hinge on resolved trust. Once domain contracts start consuming `manifest.trust` to decide what a `product_adapter` is allowed to do at runtime, the `FirstPartyRequested → Sandbox` silent downgrade documented in #3604 becomes load-bearing. Two requests:
   - Add a rustdoc note on `ExtensionManifestV2::host_apis` (or a top-level `v2.rs` module doc) calling out that `descriptor_trust_default` is **not** the runtime trust and that host-API contract handlers must consult composition policy, not the manifest. Right now nothing prevents a future contract handler from reading `manifest.descriptor_trust_default` and treating it as authoritative.
   - Track #3604 as a blocker (or hard dependency) for the first domain contract PR that wires `product_adapter` through to authority grants. Doing the doc-only fixup in this PR is fine; the test/runtime side belongs with the contract handler.

2. **`HostApiContractRegistry` reference lifetimes constrain composition shape.** `&'a dyn HostApiManifestContract` requires every contract to live at least as long as the registry. In practice composition wires this once at startup, but several other registries in this codebase (host port catalog, etc.) take owned `Box<dyn>` or `Arc<dyn>` to avoid forcing `Box::leak` or static singletons. The test helper resorts to `Box::leak(Box::new(...))` precisely because of this. Consider:
   ```rust
   contracts: BTreeMap<HostApiId, Arc<dyn HostApiManifestContract + Send + Sync>>
   ```
   This costs one allocation per contract (fine — there will be ~5-10 of these) and removes the lifetime parameter on `HostApiContractRegistry`, simplifying composition and tests. `Send + Sync` also lets the registry be shared across async tasks once runtime loading lands. Not blocking, but easier to fix now than after consumers exist.

3. **`HostApiManifestContract::validate_section` returns `Result<(), String>`.** Stringly-typed errors lose structure for downstream classification (e.g., schema-ref errors vs cardinality errors vs unknown-field errors). Recommend a thin `HostApiContractError` enum (or at minimum a `#[non_exhaustive] struct { reason: String, kind: ErrorKind }`) so callers can act on classes without parsing strings. The reason field would still survive into `ManifestV2Error::HostApiSectionRejected { reason, .. }`. Not blocking.

4. **No `#[non_exhaustive]` on the new public types.** Consistent with the rest of `v2.rs`, so not a regression — but `HostApiMultiplicity`, `HostApiRefV2`, `HostApiId`, and `ManifestV2Error` are now part of the wire/API surface that downstream contract crates will match on. Adding `#[non_exhaustive]` to `HostApiMultiplicity` and `ManifestV2Error` (at least) before external consumers land is significantly cheaper than after. Recommend doing it now.

5. **`HostApiId` accepts a fairly permissive grammar.** `/v` substring check is the only structural requirement; `ironclaw.foo/vX` (where X is non-numeric) or `ironclaw.foo/v1/v2` both pass. Suggest tightening to a single trailing `/v<digits>` segment via a small parser. Low priority — domain handlers will reject in practice — but the contract id is meant to be the stable wire discriminator and ought to validate that strictly.

6. **`is_envelope_key` is hardcoded as a `matches!` block.** Adding a new envelope field (which has already happened twice in v2's short life) requires updating this list and the `RawManifestV2` struct in lockstep, and forgetting either side fails silently (a new envelope field would be treated as an operational section and probably get rejected as unreferenced — confusing error). Consider deriving the envelope key set from `RawManifestV2`'s field names via a const, or at minimum adding a debug_assert that every `RawManifestV2` field name is present in `is_envelope_key`. Low priority.

7. **`parse_with_host_api_contracts` allows zero `[[host_api]]` entries** (it falls through to legacy capability validation since `raw.host_api.is_empty()` is true). That's intentional for cutover, but the function name suggests the caller is opting into the new shape. Consider documenting on the function that a manifest with no `[[host_api]]` entries is accepted and treated as legacy — or rejecting outright to force the caller to pick a path. Cosmetic.

### Suggestions

- Add a doctest or example on `HostApiManifestContract` showing the minimal three-method impl. The trait is the main extension point for substrate consumers and will get implemented many times.
- The docs/reborn/contracts/extensions.md addition is good. Worth also adding a one-line note that domain handlers MUST NOT read `manifest.trust` or `manifest.descriptor_trust_default` for authority decisions, mirroring issue #3604.
- `ManifestSectionPath::segments` is `pub(crate)` only because it's used by `is_prefix_of`. If downstream contract handlers want to introspect section paths (likely — they'll want to do their own subpath dispatch), expose this as `pub`.

### Verdict

**APPROVE with non-blocking nits.** The contract design is sound, fail-closed defaults are well-tested, and the substrate independence story holds. The two things I'd most like to see addressed before this lands:

1. Rustdoc note on `host_apis` and the module that domain contracts must not treat manifest trust as authoritative (links to #3604).
2. Convert `&'a dyn` to `Arc<dyn ... + Send + Sync>` on the registry to avoid forcing `Box::leak` patterns in downstream wiring.

Everything else (non_exhaustive, strict HostApiId grammar, structured contract errors) is fine as a follow-up.

PR #3664, opened 2026-05-14
feat(product-workflow): add RebornServices facade for WebUI

### LINKED ISSUE #3612: [Reborn WebUI Beta] Add WebUI-facing RebornServices facade
Parent tracker: #3607
Module: M2 — Native Product Workflow / Inbound Facade
Priority: P0

## Problem

WebUI route authors need a stable high-level facade. Without it, handlers are likely to reach into TurnCoordinator, HostRuntime, stores, dispatcher, or runtime lanes directly.

## Goal

Expose a WebUI-facing `RebornServices` API for beta route handlers.

## Proposed surface

- `submit_turn(...)`
- `get_timeline(...)`
- `stream_events(...)`
- `cancel_run(...)`
- `resolve_gate(...)` if gates are in beta

## Acceptance criteria

- WebUI route handlers can implement beta route scope using only this facade and DTOs.
- Facade owns idempotency, scope/binding handoff, and calls into ProductWorkflow/equivalent + TurnCoordinator.
- Tests prove route crate does not depend directly on `CapabilityHost`, dispatcher, run-state stores, runtime lanes, or DB stores.
- Errors returned to WebUI are sanitized/stable.

## Related

Depends on #3026/#3566 composition direction and #3280 inbound facade.

serrrfirat:
## Summary
- add WebUI-facing RebornServices facade over thread, turn, and projection ports
- expose stable DTOs/errors for submit, timeline, event stream, cancel, and gate resolution
- add contract tests for idempotent submit replay, sanitized errors, 401-vs-403 error codes, JSON metadata serialization, and forbidden runtime dependency leakage

## Follow-up for consumer PR
- when `src/channels/web/` route handlers call `RebornServicesApi`, add handler-level tests for double-submission gating, sanitization, and the unauthenticated-vs-forbidden distinction; facade-only coverage is not sufficient for the route wire-up

## Review-fix verification
- `cargo fmt`
- `CARGO_TARGET_DIR=/Volumes/NVME/ironclaw-target cargo test -p ironclaw_product_workflow --test reborn_services_contract` — passed, 12 tests
- `cargo clippy --all --benches --tests --examples --all-features` — passed
- `CARGO_TARGET_DIR=/Volumes/NVME/ironclaw-target cargo test --lib` — failed in two existing extension-manager tests requiring a built `portfolio` WASM artifact; 4,965 passed, 2 failed, 7 ignored
- `bash scripts/pre-commit-safety.sh` — failed on branch-wide pre-existing warnings outside the Reborn service files

Closes #3612


zmanian:
## Summary

Adds a `RebornServices` facade in `crates/ironclaw_product_workflow/` exposing a stable WebUI-facing API (submit turn, timeline, stream events, cancel, resolve gate, create thread) over the thread, turn, and projection ports. Wire DTOs and a sanitized error enum live in dedicated submodules. Contract tests cover idempotent submit replay, idempotent create-thread replay, sanitized validation errors, gate approve/deny/credential paths, and — importantly — a source-level grep that asserts the facade does not reach into `ironclaw_capabilities`, `ironclaw_dispatcher`, `ironclaw_host_runtime`, `ironclaw_run_state`, `ironclaw_storage`, or `RuntimeLane`.

Pure substrate change. No `src/channels/web/` consumer is added; that's the next PR's job. +1307/-5.

## Strengths

- **Substrate independence is enforced, not asserted.** Cargo deps stay within `ironclaw_host_api`, `ironclaw_product_adapters`, `ironclaw_threads`, `ironclaw_turns`. The architecture boundary test (`reborn_crate_dependency_boundaries_hold`) and the in-crate source-grep test both lock this in. This is exactly the inversion-resistance the Reborn substrate posture requires.
- **Error mapping is explicit and centralized.** `map_thread_error`, `map_turn_error`, `map_adapter_error` translate backend variants into a small stable code + HTTP status surface, and the sanitization is verified by `validation_errors_are_stable_and_sanitized` (no `"backend"` or `"TurnCoordinator"` leakage) and `credential_gate_resolution_returns_sanitized_stable_error_until_gate_port_exists` (no `credential-alpha` leakage).
- **Idempotency is handled at the facade.** `submit_turn` replays through `replay_accepted_inbound_message` and, on prior `Submitted`, re-derives the run state from `TurnCoordinator::get_run_state` rather than re-submitting — verified by `duplicate_submit_replays_prior_handoff_without_second_submission`.
- **Bounded-ref sanitization.** `bounded_ref` rejects control chars and over-length raw values by collapsing to a UUIDv5 derived from the input — defense against caller-controlled binding refs.
- **DTOs are `Serialize + Deserialize`** with explicit `#[serde(rename_all = "snake_case")]` on the tagged enums (`outcome`, error `code`), consistent with the wire-stable enum rule.

## Issues

### Important

1. **Hand-rolled JSON escaper in `escape_json_string` (reborn_services.rs:714).** Used to build `metadata_json` for `create_thread`. It handles `\\`, `"`, `\n`, `\r`, `\t` but not other control chars (`U+0000..U+001F` minus the three handled), U+2028/U+2029, or surrogate concerns. Either add `serde_json` as a direct dependency and build a `serde_json::Value` to stringify, or — preferred — make `metadata_json` a structured type upstream so callers don't hand-roll JSON at this layer. The crate's `Cargo.toml` doesn't list `serde_json` directly even though tests use it via `json!`; the production code reaching for hand-rolled escaping is the smell.

2. **`Unauthorized` collapses 401 and 403.** `RebornServicesErrorCode::Unauthorized` maps to 403 when origin is `TurnError::Unauthorized` (line 663) but to 401 when origin is `ProductAdapterError::Authentication` (line 698). These are semantically distinct (unauthenticated vs forbidden) and a beta WebUI client will key behavior off the HTTP status. Split into `Unauthenticated` (401) and `Forbidden` (403) in `RebornServicesErrorCode`, or pick one status per code and route the other origin to the right code. As written, a client cannot reliably distinguish "log in again" from "you cannot do this".

3. **Caller-level test coverage will be required when the web layer lands.** Per `.claude/rules/testing.md` ("Test Through the Caller"), once a route handler in `src/channels/web/` calls `RebornServicesApi`, the gate against double-submission, sanitization, and the 401-vs-403 distinction needs a test at the handler level — not just at the facade. Worth a tracking note in the PR description for the consumer PR.

### Suggestions

4. **`tracing` is in `[dependencies]` but unused in the new module.** Either emit a span around `submit_turn` / `cancel_run` / `resolve_gate` (useful for the eventual debugging story) or leave it to the consumer crate; if the latter, the dep can stay because other modules use it.

5. **`internal_invariant` branches (lines 196, 399, 423)** match on enum variants that were just constructed by `into_command(...)?` — they are formally unreachable. Either keep as defensive guards (current style) or replace with `unreachable!("…")` to make intent explicit. Not blocking.

6. **`create_thread` idempotency is not symmetric with `submit_turn`.** When the caller supplies an explicit `requested_thread_id`, the `client_action_id` is dropped from idempotency consideration — the test only covers the *generated* id path (`duplicate_create_thread_replays_generated_thread_for_same_client_action`). If two calls supply the same `client_action_id` but different explicit `requested_thread_id` values, the facade silently creates two threads. Either document this contract on `create_thread` or persist the `client_action_id` → `thread_id` mapping. Low priority because `create_thread` is described as "optional" — submit_turn is the primary path.

7. **PR title says "webui" but no `src/channels/web/` files are touched.** The title sets the expectation that this is the consumer wire-up. Consider renaming to `feat(product_workflow): add RebornServices facade for WebUI` or similar — the diff is entirely substrate-side. (Also: when the consumer PR lands, the dispatch-routing rule in `CLAUDE.md` will apply — mutating handlers must go through `ToolDispatcher::dispatch()` or carry a `// dispatch-exempt:` annotation.)

8. **`turn_id: turn_id.to_string()`** in `RebornSubmitTurnResponse::Submitted` (line 318) downgrades a typed `TurnId` to `String` on the wire. Since `TurnRunId` is preserved as the typed `TurnRunId` newtype in the same struct, the inconsistency is mild but visible. Either both stringly-typed or both newtyped — preferably the latter.

## Verdict

Approve with minor changes. The facade design is sound, dependency direction is correct and enforced, sanitization is tested. The two items I'd ask for before merge are (a) the 401/403 split or a documented choice, and (b) replacing the hand-rolled JSON escape — both small. Everything else is non-blocking. Look forward to the consumer PR; the dispatcher-routing and caller-level-test rules will be the next gate.

serrrfirat:
Addressed the review feedback from @zmanian in `5613a96e`:

- Replaced the hand-rolled create-thread metadata JSON escaping with `serde_json` serialization and moved `serde_json` to production dependencies.
- Split the stable service error surface into `unauthenticated` (401) and `forbidden` (403), including adapter and turn error mappings.
- Added contract coverage for metadata JSON serialization and the 401/403 distinction.
- Documented the explicit `requested_thread_id` idempotency contract on `create_thread`.
- Renamed the PR title and added the caller-level route test tracking note for the follow-up WebUI consumer PR.

Verification details are in the updated PR body.

PR #3665, opened 2026-05-14
feat(engine): IRONCLAW_DISABLE_CODEACT flag to disable v2 CodeAct

zetyquickly:
## Summary

- Adds env-toggleable `IRONCLAW_DISABLE_CODEACT` (accepts `true` or `1`) to switch engine v2 from CodeAct (Tier 1 Python via Monty) to a structured-tools-only prompt + adapter.
- Default unset/false → behavior bit-identical to today (zero risk).
- When set, swaps `CODEACT_PREAMBLE`/`POSTAMBLE` for `STRUCTURED_TOOL_PREAMBLE`/`POSTAMBLE`, and the LLM adapter never produces `LlmResponse::Code` from fenced Python text.
- Includes a per-turn `tracing::debug!` log `engine v2 prompt mode codeact_disabled=…` so the mode is visible in diagnostics.

## Motivation

In CodeAct mode the LLM has a "text response exit" after each Python step that can short-circuit multi-step recovery flows — e.g. when a tool error returns actionable hints, the model can finalize with the error narration instead of retrying with corrected args. Structured `tool_calls` mode keeps the loop running until the model emits a finished response without pending tool intent, which is closer to the production behavior on `nearai/demo/abound`.

## Behavior matrix

| Flag | Preamble/Postamble | LLM adapter |
|---|---|---|
| unset / `false` | `CODEACT_PREAMBLE` / `CODEACT_POSTAMBLE` (existing) | Existing: detect ` ```repl `/Python → `LlmResponse::Code` |
| `true` / `1` | `STRUCTURED_TOOL_PREAMBLE` / `STRUCTURED_TOOL_POSTAMBLE` | Always `LlmResponse::Text` (no code extraction) |

## What's preserved

- All existing prompt `.md` files unchanged
- `CODEACT_SYSTEM_PROMPT_MARKER` sentinel intact (refresh logic in `loop_engine.rs` still works in both modes)
- Monty / `executor/scripting.rs` / orchestrator untouched — just unreachable when flag is on
- No new dependencies, no config struct changes, no arg threading

## Files

- `crates/ironclaw_engine/src/executor/prompt.rs` — inline structured-tool constants, `pub fn codeact_disabled()`, branch preamble/postamble selection in `build_codeact_system_prompt_inner`, debug log
- `src/bridge/llm_adapter.rs` — `text_response_from_cleaned_text` early-return guard when codeact is disabled

## Test plan

- [ ] `cargo check -p ironclaw_engine` — passes
- [ ] `cargo check` — passes
- [ ] With flag unset, send a multi-turn agent task; verify orchestrator emits `__execute_code_step__` (CodeAct path active)
- [ ] With `IRONCLAW_DISABLE_CODEACT=true`, send the same task; verify orchestrator emits only `__execute_actions_parallel__` (structured tool_calls path), no `__execute_code_step__`, and trace summary has no `NameError`/CodeAct quirks

### DIFF
--- a/crates/ironclaw_engine/src/executor/prompt.rs
+++ b/crates/ironclaw_engine/src/executor/prompt.rs
@@ -32,6 +32,36 @@ const CODEACT_PREAMBLE: &str = include_str!("../../prompts/codeact_preamble.md")
 /// The strategy/closing block appended after the dynamic metadata sections.
 const CODEACT_POSTAMBLE: &str = include_str!("../../prompts/codeact_postamble.md");
 
+/// Structured-tools-only preamble used when `IRONCLAW_DISABLE_CODEACT` is set.
+const STRUCTURED_TOOL_PREAMBLE: &str = r#"You are IronClaw, a personal AI assistant.
+
+## Execution mode
+
+Use the provider's structured tool_calls interface for every action.
+Do not emit Python, repl, py, or other executable fenced code blocks.
+Do not call tools as Python functions.
+Do not write tool invocations in assistant text. Never output `[[call_tool ...]]`, `<tool_call>`, `<function_call>`, JSON tool-call blobs, or function-style calls such as `tool_name(...)`.
+Only the provider-level `tool_calls` field invokes tools. If you need a tool, return a structured tool call instead of describing or printing the call.
+When no action is needed, answer in plain text.
+"#;
+
+/// Structured-tools-only postamble used when `IRONCLAW_DISABLE_CODEACT` is set.
+const STRUCTURED_TOOL_POSTAMBLE: &str = r#"
+## Strategy
+
+Use structured tool calls when you need data, persistence, external effects, or system state.
+After tool results are available, continue with another structured tool call or return the final plain-text answer.
+Some integrations use literal UI blocks such as `[[choice_set]]...[[/choice_set]]` in final user-facing text. These are UI markup only; do not invent other bracketed control blocks, especially `[[call_tool ...]]`.
+"#;
+
+/// Whether CodeAct (Tier 1 Python execution) is disabled by env var.
+pub fn codeact_disabled() -> bool {
+    matches!(
+        std::env::var("IRONCLAW_DISABLE_CODEACT").as_deref(),
+        Ok("true" | "1")
+    )
+}
+
 /// Marker for the engine-owned CodeAct system prompt.
 const CODEACT_SYSTEM_PROMPT_MARKER: &str = "<!-- ironclaw:codeact-system-prompt -->\n";
 const CODEACT_LEGACY_OPENING: &str = "You are an AI assistant with a Python REPL environment.";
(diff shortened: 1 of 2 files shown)


serrrfirat (changes requested):
There’s a concrete reachability regression in the current wiring for `IRONCLAW_DISABLE_CODEACT`.

In disabled-CodeAct mode, the new prompt tells the model to use provider-native `tool_calls` for actions. But the bridge still only emits `FullSchema` actions into the provider tool list, while most non-allowlisted actions default to `CompactToolInfo`.

That leaves a large class of tools visible in prompt text but not actually callable in structured mode. `tool_info` can reveal their schema, but it does not make them appear in the next provider tool list, so the model still cannot issue a structured call for them.

Concretely:
- `crates/ironclaw_engine/src/executor/prompt.rs` still advertises compact tools in the prompt when CodeAct is disabled.
- `src/bridge/llm_adapter.rs` still filters provider-native tools down to `emits_full_schema_tool()`.
- `src/bridge/action_projector.rs` defaults many real tools like `mission_create`, `gmail_send`, and `notion_search` to `CompactToolInfo`.

So with `IRONCLAW_DISABLE_CODEACT=true`, those workflows regress from callable to effectively unreachable.

A fix likely needs one of these:
- emit compact actions as provider tools in disabled-CodeAct mode, or
- stop advertising compact actions in the structured-only prompt and only mention actions that are actually callable via `tool_calls`.

I think the first option is closer to the PR’s intended behavior.

zetyquickly:
> There’s a concrete reachability regression in the current wiring for `IRONCLAW_DISABLE_CODEACT`.
> 
> In disabled-CodeAct mode, the new prompt tells the model to use provider-native `tool_calls` for actions. But the bridge still only emits `FullSchema` actions into the provider tool list, while most non-allowlisted actions default to `CompactToolInfo`.
> 
> That leaves a large class of tools visible in prompt text but not actually callable in structured mode. `tool_info` can reveal their schema, but it does not make them appear in the next provider tool list, so the model still cannot issue a structured call for them.
> 
> Concretely:
> 
> * `crates/ironclaw_engine/src/executor/prompt.rs` still advertises compact tools in the prompt when CodeAct is disabled.
> * `src/bridge/llm_adapter.rs` still filters provider-native tools down to `emits_full_schema_tool()`.
> * `src/bridge/action_projector.rs` defaults many real tools like `mission_create`, `gmail_send`, and `notion_search` to `CompactToolInfo`.
> 
> So with `IRONCLAW_DISABLE_CODEACT=true`, those workflows regress from callable to effectively unreachable.
> 
> A fix likely needs one of these:
> 
> * emit compact actions as provider tools in disabled-CodeAct mode, or
> * stop advertising compact actions in the structured-only prompt and only mention actions that are actually callable via `tool_calls`.
> 
> I think the first option is closer to the PR’s intended behavior.

addressed

serrrfirat (approved):

PR #3666, opened 2026-05-14
feat(processes): route FilesystemProcessStore through unified put/get

ilblackdragon:
## Summary

First consumer migration onto the new `RootFilesystem` surface. Stacked on #3661 (indexer). Scope is deliberately tight: switch the byte-plane `read_file`/`write_file` calls inside `ironclaw_processes`' filesystem-backed store to the unified `put`/`get` ops. The on-disk JSON layout is unchanged, every existing test passes, and downstream crates that construct `FilesystemProcessStore` (`ironclaw_host_runtime` + 4 test files) don't need to change.

## Why only `Entry::bytes` + `CasExpectation::Any`

`LocalFilesystem` (used by the `ironclaw_processes` tests via `ProcessServices::filesystem(LocalFilesystem)`) still only implements the legacy bytes plane — no native `put`. The trait's default `put` impl now **bridges that case**: opaque-file entries (`kind = None`, `indexed = {}`) with `CAS::Any` route through `write_file` so legacy backends keep working unchanged.

Record-shaped writes (`kind = Some(_)` / non-empty `indexed`) and `CAS::Absent` / `CAS::Version` modes remain `Unsupported` on legacy backends. Consumers using them must run against a backend with native `put` (libSQL, Postgres, `InMemoryBackend`).

Once `LocalFilesystem` grows native `put` with sidecar metadata in a follow-up, this consumer can switch to:

```rust
let entry = Entry::record(process_record_kind(), &serde_json::to_value(&record)?)?
    .with_indexed(IndexKey::new("scope")?, IndexValue::Text(scope.to_string()));
self.filesystem.put(&path, entry, CasExpectation::Absent).await?;
```

without changing the on-disk layout.

## What changed in `ironclaw_processes`

- `FilesystemProcessStore::write_record` — uses `put(path, Entry::bytes(body), CAS::Any)`.
- `FilesystemProcessStore::start` — keeps the `transition_lock` and uses `get` (unified read) for the existence probe; once every backend supports native `put`, this can collapse to a single `CAS::Absent` call.
- `FilesystemProcessStore::update_status` — reads via `get` instead of `read_file`. The `transition_lock` still provides per-instance atomicity per the single-instance invariant in the docstring.
- `FilesystemProcessStore::get` / `records_for_scope` — read via `get` and the new `VersionedEntry.body`.
- `FilesystemProcessResultStore::{write_result, write_output, get, output}` — all migrated to the same unified pattern.
- `FilesystemProcessResultStore::write_output` — output blobs stay opaque (`Entry::bytes` with `ContentType::json()`), no record metadata, so they round-trip through legacy `read_file`/`write_file` if any consumer still uses those.

## What changed in `ironclaw_filesystem`

Single change in `root.rs`: the trait's default `put` now bridges to `write_file` instead of returning `Unsupported`, *only* when the entry is an opaque file (`kind = None`, `indexed = {}`) and CAS is `Any`. Anything richer still returns `Unsupported` since legacy backends can't honour kinds, indexed projections, or CAS preconditions.

## Test plan

- [x] `cargo test -p ironclaw_processes` — **64 tests pass** (no regressions from the migration)
- [x] `cargo test -p ironclaw_filesystem --all-features` — **65 tests pass** (foundation + port + indexer tests still green)
- [x] `cargo clippy --all-features --tests` on both crates — clean
- [x] Pre-push lib-test gate (4967 tests) — pass
- [ ] Reviewer: the `e2e_builtin_tool_coverage::tests::job_list_cancel` integration test fails on this branch but **also fails on baseline `reborn-integration`** — pre-existing, unrelated to this PR.

## Stacked PR

Base is `reborn/fs-indexer` (PR #3661). Merge order: foundation (#3659) → port (#3660) → indexer (#3661) → this. Each subsequent consumer migration will be its own PR per the agreed cadence.

### DIFF
--- a/crates/ironclaw_processes/src/filesystem_store.rs
+++ b/crates/ironclaw_processes/src/filesystem_store.rs
@@ -15,7 +15,7 @@ use std::sync::Arc;
 
 use async_trait::async_trait;
 use ironclaw_events::sanitize_error_kind;
-use ironclaw_filesystem::{FilesystemError, RootFilesystem};
+use ironclaw_filesystem::{CasExpectation, ContentType, Entry, FilesystemError, RootFilesystem};
 use ironclaw_host_api::{ProcessId, ResourceScope, VirtualPath};
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
@@ -87,8 +87,20 @@ where
 
     async fn write_record(&self, record: &ProcessRecord) -> Result<(), ProcessError> {
         let path = process_record_path(&record.scope, record.process_id)?;
-        let bytes = serialize_pretty(record)?;
-        self.filesystem.as_ref().write_file(&path, &bytes).await?;
+        let body = serialize_pretty(record)?;
+        // Entry::bytes (kind=None) keeps the legacy on-disk layout — JSON
+        // file at `process_record_path`. The unified `put` op gives us the
+        // same write semantics as `write_file` against any backend, and
+        // backends that support records natively (libSQL, Postgres,
+        // InMemoryBackend) will still store the body as an opaque entry.
+        // Once LocalFilesystem grows native put with sidecar metadata, the
+        // consumer can switch to Entry::record(process_record_kind, ...)
+        // without changing the on-disk layout.
+        let entry = Entry::bytes(body).with_content_type(ContentType::json());
+        self.filesystem
+            .as_ref()
+            .put(&path, entry, CasExpectation::Any)
+            .await?;
         Ok(())
     }
 
@@ -120,12 +132,12 @@ where
     async fn start(&self, start: ProcessStart) -> Result<ProcessRecord, ProcessError> {
         let _guard = self.transition_lock.lock().await;
         let path = process_record_path(&start.scope, start.process_id)?;
-        let existing = match self.filesystem.as_ref().read_file(&path).await {
-            Ok(_) => true,
-            Err(error) if is_not_found(&error) => false,
-            Err(error) => return Err(error.into()),
-        };
-        if existing {
+        // Existence check uses `get` (unified read) so it works regardless of
+        // whether the backend has native put. Atomicity is provided by the
+        // transition_lock per the single-instance invariant in this struct's
+        // docstring. A future migration can switch to `CasExpectation::Absent`
+        // once every backend in production exposes native put.
+        if self.filesystem.as_ref().get(&path).await?.is_some() {
             return Err(ProcessError::ProcessAlreadyExists {
                 process_id: start.process_id,
             });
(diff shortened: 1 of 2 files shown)


crates/ironclaw_filesystem/src/root.rs
```diff
@@ -48,13 +48,32 @@ pub trait RootFilesystem: Send + Sync {
...
-        unsupported(path, FilesystemOperation::WriteFile)
+        if entry.kind.is_some() || !entry.indexed.is_empty() {
+            return unsupported(path, FilesystemOperation::WriteFile);
+        }
+        if !matches!(cas, CasExpectation::Any) {
+            return unsupported(path, FilesystemOperation::WriteFile);
+        }
+        self.write_file(path, &entry.body).await?;
+        // Legacy backends don't track per-path versions; report `0` so callers
+        // that only need lost-update protection via CAS still work — they
```
serrrfirat: P1: This new default `put` implementation calls `self.write_file(...)`, while the trait's default `write_file` implementation calls `self.put(...)`. Any `RootFilesystem` implementation that overrides neither method now recurses until stack overflow instead of returning `Unsupported`. This changes the public trait contract. Please make one side non-recursive, or keep the default `put` unsupported and add the legacy bridge somewhere that can prove `write_file` is implemented natively.

crates/ironclaw_processes/src/filesystem_store.rs
```diff
@@ -215,8 +225,10 @@ where
         let mut records = Vec::new();
```
serrrfirat, line 225: P2: This silently skips a listed `.json` child when `get` returns `None`. Previously, after `list_dir` returned a child, `read_file` failures, including `NotFound`, propagated. Returning a partial process list can hide a race or backend inconsistency and cause callers to miss a process. Please preserve the old behavior by treating `None` for a listed child as an error.

serrrfirat (commented):
Reviewed the migration without using skills. I found two issues that should be addressed before merging.

ilblackdragon:
Pushed fixes for both blockers (commit on the branch addressing PR #3666 review):

1. **Recursion bug (P1)** — trait defaults for `put` and `get` are now plain `Unsupported`. The byte-only bridge moves to `LocalFilesystem::put`/`get`, where it can't recurse because it delegates to its own native `write_file`/`read_file` (not the trait defaults).

2. **Silent skip in records_for_scope (P2)** — `get` returning `None` after `list_dir` enumerated the path now propagates as `ProcessError::Filesystem` with a descriptive reason. Matches the pre-migration NotFound propagation invariant.

Test scaffold update: `BackendErrorFilesystem` now overrides `get` too, so the fault-propagation regression test continues to exercise its intended path.

ilblackdragon:
Superseded by #3679 — single unified PR on top of #3659 (foundation) that contains all the consumer-crate work applying the universal FS dispatch. This PR's commit is preserved in the unified branch.

PR #3667, opened 2026-05-14
feat(reborn): resolve host api credential accounts

serrrfirat:
## Summary
- add host-runtime credential account resolver for Host API projected requirements
- resolve active scoped credential accounts into secret staging requirements plus exact-method-and-URL WASM staged credential rules
- preserve required/optional credential semantics for future `InjectCredentialOnce` planning
- fail closed if a credential store returns an account whose id does not match the projected requirement
- document the resolver boundary and remaining `InjectCredentialOnce` product wiring

## Tests
- `cargo fmt --check`
- `cargo test -p ironclaw_host_runtime --test credential_account_resolver_contract --locked`
- `cargo test -p ironclaw_wasm --locked`
- `cargo clippy -p ironclaw_host_runtime -p ironclaw_wasm --all-targets --locked -- -D warnings`
- `cargo clippy -p ironclaw_host_runtime --test credential_account_resolver_contract --locked -- -D warnings`

Stacked on #3663 / `reborn/extension-host-api-contracts` so manifest changes stay in that PR.


zmanian:
## Summary

`CredentialAccountResolver` is a focused, fail-closed planning boundary that turns projected host-API credential requirements into (a) `InjectSecretOnce` staging requirements and (b) exact-method-and-URL `WasmStagedRuntimeCredential` rules, without ever reading raw secret material. The resolver lives in `ironclaw_host_runtime` (composition layer), the underlying account/session contracts stay in `ironclaw_secrets`, and obligation/`InjectCredentialOnce` wiring is explicitly deferred — that layering is correct and matches both the `ironclaw_secrets` guardrails ("runtime injection is not enforced here") and the secrets contract doc note added in this PR.

## Strengths

- **Substrate independence preserved.** New code only depends on `ironclaw_extensions`, `ironclaw_host_api`, `ironclaw_secrets`, and `ironclaw_wasm`. No reach back into the legacy binary, `crate::safety::*`, `state.store.*`, or any gateway/SSE/handler surface. Documentation in `docs/reborn/contracts/{host-runtime,secrets}.md` correctly assigns the new responsibility.
- **Typed-identity discipline.** Every field on `HostApiCredentialRequirement` and `CredentialAccountResolverRequest` is a newtype (`ExtensionId`, `HostApiId`, `ManifestSectionPath`, `CapabilityId`, `CredentialAccountId`, `SecretHandle`, `NetworkMethod`, `ResourceScope`). No raw `String` IDs, no reintroduction of the legacy `credential_name` / `extension_name` confusion class. The substrate already standardized on `provider_or_extension_id` and this PR consumes it directly.
- **Fail-closed semantics, distinct error variants.** Required-missing → `MissingCredential`; store returns wrong id → `CredentialScopeMismatch`; different extension → `CredentialExtensionMismatch`; expired/revoked → distinct variants; URL/method outside policy → `CredentialPolicyMismatch`; handle not bound to the account → new `MissingSecretHandle` variant with `stable_reason()` = `"CredentialPolicyMismatch"`. Each error is exercised by a dedicated test.
- **Exact-method narrowing is a real defense-in-depth win.** Even when the projected requirement is broader, the emitted WASM rule pins both method *and* URL via `for_exact_request`, and `matches_request` rejects a mismatched method. Test `resolves_projected_host_api_requirement_into_exact_url_wasm_credentials` proves a `POST` resolution cannot bleed into a `GET` to the same URL.
- **No raw secret material leaks.** Resolver returns handles only, never `consume`s leases. Errors carry account ids and handle names (both non-secret newtypes), never values. `RedactedJson` metadata stays on the account and is not exposed.
- **No `.unwrap()` / `.expect()` in production code.** All `unwrap()` calls are confined to the test file. `thiserror` + `#[from]` is used correctly.
- **Test coverage is end-to-end through the WASM caller.** The happy-path test does not stop at the resolver — it constructs `WasmStagedRuntimeCredentials::new(resolved.wasm_credentials)` and drives `credential_injections(...)` for three method/URL combinations, which is exactly the "test through the caller, not just the helper" rule. The `MismatchedAccountStore` fake also validates that the resolver does not blindly trust the store's response.

## Issues

### Important

1. **`ResourceScope` is checked in two places with different rules; the resolver's check is the stricter of the two but the divergence is implicit.** `account_matches_request` checks `tenant_id`, `user_id`, `agent_id`, `project_id` (not `mission_id`, `thread_id`, `invocation_id`), and the test `resolves_account_from_same_account_scope_under_different_invocation` documents that invocation-id drift is acceptable. That is almost certainly the right product call (credential accounts outlive a single invocation), but the contract is currently load-bearing and only encoded in field-by-field equality. Recommend extracting a named predicate — e.g. `ResourceScope::is_account_visible_to(&request_scope)` on the substrate type — so the "which scope dimensions bind an account" decision lives next to the type rather than being re-derived in the resolver. Otherwise a future scope dimension (e.g. `org_id`) will silently be ignored here.

2. **Scope-mismatch path collapses two failure modes into one error variant.** `CredentialBrokerError::CredentialScopeMismatch` is emitted both when (a) the store returns an account whose `id != requirement.account_id` (the `MismatchedAccountStore` case — a *store bug*, fail very loud) and (b) hypothetically when scope dimensions don't agree. The PR description specifically calls out "fail closed if a credential store returns an account whose id does not match" as a design goal, so this deserves its own variant (e.g. `CredentialBrokerError::StoreIdentityViolation`) rather than reusing `CredentialScopeMismatch`. The stable-reason string is also shared, which makes the two cases indistinguishable to upstream telemetry. (Defining the new variant likely belongs in `ironclaw_secrets`, but the PR is the moment the case is first exercised.)

3. **`requirements` are matched and accumulated, but there is no defense against duplicate `(account_id, handle, target, exact_url)` requirements feeding the resolver in the first place.** `push_secret_requirement` and `push_wasm_credential` dedupe by handle-and-scope on the output, but a malformed manifest projection that lists the same requirement N times still runs N store fetches via `account_store.get_account(...)`. For a hot path, recommend either dedupe-on-construction in `CredentialAccountResolver::new` or batching matching requirements by `account_id` and only fetching once per unique account.

### Suggestions

4. **Consider a `?Sized + 'static` bound or `dyn CredentialAccountStore` alias.** The current `S: CredentialAccountStore + ?Sized` generic is fine, but every call site that wants to hold a heterogeneous set of resolvers (e.g. one per backend) will end up writing `Arc<dyn CredentialAccountStore>`. A re-exported `pub type DynCredentialAccountResolver = CredentialAccountResolver<dyn CredentialAccountStore>;` would save downstream churn.

5. **`HostApiCredentialRequirement::exact_url: Option<String>` should probably be a typed URL or, at minimum, validated at construction.** The resolver currently does string equality against `request.url`; an upstream projection that normalizes one but not the other (trailing slash, percent-encoding case) will silently miss. The `CredentialTargetPolicy::matches` path on the account uses `url::Url::parse` and is robust; the requirement-side comparison is not.

6. **`HostApiCredentialRequirement` derives `PartialEq, Eq` but no `Hash`.** If callers ever want to dedupe via `HashSet` (see issue #3) they will need it. Cheap to add now.

7. **Doc comment on `CredentialAccountResolution::wasm_credentials` says "exact-method-and-URL scoped" but the type allows a `None` method via `WasmStagedRuntimeCredential::for_exact_url(...)`.** The resolver itself only ever calls `for_exact_request`, which is correct. Consider adding a debug-assert or constructor invariant on the resolution type so a future change can't accidentally emit a broader rule.

8. **Test naming nit.** `resolves_account_from_same_account_scope_under_different_invocation` is excellent intent-encoding but quite long; the file would benefit from a short module-level comment explaining the "invocation_id is intentionally not part of account scope" rule so future readers don't have to reverse-engineer it from the test body.

## Verdict

Approve with the issues above noted. This is a clean, well-tested slice that lands the credential-account resolution boundary in the right crate, preserves typed-identity discipline, fails closed on every distinguishable misuse, and explicitly defers `InjectCredentialOnce` product wiring rather than smuggling it in. Issues 1–3 are worth a follow-up but do not block this PR — none of them re-introduce the bug classes that motivated the Extension/Auth-identity invariant or the substrate-independence rule.

PR #3668, opened 2026-05-14
chore(crates): seal internal modules across service crates

ilblackdragon:
## Summary

Sweep every non-prelude service crate and seal modules whose contents
aren't consumed via the module path from outside the crate. The audit
walked each crate's lib.rs, identified `pub mod` declarations with no
external module-path consumers (consumers go through the lib.rs root
re-exports instead), demoted them to `mod`, then demoted the now-orphaned
`pub` items inside to `pub(crate)`.

The first commit (\`e3253ce12\`) was the pilot on \`ironclaw_common\` —
sealing \`attachment\`, \`platform\`, and \`trust_boundary\`.

The second commit (\`7674be823\`) sweeps 11 more crates:

| Crate | Sealed |
|---|---|
| \`ironclaw_telegram_v2_adapter\` | adapter, payload, render |
| \`ironclaw_reborn_config\` | boot, doctor, home, profile |
| \`ironclaw_skills\` | parser, selector |
| \`ironclaw_wasm_product_adapters\` | auth_verifier, egress_policy, runner |
| \`ironclaw_trust\` | clock, decision, error, invalidation, policy, sources |
| \`ironclaw_product_workflow\` | action, binding, error, fakes, inbound_turn, ledger, workflow |
| \`ironclaw_processes\` | cancellation, filesystem_store, host, memory_store, services, types, wrappers |
| \`ironclaw_product_adapters\` | adapter, egress, error, outbound, projection, workflow |
| \`ironclaw_tui\` | app, event, input, layout, render, spinner, theme, widgets |
| \`ironclaw_reborn\` | loop_driver_host, milestone_events, model_routes, text_loop_driver, model_gateway |
| \`ironclaw_turns\` | admission, checkpoint_state, coordinator, db, ids, memory, request, response, status, store |

## Out of scope (intentionally)

- **Prelude crates** (\`ironclaw_host_api\`, \`ironclaw_engine\`, \`ironclaw_llm\`):
  their crate docs explicitly say modules are an authoritative grouping
  for consumers. Sealing would conflict with that.
- **\`ironclaw_reborn::{loop_exit_applier, production_readiness, turn_runner}\`**:
  the crate's own integration tests in \`tests/\` consume these via module
  path; kept \`pub mod\`.
- **Dead-code cleanup**: scaffolding items used only in tests or staged for
  future PRs (e.g. \`identity_changed\`, \`LocalDevOverride\`, TUI helpers)
  get a file-level \`#![allow(dead_code)]\` so the seal compiles cleanly
  without dropping the staged code. Removing or restructuring that
  scaffolding is a separate audit.

## Test plan

- [ ] \`cargo clippy --workspace --all-features --tests --benches --examples\`
  — verified locally, zero warnings, zero errors
- [ ] \`cargo fmt --all -- --check\` — verified locally
- [ ] CI green on \`reborn-integration\`


serrrfirat (changes requested):
Findings:

1. [P1] `ironclaw_trust` hides types still required by public APIs.

`crates/ironclaw_trust/src/lib.rs` now makes `clock`, `sources`, and `invalidation` private, but public methods still require their types: `HostTrustPolicy::new/add_source/with_clock` take `PolicySource`/`Clock`, and `InvalidationBus::register/publish` take `TrustChangeListener`/`TrustChange`. Downstream callers can no longer name or implement these extension traits/types. Re-export these support types from the crate root, keep the modules public, or make the affected methods crate-private if they are no longer intended as API.

2. [P1] `HmacWebhookAuth::with_clock` now exposes an unnameable trait.

`crates/ironclaw_wasm_product_adapters/src/lib.rs` seals `auth_verifier` and re-exports `HmacWebhookAuth`, but not `Clock`. The struct still has `pub clock: Box<dyn Clock>` and `pub fn with_clock(... Box<dyn Clock>)` in `auth_verifier.rs`, so external tests/installations cannot inject deterministic clocks anymore. Re-export `Clock` or keep `auth_verifier` public.

3. [P2] `ironclaw_common` removes public module paths from a publishable crate.

`ironclaw_common` is not marked `publish = false`, but `attachment`, `platform`, and `trust_boundary` are now private. Root re-exports preserve some local imports, but existing downstream code using valid paths like `ironclaw_common::attachment::AttachmentKind` or `ironclaw_common::platform::PlatformInfo` breaks. Either keep those modules public or treat this as an intentional public API break with the appropriate release handling.

Verification: I checked the PR tree with `git grep` for remaining in-repo module-path consumers and did not find any. I also started `cargo check --workspace --all-features --tests --examples --benches`; it reached the affected crates, then failed compiling the root crate because `/tmp` ran out of space while writing incremental query cache, so full workspace verification is incomplete.

ilblackdragon:
Thanks for the review @serrrfirat — addressed in `2f0d3b8f9`.

**1 (P1) — `ironclaw_trust` unnameable trait/type parameters.** Fixed. Added `pub use` re-exports from the crate root for `Clock`, `PolicySource`, `TrustChange`, and `TrustChangeListener`. The modules stay sealed; the support types that appear in `HostTrustPolicy::{new,add_source,with_clock}` and `InvalidationBus::{register,publish}` signatures are now nameable from the crate root.

**2 (P1) — `HmacWebhookAuth::with_clock` unnameable `Clock`.** Fixed. Added `Clock` to the `pub use auth_verifier::{...}` re-export list in `ironclaw_wasm_product_adapters` lib.rs.

**3 (P2) — `ironclaw_common` semver.** Acknowledged. Confirmed with @ilblackdragon that `ironclaw_common` (and the other crates this PR seals) currently has no consumers outside this repo, so the path-removal is intentional and not a stable-API break in practice. If/when these crates pick up external consumers we'll add `publish = false` or version-bump deliberately.

Verified locally: `cargo clippy --workspace --all-features --tests --benches --examples` — zero warnings, zero errors.

PR #3678, opened 2026-05-15
refactor(workspace): dissolve ironclaw_storage + HsmBackend placeholder

ilblackdragon:
Two follow-on commits on top of the universal-FS-dispatch cascade. Closes tasks #9, #18, and the demonstrable portion of #19 of the rework plan.

## Summary

**Commit 1: dissolve `ironclaw_storage` crate.**

The crate predated the unified `RootFilesystem` surface introduced by PR #3659. Its `BlobStore`/`RecordStore` traits, `StorageKey`/`StorageVersion`/`PutCondition` types, and `StoredBlob`/`StoredRecord` shapes parallel the new unified `put`/`get`/`CasExpectation`/`RecordVersion` machinery — a textbook duplicate-dispatch smell from `.claude/rules/architecture.md`.

Only `ironclaw_outbound` consumed any of the crate, and only 5 small helpers (`encode_json`, `decode_json`, `redacted_backend_error`, `StorageError::Backend`, `ABSENT_SCOPE_COMPONENT`). All other types and the entire `BlobStore`/`RecordStore` surface (660 LOC) were unused — their intended consumers already moved to `RootFilesystem` directly.

- Inlined the 5 helpers into `crates/ironclaw_outbound/src/db.rs`
- Removed `ironclaw_storage = { path = ... }` from outbound `Cargo.toml`
- Removed `\"crates/ironclaw_storage\"` from workspace members
- Removed the BoundaryRule for `ironclaw_storage` from architecture tests
- Updated the `ironclaw_outbound` BoundaryRule to permit `ironclaw_filesystem` (it was stale: `FilesystemOutboundStateStore` landed in the prior cascade PR #3670)
- Deleted the crate directory

**Commit 2: `HsmBackend` placeholder + database.md scoping.**

`HsmBackend` (`crates/ironclaw_filesystem/src/hsm.rs`) demonstrates the universal-dispatch seam end-to-end: one trait, one file, declared restricted capabilities, mount-time validation rejects over-claiming descriptors, swap-by-wiring works without consumer edits. Five tests cover all of those gates including the rejection paths. A production HSM swaps the in-process placeholder for a real session handle; the trait + capability machinery is reusable as-is.

`.claude/rules/database.md` is now scoped via its `paths:` frontmatter to `src/db/**`, `src/history/**`, and `migrations/**` — exactly the legacy directories that predate the universal FS dispatch — with a 'Status & Direction' preamble pointing new persistence work at `ScopedFilesystem` and the rework plan.

## Test plan

[...]

PR #3672, opened 2026-05-15
feat(run-state): unified put/get for filesystem stores

ilblackdragon:
## Summary

Stacked on #3671. Mirrors the ironclaw_processes (#3666) and ironclaw_authorization (#3671) migrations. Switches all `read_file`/`write_file` calls in `FilesystemRunStateStore` and `FilesystemApprovalRequestStore` to the unified `get`/`put` ops with `Entry::bytes` + `CasExpectation::Any`. On-disk JSON layout unchanged.

The per-record `filesystem_record_lock` keeps transition serialization within a single instance; production shared roots still need the SQL backends or explicit CAS per the crate guardrail.

## Test plan

- [x] `cargo test -p ironclaw_run_state` — **32 tests pass** (6 lib + 26 contract)
- [x] `cargo clippy -p ironclaw_run_state --tests` — clean
- [x] `cargo fmt --check` — clean

## Stacked PR

Base is `reborn/fs-authorization` (#3671). Merge order: foundation → port → indexer → processes → outbound → authorization → this.

PR #3671, opened 2026-05-15
feat(authorization): route FilesystemCapabilityLeaseStore through unified put/get

ilblackdragon:
## Summary

Stacked on #3670. Mirrors the ironclaw_processes migration (#3666). Switches `FilesystemCapabilityLeaseStore`'s `read_file`/`write_file` calls to the unified `get`/`put` ops with `Entry::bytes` + `CasExpectation::Any`. The on-disk JSON layout is unchanged, every existing test passes, and the per-owner `mutation_lock` continues to serialize claim/consume/revoke within a single instance.

The crate guardrail's note that production shared roots need a transactional backend or explicit CAS still applies — this PR preserves lock-based single-instance semantics. Switching to backend CAS becomes a small follow-up once LocalFilesystem grows native put.

## Touch points

- `read_lease`, `read_lease_index`, `read_lease_file` — now use `get` and unwrap `VersionedEntry.body`.
- `write_lease`, `write_lease_index` — now use `put(Entry::bytes, CasExpectation::Any)`.
- Imports updated for the new types.

## Test plan

- [x] `cargo test -p ironclaw_authorization` — **47 tests pass** (no regressions)
- [x] `cargo clippy -p ironclaw_authorization --tests` — clean
- [x] `cargo fmt --check` — clean
- [x] `cargo check --workspace --all-features` — clean

## Stacked PR

Base is `reborn/fs-outbound` (PR #3670). Merge order so far: foundation → port → indexer → processes → outbound → this.

PR #3670, opened 2026-05-15
feat(outbound): FilesystemOutboundStateStore on the unified surface

ilblackdragon:
## Summary

Stacked on #3666. Adds a new `FilesystemOutboundStateStore` that persists outbound metadata under `/engine/outbound/...` through any `RootFilesystem` — additive, no removals. The existing `InMemoryOutboundStateStore` and `LibSql/Postgres` backends stay on the `OutboundStateStore` trait so callers can flip to the new store via a constructor change at their convenience.

## Why additive, not destructive

Unlike `ironclaw_processes` (#3666), the rest of the consumer crates don't currently touch `ironclaw_filesystem` at all — their SQL backends are independent. Migrating means **adding** a new filesystem-based store rather than **modifying** an existing one. The deletions of `libsql_store.rs`/`postgres_store.rs`/`db.rs` come in a final cleanup PR per consumer once the new store has run in production.

## Path scheme

- `/engine/outbound/policies/<thread-scope-hash>.json` — thread notification policy keyed by `(tenant, agent?, project?, thread)`.
- `/engine/outbound/subscriptions/<sub-key-hash>.json` — projection subscription cursor. The key hashes `(subscription_id, actor, scope, thread)` so the path doesn't leak the actor on directory listing — preserves the anti-enumeration semantics in `OutboundStateStore::load_subscription_cursor`'s docstring.
- `/engine/outbound/deliveries/<delivery_id>.json` — delivery attempt keyed by `delivery_id`. `list_delivery_attempts` currently scans the deliveries directory and filters by scope in memory; once the SQL backends' query plane is wired into this consumer, the scan becomes `query(prefix, Filter::Eq { key: "scope", ... })`.

## CLAUDE.md guardrails preserved

- No raw payloads, prompts, tool I/O, secrets, host paths, or backend error strings persisted (every value goes through the existing `validation::*` checks first).
- Backend errors collapse to `OutboundError::Backend` via `map_fs_error` so `FilesystemError` host-path detail can't leak.
- Trust-bearing sealed types (`ThreadProjectionAccessGrant`, `ValidatedReplyTargetBinding`) are unaffected — this PR only touches the store layer; the policy service is unchanged.

## Test plan

[...]

PR #3679, opened 2026-05-15
feat(reborn): apply universal FS dispatch across consumer crates

ilblackdragon:
Single PR stacking ALL the consumer-crate work that applies #3659 (universal `RootFilesystem` dispatch fabric) across the codebase. Supersedes the prior stacked PRs (#3666, #3670, #3671, #3672, #3678 — all closed).

## What's in here

13 commits, +15,214 / -929 LOC across 61 files. Every commit is independently reviewable.

| # | Commit | Scope |
|---|--------|-------|
| 1 | `feat(processes)` | `FilesystemProcessStore` on the unified put/get surface |
| 2 | `feat(outbound)` | `FilesystemOutboundStateStore` on the unified surface |
| 3 | `feat(authorization)` | `FilesystemCapabilityLeaseStore` through unified put/get |
| 4 | `feat(run-state)` | `FilesystemRunStateStore` + `FilesystemApprovalRequestStore` |
| 5 | `refactor(workspace)` | Dissolve `ironclaw_storage` (660 LOC, parallel BlobStore/RecordStore traits removed) |
| 6 | `feat(filesystem)` | `HsmBackend` placeholder + scope `database.md` to legacy |
| 7 | `feat(reborn)` | Native `append`/`tail` on libsql + postgres (V30 migration); `FilesystemDurableEventLog` |
| 8 | `feat(secrets)` | `FilesystemSecretStore` + `FilesystemCredentialBroker` |
| 9 | `feat(filesystem,memory)` | `IndexKind::Fts` + `IndexKind::Vector` on libsql + postgres; `FilesystemMemoryDocumentRepository` |
| 10 | `feat(db) conversations+jobs` | `FilesystemConversationStore` + `FilesystemJobStore` (~1700 LOC + 23 tests) |
| 11 | `feat(db) sandbox+routine+toolfail` | `FilesystemSandboxStore` + `FilesystemRoutineStore` + `FilesystemToolFailureStore` (~2000 LOC + 20 tests) |
| 12 | `feat(engine)` | `FilesystemStore` — full 33-method engine `Store` impl on the unified surface (~700 LOC + 24 tests; HybridStore untouched) |
| 13 | `feat(db) settings+workspace+rest` | `FilesystemSettingsStore` + `FilesystemUsersStore` + `FilesystemChannelPairingStore` + `FilesystemIdentityStore` + `FilesystemWorkspaceStore` (split across 5 sub-modules; ~4400 LOC + 47 tests) |

## Coverage on the universal-FS surface

- **Bytes plane**: ✅ every consumer routes through `put(Entry::bytes(...))`/`get`
- **Records plane**: ✅ every consumer writes `Entry::record(...)` with indexed projection
- **Query plane**: ✅ all consumers use `Filter::Eq`/`PrefixOn`/`And`/`Or`/`Fts`/`VectorNearest`
- **Index plane**: ✅ `ensure_index` declares Exact/Prefix at startup; FTS5 + tsvector + cosine vector advertised by libsql + postgres
- **Event plane**: ✅ libsql + postgres natively serve `append`/`tail`; FilesystemDurableEventLog + sandbox job-events use it
- **CAS**: ✅ every record store uses `CasExpectation::Version` with retry on `VersionMismatch`
- **Capability declaration + mount-time validation**: ✅ `BackendCapabilities` bitmask; descriptor over-claims surface `FilesystemError::DescriptorOverclaims`

## What's still pending (intentional follow-ups)

- **Composite `Database` trait dissolution**: each of the 10 sub-traits now has a `Filesystem*Store` facade in `src/db/filesystem_*.rs`, but the composite trait itself in `src/db/mod.rs` is unchanged — both legacy SQL impls and new FS-backed impls coexist. Final removal of the composite trait is intentionally deferred behind the wiring switch (out of scope here).
- **Host wiring switch**: `src/app.rs` still constructs `HybridStore` against the legacy `Database` impl. Switching to `CompositeRootFilesystem`-backed wiring is a single-PR follow-up.
- **`EncryptedBackend` decorator**: encryption stays embedded in `SecretsCrypto` for the migration window; documented TODO to move to a backend decorator.
- **Real HSM backend**: `HsmBackend` is a placeholder demonstrating the seam.

## Verification

- `cargo fmt --check` clean
- `cargo check --workspace --all-features` clean
- `cargo clippy --workspace --all-features --tests -- -D clippy::correctness` clean
- `cargo test` per migrated crate — all green; engine: 549 tests; secrets: 60; memory: 219 (2 pre-existing `e2e_hybrid_search` failures unrelated); event_store: 57; architecture: 14; workspace lib `db::filesystem` 86 tests
- Postgres-tier tests gracefully skip when no DB reachable; will run in integration CI

## Migration pattern (canonical, used by every commit)

```rust
pub struct FilesystemFooStore<F: RootFilesystem> {
    fs: Arc<F>,
}

impl<F: RootFilesystem> FooStore for FilesystemFooStore<F> {
    async fn upsert(&self, foo: &Foo) -> Result<(), FooError> {
        let entry = Entry::record(RecordKind::new("foo")?, &foo.to_json()?)?
            .with_indexed(IndexKey::new("status")?, IndexValue::Text(foo.status_str().into()))
            .with_indexed(IndexKey::new("user_id")?, IndexValue::Text(foo.user_id.clone()));
        self.fs.put(&foo_path(&foo.id), entry, CasExpectation::Version(foo.version)).await?;
        Ok(())
    }
}
```

Path layout: `/<root>/<scope>/<id>` for entities, `/<root>/<scope>/<id>/<child>/<child_id>` for sub-entities. Indexed projections + `Filter::Eq + And` for queries. CAS retry on `VersionMismatch`.

## Test plan

- [x] cargo fmt --check
- [x] cargo check --workspace --all-features
- [x] cargo clippy --workspace --all-features --tests -- -D clippy::correctness
- [x] Per-crate tests for every migrated crate (filesystem, secrets, event_store, memory, processes, authorization, run_state, outbound, engine, architecture)
- [x] Workspace lib db::filesystem tests (86 pass)
- [ ] Full CI run (pending)
- [ ] Postgres-tier integration CI (gracefully skipped locally)


ilblackdragon:
**Self-review round done.** Spawned a code-reviewer agent against the integrated branch; it returned 7 HIGH / 8 MEDIUM / 7 LOW findings. The HIGH-severity fixes are now in:

| # | Issue | Fix commit |
|---|-------|-----------|
| H1 | `hybrid_search` scanned every chunk in memory instead of using the native FTS/Vector indexes the libsql + postgres backends advertise | \`990c4f73e\` — rewrote `FilesystemWorkspaceStore::hybrid_search` to call `Filter::Fts` / `Filter::VectorNearest` with a graceful fallback to the scan path on `FilesystemError::Unsupported`; declared FTS+Vector indexes lazily via `OnceCell`-gated `ensure_chunk_indexes` |
| H2 | `parse_state` mapped unknown status strings silently to `JobState::Pending`, masking schema drift | \`9c292f9d0\` — returns `Result<JobState, DatabaseError>`; call site propagates with `?` |
| H3 | libsql FTS5 trigger DDL spliced path literals — single quotes allowed by `VirtualPath::new` could inject SQL | \`e2530adff\` — DDL emission site rejects any char outside `[A-Za-z0-9_/.-]`; regression test asserts `path-LIKE`-injection paths are refused at `ensure_index` time |
| H4 | `update_user_status`/`update_thread_state`/`revoke_lease`/etc. used `CasExpectation::Any` — lost concurrent updates | \`e2530adff\` — canonical CAS-Version + retry-on-`VersionMismatch` loop applied to filesystem_users, filesystem_jobs (update_job_status, mark_job_stuck), and the engine `FilesystemStore` (`update_thread_state`, `revoke_lease`, `update_mission_status`) |
| H5 | Engine store silently returned `Ok(())` on unknown ids without annotation | \`9c292f9d0\` — added explicit `// silent-ok:` annotations naming HybridStore-parity at each `let Some(versioned) = ... else { return; }` site |
| H6 | `is_engine_unsupported` substring-matched the error message; would false-positive on unrelated errors mentioning "Unsupported" | \`9c292f9d0\` — `fs_to_engine_error` now tags the typed `FilesystemError::Unsupported` discriminator with a stable `[fs:unsupported]` sentinel; check matches that sentinel |
| H7 | `FilesystemChannelPairingStore::find_pending_requests` silently dropped corrupted records | \`9c292f9d0\` — propagates `DatabaseError::Serialization` with the offending stored path |

Also (peripheral but related): vector "nearest" in libsql + postgres no longer loads every row's body to rank — first stage selects `path, indexed, version`, ranks by cosine, then `get()` materializes only the top-k bodies.

MEDIUM/LOW findings (N-scan operations missing indexes, file-size budget violations, etc.) are tracked as follow-ups; none affect correctness.

Test counts after the fix round (local, `--all-features`):
- `ironclaw_filesystem`: 101 pass
- `ironclaw_engine`: 549 pass (525 lib + 24 FilesystemStore tests)
- `db::filesystem` (workspace lib): 88 pass
- `ironclaw_architecture`: 14 pass
- `cargo clippy --workspace --all-features --tests -- -D clippy::correctness`: clean
- `cargo fmt --check`: clean

CI now running on the integrated branch (run 25911695237).

serrrfirat:
Codex review found several actionable correctness/security issues in the filesystem-backed stores:

1. **P1: Secret lease consumption needs CAS** (`crates/ironclaw_secrets/src/filesystem_store.rs`)
   `consume` reads an active lease, decrypts the secret, then writes the consumed state with `CasExpectation::Any`. The process-local lock does not protect shared filesystem backends across processes, so two consumers can both consume a one-time lease. Preserve the lease `RecordVersion` from the read and commit the consumed transition with `CasExpectation::Version` (or fail/retry on mismatch).

2. **P1: Credential session use increments need CAS** (`crates/ironclaw_secrets/src/filesystem_store.rs`)
   `consume_session_use` validates `stored.uses`, increments it, then writes with `CasExpectation::Any`. Concurrent processes can both pass `max_uses` checks and overwrite each other. Write against `versioned.version` and reject/retry on `VersionMismatch`.

3. **P2: Conversation message queries need pagination** (`src/db/filesystem_conversations.rs`)
   `list_messages_internal` fetches only `Page::MAX_LIMIT` rows before sorting/paginating in memory. Conversations with more than 1024 messages are silently truncated and `has_more` can be computed from an incomplete set. Loop through pages until a short page, or push caller pagination into the filesystem query.

4. **P2: Filesystem job queries need pagination** (`src/db/filesystem_jobs.rs`)
   The shared `run_query` helper fetches only one max page. Scopes with more than 1024 jobs/actions/estimation records can miss existing rows, affecting listings, summaries, and updates. Mirror the paged loop used by the engine store.

5. **P2: Settings path encoding is lossy** (`src/db/filesystem_settings.rs`)
   `encode_segment` maps unsafe characters to `_`, so distinct keys like `a/b` and `a_b` collide under the same path and can overwrite each other. Use a reversible/collision-free encoding such as percent-encoding or base64-url.

6. **P2: SQL index names need collision resistance** (`crates/ironclaw_filesystem/src/db.rs`)
   `sql_index_name` truncates long identifiers without adding a hash. Distinct long `(prefix, name)` specs can produce the same DDL object name, and `CREATE ... IF NOT EXISTS` can silently reuse the wrong backing index/table/triggers. Include a stable hash suffix before truncation.

7. **P2: In-memory backend should reject writes over implicit directories** (`crates/ironclaw_filesystem/src/in_memory.rs`)
   `InMemoryBackend::put` allows writing `/memory/a` while `/memory/a/b` exists, producing a tree shape SQL backends reject via `has_child_entry`. Add the same descendant check so tests cannot pass against impossible production state.

ilblackdragon:
**Review-response summary** — all HIGH-severity and P2-correctness findings from gemini-code-assist + codex review are now addressed.

| Finding | Severity | Status | Commit |
|---|---|---|---|
| Event-store `tail(0)` head probe O(N) on cold path (`crates/ironclaw_reborn_event_store/src/filesystem_store.rs:135`) | HIGH (gemini) | **Fixed** — bounded `tail(after - 1)` probe; returns ≤1 record instead of the whole log | \`0ebc22c83\` |
| `SecretStore::consume` TOCTOU on multi-process (`crates/ironclaw_secrets/src/filesystem_store.rs:412`) | HIGH (gemini) / P1 (codex) | **Fixed** — `CasExpectation::Version` + 3-attempt retry loop; new `filesystem_secret_store_consume_retries_on_version_mismatch` regression test using a `VersionRacingBackend` | \`df8ae8d1c\` |
| `consume_session_use` TOCTOU (`crates/ironclaw_secrets/src/filesystem_store.rs:728`) | HIGH (gemini) / P1 (codex) | **Fixed** — same CAS-retry pattern; regression test added | \`df8ae8d1c\` |
| `list_messages_internal` truncates at `Page::MAX_LIMIT` (`src/db/filesystem_conversations.rs`) | P2 (codex) | **Fixed** — extracted `query_all_pages` drain helper; regression test writes `MAX_LIMIT + 5` messages | \`e02cc8c4e\` |
| `run_query` truncates jobs/actions/estimations at `MAX_LIMIT` (`src/db/filesystem_jobs.rs`) | P2 (codex) | **Fixed** — same drain helper; regression tests for actions and listings | \`e02cc8c4e\` |
| `encode_segment` lossy `_` substitution (`src/db/filesystem_settings.rs`) | P2 (codex) | **Fixed** — percent-encoding; `encode_segment_is_collision_free` test proves `a/b` and `a_b` no longer collide | \`0ebc22c83\` |
| `sql_index_name` truncates without hash (`crates/ironclaw_filesystem/src/db.rs`) | P2 (codex) | **Fixed** — appends 8-char blake3 suffix before truncating | \`0ebc22c83\` |
| `InMemoryBackend::put` allows write over implicit dir (`crates/ironclaw_filesystem/src/in_memory.rs`) | P2 (codex) | **Fixed** — descendant check matches SQL backends' contract | \`0ebc22c83\` |
| `leases_for_scope` N+1 list (`crates/ironclaw_secrets/src/filesystem_store.rs:473`) | MEDIUM (gemini) | **Partial** — kept the scan; added `TODO(perf)` documenting bound (capped by per-owner path prefix + short lease TTL) and the index path that would replace it | \`df8ae8d1c\` |
| `filesystem_outbound::list` N+1 pattern (line 224) | MEDIUM (gemini) | **Deferred** — same shape applies; the scan is bounded by per-tenant cardinality. Will fold into a follow-up perf pass once the host wires production traffic against the unified store |
| `filesystem_processes::list` N+1 pattern (line 243) | MEDIUM (gemini) | **Deferred** — same reasoning as above |

**Verification** after the fix round:
- `cargo check --workspace --all-features` clean
- `cargo test -p ironclaw_filesystem -p ironclaw_secrets -p ironclaw_reborn_event_store --all-features` all pass
- `cargo test --lib --all-features db::filesystem` — 91 pass (was 88, added 3 pagination + collision-free tests)
- `cargo clippy --workspace --all-features --tests -- -D clippy::correctness` clean
- `cargo fmt --check` clean

serrrfirat (changes requested):
Correctness/security review findings:

1. High: approved request can be stranded without a lease
`crates/ironclaw_approvals/src/lib.rs` now marks the approval `Approved` before issuing the lease, then returns immediately if `self.leases.issue(...)` fails. I do not see an actual re-issue path: later approval attempts hit `NotPending`, while resume still requires a matching lease and fails with `ApprovalLeaseMissing`. This changes behavior from "no approval without usable lease" to "approved but impossible to resume." It fails closed from a security angle, but it is a correctness/availability regression.

2. High: filesystem secrets bind ciphertext to full invocation scope but store/read by owner scope
`filesystem_secret_aad` includes `mission_id`, `thread_id`, and `invocation_id`, but `secret_path` stores under only tenant/user/agent/project and `same_scope_owner` checks only those owner fields. That means a secret written under one invocation appears present to another invocation in the same owner scope, but consume decrypts with different AAD and fails as backend unavailable. A later put from another invocation also overwrites the same path. Either include full scope in the path/read predicate or bind AAD to the same owner scope the store actually keys on.

3. Medium: event append dedupe is still non-atomic
`FilesystemStore::append_events` pre-reads to skip duplicate event IDs, but the write path still uses `CasExpectation::Any`. Two processes can both observe "missing" and then last-writer-wins overwrite the event. For append-only parity, this should use `CasExpectation::Absent` and treat version conflict/existing row as the duplicate case.

Verification: `cargo test -p ironclaw_approvals -p ironclaw_secrets --all-features` passed. `cargo test -p ironclaw_engine --test filesystem_store` passed.

serrrfirat:
Second-pass finding:

- **High correctness regression: byte-only `LocalFilesystem` backends no longer satisfy filesystem-backed store contracts.** `LocalFilesystem::put` only accepts opaque entries with `CasExpectation::Any` (`crates/ironclaw_filesystem/src/local.rs:175-201`), but the new filesystem stores now issue CAS/indexed/record writes while still documenting or relying on `LocalFilesystem` compatibility. The most direct break is `FilesystemProcessStore::update_status`: the comment says unsupported versioned writes fall back to `Any` (`crates/ironclaw_processes/src/filesystem_store.rs:153-157`), but `put_with_byte_fallback` retries with the same `CasExpectation::Version(versioned.version)` (`crates/ironclaw_processes/src/filesystem_store.rs:182-187` and `:751-756`). As a result, a process can be started on a local engine filesystem, but `complete`/`fail`/`kill` fail with `operation write_file is not supported...`, leaving durable local process state stuck `Running`. This is not just theoretical: `cargo test -p ironclaw_processes filesystem_process_store_persists_under_resource_scope_engine_processes --test process_store_contract -- --exact --nocapture` fails at `crates/ironclaw_processes/tests/process_store_contract.rs:1511` with that backend error. The same compatibility assumption appears in other new facades as well: outbound says any `RootFilesystem` works (`crates/ironclaw_outbound/src/filesystem_store.rs:62-64`) but writes `Absent`/`Version` and indexed entries (`:80-92`, `:98-115`, `:210-224`, `:302-309`), and secrets says any `RootFilesystem` works (`crates/ironclaw_secrets/src/filesystem_store.rs:172-176`) but consumes/revokes via `CasExpectation::Version` (`:436-443`, `:509-516`, `:843-850`). Either the stores need a real byte-only fallback, or construction/composition/tests need to stop treating `LocalFilesystem` as a supported backend for these durable stores.

serrrfirat:
Tenant-isolation follow-up finding:

- **Medium/High multi-tenancy risk: `FilesystemStore` is not tenant-scoped even though it is the new shared-root engine-store target.** The new path layout explicitly omits tenant/user path prefixes and says isolation is enforced by the mount-table layer (`crates/ironclaw_engine/src/store/paths.rs:3-8`), while the store constructor accepts any `RootFilesystem` (`crates/ironclaw_engine/src/store/filesystem.rs:92-110`) and production composition builds a single shared SQL root filesystem for the Reborn substrates (`crates/ironclaw_reborn_composition/src/lib.rs:156-172` and `:219-235`). The engine records themselves only carry `user_id`/`project_id` for ownership, not `tenant_id` (`crates/ironclaw_engine/src/types/project.rs:73-78`, `crates/ironclaw_engine/src/types/thread.rs:203-219`, `crates/ironclaw_engine/src/types/memory.rs:51-58`, `crates/ironclaw_engine/src/types/mission.rs:133-140`), and the filesystem store indexes/queries by `user_id` only for user-facing lists (`crates/ironclaw_engine/src/store/filesystem.rs:489-530`, `:795-805`, `:844-880`, `:1004-1015`). In a hosted multi-tenant deployment where two tenants can have the same user id or deterministic project slug, this shared `/engine/projects/<project_id>.json`, `/engine/threads/<thread_id>.json`, `/engine/memory/<project_id>/...`, etc. namespace can return or overwrite another tenant’s engine state. The rest of this PR’s durable stores explicitly include `/tenants/{tenant}/users/{user}` in paths (for example process records at `crates/ironclaw_processes/src/filesystem_store.rs:613-631`), so the engine store is the odd one out. Please either include tenant id in the engine store path/index model, or make `FilesystemStore` impossible to construct except over a tenant-scoped root and add tests proving two tenants with the same user/project ids cannot see each other.

serrrfirat:
Security follow-up finding:

- **Medium: `HsmBackend` allows directory listing even though the secrets/HSM surface declares no `List` capability.** The new backend documents that it exposes only encrypted bytes (`Read` / `Write` / `Stat` / `Delete`) and explicitly says there is no query/index/event surface (`crates/ironclaw_filesystem/src/hsm.rs:16`-`22`). Its declared capabilities also omit `Capability::List` (`crates/ironclaw_filesystem/src/hsm.rs:53`-`60`). However the implementation still overrides `list_dir` and delegates to the in-memory backend (`crates/ironclaw_filesystem/src/hsm.rs:94`-`96`), and the composite dispatcher forwards `list_dir` directly to the mounted backend (`crates/ironclaw_filesystem/src/catalog.rs:305`-`306`). Mount validation does not enforce legacy `read`/`write`/`list` bits (`crates/ironclaw_filesystem/src/catalog.rs:140`-`149`), so omitting `List` from `capabilities()` is not a runtime guard. A caller with a scoped read-only/read-write grant gets `list: true` by default (`crates/ironclaw_host_api/src/mount.rs:33`-`49`), so mounting this backend for `/secrets` can expose secret/account/session path names and key existence metadata even though the HSM boundary claims not to support enumeration. Please make `HsmBackend::list_dir` return `FilesystemError::Unsupported` (or explicitly advertise and document `List` and ensure secrets mounts never grant it).

serrrfirat:
Correctness/security review on current head `440be242cfb1d3953de4541ee91bcd23622f8530`: I don't think this is ready to merge yet.

1. **High security: production wiring collapses tenant/user namespaces for the new filesystem-backed stores.** `wrap_scoped` still builds the singleton mount view that maps `/secrets`, `/authorization`, `/run-state`, `/approvals`, etc. to top-level paths with no tenant/user prefix (`crates/ironclaw_reborn_composition/src/lib.rs:126-156`, `:207-215`), and the production builders/factory wire that scoped filesystem into the long-lived secret, lease, and run-state stores (`crates/ironclaw_reborn_composition/src/factory.rs:246-280`, `:313-345`). The replacement stores explicitly omit tenant/user from their own paths because they expect the mount view to supply it: e.g. `secret_owner_alias` only appends agent/project (`crates/ironclaw_secrets/src/filesystem_store.rs:1139-1152`), lease paths omit tenant/user (`crates/ironclaw_authorization/src/lib.rs:1483-1508`), and run/approval paths do the same (`crates/ironclaw_run_state/src/lib.rs:1003-1036`). `wrap_scoped_for_invocation` exists, but it is not what the production builders use. In a multi-tenant process, same-shape records from different tenants/users now share the same backing path; at minimum this allows cross-tenant overwrite/DoS of secrets and state, and `read_lease_versioned` also returns the deserialized lease without a stored-scope check (`crates/ironclaw_authorization/src/lib.rs:461-477`). Either construct these stores from the per-invocation mount view everywhere they handle scoped data, or keep tenant/user in the store keys and add a factory-level two-tenant regression test through the actual production wiring.

2. **High correctness: version history is written before the document CAS, so losing writers leave phantom versions.** `write_document_with_options` archives the previous body via `save_document_version` before the `CasExpectation::Version/Absent` document write (`crates/ironclaw_memory/src/repo/filesystem.rs:520-556`). The append path does the same before its CAS write and then maps `VersionMismatch` to `MemoryAppendOutcome::Conflict` (`:591-627`). If another writer wins between the initial read and the final `put`, the loser returns conflict but has already inserted a `.versions/<n>` record for an overwrite/append that never happened. That corrupts version history and can also block the real writer's later archive because `save_document_version` uses `CasExpectation::Absent` for the computed next version (`:317-352`). The archive needs to be part of the same transaction/CAS success path, or be written only after proving the document update succeeded with a collision-safe version id.

3. **High correctness: stale reindexers can delete the winner's fresh chunks before returning stale.** `replace_document_chunks_if_current` verifies the hash, immediately sweeps the `.chunks` subtree, then re-checks the document hash (`crates/ironclaw_memory/src/repo/filesystem.rs:837-870`). If a writer updates the document and a fresh indexer writes chunks after the first stale read but before the stale worker's sweep, the stale worker deletes those fresh chunks and then returns `SkippedStaleContentHash`. The second hash check prevents stale chunks from being written, but it is too late to prevent data loss. Re-check currentness before deleting, or use a per-document generation/transaction so delete+replace is conditional on the same observed document version.

4. **High correctness: DB-backed memory full-text search no longer works because the repository never declares the FTS index it queries.** `FilesystemMemoryDocumentRepository::search_documents` sends `Filter::Fts { key: content }` directly to the root filesystem (`crates/ironclaw_memory/src/repo/filesystem.rs:773-796`), but there is no `ensure_index` call in that repository. The libSQL backend only translates FTS when it discovers a cataloged FTS table (`crates/ironclaw_filesystem/src/libsql.rs:1068-1138`) and otherwise returns `FilesystemError::Unsupported` (`:1444-1450`). The filesystem DB contract tests explicitly call `ensure_index` before FTS queries, while this PR deletes the native libSQL/Postgres memory repos and DB memory contract tests and removes the memory crate's SQL features (`crates/ironclaw_memory/Cargo.toml:13-40`; `.github/workflows/test.yml:206-209`). Please either ensure the content FTS index under the memory scope before search/write startup, or provide a bounded fallback, and restore SQL-backed memory repository coverage via `LibSqlRootFilesystem`/`PostgresRootFilesystem`.

5. **High correctness: internal memory sidecar paths collide with legal user document paths.** `MemoryDocumentPath` validation allows relative paths ending in `.meta` and segments ending in `.chunks`/`.versions` (`crates/ironclaw_memory/src/path.rs:299-328`). The repository stores a document at the literal virtual path (`crates/ironclaw_memory/src/path.rs:145-150`), but metadata for `foo` is also stored at `foo.meta` (`crates/ironclaw_memory/src/repo/filesystem.rs:127-142`, `:675-700`). So a legal user document `foo.meta` is the same backend path as metadata for `foo`; writing metadata for `foo` overwrites that document with JSON bytes. `list_documents` then filters these names out anyway (`crates/ironclaw_memory/src/repo/filesystem.rs:721-765`, `:895-899`), contradicting the module doc that claims literal user paths sharing sidecar names are never treated as internal (`:22-25`). Reserve/reject those names, or move sidecars into an escaped internal namespace that no user document can address.

6. **Medium security/capability correctness: `HsmBackend` exposes list despite declaring and documenting no list capability.** The module says the HSM surface is only `Read`/`Write`/`Stat`/`Delete` (`crates/ironclaw_filesystem/src/hsm.rs:16-22`), and `declared_capabilities` omits `Capability::List` (`:53-60`), but `list_dir` delegates to the inner backend (`:94-95`). Mount-time capability validation explicitly ignores legacy axes including list (`crates/ironclaw_filesystem/src/catalog.rs:150-170`), and common read/read-write mount permissions include `list` (`crates/ironclaw_host_api/src/mount.rs:33-66`). A mount that appears not to support enumeration can still enumerate names. Either return `Unsupported` from `HsmBackend::list_dir`, or advertise/validate list consistently.

Focused checks I ran on this head:

- `cargo test -p ironclaw_memory --lib --all-features` ✅
- `cargo test -p ironclaw_filesystem hsm_ --all-features -- --nocapture` ✅
- `cargo test -p ironclaw_processes filesystem_process_store_persists_under_resource_scope_engine_processes --test process_store_contract -- --exact --nocapture` ✅
- `cargo test -p ironclaw_reborn_composition invocation_mount_view --all-features -- --nocapture` ✅

These passing checks don't cover the failing scenarios above.

serrrfirat (approved):
LGTM

serrrfirat:
Deep architecture/crate hygiene review:

1. High: tenant isolation is documented but production wiring still builds singleton stores.

In `crates/ironclaw_reborn_composition/src/lib.rs`, `default_singleton_mount_view()` maps every consumer alias directly to top-level roots like `/processes`, `/secrets`, `/run-state`, `/threads`, etc. The production builders then call `wrap_scoped(...)` once and reuse that long-lived scoped filesystem for process, secret, resource, lease, run-state, and turn stores. The newer `invocation_mount_view(scope)` / `wrap_scoped_for_invocation(...)` path exists, but the production service graph does not use it.

Impact: runtime services remain shared across tenants/users despite the PR's tenant-isolation framing. Fix by making store construction request-scoped, or by carrying `ResourceScope` into every store operation and enforcing it in paths/index predicates.

2. High: production run-state/approval wiring knowingly drops an atomic contract.

`crates/ironclaw_host_runtime/src/services.rs` documents that `with_filesystem_run_state(...)` no longer carries the deleted SQL store's atomic `save_pending_and_block_approval` transition and falls back to `save_pending` followed by `block_approval`. The production builders in `ironclaw_reborn_composition` still use that path.

Impact: approval persistence can split across two records, so crash/concurrency windows can leave pending approval saved but run not blocked or related partial states. The production readiness guard also does not model combined atomic run-state/approval store as a required component. Fix with a filesystem-backed `RunStateApprovalStore` using a transaction/CAS bundle, or keep the legacy combined store until parity exists.

3. Medium: composition logic is duplicated and already drifting.

`crates/ironclaw_reborn_composition/src/factory.rs` hand-builds libSQL/Postgres production services separately from the newer substrate builders in `src/lib.rs`. The two paths construct similar stores, scoped filesystems, event-store wiring, and readiness behavior independently.

Impact: this is how contract fixes like scoped mounts, egress wiring, secret validation, or atomic stores get applied to one entrypoint but not the other. Collapse to one production builder and make `factory.rs` delegate to it.

PR #3681, opened 2026-05-15
feat:[extensions] Add first-party HTTP egress tool

serrrfirat:
## Summary

Adds a built-in first-party `builtin.http` capability for Reborn first-party tools. The tool accepts method, URL, headers, UTF-8/JSON or base64 request bodies, response limits, and timeouts, then delegates outbound calls to the existing host `RuntimeHttpEgress` path.

## Details

- Registers `builtin.http` alongside the existing built-in first-party tools.
- Passes the configured runtime HTTP egress handle into first-party dispatch requests.
- Keeps credential injection host-derived only: the tool sends an empty `credential_injections` list and cannot mint secret authority from tool input.
- Returns text responses as `body_text` and binary responses as `body_base64`.
- Adds tests for egress dispatch, fail-closed missing egress, ambiguous body inputs, zero timeout rejection, and binary response output.

## Validation

- `cargo fmt --check`
- `cargo check -p ironclaw_host_runtime --lib`
- `cargo clippy -p ironclaw_host_runtime --lib --tests -- -D warnings`
- `cargo test -p ironclaw_host_runtime --test first_party_builtin_tools`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4323,6 +4323,7 @@ name = "ironclaw_host_runtime"
 version = "0.1.0"
 dependencies = [
  "async-trait",
+ "base64 0.22.1",
  "blake3",
  "chrono",
  "chrono-tz",
--- a/crates/ironclaw_host_api/src/http.rs
+++ b/crates/ironclaw_host_api/src/http.rs
@@ -88,6 +88,17 @@ pub struct RuntimeHttpEgressResponse {
     pub redaction_applied: bool,
 }
 
+pub const RUNTIME_HTTP_REASON_RESPONSE_BODY_LIMIT_EXCEEDED: &str = "response_body_limit_exceeded";
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum RuntimeHttpEgressReasonCode {
+    CredentialUnavailable,
+    RequestDenied,
+    NetworkError,
+    ResponseError,
+    ResponseBodyLimitExceeded,
+}
+
 #[derive(Debug, Clone, PartialEq, Eq, Error)]
 pub enum RuntimeHttpEgressError {
     #[error("runtime HTTP credential error: {reason}")]
@@ -131,13 +142,30 @@ impl RuntimeHttpEgressError {
         }
     }
 
+    pub fn reason_code(&self) -> RuntimeHttpEgressReasonCode {
+        match self {
+            Self::Credential { .. } => RuntimeHttpEgressReasonCode::CredentialUnavailable,
+            Self::Request { .. } => RuntimeHttpEgressReasonCode::RequestDenied,
+            Self::Network { reason, .. } | Self::Response { reason, .. }
+                if reason == RUNTIME_HTTP_REASON_RESPONSE_BODY_LIMIT_EXCEEDED =>
+            {
+                RuntimeHttpEgressReasonCode::ResponseBodyLimitExceeded
+            }
+            Self::Network { .. } => RuntimeHttpEgressReasonCode::NetworkError,
+            Self::Response { .. } => RuntimeHttpEgressReasonCode::ResponseError,
+        }
+    }
+
     /// Stable reason token safe to expose to runtime/plugin callers.
     pub fn stable_runtime_reason(&self) -> &'static str {
-        match self {
-            Self::Credential { .. } => "credential_unavailable",
-            Self::Request { .. } => "request_denied",
-            Self::Network { .. } => "network_error",
-            Self::Response { .. } => "response_error",
+        match self.reason_code() {
+            RuntimeHttpEgressReasonCode::CredentialUnavailable => "credential_unavailable",
+            RuntimeHttpEgressReasonCode::RequestDenied => "request_denied",
+            RuntimeHttpEgressReasonCode::NetworkError => "network_error",
+            RuntimeHttpEgressReasonCode::ResponseError => "response_error",
+            RuntimeHttpEgressReasonCode::ResponseBodyLimitExceeded => {
+                RUNTIME_HTTP_REASON_RESPONSE_BODY_LIMIT_EXCEEDED
+            }
         }
     }
 }
--- a/crates/ironclaw_host_api/src/http.rs
+++ b/crates/ironclaw_host_api/src/http.rs
@@ -19,6 +19,9 @@ pub struct RuntimeHttpEgressRequest {
     pub url: String,
     pub headers: Vec<(String, String)>,
     pub body: Vec<u8>,
+    /// Request-carried fallback policy used only by legacy/test egress services.
+    /// Production first-party dispatch stages network policy in the host service
+    /// before this request is executed, so the field is ignored on that path.
     pub network_policy: NetworkPolicy,
     /// Host-derived credential injection plan.
     ///
--- a/crates/ironclaw_host_runtime/Cargo.toml
+++ b/crates/ironclaw_host_runtime/Cargo.toml
@@ -11,6 +11,7 @@ libsql = ["dep:libsql", "ironclaw_filesystem/libsql", "ironclaw_resources/libsql
 
 [dependencies]
 async-trait = "0.1"
+base64 = "0.22.1"
 blake3 = "1"
 chrono = { version = "0.4", default-features = false, features = ["clock"] }
 chrono-tz = "0.10"
--- a/crates/ironclaw_host_runtime/src/first_party.rs
+++ b/crates/ironclaw_host_runtime/src/first_party.rs
@@ -17,6 +17,10 @@ use ironclaw_host_api::{
 use serde_json::Value;
 
 /// Already-authorized first-party capability dispatch input.
+///
+/// This is host-composed first-party surface, so the struct is `non_exhaustive`:
+/// external crates may inspect fields in custom handlers but must not construct
+/// it with a struct literal.
 #[derive(Clone)]
 #[non_exhaustive]
 pub struct FirstPartyCapabilityRequest {
--- a/crates/ironclaw_host_runtime/src/first_party.rs
+++ b/crates/ironclaw_host_runtime/src/first_party.rs
@@ -12,7 +12,7 @@ use async_trait::async_trait;
 use ironclaw_filesystem::RootFilesystem;
 use ironclaw_host_api::{
     CapabilityId, MountView, ResourceEstimate, ResourceScope, ResourceUsage,
-    RuntimeDispatchErrorKind,
+    RuntimeDispatchErrorKind, RuntimeHttpEgress,
 };
 use serde_json::Value;
 
@@ -25,6 +25,7 @@ pub struct FirstPartyCapabilityRequest {
     pub estimate: ResourceEstimate,
     pub mounts: Option<MountView>,
     pub filesystem: Arc<dyn RootFilesystem>,
+    pub runtime_http_egress: Option<Arc<dyn RuntimeHttpEgress>>,
     pub input: Value,
 }
 
@@ -37,6 +38,13 @@ impl fmt::Debug for FirstPartyCapabilityRequest {
             .field("estimate", &self.estimate)
             .field("mounts", &self.mounts)
             .field("filesystem", &"<root filesystem>")
+            .field(
+                "runtime_http_egress",
+                &self
+                    .runtime_http_egress
+                    .as_ref()
+                    .map(|_| "<runtime http egress>"),
+            )
             .field("input", &self.input)
             .finish()
     }
--- a/crates/ironclaw_host_runtime/src/first_party_tools/http.rs
+++ b/crates/ironclaw_host_runtime/src/first_party_tools/http.rs
@@ -3,7 +3,8 @@ use ironclaw_extensions::{CapabilityManifest, ExtensionError};
 use ironclaw_host_api::{
     CapabilityId, EffectKind, NetworkMethod, NetworkPolicy, PermissionMode, ResourceCeiling,
     ResourceEstimate, ResourceProfile, ResourceUsage, RuntimeDispatchErrorKind,
-    RuntimeHttpEgressError, RuntimeHttpEgressRequest, RuntimeKind, SandboxQuota,
+    RuntimeHttpEgressError, RuntimeHttpEgressReasonCode, RuntimeHttpEgressRequest, RuntimeKind,
+    SandboxQuota,
 };
 use serde_json::{Map, Value, json};
 
@@ -13,7 +14,7 @@ use super::{FIRST_PARTY_MAX_OUTPUT_BYTES, input_error};
 
 pub const HTTP_CAPABILITY_ID: &str = "builtin.http";
 
-const DEFAULT_HTTP_TIMEOUT_MS: u32 = 30_000;
+const DEFAULT_HTTP_TIMEOUT_MS: u32 = 10_000;
 const MAX_HTTP_TIMEOUT_MS: u32 = 30_000;
 const DEFAULT_RESPONSE_BODY_LIMIT: u64 = 512 * 1024;
 const MAX_RESPONSE_BODY_LIMIT: u64 = 700 * 1024;
@@ -63,7 +64,7 @@ pub(super) fn manifest() -> Result<CapabilityManifest, ExtensionError> {
                     ]
                 },
                 "body": {
-                    "description": "UTF-8 string or JSON value to send as the request body"
+                    "description": "UTF-8 string or JSON value to send as the request body. No Content-Type is set automatically; pass it via headers."
                 },
                 "body_base64": {
                     "type": "string",
@@ -71,7 +72,7 @@ pub(super) fn manifest() -> Result<CapabilityManifest, ExtensionError> {
                 },
                 "response_body_limit": {
                     "type": "integer",
-                    "description": "Maximum response body bytes to return. Omit to use the built-in fail-closed default cap; values must be at least 1.",
+                    "description": "Maximum raw response body bytes to return. Omit to use the built-in fail-closed default cap; values must be at least 1. Binary responses are base64-encoded and must still fit the first-party output ceiling.",
                     "minimum": 1,
                     "maximum": MAX_RESPONSE_BODY_LIMIT
                 },
@@ -133,7 +134,12 @@ pub(super) async fn dispatch(
     };
     let response = tokio::task::spawn_blocking(move || egress.execute(http_request))
         .await
-        .map_err(|_| FirstPartyCapabilityError::new(RuntimeDispatchErrorKind::Backend))?
+        .map_err(|error| {
+            if error.is_panic() {
+                tracing::error!("first-party HTTP egress worker panicked");
+            }
+            FirstPartyCapabilityError::new(RuntimeDispatchErrorKind::Backend)
+        })?
         .map_err(http_error)?;
     let mut output = Map::new();
     output.insert("status".to_string(), json!(response.status));
@@ -218,12 +224,9 @@ fn headers(input: &Value) -> Result<Vec<(String, String)>, FirstPartyCapabilityE
 }
 
 fn header_pair(name: &str, value: &str) -> Result<(String, String), FirstPartyCapabilityError> {
-    if name.trim().is_empty()
+    if !valid_header_name(name)
         || name.len() > MAX_HTTP_HEADER_NAME_BYTES
         || value.len() > MAX_HTTP_HEADER_VALUE_BYTES
-        || name
-            .chars()
-            .any(|character| matches!(character, '\r' | '\n' | '\0'))
         || value
             .chars()
             .any(|character| matches!(character, '\r' | '\n' | '\0'))
@@ -231,6 +234,33 @@ fn header_pair(name: &str, value: &str) -> Result<(String, String), FirstPartyCa
         return Err(input_error());
     }
     Ok((name.to_string(), value.to_string()))
+}
+
+fn valid_header_name(name: &str) -> bool {
+    !name.is_empty()
+        && name.bytes().all(|byte| {
+            matches!(
+                byte,
+                b'0'..=b'9'
+                    | b'A'..=b'Z'
+                    | b'a'..=b'z'
+                    | b'!'
+                    | b'#'
+                    | b'$'
+                    | b'%'
+                    | b'&'
+                    | b'\''
+                    | b'*'
+                    | b'+'
+                    | b'-'
+                    | b'.'
+                    | b'^'
+                    | b'_'
+                    | b'`'
+                    | b'|'
+                    | b'~'
+            )
+        })
 }
 
 fn body(input: &Value) -> Result<Vec<u8>, FirstPartyCapabilityError> {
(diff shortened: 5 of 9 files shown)


zmanian:
## Summary

This PR wires a first-party `builtin.http` capability into the host runtime, delegating outbound HTTP to the existing `RuntimeHttpEgress` substrate. The boundary discipline is correct: the tool stays a thin translator and never owns transport, allowlisting, DNS, private-IP filtering, or credential injection — those remain in `HostHttpEgressService` and `crates/ironclaw_network/`. Credential authority is explicitly surrendered (`credential_injections: Vec::new()`), which matches the documented invariant on `RuntimeHttpEgressRequest::credential_injections` that tool input must not mint credential authority. Manifest declares `default_permission: Ask`, both `DispatchCapability` and `Network` effects, and resource ceilings — fail-closed defaults are in place. Tests cover the missing-egress path, ambiguous-body input, zero timeout, binary response handling, and request-byte accounting on `OutputTooLarge`.

## Strengths

- **Tool stays a translator.** No URL parsing, scheme allowlisting, DNS, or private-IP work duplicated locally. Policy resolution remains in `HostHttpEgressService::network_policy_for_request` (`StagedObligation` in production, fallback for tests). Setting `network_policy: NetworkPolicy::default()` on the outbound request is therefore safe — the staged-obligation path ignores it and the fallback path is test-only.
- **Credential authority correctly surrendered.** Tool input cannot smuggle a `RuntimeCredentialInjection`; sensitive request headers (`authorization`, `cookie`, `x-api-key`, etc.) are already rejected by `validate_runtime_request` upstream and the `manual_credentials_denied` scan catches credential-shaped header/URL content.
- **Header CRLF guard.** `header_pair` rejects `\r`/`\n` in name and value — blocks request smuggling/splitting before the host validator. Empty/whitespace name is also rejected.
- **Ambiguous body input rejected.** Sending both `body` and `body_base64` returns `InputEncode` rather than picking one silently. Base64 errors fail closed.
- **Accounting invariant preserved.** `network_egress_bytes` is request-only on success and is also stamped onto the `ResourceUsage` when `bounded_output_bytes` rejects the response — the user is charged for the egress that actually went out (covered by `builtin_http_accounts_request_bytes_when_output_is_too_large`).
- **Resource ceiling is bounded.** Timeout hard-capped at 30s, response body at 700 KB, output ≤ `FIRST_PARTY_MAX_OUTPUT_BYTES` (1 MB).
- **Test fixture is faithful.** `RecordingRuntimeHttpEgress` exercises the full first-party dispatcher (not just the helper), aligning with the "test through the caller" rule.

## Issues

### Major

1. **All `RuntimeHttpEgressError` variants collapse to `NetworkDenied`.** `http_error` maps `Credential`, `Request`, `Network`, and `Response` indiscriminately to `RuntimeDispatchErrorKind::NetworkDenied`. This destroys audit/debug fidelity that already exists upstream:
   - `Credential { reason }` (e.g., required injection unavailable) — caller has no signal to distinguish from a network failure.
   - `Request { reason: "sensitive_header_denied:..." | "manual_credentials_denied" }` — these are *input* failures and should surface as `InputEncode` or a dedicated `Client` kind, not `NetworkDenied`.
   - `Response { reason }` — host-side response-bytes-too-large/decode failures should map to `OutputTooLarge` / `OutputDecode` (these dispatch kinds already exist) so caller-side recovery logic can branch.

   The `stable_runtime_reason()` token defined on `RuntimeHttpEgressError` is also discarded — it should be surfaced into the audit channel (e.g., the dispatch error message), otherwise the operator loses the only stable signal the substrate produces. Recommend a small `match` over the four variants mapping to the closest existing `RuntimeDispatchErrorKind` rather than blanket `NetworkDenied`.

2. **`response_body_limit` `minimum: 0` in schema permits a zero limit.** `ranged_u64` accepts `0` and forwards `Some(0)` to the egress service. Whether the network layer treats `Some(0)` as "no body allowed" or "no limit" is non-obvious. Either explicitly forbid `0` in `response_body_limit` (mirroring how `timeout_ms` requires `>= 1`) or document the semantics in the manifest description so callers don't ship requests that surprise-fail. Recommend `minimum: 1` and use the existing `ranged_u64` path with min = 1.

3. **Response-body decoding charges base64 to `body_base64` even when content is text but contains a stray invalid byte.** The `from_utf8` fallback is binary all-or-nothing. For mostly-text responses with a single invalid byte, the entire payload becomes base64 — surprising for callers grepping logs. Consider `String::from_utf8_lossy` for a `body_text` (with a `body_text_lossy: true` marker) or document the all-or-nothing contract in the manifest description. Minor in practice — most APIs are clean UTF-8 or genuine binary — but flagging because there's no test for the mixed-content case.

### Minor

4. **`sandbox.network_egress_bytes` ceiling is 1 MB but `DEFAULT_NETWORK_EGRESS_BYTES` estimate is 16 KB.** The ceiling allows a ~64× overrun before the resource governor steps in. Given request bodies route through `bounded_input_size` (`MAX_FIRST_PARTY_INPUT_BYTES = 1 MB` of *JSON*), a more conservative sandbox ceiling (e.g., 256 KB) would tighten the blast radius for runaway request bodies without affecting realistic use. The current 1 MB matches `FIRST_PARTY_MAX_OUTPUT_BYTES` but the two limits cover different bytes (request vs. response).

5. **No upper bound on `headers` count or individual header length.** Only individual-pair CRLF and emptiness checks. A caller can submit thousands of headers totaling close to 1 MB through the input cap. Downstream `reqwest` will eventually error, but pre-validation at the tool layer (e.g., max 64 headers, max 8 KB per value) would fail closer to the cause and harmonize with the explicit ceilings used elsewhere.

6. **`http_error` always reports `network_egress_bytes` even for `Credential` errors.** `RuntimeHttpEgressError::request_bytes()` returns 0 for the credential variant, so this is currently harmless — but the unconditional `with_usage` invocation is brittle if `request_bytes()` ever starts returning a non-zero value for credential errors (e.g., when staged credentials get partially consumed). Either special-case or trust the upstream contract and add a comment.

7. **`response_body_limit` returning `Some(default)` is sent even when the caller didn't ask.** This is the desired fail-closed shape (always cap) — worth a one-line comment noting the intent so future readers don't "simplify" it to `None`.

8. **Schema declares `headers` accepts either array-of-`{name,value}` or object form.** The dual shape exists for ergonomics, but the array form supports duplicate header names while the object form does not. The behavioral difference (duplicate `Set-Cookie`-style headers in requests is rarely needed but legal) deserves a one-line manifest description note.

## Suggestions

- Consider exposing the substrate's `redaction_applied` flag in the failure path too, so callers can tell whether a denied response had leak-detector rewrites applied. Currently surfaced only on success.
- Add a regression test for the "sensitive request header rejected by host" path — call the tool with `headers: {"authorization": "..."}` and assert it fails with a specific kind. Today the rejection lives in `validate_runtime_request` and the tool has no direct test asserting that path triggers when invoked through the first-party dispatcher. (This is the kind of test-through-the-caller coverage that catches integration drift.)
- A doctest or comment on `dispatch` pointing readers at the substrate (`HostHttpEgressService`, `crates/ironclaw_outbound/`) for "where is the actual security?" would help future maintainers resist the temptation to add local URL parsing here.

## Verdict

**COMMENT** — substantively close to APPROVE. The boundary discipline and credential-authority handling are correct; the substrate-trust pattern is the right one and the tests exercise the dispatcher (not just helpers). The major items above are about error-mapping fidelity (issue 1) and a small fail-closed gap on `response_body_limit: 0` (issue 2) — both addressable in this PR without restructuring. Recommend landing once at minimum issue 1 (error-kind mapping) and issue 2 (response_body_limit minimum) are resolved.

henrypark133:
## Caveman review findings

**2🟠 1🟡**

- 🟠 `crates/ironclaw_host_runtime/src/first_party_tools/http.rs` (~line 187, `dispatch`): `RuntimeHttpEgress::execute` is a blocking synchronous call invoked directly from `async fn dispatch` without `spawn_blocking` or `block_in_place`. With up to 30 s timeout, this stalls the tokio worker thread for the full duration, degrading concurrency under any non-trivial load. Fix: wrap in `tokio::task::spawn_blocking`, or make `RuntimeHttpEgress::execute` async.
- 🟠 `http.rs` (~line 179): `network_policy: NetworkPolicy::default()` embeds `deny_private_ip_ranges: false` and empty `allowed_targets`. In production `StagedObligation` mode this is ignored, but all 5 new tests use `RecordingRuntimeHttpEgress` which bypasses `HostHttpEgressService` entirely — no test exercises real SSRF policy (private-IP blocking, allowlist matching) for `builtin.http`. Add an integration test that exercises a real `PolicyNetworkHttpEgress` rejection.
- 🟡 `http.rs` `header_pair` (~lines 268–273): rejects `\r` and `\n` in header names/values but not `\0` (null byte). Null in a header name passes this check and reaches reqwest, which rejects it with a confusing generic `NetworkDenied`. Fix: add `'\0'` to the rejection match arm alongside `\r | \n`.

serrrfirat:
@zmanian @henrypark133 addressed review feedback in de8f72f30:

- Mapped RuntimeHttpEgressError variants to more specific dispatch kinds and logged stable runtime reason tokens.
- Rejected response_body_limit=0 and documented fail-closed default cap behavior.
- Moved blocking RuntimeHttpEgress::execute behind spawn_blocking.
- Added header NUL/count/size validation and documented duplicate-header behavior.
- Added dispatcher-level tests for sensitive-header rejection, real private-IP policy rejection, mixed UTF-8 binary response contract, error mapping, and blocking-egress behavior.
- Tightened builtin.http network egress sandbox ceiling.

Please take another look when you have a chance.

ilblackdragon:
## Code Review

### Overview
Adds a new built-in first-party capability `builtin.http` to `ironclaw_host_runtime`. The handler is a thin translator: it validates input, builds a `RuntimeHttpEgressRequest`, and delegates to the existing `RuntimeHttpEgress` host service. URL parsing, DNS, private-IP enforcement, allowlists, transport, and credential injection remain in `HostHttpEgressService` / `ironclaw_network`.

### Strengths

- **Clear separation of concerns.** The "translator only" comment in `first_party_tools/http.rs` is honored by the implementation.
- **Fail-closed defaults.** `credential_injections: Vec::new()` is hard-coded so the tool can't mint authority from input. Missing `RuntimeHttpEgress` returns `NetworkDenied` (test: `builtin_http_fails_closed_without_runtime_egress`).
- **Bounded everywhere.** Method whitelist; header count (64), name (512 B), value (8 KiB); response body (700 KiB); egress (256 KiB); timeout (30 s). `response_body_limit` is always forwarded so the transport never inherits an unbounded cap.
- **Header injection guard.** Rejects `\r`, `\n`, `\0` in header names and values.
- **Ambiguity → reject.** Supplying both `body` and `body_base64` errors out.
- **Resource accounting on failure paths.** `wall_clock_ms` and `network_egress_bytes` are charged even when `bounded_output_bytes` rejects the response.
- **Blocking egress off the tokio worker.** `spawn_blocking` is validated by `builtin_http_runs_blocking_egress_off_tokio_worker`.
- **Tests through the caller.** All cases go through `invoke_with_context`, in line with `.claude/rules/testing.md` ("Test Through the Caller"). Coverage spans happy path, missing egress, ambiguous body, zero timeout/limit, header injection, oversized header sets, six error-source mappings, sensitive-header rejection via the real `PolicyNetworkHttpEgress`, and private-IP rejection.

### Issues / Suggestions

#### 1. Fragile substring matching for error classification (medium)
`response_body_limit_reason` and `response_error_kind` match on raw reason strings:

```rust
reason.contains("response_body_limit") || reason.contains("body_limit")
...
if response_body_limit_reason(reason) || reason.contains("too_large") { ... }
```

These rely on `RuntimeHttpEgressError::stable_runtime_reason()` strings staying stable. Risks:
- A typo or rename in `ironclaw_network`/`ironclaw_host_api` will silently re-classify `OutputTooLarge` as `OutputDecode` / `NetworkDenied`.
- `"body_limit"` is a substring of `"response_body_limit"`, so the second clause is dead — a sign of guess-and-check matching.

Consider a structured `RuntimeHttpEgressError::reason_code()` returning an enum, or extract stable reason constants in `ironclaw_host_api` that both sides import.

#### 2. Public-field addition to `FirstPartyCapabilityRequest` is a breaking change (low)
`first_party.rs` adds `runtime_http_egress: Option<Arc<dyn RuntimeHttpEgress>>` to a `pub struct` with all-public fields. Any external constructor of this struct now fails to compile. If it's intended to be host-internal only, mark fields `pub(crate)` and expose a builder; otherwise document the surface change.

#### 3. `DEFAULT_HTTP_TIMEOUT_MS == MAX_HTTP_TIMEOUT_MS` (low)
Both are `30_000`. That makes the default the same as the cap, so any explicit value <= 30 s never extends headroom. If the intent is "30 s is the absolute ceiling," `DEFAULT_HTTP_TIMEOUT_MS` should be lower (e.g. 10 s) so a misbehaving call doesn't have to wait the full 30 s. Worth a comment either way.

#### 4. Response-body cap vs. base64 expansion (low — possible footgun)
`MAX_RESPONSE_BODY_LIMIT = 700 * 1024` bytes raw. When the response isn't valid UTF-8 it is base64-encoded and the base64 string is placed in the JSON output. base64 expands ~4/3, so a 700 KiB binary body becomes ~933 KiB, which then runs `bounded_output_bytes` against `FIRST_PARTY_MAX_OUTPUT_BYTES`. If the latter is near 700 KiB, callers requesting the max binary response will predictably fail at the output-bounds check. Either shrink the response-body cap so the worst-case base64 still fits, or document the constraint in the schema description.

#### 5. `JoinError` collapses panic and cancellation (low)
```rust
.await.map_err(|_| FirstPartyCapabilityError::new(RuntimeDispatchErrorKind::Backend))?
```
A `spawn_blocking` `JoinError` distinguishes panic vs. cancellation. Today both become `Backend`; if the panic case should propagate to telemetry differently, branch on `err.is_panic()` and log.

#### 6. Mixed-UTF-8 test name is slightly misleading (nit)
`builtin_http_documents_mixed_utf8_response_as_binary` tests that the entire invalid response (including the valid prefix) is base64-encoded. Add a one-line comment explaining *why* the response isn't split into UTF-8 vs binary chunks (consistency, single output field). Otherwise a future reader may "improve" it.

#### 7. JSON body content-type is the caller's responsibility (nit)
When `body` is a JSON object, the bytes are `serde_json::to_vec(value)` but no `Content-Type: application/json` header is auto-injected. That's a defensible choice, but the schema description doesn't warn callers. Add one line: "no `Content-Type` is set automatically — pass it via `headers`."

### Project-conventions check
- `cargo fmt` / `cargo clippy -- -D warnings` per PR description ✓
- `unwrap()` only in tests ✓
- Comments are sparse and only on non-obvious behavior ✓
- "Test Through the Caller" rule honored ✓

### Risk summary
Low-to-medium. The change is well-scoped and the threat model around credentials/headers/private IPs is enforced both at the tool boundary and re-checked at the host transport. The two items I'd most want addressed before merging are (1) brittle reason-string matching for error classification, and (4) the base64 / `FIRST_PARTY_MAX_OUTPUT_BYTES` interaction.

serrrfirat:
@ilblackdragon addressed the review findings in 1137b2763:

- Replaced brittle response reason substring checks with structured `RuntimeHttpEgressReasonCode` and exact response-body-limit classification.
- Lowered default HTTP timeout to 10s while keeping 30s hard cap.
- Documented `response_body_limit` binary/base64 output-ceiling behavior and caller-owned `Content-Type`.
- Logged `spawn_blocking` panics distinctly.
- Added mixed UTF-8 test comment and tightened header-name validation with regression cases.
- Documented `FirstPartyCapabilityRequest` as `non_exhaustive` host-composed surface.

Validation run:
- `cargo fmt --check`
- `cargo check -p ironclaw_host_api -p ironclaw_host_runtime --tests`
- `cargo test -p ironclaw_host_runtime --test first_party_builtin_tools builtin_http_ -- --nocapture`
- `cargo clippy -p ironclaw_host_api -p ironclaw_host_runtime --tests --all-features -- -D warnings`
- `cargo test -p ironclaw_host_api -p ironclaw_host_runtime --lib`

Note: `cargo test --lib` hit the local MCP 120s timeout, and `scripts/pre-commit-safety.sh` still reports pre-existing PR-wide findings outside these touched files.

zmanian (commented):
## Review

First, the title/description mismatch worth flagging: my own description said "refactor(reborn): project ProductAdapter from single ExtensionManifestV2", but the actual diff is the `builtin.http` first-party egress tool. None of the files in this PR touch `ProductAdapter` or `ExtensionManifestV2` projection. If there was an intent to bundle a ProductAdapter refactor here, it's missing; otherwise the GitHub title is correct and my description was stale.

The HTTP tool itself is the security-load-bearing part of this change, so the rest of this review focuses there.

### What's solid

- Translator-only handler. The comment at the top of `dispatch` is the right invariant: URL parsing, DNS / private-IP enforcement, allowlist, transport, redaction all stay in `HostHttpEgressService` / `ironclaw_network`. The tool is a thin shaping layer over an already-hardened path.
- `credential_injections: Vec::new()` is correct for v1. The host must remain the sole authority for minting credentials; tool input cannot smuggle in `SecretRef`s. The Gemini bot's suggestion to let callers specify "authorized secret handles" should be deferred — when we add it, the obligation/grant binding has to be host-derived, not caller-derived, or we've reopened the door this PR is intentionally closing.
- Tightened header-name validation to the RFC 9110 token set, plus header count (64), name length (512B), value length (8KB), and CR/LF/NUL rejection. The `builtin_http_rejects_invalid_header_names_and_oversized_header_sets` test covers the bad-input matrix well.
- Real-policy SSRF coverage: `builtin_http_exercises_real_policy_private_ip_rejection` and `builtin_http_rejects_sensitive_headers_through_host_validator` both drive `PolicyNetworkHttpEgress` end-to-end rather than mocking around it. Good defense against regressions where someone "simplifies" the trait wiring and the policy stops getting consulted.
- `spawn_blocking` around `egress.execute`, plus the `builtin_http_runs_blocking_egress_off_tokio_worker` regression test. Right call.
- Binary-vs-text response disambiguation is conservative: any invalid UTF-8 byte returns the entire body as `body_base64`. No lossy splits.

### Concerns

1. **`network_policy: NetworkPolicy::default()` in the dispatch request is misleading.** Copilot already flagged this. Today the handler appears to work because policy is sourced from the grant/execution context elsewhere in `HostHttpEgressService`. But the field on `RuntimeHttpEgressRequest` is exactly the wrong knob to leave at "deny-all default" — if a future refactor of `HostHttpEgressService` ever falls back to `request.network_policy`, this tool becomes silently allow-nothing or, worse, silently allow-something-unintended depending on which side of the merge wins. Either populate it with the staged policy for this scope/capability, or remove the field entirely from this code path and document that it is ignored for first-party dispatch. Leaving a dead field with a security-relevant default is a footgun.

2. **No explicit cap on request body size in the handler.** `MAX_NETWORK_EGRESS_BYTES = 256KB` lives on the `SandboxQuota` and is enforced at accounting time, but the `body` / `body_base64` validators don't reject oversized inputs before we hand them to `spawn_blocking`. For a `body_base64` payload near 700KB this means a base64 decode allocation in the input path. Not exploitable today (input is already bounded by `bounded_input_size`), but worth either (a) explicitly capping the decoded body length at `MAX_NETWORK_EGRESS_BYTES`, or (b) adding a test that demonstrates the input-size ceiling rejects oversized bodies before transport is engaged.

3. **Redirect handling is invisible in this PR.** The classic SSRF-via-redirect (HTTPS → `http://169.254.169.254/`) is handled in `ironclaw_network` if `PolicyNetworkHttpEgress` re-resolves and re-checks `deny_private_ip_ranges` on each hop. That contract is not exercised by any test in this PR. Please add a test (or point to one in `ironclaw_network`) that proves a 30x to a private IP is rejected. If redirects aren't followed at all, document that in the capability description so callers don't assume they are.

4. **Response-size accounting vs. base64 expansion.** `MAX_RESPONSE_BODY_LIMIT = 700KB` is the raw response cap, but binary responses get base64-encoded and counted against `FIRST_PARTY_MAX_OUTPUT_BYTES`. A 700KB binary body expands to ~933KB encoded. If `FIRST_PARTY_MAX_OUTPUT_BYTES` is at or below ~933KB, large binary responses will fetch successfully and then trip `OutputTooLarge` at the very end, burning the full round-trip cost. `builtin_http_accounts_request_bytes_when_output_is_too_large` confirms this is the chosen behavior — fine, but the doc string on `response_body_limit` should call it out: "binary responses are base64-encoded, raise effective cost ~33%, and must still fit the first-party output ceiling."

5. **No `Content-Type` defaulting.** Schema description says "No Content-Type is set automatically; pass it via headers." Good that it's documented, but in practice every JSON-body caller will forget this. Consider a follow-up to default `Content-Type: application/json` when `body` is a non-string JSON value and no `content-type` header was supplied. Not blocking.

6. **`http_error` mapping for `RuntimeHttpEgressError::Credential`** maps to `RuntimeDispatchErrorKind::Client` in the handler, but the test at line 886 asserts it maps to `RuntimeFailureKind::Backend`. Either there's a translation between `Client` and `Backend` at a higher layer, or one of these is wrong. Worth double-checking — credential-unavailable shouldn't surface to the LLM as a user-input error.

### Substrate / binary boundary

No `ProductAdapter` work in this PR, so the substrate/binary concern doesn't apply here. The new code lives in `ironclaw_host_runtime` and depends only on `ironclaw_host_api`, `ironclaw_extensions`, `ironclaw_network` — all substrate-side. Clean.

### Verdict

Approve in principle once #1 (dead `network_policy` field) is either populated or explicitly removed, and #3 (redirect SSRF coverage) is either tested here or pointed to in the network crate. The remaining items are non-blocking follow-ups.

— Zaki

serrrfirat:
@ilblackdragon addressed review 4304829539 in afc8d01c3:

- Documented `RuntimeHttpEgressRequest.network_policy` as request-policy fallback only; first-party dispatch uses a named staged-policy placeholder with inline invariant comment.
- Added explicit request body cap at `MAX_NETWORK_EGRESS_BYTES` for both `body` and `body_base64`, with caller-through regression coverage that egress is not invoked.
- Documented redirect behavior in the capability description and added coverage that 30x responses are returned without following a private `Location`.
- Expanded `response_body_limit` docs to call out base64 ~33% output expansion and first-party output ceiling interaction.
- Defaulted `Content-Type: application/json` for non-string JSON bodies when caller did not supply content-type.
- Clarified credential egress errors: handler emits `RuntimeDispatchErrorKind::Client`, which production maps to `RuntimeFailureKind::Backend`.

Validation run:
- `cargo fmt --check`
- `cargo test -p ironclaw_host_runtime --test first_party_builtin_tools builtin_http_ -- --nocapture`
- `cargo check -p ironclaw_host_api -p ironclaw_host_runtime --tests`
- `cargo clippy -p ironclaw_host_api -p ironclaw_host_runtime --tests --all-features -- -D warnings`
- `cargo test -p ironclaw_host_api -p ironclaw_host_runtime --lib`

Note: `scripts/pre-commit-safety.sh` still reports existing PR-wide findings outside these touched files.

serrrfirat:
@ilblackdragon follow-up review fixes are pushed in afc8d01c3. Addressed network policy footgun docs, explicit request body cap, redirect behavior/test, base64 output-cost docs, JSON Content-Type defaulting, and credential-error mapping clarification. Targeted fmt/check/clippy/tests passed locally; CI is running.

zmanian (commented):
## Re-review (commit `afc8d01c`)

Walked the single new commit since my last pass against all six findings.

### Status of prior findings

1. **Dead `network_policy` field (footgun)** — Addressed via documentation, not removal. `NetworkPolicy::default()` is now wrapped in `staged_policy_placeholder()` with a comment, and the field on `RuntimeHttpEgressRequest` got a doc-comment in `ironclaw_host_api/src/http.rs` clarifying it's a fallback for legacy/test egress services. Production stages policy in `HostHttpEgressService`. Acceptable — the wire value is unchanged but the contract is now legible. The footgun is reduced to "future contributor must read the doc" rather than "future contributor stares at `Default::default()` and assumes deny-all is intended."
2. **Request-body size cap before `spawn_blocking`** — Addressed. `body()` now rejects with `InvalidInput` when decoded bytes exceed `MAX_NETWORK_EGRESS_BYTES` (256 KiB), and `builtin_http_rejects_request_bodies_over_network_egress_cap` covers both `body` and `body_base64` paths. Minor: the check runs *after* base64 decode, so an attacker-controlled ~341 KiB base64 string still allocates a ~256 KiB+1 buffer before rejection. That's bounded by `bounded_input_size` upstream, so not exploitable; flagging as a follow-up only.
3. **Redirect SSRF coverage** — Addressed. `builtin_http_returns_redirects_without_following_private_location` proves the host transport returns the 302 untouched rather than auto-following to `169.254.169.254`. Manifest description now states "the host transport does not follow them." This is the right contract — auto-follow is what makes redirect SSRF dangerous, and not following pushes the decision to the caller.
4. **Base64 expansion vs `FIRST_PARTY_MAX_OUTPUT_BYTES`** — Addressed. `response_body_limit` description now includes "raise effective output cost by about 33%". Sufficient.
5. **`Credential -> Client` vs test asserting `Backend`** — Addressed via inline comment explaining the higher-layer translation. Code unchanged. Fine — the mapping is now self-documenting at the call site.
6. **`Content-Type` defaulting** — Implemented. Non-string JSON bodies (`Array`/`Bool`/`Number`/`Object`) get `content-type: application/json` injected when the caller didn't supply one, with `eq_ignore_ascii_case` matching. Test `builtin_http_defaults_json_body_content_type` covers it. The synthesized header bypasses `header_pair` validation, but both name and value are constants under our control — non-issue.

### Verdict

Approve. All blocking items from the prior review are resolved or explicitly documented. The two residual nits (decode-then-reject ordering for oversized base64 bodies, and `staged_policy_placeholder()` still returning `default()`) are non-blocking follow-ups; the second only matters if someone wires production through `new_with_request_policy_for_tests`, which the doc comment now warns against.

Nice work on the redirect test — that's exactly the contract I wanted nailed down.

— Zaki

Issue #3533, opened 2026-05-12
[QA] Telegram in v 0.28.1 does not automatically setup from UI


joe-rlo:
### Environment

hosted-staging (crab shack)

### Version / Commit Hash

0.28.1

### QA Test Date

2026-05-12

### Feature Area

Telegram pairing

### Steps to Reproduce

See images but the directions are not up-to-date, is asking user to complete other steps where it used to setup and just prompt for the token.

<img width="1152" height="1128" alt="Image" src="https://github.com/user-attachments/assets/53ca99a5-a181-4324-b96e-28c086a571a8" />
<img width="1161" height="1082" alt="Image" src="https://github.com/user-attachments/assets/be1c4dfc-f40d-414f-b54c-1d6a6844b0f6" />

### Expected Behavior

"connect my telegram" should setup and then prompt user for token

### Actual Behavior

See steps to reproduce

### Pre-submit checklist

- [x] Title is specific (not "fix Google Suite" but "Google Suite install throws invalid_grant on OAuth step")
- [x] Commit hash is filled in
- [x] Steps are numbered and reproducible

joe-rlo:
logs are sparse:
`16:04:12.889
INFO
ironclaw::channels::web::features::logs
Log level changed to 'debug' user_id=default
14:13:09.519
ERROR
ironclaw_tui::app
TUI error: No such device or address (os error 6)
14:13:08.525
INFO
ironclaw::orchestrator::api
Orchestrator internal API listening on 0.0.0.0:50051
14:13:08.525
INFO
ironclaw::orchestrator
Docker is available
14:13:08.022
INFO
ironclaw::registry::catalog
Loaded embedded registry catalog (27 extensions, 3 bundles)
14:13:08.021
WARN
ironclaw::app
MCP server 'nearai' rejected its configured Authorization header. Update the configured credential and try again.
14:13:07.723
INFO
ironclaw::app
bootstrap_ownership: owner user ensured, default rows migrated owner_id=default
14:13:07.716
INFO
ironclaw::db::libsql_migrations
libSQL: applied 14 incremental migrations
14:13:09.519
ERROR
ironclaw_tui::app
TUI error: No such device or address (os error 6)
14:13:08.525
INFO
ironclaw::orchestrator::api
Orchestrator internal API listening on 0.0.0.0:50051
14:13:08.525
INFO
ironclaw::orchestrator
Docker is available
14:13:08.022
INFO
ironclaw::registry::catalog
[...]

PR #3682, opened 2026-05-15
fix(canary): accurate test counts, chat-install probe, strict xfails

nickpismenkov:
## Summary

Three related fixes that came out of digging into why the May 8–13 chat-driven `tool_install` regression (PR #3366, fixed in #3533/#3559) slipped past every canary cycle for five days.

- **Lane counts are real now.** `notify_slack.py` was reporting `tests=0 passed=0 failed=0 status=unknown` for six of twelve lanes (every cargo lane plus the summaries-only `private-oauth` lane) because it only knew how to count from JUnit XML or workflow-canary's `results.json`. New `emit_results_json.py` scrapes cargo `--nocapture` output into the same schema, the `run.sh` `finish` trap wires it in, and `notify_slack.py` falls back to the `summary.md` exit code when no counts exist. The `private-oauth` job also gets a preflight disk-space check on the self-hosted runner (no current cleanup logic).
- **New `tool_install_chat` canary probe.** Drives the LLM-callable install path through the existing mock LLM `check gmail unread` canned flow and asserts (1) chat-send returns 202, (2) `gmail` reaches `installed=true` within 60s, (3) a `tool_install` invocation appears in history, (4) no forbidden error fragments. Would have failed loudly on 2026-05-08 instead of nothing failing for five days.
- **`xfail_strict = true` globally.** Two stale `xfail(strict=False)` markers in `test_v2_auth_oauth_matrix.py` were the closest coverage we had for the broken surface; with `strict=False`, both XFAIL and XPASS silently count as pass, so the mark was documentation, not an assertion. Unxfailed in #3589. Flipping the global default is policy-only — current tests don't change behavior (the remaining xfails in `test_channel_approval_gates.py` are already explicit `strict=True`), but future xfails default to strict.

## Test plan

- [x] Dry-ran the updated `notify_slack.py` against the merged artifact tree from canary run `25894532227`. The two real failures (`public-smoke/anthropic`, `provider-matrix/anthropic`) now report `tests=2 passed=1 failed=1 status=fail` with panic messages captured; slack header reads `10 passed, 2 failed of 12 lanes` instead of leaving 6 lanes as `unknown`.
- [x] Strict-scrub round trip on `results.json`, including a synthesized panic message containing `sk-ant-…` — file preserved, token redacted.
- [x] `python3 -m py_compile` on `emit_results_json.py`, `notify_slack.py`, `tool_install_chat.py`, `run_workflow_canary.py`. `bash -n run.sh`. `tomllib` parse on `tests/e2e/pyproject.toml`.
- [x] `SCENARIOS` dict verified: `tool_install_chat` is registered immediately after `auth_recovery` and before `log_assertions` (which must remain last).
- [x] Audit confirms zero implicit-strict xfails anywhere in `tests/` — the three remaining markers all use explicit `strict=True`.
- [ ] Next live canary cycle (scheduled cron) produces the expected per-lane counts in slack.
- [ ] First `workflow-canary` run with the new probe registered runs `tool_install_chat` end-to-end against the mock LLM stack.

### DIFF
--- a/.github/workflows/live-canary.yml
+++ b/.github/workflows/live-canary.yml
@@ -504,6 +504,19 @@ jobs:
       LLM_BACKEND: anthropic
       ANTHROPIC_API_KEY: ${{ secrets.ANTHROPIC_API_KEY }}
       ANTHROPIC_MODEL: ${{ vars.LIVE_ANTHROPIC_MODEL || 'claude-sonnet-4-6' }}
+      # Real third-party credentials forwarded into the persona harness.
+      # The harness reads these env vars per
+      # tests/e2e_live_personas.rs::PERSONA_CREDENTIALS — present →
+      # real API call lands; absent → dummy fallback satisfies the
+      # auth pre-flight and the agent recovers via workspace tools.
+      # Add each secret in repo settings (Settings → Secrets and
+      # variables → Actions) if you want the lane to exercise that
+      # integration end-to-end.
+      LIVE_CANARY_GITHUB_TOKEN: ${{ secrets.LIVE_CANARY_GITHUB_TOKEN }}
+      LIVE_CANARY_GOOGLE_OAUTH_TOKEN: ${{ secrets.LIVE_CANARY_GOOGLE_OAUTH_TOKEN }}
+      LIVE_CANARY_SLACK_BOT_TOKEN: ${{ secrets.LIVE_CANARY_SLACK_BOT_TOKEN }}
+      LIVE_CANARY_TELEGRAM_BOT_TOKEN: ${{ secrets.LIVE_CANARY_TELEGRAM_BOT_TOKEN }}
+      LIVE_CANARY_COMPOSIO_API_KEY: ${{ secrets.LIVE_CANARY_COMPOSIO_API_KEY }}
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
         with:
--- a/.github/workflows/live-canary.yml
+++ b/.github/workflows/live-canary.yml
@@ -550,6 +550,21 @@ jobs:
       COMMAND_TIMEOUT: 60m
       STRICT_ARTIFACT_SCRUB: "true"
     steps:
+      - name: Preflight disk space
+        # The only self-hosted runner the canary uses. Nothing in this
+        # workflow cleans ~/.cargo, ~/.rustup, or runner _diag between
+        # jobs, so a wedged disk shows up as opaque mid-build failures
+        # we can't see (test-output.log isn't uploaded). Fail loudly
+        # with a clear marker instead — at least 10 GiB free on /.
+        run: |
+          df -h /
+          avail_kb=$(df --output=avail / | tail -1 | tr -d ' ')
+          min_kb=$((10 * 1024 * 1024))
+          if [ "$avail_kb" -lt "$min_kb" ]; then
+            echo "::error::Self-hosted runner low on disk: ${avail_kb} KiB available, need ${min_kb} KiB (~10 GiB)."
+            exit 1
+          fi
+          echo "Disk OK: ${avail_kb} KiB available on /."
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
         with:
           persist-credentials: false
@@ -570,11 +585,16 @@ jobs:
         uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4
         with:
           name: live-canary-private-oauth-summary
+          # results.json is emitted by scripts/live-canary/emit_results_json.py
+          # and contains only test names + outcomes + bounded panic messages
+          # — no URLs, tokens, or payloads — so it is scrub-safe to upload
+          # even under STRICT_ARTIFACT_SCRUB=true.
           path: |
             artifacts/live-canary/**/summary.md
             artifacts/live-canary/**/env-summary.txt
             artifacts/live-canary/**/trace-fixture-status.txt
             artifacts/live-canary/**/scrub-matches.txt
+            artifacts/live-canary/**/results.json
           retention-days: 7
 
   provider-matrix:
--- a/scripts/live-canary/emit_results_json.py
+++ b/scripts/live-canary/emit_results_json.py
@@ -1,0 +1,225 @@
+#!/usr/bin/env python3
+"""Scrape `cargo test --nocapture` output into a results.json file that
+notify_slack.py::parse_results_json can already consume.
+
+Invoked from scripts/live-canary/run.sh at the end of every cargo-based
+lane. Skipped (early-exits as no-op) when:
+
+  * --out already exists — workflow-canary writes its own results.json
+    via scripts/workflow_canary/run_workflow_canary.py, and that file
+    must not be overwritten.
+  * The log contains no `test result:` line — auth-* lanes use pytest +
+    JUnit XML and produce no cargo-style output, so there is nothing to
+    scrape.
+
+Schema matches the workflow-canary contract (see parse_results_json):
+
+    {"results": [
+        {"provider": "...", "mode": "<test_name>",
+         "success": bool, "latency_ms": int,
+         "details": {"error": "<short panic msg>"}},
+        ...
+    ]}
+
+We only emit one entry per executed test (cargo `ok` or `FAILED`).
+`ignored` tests are not results — they didn't run — so they are skipped
+entirely.
+
+Per-test latency is unknowable from cargo's plain stdout, so we leave
+``latency_ms`` at 0 and put the lane-level wall-clock duration on a
+single ``meta`` entry consumed only by future tooling. notify_slack.py
+sums ``latency_ms`` per entry so the 0 values are harmless — lane-level
+duration is captured by the workflow itself.
+"""
+
+from __future__ import annotations
+
+import argparse
+import json
+import re
+import sys
+from pathlib import Path
+
+# Cargo test status line — emitted by cargo at the end of each test
+# binary invocation, after all per-test output. Its presence is also our
+# gate for "is this a cargo lane".
+#
+#   test result: FAILED. 1 passed; 1 failed; 0 ignored; 0 measured; 14 filtered out; finished in 236.39s
+#   test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 12.34s
+RESULT_RE = re.compile(
+    r"test result: (?P<outcome>ok|FAILED)\. "
+    r"(?P<passed>\d+) passed; "
+    r"(?P<failed>\d+) failed; "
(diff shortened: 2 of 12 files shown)


zmanian (commented):
## Review

Solid post-mortem PR. The three fixes are independently justified and traceable back to a real, specific bug (the May 8 `tool_install` regression). A few concerns worth addressing before this is the last word on canary fidelity.

### `xfail_strict = true` — is CI actually going to fail?

The PR body claims the policy is "policy-only" because the only remaining xfails in `test_channel_approval_gates.py` already use `strict=True`. That's fine *today*, but two things to confirm so the guardrail isn't decorative:

1. The `pyproject.toml` `xfail_strict = true` only fires when pytest is invoked from `tests/e2e/` (where the `pyproject.toml` lives) and not via a `pytest.ini` at the repo root that would take precedence. The PR doesn't show a CI invocation that proves the new default reaches the test runner. Worth eyeballing the `auth-canary` workflow steps to confirm `cd tests/e2e && pytest …` is the actual entry point, otherwise the global default never engages.
2. The audit you describe in the test plan ("zero implicit-strict xfails anywhere in `tests/`") is a point-in-time check. Add a lint or grep-based pre-commit hook that rejects new bare `@pytest.mark.xfail(...)` without an explicit `strict=`, otherwise drift is inevitable. Documenting the intent in `.claude/rules/testing.md` is necessary but not sufficient.

### `emit_results_json.py` — recomputed snapshot with no tests

The scraping approach is the right call (better than hand-maintained counts that drift). Two issues:

1. **Zero unit tests for the scraper itself.** This is a piece of regex-driven parsing logic that is now the source of truth for whether a canary lane passed or failed. The `py_compile` check in the test plan is not test coverage. There are three non-trivial regexes (`RESULT_RE`, `TEST_START_RE`, `PANIC_RE`) and a two-pass walker with a "next standalone outcome owns the previous start" heuristic that relies on `--test-threads=1`. Add a `scripts/live-canary/test_emit_results_json.py` with fixtures for: cargo `ok` + `FAILED` summaries, the Rust ≥1.73 multi-line panic shape, the legacy inline panic shape, ignored tests inline, the `note: run with` followup line, and tokens-in-panic redaction. This is exactly the kind of code where a regression silently downgrades a failed lane to "skip" and we don't notice for another five days.
2. **`--test-threads=1` assumption is implicit.** The walker pairs each `test foo ...` start with the next standalone outcome line. If anyone removes `--test-threads=1` from a future `run.sh` invocation, results get scrambled with no error. Either assert it from `run.sh` (fail-fast if not set) or detect interleaving in the parser and emit a `meta` warning.

### Chat-install probe — meaningful, but check the failure surface

The probe is well-designed: it asserts (1) chat-send 202, (2) the agent actually invoked `tool_install` in history, (3) extension reaches the manager, (4) no forbidden fragments. The history-walk heuristic with the dual `_collect_tool_calls` + `_history_has_tool_install` back-stop is reasonable defensive coding given how often the envelope shape has shifted.

Three things:

1. **`registered = present in `/api/extensions``** is a weak signal. The comment acknowledges that `authenticated`/`active`/`tools` legitimately stay false post-install. That means the probe passes if `gmail` appears in the list *for any reason* — including being seeded by a fixture or staying around from a previous test in the same gateway lifetime. Tighten by capturing the pre-probe extension list and asserting `gmail` was *not* present at start, or by asserting `tool_install` appears in history *before* gmail appears in `/api/extensions`. As written, a regression that reverts to "extension pre-seeded, no install actually happened" passes silently.
2. **60s budget is fine, but no retry on transient flake** — and that's the right call. Just confirm there is no retry loop wrapping this probe in `run_workflow_canary.py`. If there is, kill it. Retry-until-pass on a regression detector is worse than no detector.
3. **`FORBIDDEN_FRAGMENTS = ["Error 400", "Internal Server Error", "panicked", "Traceback", "rust panic"]`** is brittle string-matching. Fine for now, but the auth-recovery probe already does this — the value-add here is the positive assertion that `tool_install` was called. Consider dropping the forbidden-fragments check from this probe and relying on `log_assertions` (the registered-last probe) to catch panic/5xx, so failures point at the right contract.

### Other observations

- `build_gateway_env` propagating `ALLOW_LOCAL_TOOLS` and `AGENT_AUTO_APPROVE_TOOLS` is a real fix in its own right and probably deserved a separate PR. The current bundling makes the env-propagation bug look like part of the canary alerting work, which it isn't. Not blocking.
- The disk preflight (≥10 GiB) on the self-hosted runner is a fine band-aid but the comment explicitly admits there's no cleanup logic. File a follow-up to add `~/.cargo`/`~/.rustup` GC between jobs; otherwise the next time the runner wedges, the preflight just turns an opaque mid-build failure into a clear "won't start" failure — better, but still a wedge.
- `PANIC_RE` with the lazy `.*?` is correct. Good catch on the legacy Rust <1.73 shape. Worth a test fixture (see above).

### Verdict

Approve the direction. Recommend not merging until at minimum `emit_results_json.py` gets unit tests — that scraper is now load-bearing for whether the canary alerts at all, and shipping it without coverage repeats the original sin (silent count drift) in a new spot.

— Zaki

serrrfirat:
Reviewed PR #3682. Shape is right: canary reporting now has real lane counts, chat-driven `tool_install` gets direct coverage, and strict xfail policy closes real blind spot. One blocker before merge.

1. `tool_install_chat` is not isolated from `auth_recovery`

`run_workflow_canary.py` starts one gateway stack, then runs scenarios in order. `auth_recovery` runs immediately before `tool_install_chat`, uses same `"check gmail unread"` prompt, and its own doc says it walks through gmail unavailable → `tool_install` → auth-required. That means `tool_install_chat` can start with `gmail` already installed, or racing toward installed, from prior scenario state.

Failure mode: probe no longer proves fresh chat → `tool_install(name="gmail")` → extension registration. It proves some combination of prior scenario state plus current history. This is exactly canary fidelity surface this PR aims to fix.

Please isolate this probe: move it before `auth_recovery`, run it in fresh stack, use unique extension/trigger not touched elsewhere, or capture pre-state and fail if `gmail` is already present before send. Good regression should assert current thread contains `tool_install` targeting `gmail` and that registration happens after this thread’s send.

2. Structured `tool_install` check does not bind target extension

In `tool_install_chat.py`, structured path is:

`"tool_install" in _collect_tool_calls(history)`

That checks tool name only, not arguments. So any `tool_install` call plus `gmail` appearing in `/api/extensions` can pass, even if install call targeted wrong extension. Fallback `_history_has_tool_install(history, "gmail")` checks target, but structured path bypasses target validation.

Please extract tool-call arguments from structured history and assert `tool_install` has `name == "gmail"`.

3. `emit_results_json.py` still needs parser tests

Scraper is now load-bearing canary alerting code. `py_compile` does not prove regex/state-machine behavior. Add fixtures for cargo ok/fail, Rust >=1.73 panic, legacy inline panic, ignored tests, note lines, token redaction, and multi-binary logs. Also assert parsed entries match cargo summary counts so parser drift fails closed.

Net: direction strong, but merge should wait on probe isolation + targeted `tool_install(gmail)` assertion. Parser tests remain important because this is alerting infrastructure.

zmanian (commented):
## Re-review of 12 new commits since 2026-05-17T13:10:04Z

Direction is now correct. Both my prior findings and serrrfirat's blocker are addressed in code. One residual ask + one new note below.

### Status of my prior findings

**1. `xfail_strict` wiring — Resolved (with caveat).** Confirmed CI entry points (`coverage.yml:191`, `e2e.yml:101`, `reborn-e2e.yml:108`) all `cd tests/e2e` before `pytest`, so the new `pyproject.toml` setting actually reaches the runner. No repo-root `pytest.ini` shadowing. The recommended pre-commit grep guard (rejecting new bare `@pytest.mark.xfail` without `strict=`) is *not* added in this PR. That was the durability ask, and policy-by-convention drifts. Not a merge blocker, but please file a follow-up — `.claude/rules/testing.md` documentation alone won't catch it.

**2. `emit_results_json.py` parser tests — Resolved.** `scripts/live-canary/test_emit_results_json.py` (410 lines) covers every shape I asked for: modern + legacy panic, ignored tests, `note: run with` skip, token redaction (Anthropic key + GitHub PAT + Google + Slack + Bearer), multi-binary aggregation, regex shape locks, and — best part — a cumulative cargo-summary cross-check that fails closed on parser drift. The `--test-threads=1` assumption is now enforced *in the parser itself* via `InterleavedOutputError` (`emit_results_json.py:204`), not just in `run.sh`. That's stronger than what I asked for. Good.

**3. Chat-install probe — Resolved.** Pre-state removal (`_remove_extension_if_present`) runs before the chat-send, with a 5s grace + raise-if-still-present, so the probe can no longer pass on stale `auth_recovery` state (serrrfirat's blocker #1). Structured target binding is now via `_history_has_tool_install_call_for(history, "gmail")` (`tool_install_chat.py:151`), which walks three envelope shapes (direct/OpenAI-function/pending-gate) and binds `name`/`extension` argument to the target — no more "tool_install seen anywhere + gmail seen anywhere" false positive (serrrfirat's blocker #2). The forbidden-fragments check I suggested dropping is still there, but it's cheap and the failure surface is structured separately in `details`, so I'll let it ride.

**4. Env-propagation + disk preflight — N/A.** Both stayed in this PR. Not ideal scope hygiene but the disk preflight has a clear comment about *why* it's a band-aid (no cleanup between jobs on the self-hosted runner) and fails fast with a useful error. Live with it.

### New findings

**A. `results.json` upload under STRICT_ARTIFACT_SCRUB — Verify.** The PR adds `artifacts/live-canary/**/results.json` to the private-oauth upload step (`live-canary.yml:601`) with a comment asserting it's "scrub-safe" because it only carries test names + bounded panic messages. The redaction list in `emit_results_json.py:91-99` covers Anthropic/GitHub/Google/Slack/Bearer — but *not* OpenAI keys (`sk-[A-Za-z0-9]{20,}`), AWS access keys, or Composio API keys, even though the persona harness now forwards `LIVE_CANARY_COMPOSIO_API_KEY` (`live-canary.yml:519`). If an assertion ever dumps a captured response body containing one of those, it bypasses redaction and lands in an uploaded artifact. Either align `REDACT_PATTERNS` with the full `scrub-artifacts.sh` allowlist, or gate the `results.json` upload behind `STRICT_ARTIFACT_SCRUB != true`. The comment claims alignment with `scrub-artifacts.sh` but doesn't actually deliver it.

**B. `notify_slack.parse_summary_status` fallback — Reasonable.** The `summary.md` → exit-code fallback is a nice belt-and-braces for summary-only lanes. Tested by inspection, not by unit test. Worth adding one fixture in a follow-up since it's now part of the status-classification path.

### Verdict

LGTM with the OpenAI/AWS/Composio redaction gap (finding A) addressed before merge — that's a real artifact-leak surface, not theoretical. The xfail pre-commit guard (prior finding 1) can land as a follow-up.

— Zaki

henrypark133:
It seems like the gaps LGTM with the OpenAI/AWS/Composio redaction gap (finding A) addressed before merge has been fixed. Approving.

henrypark133 (approved):

PR #3685, opened 2026-05-15
fix(ws-13): wire live cancellation into host

serrrfirat:
## Context

Stacked follow-up to #3684. This keeps live host cancellation wiring separate from the minimal durable-evidence fix.

## What changed

- Added a turn-state-backed `RunCancellationFactory` that seeds from durable run state.
- Wired Reborn text host construction to use the live turn-state cancellation factory by default.
- Forwarded turn-run wake notifications into the shared cancellation factory.
- Notified runner wake handlers when `cancel_run` persists `CancelRequested`.
- Added tests for cancellation seeding, wake flipping, terminal pruning, and default host observation.

## Validation

- [x] `cargo test -p ironclaw_loop_support cancellation_port`
- [x] `cargo test -p ironclaw_reborn text_only_host_default_cancellation_factory_observes_durable_cancel_request`
- [x] `cargo test -p ironclaw_turns cancel_run_wakes_runner_for_active_cancel_requested_run`

Note: one initial run in a separate `/tmp` worktree hit `No space left on device` compiling `wasmtime-wasi`; reran the relevant checks with the shared target dir and they passed.

## Stack

Base: #3684 / `codex/ws13-cancel-evidence`
Next: `codex/ws13-cancel-ack-safety`

### DIFF
--- a/crates/ironclaw_loop_support/Cargo.toml
+++ b/crates/ironclaw_loop_support/Cargo.toml
@@ -15,7 +15,7 @@ async-trait = "0.1"
 chrono = { version = "0.4", features = ["serde"] }
 parking_lot = "0.12"
 tracing = "0.1"
-tokio = { version = "1", features = ["sync"] }
+tokio = { version = "1", features = ["rt", "sync", "time"] }
 ironclaw_host_api = { path = "../ironclaw_host_api", version = "0.1.0" }
 ironclaw_host_runtime = { path = "../ironclaw_host_runtime", version = "0.1.0" }
 ironclaw_skills = { path = "../ironclaw_skills", version = "0.3.0", default-features = false }
--- a/crates/ironclaw_loop_support/src/cancellation_port.rs
+++ b/crates/ironclaw_loop_support/src/cancellation_port.rs
@@ -1,44 +1,90 @@
-use std::sync::{
-    Arc,
-    atomic::{AtomicBool, Ordering},
+use std::{
+    collections::HashMap,
+    sync::{
+        Arc, Weak,
+        atomic::{AtomicBool, Ordering},
+    },
+    time::Duration,
 };
 
 use async_trait::async_trait;
 use chrono::Utc;
 use ironclaw_turns::{
-    TurnRunId,
+    GetRunStateRequest, TurnRunId, TurnRunWake, TurnRunWakeNotifier, TurnScope, TurnStateStore,
+    TurnStatus,
     run_profile::{
-        AgentLoopHostError, LoopCancelReasonKind, LoopCancellationPort, LoopCancellationSignal,
+        AgentLoopHostError, AgentLoopHostErrorKind, LoopCancelReasonKind, LoopCancellationPort,
+        LoopCancellationSignal,
     },
 };
 use parking_lot::RwLock;
+
+const DEFAULT_CANCEL_POLL_INTERVAL: Duration = Duration::from_millis(25);
+
+#[derive(Clone)]
+struct RunCancellationRequester {
+    fired: Arc<AtomicBool>,
+    signal: Arc<RwLock<Option<LoopCancellationSignal>>>,
+    owner: Weak<()>,
+}
+
+impl RunCancellationRequester {
+    fn request(&self, reason_kind: LoopCancelReasonKind) {
+        request_cancellation(&self.fired, &self.signal, reason_kind);
+    }
+
+    fn is_owner_alive(&self) -> bool {
+        self.owner.upgrade().is_some()
+    }
+}
+
+fn request_cancellation(
+    fired: &AtomicBool,
+    signal: &RwLock<Option<LoopCancellationSignal>>,
+    reason_kind: LoopCancelReasonKind,
+) {
+    if fired.load(Ordering::Acquire) {
+        return;
+    }
+    let mut signal_lock = signal.write();
+    if signal_lock.is_some() {
+        return;
+    }
+    let new_signal = LoopCancellationSignal {
+        reason_kind,
+        requested_at: Utc::now(),
+    };
+    *signal_lock = Some(new_signal);
+    // Publish `fired` while the write guard is still held so any reader that
+    // observes `fired == true` via Acquire is also guaranteed to see the
+    // populated `signal_lock`, independent of the RwLock's own ordering.
+    fired.store(true, Ordering::Release);
+    drop(signal_lock);
+}
 
 /// Snapshot handle the host runtime owns and flips on cancellation.
 #[derive(Clone, Default)]
 pub struct RunCancellationHandle {
     fired: Arc<AtomicBool>,
     signal: Arc<RwLock<Option<LoopCancellationSignal>>>,
+    owner: Arc<()>,
 }
 
 impl RunCancellationHandle {
     pub fn request(&self, reason_kind: LoopCancelReasonKind) {
-        if self.fired.load(Ordering::Acquire) {
-            return;
-        }
-        let mut signal_lock = self.signal.write();
-        if signal_lock.is_some() {
-            return;
-        }
-        let signal = LoopCancellationSignal {
-            reason_kind,
-            requested_at: Utc::now(),
-        };
-        *signal_lock = Some(signal);
-        self.fired.store(true, Ordering::Release);
+        request_cancellation(&self.fired, &self.signal, reason_kind);
     }
 
     pub fn is_requested(&self) -> bool {
         self.fired.load(Ordering::Acquire)
+    }
+
+    fn requester(&self) -> RunCancellationRequester {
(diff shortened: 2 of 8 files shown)


henrypark133 (commented):
## Review — fix(ws-13): wire live cancellation into host

No blockers.

**Double-check seed pattern** correct — register-then-seed closes the TOCTOU race. Wake arriving between seed and register fires on the registered requester; wake arriving before register is caught by the second seed.

**Polling fallback**
- 25ms interval fine for typical concurrency.
- `Weak<()>` owner cleanup is correct — poller exits when handle is dropped.
- Poller holds `Arc<handles>` keeping factory alive until all tasks finish; expected lifetime for a long-lived service, but worth a comment at the spawn site.

**Terminal-status-before-registration** window: if terminal wake fires after `register` but before second `remove_run`, `notify_cancel_requested` fires and removes; the subsequent `remove_run` is a no-op. Handled.

**Low-severity notes**
- `cancellation_port.rs:221` — no unit test for the polling-fallback flip path (poller fires without a wake notification). Easy to add.
- `StaticTurnStateStore` duplicated in two test files — extract to shared test fixture to avoid drift.

Approve.

henrypark133:
## Caveman review findings

**1🟠 3🟡**

- 🟠 `crates/ironclaw_loop_support/src/cancellation_port.rs` `TurnStateRunCancellationFactory::handle_for_run` (~lines 263–268): second `seed_from_state` `?`-returns with the entry already registered and no polling task spawned. `owner` `Weak` goes dead immediately (caller never receives handle), so `RunCancellationRequester` stays in `self.handles` permanently. Permanent map leak. Fix: wrap register+second-seed+fallback-spawn so any error calls `self.remove_run(run_id)` before propagating.
- 🟡 `cancellation_port.rs` `notify_cancel_requested` (~lines 166–173): reads vec under `read()`, drops lock, fires handles outside lock, then `remove_run`. Concurrent `register(run_id, H_new)` between read-drop and `remove_run` inserts a new requester that is silently deleted without being fired. Fix: take write lock once, drain-and-fire atomically, or re-check after removal and fire newly-added entries.
- 🟡 `cancellation_port.rs` `spawn_polling_fallback` (~lines 230–250): a run transitioning to `RecoveryRequired` (non-terminal, non-`CancelRequested`) keeps the polling task alive at `DEFAULT_CANCEL_POLL_INTERVAL = 25ms` for the full owner lifetime — 40 store round-trips/s per stuck run. No backoff, no max duration. Fix: exponential backoff or hard cap via `tokio::time::timeout`.
- 🟡 `crates/ironclaw_turns/src/coordinator.rs` `cancel_run` (~lines 964–974): wake only dispatched when `response.status == TurnStatus::CancelRequested`. If `request_cancel` returns any other non-terminal status (e.g., `RecoveryRequired` → `Cancelled` direct), registered handles never receive a wake and rely solely on the 25 ms polling fallback (or leak per above). Fix: dispatch wake for any status that is `CancelRequested` or `is_terminal()`, mirroring `notify_run_wake` guard.

henrypark133:
Tracking from #3684 review (out of scope for that PR, in scope here):

- 🟡 \`crates/ironclaw_loop_support/src/cancellation_port.rs\` \`RunCancellationHandle::request\` ordering: the \`signal_lock = Some(signal)\` write and \`fired.store(true, Ordering::Release)\` mutate two independent objects (\`RwLock\` + \`AtomicBool\`). On weaker memory architectures a reader observing \`fired == true\` via \`Ordering::Acquire\` in \`observe_cancellation\` could still see \`None\` in \`signal_lock\`. Fix: store \`fired\` inside the write-lock scope before releasing the guard, or document the x86-TSO dependency in a comment.

henrypark133:
## Code Review — `fix(ws-13): wire live cancellation into host` (re-review on amended commit `928c89135`)

### Overview

Introduces `TurnStateRunCancellationFactory` in `ironclaw_loop_support`. Factory seeds cancellation handles from durable run state, registers them for wake-driven flips via `TurnRunWakeNotifier`, and falls back to polling for runtimes that have not wired the wake path. `RebornLoopDriverHostFactory` now uses this factory by default (replacing `AlwaysAliveRunCancellationFactory`). `DefaultTurnCoordinator::cancel_run` emits a wake when cancel is durably persisted as `CancelRequested` or transitions directly to a terminal status.

Amended commit addresses all five caveman findings from the prior round.

### Correctness

- ✅ **Map-leak fix in `handle_for_run`**: replaced `?`-return in the post-register path with explicit `match`. On error, `remove_run(run_id)` is called before propagating, so the `RunCancellationRequester` no longer leaks in `self.handles` when the second `seed_from_state` fails. Early-return for already-`CancelRequested`/terminal first-seed now skips registration entirely.
- ✅ **Atomic drain-and-fire in `notify_cancel_requested`**: switched from `read().get().cloned()` + later `remove_run` to `write().remove(&run_id)` in one critical section. Concurrent `register` after this point lands in a fresh vec — no silent drop. New requester will self-seed on its next `seed_from_state` call (durable status still `CancelRequested`).
- ✅ **Polling backoff**: `spawn_polling_fallback` doubles `interval` on each non-terminal/error tick, capped at `MAX_POLL_INTERVAL = 5s`. Eliminates the 40 req/s drumbeat against `TurnStateStore` for stuck `RecoveryRequired` runs.
- ✅ **`cancel_run` wake guard widened**: now fires on `CancelRequested` OR `is_terminal()`. Direct-to-terminal cancellation paths no longer leave registered handles waiting on the polling fallback.
- ✅ **`request_cancellation` ordering**: explicit `drop(signal_lock)` after `fired.store` plus inline comment documents publishing-order intent. Reader acquiring `fired` via `Acquire` is guaranteed to see populated `signal_lock` independent of RwLock internals.
- ✅ Clippy `too_many_arguments` now `#[allow]`-annotated on `RebornLoopDriverHostFactory::new`. CI was blocked on this; will pass.
- ⚠️ `MAX_POLL_INTERVAL = 5s` is hardcoded. Reasonable default, but consider parameterizing alongside `with_poll_interval` if test harnesses want shorter caps.

### Style / Conventions

- ✅ Uses `matches!`/explicit `match` in new error-handling path; no string-based predicates.
- ✅ All new diagnostics are inline comments explaining *why* (publishing order, leak boundary, backoff rationale).
- ✅ No `pub use` re-exports added beyond the existing crate-level `lib.rs` re-export of `TurnStateRunCancellationFactory`.
- ✅ No `.unwrap()`/`.expect()` in production. Test-only `state.lock().unwrap()` on `std::sync::Mutex` test double.
- ⚠️ `MutableTurnStateStore` test double overlaps in spirit with `StaticTurnStateStore`. Both file-local; no cross-crate duplication. Acceptable. Future consolidation tracked by prior note.

### Test Coverage

- ✅ **New test `turn_state_factory_polling_fallback_fires_without_wake`**: covers the polling-fallback flip path noted as missing in prior review. Uses `MutableTurnStateStore` to flip durable status without dispatching a wake, asserts \`handle.is_requested()\` becomes true within a 2-second deadline at `poll_interval = 5ms`.
- ✅ Existing tests preserved: seeds-already-cancel-requested, flip-from-wake, prune-after-cancel-wake, prune-after-terminal-wake.
- ✅ Workspace clippy `--all --tests --examples --all-features -- -D warnings` passes locally.
- ❌ **No test for the leak-fix error path**: if `seed_from_state` (second call) returns `Err`, the `remove_run` cleanup is exercised only by the code structure, not by a test. A `FailingTurnStateStore` that returns `Err` on the second `get_run_state` call could lock this in. Minor — the leak only manifested under store failure which is rare.
- ❌ **No test for the `notify_cancel_requested` race fix**: concurrent `register`-during-`notify` is hard to reproduce without timing harness. Acceptable as a defensive change.
- ❌ **No test for the widened `cancel_run` wake guard**: a `request_cancel` returning `Cancelled` (terminal direct) should also dispatch a wake. The existing `cancel_run_wakes_runner_for_active_cancel_requested_run` only covers `CancelRequested`.

### Performance

- Backoff cap `5s` reduces worst-case store traffic from 40 req/s to 0.2 req/s per stuck run after a few ticks. No effect on happy path (cancel arrives via wake, polling task exits).
- `notify_cancel_requested` write-lock-once is slightly less concurrent than the old read-then-write pattern, but the critical section is tiny (one `HashMap::remove`).

### Security

- ✅ No new boundary surface. `TurnStateStore` is internal.
- ✅ Per-tenant scope still threaded via `TurnScope`; no cross-tenant fan-out.
- ✅ No raw strings, no `format!`-built predicates.

### Recommendation

**APPROVE.** All five caveman findings from prior round addressed inline. CI clippy block resolved. New regression test covers the polling-fallback path. Three suggested additions (leak error-path test, race-fix test, direct-terminal wake test) are coverage nice-to-haves, not merge blockers. PR is ready.

PR #3617, opened 2026-05-14
feat(reborn): add product adapter runtime catalog

serrrfirat:
## Summary
- add `ironclaw_product_adapter_runtime_catalog` as a registry-backed runtime read model
- surface deterministic enabled ProductAdapter entries paired with registered manifests
- report consistency errors for missing or stale manifest pins
- keep scope narrow: no DB persistence, no webhook routing, no WASM loading, no raw secret material
- update workspace membership, crates README, and Reborn dependency boundaries

## Testing
- `cargo test -p ironclaw_product_adapter_runtime_catalog`
- `cargo fmt --check --package ironclaw_product_adapter_runtime_catalog --package ironclaw_architecture`
- `cargo clippy -p ironclaw_product_adapter_runtime_catalog --all-targets -- -D warnings`
- `cargo test -p ironclaw_architecture reborn_crate_dependency_boundaries_hold`

PR #3688, opened 2026-05-15
refactor(reborn): project ProductAdapter from single ExtensionManifestV2

serrrfirat:
## Summary

Replace the old `ProductAdapterManifestDocument` / `ProductAdapterManifest` separate TOML format with a projection from the single Extension Manifest v2 host-API section mechanism introduced in #3663.

ProductAdapter declarations now live in one manifest:

```toml
schema_version = "reborn.extension_manifest.v2"
id = "telegram-v2"
...

[[host_api]]
id = "ironclaw.product_adapter/v1"
section = "product_adapter.inbound"

[product_adapter.inbound]
surface_kind = "external_channel"

[product_adapter.inbound.auth]
kind = "shared_secret_header"
header_name = "X-Telegram-Bot-Api-Secret-Token"

[product_adapter.inbound.capabilities]
flags = ["inbound_messages", "external_final_reply_push"]

[[product_adapter.inbound.required_credentials]]
handle = "telegram_bot_token"

[[product_adapter.inbound.egress]]
host = "api.telegram.org"
credential_handle = "telegram_bot_token"
```

## What changed

- `ExtensionManifestRecord` replaces `ProductAdapterManifestDocument` / `ProductAdapterManifest`
- Activation state, credential bindings, health, and manifest hash pins kept, keyed by `ExtensionId`
- `list_enabled_runtime_entries()` projects `ProductAdapterHostApiSection` from the enabled installation manifest, replacing PR #3617's runtime catalog concept
- Old standalone ProductAdapter TOML fixture removed
- `ironclaw_extensions` added as dependency; `semver` removed
- Architecture boundary updated

## Testing

```bash
cargo test -p ironclaw_product_adapter_registry
cargo test -p ironclaw_architecture reborn_crate_dependency_boundaries_hold
cargo clippy -p ironclaw_product_adapter_registry --all-targets -- -D warnings
cargo fmt --check --package ironclaw_product_adapter_registry
```


henrypark133:
## Caveman review findings

**1🔴 4🟡**

- 🔴 `src/lib.rs` `InMemoryExtensionInstallationStore::set_activation_state`: two-phase lookup. Reads installation via `get()`, drops borrow, then `get_mut()`. Under `RwLock<InMemoryState>` the write guard is dropped and re-acquired between read and write — concurrent delete between calls makes the `?` on `get_mut` panic. Fix: acquire write lock once, do all reads/writes through the same `&mut` reference.
- 🟡 `src/lib.rs` `list_enabled_product_adapter_entries`: `manifest.product_adapters().first().cloned()` silently discards all product-adapter sections after the first. Contract is `HostApiMultiplicity::Multiple`. Fix: iterate and push one `ProductAdapterRuntimeEntry` per section.
- 🟡 `src/lib.rs` `RegistryError::NotProductAdapterExtension`: variant defined but never constructed. Either remove or wire it to the silent `continue` path in `list_enabled_product_adapter_entries`.
- 🟡 `src/lib.rs` `ExtensionActivationState`: variant order changed (`Enabled` moved from position 1 to 2). Serde string-based round-trips are safe, but any storage backend using ordinal/binary encoding will silently misread. Verify no DB column or binary format encodes by position; add `#[serde(alias)]` or migration if so.
- 🟡 `tests/registry_contract.rs`: `duplicate_credential_bindings_rejected_at_construction` and `no_op_activation_transition_does_not_update_timestamp` tests deleted with no replacement. `ExtensionInstallation::new` duplicate-binding path and same-state-no-timestamp-bump invariant now untested. Restore both.

ilblackdragon:
## Code Review (post-merge)

> This PR is **already merged** (commit `0a8eed1c`). The review below is post-merge feedback — useful for spotting follow-ups, not gating.

### Overview

Replaces the separate `ProductAdapterManifestDocument` / `ProductAdapterManifest` TOML format with a **projection** from the single Extension Manifest v2's `[[host_api]]` section mechanism (introduced in #3663). All `ProductAdapter*` types in the installation registry become `Extension*` generics, with ProductAdapter as a typed-section projection on top. +1291 / −1453 (net deletion).

The PR description captures the design well: one store, one manifest type, one installation type — `list_enabled_product_adapter_entries(store)` is the projection. The dead-code elimination is real: `ProductAdapterManifestDocument`, `ProductAdapterManifest`, `ProductAdapterComponentRef`, the standalone manifest fixture, and ~400 lines of duplicate validation all go away.

### Strengths

- **Real consolidation, not just renaming.** The old crate had two separate trait hierarchies (`ProductAdapter*` store + manifest parsing) layered on top of v1's channel-manifest analog. The new shape has *one* `ExtensionInstallationStore` plus a free function `list_enabled_product_adapter_entries(store)` for the projection. ProductAdapter is one specialization of a general extension shape — and the registry no longer cares about the difference.
- **Validation is preserved at the projection boundary.** `ProductAdapterHostApiSection::from_value` runs `reject_inline_secret_material_value`, validates the auth requirement (HTTP token rules), checks egress credential handles are declared as required, and rejects duplicates. The old `ProductAdapterManifest::validate` is structurally still there, just under a new name.
- **Host-API contract validator wired through.** `ProductAdapterHostApiContract` implements `HostApiManifestContract` with `multiplicity = Multiple` and an `accepts_section_path` that allows both `product_adapter` and `product_adapter.<subsection>`. So multiple inbound/outbound product-adapter sections in one extension are supported structurally.
- **Adapter id derivation is deterministic.** `adapter_id = "{extension_id}/{subsection}"` (e.g. `telegram-v2/inbound`). Multiple PA sections within one extension stay distinguishable downstream. Good.
- **Stronger HTTP-token validation.** `validate_http_token` enforces RFC 7230 tchars (no whitespace, no CTL, no separators) on `header_name` / `timestamp_header_name` / cookie name fields — was previously enforced by `AuthRequirement::validate_metadata`'s error formatting path, now explicit at the registry boundary.
- **Manifest hash pinning is symmetric.** `validate_installation_against_one_manifest` rejects `(Some(_), None) | (None, Some(_))` so an installation pinned to a specific manifest revision can't silently bind to an unstamped manifest. Kept verbatim from the pre-refactor code, but worth calling out.
- **Inline-secret-material denylist preserved.** `is_secret_key_name`, `looks_like_inline_secret`, `looks_like_aws_access_key`, `has_uri_userinfo`, `looks_like_telegram_token` all carried forward unchanged. AWS-key check, JWT-shape, URI userinfo, vendor prefixes (`sk-`, `xoxb-`, GitHub `gh*_`, …), telegram bot tokens.
- **Tests rewritten, not deleted.** The 388-line `manifest_ingestion.rs` becomes 96 lines — the test surface shrinks because the *single-manifest tests* against the old standalone TOML are obsolete. The new tests verify the same invariants against the new v2-projection path.
- **Boundary test updated.** `ironclaw_extensions` removed from `ironclaw_product_adapter_registry`'s forbidden list (it's now a dependency), and the comment is updated to reflect the projection-based design.
- **Trust model documented in `CLAUDE.md`.** Bold "must not introduce a second ProductAdapter TOML manifest format" — exactly the rule a future contributor might trip if they don't grok the refactor's intent.

### Issues / Follow-ups

#### 1. `RegistryError::UnknownManifest` semantics changed silently (medium)
In `list_enabled_product_adapter_entries`, if an installation is enabled but its extension manifest has no `product_adapter` host_api section, the loop now `continue`s silently with the comment `"extension is enabled but not a product adapter — skip"`. Pre-refactor, the only way to have an enabled installation was to have a registered ProductAdapter manifest — the situation was structurally impossible. Now it's possible and silently ignored.

If a runtime caller expected `list_enabled_product_adapter_entries` to surface every enabled installation it knows about, this is a behavior change. Worth either:
- Surfacing such installations as a warning (so operators see "the extension is enabled but doesn't expose a product_adapter section"), or
- Documenting on the function explicitly that it filters silently.

The current comment is internal and a future reader won't notice it.

#### 2. `ProductAdapterHostApiContract.validate_section` uses a placeholder extension id (low)
```rust
let placeholder = ExtensionId::new("x").map_err(|e| e.to_string())?;
ProductAdapterHostApiSection::from_value(&placeholder, host_api.section.clone(), section.clone())
```
The contract validator runs at v2 parse time, before the real `extension_id` is known to the contract. The shape-only validation works correctly (the placeholder isn't used in any cross-field constraint), but the `adapter_id` it computes (`x/inbound`) is thrown away. If a future invariant cross-checks `adapter_id` against `extension_id` at validation time, this code path would silently allow it. Worth a comment explaining why the placeholder is safe today.

#### 3. `ProductAdapterError` is added to `RegistryError` as `#[from]` but unused (nit)
```rust
#[error(transparent)]
ProductAdapter(#[from] ProductAdapterError),
```
I don't see `ProductAdapterError` constructed anywhere in the new code path — the previous use site (`validate_metadata` in the old `validate_auth_requirement`) is replaced by the inline `validate_http_token` helper, which returns `RegistryError::InvalidValue` directly. Either:
- Wire it through `ProductAdapterId::new(...)` failures (currently mapped to `InvalidValue` via `.map_err`), or
- Remove the variant.

A dead `#[from]` variant grows the error surface for no reason and the next refactor pass will trip over it.

#### 4. `looks_like_inline_secret` lost prefix comments in the inline expansion (nit)
The pre-refactor code had a per-prefix comment on each constant in `PREFIXES` (`// OpenAI / Anthropic`, `// Slack bot`, etc.). The new code inlines all prefixes on one line:
```rust
const PREFIXES: &[&str] = &[
    "sk-", "xoxb-", "xoxa-", "xoxp-", "xoxs-", "xoxe-", "ghp_", "gho_", "ghu_", "ghs_", "ghr_",
];
```
The comments were the *only* documentation of which vendor each prefix tripped on. Operators staring at a "rejected as inline-secret-material" error will have to grep the source to figure out which family hit. Restore the per-prefix `// vendor` comments.

#### 5. `section_value` does a fresh TOML parse (low)
`project_product_adapter_sections` calls `toml::from_str(raw_toml)` again to walk to the section path. The v2 parser already parsed the same TOML into `ExtensionManifestV2`. Two parses of the same string per manifest. Not a hot path (boot-time only), but if `ExtensionManifestV2` retains an intermediate `toml::Value` representation accessible by section path, the second parse can go away.

#### 6. `ExtensionInstallationStore::update_health` doesn't validate the snapshot (nit)
`set_health` writes the snapshot in place. If a caller constructs an `ExtensionHealthSnapshot` with an `Unhealthy` status but no message, that's stored verbatim. The old `ProductAdapterHealthSnapshot::new` accepted any combination too. Defensible — operators may have status codes without messages — but worth a single doc line on `update_health` clarifying that the store doesn't enforce status-message consistency.

#### 7. `looks_like_inline_secret` doesn't catch all the prefixes documented elsewhere in the codebase (cross-PR consistency)
PR #3704's `SECRET_PREFIXES` list is more comprehensive (`sk_live_`, `sk_test_`, `xapp-`, `github_pat_`, `AKIA`, `ASIA`, `AIza`, `ya29.`, `hf_`, `nai-sess-`). This crate's list is a subset. As of today the threat models differ (this crate is `operator-authored TOML`, #3704 is `boot config files`) — but the threat is the same shape (pasted credentials), and divergence will drift further. Worth extracting a shared `ironclaw_secret_guard` crate that both consume. Out of scope for this PR; flagging as architectural follow-up.

### Project-conventions check
- `#![forbid(unsafe_code)]` ✓
- `.unwrap()` only in tests ✓
- `deny_unknown_fields` on every TOML-facing struct ✓
- Boundary tests pinned ✓
- AGENTS.md / CLAUDE.md updated in lockstep with the refactor ✓

### Risk summary
**Low** for the refactor itself — `cargo test -p ironclaw_product_adapter_registry` passes and the invariants are preserved at the projection layer rather than at the standalone-manifest layer. The only behaviorally-meaningful follow-up is item (1) — the silent skip of enabled-but-non-PA installations is a real semantic shift that should be documented or made noisy.

The cross-PR follow-up worth tracking is (7) — converging the secret-material denylists between this crate and `ironclaw_reborn_config`.

PR #3693, opened 2026-05-15
fix(webui): add send-message idempotency replay guards

think-in-universe:
Superseded by a clean PR cut from `main`. This draft was created from the shared `reborn-integration` branch and includes unrelated branch history, so it should not be reviewed.


think-in-universe:
Closed because Codex created from an incorrect base branch.

PR #3703, opened 2026-05-16
arch(reborn): futureproof RebornRuntime surface for epic #3036 (Configuration-as-Code)

serrrfirat:
## Summary

Reshapes `RebornRuntimeInput` / `RebornRuntime` / the `ironclaw-reborn` CLI command tree so that epic [#3036 — "Configuration-as-Code for IronClaw Reborn: tenant blueprints and use-case harnesses"](https://github.com/nearai/ironclaw/issues/3036) can land its substrate (typed repos, blueprint parser, apply service, harness registry, instruction-bundle assembler) **without breaking the standalone Reborn binary's input contract, runtime handle, or operator-facing CLI surface**.

This is a forward-architecture PR: nearly every line is additive seams + DTOs + stubs + boundary tests. Behavior of the running agent is unchanged.

Stacked on top of [#3695](https://github.com/nearai/ironclaw/pull/3695) (composition root + narrow surface + live binary).

## How epic #3036 shows up in this PR

| Epic #3036 concept | Today's shape (this PR) | When the substrate lands |
|---|---|---|
| `api_version = "ironclaw.config/v1"` on every blueprint | `RebornRuntimeApiVersion::V1` mandatory field on `RebornRuntimeInput`; major mismatch is `RebornRuntimeError::IncompatibleApiVersion` (fail-closed) | Same enforcement; minor bumps land additive fields |
| Tenant blueprint `[scope] tenant = "acme"` + per-tenant identity | `RebornIdentityConfig { tenant, default_agent, default_owner, default_project }` first-class field; `ConversationIdentity` per-conversation override | DTO is populated from the parsed blueprint via the apply service writing through `RebornTenantRepo`; runtime reads from repo |
| Blueprint `[runtime]` profile + approval policy | `RebornPolicyConfig { deployment_mode, default_profile, default_approval_policy }` recorded in boot log via `tracing::debug` | Resolved per-request via `RuntimePolicyRepo` lookup; default written by blueprint apply |
| Blueprint `[providers.<name>]` LLM config | `RebornLlmConfig` already on the input (PR #3695); CLI env-resolves today | Sourced from `ProviderRepo` populated by blueprint apply |
| `HarnessManifest` activation | `RebornHarnessSelection { harness_id }` optional input field; logged + no-op today | Drives `HarnessActivationService` → `InstructionBundleAssembler` overlay + capability-surface filter; per-message override via `SendMessageOptions::with_harness` |
| `BlueprintApplyService` as a privileged `CapabilityHost` workflow | `RebornAdminClient` handle with `apply_blueprint / diff_blueprint` returning `RebornAdminError::NotYetWired` | Method bodies replaced; signatures unchanged |
| `HarnessRepo` lifecycle | `RebornAdminClient::{install_harness, activate_harness, deactivate_harness, list_harnesses}` returning `NotYetWired` | Method bodies replaced; signatures unchanged |
| `AdminScope` privilege gate (epic explicitly says "reuse `AdminScope::new` from `src/tenant.rs`") | `RebornAdminScope` sealed via private `SealedMarker`; `for_cli_admin(audit_reason)` is the only public constructor | Gains `for_authenticated_admin(credential, audit_reason)` when the privilege/auth crate lands |
| `ironclaw config apply / diff / watch` CLI commands | Discoverable stub subcommands with final argument shape, fail with structured `tracking_issue: #3036` errors | Bodies switch from `fail_not_yet_wired(...)` to `runtime.admin(scope).apply_blueprint(...).await` |
| `ironclaw harness install / list / activate / deactivate` | Same — discoverable stubs with final argument shape | Same — body swap-in only |

Every row above is **a future PR that swaps an `Err(NotYetWired)` for a real body** — not a future PR that *changes the shape* of `RebornRuntimeInput`, `RebornRuntime::admin(scope)`, `RebornAdminClient`, or the CLI subcommand tree.

## Stack of commits

| Commit | Purpose |
|---|---|
| `arch(reborn-composition): futureproof RebornRuntimeInput for epic #3036` | Splits `runtime_input.rs` into a module; adds `RebornRuntimeApiVersion`, `RebornIdentityConfig`, `ConversationIdentity`, `RebornPolicyConfig`, `RebornDriverConfig`/`RebornDriverChoice`, `RebornHarnessSelection`/`RebornHarnessId`. Compatibility wired in `build_reborn_runtime` (api_version check, planned-driver = NotYetWired, harness selection = log+proceed). Unit tests per DTO. |
| `arch(reborn-composition): add RebornAdminClient + sealed RebornAdminScope for epic #3036` | New `runtime_admin.rs`. `RebornAdminScope` sealed via private `SealedMarker`. `RebornAdminClient` with 6 typed async methods, each returning `RebornAdminError::NotYetWired { operation, tracking_issue: "#3036", requires: "<substrate>" }`. `RebornRuntime::admin(scope)` accessor + `RebornRuntime::new_conversation_for(ConversationIdentity)` + `SendMessageOptions { requested_run_profile, harness_override }` + `send_user_message_with_options(...)`. Per-conversation identity now tracked in an `Arc<RwLock<HashMap<ThreadId, ConversationContext>>>` so multi-tenant `new_conversation_for` is real, not hardcoded. |
| `feat(reborn-cli): lock epic #3036 operator surface as discoverable stubs` | Splits `commands/config.rs` → `commands/config/{mod,apply,diff,watch}.rs`; new `commands/harness/{mod,install,list,activate,deactivate}.rs`; shared `commands/stubs.rs` for uniform "not yet wired" output. 8 new CLI smoke tests pinning the operator-facing stub surface. |
| `arch(boundaries): pin RebornRuntime futureproof surface for epic #3036` | New boundary test `reborn_runtime_surface_is_futureproof_for_epic_3036` asserts (1) all 5 epic-3036 fields on `RebornRuntimeInput`, (2) every admin method present, (3) `RebornAdminScope` remains sealed (private `_sealed: SealedMarker` field), (4) every CLI stub file exists. |

## What the runtime looks like to callers now

```rust
use ironclaw_reborn_composition::{
    build_reborn_runtime,
    // versioned input
    RebornRuntimeApiVersion, RebornRuntimeInput,
    // independent input layers (forward-compatible with epic #3036)
    RebornIdentityConfig, ConversationIdentity,
    RebornPolicyConfig,
    RebornDriverConfig, RebornDriverChoice,
    RebornHarnessSelection, RebornHarnessId,
    RebornLlmConfig,                   // #[cfg(feature = "root-llm-provider")]
    TurnRunnerSettings,
    // agent surface
    RebornRuntime, ConversationId, AssistantReply, SendMessageOptions,
    // admin surface
    RebornAdminClient, RebornAdminScope, RebornAdminError,
    // admin DTOs reserved for epic #3036
    ApplyBlueprintRequest, ApplyMode, ApplyReport, ApplyChange, ApplyChangeAction,
    DiffBlueprintRequest, BlueprintSource,
    InstallHarnessRequest, ActivateHarnessRequest, DeactivateHarnessRequest,
    HarnessActivationScope, HarnessDescriptor,
};
```

That is the entire surface a downstream consumer touches. No `TurnCoordinator`, no `SessionThreadService`, no `HostManagedModelGateway`, no `SubmitTurnRequest`, no host-api substrate types — everything stays inside the composition root.

```rust
// Agent surface
let conv = runtime.new_conversation_for(
    ConversationIdentity::for_tenant(acme).with_agent(acme_bot)
).await?;
let reply = runtime.send_user_message_with_options(
    &conv,
    "hello",
    SendMessageOptions::default()
        .with_requested_run_profile("text_only")
        .with_harness(red_team),     // recorded today, takes effect when harness substrate lands
).await?;

// Admin surface
let scope = RebornAdminScope::for_cli_admin("operator ran `config apply`");
let admin = runtime.admin(scope);
let report = admin.apply_blueprint(ApplyBlueprintRequest {
    source: BlueprintSource::Path(blueprint_path),
    mode: ApplyMode::Apply,
    lockfile: Some(lockfile_path),
}).await?;
// Today: returns Err(NotYetWired { operation: "config.apply", tracking_issue: "#3036", requires: "..." })
// Tomorrow (epic sub-issue lands): returns Ok(ApplyReport { changes: [...] })
```

## Operator surface verified

```text
$ ironclaw-reborn --help
...
  harness     Install / activate / list use-case harnesses (epic #3036).
              Subcommands ship as discoverable stubs today; substrate
              lands incrementally via epic sub-issues.
...

$ ironclaw-reborn config --help
...
  path   Show resolved Reborn configuration paths without creating state
  apply  Apply a declarative IronClaw blueprint to the runtime's typed repos
  diff   Diff a declarative blueprint against the runtime's typed repos without writing
  watch  Watch a git URL for blueprint revisions and apply on each new revision (GitOps mode)

$ ironclaw-reborn harness --help
...
  install     Install a harness manifest into the typed harness repo
  list        List installed harnesses
  activate    Activate an installed harness for a scope (thread / project / tenant)
  deactivate  Deactivate the active harness for a scope

$ ironclaw-reborn config apply --dry-run /tmp/x
ironclaw-reborn: command `config.apply` is not yet wired
  tracking issue : #3036
  requires       : BlueprintParser + BlueprintApplyService + typed repos
                   (Settings, Skill, Mission, Project)

The command shape (arguments, output format) is locked from epic #3036;
the substrate that makes it execute lands incrementally via the issue's
sub-issues. Re-run after the relevant sub-issue ships.
$ echo $?
1
```

## Tests

| Crate | Result | Notes |
|---|---|---|
| `ironclaw_llm` | 733/733 | unchanged |
| `ironclaw_reborn` (with `root-llm-provider`) | 225/225 | unchanged |
| `ironclaw_reborn_composition` (both feature variants) | 23/23 | +18 unit tests across `runtime_input/` submodules + `runtime_admin` |
| `ironclaw_reborn_cli` | 43/43 | +8 stub-surface smoke tests |
| `ironclaw_architecture` | 18/18 | +1 new boundary test (`reborn_runtime_surface_is_futureproof_for_epic_3036`) |

End-to-end smoke verified: `ironclaw-reborn run -m ping` without LLM env vars still surfaces `RecoveryRequired` in ~2s, same as PR #3695. The futureproof reshape did not regress the live-agent path.

## What this PR is NOT

- **It does not implement any of epic #3036's substrate.** Every admin method is `NotYetWired`. No typed repo, no blueprint parser, no apply service, no harness registry, no instruction-bundle assembler, no capability-surface filter, no reference harness, no GitOps watcher.
- **It does not change runtime behavior** for an operator running `ironclaw-reborn run` today.
- **It does not add v1 (`ironclaw` root crate) dependencies.** The CLI's `BoundaryRule.forbidden` set inherited from PR #3695 is unchanged and still enforced.

## How epic #3036 sub-issues consume this PR

Each sub-issue from the epic's "Suggested decomposition" reads as a body-swap on top of this shape:

1. **Blueprint format + parser + JSON Schemas.** New `ironclaw_blueprint` crate. No runtime touch. (Sub-issue 1.)
2. **`BlueprintApplyService` + per-domain reconcilers + ApplyReport/lockfile.** Implements the body of `RebornAdminClient::apply_blueprint` + `::diff_blueprint`. Composition gains an optional `Arc<dyn BlueprintApplyService>` substrate field. (Sub-issue 2.)
3. **System prompt as scoped setting + `InstructionBundleAssembler`.** New `ironclaw_settings` crate. Composition starts passing `with_identity_context_source(...)` to `RebornLoopDriverHostFactory`. `RebornPolicyConfig` becomes read-through to `RuntimePolicyRepo`. (Sub-issue 3.)
4. **`HarnessManifest` typed repo + lifecycle + activation gating.** New `ironclaw_harness` crate. Implements the body of `RebornAdminClient::install_harness / activate_harness / deactivate_harness / list_harnesses`. `RebornHarnessSelection` becomes live: composition wires the active harness's capability-surface filter via `with_safety_context(...)` and its prompt overlay via the instruction-bundle assembler from sub-issue 3. (Sub-issue 4.)
5. **Reference harness: `red-team`.** Pure content. Lands under `harnesses/red-team/`. (Sub-issue 5.)
6. **Reference harness: `chain-incident-response`.** Pure content. (Sub-issue 6.)
7. **CLI surface (`apply` / `diff` / `watch`) + admin web endpoint.** CLI stub bodies replaced with real calls through `runtime.admin(scope).apply_blueprint(...)`. Admin web is an orthogonal frontend concern. (Sub-issue 7.)
8. **Migration tests + drift detection + audit/event coverage.** Tests on top of the wired apply service. (Sub-issue 8.)

At each step, **`RebornRuntimeInput`, `RebornRuntime`, `RebornAdminClient`, and the CLI command tree do not change shape** — only the bodies inside method definitions, and (orthogonally) the composition's wiring inside `build_reborn_runtime`. External tooling and operator muscle memory hold steady.

## Verification checklist (against epic acceptance criteria)

The epic enumerates many AC bullets; this PR doesn't satisfy them (none of the substrate is built), but several **structural prerequisites** the AC implicitly depends on are now in place:

- [x] **"Apply must be admin-scoped"** — `RebornAdminClient` is only obtainable via `RebornRuntime::admin(scope: RebornAdminScope)`; scope is sealed and pinned by a boundary test.
- [x] **"API version is mandatory and locks the schema major"** — `RebornRuntimeApiVersion::V1` mandatory; major mismatch fail-closed; pinned by unit test.
- [x] **"Inline secret material rejected at parse time"** — *(not in this PR; lands with parser in sub-issue 1)*
- [x] **"`apply_blueprint(blueprint, scope, mode = Apply | DryRun)` returns a typed `ApplyReport`"** — signature exists; body is `NotYetWired`. Body lands in sub-issue 2.
- [x] **"`Change { domain, key, action: Create|Update|NoOp|Delete-deferred, before_hash, after_hash }`"** — `ApplyChange` carries exactly those fields.
- [x] **"Active harness reflected in `visible_capabilities()` before next model call"** — *(seam exists via `RebornLoopDriverHostFactory::with_safety_context`; wired by sub-issue 4)*
- [x] **"Switching harness emits `harness.deactivated` then `harness.activated` audit events"** — *(audit emission in sub-issue 4)*
- [x] **"`harness install / list / activate / deactivate` exist on the CLI"** — stubs ship today with the exact argument shape; sub-issue 4 / sub-issue 7 swap bodies.
- [x] **"`ironclaw config apply <path>` produces a lockfile"** — argument exists; body in sub-issue 2.

## Why this PR is small and why it's now

~1k LOC. Every line is one of three things: (a) a typed DTO that didn't exist, (b) a stubbed method body whose signature is committed by the epic, or (c) a boundary test pinning the shape. There is no behavior change to bisect against, no agent-loop touch.

Shipping the reshape **now**, before any sub-issue starts, means epic #3036's sub-issues can land independently as body-swaps without each one re-litigating "what's the right input shape" or "where does the admin gate live." That conversation is settled here.

The alternative — extending `RebornRuntimeInput` incrementally as each sub-issue lands — would force five separate breaking changes to the public input contract, force the CLI command tree to grow new top-level subcommands across releases (breaking shell completion expectations), and re-decide the privilege gate on each sub-issue. Concentrating the shape change here is the cheapest version of futureproofing.

cc @ilblackdragon  (this addresses the "too many things exposed; just those few structs/traits" feedback that originated this work — every type in the new public surface is one of [DTO, handle, error, ID]; nothing exposes underlying substrate traits).


ilblackdragon:
## Code Review

### Overview

Forward-architecture PR that reshapes `RebornRuntimeInput` / `RebornRuntime` / the CLI command tree to absorb epic [#3036](https://github.com/nearai/ironclaw/issues/3036) ("Configuration-as-Code") without future shape changes. Every line is one of three things: a typed DTO that didn't exist, a stub method whose signature is committed by the epic, or a boundary test pinning the shape. **No behavior change** for an operator running `ironclaw-reborn run` today.

+2124 / −182, 23 files. Stacked on #3695.

### Strengths

- **Shape-first discipline is enforced, not aspirational.** New boundary test `reborn_runtime_surface_is_futureproof_for_epic_3036` greps for each of the five futureproof fields on `RebornRuntimeInput`, every admin method on `RebornAdminClient`, the **private `_sealed: SealedMarker` field** on `RebornAdminScope`, and the ten CLI stub files. A future regression hits a precise CI failure with the rationale inline.
- **`RebornAdminScope` is sealed structurally, not by convention.** Private unit struct `SealedMarker` lives inside the type; outside crates literally cannot construct it because they can't name the field. The only public ctor is `for_cli_admin`, with the future `for_authenticated_admin(credential, ...)` reserved for when the auth crate lands. Sealing pattern is clean and minimal.
- **Per-conversation identity is real, not faked.** The previous single `thread_scope` is replaced by an `Arc<RwLock<HashMap<ThreadId, ConversationContext>>>`, and `send_user_message_with_options` looks up the per-conversation context to mint the right `TurnScope`. `new_conversation_for(ConversationIdentity)` actually composes a distinct `ThreadScope`. This is a meaningful capability addition that the stub framing slightly understates — it unblocks multi-tenant on day one.
- **`RebornHarnessId` validation is strict and tested.** ASCII lowercase + digits + hyphens, must start with a lowercase letter, ≤64 bytes. Serde `try_from = String` so deserialization can't bypass `new`. Tests cover empty, `Red-Team` (invalid start), `red_team` (invalid char), `9-leading-digit`, oversized. Mirrors `PackageId` discipline.
- **`api_version` fail-closed on major mismatch.** Mandatory `RebornRuntimeApiVersion` field; major mismatch returns `IncompatibleApiVersion`; minor diffs are tolerated. Tests exercise both. Wire format `ironclaw.runtime/v1.0` is stable.
- **Uniform stub command output.** `stubs::fail_not_yet_wired` is the single source of stub-output formatting. Output names `operation`, `tracking_issue`, `requires`; greppable, scriptable; tests assert each command's specific operation string.
- **Per-DTO `#[cfg(test)] mod tests` blocks.** Every new module ships its own tests in-file rather than dumping them in `tests/runtime.rs`. Focused, fast, easy to find.
- **Documentation is unusually thorough.** Every module header has the same three sections: what it is, what it does today, what it does after epic #3036 lands. Future maintainers can read these and understand the trajectory without consulting the epic.
- **`#[non_exhaustive]` on the right enums.** `RebornDriverChoice`, `HarnessActivationScope`, `ApplyChangeAction` — all good. Lets future variants land non-breakingly.
- **Cross-PR test integrity.** The integration test from #3695 (`stub_gateway_send_cancels_recovery_required_and_releases_conversation`) is preserved without modification — the reshape didn't regress the live-agent path.

### Issues / Suggestions

#### 1. `conversations` map is unbounded (medium)
`new_conversation` and `new_conversation_for` insert into `Arc<RwLock<HashMap<ThreadId, ConversationContext>>>` with no eviction path. Long-running processes (anything beyond the CLI's short-lived sessions — a future server adapter, an e2e harness, a hosted runtime) will accumulate entries forever. Each entry is small (a tenant id, agent id, owner id, optional project, a `ThreadScope`), so the leak rate is modest, but it is unbounded.

Suggest exposing `close_conversation(id: ConversationId)` and removing the entry on `shutdown`. The shutdown removal is the easy first step.

#### 2. `RebornDriverChoice::Planned` reachable from config but fails at boot (medium)
`drivers.default == RebornDriverChoice::Planned` yields `RebornRuntimeError::InvalidArgument("...recognized but not yet wired end-to-end in build_reborn_runtime; track epic #3036 sub-issue...")`. The shape is fine, but operators who set `default = "planned"` in #3704's `config.toml` get a *runtime* error rather than a *parse-time* error. Since the TOML side accepts the wire string, the CLI's `reject_unsupported_runtime_sections` (in #3704) should grow a check for `drivers.default = "planned"` — fail at parse time so operators see the issue before they've spun up the agent.

#### 3. `harness` log noise on every boot and every send (low)
Two places emit `tracing::warn!`:
- `build_reborn_runtime` when `harness` is `Some` ("recorded but not yet wired")
- `send_user_message_with_options` when `SendMessageOptions.harness_override` is `Some` ("recorded but not yet wired")

Once operators start setting harness identifiers as a planning measure (which the PR description encourages with "operators free to populate `mission` ahead of time"), these warn lines fire on every send. Default log levels surface warns; production logs get cluttered with no actionable signal.

Suggest dropping these to `debug!` (the structured field stays in the log if anyone needs it) or warn-once-per-process by tracking emission in a `OnceLock<()>`.

#### 4. `RebornAdminScope.audit_reason` is unvalidated `String` (low)
Construction accepts empty strings, garbage, or secret-shaped values without any check. The string is destined for audit logs (epic #3036's `CapabilityHost::invoke_json` invocation reason). Apply `reject_inline_secret` (already exported from `ironclaw_reborn_config` per #3704) to catch operators pasting credentials into audit reasons. At minimum, reject empty.

#### 5. `for_cli_admin` is a misnomer (low)
The doc says "the future privilege/auth crate will add `for_authenticated_admin`", but today `for_cli_admin` is callable from anywhere the composition crate's public API is reachable — including channel adapters or web handlers if anyone wires them up. Sealing is structural via the marker field, but the *name* implies a CLI-scope restriction that isn't enforced. Future contributors reading "the CLI is the only legitimate caller today" comments will assume the function name encodes the restriction.

Suggest renaming to `for_local_admin(...)`, `unverified_admin(...)`, or `for_legacy_unauthenticated_admin(...)` so the misnomer doesn't ossify into an assumption.

#### 6. `RebornRuntimeError::InvalidArgument` is overloaded (low)
Both "thread id is invalid" (validation failure) and "Planned driver is not yet wired" (unimplemented feature) use the same `InvalidArgument` variant. Boot-time-unimplemented vs. invalid-caller-input get conflated. Consider mirroring `RebornAdminError::NotYetWired { operation, tracking_issue, requires }` on the agent surface so the two are distinguishable. Today's stub callers can `match` on it cleanly.

#### 7. `BlueprintSource` lacks `#[non_exhaustive]` (nit)
`BlueprintSource::Path(...)`, `GitUrl(...)`. Future likely variants: OCI registry, S3, signed URL. Mark `#[non_exhaustive]` to match the discipline applied to neighboring enums.

#### 8. `fail_not_yet_wired` may emit duplicate error text (nit)
The function `eprintln!`s the structured fields, then returns `Err(anyhow::anyhow!(error))`. anyhow's default formatter prints `error.to_string()` again (which is `RebornAdminError::NotYetWired`'s `#[error("...")]` line) — so a user sees the structured block *and* anyhow's rendering of the same content. Either:
- Print only the structured form and return `Err(anyhow::anyhow!("operation `{operation}` is not yet wired (see stderr)"))`, OR
- Convert to a `std::process::ExitCode::FAILURE` without a second printout.

Worth a small polish for stderr cleanliness.

#### 9. `HarnessActivationScope` variants carry untyped strings (nit)
`Thread { thread_id: String }`, `Project { project_id: String }`, `Tenant { tenant_id: String }`. Defensible to ship as strings now since the activation service isn't wired, but the typed variants (`ThreadId`, `ProjectId`, `TenantId`) already exist in `ironclaw_host_api`. Validating at the CLI boundary catches typos earlier and matches the discipline applied to `RebornHarnessId`.

#### 10. `with_default` doesn't migrate the previous default into `additional` (nit)
`RebornDriverConfig::with_additional` skips duplicates against both `default` and `additional`. `with_default` overwrites `default` without checking what happens to the previous default. The chain `config.with_default(TextOnly).with_additional(Planned).with_default(Planned)` ends with `default=Planned, additional=[Planned]` — both. `all_choices` dedupes so no functional bug, but a maintainer reading the struct would expect `with_default` to either (a) keep `additional` in sync or (b) document that it's the caller's job. Add the comment.

#### 11. `ConversationIdentity::with_owner` lacks test coverage (nit)
`identity.rs` test `conversation_identity_builders_compose` exercises `for_tenant + with_agent + with_project`. Missing `with_owner`. Trivial to add.

#### 12. CLI stub tests don't pass `env_clear()` (nit)
Each smoke test sets `IRONCLAW_REBORN_HOME` and removes `USERPROFILE`, but the rest of the parent process env leaks in. The stubs short-circuit before any provider env lookup, so it's not a real issue today, but `env_clear()` would harden the tests against env-sensitive logic added later.

### Project-conventions check
- `.unwrap()` only in tests ✓
- Comments are dense but on-target — each documents the *trajectory* (today vs. after substrate) not the *code* ✓
- New crate dependencies: none — pure restructuring within `ironclaw_reborn_composition` ✓
- Boundary tests pinned ✓
- Sealing pattern (`SealedMarker` + private field) ✓

### Risk summary
**Low.** This PR is genuinely additive — every existing code path either stays the same or gains an additional code path behind a feature/flag. The two items worth picking up before too many sub-issues land on top of this shape are (1) bounding the `conversations` map and (2) the `Planned` driver / TOML cross-validation.

Strong forward-architecture work. The combination of shape-locked stubs + boundary tests + uniform `NotYetWired` errors gives sub-issues a clean drop-in target.

zmanian (commented):
Thanks for the detailed write-up. The narrative is clear and I appreciate the table mapping epic #3036 concepts to today's seams. That said, I want to push back hard on the framing. "Futureproofing" is exactly the kind of work I'm most skeptical of, and the PR-as-shipped bakes in a lot of shape we may regret.

## What is load-bearing today and worth keeping

- `RebornIdentityConfig` + `ConversationIdentity` + `new_conversation_for(...)` + the `conversations` HashMap in `RebornRuntime`. This is real, solves a real problem (the old code hardcoded `tenant_id = "reborn-cli"` and shared one `ThreadScope` across all conversations), and the refactor of `send_user_message` to look up per-conversation context is a genuine fix, not speculation. Keep this.
- The deletion of `actor_user_id` and `thread_scope` fields on `RebornRuntime` in favor of identity-driven scope construction is also good — that was a latent multi-tenant bug.

Cut those out and ship them on their own. The rest concerns me.

## Speculative surface I'd push back on

1. **`RebornHarnessSelection` / `RebornHarnessId` / `SendMessageOptions::harness_override`.** Zero consumers today. `build_reborn_runtime` logs a `warn!` and ignores it. `send_user_message_with_options` logs a `warn!` and ignores it. We are committing to (a) the harness-id stringly-typed namespace, (b) per-message override being the activation granularity (vs. per-thread or per-tenant), and (c) the name "harness" — all before the harness substrate exists and can talk back about what shape it actually wants. When sub-issue 4 from #3036 lands, the harness service will have opinions about activation scope that may not match this DTO. Better to add this field when the substrate lands and we know the shape, in one PR. Removing a no-op field later is cheap; reshaping a public field with consumers is expensive.

2. **`RebornDriverChoice::Planned`.** This is a variant that, if selected, returns `Err(InvalidArgument)` with a "not yet wired" message. So the only valid value today is `TextOnly`. A `Vec<RebornDriverChoice>` of `additional` drivers with one possible value, plus a `default` field, is solving zero problems today. When planned driver lands (#3651 — a separate slice), it brings its own host wiring requirements that may or may not fit this DTO. Drop `RebornDriverChoice::Planned` and `RebornDriverConfig` entirely; add a driver-config DTO at the point planned-driver actually wires.

3. **`RebornPolicyConfig`.** All three fields (`deployment_mode`, `default_profile`, `default_approval_policy`) hit `let _ = (&policy, &drivers);` and a `tracing::debug!`. Nothing reads them at request time. The comment honestly admits "field-level enforcement is downstream work." Same argument: add the input when there's a consumer. Today this is shape-bikeshedding without a tiebreaker — the `RuntimePolicyRepo` lookup path doesn't exist, so we're guessing what input it wants.

4. **`RebornAdminClient` + every DTO under it.** Six async methods, all returning `Err(NotYetWired { tracking_issue: "#3036" })`. `ApplyBlueprintRequest`, `DiffBlueprintRequest`, `InstallHarnessRequest`, `ActivateHarnessRequest`, `DeactivateHarnessRequest`, `HarnessActivationScope`, `HarnessDescriptor`, `ApplyMode`, `ApplyReport`, `ApplyChange`, `ApplyChangeAction`, `BlueprintSource` — every one of these is exported, every one is a guess at a shape the substrate will dictate. The PR description acknowledges this: "the parsed-blueprint DTO will live in the future `ironclaw_blueprint` crate ... For now this is a path/url placeholder so the admin surface already takes a typed argument and won't break when the parser lands." That's not stability, that's saving on one rename in a single PR while paying for it with ~400 lines of dead public API right now.

5. **`RebornAdminScope` sealing.** I like the *direction* — privilege gating belongs here — but the `SealedMarker` mechanism does nothing today. Any caller with the type in scope can call `RebornAdminScope::for_cli_admin("whatever")`. The string `audit_reason` has no enforcement. The seal only prevents non-`pub` field construction, and the only `pub` constructor accepts any string. This is the ergonomic shape of a privilege gate without the actual gating. When the auth crate lands, the constructor changes anyway, so this is essentially comment-driven design. Either ship it with real auth (credential parameter on `for_cli_admin`, even if the only valid credential today is "I'm the CLI binary") or wait.

6. **`reborn_runtime_surface_is_futureproof_for_epic_3036` boundary test.** This test reads source files as strings and `assert!(contains(...))` on struct field names and the existence of CLI stub files. I understand the intent — pin the surface so a contributor can't drift back to implicit shape — but it pins *today's guess at* the surface, and it pins it by string match on source. Any reshape (which sub-issue 2/3/4 may demand) trips this test and the contributor either modifies the test (rendering it ceremonial) or argues with it. A test like this only makes sense when the shape is settled by something external (an RFC, a wire format, an upstream API). Epic #3036 sub-issues haven't been written; the shape isn't settled. Drop this test or scope it tightly to the load-bearing identity layer.

7. **The CLI stub commands.** I'm more sympathetic here — `harness install/list/activate/deactivate` + `config apply/diff/watch` as discoverable stubs that point operators to the tracking issue is reasonable UX. But: shell completion as justification is thin (operator muscle memory has nothing to anchor on yet either), and locking the *flag shape* of these commands (`--thread`/`--project`/`--tenant`, `--require-signature`, `--interval-secs`) is the same kind of guess as #4. If you keep the stubs, drop the flags and accept positional args + free-form ones until the substrate constrains them.

## On the "concentrate the reshape" argument

The PR description argues this is cheaper than five future breaking changes. I'd argue the opposite. Today's PR commits to specific DTO shapes for 5+ subsystems whose substrate doesn't exist. Each substrate sub-issue will discover real constraints — what the parser actually returns, what the harness service actually needs to filter on, what the policy repo's per-request resolver actually keys off — and at that point either (a) the DTO matches by luck, or (b) we change it. Option (b) is the same breaking change we were trying to avoid, just earlier and with less information.

The cheapest version of futureproofing is **not doing it**, and adding each input field at the same PR that adds its consumer. That gives us tiebreakers (the consumer's API) instead of vibes (what we think the consumer's API will probably look like).

## What I'd ship instead

Split this into:

1. **Identity refactor** (~300 LOC): `RebornIdentityConfig`, `ConversationIdentity`, `new_conversation_for`, `conversations` map, removal of hardcoded `reborn-cli` tenant. Ship this now, it's a real fix.
2. **Api-version stamp** (optional, ~50 LOC): If you want the version-stamp discipline, add it minimally — one variant, mandatory field, fail-closed on major mismatch. Skip the `RebornRuntimeApiVersionError` enum elaboration until there's a second variant.
3. **Everything else**: defer to the sub-issue that adds the consumer.

## Substrate/binary boundary

To answer the explicit question: no, this PR doesn't entangle binary concerns into substrate. Everything new lives in `ironclaw_reborn_composition` and the CLI crate. The substrate crates (`ironclaw_reborn`, `ironclaw_turns`, `ironclaw_host_api`) are not touched. That much is clean.

## Bottom line

REQUEST CHANGES. The identity refactor is a clear win; please carve it out. The rest is too much shape committed too early on epic #3036 contracts that don't exist yet. I'd rather take five smaller breaking-change PRs that each follow their consumer than one large PR that pre-commits to all the shapes.

— Zaki

PR #3711, opened 2026-05-16
test(reborn): port product-live fixtures onto planned harness

serrrfirat:
## Summary

Stacks on #3710 and ports the product-live planned runtime fixtures onto the shared `ProductLiveAgentLoopHarness`.

This PR keeps production behavior unchanged. Its purpose is to prove that the harness from the previous PR can carry existing integration-style assertions instead of being a one-off happy-path smoke test.

## What this achieves

- Extends the harness with a pausable model gateway for cancellation tests.
- Adds public harness helpers for model-call observation, turn cancellation, and cancellation-probe observation.
- Ports the no-profile product-live reply fixture onto the harness.
- Ports the product-live cancellation fixture onto the harness through the public turn coordinator path.

## What this enables next

- Capability/tool execution tests can reuse the same submit/wait/history/cancel operations.
- Existing E2E fixtures can move one behavior at a time without reconstructing runtime composition boilerplate.
- Follow-up adapter wiring PRs can focus on real adapters rather than test setup.

## Coverage / validation

- `cargo test -p ironclaw_product_workflow --test planned_agent_loop_harness -- --nocapture`
- `cargo clippy -p ironclaw_product_workflow --all-targets -- -D warnings`

Coverage note: attempted `cargo llvm-cov -p ironclaw_product_workflow --test planned_agent_loop_harness --summary-only`, but local `llvm-tools-preview` is not installed in this worktree. The PR adds branch coverage for both completed and cancelled product-live runs through the shared harness.

### DIFF
--- a/crates/ironclaw_product_workflow/tests/planned_agent_loop_harness.rs
+++ b/crates/ironclaw_product_workflow/tests/planned_agent_loop_harness.rs
@@ -44,3 +44,88 @@ async fn product_live_harness_runs_planned_loop_and_persists_reply() {
 
     harness.shutdown().await;
 }
+
+#[tokio::test]
+async fn ported_product_live_fixture_uses_shared_harness_for_no_profile_reply() {
+    let harness = ProductLiveAgentLoopHarness::new(ProductLiveAgentLoopHarnessConfig {
+        assistant_reply: "planned product reply".to_string(),
+        user_id: "user:product-live".to_string(),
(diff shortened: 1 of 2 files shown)


serrrfirat:
Superseded by #3730, which restacks this product-live planned AgentLoop harness work on top of Henry's provider-tool-calls branch without carrying the older parallel stack forward.

PR #3713, opened 2026-05-16
test(reborn): add capability path to planned AgentLoop harness

serrrfirat:
## Summary

Stacks on #3711 and adds the first planned AgentLoop capability/tool path coverage through the product-live harness.

This PR does not cut over the app or gateway. It proves the planned runtime can run the actual agent loop shape beyond simple text replies:

1. product inbound message submits a run
2. planned loop calls the model
3. model returns a capability call
4. harness capability port is invoked
5. model is called again
6. final assistant reply is persisted to thread history

## Dependency

This PR also pulls in #3712 (`fix(reborn): trust result refs with durable replies`) because the capability completion path produces both `result_refs` and a durable assistant reply ref. Without that fix, product-live loop-exit evidence correctly rejects the run as `RecoveryRequired`.

Once #3712 lands in `reborn-integration` and the stack is refreshed, this PR can drop the copied fix commit and keep only the harness capability changes.

## What this enables

- Existing E2E fixtures can start asserting real agent/tool behavior rather than only text reply behavior.
- Adapter-wiring PRs can target real capability invocation seams with regression coverage already in place.
- The next slice can replace harness capability mocks with production composition adapters incrementally.

## Coverage / validation

- `cargo test -p ironclaw_product_workflow --test planned_agent_loop_harness -- --nocapture`
- `cargo clippy -p ironclaw_product_workflow --all-targets -- -D warnings`
- `cargo test -p ironclaw_reborn loop_exit_applier -- --nocapture`

Coverage note: local `cargo llvm-cov` remains unavailable because `llvm-tools-preview` is not installed in this worktree. This PR adds branch-level behavioral coverage for the capability path: model call -> capability invocation -> second model call -> persisted final reply.

### DIFF
--- a/crates/ironclaw_product_workflow/tests/planned_agent_loop_harness.rs
+++ b/crates/ironclaw_product_workflow/tests/planned_agent_loop_harness.rs
@@ -5,7 +5,12 @@ use ironclaw_reborn::planned_driver_factory::PLANNED_DEFAULT_PROFILE_ID;
 use ironclaw_threads::MessageStatus;
 use ironclaw_turns::TurnStatus;
 
-use support::planned_agent_loop::{ProductLiveAgentLoopHarness, ProductLiveAgentLoopHarnessConfig};
+use ironclaw_loop_support::HostManagedModelResponse;
+
+use support::planned_agent_loop::{
+    HarnessCapabilityConfig, ProductLiveAgentLoopHarness, ProductLiveAgentLoopHarnessConfig,
+    capability_call_response,
+};
 
 #[tokio::test]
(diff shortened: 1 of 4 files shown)


serrrfirat:
Opened the next stacked slice for production adapter wiring: #3714. It keeps the app/gateway cutover out of scope, uses the RebornServices host-runtime facade for capability execution, requires caller-supplied product-live adapters for cancellation/input/identity/policy/budget/safety, and preserves the existing product-live readiness gates instead of loosening them.

serrrfirat:
Superseded by #3730, which restacks this capability-harness work on top of Henry's provider-tool-calls branch and keeps the tool path in one replacement PR.

PR #3714, opened 2026-05-16
feat(reborn): add product-live planned adapter bundle

serrrfirat:
## Summary

Adds the next stacked slice after the planned AgentLoop capability harness: a composition-owned product-live adapter bundle for the planned runtime.

This PR does **not** cut app, gateway, or binary traffic over. It only gives the production composition layer a facade-shaped bundle that can be passed into `build_product_live_planned_runtime` once the eventual entrypoint supplies durable thread/checkpoint stores.

## What this wires

- Host-runtime capability execution through `HostRuntimeLoopCapabilityPortFactory`, using the `RebornServices` host-runtime facade.
- Capability surface resolution through an explicit allow set.
- Model route resolution through typed product-live route settings for default and optional mission slots.
- Cancellation, input queue, identity context, model policy guard, model budget accountant, and safety context as required caller-supplied product adapters.
- Fail-closed behavior when the host runtime facade is absent.

## Contract choices

- No app/gateway cutover in this PR.
- No local-only fallback is accepted for missing host runtime.
- Cancellation is **not** synthesized from a store here. The caller must provide a product-live-capable cancellation factory so the existing readiness probe remains meaningful.
- Policy, budget, input, identity, and safety adapters are required in the config instead of defaulting to no-op production behavior.
- The adapter uses existing runtime/loop-support facades; it does not construct raw capability ports outside `ironclaw_loop_support`.

## Enables

The next contributor can wire a production entrypoint by selecting the durable thread/checkpoint stores and passing this adapter bundle into the product-live planned runtime builder, without reaching into host-runtime internals or weakening the product-live readiness gate.

## Verification

- `cargo test -p ironclaw_reborn_composition --test product_live_adapters -- --nocapture`
- `cargo test -p ironclaw_reborn_composition`
- `cargo clippy -p ironclaw_reborn_composition --all-targets -- -D warnings`

Coverage command attempted but blocked by the local toolchain:

- `cargo llvm-cov -p ironclaw_reborn_composition --test product_live_adapters --summary-only`
- failed because `llvm-tools-preview` is not installed and `LLVM_COV` / `LLVM_PROFDATA` are not configured.

## Stack

Stacked on #3713 (`reborn/product-live-agentloop-capability-harness`). #3713 currently includes the result-ref evidence fix from #3712 until #3712 lands and the stack is refreshed.

### DIFF
--- a/crates/ironclaw_reborn_composition/src/lib.rs
+++ b/crates/ironclaw_reborn_composition/src/lib.rs
@@ -21,6 +21,7 @@
 mod error;
 mod factory;
 mod input;
+mod product_live_adapters;
 mod profile;
 mod readiness;
 mod runtime;
@@ -29,6 +30,10 @@ mod runtime_input;
 pub use error::RebornBuildError;
 pub use factory::{RebornServices, build_reborn_services};
 pub use input::RebornBuildInput;
+pub use product_live_adapters::{
+    ProductLiveModelRouteSettings, ProductLivePlannedRuntimeAdapterConfig,
+    ProductLivePlannedRuntimeAdapterError, ProductLivePlannedRuntimeAdapters, capability_allowlist,
+};
 pub use profile::{RebornCompositionProfile, RebornCompositionProfileParseError};
 pub use readiness::{RebornFacadeReadiness, RebornReadiness, RebornReadinessState};
 pub use runtime::{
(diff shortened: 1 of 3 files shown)


serrrfirat:
Superseded by #3730, which carries the product-live planned adapter bundle on top of Henry's provider-tool-calls branch.

PR #3715, opened 2026-05-16
feat(reborn): add product-live capability IO adapters

serrrfirat:
## Summary

Adds the first e2e-enabling adapter slice on top of #3714: composition-owned capability IO and scoped visible-capability request construction.

This is deliberately not an app/gateway cutover. It gives local/product-live runtime tests a real tool-call boundary without reaching into lower-level internals.

## TDD flow

1. Added failing tests for:
   - staged capability input resolution
   - run-scoped result ref materialization
   - cross-run input/result ref rejection
   - visible capability request construction from an actual `LoopRunContext`
2. Implemented the minimal adapter code to satisfy those tests.
3. Re-ran focused and package validation.

## What this wires

- `ProductLiveCapabilityIo`
  - stages JSON payloads behind `CapabilityInputRef`
  - implements `LoopCapabilityInputResolver`
  - implements `LoopCapabilityResultWriter`
  - materializes `LoopResultRef`s scoped to the current run
  - fails closed for cross-run refs
- `ProductLiveVisibleCapabilityRequestConfig`
- `visible_capability_request_for_run(...)`
  - derives tenant/agent/project/thread scope from `LoopRunContext`
  - keeps the visibility policy explicit
  - requires provider trust to be supplied by the caller

## Why this shape

For e2e, the immediate blocker is moving model-emitted tool calls through `HostRuntime` without test-only inline fakes. This PR keeps that adapter in the composition layer and uses existing loop-support/host-runtime facades. It does not invent durable result storage; the store is process-local and explicitly scoped to enabling local/e2e runtime execution. Durable production result storage remains a follow-up.

## Enables

The next stacked PR can assemble a runnable local product-live planned runtime that:

1. builds a scoped capability surface,
2. stages a model tool-call input,
3. dispatches through host-runtime capability execution,
4. stores the result behind a `LoopResultRef`, and
5. lets the model continue to a final reply.

## Verification

- `cargo test -p ironclaw_reborn_composition --test product_live_adapters -- --nocapture`
- `cargo test -p ironclaw_reborn_composition`
- `cargo clippy -p ironclaw_reborn_composition --all-targets -- -D warnings`

## Stack

Stacked on #3714.

PR #3716, opened 2026-05-17
test(reborn): exercise product-live builtin tool path

serrrfirat:
## What this PR does

This is the next stacked layer after #3715. It proves the product-live planned adapter bundle can drive a real host-runtime capability, not only staged IO and fake/recording ports.

The wiring path exercised here is:

1. `build_reborn_services(RebornCompositionProfile::LocalDev)` now registers the built-in first-party package in the local host-runtime extension registry.
2. The same local-dev composition installs the built-in first-party handlers and a host trust policy for the `builtin` provider.
3. `ProductLiveCapabilityIo` stages a run-scoped input ref for `builtin.echo`.
4. `visible_capability_request_for_run` scopes a host visible capability request to the planned `LoopRunContext`.
5. `ProductLivePlannedRuntimeAdapters::from_services` builds the host-runtime-backed `LoopCapabilityPortFactory`.
6. The loop capability port lists the real `builtin.echo` descriptor and invokes it through `HostRuntime`.
7. The completed runtime output is written back through `ProductLiveCapabilityIo` as a run-scoped result ref.

## Why this exists

The previous PR gave us the adapter bundle and process-local capability IO needed by product-live e2e tests. This PR verifies that the same bundle can cross the actual host-runtime capability boundary and execute a concrete built-in tool.

This enables the next PR to port an agent-loop e2e fixture that lets the planned model path request a tool call and consume the resulting `LoopResultRef`.

## Boundary fix included

While writing the failing test first, the planned driver hit a real boundary bug: `reborn:planned-default` is a valid `LoopDriverId`, but not a valid `ExtensionId` for host-runtime execution context. The loop-support facade now derives a valid execution principal by preserving valid driver ids and replacing `:` with `-` when needed. This keeps the conversion private to the capability-port boundary instead of exposing lower-level host-runtime internals to composition callers.

## TDD / verification

Initial failing proof:

```text
cargo test -p ironclaw_reborn_composition --test product_live_adapters local_dev_adapter_invokes_builtin_echo_through_host_runtime_port -- --nocapture
# failed because builtin echo was not visible through the product-live adapter surface
```

After implementation:

```text
cargo test -p ironclaw_loop_support
cargo test -p ironclaw_reborn_composition
cargo clippy -p ironclaw_loop_support -p ironclaw_reborn_composition --all-targets -- -D warnings
cargo fmt --check
LLVM_COV=$HOME/.rustup/toolchains/stable-aarch64-apple-darwin/lib/rustlib/aarch64-apple-darwin/bin/llvm-cov LLVM_PROFDATA=$HOME/.rustup/toolchains/stable-aarch64-apple-darwin/lib/rustlib/aarch64-apple-darwin/bin/llvm-profdata cargo llvm-cov -p ironclaw_loop_support -p ironclaw_reborn_composition --summary-only
```

Coverage summary for the touched package set:

```text
TOTAL regions: 81.98%
TOTAL lines:   80.89%
```

## Not included

This does not cut over the app, gateway, or binary. It also does not yet run the full planned AgentLoop with a model-produced tool call; that is the next stacked PR now that the real host-runtime tool boundary is proven.

### DIFF
--- a/crates/ironclaw_loop_support/src/capability_port.rs
+++ b/crates/ironclaw_loop_support/src/capability_port.rs
@@ -690,13 +690,7 @@ fn invocation_context_from_visible(
     allowed_effects: &[EffectKind],
 ) -> Result<ExecutionContext, AgentLoopHostError> {
     let mut context = base.clone();
-    let loop_driver_extension =
-        ExtensionId::new(run_context.loop_driver_id.as_str()).map_err(|_| {
-            AgentLoopHostError::new(
-                AgentLoopHostErrorKind::Internal,
-                "loop driver id could not be represented as an execution extension",
-            )
-        })?;
+    let loop_driver_extension = loop_driver_execution_extension_id(run_context)?;
     context.extension_id = loop_driver_extension.clone();
     context.runtime = capability.runtime;
     context.trust = trust;
@@ -720,6 +714,20 @@ fn invocation_context_from_visible(
         )
     })?;
     Ok(context)
+}
+
+fn loop_driver_execution_extension_id(
+    run_context: &LoopRunContext,
+) -> Result<ExtensionId, AgentLoopHostError> {
+    let raw = run_context.loop_driver_id.as_str();
+    ExtensionId::new(raw)
+        .or_else(|_| ExtensionId::new(raw.replace(':', "-")))
+        .map_err(|_| {
+            AgentLoopHostError::new(
+                AgentLoopHostErrorKind::Internal,
+                "loop driver id could not be represented as an execution extension",
+            )
+        })
 }
 
 fn invocation_grants_from_visible(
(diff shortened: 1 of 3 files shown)


serrrfirat:
Opened the next stacked PR: #3718.\n\nThis layer takes the #3716 direct host-runtime tool proof and runs it through the product-live planned AgentLoop harness: scripted model -> `builtin.echo` capability call -> host-runtime-backed product-live adapters -> run-scoped result materialization -> final assistant reply.\n\nImportant boundary note: the planned loop currently keeps capability result refs in loop state; it does not yet append them into the next model prompt. #3718 records and asserts the actual host-runtime JSON result in harness test support, and calls out prompt materialization as follow-up if an e2e fixture needs the model to inspect tool output before finalizing.\n\nValidation on #3718:\n- `cargo test -p ironclaw_product_workflow`\n- `cargo clippy -p ironclaw_product_workflow --all-targets -- -D warnings`\n- `cargo fmt --check`\n- llvm-cov product workflow: 81.25% line coverage / 83.39% region coverage

serrrfirat:
Superseded by #3730, which includes the product-live builtin tool-path coverage on top of Henry's provider-tool-calls branch.

PR #3718, opened 2026-05-17
test(reborn): run product-live harness with real builtin tool

serrrfirat:
## What this PR does

This is the next stacked layer after #3716. It upgrades the product-live planned AgentLoop harness so one fixture can execute a real host-runtime built-in tool instead of the recording capability port.

The new fixture proves this wiring:

1. Product workflow accepts a normal user message and submits the planned Reborn run.
2. The planned AgentLoop asks the model gateway for a response.
3. The scripted model response requests `builtin.echo` using the real visible surface version and a run-scoped input ref staged by `ProductLiveCapabilityIo`.
4. The harness capability factory builds `ProductLivePlannedRuntimeAdapters` for the actual `LoopRunContext` and delegates to the host-runtime-backed capability port.
5. `builtin.echo` executes through `HostRuntime` and writes its output back through `ProductLiveCapabilityIo`.
6. The loop continues and finalizes the assistant reply through the product workflow path.

This keeps the prior recording capability fixture for narrow driver tests, but adds a real-tool fixture for e2e porting.

## Why this exists

#3716 proved the adapter bundle can invoke a built-in tool directly through the host-runtime port. This PR proves the same path inside the product-live planned AgentLoop harness, which is the shape needed to port existing e2e fixtures with tools.

## Boundary notes

The fixture records delegated tool invocations and materialized JSON results in test support so assertions do not reach into production internals. The production composition still owns capability execution through the adapter bundle; the harness only supplies local-dev storage and scripted model responses.

Current planned-loop behavior does not yet append capability result references into the next model prompt. The test therefore asserts the actual host-runtime result was produced and that the loop continued to final reply. Prompt materialization of tool-result refs should be a follow-up if product e2e fixtures require the model to inspect tool output before finalizing.

## Coverage / validation

TDD first failure:

```text
cargo test -p ironclaw_product_workflow --test planned_agent_loop_harness product_live_harness_invokes_builtin_echo_through_host_runtime -- --nocapture
# failed to compile because the harness had no host-runtime capability mode
```

After implementation:

```text
cargo test -p ironclaw_product_workflow
cargo clippy -p ironclaw_product_workflow --all-targets -- -D warnings
cargo fmt --check
LLVM_COV=$HOME/.rustup/toolchains/stable-aarch64-apple-darwin/lib/rustlib/aarch64-apple-darwin/bin/llvm-cov LLVM_PROFDATA=$HOME/.rustup/toolchains/stable-aarch64-apple-darwin/lib/rustlib/aarch64-apple-darwin/bin/llvm-profdata cargo llvm-cov -p ironclaw_product_workflow --summary-only
```

Coverage summary:

```text
TOTAL regions: 83.39%
TOTAL lines:   81.25%
```

## Not included

No app/gateway/binary cutover. No durable production result storage beyond the existing local harness IO. No prompt materialization change for capability result refs.


serrrfirat:
Superseded by #3730, which keeps the real builtin-tool fixture in the restacked Henry-based e2e-enabling slice.

Issue #3622, opened 2026-05-14
Reborn: verify internal tool-result completion evidence and result refs


henrypark133:
## Background
The loop protocol has `LoopResultRef` and transcript APIs for capability result refs, and earlier briefs expect completed tool calls to produce `result_refs`. WS17 intentionally proves a reply-only completion path. The current Reborn thread/checkpoint evidence adapter rejects non-empty `result_refs`, and the text-only driver completes with an empty result-ref list, so product-visible tool-result completion evidence is not implemented yet.

## Evidence
- `crates/ironclaw_turns/src/run_profile/host.rs:1132-1164` defines `AppendCapabilityResultRef`, but the default `LoopTranscriptPort::append_capability_result_ref(...)` implementation is unsupported.
- `crates/ironclaw_reborn/src/loop_exit_applier.rs:148-150` documents that capability-result evidence remains untrusted until dedicated durable stores are wired.
- `crates/ironclaw_reborn/src/loop_exit_applier.rs:193-199` makes `ThreadCheckpointLoopExitEvidencePort::verify_completion_refs(...)` return `false` for any non-empty `result_refs`.
- `crates/ironclaw_reborn/src/text_loop_driver.rs:153-157` completes final replies with `result_refs: Vec::new()`.
- `docs/reborn/agent-loop-briefs/e2e-integration-tests.md:192-194` lists expected happy-path tests where tool calls produce `result_refs.len() == 1` or `== 2`, but those are broader E2E expectations rather than WS17 coverage.
- `crates/ironclaw_product_workflow/tests/inbound_turn_contract.rs:381-525` is the WS17 product-live test and only asserts a finalized assistant reply; it does not drive a tool call or verify a product-visible result ref.

## Why this blocks claiming agent loop is live
A tool-capable live loop needs durable, product-visible evidence that a capability result actually completed and belongs to the run before the loop exit is trusted. If non-empty `result_refs` are rejected, completed tool work cannot be part of trusted loop completion. If result refs are empty, product surfaces cannot correlate final replies with tool-result evidence. That blocks a live claim for tool-using agent behavior.

## Expected architecture direction
[...]

Issue #3620, opened 2026-05-14
Reborn: convert provider tool calls into ParentLoopOutput::CapabilityCalls


henrypark133:
## Background
The Reborn planned loop supports `ParentLoopOutput::CapabilityCalls`, and the lower-level provider stack already has tool-call response types. However, the production model gateway path used by Reborn currently calls the text-only provider completion API and rejects `FinishReason::ToolUse` as an invalid text-only response. That means real provider tool calls do not enter the agent loop as capability calls.

## Evidence
- `crates/ironclaw_loop_support/src/lib.rs:955-968` defines `HostManagedModelResponse` with `output: ParentLoopOutput`, but its only constructor is `assistant_reply(...)`.
- `crates/ironclaw_agent_loop/src/test_support/mod.rs:307-317` and `crates/ironclaw_agent_loop/src/test_support/mod.rs:804-813` show tests can script `ParentLoopOutput::CapabilityCalls`, proving the loop model exists.
- `crates/ironclaw_llm/src/provider.rs:419-428` defines `ToolCompletionResponse` with `tool_calls` and `finish_reason`.
- `crates/ironclaw_llm/src/provider.rs:461-465` exposes `complete_with_tools(...)` on `LlmProvider`.
- `crates/ironclaw_reborn/src/model_gateway.rs:166-175` builds a plain `CompletionRequest`, calls `provider.complete(...)`, and converts only that response.
- `crates/ironclaw_reborn/src/model_gateway.rs:501-517` maps `FinishReason::ToolUse` to `HostManagedModelErrorKind::InvalidRequest` with "model returned unsupported tool calls for a text-only loop".
- `crates/ironclaw_llm/src/provider.rs:146-160` and `crates/ironclaw_llm/src/provider.rs:284-301` show the shared provider message/call types already represent assistant tool calls and provider tool-call metadata.

## Why this blocks claiming agent loop is live
The agent loop is not live in the meaningful tool-capable sense if production model responses that ask for tools are rejected before reaching the loop executor. WS17 proves a reply-only path with a stub model gateway; it does not prove the real provider gateway can transform model tool calls into `CapabilityCalls`, dispatch capabilities, append result refs, and continue to a final reply.

## Expected architecture direction
[...]

PR #3720, opened 2026-05-17
[codex] verify durable tool result refs

henrypark133:
## Summary
- Adds a typed Reborn thread-service operation for finalized `ToolResultReference` transcript records, with in-memory/libSQL/Postgres parity.
- Writes validated safe-summary tool-result reference envelopes before executor result refs are recorded, including retry-success paths.
- Verifies completion `result_refs` against same-run finalized durable thread records, while reply refs now require finalized assistant messages.
- Moves tool-result envelope validation out of `contract.rs` into `ironclaw_threads::tool_result_reference` and rejects unsafe summaries at append/verification boundaries.
- Adds durable evidence-port coverage for libSQL reopen, Postgres wiring, and same `result_ref` isolation across different `turn_run_id` values.

Advances #3622.

## Scope
This PR establishes durable, typed tool-result evidence inside the Reborn thread-service boundary. It does not complete the provider-native round-trip work tracked separately: provider call IDs, provider tool names, capability linkage, reasoning/signature metadata, and assistant tool-call records are intentionally left for follow-up work.

## Tests
- `cargo fmt --check`
- `cargo check -p ironclaw_reborn --all-features --tests`
- `cargo test -p ironclaw_threads safe_summary -- --nocapture`
- `cargo test -p ironclaw_threads append_tool_result_reference_is_finalized_and_idempotent_per_run_result_ref -- --nocapture`
- `cargo test -p ironclaw_loop_support transcript_port_rejects_unsafe_tool_result_summary --test thread_loop_support_contract`
- `cargo test -p ironclaw_loop_support transcript_port_appends_tool_result_reference_envelope_idempotently --test thread_loop_support_contract -- --nocapture`
- `cargo test -p ironclaw_reborn loop_exit_applier::tests`
- `cargo test -p ironclaw_reborn thread_checkpoint_evidence_isolates_same_result_ref_across_runs`
- `cargo test -p ironclaw_reborn --features libsql-restart-tests libsql_thread_checkpoint_evidence_verifies_result_ref_after_reopen`
- `IRONCLAW_SKIP_POSTGRES_TESTS=1 cargo test -p ironclaw_reborn --features postgres-thread-tests postgres_thread_checkpoint_evidence_verifies_result_ref_after_reopen_when_configured`
- `IRONCLAW_SKIP_POSTGRES_TESTS=1 cargo test -p ironclaw_reborn --all-features thread_checkpoint_evidence -- --nocapture`
- `cargo test -p ironclaw_agent_loop calls_then_reply_completes --no-default-features -- --nocapture`
- `cargo test -p ironclaw_agent_loop retries_do_not_push_signatures_again --no-default-features`
- `cargo test -p ironclaw_architecture reborn_crate_dependency_boundaries_hold --no-default-features -- --nocapture`
- `python3 scripts/check_no_panics.py --base origin/reborn-integration --head HEAD`
- `cargo clippy -p ironclaw_threads -p ironclaw_reborn --tests --all-features -- -D warnings`
- `git diff --check`

## Follow-up
- Persist provider-native tool-call/result metadata for #3620, including provider call IDs, provider tool names, capability linkage, and reasoning/signature continuity.
- Add a typed assistant-tool-call record and map tool-result context back to providers through native tool/result roles rather than system-style transcript context.

## Notes
The Postgres evidence-port test compiles and its explicit skip path is covered in this branch. It still needs live Postgres validation in an environment with a configured test database.


zmanian (commented):
Reviewed against #3622 acceptance criteria. The shape is right - typed op, envelope validation pushed out of `contract.rs` into a dedicated module, and `verify_completion_refs` no longer fails closed on non-empty `result_refs`. A few things I want addressed before this comes out of draft.

**1. Dual-backend parity evidence is thinner than the summary claims.**

`durable_history_flow` in `db_session_thread_contract.rs` is shared by libSQL and Postgres test entry points and now exercises `append_tool_result_reference` + idempotency + envelope round-trip in `LoadContextWindow`. Good. But the *verification* path (`ThreadCheckpointLoopExitEvidencePort::verify_completion_refs` against finalized records) is only driven against `InMemorySessionThreadService` in `loop_exit_applier/tests`. There is no integration test that wires the libSQL or Postgres service through the evidence port. If `LoadContextWindow` ever differs from `list_thread_history` on either backend with respect to `tool_result_ref` / `content` envelope serialization, the verification helper silently returns `false` and completions fail closed in prod but pass in-memory. Please add at least one libSQL- and Postgres-feature-gated test that drives the full path: append tool result -> reopen service -> `verify_completion_refs` -> Ok(true).

**2. "Safe" in `ToolResultSafeSummary` is a denylist, not a redactor.**

`validate_tool_result_safe_summary` rejects on a hardcoded substring list (`secret`, `bearer `, `api_key`, `sk-` tokens, control chars, brace/bracket/backtick/slash). It does not redact, and it does not derive from the loop safe-summary policy referenced in the doc comment - it duplicates a stricter, hand-rolled list. Concerns:

- `case "secret"` collides with legitimate summaries like "fetched secrets manager status" or any product name containing "secret". A retried tool whose summary trips this list is silently dropped at `LoopSafeSummary::new` in `loop_support` (mapped to a generic `map_err(|_| ...)` - no diagnostic surfaces). What is the operator/user signal when a well-intended summary is rejected at append time?
- `sk-` token check splits on `is_ascii_alphanumeric() && != '-'`, so it catches `sk-live-token` (good) but also `sk-foo` in any non-key context. Acceptable, but please document this is intentional over-rejection.
- The path/payload delimiter ban (`{ } [ ] < > / \``) means a summary cannot contain a URL path or a JSON-ish phrase like "returned []". This is going to bite real tool authors. Either widen the policy or document the contract upstream so callers know to phrase summaries as prose.

If the intent is "this is replayed into model-visible context, so reject anything that looks like raw payload," I'd rather see this stated in the doc and the list curated against a real corpus rather than ad-hoc strings. At minimum: pull these constants into a named table with a comment justifying each entry, and unify with `LoopSafeSummary` rather than duplicating.

**3. Race / ordering on result_ref verification.**

`verify_tool_result_ref` requires `status == Finalized` AND `turn_run_id == expected_run_id` AND a parseable `ToolResultReferenceEnvelope` in `content`. Two questions:

- The PR claims "writes validated safe-summary envelopes *before* executor result refs are recorded." Where is the ordering invariant enforced? If the executor records a `LoopResultRef` and then the agent emits completion before the transcript append commits (in-memory vs. DB write ordering, or a failed append that gets retried), `verify_completion_refs` returns `false` and the loop exit is rejected as untrusted. Is that a hard failure that surfaces to the user, or does the loop retry? I don't see the recovery path in the diff.
- Idempotency is keyed on `(turn_run_id, result_ref)`. What happens if a tool with the same `result_ref` is legitimately re-emitted in a different `turn_run_id` (e.g. routine re-execution)? The contract test only covers same-run dedup. Add a negative test for cross-run isolation.

**4. "Fixes #3622" is overstated.**

The issue's acceptance criteria include "Product/gateway-facing test verifies a completed tool result is visible/correlatable in the user-facing transcript or activity stream without exposing raw tool payloads" and "Caller-level planned runtime test drives model tool call -> capability completion -> result ref append -> final reply -> trusted completed loop exit." Neither lands in this PR. The diff adds the durable evidence primitive and verification, but the text-only driver path in `text_loop_driver.rs:153-157` (`result_refs: Vec::new()`) is not touched, and there is no product_workflow test that observes a tool result through the gateway transcript. Mark this as "advances #3622" rather than "Fixes #3622" and keep the issue open for the driver wiring + product surface test, or split a follow-up issue and reference it explicitly in the description.

**Smaller items:**

- `tool_result_reference.rs` doc comment says "mirrors the loop safe-summary policy" but the policy is not in fact mirrored. Either import the shared validator or drop the claim.
- `ToolResultSafeSummary` has `Deserialize` that runs validation - good - but the `Serialize` is `transparent`, so a record persisted with stricter validation then re-deserialized under a looser future validator will silently re-accept. Worth a comment that the validator is the canonical schema for stored content.
- The 512-byte limit on `safe_summary` and `result_ref` is unmotivated. Why 512? Document the rationale or pull from a shared constant.
- `verify_tool_result_ref` parses the envelope on every history scan via `serde_json::from_str` inside an `any(...)`. For threads with many messages this is O(n) JSON parses on the completion hot path. Consider short-circuiting on `tool_result_ref.as_deref() == Some(result_ref.as_str())` before the parse - actually you do, the comment says "Cheap metadata checks run before this helper" but the `any` closure still parses for every candidate match. Fine for now, but flag for later if hot.

Net: the core mechanism is sound and the test surface against in-memory is solid. Two things block "ready for review" for me: (a) real dual-backend coverage of the verification path, not just the append path, and (b) honesty about scope vs. #3622. The safe-summary denylist deserves a second pass but is not a blocker.

Zaki

henrypark133:
@zmanian addressed the review feedback in `a60b062b3`.

- Added full evidence-port coverage for the durable verification path: libSQL now drives append -> reopen service -> `ThreadCheckpointLoopExitEvidencePort::verify_completion_refs` -> `Ok(true)`, and Postgres has the same feature-gated test path when a test database is configured.
- Added same-`result_ref` cross-run isolation coverage so idempotency remains scoped to `(turn_run_id, result_ref)` and a repeated result ref in another run is treated as distinct durable evidence.
- Cleaned up `ToolResultSafeSummary` by moving policy details into named constants, documenting that this is a model-visible stored-content contract rather than a redactor, documenting intentional `sk-...` over-rejection, and clarifying that the validator is the canonical schema for stored envelope content.
- Updated the PR body from `Fixes #3622` to `Advances #3622`, with the provider-native/tool-call metadata and product/gateway surface work called out as follow-up scope.

Validation after the changes:

- `cargo fmt --check`
- `git diff --check origin/reborn-integration`
- `cargo check -p ironclaw_reborn --all-features --tests`
- `IRONCLAW_SKIP_POSTGRES_TESTS=1 cargo test -p ironclaw_reborn --all-features thread_checkpoint_evidence -- --nocapture`
- `cargo test -p ironclaw_threads safe_summary -- --nocapture`
- `cargo test -p ironclaw_threads append_tool_result_reference_is_finalized_and_idempotent_per_run_result_ref -- --nocapture`
- `cargo test -p ironclaw_loop_support transcript_port_appends_tool_result_reference_envelope_idempotently --test thread_loop_support_contract -- --nocapture`
- `cargo test -p ironclaw_agent_loop calls_then_reply_completes --no-default-features -- --nocapture`
- `cargo test -p ironclaw_architecture reborn_crate_dependency_boundaries_hold --no-default-features -- --nocapture`

The only remaining caveat is that the Postgres evidence-port test was compiled and its explicit skip path was exercised locally; I did not run it against a live Postgres instance here.

crates/ironclaw_agent_loop/src/executor.rs
```diff
@@ -713,6 +713,7 @@ impl CanonicalAgentLoopExecutor {
     ) -> Result<BatchStep, AgentLoopExecutorError> {
         match outcome {
             CapabilityOutcome::Completed(result) => {
+                append_capability_result_ref(host, &result).await?;
```
serrrfirat, line 716: P1 correctness: this now writes durable result evidence, but the canonical executor still emits `LoopCompletionKind::NoReply` whenever there is no assistant ref, even when `state.result_refs` is non-empty. Default profiles keep `allow_no_reply_completion = false`, and `validate_completed_exit` rejects `NoReply` before durable result evidence can matter. A terminate-hint capability can therefore append durable evidence but still fail end-to-end as a protocol violation. Please add a caller-level `LoopExitApplier` test for result-only completion from the executor path and either avoid `NoReply` when durable result refs exist or intentionally update the validation policy.

crates/ironclaw_threads/src/tool_result_reference.rs
```diff
@@ -0,0 +1,156 @@
...
+        validate_tool_result_ref(&result_ref)?;
+        Ok(Self {
+            version: 1,
+            result_ref,
+            safe_summary,
+        })
+    }
+}
+
+fn validate_tool_result_ref(value: &str) -> Result<(), String> {
```
serrrfirat, line 82: P3 correctness: this validator does not mirror the canonical `LoopResultRef` rules. It accepts `result:` with an empty suffix, allows extra `:` characters, and allows up to 512 bytes while `LoopResultRef` is capped at 256. That lets the thread service persist finalized tool-result references that no legal `LoopCompleted.result_refs` can claim. Please reuse or exactly mirror `LoopResultRef` validation so durable evidence cannot be written in an unclaimable shape.

serrrfirat (commented):
Focused review pass for security/correctness/multi-tenancy. I found two issues that should be addressed before merge; details inline.

serrrfirat (changes requested):
Security/correctness/multi-tenancy review complete.\n\nBlocking finding:\n- P1 correctness: the canonical executor now persists durable result evidence, but result-only completion still emits LoopCompletionKind::NoReply. Default profiles reject NoReply before result-ref evidence is considered, so terminate-hint capability completions can still fail as protocol violations. See inline comment on crates/ironclaw_agent_loop/src/executor.rs.\n\nNon-blocking finding:\n- P3 correctness/schema drift: ToolResultReferenceEnvelope validation accepts result-ref shapes that canonical LoopResultRef cannot represent. See inline comment on crates/ironclaw_threads/src/tool_result_reference.rs.\n\nI did an additional pass specifically for forged transcript evidence, scope isolation, and tenant/user boundary checks. I did not find additional medium-or-higher security or multi-tenancy issues in the latest head.

serrrfirat:
Paranoid review pass focused on security correctness and multi-tenancy.

Findings:

1. **P1 correctness: result-only completion is still emitted as `NoReply`.**
   `CanonicalAgentLoopExecutor::completed_exit` still sets `LoopCompletionKind::NoReply` whenever `assistant_refs` is empty, even when `state.result_refs` contains durable tool-result evidence. Default run profiles disallow no-reply completion, so `validate_completed_exit` rejects the exit before the new durable result-ref evidence can be trusted. This breaks the terminate-hint/result-only completion path this PR is intended to enable.

   Suggested fix: treat non-empty `result_refs` as durable completion evidence in the completion kind/validation path, and add an executor-to-applier regression test for terminate-hint result-only completion.

2. **P3 correctness/schema consistency: stored tool result refs are looser than `LoopResultRef`.**
   `ToolResultReferenceEnvelope::new` validates `result_ref` with a local validator that allows empty suffixes, extra `:` characters, and 512 bytes, while canonical `LoopResultRef` requires a non-empty suffix, only `[A-Za-z0-9_.-]` after `result:`, and a 256-byte max. This can persist finalized thread records that no legal `LoopCompleted.result_refs` can claim.

   Suggested fix: reuse `LoopResultRef::new` at this boundary or mirror its exact validation rules.

Security/multi-tenancy notes: I did not find a tenant isolation bypass in the changed write/read/evidence paths. The thread service paths are scoped by `ThreadScope`, and evidence verification checks same run, finalized status, message kind, stored metadata, and parseable envelope content.

Verification run locally against the PR worktree:

```text
cargo test -p ironclaw_threads tool_result
cargo test -p ironclaw_loop_support transcript_port_appends_tool_result_reference_envelope_idempotently
cargo test -p ironclaw_reborn thread_checkpoint_evidence
cargo test -p ironclaw_agent_loop calls_then_reply_completes --no-default-features
```

All passed. P1 remains blocking; no additional medium-or-higher findings found in this pass.

zmanian (commented):
Re-review of 5 commits since my prior pass (9791c245..11999e0b). Going point by point against my prior findings, plus echoing what serrrfirat surfaced.

**1. Dual-backend parity for `verify_completion_refs` — Resolved.**

`a60b062b` adds `libsql_thread_checkpoint_evidence_verifies_result_ref_after_reopen` (gated on `libsql-restart-tests`) and `postgres_thread_checkpoint_evidence_verifies_result_ref_after_reopen_when_configured` (gated on `postgres-thread-tests`), both driving append → reopen service → `verify_completion_refs` → `Ok(true)`. The shared `create_durable_tool_result` / `assert_durable_tool_result_verifies` helpers route through the trait, so the libSQL and Postgres impls are exercised through the same evidence port as in-memory. The Postgres test has a documented `IRONCLAW_SKIP_POSTGRES_TESTS=1` skip path; acceptable, but note it was not run against a live DB in this branch per the author's own validation log.

**2. Safe-summary denylist — Partially resolved.**

`c9a15459` pulls the substrings into named constants (`SENSITIVE_SUMMARY_MARKERS`, `RAW_PAYLOAD_OR_PATH_DELIMITERS`), documents the intent ("boundary rejects summaries that look like raw payloads, paths, stack traces, or credentials") and explicitly calls out the `sk-...` over-rejection as intentional. Doc claim that it "mirrors the loop safe-summary policy" was dropped in favor of an honest stored-content-schema framing. Good.

Still open: no diagnostic surface when a legitimate summary trips the denylist — `loop_support` still uses `.map_err(|_| ...)` so an operator gets a generic rejection at append time with no signal about which marker fired. Not a merge blocker, but worth a structured error variant or at minimum a `debug!` log carrying the marker name. Leaving the wide-net substrings (`secret`, `password`) as-is is a deliberate call; I'll defer.

**3. Race/ordering + cross-run isolation — Partially resolved.**

`a60b062b` adds `thread_checkpoint_evidence_isolates_same_result_ref_across_runs`, which proves `(turn_run_id, result_ref)` keying lets the same ref verify under two distinct run IDs. That closes the cross-run negative-test gap I called out.

Still open: the *ordering* invariant ("envelopes written before result refs are recorded") still has no test. The executor appends durable evidence before adding the executor result ref, but there is no recovery path documented for a transient append failure — `verify_completion_refs` will simply return `false` and the exit will be rejected as untrusted. The behavior may be correct (fail closed), but the recovery story is undocumented. I'll accept this as a follow-up rather than a blocker.

**4. "Fixes #3622" overstatement — Resolved.**

PR body now reads "Advances #3622" with the driver-wiring and product/gateway surface called out as explicit follow-up scope. `text_loop_driver.rs:155` still emits `result_refs: Vec::new()` — that's now correctly out of scope rather than misleadingly claimed.

**5. Smaller items — Mostly resolved.**

- 512-byte limits pulled into `MAX_TOOL_RESULT_REF_BYTES` / `MAX_TOOL_RESULT_SUMMARY_BYTES` constants with formatted error messages. Rationale still not in a doc comment, but the constants make a future tightening trivial. Accept.
- `Serialize transparent` schema-drift concern: doc comment now states "The validator below is the canonical stored-content schema for this type." Accept.
- NUL/control-character rejection added in `c9a15459` with a dedicated test (`safe_summary_rejects_control_characters`). Strict improvement.
- The per-message JSON parse in `verify_tool_result_ref` was not touched. Flagged for later if hot; not a blocker.

---

**New findings (independent of my prior pass; reinforces serrrfirat's review):**

**P1 — Result-only completion is still emitted as `NoReply` (blocking).**

`crates/ironclaw_agent_loop/src/executor.rs:1318-1326` in `completed_exit`:

```rust
completion_kind: if state.assistant_refs.is_empty() {
    LoopCompletionKind::NoReply
} else {
    LoopCompletionKind::FinalReply
},
```

This ignores `state.result_refs` entirely. Default run profiles reject `NoReply` in `validate_completed_exit` before the new durable result-ref evidence is even consulted, so the terminate-hint / result-only completion path this PR is built to enable still fails the protocol gate. The whole point of the durable evidence machinery is to let a non-empty `result_refs` set carry a completed exit when no assistant reply was emitted. Fix: treat `result_refs` non-empty as durable-completion evidence in the kind selection (e.g. a `ResultOnly` kind or extending `FinalReply` semantics) and add an executor→applier regression test that drives a terminate-hint, result-only completion through to a trusted exit.

**P3 — `validate_tool_result_ref` is looser than canonical `LoopResultRef::new`.**

`crates/ironclaw_threads/src/tool_result_reference.rs` allows 512 bytes, empty suffix after `result:`, and extra `:` characters. Canonical `LoopResultRef::new` requires non-empty suffix, only `[A-Za-z0-9_.-]` after the prefix, and a 256-byte cap. Result: a finalized record can be persisted with a `result_ref` shape that no legal `LoopCompleted.result_refs` can claim — silent drift between stored evidence and verifier input. Fix is small: call `LoopResultRef::new` at the envelope boundary, or mirror its rules exactly. Not blocking on its own but I'd rather see it land before merge given how cheap it is.

---

Net: the parity/scope blockers from my prior pass are addressed. The denylist diagnostic surface and ordering test are accepted as non-blocking follow-ups. The new P1 from serrrfirat is the real merge blocker — without it, the durable evidence path this PR builds is unreachable in the default profile. I'd like to see (a) `completed_exit` respect `result_refs` for the completion-kind selection with an executor→applier regression test, and (b) the schema-drift fix in `validate_tool_result_ref` before this comes out of draft.

Zaki

crates/ironclaw_threads/src/tool_result_reference.rs
henrypark133: Addressed in `ff54296c4`.

`ToolResultReferenceEnvelope` now mirrors the canonical `LoopResultRef` grammar without adding a new `ironclaw_threads -> ironclaw_turns` dependency:

- 256-byte max, matching loop refs.
- Must start with `result:`.
- Must include a non-empty opaque suffix.
- Suffix allows only ASCII letters, digits, `_`, `-`, or `.`.
- Extra `:` characters are rejected.

I added tests for valid canonical refs and the invalid shapes called out here: empty suffix, extra `:`, slash, spaces, control chars, and over-256-byte values.

Validation:

- `cargo test -p ironclaw_threads tool_result_ref`
- `cargo clippy --all --tests --examples --all-features -- -D warnings`
- GitHub Code Style / clippy is green on `ff54296c4`.

henrypark133:
@serrrfirat @zmanian the two latest review findings are addressed in `ff54296c4`.

- Added explicit `LoopCompletionKind::ResultOnly` instead of overloading `DelegatedResult`; the executor now emits `ResultOnly` for completed exits with durable `result_refs` and no assistant refs, while `NoReply` remains reserved/profile-gated for exits with no durable reply/result evidence.
- Added result-only contract/applier/executor coverage, including validation with `allow_no_reply_completion = false`.
- Tightened `ToolResultReferenceEnvelope` result-ref validation to mirror `LoopResultRef` without introducing a `threads -> turns` dependency.
- Updated the Reborn loop-exit contract doc for the new result-only completion vocabulary.

Local and CI validation are green, including all-features clippy. I also had a read-only subagent re-check PR scope and Reborn boundaries; it found no blockers or contract violations. The only unchanged caveat is no fresh live-Postgres run in this branch.

serrrfirat (approved):

serrrfirat (approved):
## Review: durable tool result refs

Re-reviewed current head `ff54296` after the prior blocker fixes. The shape is now right: result-only completion has explicit vocabulary, durable result refs are verified against finalized same-run thread evidence, and the PR body is honest that this advances #3622 rather than closes the whole provider/product surface.

### Strengths
- `LoopCompletionKind::ResultOnly` fixes the prior default-profile failure mode where durable result evidence was emitted but still classified as `NoReply`.
- `ToolResultReferenceEnvelope` now mirrors the canonical `LoopResultRef` grammar without adding the forbidden `threads -> turns` dependency.
- The durable verification path now has libSQL reopen coverage and a Postgres feature-gated path; the remaining live-Postgres caveat is stated clearly.
- Scope is now explicit: provider-native roundtrip metadata and product/gateway visibility are follow-up work.

### Remaining Non-Blocking Notes
1. Safe-summary rejection diagnostics are still generic at the loop transcript boundary. `append_capability_result_ref` maps both `LoopSafeSummary` and `ToolResultSafeSummary` failures to `"tool result reference summary is not safe"`, so operators cannot tell whether a legitimate summary tripped `secret`, a path delimiter, `sk-...`, or another marker. This is acceptable for merge, but I’d like a follow-up structured error/log that carries the rejector category without echoing the rejected value.

2. The Postgres evidence-port test is compiled and has an explicit skip path, but this branch still has no live Postgres validation. Fine as long as the PR does not claim live DB parity was exercised.

### Risk
The main correctness blockers from the previous passes are addressed. The remaining risk is operational ergonomics around over-rejected summaries, not trust-boundary correctness.

Approve.

PR #3721, opened 2026-05-17
Gate Reborn personal context by run profile

### LINKED ISSUE #3692: Reborn: add policy-gated personal identity and heartbeat prompt context
## Context

PR #3649 intentionally keeps WS-15 scoped to stable identity-file prompt context. The implementation wires stable files through `HostIdentityContextSource` and `LoopContextBundle.identity_messages`, but defers two identity/context surfaces that need additional prompt-context shape and run-context policy signals:

- `USER.md` and `context/assistant-directives.md`
- `HEARTBEAT.md`

The WS-15 docs now explicitly call these out as deferred so default-driver follow-up work does not accidentally treat them as shipped behavior.

## Design direction

Provider APIs do not give us a neutral first-class "context" lane with separate authority semantics:

- OpenAI Chat Completions expects prompt-visible material in `messages` roles such as `developer`/`system`, `user`, `assistant`, and `tool`.
- OpenAI Responses separates `instructions` from `input`, conversation state, and previous response references, but still does not provide a generic append-only personal-context lane.
- Anthropic Messages has top-level `system` plus `messages`; it does not have a generic `context` prompt field.

Reborn should therefore own this distinction in the prompt contract. We need a dedicated prompt-context field for user-specific/personal context, plus profile/run policy that decides whether that field is populated for a given prompt construction.

Expected contract shape:

- Add an explicit user/personal context lane, separate from transcript and generic memory, e.g. `personal_context_snippets` or `user_context`.
- Populate it only through a host-owned source that can enforce scope: profile policy, run kind, channel/surface, single-user versus shared/group context, and any redaction/summarization requirements.
- Keep it lower authority than stable identity/system instructions unless content is intentionally promoted into `identity_messages` by contract.
- Render it into provider-native prompt shapes during prompt assembly, since provider APIs do not preserve this lane for us.
- Default to fail-closed exclusion when policy is missing or ambiguous.

### 1. Personal/profile-derived identity files

`USER.md` and `context/assistant-directives.md` should not be injected as trusted raw system prompt content until the run context can express whether personal/profile-derived context is allowed for the current run.

Why:
- These files can contain user/profile-specific content.
- Future shared/group contexts need an explicit privacy policy before personal context is included.
- Without that signal, the safe behavior is fail-closed exclusion.
- A dedicated user-context lane lets prompt assembly preserve authority boundaries instead of smuggling personal context through transcript messages or stable identity files.

Expected direction:
- Add an explicit run/profile policy signal, e.g. whether personal identity/profile context is allowed for the current thread/run.
- Add a dedicated user/personal context field to the Reborn prompt context contract.
- Update the host context source to include or exclude personal files based on that policy.
- Map included personal context into the final provider prompt at the appropriate authority level.
- Add caller-level tests proving personal files are excluded for shared/group contexts and included only when policy allows them.

### 2. Volatile heartbeat context

`HEARTBEAT.md` should remain out of the stable `identity_messages` prefix. It is frequently-changing context and would break prompt-prefix stability if included with stable identity files.

Why:
- Stable identity context is intended to be byte-stable within a run.
- `HEARTBEAT.md` can contain timestamped or proactive findings that change often.
- Heartbeat context needs a run-kind/heartbeat signal so it is injected only for appropriate runs.

Expected direction:
- Add an explicit heartbeat/run-kind signal to `LoopRunContext` or an equivalent host-owned request shape.
- Add a volatile instruction/context source evaluated per context load, after the stable identity prefix.
- Keep heartbeat content out of `identity_messages` unless a future contract explicitly separates stable and volatile buckets.
- Add caller-level tests proving heartbeat is excluded from stable identity context and included only for heartbeat-eligible runs.

## Acceptance criteria

- Reborn has a dedicated user/personal prompt-context lane separate from transcript, stable identity, and generic memory.
- Personal/profile-derived files are policy-gated and fail closed by default.
- Profile/run policy controls whether prompts are constructed with user-specific context.
- Shared/group runs do not receive `USER.md` or assistant directives unless explicitly authorized.
- `HEARTBEAT.md` does not destabilize the stable prompt prefix.
- Provider rendering preserves the intended authority boundary when mapping Reborn context lanes into OpenAI/Anthropic request shapes.
- Tests drive the caller path, not only helper functions.
- WS-15 / Reborn docs are updated to reflect the final contract.

Related: #3649

henrypark133:
## Summary
- add `ResolvedRunProfile.personal_context_policy` with backward-compatible default exclusion
- keep builtin and planned default profiles excluded by default; explicitly allowed profiles remain direct-authority only, while shared authority downgrades them to excluded
- gate `USER.md` and `context/assistant-directives.md` through the resolved profile, without owner-id reauthorization
- re-check cached identity refs on dereference so stale personal refs fail closed under excluded profiles
- emit a safe `context` driver-note milestone when personal context is admitted, carrying source count + sanitized source labels only
- make personal-context admission milestones best-effort, retryable after sink failure, and deduped after successful publish
- preserve and document the existing `HEARTBEAT.md` default-loop exclusion as routine/mission/proactive context

Fixes #3692

## Follow-up split
This PR closes the default-loop personal identity / heartbeat contract in #3692. The stricter owner/shared-context authorization layer and privileged profile-upgrade dimension are intentionally split to #3745, because they need owner/surface authority threaded through both identity candidate admission and identity ref dereference rather than a partial check in this PR.

## Verification
- `cargo test -p ironclaw_turns -p ironclaw_loop_support`
- `cargo test -p ironclaw_reborn planned_driver_factory --lib`
- `cargo test -p ironclaw_loop_support --test thread_loop_support_contract personal_context`
- `cargo test -p ironclaw_turns --test run_profile_contract`
- `git diff --check`

## Review follow-up
- addressed the 2026-05-18 review comment by aligning personal-context admission milestone failure with best-effort milestone behavior and adding sink-failure coverage
- added a safety comment for future personal identity path labels to avoid basename collisions and forbidden safe-summary tokens
- added explicit profile-policy assertions for excluded named profiles
- confirmed #3745 is filed and open for owner-aware personal context authorization

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -63,6 +63,3 @@ tests/fixtures/llm_traces/live/github_dev_workflow_full_loop.log
 # tests locally. Only the .json fixtures for each scenario are checked
 # in; the .log files are local debugging artifacts.
 tests/fixtures/llm_traces/live/*.log
-
-# Multi-agent code review findings (local only, never commit)
-.review/
--- a/.gitignore
+++ b/.gitignore
@@ -63,3 +63,6 @@ tests/fixtures/llm_traces/live/github_dev_workflow_full_loop.log
 # tests locally. Only the .json fixtures for each scenario are checked
 # in; the .log files are local debugging artifacts.
 tests/fixtures/llm_traces/live/*.log
+
+# Multi-agent code review findings (local only, never commit)
+.review/
--- a/crates/ironclaw_loop_support/src/identity_context.rs
+++ b/crates/ironclaw_loop_support/src/identity_context.rs
@@ -157,6 +157,8 @@ pub enum HostIdentityContextBuildError {
     UnknownIdentityFile,
     #[error("identity context path is invalid")]
     InvalidIdentityFile,
+    #[error("identity context denied by run policy")]
+    PolicyDenied,
     /// Reserved for a future hard-limit mode. Currently, `build_identity_messages_for_run`
     /// truncates silently on budget overflow rather than returning this error.
     #[error("identity context budget exceeded")]
@@ -171,7 +173,7 @@ impl HostIdentityContextBuildError {
     pub fn into_host_error(self) -> AgentLoopHostError {
         let kind = match self {
             Self::SourceUnavailable => AgentLoopHostErrorKind::Unavailable,
-            Self::UnknownIdentityFile | Self::InvalidIdentityFile => {
+            Self::UnknownIdentityFile | Self::InvalidIdentityFile | Self::PolicyDenied => {
                 AgentLoopHostErrorKind::PolicyDenied
             }
             Self::ContextBudgetExceeded => AgentLoopHostErrorKind::BudgetExceeded,
@@ -194,11 +196,11 @@ pub async fn build_identity_messages(
     build_identity_messages_for_run(&candidates, run_context, mode, budget)
 }
 
-fn identity_candidate_allowed_for_run(
-    candidate: &HostIdentityContextCandidate,
+pub fn identity_applicability_allowed_for_run(
+    applicability: IdentityApplicability,
     run_context: &LoopRunContext,
 ) -> bool {
-    match candidate.applies_when {
+    match applicability {
         IdentityApplicability::Always | IdentityApplicability::OnCodeAct => true,
         IdentityApplicability::OnPersonalContextAllowed => {
             match run_context.resolved_run_profile.personal_context_policy {
@@ -218,7 +220,7 @@ pub fn build_identity_messages_for_run(
     let mut out = Vec::with_capacity(candidates.len());
     let mut used = 0u32;
     for candidate in candidates {
-        if !identity_candidate_allowed_for_run(candidate, run_context) {
+        if !identity_applicability_allowed_for_run(candidate.applies_when, run_context) {
             continue;
         }
         if !applies(candidate.applies_when, mode) {
--- a/crates/ironclaw_loop_support/src/identity_context.rs
+++ b/crates/ironclaw_loop_support/src/identity_context.rs
@@ -422,6 +422,17 @@ mod tests {
         assert!(messages.is_empty());
     }
 
+    #[test]
+    fn identity_budget_rejects_zero_token_ceiling() {
+        let error = IdentityBudget::new(0).unwrap_err();
+
+        assert_eq!(error, HostIdentityContextBuildError::BudgetMisconfigured);
+        assert_eq!(
+            error.into_host_error().kind,
+            AgentLoopHostErrorKind::Internal
+        );
+    }
+
     #[tokio::test]
     async fn installed_trust_summary_only() {
         let context = run_context().await;
--- a/crates/ironclaw_loop_support/src/identity_context.rs
+++ b/crates/ironclaw_loop_support/src/identity_context.rs
@@ -357,10 +357,40 @@ mod tests {
     use ironclaw_host_api::{TenantId, ThreadId};
     use ironclaw_turns::{
         RunProfileResolutionRequest, RunProfileResolver, TurnId, TurnRunId, TurnScope,
-        run_profile::{InMemoryRunProfileResolver, LoopRunContext},
+        run_profile::{InMemoryRunProfileResolver, LoopRunContext, PersonalContextPolicy},
     };
 
     use super::*;
+
+    #[tokio::test]
+    async fn identity_applicability_allowed_for_run_all_variants_covering_excluded_and_allowed() {
+        let excluded_context = run_context().await;
+        let mut allowed_context = run_context().await;
+        allowed_context.resolved_run_profile.personal_context_policy =
+            PersonalContextPolicy::Allowed;
+
+        for applicability in [
+            IdentityApplicability::Always,
+            IdentityApplicability::OnCodeAct,
+        ] {
+            assert!(identity_applicability_allowed_for_run(
+                applicability,
+                &excluded_context
+            ));
+            assert!(identity_applicability_allowed_for_run(
+                applicability,
+                &allowed_context
+            ));
+        }
+        assert!(!identity_applicability_allowed_for_run(
+            IdentityApplicability::OnPersonalContextAllowed,
+            &excluded_context
+        ));
+        assert!(identity_applicability_allowed_for_run(
+            IdentityApplicability::OnPersonalContextAllowed,
+            &allowed_context
+        ));
+    }
 
     #[tokio::test]
     async fn filters_by_applies_when() {
@@ -484,6 +514,35 @@ mod tests {
             serde_json::to_vec(&second).unwrap()
         );
         assert_eq!(source.calls.load(Ordering::SeqCst), 2);
+    }
+
+    #[tokio::test]
+    async fn build_identity_messages_for_run_detailed_mixed_candidates_excluded_covering_filtering()
+    {
+        let context = run_context().await;
+        let candidates = vec![
+            trusted("AGENTS.md", "agent", IdentityApplicability::Always),
+            trusted(
+                "USER.md",
+                "private user profile",
+                IdentityApplicability::OnPersonalContextAllowed,
+            ),
+        ];
+
+        let outcome = build_identity_messages_for_run_detailed(
+            &candidates,
+            &context,
+            PromptMode::TextOnly,
+            IdentityBudget::default(),
+        )
+        .unwrap();
+
+        assert_eq!(outcome.messages.len(), 1);
+        assert_eq!(
+            outcome.messages[0].safe_summary,
+            "identity file AGENTS.md available"
+        );
+        assert!(outcome.admitted_personal_context_paths.is_empty());
     }
 
     fn trusted(
--- a/crates/ironclaw_loop_support/src/identity_context.rs
+++ b/crates/ironclaw_loop_support/src/identity_context.rs
@@ -194,17 +194,6 @@ pub async fn build_identity_messages(
     build_identity_messages_for_run(&candidates, run_context, mode, budget)
 }
 
-fn filter_identity_candidates_for_run(
-    candidates: &[HostIdentityContextCandidate],
-    run_context: &LoopRunContext,
-) -> Vec<HostIdentityContextCandidate> {
-    candidates
-        .iter()
-        .filter(|candidate| identity_candidate_allowed_for_run(candidate, run_context))
-        .cloned()
-        .collect()
-}
-
 fn identity_candidate_allowed_for_run(
     candidate: &HostIdentityContextCandidate,
     run_context: &LoopRunContext,
@@ -226,21 +215,12 @@ pub fn build_identity_messages_for_run(
     mode: PromptMode,
     budget: IdentityBudget,
 ) -> Result<Vec<LoopContextMessage>, AgentLoopHostError> {
-    build_identity_messages_from_candidates(
-        &filter_identity_candidates_for_run(candidates, run_context),
-        mode,
-        budget,
-    )
-}
-
-fn build_identity_messages_from_candidates(
-    candidates: &[HostIdentityContextCandidate],
-    mode: PromptMode,
-    budget: IdentityBudget,
-) -> Result<Vec<LoopContextMessage>, AgentLoopHostError> {
     let mut out = Vec::with_capacity(candidates.len());
     let mut used = 0u32;
     for candidate in candidates {
+        if !identity_candidate_allowed_for_run(candidate, run_context) {
+            continue;
+        }
         if !applies(candidate.applies_when, mode) {
             continue;
         }
--- a/crates/ironclaw_loop_support/src/identity_context.rs
+++ b/crates/ironclaw_loop_support/src/identity_context.rs
@@ -260,6 +260,8 @@ fn applies(applicability: IdentityApplicability, mode: PromptMode) -> bool {
     match applicability {
         IdentityApplicability::Always => true,
         IdentityApplicability::OnCodeAct => mode == PromptMode::CodeAct,
+        // Personal-context applicability is policy-only today. Add a combined
+        // variant if a future personal directive must also be mode-specific.
         IdentityApplicability::OnPersonalContextAllowed => true,
     }
 }
--- a/crates/ironclaw_loop_support/src/identity_context.rs
+++ b/crates/ironclaw_loop_support/src/identity_context.rs
@@ -3,7 +3,8 @@ use ironclaw_memory::DEFAULT_PROMPT_PROTECTED_PATHS;
 use ironclaw_turns::{
     LoopMessageRef,
     run_profile::{
-        AgentLoopHostError, AgentLoopHostErrorKind, LoopContextMessage, LoopRunContext, PromptMode,
+        AgentLoopHostError, AgentLoopHostErrorKind, LoopContextMessage, LoopRunContext,
+        PersonalContextPolicy, PromptMode,
     },
 };
 use thiserror::Error;
@@ -90,6 +91,7 @@ pub enum IdentityTrustLevel {
 pub enum IdentityApplicability {
     Always,
     OnCodeAct,
+    OnPersonalContextAllowed,
 }
 
 #[derive(Debug, Clone, PartialEq, Eq)]
@@ -155,7 +157,7 @@ pub enum HostIdentityContextBuildError {
     UnknownIdentityFile,
     #[error("identity context path is invalid")]
     InvalidIdentityFile,
-    /// Reserved for a future hard-limit mode. Currently, `build_identity_messages_from_candidates`
+    /// Reserved for a future hard-limit mode. Currently, `build_identity_messages_for_run`
     /// truncates silently on budget overflow rather than returning this error.
     #[error("identity context budget exceeded")]
     ContextBudgetExceeded,
@@ -189,10 +191,49 @@ pub async fn build_identity_messages(
         .load_identity_candidates(run_context, mode)
         .await
         .map_err(HostIdentityContextBuildError::into_host_error)?;
-    build_identity_messages_from_candidates(&candidates, mode, budget)
+    build_identity_messages_for_run(&candidates, run_context, mode, budget)
 }
 
-pub fn build_identity_messages_from_candidates(
+fn filter_identity_candidates_for_run(
+    candidates: &[HostIdentityContextCandidate],
+    run_context: &LoopRunContext,
+) -> Vec<HostIdentityContextCandidate> {
+    candidates
+        .iter()
+        .filter(|candidate| identity_candidate_allowed_for_run(candidate, run_context))
+        .cloned()
+        .collect()
+}
+
+fn identity_candidate_allowed_for_run(
+    candidate: &HostIdentityContextCandidate,
+    run_context: &LoopRunContext,
+) -> bool {
+    match candidate.applies_when {
+        IdentityApplicability::Always | IdentityApplicability::OnCodeAct => true,
+        IdentityApplicability::OnPersonalContextAllowed => {
+            match run_context.resolved_run_profile.personal_context_policy {
+                PersonalContextPolicy::Excluded => false,
+                PersonalContextPolicy::Allowed => true,
+            }
+        }
+    }
+}
+
+pub fn build_identity_messages_for_run(
+    candidates: &[HostIdentityContextCandidate],
+    run_context: &LoopRunContext,
+    mode: PromptMode,
+    budget: IdentityBudget,
+) -> Result<Vec<LoopContextMessage>, AgentLoopHostError> {
+    build_identity_messages_from_candidates(
+        &filter_identity_candidates_for_run(candidates, run_context),
+        mode,
+        budget,
+    )
+}
+
+fn build_identity_messages_from_candidates(
     candidates: &[HostIdentityContextCandidate],
     mode: PromptMode,
     budget: IdentityBudget,
@@ -239,6 +280,7 @@ fn applies(applicability: IdentityApplicability, mode: PromptMode) -> bool {
     match applicability {
         IdentityApplicability::Always => true,
         IdentityApplicability::OnCodeAct => mode == PromptMode::CodeAct,
+        IdentityApplicability::OnPersonalContextAllowed => true,
     }
 }
 
--- a/crates/ironclaw_loop_support/src/identity_context.rs
+++ b/crates/ironclaw_loop_support/src/identity_context.rs
@@ -100,6 +100,12 @@ pub struct HostIdentityMessageContent {
     pub content: String,
 }
 
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct IdentityMessageBuildOutcome {
+    pub messages: Vec<LoopContextMessage>,
+    pub admitted_personal_context_paths: Vec<String>,
+}
+
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub struct IdentityBudget {
     pub token_ceiling: u32,
@@ -217,7 +223,17 @@ pub fn build_identity_messages_for_run(
     mode: PromptMode,
     budget: IdentityBudget,
 ) -> Result<Vec<LoopContextMessage>, AgentLoopHostError> {
+    Ok(build_identity_messages_for_run_detailed(candidates, run_context, mode, budget)?.messages)
+}
+
+pub fn build_identity_messages_for_run_detailed(
+    candidates: &[HostIdentityContextCandidate],
+    run_context: &LoopRunContext,
+    mode: PromptMode,
+    budget: IdentityBudget,
+) -> Result<IdentityMessageBuildOutcome, AgentLoopHostError> {
     let mut out = Vec::with_capacity(candidates.len());
+    let mut admitted_personal_context_paths = Vec::new();
     let mut used = 0u32;
     for candidate in candidates {
         if !identity_applicability_allowed_for_run(candidate.applies_when, run_context) {
@@ -231,13 +247,19 @@ pub fn build_identity_messages_for_run(
             break;
         }
         used = used.saturating_add(cost);
+        if candidate.applies_when == IdentityApplicability::OnPersonalContextAllowed {
+            admitted_personal_context_paths.push(candidate.name.as_str().to_string());
+        }
         out.push(LoopContextMessage {
             message_ref: candidate.message_ref.clone(),
             role: LOOP_SYSTEM_ROLE.to_string(),
             safe_summary: candidate.safe_summary.clone(),
         });
     }
-    Ok(out)
+    Ok(IdentityMessageBuildOutcome {
+        messages: out,
+        admitted_personal_context_paths,
+    })
 }
 
 pub fn identity_message_ref(
--- a/crates/ironclaw_loop_support/src/identity_context.rs
+++ b/crates/ironclaw_loop_support/src/identity_context.rs
@@ -103,7 +103,7 @@ pub struct HostIdentityMessageContent {
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct IdentityMessageBuildOutcome {
     pub messages: Vec<LoopContextMessage>,
-    pub admitted_personal_context_paths: Vec<String>,
+    pub admitted_personal_context_paths: Vec<IdentityFileName>,
 }
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
@@ -248,7 +248,7 @@ pub fn build_identity_messages_for_run_detailed(
         }
         used = used.saturating_add(cost);
         if candidate.applies_when == IdentityApplicability::OnPersonalContextAllowed {
-            admitted_personal_context_paths.push(candidate.name.as_str().to_string());
+            admitted_personal_context_paths.push(candidate.name.clone());
         }
         out.push(LoopContextMessage {
             message_ref: candidate.message_ref.clone(),
--- a/crates/ironclaw_loop_support/src/lib.rs
+++ b/crates/ironclaw_loop_support/src/lib.rs
@@ -206,8 +206,9 @@ where
                             .map_err(HostIdentityContextBuildError::into_host_error)
                     })
                     .await?;
-                identity_context::build_identity_messages_from_candidates(
+                identity_context::build_identity_messages_for_run(
                     candidates,
+                    &self.run_context,
                     mode,
                     self.identity_budget,
                 )?
--- a/crates/ironclaw_loop_support/src/lib.rs
+++ b/crates/ironclaw_loop_support/src/lib.rs
@@ -329,18 +329,7 @@ fn personal_context_admitted_summary(
 ) -> Result<LoopSafeSummary, AgentLoopHostError> {
     let source_labels = admitted_paths
         .iter()
-        .map(|path| {
-            path.as_str()
-                .rsplit('/')
-                .next()
-                .unwrap_or(path.as_str())
-                .chars()
-                .filter(|character| {
-                    character.is_ascii_alphanumeric() || matches!(character, '.' | '_' | '-')
-                })
-                .collect::<String>()
-        })
-        .filter(|label| !label.is_empty())
+        .filter_map(|path| personal_context_source_label(path.as_str()))
         .collect::<Vec<_>>()
         .join(" ");
     let summary = if source_labels.is_empty() {
@@ -358,6 +347,21 @@ fn personal_context_admitted_summary(
             format!("personal context milestone summary invalid: {reason}"),
         )
     })
+}
+
+fn personal_context_source_label(path: &str) -> Option<String> {
+    let basename = path
+        .rsplit(['/', '\\'])
+        .next()
+        .filter(|label| !label.is_empty())
+        .unwrap_or(path);
+    let label = basename
+        .chars()
+        .filter(|character| {
+            character.is_ascii_alphanumeric() || matches!(character, '.' | '_' | '-')
+        })
+        .collect::<String>();
+    (!label.is_empty()).then_some(label)
 }
 
 /// Thread-backed transcript adapter for text-only assistant replies.
@@ -1529,5 +1533,48 @@ fn safe_model_summary(kind: HostManagedModelErrorKind) -> &'static str {
         HostManagedModelErrorKind::CredentialUnavailable => "model credentials are unavailable",
         HostManagedModelErrorKind::Unavailable => "model service is unavailable",
         HostManagedModelErrorKind::Cancelled => "model request was cancelled",
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn personal_context_admitted_summary_empty_paths_uses_count_only() {
+        let summary = personal_context_admitted_summary(&[]).unwrap();
+
+        assert_eq!(summary.as_str(), "personal context admitted count 0");
+    }
+
+    #[test]
+    fn personal_context_admitted_summary_uses_safe_basenames_only() {
+        let paths = vec![
+            IdentityFileName::new("USER.md").unwrap(),
+            IdentityFileName::new("context/assistant-directives.md").unwrap(),
+        ];
+
+        let summary = personal_context_admitted_summary(&paths).unwrap();
+
+        assert_eq!(
+            summary.as_str(),
+            "personal context admitted count 2 sources USER.md assistant-directives.md"
+        );
+        assert!(!summary.as_str().contains("context/assistant-directives.md"));
+        assert!(!summary.as_str().contains('/'));
+        assert!(!summary.as_str().contains('\\'));
+    }
+
+    #[test]
+    fn personal_context_source_label_drops_empty_and_separator_only_labels() {
+        assert_eq!(
+            personal_context_source_label(r"private\USER.md").as_deref(),
+            Some("USER.md")
+        );
+        assert_eq!(
+            personal_context_source_label("context/%2Fassistant-directives.md").as_deref(),
+            Some("2Fassistant-directives.md")
+        );
+        assert_eq!(personal_context_source_label("///"), None);
     }
 }
--- a/crates/ironclaw_loop_support/src/lib.rs
+++ b/crates/ironclaw_loop_support/src/lib.rs
@@ -39,7 +39,8 @@ pub use capability_surface_filter::{
 pub use identity_context::{
     HostIdentityContextBuildError, HostIdentityContextCandidate, HostIdentityContextSource,
     HostIdentityMessageContent, IdentityApplicability, IdentityBudget, IdentityFileName,
-    IdentityTrustLevel, build_identity_messages, identity_message_ref,
+    IdentityTrustLevel, build_identity_messages, identity_applicability_allowed_for_run,
+    identity_message_ref,
 };
 pub use input_port::HostQueueLoopInputPort;
 pub use input_queue::{HostInputBatch, HostInputEnvelope, HostInputQueue, HostInputQueueError};
--- a/crates/ironclaw_loop_support/src/lib.rs
+++ b/crates/ironclaw_loop_support/src/lib.rs
@@ -40,7 +40,8 @@ pub use identity_context::{
     HostIdentityContextBuildError, HostIdentityContextCandidate, HostIdentityContextSource,
     HostIdentityMessageContent, IdentityApplicability, IdentityBudget, IdentityFileName,
     IdentityMessageBuildOutcome, IdentityTrustLevel, build_identity_messages,
-    identity_applicability_allowed_for_run, identity_message_ref,
+    build_identity_messages_for_run_detailed, identity_applicability_allowed_for_run,
+    identity_message_ref,
 };
 pub use input_port::HostQueueLoopInputPort;
 pub use input_queue::{HostInputBatch, HostInputEnvelope, HostInputQueue, HostInputQueueError};
@@ -100,6 +101,8 @@ where
 struct IdentityCandidateCache {
     text_only: OnceCell<Vec<HostIdentityContextCandidate>>,
     codeact: OnceCell<Vec<HostIdentityContextCandidate>>,
+    text_only_personal_context_admitted: OnceCell<()>,
+    codeact_personal_context_admitted: OnceCell<()>,
 }
 
 impl IdentityCandidateCache {
@@ -107,6 +110,8 @@ impl IdentityCandidateCache {
         Self {
             text_only: OnceCell::new(),
             codeact: OnceCell::new(),
+            text_only_personal_context_admitted: OnceCell::new(),
+            codeact_personal_context_admitted: OnceCell::new(),
         }
     }
 
@@ -114,6 +119,13 @@ impl IdentityCandidateCache {
         match mode {
             PromptMode::TextOnly => &self.text_only,
             PromptMode::CodeAct => &self.codeact,
+        }
+    }
+
+    fn personal_context_admitted_cell_for_mode(&self, mode: PromptMode) -> &OnceCell<()> {
+        match mode {
+            PromptMode::TextOnly => &self.text_only_personal_context_admitted,
+            PromptMode::CodeAct => &self.codeact_personal_context_admitted,
         }
     }
 }
@@ -221,8 +233,11 @@ where
                     mode,
                     self.identity_budget,
                 )?;
-                self.publish_personal_context_admitted(&outcome.admitted_personal_context_paths)
-                    .await?;
+                self.publish_personal_context_admitted(
+                    mode,
+                    &outcome.admitted_personal_context_paths,
+                )
+                .await;
                 outcome.messages
             }
             None => Vec::new(),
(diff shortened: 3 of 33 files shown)


crates/ironclaw_loop_support/src/identity_context.rs
```diff
@@ -189,10 +191,49 @@ pub async fn build_identity_messages(
...
+    run_context: &LoopRunContext,
+    mode: PromptMode,
+    budget: IdentityBudget,
+) -> Result<Vec<LoopContextMessage>, AgentLoopHostError> {
+    build_identity_messages_from_candidates(
+        &filter_identity_candidates_for_run(candidates, run_context),
+        mode,
+        budget,
+    )
+}
```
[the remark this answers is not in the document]
henrypark133: Addressed in 89d0753af: build_identity_messages_for_run now filters by run policy and prompt mode in a single pass, so it avoids the intermediate filtered Vec and candidate cloning.

crates/ironclaw_reborn/src/planned_driver_factory.rs
```diff
@@ -169,7 +170,8 @@ pub fn planned_default_profile_definition() -> Result<RunProfileDefinition, RunP
         checkpoint_schema_id,
         planned_driver_checkpoint_schema_version(),
         capability_surface_profile_id,
-    ))
+    )
+    .with_personal_context_policy(PersonalContextPolicy::Allowed))
```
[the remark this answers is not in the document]
henrypark133: Keeping reborn-planned-default explicitly Allowed is intentional: this is the normal user-facing default loop, and existing behavior expects USER.md-style profile context there. The fail-closed compatibility boundary is still preserved because missing/old ResolvedRunProfile payloads deserialize to personal_context_policy = Excluded, and shared/group/routine/mission profiles must explicitly resolve to Excluded rather than using the personal default.

zmanian (commented):
Overall direction is right — fail-closed default, single policy bit on `ResolvedRunProfile`, gate applied both at the workspace source and again in `build_identity_messages_for_run` as defense-in-depth. The serde `#[serde(default)]` + the explicit `resolved_profile_deserializes_old_payload_with_personal_context_excluded` test correctly handles old payloads, and the fingerprint includes the new policy so caches don't bleed across modes. A few things I'd want addressed before this lands:

1. Default safety / call-site audit. `ResolvedRunProfile::legacy_compatibility` and the two named profiles in `resolver.rs` (`interactive_profile`, `long_running_mission_profile`) all hard-code `Excluded`. That is the right default, but the `interactive_profile` is the *current* user-facing interactive run today. With this PR merged, the existing interactive path stops seeing `USER.md` and `context/assistant-directives.md` and only the *planned* default Reborn profile sees them. Please confirm explicitly in the PR description which production turn entrypoints resolve to `planned_default_profile_definition()` vs `interactive_profile()` after this lands, and whether there is a window where the current user loop silently loses `USER.md`. If yes, that is a behavior change worth calling out in the release notes; if no, the test in `planned_driver_factory.rs` should be paired with an integration assertion that the no-profile production resolver actually returns `Allowed`.

2. Opt-in visibility to the user. `personal_context_policy = Allowed` on the planned default profile means a third-party extension or any caller that requests "planned default Reborn" implicitly gets `USER.md` and `context/assistant-directives.md` rendered into the model prompt, with no surface for the user to see that their profile is being exfiltrated to whichever LLM the run picks. The "## User Profile Context — Informational; not authoritative runtime policy." preface is good for the model's understanding of authority, but it does nothing for the *human*. Two concrete asks: (a) emit a progress/observability event when personal context is injected, so the UI/audit log can show "this turn included your USER.md", and (b) on the resolver side, treat `Allowed` as a privileged dimension (similar to `PrivilegedRunProfileDimension`) so a caller cannot silently upgrade a request from `Excluded` to `Allowed` without authority. Right now `with_personal_context_policy` is an unguarded builder.

3. Test coverage of the gating logic. The unit coverage is good (`run_profile_contract`, `thread_loop_support_contract`, `workspace_identity_context_excludes_personal_files_until_policy_allows`, `planned_driver_factory`). What's missing is a negative test in `ironclaw_reborn` proving that a *non-default* resolved profile actually produces `Excluded` end-to-end through `text_only_host_factory_threads_identity_source_to_prompt_and_model`-style integration, and a test that the workspace source's outer-loop policy gate is not bypassable if a hostile `HostIdentityContextSource` returns `OnPersonalContextAllowed` candidates anyway — that's the whole point of the second gate in `identity_candidate_allowed_for_run`, and there's no test that exercises it (the unit test in `thread_loop_support_contract` uses the static source but the workspace source already filters, so the defense layer is untested).

4. Minor: `applies()` in `identity_context.rs` returns `true` for `OnPersonalContextAllowed` under any `PromptMode`. That's fine today, but it means if someone later adds a `CodeAct`-only personal directive they'll need a new applicability variant. Worth a comment.

5. Doc nit: the prompt-context-assembly brief still has both "stays excluded until run-context privacy policy exists" (old framing) and the new framing in adjacent paragraphs; a pass to make sure no stale text survives would help future readers.

The HEARTBEAT.md exclusion in this PR is a documentation/test assertion rather than a code change — the file was already never wired into the default identity surface. That's fine, but the PR description should say "documents existing exclusion" rather than implying new gating code, otherwise reviewers will look for a kill-switch that isn't there.

Zaki

src/workspace/reborn_identity_context.rs
```diff
@@ -94,15 +109,31 @@ impl WorkspaceIdentityContextSource {
...
-            if let Some(candidate) = self.candidate_for_path(path).await? {
+            if let Some(candidate) = self
+                .candidate_for_path(path, applicability_for_path(path))
+                .await?
+            {
                 candidates.push(candidate);
             }
         }
+
+        if run_context.resolved_run_profile.personal_context_policy
```
serrrfirat, line 138: `load_identity_candidates` now gates personal files, but the actual materialization sink below (`resolve_identity_message_content`) still ignores `run_context` and returns any cached `LoopMessageRef`. `WorkspaceIdentityContextSource` is shared through the runtime host factory and caches refs in `loaded_identity_content`, so once an allowed run has loaded `USER.md`/`context/assistant-directives.md`, the source itself no longer fails closed for an excluded run that asks to resolve the same ref. Today the prompt authority helps in the normal path, but this policy boundary is split across discovery and dereference, which is easy to bypass from another host path or a future resume/materialization flow.

Please enforce `PersonalContextPolicy` when resolving content too, e.g. store the identity path/applicability alongside the cached content and reject `PERSONAL_IDENTITY_PATHS` unless the current run profile is `Allowed`. A regression should populate the cache with an allowed context, then assert that resolving that personal ref under an excluded context returns `None`/`PolicyDenied`.

zmanian (commented):
Re-review of 3 new commits since 89d0753a (`bba56e3c9`, `8bd9eab04`, `20c8f8868`). Branch was force-rebased onto `2a8dc7d5e` but the relevant policy code is unchanged across the rebase.

**Status of prior findings**

1. **Call-site audit / regression window — Partially addressed.** Author replied (to the codex bot at `planned_driver_factory.rs:174`) that `reborn-planned-default = Allowed` is intentional as the normal user-facing default loop and that the fail-closed boundary is preserved via `#[serde(default)]` deserializing to `Excluded`. That answers the *intent* question but does not answer my concrete ask: which production turn entrypoints actually resolve to `planned_default_profile_definition()` today vs `interactive_profile()`, and is there a window where an existing live user loop now silently drops `USER.md`? The new `assert_eq!(snapshot.personal_context_policy, Allowed)` in `planned_driver_factory.rs:302` is good as a unit guard but it's not the integration assertion I asked for (no-profile production resolver returning `Allowed` end-to-end). Codex bot raised the matching shared/group concern at `inbound_turn.rs:353-360` (planned default fires when `requested_run_profile: None`, with no shared/direct discriminator). Author dismissed it; I am not yet convinced. Please add a one-line statement in the PR description naming the production resolver path and confirming the shared-thread case routes through an `Excluded` profile, not the planned default.

2. **Opt-in visibility / privileged dimension — Not addressed.** No observability/progress event when personal context is injected, and `with_personal_context_policy` is still an unguarded builder on `RunProfileDefinition`. The "## User Profile Context" preface helps the model, not the human. This is the highest-leverage remaining gap and I'd like at least the progress event before merge — even a single `LoopProgressEvent` variant fired from `build_identity_messages_for_run` when any `OnPersonalContextAllowed` candidate is admitted would be enough to drive UI/audit. The privileged-dimension ask can be a follow-up issue if you'd rather not expand scope, but please file it.

3. **Defense-in-depth gate test — Addressed.** Commit `20c8f8868` adds `text_only_host_factory_excludes_personal_identity_when_profile_excludes_it` in `crates/ironclaw_reborn/tests/loop_driver_host.rs:2909`, which drives a `StaticIdentityContextSource` returning a personal candidate under an `Excluded` profile through the actual factory-built host and asserts `identity_message_count == 0` and no `msg:identity.` refs. That is exactly the defense-layer regression I asked for. The renamed `context_port_defense_gate_excludes_personal_identity_from_host_source_by_default_policy` also makes the intent explicit. Good.

4. **HEARTBEAT.md docs framing — Acknowledged in PR description.** Fine.

5. **Doc nit (stale framing) — Not visibly addressed in this diff range; low priority.**

6. **Minor `applies()` comment — Addressed** in commit `20c8f8868` with the inline comment about adding a combined variant when a personal directive needs mode-gating.

**New finding from this round (P0, blocking)**

`WorkspaceIdentityContextSource::resolve_identity_message_content` still ignores `run_context` and returns any cached `LoopMessageRef` from `loaded_identity_content` (see `src/workspace/reborn_identity_context.rs:142-150`). This is the same boundary serrrfirat flagged at `src/workspace/reborn_identity_context.rs:125` at 10:41Z UTC — the `load_identity_candidates` gate filters discovery, but the dereference path does not. Because `WorkspaceIdentityContextSource` is shared through the runtime host factory and the cache is keyed only on `LoopMessageRef`, once an `Allowed` run has loaded `USER.md`/`context/assistant-directives.md`, an `Excluded` run that resolves the same ref (via resume, materialization, or any other host path that reuses the source) gets the personal content back. The new defense gate in `build_identity_messages_for_run` only fires when `OnPersonalContextAllowed` is *declared on the candidate* — a stale or re-fetched ref that was previously surfaced as personal will still resolve.

This is split policy enforcement: discovery is gated by policy, dereference is not. Please either (a) store the path/applicability alongside the cached content and reject `PERSONAL_IDENTITY_PATHS` from `resolve_identity_message_content` unless the run profile is `Allowed`, or (b) key the cache on `(run_profile_personal_context_policy, message_ref)` so an excluded run cannot read an allowed run's slot. Add a regression: populate the cache via an allowed run, then assert resolving the same personal ref under an excluded run returns `Ok(None)` (or a `PolicyDenied` error).

**Verdict**

Hold for finding 2 (observability) and the P0 from finding 7 (resolve-path policy). Finding 1's clarification can land as a PR-description edit. Finding 3 is cleanly closed.

Zaki

src/workspace/reborn_identity_context.rs
henrypark133: Addressed in `813551157`.

The workspace identity cache now stores the cached content together with its `IdentityApplicability`, and `resolve_identity_message_content` re-checks that applicability against the current `LoopRunContext` before returning cached content. If an excluded run tries to dereference a personal identity ref that was populated by an allowed run, it now returns `HostIdentityContextBuildError::PolicyDenied` instead of returning the cached `USER.md`/assistant-directives content.

Added regressions:
- `workspace_identity_context_denies_cached_personal_ref_when_policy_excludes`: allowed context loads `USER.md`, excluded context resolves the same cached ref and gets `PolicyDenied`.
- `model_port_rejects_policy_denied_identity_ref_before_gateway_call`: caller-level model materialization fails with `PolicyDenied` before the model gateway is called.

Verified with:
- `cargo test -p ironclaw_loop_support model_port_rejects_policy_denied_identity_ref_before_gateway_call`
- `cargo test --features libsql workspace_identity_context_denies_cached_personal_ref_when_policy_excludes`
- `git diff --check`

henrypark133:
@zmanian I pushed `813551157` to address the resolve-path cache policy blocker from your re-review.

Summary:
- `WorkspaceIdentityContextSource` now caches identity content with its `IdentityApplicability`.
- `resolve_identity_message_content` now re-checks that applicability against the current `LoopRunContext` and returns `PolicyDenied` for stale personal refs under an excluded profile.
- The shared applicability predicate is reused for both candidate filtering and dereference enforcement.
- Added both the direct cache regression and a caller-level model-port regression proving the gateway is not called after identity dereference is policy-denied.

Verified:
- `cargo test -p ironclaw_loop_support model_port_rejects_policy_denied_identity_ref_before_gateway_call`
- `cargo test --features libsql workspace_identity_context_denies_cached_personal_ref_when_policy_excludes`
- `git diff --check`

This addresses the P0 split discovery/dereference enforcement issue. The observability/progress-event item from finding 2 is still separate unless we want to expand this PR further.

serrrfirat:
Maintainer-style re-review at `8135511577983b3661f25aa23628e59dda795a00`.

The prior P0 cache/dereference concern looks addressed. `WorkspaceIdentityContextSource` now caches identity content with its `IdentityApplicability`, and `resolve_identity_message_content` re-checks that applicability against the current `LoopRunContext` before returning content. I verified the stale-ref regression through both the direct workspace test and the model-port caller path.

Remaining finding:

| Severity | Category | File:Line | Finding | Suggested fix |
|---|---|---|---|---|
| Medium | Privacy / observability | `crates/ironclaw_loop_support/src/identity_context.rs:223` / `crates/ironclaw_loop_support/src/lib.rs:195` | When an `OnPersonalContextAllowed` candidate is admitted, `USER.md` / assistant directives can be added into the prompt path silently. The model gets the informational prefix, but the user/audit surface still cannot tell that personal context was included. This leaves the earlier opt-in visibility concern unresolved. | Emit a safe milestone/progress/audit signal when any personal-context candidate is admitted, carrying only safe metadata such as count/path names, not raw content. The natural caller-side place is around identity message construction in `ThreadBackedLoopContextPort::load_loop_context`. |

Verification run locally:

- `cargo test -p ironclaw_turns --test run_profile_contract`
- `cargo test -p ironclaw_loop_support --test thread_loop_support_contract personal_identity`
- `cargo test -p ironclaw_loop_support --test thread_loop_support_contract model_port_rejects_policy_denied_identity_ref_before_gateway_call`
- `cargo test -p ironclaw_reborn planned_driver_factory --lib`
- `cargo test -p ironclaw_reborn --test loop_driver_host text_only_host_factory_excludes_personal_identity_when_profile_excludes_it`
- `cargo test --lib --features libsql workspace_identity_context_excludes_personal_files_until_policy_allows`
- `cargo test --lib --features libsql workspace_identity_context_denies_cached_personal_ref_when_policy_excludes`
- `git diff origin/reborn-integration...origin/pr/3721 --check`

Verdict: correctness/privacy gating looks materially improved after the cache-policy fix. I would still ask for the observability signal before merge unless we explicitly accept that as a follow-up.

serrrfirat:
Extended this PR with the remaining #3692 closeout items that fit the current design:

- emits a safe `context` driver-note milestone when personal context is admitted, using source count + sanitized source labels only
- keeps raw `USER.md` / assistant-directive contents and full nested paths out of the milestone/audit surface
- wires the milestone sink through the Reborn text-only host path
- documents the final default-loop contract: `USER.md` and `context/assistant-directives.md` are profile-gated personal context, while `HEARTBEAT.md` stays out of the default planned/interactive loop prefix and belongs to routine/mission/proactive semantics

Verification run on commit `313bd369`:

- `cargo test -p ironclaw_turns --test run_profile_contract -- --nocapture`
- `cargo test -p ironclaw_loop_support --test thread_loop_support_contract context_port_emits_safe_milestone_when_personal_identity_is_admitted -- --nocapture`
- `cargo test -p ironclaw_loop_support --test thread_loop_support_contract personal_identity -- --nocapture`
- `cargo test -p ironclaw_reborn --test loop_driver_host text_only_host_factory_excludes_personal_identity_when_profile_excludes_it -- --nocapture`
- `cargo test --lib --features libsql workspace_identity_context -- --nocapture`
- `cargo test -p ironclaw_turns --test agent_loop_host_contract loop_host_refs_are_bounded_opaque_tokens -- --nocapture`
- `cargo fmt --check && git diff --check`

Follow-up note: I did not add an `OwnerOnly` / thread-owner-aware personal-context policy or a separate privileged-profile dimension in this PR. Those need a small extension PR/issue if we want them, because the current `HostIdentityContextSource` dereference contract only carries `LoopRunContext`; a correct implementation should thread owner/authority context through candidate admission and ref dereference together instead of adding a partial check here.

zmanian (commented):
Reviewed against head `313bd369` (the "observe admitted personal context" commit) on top of the prior two rounds.

## Prior findings — status

| # | Finding | Status |
|---|---|---|
| Z1 | Call-site / regression-window audit for `interactive_profile()` vs `planned_default_profile_definition()` | **Partial.** `implicit_default_resolves_to_planned_driver` (planned_driver_factory.rs:294-309) asserts the *no-requested-profile* path resolves to `Allowed` end-to-end. The shared-thread / `requested_run_profile: None` concern from the codex thread at inbound_turn.rs:353-360 still has no integration assertion that group-chat submissions land on `Excluded`. A one-line PR-description note naming the production resolver path for shared threads would close this. |
| Z2a | Opt-in visibility / progress event when personal context is admitted | **Addressed.** `ThreadBackedLoopContextPort::publish_personal_context_admitted` fires a `LoopDriverNoteKind::Context` driver-note milestone via the configured sink (lib.rs:245-259); summary carries only count + sanitized basenames. `context_port_emits_safe_milestone_when_personal_identity_is_admitted` asserts both inclusion of `USER.md` / `assistant-directives.md` and explicit absence of raw content and full nested paths. |
| Z2b | `with_personal_context_policy` is an unguarded builder; no privileged-dimension gating | **Deferred** to #3745 per PR description. Reasonable given the dependency on the owner/authority threading; please make sure #3745 is filed and referenced once. |
| Z3 | Defense-in-depth gate test | **Addressed** in `text_only_host_factory_excludes_personal_identity_when_profile_excludes_it` (loop_driver_host.rs:2909-2951) and `context_port_defense_gate_excludes_personal_identity_from_host_source_by_default_policy`. |
| Z4 | `applies()` semantics for `OnPersonalContextAllowed` | **Addressed** with the explanatory comment at identity_context.rs:287-289. |
| Z6 / serrrfirat P0 | Cache/dereference split-policy enforcement | **Addressed.** `CachedIdentityMessageContent` now stores `(content, applies_when)` (reborn_identity_context.rs:38-42); `resolve_identity_message_content` re-checks via the shared `identity_applicability_allowed_for_run` and returns `PolicyDenied` (lines 150-168); regression `workspace_identity_context_denies_cached_personal_ref_when_policy_excludes` (lines 322-358) walks an allowed→excluded cache hit and asserts `PolicyDenied`. Worth also noting the supporting model-port test in `thread_loop_support_contract.rs`. |
| Z5 | Stale "until run-context privacy policy exists" framing | Could not confirm a sweep in this diff range. Low priority. |

## New observations against `313bd369`

**1. Milestone-publication failure is now load-bearing, inconsistent with the rest of the milestone surface.** `publish_personal_context_admitted` (lib.rs:245-259) propagates `sink.driver_note(...)` errors all the way up out of `load_loop_context`. By contrast, `ThreadBackedLoopTranscriptPort::emit_assistant_reply_finalized` (lib.rs:476-486) treats a sink error as best-effort: it logs at `debug` and returns `Ok(())`. The transcript port's posture matches the inline comment on `LoopProgressPort::emit_loop_progress` saying "progress events are best-effort and must not be used as recoverability-critical durability markers."

If the intent here is to fail closed on personal-context admission when audit is unavailable (a defensible privacy stance), please add a comment saying so explicitly, since this departs from the established pattern. Otherwise, prefer the transcript port's pattern: log on failure, do not block the bundle from returning. Either way the *test* for this behavior is missing — only the happy path is covered.

**2. Sanitized-basename labelling has two latent issues for future personal paths.** `personal_context_admitted_summary` (lib.rs:262-295) takes `path.rsplit('/').next()` and then strips to `[A-Za-z0-9._-]`, producing labels like `USER.md` / `assistant-directives.md`. Two follow-ons worth documenting now while the surface is only two paths:

   a. *Basename collisions.* The label set drops directory context, so a future personal identity at `something/assistant-directives.md` would alias the existing `context/assistant-directives.md`. The test asserts the *path* is stripped on purpose, but the audit reader has no way to disambiguate.
   
   b. *Forbidden-token interaction with `LoopSafeSummary`.* `validate_loop_safe_summary` (run_profile/host.rs:126-160) rejects substrings like `secret`, `password`, `bearer ` (with trailing space — triggers if a `bearer` label is followed by another label in the joined source list), `api key`, `tool input`, etc. If a future personal identity file is named, e.g., `secret-prompt.md` or `bearer-rules.md`, the milestone summary build will fail with `Internal`, which under finding 1 means `load_loop_context` itself fails and personal context cannot be admitted at all. Worth a `// dev_assert!` or a doc-comment on `PERSONAL_IDENTITY_PATHS` (reborn_identity_context.rs:30) saying the path basenames must not match `validate_loop_safe_summary`'s forbidden-token list.

**3. Milestone fires per `load_loop_context` invocation.** If the context port is called multiple times per turn (e.g., across iterations), the milestone fires multiple times for the same admitted set. That looks intentional given how the test asserts `recorded.len() == 1` for a single call, but the contract docs (memory.md, turns-agent-loop.md) say "the loop emits safe observability metadata" — they don't disambiguate per-iteration vs per-turn. A one-line note in the contract clarifying which would help downstream audit consumers de-duplicate.

## Smaller items

- `personal_context_policy: PersonalContextPolicy::Excluded` is hard-coded in `interactive_profile()` and `long_running_mission_profile()` (resolver.rs:323, 377) but neither is asserted in `run_profile_contract.rs`. The default-deserialization test catches *old-payload* fallback, but a positive `assert_eq!(snapshot.personal_context_policy, Excluded)` for the explicit named profiles would defend against an accidental future flip.
- The `applies()` change at identity_context.rs:287-289 is correct (policy gate runs first in `build_identity_messages_for_run_detailed`), but consider expressing the invariant by inverting the dependency: have `identity_applicability_allowed_for_run` short-circuit *before* `applies()` is reached. That's what the code already does, so the comment is the only documentation tying them — fine as-is, just worth knowing future readers depend on call-order in the loop body.

## Verification

I read the diff and key files at PR head `313bd369`; I did not re-run the test suite locally. CI status I last saw on the PR was clean for the previously-validated tests cited in Henry's `4477622265` comment.

## Verdict

The substantive correctness/privacy story is in good shape: the cache-path P0 from the prior round is properly closed, the defense-in-depth integration test exists, and an audit signal now fires when personal context is admitted (with the right safety properties on the summary). The remaining gaps are:

- **(Should fix)** Decide whether milestone-publish failure should fail-closed or best-effort, and either document the fail-closed intent or align with the transcript port's logged-debug pattern. Adding a one-line test for that branch closes the matrix.
- **(Should fix)** Add a comment near `PERSONAL_IDENTITY_PATHS` noting the basename constraints from `validate_loop_safe_summary` so future personal-identity additions don't silently break the milestone path.
- **(Nit)** Add positive `assert_eq!(..., Excluded)` cases for the two named non-default profiles in `run_profile_contract.rs`.
- **(Tracking)** Confirm #3745 is filed for the privileged-dimension / shared-thread-routing audit, and add a sentence to the PR description naming the production resolver path that shared threads land on today.

None of these are correctness regressions on the privacy boundary itself, so this is a COMMENT rather than a request-for-changes.

---

henrypark133:
Addressed the review from https://github.com/nearai/ironclaw/pull/3721#pullrequestreview-4312393940 in latest head `46c843915`.

Latest changes:
- Personal-context admission observability now follows the existing best-effort milestone posture: `ThreadBackedLoopContextPort::publish_personal_context_admitted` logs sink failures and still returns the context bundle instead of making milestone publication load-bearing.
- Admission milestone emission is deduped after a successful publish and remains retryable after a transient sink failure.
- Added caller-level coverage for the failure path in `context_port_survives_personal_context_admitted_milestone_sink_failure`, plus the existing safe-summary test still verifies no raw `USER.md` / assistant-directive contents or nested paths are exposed.
- Added the `PERSONAL_IDENTITY_PATHS` comment documenting future basename/safe-summary constraints so new personal identity paths do not collide or trip `LoopSafeSummary` forbidden-token validation.
- Added explicit `PersonalContextPolicy::Excluded` assertion for the long-running named profile in `run_profile_contract`.
- Confirmed #3745 is open and now referenced in the PR body for the owner-aware / privileged-upgrade follow-up.
- Updated the PR description to match the current behavior: builtin and planned defaults are excluded by default; explicitly allowed profiles stay direct-authority only, and shared authority downgrades them to excluded.

Verification run after the rebase:
- `cargo test -p ironclaw_turns -p ironclaw_loop_support`
- `cargo test -p ironclaw_reborn planned_driver_factory --lib`
- `cargo test -p ironclaw_loop_support --test thread_loop_support_contract personal_context`
- `cargo test -p ironclaw_turns --test run_profile_contract`
- `git diff --check`

crates/ironclaw_reborn/src/planned_driver_factory.rs
serrrfirat: [High] `planned_default_profile_definition()` marks the planned default as `PersonalContextPolicy::Allowed`, and `default_planned_run_profile_resolver()` also uses this profile as the implicit default. That means normal no-profile Reborn turns can admit `USER.md` and `context/assistant-directives.md`, which contradicts the PR's stated constraint that planned defaults remain excluded by default. Keep the planned default excluded and add a separate opt-in profile/path for personal context admission after verified direct authority.

Also flagged by: Security, Bugs.

crates/ironclaw_loop_support/src/lib.rs
```diff
@@ -226,6 +253,70 @@ where
...
+        let Some(milestone_sink) = self.milestone_sink.as_ref() else {
+            return;
+        };
+        let emitted_cell = self
+            .identity_candidates
+            .personal_context_admitted_cell_for_mode(mode);
+        let publish_result = emitted_cell
+            .get_or_try_init(|| async {
+                let summary = personal_context_admitted_summary(admitted_paths)?;
+                LoopHostMilestoneEmitter::new(self.run_context.clone(), Arc::clone(milestone_sink))
```
serrrfirat: [Medium] The best-effort admission milestone is awaited inside the `OnceCell` initializer on the context-load path. In Reborn the milestone sink can perform durable event-log I/O, so a slow or hung publish stalls prompt/context construction, and concurrent callers for the same mode wait behind the same initializer. Publish this off the critical path or bound it with a timeout, while still recording the dedupe cell only after successful publish.

crates/ironclaw_loop_support/src/identity_context.rs
```diff
@@ -98,6 +100,12 @@ pub struct HostIdentityMessageContent {
     pub content: String,
 }
 
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct IdentityMessageBuildOutcome {
+    pub messages: Vec<LoopContextMessage>,
+    pub admitted_personal_context_paths: Vec<String>,
```
serrrfirat: [Medium] `IdentityMessageBuildOutcome` carries admitted personal context paths as raw `String`s even though this module already has the `IdentityFileName` domain type. `.claude/rules/types.md` says internal domain values should use specialized types rather than raw strings. Store `Vec<IdentityFileName>` here and let milestone formatting read `as_str()` at the boundary.

crates/ironclaw_loop_support/src/identity_context.rs
```diff
@@ -208,13 +247,19 @@ pub fn build_identity_messages_from_candidates(
             break;
         }
         used = used.saturating_add(cost);
+        if candidate.applies_when == IdentityApplicability::OnPersonalContextAllowed {
+            admitted_personal_context_paths.push(candidate.name.as_str().to_string());
```
serrrfirat: [Medium] The admission outcome records personal paths only after the budget check, which is the right shape, but there is no test proving a budget-skipped personal file is omitted from the milestone count/labels. Add `context_port_milestone_counts_only_budget_admitted_personal_identity` covering a personal candidate skipped by `IdentityBudget`.

src/workspace/reborn_identity_context.rs
```diff
@@ -94,27 +122,62 @@ impl WorkspaceIdentityContextSource {
...
```
serrrfirat, line 138: [Medium] Allowed personal context now adds `USER.md` and assistant directives, but the adjacent tests do not write `HEARTBEAT.md` under an allowed profile to prove it still remains excluded from this new personal-context path. Add `workspace_identity_context_allowed_policy_still_excludes_heartbeat` covering that case.

crates/ironclaw_loop_support/src/lib.rs
```diff
@@ -226,6 +253,70 @@ where
...
+                    .driver_note(LoopDriverNoteKind::Context, summary)
+                    .await?;
+                Ok::<(), AgentLoopHostError>(())
+            })
+            .await;
+        if let Err(error) = publish_result {
+            tracing::warn!("failed to emit personal context admitted milestone: {error}");
```
serrrfirat: [Low] This failure is intentionally non-fatal, but it logs at `warn!`. Root `CLAUDE.md` says `warn!` appears in the REPL/TUI and internal diagnostics should use `debug!`; switch this to `tracing::debug!` unless the message is intentionally user-visible.

.gitignore
```diff
@@ -63,3 +63,6 @@ tests/fixtures/llm_traces/live/github_dev_workflow_full_loop.log
 # tests locally. Only the .json fixtures for each scenario are checked
 # in; the .log files are local debugging artifacts.
 tests/fixtures/llm_traces/live/*.log
+
+# Multi-agent code review findings (local only, never commit)
+.review/
```
serrrfirat: [Low] This `.review/` ignore entry is unrelated to the personal-context policy change. Root `AGENTS.md` asks PRs to avoid unrelated file churn; please drop this from this PR or move it to a separate hygiene change.

crates/ironclaw_loop_support/src/lib.rs
```diff
@@ -226,6 +253,70 @@ where
...
+        if admitted_paths.is_empty() {
+            return;
+        }
+        let Some(milestone_sink) = self.milestone_sink.as_ref() else {
+            return;
+        };
+        let emitted_cell = self
+            .identity_candidates
+            .personal_context_admitted_cell_for_mode(mode);
+        let publish_result = emitted_cell
```
serrrfirat: [Low] The `OnceCell` dedupe for the new admission milestone has sequential coverage, but no contention test. Add `context_port_dedupes_personal_context_admitted_milestone_under_concurrent_loads` covering two concurrent `load_loop_context` calls that should emit one admission milestone.

serrrfirat (changes requested):
Multi-agent code review for `nearai/ironclaw#3721` at `46c843915e6021b963a7fecbd5a48b08f09a1798`.

Result: **request changes**.

Stats: 5 reviewer lenses completed. Aggregated 9 raw findings into 8 deduped findings after confidence filtering; the default-policy issue was independently flagged by Security and Bugs.

Key blocker:
- High: default planned profile still resolves to `PersonalContextPolicy::Allowed`, which contradicts this PR's stated constraint that planned defaults remain excluded by default and creates a personal-context leakage path for implicit default runs.

Additional findings are inline: milestone publication can block context loading, two objective missing edge tests, a domain-typing rule violation, logging-level noise, and unrelated `.gitignore` churn.

henrypark133:
Addressed the latest review round in `51eac3c7d`.

What changed:
- Removed the implicit `PersonalContextPolicy::Allowed` from `planned_default_profile_definition`; planned defaults now resolve to `Excluded`, while the explicit direct-authority allowed test path still covers opt-in behavior and shared authority still downgrades to `Excluded`.
- Moved the `PersonalContextAdmitted` milestone publish off the `load_loop_context` critical path. The publish now runs in a background task, records the dedupe cell only after successful publish, retries after sink failure, and logs non-fatal failures at `debug`.
- Changed admitted personal context tracking from raw `String` paths to `IdentityFileName`, with formatting pushed to the milestone boundary.
- Added regression coverage for budget-skipped personal files not appearing in milestone counts/labels, concurrent context loads deduping to one milestone, sink failure resilience/retry, and `HEARTBEAT.md` remaining excluded even when personal context is allowed.
- Removed the unrelated `.review/` ignore entry from `.gitignore`.

Verification run locally:
- `cargo fmt --check`
- `git diff --cached --check`
- `cargo test -p ironclaw_reborn planned_driver_factory --lib`
- `cargo test -p ironclaw_loop_support --test thread_loop_support_contract personal_context`
- `cargo test -p ironclaw_loop_support --test thread_loop_support_contract context_port_milestone_counts_only_budget_admitted_personal_identity`
- `cargo test --lib --features libsql workspace_identity_context_allowed_policy_still_excludes_heartbeat`

CI has started on the pushed commit and is currently pending.

crates/ironclaw_loop_support/src/lib.rs
```diff
@@ -92,18 +97,27 @@ where
     identity_context_source: Option<Arc<dyn HostIdentityContextSource>>,
     identity_budget: IdentityBudget,
     identity_candidates: Arc<IdentityCandidateCache>,
+    milestone_sink: Option<Arc<dyn LoopHostMilestoneSink>>,
 }
 
 struct IdentityCandidateCache {
     text_only: OnceCell<Vec<HostIdentityContextCandidate>>,
     codeact: OnceCell<Vec<HostIdentityContextCandidate>>,
+    text_only_personal_context_admitted: OnceCell<()>,
```
henrypark133, line 106: **Medium** — Personal context milestone path labels may leak directory structure.

personal_context_admitted_summary uses rsplit('/') to extract basenames from IdentityFileName values. The basename extraction assumes paths use forward slashes. If any future personal identity path uses a different separator or contains encoded slashes, the full path could leak into the milestone summary.

**Fix:** Use a dedicated path-basename utility instead of rsplit('/').next(). Add a test that verifies milestone summaries never contain path separators.

crates/ironclaw_loop_support/src/lib.rs
```diff
@@ -228,6 +268,98 @@ where
     }
 }
 
+impl<S> ThreadBackedLoopContextPort<S>
+where
+    S: SessionThreadService + ?Sized + Send + Sync,
+{
+    fn publish_personal_context_admitted(
```
henrypark133, line 275: **High** — Dedup guard race allows duplicate milestone emission after sink failure.

publish_personal_context_admitted uses a OnceCell and AtomicBool as a two-part dedup guard, but the check-then-act on these two primitives is not atomic. When the async publish fails, in_flight is reset to false while the OnceCell remains unset. A concurrent caller can slip through between the failure reset and a subsequent successful set, emitting a duplicate milestone.

**Fix:** Replace the two-primitive guard with a single AtomicU8 state machine (IDLE=0, IN_FLIGHT=1, EMITTED=2) using compare_exchange for atomic transitions.

crates/ironclaw_loop_support/src/lib.rs
```diff
@@ -228,6 +268,98 @@ where
...
+        if emitted_cell.get().is_some() {
+            return;
+        }
+        let in_flight = self
+            .identity_candidates
+            .personal_context_admitted_in_flight_for_mode(mode);
+        if in_flight.swap(true, Ordering::AcqRel) {
+            return;
+        }
+        let summary = match personal_context_admitted_summary(admitted_paths) {
```
henrypark133, line 298: **Medium** — personal_context_admitted_summary has no unit tests for edge cases.

The private function personal_context_admitted_summary builds safe milestone summaries by filtering path characters. No tests cover: empty admitted_paths slice, paths where all characters are stripped by the filter, or paths that would produce an empty label after basename extraction.

**Fix:** tests::lib::personal_context_admitted_summary_empty_paths_covering_empty_slice

Also flagged by: tests/Medium

crates/ironclaw_loop_support/src/identity_context.rs
```diff
@@ -189,17 +199,46 @@ pub async fn build_identity_messages(
         .load_identity_candidates(run_context, mode)
         .await
         .map_err(HostIdentityContextBuildError::into_host_error)?;
-    build_identity_messages_from_candidates(&candidates, mode, budget)
+    build_identity_messages_for_run(&candidates, run_context, mode, budget)
 }
 
-pub fn build_identity_messages_from_candidates(
+pub fn identity_applicability_allowed_for_run(
```
henrypark133, line 205: **Medium** — identity_applicability_allowed_for_run has no direct unit test.

The public function identity_applicability_allowed_for_run gates OnPersonalContextAllowed candidates based on the resolved policy. While integration tests cover the end-to-end flow, no unit test directly exercises all three applicability variants against both Excluded and Allowed policies.

**Fix:** tests::identity_context::identity_applicability_allowed_for_run_all_variants_covering_excluded_and_allowed

crates/ironclaw_loop_support/src/identity_context.rs
```diff
@@ -189,17 +199,46 @@ pub async fn build_identity_messages(
...
+    }
+}
+
+pub fn build_identity_messages_for_run(
     candidates: &[HostIdentityContextCandidate],
+    run_context: &LoopRunContext,
     mode: PromptMode,
     budget: IdentityBudget,
 ) -> Result<Vec<LoopContextMessage>, AgentLoopHostError> {
+    Ok(build_identity_messages_for_run_detailed(candidates, run_context, mode, budget)?.messages)
```
henrypark133, line 226: **Medium** — build_identity_messages_for_run_detailed untested with mixed candidates under excluded policy.

No test verifies that when both Always and OnPersonalContextAllowed candidates are present and the policy is Excluded, the Always candidates are included while OnPersonalContextAllowed ones are filtered out.

**Fix:** tests::identity_context::build_identity_messages_for_run_detailed_mixed_candidates_excluded_covering_filtering

crates/ironclaw_loop_support/tests/thread_loop_support_contract.rs
```diff
@@ -345,6 +346,305 @@ async fn context_port_caches_identity_candidates_per_prompt_mode() {
     assert_eq!(source.load_calls(), 2);
 }
 
+#[tokio::test]
```
henrypark133, line 349: **Medium** — No integration test for personal context admission in CodeAct mode.

All personal context integration tests use PromptMode::TextOnly. The applies() function returns true for OnPersonalContextAllowed in both TextOnly and CodeAct modes, but no test verifies personal identity files are admitted in CodeAct mode.

**Fix:** tests::thread_loop_support_contract::context_port_includes_personal_identity_in_codeact_mode_covering_codeact_prompt_mode

crates/ironclaw_loop_support/src/lib.rs
```diff
@@ -206,11 +240,17 @@ where
...
                     .await?;
-                identity_context::build_identity_messages_from_candidates(
+                let outcome = identity_context::build_identity_messages_for_run_detailed(
                     candidates,
+                    &self.run_context,
                     mode,
                     self.identity_budget,
-                )?
+                )?;
+                self.publish_personal_context_admitted(
```
henrypark133, line 249: **Medium** — publish_personal_context_admitted untested when milestone summary construction fails.

The publish_personal_context_admitted method handles the case where personal_context_admitted_summary returns an error by logging and resetting the in-flight flag. No test exercises this error path.

**Fix:** tests::thread_loop_support_contract::context_port_resets_in_flight_flag_when_summary_build_fails_covering_summary_error_recovery

henrypark133 (commented):
## Code Review (multi-agent)

**Intent:** Gate personal context files by run profile policy, adding backward-compatible exclusion defaults and safe milestone emission for admitted context sources.

**Stats:** 9 findings (from 12 raw, 2 after same-line dedup) across 5 files. Reviewers run: security, bugs, performance, tests, conventions, design. Reviewers failed: conventions (empty response). Body-only: 2

### Security

1. **Medium** personal_context_authority defaults to Direct in all production callers (`crates/ironclaw_turns/src/run_profile/resolver.rs:30-38`, confidence 75) — anchor: crates/ironclaw_turns/src/run_profile/resolver.rs:38 _(no diff position — body only)_
   
   RunProfileResolutionRequest::interactive_default() sets personal_context_authority to Direct by default. All production callers use ..interactive_default() without overriding it. The authority downgra...
2. **Medium** Personal context milestone path labels may leak directory structure (`crates/ironclaw_loop_support/src/lib.rs:268-330`, confidence 75) — anchor: crates/ironclaw_loop_support/src/lib.rs:300
   
   personal_context_admitted_summary uses rsplit('/') to extract basenames from IdentityFileName values. The basename extraction assumes paths use forward slashes. If any future personal identity path us...


### Bugs

1. **High** Dedup guard race allows duplicate milestone emission after sink failure (`crates/ironclaw_loop_support/src/lib.rs:275-324`, confidence 100) — anchor: crates/ironclaw_loop_support/src/lib.rs:289
   
   publish_personal_context_admitted uses a OnceCell and AtomicBool as a two-part dedup guard, but the check-then-act on these two primitives is not atomic. When the async publish fails, in_flight is res...


### Performance

No findings.


### Tests

1. **Medium** render_identity_content has no unit tests (`src/workspace/reborn_identity_context.rs:174-182`, confidence 75) — anchor: src/workspace/reborn_identity_context.rs:174 _(no diff position — body only)_
   
   The private function render_identity_content wraps personal identity paths with a safety header but returns content unchanged for non-personal paths. No tests verify either branch.
   
   Also flagged by: tests/Medium, security/Low
2. **Medium** personal_context_admitted_summary has no unit tests for edge cases (`crates/ironclaw_loop_support/src/lib.rs:327-361`, confidence 75) — anchor: crates/ironclaw_loop_support/src/lib.rs:327
   
   The private function personal_context_admitted_summary builds safe milestone summaries by filtering path characters. No tests cover: empty admitted_paths slice, paths where all characters are stripped...
   
   Also flagged by: tests/Medium
3. **Medium** identity_applicability_allowed_for_run has no direct unit test (`crates/ironclaw_loop_support/src/identity_context.rs:205-218`, confidence 75) — anchor: crates/ironclaw_loop_support/src/identity_context.rs:205
   
   The public function identity_applicability_allowed_for_run gates OnPersonalContextAllowed candidates based on the resolved policy. While integration tests cover the end-to-end flow, no unit test direc...
4. **Medium** build_identity_messages_for_run_detailed untested with mixed candidates under excluded policy (`crates/ironclaw_loop_support/src/identity_context.rs:229-265`, confidence 75) — anchor: crates/ironclaw_loop_support/src/identity_context.rs:229
   
   No test verifies that when both Always and OnPersonalContextAllowed candidates are present and the policy is Excluded, the Always candidates are included while OnPersonalContextAllowed ones are filter...
5. **Medium** No integration test for personal context admission in CodeAct mode (`crates/ironclaw_loop_support/tests/thread_loop_support_contract.rs:370-430`, confidence 75) — anchor: crates/ironclaw_loop_support/tests/thread_loop_support_contract.rs:370
   
   All personal context integration tests use PromptMode::TextOnly. The applies() function returns true for OnPersonalContextAllowed in both TextOnly and CodeAct modes, but no test verifies personal iden...
6. **Medium** publish_personal_context_admitted untested when milestone summary construction fails (`crates/ironclaw_loop_support/src/lib.rs:268-323`, confidence 75) — anchor: crates/ironclaw_loop_support/src/lib.rs:298
   
   The publish_personal_context_admitted method handles the case where personal_context_admitted_summary returns an error by logging and resetting the in-flight flag. No test exercises this error path.


### Conventions

No findings.


### Design

No findings.

zmanian (commented):
## Re-review of 3 new commits

Walked `b6465e4e`, `42940781`, `0072dab1` against my prior findings.

### F2 — observability + privileged dimension: RESOLVED

- **Observability**: `publish_personal_context_admitted` now emits a `LoopDriverNoteKind::Context` milestone with the admitted basenames, deduped per `PromptMode` via `OnceCell` + an `AtomicBool` in-flight latch (`crates/ironclaw_loop_support/src/lib.rs`). Failures release the latch so a later prepare can retry — good. Test coverage in `thread_loop_support_contract.rs` exercises both the emit-once invariant and the empty-admission no-op path.
- **Privileged dimension**: new `PersonalContextAuthority::{Direct, Shared}` enum in `crates/ironclaw_turns/src/run_profile/policy.rs`, threaded through `RunProfileResolutionRequest` and consumed in `RunProfileDefinition::resolve`. When the request is `Shared` (the `Default::default()` value), an `Allowed` profile is downgraded to `Excluded` *and* the resolution fingerprint changes. `interactive_default()` opts into `Direct`, so the REPL/binary path retains personal context while every other construction site (workspace/background/heartbeat callers using `Default`) fails closed. Three new resolver tests pin the downgrade, the direct preservation, and the fingerprint divergence. This is exactly the privilege axis I was asking for.

### P0 (shared cache cross-contamination): RESOLVED

`WorkspaceIdentityContextSource::resolve_identity_message_content` now re-checks `identity_applicability_allowed_for_run` against the *current* `run_context` before returning the cached entry and surfaces `HostIdentityContextBuildError::PolicyDenied` on mismatch (`src/workspace/reborn_identity_context.rs:153`). Regression test `workspace_identity_context_denies_cached_personal_ref_when_policy_excludes` writes USER.md, primes the cache with an `Allowed` context, then asserts a default (`Excluded`) context returns `PolicyDenied` on the same ref — the precise scenario I described.

### F1 (call-site audit): RESOLVED

The `Default` → `Shared` mapping combined with `interactive_default()` → `Direct` is a clean fail-closed default. No production caller currently opts into `Direct` outside `interactive_default()`, which is fine — but worth noting that any future caller adding a non-interactive entrypoint that legitimately wants personal context will need to explicitly call `with_personal_context_authority(Direct)`. Consider an `AGENTS.md`/contract note in `crates/ironclaw_turns/` so this isn't lost.

### New finding (NIT, non-blocking): milestone is now fire-and-forget

The follow-up commit converted `publish_personal_context_admitted` from `await`ed to `tokio::spawn`. The admission milestone is no longer ordered before the assistant consumes the personal-context messages — emit can land after model output that referenced the data. For a security-relevant audit event this is a minor degradation. Not a blocker (the OnceCell guard still ensures eventual emission), but I'd file an issue: either re-await the milestone, or document that the personal-context admission event is best-effort/eventually-consistent in the contracts doc you updated (`docs/reborn/contracts/turns-agent-loop.md`).

### Verdict

LGTM modulo the nit. All three of my prior blocking findings — including the P0 cache cross-contamination — are properly addressed with regression coverage. Ship it.

— Zaki

